PR #7946, opened 2024-09-05
Fix conversion in ssl* files and in internal.c

gasbytes:
Fix issues related to -Wconversion and -Wsign-conversion in the ssl* files and in internal.c.

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -7817,7 +7817,7 @@ void FreeKey(WOLFSSL* ssl, int type, void** pKey)
 int AllocKey(WOLFSSL* ssl, int type, void** pKey)
 {
     int ret = WC_NO_ERR_TRACE(BAD_FUNC_ARG);
-    int sz = 0;
+    size_t sz = 0;
 #ifdef HAVE_ECC
     ecc_key* eccKey;
 #endif /* HAVE_ECC */
@@ -7894,7 +7894,7 @@ int AllocKey(WOLFSSL* ssl, int type, void** pKey)
     }
 
     /* Allocate memory for key */
-    *pKey = (void *)XMALLOC((size_t)sz, ssl->heap, type);
+    *pKey = (void *)XMALLOC(sz, ssl->heap, type);
     if (*pKey == NULL) {
         return MEMORY_E;
     }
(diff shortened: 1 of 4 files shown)


src/internal.c
```diff
@@ -7894,7 +7894,7 @@ int AllocKey(WOLFSSL* ssl, int type, void** pKey)
     }
 
     /* Allocate memory for key */
-    *pKey = (void *)XMALLOC(sz, ssl->heap, type);
+    *pKey = (void *)XMALLOC((size_t)sz, ssl->heap, type);
```
bandi13, line 7897: Better to change the type of the variable declaration.

src/internal.c
```diff
@@ -10697,8 +10697,8 @@ int SendBuffered(WOLFSSL* ssl)
             return SEND_OOB_READ_E;
         }
 
-        ssl->buffers.outputBuffer.idx += sent;
-        ssl->buffers.outputBuffer.length -= sent;
+        ssl->buffers.outputBuffer.idx += (word32)sent;
```
bandi13, line 10700: Change variable type instead

src/internal.c
```diff
@@ -10821,7 +10821,7 @@ int GrowInputBuffer(WOLFSSL* ssl, int size, int usedLength)
         return BAD_FUNC_ARG;
     }
 
-    tmp = (byte*)XMALLOC(size + usedLength + align,
+    tmp = (byte*)XMALLOC((size_t)(size + usedLength + align),
```
bandi13, line 10824: I wonder if it is possible to change the function input variable types

src/internal.c
```diff
@@ -20727,7 +20727,7 @@ static int DoAlert(WOLFSSL* ssl, byte* input, word32* inOutIdx, int* type)
             if (CipherHasExpIV(ssl))
                 ivExtra = AESGCM_EXP_IV_SZ;
         }
-        dataSz -= ivExtra;
+        dataSz -= (word32)ivExtra;
```
bandi13, line 20730: Change variable type instead

src/internal.c
```diff
@@ -20454,7 +20454,7 @@ int DoApplicationData(WOLFSSL* ssl, byte* input, word32* inOutIdx, int sniff)
             if (dataSz < 0) return dataSz;
         }
 #endif
-        idx += rawSz;
+        idx += (word32)rawSz;
```
bandi13, line 20457: Change variable type instead

bandi13 (changes requested):
Please go through and check if the variable declaration type should be changed instead of being cast.

bandi13 (approved):

douzzer (approved):

PR #7959, opened 2024-09-09
Update to use C99 instead of C11 because of lack of SGX support

billphipps:
# Description

After adding C11 memory fence support, the SGX nightly test failed with a missing stdatomic.h file.  Our IDE/LINUX-SGX makefile erroneously forced C11 standard compliance, but SGX is only compatible with C99 and C++11.  This change corrects the C standard to be C99 to match the libc that is embedded within the SGX include libraries.

Fixes job nightly-SGX_Linux-v2. 

Note: I'm apply the same correction to wolfssl-examples SGX_LINUX.

# Testing

Tested using boz-sgx, which is Ubuntu 18 with SGX SDK and matched build commands with script.  

# Checklist

 - [ X ] added tests -CORRECTED tests

### DIFF
--- a/IDE/LINUX-SGX/sgx_t_static.mk
+++ b/IDE/LINUX-SGX/sgx_t_static.mk
@@ -115,7 +115,7 @@ ifeq ($(HAVE_WOLFSSL_SP), 1)
 endif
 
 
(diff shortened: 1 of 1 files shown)


billphipps:
retest this please

douzzer (approved):

PR #7964, opened 2024-09-11
Avoid checks if key is not populated.

anhu:


anhu:
Not the right solution. Closing.

PR #7965, opened 2024-09-11
Use context specific for PKCS7 encrypted data encoding

ColtonWilley:
# Description
While the current encoding for PKCS8 PBE is within specification, openssl will only parse the indefinite length form which the old ASN code used to generate before the move to the ASN template. Modify ASN template for PKCS8 PBE encoding to use indefinite length, making it consistent with both old ASN code and openssl expectations. 

Fixes zd 18515

### DIFF
--- a/wolfcrypt/src/asn.c
+++ b/wolfcrypt/src/asn.c
@@ -9156,7 +9156,7 @@ static const ASNItem p8EncPbes1ASN[] = {
 /* ENCALGO_PBEPARAM_SALT */             { 3, ASN_OCTET_STRING, 0, 0, 0 },
                         /* Iteration Count */
 /* ENCALGO_PBEPARAM_ITER */             { 3, ASN_INTEGER, 0, 0, 0 },
-/* ENCDATA               */     { 1, ASN_INDEF_LENGTH, 0, 0, 0 },
+/* ENCDATA               */     { 1, (ASN_CONTEXT_SPECIFIC | 0), 0, 0, 0 },
 };
 enum {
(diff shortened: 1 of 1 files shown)


wolfcrypt/src/asn.c
```diff
@@ -9156,7 +9156,7 @@ static const ASNItem p8EncPbes1ASN[] = {
 /* ENCALGO_PBEPARAM_SALT */             { 3, ASN_OCTET_STRING, 0, 0, 0 },
                         /* Iteration Count */
 /* ENCALGO_PBEPARAM_ITER */             { 3, ASN_INTEGER, 0, 0, 0 },
-/* ENCDATA               */     { 1, ASN_OCTET_STRING, 0, 0, 0 },
```
SparkiDev, line 9159: Do you mean CONTEXT_SPECIFIC?
PKCS#8 RFC 5958, 3 - EncryptedPrivateKeyInfo says it will be an OCTECT_STRING.

Why is OpenSSL different?

SparkiDev (changes requested):

wolfcrypt/src/asn.c
ColtonWilley: Yes I believe you are right, CONTEXT_SPECIFIC is actually what I need here.

I completely agree that our encoding is exactly to specification. I do not know why OpenSSL does things differently, but its also how the original code before WOLFSSL_ASN_TEMPLATE worked as well.

SparkiDev (approved):

PR #7980, opened 2024-09-17
FP SmallStack Fix

ejohnstown:
# Description

The function _fp_exptmod_nct() is using WOLFSSL_NO_MALLOC to guard using stack allocation vs malloc. It's twin function _fp_exptmod_ct() is using WOLFSSL_SMALL_STACK for this. This is causing inappropriate use of malloc() in a small stack environment. The no-malloc case will also be kept so static memory and no-malloc fix still works.

1. Change the guards for `#ifndef WOLFSSL_NO_MALLOC` in the function `_fp_exptmod_nct()` to `#if defined(WOLFSSL_SMALL_STACK) && !defined(WOLFSSL_NO_MALLOC)`.

# Testing

Verified the code path is correctly set. Ran the wolfcrypttest in the debugger to confirm allocation and no allocation in the function `_fp_exptmod_nct()`. Ran the wolfcrypttest in valgrind.

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/wolfcrypt/src/tfm.c
+++ b/wolfcrypt/src/tfm.c
@@ -2430,7 +2430,7 @@ static int _fp_exptmod_nct(fp_int * G, fp_int * X, fp_int * P, fp_int * Y)
   fp_int  *res;
   fp_digit buf, mp;
   int      err, bitbuf, bitcpy, bitcnt, mode, digidx, x, y, winsize;
-#ifndef WOLFSSL_NO_MALLOC
+#if defined(WOLFSSL_SMALL_STACK) && !defined(WOLFSSL_NO_MALLOC)
   fp_int  *M;
 #else
   fp_int   M[(1 << 6) + 1];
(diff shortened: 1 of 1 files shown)


wolfcrypt/src/tfm.c
```diff
@@ -2430,7 +2430,7 @@ static int _fp_exptmod_nct(fp_int * G, fp_int * X, fp_int * P, fp_int * Y)
   fp_int  *res;
   fp_digit buf, mp;
   int      err, bitbuf, bitcpy, bitcnt, mode, digidx, x, y, winsize;
-#ifndef WOLFSSL_NO_MALLOC
+#ifdef WOLFSSL_SMALL_STACK
```
dgarske, line 2433: This change makes sense. Can you please use `#if defined(WOLFSSL_SMALL_STACK) && !defined(WOLFSSL_NO_MALLOC)`. It ensures both cases are covered, even though small stack and no malloc are not compatible.

dgarske (changes requested):

dgarske (approved):

ejohnstown:
retest this please

SparkiDev (approved):

douzzer (approved):

PR #7989, opened 2024-09-18
Update to use memory_order_seq_cst instead of GCC internal for C11

billphipps:
# Description

Update the memory order constant from the GCC-specific to the C11 standard enumeration identifier.

Fixes zd18640

# Testing

Tested on MacOS (clang) and IAR 8.6.0 compiler.  I suspect we may have additional compilers that may need different workarounds.

# Checklist

 - [ N/A ] added tests


bandi13:
retest this please

bandi13 (approved):

PR #8002, opened 2024-09-21
make `wolfSSL_UseALPN` take `const char*` rather than `char*`

ycdev1:
Currently `wolfSSL_UseALPN` takes `char*` argument which suggests that it mutates the data it receives while it just copies the data to a temporary allocated buffer so it should take `const char*` instead.

Also `wolfSSL_ALPN_GetProtocol` should output `const char*` but this may break existing code.


ycdev1:
Does wolfssl currently have an api to clear the set alpn? If no, the `wolfSSL_UseALPN` can clear the alpn if passed `nullptr` and 0

embhorn:
Hi @ycdev1 

Thanks for this suggestion. Could you tell us a bit about your project using wolfSSL? We do not typically accept small changes to the library. Do you think you will propose more changes in the future?

Thanks,
Eric -wolfSSL Support

PR #8005, opened 2024-09-23
New option to always copy over key to SSL object

ColtonWilley:
# Description

Initial implementation of new option to always copy over key to SSL ctx

Fixes zd 18662

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -6829,7 +6829,14 @@ int SetSSL_CTX(WOLFSSL* ssl, WOLFSSL_CTX* ctx, int writeDup)
     ssl->buffers.certChainCnt = ctx->certChainCnt;
 #endif
 #ifndef WOLFSSL_BLIND_PRIVATE_KEY
+#ifdef WOLFSSL_COPY_KEY
+    AllocCopyDer(&ssl->buffers.key, ctx->privateKey->buffer,
+        ctx->privateKey->length, ctx->privateKey->type,
+        ctx->privateKey->heap);
+    ssl->buffers.weOwnKey = 1;
+#else
     ssl->buffers.key      = ctx->privateKey;
+#endif
 #else
     if (ctx->privateKey != NULL) {
         AllocCopyDer(&ssl->buffers.key, ctx->privateKey->buffer,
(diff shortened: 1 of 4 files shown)


wolfssl/wolfcrypt/settings.h
```diff
@@ -3581,6 +3581,11 @@ extern void uITRON4_free(void *p) ;
     #define WOLFSSL_COPY_CERT
 #endif
 
+#if defined(OPENSSL_ALL) && !defined(WOLFSSL_NO_COPY_KEY)
+    #undef WOLFSSL_COPY_KEY
```
dgarske, line 3585: Please add a code comment for purpose of `WOLFSSL_COPY_KEY` and use-case it resolves.

dgarske (changes requested):

wolfssl/wolfcrypt/settings.h
ColtonWilley: Comments have been added.

dgarske (approved):

src/internal.c
```diff
@@ -6829,7 +6829,22 @@ int SetSSL_CTX(WOLFSSL* ssl, WOLFSSL_CTX* ctx, int writeDup)
     ssl->buffers.certChainCnt = ctx->certChainCnt;
 #endif
 #ifndef WOLFSSL_BLIND_PRIVATE_KEY
+#ifdef WOLFSSL_COPY_KEY
+    if (ctx->privateKey != NULL) {
+        if (ssl->buffers.key != NULL) {
+            FreeDer(&ssl->buffers.key);
```
JacobBarthelmeh, line 6835: I think if free'ing a previous key here we need a check for ssl->buffers.weOwnKey and follow the same pattern as wolfSSL_UnloadCertsKeys().

JacobBarthelmeh (changes requested):

src/internal.c
ColtonWilley: I actually disagree. Unlike wolfSSL_UnloadCertsKeys() we are with certainty going to set the ssl->buffers.key. Checking on weOwnCert as well just introduces the chance that through mismanagement if weOwnCert is incorrectly not set at the time of this check and we leak memory. This flag is also partially to ensure behavior replication for openssl compat. Regardless of the state of weOwnCert, if this flag is enabled we should always be copying the key material in my opinion as that is what openssl does.

JacobBarthelmeh: Biggest concern was a case state where the pointer is pointing to the key from the WOLFSSL_CTX. I don't think that state could be hit here though when looking more. We do not check before assigning for the blinding case.

JacobBarthelmeh (approved):

PR #8007, opened 2024-09-23
Update to separate CMAC and AES conditional compiles.  Correct update.

billphipps:
# Description

This provides a few cryptocb and CMAC updates:
1. Correct cryptocb debug prints to correctly handle when NO_AES is defined
2. Correct cryptocb CMAC update to NOT set devCtx to NULL
3. Update CMAC to correctly support NO_AES and allow alternate crypto types
4. Update internal usage to ensure NO_AES and WOLFSSL_CMAC are independent

Note that for proper functionality with AES CMAC, the defines WOLFSSL_CMAC, WOLFSSL_AES_DIRECT must be set, while NO_AES cannot be set.  This matches current functionality.

# Testing
1. Numerous configurations with and without NO_AES using wolfHSM
2. testwolfcrypt with --enable-all
3. commit-tests.sh

# Checklist

 - [ N/A ] added tests

### DIFF
--- a/wolfcrypt/src/cmac.c
+++ b/wolfcrypt/src/cmac.c
@@ -178,7 +178,6 @@ int wc_InitCmac_ex(Cmac* cmac, const byte* key, word32 keySz,
         break;
 #endif /* !NO_AES && WOLFSSL_AES_DIRECT */
     default:
-
         return BAD_FUNC_ARG;
     }
 
@@ -249,7 +248,7 @@ int wc_CmacUpdate(Cmac* cmac, const byte* in, word32 inSz)
         }
     }; break;
 #endif /* !NO_AES && WOLFSSL_AES_DIRECT */
-    default :
+    default:
         ret = BAD_FUNC_ARG;
     }
     return ret;
@@ -271,7 +270,7 @@ int wc_CmacFree(Cmac* cmac)
         wc_AesFree(&cmac->aes);
         break;
 #endif /* !NO_AES && WOLFSSL_AES_DIRECT */
-    default :
+    default:
         /* Nothing to do */
         (void)cmac;
     }
@@ -344,7 +343,7 @@ int wc_CmacFinalNoFree(Cmac* cmac, byte* out, word32* outSz)
             }
         }; break;
     #endif /* !NO_AES && WOLFSSL_AES_DIRECT */
-        default :
+        default:
             ret = BAD_FUNC_ARG;
         }
     }
--- a/wolfcrypt/src/cmac.c
+++ b/wolfcrypt/src/cmac.c
@@ -297,9 +297,11 @@ int wc_CmacFinalNoFree(Cmac* cmac, byte* out, word32* outSz)
         ret = wc_CryptoCb_Cmac(cmac, NULL, 0, NULL, 0, out, outSz, cmac->type, NULL);
         if (ret != WC_NO_ERR_TRACE(CRYPTOCB_UNAVAILABLE))
             return ret;
+        /* Clear CRYPTOCB_UNAVAILABLE return code */
+       ret = 0;
+
         /* fall-through when unavailable */
     }
-    ret = 0;
 #endif
     switch (cmac->type) {
 #if !defined(NO_AES) && defined(WOLFSSL_AES_DIRECT)
@@ -343,7 +345,8 @@ int wc_CmacFinalNoFree(Cmac* cmac, byte* out, word32* outSz)
     return ret;
 }
 
-int wc_CmacFinal(Cmac* cmac, byte* out, word32* outSz) {
+int wc_CmacFinal(Cmac* cmac, byte* out, word32* outSz)
+{
     int ret = 0;
 
     if (cmac == NULL)
--- a/wolfcrypt/src/cmac.c
+++ b/wolfcrypt/src/cmac.c
@@ -299,53 +299,55 @@ int wc_CmacFinalNoFree(Cmac* cmac, byte* out, word32* outSz)
                 NULL);
         if (ret != WC_NO_ERR_TRACE(CRYPTOCB_UNAVAILABLE))
             return ret;
+
         /* Clear CRYPTOCB_UNAVAILABLE return code */
        ret = 0;
 
         /* fall-through when unavailable */
     }
 #endif
-    switch (cmac->type) {
-#if !defined(NO_AES) && defined(WOLFSSL_AES_DIRECT)
-    case WC_CMAC_AES:
-    {
-        const byte* subKey;
-        word32 remainder;
+    if (ret == 0) {
+        switch (cmac->type) {
+    #if !defined(NO_AES) && defined(WOLFSSL_AES_DIRECT)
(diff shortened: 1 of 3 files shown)


wolfssl/wolfcrypt/cmac.h
```diff
@@ -24,9 +24,10 @@
 #define WOLF_CRYPT_CMAC_H
 
 #include <wolfssl/wolfcrypt/types.h>
-#include <wolfssl/wolfcrypt/aes.h>
 
-#if !defined(NO_AES) && defined(WOLFSSL_CMAC)
+#ifdef WOLFSSL_CMAC
+
+#include <wolfssl/wolfcrypt/aes.h>
```
bigbrett, line 30: it is a header include, so most likely innocuous, but do we want to be guarding the inclusion of `aes.h` with 
`!NO_AES`?

wolfssl/wolfcrypt/cmac.h
```diff
@@ -74,12 +82,15 @@ struct Cmac {
...
-} CmacType;
 
+#ifndef NO_AES
 #define WC_CMAC_TAG_MAX_SZ AES_BLOCK_SIZE
 #define WC_CMAC_TAG_MIN_SZ (AES_BLOCK_SIZE/4)
+#else
+/* Reasonable defaults */
+#define WC_CMAC_TAG_MAX_SZ 16
+#define WC_CMAC_TAG_MIN_SZ 4
+#endif
```
bigbrett, line 93: Since AES is the only cipher supported by CMAC for now, should we consider adding a compile time error if CMAC is enabled without AES? Or was this deliberate to allow for cryptoCbs to still work?

bigbrett (approved):
approved but consider adding compile error for non-AES enabled CMAC builds for now, unless that was deliberate to facilitate a cryptocb only approach.

bigbrett (changes requested):
CI failing

wolfssl/wolfcrypt/cmac.h
billphipps: Yeah, I had that in there and removed looking at a few other spots.  The first thing it does is check, but I think you are right.  I'll update.

wolfssl/wolfcrypt/cmac.h
billphipps: This was deliberate to support other types if the cryptocb wanted to support it.  The functions return FUNC_BAD_ARG if an unknown type is passed, so at least this will be detected at run time

wolfcrypt/src/cmac.c
```diff
@@ -267,44 +295,64 @@ int wc_CmacFinalNoFree(Cmac* cmac, byte* out, word32* outSz)
...
+                cmac->buffer[AES_BLOCK_SIZE - remainder] = 0x80;
+                subKey = cmac->k2;
+            }
+            xorbuf(cmac->buffer, cmac->digest, AES_BLOCK_SIZE);
+            xorbuf(cmac->buffer, subKey, AES_BLOCK_SIZE);
+            ret = wc_AesEncryptDirect(&cmac->aes, cmac->digest, cmac->buffer);
+            if (ret == 0) {
+                XMEMCPY(out, cmac->digest, *outSz);
+            }
+        }; break;
```
philljj, line 345: nits picked:
- `}; break;` break on newline would be more readable.
- `default :` space between default and :

wolfcrypt/src/cryptocb.c
```diff
@@ -141,6 +144,16 @@ static const char* GetHashTypeStr(int hash)
     return NULL;
 }
 
+#ifdef WOLFSSL_CMAC
+static const char* GetCmacTypeStr(int type)
+{
+    switch (type) {
+        case WC_CMAC_AES: return "AES";
```
philljj, line 151: case without break or default might trigger analyzer warnings.

wolfcrypt/src/cryptocb.c
```diff
@@ -1812,7 +1838,7 @@ int wc_CryptoCb_Cmac(Cmac* cmac, const byte* key, word32 keySz,
 
     return wc_CryptoCb_TranslateErrorCode(ret);
 }
-#endif
+#endif /* WOLFSSL_CMAC && !NO_AES */
```
philljj, line 1841: Isn't this endif just `/* WOLFSSL_CMAC */`?

wolfssl/wolfcrypt/cmac.h
```diff
@@ -40,16 +43,23 @@
...
+typedef enum CmacType {
+    WC_CMAC_AES = 1
+} CmacType;
+
 #ifndef WC_CMAC_TYPE_DEFINED
     typedef struct Cmac Cmac;
     #define WC_CMAC_TYPE_DEFINED
 #endif
 struct Cmac {
+    CmacType type;
```
dgarske, line 55: Enum types are hard to predict size for and this might break build-time compatibility for some folks. I would recommend adding the type lower down in the structure.

wolfcrypt/src/cmac.c
```diff
@@ -211,26 +223,35 @@ int wc_CmacUpdate(Cmac* cmac, const byte* in, word32 inSz)
...
+                }
+                ret = wc_AesEncryptDirect(&cmac->aes, cmac->digest,
+                        cmac->buffer);
+                if (ret == 0) {
+                    cmac->totalSz += AES_BLOCK_SIZE;
+                    cmac->bufferSz = 0;
+                }
             }
         }
+    }; break;
```
philljj, line 250: Same picky suggestion on break on newline and `default :` extra space.

bigbrett (dismissed):

philljj (dismissed):

dgarske (changes requested):
Really nice cleanups here. Just a minor item. Not required to fix, but it might improve things.

wolfssl/wolfcrypt/cmac.h
billphipps: Concur.  Moved down.

wolfcrypt/src/cryptocb.c
billphipps: Break shouldn't be necessary since this directly returns.  Concur that we could add defaults to all of these, but I'll follow the pattern for now.

wolfcrypt/src/cryptocb.c
billphipps: Good catch.  Updated.

wolfcrypt/src/cmac.c
billphipps: Concur on the extra default space. Updated.
I manually created a section for this case so I could declare variables that are only used for this CMAC type.  I wanted to make it clear this was the end of this scope as well as well as the case, so I place the break on the same line as the closing of that section.  I should probably pull out this case into a static function.  Maybe we'll do that when we add the next CMAC type.  I've left it as is.

dgarske (approved):

bigbrett (approved):

philljj (approved):

PR #8008, opened 2024-09-23
Fix '--depth=1' repos

bandi13:
When the repo was checked out as a shallow copy, we need to unshallow so FIPS builds can successfully find all the required tags and branches.


bandi13:
retest this please

JacobBarthelmeh (approved):

PR #8045, opened 2024-10-07
sniffer: set ssl->curSize before invoking Do* routines

rizlik:
# Description
commit 99a99e3d6e470c6eecb536d18f4042f6433afe93 changes DoApplication to use ssl->curSize as the size of the current decrypted record. Fix sniffer code to set this value.


bandi13 (approved):
This works. Thanks!

dgarske (approved):

PR #8063, opened 2024-10-10
ARM32 ASM: regenerated with fixes

SparkiDev:
# Description

Fix thumb interwork def check to be consistent and update #endif.
Remove duplicate check in generated C files.

# Testing

Tested ARM32 platforms with all algorithms and with inline assembly.

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


danielinux (approved):
Tested again (sha2-aes on thumb-interwork platform.

PR #8074, opened 2024-10-15
Revert "Merge pull request #8072 from rizlik/github-fix"

bandi13:
This reverts commit 0f8b4dbc6391743c8046933005cb05021c259532, reversing changes made to 743a78dc853a84ab44c72f80523398d1b6ec30dd.

### DIFF
--- a/.github/workflows/jwt-cpp.yml
+++ b/.github/workflows/jwt-cpp.yml
@@ -68,7 +68,8 @@ jobs:
       - name: Checkout OSP
         uses: actions/checkout@v4
         with:
-          repository: wolfssl/osp
+          repository: bandi13/wolfssl-osp
+          ref: 'updateJWTpatch'
           path: osp
 
       - name: Checkout jwt-cpp
--- a/.github/workflows/jwt-cpp.yml
+++ b/.github/workflows/jwt-cpp.yml
@@ -68,8 +68,7 @@ jobs:
       - name: Checkout OSP
         uses: actions/checkout@v4
         with:
-          repository: bandi13/wolfssl-osp
-          ref: 'updateJWTpatch'
+          repository: wolfssl/osp
           path: osp
 
       - name: Checkout jwt-cpp
--- a/.github/workflows/jwt-cpp.yml
+++ b/.github/workflows/jwt-cpp.yml
@@ -41,13 +41,17 @@ jobs:
           retention-days: 5
 
   build_pam-ipmi:
+    if: github.repository_owner == 'wolfssl'
     strategy:
       fail-fast: false
       matrix:
-        ref: [ 0.7.0 ]
-    name: ${{ matrix.ref }}
-    if: github.repository_owner == 'wolfssl'
-    runs-on: ubuntu-latest
+        config:
+          - ref: 0.7.0
+            runner: ubuntu-latest
+          - ref: 0.6.0
+            runner: ubuntu-22.04
+    name: ${{ matrix.config.ref }}
+    runs-on: ${{ matrix.config.runner }}
     needs: build_wolfssl
     steps:
       - name: Install dependencies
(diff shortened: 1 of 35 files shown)


bandi13:
Depends on https://github.com/wolfSSL/osp/pull/205.

bandi13:
Note that `jwt-cpp` tests pass if using the PR of https://github.com/wolfSSL/osp/pull/205.

.github/workflows/jwt-cpp.yml
```diff
@@ -44,10 +44,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ref: [ 0.6.0 ]
+        ref: [ 0.7.0 ]
```
julek-wolfssl, line 47: Why are we not testing `0.6.0` anymore?

.github/workflows/sssd.yml
```diff
@@ -13,46 +13,21 @@ concurrency:
...
-        run: tar -zcf build-dir.tgz build-dir
-
-      - name: Upload built lib
-        uses: actions/upload-artifact@v4
-        with:
-          name: wolf-install-sssd
-          path: build-dir.tgz
-          retention-days: 5
-
   sssd_check:
```
julek-wolfssl, line 16: Why are we including building wolfSSL in the `sssd_check` job?

julek-wolfssl (changes requested):

.github/workflows/jwt-cpp.yml
bandi13: 0.6.0 does not compile on the latest compiler because of a build error.

.github/workflows/sssd.yml
bandi13: I was just consolidating the job to be simpler. I didn't see a need to upload an artifact only to download it again. I could change the name of the stage if that's better.

.github/workflows/jwt-cpp.yml
julek-wolfssl: We should fix that error. What is the error?

julek-wolfssl: https://github.com/wolfSSL/wolfssl/actions/runs/11485310894 for example passes.

.github/workflows/sssd.yml
bandi13: Agreed, but we're only testing a single version so it seemed superfluous. I'll revert this change.

.github/workflows/jwt-cpp.yml
bandi13: Note that the example you gave uses the `ubuntu-22.04` runner. It will fail on `ubuntu-latest`. See: https://github.com/wolfSSL/wolfssl/actions/runs/11350595837/job/31569431321.

julek-wolfssl: If you don't want to invest time investigating the source of the error, we should at least run those tests on the version of Ubuntu it passes on. You can add the runner choice to the strategy to run 0.6 on the lts version and 0.7 on latest.

.github/workflows/sssd.yml
julek-wolfssl: It decreases the workload on the person updating the support. I think its worth it to separate on the initial write.

.github/workflows/jwt-cpp.yml
bandi13: It's not that I don't want to investigate the source of the error. It's simply adding in `#include <cstdint.h>` into `base.h`, which is what `jwt-cpp` has [already been done](https://github.com/Thalhammer/jwt-cpp/compare/v0.6.0...v0.7.0#diff-7884d932b55bcb4b9bbf516c2f3f7e55624c0dfdf8656783ec42c8d3d694db5eR6). I have done as you suggested and just added a `matrix`.

douzzer (approved):

PR #8087, opened 2024-10-16
Initial rewrite of X509 STORE to replicate openssl behavior

ColtonWilley:
# Description
Rewrite the X509 STORE internals to replicate openssl behavior. 

Fixes zd 18689

# Testing
 Added 9 additional test cases to api.c

# Checklist

 - [x] added tests
 - [ ] updated/added doxygen


src/x509_str.c
```diff
@@ -1010,22 +1212,56 @@ WOLFSSL_X509_LOOKUP* wolfSSL_X509_STORE_add_lookup(WOLFSSL_X509_STORE* store,
...
+            }
+    #endif
+        }
+        else {
+            if (store->certs != NULL) {
+                result = wolfSSL_sk_X509_push(store->certs, x509);
+                result = (result > 0) ? WOLFSSL_SUCCESS : WOLFSSL_FATAL_ERROR;
+            }
+            else {
+                result = wolfSSL_X509_STORE_add_ca(
```
SparkiDev, line 1262: We add the certificate to CA if we don't have a store?
Kinda confusing given the comment above.

SparkiDev (dismissed):

src/x509_str.c
ColtonWilley: So this is to support the SSL_CTX current behavior. SSL_CTX has a X509_STORE that is not created with new/free, and allocates internal members of X509_STORE  itself. It also shares a single underlying CM and does operations with it directly. store->certs is allocated in `wolfSSL_X509_STORE_new,` so the only way it can be NULL is the SSL_CTX using it. For those cases I opted to leave the behavior as is, because the SSL_CTX will be expecting the underlying CM to be in certain states after using `wolfSSL_X509_STORE_add_cert`.

SparkiDev: Add a comment

ColtonWilley: Ok, I do something similar in a few other places as well. Alternatively we could also check on `store->isDynamic` instead.

julek-wolfssl (changes requested):
Do we have tests for inter cert attributes like key usage and path length?

tests/api.c
```diff
@@ -59980,9 +60345,9 @@ static int test_X509_STORE_untrusted(void)
     /* Succeeds because path to loaded CA is available. */
     ExpectIntEQ(test_X509_STORE_untrusted_certs(untrusted2, 1, 0, 1),
             TEST_SUCCESS);
-    /* Fails because root CA is in the untrusted stack */
-    ExpectIntEQ(test_X509_STORE_untrusted_certs(untrusted3, 0,
-            X509_V_ERR_UNABLE_TO_GET_ISSUER_CERT_LOCALLY, 0), TEST_SUCCESS);
+    /* Root CA in untrusted chain is OK */
+    ExpectIntEQ(test_X509_STORE_untrusted_certs(untrusted3, 1, 0, 1),
+            TEST_SUCCESS);
```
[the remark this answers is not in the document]
ColtonWilley: The previous test did not mirror openssl behavior. Given those inputs and function calls, openssl is ok with it and verifies the chain.

src/x509_str.c
```diff
@@ -192,10 +197,14 @@ void wolfSSL_X509_STORE_CTX_cleanup(WOLFSSL_X509_STORE_CTX* ctx)
...
 
 
-void wolfSSL_X509_STORE_CTX_trusted_stack(WOLFSSL_X509_STORE_CTX *ctx, WOLF_STACK_OF(WOLFSSL_X509) *sk)
+void wolfSSL_X509_STORE_CTX_trusted_stack(WOLFSSL_X509_STORE_CTX *ctx,
+                                          WOLF_STACK_OF(WOLFSSL_X509) *sk)
 {
     if (ctx != NULL) {
-        ctx->chain = sk;
+        if (ctx->setTrustedSk != NULL) {
+            wolfSSL_sk_X509_free(ctx->setTrustedSk);
```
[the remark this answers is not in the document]
ColtonWilley: No, and in fact it should not be a free at all. This does not transfer ownership, caller maintains responsibility to free sk after using the STORE_CTX.

src/x509_str.c
```diff
@@ -849,6 +1042,20 @@ void wolfSSL_X509_STORE_free(WOLFSSL_X509_STORE* store)
                 wolfSSL_CertManagerFree(store->cm);
                 store->cm = NULL;
             }
+            if (store->certs != NULL) {
+                wolfSSL_sk_X509_free(store->certs);
```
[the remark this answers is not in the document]
ColtonWilley: No, when certs are added to that list through `X509_STORE_add_cert()`, the caller retains ownership of the X509 object. Same with `trusted`. We only own the list in `owned` which are certs we read from file in `X509_STORE_load_locations()`, those are freed with a pop_free()

tests/api.c
```diff
@@ -80131,7 +80496,7 @@ static int test_wolfSSL_X509_load_crl_file(void)
 #ifdef WC_RSA_PSS
         ExpectIntEQ(wolfSSL_CertManagerVerify(store->cm,
             "certs/rsapss/server-rsapss-cert.pem", WOLFSSL_FILETYPE_PEM),
-            WC_NO_ERR_TRACE(CRL_CERT_REVOKED));
+            WC_NO_ERR_TRACE(ASN_NO_SIGNER_E));
 #endif
```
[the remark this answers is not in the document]
ColtonWilley: So this one is a bit tricky. This test mixes use of X509_STORE and CertManager directly, which is not really designed that way for external use. The certs are added using `X509_LOOKUP_load_file()` but then instead of using associated X509_* interfaces to verify, the test reaches directly in to the CertManager and does a verify. This test does not form a chain to a self signed CA, previous behavior that was OK and so it proceeded to the CRL checks which failed. New behavior fails before CRL checks even happen as a chain cannot be made. I believe this is consistent with openssl error codes for the same inputs using X509_verify_cert().

src/x509_str.c
```diff
@@ -106,51 +125,37 @@ int wolfSSL_X509_STORE_CTX_init(WOLFSSL_X509_STORE_CTX* ctx,
...
+                }
+            }
+
+            for (i = 0; i < wolfSSL_sk_X509_num(sk); i++) {
+                ret = wolfSSL_sk_X509_push(ctx->ctxIntermediates,
+                                           wolfSSL_sk_X509_value(sk, i));
+                if (ret <= 0) {
+                    return WOLFSSL_FAILURE;
                 }
             }
```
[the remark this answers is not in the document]
ColtonWilley: Great catch, after review it does not need to be copied. I will make the changes.

src/x509_str.c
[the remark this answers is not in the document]
ColtonWilley: Fixed in latest commit.

src/x509_str.c
```diff
@@ -255,19 +264,15 @@ static void SetupStoreCtxError(WOLFSSL_X509_STORE_CTX* ctx, int ret)
     wolfSSL_X509_STORE_CTX_set_error_depth(ctx, depth);
 }
 
-/* Verifies certificate chain using WOLFSSL_X509_STORE_CTX
- * returns 0 on success or < 0 on failure.
- */
-int wolfSSL_X509_verify_cert(WOLFSSL_X509_STORE_CTX* ctx)
+static int wolfSSL_X509_verify_cert_ex(WOLFSSL_X509_STORE_CTX* ctx)
```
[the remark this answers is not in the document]
ColtonWilley: Reworked internal naming.

src/x509_str.c
```diff
@@ -743,7 +900,7 @@ int wolfSSL_X509_STORE_CTX_get1_issuer(WOLFSSL_X509 **issuer,
         for (node = ctx->chain; node != NULL; node = node->next) {
             if (wolfSSL_X509_check_issued(node->data.x509, x) ==
                                                             WOLFSSL_X509_V_OK) {
-                *issuer = x;
+                *issuer = node->data.x509;
```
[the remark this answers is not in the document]
ColtonWilley: No but investigating if we did need one led to discovering even more discrepancy with openssl in memory handling for get issuer. Rewrote get issuer internals entirely.

ColtonWilley:
Retest this please.

src/x509_str.c
```diff
@@ -1010,22 +1238,68 @@ WOLFSSL_X509_LOOKUP* wolfSSL_X509_STORE_add_lookup(WOLFSSL_X509_STORE* store,
...
+                    result = WOLFSSL_SUCCESS;
+                }
+                else {
+                    result = WOLFSSL_FATAL_ERROR;
+                }
+            }
+    #endif
+        }
+        else {
+            if (store->certs != NULL) {
```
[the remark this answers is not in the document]
ColtonWilley: I strongly disagree. Previous behavior is simply wrong, and a security vulnerability. If your same code and certs break with this new behavior, then the chain is incorrect and should not verify. There are multiple options to work around this already with openssl compat APIs, you can set a partial chain verification flag or set a trusted stack to specify trusted certs to use. In those options are not used, we should not expose additional functionality to skirt the proper verification behavior.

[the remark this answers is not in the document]
JacobBarthelmeh: Supporting the flag X509_V_FLAG_PARTIAL_CHAIN would work and follow OpenSSL behavior. Currently that flag would have no effect though. So it would take an adjustment to account for if the flag is set or a macro in the meantime.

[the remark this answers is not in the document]
julek-wolfssl: If we change functionality this should be high up on the release notes.

tests/api.c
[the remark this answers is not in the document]
julek-wolfssl: So OpenSSL trusts a self signed CA in the untrusted chain? I would assume that it should fail if it does not authenticate to something in the trusted chain.

Also, you are changing how the test works by passing `1` to `loadCA`. This is now a different test. Please test for both cases where the CA is and isn't present in the trusted chain/store.

tests/api.c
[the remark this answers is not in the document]
julek-wolfssl: Ok. Since this is an older test and I don't believe that it was used for a specific case where the exact return code is expected.

JacobBarthelmeh (approved):

src/x509_str.c
```diff
@@ -849,6 +1042,20 @@ void wolfSSL_X509_STORE_free(WOLFSSL_X509_STORE* store)
...
+                wolfSSL_sk_X509_free(store->certs);
+                store->certs = NULL;
+            }
+            if (store->owned != NULL) {
+                wolfSSL_sk_X509_pop_free(store->owned, wolfSSL_X509_free);
+                store->owned = NULL;
+            }
+#if !defined(WOLFSSL_SIGNER_DER_CERT)
+            if (store->trusted != NULL) {
+                wolfSSL_sk_X509_free(store->trusted);
```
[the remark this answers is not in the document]
ColtonWilley: No, only ->owned gets a pop free. These are either owned by caller for add_cert() or ->owned for load_locations().

src/x509_str.c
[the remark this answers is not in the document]
ColtonWilley: Confirmed that the flag works properly, and added a test for it.

ColtonWilley:
> Do we have tests for inter cert attributes like key usage and path length?

We do for direct usage of the CM, but I dont think anything for X509 STORE.

src/x509_str.c
```diff
@@ -300,17 +286,170 @@ int wolfSSL_X509_verify_cert(WOLFSSL_X509_STORE_CTX* ctx)
...
+            if (added == 1) {
+                wolfSSL_CertManagerUnloadTempIntermediateCerts(ctx->store->cm);
+            }
+        }
+        if (orig != NULL) {
+            ctx->current_cert = orig;
+        }
+    }
+
+    return ret == WOLFSSL_SUCCESS ? WOLFSSL_SUCCESS : WOLFSSL_FAILURE;
```
dgarske, line 445: Can you confirm you only want to return success or failure and not any other possible return codes? Old logic was >= 0 success else failure.
@douzzer does this need to be `WC_NO_ERR_TRACE(WOLFSSL_FAILURE)`, etc?

src/x509_str.c
```diff
@@ -300,17 +286,170 @@ int wolfSSL_X509_verify_cert(WOLFSSL_X509_STORE_CTX* ctx)
...
+        certs = ctx->ctxIntermediates;
+    }
+    else {
+        /* Add the intermediates provided on init to the list of untrusted
+         * intermediates to be used */
+        for (i = 0; i < wolfSSL_sk_X509_num(ctx->ctxIntermediates); i++) {
+            ret = wolfSSL_sk_X509_push(certs,
+                wolfSSL_sk_X509_value(ctx->ctxIntermediates, i));
+            if (ret <= 0) {
+                return WOLFSSL_FAILURE;
```
dgarske, line 336: Should this be a `goto exit`?

src/x509_str.c
```diff
@@ -278,9 +264,9 @@ int wolfSSL_X509_verify_cert(WOLFSSL_X509_STORE_CTX* ctx)
         if (ret != WC_NO_ERR_TRACE(ASN_BEFORE_DATE_E) &&
             ret != WC_NO_ERR_TRACE(ASN_AFTER_DATE_E)) {
             /* wolfSSL_CertManagerVerifyBuffer only returns ASN_AFTER_DATE_E or
-             ASN_BEFORE_DATE_E if there are no additional errors found in the
-             cert. Therefore, check if the cert is expired or not yet valid
-             in order to return the correct expected error. */
+            ASN_BEFORE_DATE_E if there are no additional errors found in the
```
dgarske, line 267: NIT: Please use * at start of each line.

src/x509_str.c
```diff
@@ -36,9 +36,20 @@
...
-/*******************************************************************************
+#ifdef OPENSSL_EXTRA
+static int X509StoreGetIssuerEx(WOLFSSL_X509 **issuer,
+                            WOLFSSL_STACK *certs, WOLFSSL_X509 *x);
+static int X509StoreAddCa(WOLFSSL_X509_STORE* store,
+                                          WOLFSSL_X509* x509, int type);
+#endif
+
+#ifndef WOLFSSL_X509_STORE_DEFAULT_MAX_DEPTH
+#define WOLFSSL_X509_STORE_DEFAULT_MAX_DEPTH 100
```
dgarske, line 47: How was the value of 100 choose? Perhaps a code comment?

dgarske (changes requested):

src/x509_str.c
ColtonWilley: Chosen because openssl uses that as its default max depth. I will leave a comment.

src/x509_str.c
ColtonWilley: Yes it absolutely should be, great catch I will make the change.

wolfssl/ssl.h
```diff
@@ -3304,7 +3312,8 @@ enum {
     WOLFSSL_DTLSV1_3 = 7,
 
     WOLFSSL_USER_CA  = 1,          /* user added as trusted */
-    WOLFSSL_CHAIN_CA = 2           /* added to cache from trusted chain */
+    WOLFSSL_CHAIN_CA = 2,          /* added to cache from trusted chain */
+    WOLFSSL_INTER_CA = 3           /* Intermediate CA */
```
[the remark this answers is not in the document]
ColtonWilley: So I basically just needed a separate type so I can unload only my newly added temp certs. I am open to other naming schemes on this if anybody has a suggestion, I had also considered WOLFSSL_TEMP_CA

src/x509_str.c
ColtonWilley: Talked with douzzer, he said this should be OK.

dgarske (approved):

src/x509_str.c
douzzer: @ColtonWilley the first half of the question still stands -- do we really want to be masking the true return code here?

ColtonWilley: Yes actually I believe we do. I believe openssl only returns 0 or 1, and uses `X509_STORE_CTX_get_error` for error reporting.

julek-wolfssl: OpenSSL returns `-1` on an error, `0` on verification failure, and `1` on verification success. This rework conforms to that API.

wolfssl/ssl.h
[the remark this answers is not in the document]
julek-wolfssl: I think `WOLFSSL_TEMP_CA` or `WOLFSSL_X509_STORE_TEMP_CA` is a better name to indicate that these certs will be cleaned up soon.

src/x509_str.c
```diff
@@ -53,7 +65,12 @@ WOLFSSL_X509_STORE_CTX* wolfSSL_X509_STORE_CTX_new_ex(void* heap)
...
-        if (wolfSSL_X509_STORE_CTX_init(ctx, NULL, NULL, NULL) !=
+        if ((ctx->owned = wolfSSL_sk_X509_new_null()) == NULL) {
+            XFREE(ctx, heap, DYNAMIC_TYPE_X509_CTX);
+            ctx = NULL;
+        }
+        if (ctx != NULL &&
+            wolfSSL_X509_STORE_CTX_init(ctx, NULL, NULL, NULL) !=
                 WOLFSSL_SUCCESS) {
             XFREE(ctx, heap, DYNAMIC_TYPE_X509_CTX);
             ctx = NULL;
```
julek-wolfssl, line 76: I think its safer to call `wolfSSL_X509_STORE_CTX_free` on error here. The pain point would be if `wolfSSL_X509_STORE_CTX_init` allocates something then we don't free it here.

src/x509_str.c
```diff
@@ -78,6 +95,18 @@ void wolfSSL_X509_STORE_CTX_free(WOLFSSL_X509_STORE_CTX* ctx)
...
+            wolfSSL_sk_X509_free(ctx->chain);
+        }
+        if (ctx->owned != NULL) {
+            wolfSSL_sk_X509_pop_free(ctx->owned, NULL);
+        }
+
+        if (ctx->current_issuer != NULL) {
+            wolfSSL_X509_free(ctx->current_issuer);
+            ctx->current_issuer = NULL;
+        }
```
julek-wolfssl, line 109: Either all of the members needs to be set to `NULL` after free or none of them.

src/x509_str.c
```diff
@@ -734,34 +912,65 @@ WOLF_STACK_OF(WOLFSSL_X509)* wolfSSL_X509_STORE_get1_certs(
...
+
+#ifdef WOLFSSL_SIGNER_DER_CERT
+    ret = x509GetIssuerFromCM(issuer, ctx->store->cm, x);
+#else
+    ret = X509StoreGetIssuerEx(issuer, ctx->store->trusted, x);
+    if ((ret == WOLFSSL_SUCCESS) && (*issuer != NULL)) {
+        *issuer = wolfSSL_X509_dup(*issuer);
+        return (*issuer != NULL) ? WOLFSSL_SUCCESS : WOLFSSL_FAILURE;
+    }
+#endif
```
julek-wolfssl, line 935: I think its better and easier to use `wolfSSL_X509_up_ref` here instead of `wolfSSL_X509_dup`.

src/x509_str.c
```diff
@@ -827,6 +1051,30 @@ WOLFSSL_X509_STORE* wolfSSL_X509_STORE_new(void)
...
+    while (cnt > 0 && i > 0) {
+        /* The inner X509 is owned by somebody else, NULL out the reference */
+        obj = wolfSSL_sk_X509_OBJECT_value(objs, i);
+        if (obj != NULL) {
+            obj->type = 0;
+            obj->data.x509 = NULL;
+        }
+        cnt--;
+        i--;
+    }
```
julek-wolfssl, line 1072: This could also be a `crl` so maybe better to use `obj->data.ptr`. Same effect just showing that we don't want to free anything the `WOLFSSL_X509_OBJECT` holds.

julek-wolfssl: In the future it would be nice to up ref what gets put into the object list so that we can just call `pop_free` but that is not necessary on this PR.

src/x509_str.c
```diff
@@ -1010,22 +1275,68 @@ WOLFSSL_X509_LOOKUP* wolfSSL_X509_STORE_add_lookup(WOLFSSL_X509_STORE* store,
...
+
+    WOLFSSL_ENTER("wolfSSL_X509_STORE_add_cert");
+    if (store != NULL && store->cm != NULL && x509 != NULL
+                                                && x509->derCert != NULL) {
+        /* Mimic the openssl behavior, must be self signed to be considered
+         * trusted, addCA() internals will do additional checks for
+         * CA=TRUE */
+        if (wolfSSL_X509_NAME_cmp(&x509->issuer, &x509->subject) == 0) {
+            result = X509StoreAddCa(store, x509, WOLFSSL_USER_CA);
+    #if !defined(WOLFSSL_SIGNER_DER_CERT)
```
julek-wolfssl, line 1312: Why `!WOLFSSL_SIGNER_DER_CERT` here?

julek-wolfssl (changes requested):
Almost there. Last round of comments.

src/x509_str.c
ColtonWilley: I agree, good catch.

src/x509_str.c
ColtonWilley: Great catch on the CRLs. Agreed this whole thing could use a second pass at some point to take a close look at better using refcounts, but I think its large enough as is.

src/x509_str.c
ColtonWilley: If !WOLFSSL_SIGNER_DER_CERT, CM internals do not hold onto the full DER buffer of the cert. In this case we need to keep a copy of it ourselves to meet requirements to return the chain. If WOLFSSL_SIGNER_DER_CERT, is defined, no need to keep a separate trusted stack as we can retrieve DER buffer from the underlying CM.

src/x509_str.c
ColtonWilley: Changes made

JacobBarthelmeh (approved):

dgarske (dismissed):

wolfssl/openssl/ssl.h
```diff
@@ -643,8 +643,7 @@ typedef WOLFSSL_X509_NAME_ENTRY X509_NAME_ENTRY;
 #define X509_V_FLAG_CRL_CHECK     WOLFSSL_CRL_CHECK
 #define X509_V_FLAG_CRL_CHECK_ALL WOLFSSL_CRL_CHECKALL
 
-#define X509_V_FLAG_PARTIAL_CHAIN 0x80000
-#define X509_V_FLAG_TRUSTED_FIRST 0
```
JacobBarthelmeh, line 647: Why was X509_V_FLAG_TRUSTED_FIRST removed here?

douzzer: misunderstanding -- fixed.  it's needed for gRPC.

ColtonWilley: Its defined value is wrong. Also @douzzer and I discussed and decided only flags with wolfssl equivalents should be here.

JacobBarthelmeh (approved):

douzzer:
retest this please

douzzer (approved):

PR #8090, opened 2024-10-17
Add Visual Studio 2022 Project Files for wolfSSL, Test, & Benchmark

gojimmypi:
# Description

This is a no-code PR that adds Visual Studio 2022 Project Files to the wolfCrypt `test` and `benchmark` apps, as well as a root-level `wolfssl` as a reference library project.

I've used Visual Studio for quite a long time, and even I struggled with the existing VS2010 files that attempt to convert when opening in VS2022, but in the end require some non-intuitive changes to actually compile and run.

This PR allows the  `test` and `benchmark` apps to work out of the box for VS2022 users. Open the project file, run. The only thing that might need attention is the specific toolchain version, but VS2022 should handle this gracefully.

Note that although usually excluded in `.gitignore`, the respective `[app]-VS2022.vcxproj.user`  file is needed for this line:

```
<LocalDebuggerWorkingDirectory>$(ProjectDir)../../</LocalDebuggerWorkingDirectory>
```

This is needed as the Linux command-line app usually runs from the root, and (for instance) the default certs location is currently hard-coded as `./certs`, which is not found from the `wolfssl\wolfcrypt\benchmark`  and `wolfssl\wolfcrypt\test` project directories.  The code of course could be gated or changed, but including the the user settings file for VS2022 users is vastly simpler.

* edit:

# Background:

Opening the existing `wolfssl.vcxproj` results in an as-expected dialog box, warning of the old version:

image

The other examples mentioned in the above dialog box have not been addressed in this PR, but could be considered in a future PR.

The warning alone is perhaps just cosmetically undesired, but there are more serious problems, too.

Changing to DLL Release gives this warning:

image

However, even after compiling there, when opening the respective `benchmark` app and attempting to compile, this error is observed:

image

Yes, there are a variety of manual settings that can address these issues, but having the VS2022 files provides a much nicer first experience when first using wolfSSL.

A key, non-unintuitive edit needs to be made to the `Working Directory` when using the existing project files (_not_ those in this PR):

image

Additionally, the enclosed project files includes both the root level wolfSSL project *and* the local project in the solution file:

image

For completeness: the `wolfssl-gojimmgpi-pr` portion of the solution file shown in my screen snip above, is _not_ native to Visual Studio. For that I am using the [File Explorer Extension for Visual Studio](https://marketplace.visualstudio.com/items?itemName=MadsKristensen.WorkflowBrowser). 

Fixes zd# n/a

# Testing

Tested on Windows 11, Visual Studio 2022 v17.11.5

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ ctaocrypt/src/src/
 *.cache
 .dirstamp
 *.user
+!*-VS2022.vcxproj.user
 configure
 config.*
 !cmake/config.in
--- a/Makefile.am
+++ b/Makefile.am
@@ -141,6 +141,7 @@ ACLOCAL_AMFLAGS= -I m4
 EXTRA_DIST+= .cyignore
 EXTRA_DIST+= wolfssl.vcproj
 EXTRA_DIST+= wolfssl.vcxproj
+EXTRA_DIST+= wolfssl-VS2022.vcxproj
 EXTRA_DIST+= wolfssl64.sln
 EXTRA_DIST+= valgrind-error.sh
 EXTRA_DIST+= valgrind-bash.supp
--- a/wolfcrypt/benchmark/benchmark-VS2022.sln
+++ b/wolfcrypt/benchmark/benchmark-VS2022.sln
@@ -1,0 +1,87 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version 17
+VisualStudioVersion = 17.11.35327.3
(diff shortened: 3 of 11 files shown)


gojimmypi:
I've force-pushed a commit that appends manually-adding CR/LF at the end of Visual Studio project files.

The files were included in the original PR as-generated by Visual Studio.

The edit fixes the no-ending-linebreak warning.

douzzer (approved):

PR #8097, opened 2024-10-22
Fix TLS v1.2 session resumption edge cases

julek-wolfssl:
- Fix for TLS v1.2 session resumption with tickets where the server decides to do a full handshake. The wrong sessionIDSz was being checked and should be the arrays one since it get set from the server_hello.
- Fixes issue with server_hello including session_id when not doing resumption.
- Optimize entropy use. Only generate the exact amount of random data that we will actually keep.
- Send a new session ticket when rejecting a session ticket and session tickets are enabled

Refactor done as part of work on ZD18822

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -34510,6 +34510,29 @@ static int DoSessionTicket(WOLFSSL* ssl, const byte* input, word32* inOutIdx,
 
 #ifndef WOLFSSL_NO_TLS12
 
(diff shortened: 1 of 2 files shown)


julek-wolfssl:
Retest this please

dgarske (approved):

dgarske (approved):

JacobBarthelmeh (approved):

douzzer (approved):

PR #8108, opened 2024-10-23
Fix for TLS v1.2 session resumption with tickets where the server decides to do full handshake

dgarske:
# Description

Fix for TLS v1.2 session resumption with tickets where the server decides to do a full handshake. The wrong sessionIDSz was being checked and should be the arrays one since it get set from the server_hello.

Fixes ZD 18822

# Testing

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -17471,6 +17471,18 @@ int DoHandShakeMsgType(WOLFSSL* ssl, byte* input, word32* inOutIdx,
         case certificate_request:
         case server_hello_done:
             if (ssl->options.resuming) {
+                /* Client requested resumption, but server is doing a
+                 * full handshake */
+
+                /* The server's decision to resume isn't known until after the
+                 * "server_hello". If subsequent handshake messages like
+                 * "certificate" or "server_key_exchange" are recevied then we
+                 * are doing a full handshake */
+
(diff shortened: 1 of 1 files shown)


src/internal.c
```diff
@@ -17471,6 +17471,18 @@ int DoHandShakeMsgType(WOLFSSL* ssl, byte* input, word32* inOutIdx,
         case certificate_request:
         case server_hello_done:
             if (ssl->options.resuming) {
+                /* Client requested resumption, but server is doing a
+                 * full handshake */
+
+                /* The server's decision to resume isn't known until after the
+                 * "server_hello". It subsequent handshake messages like
```
julek-wolfssl, line 17478:
```suggestion
                 * "server_hello". If subsequent handshake messages like
```

julek-wolfssl (approved):

src/internal.c
dgarske: Fixed

dgarske:
Applied commit on top of https://github.com/wolfSSL/wolfssl/pull/8097. That also resolves an issue with the server_hello session id being populated when it should not. Resolves the `964: test_session_ticket_no_id` issue.

PR #8104, opened 2024-10-23
Name VS2022 binary output wolfssl via project name

gojimmypi:
# Description

This is a no-code-change PR that updates the project name of the new VS 2022 project files introduced in https://github.com/wolfSSL/wolfssl/pull/8090.

Currently the  output binary name would be `wolfssl-VS2022.lib` and/or `wolfssl-VS2022.dll`.

The new name will be `wolfssl.lib` and/or `wolfssl.dll`.

For instance, when compiling the [benchmark example](https://github.com/wolfSSL/wolfssl/tree/master/wolfcrypt/benchmark), note the new output file when compiling the wolfssl project is instead `wolfssl.lib` with this PR. 

```
1>Generating Code...
1>wolfssl-VS2022.vcxproj -> C:\workspace\wolfssl-gojimmypi-pr\wolfcrypt\benchmark\Debug\x64\wolfssl.lib
2>------ Rebuild All started: Project: benchmark-VS2022, Configuration: Debug x64 ------
2>benchmark.c
2>benchmark-VS2022.vcxproj -> C:\workspace\wolfssl-gojimmypi-pr\wolfcrypt\benchmark\x64\Debug\benchmark-VS2022.exe
========== Rebuild All: 2 succeeded, 0 failed, 0 skipped ==========
========== Rebuild completed at 9:33 AM and took 07.492 seconds ==========
```

Fixes zd# n/a

# Testing

Tested on Windows 11, Visual Studio 2022

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen
 - [ ] updated appropriate READMEs
 - [ ] Updated manual and documentation

PR #8112, opened 2024-10-25
Add VS2022 DLL Release options.h build configuration

gojimmypi:
# Description

Building on  https://github.com/wolfSSL/wolfssl/pull/8090 and https://github.com/wolfSSL/wolfssl/pull/8104, this PR adds a new `DLL Release (config.h)` Build Configuration to the Visual Studio 2022 Project file that instead of using the `user_settings.h` from the [IDE/WIN](https://github.com/wolfSSL/wolfssl/tree/master/IDE/WIN) directory, uses the generated wolfssl `config.h` file.

Let's say, for whatever reason, you'd like to used the same wolfssl `options.h` created by a Linux `./configure` command, also for a Windows build of a DLL.

One place that this might be useful is a NodeJS/NPM module such as my [wolfcrypt_nodejs](https://github.com/gojimmypi/wolfcrypt_nodejs/blob/dev/README.md) branch used in both Windows and Linux. See also https://github.com/wolfSSL/wolfcrypt_nodejs/pull/8

Without the capability added in this PR, one would need to know both the Linux `./configure` command-line options, as well as the respective settings to exactly match in the Windows `user_settings.h`. 

Alternatively, the option of course exists for use a `user_settings.h` in the Linux environment. This PR is only for using  `config.h`  for a Windows build.

We'll assume that the respective `./configure` options in WSL are provided to give compatibility in both Windows and Linux.

Otherwise, for an example here, if in WSL on 64bit Windows 11 using the `--enable-all`:

```bash
./autogen.sh
./configure --enable-all
```

Add this to the respective generated `options.h` to also use in Windows to address incompatibilities observed for this PR.

```c
/* Disable some items from Ubuntu (WSL) ./configure --enable-all that were
 * observed to cause problems in Windows when sharing the same options.h */

#undef HAVE___UINT128_T
#undef WOLFSSL_QUIC

#undef HAVE_HPKE
#undef HAVE_ECH

#undef WOLFSSL_SP_X86_64_ASM
#undef WOLFSSL_SP_X86_64
#undef WOLFSSL_SP_ASM

#undef HAVE_INTEL_AVX2
#undef HAVE_INTEL_AVX1
#undef USE_INTEL_SPEEDUP

#define WOLFSSL_UINT128_T_DEFINED
#define WOLFSSL_NO_ATOMICS
#ifndef INT32_MAX
    /* Maximum value for a 32-bit signed integer, used in evp.c */
    #define INT32_MAX 2147483647
#endif

/* WOLFSSL_DLL is typically defined in the project build config.
 * undef here to avoid "Duplicate definition" warnings. */
#undef WOLFSSL_DLL
```

Fixes zd# n/a

# Testing

Tested on Windows 11, Visual Studio 17.11.5, Platform Toolset v143

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ config.*
 !cmake/config.in
 *Debug/
 *Release/
+*Release (options.h)/
 *.ncb
 *.suo
 *.sdf
--- a/wolfssl-VS2022.vcxproj
+++ b/wolfssl-VS2022.vcxproj
@@ -25,11 +25,27 @@
       <Configuration>DLL Debug</Configuration>
       <Platform>ARM64</Platform>
     </ProjectConfiguration>
+    <ProjectConfiguration Include="DLL Release (options.h)|ARM64">
+      <Configuration>DLL Release (options.h)</Configuration>
+      <Platform>ARM64</Platform>
+    </ProjectConfiguration>
+    <ProjectConfiguration Include="DLL Release (options.h)|Win32">
+      <Configuration>DLL Release (options.h)</Configuration>
+      <Platform>Win32</Platform>
+    </ProjectConfiguration>
+    <ProjectConfiguration Include="DLL Release (options.h)|x64">
+      <Configuration>DLL Release (options.h)</Configuration>
+      <Platform>x64</Platform>
+    </ProjectConfiguration>
     <ProjectConfiguration Include="DLL Release|Win32">
       <Configuration>DLL Release</Configuration>
       <Platform>Win32</Platform>
     </ProjectConfiguration>
     <ProjectConfiguration Include="DLL Release|x64">
+      <Configuration>DLL Release</Configuration>
+      <Platform>x64</Platform>
+    </ProjectConfiguration>
+    <ProjectConfiguration Include="DLL Release (options.h)|x64">
       <Configuration>DLL Release</Configuration>
       <Platform>x64</Platform>
     </ProjectConfiguration>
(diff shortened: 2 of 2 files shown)


gojimmypi:
Jenkins retest this please

.gitignore
```diff
@@ -16,6 +16,7 @@ config.*
 !cmake/config.in
 *Debug/
 *Release/
+*Release (options.h)/
```
dgarske, line 19: Why only for release? Where is the README to go along with these new VS 2022 projects? Would be nice to explain the various configuration options and where the user_settings.h is coming from.

dgarske (changes requested):

.gitignore
gojimmypi: > Why only for release? 

The initial settings for only release build was as related to https://github.com/wolfSSL/wolfcrypt_nodejs/pull/8 where debugging was unlikely.

It was also an exploratory PR, to see if there was interest in an `options.h` build for Windows. 

But you are right; there should be a debug option. Potentially also `options.h` configs for 32bit debug and release, too.

> Where is the README to go along with these new VS 2022 projects? 

I neglected that, as I was documenting in an upcoming blog that I would have referred to. 

I'll add some README info, as well.

> Would be nice to explain the various configuration options and where the user_settings.h is coming from.

That's also in the upcoming blog, but you ae correct it should also be in the README.

dgarske: Seems like documentation needs to be added similar to PR description?

gojimmypi:
I converted to draft while I work on adding `debug` build configs and related documentation.

PR #8118, opened 2024-10-28
wolfcrypt tests: disable ecc sign/verify of all zero digest

bigbrett:
Adds a non-feature-specific macro to disable ECC sign/verify of an all-zero digest in wolfCrypt tests. 

Lots of ECC crypto hardware accelerators cannot handle a zero input digest for sign/verify. This flag is required when using wolfCrypt tests to exercise platform-specific crypto callbacks for hardware accelerators on these platforms that aren't natively supported in wolfCrypt.

We already disable this on a feature-by-feature basis (KCAPI, CRYPTOCELL, ASYNC, etc), but clearly enough hardware has issues with this feature that we should allow a generic way to turn it off. 

I will add the macro to chapter 2 of the documentation when this merges. I don't think this is worthwhile to add as a configure flag, given the niche usage.

### DIFF
--- a/wolfcrypt/test/test.c
+++ b/wolfcrypt/test/test.c
@@ -30869,9 +30869,8 @@ static wc_test_ret_t ecc_test_curve_size(WC_RNG* rng, int keySize, int testVerif
 #if !defined(ECC_TIMING_RESISTANT) || (defined(ECC_TIMING_RESISTANT) && \
     !defined(WC_NO_RNG) && !defined(WOLFSSL_KCAPI_ECC))
 #ifdef HAVE_ECC_SIGN
-    /* ECC w/out Shamir has issue with all 0 digest */
-    /* WC_BIGINT doesn't have 0 len well on hardware */
-#if defined(ECC_SHAMIR) && !defined(WC_TEST_NO_ECC_SIGN_VERIFY_ZERO_DIGEST)
+    /* some hardware doesn't support sign/verify of all zero digest */
+#if !defined(WC_TEST_NO_ECC_SIGN_VERIFY_ZERO_DIGEST)
     /* test DSA sign hash with zeros */
     for (i = 0; i < (int)ECC_DIGEST_SIZE; i++) {
         digest[i] = 0;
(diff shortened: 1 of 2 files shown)


wolfcrypt/test/test.c
```diff
@@ -30873,7 +30873,8 @@ static wc_test_ret_t ecc_test_curve_size(WC_RNG* rng, int keySize, int testVerif
     /* WC_BIGINT doesn't have 0 len well on hardware */
     /* Cryptocell has issues with all 0 digest */
 #if defined(ECC_SHAMIR) && !defined(WOLFSSL_ASYNC_CRYPT) && \
-    !defined(WOLFSSL_CRYPTOCELL)
+    !defined(WOLFSSL_CRYPTOCELL) && \
```
dgarske, line 30876: I wonder if ECC_SHAMIR is actually still required. Also it would be great to have settings.h set WC_TEST_NO_ECC_SIGN_VERIFY_ZERO_DIGEST if WOLFSSL_ASYNC_CRYPT or WOLFSSL_CRYPTOCELL

dgarske (changes requested):

wolfssl/wolfcrypt/settings.h
```diff
@@ -3168,6 +3168,14 @@ extern void uITRON4_free(void *p) ;
     #undef NO_DH
 #endif
 
+/* CryptoCell defines */
+#ifdef WOLFSSL_CRYPTOCELL
```
dgarske, line 3172: I think you may have lost the `WOLFSSL_ASYNC_CRYPT` one? This section can just be about which ECC signing hardware support zero digest.

dgarske (changes requested):

wolfssl/wolfcrypt/settings.h
bigbrett: @dgarske I added it later on in the WOLFSSL_ASYNC_CRYPT section of settings.h, see line [3206](https://github.com/wolfSSL/wolfssl/pull/8118/files#diff-9efb2cca94d744438b9691461eb2bf010ecf7b2e9a6e24c8d6fd7872a663197aR3206).

That is fair, I was thinking I'd add the disable to the section of settings.h that corresponds to each higher level feature that wants to disable it, since that is what would drive the decision.

That way all the dependent options for ASYNC are set in the ASYNC section, the dependent options for cryptocell are set in the cryptocell section, etc. I noticed there wasn't a centralized cryptocell section for settings.h, so I created one.

Let me know if you would rather have it relocated to the ECC section?

bigbrett:
Jenkins retest this please

wolfcrypt/test/test.c
bigbrett: fixed

dgarske (approved):

SparkiDev (approved):

PR #8126, opened 2024-10-30
Introduce IDE-FIPS directory

gojimmypi:
# Description

Upon reconsideration after the merge of https://github.com/wolfSSL/wolfssl/pull/8090, I think it would be best to move all project files to the respective [IDE/\<environment\>](https://github.com/wolfSSL/wolfssl/tree/master/IDE) directory, out of the respective source directories such as [wolfcrypt/benchmark](https://github.com/wolfSSL/wolfssl/tree/master/wolfcrypt/benchmark).

Additionally, it may be best to isolate FIPS and non-FIPS environments.

This PR proposes a new `IDE-FIPS` for all the FIPS-related builds in preparation for the moves.

Fixes zd# n/a

# Testing

Not tested.

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/IDE-FIPS/README.md
+++ b/IDE-FIPS/README.md
@@ -1,0 +1,6 @@
+# IDE-FIPS
+
+This directory contains FIPS-only IDE builds to better isolate from non-FIPS environments.
+
+When adding a new environment, remember to edit the local [include.am](./include.am) file
+and probably create a new `IDE-FIPS/<NEW-OE>/include.am` file as well.
--- a/IDE-FIPS/include.am
+++ b/IDE-FIPS/include.am
@@ -1,0 +1,8 @@
+# vim:ft=automake
+# included from Top Level Makefile.am
+# All paths should be given relative to the root
+
+# For example, when WIN-SRTP-KDF-140-3 is ready:
+# include IDE-FIPS/WIN-SRTP-KDF-140-3/include.am
+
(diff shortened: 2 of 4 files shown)


gojimmypi:
Jenkins retest this please

kaleb-himes (approved):
Excellent prep work on getting the transition space ready @gojimmypi, thank you!

douzzer (changes requested):
I think it may be an unforced error to separate FIPS from non-FIPS artifacts for a given IDE.

My thinking is that the big variable here is the IDE, not FIPSness, and we go to a lot of trouble to assure that things work the same API-wise, FIPS vs non-FIPS, modulo the reduced set of available crypto algorithms.

I could be convinced otherwise, but currently I find it more intuitive to have side-by-side FIPS and non-FIPS project files grouped under the same IDE.

JacobBarthelmeh:
Reviewed these changes. My thinking is more aligned with @douzzer on this. I also think that in most IDE projects there should not be enough of a difference between a FIPS and non-FIPS build to merit a new IDE project. Just build configuration options, possibly linker file changes and handling a couple of FIPS API calls on initialization.

gojimmypi:
After feedback, I'm closing this unmerged. We can revisit later if there are compelling reasons for separation of IDE project files.

PR #8129, opened 2024-10-30
Curve25519 generic keyparsing

bigbrett:
Adds API for curve25519 `der` import/export that can handle both private and public keys in one step. I realize that typically curve25519 keys are distributed as public/private only, since you can recover the public key from the private key, but this API simplifies usage for some cases in that it doesn't require multiple import attempts if you don't know at runtime what the contents of a `der` key holds. 

This API is particularly useful for wolfHSM, where the server doesn't know what the contents of a `DER` key contain when it loads/stores from NVM, but needs to import the key into a `curve25519_key` struct that then will have the proper public/private key fields set. The current API doesn't provide a straightforward way to do this.

Added doxygen, and will update the manual/docs once this merges

# Testing

- Added test to ensure new API can handle private only/public only keys in the same exact way as the existing API
- Added test to ensure a combined keypair import/export yields the same exported der as public/private only

# Checklist

 - [x] added tests
 - [x] updated/added doxygen

### DIFF
--- a/doc/dox_comments/header_files/asn_public.h
+++ b/doc/dox_comments/header_files/asn_public.h
@@ -1557,6 +1557,137 @@ int wc_EccPublicKeyToDer(ecc_key* key, byte* output,
 int wc_EccPublicKeyToDer_ex(ecc_key* key, byte* output,
                                      word32 inLen, int with_AlgCurve, int comp);
 
+
+/*!
+    \ingroup ASN
+
+    \brief This function decodes a Curve25519 private key (only) from a DER
+    encoded buffer
+
+    \return 0 Success
+    \return BAD_FUNC_ARG Returns if input, inOutIdx or key is null
+    \return ASN_PARSE_E Returns if there is an error parsing the DER encoded
+    data
+    \return ECC_BAD_ARG_E Returns if the key length is not CURVE25519_KEYSIZE or
+    the DER key contains other issues despite being properly formatted.
+    \return BUFFER_E Returns if the input buffer is too small to contain a
+    valid DER encoded key.
+
+    \param input Pointer to buffer containing DER encoded private key
+    \param inOutIdx Index to start reading input buffer from.  On output,
+    index is set to last position parsed of input buffer.
+    \param key Pointer to curve25519_key structure to store decoded key
+    \param inSz Size of input DER buffer
+*/
+int wc_Curve25519PrivateKeyDecode(const byte* input, word32* inOutIdx,
+                                  curve25519_key* key, word32 inSz);
+
+/*!
+    \ingroup ASN
+
+    \brief This function decodes a Curve25519 public key (only) from a DER
+    encoded buffer.
+
+    \return 0 Success
+    \return BAD_FUNC_ARG Returns if input, inOutIdx or key is null
+    \return ASN_PARSE_E Returns if there is an error parsing the DER encoded
+    data
+    \return ECC_BAD_ARG_E Returns if the key length is not CURVE25519_KEYSIZE or
+    the DER key contains other issues despite being properly formatted.
+    \return BUFFER_E Returns if the input buffer is too small to contain a
+    valid DER encoded key.
+
+    \param input Pointer to buffer containing DER encoded public key
+    \param inOutIdx Index to start reading input buffer from.  On output,
+    index is set to last position parsed of input buffer.
+    \param key Pointer to curve25519_key structure to store decoded key
+    \param inSz Size of input DER buffer
+*/
+int wc_Curve25519PublicKeyDecode(const byte* input, word32* inOutIdx,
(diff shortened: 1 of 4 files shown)


dgarske (approved):
Over to @SparkiDev

wolfcrypt/src/asn.c
```diff
@@ -35661,6 +35661,55 @@ int wc_Curve25519PublicKeyDecode(const byte* input, word32* inOutIdx,
...
+
+/* Decode Curve25519 key from DER format - can handle private only,
+ * public only, or private+public key pairs.
+ * return 0 on success, negative on error */
+int wc_Curve25519KeyDecode(const byte* input, word32* inOutIdx,
+                          curve25519_key* key, word32 inSz)
+{
+    int ret;
+    byte privKey[CURVE25519_KEYSIZE];
+    byte pubKey[CURVE25519_KEYSIZE];
```
SparkiDev, line 35673: CURVE25519_PUB_KEY_SIZE
(same value but more correct name)

wolfcrypt/src/asn.c
```diff
@@ -35868,6 +35917,63 @@ int wc_Curve25519PublicKeyToDer(curve25519_key* key, byte* output, word32 inLen,
...
 }
+
+/* Export Curve25519 key to DER format - handles private only, public only,
+ * or private+public key pairs based on what's set in the key structure.
+ * Returns length written on success, negative on error */
+int wc_Curve25519KeyToDer(curve25519_key* key, byte* output, word32 inLen, int withAlg)
+{
+    int ret;
+    byte privKey[CURVE25519_KEYSIZE];
+    byte pubKey[CURVE25519_KEYSIZE];
```
SparkiDev, line 35928: CURVE25519_PUB_KEY_SIZE

wolfcrypt/test/test.c
```diff
@@ -35042,6 +35042,163 @@ static wc_test_ret_t curve255519_der_test(void)
...
+        else {
+            ret = WC_TEST_RET_ENC_EC(ret);
+        }
+    }
+    if (ret == 0 && (outputSz != (word32)sizeof(kCurve25519PubDer) ||
+                    XMEMCMP(output, kCurve25519PubDer, outputSz) != 0)) {
+        ret = WC_TEST_RET_ENC_NC;
+    }
+
+    /* Test decode/encode of a key file containing both public and private
```
SparkiDev, line 35101: key file -> key data?

SparkiDev (changes requested):

doc/dox_comments/header_files/asn_public.h
```diff
@@ -1557,6 +1557,137 @@ int wc_EccPublicKeyToDer(ecc_key* key, byte* output,
...
+    the DER key contains other issues despite being properly formatted.
+    \return BUFFER_E Returns if the input buffer is too small to contain a
+    valid DER encoded key.
+
+    \param input Pointer to buffer containing DER encoded private key
+    \param inOutIdx Index to start reading input buffer from.  On output,
+    index is set to last position parsed of input buffer.
+    \param key Pointer to curve25519_key structure to store decoded key
+    \param inSz Size of input DER buffer
+*/
```
SparkiDev, line 1581: Add \sa comments for see also.

doc/dox_comments/header_files/asn_public.h
```diff
@@ -1557,6 +1557,137 @@ int wc_EccPublicKeyToDer(ecc_key* key, byte* output,
...
+int wc_Curve25519PrivateKeyDecode(const byte* input, word32* inOutIdx,
```
SparkiDev, line 1582: Missing examples too.

SparkiDev (dismissed):

wolfcrypt/src/asn.c
bigbrett: fixed

wolfcrypt/src/asn.c
bigbrett: fixed

wolfcrypt/test/test.c
bigbrett: fixed

doc/dox_comments/header_files/asn_public.h
bigbrett: done

doc/dox_comments/header_files/asn_public.h
bigbrett: added examples

bigbrett:
@dgarske @SparkiDev addressed all feedback

dgarske (approved):

SparkiDev (approved):
Great work by the way.

bigbrett:
Thanks @SparkiDev!

PR #8136, opened 2024-10-31
Fix for setting wrong version in CSRs.

anhu:
Fixes ZD 18873

# Testing
```
$ cd wolfssl
$ ./configure --enable-wolfclu
$ make all check
$ sudo make install
$cd ../wolfCLU
$ ./configure (all tried with CFLAG=-DNO_WOLFSSL_REQ_PRINT) 
$ make all
$ ./wolfssl -genkey rsa -size 2048 -out temp.key -outform pem  -output KEYPAIR
$ ./wolfssl req -new -days 1095 -config csr.conf -key temp.key.priv -out service.csr  -outform PEM -sha256
$ openssl req -in service.csr -text
        Version: 1 (0x0)
$ ./wolfssl req -in service.csr  -noout -text
        Version: 1 (0x0)
```

### DIFF
--- a/src/x509.c
+++ b/src/x509.c
@@ -7070,7 +7070,7 @@ int wolfSSL_X509_REQ_print(WOLFSSL_BIO* bio, WOLFSSL_X509* x509)
     /* print version of cert.  Note that we increment by 1 because for REQs,
      * the value stored in x509->version is the actual value of the field; not
      * the version. */
-    if (X509PrintVersion(bio, wolfSSL_X509_REQ_get_version(x509) + 1, 8)
+    if (X509PrintVersion(bio, wolfSSL_X509_REQ_get_version(x509) + (long)1, 8)
             != WOLFSSL_SUCCESS) {
         return WOLFSSL_FAILURE;
     }
--- a/src/x509.c
+++ b/src/x509.c
@@ -7067,8 +7067,10 @@ int wolfSSL_X509_REQ_print(WOLFSSL_BIO* bio, WOLFSSL_X509* x509)
             return WOLFSSL_FAILURE;
     }
 
-    /* print version of cert */
-    if (X509PrintVersion(bio, wolfSSL_X509_version(x509), 8)
+    /* print version of cert.  Note that we increment by 1 because for REQs,
+     * the value stored in x509->version is the actual value of the field; not
+     * the version. */
+    if (X509PrintVersion(bio, wolfSSL_X509_REQ_get_version(x509) + 1, 8)
             != WOLFSSL_SUCCESS) {
         return WOLFSSL_FAILURE;
     }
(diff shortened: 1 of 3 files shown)


src/x509.c
```diff
@@ -14840,6 +14842,23 @@ void wolfSSL_X509_REQ_free(WOLFSSL_X509* req)
     wolfSSL_X509_free(req);
 }
 
+int wolfSSL_X509_REQ_set_version(WOLFSSL_X509 *x, long version) {
```
dgarske, line 14845: NIT: Functions use brace on new line.

src/x509.c
```diff
@@ -14840,6 +14842,23 @@ void wolfSSL_X509_REQ_free(WOLFSSL_X509* req)
...
+        return WOLFSSL_FAILURE;
+    }
+    x->version = (int)version;
+    return WOLFSSL_SUCCESS;
+}
+
+long wolfSSL_X509_REQ_get_version(const WOLFSSL_X509 *req) {
+    WOLFSSL_ENTER("wolfSSL_X509_REQ_get_version");
+    if (req == NULL) {
+        return WOLFSSL_FAILURE;
```
dgarske, line 14857: Would prefer return `return 0; /* invalid arg */` here instead of `WOLFSSL_FAILURE = 0`.

dgarske (changes requested):
Looks good otherwise

src/x509.c
```diff
@@ -14840,6 +14842,25 @@ void wolfSSL_X509_REQ_free(WOLFSSL_X509* req)
     wolfSSL_X509_free(req);
 }
 
+int wolfSSL_X509_REQ_set_version(WOLFSSL_X509 *x, long version)
```
dgarske, line 14845: See `wolfSSL_X509_set_version` -> `x509->version = (int) v + 1;`

src/x509.c
```diff
@@ -7067,8 +7067,10 @@ int wolfSSL_X509_REQ_print(WOLFSSL_BIO* bio, WOLFSSL_X509* x509)
             return WOLFSSL_FAILURE;
     }
 
-    /* print version of cert */
-    if (X509PrintVersion(bio, wolfSSL_X509_version(x509), 8)
+    /* print version of cert.  Note that we increment by 1 because for REQs,
+     * the value stored in x509->version is the actual value of the field; not
+     * the version. */
+    if (X509PrintVersion(bio, (int)wolfSSL_X509_REQ_get_version(x509) + 1, 8)
```
dgarske, line 7073: Is this really the right way/place to adjust the version? The `wolfSSL_X509_get_version` and `wolfSSL_X509_set_version` do it differently.

dgarske (changes requested):

bandi13:
retest this please

src/x509.c
anhu: Yes, I think the reason this is correct is because this is specific for REQs; not certificates.

dgarske: Shouldn’t the +1 be inside the get function? Or is this the same way openssl does it?

anhu: If the +1 was in the wolfSSL_X509_REQ_get_version() then it would return the wrong value.  This behaviour matches OpenSSL.

dgarske (approved):

PR #8174, opened 2024-11-11
cmake: new add_option_defs helper, add KEYLOG_EXPORT build param

### LINKED ISSUE #8165: CMake buid scripts don't offer `--enable-keylog-export` like option
### Version

5.7.4

### Description

configure script has `--enable-keylog-export` param which defines `WOLFSSL_SSLKEYLOGFILE`. This feature doesn't seem to be supported by CMake build scripts.

redbaron:
# Description

Add `add_option_defs` helper which cover simple case of setting compile definitions when build option is selected. New helper uses named arguments and adds following QoL improvements:
- optionally infer default values from list of possible values. First possible value becomes default.
- optionally append default value to helper string. Helper string and default values can't go out of sync
- support adding defines if build option is set. It covers many simple cases like new `WOLFSSL_KEYLOG_EXPORT`

Fixes #8165

# Testing

1. Build before and after patch, diffed CMakeCache.txt files to find only new WOLFSSL_KEYLOG_EXPORT function is different.

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/cmake/functions.cmake
+++ b/cmake/functions.cmake
@@ -9,19 +9,27 @@ function(override_cache VAR VAL)
     set_property(CACHE ${VAR} PROPERTY VALUE ${VAL})
 endfunction()
 
-function(add_option NAME HELP_STRING DEFAULT VALUES)
-    if(VALUES STREQUAL "yes;no")
+function(add_option_defs)
+    cmake_parse_arguments(PARSE_ARGV 0 arg "" "NAME;HELP_STRING;DEFAULT" "VALUES;DEFINITIONS")
+    if(NOT DEFINED arg_DEFAULT)
+        # Pick first value from list of values as default
+        list(GET arg_VALUES 0 arg_DEFAULT)
+    endif()
+    if(NOT arg_HELP_STRING MATCHES "default:")
+        string(APPEND arg_HELP_STRING ". (default: " "${arg_DEFAULT}" ")")
+    endif()
+    if(arg_VALUES STREQUAL "yes;no" OR arg_VALUES STREQUAL "no;yes")
         # Set the default value for the option.
-        set(${NAME} ${DEFAULT} CACHE BOOL ${HELP_STRING})
+        set(${arg_NAME} ${arg_DEFAULT} CACHE BOOL ${arg_HELP_STRING})
     else()
         # Set the default value for the option.
-        set(${NAME} ${DEFAULT} CACHE STRING ${HELP_STRING})
+        set(${arg_NAME} ${arg_DEFAULT} CACHE STRING ${arg_HELP_STRING})
         # Set the list of allowed values for the option.
-        set_property(CACHE ${NAME} PROPERTY STRINGS ${VALUES})
+        set_property(CACHE ${arg_NAME} PROPERTY STRINGS ${arg_VALUES})
     endif()
 
-    if(DEFINED ${NAME})
-        list(FIND VALUES ${${NAME}} IDX)
+    if(DEFINED ${arg_NAME})
+        list(FIND arg_VALUES ${${arg_NAME}} IDX)
         #
         # If the given value isn't in the list of allowed values for the option,
         # reduce it to yes/no according to CMake's "if" logic:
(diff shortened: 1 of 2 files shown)


cmake/functions.cmake
```diff
@@ -9,19 +9,27 @@ function(override_cache VAR VAL)
...
-        set(${NAME} ${DEFAULT} CACHE STRING ${HELP_STRING})
+        set(${arg_NAME} ${arg_DEFAULT} CACHE STRING ${arg_HELP_STRING})
         # Set the list of allowed values for the option.
-        set_property(CACHE ${NAME} PROPERTY STRINGS ${VALUES})
+        set_property(CACHE ${arg_NAME} PROPERTY STRINGS ${arg_VALUES})
     endif()
 
-    if(DEFINED ${NAME})
-        list(FIND VALUES ${${NAME}} IDX)
+    if(DEFINED ${arg_NAME})
```
redbaron, line 31: I don't understand purpose this if() in original function. I transformed it mechanically to new args so it should work as before.

when building there is no difference in CMakeCache.txt before and after, so this branch either never triggered to mechanical transformation works as before.

redbaron: OK, I think I get it. It triggers when build options is passed via command line with value which is not among the list of supported values, like we pass "ON" for values like "yes|no" This works with booleans, but doesn't it break build options where string input is expected?

bigbrett: @redbaron You are correct, this seems like it will erroneously mess up string input arguments. What are your thoughts on adding something like this? 

```cmake

if(DEFINED ${NAME})
    list(FIND VALUES ${${NAME}} IDX)

    # Apply override only if VALUES is yes/no
    if (VALUES STREQUAL "yes;no" OR VALUES STREQUAL "no;yes")
        if (${IDX} EQUAL -1)
            if(${${NAME}})
                override_cache(${NAME} "yes")
            else()
                override_cache(${NAME} "no")
            endif()
        endif()
    else()
        # For non-boolean VALUES, reset to default or error on invalid input
        if (${IDX} EQUAL -1)
            message(WARNING "Invalid value for ${NAME}. Resetting to default: ${DEFAULT}")
            set(${NAME} ${DEFAULT} CACHE STRING "${HELP_STRING}")
        endif()
    endif()
endif()

```

This logic effectively:

- Coerces boolean-like inputs (ON, OFF, etc.) only when VALUES is yes;no.
- Resets to the default or warns the user for invalid non-boolean values in other cases, preserving the original intent and behavior for each scenario.

bigbrett: @redbaron sorry accidentally deleted my previous response.

You are correct, this looks like it squashes non-boolean values inappropriately. Could you add a check for this case, to ensure this doesn't happen? Something like this to fall back to defaults maybe?

```cmake
if(DEFINED ${NAME})
    list(FIND VALUES ${${NAME}} IDX)

    # Apply override only if VALUES is yes/no
    if (VALUES STREQUAL "yes;no" OR VALUES STREQUAL "no;yes")
        if (${IDX} EQUAL -1)
            # Interpret CMake truthiness for boolean-like values
            if(${${NAME}})
                override_cache(${NAME} "yes")
            else()
                override_cache(${NAME} "no")
            endif()
        endif()
    else()
        # For non-boolean VALUES, reset to default or error on invalid input
        if (${IDX} EQUAL -1)
            message(WARNING "Invalid value for ${NAME}. Resetting to default: ${DEFAULT}")
            set(${NAME} ${DEFAULT} CACHE STRING "${HELP_STRING}")
        endif()
    endif()
endif()

```

bigbrett (changes requested):

dgarske:
Hi @redbaron , we had some feedback on this PR. Can you update it? Thank you again for these changes. Thanks, David Garske, wolfSSL

redbaron:
I'll circle back to updating wolfssl in our project, this will be a closest time I could have a look. If you get to it quicker than me, feel free to take over this PR

dgarske:
Closing in favor of: https://github.com/wolfSSL/wolfssl/pull/8596

PR #8179, opened 2024-11-13
Support for the MAXQ1065 wolfPKCS11 effort.

anhu:
- in wc_ecc_import_private_key_ex, check to make sure devid is not invalid before calling wc_MAXQ10XX_EccSetKey. This is because we sometimes need to locally sign material.
- in various places, delete a key in case it already exists; ignore error since it might not exist.
- in various places, unlock and then lock the HW mutex around ECDSA_sign() because it needs access to rng
- in wolfSSL_MAXQ10XX_CryptoDevCb allow 1065 to call the crypto callback.
- in wolfSSL_MAXQ10XX_CryptoDevCb don't set the key during signing; use pre-provisioned one instead; DEVICE_KEY_PAIR_OBJ_ID
- add ecc_establish and modifications to key creation to support ECDH
- support AES CBC
- support AES CCM

### DIFF
--- a/wolfcrypt/src/ecc.c
+++ b/wolfcrypt/src/ecc.c
@@ -11273,7 +11273,7 @@ int wc_ecc_import_private_key_ex(const byte* priv, word32 privSz,
 #endif
 
 #ifdef WOLFSSL_MAXQ10XX_CRYPTO
-    if (ret == 0) {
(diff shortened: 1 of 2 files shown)


anhu:
This has been superseded by a newer PR.  Closing.

PR #8182, opened 2024-11-13
Support for building without wolfssl/openssl header files

dgarske:
# Description

Support for building without wolfssl/openssl header files.
* Fix for `TlsSessionCacheGetAndLock` that was not checking the sessionIDSz, so could return a pointer to an invalid session (if 0's). Resolves issue with `test_wolfSSL_CTX_sess_set_remove_cb` test.
* Fix for size increase on X509 small. 
* Fix for CRL test with NO_RSA.
* Fix for building with `HAVE_SECRET_CALLBACK` only.
* Fix cast warning with `HAVE_EX_DATA` in Windows VS.
* Fix openssl_extra without PKCS12.
* Refactor the EX data crypto and session API's to gate on `HAVE_EX_DATA_CRYPTO`.
* Grouped the EX data API's in ssl.h
* Moved API's in ssl.h to separate the compatibility ones from ours.

Fixes ZD 18465

# Testing

Support building with wolfssl/openssl/*.h files removed.

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


dgarske:
Retest this please

src/ssl_sess.c
```diff
@@ -1120,7 +1120,9 @@ static int TlsSessionCacheGetAndLock(const byte *id,
 #else
         s = &sessRow->Sessions[idx];
 #endif
-        if (s && XMEMCMP(s->sessionID, id, ID_LEN) == 0 && s->side == side) {
+        /* match session ID value and length */
+        if (s && s->sessionIDSz == ID_LEN && s->side == side &&
```
dgarske, line 1124: @julek-wolfssl , please review this change. I spent over 8 hours tracking down this long standing bug. Occasionally this would be called with a session id of 0's and return an old releases session and messing up the ex_data. The only solution I could find was to also check `sessionIDSz` since it is set to 0 when the session is released.

JacobBarthelmeh (dismissed):
Nice refactor!

The only thing that I had to bring up is that this slightly increases the code size of the OPENSSL_EXTRA_X509_SMALL build. Adding the API's

```
wolfSSL_CTX_get_ex_data
wolfSSL_CTX_set_ex_data
```

Tested with `./configure --enable-opensslextra=x509small --enable-static --disable-shared && make && nm ./src/.libs/libwolfssl.a`

JacobBarthelmeh (approved):
Thanks David!

tests/api.c
```diff
@@ -71980,15 +71980,12 @@ static int test_wolfSSL_SESSION_expire_downgrade(void)
...
-
-    WOLFSSL_CTX* ctx = NULL;
     callback_functions server_cbf, client_cbf;
 
     XMEMSET(&server_cbf, 0, sizeof(callback_functions));
     XMEMSET(&client_cbf, 0, sizeof(callback_functions));
 
     /* force server side to use TLS 1.2 */
-    server_cbf.ctx = ctx;
     server_cbf.method = wolfTLSv1_2_server_method;
```
julek-wolfssl, line 71990: In `test_wolfSSL_SESSION_expire_downgrade` we are trying to test that we correctly reject an expired session. Now that the `ctx` is being changed we will always reject the session since it doesn't match with the `ctx`. Are you sure we are still testing the same thing?

julek-wolfssl: I just checked and we do hit the line `Invalid session: timed out` so it looks like we are good. I don't remember then why I used a shared context here.

src/ssl_sess.c
julek-wolfssl: I suppose the same improvement can be made to `AddSessionToCache` in the `Session already exists. Overwriting.` section. This is a major oversight probably stemming all the way back from when session ID's were the only way to renegotiate. Thanks for finding this.

PR #8197, opened 2024-11-18
ifdef out all code in isDGramSock() when WOLFSSL_NO_SOCK defined.

anhu:
In that case, simply return false. 

Fixes ZD18962

### DIFF
--- a/src/wolfio.c
+++ b/src/wolfio.c
@@ -619,6 +619,9 @@ static int PeerIsIpv6(const SOCKADDR_S *peer, XSOCKLENT len)
 
 static int isDGramSock(int sfd)
 {
+#ifdef WOLFSSL_NO_SOCK
+    return 0;
+#else
     int type = 0;
     /* optvalue 'type' is of size int */
     XSOCKLENT length = (XSOCKLENT)sizeof(type);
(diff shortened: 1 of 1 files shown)


JacobBarthelmeh (changes requested):
Am not clear on how this case could be hit. Lets dig into the defines used when compiling to make sure we have a complete fix.

JacobBarthelmeh:
The relevant user_settings.h defines are:

```
WOLFSSL_DTLS
WOLFSSL_NO_SOCK
USE_WOLFSSL_IO
```

If allowing WOLFSSL_NO_SOCK and USE_WOLFSSL_IO at the same time then we need to have isDGramSock return correctly.

Previous versions (back to v5.4.0) we did the following in EmbedSendTo and did not have a isDGramSock internal API:

```
    if (getsockopt(sd, SOL_SOCKET, SO_TYPE, &type, &length) == 0 &&             
            type != SOCK_DGRAM) {                                               
        /* Probably a TCP socket. peer and peerSz MUST be NULL and 0 */         
    }  
```

I think a more complete fix is either we make an assumption on the type for XSOCKOPT_TYPE_OPTVAL_TYPE when not already defined. Or if the user is setting up custom socket calls with they define it.

With either fix though we should add a user_settings.h config file to our CI testing that builds this use case if supporting it going forward.

PR #8177, opened 2024-11-12
ML-DSA/Dilithium: obtain security level from DER when decoding

bigbrett:
- Add ability to obtain Dilithium security level (parameters) from a DER  encoded key based on the algorithm type OID
- Add test coverage for decoding keys without level specified

# Description

Similar to #8129, wolfHSM needs to be able to "blindly" decode a DER key without knowing its exact contents. In this case it just knows that an item is an ML-DSA key, and nothing more. This change allows the security level to be set dynamically in the output `MlDsaKey` struct based on the algorithm OID in the DER file. 

# Testing

Added new wolfCrypt test coverage for decoding keys without specifying a level

# Checklist

 - [x] added tests
 - [x] updated/added doxygen
 - [x] updated appropriate READMEs
 - [ ] Updated manual and documentation

PR #8207, opened 2024-11-20
dilithium fixes for FIPS 204 draft mode

bigbrett:
# Description

Follow up to #8177, adding fixes for broken Dilithium nightly tests that use FIPS 204 draft mode. Prevents draft mode from using OID autodetection and updates documentation.

Also add the auto-generated error code trace macros to .gitignore so it doesn't show up as unstaged when generated


bigbrett:
I'm pretty sure the current PRB failures aren't related to this PR, as they all seem to have to do with CRL stuff which I think is temporarily broken?

dgarske:
Retest this please. CRL issues

PR #8210, opened 2024-11-21
/dev/crypto auth error fix/adjustment for benchmark

night1rider:
When configuring wolfSSL to use `/dev/crypto` and running our benchmark tool, GMAC benchmarks will throw the `-265` error code after calling this [line](https://github.com/wolfSSL/wolfssl/blob/42825e82d2ee849526a8c1405f9cd256ccfb5a5d/wolfcrypt/src/port/devcrypto/devcrypto_aes.c#L325).

This is because is due to this [line](https://github.com/cryptodev-linux/cryptodev-linux/blob/277d4574c10bb8e16ab6ab3f38b8e1cb6cd6c678/authenc.c#L708) in `/dev/crypto` driver.

This PR adds better debugging and will set size when running GMAC to that of the system's page size. It also allows for an override for this size by setting the macro `WOLFSSL_AUTHSZ_BENCH`

### DIFF
--- a/wolfcrypt/benchmark/benchmark.c
+++ b/wolfcrypt/benchmark/benchmark.c
@@ -1991,6 +1991,11 @@ static const char* bench_result_words2[][5] = {
 #endif
 
 
+
+#if defined(WOLFSSL_DEVCRYPTO) && defined(WOLFSSL_AUTHSZ_BENCH)
+    #warning Large/Unalligned AuthSz could result in errors with /dev/crypto
+#endif
+
 /* use kB instead of mB for embedded benchmarking */
 #ifdef BENCH_EMBEDDED
     #ifndef BENCH_NTIMES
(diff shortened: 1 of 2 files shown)


dgarske:
Retest this please - Python test

JacobBarthelmeh (approved):
Looks like sysconf is in QNX too (another system which has /dev/crypto) which is good, so seems portable enough for a call wrapped with the debug macro guard. Assign to wolfssl-bot when ready for a final review and potential merge into master branch.

JacobBarthelmeh (changes requested):
Works as expected. Please resolve this warning though:

```
  CC       examples/server/examples_server_server-server.o
wolfcrypt/benchmark/benchmark.c: In function ‘bench_gmac’:
wolfcrypt/benchmark/benchmark.c:4846:71: warning: format ‘%d’ expects argument of type ‘int’, but argument 3 has type ‘long int’ [-Wformat=]
         printf("authIn Buffer Size[%d] greater than System Page Size[%d]\n",
                                                                      ~^
                                                                      %ld
                         bench_size, sysconf(_SC_PAGESIZE));
                                     ~~~~~~~~~~~~~~~~~~~~~              
  CC       examples/pem/pem.o
```

PR #8223, opened 2024-11-25
Initial support for STM32MP13 HAL

LinuxJedi:
# Description

This adds support for the STM32MP13 HAL, tested on the STM32MP135F MPU.

Using the HAL this modifies our previous RNG, AES-CBC, AES-GCM, HASH, ECDSA and DES3 ST HAL acceleration to work with the MPU. It also works around bugs found in the AES-GCM code of the HAL.

The HAL does not appear to have support for MD5 HASH at the moment, so this has been given a flag to disable it on this MPU.

# Testing

On an STM32MP135-DK in Engineering Mode. The included additional documentation file outlines how to configure the board to work.

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


wolfssl/wolfcrypt/settings.h
```diff
@@ -2123,6 +2123,12 @@ extern void uITRON4_free(void *p) ;
             #include "stm32u5xx_hal.h"
         #elif defined(WOLFSSL_STM32H5)
             #include "stm32h5xx_hal.h"
+        #elif defined(WOLFSSL_STM32MP13)
+            /* HAL headers error on our ASM files */
+            #ifndef __ASSEMBLER__
```
dgarske, line 2128:
```
[check-source-text] [2 of 7] [wolfssl]
    autogen.sh wolfssl...   real 0m9.678s  user 0m8.255s  sys 0m0.289s
    configure...   real 0m12.585s  user 0m9.502s  sys 0m2.338s
unrecognized macros used:
__ASSEMBLER__
add well-formed but unknown macros to /tmp/workspace/PRB-multi-test-script@2/wolfssl/.wolfssl_known_macro_extras.
    [skipping "codespell" subtest in check-source-text -- codespell is missing or too old]
    check-source-text fail_K
    failed config: '--srcdir' '.' '--disable-jobserver' '--enable-option-checking=fatal' '--enable-all' '--enable-testcert' '--enable-acert' '--enable-dtls13' '--enable-dtls-mtu' '--enable-dtls-frag-ch' '--enable-quic'
```

dgarske (commented):
Changes look great! I'll test them out here too.

You'll need to add `__ASSEMBLER__` to `.wolfssl_known_macro_extras`.

LinuxJedi:
Many thanks, should be fixed now.

wolfcrypt/src/port/st/STM32MP13.md
```diff
@@ -0,0 +1,212 @@
...
+In the configuration tool, enable and activate the following:
+
+```
+CRYP1
+HASH1
+PKA
+RNG1
+```
+
+#### DDR RAM
```
dgarske, line 48: Instead of forcing users to start from the DDR_Init example please setup a new example at http://github.com/wolfSSL/wolfssl-examples-stm32

wolfcrypt/src/port/st/STM32MP13.md
```diff
@@ -0,0 +1,212 @@
...
+  SystemClock_Config();
+
+  /* USER CODE BEGIN SysInit */
+#endif
+  /* USER CODE END SysInit */
+```
+
+### Benchmark
+
+To use the wolfCrypt benchmark, enable and activate the RTC in the Device
```
dgarske, line 163: Consider putting RTC in the list above, so don't have to re-generate again

wolfcrypt/src/port/st/STM32MP13.md
```diff
@@ -0,0 +1,212 @@
...
+If you want ECDSA acceleration, you should also add:
+
+```c
+#define WOLFSSL_STM32_PKA
+#define WOLFSSL_STM32_PKA_V2
+```
+
+### Running
+
+Once you have compiled everything, to run your code you will first need to run
```
dgarske, line 208: Consider manually including the DDR init code required in the example. This README.md does not include steps for adding the calls to wolfCrypt_test / bench. That usually come with `IDE/STM32Cube/wolfssl_example.c`

dgarske (changes requested):
Also I started a thread with ST to get our Cube pack working with the MP1 parts.

wolfcrypt/src/port/st/STM32MP13.md
LinuxJedi: I'll rewrite to make this to make it clearer, but the DDR_Init project is not a starting point. It is a separate project you need to run first as-is before running your wolfSSL project, every time you power the board on. There is no way around this for engineering mode.

The DDR_Init project sets up the clocks and initializes the DDR RAM only. Then a blue LED flashes to say it is done. When you run the wolfSSL project, the ST-Link loads this directly into the DDR RAM and runs it from there. This is why the DDR_Init project is required to run first, otherwise the board will immediately crash :)

wolfcrypt/src/port/st/STM32MP13.md
LinuxJedi: We would only be able to integrate the DDR Init code if we were running from the SD card (I don't know how to do it yet, but it must be possible that way), that requires signed binaries and is not a good place to do development / debugging.

Good point on the calls. I'll see what I can do there.

LinuxJedi:
Updated some of the MD doc items to clarify things.

dgarske (approved):

dgarske (approved):

Issue #8222, opened 2024-11-25
[Bug]: HAProxy ca-file manipulation broken since 17c9e92b 


wlallemand:
### Contact Details

wlallemand@haproxy.com

### Version

v5.7.4

### Description

Hello,

After updating our CI with the latest WolfSSL version, I found out that 2 of our regression tests were broken. v5.7.2 is behaving correctly, but v5.7.4 is broken.

After bisecting this is the first bad commit that introduced the problem:

```
commit 17c9e92b7f780c7e122b2bfb2a27985a3273354a
Author: Colton Willey <colton@wolfssl.com>
Date:   Wed Oct 16 16:34:57 2024 -0700

    Initial rewrite of X509 STORE to replicate openssl behavior
```

The behavior of the X509_STORE  does not seem  to match the one of OpenSSL anymore, we don't have specific Wolfssl code in this part, we are only using the OpenSSL API.

I didn't try to reproduce the problem outside HAProxy yet, but I could try if that helps.

### Reproduction steps

After cloning the master of haproxy and wolfssl in /tmp:

git bisect run sh testhaproxy.sh    

testhaproxy.sh:
#!/bin/sh
./autogen.sh
./configure --enable-quic --enable-haproxy --enable-debug
make
cd /tmp/haproxy/
make -j8 TARGET=linux-glibc USE_OPENSSL_WOLFSSL=1  SSL_LIB=/tmp/wolfssl/src/.libs/ SSL_INC=/tmp/wolfssl/ ADDLIB=-Wl,-rpath,/tmp/wolfssl/src/.libs/
make reg-tests -- --debug reg-tests/ssl/new_del_ssl_cafile.vtc

wlallemand:
[haproxy.wolfssl.debug.txt](https://github.com/user-attachments/files/17904819/haproxy.wolfssl.debug.txt)

Here the output of the reg-test with --enable-debug on wolfssl.

ColtonWilley:
Hello @wlallemand 

I am trying to reproduce using your provided steps but am running into a problem. I tried to google around for a solution but did not find anything, I was wondering if you knew how to resolve these errors so I can reproduce?


```
make reg-tests -- --debug reg-tests/ssl/new_del_ssl_cafile.vtc

########################## Preparing to run tests ##########################
vtest not found in path, please specify VTEST_PROGRAM environment variable
make: *** [Makefile:1218: reg-tests] Error 1
```

wlallemand:
[...]

PR #8226, opened 2024-11-25
Fix wolfSSL_X509_STORE_get0_objects to handle no CA

ColtonWilley:
# Description

Fix wolfSSL_X509_STORE_get0_objects to handle case where no CA has been loaded

Fixes github issue #8222

### DIFF
--- a/src/x509_str.c
+++ b/src/x509_str.c
@@ -1850,6 +1850,13 @@ WOLF_STACK_OF(WOLFSSL_X509_OBJECT)* wolfSSL_X509_STORE_get0_objects(
 #if defined(WOLFSSL_SIGNER_DER_CERT) && !defined(NO_FILESYSTEM)
     cert_stack = wolfSSL_CertManagerGetCerts(store->cm);
     store->numAdded = 0;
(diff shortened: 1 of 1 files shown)


src/x509_str.c
```diff
@@ -1850,6 +1850,13 @@ WOLF_STACK_OF(WOLFSSL_X509_OBJECT)* wolfSSL_X509_STORE_get0_objects(
 #if defined(WOLFSSL_SIGNER_DER_CERT) && !defined(NO_FILESYSTEM)
     cert_stack = wolfSSL_CertManagerGetCerts(store->cm);
```
julek-wolfssl, line 1851: Why would we get `NULL` here? Isn't `NULL` supposed to be an error return? The way I'm reading `wolfSSL_CertManagerGetCerts`, it can return an empty stack but a `NULL` return should signify an error occurred.

julek-wolfssl (changes requested):

src/x509_str.c
ColtonWilley: In wolfssl_cm_get_certs_der() if numCerts is zero then it causes an error. I thought the current fix I made was easier than rewriting the internals of wolfssl_cm_get_certs_der() to allow zero certs.

PR #8231, opened 2024-11-27
Add STM32MP13 HAL support for more SHA types

LinuxJedi:
# Description

This adds STM32 HAL support for:

* SHA384
* SHA512 (with -224 and -256)
* SHA3 (all variants apart from SHAKE)

The partial FIFO block calculations have been adjusted based in the STM32 code to support the larger hash sizes.

This should work with other chips such as the STM32U5xx, but is not enabled for that yet.

# Testing

Tested using the wolfCrypt test suite and benchmark suite on an STM32MP135F-DK.

```
error    test passed!
MEMORY   test passed!
base64   test passed!
asn      test passed!
RANDOM   test passed!
MD5      test passed!
SHA      test passed!
SHA-224  test passed!
SHA-256  test passed!
SHA-384  test passed!
SHA-512  test passed!
SHA-512/224  test passed!
SHA-512/256  test passed!
SHA-3    test passed!
Hash     test passed!
HMAC-MD5 test passed!
HMAC-SHA test passed!
HMAC-SHA224 test passed!
HMAC-SHA256 test passed!
HMAC-SHA384 test passed!
HMAC-SHA512 test passed!
HMAC-SHA3   test passed!
HMAC-KDF    test passed!
PRF         test passed!
TLSv1.3 KDF test passed!
GMAC     test passed!
Chacha   test passed!
POLY1305 test passed!
ChaCha20-Poly1305 AEAD test passed!
DES      test passed!
DES3     test passed!
AES      test passed!
AES192   test passed!
AES256   test passed!
AES-CBC  test passed!
AES-GCM  test passed!
RSA      test passed!
DH       test passed!
ECC      test passed!
ECC buffer test passed!
CURVE25519 test passed!
ED25519  test passed!
logging  test passed!
time     test passed!
mutex    test passed!
Test complete
```

```
SHA-384                     86 MiB took 1.000 seconds,   85.938 MiB/s
SHA-512                     86 MiB took 1.000 seconds,   85.840 MiB/s
SHA-512/224                 86 MiB took 1.000 seconds,   86.035 MiB/s
SHA-512/256                 86 MiB took 1.000 seconds,   85.938 MiB/s
SHA3-224                    84 MiB took 1.000 seconds,   84.473 MiB/s
SHA3-256                    85 MiB took 1.000 seconds,   85.156 MiB/s
SHA3-384                    89 MiB took 1.000 seconds,   89.453 MiB/s
SHA3-512                    92 MiB took 1.000 seconds,   91.699 MiB/s
```

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/wolfcrypt/src/port/st/stm32.c
+++ b/wolfcrypt/src/port/st/stm32.c
@@ -137,6 +137,9 @@ static void wc_Stm32_Hash_SaveContext(STM32_HASH_Context* ctx)
     ctx->HASH_IMR = HASH->IMR;
     ctx->HASH_STR = HASH->STR;
     ctx->HASH_CR  = HASH->CR;
+#ifdef STM32_HASH_SHA3
+    ctx->SHA3CFGR  = HASH->SHA3CFGR;
+#endif
     for (i=0; i<HASH_CR_SIZE; i++) {
         ctx->HASH_CSR[i] = HASH->CSR[i];
     }
@@ -184,6 +187,9 @@ static void wc_Stm32_Hash_RestoreContext(STM32_HASH_Context* ctx, int algo)
         HASH->IMR = ctx->HASH_IMR;
         HASH->STR = ctx->HASH_STR;
         HASH->CR = ctx->HASH_CR;
+#ifdef STM32_HASH_SHA3
+        HASH->SHA3CFGR = ctx->SHA3CFGR;
+#endif
 
         /* Initialize the hash processor */
         HASH->CR |= HASH_CR_INIT;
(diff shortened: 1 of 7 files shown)


LinuxJedi:
SHAKE hardware acceleration is also possible if we need it.

wolfcrypt/src/port/st/stm32.c
```diff
@@ -329,11 +335,11 @@ int wc_Stm32_Hash_Update(STM32_HASH_Context* stmCtx, word32 algo,
     while (len) {
         word32 add;
 
-        /* fill the FIFO plus one additional to flush the block */
-        chunkSz = ((STM32_HASH_FIFO_SIZE + 1) * STM32_HASH_REG_SIZE);
-        /* account for extra bytes in the FIFO (use mask 0x3F to get remain) */
-        chunkSz -= (stmCtx->fifoBytes &
-            ((STM32_HASH_FIFO_SIZE * STM32_HASH_REG_SIZE)-1));
+        chunkSz = blockSize;
```
dgarske, line 338: I'll need to re-run some tests on a few STM32 targets to make sure this doesn't break anything. Some of the hashing hardware FIFO 16 * 32-bit words would not save/restore correctly if the FIFO was filled and not flushed. This logic was here to support that edge case. Its possible you've signified the old logic in a good way, but I'll need to confirm before I can merge.

dgarske (commented):
Changes all look good. I'll need to do some regression testing on other STM32 parts to make sure the hashing changes do not break anything.

wolfcrypt/src/port/st/stm32.c
LinuxJedi: Please do, digging into the HAL, I think I see why you did it and I think what I did equates to the same thing. But if it is wrong we can work towards a better way.

dgarske (approved):
Thank you @LinuxJedi 
Tested successfully on STM32U585, STM32U5A9 and STM32H7.
Also pushed a small issue with wolfssl_examples.c broken from a previous PR.

LinuxJedi:
Thanks for running those tests. I have some STM32Fs here if you want me to run additional tests, but I don't envisage them being different. Good catch on your commit.

dgarske (approved):

dgarske (approved):

dgarske:
@wolfSSL-Bot this PR is ready for merge but it needs an additional reviewer.

JacobBarthelmeh (approved):

PR #8233, opened 2024-11-27
Use proper ref count handling when adding to x509 store

ColtonWilley:
# Description

Use proper ref count handling when adding to x509 store

Fixes github issue #8222

### DIFF
--- a/src/x509_str.c
+++ b/src/x509_str.c
@@ -1408,24 +1408,30 @@ int wolfSSL_X509_STORE_add_cert(WOLFSSL_X509_STORE* store, WOLFSSL_X509* x509)
             result = X509StoreAddCa(store, x509, WOLFSSL_USER_CA);
     #if !defined(WOLFSSL_SIGNER_DER_CERT)
             if (result == WOLFSSL_SUCCESS && store->trusted != NULL) {
-                result = wolfSSL_sk_X509_push(store->trusted, x509);
-                if (result > 0) {
-                    result = WOLFSSL_SUCCESS;
-                }
-                else {
-                    result = WOLFSSL_FATAL_ERROR;
+                result = wolfSSL_X509_up_ref(x509);
+                if (result == WOLFSSL_SUCCESS) {
+                    result = wolfSSL_sk_X509_push(store->trusted, x509);
+                    if (result > 0) {
+                        result = WOLFSSL_SUCCESS;
+                    }
+                    else {
+                        result = WOLFSSL_FATAL_ERROR;
+                    }
                 }
             }
     #endif
         }
         else {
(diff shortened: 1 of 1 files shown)


src/x509_str.c
```diff
@@ -1408,24 +1408,30 @@ int wolfSSL_X509_STORE_add_cert(WOLFSSL_X509_STORE* store, WOLFSSL_X509* x509)
...
-                else {
-                    result = WOLFSSL_FATAL_ERROR;
+                result = wolfSSL_X509_up_ref(x509);
+                if (result == WOLFSSL_SUCCESS) {
+                    result = wolfSSL_sk_X509_push(store->trusted, x509);
+                    if (result > 0) {
+                        result = WOLFSSL_SUCCESS;
+                    }
+                    else {
+                        result = WOLFSSL_FATAL_ERROR;
```
SparkiDev, line 1418: Should we be freeing x509 on failure to push?

ColtonWilley: Possibly, there are still additional issues with the implementation in this PR the automated tests exposed. I am currently working on resolving these new memory leaks.

src/x509_str.c
```diff
@@ -1196,20 +1194,18 @@ void wolfSSL_X509_STORE_free(WOLFSSL_X509_STORE* store)
             }
 #if defined(OPENSSL_EXTRA)
             if (store->certs != NULL) {
-                wolfSSL_sk_X509_free(store->certs);
+                wolfSSL_sk_X509_pop_free(store->certs, wolfSSL_X509_free);
```
julek-wolfssl, line 1197: Second param can be NULL. We should be recognizing the correct function to use internally.

julek-wolfssl (changes requested):

src/x509_str.c
julek-wolfssl: Very good catch @SparkiDev. We should call free to decrement the ref counter.

julek-wolfssl: In both cases.

ColtonWilley: Agreed, a great catch. The changes have been made.

ColtonWilley:
Retest this please.

SparkiDev (approved):

douzzer (approved):
this all LGTM.

A separate issue, which you probably noticed yourself, is that we have a `wolfSSL_X509_up_ref()` but no corresponding `wolfSSL_X509_down_ref()`.  Hence, the `*free()` routines need to use _ad hoc_ `wolfSSL_RefDec(&store->ref, ...)` calls.

Cleanup opportunity for another time.

PR #8252, opened 2024-12-04
wolfSSL_CTX_set_tlsext_use_srtp() should return 1 on failure and 0 up…

anhu:
…on success. Same with wolfSSL_set_tlsext_use_srtp().
Fixes ZD 19026


anhu:
See https://docs.openssl.org/1.1.1/man3/SSL_CTX_set_tlsext_use_srtp/

JacobBarthelmeh (changes requested):
We need a test case for this so a regression does not happen. Looks like simply passing null in for profile_str should trigger the failure? Could you add a call to each function in tests/api.c with a fail case to complete the fix?

anhu:
failed test: [make check (macos-latest, --enable-tls13 --enable-session-ticket --enable-dtls --enable-dtls13 --...]  Bad socket state.

JacobBarthelmeh (approved):

PR #8256, opened 2024-12-05
Fix broken verify on Ada wrapper

### LINKED ISSUE #7461: Ada wrapper incorrectly sets verification method for remote peers
I think there is a bug in the logic for setting the verification method in the Ada wrapper [here](https://github.com/wolfSSL/wolfssl/blob/c3d9fb61a88322eaaa237545e1027e05ac6c6d2f/wrapper/Ada/wolfssl.ads#L103).

The documentation suggests it should be a logical OR operator [here](https://github.com/wolfSSL/wolfssl/blob/master/wrapper/Ada/wolfssl.adb#L174).

However, it performs an AND operator [here](https://github.com/wolfSSL/wolfssl/blob/master/wrapper/Ada/wolfssl.adb#L211).

Unless the documentation is incorrect! But the C examples suggest it is a logical OR [here](https://github.com/wolfSSL/wolfssl-examples/blob/master/tls/server-tls13.c#L220).

LinuxJedi:
# Description

The Ada wrapper had an `&` operator for the verification mode. This effectively caused the verification mode to equal `0`.

The operator has been switched to `or` now, in addition, a getter has been added to the API. This allows for the test I've added to the server code to verify that it is being set correctly.

`OPENSSL_ALL` flag added to Ada so that the verify mode getter function is compiled in.

Fixes #7461

Thanks to @dalybrown for reporting it.

# Testing

Test added to the `tls_server_main` code, it will abort with an error if the mode is not set correctly. Also confirmed by inspecting the parameters given to `wolfSSL_CTX_set_verify()` and `wolfSSL_CTX_get_verify()` when they were called.

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/wrapper/Ada/user_settings.h
+++ b/wrapper/Ada/user_settings.h
@@ -260,9 +260,9 @@ extern "C" {
 
 
 /* Openssl compatibility */
+#define OPENSSL_EXTRA
 #if 0 /* DG Disabled */
     /* Openssl compatibility API's */
-    #define OPENSSL_EXTRA
     #define OPENSSL_ALL
     #define HAVE_OPENSSL_CMD
     #define SSL_TXT_TLSV1_2
--- a/wrapper/Ada/tls_client.adb
+++ b/wrapper/Ada/tls_client.adb
@@ -219,7 +219,7 @@ package body Tls_Client with SPARK_Mode is
       --  Require mutual authentication.
       WolfSSL.Set_Verify
          (Context => Ctx,
-          Mode    => WolfSSL.Verify_Peer & WolfSSL.Verify_Fail_If_No_Peer_Cert);
+          Mode    => WolfSSL.Verify_Peer or WolfSSL.Verify_Fail_If_No_Peer_Cert);
 
       --  Load client certificate into WOLFSSL_CTX.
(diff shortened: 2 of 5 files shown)


dgarske (approved):

dgarske:
Will @joakim-strandberg and @dalybrown please review also? Thank you

wrapper/Ada/user_settings.h
```diff
@@ -260,10 +260,10 @@ extern "C" {
 
 
 /* Openssl compatibility */
+#define OPENSSL_ALL
```
dgarske, line 263: Not a fan of having to enable and pull in 100% of the compatibility layer. Once https://github.com/wolfSSL/wolfssl/pull/8258 is merged just use OPENSSL_EXTRA.
Unless you can find a way to avoid it.

LinuxJedi: I wasn't a fan of it either, but couldn't see another way without removing the guard on that function or completely re-implementing it in the Ada C wrapper. I'll update this with OPENSSL_EXTRA today.

dgarske (approved):

PR #8275, opened 2024-12-11
Aarch64 Poly1305: fix corner case

SparkiDev:
# Description

Don't mask top 26 bits as it may have next bit set as reduction step was only approximate.

Fixes zd#19030

# Testing

./configure '--host=aarch64' 'CC=aarch64-linux-gnu-gcc' 'LDFLAGS=--static' '--enable-armasm' '--enable-poly1305' '--enable-chacha'
while ./examples/client/client -h 142.251.221.67 -p 443 -x -g -j -l TLS13-CHACHA20-POLY1305-SHA256 -v 4 2>&1 | tee l ; do echo . ; done
(IP address is for google.com.au)

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


douzzer (approved):
tested _pro forma_ with `wolfssl-multi-test.sh ... quantum-safe-wolfssl-all-cross-aarch64-armasm-unittest-sanitizer`.

PR #8282, opened 2024-12-12
MacOS: allow SHA-3 instructions to be explicitly not used

SparkiDev:
# Description

Some iPads and iPhones don't support SHA-3 instructions. Allow SHA-3 instructions to explicitly not be used for these devices.

# Testing

On Mac:

./configure --enable-armasm
[WOLFSSL_ARMASM_CRYPTO_SHA3 in Makefile]

./configure --enable-armasm=no-sha3-crypto
[WOLFSSL_ARMASM_CRYPTO_SHA3 not in Makefile]

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2973,7 +2973,7 @@ fi
 
 
 ENABLED_ARMASM_INLINE="no"
-ENABLED_ARMASM_SHA3="no"
+ENABLED_ARMASM_SHA3="unknown"
 ENABLED_ARMASM_CRYPTO_SM4="no"
 # ARM Assembly
 # Both SHA3 and SHA512 instructions available with ARMV8.2-a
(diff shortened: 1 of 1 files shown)


bandi13 (changes requested):
We should actually just test whether the instruction exists with configure.ac using `AC_LANG_PROGRAM([[ ]], [[ asm(""EOR3 v31.16b, v31.16b, v18.16b, v15.16b") ]], [ WOLFSSL_ARMASM_CRYPTO_SHA3=yes ], [ WOLFSSL_ARMASM_CRYPTO_SHA3=no ])` or similar.

SparkiDev:
We could but I prefer the flexibility of being able to explicitly turn it off or on so compiling on different targets works.

douzzer (approved):

PR #8284, opened 2024-12-13
DTLS decryption threaded

SparkiDev:
# Description

Add support for decryption in threads for DTLS.

Fixes zd#18975

# Testing

Thrashing with custom app that has encryption and decryption threads.

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


src/dtls13.c
```diff
@@ -956,8 +956,18 @@ static int Dtls13SendOneFragmentRtx(WOLFSSL* ssl,
...
+            int lockRet = wc_LockMutex(&ssl->dtls13Rtx.mutex);
+            if (lockRet < 0) {
+                return lockRet;
+            }
+        #endif
             Dtls13RtxAddRecord(&ssl->dtls13Rtx, rtxRecord);
+        #ifdef WOLFSSL_RW_THREADED
+            wc_UnLockMutex(&ssl->dtls13Rtx.mutex);
+        #endif
+        }
```
julek-wolfssl, line 970: I still think its better to use a high level lock. That will avoid any other possible races as well as handle other post-hs messages.

src/tls13.c
```diff
@@ -2425,8 +2425,10 @@ static WC_INLINE void WriteSEQTls13(WOLFSSL* ssl, int verifyOrder, byte* out)
  * iv     The derived IV.
  * order  The side on which the message is to be or was sent.
  */
-static WC_INLINE void BuildTls13Nonce(WOLFSSL* ssl, byte* nonce, const byte* iv,
-                                   int order)
+#ifndef WOLFSSL_THREADED_CRYPT
+static WC_INLINE
+#endif
+void BuildTls13Nonce(WOLFSSL* ssl, byte* nonce, const byte* iv, int order)
```
julek-wolfssl, line 2431: Why not just make this never static?

julek-wolfssl (changes requested):
First comments. More tomorrow. This PR would be a good place to add unit testing for threaded crypt and rw.

rizlik:
sorry, force pushed by mistake while reviewing, restored to the original commit

rizlik (changes requested):
This includes the review of commit e4a661ff6e as this PR is strictly related.
These are some high level comments:

1. **Thread Keys Not Updated After Key Update**
   Thread keys are set in `SendData` the first time a thread is used but are never updated after a key update. This desynchronization leads to errors during decryption.
   I created a small unit test here: https://github.com/rizlik/wolfssl/blob/68d5cfbed6f2bd18a88b2ee5654015bb6598519c/tests/api/test_dtls.c#L604 that should reproduce the issue.

2. **Changed Return Value Semantic of `wolfSSL_write`**
   The return value semantics of `wolfSSL_write` changed, returning a size even if no data is sent, also the size returned is not equal to the plaintext size. This may apply to decryption side as well. As the application already needs to do some work to setup the threads, it can be an idea to use a separate API for threaded encryption/decryption.

3. **Unit Test Failures with `WOLFSSL_THREADED_CRYPT`**
   Enabling `WOLFSSL_THREADED_CRYPT` causes multiple unit tests to fail.

4. **Lack of Testing**
   These are two very complex features (RW_THREADED and THREADED_CRYPT), but no tests have been added and the current tests are failing.

5. **Lingering Data Sending from Threads**
   After the application calls wolfSSL_write(), threads run and encrypt the data. How should the application trigger the sending of the data when ready? Should it use a 0-sized write? If so, it's better to document it. Overall, it might be beneficial to add a new API.

6. **Unclear Feature Scope (DTLS vs. TLS 1.3)**
   It's unclear if the feature is only for DTLS or also for TLS 1.3. The code should be modularized better if it's only for DTLS. If it can be used also for TLS 1.3, then the sending and encryption order needs to be maintained AFAIU.

7. **Handling Busy Threads**
   If all threads are busy, the current implementation might loop indefinitely. Suggestion: Return `WANT_WRITE` instead.

8. **Signal Field in `ThreadCrypt` Struct**
   There is a signal field per `ThreadCrypt` struct, but only a global signal can be set. Suggestion: Either drop the field or allow enabling each thread individually.

9. **Return Values of `wolfSSL_Async*` API**
   The `wolfSSL_Async*` API should return `WOLFSSL_SUCCESS` instead of zero.

10. **Lingering Messages in `WOLFSSL_RW_THREADED`**
    There are problems with stale messages if the application is not continuously calling `wolfSSL_read()`/`wolfSSL_write()`. Post-handshake messages may remain stale and not be sent.

PR #8306, opened 2024-12-20
ML-KEM/Kyber: fix kyber_prf() for when no AVX2

SparkiDev:
# Description

When no AVX2 available, kyber_prf() is called to produce more than one SHAKE-256 blocks worth of ouput. Otherwise only one block is needed. Changed function to support an outlen of greater than one block.

Fixes zd#19125

# Testing

./configure --disable-shared --enable-kyber=all,original,ml-kem --enable-intelasm
Disable support for AVX2 in CPU detection.
./wolfcypt/test/testwolfcrypt
KYBER test failed!

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


JacobBarthelmeh (approved):
Tested testwolfcrypt before and after the fix. Thanks Sean!

PR #8320, opened 2024-12-28
src/bio.c: add missing functions for libcurl

mscdex:
# Description

Without this, linking libcurl 8.11.1 (and possibly other versions) fails due to missing symbols which are only included with `OPENSSL_ALL`, whereas libcurl is supposedly buildable with just `OPENSSL_EXTRA`.

# Testing

Linked an application with libcurl 8.11.1 and wolfssl 5.7.4 with this patch.

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/src/bio.c
+++ b/src/bio.c
@@ -3554,17 +3554,7 @@ int wolfSSL_BIO_new_bio_pair(WOLFSSL_BIO **bio1_p, size_t writebuf1,
 
 #endif
 
-#ifdef OPENSSL_ALL
-
-#ifndef NO_WOLFSSL_STUB
-void wolfSSL_BIO_set_init(WOLFSSL_BIO* bio, int init)
-{
-    WOLFSSL_STUB("wolfSSL_BIO_set_init");
-    (void)bio;
-    (void)init;
-}
-#endif /* NO_WOLFSSL_STUB */
-
+#if defined(OPENSSL_EXTRA) || defined(OPENSSL_ALL)
 void wolfSSL_BIO_set_shutdown(WOLFSSL_BIO* bio, int shut)
 {
     WOLFSSL_ENTER("wolfSSL_BIO_set_shutdown");
@@ -3586,6 +3576,18 @@ void wolfSSL_BIO_clear_retry_flags(WOLFSSL_BIO* bio)
         bio->flags &= ~(WOLFSSL_BIO_FLAG_READ | WOLFSSL_BIO_FLAG_WRITE |
                         WOLFSSL_BIO_FLAG_RETRY);
 }
+#endif
+
+#ifdef OPENSSL_ALL
+
+#ifndef NO_WOLFSSL_STUB
+void wolfSSL_BIO_set_init(WOLFSSL_BIO* bio, int init)
+{
+    WOLFSSL_STUB("wolfSSL_BIO_set_init");
+    (void)bio;
+    (void)init;
+}
+#endif /* NO_WOLFSSL_STUB */
 
 int wolfSSL_BIO_should_retry(WOLFSSL_BIO *bio)
 {


embhorn:
Hi @mscdex 

Thanks for using wolfSSL. Are you using configuring wolfSSL with `--enable-curl` ? 

Could you please share the full wolfSSL configuration that lead to this issue?

Kind regards,
@embhorn

mscdex:
> Are you using configuring wolfSSL with `--enable-curl` ?

Previously I was not, but I also checked `configure` to see what it was enabling with `--enable-curl`, and it was only enabling `OPENSSL_EXTRA` as far as OpenSSL compatibility goes. Additionally for whatever reason while `configure` should have been implicitly enabling ALPN support with `--enable-curl`, the linker was still complaining about missing ALPN functionality.

The other reason I probably wouldn't want to use `--enable-curl` is that it seems to enable some features that I would rather not have (both for security and additional code size reasons) and are apparently not used by the components I'm using in libcurl.

> Could you please share the full wolfSSL configuration that lead to this issue?

These are the `configure` arguments I'm using:

```
  --disable-crypttests
  --disable-examples
  --disable-shared
  --enable-static
  --prefix=/tmp/build

  --disable-sslv3
  --enable-aesni
  --enable-alpn
  --enable-altcertchains
  --enable-asm
  --enable-curve25519
  --enable-ed25519
  --enable-ed25519-stream
  --enable-opensslextra
```

FWIW before the patch, with the above configuration I get these linker errors:

```
/usr/bin/ld: deps/curl/curl/lib/.libs/libcurl.a(libcurl_la-wolfssl.o): in function `wolfssl_bio_cf_create':
wolfssl.c:(.text.wolfssl_bio_cf_create+0xe): undefined reference to `wolfSSL_BIO_set_shutdown'
/usr/bin/ld: deps/curl/curl/lib/.libs/libcurl.a(libcurl_la-wolfssl.o): in function `wolfssl_bio_cf_ctrl':
wolfssl.c:(.text.wolfssl_bio_cf_ctrl+0x27): undefined reference to `wolfSSL_BIO_get_shutdown'
/usr/bin/ld: wolfssl.c:(.text.wolfssl_bio_cf_ctrl+0x57): undefined reference to `wolfSSL_BIO_set_shutdown'
/usr/bin/ld: deps/curl/curl/lib/.libs/libcurl.a(libcurl_la-wolfssl.o): in function `wolfssl_bio_cf_in_read':
wolfssl.c:(.text.wolfssl_bio_cf_in_read+0xcc): undefined reference to `wolfSSL_BIO_clear_retry_flags'
/usr/bin/ld: deps/curl/curl/lib/.libs/libcurl.a(libcurl_la-wolfssl.o): in function `wolfssl_bio_cf_out_write':
wolfssl.c:(.text.wolfssl_bio_cf_out_write+0xa9): undefined reference to `wolfSSL_BIO_clear_retry_flags'
```

and also for reference, replacing at least `--enable-alpn` and `--enable-opensslextra` with `--enable-curl` results in these *additional* linker errors:

```
/usr/bin/ld: deps/curl/curl/lib/.libs/libcurl.a(libcurl_la-wolfssl.o): in function `wolfssl_connect_step2':
wolfssl.c:(.text.wolfssl_connect_step2+0x26f): undefined reference to `wolfSSL_ALPN_GetProtocol'
/usr/bin/ld: deps/curl/curl/lib/.libs/libcurl.a(libcurl_la-wolfssl.o): in function `wolfssl_connect_common':
wolfssl.c:(.text.wolfssl_connect_common+0x525): undefined reference to `wolfSSL_UseALPN'
```

embhorn:
Thanks for sharing the wolfSSL config. What options are you using with Curl? There are checks in the curl autoconfig to check for these wolfSSL features.

mscdex:
> What options are you using with Curl?

```
--disable-debug
--disable-docs
--disable-manual
--disable-shared
--enable-optimize
--enable-static
--enable-symbol-hiding
--with-wolfssl=/path/to/wolfssl
--without-bearssl
--without-brotli
--without-fish-functions-dir
--without-gnutls
--without-hyper
--without-libidn2
--without-libpsl
--without-librtmp
--without-libssh
--without-libssh2
--without-mbedtls
--without-msh3
--without-nghttp2
--without-nghttp3
--without-ngtcp2
--without-openssl
--without-quiche
--without-rustls
--without-winidn
--without-wolfssh
--without-zlib
--without-zsh-functions-dir
--without-zstd

--disable-aws
--disable-bindlocal
--disable-cookies
--disable-curldebug
--disable-dateparse
--disable-dict
--disable-doh
--disable-form-api
--disable-gopher
--disable-file
--disable-ftp
--disable-headers-api
--disable-hsts
--disable-http
--disable-http-auth
--disable-imap
--disable-kerberos-auth
--disable-libcurl-option
--disable-ldap
--disable-ldaps
--disable-mime
--disable-mqtt
--disable-netrc
--disable-ntlm
--disable-ntlm-wb
--disable-pop3
--disable-progress-meter
--disable-proxy
--disable-rtsp
--disable-smb
--disable-socketpair
--disable-sspi
--disable-telnet
--disable-tftp
--disable-tls-srp
--disable-websockets
--enable-smtp
```

> Also, are you running `autoreconf -fi` for Curl after building wolfSSL?

No.

embhorn:
With your curl config I get this error:
```
configure: error: select TLS backend(s) or disable TLS with --without-ssl.

Select from these:

  --with-amissl
  --with-bearssl
  --with-gnutls
  --with-mbedtls
  --with-openssl (also works for BoringSSL and LibreSSL)
  --with-rustls
  --with-schannel
  --with-secure-transport
  --with-wolfssl
```

embhorn:
@mscdex 

Please try running the autoconf tool prior to configuring curl
`autoreconf -fi`

mscdex:
@embhorn That made no difference.

However, it seems the problem now was that the `--with-wolfssl=...` during curl's `configure` process was pointing to the wrong directory and `configure` happily and silently ignored that fact. My guess is that it was then silently falling back to a global copy of wolfssl which had the full OpenSSL compatibility, which led to the linker errors because of the mismatched configurations between the global and local copies of wolfssl.

This worked previously (by chance) with an older version of libcurl because it had a much less granular compile-time guard around the code that used these few specific BIO functions.

PR #8324, opened 2024-12-31
ntp 4.2.8p17 additions

julek-wolfssl:
- OBJ_sn2nid: use correct short names
- Add MD4 to EVP layer
- Add ntp 4.2.8p17 to tested versions

### DIFF
--- a/src/ssl.c
+++ b/src/ssl.c
@@ -18124,6 +18124,9 @@ const WOLFSSL_ObjectInfo wolfssl_object_info[] = {
     #ifdef WOLFSSL_MD2
         { WC_NID_md2, MD2h, oidHashType, "MD2", "md2"},
     #endif
+    #ifndef NO_MD4
+        { WC_NID_md4, MD4h, oidHashType, "MD4", "md4"},
+    #endif
     #ifndef NO_MD5
         { WC_NID_md5, MD5h, oidHashType, "MD5", "md5"},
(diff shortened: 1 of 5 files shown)


julek-wolfssl:
Depends on https://github.com/wolfSSL/osp/pull/220

src/ssl.c
```diff
@@ -202,79 +202,20 @@
...
+            return obj_info->nid;
     }
-
-#ifdef HAVE_ECC
-    if (XSTRLEN(sn) > ECC_MAXNAME)
-        return WC_NID_undef;
-
-    /* Nginx uses this OpenSSL string. */
-    if (XSTRCMP(sn, "prime256v1") == 0)
-        sn = "SECP256R1";
```
JacobBarthelmeh, line 256: Are we confident that this and also removing the check on WOLFSSL_*_NAME strings will not cause issues? Would prefer keeping the extra handling of those additional short names if possible.

JacobBarthelmeh (changes requested):

src/ssl.c
julek-wolfssl: These extra names are not OpenSSL compliant. That is why I removed it to simplify the implementation.

dgarske (approved):

JacobBarthelmeh (approved):

PR #8332, opened 2025-01-04
Ada binding: add support  for PSK client callback

mgrojo:
# Description

This adds support in the Ada binding for `wolfSSL_set_psk_client_callback`.

# Testing

Tested with:
`wolfSSL/wolfssl-examples/psk/server-psk.c`
after changing `DTLSv1_3_Client_Method` to `DTLSv1_2_Client_Method` to comply with the server example.

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/wrapper/Ada/tls_client.adb
+++ b/wrapper/Ada/tls_client.adb
@@ -63,17 +63,23 @@ package body Tls_Client with SPARK_Mode is
       Key            : chars_ptr;
       Key_Max_Length : unsigned) return unsigned
    with
-     SPARK_Mode => Off      
+     SPARK_Mode => Off
    is
       use type Interfaces.C.unsigned;
 
       Hint_String : constant String := Interfaces.C.Strings.Value (Hint);
+
+      --  Identity is OpenSSL testing default for openssl s_client, keep same
       Identity_String : constant String := "Client_identity";
+      --  Test key in hex is 0x1a2b3c4d, in decimal 439,041,101
       Key_String : constant String :=
         Character'Val   (26)
         & Character'Val (43)
         & Character'Val (60)
         & Character'Val (77);
+      --  These values are aligned with test values in wolfssl/wolfssl/test.h
+      --  and wolfssl-examples/psk/server-psk.c for testing interoperability.
+
    begin
 
       Ada.Text_IO.Put_Line ("Hint: " & Hint_String);
@@ -199,6 +205,7 @@ package body Tls_Client with SPARK_Mode is
 
       Result : WolfSSL.Subprogram_Result;
       DTLS   : Boolean;
+      PSK    : Boolean;
    begin
       Result := WolfSSL.Initialize;
       if Result /= Success then
@@ -208,13 +215,19 @@ package body Tls_Client with SPARK_Mode is
 
       if Argument_Count < 1
          or Argument_Count > 2
-         or (Argument_Count = 2 and then Argument (2) /= "--dtls")
+         or (Argument_Count = 2 and then
+             Argument (2) /= "--dtls" and then
+             Argument (2) /= "--psk")
       then
-         Put_Line ("usage: tls_client_main <IPv4 address> [--dtls]");
+         Put_Line ("usage: tls_client_main <IPv4 address> [--dtls | --psk]");
          return;
       end if;
 
-      DTLS := (SPARK_Terminal.Argument_Count = 2);
+      DTLS := (SPARK_Terminal.Argument_Count = 2 and then
+                Argument (2) = "--dtls");
+
+      PSK := (SPARK_Terminal.Argument_Count = 2 and then
+                Argument (2) = "--psk");
 
       if DTLS then
          SPARK_Sockets.Create_Datagram_Socket (C);
@@ -276,8 +289,7 @@ package body Tls_Client with SPARK_Mode is
          (Context => Ctx,
           Mode    => WolfSSL.Verify_Peer or WolfSSL.Verify_Fail_If_No_Peer_Cert);
 
-      if Ada.Directories.Exists (CERT_FILE) and then
-         Ada.Directories.Exists (KEY_FILE) then
+      if not PSK then
 
          --  Load client certificate into WOLFSSL_CTX.
          Result := WolfSSL.Use_Certificate_File (Context => Ctx,
@@ -335,10 +347,9 @@ package body Tls_Client with SPARK_Mode is
          return;
       end if;
 
-      if not (Ada.Directories.Exists (CERT_FILE) and then
-              Ada.Directories.Exists (KEY_FILE)) then
+      if PSK then
 
-         -- Use PSK for authentication.
+         --  Use PSK for authentication.
          WolfSSL.Set_PSK_Client_Callback
            (Ssl      => Ssl,
             Callback => PSK_Client_Callback'Access);
--- a/wrapper/Ada/tls_client.adb
+++ b/wrapper/Ada/tls_client.adb
@@ -168,9 +168,9 @@ package body Tls_Client with SPARK_Mode is
 
    Any_Inet_Addr : Inet_Addr_Type renames SPARK_Sockets.Any_Inet_Addr;
 
-   CERT_FILE : constant String := "../../../certs/client-cert.pem";
-   KEY_FILE  : constant String := "../../../certs/client-key.pem";
-   CA_FILE   : constant String := "../../../certs/ca-cert.pem";
+   CERT_FILE : constant String := "../../certs/client-cert.pem";
+   KEY_FILE  : constant String := "../../certs/client-key.pem";
+   CA_FILE   : constant String := "../../certs/ca-cert.pem";
 
    subtype Byte_Array is WolfSSL.Byte_Array;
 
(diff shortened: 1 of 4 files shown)


dgarske:
Hi @mgrojo ,

Thank you for your code contribution. In order to consider this we will need to have an agreement on file. Can you please email support at wolfssl dot com and reference this PR requesting a contributor agreement. Please include information about your location and current project.

Thanks,
David Garske, wolfSSL

dgarske:
Hi @mgrojo ,

Your contributor agreement has been approved. We will review these changes and provide feedback shortly.

Okay to test

Thanks,
David Garske, wolfSSL

dgarske:
Tagging some of our Ada contributors for awareness : @joakim-strandberg @dalybrown @Irvise

wrapper/Ada/tls_client.adb
```diff
@@ -40,8 +41,62 @@ package body Tls_Client with SPARK_Mode is
...
+      Id_Max_Length  : unsigned;
+      Key            : chars_ptr;
+      Key_Max_Length : unsigned) return unsigned
+   with
+     SPARK_Mode => Off      
+   is
+      use type Interfaces.C.unsigned;
+
+      Hint_String : constant String := Interfaces.C.Strings.Value (Hint);
+      Identity_String : constant String := "Client_identity";
```
dgarske, line 71: Might be useful to add a code comment...
`/* identity is OpenSSL testing default for openssl s_client, keep same */`

wrapper/Ada/tls_client.adb
```diff
@@ -40,8 +41,62 @@ package body Tls_Client with SPARK_Mode is
...
+      Key_String : constant String :=
```
dgarske, line 72: This is just a sample test key from https://github.com/wolfSSL/wolfssl/blob/master/wolfssl/test.h#L1909
`/* test key in hex is 0x1a2b3c4d , in decimal 439,041,101 , we're using unsigned binary */`

wrapper/Ada/tls_client.adb
```diff
@@ -40,8 +41,62 @@ package body Tls_Client with SPARK_Mode is
...
+   function PSK_Client_Callback
+     (Unused         : WolfSSL.WolfSSL_Type;
+      Hint           : chars_ptr;
+      Identity       : chars_ptr;
+      Id_Max_Length  : unsigned;
+      Key            : chars_ptr;
+      Key_Max_Length : unsigned) return unsigned
+   with Convention => C;
+
+   function PSK_Client_Callback
```
dgarske, line 58: Can you confirm this is in the tls_client example only and not part of the wolfSSL ADA wrapper? The PSK callback is meant to be customized for each customer.

wrapper/Ada/tls_client.adb
```diff
@@ -276,6 +335,15 @@ package body Tls_Client with SPARK_Mode is
          return;
       end if;
 
+      if not (Ada.Directories.Exists (CERT_FILE) and then
```
dgarske, line 338: Is this the best way to gate using PSK vs loading key/cert? A build option or run-time argument seems like it would be ideal (if possible). 

Note: It is possible to set the PSK callback regardless, assuming NO_PSK is not set when building wolfSSL.

wrapper/Ada/wolfssl.ads
```diff
@@ -297,6 +300,31 @@ package WolfSSL with SPARK_Mode is
    --  This function wraps the corresponding WolfSSL C function to allow
    --  clients to use Ada socket types when implementing a DTLS client.
 
+   type PSK_Client_Callback is access function
```
dgarske, line 303: Does the documentation need updated to indicate how to enable PSK for users who want to test it?

dgarske (changes requested):

wrapper/Ada/tls_client.adb
mgrojo: I chose these value because I took them from
https://github.com/wolfSSL/wolfssl-examples/blob/63893d5af9caed265c88708954d3dba11047c768/psk/server-psk.c#L51

I gather from your comment that these values ultimately come from "OpenSSL testing default for openssl s_client" and you suggest me to make that apparent in the comments. Is that right?

wrapper/Ada/tls_client.adb
mgrojo: As said, my inspiration (and test counterpart) was https://github.com/wolfSSL/wolfssl-examples/blob/master/psk/server-psk.c

I can change the values to hexadecimal as in `test.h` if that is what you are suggesting.

dgarske: You don't need to change any code. Please add inline code comment referencing why these test values were chosen.

wrapper/Ada/wolfssl.ads
mgrojo: For me, PSK was working without changes. 

As I see, `NO_PSK` is not defined in this file, which sets up the Ada build (that line is disabled by a previous `#ifdef`):

https://github.com/wolfSSL/wolfssl/blob/78c4a04cac0647506807023b4d27b388ae4574c3/wrapper/Ada/user_settings.h#L355

Consequently, there's nothing to do to get the Ada library to import the function `wolfSSL_set_psk_client_callback` and to work with PSK.

On the other hand, I had to define `WOLFSSL_STATIC_PSK` in the build for my particular use case. I didn't find a way to do that without changing the `user_settings.h` file in the repository (which doesn't seem appropriate). Ultimately, I added an extension of the `wolfssl.gpr` file in my own repository to add the `-DWOLFSSL_STATIC_PSK` compilation switch:

```
-- Extend the wolfssl.gpr project file to customize the build for the CoAP_SPARK
-- library.
project WolfSSL_CoAP extends "libs/wolfssl/wrapper/Ada/wolfssl.gpr" is

   for Object_Dir use "obj/wolfssl_coap";

   -- TODO: find a way to not repeat the source dirs from wolfssl.gpr
   for Source_Dirs use ("libs/wolfssl/wrapper/Ada",
                        "libs/wolfssl/",
                        "libs/wolfssl/src",
                        "libs/wolfssl/wolfcrypt/src");

   for Library_Dir use "lib/wolfssl_coap";

   C_Compiler_Switches :=
      WolfSSL.Compiler'Switches ("C") & (
         "-DWOLFSSL_STATIC_PSK" -- Enable the static PSK cipher support
                                -- (necessary for TRADFRI interoperability)
         );

   package Compiler is
      for Switches ("C") use C_Compiler_Switches;
   end Compiler;

   for Excluded_Source_Files use WolfSSL'Excluded_Source_Files;

end WolfSSL_CoAP;
```
There should be a simpler way to set up the Ada build, but it's not obvious. Maybe the previous Ada contributors can give some insight.

wrapper/Ada/tls_client.adb
mgrojo: Yes, this is only an example for a (D)TLS client. In fact, I think these files should be put in an `examples/` directory and that would make it more obvious, and spare these lines:
https://github.com/wolfSSL/wolfssl/blob/78c4a04cac0647506807023b4d27b388ae4574c3/wrapper/Ada/wolfssl.gpr#L18-L23

mgrojo:
> Retest this please: "FIPSv2-regression" failure.

Where can I see that failure? Nevertheless, it's probably unrelated, since these are only additions. Could it be the same case as I have found here: https://github.com/wolfSSL/wolfssl/pull/5579#issuecomment-1243166692?
> The "pending" on Options-A and FIPSv2 Regression are Jenkins->github integration glitches.

dgarske:
> > Retest this please: "FIPSv2-regression" failure.
> 
> Where can I see that failure? Nevertheless, it's probably unrelated, since these are only additions. Could it be the same case as I have found here: [#5579 (comment)](https://github.com/wolfSSL/wolfssl/pull/5579#issuecomment-1243166692)?
> 
> > The "pending" on Options-A and FIPSv2 Regression are Jenkins->github integration glitches.

That's from our internal Jenkins and not related to your changes. So just ignore for now. 
As far as I am concerned the state of this PR is good. I'm going to have another engineer review and run some tests on it before we merge. I would also appreciate some feedback from some of our third-party ADA contributors I tagged.

mgrojo:
I've also added the server side PSK callbacks on top of this branch. May I add it to this pull request, or should a new one be opened once this one is merged?

mgrojo:
Done.

Irvise:
I am mostly an Ada connoisseur, but by just taking a look at the source code (without checking that the C signatures do indeed match those being imported), it looks fine :) I trust Manu, the compiler and the test to be good enough. Though I would recommend to also wait for at least someone with more experience to voice their point of view.

Best regards,
Fer

dgarske:
Retest this please. Build was aborted.

mgrojo:
> Retest this please. Build was aborted.

I don't understand. The pull request says: "All checks have passed. 151 successful checks".

JacobBarthelmeh:
@mgrojo you're good, the comment from David about restarting tests is directed at our internal testing system. Which it looks like it did re-run and now has the 151 successful checks.

dgarske (approved):

PR #8345, opened 2025-01-09
Python update to 3.12.6

JacobBarthelmeh:
ZD18885


JacobBarthelmeh:
Failing test case for Python 3.8.14 port needs these changes https://github.com/wolfSSL/osp/pull/225

JacobBarthelmeh:
Retest this please Jenkins

wolfcrypt/src/evp.c
```diff
@@ -10723,17 +10733,28 @@ int wolfSSL_EVP_MD_type(const WOLFSSL_EVP_MD* type)
...
                 break;
         #endif
-            case WC_HASH_TYPE_NONE:
-            case WC_HASH_TYPE_MD2:
-            case WC_HASH_TYPE_MD5_SHA:
-            case WC_HASH_TYPE_BLAKE2B:
-            case WC_HASH_TYPE_BLAKE2S:
         #if defined(WOLFSSL_SHA3) && defined(WOLFSSL_SHAKE128)
             case WC_HASH_TYPE_SHAKE128:
+                if (wc_Shake128_Update(&ctx->hash.digest.shake, data,
```
dgarske, line 10738:
```[quantum-safe-wolfssl-all-g++-latest-debug] [6 of 39] [d683165f19]
    configure...   real 0m23.820s  user 0m12.770s  sys 0m13.299s
    build...In file included from src/ssl.c:185:
./wolfcrypt/src/evp.c: In function ‘int wolfSSL_EVP_DigestUpdate(WOLFSSL_EVP_MD_CTX*, const void*, size_t)’:
./wolfcrypt/src/evp.c:10738:65: error: invalid conversion from ‘const void*’ to ‘const byte*’ {aka ‘const unsigned char*’} [-fpermissive]
10738 |                 if (wc_Shake128_Update(&ctx->hash.digest.shake, data,
      |                                                                 ^~~~
      |                                                                 |
      |                                                                 const void*
In file included from ./wolfssl/wolfcrypt/hash.h:47,
                 from ./wolfssl/wolfcrypt/sp_int.h:35,
                 from ./wolfssl/wolfcrypt/wolfmath.h:48,
                 from ./wolfssl/wolfcrypt/dsa.h:33,
                 from ./wolfssl/wolfcrypt/asn_public.h:35,
                 from ./wolfssl/ssl.h:36,
                 from ./wolfssl/internal.h:28,
                 from src/ssl.c:36:
./wolfssl/wolfcrypt/sha3.h:196:65: note:   initializing argument 2 of ‘int wc_Shake128_Update(wc_Shake*, const byte*, word32)’
  196 | WOLFSSL_API int wc_Shake128_Update(wc_Shake* shake, const byte* data, word32 len);
      |                                                     ~~~~~~~~~~~~^~~~
./wolfcrypt/src/evp.c:10747:65: error: invalid conversion from ‘const void*’ to ‘const byte*’ {aka ‘const unsigned char*’} [-fpermissive]
10747 |                 if (wc_Shake256_Update(&ctx->hash.digest.shake, data,
      |                                                                 ^~~~
      |                                                                 |
      |                                                                 const void*
./wolfssl/wolfcrypt/sha3.h:208:65: note:   initializing argument 2 of ‘int wc_Shake256_Update(wc_Shake*, const byte*, word32)’
  208 | WOLFSSL_API int wc_Shake256_Update(wc_Shake* shake, const byte* data, word32 len);
      |                                                     ~~~~~~~~~~~~^~~~
make[2]: *** [Makefile:8282: src/libwolfssl_la-ssl.lo] Error 1
make[2]: *** Waiting for unfinished jobs....
make[1]: *** [Makefile:9077: all-recursive] Error 1
make: *** [Makefile:5427: all] Error 2
   real 0m24.291s  user 1m23.288s  sys 0m9.056s
    quantum-safe-wolfssl-all-g++-latest-debug fail_build```

**Review by dgarske — CHANGES_REQUESTED** [review]
Please split out the codespell stuff to give more time to review.

```overlong lines added:
src/ssl_load.c:2818                 wolfSSL_X509_STORE_add_crl(wolfSSL_CTX_get_cert_store(ctx), &crl);```

**Review by dgarske — DISMISSED** [review]

**Comment by JacobBarthelmeh** [comment]
@dgarske I don't understand the only failing test here with codespell. A grep for `sytem` in wolfssl/ssl.h turned up no hits.

File: wolfcrypt/src/evp.c [hunk 2]
```diff
@@ -10874,6 +10903,43 @@ int wolfSSL_EVP_MD_type(const WOLFSSL_EVP_MD* type)
...
+     * returns 1 (WOLFSSL_SUCCESS) on success and 0 (WOLFSSL_FAILURE) on fail */
+    int wolfSSL_EVP_DigestFinalXOF(WOLFSSL_EVP_MD_CTX *ctx, unsigned char *md,
+        size_t sz)
+    {
+        unsigned int len;
+
+        WOLFSSL_ENTER("wolfSSL_EVP_DigestFinalXOF");
+        len = (unsigned int)sz;
+        return wolfSSL_EVP_DigestFinal(ctx, md, &len);
+    }
```
julek-wolfssl, line 10919: Should `wolfSSL_EVP_DigestFinal` block if an XOF is called outside of `wolfSSL_EVP_DigestFinalXOF` and vice versa?

julek-wolfssl (changes requested):
Review so far. Still need to look at `src` and `tests`.

wolfssl/internal.h
```diff
@@ -6224,6 +6226,7 @@ struct WOLFSSL {
 #if defined(OPENSSL_EXTRA)
     WOLFSSL_STACK* supportedCiphers; /* Used in wolfSSL_get_ciphers_compat */
     WOLFSSL_STACK* peerCertChain;    /* Used in wolfSSL_get_peer_cert_chain */
+    WOLFSSL_STACK* verifiedChain;    /* peer cert chain to CA */
```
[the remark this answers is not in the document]
JacobBarthelmeh: Whether the chain contains a verified chain up to and including the CA (verifiedChain) or just the chain that the peer sent which could possibly not contain the CA also used with verification (peerCertChain). Difference of the return of  `SSL_get0_verified_chain` vs `SSL_get_peer_cert_chain`.

wolfcrypt/src/evp.c
JacobBarthelmeh: Likely yes. Will look at adding a test case for the mix and match of starting with a XOF type of cipher like shake and calling `EVP_DigestFinal` instead of `EVP_DigestFinalXOF`.

wolfcrypt/src/evp.c
```diff
@@ -9936,54 +9936,44 @@ static const struct alias {
...
-        return NULL;
-
-    name = nameUpper;
-    for (al = digest_alias_tbl; al->name != NULL; al++)
+    for (al = digest_alias_tbl; al->name != NULL; al++) {
         if(XSTRCMP(name, al->alias) == 0) {
             name = al->name;
             break;
         }
+    }
```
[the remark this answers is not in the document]
JacobBarthelmeh: Removed it for cases like `WC_SN_blake2s256 "BLAKE2s256" ` Where I think it would fail to match unless we also do XTOUPPER on the SN before comparing. Which I could add in.

wolfcrypt/src/evp.c
julek-wolfssl: I believe that the `s` in `wolfSSL_EVP_DigestFinal_ex` could be uninited. In that case we should definitely require the user to use the correct API so that they are aware of the difference.

src/internal.c
```diff
@@ -7785,6 +7785,11 @@ int InitSSL(WOLFSSL* ssl, WOLFSSL_CTX* ctx, int writeDup)
             return MEMORY_E;
         }
         XMEMSET(ssl->param, 0, sizeof(WOLFSSL_X509_VERIFY_PARAM));
+
+        /* pass on PARAM flags value from ctx to ssl */
+        wolfSSL_X509_VERIFY_PARAM_set_flags(wolfSSL_get0_param(ssl),
+            (unsigned long)wolfSSL_X509_VERIFY_PARAM_get_flags(
+            wolfSSL_CTX_get0_param(ctx)));
```
julek-wolfssl, line 7792: Missing error check.

src/internal.c
```diff
@@ -14995,6 +15011,25 @@ static int ProcessPeerCertsChainCRLCheck(WOLFSSL* ssl, ProcPeerCertArgs* args)
...
+
+    flags = wolfSSL_X509_VERIFY_PARAM_get_flags(param);
+
+    if ((flags & WOLFSSL_CRL_CHECK) == WOLFSSL_CRL_CHECK ||
+        (flags & WOLFSSL_CRL_CHECKALL) == WOLFSSL_CRL_CHECKALL) {
+        ret = wolfSSL_CertManagerEnableCRL(SSL_CM(ssl), flags &
+            (WOLFSSL_CRL_CHECK | WOLFSSL_CRL_CHECKALL));
+    }
+    return ret;
+}
```
julek-wolfssl, line 15030: Does anything need to be done here for `X509_V_FLAG_PARTIAL_CHAIN`, `X509_V_FLAG_USE_CHECK_TIME`, and `X509_V_FLAG_NO_CHECK_TIME`?

src/internal.c
```diff
@@ -15063,6 +15098,11 @@ int ProcessPeerCerts(WOLFSSL* ssl, byte* input, word32* inOutIdx,
     #endif
     }
 
+#ifdef OPENSSL_EXTRA
+    /* account for verify params flag set */
+    AdjustCMForParams(ssl);
+#endif
```
julek-wolfssl, line 15104: Either make `AdjustCMForParams` a `void` return or check the returned error.

src/internal.c
```diff
@@ -37553,7 +37593,7 @@ static int DoSessionTicket(WOLFSSL* ssl, const byte* input, word32* inOutIdx,
     {
         byte            b;
         ProtocolVersion pv;
-#ifdef WOLFSSL_SMALL_STACK
+#if defined(WOLFSSL_SMALL_STACK) || defined(OPENSSL_EXTRA)
         Suites*         clSuites = NULL;
 #else
         Suites          clSuites[1];
```
julek-wolfssl, line 37599: This is getting too confusing for no reason. Let's change it to `Suites* clSuites = NULL` in all cases. I think it would be a good idea to just get rid of the local `clSuites` var and stick with `ssl->clSuites` in all cases. For optimization we can consider free'ing it immediately in this func without `OPENSSL_EXTRA`.

src/ssl_load.c
```diff
@@ -2803,9 +2803,27 @@ int wolfSSL_CTX_load_verify_locations_ex(WOLFSSL_CTX* ctx, const char* file,
...
+            if (ret == 1 && crl.crlList != NULL) {
+                wolfSSL_X509_STORE_add_crl(wolfSSL_CTX_get_cert_store(ctx),
+                    &crl);
+            }
+        #else
+            /* Load the PEM formatted CA file. No CRL support, only expecting
+             * CA's */
             ret = ProcessFile(ctx, file, WOLFSSL_FILETYPE_PEM, CA_TYPE, NULL, 0,
                 NULL, verify);
+        #endif
```
julek-wolfssl, line 2826: I don't see anything in https://docs.openssl.org/master/man3/SSL_CTX_load_verify_locations about needing to support CRL's here.

src/ssl.c
```diff
@@ -3142,6 +3142,23 @@ int wolfSSL_inject(WOLFSSL* ssl, const void* data, int sz)
...
+    ret = wolfSSL_write(ssl, data, sz);
+    if (ret >= 0 && wr != NULL) {
+        *wr = (size_t)ret;
+        ret = 1;
+    }
+    else {
+        ret = 0;
+    }
+    return ret;
+}
```
julek-wolfssl, line 3159: > SSL_write_ex() will return 1 for success or 0 for failure. Success means that all requested application data bytes have been written to the SSL connection or, if SSL_MODE_ENABLE_PARTIAL_WRITE is in use, at least 1 application data byte has been written to the SSL connection. Failure means that not all the requested bytes have been written yet (if SSL_MODE_ENABLE_PARTIAL_WRITE is not in use) or no bytes could be written to the SSL connection (if SSL_MODE_ENABLE_PARTIAL_WRITE is in use).
https://docs.openssl.org/3.0/man3/SSL_write/#warnings

I don't see these partial writes being handled here. This might become a very annoying edge case to debug down the road.

wolfssl/internal.h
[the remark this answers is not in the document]
julek-wolfssl: Oh so its important to differentiate between the two?

src/ssl.c
```diff
@@ -14405,13 +14385,14 @@ static int PushCAx509Chain(WOLFSSL_CERT_MANAGER* cm,
...
 
 /* Builds up and creates a stack of peer certificates for ssl->peerCertChain
-    based off of the ssl session chain. Attempts to place CA certificates
-    at the bottom of the stack. Returns stack of WOLFSSL_X509 certs or
-    NULL on failure */
-WOLF_STACK_OF(WOLFSSL_X509)* wolfSSL_set_peer_cert_chain(WOLFSSL* ssl)
+    or ssl->verifiedChain based off of the ssl session chain. Attempts to place
+    CA certificates at the bottom of the stack for a verified chain. Returns
+    stack of WOLFSSL_X509 certs or NULL on failure */
+static WOLF_STACK_OF(WOLFSSL_X509)* CreatePeerCertChain(WOLFSSL* ssl,
```
julek-wolfssl, line 14394: Can we make `ssl` `const`?

src/ssl.c
```diff
@@ -14456,19 +14435,93 @@ WOLF_STACK_OF(WOLFSSL_X509)* wolfSSL_set_peer_cert_chain(WOLFSSL* ssl)
...
+        WOLFSSL_MSG("wolfSSL_X509_STORE_CTX_init error");
+        wolfSSL_X509_STORE_CTX_free(storeCtx);
+        return NULL;
+    }
+    if (wolfSSL_X509_verify_cert(storeCtx) <= 0) {
+        WOLFSSL_MSG("wolfSSL_X509_verify_cert error");
+        wolfSSL_X509_STORE_CTX_free(storeCtx);
+        return NULL;
+    }
+    wolfSSL_X509_STORE_CTX_free(storeCtx);
```
julek-wolfssl, line 14522: Do you think the verification step is necessary? I don't recall why this was added.

src/tls13.c
```diff
@@ -6682,17 +6682,19 @@ typedef struct Dch13Args {
...
     if (args && args->clSuites) {
         XFREE(args->clSuites, ssl->heap, DYNAMIC_TYPE_SUITES);
         args->clSuites = NULL;
     }
-#ifdef OPENSSL_EXTRA
-    ssl->clSuites = NULL;
 #endif
+    (void)ssl;
+    (void)pArgs;
+
```
julek-wolfssl, line 6697: As I mentioned in another comment. Let's just move this into the `ssl` object. The ownership is becoming a bit of a mess.

src/x509.c
```diff
@@ -481,6 +481,24 @@ int wolfSSL_X509_get_ext_by_OBJ(const WOLFSSL_X509 *x,
...
+}
+
+int wolfSSL_X509_OBJECT_set1_X509_CRL(WOLFSSL_X509_OBJECT *a,
+    WOLFSSL_X509_CRL *obj)
+{
+    WOLFSSL_STUB("wolfSSL_X509_OBJECT_set1_X509_CRL");
+    (void)a;
+    (void)obj;
+    return 1;
+}
```
julek-wolfssl, line 500: Stubs shouldn't return success.

tests/api.c
```diff
@@ -59072,21 +59072,21 @@ static int test_wolfSSL_EVP_Digest_all(void)
...
 #ifndef WOLFSSL_NOSHA3_256
-        "SHA3_256",
+        "SHA3-256",
 #endif
-        "SHA3_384",
+        "SHA3-384",
 #ifndef WOLFSSL_NOSHA3_512
-        "SHA3_512",
+        "SHA3-512",
 #endif
```
julek-wolfssl, line 59090: Any idea why this was testing `_` before? I know its wrong but its just weird that it was passing previously.

tests/api.c
```diff
@@ -61370,6 +61370,43 @@ static int test_wolfSSL_X509_STORE_CTX_get0_store(void)
...
 
+#if defined(OPENSSL_ALL) && !defined(NO_RSA) && !defined(NO_CERTS) && \
+    !defined(NO_WOLFSSL_CLIENT) && !defined(NO_BIO) && !defined(NO_TLS)
+static int test_wolfSSL_get_client_ciphers_on_result(WOLFSSL* ssl) {
+    EXPECT_DECLS;
+    if (wolfSSL_is_server(ssl) == 0) {
+        ExpectNull(SSL_get_client_ciphers(ssl));
+    }
+    else {
+        ExpectNotNull(SSL_get_client_ciphers(ssl));
```
julek-wolfssl, line 61381: It would be nice to iterate through the stack and see if we have some sane expected values for the ciphers.

src/ssl_load.c
JacobBarthelmeh: This one surprised me. It looks like it was handling a PEM file that has a CRL header though. Test case it was done for is here (https://github.com/python/cpython/blob/960936fe90f9a0b947c7ed10f91582ff6664c0b8/Lib/test/test_ssl.py#L2932)

tests/api.c
```diff
@@ -62478,6 +62516,8 @@ static int test_X509_STORE_get0_objects(void)
     ExpectIntEQ(wolfSSL_sk_X509_OBJECT_push(NULL, NULL), WOLFSSL_FAILURE);
     ExpectIntEQ(wolfSSL_sk_X509_OBJECT_push(objs, NULL), WOLFSSL_FAILURE);
     ExpectIntEQ(wolfSSL_sk_X509_OBJECT_push(NULL, obj), WOLFSSL_FAILURE);
+    ExpectNotNull(objsCopy = sk_X509_OBJECT_deep_copy(objs, NULL, NULL));
```
julek-wolfssl, line 62519: Add a test copying a non-empty stack too.

julek-wolfssl (changes requested):
Nice work. Missing tests for `wolfSSL_write_ex` and `wolfSSL_read_ex`.

wolfcrypt/src/evp.c
JacobBarthelmeh: It looks like at least without calling the _ex version of digest final it makes the assumption of 16 bytes for shake128 and 32 bytes for shak256:

Tested with:

```
    mdCtx = EVP_MD_CTX_new();                                                   
    EVP_MD_CTX_init(mdCtx);                                                     
    printf("ret = %d\n", EVP_DigestInit(mdCtx, EVP_shake128()));                
    printf("ret = %d\n", EVP_DigestUpdate(mdCtx, data, 1));                     
    printf("ret = %d\n", EVP_DigestFinal(mdCtx, shake, &sz));                   
    printf("sz = %d\n", sz);   
```

```
$ ./a.out 
ret = 1
ret = 1
ret = 1
sz = 16
result : 9DD7847B4C29404DA0CBAA5D1CB69DF4
```

I can add the test to api.c to make sure we conform to this and check with the _ex version also.

src/ssl_load.c
julek-wolfssl: I'm not convinced this is correct. Please go through OpenSSL to see if this is intentional or just coincidence that it works.

wolfcrypt/src/evp.c
JacobBarthelmeh: *note sz was initialized to a value of 15 which got stomped on with both shake128 and shak256 if not calling the XOF version of digest final

wolfssl/internal.h
```diff
@@ -5780,16 +5780,18 @@ struct WOLFSSL {
     WOLFSSL_CTX*    initial_ctx; /* preserve session key materials */
 #endif
     Suites*         suites; /* Only need during handshake. Can be NULL when
-                             * re-using the context's object. When WOLFSSL
+                             * reusing the context's object. When WOLFSSL
                              * object needs separate instance of suites use
                              * AllocateSuites(). */
 #ifdef OPENSSL_EXTRA
-    const Suites*   clSuites;
+    Suites*   clSuites;
```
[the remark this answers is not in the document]
julek-wolfssl: Ok in that case let's get rid of the local `clSuites` and make it an `ssl` object member always.

tests/api.c
JacobBarthelmeh: Agreed I'll take a look. The testing was light though because it has sanity checks when ran in the updated python 3.12.6 test_ssl.py tests (updated script here https://github.com/wolfSSL/testing/pull/837). Having the cases in our unit test also is useful though, will work on adding some in.

tests/api.c
JacobBarthelmeh: I'm not sure on this one. I think was likely a case where we were handling it wrong but also had mappings for the right strings too. So handling more than needed to. It dates back some time in the history, will trace it down and see if it's still needed.

src/ssl.c
JacobBarthelmeh: I believe so, when in the `wolfSSL_get0_verified_chain` function.

wolfcrypt/src/evp.c
julek-wolfssl: Is that test with OpenSSL? I don't see where we would be setting `sz` in our implementation.

src/ssl_load.c
JacobBarthelmeh: In OpenSSL, function trace:
`SSL_CTX_load_verify_locations` -> `SSL_CTX_load_verify_file` -> `X509_STORE_load_file_ex` -> `X509_LOOKUP_load_file_ex(l, f, X509_FILETYPE_PEM, lib, arg)`

At that point it's treating it as a simple X509_FILETYPE_PEM file being passed in. I think it's something that is a coincidence that it works. Regardless of that though there is a Python test expecting the behavior. I can limit this to the WOLFSSL_PYTHON macro.

julek-wolfssl: It was added here and its definitely an error. https://github.com/python/cpython/commit/225877917e002df4b2d87e965ddd30226aa209ec#diff-d526ded1c360bed6b222de46f4ca92b834f978ebed992fb3189bf9a94a198578R1826

wolfcrypt/src/evp.c
JacobBarthelmeh: Yes, was a small test application created and linked against OpenSSL version 1.1.1.

src/ssl_load.c
JacobBarthelmeh: Fixed by adding a case to the Python patch for it to fall back and try loading it as a CRL if the call to load verify locations fails.

wolfcrypt/src/evp.c
```diff
@@ -10847,23 +10865,84 @@ int wolfSSL_EVP_MD_type(const WOLFSSL_EVP_MD* type)
...
+                          * default for SHAKE256 */
+                break;
         #endif
             default:
                 return WOLFSSL_FAILURE;
         }
-
-        return ret;
+        return wolfSSL_EVP_DigestFinal_Common(ctx, md, s, macType);
     }
```
julek-wolfssl, line 10946: The `sz` arg passed into `wolfSSL_EVP_DigestFinalXOF` is being stomped. I don't think that is correct unless I'm reading something wrong.

src/internal.c
```diff
@@ -7785,6 +7785,15 @@ int InitSSL(WOLFSSL* ssl, WOLFSSL_CTX* ctx, int writeDup)
...
         XMEMSET(ssl->param, 0, sizeof(WOLFSSL_X509_VERIFY_PARAM));
+
+        /* pass on PARAM flags value from ctx to ssl */
+        if (wolfSSL_X509_VERIFY_PARAM_set_flags(wolfSSL_get0_param(ssl),
+            (unsigned long)wolfSSL_X509_VERIFY_PARAM_get_flags(
+            wolfSSL_CTX_get0_param(ctx))) != WOLFSSL_SUCCESS) {
+
+            XFREE(ssl->param, ssl->heap, DYNAMIC_TYPE_OPENSSL);
+            return WOLFSSL_FAILURE;
+        }
```
julek-wolfssl, line 7796: Don't `XFREE`. Just log and let `FreeSSL` take care of `free`'ing anything that got allocated.

src/ssl_load.c
```diff
@@ -2803,7 +2803,6 @@ int wolfSSL_CTX_load_verify_locations_ex(WOLFSSL_CTX* ctx, const char* file,
         }
 
         if (file != NULL) {
-            /* Load the PEM formatted CA file. */
```
julek-wolfssl, line 2806: Reinstate comment.

julek-wolfssl (changes requested):
Almost there. Just a few remaining comments.

wolfcrypt/src/evp.c
JacobBarthelmeh: You're right. The test should have been able to catch it, updated the test coverage and fixed the refactor.

julek-wolfssl (approved):

dgarske (approved):

PR #8351, opened 2025-01-10
Better guarding for LMS SHA256_256 vs LMS SHA256_192

anhu:
Test case: `./configure --enable-lms=sha256-192,no-sha256-256 && make check`

### DIFF
--- a/wolfcrypt/src/wc_lms.c
+++ b/wolfcrypt/src/wc_lms.c
@@ -351,7 +351,9 @@ static const wc_LmsParamsMap wc_lms_map[] = {
       LMS_PARAMS(1, 10, 8, 3, LMS_SHA256_M24_H10, LMOTS_SHA256_N24_W8,
                  WC_SHA256_192_DIGEST_SIZE) },
 #endif
+#endif /* WOLFSSL_LMS_SHA256_192 */
 #if LMS_MAX_HEIGHT >= 20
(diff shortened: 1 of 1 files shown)


anhu:
Fixes a compile error when the following configure time flag is used : `--enable-lms=sha256-192,no-sha256-256`

wolfcrypt/src/wc_lms.c
```diff
@@ -351,7 +351,9 @@ static const wc_LmsParamsMap wc_lms_map[] = {
       LMS_PARAMS(1, 10, 8, 3, LMS_SHA256_M24_H10, LMOTS_SHA256_N24_W8,
                  WC_SHA256_192_DIGEST_SIZE) },
 #endif
+#endif /* WOLFSSL_LMS_SHA256_192 */
 #if LMS_MAX_HEIGHT >= 20
+#ifndef WOLFSSL_NO_LMS_SHA256_256
```
SparkiDev, line 356: Why can't we have both?
You need to use the name explicitly if you want both.

SparkiDev (changes requested):

wolfcrypt/src/wc_lms.c
anhu: Not sure what you mean Sean.  I tried to build with the flags specified above and there were build failures. This fixes that build failure.

dgarske (approved):

PR #8400, opened 2025-01-30
Add support for parsing trusted PEM certs

ColtonWilley:
Add support for parsing PEM certificates with begin trusted cert header/footer, needed for wolfProvider.

Fixes zd 18906


ColtonWilley:
Retest this please.

JacobBarthelmeh (changes requested):
Code change looks good. Needs a test case added though before merging.

JacobBarthelmeh (changes requested):
Test case looks good. One last item, is having new certificates be generated by the appropriate .sh script. In this case likely ./certs/test/gen-testcerts.sh. Not sure if ossl-trusted-cert.pem also needs added to the certs/test/include.am, but please try `make distcheck` to be sure.

ColtonWilley:
Dang, I thought since I am only testing PEM to DER conversion it would be ok to even expire in the future, but you are right the diskcheck fails. I will add it to the generation script.

JacobBarthelmeh (approved):

PR #8409, opened 2025-02-01
Fix misuse of TLSX_KeyShare_Find()

egorpugin:
TLSX_KeyShare_Find() returns 1 on success and condition has incorrect non-inverted logic.

# Description

Here you can see correct condition logic.
https://github.com/wolfSSL/wolfssl/blob/master/src/tls.c#L9235

# Testing

I tried to connect from example client to example server using single TLS13-SM4-GCM-SM3 suite.

### DIFF
--- a/src/tls.c
+++ b/src/tls.c
@@ -9277,7 +9277,7 @@ int TLSX_KeyShare_Parse(WOLFSSL* ssl, const byte* input, word16 length,
             }
 
             /* Check if the group was sent. */
-            if (TLSX_KeyShare_Find(ssl, group)) {
+            if (!TLSX_KeyShare_Find(ssl, group)) {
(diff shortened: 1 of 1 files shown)


embhorn:
Hi @egorpugin 

Thanks for sharing this issue. We do not typically accept small changes, but I can fix this in another PR. Do you intend to create more PR's for wolfSSL? If so, email us at support@wolfssl.com to request a contributor agreement.

Thanks,
@embhorn

dgarske:
@SparkiDev can you please review this bug report from @egorpugin (third party / not approved contributor) and see if this change makes sense?

SparkiDev:
Hi @egorpugin

I've updated the comments for the code in this area to more clearly explain what is going on: #8417
The code here is correct.

Sean

PR #8414, opened 2025-02-03
Fix potential memory leak on error in wolfSSL_EVP_SignFinal()

nielsdos:
# Description

`ecdsaSig` is a local variable pointing to new memory, but the error path in `ret <= 0 || ret > (int)*siglen` does not free this. Fix it by adding a wolfSSL_ECDSA_SIG_free() in this path.

Note: this was found by an experimental static analyser I'm developing, so it's possible that I'm missing some details as I'm not very familiar with the code.

# Testing

Tested by hand by hardcoding the error condition to true.

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen
 - [ ] updated appropriate READMEs
 - [ ] Updated manual and documentation

PR #8415, opened 2025-02-03
Fix potential memory leak on error in LoadCRL()

nielsdos:
# Description

If ProcessFile() fails, the code does not clean up readCtx. Perform the same cleanup as after the loop to fix this.

Note: this was found by an experimental static analyser I'm developing, so it's possible that I'm missing some details as I'm not very familiar with the code.

# Testing

It appears that this if condition is not hit when running `make test`, unless I miss something. So I didn't know how to test this.

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


embhorn:
Hi @nielsdos 

Please let us know the details requested from your previous PR #8414

PR #8485, opened 2025-02-21
fix failing build if not found standard threads support

jmalak:
# Description

Threads::Threads is not defined, if standard support not found for some reason
if custom threads support is used then it happen always
of cause some tests relates to standard threads support then it fails during build, but build is started and only some tests fail or can be suppressed

# Testing

How did you test?

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


embhorn:
@jmalak is an approved contributor

embhorn (approved):

dgarske (approved):

PR #8491, opened 2025-02-23
correct comment for _WINSOCKAPI_ macro manipulation

jmalak:
# Description

add correct comment

The issue is with MINGW winsock2.h header file which is not compatible with Miscrosoft version
and handle **`_WINSOCKAPI_`** macro differently

# Testing

How did you test?

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


embhorn:
@jmalak is an approved contributor

PR #8511, opened 2025-02-26
SHA256: Intel flags has SHA but not AVX1 or AVX2

SparkiDev:
# Description

Reversal of bytes when IS_INTEL_SHA only is same as when AVX1 or AVX2.

Fixes zd#19439

# Testing

Try with a machine that supports Intel SHA instructions but isn't supporting AVX1.

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


SparkiDev:
retest this please

Hudson:

dgarske (approved):

PR #8527, opened 2025-03-04
SP int: inline asm improvements and mont reduce simplifications

### LINKED ISSUE #8523: [Bug]: Invalid inline asm for ARM32
### Contact Details

 aribes96@gmail.com

### Version

5.7.6-stable

### Description

I am compiling for `ARM32` using gcc `4.3.3`

I run configure like this
~~~sh
./configure --host=arm-linux
~~~

At the end of the configure there is this line
~~~sh
---
./configure flags: '--host=arm-linux' 'host_alias=arm-linux' 'CC=arm-unknown-linux-gnueabi-gcc' 'CFLAGS=-march=armv5te' 'CPP=arm-unknown-linux-gnueabi-cpp'
---
~~~

When running `make`, it complains with this

~~~sh
wolfcrypt/src/sp_int.c: In function '_sp_mul_d':
wolfcrypt/src/sp_int.c:6561: error: expected string literal before ')' token
~~~

The relevant line is this
~~~c
        /* Multiply and add into low and high from previous result.
         * No overflow of possible with add. */
        SP_ASM_MUL_ADD_NO(l, h, a->dp[i], d);
~~~

And for my particular build, the macro `SP_ASM_MUL_ADD_NO` uses this expansion

~~~c
/* Multiply va by vb and add double size result into: vh | vl */
#define SP_ASM_MUL_ADD_NO(vl, vh, va, vb)                \
    __asm__ __volatile__ (                               \
        "umlal	%[l], %[h], %[a], %[b]	\n\t"            \
        : [l] "+r" (vl), [h] "+r" (vh)                   \
        : [a] "r" (va), [b] "r" (vb)                     \
        :                                                \
    )
~~~
(this is `wolfcrypt/src/sp_int.c:1161`)

I am not very familiar with inline assembly, but some googling shows that using `cc` for the clobbers would help

~~~c
/* Multiply va by vb and add double size result into: vh | vl */
#define SP_ASM_MUL_ADD_NO(vl, vh, va, vb)                \
    __asm__ __volatile__ (                               \
        "umlal	%[l], %[h], %[a], %[b]	\n\t"            \
        : [l] "+r" (vl), [h] "+r" (vh)                   \
        : [a] "r" (va), [b] "r" (vb)                     \
        : "cc"                                           \
    )
~~~

Indeed, when I add the `cc` the code does compile fine.

In my case, I also had to fix macro `SP_ASM_MUL_SET`

~~~c
/* Multiply va by vb and store double size result in: vo | vh | vl */
#define SP_ASM_MUL_SET(vl, vh, vo, va, vb)               \
    __asm__ __volatile__ (                               \
        "umull	%[l], %[h], %[a], %[b]	\n\t"            \
        "mov	%[o], #0		\n\t"            \
        : [l] "+r" (vl), [h] "+r" (vh), [o] "=r" (vo)    \
        : [a] "r" (va), [b] "r" (vb)                     \
        : "cc"                                           \
    )
~~~
(this is `wolfcrypt/src/sp_int.c:1141`)

I'm not sure if this is a problem only for old versions of gcc, or if the code is indeed wrong. If the problem is in the code, maybe the fix should be applied also to other parts that I don't happen to catch because my settings don't compile it

### Reproduction steps

_No response_

### Relevant log output

```shell

```

SparkiDev:
# Description

SP int inline asm:
- allow input variables to be either registers or memory for Intel x86/x64 (minor performance improvement)
  - don't have memory in clobber list if output variables are registers
- remove empty clobber line in arm32/thumb2 code for old versions of gcc
_sp_mont_red():
  - simplify the code by not using extra variables
  - don't add to j in for loop check.

Fixes #8523

# Testing

Regression tested SP int code.

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/wolfcrypt/src/sp_int.c
+++ b/wolfcrypt/src/sp_int.c
@@ -353,8 +353,8 @@ while (0)
         "movq	%%rax, %[l]	\n\t"                    \
         "movq	%%rdx, %[h]	\n\t"                    \
         : [h] "+r" (vh), [l] "+r" (vl)                   \
-        : [a] "m" (va), [b] "m" (vb)                     \
-        : "memory", "%rax", "%rdx", "cc"                 \
+        : [a] "rm" (va), [b] "rm" (vb)                   \
+        : "%rax", "%rdx", "cc"                           \
     )
 /* Multiply va by vb and store double size result in: vo | vh | vl */
 #define SP_ASM_MUL_SET(vl, vh, vo, va, vb)               \
@@ -377,7 +377,7 @@ while (0)
         "adcq	%%rdx, %[h]	\n\t"                    \
         "adcq	$0   , %[o]	\n\t"                    \
         : [l] "+r" (vl), [h] "+r" (vh), [o] "+r" (vo)    \
-        : [a] "m" (va), [b] "m" (vb)                     \
+        : [a] "rm" (va), [b] "rm" (vb)                   \
         : "%rax", "%rdx", "cc"                           \
     )
 /* Multiply va by vb and add double size result into: vh | vl */
@@ -388,7 +388,7 @@ while (0)
         "addq	%%rax, %[l]	\n\t"                    \
         "adcq	%%rdx, %[h]	\n\t"                    \
         : [l] "+r" (vl), [h] "+r" (vh)                   \
-        : [a] "m" (va), [b] "m" (vb)                     \
+        : [a] "rm" (va), [b] "rm" (vb)                   \
         : "%rax", "%rdx", "cc"                           \
     )
 /* Multiply va by vb and add double size result twice into: vo | vh | vl */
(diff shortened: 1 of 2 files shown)


SparkiDev:
retest this please

Hudson

wolfcrypt/src/sp_int.c
```diff
@@ -3482,7 +3475,7 @@ static WC_INLINE sp_int_digit sp_div_word(sp_int_digit hi, sp_int_digit lo,
         "mulhdu	%[h], %[a], %[b]	\n\t"            \
         : [h] "+r" (vh), [l] "+r" (vl)                   \
         : [a] "r" (va), [b] "r" (vb)                     \
-        : "memory"                                       \
+        :                                                \
```
ribes96, line 3478: Aren't other changes in this same commit precisely removing empty clobbers?

SparkiDev: For other platforms.

dgarske (approved):

PR #7127, opened 2024-01-10
Fix for crypto callbacks to allow invalid devId for non _ex single shot hash functions

dgarske:
# Description

Fix for crypto callbacks to allow invalid devId for non _ex single shot hash functions.

Resolves recursive issue with https://github.com/wolfSSL/wolfssl-examples/pull/413

# Testing

https://github.com/wolfSSL/wolfssl-examples/pull/413

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen
 - [ ] updated appropriate READMEs
 - [ ] Updated manual and documentation

PR #4916, opened 2022-03-02
Add SECO use and expand cryptodev

JacobBarthelmeh:
# Description
- Add ECB crypto callback
- Adds support for using NXP SECO with algorithms (AES CBC,ECB,CCM,GCM,CMAC / ECC P256,P384 sign and verify / SHA256 / SHA384 / TRNG)
- Adds support for import keys with SECO
- Adds support for using an expanded cryptodev-linux with algorithms (Curve25519 / RSA (black keys) / ECC / HMAC)
- Touches up the benchmark app for crypto callback cases
- Adds a WOLFSSL_HASH_KEEP option for cases where "update" with CMAC or Hashing is not possible
- Adds the function wc_CryptoCb_DefaultDevID to get what devId should be defaulted to when not specified

Configure options added:
--enable-devcrypto=seco
--enable-caam=seco
--with-seco=

WOLFSSL_HASH_KEEP
DEBUG_SECO
DEBUG_DEVCRYPTO

# Testing

How did you test? On i.MX8 device

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen
 - [ ] updated appropriate READMEs
 - [ ] Updated manual and documentation

PR #8555, opened 2025-03-12
Add option to disallow automatic use of "default" devId

bigbrett:
# Description

Adds a build option preventing any API that currently uses `wc_CryptoCb_DefaultDevID` to fall back to a "default" devId (which currently is just the first registered global devId) from doing so.

The `wc_CryptoCb_DefaultDevID()` function is currently used by [SHA](https://github.com/wolfSSL/wolfssl/blame/517f4bd561ce1dd1c2404c8c9cc2b06eec6ff97a/wolfcrypt/src/sha256.c#L2226) and [hash.c](https://github.com/wolfSSL/wolfssl/blame/517f4bd561ce1dd1c2404c8c9cc2b06eec6ff97a/wolfcrypt/src/hash.c#L1396) implementations, which causes them to automatically choose a devId even when not requested. I think this behavior is incorrect, as I might not want my cryptoCb devId to be used when I don't explicitly choose to use it, and if one is to be used, it should definitely not just be the somewhat arbitrary choice of the first one in the global list. 

IMO, the correct behavior would be to always use `INVALID_DEVID` unless you explicitly opt in to a fixed platform-wide devID by building with a special platform configuration flag (`WOLFSSL_CAAM_DEVID`, etc.). We should never arbitrarily choose the first devId from the list. However, I realize that certain platforms may be relying on this functionality currently, so I've elected to make the change opt-in to prevent breaking on platforms for which we don't have CI coverage (the only real places this is relevant).

If we are okay with the slightly more intrusive step of defaulting to the new "`INVALID_DEVID` as default" behavior but provide a build macro to keep the current behavior, I think that would be even better, so let me know and I can change the default.

Relevant: #4916 and #7127

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -8791,10 +8791,20 @@ AC_ARG_ENABLE([cryptodev],
 
 # Support for crypto callbacks
 AC_ARG_ENABLE([cryptocb],
-    [AS_HELP_STRING([--enable-cryptocb],[Enable crypto callbacks (default: disabled)])],
-    [ ENABLED_CRYPTOCB=$enableval ],
-    [ ENABLED_CRYPTOCB=no ]
-    )
+    [AS_HELP_STRING([--enable-cryptocb],
+        [Enable crypto callbacks (default: disabled). Use 'no-default-devid' to enable without a platform-specific default device ID])],
+[
+    case "$enableval" in
+        no-default-devid)
+            ENABLED_CRYPTOCB=yes
+            AM_CPPFLAGS="$AM_CPPFLAGS -DWC_NO_DEFAULT_DEVID"
+            ;;
+        *)
+            ENABLED_CRYPTOCB="$enableval"
+            ;;
+    esac
+],
+[ ENABLED_CRYPTOCB=no ])
 
 # Enable testing of cryptoCb using software crypto. On platforms where wolfCrypt tests
 # are used to test a custom cryptoCb, it may be desired to disable this so wolfCrypt tests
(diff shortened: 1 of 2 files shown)


bigbrett:
Jenkins retest this please

dgarske (changes requested):
What happens with the tests when the default is disabled? Do any tests still pass? Should we add a test? As I mentioned the ASN HashCalc and one shot hash functions are why we originally added this.

bigbrett:
> What happens with the tests when the default is disabled? Do any tests still pass? Should we add a test? As I mentioned the ASN HashCalc and one shot hash functions are why we originally added this.

Tests all pass for me (`make check`) when I disable the default (e.g. fall back to `INVALID_DEVID`). I was thinking of adding a test to multi-test to exercise this option

configure.ac
```diff
@@ -9265,6 +9265,14 @@ then
     AM_CFLAGS="$AM_CFLAGS -DALT_ECC_SIZE"
 fi
 
+AC_ARG_ENABLE([default-devid],
```
JacobBarthelmeh, line 9268: I almost think this should be an option on `--enable-cryptocb` such as `--enable-cryptocb=default` or `--enable-cryptocb=no-default-devid`. Someone first picking up wolfssl and running `./configure -h ` might not know what a devid is or that it is related to cryptocb.

dgarske: I agree. Something like `--enable-cryptocb=no-default-devid`.

bigbrett:
Jenkins retest this please

JacobBarthelmeh (approved):

dgarske (approved):

PR #8566, opened 2025-03-17
Only perform ARM assembly CPUID checks if support was enabled at build time.

kareem-wolfssl:
# Description

Fixes zd#19572


SparkiDev:
retest this please

PR #8601, opened 2025-03-27
Fix for STM32 PKA with P521 and shared secret

dgarske:
# Description

Fix for STM32 PKA with P521 and shared secret. The old code might get 65 length for `szkbin`, which is not modulus size and caused ecc_mulmod to have an invalid result.
ZD 19422

# Testing

Actual STM32U585 hardware with PKA enabled.

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


dgarske:
Retest this please: "Found unhandled org.jenkinsci.plugins.workflow.support.steps.AgentOfflineException exception:" known configs 40

PR #8642, opened 2025-04-07
DTLS: drop records that span datagrams

rizlik:
# Description

DTLS records should not span UDP datagrams, this PR adds an API to enforce it.


doc/dox_comments/header_files/ssl.h
```diff
@@ -3788,6 +3788,26 @@ void  wolfSSL_dtls13_set_send_more_acks(WOLFSSL *ssl, int value);
...
+
+    \return WOLFSSL_SUCCESS returned if the function executes without an error.
+    \return A negative error code returned on failure.
+
+    \param ssl a pointer to a WOLFSSL structure, created using wolfSSL_new().
+    \param allowSpan a flag indicating whether records can span datagrams:
+                     - 0: Records cannot span datagrams.
+                     - 1: Records can span datagrams (default behavior).
+*/
+int wolfSSL_dtls_set_records_can_span_datagrams(WOLFSSL* ssl, int value);
```
julek-wolfssl, line 3809: Since the RFC explicitly prohibits this, I would make this a compile time const. Its not something that users are going to be changing dynamically.

> Each DTLS record MUST fit within a single datagram.

https://www.rfc-editor.org/rfc/rfc9147.html#section-4.3

julek-wolfssl (changes requested):

gojimmypi:
Jenkins retest this please

for long-running job

rizlik:
retest this please

rizlik:
retest this please

kaleb-himes:
retest this please

rizlik:
retest this please

src/internal.c
```diff
@@ -21543,9 +21579,18 @@ static int GetInputData(WOLFSSL *ssl, word32 size)
...
+            if (!DtlsRecordsCanSpanDatagrams(ssl)) {
+                WOLFSSL_MSG("DTLS: Received partial record, ignoring");
+#ifdef WOLFSSL_DTLS_DROP_STATS
+                ssl->replayDropCount++;
+#endif /* WOLFSSL_DTLS_DROP_STATS */
+                continue;
+            }
+        }
+
         ssl->buffers.inputBuffer.length += (word32)in;
```
julek-wolfssl, line 21592: I don't know if dropping and continuing to read is the best option. I think we should check if we are non-blocking. If we are non-blocking then return `WANT_READ`. Also this section needs to be protected with `IsDtlsNotSctpMode`.

src/internal.c
```diff
@@ -21931,6 +22095,16 @@ static int DoProcessReplyEx(WOLFSSL* ssl, int allowSocketErr)
...
                        ssl->buffers.inputBuffer.idx;
                 if (used < readSz) {
+                    if (used > 0 && !DtlsRecordsCanSpanDatagrams(ssl)) {
+                        WOLFSSL_MSG("DTLS: Partial record in buffer, dropping");
+                        ssl->options.processReply = doProcessInit;
+                        ssl->buffers.inputBuffer.length = 0;
+                        ssl->buffers.inputBuffer.idx = 0;
+#ifdef WOLFSSL_DTLS_DROP_STATS
+                        ssl->replayDropCount++;
+#endif /* WOLFSSL_DTLS_DROP_STATS */
```
julek-wolfssl, line 22105: This exact pattern is repeated multiple times in multiple locations. It would be nice to refactor it into either a `void func` or a macro.

src/internal.c
```diff
@@ -215,6 +217,12 @@ static int DoAppleNativeCertValidation(const WOLFSSL_BUFFER_INFO* certs,
                                             int totalCerts);
 #endif /* #if defined(__APPLE__) && defined(WOLFSSL_SYS_CA_CERTS) */
 
+#ifdef WOLFSSL_DTLS
+#ifndef DTLS_RECORDS_CAN_SPAN_DATAGRAMS_DEFAULT
+#define DTLS_RECORDS_CAN_SPAN_DATAGRAMS_DEFAULT 1
+#endif
+#endif
```
julek-wolfssl, line 224: What is this for? I don't see it used anywhere.

julek-wolfssl (changes requested):
Please add `WOLFSSL_DTLS_RECORDS_CAN_SPAN_DATAGRAMS` (with all dtls relevant options) to the gh actions.

I haven't reviewed the tests yet. This PR is the right direction.

src/internal.c
rizlik: I think the end result will be the same as the next invocation of `wolfSSLReceive` will return `WANT_READ` in that case. I would like to avoid returning `WANT_READ` outside of the real networking function.  
`DtlsRecordsCanSpanDatagrams` returns always 1 for the `IsDtlsNotSctpMode`

src/internal.c
rizlik: good catch! removed.

src/internal.c
rizlik: done

rizlik:
retest this please

src/internal.c
julek-wolfssl: `DtlsRecordsCanSpanDatagrams` implies that this is implemented for DTLS only. If you don't want to protect the call with dtls checks I would suggest changing the name to just `RecordsCanSpanMultipleReads` or something simpler.

julek-wolfssl (changes requested):
One comment unresolved and rebase and fix conflict.

src/internal.c
rizlik: renamed and simplified nested ifdef

rizlik:
retest this please

julek-wolfssl (dismissed):

rizlik:
retest this please

julek-wolfssl (approved):

SparkiDev (approved):

PR #8643, opened 2025-04-07
Explicit API redirects for FIPS moving forward

kaleb-himes:
# Description

Explicit remap of _clear to _forcezero for all FIPS uses moving forward. Best security practice.

# Testing

Stressing use cases involving various edge cases with CSTL.

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/wolfcrypt/src/integer.c
+++ b/wolfcrypt/src/integer.c
@@ -177,6 +177,9 @@ int mp_init (mp_int * a)
 /* clear one (frees)  */
 void mp_clear (mp_int * a)
 {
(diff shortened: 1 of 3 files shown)


kaleb-himes:
retest this please, one of the agents went offline mid-test.

dgarske (approved):
Over to @SparkiDev

SparkiDev (approved):

SparkiDev:
DH should be changed to call mp_forcezero().
Since we can't change the boundary ... this will have to do.

PR #8651, opened 2025-04-08
Update to expose reasonable DER buffer sizes for Curve25519

billphipps:
# Description

Following the pattern in ecc.h for ECC_BUFSIZE to expose a reasonable buffer size when exporting a Curve25519 key to DER, an enum was added to allow static allocation of DER buffers: CURVE25519_MAX_KEY_TO_DER_SZ.

# Testing

Updates were made to the curve25519 API tests to use this new constant, rather than the magic number of 128 or the overestimation of ONEK_BUF.

# Checklist

 - [ x ] added tests

### DIFF
--- a/wolfssl/wolfcrypt/curve25519.h
+++ b/wolfssl/wolfcrypt/curve25519.h
@@ -45,6 +45,12 @@
 #define CURVE25519_KEYSIZE 32
 #define CURVE25519_PUB_KEY_SIZE 32
 
+enum {
+    CURVE25519_BUFSIZE = 128,  /* for exported DER keys temp buffer */
+
+    WOLF_ENUM_DUMMY_LAST_ELEMENT(CURVE25519)
+};
+
 #ifdef WOLFSSL_NAMES_STATIC
 typedef char curve25519_str[12];
 #else
--- a/wolfssl/wolfcrypt/curve25519.h
+++ b/wolfssl/wolfcrypt/curve25519.h
@@ -46,7 +46,7 @@
 #define CURVE25519_PUB_KEY_SIZE 32
 
 enum {
-    CURVE25519_BUFSIZE = 128,  /* for exported DER keys temp buffer */
+    CURVE25519_MAX_KEY_TO_DER_SZ = 82,  /* for exported DER keys temp buffer */
 
     WOLF_ENUM_DUMMY_LAST_ELEMENT(CURVE25519)
 };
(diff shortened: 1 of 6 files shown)


wolfssl/wolfcrypt/curve25519.h
```diff
@@ -45,6 +45,12 @@
 #define CURVE25519_KEYSIZE 32
 #define CURVE25519_PUB_KEY_SIZE 32
 
+enum {
+    CURVE25519_BUFSIZE = 128,  /* for exported DER keys temp buffer */
```
SparkiDev, line 49: The DER encoding of the public and private keys are not the same and the private key can be wrapped in PKCS#8.
Is this CURVE25519_MAX_PUB_DER_SZ and CURVE25519_MAX_PRIV_DER_SZ?

SparkiDev (changes requested):

wolfssl/wolfcrypt/curve25519.h
billphipps: In wolfHSM, we needed a reasonable size to allocate exported keys, which may be pub, priv, or pub+priv.  We are looking for the maximum size of encoding output from wc_Curve25519KeyToDer with Alg.  I think its a bit less than 128, but it seemed like we haven't had any issues.

billphipps:
Retest this please Jenkins

wolfssl/wolfcrypt/curve25519.h
SparkiDev: CURVE25519_MAX_KEY_DER_SZ = 82

billphipps: Since this size is strictly what could be exported from wolfCrypt (no PKCS8), perhaps the final name should be CURVE25519_MAX_KEY_TO_DER_SZ?

Not super familiar with DER encoding sizes, but would I be correct in assuming that CURVE448_MAX_KEY_TO_DER_SZ would be 130?  18 byte header + 56 pub + 56 priv

SparkiDev: If the DER is coming out of the Curve25519 APIs then it won't be PKCS#8 encoded.
The maximum length of a private key with public key is 82.

SparkiDev (approved):

SparkiDev:
retest this please

PR #8653, opened 2025-04-08
Make trusted_ca_keys check opt-in. 

kareem-wolfssl:
It is not required according to the RFC.

RFC 6066 section 6:
``` 
Servers that receive a client hello containing the "trusted_ca_keys"
extension MAY use the information contained in the extension to guide
their selection of an appropriate certificate chain to return to the
client.  In this event, the server SHALL include an extension of type
"trusted_ca_keys" in the (extended) server hello.  The
"extension_data" field of this extension SHALL be empty.
```

# Description

Fixes zd#19696

# Testing

Built in tests

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


kareem-wolfssl:
Question for reviewer: should this be included under WOLFSSL_HARDEN_TLS rather than making a new flag WOLFSSL_REQUIRE_TCA?

SparkiDev:
This isn't a harden thing.
The client is sending a list of known CAs.
It doesn't matter if the server uses the list or not.
If it ignores and sends certificates that don't verify, the client will fail the handshake.

SparkiDev (approved):

PR #8675, opened 2025-04-16
Entropy MemUse: fix for when block size less than update bits

SparkiDev:
# Description

When the block size is less than the number of update bits, adding the update value will make the index larger than ENTROPY_NUM_WORDS. The update bits, ENTROPY_NUM_UPDATES_BITS, should be less than or equal to ENTROPY_BLOCK_SZ but is not practical.
Add extra elements to the entropy state to accomadate this.

Call the custom high resolution time function when CUSTOM_ENTROPY_TIMEHIRES is defined with the function name.

Fixes zd#19728, zd#19727

# Testing

./configure --disable-shared --enable-debug --enable-entropy-memuse 'CFLAGS=-DENTROPY_NUM_WORDS_BITS=9 -DWOLFSSL_DEBUG_ENTROPY_MEMUSE' 'CC=clang -fsanitize=address'
./wolfccrypt/test/testwolfcrypt

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


dgarske (dismissed):

SparkiDev:
retest this please

SparkiDev:
retest this please

wolfcrypt/src/random.c
```diff
@@ -1038,8 +1043,14 @@ static void Entropy_StopThread(void)
     #error "ENTROPY_NUM_64BIT_WORDS must be <= SHA3-256 digest size in bytes"
 #endif
 
+#if ENTROPY_BLOCK_SZ < ENTROPY_NUM_UPDATES_BITS
```
dgarske, line 1046: Double check possible divide by zero:

Setting the `ENTROPY_NUM_UPDATES_BITS` could end up with a division by zero:
`CFLAGS="-DENTROPY_NUM_WORDS_BITS=8 -DENTROPY_NUM_UPDATES=8"`

```
wolfcrypt/src/random.c: In function 'Entropy_MemUse':
wolfcrypt/src/random.c:1037:23: error: division by zero [-Werror=div-by-zero]
1037 |     (ENTROPY_BLOCK_SZ / ENTROPY_NUM_UPDATES_BITS)
```

dgarske: Is it possible to document a bit more about how to configure wolfEntropy?

It seems to be controlled by these defines:
* ENTROPY_NUM_UPDATES, default 18
* ENTROPY_NUM_UPDATES_BITS, depends on ENTROPY_NUM_UPDATES:  /* Upper round of log2(ENTROPY_NUM_UPDATES) */
* ENTROPY_NUM_WORDS_BITS, range [8 … 30], default 14
 ```

Any hints/suggestions that we should take into consideration when choosing the values for the above?

dgarske (changes requested):
Feedback from the ticket

wolfcrypt/src/random.c
SparkiDev: Added check for when ENTROPY_NUM_UPDATES_BITS is 0

SparkiDev:
retest this polease

dgarske (approved):

PR #8678, opened 2025-04-16
Fix types.h for MSVC6.0

embhorn:
# Description

MSVC6 does not have the "long long" data type. However, MSVC6 does have built-in support for `__int64` and `unsigned __int64`. Windows.h is not necessary for this.

Fixes zd19742

# Testing

Customer conifirmed

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/wolfssl/wolfcrypt/types.h
+++ b/wolfssl/wolfcrypt/types.h
@@ -252,7 +252,8 @@ decouple library dependencies with standard string, memory and so on.
          #endif
     #endif
 
-    #if (defined(_MSC_VER) && !defined(WOLFSSL_NOT_WINDOWS_API)) || \
+    #if (defined(_MSC_VER) && (_MSC_VER == 1200)) ||  /* MSVC6 */ \
+        (defined(_MSC_VER) && !defined(WOLFSSL_NOT_WINDOWS_API)) || \
            defined(__BCPLUSPLUS__) || \
            (defined(__WATCOMC__) && defined(__WATCOM_INT64__))
         /* windows types */


embhorn:
Retest this please

```
https://cloud.wolfssl-test.com/jenkins/job/PRB-generic-config-parser/8120/
Also:   org.jenkinsci.plugins.workflow.actions.ErrorAction$ErrorId: 02e1fc80-1e96-4810-ac9e-b0f00b974314
org.jenkinsci.plugins.workflow.support.steps.AgentOfflineException: Unable to create live FilePath for wolf-linux-cloud-node-n6ec9n; wolf-linux-cloud-node-n6ec9n was marked offline: Connection was broken
```

wolfssl/wolfcrypt/types.h
```diff
@@ -252,7 +252,8 @@ decouple library dependencies with standard string, memory and so on.
          #endif
     #endif
 
-    #if (defined(_MSC_VER) && !defined(WOLFSSL_NOT_WINDOWS_API)) || \
+    #if (defined(_MSC_VER) && (_MSC_VER == 1200)) ||  /* MSVC6 */ \
+        (defined(_MSC_VER) && !defined(WOLFSSL_NOT_WINDOWS_API)) || \
```
dgarske, line 256: Is the customer reporting this defining `WOLFSSL_NOT_WINDOWS_API`? If they are building in MSVC 6.0 shouldn't that not be defined and shouldn't the code use this code path?

dgarske (changes requested):

wolfssl/wolfcrypt/types.h
embhorn: They are defining `WOLFSSL_NOT_WINDOWS_API`, which they need because windows.h is not needed. However, MSVC 6.0 does not have `long long`, but it does support `__int64` and `unsigned __int64`, which are independent of windows.h

So this is specific to the MSVC 6.0 environment.

dgarske (approved):

douzzer (approved):

PR #8692, opened 2025-04-17
Update wolfSSL_X509_verify_cert to retry all certs until a valid chain is found.

kareem-wolfssl:
# Description

Previously wolfSSL_X509_verify_cert was failing out after finding an invalid chain, this commit updates it to remove the invalid cert and retry until a valid chain is found or all certs are exhausted.  This matches OpenSSL's behavior according to the documentation and testing.
This allows wolfSSL_X509_verify_cert to work with multiple CA certs with the same subject key, ie. alt cert chains.

Fixes zd#19563

# Testing

Tested with customer provided test.
Unit tests to be added.  Confirming it passes CI/CD tests first.

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


JacobBarthelmeh:
Looks like it's currently failing unit tests with this configure `--disable-sha --enable-opensslextra`.

JacobBarthelmeh:
Needs a test case added to avoid regression and to excersice the code with the CI `make check`'s.

dgarske (changes requested):
Lots of changes here and the area in x509_str.c (566-589) needs a closer look.

```
overlong lines added:
src/ssl.c:6103 /* Removes the CA with the passed in subject hash from the cert manager's CA cert store. */
src/x509_str.c:37 static int X509StorePopCert(WOLFSSL_STACK *certs_stack, WOLFSSL_STACK *dest_stack,
src/x509_str.c:569         /* Current certificate failed, but it is possible there is an alternative
src/x509_str.c:572          WOLFSSL_MSG("X509_verify_cert current cert failed, retrying with other certs.");
src/x509_str.c:573          ret = X509StoreRemoveCa(ctx->store, ctx->current_cert, WOLFSSL_TEMP_CA);
src/x509_str.c:1097 static int X509StorePopCert(WOLFSSL_STACK *certs_stack, WOLFSSL_STACK *dest_stack, WOLFSSL_X509 *cert) {
src/x509_str.c:1105             wolfSSL_sk_X509_push(dest_stack, (WOLFSSL_X509*)wolfSSL_sk_pop_node(certs_stack, i));
src/x509_str.c:1459         wc_InitDecodedCert(dCert, x509->derCert->buffer, x509->derCert->length, NULL);
```

src/x509_str.c
```diff
@@ -491,12 +503,14 @@ int wolfSSL_X509_verify_cert(WOLFSSL_X509_STORE_CTX* ctx)
             ret = X509StoreAddCa(ctx->store, issuer,
                                             WOLFSSL_TEMP_CA);
             if (ret != WOLFSSL_SUCCESS) {
-                goto exit;
+                goto retry;
```
JacobBarthelmeh: In "retry" the depth is incremented, it appears to me that this increment is done expecting that the depth had previously been decremented on an attempt to get the chain. I think that the way this is currently setup "depth" can end up being off. For example multiple chains of depth 3 where depth is decremented a couple times then the fail to verify happens (or adding the X509 to the store fails).

src/x509_str.c
```diff
@@ -548,9 +563,31 @@ int wolfSSL_X509_verify_cert(WOLFSSL_X509_STORE_CTX* ctx)
...
 
         depth--;
+        continue;
+
+retry:
+        /* Current certificate failed, but it is possible there is an alternative
+         * cert with the same subject key which will work.  Retry until all
+         * possible candidate certs are exhausted. */
+         WOLFSSL_MSG("X509_verify_cert current cert failed, retrying with other certs.");
+         ret = X509StoreRemoveCa(ctx->store, ctx->current_cert, WOLFSSL_TEMP_CA);
```
JacobBarthelmeh: It looks like this function removes a single temp CA? Is there cases where an intermediate CA unintentionally stays in the list of CA's after a failed attempt at verification and is used with the next attempt?

kareem-wolfssl: Yes, this is removing the current cert from the CA list if it is present with type WOLFSSL_TEMP_CA.  Otherwise known invalid CAs will stay in the cert manager's CA list and potentially cause false verifications.

>Is there cases where an intermediate CA unintentionally stays in the list of CA's after a failed attempt at verification and is used with the next attempt?

That is correct, since the intermediate CA is potentially added to the CA list earlier with the call to X509StoreAddCa, the intermediate CA would stay in the CA list and be used for verification for all future attempts in this function despite being invalid.  This call prevents that.  Otherwise the CA isn't removed until the call to wolfSSL_CertManagerUnloadTempIntermediateCerts in exit.

src/x509_str.c
kareem-wolfssl: Basically, the intention of the retry block is to completely undo the previous iteration of the while loop, which should always have decremented depth by one.  The only cases where it wouldn't reach depth-- are either from a break statement (which will end the function), goto retry or goto exit.
To be clear, this function is forming one chain per call, there aren't multiple chains.

src/x509_str.c
```diff
@@ -1390,6 +1450,36 @@ static int X509StoreAddCa(WOLFSSL_X509_STORE* store,
...
+
+        if (dCert == NULL) {
+            return result;
+        }
+        XMEMSET(dCert, 0, sizeof(DecodedCert));
+        wc_InitDecodedCert(dCert, x509->derCert->buffer,
+                           x509->derCert->length, NULL);
+        result = wc_ParseCert(dCert, CA_TYPE, NO_VERIFY, store->cm);
+        if (result)
+            return WOLFSSL_FATAL_ERROR;
```
dgarske: Can leak `dCert` here, please fix.

tests/api.c
```diff
@@ -27374,6 +27374,46 @@ static int test_wolfSSL_X509_STORE_CTX_ex11(X509_STORE_test_data *testData)
     X509_STORE_free(store);
     return EXPECT_RESULT();
 }
+
+#ifdef HAVE_ECC
+static int test_wolfSSL_X509_STORE_CTX_ex12(void)
+{
+    EXPECT_DECLS;
+    X509_STORE* store = NULL;
```
julek-wolfssl: Moving the guard like this will allow us to omit the guard in the call to `test_wolfSSL_X509_STORE_CTX_ex12` later.
```suggestion
static int test_wolfSSL_X509_STORE_CTX_ex12(void)
{
    EXPECT_DECLS;
#ifdef HAVE_ECC
    X509_STORE* store = NULL;
```

tests/api.c
```diff
@@ -27374,6 +27374,46 @@ static int test_wolfSSL_X509_STORE_CTX_ex11(X509_STORE_test_data *testData)
...
+    ExpectNotNull(store = X509_STORE_new());
+    ExpectNotNull(rootEccX509 = test_wolfSSL_X509_STORE_CTX_ex_helper(intCARootECCFile));
+    ExpectIntEQ(X509_STORE_add_cert(store, rootEccX509), 1);
+    ExpectNotNull(badAkiX509 = test_wolfSSL_X509_STORE_CTX_ex_helper(intCABadAKIECCFile));
+    ExpectIntEQ(X509_STORE_add_cert(store, badAkiX509), 1);
+    ExpectNotNull(ctx = X509_STORE_CTX_new());
+    ExpectNotNull(ca1X509 = test_wolfSSL_X509_STORE_CTX_ex_helper(intCA1ECCFile));
+    ExpectIntEQ(X509_STORE_CTX_init(ctx, store, ca1X509, NULL), 1);
+    ExpectIntEQ(X509_verify_cert(ctx), 1);
+    ExpectNotNull(chain = X509_STORE_CTX_get_chain(ctx));
```
julek-wolfssl: Would be good to have a test case to make sure verification fails when only `badAkiX509` is loaded.

src/x509_str.c
```diff
@@ -491,12 +504,14 @@ int wolfSSL_X509_verify_cert(WOLFSSL_X509_STORE_CTX* ctx)
...
             added = 1;
             ret = X509StoreVerifyCert(ctx);
             if (ret != WOLFSSL_SUCCESS) {
-                goto exit;
+                if ((origDepth - depth) <= 1)
+                    added = 0;
+                goto retry;
             }
             /* Add it to the current chain and look at the issuer cert next */
             wolfSSL_sk_X509_push(ctx->chain, ctx->current_cert);
```
julek-wolfssl: `else` section needs to be indented by one. Move the `{}` of `else` outside of the `#if` to improve readability and not have to guard the `}`.

src/x509_str.c
```diff
@@ -548,9 +564,34 @@ int wolfSSL_X509_verify_cert(WOLFSSL_X509_STORE_CTX* ctx)
...
+            depth++;
     }
 
 exit:
+    /* Copy back failed certs if verification failed. */
+    if (ret != WOLFSSL_SUCCESS) {
+        while (wolfSSL_sk_X509_num(failedCerts) > 0)
+        {
+            wolfSSL_sk_X509_push(certs, wolfSSL_sk_X509_pop(failedCerts));
+        }
```
julek-wolfssl: This has infinite loop potential if for any reason `wolfSSL_sk_X509_pop` doesn't decrement the number of certs. It would be better to do error checking if we are going to rely on the counter for the loop exit.

src/x509_str.c
```diff
@@ -548,9 +564,34 @@ int wolfSSL_X509_verify_cert(WOLFSSL_X509_STORE_CTX* ctx)
...
+    }
+    if (failedCerts) {
+        wolfSSL_sk_X509_pop_free(failedCerts, NULL);
+    }
```
julek-wolfssl:
```suggestion
    wolfSSL_sk_X509_pop_free(failedCerts, NULL);
```

src/x509_str.c
```diff
@@ -34,8 +34,13 @@
 #ifdef OPENSSL_EXTRA
 static int X509StoreGetIssuerEx(WOLFSSL_X509 **issuer,
                             WOLFSSL_STACK *certs, WOLFSSL_X509 *x);
+static int X509StorePopCert(WOLFSSL_STACK *certs_stack,
+                            WOLFSSL_STACK *dest_stack,
+                            WOLFSSL_X509 *cert);
 static int X509StoreAddCa(WOLFSSL_X509_STORE* store,
                                           WOLFSSL_X509* x509, int type);
+static int X509StoreRemoveCa(WOLFSSL_X509_STORE* store,
+                                            WOLFSSL_X509* x509, int type);
```
julek-wolfssl: Place new `static` functions next to `wolfSSL_X509_verify_cert` so that we don't need forward declarations. They're only used there.

src/x509_str.c
```diff
@@ -1390,6 +1450,39 @@ static int X509StoreAddCa(WOLFSSL_X509_STORE* store,
...
+    }
+
+    if (dCert) {
+        wc_FreeDecodedCert(dCert);
+        XFREE(dCert, NULL, DYNAMIC_TYPE_DCERT);
+    }
+
+    return result;
+}
+
```
julek-wolfssl: Would be easier to call `GetHashId` (you can make it non-static) on `x509->subjKeyId`. Look at `DecodeSubjKeyId` and `CopyDecodedToX509` to see which fields to use.

src/x509_str.c
```diff
@@ -1057,6 +1098,25 @@ static int X509StoreGetIssuerEx(WOLFSSL_X509 **issuer,
     return WOLFSSL_FAILURE;
 }
 
+static int X509StorePopCert(WOLFSSL_STACK *certs_stack,
+                            WOLFSSL_STACK *dest_stack,
+                            WOLFSSL_X509 *cert) {
```
julek-wolfssl: This will be a better name.
```suggestion
static int X509StoreMoveCert(WOLFSSL_STACK *certs_stack,
                            WOLFSSL_STACK *dest_stack,
                            WOLFSSL_X509 *cert) {
```

src/ssl.c
```diff
@@ -6100,6 +6100,59 @@ int AddCA(WOLFSSL_CERT_MANAGER* cm, DerBuffer** pDer, int type, int verify)
...
+        }
+        prev = current;
+        current = current->next;
+    }
+    wc_UnLockMutex(&cm->caLock);
+
+    WOLFSSL_LEAVE("RemoveCA", ret);
+
+    return ret;
+}
```
julek-wolfssl, line 6190: The `current == cm->caTable[row]` case can be simplified by using a double pointer.
```suggestion
int RemoveCA(WOLFSSL_CERT_MANAGER* cm, byte* hash, int type)
{
    Signer* current;
    Signer** prev;
    int     ret = WC_NO_ERR_TRACE(WOLFSSL_FAILURE);
    word32  row;

    WOLFSSL_MSG("Removing a CA");

    if (cm == NULL || hash == NULL) {
        return ret;
    }

    row = HashSigner(hash);

    if (wc_LockMutex(&cm->caLock) != 0) {
        return ret;
    }
    current = cm->caTable[row];
    prev = &cm->caTable[row];
    while (current) {
        byte* subjectHash;

    #ifndef NO_SKID
        subjectHash = current->subjectKeyIdHash;
    #else
        subjectHash = current->subjectNameHash;
    #endif

        if ((current->type == type) &&
            (XMEMCMP(hash, subjectHash, SIGNER_DIGEST_SIZE) == 0)) {
            *prev = current->next;
            FreeSigner(current, cm->heap);
            ret = WOLFSSL_SUCCESS;
            break;
        }
        prev = &current->next;
        current = current->next;
    }
    wc_UnLockMutex(&cm->caLock);

    WOLFSSL_LEAVE("RemoveCA", ret);

    return ret;
}
```

src/x509_str.c
```diff
@@ -548,9 +564,34 @@ int wolfSSL_X509_verify_cert(WOLFSSL_X509_STORE_CTX* ctx)
         }
 
         depth--;
+        continue;
+
+retry:
```
julek-wolfssl: We limit goto only to error handling. Use a func/macro and continue.

julek-wolfssl (changes requested):

src/x509_str.c
```diff
@@ -551,6 +626,16 @@ int wolfSSL_X509_verify_cert(WOLFSSL_X509_STORE_CTX* ctx)
...
+    /* Copy back failed certs if verification failed. */
+    if (ret != WOLFSSL_SUCCESS) {
+        numFailedCerts = wolfSSL_sk_X509_num(failedCerts);
+        for (i = 0; i < numFailedCerts; i++)
+        {
+            wolfSSL_sk_X509_push(certs, wolfSSL_sk_X509_pop(failedCerts));
+        }
+    }
+    wolfSSL_sk_X509_pop_free(failedCerts, NULL);
+
```
julek-wolfssl: Don't we always want to copy back the failed certs? This way the store is left in an unknown state after each run.

julek-wolfssl (changes requested):

src/x509_str.c
julek-wolfssl: After a successful run, the store's cert list is left in an unpredictable state since the user won't know which certs failed.

douzzer (approved):

PR #8705, opened 2025-04-23
Fix for STM32 Hash with IRQ enabled

dgarske:
# Description

Fix for STM32 Hash with NVIC (IRQ) enabled that can cause a DINIS interrupt that does not get cleared. If the HASH NVIC tab has Interrupts enabled it can cause an IRQ to be triggered that is not cleared. This is because the wolfSSL implementation of STM32 Hash does not call the HAL HASH API's and does not use interrupts yet. 

Fixes ZD 19778

# Testing

Tested on STM32H753.
Screenshot 2025-04-23 at 12 00 53 PM
Screenshot 2025-04-23 at 12 01 10 PM

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


dgarske:
Retest this please: "Found unhandled org.jenkinsci.plugins.workflow.support.steps.AgentOfflineException exception:"

SparkiDev (approved):

PR #8706, opened 2025-04-23
New build option to allow reuse of the windows crypt provider handle …

dgarske:
# Description

New build option to allow reuse of the windows crypt provider handle (`WIN_REUSE_CRYPT_HANDLE`). Documentation from Windows suggests using the same handle from different threads in the same process is safe when `CRYPT_VERIFYCONTEXT` is used.

Fixes ZD 19754

# Testing

Tested on Windows 10 with Visual Studio 2022 with and without `WIN_REUSE_CRYPT_HANDLE`.

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


dgarske:
Retest this please: "Found unhandled org.jenkinsci.plugins.workflow.support.steps.AgentOfflineException exception:"

wolfcrypt/src/random.c
```diff
@@ -2741,14 +2745,26 @@ int wc_GenerateSeed(OS_Seed* os, byte* output, word32 sz)
         }
     #endif /* HAVE_INTEL_RDSEED */
 
+#ifdef WIN_REUSE_CRYPT_HANDLE
```
redbaron, line 2748: `OS_Seed` already contains `ProviderHandle`,  wouldn't it be enough _not_ to release it with `CryptReleaseContext` and init only when not initialised?  This will avoid global static

wolfcrypt/src/random.c
```diff
@@ -2711,6 +2711,10 @@ int wc_GenerateSeed(OS_Seed* os, byte* output, word32 sz)
 
 #elif defined(USE_WINDOWS_API)
 
+#ifdef WIN_REUSE_CRYPT_HANDLE
```
redbaron, line 2714: if it is safe optimisation, can it be enabled unconditionally without build option? If it is not and users must make choice what are consequences here?

wolfcrypt/src/random.c
```diff
@@ -2741,14 +2745,26 @@ int wc_GenerateSeed(OS_Seed* os, byte* output, word32 sz)
         }
     #endif /* HAVE_INTEL_RDSEED */
 
+#ifdef WIN_REUSE_CRYPT_HANDLE
+    if (gHandle == 0) {
+        if(!CryptAcquireContext(&gHandle, 0, 0, PROV_RSA_FULL,
+                                CRYPT_VERIFYCONTEXT))
+            return WINCRYPT_E;
+    }
+    os->handle = gHandle;
```
redbaron, line 2754: if `WIN_REUSE_CRYPT_HANDLE` is enabled `os->handle` seems to be unused and can be removed from `OS_Seed`

wolfcrypt/src/random.c
```diff
@@ -2741,14 +2745,26 @@ int wc_GenerateSeed(OS_Seed* os, byte* output, word32 sz)
         }
     #endif /* HAVE_INTEL_RDSEED */
 
+#ifdef WIN_REUSE_CRYPT_HANDLE
+    if (gHandle == 0) {
```
redbaron, line 2749: is `wc_GenerateSeed` supposed to be thread safe? if it is, then this check is not safe.

dgarske: Hi @redbaron , no this is not thread-safe and it will need fixed before this can be considered. It was already pointed out by a few other wolfSSL engineers.

wolfcrypt/src/random.c
dgarske: @redbaron , in order to consider making this default I'll have to do some more testing with multiple threads to confirm there aren't any issues.

wolfcrypt/src/random.c
dgarske: There could be multiple OS_Seed provider handles and without a global static there would be no way to properly cleanup handles.

wolfcrypt/src/random.c
dgarske: Keeping for compatibility (structure needs a member)

dgarske:
Retest this please: "Found unhandled org.jenkinsci.plugins.workflow.support.steps.AgentOfflineException exception:"

SparkiDev (approved):

PR #8708, opened 2025-04-23
In wolfSSL_CTX_set_cert_store, send certificates into the CertMgr

anhu:
# Description

Fixes zd#

# Testing

How did you test?

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


anhu:
Jenkins retest this please

(lost the build data)

anhu:
Jenkins retest this please

> java.io.StreamCorruptedException: invalid stream header: 636F7272

anhu:
jenkins retest this please

anhu:
Jenkins retest this please

tests/api.c
```diff
@@ -28772,6 +28772,47 @@ static int test_wolfSSL_CTX_set_srp_password(void)
...
+
+    /* Create a new X509_STORE */
+    ExpectNotNull(store = X509_STORE_new());
+
+    /* Load a certificate */
+    ExpectNotNull(cert = wolfSSL_X509_load_certificate_file(caCert,
+                                                            SSL_FILETYPE_PEM));
+
+    /* Add the certificate to the store */
+    ExpectIntEQ(X509_STORE_add_cert(store, cert), SSL_SUCCESS);
```
JacobBarthelmeh, line 28794: I was looking over the support case referenced and I don't think this test reflects the edge case reported. It looks like we should be testing the order of :

`wolfSSL_CTX_set_cert_store(ctx, store);`

then

`X509_STORE_add_cert(store, cert)`

Where a cert store is assigned to the WOLFSSL_CTX struct and after it is assigned to it there is certificates added.

src/ssl.c
```diff
@@ -12946,6 +12947,26 @@ int wolfSSL_set_compression(WOLFSSL* ssl)
...
 
+        /* wolfSSL_CTX_set_cert_store() (this function) associates str with the
+         * wolfSSL_CTX. It is clear that this is a TLS use case which means we
+         * should move all the certs, if any, into the CertMgr and set
+         * str->certs to NULL as that will allow the certs to be properly
+         * processed. */
+        if (str->certs != NULL) {
+            while (wolfSSL_sk_X509_num(str->certs) > 0) {
+                int ret;
+                x = wolfSSL_sk_X509_pop(str->certs);
```
JacobBarthelmeh, line 12958: The cert store has 3 X509 stacks, one for `certs`, one for `trusted`, and one named `owned`. Are we sure that it is the `certs` stack that should get added as CA's?

src/ssl.c
```diff
@@ -12946,6 +12947,26 @@ int wolfSSL_set_compression(WOLFSSL* ssl)
...
+                ret = X509StoreAddCa(str, x, WOLFSSL_USER_CA);
+                if (ret != WOLFSSL_SUCCESS) {
+                    WOLFSSL_MSG("Error adding CA certificate to store");
+                    wolfSSL_X509_free(x); /* Free the certificate to avoid memory leaks */
+                    break; /* Exit the loop on failure */
+                }
+            }
+            wolfSSL_sk_X509_pop_free(str->certs, NULL);
```
JacobBarthelmeh, line 12966: Why does the `certs` stack need free'd and emptied here? Is it being converted over to a WOLFSSL_CTX cert store, and that future calls to wolfSSL_X509_STORE_add_cert would then load `certs` as a CA?

tests/api.c
```diff
@@ -28772,6 +28772,47 @@ static int test_wolfSSL_CTX_set_srp_password(void)
...
+
+    /* Create a new SSL_CTX */
+    ExpectNotNull(method = wolfSSLv23_server_method());
+    ExpectNotNull(ctx = wolfSSL_CTX_new(method));
+
+    /* Set the store in the SSL_CTX */
+    wolfSSL_CTX_set_cert_store(ctx, store);
+
+    /* Verify that the certs member of the store is null */
+    ExpectNull(store->certs);
```
JacobBarthelmeh, line 28805: We should be testing a successful verification making use of the loaded CA's.

src/ssl.c
JacobBarthelmeh: Also it would be worth checking if this behavior conforms to the expected OpenSSL compat behavior. Attempting untrusted vs trusted self signed certificate loads after the store is set to the WOLFSSL_CTX struct.

JacobBarthelmeh (changes requested):

src/ssl.c
ColtonWilley: In my opinion, certs should get added but trusted should also.

src/ssl.c
ColtonWilley: There is a concept of SSL_CTX owned X509_STORE vs non-owned. While confusing, it was done this way to leave all existing code for TLS and CM usage unmodified. To determine if X509_STORE is owned by SSL_CTX or not, we do NULL checks on str->certs. So that is why it needs to be pop_freed here, otherwise logic in X509_STORE wont work right.

tests/api.c
ColtonWilley: This is going to be tricky or even impossible to fully support with the current design. X509_STORE owned by SSL_CTX expects everything to be in underlying CM, while X509_STORE not-owned expects them in str->certs. I should say this flow will work for TLS, but setting a X509_STORE inside a SSL_CTX, then expecting direct X509_STORE APIs to work right is likely not going to work with current model.

anhu:
Assigned to @ColtonWilley as agreed in offline conversation.

anhu:
started over.  Closing in favor of new one.

PR #8723, opened 2025-04-30
Sniffer: Add multiple sessions by removing cached check

lealem47:
# Description

By removing the caching check, we can store more than one TLS session for the sniffer. This fixes a bug with the sniffer not being able to decrypt messages after a session ticket resumption if a session ID was also present in the client hello. We add/store the session by the sessionID if it's present, but then when we see the session ticket message we don't use the session ticket to store the session if we already added it by it's sessionID. With this PR, we just add it both times.

Fixes zd#19721

# Testing

Tested with the snifftest and the pcap provided by the customer.

`./sslSniffer/sslSnifferTest/snifftest -pcap test.mpn2v.2y.pcap -key rsa.pem.key -server 10.128.2.247 -port 9443`

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


lealem47:
Jenkins retest this please

src/sniffer.c
```diff
@@ -3475,17 +3474,14 @@ static int ProcessSessionTicket(const byte* input, int* sslBytes,
...
-                NULL, 0);
-            if (sess == NULL) {
-                SetupSession(session->sslServer);
-                AddSession(session->sslServer); /* don't re add */
-            #ifdef WOLFSSL_SNIFFER_STATS
-                INC_STAT(SnifferStats.sslResumptionInserts);
-            #endif
-            }
-            session->flags.cached = 1;
+        WOLFSSL_SESSION* sess = wolfSSL_GetSession(session->sslServer,
```
SparkiDev, line 3477: Can't declare a variable in the middle of a block with old C versions.
Still have customers using compilers that are old C versions.

SparkiDev (changes requested):

SparkiDev:
Are there times when we want only one cached session?
If so, add a compile time check to enable that behaviour.

src/sniffer.c
lealem47: Fixed

lealem47:
> Are there times when we want only one cached session? If so, add a compile time check to enable that behaviour.

Ideally we wouldn't, since we want the sniffer to be able to decrypt what it should. I do think this should be the default but I've added the `SNIFFER_SINGLE_SESSION_CACHE` macro for cases sensitive in  resources usage deltas (especially if decoding huge heaps of TLS connections)

SparkiDev (approved):

SparkiDev:
retest this plesae
Hudson failure

PR #8724, opened 2025-05-01
Fix an off-by-one buffer error in asn.c.

seego5555:
Need to ensure "finish-start" is less than the size before using it as an index a few lines down.

# Description

On line 26156 of asn.c, a check is made that "finish-start" is greater than the size of the buffer and if so, it returns an error.  A few lines down the buffer is accessed at index "finish-start", which will write off the end of the buffer if "finish-start" equals the size of the buffer.  The fix is to ensure that "finish-start" is less than the buffer size by changing "<" to "<=" on the comparison.

ZD 19834

### DIFF
--- a/wolfcrypt/src/asn.c
+++ b/wolfcrypt/src/asn.c
@@ -26153,7 +26153,7 @@ int wc_EncryptedInfoParse(EncryptedInfo* info, const char** pBuffer,
             newline = XSTRNSTR(finish, "\r", min(finishSz, PEM_LINE_LEN));
 
             /* get cipher name */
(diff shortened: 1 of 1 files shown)


dgarske:
Hi @seego5555 ,

Thank for your this PR. I do not see you setup as a contributor. Typically for a small PR like this we will treat as bug report and then reproduce fix ourselves. Can you tell us more about your project, where you are located and how you discovered this issue? If you would like to keep details private please email support at wolfssl dot com and reference this PR.

Thanks,
David Garske, wolfSSL

dgarske:
Fixed in https://github.com/wolfSSL/wolfssl/pull/8775

PR #8733, opened 2025-05-02
RISC-V 64-bit: fix raw hash when using crypto instructions

SparkiDev:
# Description

Digest state is not always stored in a way that can be directly copied out.

# Testing

./configure CC=riscv64-linux-gnu-gcc --host=riscv64 --disable-shared LDFLAGS=--static --enable-riscv-asm=zvkned 

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


SparkiDev:
retest this please
Hudson error

dgarske (approved):

PR #8748, opened 2025-05-08
Update PKCS7 to use X509 STORE for internal verification

ColtonWilley:
Update PKCS7 to use X509 STORE for internal verification instead of underlying CM.

# Description
Current PKCS7 uses underlying CM for verification, which will not account for intermediate certs since the X509_STORE rewrite. Change PKCS7 to use X509 store for verification. 

Fixes zd 19835


ColtonWilley:
Retest this please.

dgarske (approved):

PR #8759, opened 2025-05-09
Rename variable index to idx to avoid conflicting declaration.

anhu:
Fixes https://github.com/wolfSSL/wolfssl/issues/8731

Portability fix: Issue with shadow declaration of "index" from `/usr/include/string.h` which is: `extern char *index (__const char *__s, int __c) __THROW __attribute_pure__;`.

### DIFF
--- a/wolfcrypt/src/wc_pkcs11.c
+++ b/wolfcrypt/src/wc_pkcs11.c
@@ -535,7 +535,7 @@ static int Pkcs11Slot_FindByTokenName(Pkcs11Dev* dev,
     CK_RV         rv;
     CK_ULONG      slotCnt = 0;
     CK_TOKEN_INFO tinfo;
-    int           index = -1;
+    int           idx = -1;
     CK_SLOT_ID*   slot = NULL;
 
     rv = dev->func->C_GetSlotList(CK_TRUE, NULL, &slotCnt);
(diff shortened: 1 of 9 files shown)


dgarske (changes requested):
Looking through the code I also found occurrences in sniffer.c and kdf.c. Look for `int index` and `const byte* index`.

dgarske:
> Looking through the code I also found occurrences in sniffer.c and kdf.c. Look for `int index` and `const byte* index`.

Also please update the coding standard to mention reserved keywords... There are a few others... min, max, send, dup, cipherType.

wolfcrypt/src/wc_pkcs11.c
```diff
@@ -535,7 +535,7 @@ static int Pkcs11Slot_FindByTokenName(Pkcs11Dev* dev,
     CK_RV         rv;
     CK_ULONG      slotCnt = 0;
     CK_TOKEN_INFO tinfo;
-    int           index = -1;
+    int           idx = -1;
```
dgarske, line 538: This isn't changing the calling code below. See line 550, 555.

dgarske (changes requested):

anhu:
Jenkins retest this please

anhu:
jenkins retest this please

PR #8792, opened 2025-05-21
Coverity: remove dead code issue in ssl.c

rlm2002:
# Description

Remove dead code coverity issue in ssl.c with NO_OLD_TLS macro

### DIFF
--- a/src/ssl.c
+++ b/src/ssl.c
@@ -13545,10 +13545,12 @@ static int Set_CTX_max_proto_version(WOLFSSL_CTX* ctx, int ver)
     }
 
     switch (ver) {
+#ifndef NO_TLS
+#ifndef NO_OLD_TLS
     case SSL2_VERSION:
         WOLFSSL_MSG("wolfSSL does not support SSLv2");
         return WOLFSSL_FAILURE;
-#ifndef NO_TLS
+#endif
     case SSL3_VERSION:
         wolfSSL_CTX_set_options(ctx, WOLFSSL_OP_NO_TLSv1);
         FALL_THROUGH;
@@ -13603,9 +13605,6 @@ static int Set_CTX_max_proto_version(WOLFSSL_CTX* ctx, int ver)
         }
         /* Update the method */
         switch (ver) {
-        case SSL2_VERSION:
-            WOLFSSL_MSG("wolfSSL does not support SSLv2");
-            return WOLFSSL_FAILURE;
     #ifndef NO_TLS
         case SSL3_VERSION:
             ctx->method->version.minor = SSLv3_MINOR;


src/ssl.c
```diff
@@ -13603,9 +13603,6 @@ static int Set_CTX_max_proto_version(WOLFSSL_CTX* ctx, int ver)
         }
         /* Update the method */
         switch (ver) {
-        case SSL2_VERSION:
```
dgarske, line 13606: Instead of removing please wrap with `#ifndef NO_OLD_TLS`.

dgarske (changes requested):

rlm2002:
Retest this please Jenkins
`Found unhandled org.jenkinsci.plugins.workflow.support.steps.AgentOfflineException exception:
Unable to create live FilePath for wolf-linux-cloud-node-7y4k14; wolf-linux-cloud-node-7y4k14 was marked offline: Connection was broken`

src/ssl.c
```diff
@@ -13603,9 +13603,11 @@ static int Set_CTX_max_proto_version(WOLFSSL_CTX* ctx, int ver)
         }
         /* Update the method */
         switch (ver) {
+    #ifndef NO_OLD_TLS
```
dgarske, line 13606: Can you move `#ifndef NO_TLS` at line 13610 to above at 13606? I think the `case SSL2_VERSION:` should be included in the NO_TLS logic.

dgarske (changes requested):

src/ssl.c
```diff
@@ -13603,9 +13605,6 @@ static int Set_CTX_max_proto_version(WOLFSSL_CTX* ctx, int ver)
         }
         /* Update the method */
         switch (ver) {
-        case SSL2_VERSION:
```
dgarske, line 13606: Please don't remove. Apply same pattern...

```
#ifndef NO_TLS
#ifndef NO_OLD_TLS
```

dgarske (changes requested):

src/ssl.c
rlm2002: After rereading the file, I came to the conclusion that  `Set_CTX_max_proto_version()` will return `WOLFSSL_FAILURE` if `ver` matches the `SSL2_VERSION` case at the beginning of the function. If this occurs, then the second switch statement will never be reached. The second check occurs in an if statement, but `ver` is not changed at this point and which means the function should have already been returned if ver is `SSL2_VERSION`. I can keep the case to be exhaustive if that is preferred.

dgarske: I agree. Sorry about that. Looked like those were in different functions. This looks good. marking resolved/approved.

dgarske (approved):

rlm2002:
retest this please Jenkins

SparkiDev (approved):

PR #8799, opened 2025-05-22
Fix issue with CSharp and Windows CE with conversion of ASCII and Unicode

dgarske:
# Description

Fix issue with CSharp and Windows CE with conversion of WideChar->MultiByte and MultiByte->WideChar with odd length and extra null terminator. Fixed issue with WinCE wolfSSL.get_error.

# Testing

Tested with customer hardware on Win CE 6.0 and VS 2005. Test code:

```
/* string conversion tests */
/* WinCE:       using Unicode 16-bit
 * wolfSSL DLL: using multi-byte (8-bit) */
string caCert1 = Path.GetFullPath("\\Cert\\ca-cert.pem");
string uCaCert1 = wolfssl.WideCharToMultiByte(caCert1);
string bCaCert1 = wolfssl.MultiByteToWideChar(uCaCert1);
Console.WriteLine("Before: " + caCert1 + ", After: " + bCaCert1);
/* odd length test */
string caCert2 = Path.GetFullPath("\\Certs\\ca-cert.pem");
string uCaCert2 = wolfssl.WideCharToMultiByte(caCert2);
string bCaCert2 = wolfssl.MultiByteToWideChar(uCaCert2);
Console.WriteLine("Before: " + caCert2 + ", After: " + bCaCert2);
```

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/wrapper/CSharp/wolfSSL_CSharp/wolfSSL.cs
+++ b/wrapper/CSharp/wolfSSL_CSharp/wolfSSL.cs
@@ -42,41 +42,43 @@ namespace wolfSSL.CSharp
          */
 #if WindowsCE
         /// <summary>
-        /// Convert unicode string to ASCII
+        /// Convert MBCS (8-bit single/multi byte) to Wide Char/Unicode (16-bit) character set
         /// </summary>
-        public static string UnicodeToAscii(string msg)
+        public static string MultiByteToWideChar(string msg)
         {
             if (msg == null)
                 return null;
-            /* Convert Unicode to Bytes */
+            /* Convert to Byte Array */
             byte[] bytes = Encoding.Unicode.GetBytes((string)msg.ToString());
-            /* Convert to ASCII */
+            /* Convert to String */
             string ret = Encoding.ASCII.GetString(bytes, 0, bytes.Length);
-            /* odd length unicode might have extra null terminator, so remove */
-            return ret.Replace("\0", "");
+            /* Remove possible extra null terminator */
+            int len = 0;
+            while (len < ret.Length && ret[len] != 0) len++;
+            return ret.Substring(0, len);
         }
 
         /// <summary>
-        /// Convert string to Unicode
+        /// Convert Unicode/Wide Char (16-bit) to MBCS (8-bit single/multi byte) character set
         /// </summary>
-        public static string AsciiToUnicode(string msg)
+        public static string WideCharToMultiByte(string msg)
         {
             if (msg == null)
                 return null;
-            /* Get length and round up to even unicode */
+            /* Get length and round up to even for multibyte / unicode */
             int msgLen = msg.Length;
             msgLen = ((msgLen + 1) & ~1);
             byte[] bytes = new byte[msgLen];
-            /* Convert Ascii to Bytes */
+            /* Convert to Byte Array */
             byte[] msgBytes = Encoding.ASCII.GetBytes((string)msg.ToString());
             msgBytes.CopyTo(bytes, 0);
-            /* Convert to Unicode */
+            /* Convert to String */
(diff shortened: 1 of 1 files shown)


dgarske:
Retest this please

dgarske:
@gojimmypi I tested your proposed solution on customer hardware and it does not work. The result from your test just truncates non ascii characters. 
The wolfSSL DLL is expecting mutli-byte unicode. The strings it returns look like your screen capture uCaCert and it expects strings like that too.
Please consider this PR good as-is.

gojimmypi:
Hi @dgarske 

Perhaps I am misunderstanding the objective.

> result from your test just truncates non ascii characters.

Well, yes. I'm assuming your `UnicodeToAscii` intended to do just that: takes a Unicode string such as "`Hello, 世界`" and converts it to ASCII. In this case _truncating_ the non-ASCII `世界` part. 

My proposed alternative only looks at ASCII < 128  (and UnicodeCategory.NonSpacingMark values such as `é` vs `e`)

> The wolfSSL DLL is expecting mutli-byte unicode.

It is my understanding from the [docs](https://learn.microsoft.com/en-us/dotnet/api/system.string) that the `System.String` class "Represents text as a sequence of UTF-16 code units"  for all versions of .Net, including the CE version.

> I tested your proposed solution on customer hardware and it does not work

Here's what the current master branch looks like, clearly a problem with the trailing `m` char:

image

Your proposed change fixes the final result of `bCaCert`:

image

However, my concern is regarding `string uCaCert = wolfssl.AsciiToUnicode(caCert)`; The `uCaCert` value should be recognized as "`\\Cert\\ca-cert.pem`" when converting from ASCII to Unicode, no? Instead the result is garbled.

Isn't `Encoding.Unicode.GetString(bytes, 0, bytes.Length)` [here](https://github.com/dgarske/wolfssl/blob/db0b0e28d259cc3bae4597eda543bfef4d1585c2/wrapper/CSharp/wolfSSL_CSharp/wolfSSL.cs#L74)  expecting a byte array that's _already_ Unicode byte array, not ASCII?

I respect that the fix may be working for a given use case. My concern is the long term consequences having `wolfssl.AsciiToUnicode` come back to bite us later.

gojimmypi:
Hi @dgarske 

Is the intention of your `AsciiToUnicode` to convert a c-ctyle string into a binary storage in the .Net `String` type? If so, do you think it would be better to use a different type, other than `String` for the return parameter?

dgarske:
I spent a good part of today understanding this issue and made some additional improvements to the code and spent more time testing on hardware.

The C# code is using wide-char (16-bit) and VS2005 C is using single-byte (MBCS). The conversion routines are clearly documented now.

Screenshot 2025-05-28 at 9 31 50 AM
Screenshot 2025-05-28 at 9 44 10 AM

If available I would have used MultiByteToWideChar and WideCharToMultiByte, but from C# WinCE 6 there is not a way.

gojimmypi (approved):
Confirmed working and reversable. A little unusual to have `WideCharToMultiByte()` return type of `String` and not be a valid C# string of the same value, but I see what's going on now & why.

gojimmypi:
Hi @dgarske 

The function name change is helpful. There might be an argument that the `String` return type does not actually contain a valid C# string for `WideCharToMultiByte()` return value, but it is much more clear now that the assignment is not Unicode.

The "search for zero" rather than "replace all zeroes" is also a good improvement.

Nice catch on the number of params for `wolfSSL_ERR_reason_error_string`.

I don't have customer hardware, so I trust that this is working there.

I retested in VS2005 PocketPC 2003 SE and confirmed the values are functional and reversable.

image

Thank you

dgarske:
Retest this please. PPC test abort.

PR #8805, opened 2025-05-27
Fixed bug with set_groups for tls 1.2

BridgerVoss:
# Description

Fixed bug for wolfssl_set_groups choosing curves on tls 1.2 on the server that the client did not have.

Fixes zd#

# Testing

How did you test?

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


julek-wolfssl (changes requested):
This is a valid solution but I would prefer you to keep the entire PR only inside of `TLSX_SupportedCurve_Parse` by:
- if `TLSX_Find(*extensions, TLSX_SUPPORTED_GROUPS)` returns non-`NULL` then storing only the intersect or returning an error if no common groups are found
- storing the incoming list if `TLSX_Find(*extensions, TLSX_SUPPORTED_GROUPS)` returns `NULL`

This will bypass the need to make changes elsewhere and you won't have struct members that only apply sometimes (talking about `!IsAtLeastTLSv1_3(ssl->version) && (curve)->isSharedCurve != 1`).

Please also add unit tests to `test_tls.c` to check that the connection fails when there is no group overlap and succeeds when a common group is found.

julek-wolfssl:
Continued in https://github.com/wolfSSL/wolfssl/pull/8824

PR #8822, opened 2025-05-30
Support PBE_AES(256|128)_CBC certificate encryptions on wc_PKCS12_create()

kojiws:
# Description

This PR adds `PBE_AES256_CBC` and `PBE_AES128_CBC` certificate encryption supports on `wc_PKCS12_create()`.

Fixes zd#19747

# Testing
./configure CFLAGS="-DUSE_CERT_BUFFERS_2048" --enable-debug --enable-des3 --enable-asn && make test

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/tests/api.c
+++ b/tests/api.c
@@ -18924,8 +18924,9 @@ static int test_wc_i2d_PKCS12(void)
 static int test_wc_PKCS12_create_once(int keyEncType, int certEncType)
 {
     EXPECT_DECLS;
-#if !defined(NO_ASN) && defined(HAVE_PKCS12) && !defined(NO_PWDBASED) \
-    && !defined(NO_HMAC) && !defined(NO_CERTS) && defined(USE_CERT_BUFFERS_2048)
+#if !defined(NO_ASN) && defined(HAVE_PKCS12) && !defined(NO_PWDBASED) && \
+    !defined(NO_RSA) && !defined(NO_ASN_CRYPT) && \
+    !defined(NO_HMAC) && !defined(NO_CERTS) && defined(USE_CERT_BUFFERS_2048)
 
     byte* inKey = (byte*) server_key_der_2048;
     const word32 inKeySz= sizeof_server_key_der_2048;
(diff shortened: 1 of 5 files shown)


tests/api.c
```diff
@@ -18989,18 +18990,17 @@ static int test_wc_PKCS12_create(void)
 {
     EXPECT_DECLS;
 #if !defined(NO_DES3) && !defined(NO_SHA)
-    ExpectIntEQ(test_wc_PKCS12_create_once(PBE_SHA1_DES3, PBE_SHA1_DES3),
-        TEST_SUCCESS);
+    EXPECT_TEST(test_wc_PKCS12_create_once(PBE_SHA1_DES3, PBE_SHA1_DES3));
 #endif
-#if defined(HAVE_AES_CBC) && !defined(NO_AES_256) && !defined(NO_DES3) && \
-    !defined(NO_SHA)
-    ExpectIntEQ(test_wc_PKCS12_create_once(PBE_AES256_CBC, PBE_SHA1_DES3),
```
JacobBarthelmeh, line 18997: Is this removing the mix match test?

`ExpectIntEQ(test_wc_PKCS12_create_once(PBE_AES256_CBC, PBE_SHA1_DES3)`

We should still test that case.

JacobBarthelmeh (changes requested):
Nice addition! My only comment is if we are reducing code coverage with the test case?

tests/api.c
kojiws: I added the `ExpectIntEQ(test_wc_PKCS12_create_once(PBE_AES256_CBC, PBE_SHA1_DES3)` before to test the `PBE_AES256_CBC` for key encryption. But, for cert encryption, `PBE_AES_256_CBC` wasn't supported then so that I set it to `PBE_SHA1_DES3` instead of `PBE_AES256_CBC` temporarily. In this pull request, I restored from the temporal `PBE_SHA1_DES3` to `PBE_AES_256_CBC` because it has been supported for the cert encryption. 

But, now I realized that keeping mix match test would be better after you mentioned. So, I will keep one mix match test. Probably, one test is good because testing all mixed pattern needs resources.

JacobBarthelmeh (approved):

PR #8830, opened 2025-06-02
add option to not use CT code with min/max

JacobBarthelmeh:
Ran into a case where an old compiler (gcc version 4.7.2 (GCC)) was returning the wrong values for max/min with the constant time code.


dgarske:
Retest this please: "Found unhandled java.io.InterruptedIOException exception:"

dgarske (dismissed):
For this use case on the Renesas RX and old compiler does it make sense to just enable `WOLFSSL_NO_CT_OPS` instead? If ctMaskWord is misbehaving how do we know it won't show up elsewhere?

JacobBarthelmeh:
> For this use case on the Renesas RX and old compiler does it make sense to just enable `WOLFSSL_NO_CT_OPS` instead? If ctMaskWord is misbehaving how do we know it won't show up elsewhere?

True. I think that is what I started with, but there is spots in wolfSSL that do not have the proper macro guard for a build with WOLFSSL_NO_CT_OPS. I can change this PR to instead be a fix for that build and then use it instead.

```
 ./configure --enable-all CPPFLAGS=-DWOLFSSL_NO_CT_OPS && make
 
  CC       examples/echoclient/echoclient.o
src/internal.c:5396:12: error: call to undeclared function 'ctMaskGT'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
    mask = ctMaskGT(ret, 0);
           ^

```

dgarske:
> > For this use case on the Renesas RX and old compiler does it make sense to just enable `WOLFSSL_NO_CT_OPS` instead? If ctMaskWord is misbehaving how do we know it won't show up elsewhere?
> 
> True. I think that is what I started with, but there is spots in wolfSSL that do not have the proper macro guard for a build with WOLFSSL_NO_CT_OPS. I can change this PR to instead be a fix for that build and then use it instead.
> 
> ```
>  ./configure --enable-all CPPFLAGS=-DWOLFSSL_NO_CT_OPS && make
>  
>   CC       examples/echoclient/echoclient.o
> src/internal.c:5396:12: error: call to undeclared function 'ctMaskGT'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
>     mask = ctMaskGT(ret, 0);
>            ^
> ```

I just saw this too. Places in rsa.c that assume those constant time functions are available. Yes let's fix. No rush

JacobBarthelmeh:
The locations where constant time mask functions were being used I think should be preserved (better that wolfSSL does not have the option to build without it). I added a preprocessor warning and update the macro guard.

SparkiDev (approved):

PR #8832, opened 2025-06-02
Aarch64 XFENCE

### LINKED ISSUE #8804: Switch at runtime from 'isb' to 'sb' instruction for ARM v8.5 onward
We would like to propose this enhancement for ARM architecture that, at runtime, it switches from ISB to SB instruction if supported by the system. SB (Speculation Barrier) is a modern barrier which is available from armv8.5a. It achieves the same result as issuing ISB, but instead of flushing the CPU it does so by serializing older instructions to be non-speculative before it completes. This is less disruptive than an "isb" to high performance CPUs.

We have noticed the implementation of the ISB instruction on the XFENCE code:
https://github.com/wolfSSL/wolfssl/blob/master/wolfssl/wolfcrypt/wc_port.h#L1539-L1540

```
/* ARM specific for HWCAP_SB detection at runtime */
#if defined(__aarch64__) && defined(__linux__)
#include <sys/auxv.h>
#ifndef HWCAP_SB
#define HWCAP_SB (1 << 29)
#endif

...
#elif (defined (__arm__) && (__ARM_ARCH > 6)) || defined(__aarch64__)
#if defined(__linux__)
static inline void arm_arch_pause(void) {
  static int use_spin_delay_sb = -1;

  if (__builtin_expect(use_spin_delay_sb == 1, 1)) {
    XASM_VOLATILE(".inst 0xd50330ff");  // SB instruction encoding
  }
  else if (use_spin_delay_sb == 0) {
    XASM_VOLATILE("isb");
  }
  else {
    // Initialize variable and check if SB is supported
    if (getauxval(AT_HWCAP) & HWCAP_SB)
      use_spin_delay_sb = 1;
    else
      use_spin_delay_sb = 0;
  }
}
#else
static inline void arm_arch_pause(void) {
    XASM_VOLATILE("isb");
}
#endif // __linux__

#define XFENCE()  arm_arch_pause()
#else
...
```

SparkiDev:
# Description

Use sb instruction instead of isb if available.

Fixes #8804

# Testing

./configure --disable-shared LDFLAGS=--static --host=aarch64 CC=aarch64-linux-gnu-gcc --enable-armasm=barrier-sb

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -3225,6 +3225,26 @@ then
             ENABLED_ARMASM_CRYPTO_SM4=yes
             ENABLED_ARMASM_PLUS=yes
             ;;
+        barrier-sb)
+            case $host_cpu in
+            *aarch64*)
+                ;;
+            *)
+                AC_MSG_ERROR([SB instructions only available on Aarch64 CPU.])
+                break;;
+            esac
+            ENABLED_ARMASM_BARRIER_SB=yes
+            ;;
+        barrier-detect)
+            case $host_cpu in
+            *aarch64*)
+                ;;
+            *)
+                AC_MSG_ERROR([SB instructions only available on Aarch64 CPU.])
+                break;;
+            esac
+            ENABLED_ARMASM_BARRIER_DETECT=yes
+            ;;
         *)
             AC_MSG_ERROR([Invalid choice of ARM asm inclusions (yes, sha512-crypto, sha3-crypto): $ENABLED_ARMASM.])
             break;;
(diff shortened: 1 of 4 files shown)


configure.ac
```diff
@@ -3228,6 +3228,26 @@ then
             ENABLED_ARMASM_CRYPTO_SM4=yes
             ENABLED_ARMASM_PLUS=yes
             ;;
+        barrier-sb)
+            case $host_cpu in
+            *aarch64*)
+                ;;
+            *)
+                AC_MSG_ERROR([SB instructions only available on Aarch64 CPU.])
```
salvatoredipietro: nit: "SB instructions only available on aarch64 v8.5+ CPU."

SparkiDev:
retest this please

configure.ac
SparkiDev: changed

SparkiDev:
retest this please

julek-wolfssl (changes requested):
Now that github has [arm based runners](https://docs.github.com/en/actions/writing-workflows/choosing-where-your-workflow-runs/choosing-the-runner-for-a-job#standard-github-hosted-runners-for-public-repositories) it would be nice to add these configurations to CI. Preferably some config where the tests hit these fence calls.

SparkiDev:
Not sure we can test the new code as other implementations will be used.

salvatoredipietro:
looks good to me

PR #8835, opened 2025-06-03
Use simpler sizeof expressions

kojiws:
# Description

Original `sizeof(*var)` expression is a bit confusing when `*var` == NULL;

Fixes zd#

# Testing

./configure --enable-smallstack && make test

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


kojiws:
I doubted that this modification in this PR seems related to a problem on a zendesk ticket.
However, I found out that the modification was not related to the problem.
So, I close this PR.

PR #8848, opened 2025-06-05
dtlsProcessPendingPeer: correctly set the current peer

### LINKED ISSUE #8841: [Bug]: dtlsProcessPendingPeer incorrect update of the peer when using CID
### Contact Details

teozwiebel@protonmail.com

### Version

latest master

### Description

./configure --enable-all --enable-debug

OS : Ubuntu 24.04 LTS

Hello,

While testing the DTLS examples from wolfssl-examples (with the fix referenced in [this issue](https://github.com/wolfSSL/wolfssl-examples/issues/505)), I encountered a bug that appears to affect both DTLS 1.2 and 1.3 with the CID enabled.

Image

Image

It looks like the call to wolfSSL_dtls_set_peer in the dtlsProcessPendingPeer function is using the wrong address. This causes the peer address (sockaddr_in) to be set incorrectly.

Here we can see that the SockAddrSet is called with the arg "peer=0x5555555e948" which is the address *&ssl->buffers.dtlsCtx.pendingPeer.sa*, but it looks like it should be the address **pointed by** *ssl->buffers.dtlsCtx.pendingPeer.sa* which is a sockaddr_in structure

Image

The problematic code is located at line 21861 in internal.c:

```
            (void)wolfSSL_dtls_set_peer(ssl,
                    &ssl->buffers.dtlsCtx.pendingPeer.sa,
                    ssl->buffers.dtlsCtx.pendingPeer.sz);
```

I was able to resolve the issue by removing the &, so that the pointer passed corresponds directly to the structure:

```
            (void)wolfSSL_dtls_set_peer(ssl,
                    ssl->buffers.dtlsCtx.pendingPeer.sa,
                    ssl->buffers.dtlsCtx.pendingPeer.sz);
```

Best regards,

Téo Zwiebel

### Reproduction steps

For Wolfssl : 
1. ./configure --enable-all --enable-debug
2. sudo make
3. sudo make install

For the Client and Server examples : 
1. Fix the call to wolfSSL_dtls_cid_parse ([PR](https://github.com/wolfSSL/wolfssl-examples/pull/506))
2. make debug
3. ./server-dtls-demux
5. ./client-dtls(13)-cid 127.0.0.1
6. type multiple messages

For the fix : 
1. Change the line mentionned 
2. Wolfssl : make & install
3. Examples : make clean & make
4. run the examples

### Relevant log output

```shell

```

julek-wolfssl:


julek-wolfssl:
AgentOfflineException

julek-wolfssl:
Retest this please

JacobBarthelmeh (approved):

PR #8849, opened 2025-06-06
Reseed DRBG in RAND_poll()

holtrop:
# Description

Reseed DRBG in RAND_poll()
Fixes ZD#20016

Update: I created a separate PR https://github.com/wolfSSL/wolfssl/pull/8852 which includes this fix and adds a fork "unit" test to api.c. I did that separately in case there were concerns with that test and it was desired to merge a fix for this issue without the test for now. However, if that test is acceptable, then this PR can be closed and we can merge that one instead.

# Testing

Ran the r0 test application from ZD#20016 before and after the fix.

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/src/ssl.c
+++ b/src/ssl.c
@@ -26041,11 +26041,25 @@ int wolfSSL_RAND_poll(void)
         return  WOLFSSL_FAILURE;
     }
     ret = wc_GenerateSeed(&globalRNG.seed, entropy, entropy_sz);
-    if (ret != 0){
+    if (ret != 0) {
         WOLFSSL_MSG("Bad wc_RNG_GenerateBlock");
         ret = WOLFSSL_FAILURE;
-    }else
-        ret = WOLFSSL_SUCCESS;
+    }
+    else {
+#ifdef HAVE_HASHDRBG
+        ret = wc_RNG_DRBG_Reseed(&globalRNG, entropy, entropy_sz);
+        if (ret != 0) {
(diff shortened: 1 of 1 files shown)


holtrop:
Retest this please Jenkins

src/ssl.c
```diff
@@ -26041,11 +26041,24 @@ int wolfSSL_RAND_poll(void)
...
     }
     ret = wc_GenerateSeed(&globalRNG.seed, entropy, entropy_sz);
-    if (ret != 0){
+    if (ret != 0) {
         WOLFSSL_MSG("Bad wc_RNG_GenerateBlock");
         ret = WOLFSSL_FAILURE;
-    }else
+    }
+    else {
+#ifdef HAVE_HASHDRBG
```
holtrop, line 26049: It looks like wc_RNG_DRBG_Reseed() is only available if HAVE_HASHDRBG is set. So if HAVE_HASHDRBG is not set, this is just using the old behavior of returning SUCCESS. If this is not correct please let me know.

JacobBarthelmeh: It's a rare case wher the RNG functions are used with out HAVE_HASHDRBG. This should emmit a debug message and return a failure though if called and unable to use wc_RNG_DRBG_Reseed().

JacobBarthelmeh (changes requested):

holtrop:
retest this please

JacobBarthelmeh (approved):

PR #8850, opened 2025-06-06
Fix X509 load locations to handle PEM files with multiple certs

ColtonWilley:
# Description

wolfSSL_X509_load_locations() previously did not handle PEM files with multiple certs in one file, only the first cert would be loaded. Modify wolfSSL_X509_load_locations() to fix this by using existing logic in x509.c to loop over the PEM headers and load each item, cert or CRL. 

Fixes zd 19835

# Checklist

 - [x] added tests

### DIFF
--- a/src/x509_str.c
+++ b/src/x509_str.c
@@ -1878,74 +1878,6 @@ int X509StoreLoadCertBuffer(WOLFSSL_X509_STORE *str,
 
 #if !defined(NO_FILESYSTEM) && !defined(NO_WOLFSSL_DIR)
 
-static int X509StoreReadFile(const char *fname,
-                StaticBuffer *content, word32 *bytesRead, int *type)
-{
-    int ret = -1;
-    long sz = 0;
-#ifdef HAVE_CRL
-    const char* header = NULL;
-    const char* footer = NULL;
-#endif
-
-    ret = wolfssl_read_file_static(fname, content, NULL, DYNAMIC_TYPE_FILE,
-        &sz);
-    if (ret == 0) {
-        *type = CERT_TYPE;
-        *bytesRead = (word32)sz;
-#ifdef HAVE_CRL
-        /* Look for CRL header and footer. */
-        if (wc_PemGetHeaderFooter(CRL_TYPE, &header, &footer) == 0 &&
-                (XSTRNSTR((char*)content->buffer, header, sz) !=
-                    NULL)) {
-            *type = CRL_TYPE;
-        }
-#endif
-    }
-
-    return (ret == 0 ? WOLFSSL_SUCCESS : WOLFSSL_FAILURE);
-}
-
-static int X509StoreLoadFile(WOLFSSL_X509_STORE *str,
-                                        const char *fname)
-{
-    int ret = WOLFSSL_SUCCESS;
-    int type = 0;
-#ifndef WOLFSSL_SMALL_STACK
-    byte   stackBuffer[FILE_BUFFER_SIZE];
-#endif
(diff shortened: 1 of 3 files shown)


ColtonWilley:
Retest this please.  

tests/api.c
```diff
@@ -28968,6 +28968,39 @@ static int test_wolfSSL_X509_STORE_load_locations(void)
     return EXPECT_RESULT();
 }
 
+static int test_wolfSSL_X509_STORE_load_multiple_certs(void)
```
julek-wolfssl: Move it to `test_x509.c`.

tests/api.c
```diff
@@ -28968,6 +28968,39 @@ static int test_wolfSSL_X509_STORE_load_locations(void)
...
+        WOLFSSL_SUCCESS);
+
+    /* Count X509 certificate objects in store */
+    ExpectNotNull(objs = X509_STORE_get0_objects(store));
+    for (i = 0; i < sk_X509_OBJECT_num(objs) && EXPECT_SUCCESS(); i++) {
+        X509_OBJECT *obj = (X509_OBJECT*)sk_X509_OBJECT_value(objs, i);
+        if (obj && X509_OBJECT_get_type(obj) == X509_LU_X509) {
+            cert_count++;
+        }
+    }
```
julek-wolfssl: I would add a subject name check or something similar in the loop to make sure that the objects are correctly populated.

src/x509_str.c
```diff
@@ -1607,23 +1539,9 @@ WOLFSSL_API int wolfSSL_X509_STORE_load_locations(WOLFSSL_X509_STORE *str,
...
-                != WOLFSSL_SUCCESS ||
-                wolfSSL_CertManagerDisableCRL(str->cm) != WOLFSSL_SUCCESS) {
-            WOLFSSL_MSG("Enable CRL failed");
-            wolfSSL_CTX_free(ctx);
-            return WOLFSSL_FAILURE;
-        }
-    }
-#endif
-
     /* Load individual file */
```
julek-wolfssl, line 1906: Are CRL's still loaded without this? I'm not sure if we have a test for this. Please check.

julek-wolfssl (changes requested):

src/x509_str.c
ColtonWilley: Yes, this check is now handled by existing similar block in X509LoadPemFile(). I will have to double check on tests for it.

tests/api.c
ColtonWilley: I mean I agree we should start putting more x509 testing in there, but as is it  would be pretty inconsistent as all other X509_STORE* tests are in api.c I would prefer to keep them together for now and split them all off into x509_test.c in the future.

julek-wolfssl: We don't have to move all tests at once. We can move them one by one and make sure that new tests are written there. We want to stop the growth of `api.c`. It already takes long enough to compile.

julek-wolfssl (dismissed):
Please move the new tests and check if the CRL functionality is tested somewhere.

tests/api.c
ColtonWilley: Tests moved to test_x509.c

tests/api.c
ColtonWilley: Subject name check added.

dgarske (dismissed):

SparkiDev (dismissed):

julek-wolfssl (dismissed):

julek-wolfssl:
@ColtonWilley please rebase to run all the latest tests.

PR #8866, opened 2025-06-10
New macro FILE_LINE_UNAVAILABLE

anhu:
Define it if  `__FILE__` nor `__LINE__` are not available. 

Fixes ZD 19956


wolfssl/wolfcrypt/logging.h
```diff
@@ -225,8 +225,10 @@ WOLFSSL_API void wolfSSL_SetLoggingPrefix(const char* prefix);
 
 #endif /* DEBUG_WOLFSSL && !WOLFSSL_DEBUG_ERRORS_ONLY */
 
-#if defined(DEBUG_WOLFSSL) || defined(OPENSSL_ALL) || defined(WOLFSSL_NGINX) ||\
-    defined(WOLFSSL_HAPROXY) || defined(OPENSSL_EXTRA)
+#if (defined(DEBUG_WOLFSSL) || defined(OPENSSL_ALL) || \
```
dgarske, line 228: Consider testing build with `FILE_LINE_UNAVAILABLE`. Or I think you'll need to add it to `.wolfssl_known_macro_extras`

anhu:
Jenkins retest this please. 

Oops! Not found.

wolfssl/wolfcrypt/logging.h
```diff
@@ -225,8 +225,10 @@ WOLFSSL_API void wolfSSL_SetLoggingPrefix(const char* prefix);
...
+     defined(WOLFSSL_NGINX) || defined(WOLFSSL_HAPROXY) || \
+     defined(OPENSSL_EXTRA)) && \
+    !defined(FILE_LINE_UNAVAILABLE)
```
dgarske, line 231: My preference is a macro named `WOLFSSL_FILE_LINE_UNAVAILABLE`.
Aren't there only places we use `__FILE__` and `__LINE__`? Couldn't someone just work around this by defining these themselves? I've never heard of a compiler not supporting these, but I have seen variations in the naming format. Are you sure this is the best fix?

anhu: The reporter stated that they could override `__func__` but not `__FILE__` and `__LINE__` . I'm not sure this is the best fix.  Let me talk to the customer and see if we can come up with a better solution.   I'll come back to this  later.

anhu:
Customer found an alternative solution.  Closing.

PR #8867, opened 2025-06-10
Improvements to RNG and compatibility layer

JacobBarthelmeh:
ZD#20016

### DIFF
--- a/src/ssl.c
+++ b/src/ssl.c
@@ -25500,6 +25500,12 @@ static int wolfSSL_RAND_InitMutex(void)
 
 #ifdef OPENSSL_EXTRA
 
+#if defined(HAVE_GETPID) && defined(HAVE_FIPS) && FIPS_VERSION3_LT(6,0,0)
+/* In older FIPS bundles add check for reseed here since it does not exist in
+ * the older random.c certified files. */
+static pid_t currentRandPid = 0;
+#endif
+
 /* Checks if the global RNG has been created. If not then one is created.
  *
  * Returns WOLFSSL_SUCCESS when no error is encountered.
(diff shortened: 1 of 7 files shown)


JacobBarthelmeh:
Retest this please Jenkins

src/ssl.c
```diff
@@ -25989,6 +26000,26 @@ int wolfSSL_RAND_bytes(unsigned char* buf, int num)
          * have the lock.
          */
         if (initGlobalRNG) {
+        #if defined(HAVE_GETPID) && !defined(WOLFSSL_NO_GETPID) && \
```
dgarske, line 26003: Is this HSVE_FIPS logic correct? Its forcing on FIPS.

tests/api.c
```diff
@@ -33124,6 +33130,46 @@ static int test_wolfSSL_RAND_bytes(void)
     ExpectIntEQ(RAND_bytes(my_buf, size3), 1);
     ExpectIntEQ(RAND_bytes(my_buf, size4), 1);
 
+#if defined(OPENSSL_EXTRA) && defined(HAVE_GETPID)
```
dgarske, line 33133: Add `&& !defined(WOLFSSL_NO_GETPID)`

JacobBarthelmeh: There is a macro check for WOLFSSL_NO_GETPID later on in the test to adjust the expected results.

src/ssl.c
JacobBarthelmeh: This code path will be used when FIPS is enabled and the FIPS version is less than 6.0.0. I've added the macro guard and code section in to account for random.c being inside the FIPS boundary. Was tested with `./fips-check linuxv5 keep` and then using `--enable-opensslextra` when configuring with the resulting XXX-fips-test.

PR #8879, opened 2025-06-13
Updates for OpenSSH 10.0p2

julek-wolfssl:
- random.c: use getrandom when available and fall back to direct file access
- openssh.yml: run more tests
- openssh.yml: add 10.0p2 and 9.9p2
- configure.ac: detect if `getrandom` is available on the system
- configure.ac: openssh requires WC_RNG_SEED_CB to always use `getrandom` so that the RNG doesn't get killed by SECCOMP


julek-wolfssl:
Retest this please

dgarske (approved):

Issue #8839, opened 2025-06-04
[Bug]: wolfSSL_X509_print returns error and fails to print certificate details when certificate created without common name


holtrop:
### Version

b077c81eb635392e694ccedbab8b644297ec0285

### Description

Create RSA public/private key pair per https://github.com/wolfSSL/wolfCLU README with:

    wolfssl genkey rsa -size 2048 -out mykey -outform pem -output KEY

Create certificate per https://github.com/wolfSSL/wolfCLU README with:

    wolfssl req -new -days 3650 -key mykey.priv -out test.cert -x509

Press enter for all fields (do not enter a CN).

Try to dump certificate with wolfCLU:

```
$ wolfssl x509 -inform pem -in test.cert -text
Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number:
            7b:71:31:af:08:19:18:01:89:c7:70:d9:cb:64:8f:2d
        Signature Algorithm: sha256WithRSAEncryption
unable to print certificate out
Error returned: -1.
        Issuer:
```

### Reproduction steps

See issue description.

### Relevant log output

```shell

```

holtrop:
Fixed via https://github.com/wolfSSL/wolfssl/pull/8880

PR #8880, opened 2025-06-13
Fix printing empty names in certificates

holtrop:
The empty-issuer-cert.pem certificate was created with:

    wolfssl genkey rsa -size 2048 -out mykey -outform pem -output KEY
    wolfssl req -new -days 3650 -key mykey.priv -out empty-issuer-cert.pem -x509

Prior to this fix this command would error printing the certificate:

    wolfssl x509 -inform pem -in empty-issuer-cert.pem -text

This fixes issue #8839.

# Testing

- Added automated test in `tests/api.c`
- Built wolfCLU with fix and verified `wolfssl` command line can now print the test certificate.

# Checklist

 - [X] added tests
 - [ ] updated/added doxygen


holtrop:
retest this please

dgarske (approved):

certs/empty-issuer-cert.pem
```diff
@@ -0,0 +1,17 @@
+-----BEGIN CERTIFICATE-----
```
dgarske, line 1: How did you create this certificate? Can you update the `certs/test/gen-testcerts.sh` or similar script when it comes time to update this (when it expires)? Thank you

dgarske (changes requested):

certs/empty-issuer-cert.pem
holtrop: The commands I used to create the certificate are in both the git commit message and this PR description.

I tried using `gen-testcerts.sh` to generate this, but openssl gives an error when CN is empty, which is exactly what I need to create this test certificate.

```
Error making certificate request
802BD70F0E7F0000:error:06800098:asn1 encoding routines:ASN1_mbstring_ncopy:string too short:crypto/asn1/a_mbstr.c:100:minsize=1
```

I created the certificate to expire after 10 years. Perhaps I could just recreate it with 100 years so we wouldn't need to worry about it expiring for a very long time?

dgarske: Please document the steps to create it in the library somewhere. Tracing back to PR is difficult. Yes please extend expiration date. Thank you

holtrop: Would an empty-issuer-cert.pem.txt file with instructions next to the .pem file be sufficient or do you have a better suggestion for where to document within the library?

dgarske: Put it in the .PEM or put it next to the code that calls it.

dgarske: My preference is documented in the gen-testcerts.sh. Second would be instructions next to the calling code. 
Note: You can have free text in the PEM, but that might confuse people.

holtrop:
retest this please

PR #8889, opened 2025-06-17
Remove DTLS from echo examples

holtrop:
# Description

Remove DTLS from echo examples

This fixes some intermittent CI testsuite failures.

# Testing

- `make check`
- testsuite-loop.sh from @julek-wolfssl (prior to this change it would fail within a couple minutes; with this in place it ran over 3 hours without failing)

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -213,7 +213,7 @@ AS_IF([test "$ax_enable_debug" = "yes"],
       [AM_CCASFLAGS="$AM_CCASFLAGS -DNDEBUG"])
 
 AC_ARG_ENABLE([debug-code-points],
-    [ AS_HELP_STRING([--enable-debug-code-points],[Include source file and line number in --enable-verbose messages.]) ],
+    [ AS_HELP_STRING([--enable-debug-code-points],[Include source file and line number in --enable-debug messages.]) ],
     [ ENABLED_DEBUG_CODEPOINTS=$enableval ],
     [ ENABLED_DEBUG_CODEPOINTS=no ]
     )


holtrop:
retest this please

rizlik:
do you know why is failing?

holtrop:
> I wonder if this doesn't break the echo functionality in the examples apps.

The echo functionality in the apps still works, it just does not use DTLS. That was what we had discussed in Slack.

holtrop:
> do you know why is failing?

There was a different type of intermittent error that I've seen less frequently than the one that I fixed in this PR.

dgarske:
> retest this please

I don't see additional comments as to why you had to retry testing. How do we know this will improve our testing reliability?

Do you think removing DTLS from the echoserver/echoclient is the best solution? Couldn't we still prioritize TLS even if DTLS is enabled?

holtrop:
> > retest this please
> 
> I don't see additional comments as to why you had to retry testing. How do we know this will improve our testing reliability?
> 
> Do you think removing DTLS from the echoserver/echoclient is the best solution? Couldn't we still prioritize TLS even if DTLS is enabled?

I requested a retest due to a different failure (error connecting). I'm happy to investigate that one separately but it did not seem related to the failure I was trying to fix in this PR.

I believe this change will improve our testing reliability because I was able to repeatedly quickly get the testsuite.test errors I fixed here within a couple minutes each time using the loop test from Juliusz but after this change I ran it over 3 hours and did not see it fail.

I was initially going to keep DTLS support in echoclient/echoserver but just not enable them for the CI tests, but Juliusz was concerned it would become untested dead code in our Slack thread. If you feel otherwise, I can go back to my initial idea to protect enabling DTLS in the echo examples with a different build option instead of just always enabling DTLS if support for it is built in.

dgarske:
> > > retest this please
> > 
> > 
> > I don't see additional comments as to why you had to retry testing. How do we know this will improve our testing reliability?
> > Do you think removing DTLS from the echoserver/echoclient is the best solution? Couldn't we still prioritize TLS even if DTLS is enabled?
> 
> I requested a retest due to a different failure (error connecting). I'm happy to investigate that one separately but it did not seem related to the failure I was trying to fix in this PR.
> 
> I believe this change will improve our testing reliability because I was able to repeatedly quickly get the testsuite.test errors I fixed here within a couple minutes each time using the loop test from Juliusz but after this change I ran it over 3 hours and did not see it fail.
> 
> I was initially going to keep DTLS support in echoclient/echoserver but just not enable them for the CI tests, but Juliusz was concerned it would become untested dead code in our Slack thread. If you feel otherwise, I can go back to my initial idea to protect enabling DTLS in the echo examples with a different build option instead of just always enabling DTLS if support for it is built in.

No action needed. As long as the DTLS tests are being done properly in the tests/unit.test I am fine with this. For the agent offline issues don't worry about those for now. Let's let the dust settle before we dig further.

PR #8890, opened 2025-06-17
Allow larger pathLen values in Basic Constraints.

kareem-wolfssl:
# Description

Allow higher max_path_len values.  We are arbitrarily limiting this to 8 bits, the RFC has no such restriction (it is only limited to an integer there).

Maintains the default WOLFSSL_MAX_PATH_LEN value of 127, users will have to override this value at build time for higher values.

Fixes zd#20022

# Testing

Customer provided certs with pathLen=1000.

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/wolfcrypt/src/asn.c
+++ b/wolfcrypt/src/asn.c
@@ -20631,7 +20631,7 @@ static int DecodeBasicCaConstraint(const byte* input, int sz, DecodedCert* cert)
     if (ret == 0) {
         /* Get the CA boolean and path length when present. */
         GetASN_Boolean(&dataASN[BASICCONSASN_IDX_CA], &isCA);
-        GetASN_Int8Bit(&dataASN[BASICCONSASN_IDX_PLEN], &cert->pathLength);
+        GetASN_Int16Bit(&dataASN[BASICCONSASN_IDX_PLEN], &cert->pathLength);
 
         ret = GetASN_Items(basicConsASN, dataASN, basicConsASN_Length, 1, input,
                            &idx, (word32)sz);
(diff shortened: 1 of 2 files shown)


SparkiDev (approved):

wolfcrypt/src/asn.c
```diff
@@ -20590,7 +20590,7 @@ static int DecodeBasicCaConstraint(const byte* input, int sz, DecodedCert* cert)
     if (ret == 0) {
         /* Get the CA boolean and path length when present. */
         GetASN_Boolean(&dataASN[BASICCONSASN_IDX_CA], &isCA);
-        GetASN_Int8Bit(&dataASN[BASICCONSASN_IDX_PLEN], &cert->pathLength);
+        GetASN_Int16Bit(&dataASN[BASICCONSASN_IDX_PLEN], &cert->pathLength);
```
SparkiDev: Similar change required for non-ASN template code.

SparkiDev (changes requested):

dgarske:
Retest this please: "history lost"

dgarske (dismissed):

wolfcrypt/src/asn.c
```diff
@@ -20631,7 +20636,7 @@ static int DecodeBasicCaConstraint(const byte* input, int sz, DecodedCert* cert)
...
```
dgarske, line 20639: Why aren't we testing WOLFSSL_MAX_PATH_LEN here? It might be useful to add comment on these indicating that WOLFSSL_MAX_PATH_LEN can be overridden if someone needs more than 127. The reason for the limit has to do with memory usage.

dgarske (changes requested):

wolfcrypt/src/asn.c
kareem-wolfssl: WOLFSSL_MAX_PATH_LEN is tested in the ASN template case below on line 20656.  I can add a comment, now that this is always a 16-bit variable we are slightly increasing memory usage.
I can look into making this dynamic where we only use a 16-bit integer if WOLFSSL_MAX_PATH_LEN > 127.

dgarske (approved):

PR #8891, opened 2025-06-17
Coverity Warnings Fix

JeremiahM37:
# Description

Potential fixes for the Coverity warnings addressed in the support ticket. So far, I think I have fixes for the mixed Enums issue in hmac.c as well as the unused value warnings. The undefined min warning can be ignored since there is a fallback it.
Fixes zd#20029

# Testing

How did you test?

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


wolfcrypt/src/pkcs12.c
```diff
@@ -1716,7 +1716,6 @@ int wc_PKCS12_parse_ex(WC_PKCS12* pkcs12, const char* psw,
 
         /* free temporary buffer */
         XFREE(buf, pkcs12->heap, DYNAMIC_TYPE_PKCS);
-        buf = NULL;
```
dgarske, line 1719: Clang-tidy is very unhappy with this change. Please find a better way to fix:

```
[clang-tidy-defaults] [7 of 7] [wolfssl]
    configure...   real 0m5.920s  user 0m4.165s  sys 0m2.072s
    build.../tmp/workspace/PRB-multi-test-script/wolfssl/wolfcrypt/src/pkcs12.c:1718:9: warning: Attempt to free released memory [clang-analyzer-unix.Malloc]
XFREE(buf, pkcs12->heap, DYNAMIC_TYPE_PKCS);
^```

File: wolfcrypt/src/hmac.c [hunk 2]
```diff
@@ -152,7 +152,7 @@ static int HmacKeyInitHash(wc_HmacHash* hash, int type, void* heap, int devId)
 {
     int ret = 0;
 
-    switch (type) {
+    switch ((int)type) {
```
dgarske: `type` is already an `int`, so this change isn't doing anything. Please revert.

dgarske (changes requested):

wolfcrypt/src/hmac.c
JeremiahM37: 233861-hmac c

Coverity is giving this warning due to inferred misuse of enums. I brough up potential solutions with Kareem, either changing the function declaration from (int type) to (enum wc_HashType type) or just casting type to int, and he suggested that simply casting to int would likely work and be the easier solution. Should I go try converting them all to enum wc_hashType and change WC_MD5 to WC_HASH_TYPE_MD5 in the function declaration or try to find a different solution entirely?

dgarske: The fix is to cast to `enum wc_HashType`

wolfcrypt/src/hmac.c
```diff
@@ -147,85 +147,84 @@ int wc_HmacSizeByType(int type)
...
 }
+
 static int HmacKeyInitHash(wc_HmacHash* hash, int type, void* heap, int devId)
 {
     int ret = 0;
 
-    switch ((enum wc_HashType)type) {
-    #ifndef WOLFSSL_NO_SHA224
-        case WC_HASH_TYPE_SHA224:
+    switch (type) {
```
dgarske: The old code was fine. I think you'll have issues with some of the older FIPS using this technique. The warnings reported are harmless and have to do with a mix of us passing `int type` and `wc_HashType type`. The important thing to identify is if you are actually comparing the same types, since we also have OID's and NID's.

PR #8893, opened 2025-06-18
ARM32/Thumb2/RISC-V 64 ASM: omit frame pointer

SparkiDev:
# Description

ARM32/Thumb2: Generated code now omits the frame pointer attribute on each function. Remove global use in configure.ac.
RISC-V 64: Omit the frame pointer on the one function that uses the register 's0'.

Fixes zd#20080

# Testing

Tested all inline ARM32/Thumb2/RISC-V 64 assembly.

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -3367,7 +3367,7 @@ then
         armv7m*)
             # QEMU doesn't work with armv7-m
             AM_CFLAGS="$AM_CFLAGS -DWOLFSSL_ARMASM_THUMB2"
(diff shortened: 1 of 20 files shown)


SparkiDev:
retest this please

dgarske:
I am going to test this on Cortex-M as well, since in debug mode we were getting the r7 conflict and this might be a good solution for that as well. I posted a comment to customer in ZD20080.

PR #8896, opened 2025-06-18
Fix memory leak in ParseCRL_Extensions

holtrop:
# Description

Fix memory leak in ParseCRL_Extensions

# Testing

Ran wolfssl-multi-test.sh sanitizer-all-asm-smallstack-async

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


wolfcrypt/src/asn.c
```diff
@@ -40593,9 +40593,13 @@ static int ParseCRL_Extensions(DecodedCRL* dcrl, const byte* buf, word32 idx,
...
 
-                if (ret == 0 && mp_toradix(m, (char*)dcrl->crlNumber,
-                             MP_RADIX_HEX) != MP_OKAY)
-                    ret = BUFFER_E;
+                if (ret == 0) {
+                    if (mp_toradix(m, (char*)dcrl->crlNumber,
+                             MP_RADIX_HEX) != MP_OKAY) {
+                        ret = BUFFER_E;
+                    }
+                    mp_free(m);
```
holtrop, line 40601: It looks like commit 02a49693e2bf63b76a95cba060bf5c9594af296d reworked this code and erroneously dropped the mp_free().

dgarske: Looks like we thought the `FREE_MP_INT_SIZE(m, NULL, DYNAMIC_TYPE_TMP_BUFFER);` below would also call `mp_free`. It does pass now with `../testing/git-hooks/wolfssl-multi-test.sh sanitizer-all-asm-smallstack-async` after this fix. But I would prefer the mp_free and `FREE_MP_INT_SIZE` be next to each other. Move one of the other please.

holtrop: It does not look like `wc_bigint_free()` checks if `a->buf` is non-`NULL` before attempting to `XFREE()` it. That is why I put the `mp_free()` call where I did, because it should only be called if `GetInt()` returns 0 and thus populates the `buf` field. And since the `ret` variable may be overwritten to `BUFFER_E` just above I didn't want to use it to guard whether or not to call `mp_free()` later.

I suppose I could add a second variable to hold the `GetInt()` result status and check it immediately before `FREE_MP_INT_SIZE()`. Would that be preferable?

dgarske: XFREE does NULL checks. It’s required. I just mean to keep the mp_free and FREE_MP_INT_SIZE next to each other.

holtrop: Ah, ok. I didn't know XFREE() was safe for the arg to be NULL in all implementations. I'll move it down then!

dgarske (approved):
Thank you. It passes `sanitizer-all-asm-smallstack-async`

dgarske:
Retest this please. Agent offline exception

PR #8907, opened 2025-06-20
Coverity warning fixes

JeremiahM37:
# Description
Potential fixes for the Coverity warnings addressed in the support ticket. 

Accidentally closed last pr for the same changes and it won't let me reopen for some reason.

Fixes zd#20029

# Testing

How did you test?

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


wolfcrypt/src/asn.c
```diff
@@ -25028,8 +25028,10 @@ Signer* findSignerByName(Signer *list, byte *hash)
     }
     return NULL;
 }
+/*declared with WOLFSSL_LOCAL due to coverity warning "routine not emitted",
+likely due to inconsistencies between its declaration and definition. */
 
-int ParseCertRelative(DecodedCert* cert, int type, int verify, void* cm,
+WOLFSSL_LOCAL int ParseCertRelative(DecodedCert* cert, int type, int verify, void* cm,
```
dgarske: This needs to be in the header, not the .c file.

dgarske: Make sure the header and .c definitions match. Remove the comment.

src/tls.c
```diff
@@ -9384,8 +9384,9 @@ static int TLSX_KeyShare_ProcessEcc_ex(WOLFSSL* ssl,
             break;
     #endif
         default:
-            /* unsupported curve */
+            curveId = ECC_CURVE_INVALID;
```
dgarske: Please revert this. The curveId not used can just be solved with the `(void)curveId` below.

wolfcrypt/src/pkcs12.c
```diff
@@ -830,7 +830,10 @@ int wc_d2i_PKCS12_fp(const char* file, WC_PKCS12** pkcs12)
         wc_PKCS12_free(*pkcs12);
         *pkcs12 = NULL;
     }
-    XFREE(buf, NULL, DYNAMIC_TYPE_TMP_BUFFER);
+    if (buf != NULL) {
```
dgarske: XFREE already checks NULL. Just add the `buf = NULL;`

dgarske (changes requested):
These fixes are all trying to appease a static analyzer but do not improve the code.

PR #8911, opened 2025-06-21
Allow configuration with only RSA cipher suites

gojimmypi:
# Description

Updates `settings.h` sanity check rule to allow all cipher suites except RSA to be disabled.

Currently the error is triggered when only RSA is enabled.

Fixes zd#

# Testing

How did you test? 

Mostly test in embedded using manual `user_settings.h`, for instance adding these settings at the end of [this ESP32 user_settings.h](https://github.com/wolfSSL/wolfssl/blob/master/IDE/Espressif/ESP-IDF/examples/wolfssl_server/components/wolfssl/include/user_settings.h) and toggling the `TEST_CASE` for only RSA or not:

```c
// #define TEST_CASE
#ifdef TEST_CASE
    #undef  NO_RSA
    #define NO_RSA
#else
    #undef  NO_RSA
    #define HAVE_RSA
    #define FP_MAX_BITS (2 * 4096)
#endif

#undef NO_OLD_TLS 
#undef NO_SHA
#undef NO_SHA1
#undef HAVE_ECC
#undef HAVE_CURVE25519
#undef HAVE_CURVE448
#undef HAVE_ED25519
#undef HAVE_ED448
#undef HAVE_SUPPORTED_CURVES
#undef HAVE_ECC
#undef HAVE_CURVE25519
#undef HAVE_CURVE448
#undef HAVE_ED25519
#undef HAVE_ED448
```

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen
 - [ ] updated appropriate READMEs
 - [ ] Updated manual and documentation

PR #8916, opened 2025-06-23
Revert PR #8911

dgarske:
# Description

Revert PR #8911. I should not have merged it.

After running some more tests on TLS v1.2 and TLS v1.3 the original error was correct.

* For TLS v1.2 RSA only is only supported with `WOLFSSL_STATIC_RSA`. 
* For TLS v1.3 RSA only is not supported (must be PFS).

# Testing

With and without `WOLFSSL_STATIC_RSA`.
```
./configure --disable-ecc --disable-dh --disable-curve448 --disable-ed448 --disable-curve25519 --disable-ed25519 --enable-debug --disable-shared --disable-tls13 --enable-sni CFLAGS="-DWOLFSSL_STATIC_RSA" && make
./examples/client/client -v 3 -h www.google.com -p 443 -g -x -S www.google.com -d
```

Without static RSA the result is -313. Works with static RSA.

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -19621,8 +19621,8 @@ static WC_INLINE int EncryptDo(WOLFSSL* ssl, byte* out, const byte* input,
 #endif
 
     (void)out;
-    (void)input;
     (void)sz;
+    (void)type;
 
     if (input == NULL) {
         return BAD_FUNC_ARG;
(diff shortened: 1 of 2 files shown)


dgarske:
Retest this please

dgarske:
Retest this please. History lost (can't tell why).

dgarske:
Jenkins retest this please. Again history is already deleted.... We might consider keeping more.

PR #8929, opened 2025-06-26
Regression testing

SparkiDev:
# Description

Fixes to get WOLFSSL_PUBLIC_MP testing passing.
Fix DH constant time agreement:
  - implement constant time encoding to big-endian byte array in TFM
- only force x to be zero for SP math as others implementations ensure unused words are zero
- exponentiate in constant time to the smallest number of words possible
- no need to encode into separate buffer anymore as encoding is constant time and front padded
- make requested_sz be the maximum size for the parameters and check against agreeSz
- update agreeSz to be the maximum valid size instead of filling all the buffer which may be many times too big

# Testing

Regression testing enabling and disabling PK algorithms with different math implementations.

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/wolfcrypt/src/dh.c
+++ b/wolfcrypt/src/dh.c
@@ -2058,80 +2058,19 @@ static int wc_DhAgree_Sync(DhKey* key, byte* agree, word32* agreeSz,
 #endif
 
 #ifdef WOLFSSL_HAVE_SP_DH
+    if (0
 #ifndef WOLFSSL_SP_NO_2048
-    if (mp_count_bits(&key->p) == 2048) {
-        if (mp_init(y) != MP_OKAY)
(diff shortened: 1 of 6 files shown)


SparkiDev:
mp_forcezero is void return

SparkiDev:
retest this please

FIPS windows testing: too many retries
```
Didn't fail enough times, try again
ret of generating responses = 2
Changed ret code from CAVP_RUN_IT_AGAIN to CAVP_FAILUREExiting harness with ret: -1
```

FIPS:
```
ERROR: Checkout failed
```

valgrind:
```
ERROR: Checkout failed
```

dgarske (approved):
Reviewed and tested. Thanks Sean

PR #8938, opened 2025-06-28
Fix x509 memory leak

Mohre:
# Description

The definition of `WOLFSSL_X509` class in Internal.h file adds `wolfSSL_Ref` by one of these two flags:
- OPENSSL_EXTRA
- OPENSSL_EXTRA_X509_SMALL

But the statement in `FreeX509` doesn't match the definition. So it doesn't free `ref` and causes memory leak.

# Testing

Run with `OPENSSL_EXTRA_X509_SMALL` and check memory with valgrind.

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -4849,7 +4849,7 @@ void FreeX509(WOLFSSL_X509* x509)
     }
     #endif /* WOLFSSL_DUAL_ALG_CERTS */
 
-    #if defined(OPENSSL_EXTRA) || defined(OPENSSL_ALL)
+    #if defined(OPENSSL_EXTRA) || defined(OPENSSL_EXTRA_X509_SMALL)
         wolfSSL_RefFree(&x509->ref);
     #endif
 }


dgarske:
Hi @Mohre ,

Thank you for this report. We are reviewing the testing around x509 small to see if we can find anything else. Since you are not an existing contributor and this is a small patch we will likely treat as a bug report.

Can you tell us more about your project and uses of our products?

Thanks,
David Garske, wolfSSL

Mohre:
It is a small change. Couldn't you fix that in your next release?

dgarske:
Hi @Mohre ,

Yes will will make sure this gets fixed. We will treat this as a bug report and put up another PR shortly. I've got @douzzer assigned to this and he will have a PR up soon.

Thanks,
David Garske, wolfSSL

dgarske:
Thank you! Replaced with https://github.com/wolfSSL/wolfssl/pull/8982

PR #8944, opened 2025-07-01
EVP HMAC: get working with WOLFSSL_HMAC_COPY_HASH

SparkiDev:
# Description

Get the EVP layer working with the wolfSSL HMAC implementation when WOLFSSL_HMAC_COPY_HASH is defined.

This define hashes the ipad and opad into temporary hashes and copies the required hash into the working hash when needed. Uses more memory but is faster when starting a new hash with the same key.

# Testing

./configure --disable-shared --disable-tls --enable-all --enable-hmac-copy

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/src/ssl_crypto.c
+++ b/src/ssl_crypto.c
@@ -1472,56 +1472,136 @@ int wolfSSL_HmacCopy(Hmac* dst, Hmac* src)
     #ifndef NO_MD5
         case WC_MD5:
             rc = wc_Md5Copy(&src->hash.md5, &dst->hash.md5);
+        #ifdef WOLFSSL_HMAC_COPY_HASH
+            if (rc == 0) {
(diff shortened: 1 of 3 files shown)


SparkiDev:
> 
>     * `declare-const-pointers` [snippet](https://github.com/wolfSSL/wolfssl/blob/109ebb2f94868a57bd83de2462ba4ebcc2cc5334/src/ssl_crypto.c#L1445): Declare the read-only parameter as const: `static int wolfssl_hmac_hash_copy(const wc_HmacHash* src, wc_HmacHash* dst, int type)`.
> 
> 

The hash functions that copy don't take a const for the source hash.

SparkiDev:
retest this please

PR #8946, opened 2025-07-01
CSharp Wrapper Improvements

gojimmypi:
# Description

C# Improvements

(more description soon)

Fixes zd# n/a

# Testing

How did you test?

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


gojimmypi:
Jenkins retest this please.

For missing results, SSL_read input error -308, error state on socket

wrapper/CSharp/wolfSSL_CSharp/wolfSSL.cs
```diff
@@ -36,6 +36,117 @@ public class wolfssl
...
+        /// <returns></returns>
+        [DllImport("kernel32.dll", SetLastError = true)]
+        static extern bool SetDllDirectory(string lpPathName);
+
+        /// <summary>
+        ///
+        /// </summary>
+        /// <param name="p"></param>
+        /// <returns></returns>
+        public static bool LoadDLL(string wolfsslPath) {
```
dgarske, line 149: This is not compatible with `#if WindowsCE`. Please fix. Thank you

```
1>C:\Windows\Microsoft.NET\Framework\v3.5\Csc.exe /noconfig /nowarn:1701,1702 /nostdlib+ /errorreport:prompt /warn:4 /define:DEBUG;TRACE;WindowsCE /reference:"C:\Program Files (x86)\Microsoft.NET\SDK\CompactFramework\v3.5\WindowsCE\mscorlib.dll" /reference:"C:\Program Files (x86)\Microsoft.NET\SDK\CompactFramework\v3.5\WindowsCE\System.Core.dll" /reference:"C:\Program Files (x86)\Microsoft.NET\SDK\CompactFramework\v3.5\WindowsCE\System.Data.DataSetExtensions.dll" /reference:"C:\Program Files (x86)\Microsoft.NET\SDK\CompactFramework\v3.5\WindowsCE\System.Data.dll" /reference:"C:\Program Files (x86)\Microsoft.NET\SDK\CompactFramework\v3.5\WindowsCE\System.dll" /reference:"C:\Program Files (x86)\Microsoft.NET\SDK\CompactFramework\v3.5\WindowsCE\System.Xml.dll" /reference:"C:\Program Files (x86)\Microsoft.NET\SDK\CompactFramework\v3.5\WindowsCE\System.Xml.Linq.dll" /debug+ /debug:full /filealign:512 /optimize- /out:obj\Debug\wolfSSLWrapper.dll /target:library Properties\AssemblyInfo.cs wolfCrypt.cs wolfSSL.cs X509.cs
wolfSSLWrapper\wolfSSL.cs(63,103): error CS1061: 'System.AppDomain' does not contain a definition for 'BaseDirectory' and no extension method 'BaseDirectory' accepting a first argument of type 'System.AppDomain' could be found (are you missing a using directive or an assembly reference?)
wolfSSLWrapper\wolfSSL.cs(87,42): error CS0117: 'System.Environment' does not contain a definition for 'CurrentDirectory'
wolfSSLWrapper\wolfSSL.cs(114,54): error CS0117: 'System.IO.Directory' does not contain a definition for 'GetParent'
wolfSSLWrapper\wolfSSL.cs(919,49): error CS0117: 'System.IO.Directory' does not contain a definition for 'GetParent'
```

dgarske (changes requested):

wrapper/CSharp/wolfSSL_CSharp/wolfSSL.cs
```diff
@@ -575,6 +703,8 @@ public void free()
         private extern static IntPtr wolfSSL_ERR_reason_error_string(uint err);
         [DllImport(wolfssl_dll)]
         private extern static int wolfSSL_get_error(IntPtr ssl, int err);
+        [DllImport(wolfssl_dll)]
+        private extern static int wolfSSL_get_alert_history(IntPtr ssl, ref WOLFSSL_ALERT_HISTORY h);
```
dgarske, line 887: Why is `wolfSSL_get_alert_history` in here twice?

dgarske (changes requested):

wrapper/CSharp/wolfSSL_CSharp/wolfSSL.cs
gojimmypi: I only see `wolfSSL_get_alert_history` for (1) `WindowsCE`  and (2) everything else, each with different declaration decorators. 

Did you see an additional one someplace?

image

dgarske: Sorry that's fine. Missed one was WindowsCE. Consider this item resolved.

JacobBarthelmeh (approved):

wrapper/CSharp/wolfSSL_CSharp/wolfSSL.cs
dgarske: @JacobBarthelmeh We should not merge this until these errors are resolved.

gojimmypi:
@dgarske @JacobBarthelmeh sorry for the delay, I ended up down a rabbit hole & should be done by the end of the day.

I've implemented a new `StringToAnsiPtr` and I'm converting callbacks from

```
public static void standard_log(int lvl, StringBuilder msg)
```

to

```
public static void standard_log(int lvl, IntPtr msg) {
```

wrapper/CSharp/wolfSSL_CSharp/wolfSSL.cs
```diff
@@ -575,6 +703,8 @@ public void free()
         private extern static IntPtr wolfSSL_ERR_reason_error_string(uint err);
         [DllImport(wolfssl_dll)]
```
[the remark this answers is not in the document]
gojimmypi: There's only one `wolfSSL_get_error` at a given time. The two instances are gated on `#if WindowsCE` or not.

wrapper/CSharp/wolfSSL_CSharp/wolfSSL.cs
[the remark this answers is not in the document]
gojimmypi: There's only one `wolfSSL_get_error` at a given time. The two instances are gated on #if WindowsCE or not. 

Each has a different decorator: 

- `[DllImport(wolfssl_dll)]` for CE
- `[DllImport(wolfssl_dll, CallingConvention = CallingConvention.Cdecl)]` otherwise.

wrapper/CSharp/wolfSSL_CSharp/wolfSSL.cs
```diff
@@ -575,6 +703,8 @@ public void free()
         private extern static IntPtr wolfSSL_ERR_reason_error_string(uint err);
         [DllImport(wolfssl_dll)]
         private extern static int wolfSSL_get_error(IntPtr ssl, int err);
+        [DllImport(wolfssl_dll)]
```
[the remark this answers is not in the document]
gojimmypi: `CallingConvention.Cdecl` is not supported on Visual Studio 2008 Windows CE, thus the conditional `WindowsCE `.

Otherwise this error is encountered:

```
Error	1	'System.Runtime.InteropServices.CallingConvention' does not contain a definition for 'Cdecl'	C:\workspace\wolfssl-gojimmypi\wrapper\CSharp\wolfSSL_CSharp\wolfSSL.cs	751	71	CSharp_CE_2008
```

See numerous examples throughout the code.

wrapper/CSharp/wolfSSL_CSharp/wolfSSL.cs
```diff
@@ -36,6 +36,117 @@ public class wolfssl
...
+            if (wolfsslPath == "") {
+                /* wolfSSL project should have created a DLL in [WOLFSSL_ROOT]\Debug, some number of directories up from here:  */
+                /* baseDir = Path.GetFullPath(Path.Combine(AppDomain.CurrentDomain.BaseDirectory, @"..\..\..\..\..\..\Debug")); */
+                string dir = Environment.CurrentDirectory;
+                if (verbose) {
+                    Console.WriteLine("Looking for wolfSSL in parents, starting from: " + dir);
+                }
+
+                while ((!string.IsNullOrEmpty(dir) && Directory.Exists(dir)) || !ret) {
+                    for (int i = 0; i < subDirsToCheck.Length; i++) {
```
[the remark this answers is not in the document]
gojimmypi: The for-loop is constrained by `subDirsToCheck.Length`. 

The while-loop is constrained by ` Directory.GetParent(dir) != null`

It is unlikely that `dir` will be invalid or empty, given that it is assigned `Environment.CurrentDirectory`.

When updated, we assume that Windows would not allow a bad parent directory in `dir = parent.FullName`

gojimmypi:
@dgarske @JacobBarthelmeh This PR has changed substantially since submitted. I've left the new changes in a separate commit for now.

I've addressed the _compile_ issues for CE, but I have been still  unable to figure out how to _run_ native C wolfSSL in any of the CE emulators with the `wolfSSL_CSharp` C# wrapper class.

@dgarske please test on customer hardware. Thank you.

wrapper/CSharp/wolfSSL_CSharp/wolfSSL.cs
gojimmypi: @dgarske this was my mistake; I've addressed the CE compile errors in https://github.com/wolfSSL/wolfssl/pull/8946/commits/acd8fedebcb5742df96311a8e01e37036486fd2c 

Latest updates include improved overall CE detection and compile-time suggestions.

gojimmypi:
I've addressed the [do-not-change-external-apis comment](https://github.com/wolfSSL/wolfssl/pull/8946#issuecomment-3029657139) in https://github.com/wolfSSL/wolfssl/pull/8946/commits/b987303631ccd6e985688ac407a2e6aadbd3d6e5:

Introduced `loggingCbEx` (taking parameters ` int level, IntPtr msg ` instead of ` int lvl, StringBuilder msg `).

The `IntPtr msg` vs `StringBuilder msg` logging parameter changes should also be more robust and intuitive with regards to memory use and prevention of leaks.

Included is a consolidation of logging to use only the `wolfssl` class library. Duplicate code removed from `wolfCrypt.cs`.

gojimmypi:
Jenkins retest this please

For Unable to create live FilePath for wolf-linux-cloud-node-[n]; wolf-linux-cloud-node[n] was marked offline: Connection was broken

wrapper/CSharp/wolfSSL_CSharp/X509.cs
```diff
@@ -19,10 +19,22 @@
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1335, USA
  */
 
+
+/* CE Not always reliably detected. Define your own WindowsCE as needed */
+#if _WIN32_WCE || WINCE || PocketPC
+    #if !WindowsCE
```
dgarske, line 26: Please don't raise a warning here, just force define `WindowsCE` and remove your NEED_WINDOWS_CE and warnings.

```
#undef WindowsCE
#define WindowsCE
```

wolfssl/wolfio.h
```diff
@@ -50,6 +50,23 @@
     #endif
 #endif
 
+/* CE Not always reliably detected. Define your own WindowsCE as needed */
+#if defined(_WIN32_WCE) || defined(WINCE) || defined(PocketPC) || \
+   (defined(_MSC_VER) && (_MSC_VER < 1600))
+    #if !defined(WindowsCE)
```
dgarske: Again just force define WindowsCE, but warn

```
2>wolfio.c
2>Warning WindowsCE should be defined in your user_settings.h file AND project
```

wrapper/CSharp/wolfSSL_CSharp/wolfSSL.cs
```diff
@@ -2641,17 +2886,57 @@ public static void Debugging_OFF()
         /// <summary>
         /// Set the function to use for logging
         /// </summary>
-        /// <param name="input">Function that conforms as to loggingCb</param>
+        /// <param name="input">Function that conforms as to loggingCb StringBuilder msg</param>
```
dgarske: Breaks compatibility. Please provide support for both versions...

```
2>x509_str.c
WOLFSSL\WolfSSLClient.cs(85,13): error CS1502: The best overloaded method match for 'wolfSSL.CSharp.wolfssl.SetLogging(wolfSSL.CSharp.wolfssl.loggingCb)' has some invalid arguments
WOLFSSL\WolfSSLClient.cs(85,32): error CS1503: Argument '1': cannot convert from 'method group' to 'wolfSSL.CSharp.wolfssl.loggingCb'
```

dgarske: Example calling code:

```
/// <summary>
/// Example of a logging function
/// </summary>
/// <param name="lvl">level of log</param>
/// <param name="msg">message to log</param>
public static void standard_log(int lvl, string msg)
{
    string logMsg = wolfssl.MultiByteToWideChar(msg);
    Console.WriteLine(logMsg);
}

// example of function used for setting logging
Console.WriteLine("Setting Logging");
wolfssl.SetLogging(standard_log);
```

dgarske (changes requested):

wrapper/CSharp/wolfSSL_CSharp/wolfSSL.cs
gojimmypi: Hi @dgarske - there's currently an inconsistency in `loggingCb` parameter types between [WindowsCE](https://github.com/wolfSSL/wolfssl/blob/6be8a3710d18f211bcd5fa53e5715d4db2455fd5/wrapper/CSharp/wolfSSL_CSharp/wolfSSL.cs#L578) `string msg`:

```
    public delegate void loggingCb(int lvl, string msg);
```

And [non-WindowsCE:](https://github.com/wolfSSL/wolfssl/blob/6be8a3710d18f211bcd5fa53e5715d4db2455fd5/wrapper/CSharp/wolfSSL_CSharp/wolfSSL.cs#L591) `StringBuilder msg`

```
    public delegate void loggingCb(int lvl, StringBuilder msg);
```

Three options:

1) Can I interest you in a change to your calling parameter type?

```
    public static void standard_log(int lvl, StringBuilder msg)
    {
        string logMsg = wolfssl.MultiByteToWideChar(msg.ToString());
        Console.WriteLine(logMsg);
    }
```

2) Alternatively: Given a choice, I'd prefer to convert them all to  `string msg`. I didn't think the `StringBuilder msg` parameter was ideal, and kept it only for compatibility.  This option to change all to `string msg` would allow your example to work, but potentially break unknown other non-CE uses. 

3) We could keep the different types, but I'd also prefer there be the _same_ types for CE or not, allowing maximum code compatibility when upgrade, sharing libraries, etc.

Which do you recommend?

dgarske: Keep different types please.

wrapper/CSharp/wolfSSL_CSharp/X509.cs
gojimmypi: Revised as suggested in most recent commit

wolfssl/wolfio.h
gojimmypi: Revised as suggested in most recent commit

wrapper/CSharp/wolfSSL_CSharp/wolfSSL.cs
gojimmypi: Hi @dgarske - 

Different types it is. I just pushed update commit https://github.com/wolfSSL/wolfssl/pull/8946/commits/e5d70faa6f7b11a8e143c262da5d048f34af67f5 for that.

I also added more explicit cleanup when `SetLogging()` called repeatedly. See new `ClearLoggingCallbacks()`

Please retest on customer hardware, thank you.

wolfssl/wolfio.h
gojimmypi: > Again just force define WindowsCE, but warn

@dgarske I first read this as "_same as other comment, no warn_" 

When re-reading - shall I force  `WindowsCE` in `wolfio.c` _and_ warn? (just for C code, not .cs files?)

dgarske: Is there a reason you added it to the .h files? I don't want to introduce a new warning. Customers who are using us on those targets are already setting the right macros. If its not required please remove it altogether.

dgarske: The `WindowsCE` never existed before in the .c/.h world, so please remove it

IDE/WIN/user_settings.h
```diff
@@ -6,6 +6,10 @@
 #error This user_settings.h header is only designed for Windows
 #endif
 
+/* Optional WindowsCE, needed here for all CE configurations:
```
dgarske: Remove this

wolfssl/wolfio.h
```diff
@@ -920,7 +927,11 @@ WOLFSSL_API void wolfSSL_SetIOWriteFlags(WOLFSSL* ssl, int flags);
         #if defined(__MINGW64__) && !defined(UNICODE)
             #define XINET_PTON(a,b,c)   InetPton((a),(b),(c))
         #else
-            #define XINET_PTON(a,b,c)   InetPton((a),(PCWSTR)(b),(c))
+            #if (defined(_MSC_VER) && (_MSC_VER >= 1600)) || defined(WindowsCE)
```
dgarske: Remove `|| defined(WindowsCE)` or use the existing `_WIN32_WCE` instead...

wrapper/CSharp/wolfCrypt-Test/wolfCrypt-Test.cs
```diff
@@ -1,928 +1,928 @@
...
-            Console.WriteLine("\nAll tests completed successfully");
-        }
-        catch (Exception ex)
-        {
-            Console.WriteLine($"An error occurred: {ex.Message}");
-            Environment.Exit(-1);
-        }
-    }
-}
+/* wolfCrypt-Test.cs
```
dgarske: This file shows a full replace. Guessing a CRLF issue. Please try and avoid that change if possible.

wrapper/CSharp/wolfSSL-TLS-Client/wolfSSL-TLS-Client.cs
```diff
@@ -188,7 +268,18 @@ public static void Main(string[] args)
...
         try
         {
-            tcp.Connect("localhost", 11111);
+#if WindowsCE
+            IPAddress[] addresses = Dns.GetHostEntry(SERVER_NAME).AddressList;
+            foreach (IPAddress addr in addresses)
+            {
+                if (addr.AddressFamily == AddressFamily.InterNetwork)
+                {
+                    tcp.Connect(new IPEndPoint(addr, SERVER_PORT));
```
dgarske, line 313: Why is this new logic needed? Where is the `break;`?

dgarske (changes requested):

wrapper/CSharp/wolfSSL-TLS-Client/wolfSSL-TLS-Client.cs
```diff
@@ -238,6 +333,8 @@ public static void Main(string[] args)
         Console.WriteLine("SSL version is " + wolfssl.get_version(ssl));
         Console.WriteLine("SSL cipher suite is " + wolfssl.get_current_cipher(ssl));
 
+        /* Optional code & level history */
+        show_alert_history(ssl);
```
dgarske: This belongs at the end before cleanup.

wrapper/CSharp/wolfSSL-TLS-Client/wolfSSL-TLS-Client.cs
```diff
@@ -29,16 +38,70 @@
...
     /// </summary>
     /// <param name="lvl">level of log</param>
     /// <param name="msg">message to log</param>
     public static void standard_log(int lvl, StringBuilder msg)
     {
         Console.WriteLine(msg);
     }
+#endif
+
+    public static void show_alert_history_code(WOLFSSL_ALERT h, string m)
```
dgarske: Make show_alert_history and show_alert_history_code private.

wrapper/CSharp/wolfSSL_CSharp/wolfSSL.cs
```diff
@@ -20,22 +20,180 @@
...
 using System.Runtime.InteropServices;
 using System.Text;
-using System.Threading;
 using System.IO;
 using System.Net;
 using System.Net.Sockets;
 
 namespace wolfSSL.CSharp
 {
+    /********************************
```
dgarske: Shouldn't these be part of the `public class wolfssl`? I'd prefer use wolfssl.WOLFSSL_ALERT, etc in the calling code.

wrapper/CSharp/wolfSSL-TLS-Client/wolfSSL-TLS-Client.cs
```diff
@@ -29,16 +38,70 @@
...
+    {
+        /* VS initializes .code and .level to zero; wolfSSL sets to -1 until there's a valid value. */
+        if ((h.code > 0) || (h.level > 0)) {
```
dgarske, line 63: Please cleanup example. You have `if ((h.code > 0) || (h.level > 0))` twice.

dgarske (changes requested):

IDE/WIN/user_settings.h
gojimmypi: Done

wolfssl/wolfio.h
gojimmypi: Without `WindowsCE`, the `_WIN32_WCE ` is not automatically assigned for me, and I see these compile-time warnings:

```
2>c:\workspace\wolfssl-gojimmypi\src\ssl.c(1928) : warning C4133: 'function' : incompatible types - from 'PCSTR' to 'PCWSTR'
2>c:\workspace\wolfssl-gojimmypi\src\ssl.c(22180) : warning C4133: 'function' : incompatible types - from 'const char *' to 'PCWSTR'
```

I've removed the `WindowsCE`  and added explicit known CE macros:

```
            #if (defined(_MSC_VER) && (_MSC_VER >= 1600)) || \
                 defined(_WIN32_WCE) || defined(WINCE) || defined(PocketPC)
                #define XINET_PTON(a,b,c)   InetPton((a),(PCWSTR)(b),(c))
            #else
                #define XINET_PTON(a,b,c)   InetPton((a),(PCSTR)(b),(c))
            #endif
```

wrapper/CSharp/wolfCrypt-Test/wolfCrypt-Test.cs
gojimmypi: Yes, the prior file was LF only. Windows / Visual Studio default is CRLF.

I'll revert for this PR, but I recommend we change line endings at some future time.

wrapper/CSharp/wolfSSL-TLS-Client/wolfSSL-TLS-Client.cs
gojimmypi: I've been still trying to get my CE emulator to work and use the wolfssl-tls-client example.

> Why is this new logic needed?

Because this:

```
tcp.Connect(SERVER_NAME, SERVER_PORT);
```
on WindowsCE gives this error (`connect` takes only 1 param on CE, 2 on others):

```
Error	1	No overload for method 'Connect' takes '2' arguments	C:\workspace\wolfssl-gojimmypi\wrapper\CSharp\wolfSSL-TLS-Client\wolfSSL-TLS-Client.cs	281	13	wolfSSL-TLS_CE_2008
```

>  Where is the break;?

added

wrapper/CSharp/wolfSSL-TLS-Client/wolfSSL-TLS-Client.cs
gojimmypi: Moved

wrapper/CSharp/wolfSSL-TLS-Client/wolfSSL-TLS-Client.cs
gojimmypi: Done

wrapper/CSharp/wolfSSL-TLS-Client/wolfSSL-TLS-Client.cs
gojimmypi: Good catch, changed.

wrapper/CSharp/wolfCrypt-Test/wolfCrypt-Test.cs
dgarske: Are you saying this file specifically is different? If so go ahead and keep it.

gojimmypi:
Jenkins retest this please

For ERROR: Unable to tear down: java.io.IOException: Unexpected EOF; Python Remote call to wolf-linux-cloud-node no workspace for PRB-python-port #8458

wolfssl/wolfio.h
```diff
@@ -50,6 +50,12 @@
     #endif
 #endif
 
+/* CE Not always reliably detected. Define our own _WIN32_WCE as needed. */
+#if defined(_WIN32_WCE) || defined(WINCE) || defined(PocketPC)
+    #if !_WIN32_WCE
```
dgarske: Please change this to `#if !defined(_WIN32_WCE)`. Why add another macro `WINCE` or `PocketPC` here? Are those supposed to supplied by the compiler or optionally by the user?

wolfssl/wolfio.h
```diff
@@ -920,7 +926,12 @@ WOLFSSL_API void wolfSSL_SetIOWriteFlags(WOLFSSL* ssl, int flags);
         #if defined(__MINGW64__) && !defined(UNICODE)
             #define XINET_PTON(a,b,c)   InetPton((a),(b),(c))
         #else
-            #define XINET_PTON(a,b,c)   InetPton((a),(PCWSTR)(b),(c))
+            #if (defined(_MSC_VER) && (_MSC_VER >= 1600)) || \
+                 defined(_WIN32_WCE) || defined(WINCE) || defined(PocketPC)
```
dgarske: Just use `_WIN32_WCE`. You don't need WINCE or PocketPC

wrapper/CSharp/wolfSSL_CSharp/wolfCrypt.cs
```diff
@@ -511,38 +532,11 @@ public class wolfcrypt
 #if WindowsCE
         [DllImport(wolfssl_dll)]
         private extern static IntPtr wc_GetErrorString(int error);
-        public delegate void loggingCb(int lvl, string msg);
+        /* No Windows CE decorator for logging call-back */
```
dgarske, line 598: I think this could break some users. I would prefer the logging continue to be duplicated. Technically its in wolfCrypt not wolfSSL. I don't want a refactor on this. Please duplicate the logging.

dgarske (changes requested):

wolfssl/wolfio.h
gojimmypi: I will change this to only `_WIN32_WCE`. Neither that, nor `WINCE` have ever been auto-defined for me. I added them for likely completeness, although never actually tested.

edit: I'll remove this entire section and use only `_WIN32_WCE` for `XINET_PTON` detection of `PCWSTR` in this file.

```
    #elif defined(USE_WINDOWS_API) /* Windows-friendly definition */
        #if defined(__MINGW64__) && !defined(UNICODE)
            #define XINET_PTON(a,b,c)   InetPton((a),(b),(c))
        #else
            #if (defined(_MSC_VER) && (_MSC_VER >= 1600)) || defined(_WIN32_WCE)
                #define XINET_PTON(a,b,c)   InetPton((a),(PCWSTR)(b),(c))
            #else
                #define XINET_PTON(a,b,c)   InetPton((a),(PCSTR)(b),(c))
            #endif
        #endif
```

The `PocketPC` was auto-added at new project time here. I manually added the WindowsCE.

image

wolfssl/wolfio.h
gojimmypi: Ok, I'll use only `_WIN32_WCE`. It may need to be manually defined for some projects.

dgarske: No, your logic a few lines up makes sure _WIN32_WCE is defined... So you can remove this duplicate logic.

wrapper/CSharp/wolfCrypt-Test/wolfCrypt-Test.cs
gojimmypi: Correct. All of the other examples are CRLF. I'll change the `wolfCrypt-Test.cs` from LF to CRLF.

wrapper/CSharp/wolfSSL_CSharp/wolfSSL.cs
gojimmypi: > Shouldn't these be part of the public class wolfssl? 

Yes, it was at one point, but there was a scoping problem.  I'll see if I can move it back or document why not.

wrapper/CSharp/wolfSSL_CSharp/wolfCrypt.cs
gojimmypi: Ok. Do you have an example of what might be problematic & why the duplication is preferred? 

I'd like to include a comment to avoid a future de-dupe.

gojimmypi:
@dgarske you removed the `For This Release` tag today; If I have more time on this PR, would you like the other examples cleaned up and expanded (e.g. add new alert history)? I also have some new examples that are missing, such as `wolfSSL-DTLS-PSK-Client`.

We may also consider _moving_ some of the C# examples to https://github.com/wolfSSL/wolfssl-examples

wrapper/CSharp/wolfSSL_CSharp/wolfCrypt.cs
dgarske: WolfCrypt only users who are already calling wolfcrypt.setLogging

gojimmypi: Hm. Interesting. 

What do you think of moving it all from wolfSSL to wolfcrypt, no duplicating, but including API references to maintain current functionality?

dgarske: That’s fine. As long as we keep the existing API’s. I do not want to break compatibility

wolfssl/wolfio.h
gojimmypi: @dgarske what do you think of revising the [Windows API logic](https://github.com/wolfSSL/wolfssl/blob/fb691fac94f079170a7fd90b75bb1cffa2a5846c/wolfssl/wolfio.h#L919) and first check for `UNICODE` like this:

```
#ifndef XINET_PTON
    #if defined(__WATCOMC__)
        #if defined(__OS2__) || defined(__NT__) && \
                (NTDDI_VERSION >= NTDDI_VISTA)
            #define XINET_PTON(a,b,c)   inet_pton((a),(b),(c))
        #else
            #define XINET_PTON(a,b,c)   *(unsigned *)(c) = inet_addr((b))
        #endif
    #elif defined(USE_WINDOWS_API) /* Windows-friendly definition */
        #if defined(UNICODE)
            /* Use Win API Pointer to a constant wide-character string */
            #define XINET_PTON(a,b,c)   InetPton((a),(PCWSTR)(b),(c))
        #elif defined(__MINGW64__) && !defined(UNICODE)
            #define XINET_PTON(a,b,c)   InetPton((a),(b),(c))
        #else
            #if (defined(_MSC_VER)  && (_MSC_VER >= 1600)) || defined(_WIN32_WCE)
                #define XINET_PTON(a,b,c)   InetPton((a),(PCWSTR)(b),(c))
            #else
                #define XINET_PTON(a,b,c)   InetPton((a),(PCSTR)(b),(c))
            #endif
        #endif
    #else
        #define XINET_PTON(a,b,c)   inet_pton((a),(b),(c))
    #endif
#endif
```

PR #8948, opened 2025-07-01
Disable MD5 by default for cmake builds

AlexLanzano:
# Description

Disables MD5 by default for builds using cmake. This will also disable the use of old TLS by default as well


AlexLanzano:
Retest this please.

PRB-dtls.txt_33 failed on last stage due to not receiving the test results

JacobBarthelmeh (approved):

PR #8949, opened 2025-07-01
wc/asn: fix several inconsistent function prototype parameter names

holtrop:
# Description

wc/asn: fix several inconsistent function prototype parameter names

# Testing

CI build only; only function parameter names in prototypes changed to be consistent with definitions

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/doc/dox_comments/header_files/asn_public.h
+++ b/doc/dox_comments/header_files/asn_public.h
@@ -1192,7 +1192,7 @@ int wc_PemCertToDer(const char* fileName, unsigned char* derBuf, int derSz);
     \sa wc_PemCertToDer
 */
 int wc_DerToPem(const byte* der, word32 derSz, byte* output,
-                                word32 outputSz, int type);
+                                word32 outSz, int type);
 
 /*!
     \ingroup ASN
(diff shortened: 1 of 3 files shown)


dgarske (approved):
It would be nice to have some CI (even Devin) to catch these going forward. What tool did you use to catch these?

holtrop:
> It would be nice to have some CI (even Devin) to catch these going forward. What tool did you use to catch these?

I noticed a couple manually while reviewing the API for my project. After I fixed a couple that I had noticed I figured I may as well fix all of them for asn.c to make a PR so I ran:

```
clang-tidy -checks='-*,readability-inconsistent-declaration-parameter-name' wolfcrypt/src/asn.c -- -I.
```

It found several more which I fixed.

I think it would be great to run this in an automated fashion across our entire code base but it might require a lot of cleanup first!

Also, unfortunately due to the way we have redundant prototypes in the doc/dox_comments directory, I don't think it would find issues in any of those. Perhaps we could run two passes, one where we use the regular header files and one where we use the doc/dox_comments/header_files. I didn't try this...

holtrop:

PR #8963, opened 2025-07-03
CodeQL: o is larger type and could cause issues

### LINKED ISSUE #8958: CWE-190: wolfssl/wolfcrypt/src/sp_int.c:7616
### Version

5.8.0

### Description

CodeQL flagged this with CWE-190, CWE-197, CWE-835.

>Comparison of narrow type with wide type in loop condition:

>In a loop condition, comparison of a value of a narrow type with a value of a wide type may result in unexpected behavior if the wider value is sufficiently large (or small). This is because the narrower value may overflow. This can lead to an infinite loop.

libs/wolfssl/wolfcrypt/src/sp_int.c:7616: Comparison between i of type sp_size_t and o of wider type unsigned int.
```
   7614     /* Need to copy digits up to offset into result. */
   7615     if (r != a) {
   7616         for (; (i < o) && (i < a->used); i++) {
   7617             r->dp[i] = a->dp[i];
   7618         }
   7619     }
```

Not sure if this is causing any issue in this case? But technically finding seems correct, if _sp_sub_off() is ever called with o parameter set to >= 2^16, then comparison "i < o" will always be true regardless value of i...

SparkiDev:
# Description

Make 'o' sp_size_t as the callers are passing 0 or explicit cast to sp_size_t

Fixes #8958

# Testing

Standard

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/wolfcrypt/src/sp_int.c
+++ b/wolfcrypt/src/sp_int.c
@@ -7600,7 +7600,7 @@ static void _sp_add_off(const sp_int* a, const sp_int* b, sp_int* r, int o)
  * @param  [in]   o  Number of digits to offset b.
  */
 static void _sp_sub_off(const sp_int* a, const sp_int* b, sp_int* r,
-    unsigned int o)
+    sp_size_t o)
 {
     sp_size_t i = 0;
     sp_size_t j;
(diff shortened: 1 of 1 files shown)


SparkiDev:
retest this please

dgarske (approved):

PR #8965, opened 2025-07-04
PPC32 ARM ASM SHA-256: SPE impl, tidy up original

SparkiDev:
# Description

Implement using SPE instructions that allow for 64-bit registers as a vector of 2 32-bit values.
Tidy up original implementation to not use stack.

# Testing

./configure --disable-shared LDFLAGS=--static --host=powerpc 'CC=powerpc-linux-gnu-gcc -mcpu=8540' --enable-ppc32-asm=spe
./configure --disable-shared LDFLAGS=--static --host=powerpc 'CC=powerpc-linux-gnu-gcc -mcpu=8540' --enable-ppc32-asm=inline,spe
(Had to compile a special QEMU that allows a CPU with hard-float to support SPE instructions as I don't have soft-float libc.)

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


SparkiDev:
> 
> * `no-void-functions` [snippet](https://github.com/wolfSSL/wolfssl/blob/f7e2cd1a3348af8c391823f4f8e6cf48b83e97d5/wolfcrypt/src/port/ppc32/ppc32-sha256-asm_c.c#L69): Change `void Transform_Sha256_Len( … )` to return an `int` (or other appropriate status type) and ensure every exit path returns 0 on success or a negative error code on failure, propagating that value to the caller.
> 

All versions of Transform_Sha256_Len are void return.

SparkiDev:
retest this please

SparkiDev:
retest this please

dgarske (approved):

PR #8967, opened 2025-07-04
Store in extensions the full octet string

### LINKED ISSUE #8941: Get full OCTET STRING of an extension
### Version

5.7.6-stable

### Description

RFC5280 defines an X.509 extension as
~~~asn1
   Extension  ::=  SEQUENCE  {
        extnID      OBJECT IDENTIFIER,
        critical    BOOLEAN DEFAULT FALSE,
        extnValue   OCTET STRING
                    -- contains the DER encoding of an ASN.1 value
                    -- corresponding to the extension type identified
                    -- by extnID
        }
~~~

OpenSSL's function `X509_EXTENSION_get_data` returns the full `extnValue   OCTET STRING`.
But WolfSSL's function `wolfSSL_X509_EXTENSION_get_data` doesn't seem to get the full `OCTET STRING`, but some data specific for each extension.

For example, for `AuthorityKeyIdentifier`, which is defined as
~~~asn1
   AuthorityKeyIdentifier ::= SEQUENCE {
      keyIdentifier             [0] KeyIdentifier           OPTIONAL,
      authorityCertIssuer       [1] GeneralNames            OPTIONAL,
      authorityCertSerialNumber [2] CertificateSerialNumber OPTIONAL  }
~~~
it directly gives the `keyIdentifier` array (in my case `{0x3d, 0x1, 0xd3, 0x87,...`) instead of the SEQUENCE, which should start with `{0x30, 0x16, 0x80, 0x14, 0x3d, 0x1, 0xd3, 0x87,...`

`wolfSSL_X509_EXTENSION_get_data` is aliased to `X509_EXTENSION_get_data`, so I guess the intention is for them to have the same behavior. If they are not, is there a way to get the full `extnValue   OCTET STRING`?

ribes96:
Store in WOLFSSL_X509_EXTENSION.value always the full contents of the OCTET STRING of the extension, instead of different type of data depending on the type of extension. Previously this was only done for unknown extensions.

# Description

It is clear from Issue #8941 that `X509_EXTENSION_get_data` has a clear different behavior on OpenSSL than in WolfSSL. [RFC8280](https://www.rfc-editor.org/rfc/rfc5280#appendix-A.1) defines an extension as:

~~~asn1
Extension  ::=  SEQUENCE  {
     extnID      OBJECT IDENTIFIER,
     critical    BOOLEAN DEFAULT FALSE,
     extnValue   OCTET STRING
                 -- contains the DER encoding of an ASN.1 value
                 -- corresponding to the extension type identified
                 -- by extnID
     }
~~~

Whereas OpenSSL always returns the full OCTET STRING when calling `X509_EXTENSION_get_data`, WolfSSL returns a different type of object depending on the type of extension parsed. This can be observed in the following program:

~~~c
#include <stdio.h>
#include <err.h>

#ifdef USE_WOLFSSL

#define WOLFSSL_USE_OPTIONS_H

#include <wolfssl/ssl.h>
#include <wolfssl/openssl/pem.h>
#include <wolfssl/openssl/ssl.h>
#include <wolfssl/openssl/buffer.h>

#endif /* USE_WOLFSSL */

#ifdef USE_OPENSSL

#include <openssl/x509.h>
#include <openssl/pem.h>
#include <openssl/err.h>

#endif /* USE_OPENSSL */

int main (int argc, char **argv)
{
	const char     *filename;
	FILE           *f;
	X509           *x509;
	int            n_extensions;
	int            i;

	if (argc < 2)
		errx (-1, "Error: Usage: <filename>");

	filename = argv[1];

	if ((f = fopen (filename, "r")) == NULL)
		err (-1, "%s", filename);

	if ((x509 = PEM_read_X509 (f, NULL, NULL, NULL)) == NULL)
		err (-1, NULL);

	fclose (f);

	n_extensions = X509_get_ext_count (x509);
	for (i = 0; i < n_extensions; ++i)
	{
		X509_EXTENSION         *ext;
		ASN1_OBJECT            *object;
		char                   oid[32];
		int                    j;
		#ifdef USE_WOLFSSL
		ASN1_STRING            *data;
		#endif /* USE_WOLFSSL */

		#ifdef USE_OPENSSL
		ASN1_OCTET_STRING      *data;
		#endif /* USE_OPENSSL */

		if ((ext = X509_get_ext (x509, i)) == NULL)
			errx (-1, "Error: No extension at %d\n", i);

		if ((object = X509_EXTENSION_get_object (ext)) == NULL)
			errx (-1, "Error: No object for extension at %d\n", i);

		OBJ_obj2txt (oid, sizeof (oid), object, 1);

		if ((data = X509_EXTENSION_get_data (ext)) == NULL)
			errx (-1, "Error: No data for extension at %d\n", i);

		printf ("i=%d\n", i);
		printf ("oid=%s\n", oid);
		printf ("data->length=%d\n", data->length);
		printf ("data->data={");
		for (j = 0; j < data->length; ++j)
			printf ("%02X:", (unsigned char) data->data[j]);
		printf ("}\n");
		printf ("\n");
	}


	X509_free (x509);
	return 0;
}
~~~

For example with this extension

~~~openssl
-----BEGIN CERTIFICATE-----
MIIFYjCCBEqgAwIBAgIQd70NbNs2+RrqIQ/E8FjTDTANBgkqhkiG9w0BAQsFADBX
MQswCQYDVQQGEwJCRTEZMBcGA1UEChMQR2xvYmFsU2lnbiBudi1zYTEQMA4GA1UE
CxMHUm9vdCBDQTEbMBkGA1UEAxMSR2xvYmFsU2lnbiBSb290IENBMB4XDTIwMDYx
OTAwMDA0MloXDTI4MDEyODAwMDA0MlowRzELMAkGA1UEBhMCVVMxIjAgBgNVBAoT
GUdvb2dsZSBUcnVzdCBTZXJ2aWNlcyBMTEMxFDASBgNVBAMTC0dUUyBSb290IFIx
MIICIjANBgkqhkiG9w0BAQEFAAOCAg8AMIICCgKCAgEAthECix7joXebO9y/lD63
ladAPKH9gvl9MgaCcfb2jH/76Nu8ai6Xl6OMS/kr9rH5zoQdsfnFl97vufKj6bwS
iV6nqlKr+CMny6SxnGPb15l+8Ape62im9MZaRw1NEDPjTrETo8gYbEvs/AmQ351k
KSUjB6G00j0uYODP0gmHu81I8E3CwnqIiru6z1kZ1q+PsAewnjHxgsHA3y6mbWwZ
DrXYfiYaRQM9sHmklCitD38m5agI/pboPGiUU+6DOogrFZYJsuB6jC511pzrp1Zk
j5ZPaK49l8KEj8C8QMALXL32h7M1bKwYUH+E4EzNktMg6TO8UpmvMrUpsyUqtEj5
cuHKZPfmghCN6J3Cioj6OGaK/GP5Afl4/Xtcd/p2h/rs37EOeZVXtL0m79YB0esW
CruOC7XFxYpVq9Os6pFLKcwZpDIlTirxZUTQAs6qzkm06p98g7BAe+dDq6dso499
iYH6TKX/1Y7DzkvgtdizjkXPdsDtQCv9Uw+wp9U7DbGKogPeMa3Md+pvez7W35Ei
Eua++tgy/BBjFFFy3l3WFpO9KWgz7zpm7AeKJt8T11dleCfeXkkUAKIAf5qoIbap
sZWwpbkNFhHax2xIPEDgfg1azVY80ZcFuctL7TlLnMQ/0lUTbiSw1nH69MG6zO0b
9f6BQdgAmD06yK56mDcYBZUCAwEAAaOCATgwggE0MA4GA1UdDwEB/wQEAwIBhjAP
BgNVHRMBAf8EBTADAQH/MB0GA1UdDgQWBBTkrysmcRorSCeFL1JmLO/wiRNxPjAf
BgNVHSMEGDAWgBRge2YaRQ2XyolQL30EzTSo//z9SzBgBggrBgEFBQcBAQRUMFIw
JQYIKwYBBQUHMAGGGWh0dHA6Ly9vY3NwLnBraS5nb29nL2dzcjEwKQYIKwYBBQUH
MAKGHWh0dHA6Ly9wa2kuZ29vZy9nc3IxL2dzcjEuY3J0MDIGA1UdHwQrMCkwJ6Al
oCOGIWh0dHA6Ly9jcmwucGtpLmdvb2cvZ3NyMS9nc3IxLmNybDA7BgNVHSAENDAy
MAgGBmeBDAECATAIBgZngQwBAgIwDQYLKwYBBAHWeQIFAwIwDQYLKwYBBAHWeQIF
AwMwDQYJKoZIhvcNAQELBQADggEBADSkHrEoo9C0dhemMXoh6dFSPsjbdBZBiLg9
NR3t5P+T4Vxfq7vqfM/b5A3Ri1fyJm9bvhdGaJQ3b2t6yMAYN/olUazsaL+yyEn9
WprKASOshIArAoyZl+tJaox118fessmXn1hIVw41oeQa1v1vg4Fv74zPl6/AhSrw
9U5pCZEt4Wi4wStz6dTZ/CLANx8LZh1J7QJVj2fhMtfTJr9w4z30Z209fOU0iOMy
+qduBmpvvYuR7hZL6Dupszfnw0Skfths18dG9ZKb59UhvmaSGZRVbNQpsg3BZlvi
d0lIKO2d1xozclOzgjXPYovJJIultzkMu34qQb9Sz/yilrbCgj8=
-----END CERTIFICATE-----
~~~

Given that `wolfSSL_X509_EXTENSION_get_data` is currently aliased to `X509_EXTENSION_get_data`, I would say the intention is for them to have the same behaviour, so this PR changes how an extension is stored in WolfSSL.

Now, this is a clear change of behavior, and will probably break a lot of tests. To start, the API of the public function would need to be changed to return an `ASN1_OCTET_STRING` instead of an `ASN1_STRING`. But I don't really see a utility for a function that returns different type of data depending on the type of extension, other than to print it in a console.

So, this PR is not really intended to be merged yet, but to trigger a discussion about how should WolfSSL behave. If it is acknowledged that they should **not** have the same behavior, I think the alias to OpenSSL alternatives should be removed.

The storage format of an X.509 extension is changed

Fixes #8941 

# Testing

How did you test?

No testing other than running the attached program

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/src/x509.c
+++ b/src/x509.c
@@ -602,136 +602,6 @@ err:
 #endif /* OPENSSL_ALL || WOLFSSL_WPAS_SMALL */
 
 #if defined(OPENSSL_ALL) || defined(OPENSSL_EXTRA)
-static int DNS_to_GENERAL_NAME(WOLFSSL_GENERAL_NAME* gn, DNS_entry* dns)
-{
-    gn->type = dns->type;
-    switch (gn->type) {
-        case WOLFSSL_GEN_OTHERNAME:
-                if (!wolfssl_dns_entry_othername_to_gn(dns, gn)) {
-                    WOLFSSL_MSG("OTHERNAME set failed");
-                    return WOLFSSL_FAILURE;
-                }
-            break;
-
-        case WOLFSSL_GEN_EMAIL:
-        case WOLFSSL_GEN_DNS:
-        case WOLFSSL_GEN_URI:
-        case WOLFSSL_GEN_IPADD:
-        case WOLFSSL_GEN_IA5:
-                gn->d.ia5->length = dns->len;
-                if (wolfSSL_ASN1_STRING_set(gn->d.ia5, dns->name,
-                        gn->d.ia5->length) != WOLFSSL_SUCCESS) {
-                    WOLFSSL_MSG("ASN1_STRING_set failed");
-                    return WOLFSSL_FAILURE;
-                }
-                break;
-
-
-        case WOLFSSL_GEN_DIRNAME:
-            /* wolfSSL_GENERAL_NAME_new() mallocs this by default */
-            wolfSSL_ASN1_STRING_free(gn->d.ia5);
-            gn->d.ia5 = NULL;
-
-            gn->d.dirn = wolfSSL_X509_NAME_new();;
-            /* @TODO extract dir name info from DNS_entry */
-            break;
-
-#ifdef WOLFSSL_RID_ALT_NAME
-        case WOLFSSL_GEN_RID:
-            /* wolfSSL_GENERAL_NAME_new() mallocs this by default */
-            wolfSSL_ASN1_STRING_free(gn->d.ia5);
-            gn->d.ia5 = NULL;
-
-            gn->d.registeredID = wolfSSL_ASN1_OBJECT_new();
-            if (gn->d.registeredID == NULL) {
-                return WOLFSSL_FAILURE;
-            }
-            gn->d.registeredID->obj = (const unsigned char*)XMALLOC(dns->len,
-                gn->d.registeredID->heap, DYNAMIC_TYPE_ASN1);
-            if (gn->d.registeredID->obj == NULL) {
-                /* registeredID gets free'd up by caller after failure */
-                return WOLFSSL_FAILURE;
-            }
-            gn->d.registeredID->dynamic |= WOLFSSL_ASN1_DYNAMIC_DATA;
-            XMEMCPY((byte*)gn->d.registeredID->obj, dns->ridString, dns->len);
-            gn->d.registeredID->objSz = dns->len;
-            gn->d.registeredID->grp = oidCertExtType;
-            gn->d.registeredID->nid = WC_NID_registeredAddress;
-            break;
-#endif
-
-        case WOLFSSL_GEN_X400:
-            /* Unsupported: fall through */
-        case WOLFSSL_GEN_EDIPARTY:
-            /* Unsupported: fall through */
-        default:
-            WOLFSSL_MSG("Unsupported type conversion");
-            return WOLFSSL_FAILURE;
-    }
-    return WOLFSSL_SUCCESS;
-}
-
-
-static int wolfssl_x509_alt_names_to_gn(WOLFSSL_X509* x509,
-    WOLFSSL_X509_EXTENSION* ext)
-{
-    int ret = 0;
-    WOLFSSL_GENERAL_NAME* gn = NULL;
-    DNS_entry* dns = NULL;
-    WOLFSSL_STACK* sk;
-
-#ifdef OPENSSL_ALL
-    ret = wolfSSL_ASN1_STRING_set(&ext->value, x509->subjAltNameSrc,
-              x509->subjAltNameSz);
-    if (ret != WOLFSSL_SUCCESS) {
-        WOLFSSL_MSG("ASN1_STRING_set() failed");
-        goto err;
-    }
-#endif
-
-    sk = (WOLFSSL_GENERAL_NAMES*)XMALLOC(sizeof(WOLFSSL_GENERAL_NAMES), NULL,
-        DYNAMIC_TYPE_ASN1);
-    if (sk == NULL) {
-        goto err;
-    }
-    XMEMSET(sk, 0, sizeof(WOLFSSL_GENERAL_NAMES));
-    sk->type = STACK_TYPE_GEN_NAME;
-
-    if (x509->subjAltNameSet && x509->altNames != NULL) {
-        /* alt names are DNS_entry structs */
-        dns = x509->altNames;
-        /* Currently only support GEN_DNS type */
-        while (dns != NULL) {
-            gn = wolfSSL_GENERAL_NAME_new();
-            if (gn == NULL) {
-                WOLFSSL_MSG("Error creating GENERAL_NAME");
-                wolfSSL_sk_pop_free(sk, NULL);
-                goto err;
-            }
-
-            if (DNS_to_GENERAL_NAME(gn, dns) != WOLFSSL_SUCCESS) {
-                wolfSSL_GENERAL_NAME_free(gn);
-                wolfSSL_sk_pop_free(sk, NULL);
-                goto err;
-            }
-
-            if (wolfSSL_sk_GENERAL_NAME_push(sk, gn) <= 0) {
-                WOLFSSL_MSG("Error pushing onto stack");
-                wolfSSL_GENERAL_NAME_free(gn);
-                wolfSSL_sk_pop_free(sk, NULL);
-                goto err;
-            }
-
-            dns = dns->next;
-        }
-    }
-    ext->ext_sk = sk;
-    ext->crit = x509->subjAltNameCrit;
-
-    ret = 1;
-err:
-    return ret;
-}
 
 /* Pushes a new X509_EXTENSION* ext onto the stack inside WOLFSSL_X509* x509.
  * This is currently a helper function for wolfSSL_X509_get_ext
@@ -740,14 +610,12 @@ err:
 WOLFSSL_X509_EXTENSION* wolfSSL_X509_set_ext(WOLFSSL_X509* x509, int loc)
 {
     int extCount = 0, length = 0, outSz = 0, sz = 0, ret = 0;
-    int objSz = 0, isSet = 0;
+    int objSz = 0;
     const byte* rawCert;
     const byte* input;
     byte* oidBuf;
     word32 oid, idx = 0, tmpIdx = 0, nid;
     WOLFSSL_X509_EXTENSION* ext = NULL;
-    WOLFSSL_ASN1_INTEGER* a;
-    WOLFSSL_STACK* sk;
 #ifdef WOLFSSL_SMALL_STACK
     DecodedCert* cert = NULL;
 #else
@@ -887,8 +755,6 @@ WOLFSSL_X509_EXTENSION* wolfSSL_X509_set_ext(WOLFSSL_X509* x509, int loc)
             continue;
         }
         /* extCount == loc. Now get the extension. */
-        /* Check if extension has been set */
-        isSet = wolfSSL_X509_ext_isSet_by_NID((WOLFSSL_X509*)x509, (int)nid);
 
         if (wolfSSL_OBJ_nid2ln((int)nid) != NULL) {
             /* This is NOT an unknown OID. */
@@ -908,376 +774,143 @@ WOLFSSL_X509_EXTENSION* wolfSSL_X509_set_ext(WOLFSSL_X509* x509, int loc)
             ext->obj->nid = (int)nid;
         }
 
-        switch (oid) {
-            case BASIC_CA_OID:
-            {
-                word32 dataIdx = idx;
-                word32 dummyOid;
-                int dataLen = 0;
+        /* The ASN1_OBJECT in the extension is set in the same way
+         * for recognized and for unrecognized extension types, as
+         * the full OCTET STRING */
 
-                if (!isSet)
-                    break;
-                /* Set pathlength */
-                a = wolfSSL_ASN1_INTEGER_new();
+        /* Get OID from input */
+        if (GetASNObjectId(input, &idx, &length, (word32)sz) != 0) {
+            WOLFSSL_MSG("Failed to Get ASN Object Id");
+            wolfSSL_X509_EXTENSION_free(ext);
+            FreeDecodedCert(cert);
+        #ifdef WOLFSSL_SMALL_STACK
+            XFREE(cert, NULL, DYNAMIC_TYPE_DCERT);
+        #endif
+            return NULL;
+        }
+        oidBuf = (byte*)XMALLOC(length+1+MAX_LENGTH_SZ, NULL,
+                            DYNAMIC_TYPE_TMP_BUFFER);
+        if (oidBuf == NULL) {
+            WOLFSSL_MSG("Failed to malloc tmp buffer");
+            wolfSSL_X509_EXTENSION_free(ext);
+            FreeDecodedCert(cert);
+        #ifdef WOLFSSL_SMALL_STACK
+            XFREE(cert, NULL, DYNAMIC_TYPE_DCERT);
+        #endif
+            return NULL;
+        }
+        oidBuf[0] = ASN_OBJECT_ID;
+        objSz++;
+        objSz += SetLength(length, oidBuf + 1);
+        objSz += length;
 
-                /* Set the data */
-                ret = GetObjectId(input, &dataIdx, &dummyOid, oidCertExtType,
-                        (word32)sz) == 0;
-                if (ret && dataIdx < (word32)sz) {
-                    /* Skip the critical information */
-                    if (input[dataIdx] == ASN_BOOLEAN) {
-                        dataIdx++;
-                        ret = GetLength(input, &dataIdx, &dataLen, sz) >= 0;
-                        dataIdx += dataLen;
-                    }
-                }
-                if (ret) {
-                    ret = GetOctetString(input, &dataIdx, &dataLen,
-                            (word32)sz) > 0;
-                }
-                if (ret) {
-                    ret = wolfSSL_ASN1_STRING_set(&ext->value, input + dataIdx,
-                            dataLen) == 1;
-                }
-
-                if (a == NULL || !ret) {
-                    wolfSSL_X509_EXTENSION_free(ext);
-                    FreeDecodedCert(cert);
-                #ifdef WOLFSSL_SMALL_STACK
-                    XFREE(cert, NULL, DYNAMIC_TYPE_DCERT);
-                #endif
-                    return NULL;
-                }
-                a->length = (int)x509->pathLength;
-
-                /* Save ASN1_INTEGER in x509 extension */
-                ext->obj->pathlen = a;
-
-                ext->obj->ca = x509->isCa;
-                ext->crit = x509->basicConstCrit;
-                break;
-            }
-            case AUTH_INFO_OID:
-                if (!isSet)
-                    break;
-
-                /* Create a stack to hold both the caIssuer and ocsp objects
-                    in X509_EXTENSION structure */
-                sk = wolfSSL_sk_new_asn1_obj();
-                if (sk == NULL) {
-                    WOLFSSL_MSG("Failed to malloc stack");
-                    wolfSSL_X509_EXTENSION_free(ext);
-                    FreeDecodedCert(cert);
-                #ifdef WOLFSSL_SMALL_STACK
-                    XFREE(cert, NULL, DYNAMIC_TYPE_DCERT);
-                #endif
-                    return NULL;
-                }
-
-                /* Add CaIssuers object to stack */
-                if (x509->authInfoCaIssuer != NULL &&
-                    x509->authInfoCaIssuerSz > 0)
-                {
-                    WOLFSSL_ASN1_OBJECT* obj;
-                    obj = wolfSSL_ASN1_OBJECT_new();
-                    if (obj == NULL) {
-                        WOLFSSL_MSG("Error creating ASN1 object");
-                        wolfSSL_sk_ASN1_OBJECT_pop_free(sk, NULL);
-                        wolfSSL_X509_EXTENSION_free(ext);
-                        FreeDecodedCert(cert);
-                    #ifdef WOLFSSL_SMALL_STACK
-                        XFREE(cert, NULL, DYNAMIC_TYPE_DCERT);
-                    #endif
-                        return NULL;
-                    }
-                    obj->obj = (byte*)x509->authInfoCaIssuer;
-                    obj->objSz = (unsigned int)x509->authInfoCaIssuerSz;
-                    obj->grp = oidCertAuthInfoType;
-                    obj->nid = WC_NID_ad_ca_issuers;
-
-                    ret = wolfSSL_sk_ASN1_OBJECT_push(sk, obj) > 0
-                            ? WOLFSSL_SUCCESS : WOLFSSL_FAILURE;
-                    if (ret != WOLFSSL_SUCCESS) {
-                        WOLFSSL_MSG("Error pushing ASN1 object onto stack");
-                        wolfSSL_ASN1_OBJECT_free(obj);
-                        wolfSSL_sk_ASN1_OBJECT_pop_free(sk, NULL);
-                        wolfSSL_X509_EXTENSION_free(ext);
-                        FreeDecodedCert(cert);
-                    #ifdef WOLFSSL_SMALL_STACK
-                        XFREE(cert, NULL, DYNAMIC_TYPE_DCERT);
-                    #endif
-                        return NULL;
-                    }
-                }
-
-                /* Add OCSP object to stack */
-                if (x509->authInfo != NULL &&
-                    x509->authInfoSz > 0)
-                {
-                    WOLFSSL_ASN1_OBJECT* obj;
-                    obj = wolfSSL_ASN1_OBJECT_new();
-                    if (obj == NULL) {
-                        WOLFSSL_MSG("Error creating ASN1 object");
-                        wolfSSL_sk_ASN1_OBJECT_pop_free(sk, NULL);
-                        wolfSSL_X509_EXTENSION_free(ext);
-                        FreeDecodedCert(cert);
-                    #ifdef WOLFSSL_SMALL_STACK
-                        XFREE(cert, NULL, DYNAMIC_TYPE_DCERT);
-                    #endif
-                        return NULL;
-                    }
-                    obj->obj = x509->authInfo;
-                    obj->objSz = (unsigned int)x509->authInfoSz;
-                    obj->grp = oidCertAuthInfoType;
-                    obj->nid = WC_NID_ad_OCSP;
-
-                    ret = wolfSSL_sk_ASN1_OBJECT_push(sk, obj) > 0
-                            ? WOLFSSL_SUCCESS : WOLFSSL_FAILURE;
-                    if (ret != WOLFSSL_SUCCESS) {
-                        WOLFSSL_MSG("Error pushing ASN1 object onto stack");
-                        wolfSSL_ASN1_OBJECT_free(obj);
-                        wolfSSL_sk_ASN1_OBJECT_pop_free(sk, NULL);
-                        wolfSSL_X509_EXTENSION_free(ext);
-                        FreeDecodedCert(cert);
-                    #ifdef WOLFSSL_SMALL_STACK
-                        XFREE(cert, NULL, DYNAMIC_TYPE_DCERT);
-                    #endif
-                        return NULL;
-                    }
-                }
-                ext->ext_sk = sk;
-                ext->crit = x509->authInfoCrit;
-                break;
-
-            case AUTH_KEY_OID:
-                if (!isSet)
-                    break;
-
-                ret = wolfSSL_ASN1_STRING_set(&ext->value, x509->authKeyId,
-                                        x509->authKeyIdSz);
-                if (ret != WOLFSSL_SUCCESS) {
-                    WOLFSSL_MSG("ASN1_STRING_set() failed");
-                    wolfSSL_X509_EXTENSION_free(ext);
-                    FreeDecodedCert(cert);
-                #ifdef WOLFSSL_SMALL_STACK
-                    XFREE(cert, NULL, DYNAMIC_TYPE_DCERT);
-                #endif
-                    return NULL;
-                }
-                ext->crit = x509->authKeyIdCrit;
-                break;
-
-            case SUBJ_KEY_OID:
-                if (!isSet)
-                    break;
-
-                ret = wolfSSL_ASN1_STRING_set(&ext->value, x509->subjKeyId,
-                                        x509->subjKeyIdSz);
-                if (ret != WOLFSSL_SUCCESS) {
-                    WOLFSSL_MSG("ASN1_STRING_set() failed");
-                    wolfSSL_X509_EXTENSION_free(ext);
-                    FreeDecodedCert(cert);
-                #ifdef WOLFSSL_SMALL_STACK
-                    XFREE(cert, NULL, DYNAMIC_TYPE_DCERT);
-                #endif
-                    return NULL;
-                }
-                ext->crit = x509->subjKeyIdCrit;
-                break;
-
-            case CERT_POLICY_OID:
-                if (!isSet)
-                    break;
-            #ifdef WOLFSSL_SEP
-                ext->crit = x509->certPolicyCrit;
-            #endif
-                break;
-
-            case KEY_USAGE_OID:
-                if (!isSet)
-                    break;
-
-                ret = wolfSSL_ASN1_STRING_set(&ext->value,
-                                  (byte*)&(x509->keyUsage), sizeof(word16));
-                if (ret != WOLFSSL_SUCCESS) {
-                    WOLFSSL_MSG("ASN1_STRING_set() failed");
-                    wolfSSL_X509_EXTENSION_free(ext);
-                    FreeDecodedCert(cert);
-                #ifdef WOLFSSL_SMALL_STACK
-                    XFREE(cert, NULL, DYNAMIC_TYPE_DCERT);
-                #endif
-                    return NULL;
-                }
-                ext->crit = x509->keyUsageCrit;
-                break;
-
-            case EXT_KEY_USAGE_OID:
-                if (!isSet)
-                    break;
-
-                ret = wolfSSL_ASN1_STRING_set(&ext->value, x509->extKeyUsageSrc,
-                                              x509->extKeyUsageSz);
-                if (ret != WOLFSSL_SUCCESS) {
-                    WOLFSSL_MSG("ASN1_STRING_set() failed");
-                    wolfSSL_X509_EXTENSION_free(ext);
-                    FreeDecodedCert(cert);
-                #ifdef WOLFSSL_SMALL_STACK
-                    XFREE(cert, NULL, DYNAMIC_TYPE_DCERT);
-                #endif
-                    return NULL;
-                }
-                ext->crit = x509->extKeyUsageCrit;
-                break;
-
-            case CRL_DIST_OID:
-                if (!isSet)
-                    break;
-                ext->crit = x509->CRLdistCrit;
-                break;
-
-            case ALT_NAMES_OID:
-                if (!isSet)
-                    break;
-                if (!wolfssl_x509_alt_names_to_gn(x509, ext)) {
-                    wolfSSL_X509_EXTENSION_free(ext);
-                    FreeDecodedCert(cert);
-                #ifdef WOLFSSL_SMALL_STACK
-                    XFREE(cert, NULL, DYNAMIC_TYPE_DCERT);
-                #endif
-                    return NULL;
-                }
-                break;
-
-            default:
-                WOLFSSL_MSG("Unknown extension type found, parsing OID");
-                /* If the extension type is not recognized/supported,
-                 *  set the ASN1_OBJECT in the extension with the
-                 *  parsed oid for access in later function calls */
-
-                /* Get OID from input */
-                if (GetASNObjectId(input, &idx, &length, (word32)sz) != 0) {
-                    WOLFSSL_MSG("Failed to Get ASN Object Id");
-                    wolfSSL_X509_EXTENSION_free(ext);
-                    FreeDecodedCert(cert);
-                #ifdef WOLFSSL_SMALL_STACK
-                    XFREE(cert, NULL, DYNAMIC_TYPE_DCERT);
-                #endif
-                    return NULL;
-                }
-                oidBuf = (byte*)XMALLOC(length+1+MAX_LENGTH_SZ, NULL,
-                                    DYNAMIC_TYPE_TMP_BUFFER);
-                if (oidBuf == NULL) {
-                    WOLFSSL_MSG("Failed to malloc tmp buffer");
-                    wolfSSL_X509_EXTENSION_free(ext);
-                    FreeDecodedCert(cert);
-                #ifdef WOLFSSL_SMALL_STACK
-                    XFREE(cert, NULL, DYNAMIC_TYPE_DCERT);
-                #endif
-                    return NULL;
-                }
-                oidBuf[0] = ASN_OBJECT_ID;
-                objSz++;
-                objSz += SetLength(length, oidBuf + 1);
-                objSz += length;
-
-                /* Set object size and reallocate space in object buffer */
-                if (ext->obj == NULL) {
-                    ext->obj = wolfSSL_ASN1_OBJECT_new();
-                    if (ext->obj == NULL) {
-                        XFREE(oidBuf, NULL, DYNAMIC_TYPE_TMP_BUFFER);
-                        wolfSSL_X509_EXTENSION_free(ext);
-                        FreeDecodedCert(cert);
-                #ifdef WOLFSSL_SMALL_STACK
-                        XFREE(cert, NULL, DYNAMIC_TYPE_DCERT);
-                #endif
-                        return NULL;
-                    }
-                }
-
-                if (((ext->obj->dynamic & WOLFSSL_ASN1_DYNAMIC_DATA) != 0) ||
-                    (ext->obj->obj == NULL)) {
-                #ifdef WOLFSSL_NO_REALLOC
-                    byte* tmp = NULL;
-
-                    tmp = (byte*)XMALLOC(objSz, NULL, DYNAMIC_TYPE_ASN1);
-                    if (tmp != NULL && ext->obj->obj != NULL) {
-                        XMEMCPY(tmp, ext->obj->obj, ext->obj->objSz);
-                        XFREE((byte*)ext->obj->obj, NULL, DYNAMIC_TYPE_ASN1);
-                    }
-                    else if (tmp == NULL) {
-                        ext->obj->obj = tmp;
-                    }
-                    ext->obj->obj = tmp;
-                #else
-                    ext->obj->obj = (byte*)XREALLOC((byte*)ext->obj->obj, objSz,
-                                           NULL, DYNAMIC_TYPE_ASN1);
-                #endif
-                    if (ext->obj->obj == NULL) {
-                        wolfSSL_X509_EXTENSION_free(ext);
-                        FreeDecodedCert(cert);
-                        XFREE(oidBuf, NULL, DYNAMIC_TYPE_TMP_BUFFER);
-                    #ifdef WOLFSSL_SMALL_STACK
-                        XFREE(cert, NULL, DYNAMIC_TYPE_DCERT);
-                    #endif
-                        return NULL;
-                    }
-                    ext->obj->dynamic |= WOLFSSL_ASN1_DYNAMIC_DATA;
-                }
-                else {
-                    ext->obj->dynamic &= ~WOLFSSL_ASN1_DYNAMIC_DATA;
-                }
-                ext->obj->objSz = (unsigned int)objSz;
-
-                /* Get OID from input and copy to ASN1_OBJECT buffer */
-                XMEMCPY(oidBuf+2, input+idx, length);
-                XMEMCPY((byte*)ext->obj->obj, oidBuf, ext->obj->objSz);
+        /* Set object size and reallocate space in object buffer */
+        if (ext->obj == NULL) {
+            ext->obj = wolfSSL_ASN1_OBJECT_new();
+            if (ext->obj == NULL) {
                 XFREE(oidBuf, NULL, DYNAMIC_TYPE_TMP_BUFFER);
-                oidBuf = NULL;
-                ext->obj->grp = oidCertExtType;
-                ext->crit = 0;
+                wolfSSL_X509_EXTENSION_free(ext);
+                FreeDecodedCert(cert);
+        #ifdef WOLFSSL_SMALL_STACK
+                XFREE(cert, NULL, DYNAMIC_TYPE_DCERT);
+        #endif
+                return NULL;
+            }
+        }
 
-                /* Get extension data and copy as ASN1_STRING */
-                tmpIdx = idx + length;
-                if ((tmpIdx >= (word32)sz) ||
-                    (input[tmpIdx] != ASN_OCTET_STRING))
-                {
-                    WOLFSSL_MSG("Error decoding unknown extension data");
-                    wolfSSL_ASN1_OBJECT_free(ext->obj);
-                    wolfSSL_X509_EXTENSION_free(ext);
-                    FreeDecodedCert(cert);
-                #ifdef WOLFSSL_SMALL_STACK
-                    XFREE(cert, NULL, DYNAMIC_TYPE_DCERT);
-                #endif
-                    return NULL;
-                }
+        ext->obj->objSz = (unsigned int)objSz;
+        if (((ext->obj->dynamic & WOLFSSL_ASN1_DYNAMIC_DATA) != 0) ||
+            (ext->obj->obj == NULL)) {
+                ext->obj->obj =(byte*)XREALLOC((byte*)ext->obj->obj,
+                                     ext->obj->objSz,
+                                     NULL,DYNAMIC_TYPE_ASN1);
+            if (ext->obj->obj == NULL) {
+                wolfSSL_X509_EXTENSION_free(ext);
+                FreeDecodedCert(cert);
+                XFREE(oidBuf, NULL, DYNAMIC_TYPE_TMP_BUFFER);
+            #ifdef WOLFSSL_SMALL_STACK
+                XFREE(cert, NULL, DYNAMIC_TYPE_DCERT);
+            #endif
+                return NULL;
+            }
+            ext->obj->dynamic |= WOLFSSL_ASN1_DYNAMIC_DATA;
+        }
+        else {
+            ext->obj->dynamic &= ~WOLFSSL_ASN1_DYNAMIC_DATA;
+        }
+        /* Get OID from input and copy to ASN1_OBJECT buffer */
+        XMEMCPY(oidBuf+2, input+idx, length);
+        XMEMCPY((byte*)ext->obj->obj, oidBuf, ext->obj->objSz);
+        XFREE(oidBuf, NULL, DYNAMIC_TYPE_TMP_BUFFER);
+        oidBuf = NULL;
+        ext->obj->grp = oidCertExtType;
+        ext->crit = 0;
 
-                tmpIdx++;
+        tmpIdx = idx + length;
 
-                if (GetLength(input, &tmpIdx, &length, (word32)sz) <= 0) {
-                    WOLFSSL_MSG("Error: Invalid Input Length.");
-                    wolfSSL_ASN1_OBJECT_free(ext->obj);
-                    wolfSSL_X509_EXTENSION_free(ext);
-                    FreeDecodedCert(cert);
-                #ifdef WOLFSSL_SMALL_STACK
-                    XFREE(cert, NULL, DYNAMIC_TYPE_DCERT);
-                #endif
-                    return NULL;
-                }
-                ext->value.data = (char*)XMALLOC(length, NULL,
-                    DYNAMIC_TYPE_ASN1);
-                ext->value.isDynamic = 1;
-                if (ext->value.data == NULL) {
-                    WOLFSSL_MSG("Failed to malloc ASN1_STRING data");
-                    wolfSSL_X509_EXTENSION_free(ext);
-                    FreeDecodedCert(cert);
-                #ifdef WOLFSSL_SMALL_STACK
-                    XFREE(cert, NULL, DYNAMIC_TYPE_DCERT);
-                #endif
-                    return NULL;
-                }
-                XMEMCPY(ext->value.data,input+tmpIdx,length);
-                ext->value.length = length;
-        } /* switch(oid) */
+        /* Get CRITICAL. If not present, defaults to false.
+         * It present, must be a valid TRUE */
+        if ((tmpIdx < (word32)sz) &&
+            (input[tmpIdx] == ASN_BOOLEAN))
+        {
+            if (((tmpIdx + 2) >= (word32)sz) ||
+                (input[tmpIdx+1] != 1) ||
+                (input[tmpIdx+2] != 0xff))
+            {
+                WOLFSSL_MSG("Error decoding unknown extension data");
+                wolfSSL_ASN1_OBJECT_free(ext->obj);
+                wolfSSL_X509_EXTENSION_free(ext);
+                FreeDecodedCert(cert);
+            #ifdef WOLFSSL_SMALL_STACK
+                XFREE(cert, NULL, DYNAMIC_TYPE_DCERT);
+            #endif
+                return NULL;
+            }
+
+            ext->crit = 1;
+            tmpIdx += 3;
+        }
+
+        /* Get extension data and copy as ASN1_STRING */
+        if ((tmpIdx >= (word32)sz) ||
+            (input[tmpIdx] != ASN_OCTET_STRING))
+        {
+            WOLFSSL_MSG("Error decoding unknown extension data");
+            wolfSSL_ASN1_OBJECT_free(ext->obj);
+            wolfSSL_X509_EXTENSION_free(ext);
+            FreeDecodedCert(cert);
+        #ifdef WOLFSSL_SMALL_STACK
+            XFREE(cert, NULL, DYNAMIC_TYPE_DCERT);
+        #endif
+            return NULL;
+        }
+
+        tmpIdx++;
+
+        if (GetLength(input, &tmpIdx, &length, (word32)sz) <= 0) {
+            WOLFSSL_MSG("Error: Invalid Input Length.");
+            wolfSSL_ASN1_OBJECT_free(ext->obj);
+            wolfSSL_X509_EXTENSION_free(ext);
+            FreeDecodedCert(cert);
+        #ifdef WOLFSSL_SMALL_STACK
+            XFREE(cert, NULL, DYNAMIC_TYPE_DCERT);
+        #endif
+            return NULL;
+        }
+        ext->value.data = (char*)XMALLOC(length, NULL,
+            DYNAMIC_TYPE_ASN1);
+        ext->value.isDynamic = 1;
+        if (ext->value.data == NULL) {
+            WOLFSSL_MSG("Failed to malloc ASN1_STRING data");
+            wolfSSL_X509_EXTENSION_free(ext);
+            FreeDecodedCert(cert);
+        #ifdef WOLFSSL_SMALL_STACK
+            XFREE(cert, NULL, DYNAMIC_TYPE_DCERT);
+        #endif
+            return NULL;
+        }
+        XMEMCPY(ext->value.data,input+tmpIdx,length);
+        ext->value.length = length;
 
         break; /* Got the Extension. Now exit while loop. */
 


dgarske:
Okay to test. No approved contributor agreement yet but process started.

ribes96:
Last pushed commit (979d73228 (Add missing ext_sk to ALT_NAMES extension, 2025-07-08)) makes the LDAP tests pass.

`libspdm` is failing because the 'wolfSSL/osp' project has a [patch](https://github.com/wolfSSL/osp/blob/master/libspdm/3.7.0/libspdm-3.7.0.patch) with the following contents:

~~~patch
--- a/library/spdm_crypt_lib/libspdm_crypt_cert.c
+++ b/library/spdm_crypt_lib/libspdm_crypt_cert.c
@@ -911,11 +911,13 @@ static bool libspdm_verify_leaf_cert_spdm_eku(const uint8_t *cert, size_t cert_s
     req_auth_oid_find_success = false;
     rsp_auth_oid_find_success = false;
 
+#ifndef SPDM_USING_WOLFSSL /* wolfssl returns data without sequence tag */
     status = libspdm_asn1_get_tag(&ptr, eku + eku_size, &obj_len,
                                   LIBSPDM_CRYPTO_ASN1_SEQUENCE | LIBSPDM_CRYPTO_ASN1_CONSTRUCTED);
     if (!status) {
         return false;
     }
+#endif
 
     while(ptr < eku + eku_size) {
         status = libspdm_asn1_get_tag(&ptr, eku + eku_size, &obj_len, LIBSPDM_CRYPTO_ASN1_OID);
~~~

So it is working around precisely the behavior that this PR is fixing. With the patch fixed, the `libspdm` tests pass in my workstation.

dgarske:
Okay to test. Contributor agreement approved and on file.

dgarske:
> Last pushed commit ([979d732](https://github.com/wolfSSL/wolfssl/commit/979d73228bcedd94c50f0141e57408629fc9a4b8) (Add missing ext_sk to ALT_NAMES extension, 2025-07-08)) makes the LDAP tests pass.
> 
> `libspdm` is failing because the 'wolfSSL/osp' project has a [patch](https://github.com/wolfSSL/osp/blob/master/libspdm/3.7.0/libspdm-3.7.0.patch) with the following contents:
> 
> ```diff
> --- a/library/spdm_crypt_lib/libspdm_crypt_cert.c
> +++ b/library/spdm_crypt_lib/libspdm_crypt_cert.c
> @@ -911,11 +911,13 @@ static bool libspdm_verify_leaf_cert_spdm_eku(const uint8_t *cert, size_t cert_s
>      req_auth_oid_find_success = false;
>      rsp_auth_oid_find_success = false;
>  
> +#ifndef SPDM_USING_WOLFSSL /* wolfssl returns data without sequence tag */
>      status = libspdm_asn1_get_tag(&ptr, eku + eku_size, &obj_len,
>                                    LIBSPDM_CRYPTO_ASN1_SEQUENCE | LIBSPDM_CRYPTO_ASN1_CONSTRUCTED);
>      if (!status) {
>          return false;
>      }
> +#endif
>  
>      while(ptr < eku + eku_size) {
>          status = libspdm_asn1_get_tag(&ptr, eku + eku_size, &obj_len, LIBSPDM_CRYPTO_ASN1_OID);
> ```
> 
> So it is working around precisely the behavior that this PR is fixing. With the patch fixed, the `libspdm` tests pass in my workstation.

@julek-wolfssl will you please work on updating the libspdm patch and reviewing this PR?

src/x509.c
```diff
@@ -1539,7 +1455,7 @@ int wolfSSL_X509V3_EXT_print(WOLFSSL_BIO *out, WOLFSSL_X509_EXTENSION *ext,
     WOLFSSL_ASN1_STRING* str;
     int nid;
     int rc = WC_NO_ERR_TRACE(WOLFSSL_FAILURE);
-    char tmp[CTC_NAME_SIZE*2 + 1];
+    char tmp[1024];
```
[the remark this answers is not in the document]
ribes96: I don't think it is possible to find a value suitable for any extension we can find, since they can be arbitrarily large. But given that the only errors inside this function are due to a short buffer, maybe we could directly use `wolfSSL_BIO_printf`, and then we would not need a temporary buffer

src/x509.c
```diff
@@ -1147,137 +1055,145 @@ WOLFSSL_X509_EXTENSION* wolfSSL_X509_set_ext(WOLFSSL_X509* x509, int loc)
...
+            }
+        }
 
-                /* Get OID from input and copy to ASN1_OBJECT buffer */
-                XMEMCPY(oidBuf+2, input+idx, length);
-                XMEMCPY((byte*)ext->obj->obj, oidBuf, ext->obj->objSz);
+        ext->obj->objSz = (unsigned int)objSz;
+        if (((ext->obj->dynamic & WOLFSSL_ASN1_DYNAMIC_DATA) != 0) ||
+            (ext->obj->obj == NULL)) {
+                ext->obj->obj =(byte*)XREALLOC((byte*)ext->obj->obj,
```
SparkiDev: The code to handle no XREALLOC implementation was removed.
Please put it back.

src/x509.c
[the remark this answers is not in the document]
SparkiDev: Leave it at CTC_NAME_SIZE*2+1 unless there is unreasonably small. If too small leave it at 1024.
This is only a maximum to print - it doesn't fail if the data is cut-off.
Put a TODO to use BIO_printf (or you can do that it if you want.)

src/x509.c
```diff
@@ -1953,31 +1869,75 @@ void* wolfSSL_X509V3_EXT_d2i(WOLFSSL_X509_EXTENSION* ext)
...
             WOLFSSL_MSG("AuthorityKeyIdentifier");
 
+            asn1String = wolfSSL_X509_EXTENSION_get_data(ext);
+            if (asn1String == NULL) {
+                WOLFSSL_MSG("X509_EXTENSION_get_data() failed");
+                return NULL;
+            }
+
+            idx = 0;
+            if (GetSequence((byte *)asn1String->data, &idx, &len, asn1String->length) < 0) {
```
SparkiDev: Please ensure lines are no longer than 80 characters.
Thanks!

src/x509.c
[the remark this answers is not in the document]
ribes96: I increased it because after the changes the extension of some test required 611 bytes, failing because the buffer had a size of 129. Then, I'll leave it at 1024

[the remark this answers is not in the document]
dgarske: Hi @ribes96 , please revert this line. If you want a larger tmp in your application then override `WC_CTC_NAME_SIZE` at build-time.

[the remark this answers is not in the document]
ribes96: Okay, done. What should I do now with the test that fails due to the revert. I'm talking about `scripts/unit.test`, `tests/api.c:68860`, `test_wolfSSL_X509V3_EXT_print`

src/x509.c
```diff
@@ -1046,238 +1040,174 @@ WOLFSSL_X509_EXTENSION* wolfSSL_X509_set_ext(WOLFSSL_X509* x509, int loc)
...
-                #ifdef WOLFSSL_NO_REALLOC
-                    byte* tmp = NULL;
+        /* Get CRITICAL. If not present, defaults to false.
+         * It present, must be a valid TRUE */
+        if ((tmpIdx < (word32)sz) &&
+            (input[tmpIdx] == ASN_BOOLEAN))
+        {
+            if (((tmpIdx + 2) >= (word32)sz) ||
+                (input[tmpIdx+1] != 1) ||
+                (input[tmpIdx+2] != 0xff))
```
julek-wolfssl:
```suggestion
            if (((tmpIdx + 2) >= (word32)sz) ||
                /* Check bool length */
                (input[tmpIdx+1] != 1) ||
                /* Assert true if CRITICAL present */
                (input[tmpIdx+2] != 0xff))
```

src/x509.c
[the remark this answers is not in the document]
dgarske: The code should truncate and the test should accommodate a smaller response size.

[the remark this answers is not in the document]
ribes96: Okay, I can do that if truncation is allowed. If fact the test only checks if the call is successful, so it won't need accommodation

ribes96:
> I don't like the growing complexity of `wolfSSL_X509V3_EXT_d2i`. At this point I would prefer to use `DecodeCertExtensions` to parse the extension and then set the appropriate fields from `DecodedCert`. @ribes96 are you up for such a refactor or should one of us tackle this section?

In fact this this function is just replicating what the following functions do:
~~~c
DecodeBasicCaConstraint
DecodeSubjKeyId
DecodeAuthKeyId
DecodeKeyUsage
~~~

But they couldn't be used on the one hand because they are private of `wolfcrypt`, and on the other because they all receive as parameter a `DecodedCert`, which is not available in `wolfSSL_X509V3_EXT_d2i` because there is simply no certificate, just an extension.

So the path forward would be to add a version of these functions to the public API of `wolfcrypt`, that doesn't receive a `DecodedCert` but some specific data, and use them both inside current functions `DecodeBasicCaConstraint`, etc, and from `wolfSSL_X509V3_EXT_d2i`.

I can do that

dgarske:
```
./configure CC="clang -fsanitize=address,undefined -g" --enable-all --enable-sp-math-all
FAIL: scripts/unit.test
```

See: https://github.com/wolfSSL/wolfssl/actions/runs/16224294061/job/45819565451?pr=8967

```
    710: test_wolfSSL_X509V3_EXT                            :tests/api.c:43770:5: runtime error: call to function (unknown) through pointer to incorrect function type 'char *(*)(struct WOLFSSL_v3_ext_method *, void *)'
(/home/davidgarske/GitHub/wolfssl/src/.libs/libwolfssl.so.43+0x6cd898): note: (unknown) defined here
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior tests/api.c:43770:5 
 passed (  0.00069)
    711: test_wolfSSL_X509V3_EXT_bc                         : passed (  0.00000)
    712: test_wolfSSL_X509V3_EXT_san                        : passed (  0.00000)
    713: test_wolfSSL_X509V3_EXT_aia                        : passed (  0.00001)
    714: test_wolfSSL_X509V3_EXT_print                      :=================================================================
==2033422==ERROR: AddressSanitizer: stack-buffer-overflow on address 0x7bafba8f4aa1 at pc 0x0000004af87f bp 0x7ffec4fbc410 sp 0x7ffec4fbbbd0
READ of size 611 at 0x7bafba8f4aa1 thread T0
    #0 0x0000004af87e  (/home/davidgarske/GitHub/wolfssl/tests/.libs/unit.test+0x4af87e) (BuildId: 0fc9b9dcb8ce3765825afd11856edfdf479144ea)
    #1 0x7fafbd0c2a15  (/home/davidgarske/GitHub/wolfssl/src/.libs/libwolfssl.so.43+0x6c2a15) (BuildId: b849fc0057905a409e0b93d1836dbe2acfb1bc78)
    #2 0x7fafbd1c4010  (/home/davidgarske/GitHub/wolfssl/src/.libs/libwolfssl.so.43+0x7c4010) (BuildId: b849fc0057905a409e0b93d1836dbe2acfb1bc78)
    #3 0x0000006f849e  (/home/davidgarske/GitHub/wolfssl/tests/.libs/unit.test+0x6f849e) (BuildId: 0fc9b9dcb8ce3765825afd11856edfdf479144ea)
    #4 0x000000974389  (/home/davidgarske/GitHub/wolfssl/tests/.libs/unit.test+0x974389) (BuildId: 0fc9b9dcb8ce3765825afd11856edfdf479144ea)
    #5 0x0000004f7476  (/home/davidgarske/GitHub/wolfssl/tests/.libs/unit.test+0x4f7476) (BuildId: 0fc9b9dcb8ce3765825afd11856edfdf479144ea)
    #6 0x7fafbc6c65f4  (/lib64/libc.so.6+0x35f4) (BuildId: c4b06a608071b2c9852fae62ca3b69cdc22cd022)
    #7 0x7fafbc6c66a7  (/lib64/libc.so.6+0x36a7) (BuildId: c4b06a608071b2c9852fae62ca3b69cdc22cd022)
    #8 0x00000040e074  (/home/davidgarske/GitHub/wolfssl/tests/.libs/unit.test+0x40e074) (BuildId: 0fc9b9dcb8ce3765825afd11856edfdf479144ea)

Address 0x7bafba8f4aa1 is located in stack of thread T0 at offset 161 in frame
    #0 0x7fafbd1c37c7  (/home/davidgarske/GitHub/wolfssl/src/.libs/libwolfssl.so.43+0x7c37c7) (BuildId: b849fc0057905a409e0b93d1836dbe2acfb1bc78)

  This frame has 3 object(s):
    [32, 161) 'tmp' (line 1474)
    [240, 245) 'isCa' (line 1501) <== Memory access at offset 161 partially underflows this variable
    [272, 278) 'notCa' (line 1502) <== Memory access at offset 161 partially underflows this variable
HINT: this may be a false positive if your program uses some custom stack unwind mechanism, swapcontext or vfork
      (longjmp and C++ exceptions *are* supported)
SUMMARY: AddressSanitizer: stack-buffer-overflow (/home/davidgarske/GitHub/wolfssl/tests/.libs/unit.test+0x4af87e) (BuildId: 0fc9b9dcb8ce3765825afd11856edfdf479144ea) 
Shadow bytes around the buggy address:
  0x7bafba8f4800: f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5
  0x7bafba8f4880: f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5
  0x7bafba8f4900: f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5
  0x7bafba8f4980: f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5
  0x7bafba8f4a00: f1 f1 f1 f1 00 00 00 00 00 00 00 00 00 00 00 00
=>0x7bafba8f4a80: 00 00 00 00[01]f2 f2 f2 f2 f2 f2 f2 f2 f2 f8 f2
  0x7bafba8f4b00: f2 f2 f8 f3 f3 f3 f3 f3 00 00 00 00 00 00 00 00
  0x7bafba8f4b80: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x7bafba8f4c00: f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5
  0x7bafba8f4c80: f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5
  0x7bafba8f4d00: f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07 
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
==2033422==ABORTING
```

SparkiDev:
Consider putting in #ifdefs to have the library do the old behaviour for those that are expecting it.

dgarske:
Retest this please: "AgentOfflineException"

julek-wolfssl:
@ribes96 My plan was to expose `DecodeCertExtensions` and to allocate a `DecodedCert` on the stack to decode the extensions. This wasy `asn.c` changes stay minimal.

ribes96:
> @ribes96 My plan was to expose `DecodeCertExtensions` and to allocate a `DecodedCert` on the stack to decode the extensions. This wasy `asn.c` changes stay minimal.

I'm not seeing how that can be done. `DecodeCertExtensions` expects a full extensions sequence inside `DecodedCert->{extensions, extensionsSz}` and then fills `DecodedCert` with the extensions found there. Are you proposing to construct a dummy extensions sequence, containing the only one extension available in `wolfSSL_X509V3_EXT_d2i`, initialize a new `DecodedCert` by just filling `extensions` and `extensionsSz` with the dummy extensions sequence, then call `DecodeCertExtensions` and extract from `DecodedCert` the appropriate fields depending on the extension?

ribes96:
> Consider putting in #ifdefs to have the library do the old behaviour for those that are expecting it.

I am now working on this
I am considering adding a new private function that transforms a given `WOLFSSL_ASN1_STRING` from containing the full extension `OCTET STRING` into the old contents. Then function `wolfSSL_X509_EXTENSION_get_data` would call it in the end depending on the #ifdefs. `wolfSSL_X509_EXTENSION_get_data` should be split into an internal version and an external version, to ensure that internal callers get the expected behavior independently of the #ifdef.

Any proposal for the name of the feature flag?

Strictly speaking, `wolfSSL_X509_EXTENSION_get_data` should be updated to return an `ASN1_OCTET_STRING` instead of an `ASN1_STRING`. They are currently aliased:
~~~c
#define ASN1_OCTET_STRING         WOLFSSL_ASN1_STRING
~~~

Should we change the function signature? I'm not sure how will this affect existing callers

SparkiDev (dismissed):

src/x509.c
```diff
@@ -1953,31 +1898,62 @@ void* wolfSSL_X509V3_EXT_d2i(WOLFSSL_X509_EXTENSION* ext)
...
+            asn1String = wolfSSL_X509_EXTENSION_get_data_internal(ext);
             if (asn1String == NULL) {
                 WOLFSSL_MSG("X509_EXTENSION_get_data() failed");
                 return NULL;
             }
+            if (DecodeSubjKeyId((const byte *)asn1String->data,
+                    asn1String->length, &extSubjKeyId, &extSubjKeyIdSz) != 0) {
+                return NULL;
+            }
+
```
julek-wolfssl: I would prefer to call `DecodeExtensionType` before and then call the asn1 setting functions using fields from `DecodedCert`.

julek-wolfssl (changes requested):

wolfcrypt/src/asn.c
```diff
@@ -21345,154 +21197,42 @@ enum {
...
-                cert->extExtKeyUsageSsh |= EXTKEYUSE_SSH_MSCL;
-                break;
-            case EKU_SSH_KP_CLIENT_AUTH_OID:
-                cert->extExtKeyUsageSsh |= EXTKEYUSE_SSH_KP_CLIENT_AUTH;
-                break;
-            #endif /* WOLFSSL_WOLFSSH */
-            default:
-                break;
-        }
+    cert->extExtKeyUsageSrc = extExtKeyUsageSrc;
```
dgarske:
```
wolfcrypt/src/asn.c: In function ‘DecodeExtKeyUsageInternal’:

wolfcrypt/src/asn.c:21254:11: error: ‘DecodedCert’ has no member named ‘extExtKeyUsageSrc’; did you mean ‘extExtKeyUsageSet’?
21254 |     cert->extExtKeyUsageSrc = extExtKeyUsageSrc;
      |           ^~~~~~~~~~~~~~~~~
      |           extExtKeyUsageSet

wolfcrypt/src/asn.c:21255:11: error: ‘DecodedCert’ has no member named ‘extExtKeyUsageSz’; did you mean ‘extExtKeyUsage’?
21255 |     cert->extExtKeyUsageSz = extExtKeyUsageSz;
      |           ^~~~~~~~~~~~~~~~
      |           extExtKeyUsage
```

dgarske (changes requested):

ribes96:
Should I merge master to resolve the conflicts?

dgarske:
Hi @ribes96 , please resolve merge conflicts. Thank you

wolfcrypt/src/asn.c
```diff
@@ -23743,6 +23489,503 @@ int DecodeCert(DecodedCert* cert, int verify, int* criticalExt)
...
+ * @return  ASN_PARSE_E when BER encoded data does not match ASN.1 items or
+ *          is invalid.
+ * @return  BUFFER_E when data in buffer is too small.
+ */
+WOLFSSL_LOCAL int DecodeAuthKeyId(const byte* input, word32 sz,
+            const byte **extAuthKeyId, word32 *extAuthKeyIdSz,
+            const byte **extAuthKeyIdIssuer, word32 *extAuthKeyIdIssuerSz,
+            const byte **extAuthKeyIdIssuerSN, word32 *extAuthKeyIdIssuerSNSz)
+{
+    *extAuthKeyId = NULL;
```
dgarske: Mixed declaration. Please make sure all variables are declared at top of function or brace {} section.

dgarske (changes requested):

dgarske (changes requested):
```
#ifdef WOLFSSL_ASN_TEMPLATE
^
/tmp/workspace/PRB-multi-test-script/wolfssl/wolfcrypt/src/asn.c:23789:2: warning: nested redundant #ifndef; consider removing it [readability-redundant-preprocessor]
#ifndef WOLFSSL_ASN_TEMPLATE
^
/tmp/workspace/PRB-multi-test-script/wolfssl/wolfcrypt/src/asn.c:22894:2: note: previous #ifdef was here
#ifdef WOLFSSL_ASN_TEMPLATE
^
/tmp/workspace/PRB-multi-test-script/wolfssl/wolfcrypt/src/asn.c:23853:2: warning: nested redundant #ifndef; consider removing it [readability-redundant-preprocessor]
#ifndef WOLFSSL_ASN_TEMPLATE
^
/tmp/workspace/PRB-multi-test-script/wolfssl/wolfcrypt/src/asn.c:22894:2: note: previous #ifdef was here
#ifdef WOLFSSL_ASN_TEMPLATE
^
```

```
./configure --enable-jobserver=4 CC="clang -fsanitize=address,undefined -g" --enable-all --enable-sp-math-all
make check
FAIL: scripts/unit.test
```

wolfcrypt/src/asn.c
```diff
@@ -21389,147 +21234,35 @@ enum {
  *          is invalid.
  * @return  MEMORY_E on dynamic memory allocation failure.
  */
-static int DecodeExtKeyUsage(const byte* input, word32 sz, DecodedCert* cert)
+static int DecodeExtKeyUsageInternal(const byte* input, word32 sz, DecodedCert* cert)
```
dgarske:
```
[clang-tidy-defaults] [7 of 7] [wolfssl]
    configure...   real 0m9.257s  user 0m5.839s  sys 0m3.056s
    build.../tmp/workspace/PRB-multi-test-script/wolfssl/wolfcrypt/src/asn.c:21237:5: warning: Value stored to 'ret' is never read [clang-analyzer-deadcode.DeadStores]
ret = GetHashId(extSubjKeyId, extSubjKeyIdSz, cert->extSubjKeyId,
^     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/tmp/workspace/PRB-multi-test-script/wolfssl/wolfcrypt/src/asn.c:21237:5: note: Value stored to 'ret' is never read
ret = GetHashId(extSubjKeyId, extSubjKeyIdSz, cert->extSubjKeyId,
^     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

dgarske (changes requested):
Hi @ribes96 ,

This breaks some of the return code logic. For example the `test_wolfSSL_CRL_duplicate_extensions` we have a faketime test that pushes the clock a year out. We expect a `CRL_CERT_DATE_ERR`, but now we get a `ASN_PARSE_E`. Can you please make sure we aren't loosing some of the return code logic.

```
[all-check-certs-one-year] [7 of 50] [e6b39baf40] [previously fail_check]
    autogen.sh e6b39baf40...   real 0m4.007s  user 0m2.797s  sys 0m0.248s
    configure...   real 0m6.777s  user 0m4.823s  sys 0m2.537s
    build...   real 0m17.111s  user 1m26.745s  sys 0m3.905s
    check[faketime]...   real 0m0.055s  user 0m0.046s  sys 0m0.014s
    configure...   real 0m6.531s  user 0m4.536s  sys 0m2.573s
    build...   real 0m16.726s  user 1m26.467s  sys 0m4.037s
    check[faketime]...FAILURES:
   real 0m11.347s  user 0m3.251s  sys 0m0.056s
ERROR - tests/api.c line 4291 failed with:
    expected: ret == ASN_PARSE_E
    result:   -179 != -140
ERROR - tests/api.c line 68860 failed with:
    expected: Test failed
    result:   ret 0
ERROR - tests/api.c line 9892 failed with:
    expected: wolfSSL_read_ex(ssl_s, buf, sizeof(buf), &count) == WOLFSSL_SUCCESS
    result:   5 != 1
ERROR - tests/api.c line 64228 failed with:
    expected: wolfSSL_shutdown(ssl_c) == 1
    result:   0 != 1

FAILURES:
   917: test_wolfSSL_CRL_duplicate_extensions
```

Here is that specific test with the logging and WOLFSSL_DEBUG_ASN_TEMPLATE set.

```
davidgarske@i9w680:/tmp/wolfssl_test_workdir.811078/wolfssl$ faketime -f +1y scripts/unit.test -917
starting unit tests...
 Begin API Tests
wolfSSL Entering wolfSSL_Init

wolfSSL Entering wolfCrypt_Init
RNG_HEALTH_TEST_CHECK_SIZE = 128
sizeof(seedB_data)         = 128
    917: test_wolfSSL_CRL_duplicate_extensions              :wolfSSL Entering wolfSSL_CertManagerNew
heap param is null
DYNAMIC_TYPE_CERT_MANAGER Allocating = 392 bytes
Testing CRL with duplicate Authority Key Identifier extensions
wolfSSL Entering wolfSSL_CertManagerLoadCRLBuffer
wolfSSL Entering wolfSSL_CertManagerEnableCRL
wolfSSL Entering InitCRL
wolfSSL Entering BufferLoadCRL
wolfSSL Entering PemToDer
InitDecodedCRL
ParseCRL
TEMPLATE: crlASN
wolfSSL Entering GetASN_Items
 0:    0  520 +  SEQUENCE        
 1:    4  241 +   SEQUENCE        
 2:    7    1      INTEGER         
 3:   10   13 +    SEQUENCE        
 4:   12    9       OBJECT ID       
 5:   23    0       NULL            
 7:   25  121 +    SEQUENCE        
 8:  148   13      UTCTime         
10:  163   13      UTCTime         
13:  178   68 +    [0]             
14:  180   66 +     SEQUENCE        
15:  248   13 +   SEQUENCE        
16:  250    9      OBJECT ID       
17:  261    0      NULL            
19:  263  257     BIT STRING      
Date AFTER check failed
CRL after date is no longer valid
wolfSSL error occurred, error = 179 line:41004 file:wolfcrypt/src/asn.c
ParseCRL error
wolfSSL Entering FreeCRL_Entry
FreeDecodedCRL

ERROR - tests/api.c line 4291 failed with:
    expected: ret == ASN_PARSE_E
    result:   -179 != -140

wolfSSL Entering wolfSSL_CertManagerFree
wolfSSL Entering FreeCRL
 failed (  0.00010)

ERROR - tests/api.c line 68860 failed with:
    expected: Test failed

    result:   ret 0

wolfSSL Entering wolfSSL_ERR_clear_error
wolfSSL Entering wolfSSL_Cleanup
wolfSSL Entering wolfSSL_crypto_policy_disable
wolfSSL Entering wolfCrypt_Cleanup

FAILURES:
   917: test_wolfSSL_CRL_duplicate_extensions

 End API Tests
 Failed/Skipped/Passed/All: 1/0/0/1
davidgarske@i9w680:/tmp/wolfssl_test_workdir.811078/wolfssl$ 
```

ribes96:
> Hi @ribes96 ,
> 
> This breaks some of the return code logic. For example the `test_wolfSSL_CRL_duplicate_extensions` we have a faketime test that pushes the clock a year out. We expect a `CRL_CERT_DATE_ERR`, but now we get a `ASN_PARSE_E`. Can you please make sure we aren't loosing some of the return code logic.
> 
> ```
> [all-check-certs-one-year] [7 of 50] [e6b39baf40] [previously fail_check]
>     autogen.sh e6b39baf40...   real 0m4.007s  user 0m2.797s  sys 0m0.248s
>     configure...   real 0m6.777s  user 0m4.823s  sys 0m2.537s
>     build...   real 0m17.111s  user 1m26.745s  sys 0m3.905s
>     check[faketime]...   real 0m0.055s  user 0m0.046s  sys 0m0.014s
>     configure...   real 0m6.531s  user 0m4.536s  sys 0m2.573s
>     build...   real 0m16.726s  user 1m26.467s  sys 0m4.037s
>     check[faketime]...FAILURES:
>    real 0m11.347s  user 0m3.251s  sys 0m0.056s
> ERROR - tests/api.c line 4291 failed with:
>     expected: ret == ASN_PARSE_E
>     result:   -179 != -140
> ERROR - tests/api.c line 68860 failed with:
>     expected: Test failed
>     result:   ret 0
> ERROR - tests/api.c line 9892 failed with:
>     expected: wolfSSL_read_ex(ssl_s, buf, sizeof(buf), &count) == WOLFSSL_SUCCESS
>     result:   5 != 1
> ERROR - tests/api.c line 64228 failed with:
>     expected: wolfSSL_shutdown(ssl_c) == 1
>     result:   0 != 1
> 
> FAILURES:
>    917: test_wolfSSL_CRL_duplicate_extensions
> ```
> 
> Here is that specific test with the logging and WOLFSSL_DEBUG_ASN_TEMPLATE set.
> 
> ```
> davidgarske@i9w680:/tmp/wolfssl_test_workdir.811078/wolfssl$ faketime -f +1y scripts/unit.test -917
> starting unit tests...
>  Begin API Tests
> wolfSSL Entering wolfSSL_Init
> 
> wolfSSL Entering wolfCrypt_Init
> RNG_HEALTH_TEST_CHECK_SIZE = 128
> sizeof(seedB_data)         = 128
>     917: test_wolfSSL_CRL_duplicate_extensions              :wolfSSL Entering wolfSSL_CertManagerNew
> heap param is null
> DYNAMIC_TYPE_CERT_MANAGER Allocating = 392 bytes
> Testing CRL with duplicate Authority Key Identifier extensions
> wolfSSL Entering wolfSSL_CertManagerLoadCRLBuffer
> wolfSSL Entering wolfSSL_CertManagerEnableCRL
> wolfSSL Entering InitCRL
> wolfSSL Entering BufferLoadCRL
> wolfSSL Entering PemToDer
> InitDecodedCRL
> ParseCRL
> TEMPLATE: crlASN
> wolfSSL Entering GetASN_Items
>  0:    0  520 +  SEQUENCE        
>  1:    4  241 +   SEQUENCE        
>  2:    7    1      INTEGER         
>  3:   10   13 +    SEQUENCE        
>  4:   12    9       OBJECT ID       
>  5:   23    0       NULL            
>  7:   25  121 +    SEQUENCE        
>  8:  148   13      UTCTime         
> 10:  163   13      UTCTime         
> 13:  178   68 +    [0]             
> 14:  180   66 +     SEQUENCE        
> 15:  248   13 +   SEQUENCE        
> 16:  250    9      OBJECT ID       
> 17:  261    0      NULL            
> 19:  263  257     BIT STRING      
> Date AFTER check failed
> CRL after date is no longer valid
> wolfSSL error occurred, error = 179 line:41004 file:wolfcrypt/src/asn.c
> ParseCRL error
> wolfSSL Entering FreeCRL_Entry
> FreeDecodedCRL
> 
> ERROR - tests/api.c line 4291 failed with:
>     expected: ret == ASN_PARSE_E
>     result:   -179 != -140
> 
> wolfSSL Entering wolfSSL_CertManagerFree
> wolfSSL Entering FreeCRL
>  failed (  0.00010)
> 
> ERROR - tests/api.c line 68860 failed with:
>     expected: Test failed
> 
>     result:   ret 0
> 
> wolfSSL Entering wolfSSL_ERR_clear_error
> wolfSSL Entering wolfSSL_Cleanup
> wolfSSL Entering wolfSSL_crypto_policy_disable
> wolfSSL Entering wolfCrypt_Cleanup
> 
> FAILURES:
>    917: test_wolfSSL_CRL_duplicate_extensions
> 
>  End API Tests
>  Failed/Skipped/Passed/All: 1/0/0/1
> davidgarske@i9w680:/tmp/wolfssl_test_workdir.811078/wolfssl$ 
> ```

I'm sorry, I don't understand.

`test_wolfSSL_CRL_duplicate_extensions` gives me the same results in master (cc2f7927e (Merge pull request #9035 from douzzer/20250725-wc_linuxkm_relax_long_loop, 2025-07-26)) and in the last commit of this branch (b07a0e07c (Fix return correct error code in DecodeSubjKeyIdInternal, 2025-07-25)).

The test checks that `wolfSSL_CertManagerLoadCRLBuffer` returns `ASN_PARSE_E`. I see this behaviour both with and without `faketime`.

You are talking about this test, right?
~~~c
static int test_wolfSSL_CRL_duplicate_extensions(void)
{
    EXPECT_DECLS;
#if defined(WOLFSSL_ASN_TEMPLATE) && !defined(NO_CERTS) && \
    defined(HAVE_CRL) && !defined(NO_RSA) && !defined(WOLFSSL_NO_ASN_STRICT) && \
    (defined(WC_ASN_RUNTIME_DATE_CHECK_CONTROL) || defined(NO_ASN_TIME_CHECK))
    const unsigned char crl_duplicate_akd[] =
        "-----BEGIN X509 CRL-----\n"
        "MIICCDCB8QIBATANBgkqhkiG9w0BAQsFADB5MQswCQYDVQQGEwJVUzETMBEGA1UE\n"
        "CAwKQ2FsaWZvcm5pYTEWMBQGA1UEBwwNU2FuIEZyYW5jaXNjbzETMBEGA1UECgwK\n"
        "TXkgQ29tcGFueTETMBEGA1UEAwwKTXkgUm9vdCBDQTETMBEGA1UECwwKTXkgUm9v\n"
        "dCBDQRcNMjQwOTAxMDAwMDAwWhcNMjUxMjAxMDAwMDAwWqBEMEIwHwYDVR0jBBgw\n"
        "FoAU72ng99Ud5pns3G3Q9+K5XGRxgzUwHwYDVR0jBBgwFoAU72ng99Ud5pns3G3Q\n"
        "9+K5XGRxgzUwDQYJKoZIhvcNAQELBQADggEBAIFVw4jrS4taSXR/9gPzqGrqFeHr\n"
        "IXCnFtHJTLxqa8vUOAqSwqysvNpepVKioMVoGrLjFMjANjWQqTEiMROAnLfJ/+L8\n"
        "FHZkV/mZwOKAXMhIC9MrJzifxBICwmvD028qnwQm09EP8z4ICZptD6wPdRTDzduc\n"
        "KBuAX+zn8pNrJgyrheRKpPgno9KsbCzK4D/RIt1sTK2M3vVOtY+vpsN70QYUXvQ4\n"
        "r2RZac3omlT43x5lddPxIlcouQpwWcVvr/K+Va770MRrjn88PBrJmvsEw/QYVBXp\n"
        "Gxv2b78HFDacba80sMIm8ltRdqUCa5qIc6OATsz7izCQXEbkTEeESrcK1MA=\n"
        "-----END X509 CRL-----\n";

    WOLFSSL_CERT_MANAGER* cm = NULL;
    int ret;

    (void)wc_AsnSetSkipDateCheck(1);

    cm = wolfSSL_CertManagerNew();
    ExpectNotNull(cm);

    /* Test loading CRL with duplicate extensions */
    WOLFSSL_MSG("Testing CRL with duplicate Authority Key Identifier extensions");
    ret = wolfSSL_CertManagerLoadCRLBuffer(cm, crl_duplicate_akd,
                                           sizeof(crl_duplicate_akd),
                                           WOLFSSL_FILETYPE_PEM);
    ExpectIntEQ(ret, ASN_PARSE_E);

    wolfSSL_CertManagerFree(cm);

    (void)wc_AsnSetSkipDateCheck(0);
#endif
    return EXPECT_RESULT();
}
~~~

I am building with
~~~sh
./configure --enable-all CFLAGS="-DWC_ASN_RUNTIME_DATE_CHECK_CONTROL"
~~~

I am using `faketime: Version 0.9.7`

dgarske:
Hi @ribes96 ,

I was wrong, your PR is passing that `all-check-certs-one-year` test. It was after I did a rebase to master. Also I seem to have an issue with rebasing this on master. Can you please squash these commits and force push?

Thanks,
David Garske, wolfSSL

dgarske:
Jenkins retest this please. Agent Offline

dgarske:
Hi @ribes96 ,

It seems like the changes in this PR are breaking some of the tests for libspdm, socat and zephyr?

```
[  FAILED  ] spdm_crypt_lib_tests: 5 test(s), listed below:
[  FAILED  ] libspdm_test_crypt_spdm_x509_certificate_check
[  FAILED  ] libspdm_test_crypt_spdm_x509_certificate_check_ex
[  FAILED  ] libspdm_test_crypt_spdm_x509_set_cert_certificate_check_ex
[  FAILED  ] libspdm_test_crypt_spdm_verify_cert_chain_data_ex
[  FAILED  ] libspdm_test_crypt_spdm_verify_certificate_chain_buffer_ex
```

ribes96:
> Hi @ribes96 ,
> 
> It seems like the changes in this PR are breaking some of the tests for libspdm, socat and zephyr?
> 
> ```
> [  FAILED  ] spdm_crypt_lib_tests: 5 test(s), listed below:
> [  FAILED  ] libspdm_test_crypt_spdm_x509_certificate_check
> [  FAILED  ] libspdm_test_crypt_spdm_x509_certificate_check_ex
> [  FAILED  ] libspdm_test_crypt_spdm_x509_set_cert_certificate_check_ex
> [  FAILED  ] libspdm_test_crypt_spdm_verify_cert_chain_data_ex
> [  FAILED  ] libspdm_test_crypt_spdm_verify_certificate_chain_buffer_ex
> ```

`libspdm` is expected to fail. It requires changes in the `wolfSSL/osp` repo. See [here](https://github.com/wolfSSL/osp/pull/266)

I've been unable to reproduce locally the `socat` test failure. I see exactly the same behavior in master and in this branch. My bet is that it is a flaky test. Test script `socat-chain.sh` indeed seems to be sensitive to a timing race when initializing a server. Moreover, the test has nothing to do with SSL

I am working in the `zephyr` case. But it is a very complicated setup and is requiring me a lot of time

ribes96:
Hey,
I managed to run the `zephyr` workflow locally using [act](https://github.com/nektos/act) and noticed that this workflow already failed before this branch started, in `a40b56ccb (Merge pull request #8964 from douzzer/20250703-linuxkm-fixes, 2025-07-04)`.
But it doesn't seem to fail in my current local master in `26f4c968d (Merge pull request #9032 from holtrop/allow-pkcs7-without-aes-keywrap, 2025-07-29)`.
I will try to bisect between the commits in that range, which do **not** include this branch.
But do you have any idea what may have happened?

**EDIT**: Forget about this. I was not running `act` correctly. master worked fine before my changes

ribes96:
Okay, this is what I have regarding the failing zephyr tests:

They are failing due to a timeout. One of the tests is configured for at most 120 seconds. Before my changes, the test required only 75 seconds, and after my changes they require around 120 seconds.

So, my changes introduced a dramatic performance loss on some settings.

The offending commit was the one than refactored function `DecodeExtKeyUsage`. I see nothing wrong with it other than it declares a bunch of extra local variables, adds one more function call to the chain, and makes some extra variable assignments through pointers. Before the squash is was `171bc7370 wolfcrypt: Make DecodeExtKeyUsage public`

The dramatic performance loss seems to affect those configurations that have defined `WOLFSSL_NO_MALLOC`, `WOLFSSL_STATIC_MEMORY` and `USE_FAST_MATH`. Zephyr is one of them.

I'm not sure what should be the next step. I'm not quite familiar with the low level details that may be affecting here. I will try declaring fewer variables to see if that helps.
(of course, another option would be to increase the timeout of the test. But I guess this is not the right solution)

Any help here will be very welcome

ribes96:
Last pushed commit should fix the Zephyr tests.
I expect the socat tests not to fail simply by chance

dgarske:
Jenkins retest this please

src/x509.c
```diff
@@ -2012,31 +1958,58 @@ void* wolfSSL_X509V3_EXT_d2i(WOLFSSL_X509_EXTENSION* ext)
...
+            asn1String = wolfSSL_X509_EXTENSION_get_data_internal(ext);
+            if (asn1String == NULL) {
+                WOLFSSL_MSG("X509_EXTENSION_get_data() failed");
+                return NULL;
+            }
+
+            InitDecodedCert(&cert, NULL, 0, NULL);
+            ret = DecodeExtensionType((const byte*)asn1String->data,
+                asn1String->length, object->type, (byte)ext->crit, &cert, NULL);
+            if (ret != 0) {
```
dgarske: Missing some edge cases to call `FreeDecodedCert`. Please make sure this is called as soon as cert is no longer needed and in all the error paths below.

src/x509.c
```diff
@@ -2075,32 +2041,45 @@ void* wolfSSL_X509V3_EXT_d2i(WOLFSSL_X509_EXTENSION* ext)
...
             if (asn1String == NULL) {
                 WOLFSSL_MSG("X509_EXTENSION_get_data() failed");
                 return NULL;
             }
+
+            InitDecodedCert(&cert, NULL, 0, NULL);
+            ret = DecodeExtensionType((const byte*)asn1String->data,
+                asn1String->length, object->type, (byte)ext->crit, &cert, NULL);
+            if (ret != 0) {
+                WOLFSSL_MSG("DecodeExtensionType() failed");
```
dgarske: Same... needs `FreeDecodedCert`.

src/x509.c
```diff
@@ -3288,14 +3267,151 @@ int wolfSSL_X509_EXTENSION_set_object(WOLFSSL_X509_EXTENSION* ext,
 }
 #endif /* OPENSSL_ALL */
 
+#ifdef WOLFSSL_OLD_EXTDATA_FMT
+/*
+ * Replace the current string in 'asn1str', which is the full X.509
+ * extension octet string with some data specific for the extension
+ * type. The extension is the one given in 'oid'
+ */
+static void wolfSSL_ASN1_STRING_into_old_ext_fmt (WOLFSSL_ASN1_STRING *asn1str, word32 oid)
```
dgarske: Please return int and capture return codes.

dgarske: 80 char max length please.
```
overlong lines added:
src/x509.c:3276 static void wolfSSL_ASN1_STRING_into_old_ext_fmt (WOLFSSL_ASN1_STRING *asn1str, word32 oid)
src/x509.c:3307                 data = (char*)XMALLOC((size_t)(extAuthKeyIdSz), NULL, DYNAMIC_TYPE_OPENSSL);
src/x509.c:3329                 data = (char*)XMALLOC((size_t)(extSubjKeyIdSz), NULL, DYNAMIC_TYPE_OPENSSL);
src/x509.c:3354                 wolfSSL_ASN1_STRING_set(asn1str, (byte*)&extKeyUsage, sizeof(word16));
src/x509.c:3367             ret = DecodeExtKeyUsage((const byte *)asn1str->data, asn1str->length,
src/x509.c:3373                 data = (char*)XMALLOC((size_t)(extExtKeyUsageSz), NULL, DYNAMIC_TYPE_OPENSSL);
wolfcrypt/src/asn.c:20673 static int DecodeBasicCaConstraintInternal(const byte* input, int sz, DecodedCert* cert)
wolfcrypt/src/asn.c:20679     ret = DecodeBasicCaConstraint(input, sz, &isCa, &pathLength, &pathLengthSet);
wolfcrypt/src/asn.c:21136 static int DecodeAuthKeyIdInternal(const byte* input, word32 sz, DecodedCert* cert)
wolfcrypt/src/asn.c:21271 static int DecodeKeyUsageInternal(const byte* input, word32 sz, DecodedCert* cert)
[...and more...]
```

src/x509.c
```diff
@@ -3288,14 +3267,151 @@ int wolfSSL_X509_EXTENSION_set_object(WOLFSSL_X509_EXTENSION* ext,
...
+            word32 serialSz = 0;
+
+            ret = DecodeAuthKeyId((const byte *)asn1str->data, asn1str->length,
+                    &extAuthKeyId, &extAuthKeyIdSz, &issuer, &issuerSz, &serial,
+                    &serialSz);
+
+            if (ret == 0)
+            {
+                char *data;
+                data = (char*)XMALLOC((size_t)(extAuthKeyIdSz), NULL, DYNAMIC_TYPE_OPENSSL);
```
dgarske: Why allocate a new buffer to call ASN1_STRING_set then just free it? Seems like you could just pass extAuthKeyId and extAuthKeyIdSz directly?

src/x509.c
```diff
@@ -3288,14 +3267,151 @@ int wolfSSL_X509_EXTENSION_set_object(WOLFSSL_X509_EXTENSION* ext,
...
+        {
+            int ret = 0;
+            const byte *extAuthKeyId = NULL;
+            word32 extAuthKeyIdSz = 0;
+            const byte *issuer = NULL;
+            word32 issuerSz = 0;
+            const byte *serial = NULL;
+            word32 serialSz = 0;
+
+            ret = DecodeAuthKeyId((const byte *)asn1str->data, asn1str->length,
```
dgarske, line 3304: Many set but not used variables here. Could DecodeAuthKeyId allow passing NULL's for the fields not needed?

src/x509.c
```diff
@@ -3288,14 +3267,151 @@ int wolfSSL_X509_EXTENSION_set_object(WOLFSSL_X509_EXTENSION* ext,
...
+            int ret = 0;
+            const byte *extSubjKeyId = NULL;
+            word32 extSubjKeyIdSz = 0;
+
+            ret = DecodeSubjKeyId((const byte *)asn1str->data, asn1str->length,
+                    &extSubjKeyId, &extSubjKeyIdSz);
+            if (ret == 0)
+            {
+                char *data;
+                data = (char*)XMALLOC((size_t)(extSubjKeyIdSz), NULL, DYNAMIC_TYPE_OPENSSL);
```
dgarske: Again, is it really required to allocate and free this? Can't you use extSubKeyId and extSubKeyIdSz directly?

src/x509.c
```diff
@@ -3288,14 +3267,151 @@ int wolfSSL_X509_EXTENSION_set_object(WOLFSSL_X509_EXTENSION* ext,
...
+
+        case KEY_USAGE_OID:
+        {
+            int ret = 0;
+            word16 extKeyUsage = 0;
+
+            ret = DecodeKeyUsage((const byte *)asn1str->data, asn1str->length,
+                    &extKeyUsage);
+            if (ret == 0)
+                wolfSSL_ASN1_STRING_set(asn1str, (byte*)&extKeyUsage, sizeof(word16));
```
dgarske: sizeof(extKeyUsage)

src/x509.c
```diff
@@ -3288,14 +3267,151 @@ int wolfSSL_X509_EXTENSION_set_object(WOLFSSL_X509_EXTENSION* ext,
...
+            byte extExtKeyUsage = 0;
+            byte extExtKeyUsageSsh = 0;
+
+            ret = DecodeExtKeyUsage((const byte *)asn1str->data, asn1str->length,
+                    &extExtKeyUsageSrc, &extExtKeyUsageSz, &extExtKeyUsageCount,
+                    &extExtKeyUsage, &extExtKeyUsageSsh);
+            if (ret == 0)
+            {
+                char *data;
+                data = (char*)XMALLOC((size_t)(extExtKeyUsageSz), NULL, DYNAMIC_TYPE_OPENSSL);
```
dgarske: Same, is malloc/free really required?

configure.ac
```diff
@@ -9806,6 +9806,16 @@ then
     fi
 fi
 
+AC_ARG_ENABLE([old-extdata-fmt],
```
dgarske, line 9813: Please add comment for when this old option would be needed. Add that it only applies to compatibility layer and certain API's.

dgarske (changes requested):

dgarske (changes requested):
Please add test case in api.c that tests getting a full octet string.

src/x509.c
ribes96: `extAuthKeyId` is just a reference to some offset in `asn1str->data`. We are replacing the old value of `asn1str->data` with `asn1str->data[some_offset]`. `wolfSSL_ASN1_STRING_set` can't be called with the same buffer for the first and third parameter, because it will first free the buffer in the first parameter, making the third parameter invalid.

dgarske (approved):

dgarske:

ribes96:
@dgarske as I suspected, the Zephyr tests are failing again due to to commit `97b2ef4f6 (Allow NULL parameters for 'DecodeAuthKeyId', 2025-08-03)`. Reverting it makes it pass again. Should I? The only thing that commit does is add a bunch of 'ifs', and that's enough to slown down dramatically the Zephyr tests

src/x509.c
```diff
@@ -1965,6 +1910,7 @@ void* wolfSSL_X509V3_EXT_d2i(WOLFSSL_X509_EXTENSION* ext)
     WOLFSSL_AUTHORITY_KEYID* akey;
     WOLFSSL_ASN1_STRING* asn1String, *newString;
     WOLFSSL_STACK* sk;
+    DecodedCert cert;
```
dgarske: You'll need to add a small stack option for DecodedCert...

```
#ifdef WOLFSSL_SMALL_STACK
    DecodedCert *cert;
#else
    DecodedCert cert[1];
#endif

#ifdef WOLFSSL_SMALL_STACK
    cert = (DecodedCert *)XMALLOC(sizeof(*cert), x509->heap,
                                  DYNAMIC_TYPE_DCERT);
    if (cert == NULL) {
        WOLFSSL_MSG("\tout of memory");
        return WOLFSSL_FATAL_ERROR;
    }
#endif
```

```
[all-crypto-linuxkm-defaults-max-func-stack-2k-build] [39 of 50] [17964e5ff8]
    configure...   real 0m9.909s  user 0m6.574s  sys 0m3.849s
    build...In file included from src/ssl.c:25238:
./src/x509.c: In function ‘wolfSSL_X509V3_EXT_d2i’:
./src/x509.c:2131:1: error: the frame size of 2560 bytes is larger than 2048 bytes [-Werror=frame-larger-than=]
 2131 | }
      | ^
cc1: all warnings being treated as errors
make[2]: *** [Makefile:8779: src/libwolfssl_la-ssl.lo] Error 1
make[1]: *** [Makefile:10241: all-recursive] Error 1
make: *** [Makefile:5741: all] Error 2
   real 0m14.307s  user 1m2.929s  sys 0m2.883s
    all-crypto-linuxkm-defaults-max-func-stack-2k-build fail_build
```

dgarske (changes requested):

ribes96:
haproxy failures have nothing to do with these commits. Repo [https://github.com/wlallemand/VTest](https://github.com/wlallemand/VTest) has been deleted

dgarske:
> @dgarske as I suspected, the Zephyr tests are failing again due to to commit `97b2ef4f6 (Allow NULL parameters for 'DecodeAuthKeyId', 2025-08-03)`. Reverting it makes it pass again. Should I? The only thing that commit does is add a bunch of 'ifs', and that's enough to slown down dramatically the Zephyr tests

Seems like the better fix is to allow more time on the test?

```
DEBUG   - QEMU (12086): PKCS12   test passed!
DEBUG   - QEMU (12086) complete (unexpected eof) after 119.97149276733398 seconds
DEBUG   - return code from QEMU (None): 0
DEBUG   - run status: qemu_x86/zephyr/samples/wolfssl_test/sample.crypto.wolfssl_test_no_malloc failed
INFO    - 1/1 qemu_x86                  zephyr/samples/wolfssl_test/sample.crypto.wolfssl_test_no_malloc  FAILED Timeout (qemu 119.971s)
```

ribes96:
> > @dgarske as I suspected, the Zephyr tests are failing again due to to commit `97b2ef4f6 (Allow NULL parameters for 'DecodeAuthKeyId', 2025-08-03)`. Reverting it makes it pass again. Should I? The only thing that commit does is add a bunch of 'ifs', and that's enough to slown down dramatically the Zephyr tests
> 
> Seems like the better fix is to allow more time on the test?
> 
> ```
> DEBUG   - QEMU (12086): PKCS12   test passed!
> DEBUG   - QEMU (12086) complete (unexpected eof) after 119.97149276733398 seconds
> DEBUG   - return code from QEMU (None): 0
> DEBUG   - run status: qemu_x86/zephyr/samples/wolfssl_test/sample.crypto.wolfssl_test_no_malloc failed
> INFO    - 1/1 qemu_x86                  zephyr/samples/wolfssl_test/sample.crypto.wolfssl_test_no_malloc  FAILED Timeout (qemu 119.971s)
> ```

Well, if it is not a problem that some workload will run considerably slower...

The required time increased from over 85 secs to over 140

dgarske:

src/x509.c
```diff
@@ -1147,137 +1055,145 @@ WOLFSSL_X509_EXTENSION* wolfSSL_X509_set_ext(WOLFSSL_X509* x509, int loc)
...
-                #ifdef WOLFSSL_SMALL_STACK
-                    XFREE(cert, NULL, DYNAMIC_TYPE_DCERT);
-                #endif
-                    return NULL;
-                }
-                oidBuf[0] = ASN_OBJECT_ID;
-                objSz++;
-                objSz += SetLength(length, oidBuf + 1);
-                objSz += length;
+        /* The ASN1_OBJECT in the extension is set in the same way
```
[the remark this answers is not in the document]
dgarske: @ribes96 please confirm the critical extension checking code is still present.

src/x509.c
```diff
@@ -2012,36 +1962,100 @@ void* wolfSSL_X509V3_EXT_d2i(WOLFSSL_X509_EXTENSION* ext)
...
                 return NULL;
             }
-            ret = wolfSSL_ASN1_STRING_set(newString, asn1String->data,
-                                                            asn1String->length);
+            ret = wolfSSL_ASN1_STRING_set(newString, cert->extSubjKeyId,
+                                          cert->extSubjKeyIdSz);
+            FreeDecodedCert(cert);
+#ifdef WOLFSSL_SMALL_STACK
+            XFREE(cert, NULL, DYNAMIC_TYPE_X509_EXT);
+#endif
```
julek-wolfssl: I would like to move everything involving `cert` outside of the `switch`. This pattern repeats in almost all the cases. `WC_NID_basic_constraints` for some reason is a special case so `cert` decoding can be omitted in that case. There should be only one of each call (`XMALLOC`, `InitDecodedCert`, `DecodeExtensionType`, `FreeDecodedCert`, `XFREE`) in this function. This means that there should be no `return` inside the `switch` and all cleanup should be done at the end of the function. Initialize all the stack variables to `NULL` for an even simpler cleanup section.

julek-wolfssl (changes requested):

dgarske:

wolfcrypt/src/asn.c
```diff
@@ -20670,100 +20670,25 @@ enum {
  * @return  ASN_EXPECT_0_E when the INTEGER has the MSB set or NULL has a
  *          non-zero length.
  */
-static int DecodeBasicCaConstraint(const byte* input, int sz, DecodedCert* cert)
+static int DecodeBasicCaConstraintInternal(const byte* input, int sz,
```
dgarske, line 20673: Can you please keep the old function located in same area of code to make diff easier?

wolfcrypt/src/asn.c
```diff
@@ -24170,7 +23907,559 @@ static int DecodeCertReq(DecodedCert* cert, int* criticalExt)
...
+ * @return  ASN_PARSE_E when CA boolean is not present unless
+ *          WOLFSSL_X509_BASICCONS_INT is defined. Only a CA extension.
+ * @return  ASN_PARSE_E when path length more than 7 bits.
+ * @return  ASN_PARSE_E when BER encoded data does not match ASN.1 items or
+ *          is invalid.
+ * @return  BUFFER_E when data in buffer is too small.
+ * @return  ASN_EXPECT_0_E when the INTEGER has the MSB set or NULL has a
+ *          non-zero length.
+ */
+WOLFSSL_LOCAL int DecodeBasicCaConstraint(const byte* input, int sz,
```
dgarske: Please remove WOLFSSL_LOCAL on the .c files. Its only required for the .h files.

dgarske:
Jenkins retest this please: FIPS test failed with "RequestAbortedException"

dgarske:

dgarske (approved):

PR #8969, opened 2025-07-06
ALPN: don't use BIO

SparkiDev:
# Description

Fix wolfSSL_set_alpn_protos to not use BIO.
When compiling with -Os and newer gcc, the compiler gets confused with the void* cast in the wolfSSL_BIO_get_mem_data call.

# Testing

./configure --disable-shared --enable-keygen --enable-sp-math-all=small --enable-all

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


SparkiDev:
Changed code calling wolfSSL_UseALPN() to check return value.
(Original code did not.)

SparkiDev:
retest this please

dgarske (approved):

PR #8970, opened 2025-07-07
Fix Qt nightly Jenkins failure

miyazakh:
# Description
Fix Qt nightly jenkins failure
`hashType` default is `MD_5` to follow OpenSSL default. For PBKDF1, Its `hashType` is `MD_5` or `MD_SHA`. It also needs to have the matched `hashType` when decrypting key.
 
# Testing
Added New unit test. Qt Jenkins test.

# Checklist

 - [x] added tests
 - [ ] updated/added doxygen


miyazakh:
retest this please

`PRB-fsanitize-addr-v3` Timeout has been exceeded

src/pk.c
```diff
@@ -507,12 +507,10 @@ static int der_to_enc_pem_alloc(unsigned char* der, int derSz,
     byte* cipherInfo = NULL;
     int pemSz = 0;
     int hashType = WC_HASH_TYPE_NONE;
-#if !defined(NO_SHA256)
-    hashType = WC_SHA256;
+#if !defined(NO_MD5)
+    hashType = WC_MD5;
 #elif !defined(NO_SHA)
```
dgarske, line 512: Any reason we can't have a SHA256 option here too?

dgarske (changes requested):

src/pk.c
miyazakh: Regarding to [RFC8018(section 5.1)](https://datatracker.ietf.org/doc/html/rfc8018#section-5.1), PBKDF1 supports `MD2`, `MD5` or `SHA-1`, but not `SHA-2`. Therefore, I remove `SHA256`.

dgarske (approved):

PR #8976, opened 2025-07-09
Add wc_PKCS7_DecodeEncryptedKeyPackage()

holtrop:
# Description

Add wc_PKCS7_DecodeEncryptedKeyPackage()

# Testing

Unit tested

# Checklist

 - [X] added tests
 - [X] updated/added doxygen


holtrop:
retest this please (Error cloning remote repo 'origin')

holtrop:
retest this please (SSL_read input error -308, error state on socket)

dgarske:
```
overlong lines added:
wolfcrypt/src/pkcs7.c:14881 /* Unwrap and decrypt PKCS#7/CMS EncryptedKeyPackage object, return decoded size. */
wolfcrypt/src/pkcs7.c:14896         /* Expect a SEQUENCE header to start the EncryptedKeyPackage ContentInfo. */
wolfcrypt/src/pkcs7.c:14921         /* Check for EncryptedKeyPackage explicit CHOICE [0] tag, indicating EnvelopedData subtype. */
wolfcrypt/src/pkcs7.c:14925              * EnvelopedData ContentInfo object within the EncryptedKeyPackage. */
wolfcrypt/src/pkcs7.c:14932              * support AuthEnvelopedData, so check if we have an EncryptedData blob. */
```

holtrop:

wolfcrypt/src/pkcs7.c
```diff
@@ -14878,6 +14878,74 @@ int wc_PKCS7_SetDecodeEncryptedCtx(wc_PKCS7* pkcs7, void* ctx)
...
+ * decoded size. */
+WOLFSSL_API int wc_PKCS7_DecodeEncryptedKeyPackage(wc_PKCS7 * pkcs7,
+        byte * pkiMsg, word32 pkiMsgSz, byte * output, word32 outputSz)
+{
+    int ret = 0;
+    word32 pkiIndex = 0;
+    word32 contentType = 0;
+    int length = 0;
+
+    do {
```
JacobBarthelmeh: Please use checks on ret value instead of the use of break like a goto, for ease of maintaining the code. Removing the do {} while(0);

wolfcrypt/src/pkcs7.c
```diff
@@ -14878,6 +14878,74 @@ int wc_PKCS7_SetDecodeEncryptedCtx(wc_PKCS7* pkcs7, void* ctx)
 #endif /* NO_PKCS7_ENCRYPTED_DATA */
 
 
+/* Unwrap and decrypt PKCS#7/CMS EncryptedKeyPackage object, return the
+ * decoded size. */
+WOLFSSL_API int wc_PKCS7_DecodeEncryptedKeyPackage(wc_PKCS7 * pkcs7,
```
JacobBarthelmeh: WOLFSSL_API is only needed in the header file

wolfcrypt/src/pkcs7.c
```diff
@@ -14878,6 +14878,74 @@ int wc_PKCS7_SetDecodeEncryptedCtx(wc_PKCS7* pkcs7, void* ctx)
...
+
+        /* Check for an EncryptedKeyPackage explicit CHOICE [0] tag, indicating
+         * an EnvelopedData subtype. */
+        if (GetASNHeader(pkiMsg, ASN_CONTEXT_SPECIFIC | ASN_CONSTRUCTED,
+                    &pkiIndex, &length, pkiMsgSz) >= 0) {
+            /* An explicit CHOICE [0] tag was found. pkiIndex now should point
+             * to the EnvelopedData ContentInfo object within the
+             * EncryptedKeyPackage. */
+            ret = wc_PKCS7_DecodeEnvelopedData(pkcs7, &pkiMsg[pkiIndex],
+                    pkiMsgSz - pkiIndex, output, outputSz);
```
JacobBarthelmeh: Possible underflow? pkiMsgSz - pkiIndex. If so we could use a check that pkiIndex is not larger than pkiMsgSz.

JacobBarthelmeh (changes requested):

wolfcrypt/src/pkcs7.c
holtrop: Oh ok. I followed wc_PKCS7_DecodeEnvelopedData() example from the C file. Should I also remove WOLFSSL_API from the other functions in the C file?

wolfcrypt/src/pkcs7.c
holtrop: Ok. My thought was to keep the compiled size down by doing it this way. Is that not worth worrying about? Would an if .. else if chain be acceptable to avoid having to check ret == 0 many times?

wolfcrypt/src/pkcs7.c
JacobBarthelmeh: No need to change other cases in the source code for this PR. Though it should be changed to be consistent (having WOLFSSL_LOCAL/WOLFSSL_API on function declaration not implementation), I can open a PR for that.

wolfcrypt/src/pkcs7.c
holtrop: It looks like GetSequence/GetASNHeader and friends take care of checking and not consuming more of pkiMsg than pkiMsgSz. I think that makes sense for them to do so they won't advance pkiIndex beyond pkiMsgSz. And I don't advance pkiIndex myself directly in this function; it's only advanced with one of those consuming "Get" functions.

wolfcrypt/src/pkcs7.c
JacobBarthelmeh: if .. else is okay depending on how deep the number of if's go....

Code size optimizations are great! I'd like to avoid the do {} while(0); with breaks to jump out though. To help me understand more about the size tradeoff vs code change, how big of a difference was the code size for the function?

holtrop: It looks like the function was 0x1A9 bytes with the do/while(0) and 0x1B4 bytes with the ret == 0 checks, although I had --enable-debug on to debug the unit test. It's not a huge difference, at least with my compiler and target CPU. If I reconfigure without --enable-debug though, then it's 0x15C for both implementations. And the if/else if chain is the same size as the do/while(0) in either case.

holtrop:

JacobBarthelmeh (approved):

Issue #8951, opened 2025-07-02
Outside array bounds warnings from GCC: wolfcrypt/src/sp_int.c


tjko:
### Version

5.8.0

### Description

I noticed following warnings when compiling wolfSSL:

```
In function '_sp_init_size',
    inlined from 'sp_init_size' at /home/tjko/git/fanpico/libs/wolfssl/wolfcrypt/src/sp_int.c:5126:9,
    inlined from '_sp_sqrmod' at /home/tjko/git/fanpico/libs/wolfssl/wolfcrypt/src/sp_int.c:17303:15,
    inlined from 'sp_sqrmod' at /home/tjko/git/fanpico/libs/wolfssl/wolfcrypt/src/sp_int.c:17359:15,
    inlined from '_sp_exptmod_ex' at /home/tjko/git/fanpico/libs/wolfssl/wolfcrypt/src/sp_int.c:13055:19:
/home/tjko/git/fanpico/libs/wolfssl/wolfcrypt/src/sp_int.c:5105:13: warning: array subscript 'sp_int[0]' is partly outside array bounds of 'sp_int[0]' [-Warray-bounds=]
 5105 |     a->size = (sp_size_t)size;
      |     ~~~~~~~~^~~~~~~~~~~~~~~~~
In file included from /home/tjko/git/fanpico/libs/wolfssl/wolfssl/wolfcrypt/libwolfssl_sources.h:46,
                 from /home/tjko/git/fanpico/libs/wolfssl/wolfcrypt/src/sp_int.c:30:
In function '_sp_sqrmod',
    inlined from 'sp_sqrmod' at /home/tjko/git/fanpico/libs/wolfssl/wolfcrypt/src/sp_int.c:17359:15,
    inlined from '_sp_exptmod_ex' at /home/tjko/git/fanpico/libs/wolfssl/wolfcrypt/src/sp_int.c:13055:19:
/home/tjko/git/fanpico/libs/wolfssl/wolfssl/wolfcrypt/types.h:577:61: note: object of size [8, 772] allocated by 'malloc'
  577 |         #define XMALLOC(s, h, t)     ((void)(h), (void)(t), malloc((size_t)(s))) /* native heap */
      |                                                             ^~~~~~~~~~~~~~~~~~~
/home/tjko/git/fanpico/libs/wolfssl/wolfcrypt/src/sp_int.c:148:28: note: in expansion of macro 'XMALLOC'
[...]

PR #8985, opened 2025-07-10
improper access of sp_int_minimal using sp_int

sebastian-carpenter:
# Description

User was getting a compiler warning about an out of bounds access for an array. This turned out to be mostly unrelated to arrays as the issue appears due to a pointer dereference.

It would seem that an sp_int_minimal must be used instead of an sp_int and this fixes the issue.

related thread: https://bugzilla.redhat.com/show_bug.cgi?id=2047439

Also found an sp_uint8 in the sp_int_minimal struct that was not updated to sp_sign_t. This is unrelated to the compiler warning.

Fixes GH issue #8951

# Testing

GH issue #8951 describes the procedure for reproducing the issue.

I ended up using their fanpico project to build and test. Run cmake and use make to get the warning.

Some dependencies that were necessary to get cmake working for this project:
* arm-none-eabi-binutils-cs
* arm-none-eabi-gcc-cs
* arm-none-eabi-gcc-cs-c++
* arm-none-eabi-newlib

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/wolfssl/wolfcrypt/sp_int.h
+++ b/wolfssl/wolfcrypt/sp_int.h
@@ -932,7 +932,7 @@ typedef struct sp_int_minimal {
     sp_size_t    size;
 #ifdef WOLFSSL_SP_INT_NEGATIVE
     /** Indicates whether number is 0/positive or negative.  */
-    sp_uint8     sign;
+    sp_sign_t    sign;
 #endif
 #ifdef HAVE_WOLF_BIGINT
     /** Unsigned binary (big endian) representation of number. */
--- a/wolfcrypt/src/sp_int.c
+++ b/wolfcrypt/src/sp_int.c
@@ -5102,7 +5102,7 @@ static void _sp_init_size(sp_int* a, unsigned int size)
 #endif
     _sp_zero((sp_int*)am);
 
-    a->size = (sp_size_t)size;
+    am->size = (sp_size_t)size;
 }
 
 /* Initialize the multi-precision number to be zero with a given max size.


dgarske:
Retest this please: "Found unhandled hudson.remoting.RequestAbortedException exception: java.io.StreamCorruptedException: invalid stream header: 636F7272"

wolfssl/wolfcrypt/sp_int.h
```diff
@@ -932,7 +932,7 @@ typedef struct sp_int_minimal {
     sp_size_t    size;
 #ifdef WOLFSSL_SP_INT_NEGATIVE
     /** Indicates whether number is 0/positive or negative.  */
-    sp_uint8     sign;
+    sp_sign_t    sign;
```
SparkiDev, line 935: The sign is positive or negative: 1 bit.
sp_size_t is used for the size of the count of words.
Different uses. Why change the type?

sebastian-carpenter: When I looked at the declaration for `sp_int` the `sign` variable had type `sp_sign_t`. It seemed to me that these changes should be carried over to `sp_int_minimal` as well so I made its `sign` variable an `sp_sign_t` too.

SparkiDev: Sorry, not reading the code right.
sp_sign_t is correct.

dgarske (approved):

PR #8991, opened 2025-07-10
Gate some tests on MD5 because PKCS8 wants to use MD5 by default

anhu:


SparkiDev:
What configuration fails?
Put into the description.

dgarske:
@anhu please add the configure to reproduce this issue to the PR description.

anhu:
I can't seem to remember how I came about needing to do this.  I'm going assign this to myself and remove the review.

anhu:
Closing for now as there seems to be no way to reproduce what this was trying to fix.

PR #8993, opened 2025-07-11
Make properties related to TLS handshake hidden for TSIP TLS user-context structure

miyazakh:
# Description
Make properties related to TLS handshake hidden for TSIP TLS user-context structure so that user just takes account of public properties when needed. E.g. Crypto API use.

# Testing

Example program including Crypt, Bench, Client/Server and Unit test cases in the repository

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/src/keys.c
+++ b/src/keys.c
@@ -3559,7 +3559,7 @@ int SetKeysSide(WOLFSSL* ssl, enum encrypt_side side)
             cbInfo->side = side;
         #elif defined(WOLFSSL_RENESAS_TSIP_TLS)
             TsipUserCtx* cbInfo = (TsipUserCtx*)ctx;
-            cbInfo->key_side = side;
+            _ACCESSOR(cbInfo)->key_side = side;
         #endif
         ret = ssl->ctx->EncryptKeysCb(ssl, ctx);
     }
--- a/wolfssl/wolfcrypt/port/Renesas/renesas_tsip_internal.h
+++ b/wolfssl/wolfcrypt/port/Renesas/renesas_tsip_internal.h
@@ -1,0 +1,171 @@
+/* renesas_tsip_internal.h
+ *
+ * Copyright (C) 2006-2025 wolfSSL Inc.
+ *
+ * This file is part of wolfSSL.
+ *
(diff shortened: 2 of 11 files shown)


dgarske:
Retest this please

wolfssl/wolfcrypt/port/Renesas/renesas_tsip_internal.h
```diff
@@ -0,0 +1,171 @@
+/* renesas_tsip_internal.h
```
dgarske, line 1: Make sure new files are added to include.am so they can be part of the release process.

src/keys.c
```diff
@@ -3559,7 +3559,7 @@ int SetKeysSide(WOLFSSL* ssl, enum encrypt_side side)
             cbInfo->side = side;
         #elif defined(WOLFSSL_RENESAS_TSIP_TLS)
             TsipUserCtx* cbInfo = (TsipUserCtx*)ctx;
-            cbInfo->key_side = side;
+            _ACCESSOR(cbInfo)->key_side = side;
```
dgarske: From the .c side it seems like using _ACCESSOR() is just not needed. Why not reference the internal directly and avoid the macro? If a macro is needed please use something more specific like WC_WSIP_STUCT

dgarske (changes requested):

src/keys.c
miyazakh: There was no specific reason. Just a personal preference that I thought avoiding multiple arrow operators, increase readability.

miyazakh: Replace them to `xxx->internal`

wolfssl/wolfcrypt/port/Renesas/renesas_tsip_internal.h
miyazakh: Thanks. Fixed

miyazakh:
retest this please

Timeout

miyazakh:
retest this please

`Found unhandled org.jenkinsci.plugins.workflow.support.steps.AgentOfflineException exception`

SparkiDev (approved):

PR #8995, opened 2025-07-11
reduce check size to BLAKE2B_BLOCKBYTES

rlm2002:
# Description

Initial check was too large and could lead to attempt to access buffer at location outside of buffer. Note that `byte buf [2 * BLAKE2B_BLOCKBYTES]`, but max length of `S->buflen` could be 255 according to original change.

The original change caused new Coverity issue CID 531298 Out-of-bounds access. This change is expected to fix CID 218595 Overlapping buffer in memory copy as well. However, changing XMEMCPY to XMEMMOVE may be necessary to remove issue rather than ignoring it.


rlm2002:
Addressed by #9003 closing this draft.

PR #8996, opened 2025-07-11
Add logic to match IPv6 domain addresses

lealem47:
# Description

This PR fixes a bug where we fail to match IPv6 domain names when we should. For example   `1002:097c:C2BB:0820:0010:0240:0115:0083` to `1002:97c:c2bb:820:10:240:115:84` or even `0000:0000:0000:0000:0000:0000:0000:0001` to `::1` . This was leading valid TLS connections to fail. The new logic uses inet_pton to handle all the parsing and makes matching the addresses easier. Guarded by `WOLFSSL_IP_ALT_NAME` 

Fixes zd#20210

# Testing

Tested with curl built against wolfssl. I've attached the script I used to test. The TLS connection would previously fail, and should now work.
[ipv6_test.zip](https://github.com/user-attachments/files/21188904/ipv6_test.zip)

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -12947,6 +12947,39 @@ int CipherRequires(byte first, byte second, int requirement)
 #endif /* !NO_TLS */
 
 #ifndef NO_CERTS
+#ifdef WOLFSSL_IP_ALT_NAME
+static int MatchIPv6(const char* pattern, int patternLen,
+                     const char* str, word32 strLen)
+{
+    WOLFSSL_SOCKADDR_IN6 addr1, addr2;
+    char patBuf[WOLFSSL_MAX_IPSTR] = {0};
+    char strBuf[WOLFSSL_MAX_IPSTR] = {0};
+
+    if ((word32)patternLen >= sizeof(patBuf) || strLen >= sizeof(strBuf))
+        return 0;
+
+    XMEMSET(patBuf, 0, WOLFSSL_MAX_IPSTR);
+    XMEMSET(strBuf, 0, WOLFSSL_MAX_IPSTR);
+
+    /* Make sure strings are null-terminated and safely copied */
+    XMEMCPY(patBuf, pattern, patternLen);
+    patBuf[patternLen] = '\0';
+    XMEMCPY(strBuf, str, strLen);
+    strBuf[strLen] = '\0';
+
+    XMEMSET(&addr1, 0, sizeof(addr1));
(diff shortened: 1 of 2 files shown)


lealem47:
Jenkins retest this please

SparkiDev (approved):

src/internal.c
```diff
@@ -12947,6 +12947,39 @@ int CipherRequires(byte first, byte second, int requirement)
...
+                     const char* str, word32 strLen)
+{
+    WOLFSSL_SOCKADDR_IN6 addr1, addr2;
+    char patBuf[WOLFSSL_MAX_IPSTR];
+    char strBuf[WOLFSSL_MAX_IPSTR];
+
+    if ((word32)patternLen >= sizeof(patBuf) || strLen >= sizeof(strBuf))
+        return 0;
+
+    XMEMSET(patBuf, 0, WOLFSSL_MAX_IPSTR);
```
SparkiDev, line 12961: Is this needed?
You put a '\0' at then end and have checked length.

SparkiDev (dismissed):

lealem47:
Jenkins retest this please

Error in PRB-generic-config-parser
```
java.io.StreamCorruptedException: invalid stream header: 496E636F
```

src/internal.c
```diff
@@ -12947,6 +12947,36 @@ int CipherRequires(byte first, byte second, int requirement)
 #endif /* !NO_TLS */
 
 #ifndef NO_CERTS
+#if defined(WOLFSSL_IP_ALT_NAME) && !defined(WOLFSSL_USER_IO)
```
dgarske, line 12950: Should any of this IPv6 logic be wrapped with `WOLFSSL_IPV6`?

dgarske (changes requested):

src/internal.c
lealem47: From the prompt in configure, I thought that macro was only relevant to testing. Looking through the code, I see that's not the case but it seems to be used primarily in connectivity functions. I don't think we should require it in areas relevant to security

```
# IPv6 Test Apps
AC_ARG_ENABLE([ipv6],
    [AS_HELP_STRING([--enable-ipv6],[Enable testing of IPV6 (default: disabled)])],
    [ ENABLED_IPV6=$enableval ],
    [ ENABLED_IPV6=no ]
    )
```

dgarske (approved):

PR #9000, opened 2025-07-14
Import ML-DSA's seed from PKCS8 file

kojiws:
# Description

This PR will support import of ML-DSA's seed from PKCS8 file.
The importing seed is only supported when `WOLFSSL_ASN_TEMPLATE` is declared.

Note: Some tests of `wc_Dilithium_PrivateKeyToDer()` and `wc_Dilithium_KeyToDer()`, which invokes `SetAsymKeyDer()`, fails but it's not related to this PR. For the temporal solution, `SetAsymKeyDer()` returns `BAD_FUNC_ARG` for key length equal to or longer than 128 bytes without `WOLFSSL_ASN_TEMPLATE`. This will be fixed by next PR. 

# Testing
./configure --enable-dilithium --enable-wolfclu && make test
./configure --enable-dilithium --enable-wolfclu --enable-asn=original && make test

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


wolfcrypt/src/dilithium.c
```diff
@@ -9770,19 +9773,26 @@ int wc_Dilithium_PrivateKeyDecode(const byte* input, word32* inOutIdx,
                 ret = wc_dilithium_set_level(key, (byte)ret);
             }
         }
+
+        (void) dilithium_get_priv_size;
         /* If it failed to decode try alternative DER encoding. */
+        /* This should be done within DecodeAsymKey_Assign */
```
SparkiDev: Yep, remove this code now.

wolfcrypt/src/dilithium.c
```diff
@@ -9770,19 +9773,26 @@ int wc_Dilithium_PrivateKeyDecode(const byte* input, word32* inOutIdx,
...
+        /*
         else if (ret != 0) {
             word32 levelSize = dilithium_get_priv_size(key->level);
             privKey = input + *inOutIdx;
             privKeyLen = inSz - *inOutIdx;
+        */
 
             /* Check for an alternative DER encoding. */
+        /*
```
SparkiDev: And this looks handled too,
Remove

kojiws:
`WOLFSSL_LOCAL` means not external one, so no problem.

JacobBarthelmeh (approved):

PR #9014, opened 2025-07-17
Fix GetShortInt/SetShortInt handling of negative values

holtrop:
# Description

This PR fixes GetShortInt/SetShortInt handling of negative INTEGER values.

1. GetShortInt was not correctly decoding negative INTEGER values. This has been fixed.
2. SetShortInt was not correctly encoding negative INTEGER values. This has been fixed.
3. SetShortInt took an unsigned word32 input even though every call to it actually used a signed value and casted it to word32. SetShortInt now takes a sword32 input instead of word32.
4. GetShortInt returned the decoded INTEGER value both via an output pointer and as the function return value. Once it was fixed to properly decode negative INTEGER values, the return value could have conflicted with an actual error code. The return value is now 0 on success or an error code and the decoded INTEGER value is only returned via pointer.
5. GetShortInt returned the decoded INTEGER value via an int pointer. It now returns via sword32 pointer.

# Testing

How did you test?

# Checklist

 - [X] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/wolfcrypt/src/asn.c
+++ b/wolfcrypt/src/asn.c
@@ -1289,11 +1289,23 @@ static int GetASN_StoreData(const ASNItem* asn, ASNGetData* data,
             #endif
                 return ASN_PARSE_E;
             }
+            else if (zeroPadded && len == 2 && (input[idx] & 0x80U) != 0U) {
+                /* Value is >= 0x8000 which is too large for a sword16. */
+            #ifdef WOLFSSL_DEBUG_ASN_TEMPLATE
+                WOLFSSL_MSG_VSNPRINTF("Value too large for sword16");
+            #endif
+                return ASN_PARSE_E;
+            }
             /* Fill number with all of data. */
-            *data->data.u16 = 0;
-            for (i = 0; i < len; i++) {
-                *data->data.u16 = (word16)(*data->data.u16 << 8U);
-                *data->data.u16 = (word16)(*data->data.u16 | input[idx + (word32)i]);
+            else {
+                sword16 v = 0;
+                if (!zeroPadded && (input[idx] & 0x80U) != 0U) {
+                    v = -1;
+                }
+                for (i = 0; i < len; i++) {
+                    v = (sword16)(v << 8U) | (sword16)(word16)input[idx + (word32)i];
+                }
+                *data->data.u16 = (word16)v;
             }
             break;
         case ASN_DATA_TYPE_WORD32:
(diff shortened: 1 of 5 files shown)


wolfcrypt/src/asn.c
```diff
@@ -1304,11 +1316,23 @@ static int GetASN_StoreData(const ASNItem* asn, ASNGetData* data,
             #endif
                 return ASN_PARSE_E;
             }
+            else if (zeroPadded && len == 4 && (input[idx] & 0x80U) != 0U) {
```
SparkiDev, line 1319: ASN_DATA_TYPE_WORD16 is unsigned.
Add ASN_DATAT_TYPE_SWORD16 if you want to handle signed.

wolfcrypt/src/asn.c
```diff
@@ -3211,15 +3235,14 @@ int GetMyVersion(const byte* input, word32* inOutIdx,
  * @return  BUFFER_E when data in buffer is too small.
  * @return  ASN_EXPECT_0_E when the most significant bit is set.
  */
-int GetShortInt(const byte* input, word32* inOutIdx, int* number, word32 maxIdx)
+int GetShortInt(const byte* input, word32* inOutIdx, sword32 * number,
```
SparkiDev, line 3238: Add a new API GetSignedShortInt().
This API is for an unsigned value.

holtrop: This API was returning its value via an int, which is signed. Wouldn't it make sense to just have one API to decode a short int rather than two?

SparkiDev: The implementation never properly handled signed.
If you don't want a second API then add a flag to this API to say the caller will accept a signed value out.
Lots of places it must be positive.

holtrop: > The implementation never properly handled signed.

Right, that's what I fixed.

> If you don't want a second API then add a flag to this API to say the caller will accept a signed value out. Lots of places it must be positive.

This API is called in five places. It looks like two of the five call sites don't even use the retrieved INTEGER value, one just checks that it is equal to 0, and the other two use the retrieved value (as iteration count values). For those last two, I could add a check that the retrieved value is positive.

Do you think if I add those two checks it would be sufficient to not need to add a new API or a new flag parameter?

SparkiDev: Iteration count can't be negative. key size can't be negative.
Looking for 0 is most likely the version and must be positive.
Is there anything that wants a negative value?
If not, then return an error when the value is negative in this function.

PR #9020, opened 2025-07-21
Avoid bogus warning on uninitialized variables on old versions of GCC

ribes96:
gcc-4.3.3 erroneously complains that some variables may be used uninitialized. Silence it assigning NULL on declaration, as is already done with many other variables.

# Description

Compiling with `gcc-4.3.3` I get the following errors:

~~~log
src/ssl.c: In function 'wolfSSL_UseALPN':
src/ssl.c:3858: error: 'ptr' may be used uninitialized in this function

tests/api.c: In function 'test_wolfSSL_EC_KEY_private_key':
tests/api.c:56471: error: 'bn' may be used uninitialized in this function

tests/api.c: In function 'test_sk_X509_CRL':
tests/api.c:47342: error: 'sn' may be used uninitialized in this function
~~~

They are all false positives, but it can be easily avoided by assigning NULL on initialization

Assign NULL to variables when they are declared

Fixes zd#

# Testing

How did you test?

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/src/ssl.c
+++ b/src/ssl.c
@@ -3855,7 +3855,7 @@ WOLFSSL_ABI
 int wolfSSL_UseALPN(WOLFSSL* ssl, char *protocol_name_list,
                     word32 protocol_name_listSz, byte options)
 {
-    char    *list, *ptr, **token;
+    char    *list, *ptr = NULL, **token;
     word16  len;
     int     idx = 0;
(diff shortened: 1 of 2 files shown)


dgarske:
Thank you @ribes96! Jenkins okay to test. Contributor agreement on file.

dgarske:

SparkiDev (approved):

PR #9021, opened 2025-07-21
Support CFLAGS="-Wno-shadow"

ribes96:
Avoid appending "-Wshadow" in the end of compiler flags if the user provided CFLAGS="-Wno-shadow"

# Description

Old versions of `gcc` are stricter about local variables shadowing global declarations, and generate warnings like this:
~~~log
tests/api.c: In function 'test_wolfSSL_ASN1_INTEGER':
tests/api.c:19508: error: declaration of 'dup' shadows a global declaration
~~~
Fixing  these cases is easy, but since modern versions don't complain, they keep appearing from time to time.
Users of old compilers could not use `CFLAGS="-Wno-shadow"` because it was overwritten by the hardening flags.

Now, "-Wshadow" will **NOT** be appended if the user has provided `CFLAGS="-Wno-shadow"`

Fixes zd#

# Testing

How did you test?

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


dgarske:
Thank you @ribes96! Jenkins okay to test. Contributor agreement on file.

dgarske:

PR #9024, opened 2025-07-21
fix mldsa test case for buffer size and expire date

JacobBarthelmeh:
Size of certificate being signed could be larger than the test buffer. Steps to hit the failing test case is:

```
./configure --enable-kyber --enable-dilithium --enable-all CC="clang -g -fsanitize=address"
make
for i in {1..10000}; do echo "Test $i"; ./tests/unit.test -372 -373 &> stdout.txt; if [ "$?" != "0" ]; then echo "Caught a failure"; break; fi done
```


JacobBarthelmeh:
Retest this please Jenkins:

```
```

dgarske (approved):

PR #9028, opened 2025-07-23
Fixes for building with MD5 and SHA1 to support Hash `WC_HASH_TYPE_MD5_SHA`

dgarske:
# Description

Fixes for building with MD5 and SHA1 to support Hash `WC_HASH_TYPE_MD5_SHA`. ZD 20269.

Max digest for MD5 (16) + SHA1 (20) = 36, which is larger than SHA256 (32). Can cause buffer overrun in some configurations if `WC_HASH_TYPE_MD5_SHA` is used (only for TLS v1.1 and older).

# Testing

```
./configure 'CC=clang -fsanitize=address -g -fno-omit-frame-pointer' --enable-usersettings --enable-debug --disable-shared && make check
```

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


dgarske:
Jenkins retest this please: "scripts/google.test failed"

JacobBarthelmeh (approved):

PR #9036, opened 2025-07-25
Fix size used by signature context struct with WOLFSSL_NO_MALLOC.

kareem-wolfssl:
This matches the size used by sigCpy/sigSz when building without WOLFSSL_NO_MALLOC.

# Description

Fixes zd#20239

# Testing

Built in tests.

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/wolfssl/wolfcrypt/asn.h
+++ b/wolfssl/wolfcrypt/asn.h
@@ -1383,7 +1383,7 @@ struct SignatureCtx {
 #endif
 #if !defined(NO_RSA) || !defined(NO_DSA)
     #ifdef WOLFSSL_NO_MALLOC
-    byte  sigCpy[MAX_SIG_SZ];
+    byte  sigCpy[MAX_ENCODED_SIG_SZ];
     #else
     byte* sigCpy;
     #endif


dgarske (dismissed):

wolfssl/wolfcrypt/asn.h
```diff
@@ -1383,7 +1383,7 @@ struct SignatureCtx {
 #endif
 #if !defined(NO_RSA) || !defined(NO_DSA)
     #ifdef WOLFSSL_NO_MALLOC
-    byte  sigCpy[MAX_SIG_SZ];
```
dgarske, line 1386: This leaves `MAX_SIG_SZ` dead (not used). Should we leave it or remove it?

kareem-wolfssl: Before the original change that used MAX_SIG_SZ in sigCpy, MAX_SIG_SZ was present in enum Misc_ASN.  Even though wolfSSL wasn't using this constant, it's possible external applications are.  However this is in asn.h rather than asn_public.h, so it's most likely safe to remove.

dgarske: PR Cap @JacobBarthelmeh thoughts?

JacobBarthelmeh: My thoughts are it should be removed if not used, to avoid confusion in the future.

JacobBarthelmeh:
Retest this please Jenkins.

Timeout was exceded.

```
Found unhandled org.jenkinsci.plugins.workflow.support.steps.AgentOfflineException exception
```

PR #9039, opened 2025-07-30
Add _new/_delete API for ML-KEM/ML-DSA

tamasan238:
# Description

Added below to solve trouble allocating memory for the MlKemKey structure in C#.
Since the size of the MlKemKey structure depends on the build options, these are needed.

- `wolfCrypt/wc_mlkem.c`: `wc_MlKemKey_New()`, `wc_MlKemKey_Delete()`
- `wolfCrypt/wc_dilithium.c`: `wc_dilithium_new()`, `wc_dilithium_delete()`

This PR is related to another PR I'll add in the comments.

I'm not familiar with adding code to wolfCrypt.
Please feel free to point out any inappropriate parts.

# Testing

Tested.

# Checklist

 - [x] added tests
 - [x] updated/added doxygen
 - [ ] updated appropriate READMEs
 - [ ] Updated manual and documentation

PR #9040, opened 2025-07-30
Add ML-KEM/ML-DSA support for C# wrapper

tamasan238:
# Description

Implementing ML-KEM and ML-DSA APIs in the C# wrapper.
For this PR to work, you need to merge other PRs that you will add in comments.

# Testing

Tested.

# Checklist

 - [x] added tests
 - [x] updated/added doxygen

### DIFF
--- a/wrapper/CSharp/wolfCrypt-Test/wolfCrypt-Test.cs
+++ b/wrapper/CSharp/wolfCrypt-Test/wolfCrypt-Test.cs
@@ -648,6 +648,289 @@ public class wolfCrypt_Test_CSharp
         if (publicKeyB != IntPtr.Zero) wolfcrypt.Curve25519FreeKey(publicKeyB);
     } /* END curve25519_test */
 
+    private static void mlkem_test(wolfcrypt.MlKemTypes type)
+    {
+        int ret = 0;
+        IntPtr keyA = IntPtr.Zero;
+        IntPtr keyB = IntPtr.Zero;
+        IntPtr heap = IntPtr.Zero;
+        int devId = wolfcrypt.INVALID_DEVID;
+        byte[] pubA, privA, cipherText, sharedSecretA, sharedSecretB;
+
+        try
+        {
+            Console.WriteLine("\nStarting " + type + " shared secret test ...");
+
+            /* Generate Key Pair */
+            Console.WriteLine("Testing ML-KEM Key Generation...");
+
+            Console.WriteLine("Generate Key Pair A...");
+            keyA = wolfcrypt.MlKemMakeKey(type, heap, devId);
+            if (keyA == IntPtr.Zero)
+            {
+                ret = -1;
+                Console.Error.WriteLine("Failed to generate key pair A.");
+            }
+            if (ret == 0)
+            {
+                Console.WriteLine("Generate Key Pair B...");
+                keyB = wolfcrypt.MlKemMakeKey(type, heap, devId);
+                if (keyB == IntPtr.Zero)
+                {
+                    ret = -1;
+                    Console.Error.WriteLine("Failed to generate key pair B.");
+                }
+            }
+            if (ret == 0)
+            {
+                Console.WriteLine("ML-KEM Key Generation test passed.");
+            }
+
+            /* Encode */
+            if (ret == 0)
+            {
+                Console.WriteLine("Testing ML-KEM Key Encode...");
+                ret = wolfcrypt.MlKemEncodePublicKey(keyA, out pubA);
+                if (ret != 0)
+                {
+                    Console.Error.WriteLine("Failed to encode public key of A.");
+                }
+            }
+            if (ret == 0)
+            {
+                ret = wolfcrypt.MlKemEncodePrivateKey(keyA, out privA);
+                if (ret != 0)
+                {
+                    Console.Error.WriteLine("Failed to encode private key of A.");
+                }
+            }
+            if (ret == 0)
+            {
+                Console.WriteLine("ML-KEM Key Encode test passed.");
+            }
+
+            /* Encapsulate */
+            if (ret == 0)
+            {
+                Console.WriteLine("Testing ML-KEM Encapsulation...");
+                ret = wolfcrypt.MlKemEncapsulate(keyA, out cipherText, out sharedSecretA);
+                if (ret != 0)
+                {
+                    Console.Error.WriteLine("Failed to encapsulate.");
+                }
+            }
+            if (ret == 0)
+            {
+                Console.WriteLine("ML-KEM Encapsulation test passed.");
+            }
+            
+            /* Decode */
+            if (ret == 0)
+            {
+                Console.WriteLine("Testing ML-KEM Decode...");
+                ret = wolfcrypt.MlKemDecodePrivateKey(keyB, privA);
+                if (ret != 0)
+                {
+                    Console.Error.WriteLine("Failed to decode private key of A.");
+                }
+            }
+            if (ret == 0)
+            {
+                ret = wolfcrypt.MlKemDecodePublicKey(keyB, pubA);
+                if (ret != 0)
+                {
+                    Console.Error.WriteLine("Failed to decode public key of B.");
+                }
+            }
+            if (ret == 0)
+            {
+                Console.WriteLine("ML-KEM Decode test passed.");
+            }
+
+            /* Decapsulate */
+            if (ret == 0)
+            {
+                Console.WriteLine("Testing ML-KEM Decapsulation...");
+                ret = wolfcrypt.MlKemDecapsulate(keyB, cipherText, out sharedSecretB);
+                if (ret != 0)
+                {
+                    Console.Error.WriteLine("Failed to decapsulate.");
+                }
+            }
+            if (ret == 0)
+            {
+                Console.WriteLine("ML-KEM Decapsulation test passed.");
+            }
+
+            /* DEBUG CODE FOR REVIEW */
+            ret = BAD_FUNC_ARG;
+
+            /* Check */
+            if (ret == 0)
+            {
+                Console.WriteLine("Comparing Shared Secrets...");
+                if (!wolfcrypt.ByteArrayVerify(sharedSecretA, sharedSecretB))
+                {
+                    Console.Error.WriteLine("Shared secrets do not match.");
+                }
+                else
+                {
+                    Console.WriteLine("ML-KEM shared secret match.");
+                }
+            }
+        }
+        catch
+        {
+            // todo:
+            // ........ https://github.com/wolfSSL/wolfssl/pull/9040#discussion_r2386193630
+            // ........ https://github.com/wolfSSL/wolfssl/pull/9040#discussion_r2386194020
+        }
+        finally
+        {
+            /* Cleanup */
+            if (keyA != IntPtr.Zero)
+            {
+                ret = wolfcrypt.MlKemFreeKey(keyA);
+                if (ret != 0)
+                {
+                    Console.Error.WriteLine("Failed to free MlKem key A. Error code: " + ret);
+                }
+            }
+            if (keyB != IntPtr.Zero)
+            {
+                ret = wolfcrypt.MlKemFreeKey(keyB);
+                if (ret != 0)
+                {
+                    Console.Error.WriteLine("Failed to free MlKem key B. Error code: " + ret);
+                }
+            }
+        }
+    } /* END mlkem_test */
+
+    private static void mldsa_test(wolfcrypt.MlDsaTypes type)
+    {
+        int ret = 0;
+        IntPtr key = IntPtr.Zero;
+        IntPtr heap = IntPtr.Zero;
+        int devId = wolfcrypt.INVALID_DEVID;
+        byte[] privateKey;
+        byte[] publicKey;
+        byte[] message = Encoding.UTF8.GetBytes("This is some data to sign with ML-DSA");
+        byte[] signature;
+
+        try
+        {
+            Console.WriteLine("\nStarting " + type + " shared secret test ...");
+
+            /* Generate Key */
+            Console.WriteLine("Testing ML-DSA Key Generation...");
+            key = wolfcrypt.DilithiumMakeKey(heap, devId, type);
+            if (key == IntPtr.Zero)
+            {
+                ret = -1;
+                Console.Error.WriteLine("DilithiumMakeKey failed");
+            }
+            if (ret == 0)
+            {
+                Console.WriteLine("ML-DSA Key Generation test passed.");
+            }
+
+            /* Export */
+            if (ret == 0)
+            {
+                Console.WriteLine("Testing ML-DSA Key Export...");
+                ret = DilithiumExportPrivateKey(key, out privateKey);
+                if (ret != 0)
+                {
+                    Console.Error.WriteLine("DilithiumExportPrivateKey failed");
+                }
+            }
+            if (ret == 0)
+            {
+                ret = DilithiumExportPublicKey(key, out publicKey);
+                if (ret != 0)
+                {
+                    Console.Error.WriteLine("DilithiumExportPublicKey failed");
+                }
+            }
+            if (ret == 0)
+            {
+                Console.WriteLine("ML-DSA Key Export test passed.");
+            }
+
+            /* Import */
+            if (ret == 0)
+            {
+                Console.WriteLine("Testing ML-DSA Key Import...");
+                ret = DilithiumImportPrivateKey(privateKey, key);
+                if (ret != 0)
+                {
+                    Console.Error.WriteLine("DilithiumImportPrivateKey failed");
+                }
+            }
+            if (ret == 0)
+            {
+                ret = DilithiumImportPublicKey(publicKey, key);
+                if (ret != 0)
+                {
+                    Console.Error.WriteLine("DilithiumImportPrivateKey failed");
+                }
+            }
+            if (ret == 0)
+            {
+                Console.WriteLine("ML-DSA Key Import test passed.");
+            }
+
+            /* Sign */
+            if (ret == 0)
+            {
+                Console.WriteLine("Testing ML-DSA Signature Creation...");
+                ret = wolfcrypt.DilithiumSignMsg(key, message, out signature);
+                if (ret != 0)
+                {
+                    Console.Error.WriteLine("DilithiumSign failed");
+                }
+            }
+            if (ret == 0)
+            {
+                Console.WriteLine($"ML-DSA Signature Creation test passed. Signature Length: {signature.Length}");
+            }
+
+            /* Verify */
+            if (ret == 0)
+            {
+                Console.WriteLine("Testing ML-DSA Signature Verification...");
+                ret = wolfcrypt.DilithiumVerifyMsg(key, message, signature);
+                if (ret != 0)
+                {
+                    Console.Error.WriteLine("DilithiumVerify failed");
+                }
+            }
+            if (ret == 0)
+            {
+                Console.WriteLine("ML-DSA Signature Verification test passed.");
+            }
+        }
+        catch
+        {
+            // todo:
+            // ........ https://github.com/wolfSSL/wolfssl/pull/9040#discussion_r2386193630
+            // ........ https://github.com/wolfSSL/wolfssl/pull/9040#discussion_r2386194020
+        }
+        finally
+        {
+            if (key != IntPtr.Zero)
+            {
+                ret = wolfcrypt.DilithiumFreeKey(key);
+                if (ret != 0)
+                {
+                    Console.Error.WriteLine("Failed to free Dilithium key. Error code: " + ret);
+                }
+            }
+        }
+
+    } /* END mldsa_test */
+
     private static void aes_gcm_test()
     {
         IntPtr aes = IntPtr.Zero;
@@ -903,6 +1186,18 @@ public class wolfCrypt_Test_CSharp
             Console.WriteLine("\nStarting curve25519 test");
 
             curve25519_test(); /* curve25519 shared secret test */
+
+            Console.WriteLine("\nStarting ML-KEM test");
+
+            mlkem_test(wolfcrypt.MlKemTypes.ML_KEM_512); /* ML-KEM test */
+            mlkem_test(wolfcrypt.MlKemTypes.ML_KEM_768); /* ML-KEM test */
+            mlkem_test(wolfcrypt.MlKemTypes.ML_KEM_1024); /* ML-KEM test */
+
+            Console.WriteLine("\nStarting ML-DSA test");
+
+            mldsa_test(wolfcrypt.MlDsaTypes.ML_DSA_44); /* ML-DSA test */
+            mldsa_test(wolfcrypt.MlDsaTypes.ML_DSA_65); /* ML-DSA test */
+            mldsa_test(wolfcrypt.MlDsaTypes.ML_DSA_87); /* ML-DSA test */
 
             Console.WriteLine("\nStarting AES-GCM test");
 
--- a/wrapper/CSharp/wolfSSL_CSharp/wolfCrypt.cs
+++ b/wrapper/CSharp/wolfSSL_CSharp/wolfCrypt.cs
@@ -428,6 +428,147 @@ namespace wolfSSL.CSharp
         private extern static int wc_curve25519_export_private_raw(IntPtr key, IntPtr outPrivKey, IntPtr outPubKey);
 #endif
 
+        /********************************
+         * ML-KEM
+         */
+#if WindowsCE
+        [DllImport(wolfssl_dll)]
+        private static extern int wc_MlKemKey_CipherTextSize(IntPtr key, ref uint len);
+        [DllImport(wolfssl_dll)]
+        private static extern int wc_MlKemKey_SharedSecretSize(IntPtr key, ref uint len);
+        [DllImport(wolfssl_dll)]
+        private static extern int wc_MlKemKey_PrivateKeySize(IntPtr key, ref uint len);
+        [DllImport(wolfssl_dll)]
+        private static extern int wc_MlKemKey_PublicKeySize(IntPtr key, ref uint len);
+        [DllImport(wolfssl_dll)]
+        private static extern IntPtr wc_MlKemKey_New(int type, IntPtr heap, int devId);
+        [DllImport(wolfssl_dll)]
+        private static extern int wc_MlKemKey_Delete(IntPtr key, ref IntPtr key_p);
+        [DllImport(wolfssl_dll)]
+        private static extern int wc_MlKemKey_Init(IntPtr key, int type, IntPtr heap, int devId);
+        [DllImport(wolfssl_dll)]
+        private static extern void wc_MlKemKey_Free(IntPtr key);
+        [DllImport(wolfssl_dll)]
+        private static extern int wc_MlKemKey_MakeKey(IntPtr key, IntPtr rng);
+        [DllImport(wolfssl_dll)]
+        private static extern int wc_MlKemKey_MakeKeyWithRandom(IntPtr key, byte[] rand, int len);
+        [DllImport(wolfssl_dll)]
+        private static extern int wc_MlKemKey_EncodePublicKey(IntPtr key, byte[] output, uint len);
+        [DllImport(wolfssl_dll)]
+        private static extern int wc_MlKemKey_DecodePublicKey(IntPtr key, byte[] input, uint len);
+        [DllImport(wolfssl_dll)]
+        private static extern int wc_MlKemKey_Encapsulate(IntPtr key, byte[] ct, byte[] ss, IntPtr rng);
+        [DllImport(wolfssl_dll)]
+        private static extern int wc_MlKemKey_EncapsulateWithRandom(IntPtr key, byte[] ct, byte[] ss, byte[] rand, int len);
+        [DllImport(wolfssl_dll)]
+        private static extern int wc_MlKemKey_Decapsulate(IntPtr key, byte[] ss, byte[] ct, uint len);
+        [DllImport(wolfssl_dll)]
+        private static extern int wc_MlKemKey_EncodePrivateKey(IntPtr key, byte[] output, uint len);
+        [DllImport(wolfssl_dll)]
+        private static extern int wc_MlKemKey_DecodePrivateKey(IntPtr key, byte[] input, uint len);
+#else
+        [DllImport(wolfssl_dll, CallingConvention = CallingConvention.Cdecl)]
+        private static extern int wc_MlKemKey_CipherTextSize(IntPtr key, ref uint len);
+        [DllImport(wolfssl_dll, CallingConvention = CallingConvention.Cdecl)]
+        private static extern int wc_MlKemKey_SharedSecretSize(IntPtr key, ref uint len);
+        [DllImport(wolfssl_dll, CallingConvention = CallingConvention.Cdecl)]
+        private static extern int wc_MlKemKey_PrivateKeySize(IntPtr key, ref uint len);
+        [DllImport(wolfssl_dll, CallingConvention = CallingConvention.Cdecl)]
+        private static extern int wc_MlKemKey_PublicKeySize(IntPtr key, ref uint len);
+        [DllImport(wolfssl_dll, CallingConvention = CallingConvention.Cdecl)]
+        private static extern IntPtr wc_MlKemKey_New(int type, IntPtr heap, int devId);
+        [DllImport(wolfssl_dll, CallingConvention = CallingConvention.Cdecl)]
+        private static extern int wc_MlKemKey_Delete(IntPtr key, ref IntPtr key_p);
+        [DllImport(wolfssl_dll, CallingConvention = CallingConvention.Cdecl)]
+        private static extern int wc_MlKemKey_Init(IntPtr key, int type, IntPtr heap, int devId);
+        [DllImport(wolfssl_dll, CallingConvention = CallingConvention.Cdecl)]
+        private static extern void wc_MlKemKey_Free(IntPtr key);
+        [DllImport(wolfssl_dll, CallingConvention = CallingConvention.Cdecl)]
+        private static extern int wc_MlKemKey_MakeKey(IntPtr key, IntPtr rng);
+        [DllImport(wolfssl_dll, CallingConvention = CallingConvention.Cdecl)]
+        private static extern int wc_MlKemKey_MakeKeyWithRandom(IntPtr key, byte[] rand, int len);
+        [DllImport(wolfssl_dll, CallingConvention = CallingConvention.Cdecl)]
+        private static extern int wc_MlKemKey_EncodePublicKey(IntPtr key, byte[] output, uint len);
+        [DllImport(wolfssl_dll, CallingConvention = CallingConvention.Cdecl)]
+        private static extern int wc_MlKemKey_DecodePublicKey(IntPtr key, byte[] input, uint len);
+        [DllImport(wolfssl_dll, CallingConvention = CallingConvention.Cdecl)]
+        private static extern int wc_MlKemKey_Encapsulate(IntPtr key, byte[] ct, byte[] ss, IntPtr rng);
+        [DllImport(wolfssl_dll, CallingConvention = CallingConvention.Cdecl)]
+        private static extern int wc_MlKemKey_EncapsulateWithRandom(IntPtr key, byte[] ct, byte[] ss, byte[] rand, int len);
+        [DllImport(wolfssl_dll, CallingConvention = CallingConvention.Cdecl)]
+        private static extern int wc_MlKemKey_Decapsulate(IntPtr key, byte[] ss, byte[] ct, uint len);
+        [DllImport(wolfssl_dll, CallingConvention = CallingConvention.Cdecl)]
+        private static extern int wc_MlKemKey_EncodePrivateKey(IntPtr key, byte[] output, uint len);
+        [DllImport(wolfssl_dll, CallingConvention = CallingConvention.Cdecl)]
+        private static extern int wc_MlKemKey_DecodePrivateKey(IntPtr key, byte[] input, uint len);
+#endif
+
+        /********************************
+         * ML-DSA
+         */
+#if WindowsCE
+        [DllImport(wolfssl_dll)]
+        private static extern IntPtr wc_dilithium_new(IntPtr heap, int devId);
+        [DllImport(wolfssl_dll)]
+        private static extern int wc_dilithium_delete(IntPtr key, ref IntPtr key_p);
+        [DllImport(wolfssl_dll)]
+        private static extern int wc_dilithium_init_ex(IntPtr key, IntPtr heap, int devId);
+        [DllImport(wolfssl_dll)]
+        private static extern int wc_dilithium_set_level(IntPtr key, byte level);
+        [DllImport(wolfssl_dll)]
+        private static extern void wc_dilithium_free(IntPtr key);
+        [DllImport(wolfssl_dll)]
+        private static extern int wc_dilithium_make_key(IntPtr key, IntPtr rng);
+        [DllImport(wolfssl_dll)]
+        private static extern int wc_dilithium_export_private(IntPtr key, byte[] output, ref uint outLen);
+        [DllImport(wolfssl_dll)]
+        private static extern int wc_dilithium_import_private(byte[] priv, uint privSz, IntPtr key);
+        [DllImport(wolfssl_dll)]
+        private static extern int wc_dilithium_export_public(IntPtr key, byte[] output, ref uint outLen);
+        [DllImport(wolfssl_dll)]
+        private static extern int wc_dilithium_import_public(byte[] input, uint inputLen, IntPtr key);
+        [DllImport(wolfssl_dll)]
+        private static extern int wc_dilithium_sign_msg(byte[] msg, uint msgLen, byte[] sig, ref uint sigLen, IntPtr key, IntPtr rng);
+        [DllImport(wolfssl_dll)]
+        private static extern int wc_dilithium_verify_msg(byte[] sig, uint sigLen, byte[] msg, uint msgLen, ref int res, IntPtr key);
+        [DllImport(wolfssl_dll)]
+        private static extern int wc_MlDsaKey_GetPrivLen(IntPtr key, ref int len);
+        [DllImport(wolfssl_dll)]
+        private static extern int wc_MlDsaKey_GetPubLen(IntPtr key, ref int len);
+        [DllImport(wolfssl_dll)]
+        private static extern int wc_MlDsaKey_GetSigLen(IntPtr key, ref int len);
+#else
+        [DllImport(wolfssl_dll, CallingConvention = CallingConvention.Cdecl)]
+        private static extern IntPtr wc_dilithium_new(IntPtr heap, int devId);
+        [DllImport(wolfssl_dll, CallingConvention = CallingConvention.Cdecl)]
+        private static extern int wc_dilithium_delete(IntPtr key, ref IntPtr key_p);
+        [DllImport(wolfssl_dll, CallingConvention = CallingConvention.Cdecl)]
+        private static extern int wc_dilithium_init_ex(IntPtr key, IntPtr heap, int devId);
+        [DllImport(wolfssl_dll, CallingConvention = CallingConvention.Cdecl)]
+        private static extern int wc_dilithium_set_level(IntPtr key, byte level);
+        [DllImport(wolfssl_dll, CallingConvention = CallingConvention.Cdecl)]
+        private static extern void wc_dilithium_free(IntPtr key);
+        [DllImport(wolfssl_dll, CallingConvention = CallingConvention.Cdecl)]
+        private static extern int wc_dilithium_make_key(IntPtr key, IntPtr rng);
+        [DllImport(wolfssl_dll, CallingConvention = CallingConvention.Cdecl)]
+        private static extern int wc_dilithium_export_private(IntPtr key, byte[] output, ref uint outLen);
+        [DllImport(wolfssl_dll, CallingConvention = CallingConvention.Cdecl)]
+        private static extern int wc_dilithium_import_private(byte[] priv, uint privSz, IntPtr key);
+        [DllImport(wolfssl_dll, CallingConvention = CallingConvention.Cdecl)]
+        private static extern int wc_dilithium_export_public(IntPtr key, byte[] output, ref uint outLen);
+        [DllImport(wolfssl_dll, CallingConvention = CallingConvention.Cdecl)]
+        private static extern int wc_dilithium_import_public(byte[] input, uint inputLen, IntPtr key);
+        [DllImport(wolfssl_dll, CallingConvention = CallingConvention.Cdecl)]
+        private static extern int wc_dilithium_sign_msg(byte[] msg, uint msgLen, byte[] sig, ref uint sigLen, IntPtr key, IntPtr rng);
+        [DllImport(wolfssl_dll, CallingConvention = CallingConvention.Cdecl)]
+        private static extern int wc_dilithium_verify_msg(byte[] sig, uint sigLen, byte[] msg, uint msgLen, ref int res, IntPtr key);
+        [DllImport(wolfssl_dll, CallingConvention = CallingConvention.Cdecl)]
+        private static extern int wc_MlDsaKey_GetPrivLen(IntPtr key, ref int len);
+        [DllImport(wolfssl_dll, CallingConvention = CallingConvention.Cdecl)]
+        private static extern int wc_MlDsaKey_GetPubLen(IntPtr key, ref int len);
+        [DllImport(wolfssl_dll, CallingConvention = CallingConvention.Cdecl)]
+        private static extern int wc_MlDsaKey_GetSigLen(IntPtr key, ref int len);
+#endif
 
         /********************************
          * AES-GCM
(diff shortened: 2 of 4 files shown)


tamasan238:
Add _new/_delete API for ML-KEM/ML-DSA
https://github.com/wolfSSL/wolfssl/pull/9039

**Use case**: Add C# examples using ML-KEM/ML-DSA
https://github.com/wolfSSL/wolfssl-examples/pull/514

JacobBarthelmeh:
Retest this please Jenkins:

```
Found unhandled org.jenkinsci.plugins.workflow.support.steps.AgentOfflineException 
```

Codespell warning 

```
Error: ./wrapper/CSharp/README.md:106: Suppport ==> Support
Codespell found one or more problems
```

JacobBarthelmeh (commented):
@tamasan238 thanks! Looks good for the most part. Please double check the flagged WINCE build comment.

@gojimmypi could you take a look at this PR having C# dev fresh on your mind?

gojimmypi:
Hi @tamasan238  - 

Looks like some nice PQ updates for wolfSSL C#. Thank you!

I'm seeing some runtime errors:

```
Starting Cryptographic Tests...

Starting RNG test
RNG Test Passed

Starting ECC tests

[ ... snip other tests ... ]

Comparing Shared Secrets...
Curve25519 shared secret match.

Starting ML-KEM test

Starting ML_KEM_512 shared secret test ...
Testing ML-KEM Key Generation...
Generate Key Pair A...
MlKem key creation exception: System.EntryPointNotFoundException: Unable to find an entry point named 'wc_MlKemKey_New' in DLL 'wolfssl.dll'.
   at wolfSSL.CSharp.wolfcrypt.wc_MlKemKey_New(Int32 type, IntPtr heap, Int32 devId)
   at wolfSSL.CSharp.wolfcrypt.MlKemMakeKey(MlKemTypes type, IntPtr heap, Int32 devId) in C:\PR9040\wolfssl\wrapper\CSharp\wolfSSL_CSharp\wolfCrypt.cs:line 2876
An error occurred: Failed to generate key pair A.

C:\PR9040\wolfssl\wrapper\CSharp\Debug\x64\wolfCrypt-Test.exe (process 26580) exited with code -1 (0xffffffff).
To automatically close the console when debugging stops, enable Tools->Options->Debugging->Automatically close the console when debugging stops.
Press any key to close this window . . .
```


I followed the instructions in the README update:

```
## PQC Suppport

To enable ML-KEM / ML-DSA API, add following options for `wolfssl` and `wolfCrypt-Test` projects:

HAVE_MLKEM
HAVE_MLDSA
```
 <br />

As seen here:

image

image
 <br />
 <br />

It appears `wc_MlKemKey_New` is not in the wolfSSL codebase? Perhaps you meant `wc_MlKemKey_Init` or `wc_MlKemKey_MakeKey` ?

image

<br />

## Unrelated Warnings

There are some warnings that seem to have snuck in over time, unrelated to this PR:

```
1>C:\PR9040\wolfssl\wolfssl\error-ssl.h(245,1): warning C5287: operands are different enum types 'wolfCrypt_ErrorCodes' and 'wolfSSL_ErrorCodes'; use an explicit cast to silence this warning
1>(compiling source file '../../src/crl.c')
1>dtls13.c
1>C:\PR9040\wolfssl\wolfssl\error-ssl.h(245,1): warning C5287: operands are different enum types 'wolfCrypt_ErrorCodes' and 'wolfSSL_ErrorCodes'; use an explicit cast to silence this warning
1>(compiling source file '../../src/dtls13.c')
1>dtls.c
1>C:\PR9040\wolfssl\wolfssl\error-ssl.h(245,1): warning C5287: operands are different enum types 'wolfCrypt_ErrorCodes' and 'wolfSSL_ErrorCodes'; use an explicit cast to silence this warning
1>(compiling source file '../../src/dtls.c')
1>internal.c
1>C:\PR9040\wolfssl\wolfssl\error-ssl.h(245,1): warning C5287: operands are different enum types 'wolfCrypt_ErrorCodes' and 'wolfSSL_ErrorCodes'; use an explicit cast to silence this warning
1>(compiling source file '../../src/internal.c')
1>wolfio.c
1>C:\PR9040\wolfssl\wolfssl\error-ssl.h(245,1): warning C5287: operands are different enum types 'wolfCrypt_ErrorCodes' and 'wolfSSL_ErrorCodes'; use an explicit cast to silence this warning
1>(compiling source file '../../src/wolfio.c')
1>keys.c
1>C:\PR9040\wolfssl\wolfssl\error-ssl.h(245,1): warning C5287: operands are different enum types 'wolfCrypt_ErrorCodes' and 'wolfSSL_ErrorCodes'; use an explicit cast to silence this warning
1>(compiling source file '../../src/keys.c')
1>ocsp.c
1>ssl.c
1>C:\PR9040\wolfssl\wolfssl\error-ssl.h(245,1): warning C5287: operands are different enum types 'wolfCrypt_ErrorCodes' and 'wolfSSL_ErrorCodes'; use an explicit cast to silence this warning
1>(compiling source file '../../src/ssl.c')
1>tls.c
1>C:\PR9040\wolfssl\wolfssl\error-ssl.h(245,1): warning C5287: operands are different enum types 'wolfCrypt_ErrorCodes' and 'wolfSSL_ErrorCodes'; use an explicit cast to silence this warning
1>(compiling source file '../../src/tls.c')
1>tls13.c
1>C:\PR9040\wolfssl\wolfssl\error-ssl.h(245,1): warning C5287: operands are different enum types 'wolfCrypt_ErrorCodes' and 'wolfSSL_ErrorCodes'; use an explicit cast to silence this warning
1>(compiling source file '../../src/tls13.c')
1>aes.c
1>arc4.c
1>asn.c
1>C:\PR9040\wolfssl\wolfssl\error-ssl.h(245,1): warning C5287: operands are different enum types 'wolfCrypt_ErrorCodes' and 'wolfSSL_ErrorCodes'; use an explicit cast to silence this warning
1>(compiling source file '../../wolfcrypt/src/asn.c')
```

I'm addressing those warnings in https://github.com/wolfSSL/wolfssl/pull/8946

gojimmypi:
Also, I checked with anhu & confirmed additional macros need to be defined.

For instance, check out [this section](https://github.com/wolfSSL/wolfssl/blob/46347173b2cdb6579aa58fb6682d30e5452e6c9d/IDE/Espressif/ESP-IDF/examples/wolfssl_client/components/wolfssl/include/user_settings.h#L215) of the `user_settings.h` that I used for the ESP32. The `ESP_WOLFSSL_ENABLE_MLKEM` is a Kconfig setting, that ends up in a header as `CONFIG_ESP_WOLFSSL_ENABLE_MLKEM`:

```
#ifdef CONFIG_ESP_WOLFSSL_ENABLE_MLKEM
    /* Kyber typically needs a minimum 10K stack */
    #define WOLFSSL_HAVE_MLKEM
    #define WOLFSSL_WC_MLKEM
    #define WOLFSSL_SHAKE128
    #define WOLFSSL_SHAKE256

    /* Old code points to keep compatibility with Kyber Round 3. */
    /*   ./configure --enable-kyber=all --enable-experimental    */
    #if defined(CONFIG_WOLFSSL_ENABLE_KYBER)
        #define WOLFSSL_MLKEM_KYBER
        #define WOLFSSL_EXPERIMENTAL_SETTINGS
    #endif

    #if defined(CONFIG_IDF_TARGET_ESP8266)
        /* With limited RAM, we'll disable some of the Kyber sizes: */
        #define WOLFSSL_NO_KYBER1024
        #define WOLFSSL_NO_KYBER768
        #define WOLFSSL_NO_ML_KEM_1024
        #define WOLFSSL_NO_ML_KEM_768
        #define NO_SESSION_CACHE
    #else
        /* Only needed for older wolfssl versions, see mlkem.h */
        #define WOLFSSL_KYBER1024
        /* optional alternative sizes:    */
        /* #define WOLFSSL_KYBER768       */
        /* #define WOLFSSL_KYBER512       */
        /* -- or disable a specific one:  */
        /* #define WOLFSSL_NO_ML_KEM_1024 */
        /* #define WOLFSSL_NO_ML_KEM_768  */
        /* #define WOLFSSL_NO_ML_KEM_512  */
    #endif
#endif
```

tamasan238:
Thank you for your review!
I'll correct each item and contact again.

tamasan238:
1️⃣
Fixed below.

**1. typo**

- s/Suppport/Support/
- s/encalsulate/encapsulate/

**2. WinCE Error**

- type, key, ... -> key, type, ...

**3. configuration options**

added some options for wolfSSL(wolfCrypt).
// I think I wrote it before, but it seems that I blew it away when rebasing git commit. I'm sorry.

2️⃣
These APIs are currently being added in a separate PR. (https://github.com/wolfSSL/wolfssl/pull/9039)

- wolfCrypt/wc_mlkem.c: 
    - wc_MlKemKey_New()
    - wc_MlKemKey_Delete()
- wolfCrypt/wc_dilithium.c: 
    - wc_dilithium_new()
    - wc_dilithium_delete()

I would appreciate it if you could check it after referring to the above.

gojimmypi:
Hi @tamasan238 - looks like you are making good progress on the C# library and examples!

I do have a question: why is the PQ code only conditionally compiled in? The current conditional code is primarily platform-based, no? (e.g. decorators specific to Windows CE). Do you think it would be better if the PQ code was always on, always available, with no need for manually defined macros?

tamasan238:
Thank you for your review!

Yes, I think it is better to be available without any macros.
But in my understanding, some errors may occur when a user builds wolfSSL without PQC macros.

If some PQC APIs are unavailable in wolfSSL, the C# wrapper test program(`wolfCrypt-Test.cs`) will output an error.
So, I think that at least `wolfCrypt-Test.cs` requires `ifdef` (`if` in C#).
In order to avoid confusing users, I thought it would be better to apply the same style in `wolfCrypt.cs`.

However, I believe it is also important to keep the code base and build procedures as simple as possible.
I would like to hear your perspective.

gojimmypi:
Hi @tamasan238 

Yes, I agree that building the native C wolfSSL DLL would need attention. As the PQ code there is gated with macros defined in the `user_settings.h`, what do you think of updating that to always have the PQ code enable?

You could leave a brief comment around the C# declarations indicating which macros are assumed to have been enabled in the external wolfSSL DLL.

tamasan238:
Oh, I see! That is the best plan.
I'll do that. Thank you :)

tamasan238:
I applied them. Is this style correct?

// Build test will fail now because PR https://github.com/wolfSSL/wolfssl/pull/9039 has not yet been merged.

gojimmypi:
Hi @tamasan238 

I fetched your latest code from this pr2 branch, and applied the [patch from pr1](https://patch-diff.githubusercontent.com/raw/wolfSSL/wolfssl/pull/9039.patch).

I see there's a section added in the [wrapper/CSharp/user_settings.h](https://github.com/tamasan238/wolfssl/blob/for-pr-2/wrapper/CSharp/user_settings.h):

```c
/* Enable ML-KEM, ML-DSA */
#define HAVE_MLKEM
#define WOLFSSL_WC_MLKEM
#define WOLFSSL_HAVE_MLKEM
#define WOLFSSL_DTLS_CH_FRAG
#define HAVE_DILITHIUM
#define WOLFSSL_WC_DILITHIUM
#define WOLFSSL_SHAKE128
#define WOLFSSL_SHAKE256
```
This only _enables_ ML-KEM/ML-DSA support, eh?

The reason I ask, is the out-of-the-box experience is such that my C# client in this PR will not connect to my unmodified listener.

I see this error for the C#  `wolfSSL-TLS-Client`  example:

```text
Windows - ca-cert.pem
Windows - dh2048.pem
Calling ctx Init from wolfSSL
Finished init of ctx .... now load in CA
SNI IS OFF
Ciphers : TLS13-AES128-GCM-SHA256:TLS13-AES256-GCM-SHA384:DHE-RSA-AES128-SHA256:DHE-RSA-AES256-SHA256:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES128-SHA256:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES256-SHA384:ECDHE-ECDSA-AES256-SHA384
Connected TCP
Connection made wolfSSL_connect
SSL version is TLSv1.3
SSL cipher suite is TLS_AES_256_GCM_SHA384
socket connection issue System.Net.Sockets.SocketException (0x80004005): An existing connection was forcibly closed by the remote host
   at System.Net.Sockets.Socket.Send(Byte[] buffer, Int32 offset, Int32 size, SocketFlags socketFlags)
   at wolfSSL.CSharp.wolfssl.wolfSSLCbIOSend(IntPtr ssl, IntPtr buf, Int32 sz, IntPtr ctx) in C:\PR9040\wolfssl\wrapper\CSharp\wolfSSL_CSharp\wolfSSL.cs:line 943
Error in write
freeing ssl handle
freeing ctx handle

C:\PR9040\wolfssl\wrapper\CSharp\Debug\x64\wolfSSL-TLS-Client.exe (process 16152) exited with code 0 (0x0).
To automatically close the console when debugging stops, enable Tools->Options->Debugging->Automatically close the console when debugging stops.
Press any key to close this window . . .
```

Here's the message I see at the server (a stand-alone RPi with clean wolfssl):

```text
waiting... listening on port 11111
Early Data was not sent.
SSL_accept error -345, peer did not return a certificate
wolfSSL error: SSL_accept failed
Loop iteration: 25

waiting... listening on port 11111
```

My server was build like this:

```
 ./autogen.sh
 ./configure --enable-all
```

and I'm running it like this ` ./server_loop.sh`:

```
#!/bin/bash

count=1
THIS_PORT=11111

while true; do
    echo ""
    echo "waiting... listening on port $THIS_PORT"
    ./examples/server/server -v 4 -p "$THIS_PORT" -b 0.0.0.0
    ((count++))
    echo "Loop iteration: $count"
done
```

The failure occurs in `wolfSSLCbIOSend`, specifically this line 934:

```c
if (con.Send(msg, 0, msg.Length, SocketFlags.None) == 0 && sz != 0)
```

Can you please confirm what you see when running the examples? Any special instructions needed?

tamasan238:
This is strange behavior. I will check it right away.

tamasan238:
I could reproduce that issue with unmodified wolfssl v5.8.2.
image
So I think that problem is not related this PR (but it should be fixed by other PR...).

Communication between Server/Client under the C# wrapper directory should work fine.
I have attached a video. Could you please try it out?
https://youtu.be/h0vbJybUvNs
\* It needs wolfssl.com's google account

Examples of ML-KEM/ML-DSA usage can be found here.
https://github.com/wolfSSL/wolfssl-examples/pull/514

tamasan238:
jenkins retest this please

gojimmypi:
Hi @tamasan238 

I'm glad to see you were able to reproduce the connection problem that I encountered.

I don't recall seeing & resolving that issue, but it does not occur on my WIP https://github.com/wolfSSL/wolfssl/pull/8946

btw: Note your failing [haproxy Test / v3.1.0 (pull_request)](https://github.com/wolfSSL/wolfssl/actions/runs/16699148194/job/47407292960?pr=9040) is resolved with https://github.com/wolfSSL/wolfssl/pull/9061

Perhaps if you refresh from upstream to fix HAProxy, it will also resolve the C# connection error? I didn't dig into the root cause.

I'll take a closer look at the  ML-KEM/ML-DSA examples soon.

As for the code review: I see you followed the existing pattern in the `wolfCrypt-tests`. That's good. But I'm not sure I would have implemented that with all the `throw new Exception`'s. The problem here is all the subsequent tests will not even be attempted as soon as the first test fails. 

I would have implemented something differently: one that completed all the tests, showed results, then threw a single error at the end, or even just gracefully exit with a non-zero exit code. Refactoring that is well beyond the scope of the PR though, and I suspect someone had a specific objective in mind there. Perhaps you can find out why that pattern was chosen? 

In any case: thank you for your work on the C# PQ. It will be a nice addition!

Cheers

tamasan238:
Thank you for your confirmation!

> Issue where `wolfSSLCbIoSend()` fails under specific conditions

I still haven't figured out the root cause of this issue.

> Regarding `throw new Exception`

Indeed, even if a test fails, subsequent tests should still be executed.
I also think it would be better to address this in a separate PR, so can we put this PR on hold for now?
To be honest, I'm not really sure why it's implemented this way.

https://github.com/wolfSSL/wolfssl/pull/9039 should be merged soon.
After confirming this, I will rerun the GitHub Actions/Jenkins tests and get back to you.

tamasan238:
jenkins retest this please

tamasan238:
The API addition PR(https://github.com/wolfSSL/wolfssl/pull/9039) has been merged!
And by rebasing, I confirmed the following:
- The haproxy test issue now passes. (https://github.com/wolfSSL/wolfssl/pull/9061)
- Using the added API, `wolfCrypt-Test.cs` now succeeds.

Could you please re-check or/and merge this PR?

gojimmypi:
I've confirmed the ML-KEM / ML-DSA test app runs successfully in VS 2022. 

I didn't get a chance today to test client / server apps, nor a specific code review yet.

```
[ ... snip ... ]
Starting ML_KEM_1024 shared secret test ...
Testing ML-KEM Key Generation...
Generate Key Pair A...
Generate Key Pair B...
ML-KEM Key Generation test passed.
Testing ML-KEM Key Encode...
ML-KEM Key Encode test passed.
Testing ML-KEM Encapsulation...
ML-KEM Encapsulation test passed.
Testing ML-KEM Decode...
ML-KEM Decode test passed.
Testing ML-KEM Decapsulation...
ML-KEM Decapsulation test passed.
Comparing Shared Secrets...
ML-KEM shared secret match.

Starting ML-DSA test

Starting ML_DSA_44 shared secret test ...
Testing ML-DSA Key Generation...
ML-DSA Key Generation test passed.
Testing ML-DSA Key Export...
ML-DSA Key Export test passed.
Testing ML-DSA Key Import...
ML-DSA Key Import test passed.
Testing ML-DSA Signature Creation...
ML-DSA Signature Creation test passed. Signature Length: 2420
Testing ML-DSA Signature Verification...
ML-DSA Signature Verification test passed.
[ ... snip ... ]
```

wrapper/CSharp/wolfSSL_CSharp/wolfCrypt.cs
```diff
@@ -2709,6 +2850,627 @@ public static void Curve25519ExportKeyRaw(IntPtr key, out byte[] privateKey, out
...
+                    MlKemFreeKey(key);
+                    log(ERROR_LOG, "Failed to make MlKem key. Error code: " + ret);
+                    return IntPtr.Zero;
+                }
+            }
+            catch (Exception ex)
+            {
+                log(ERROR_LOG, "MlKem key creation exception: " + ex.ToString());
+                if (key != IntPtr.Zero)
+                    MlKemFreeKey(key);
```
gojimmypi: nit: be more generous with curly brackets, even for one line.

wrapper/CSharp/wolfCrypt-Test/wolfCrypt-Test.cs
```diff
@@ -648,6 +648,170 @@ private static void curve25519_test()
...
+        /* Verify */
+        Console.WriteLine("Testing ML-DSA Signature Verification...");
+        ret = wolfcrypt.DilithiumVerifyMsg(key, message, signature);
+        if (ret != 0)
+        {
+            throw new Exception("DilithiumVerify failed");
+        }
+        Console.WriteLine("ML-DSA Signature Verification test passed.");
+
+        if (key != IntPtr.Zero) wolfcrypt.DilithiumFreeKey(key);
```
gojimmypi: nit: put freekey on next line with curly brackets

gojimmypi: There's a potentially more serious logic problem here related to the excess throwing of exceptions. At the end, you call DilithiumFreeKey, but what if one of the above exceptions are triggered? Memory leak with the exception exit and no free?

You've done a nice job of following the existing code patterns. Still it is best to limit the hard exceptions and use error messages:

```
Console.Error.WriteLine("Error setting verify callback.");
```

wrapper/CSharp/wolfSSL_CSharp/wolfCrypt.cs
```diff
@@ -2709,6 +2850,627 @@ public static void Curve25519ExportKeyRaw(IntPtr key, out byte[] privateKey, out
...
+                    return ret;
+                }
+            }
+            catch (Exception e)
+            {
+                log(ERROR_LOG, "MlKem encode public key exception: " + e.ToString());
+                publicKey = null;
+                return EXCEPTION_E;
+            }
+            return 0;
```
gojimmypi: no magic numbers. Did you mean `return SUCCESS;` ?

wrapper/CSharp/wolfSSL_CSharp/wolfCrypt.cs
```diff
@@ -2709,6 +2850,627 @@ public static void Curve25519ExportKeyRaw(IntPtr key, out byte[] privateKey, out
...
+                    return ret;
+                }
+            }
+            catch (Exception e)
+            {
+                log(ERROR_LOG, "MlKem encode private key exception: " + e.ToString());
+                privateKey = null;
+                return EXCEPTION_E;
+            }
+            return 0;
```
gojimmypi: 'return SUCCESS;` ?

wrapper/CSharp/wolfSSL_CSharp/wolfCrypt.cs
```diff
@@ -2709,6 +2850,627 @@ public static void Curve25519ExportKeyRaw(IntPtr key, out byte[] privateKey, out
...
+                    log(ERROR_LOG, "Failed to decode MlKem public key. Error code: " + ret);
+                    return EXCEPTION_E;
+                }
+            }
+            catch (Exception ex)
+            {
+                log(ERROR_LOG, "MlKem decode public key exception: " + ex.ToString());
+                return EXCEPTION_E;
+            }
+            return 0;
```
gojimmypi: `return SUCCESS;` ?

wrapper/CSharp/wolfSSL_CSharp/wolfCrypt.cs
```diff
@@ -2709,6 +2850,627 @@ public static void Curve25519ExportKeyRaw(IntPtr key, out byte[] privateKey, out
...
+                return IntPtr.Zero;
+            }
+            return key;
+        }
+
+        /// <summary>
+        /// Free a MlKem key structure and release its memory
+        /// </summary>
+        /// <param name="key">Pointer to the MlKem key structure</param>
+        public static void MlKemFreeKey(IntPtr key)
```
gojimmypi: why not return the value of `wc_MlKemKey_Delete`?

wrapper/CSharp/wolfSSL_CSharp/wolfCrypt.cs
gojimmypi: Better to have only one free / return at the end, instead of repeating the code?

wrapper/CSharp/wolfSSL_CSharp/wolfSSL.cs
```diff
@@ -709,6 +713,89 @@ public void free()
         public static readonly int WOLFSSL_LOAD_FLAG_IGNORE_ZEROFILE     = 0x00000010;
         public static readonly int WOLFSSL_LOAD_VERIFY_DEFAULT_FLAGS     = WOLFSSL_LOAD_FLAG_NONE;
 
+        public enum NamedGroup
```
gojimmypi, line 717: Where did these numbers originate? 
Would it be a good idea to reference the `ssl.h` file in a comment?

wrapper/CSharp/wolfSSL_CSharp/wolfSSL.cs
```diff
@@ -1288,6 +1375,34 @@ public static int shutdown(IntPtr ssl)
...
+        /// <summary>
+        /// Creates a key share entry for the specified group on the given SSL/TLS connection.
+        /// </summary>
+        /// <param name="ssl">Pointer to the SSL structure to use.</param>
+        /// <param name="group">The key exchange group identifier to use for key share (e.g., TLS supported group ID).</param>
+        /// <returns>1 on success</returns>
+        public static int UseKeyShare(IntPtr ssl, NamedGroup group)
+        {
+            if (ssl == IntPtr.Zero)
+                return FAILURE;
```
gojimmypi, line 1389: be more generous with curly brackets

wrapper/CSharp/wolfSSL_CSharp/wolfCrypt.cs
```diff
@@ -2709,6 +2850,627 @@ public static void Curve25519ExportKeyRaw(IntPtr key, out byte[] privateKey, out
...
+            int ret = 0;
+            uint pubLen = 0;
+
+            try
+            {
+                ret = wc_MlKemKey_PublicKeySize(key, ref pubLen);
+                if (ret != 0 || pubLen == 0)
+                {
+                    log(ERROR_LOG, "Failed to get MlKem public key length. Error code: " + ret);
+                    return ret != 0 ? ret : EXCEPTION_E;
```
gojimmypi: A possible error return == 0 even when there was an error?  (pubLen == 0)

wrapper/CSharp/wolfSSL_CSharp/wolfCrypt.cs
```diff
@@ -2709,6 +2850,627 @@ public static void Curve25519ExportKeyRaw(IntPtr key, out byte[] privateKey, out
...
+        public static int MlKemEncodePrivateKey(IntPtr key, out byte[] privateKey)
+        {
+            privateKey = null;
+            int ret = 0;
+            uint privLen = 0;
+
+            try
+            {
+                ret = wc_MlKemKey_PrivateKeySize(key, ref privLen);
+                if (ret != 0 || privLen == 0)
```
gojimmypi: same as above. better to check each separately?

tamasan238:
jenkins retest this please

tamasan238:
jenkins retest this please

tamasan238:
@gojimmypi I fixed mistakes. Could you please check them?

wrapper/CSharp/wolfCrypt-Test/wolfCrypt-Test.cs
```diff
@@ -648,6 +648,201 @@ private static void curve25519_test()
...
+        {
+            Console.WriteLine("\nStarting " + type + " shared secret test ...");
+
+            /* Generate Key Pair */
+            Console.WriteLine("Testing ML-KEM Key Generation...");
+
+            Console.WriteLine("Generate Key Pair A...");
+            keyA = wolfcrypt.MlKemMakeKey(type, heap, devId);
+            if (keyA == IntPtr.Zero)
+            {
```
gojimmypi, line 670: Is there a reason why `ret = -1`  (or some failure) is not set here?

wrapper/CSharp/wolfCrypt-Test/wolfCrypt-Test.cs
```diff
@@ -648,6 +648,201 @@ private static void curve25519_test()
...
+                Console.Error.WriteLine("Failed to generate key pair A.");
+            }
+
+            Console.WriteLine("Generate Key Pair B...");
+            keyB = wolfcrypt.MlKemMakeKey(type, heap, devId);
+            if (keyB == IntPtr.Zero)
+            {
+                Console.Error.WriteLine("Failed to generate key pair B.");
```
gojimmypi: Same as keyA check, above

wrapper/CSharp/wolfCrypt-Test/wolfCrypt-Test.cs
```diff
@@ -648,6 +648,201 @@ private static void curve25519_test()
...
+            }
+
+            Console.WriteLine("ML-KEM Key Generation test passed.");
```
gojimmypi: What if either `keyA == IntPtr.Zero` or `keyB == IntPtr.Zero` ?

wrapper/CSharp/wolfCrypt-Test/wolfCrypt-Test.cs
```diff
@@ -648,6 +648,201 @@ private static void curve25519_test()
...
+            Console.WriteLine("\nStarting " + type + " shared secret test ...");
+
+            /* Generate Key Pair */
+            Console.WriteLine("Testing ML-DSA Key Generation...");
+            key = wolfcrypt.DilithiumMakeKey(heap, devId, type);
+            if (key == IntPtr.Zero)
+            {
+                Console.Error.WriteLine("DilithiumMakeKey failed");
+            }
+            Console.WriteLine("ML-DSA Key Generation test passed.");
```
gojimmypi: what if `key == IntPtr.Zero` ?

wrapper/CSharp/wolfCrypt-Test/wolfCrypt-Test.cs
```diff
@@ -648,6 +648,201 @@ private static void curve25519_test()
...
+            {
+                Console.WriteLine("ML-KEM shared secret match.");
+            }
+        }
+        finally
+        {
+            /* Cleanup */
+            if (keyA != IntPtr.Zero)
+            {
+                ret = wolfcrypt.MlKemFreeKey(keyA);
```
gojimmypi, line 793: The soft errors tracked by `ret` values (above) is good. 

At the end here, the prior value of `ret` is lost. This may be important for test monitoring.

If there are any external tests such as GitHub workflows that expect error codes upon failure, it would be lost here, no?

Although it is much better to not throw many exceptions for each failure, there's also the desire to match existing code patterns. Since the test does not return a success / fail (it should, but beyond the scope of this PR) ... do you think it would be best to put just one final `throw exception` at the end here? 

Having just one exception will make future refactoring easier.

wrapper/CSharp/wolfCrypt-Test/wolfCrypt-Test.cs
```diff
@@ -648,6 +648,201 @@ private static void curve25519_test()
...
+            if (!wolfcrypt.ByteArrayVerify(sharedSecretA, sharedSecretB))
+            {
+                Console.Error.WriteLine("Shared secrets do not match.");
+            }
+            else
+            {
+                Console.WriteLine("ML-KEM shared secret match.");
+            }
+        }
+        finally
```
gojimmypi, line 788: Would it be a good idea to have a `catch` above this, to describe the failure and decide how to proceed?

wrapper/CSharp/wolfCrypt-Test/wolfCrypt-Test.cs
```diff
@@ -648,6 +648,201 @@ private static void curve25519_test()
...
+            if (ret != 0)
+            {
+                Console.Error.WriteLine("Failed to encode public key of A.");
+            }
+            ret = wolfcrypt.MlKemEncodePrivateKey(keyA, out privA);
+            if (ret != 0)
+            {
+                Console.Error.WriteLine("Failed to encode private key of A.");
+            }
+            Console.WriteLine("ML-KEM Key Encode test passed.");
```
gojimmypi: Should this be wrapped in an `else`? Even if the test fails for `ret !=0` the `test passed` message is always printed.

(same for the pattern, below)

wrapper/CSharp/wolfCrypt-Test/wolfCrypt-Test.cs
```diff
@@ -648,6 +648,201 @@ private static void curve25519_test()
...
+
+            /* Encapsulate */
+            Console.WriteLine("Testing ML-KEM Encapsulation...");
+            ret = wolfcrypt.MlKemEncapsulate(keyA, out cipherText, out sharedSecretA);
```
gojimmypi: It is usually best to abort the entire test of an error is previously encountered. Here the prior failure is lost.

wrapper/CSharp/wolfSSL_CSharp/wolfCrypt.cs
```diff
@@ -2709,6 +2850,658 @@ public static void Curve25519ExportKeyRaw(IntPtr key, out byte[] privateKey, out
...
+        {
+            int ret = 0;
+            IntPtr heap = IntPtr.Zero;
+
+            if (key != IntPtr.Zero)
+            {
+                ret = wc_MlKemKey_Delete(key, ref heap);
+                key = IntPtr.Zero;
+            }
+            return ret;
```
gojimmypi, line 2937: If `key == IntPtr.Zero` should this return success?

wrapper/CSharp/wolfSSL_CSharp/wolfCrypt.cs
```diff
@@ -2709,6 +2850,658 @@ public static void Curve25519ExportKeyRaw(IntPtr key, out byte[] privateKey, out
...
+        public static int MlKemEncodePrivateKey(IntPtr key, out byte[] privateKey)
+        {
+            privateKey = null;
+            int ret = 0;
+            uint privLen = 0;
+
+            try
+            {
+                ret = wc_MlKemKey_PrivateKeySize(key, ref privLen);
+                if (privLen == 0)
```
gojimmypi: what if the call to `wc_MlKemKey_PrivateKeySize` failed and returned a negative value?

wrapper/CSharp/wolfSSL_CSharp/wolfCrypt.cs
```diff
@@ -2709,6 +2850,658 @@ public static void Curve25519ExportKeyRaw(IntPtr key, out byte[] privateKey, out
...
+        /// <returns>0 on success, negative value on error.</returns>
+        public static int MlKemDecodePublicKey(IntPtr key, byte[] publicKey)
+        {
+            int ret = 0;
+            uint pubLen = 0;
+
+            try
+            {
+                ret = wc_MlKemKey_PublicKeySize(key, ref pubLen);
+                if (pubLen == 0)
```
gojimmypi: same as above: what if `wc_MlKemKey_PublicKeySize` failed and returned a negative value but `pubLen != 0` ?

wrapper/CSharp/wolfSSL_CSharp/wolfCrypt.cs
```diff
@@ -2709,6 +2850,658 @@ public static void Curve25519ExportKeyRaw(IntPtr key, out byte[] privateKey, out
...
+        /// <param name="key">Pointer to the MlKem key structure</param>
+        /// <param name="ct">Ciphertext buffer</param>
+        /// <param name="ss">Output buffer for the shared secret</param>
+        /// <returns>0 on success, otherwise an error code</returns>
+        public static int MlKemDecapsulate(IntPtr key, byte[] ct, out byte[] ss)
+        {
+            int ret;
+            ss = null;
+            uint ssLen = 0;
+
```
gojimmypi, line 3175: Consider checking the parameters before proceeding.

wrapper/CSharp/wolfSSL_CSharp/wolfCrypt.cs
```diff
@@ -2709,6 +2850,658 @@ public static void Curve25519ExportKeyRaw(IntPtr key, out byte[] privateKey, out
...
+        {
+            int ret = 0;
+            IntPtr heap = IntPtr.Zero;
+
+            if (key != IntPtr.Zero)
+            {
+                ret = wc_dilithium_delete(key, ref heap);
+                key = IntPtr.Zero;
+            }
+            return ret;
```
gojimmypi, line 3308: what if `key == IntPtr.Zero` ?

wrapper/CSharp/wolfSSL_CSharp/wolfCrypt.cs
```diff
@@ -2709,6 +2850,658 @@ public static void Curve25519ExportKeyRaw(IntPtr key, out byte[] privateKey, out
...
+        /// <param name="key">Pointer to the Dilithium key structure.</param>
+        /// <param name="privateKey">Output byte array containing the private key.</param>
+        /// <returns>0 on success, negative value on error.</returns>
+        public static int DilithiumExportPrivateKey(IntPtr key, out byte[] privateKey)
+        {
+            privateKey = null;
+            int ret = 0;
+            int privLen = 0;
+            uint outLen;
+
```
gojimmypi, line 3371: Consider checking parameters before proceeding.

wrapper/CSharp/wolfCrypt-Test/wolfCrypt-Test.cs
```diff
@@ -648,6 +648,201 @@ private static void curve25519_test()
...
+            if (ret != 0)
+            {
+                Console.Error.WriteLine("Failed to encapsulate.");
+            }
+            Console.WriteLine("ML-KEM Encapsulation test passed.");
```
gojimmypi: Is this message always printed? Same question with others, below.

tamasan238:
Thank you for checking.
There are still many things that need to be fixed. I'm learning a lot from you explaining your thinking in detail.
I'll continue working on the fixes.

wrapper/CSharp/wolfSSL_CSharp/wolfCrypt.cs
tamasan238: In principle, I believe it is important to validate arguments before executing a function's functionality.
This is crucial for preventing unintended issues during function execution.
Indeed, within this function, we verify that the argument is not IntPtr.Zero before performing the Free operation.

Then, functions like `ECCFreeKey()` and `RsaFreeKey()` simply ignore `IntPtr.Zero` if passed.
If `MlKemFreeKey()`/`DilithiumFreeKey()` were to throw an error and halt processing only when used similarly, it cause inconsistent behavior across algorithms.
I thought that might be more problematic for users.
Could you please tell your thought?

wrapper/CSharp/wolfCrypt-Test/wolfCrypt-Test.cs
tamasan238: > Although it is much better to not throw many exceptions for each failure, there's also the desire to match existing code patterns. Since the test does not return a success / fail (it should, but beyond the scope of this PR) ... do you think it would be best to put just one final throw exception at the end here?

I'm sorry, but I'm currently unable to visualize the code you have in mind.
Could you show me some program code to illustrate what roles you think the `try`, `catch`, `finally`, and `exception` should each play?

tamasan238:
I've made several fixes and responded to some review comments.
Please review the changes.

Note that the system drive on my development Windows machine has failed, so I'm unable to run tests locally using Visual Studio.
As a result, there may be some basic errors present.
I plan to fix these by referencing the test results from Jenkins/GitHub Actions.
If you notice any issues, I'd appreciate it if you could let me know.

tamasan238:
Additionally, I was advised via Slack to refer to the following regarding error values, but it appears they are not referenced in `wolfCrypt.cs`.

https://github.com/wolfSSL/wolfssl/blob/master/wolfssl/wolfcrypt/error-crypt.h

Please let me know what error values you consider desirable to set, or whether you think a reference to `error-crypt.h` should be added.

```
"D:\a\wolfssl\wolfssl\wolfssl\wrapper\CSharp\wolfSSL_CSharp.sln" (default target) (1) ->
"D:\a\wolfssl\wolfssl\wolfssl\wrapper\CSharp\wolfSSL-TLS-PSK-Server\wolfSSL-TLS-PSK-Server.csproj" (default target) (6) ->
"D:\a\wolfssl\wolfssl\wolfssl\wrapper\CSharp\wolfSSL_CSharp\wolfSSL_CSharp.csproj" (default target) (10:9) ->
(CoreCompile target) -> 
 D:\a\wolfssl\wolfssl\wolfssl\wrapper\CSharp\wolfSSL_CSharp\wolfCrypt.cs(2960,24): error CS0103: The name 'BAD_FUNC_ARG' does not exist in the current context [D:\a\wolfssl\wolfssl\wolfssl\wrapper\CSharp\wolfSSL_CSharp\wolfSSL_CSharp.csproj]
 D:\a\wolfssl\wolfssl\wolfssl\wrapper\CSharp\wolfSSL_CSharp\wolfCrypt.cs(3004,24): error CS0103: The name 'BAD_FUNC_ARG' does not exist in the current context [D:\a\wolfssl\wolfssl\wolfssl\wrapper\CSharp\wolfSSL_CSharp\wolfSSL_CSharp.csproj]
 D:\a\wolfssl\wolfssl\wolfssl\wrapper\CSharp\wolfSSL_CSharp\wolfCrypt.cs(3047,24): error CS0103: The name 'BAD_FUNC_ARG' does not exist in the current context [D:\a\wolfssl\wolfssl\wolfssl\wrapper\CSharp\wolfSSL_CSharp\wolfSSL_CSharp.csproj]
 D:\a\wolfssl\wolfssl\wolfssl\wrapper\CSharp\wolfSSL_CSharp\wolfCrypt.cs(3087,24): error CS0103: The name 'BAD_FUNC_ARG' does not exist in the current context [D:\a\wolfssl\wolfssl\wolfssl\wrapper\CSharp\wolfSSL_CSharp\wolfSSL_CSharp.csproj]
 D:\a\wolfssl\wolfssl\wolfssl\wrapper\CSharp\wolfSSL_CSharp\wolfCrypt.cs(3132,24): error CS0103: The name 'BAD_FUNC_ARG' does not exist in the current context [D:\a\wolfssl\wolfssl\wolfssl\wrapper\CSharp\wolfSSL_CSharp\wolfSSL_CSharp.csproj]
 D:\a\wolfssl\wolfssl\wolfssl\wrapper\CSharp\wolfSSL_CSharp\wolfCrypt.cs(3184,24): error CS0103: The name 'BAD_FUNC_ARG' does not exist in the current context [D:\a\wolfssl\wolfssl\wolfssl\wrapper\CSharp\wolfSSL_CSharp\wolfSSL_CSharp.csproj]
 D:\a\wolfssl\wolfssl\wolfssl\wrapper\CSharp\wolfSSL_CSharp\wolfCrypt.cs(3327,24): error CS0103: The name 'BAD_FUNC_ARG' does not exist in the current context [D:\a\wolfssl\wolfssl\wolfssl\wrapper\CSharp\wolfSSL_CSharp\wolfSSL_CSharp.csproj]
 D:\a\wolfssl\wolfssl\wolfssl\wrapper\CSharp\wolfSSL_CSharp\wolfCrypt.cs(3351,24): error CS0103: The name 'BAD_FUNC_ARG' does not exist in the current context [D:\a\wolfssl\wolfssl\wolfssl\wrapper\CSharp\wolfSSL_CSharp\wolfSSL_CSharp.csproj]
 D:\a\wolfssl\wolfssl\wolfssl\wrapper\CSharp\wolfSSL_CSharp\wolfCrypt.cs(3380,24): error CS0103: The name 'BAD_FUNC_ARG' does not exist in the current context [D:\a\wolfssl\wolfssl\wolfssl\wrapper\CSharp\wolfSSL_CSharp\wolfSSL_CSharp.csproj]
 D:\a\wolfssl\wolfssl\wolfssl\wrapper\CSharp\wolfSSL_CSharp\wolfCrypt.cs(3430,24): error CS0103: The name 'BAD_FUNC_ARG' does not exist in the current context [D:\a\wolfssl\wolfssl\wolfssl\wrapper\CSharp\wolfSSL_CSharp\wolfSSL_CSharp.csproj]
 D:\a\wolfssl\wolfssl\wolfssl\wrapper\CSharp\wolfSSL_CSharp\wolfCrypt.cs(3482,24): error CS0103: The name 'BAD_FUNC_ARG' does not exist in the current context [D:\a\wolfssl\wolfssl\wolfssl\wrapper\CSharp\wolfSSL_CSharp\wolfSSL_CSharp.csproj]
 D:\a\wolfssl\wolfssl\wolfssl\wrapper\CSharp\wolfSSL_CSharp\wolfCrypt.cs(3531,24): error CS0103: The name 'BAD_FUNC_ARG' does not exist in the current context [D:\a\wolfssl\wolfssl\wolfssl\wrapper\CSharp\wolfSSL_CSharp\wolfSSL_CSharp.csproj]
 ```

**Comment by tamasan238** [comment]
Additionally, I note that the current code includes code to force the test to terminate prematurely, as suggested to me via Slack.

```
/* DEBUG CODE FOR REVIEW */
ret = BAD_FUNC_ARG;
```

File: wrapper/CSharp/wolfCrypt-Test/wolfCrypt-Test.cs (hunk 16 above)
Comment (tamasan238) [reply wrapper/CSharp/wolfCrypt-Test/wolfCrypt-Test.cs:788]: It's related to https://github.com/wolfSSL/wolfssl/pull/9040#discussion_r2462588258

File: wrapper/CSharp/wolfSSL_CSharp/wolfCrypt.cs (hunk 23 above)
Comment (tamasan238) [reply wrapper/CSharp/wolfSSL_CSharp/wolfCrypt.cs:3308]: It's related to https://github.com/wolfSSL/wolfssl/pull/9040#discussion_r2462584023

**Comment by gojimmypi** [comment]
> Additionally, I note that the current code includes code to force the test to terminate prematurely, as suggested to me via Slack.
> 
> ```
> /* DEBUG CODE FOR REVIEW */
> ret = BAD_FUNC_ARG;
> ```

See the error codes for `BAD_FUNC_ARG` and others in [wolfssl/wolfcrypt/error-crypt.h](https://github.com/wolfSSL/wolfssl/blob/master/wolfssl/wolfcrypt/error-crypt.h); it might be nice to have them all available in the C# wrapper, or you could just define the ones you use. Try to use consistent values if not importing the entire list.

You might consider something like this:

```
using System;
using System.Runtime.InteropServices;
using System.Security.Cryptography;
using System.Text;
using static wolfSSL.CSharp.WolfCryptErrorCodes; // << add this

namespace wolfSSL.CSharp
{
    public enum WolfCryptErrorCodes : int  // << add this
    {
        WC_FAILURE = -1,
                              // with all the other error codes, maybe even in a file of its own
        BAD_FUNC_ARG = -173
    }

    public class wolfcrypt
    {

        private const string wolfssl_dll = "wolfssl.dll";
```

and then either:  `return (int)BAD_FUNC_ARG;` or (probably more invasive and problematic but arguably better) update the return value types like this:

```
        public static WolfCryptErrorCodes MlKemEncodePublicKey(IntPtr key, out byte[] publicKey)
        {
```

There's also the option of something simple for now and this PR: just defined the ones you use & leave a note:

```
/* TODO: add all error codes from wolfcrypt/error-crypt.h */
public const int BAD_FUNC_ARG = -173;
```

**Comment by gojimmypi** [comment]
> Note that the system drive on my development Windows machine has failed, so I'm unable to run tests locally using Visual Studio. As a result, there may be some basic errors present.

Any luck on your Visual Studio? I'm seeing these compile-time errors:

```
Use of unassigned local variable 'privA'
Use of unassigned local variable 'pubA'
Use of unassigned local variable 'cipherText'
The name 'BAD_FUNC_ARG' does not exist in the current context
Use of unassigned local variable 'sharedSecretA'
Use of unassigned local variable 'sharedSecretB'
Use of unassigned local variable 'privateKey'
Use of unassigned local variable 'publicKey'
Use of unassigned local variable 'signature'
```

Consider:

```csharp
private static void mlkem_test(wolfcrypt.MlKemTypes type)
        byte[] pubA = { }, privA = { }, cipherText = { }, sharedSecretA = { }, sharedSecretB= { };
```

and:

```csharp
 private static void mldsa_test(wolfcrypt.MlDsaTypes type)
        byte[] privateKey = { };
        byte[] publicKey = { };
        byte[] signature = { };

```

wrapper/CSharp/wolfCrypt-Test/wolfCrypt-Test.cs
```diff
@@ -648,6 +648,289 @@ private static void curve25519_test()
...
+            {
+                ret = wolfcrypt.MlKemFreeKey(keyA);
+                if (ret != 0)
+                {
+                    Console.Error.WriteLine("Failed to free MlKem key A. Error code: " + ret);
+                }
+            }
+            if (keyB != IntPtr.Zero)
+            {
+                ret = wolfcrypt.MlKemFreeKey(keyB);
```
gojimmypi, line 801: Any prior `ret` error values are lost and overwritten here. Is this as intended?

wrapper/CSharp/wolfCrypt-Test/wolfCrypt-Test.cs
```diff
@@ -648,6 +648,289 @@ private static void curve25519_test()
...
+                if (ret != 0)
+                {
+                    Console.Error.WriteLine("Failed to free MlKem key B. Error code: " + ret);
```
gojimmypi, line 804: Should we then exit with an error code (change `private static void mlkem_test` to have a return code) and/or throw the exception here?

Remember we want the caller to programmatically know there was a failure here, not just printed to the console.

wrapper/CSharp/wolfCrypt-Test/wolfCrypt-Test.cs
```diff
@@ -648,6 +648,289 @@ private static void curve25519_test()
...
+            // ........ https://github.com/wolfSSL/wolfssl/pull/9040#discussion_r2386194020
+        }
+        finally
+        {
+            if (key != IntPtr.Zero)
+            {
+                ret = wolfcrypt.DilithiumFreeKey(key);
+                if (ret != 0)
+                {
+                    Console.Error.WriteLine("Failed to free Dilithium key. Error code: " + ret);
```
gojimmypi, line 927: Same as above: the caller needs to know there was a failure with return code and/or exception.

Return codes are preferred.

wrapper/CSharp/wolfSSL_CSharp/wolfCrypt.cs
gojimmypi: > In principle, I believe it is important to validate arguments before executing a function's functionality.

agreed.

> functions like `ECCFreeKey()` and `RsaFreeKey()` simply ignore `IntPtr.Zero` if passed.

It's more of a design-pattern and readability issue. The `ecc.c` has an [explicit check](https://github.com/wolfSSL/wolfssl/blob/5922b5def5177ad6f1632dd3b961f2336881202c/wolfcrypt/src/ecc.c#L7884):

```
    if (key == NULL) {
        return 0;
    }
```

This way, the reader (often your future self) will know the return code is intentional.

wrapper/CSharp/wolfSSL_CSharp/wolfCrypt.cs
gojimmypi: yes, see my design-pattern comment there; point being is to make it easy for the reader to know the default return code is intentional.

tamasan238:
Once preparations are complete, I will create a new PR.

PR #9051, opened 2025-07-31
Sniffer: Fix infinite recursion caused by an OOO appData packet

lealem47:
# Description

The sniffer isn't able to decode the application data if it is Out of Order and comes earlier than any `Change Cipher Spec` packets in the pcap capture. Because no data was decoded, we loop through and try to decode it again and again. With this patch, we only try and decode more data if we were able to decode the first portion otherwise we move on to decoding the rest of the capture.

Future TODO: handle parsing of OOO appData packets

Fixes zd#20288

# Testing

With the pcap provided by the customer.

```
./configure --enable-sniffer
make
./sslSniffer/sslSnifferTest/snifftest -pcap ~/Downloads/DV.TLSv12-RecordLayer.pcap -key ~/Downloads/DV_Navigator_tls.key -server 10.128.5.248 -port 9443 
```

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/src/sniffer.c
+++ b/src/sniffer.c
@@ -6496,8 +6496,8 @@ doPart:
             return WOLFSSL_FATAL_ERROR;
     }
 
-    /* do we have another msg in record ? */
-    if (sslFrame < recordEnd) {
+    /* do we have another msg in record ? did we decode the current msg ? */
+    if (sslFrame < recordEnd && decoded) {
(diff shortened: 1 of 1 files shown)


lealem47:
Jenkins retest this please

```
```

JacobBarthelmeh (approved):

PR #9054, opened 2025-08-01
Please use DEBUG_WOLFSSL not WOLFSSL_DEBUG

gojimmypi:
# Description

I've made this mistake myself on more that one occasion: `WOLFSSL_DEBUG` vs `DEBUG_WOLFSSL`.

The correct macro for wolfSSL debugging is `DEBUG_WOLFSSL`.

I've made some corrections that already exist in the codebase and added this to `settings.h` to avoid future mistakes:

```c
#ifdef WOLFSSL_DEBUG
    #error "Please use DEBUG_WOLFSSL not WOLFSSL_DEBUG"
#endif
```

As for the pattern:

```c
#ifdef DEBUG_WOLFSSL
        printf("Testing something...\n");
#endif
```

It is best to use [WOLFSSL_MSG](https://github.com/wolfSSL/wolfssl/blob/0392ee009feabcb5724b216b0d965b272fa765c7/wolfssl/wolfcrypt/logging.h#L182) from `wolfssl/wolfcrypt/logging.h`.

```c
WOLFSSL_MSG("Testing something...\n");
```

Fixes zd#

# Testing

Manual testing only

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/wolfcrypt/src/port/Espressif/esp32_mp.c
+++ b/wolfcrypt/src/port/Espressif/esp32_mp.c
@@ -2154,7 +2154,7 @@ int esp_mp_mulmod(MATH_INT_T* X, MATH_INT_T* Y, MATH_INT_T* M, MATH_INT_T* Z)
         DPORT_REG_WRITE(RSA_MULT_MODE_REG, (mph->hwWords_sz >> 4) - 1);
 #if defined(DEBUG_WOLFSSL)
         ESP_LOGV(TAG, "RSA_MULT_MODE_REG = %d", (mph->hwWords_sz >> 4) - 1);
-#endif /* WOLFSSL_DEBUG */
+#endif /* DEBUG_WOLFSSL */
 
         /* step.2 write X, M, and r_inv into memory.
          * The capacity of each memory block is 128 words.
--- a/wolfcrypt/src/port/devcrypto/devcrypto_aes.c
+++ b/wolfcrypt/src/port/devcrypto/devcrypto_aes.c
@@ -317,7 +317,7 @@ static int wc_DevCrypto_AesGcm(Aes* aes, byte* out, byte* in, word32 sz,
                       dir, (byte*)authIn, authInSz, authTag, authTagSz);
     ret = ioctl(aes->ctx.cfd, CIOCAUTHCRYPT, &crt);
     if (ret != 0) {
-        #ifdef WOLFSSL_DEBUG
+        #ifdef DEBUG_WOLFSSL
         if (authInSz > sysconf(_SC_PAGESIZE)) {
             WOLFSSL_MSG("authIn Buffer greater than System Page Size");
         }
--- a/wolfcrypt/test/test.c
+++ b/wolfcrypt/test/test.c
@@ -18663,7 +18663,7 @@ static wc_test_ret_t const_byte_ptr_test(const byte* in, word32 *outJ)
     j = *outJ; /* Found index to use in const array. */
 
     if (j == 0) {
-#ifdef WOLFSSL_DEBUG
+#ifdef DEBUG_WOLFSSL
         printf("Testing const byte ptr reference...\n");
 #endif
         /* although j is zero, in[0] does not detect the Illegal instruction */
--- a/wolfssl/test.h
+++ b/wolfssl/test.h
@@ -3124,7 +3124,7 @@ static WC_INLINE int wolfSSL_PrintStats(WOLFSSL_MEM_STATS* stats)
         return 0;
     }
 
-    /* print to stderr so is on the same pipe as WOLFSSL_DEBUG */
+    /* print to stderr so is on the same pipe as DEBUG_WOLFSSL */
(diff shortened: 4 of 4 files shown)


wolfssl/wolfcrypt/settings.h
```diff
@@ -4538,6 +4538,10 @@ extern void uITRON4_free(void *p) ;
 #error "If TLS is enabled please make sure either client or server is enabled."
 #endif
 
+#ifdef WOLFSSL_DEBUG
```
dgarske: Do we really need to throw an error? I'm okay with that macro setting DEBUG_WOLFSSL if its not set.

gojimmypi: I prefer an error over quietly auto-fixing. It is not a valid macro.

The correction will teach the user the correct macro to use.

Having an error will also ensure our own code is using the proper macro.

gojimmypi: The current error in jenkins is `Cannot contact wolf-linux-cloud-node`. 

Before I retest, I'll await word if we should quietly support either `DEBUG_WOLFSSL` or `WOLFSSL_DEBUG`.

I see the merit of both, particularly since the correct one does _not_ start with `WOLFSSL_`.

dgarske: I'm fine with just defining DEBUG_WOLFSSL, but make sure you either do a #ifndef or you #undef DEBUG_WOLFSSL to avoid redef warning. You'll need to add to the .known_macros list probably. @JacobBarthelmeh thoughts?

dgarske (changes requested):

wolfssl/wolfcrypt/settings.h
gojimmypi: Additionally, if we're ok to use `DEBUG_WOLFSSL` (the correct one) **as well as** `WOLFSSL_DEBUG` (the common misspelling) - then we should include some hacky logic like this to ensure full cross-use ability:

```x
#ifdef WOLFSSL_DEBUG
    #ifndef DEBUG_WOLFSSL
        #define DEBUG_WOLFSSL
    #endif
#endif

#ifdef DEBUG_WOLFSSL
    #ifndef WOLFSSL_DEBUG
        #define WOLFSSL_DEBUG
    #endif
#endif
```

Without this logic, we'll always have the issue of `WOLFSSL_DEBUG` vs `DEBUG_WOLFSSL` being used. (e.g. what if someone puts the incorrect `#define WOLFSSL_DEBUG` in `user_settings.h`?

I think it is really out of character for the codebase to support mistakes like this... but I do understand - as I too, have made the mistake on more than one occasion. 

Still, imho it sets a bad example. We probably don't want to continue and support things like allowing both of these:

`WOLFSSL_DEBUG_NONBLOCK` 
`DEBUG_WOLFSSL_NONBLOCK`

So if I'm overruled and we're going to allow `WOLFSSL_DEBUG` and `WOLFSSL_DEBUG` - I suggest that should be the limit.

dgarske:

wolfssl/wolfcrypt/settings.h
dgarske: Please don't start down this path of trying to introduce a better named macro. Its fine as-is.

dgarske (approved):

wolfssl/wolfcrypt/settings.h
JacobBarthelmeh: I think we should not start error'ing on it. DEBUG_WOLFSSL is a very old macro and has been set as the public debug macro for years.

JacobBarthelmeh (changes requested):
Some good changes where WOLFSSL_DEBUG was corrected to DEBUG_WOLFSSL. I don't think we should start error'ing though when encountering WOLFSSL_DEBUG.

wolfssl/wolfcrypt/settings.h
dgarske: I agree, please remove this logic.

dgarske (changes requested):

wolfssl/wolfcrypt/settings.h
gojimmypi: Done. I assume `#warning` also undesirable at compile time.

Is there a place to enter `WOLFSSL_DEBUG` as a known spelling error to be caught during Jenkins or GitHub Actions?

gojimmypi:
Refresh from upstream for https://github.com/wolfSSL/wolfssl/pull/9061 fix, squashed commits & force-pushed to retest.

This forced logic removed per code review, consider spell check instead:

```
#ifdef WOLFSSL_DEBUG
    #error "Please use DEBUG_WOLFSSL not WOLFSSL_DEBUG"
#endif
```

gojimmypi:
Jenkins retest this please

For AgentOfflineException: Unable to create live FilePath for wolf-linux-cloud-node-[n]; wolf-linux-cloud-node-[n]was marked offline: Connection was broken

PR #9065, opened 2025-08-05
Update fe_448.c from script

rlm2002:
# Description

Updating files with regenerated version from scripts repo.

changes from scripts PR [502](https://github.com/wolfSSL/scripts/pull/502). Adds additional casts to `fe448_to_bytes()`.

# Testing

### DIFF
--- a/wolfcrypt/src/fe_448.c
+++ b/wolfcrypt/src/fe_448.c
@@ -569,7 +569,7 @@ void fe448_from_bytes(sword64* r, const unsigned char* b)
  */
 void fe448_to_bytes(unsigned char* b, const sword64* a)
 {
-    sword128 t;
+    sword64 t;
     /* Mod */
     sword64 in0 = a[0];
     sword64 in1 = a[1];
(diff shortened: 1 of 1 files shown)


rlm2002:
retest this please Jenkins: PRB-fsanitize-addr-v3/PRB-generic-config-parser - Found unhandled hudson.remoting.RequestAbortedException exception

wolfcrypt/src/fe_448.c
```diff
@@ -594,13 +594,13 @@ void fe448_to_bytes(unsigned char* b, const sword64* a)
...
     in4 += o;
     in7 -= o << 56;
-    o = (in0  >> 56); in1  += o; t = o << 56; in0  -= (sword64)t;
-    o = (in1  >> 56); in2  += o; t = o << 56; in1  -= (sword64)t;
-    o = (in2  >> 56); in3  += o; t = o << 56; in2  -= (sword64)t;
-    o = (in3  >> 56); in4  += o; t = o << 56; in3  -= (sword64)t;
-    o = (in4  >> 56); in5  += o; t = o << 56; in4  -= (sword64)t;
-    o = (in5  >> 56); in6  += o; t = o << 56; in5  -= (sword64)t;
-    o = (in6  >> 56); in7  += o; t = o << 56; in6  -= (sword64)t;
+    o = (in0  >> 56); in1  += o; t = (sword128)o << 56; in0  -= (sword64)t;
```
SparkiDev: I don't know why I did this.
I can't see 't' needing to be double width.
Change 't' to be 64-bit and put the cast to 64-bit on the result of the operation.

SparkiDev (changes requested):

wolfcrypt/src/fe_448.c
rlm2002: If we change `t` to 64-bit, `t = o << 56` (assuming this is the operation you referred to) doesn't seem to need the cast since `o` would be initialized to 64-bit as well. Is this incorrect?

rlm2002: After looking at this closer. I can modify the function to remove the cast on `t`, but I don't believe the additional cast on the result of the operation is necessary as the other values are 64-bit as well.

I think these changes would also assume `o` does not contain 8 or more bits of value. Would this scenario be possible here?

SparkiDev: in0 is a 64-bit type and is shifted down 56 bits into o.
o is not more than 8 bits.

Change t to a 64-bit type and remove the casts.
See if that works.

rlm2002:
retest this please jenkins. PRB-generic-config-parser: Found unhandled org.jenkinsci.plugins.workflow.support.steps.AgentOfflineException exception:
Unable to create live FilePath for wolf-linux-cloud-node-5fr8ya; wolf-linux-cloud-node-5fr8ya was marked offline: Connection was broken

SparkiDev (approved):

douzzer (approved):

PR #8902, opened 2025-06-19
Introduce WOLFSSL_DEBUG_CERTS Certificate Debug Messages

gojimmypi:
# Description

Adds a new troubleshooting capability to view only interesting certificate-related messages:

```c
WOLFSSL_API void WOLFSSL_MSG_CERT(const char* fmt, ...)
```

Also adds `WOLFSSL_DEBUG_LINE_ENDING` to suppress LF characters on message printed for systems that supply their own line feeds during messaging, such as the Espressif `ESP_LOG`. 

Turning on `WOLFSSL_DEBUG` _always_ enables `WOLFSSL_DEBUG_CERTS`.

However `WOLFSSL_DEBUG_CERTS` can be used _without_ `WOLFSSL_DEBUG`.

## Why?

Turning on full debugging is often overly verbose. On embedded devices the delay in printing debug messages can have an adverse effect on timing-critical code, such as certificate validation during TLS connections.

## Usage
To use, add to `user_settings.h`:

```
#define WOLFSSL_DEBUG_CERTS
```

## Sample output:

```
I (15213) time lib: Successfully set time via NTP servers.
I (15222) client-tls: get target IP address
I (15262) client-tls: Host name: www.google.com
E (15262) client-tls: Create ctx success, ret = 1
I (15262) client-tls: CA cert Use SNI success
I (15263) client-tls: Loading 8408 bytes of PEM certs
I (15272) wolfssl:      Parsed new CA
I (15273) wolfssl:      -  issuer:  '/C=US/O=Google Trust Services LLC/CN=GTS Root R1'
I (15279) wolfssl:      -  subject: '/C=US/O=Google Trust Services/CN=WR2'
I (15292) wolfssl:      Parsed new CA
I (15292) wolfssl:      -  issuer:  '/C=BE/O=GlobalSign nv-sa/OU=Root CA/CN=GlobalSign Root CA'
I (15293) wolfssl:      -  subject: '/C=US/O=Google Trust Services LLC/CN=GTS Root R1'
I (15303) wolfssl:      Parsed new CA
I (15303) wolfssl:      -  issuer:  '/C=US/O=Google Trust Services LLC/CN=GTS Root R4'
I (15303) wolfssl:      -  subject: '/C=US/O=Google Trust Services/CN=WE2'
I (15312) wolfssl:      Parsed new CA
I (15312) wolfssl:      -  issuer:  '/C=BE/O=GlobalSign nv-sa/OU=Root CA/CN=GlobalSign Root CA'
[...]

PR #9069, opened 2025-08-06
Fix inconsistent function prototype parameter names for wolfcrypt

holtrop:
# Description

Fix inconsistent function prototype parameter names for wolfcrypt

# Testing

Unit/CI tests only.

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


holtrop:

wolfcrypt/src/logging.c
```diff
@@ -118,7 +118,7 @@ THREAD_LS_T void *StackSizeCheck_stackOffsetPointer = 0;
 #ifdef DEBUG_WOLFSSL
 
 /* Set these to default values initially. */
-static wolfSSL_Logging_cb log_function = NULL;
+static wolfSSL_Logging_cb LogFunction = NULL;
```
dgarske, line 121: Can you describe the MISRA rule violation for `log_function` vs `LogFunction` for a global variable name? Is it because you renamed the argument in `wolfSSL_SetLoggingCb` and then it conflicted (shadowed global)?

holtrop: Yes it was only because of the conflict between the global name and the parameter name. If it makes more sense to leave the global name, I can change both the function parameter and prototype names. I thought it should be safe to change the global name since it was static.

dgarske: I'd prefer not change those, but it really doesn't mater. It will cause a merge conflict with https://github.com/wolfSSL/wolfssl/pull/8902

douzzer (approved):

PR #9079, opened 2025-08-08
Error from GetShortInt with negative INTEGER values

holtrop:
# Description

Error from GetShortInt with negative INTEGER values.

# Testing

Unit/CI tests.

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/wolfcrypt/src/asn.c
+++ b/wolfcrypt/src/asn.c
@@ -1294,6 +1294,12 @@ static int GetASN_StoreData(const ASNItem* asn, ASNGetData* data,
             #endif
                 return ASN_PARSE_E;
             }
+            if (!zeroPadded && (input[idx] >= 0x80U)) {
+            #ifdef WOLFSSL_DEBUG_ASN_TEMPLATE
+                WOLFSSL_MSG_VSNPRINTF("Unexpected negative INTEGER value");
+            #endif
+                return ASN_EXPECT_0_E;
+            }
             /* Fill number with all of data. */
             *data->data.u16 = 0;
             for (i = 0; i < len; i++) {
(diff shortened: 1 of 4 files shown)


holtrop:

wolfcrypt/src/asn.c
```diff
@@ -1289,6 +1289,12 @@ static int GetASN_StoreData(const ASNItem* asn, ASNGetData* data,
             #endif
                 return ASN_PARSE_E;
             }
+            if (!zeroPadded && (input[idx] >= 0x80U)) {
```
SparkiDev, line 1292: Add to ASN_DATA_TYPE_WORD8 too.

holtrop: Added. Not sure why I didn't originally do that!

holtrop: So, it is this addition to the WORD8 that is causing the CI failures after I rebased last week. It seems to pass with WORD16 and WORD32.  I don't yet have a full understanding of the template mechanism yet, but do you know why it would fail for just the WORD8 case?

holtrop:
retest this please (RequestAbortedException)

wolfcrypt/src/asn.c
SparkiDev: BOOLEAN is stored as a word8.
Put the negative check in but add a check for: asn->tag == ASN_INTEGER

holtrop:

wolfcrypt/src/asn.c
holtrop: Ah, that did it!

SparkiDev (approved):

PR #9104, opened 2025-08-14
Improve original implementation on SetAsymKeyDer() and the test

kojiws:
# Description

This PR improves original (=not template) implementation on `SetAsymKeyDer()` to export longer key length than 127.
The test are also added at the same time.

# Testing

./configure --enable-wolfclu && make test
./configure --enable-wolfclu --enable-asn=original && make test

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/wolfcrypt/src/asn.c
+++ b/wolfcrypt/src/asn.c
@@ -37869,7 +37869,7 @@ int SetAsymKeyDer(const byte* privKey, word32 privKeyLen,
 {
     int ret = 0;
 #ifndef WOLFSSL_ASN_TEMPLATE
-    word32 idx = 0, seqSz, verSz, algoSz, privSz, pubSz = 0, sz;
+    word32 idx = 0, seqSz, verSz, algoSz, tmpSz, privSz, pubSz = 0, sz;
 #else
     DECL_ASNSETDATA(dataASN, privateKeyASN_Length);
     int sz = 0;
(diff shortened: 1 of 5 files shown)


wolfcrypt/src/asn.c
```diff
@@ -37916,14 +37915,14 @@ int SetAsymKeyDer(const byte* privKey, word32 privKeyLen,
...
+        idx += SetOctetString(tmpSz, output + idx);
         idx += SetOctetString(privKeyLen, output + idx);
         XMEMCPY(output + idx, privKey, privKeyLen);
         idx += privKeyLen;
         /* pubKey */
         if (pubKey) {
-            idx += SetHeader(ASN_CONTEXT_SPECIFIC | ASN_ASYMKEY_PUBKEY |
-                             1, pubKeyLen, output + idx, 0);
+            idx += SetHeader(ASN_CONTEXT_SPECIFIC | ASN_ASYMKEY_PUBKEY,
+                pubKeyLen, output + idx, 0);
```
kojiws, line 37925: The `1` isn't needed because the `ASN_ASYMKEY_PUBKEY` is itself.

kojiws:
I believe that `XFREE()` will be always called in `test_SetAsymKey_once()`.

`SetHeader()` / `SetOctetString()` / `SetAlgoID()` return word32 value, so they don't return negative value.
(However, it's better to check if `SetAlgoID()` returns 0 or not. It should be fixed in another PR.)

dgarske (approved):

PR #9116, opened 2025-08-20
wolfcrypt: test: add guard to opt-out for variable iv tests in aes-gcm

rizlik:
# Description

Add a macro to skip variable iv test in Aes-GCM. This is useful for hardware port that do not support IV length != than 12 bytes.


rizlik:
Don't needed anymore

PR #9119, opened 2025-08-22
testing effect of turning off WOLFSSL_ERROR_VERBOSE

effbiae:
# Description
turning off WOLFSSL_ERROR_VERBOSE to see if it affects tests


dgarske:
@effbiae I don't think we can accept this change, but hopefully the CI tells you the results you seek. Please close if you agree. Thanks

PR #9139, opened 2025-08-26
rename enum to avoid reserved name

JacobBarthelmeh:


dgarske (approved):

JacobBarthelmeh:
I think this is maybe part of the fix but it appears to not be the complete fix? Just saw the internal thread that @gojimmypi is investigating the CI failure, will hold off on investigating the recent new CI failures shown. Thanks @gojimmypi.

gojimmypi:
Indeed part of the fix. It is more than a compiler issue. 

After the compile fix, I saw numerous `dangerous relocation: call8: call target out of range: memset`

I'm in the home stretch. See

https://github.com/gojimmypi/wolfssl/actions/workflows/docker-Espressif.yml

gojimmypi:
I put up https://github.com/wolfSSL/wolfssl/pull/9141 to address the failing Espressif workflow.

PR #9154, opened 2025-09-01
Provide default method to get the error string

ThePseudo:
# Description

Since no method is provided through the defines on ssl.h, I think this might be important both for traceability and for compatibility to provide an interface similar to what GNUTLS offers.

Fixes zd#

# Testing

No testing is needed to add one define

# Checklist

 - [ X] added tests
 - [ ] updated/added doxygen

Since it is my first time, is there some pointer to some documentation that tells me how to properly perform the above actions? It is just a one-line change, but maybe something needs to be updated


dgarske:
Hi @ThePseudo ,

For all code contributions we require a signed contributor agreement. As this is a 1 line change we will likely consider this feature request and fix ourselves. @mattia-moffa or @julek-wolfssl would you mind reviewing this report and opening a PR to provide a way to get error strings via the compatibility layer?

Thanks,
David Garske, wolfSSL

mattia-moffa:
Hi @ThePseudo ,

The requested feature is actually already available as `ERR_reason_error_string`:

```
#define ERR_reason_error_string         wolfSSL_ERR_reason_error_string
```

Best regards,
Mattia Moffa

PR #9161, opened 2025-09-04
Make Debian packaging more Debian-like

LinuxJedi:
When you do `./configure`, the options are stored in the rules file. This then means you can use the normal Debian packaging methods.

`make deb` also now uses `dpkg-buildpackage`.

We now have three packages instead of two, because the default Debian build system makes a `dbgsym` package.


dgarske:
Jenkins retest this please. Fix for multi-test merged for tabs.

PR #2302, opened 2019-06-26
Add checks of public key for ECC and curve25519

SparkiDev:

PR #9165, opened 2025-09-05
align two portions of src/internal.c prior to refactor

effbiae:
# Description

if you run this long one-liner in a 160 column wide terminal in the root of a repo, you'll see two sections of `src/internal.c` that are very close but slightly different.

```sh
$ i=0;for x in `grep -n 'if ((args->idx - args->begin) + ENUM_LEN + OPAQUE16_LEN +' src/internal.c |cut -d: -f1`;do i=`expr $i + 1`;sed -n $x,`expr $x + 140`p src/internal.c >internal$i.c; done; diff -W160 -y internal?.c |less ; rm internal?.c
```

i think this is a case of a copy/paste and then later, bugs have been fixed in one version and not the other.  this PR is to bring both versions in line prior to pulling out the code to a function and eliminating the duplication.

that is, if the two sections should be the same

# Testing

`./autogen.sh && ./configure && make test && ./configure --enable-all && make check`

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -32603,6 +32603,9 @@ static int DoServerKeyExchange(WOLFSSL* ssl, const byte* input,
                         ERROR_OUT(ECC_CURVE_ERROR, exit_dske);
                     }
                     ssl->ecdhCurveOID = (word32)curveOid;
+                #if defined(WOLFSSL_TLS13) || defined(HAVE_FFDHE)
+                    ssl->namedGroup = 0;
+                #endif
 
                     length = input[args->idx++];
                     if ((args->idx - args->begin) + length > size) {
@@ -32617,7 +32620,7 @@ static int DoServerKeyExchange(WOLFSSL* ssl, const byte* input,
                             if (ret != 0) {
                                 goto exit_dske;
                             }
-                        } else if (ssl->peerEccKeyPresent) {
+                        } else if (ssl->peerX25519KeyPresent) {
                             ret = ReuseKey(ssl, DYNAMIC_TYPE_CURVE25519,
                                            ssl->peerX25519Key);
                             ssl->peerX25519KeyPresent = 0;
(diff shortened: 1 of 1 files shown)


dgarske:
@effbiae Can you point us to the PR's that caused these to be lost? Thank you

dgarske:
Okay to test. Contributor agreement on file

dgarske:
Jenkins retest this please. Out of space issue. @effbiae if you rebase to latest master it will resolve the Watcom C tests.

effbiae:
> Can you point us to the PR's that caused these to be lost?

these changes weren't lost, they were never applied. there are two 140 line sections of `src/internal.c` that are the same but one section didn't receive the fixes that the other did. this patch applies these fixes. the intention is then to refactor the two identical sections into one function.

the two sections are in `DoServerKeyExchange()` both start with
`                    if ((args->idx - args->begin) + ENUM_LEN + OPAQUE16_LEN +`

i've located the PRs that updated the first section but not the second:

pull request #2840
```
+                    ssl->namedGroup = 0;
commit 6fcfde06519044ca7f71ea4af9bd900383b47bee
Author: Sean Parkinson <sean@wolfssl.com>
Date:   Thu Mar 5 11:20:50 2020 +1000
    Fix to show the FFDHE group when negotiated
```

pull request #1710
```
-                        } else if (ssl->peerEccKeyPresent) {
+                        } else if (ssl->peerX25519KeyPresent) {

commit 6d3e145571d8a371655dc49a0a3e4b64b1d39f78
Author: Sean Parkinson <sean@wolfssl.com>
Date:   Mon Jul 23 10:20:18 2018 +1000

    Changes to build with X25519 and Ed25519 only
```

pull request #2302
```
+                   #ifdef WOLFSSL_EXTRA_ALERTS
+                       SendAlert(ssl, alert_fatal, illegal_parameter);
+                   #endif
commit 4778dac49f65683e6106c1642886e27805a90ee9
Author: Sean Parkinson <sean@wolfssl.com>
Date:   Wed Jun 26 11:56:26 2019 +1000

    Add checks of public key for ECC and curve25519
```

dgarske:
Jenkins retest this please: AgentOfflineException

effbiae:
the macos test errored with  [wolfSSL error: tcp connect failed: Connection refused](https://github.com/wolfSSL/wolfssl/actions/runs/17509667170/job/49896419435?pr=9165#step:2:915) that seems to happen at times

PR #9173, opened 2025-09-08
initialize values used by unsigned bin

rlm2002:
# Description

Adds loop to initialize `a->dp[i]` to 0 in `sp_read_unsigned_bin()` for `a->used`.

# Testing

`./configure && make check`

### DIFF
--- a/wolfcrypt/src/sp_int.c
+++ b/wolfcrypt/src/sp_int.c
@@ -18077,6 +18077,9 @@ int sp_read_unsigned_bin(sp_int* a, const byte* in, word32 inSz)
             a->dp[j++] = *(sp_int_digit*)(in + i - (SP_WORD_SIZEOF - 1));
         }
     #else
+        for (i = 0; i < (int)a->used; i++) {
+            a->dp[i] = 0;
+        }
         /* Construct digit from required number of bytes. */
         for (i = (int)(inSz-1); i >= SP_WORD_SIZEOF - 1; i -= SP_WORD_SIZEOF) {
             a->dp[j]  = ((sp_int_digit)in[i - 0] <<  0)


rlm2002:
Retest this please Jenkins: java.io.StreamCorruptedException and Timeout

wolfcrypt/src/sp_int.c
```diff
@@ -18077,6 +18079,9 @@ int sp_read_unsigned_bin(sp_int* a, const byte* in, word32 inSz)
             a->dp[j++] = *(sp_int_digit*)(in + i - (SP_WORD_SIZEOF - 1));
         }
     #else
+        for (k = 0; k < (int)a->used; k++) {
```
dgarske: Don't add `k`, just use `i`.

dgarske (changes requested):

wolfcrypt/src/sp_int.c
```diff
@@ -18077,6 +18077,9 @@ int sp_read_unsigned_bin(sp_int* a, const byte* in, word32 inSz)
             a->dp[j++] = *(sp_int_digit*)(in + i - (SP_WORD_SIZEOF - 1));
         }
     #else
+        for (i = 0; i < (int)a->used; i++) {
```
SparkiDev, line 18080: Don't want to do this.
We are reading in bytes to create digits and the used will be updated to the length of the created digits.
Need to analyse the problem more deeply.

SparkiDev (changes requested):

wolfcrypt/src/sp_int.c
rlm2002: Got it. Going back to the drawing board.

rlm2002:
Closing lingering PR. May reopen if issue at later date.

PR #9191, opened 2025-09-12
Create initial Rust wrapper structure

holtrop:
# Description

Create initial Rust wrapper structure

Generate bindings to C library with bindgen
Add github CI workflow to build Rust wrapper

# Testing

Tested that new github workflow executes to build Rust wrapper.

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/wrapper/include.am
+++ b/wrapper/include.am
@@ -4,5 +4,6 @@
 
 include wrapper/Ada/include.am
 include wrapper/CSharp/include.am
+include wrapper/rust/include.am
 
 EXTRA_DIST+= wrapper/python/README.md
--- a/wrapper/rust/include.am
+++ b/wrapper/rust/include.am
@@ -1,0 +1,16 @@
+# vim:ft=automake
+# included from Top Level Makefile.am
+# All paths should be given relative to the root
+
+EXTRA_DIST += wrapper/rust/Makefile
+EXTRA_DIST += wrapper/rust/README.md
+EXTRA_DIST += wrapper/rust/wolfssl-sys/Cargo.lock
+EXTRA_DIST += wrapper/rust/wolfssl-sys/Cargo.toml
+EXTRA_DIST += wrapper/rust/wolfssl-sys/Makefile
+EXTRA_DIST += wrapper/rust/wolfssl-sys/build.rs
+EXTRA_DIST += wrapper/rust/wolfssl-sys/headers.h
+EXTRA_DIST += wrapper/rust/wolfssl-sys/src/lib.rs
+EXTRA_DIST += wrapper/rust/wolfssl/Cargo.lock
+EXTRA_DIST += wrapper/rust/wolfssl/Cargo.toml
+EXTRA_DIST += wrapper/rust/wolfssl/Makefile
+EXTRA_DIST += wrapper/rust/wolfssl/src/lib.rs


holtrop:
retest this please (hudson.remoting.RequestAbortedException)

wrapper/rust/wolfssl-sys/Cargo.lock
```diff
@@ -0,0 +1,293 @@
+# This file is automatically @generated by Cargo.
+# It is not intended for manual editing.
```
JacobBarthelmeh, line 2: Are these Cargo.lock files necessary to be checked in to the GitHub repo? I tried with removing them and rebuilding and it seemed to succeed.

wrapper/rust/README.md
```diff
@@ -0,0 +1,17 @@
+# wolfSSL Rust Wrapper
+
+## Building the wolfssl Rust Wrapper
+
+First, configure and build wolfssl C library.
+
+Then build the wolfssl Rust wrapper with:
+
+    make -C wrapper/rust
+
```
JacobBarthelmeh, line 10: How should wolfSSL be configured and built? Does it matter?

Is there an example of using the rust bindings after created? Something like testwolfcrypt running.

.gitignore
```diff
@@ -33,7 +33,7 @@ aclocal.m4
 aminclude.am
 lt*.m4
 Makefile.in
-Makefile
+/Makefile
```
JacobBarthelmeh, line 36: Seems like an extra change checked in? Is this to avoid showing changes in subdirectory Makefiles?

.github/workflows/rust-wrapper.yml
```diff
@@ -0,0 +1,30 @@
...
+    timeout-minutes: 10
+    steps:
+    - name: Build wolfSSL
+      uses: wolfSSL/actions-build-autotools-project@v1
+      with:
+        path: wolfssl
+        configure: --enable-all
+    - name: Build Rust Wrapper
+      working-directory: wolfssl
+      run: make -C wrapper/rust
```
JacobBarthelmeh, line 30: I'd like to see run time addition here.

JacobBarthelmeh (changes requested):
The wrapper needs added to include.am files. To add the files to a bundle created with `make dist`

holtrop:
> The wrapper needs added to include.am files. To add the files to a bundle created with `make dist`

Thanks. I always forget that!

I've added it now.

I wonder if we could have `make dist` pull from `git ls-files` as a base rather than needing to specify all these paths manually?

wrapper/rust/wolfssl-sys/Cargo.lock
holtrop: Yes I intentionally added them. Rebuilding without them will re-generate the lock file, correct. However the point of having the lock file present is that if a third-party dependency is updated to something newer that causes a problem we don't want our build to break because of that. This ensures that known working versions of dependencies are used for CI builds. The lock file can be periodically updated as someone works with the Rust wrapper, but that isn't strictly necessary.

wrapper/rust/README.md
holtrop: The goal of using bindgen is that the FFI bindings are generated based on whatever configuration is used, so the user is free to configure with whatever options are desired.

This PR just adds the initial crates and directory structure. I have another following this that actually adds the first Rust module to wrap functionality and that includes tests and examples.

.gitignore
holtrop: This was intentional. Our .gitignore was ignoring any Makefile anywhere in the directory tree which was not desired. Indeed we already have several Makefile files versioned that match this .gitignore entry! This change tells git to only ignore the top-level generated Makefile.

.github/workflows/rust-wrapper.yml
holtrop: I have timeout-minutes specified for this entire job a few lines above. If that's not what you're asking for let me know!

holtrop:

.github/workflows/rust-wrapper.yml
JacobBarthelmeh: Ahh run time like testwolfcrypt. Where the wrapper was used during run time.

PR #9211, opened 2025-09-17
Allow building with hashdrbg disabled under fips

anhu:
Fixes ZD20493


anhu:
Not sure if this is needed anymore as they went with hashdrbg enabled.

SparkiDev (approved):

PR #9232, opened 2025-09-23
Cleanup debian build

LinuxJedi:
* CFLAGS get pulled in anyway with configure options, or part of the env, we don't need to add them
* Path handling went wrong in one specific platform test run

# Testing

A few different deb-based platforms on VMs and Docker.


LinuxJedi:
Retest this please:
```
java.io.StreamCorruptedException: invalid stream header
```

PR #9233, opened 2025-09-23
SetKeys refactor phase 1

effbiae:
# Description

I've simplified some of `SetKeys()` using a loop over the passed in `Cipher`s.

If you like this simplification I'll finish the job for the remaining algos in `SetKeys()`

# Testing

```
./configure && make check && \
./configure --enable-all \
  --enable-camellia --enable-sm4-cbc --enable-sm4-gcm --enable-sm4-ccm --enable-nullcipher --enable-chacha && \
 make check
```

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/src/keys.c
+++ b/src/keys.c
@@ -2369,32 +2369,6 @@ static int SetPrefix(byte* sha_input, int idx)
 }
 #endif
 
-#define MALLOC_ENC_DEC(name, type)\
-    (void)tls13;\
-    int enc_malloced = 0;\
-    if (enc && enc->name == NULL) {\
-        enc->name = (type*)XMALLOC(sizeof(type), heap, DYNAMIC_TYPE_CIPHER);\
-        if (enc->name == NULL)\
-            return MEMORY_E;\
-        enc_malloced = 1;\
-    }\
-    if (dec && dec->name == NULL) {\
-        dec->name = (type*)XMALLOC(sizeof(type), heap, DYNAMIC_TYPE_CIPHER);\
-        if (dec->name == NULL) {\
-            if (enc_malloced) {\
-                XFREE(enc->name, heap, DYNAMIC_TYPE_CIPHER);\
-                enc->name = NULL;\
-            }\
-            return MEMORY_E;\
-        }\
-    }
-#define FREE_ENC_DEC(name, type)\
-    if (enc && enc->name)\
-        wc_##type##Free(enc->name);\
-    if (dec && dec->name)\
-        wc_##type##Free(dec->name);
-#define IGNORE_VARS (void)decode, (void)server, (void)keys, (void)specs,\
-    (void)heap, (void)devId, (void)rng, (void)tls13
 /*
   short hand macros for *_setkey functions
 */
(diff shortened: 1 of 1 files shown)


src/keys.c
```diff
@@ -2375,50 +2375,29 @@ int SetKeys(Ciphers* enc, Ciphers* dec, Keys* keys, CipherSpecs* specs,
...
-                return ASYNC_INIT_E;
-            }
-        }
-        if (dec) {
-            if (wc_Arc4Init(dec->arc4, heap, devId) != 0) {
-                WOLFSSL_MSG("Arc4Init failed in SetKeys");
-                return ASYNC_INIT_E;
+        for (i = 0; i < 2; i++) {
+            if (ed[i]) {
+                if (ed[i]->arc4 == NULL) {
```
SparkiDev: What I would prefer to see is this block refactored out into its own function.
Could you do this for all the ciphers?

effbiae: the challenge is slight differences in the handling for each cipher.  
  - some ciphers need to deal with `keys->{client,server}_write_IV`, some dont
  - differing number of args to `wc_..._SetKey`
  - for some ciphers, there is `wc_...Free`, not for others
  - ... other differences

a complete rework i could implement is
```c
int SetKeys(Ciphers* enc, Ciphers* dec, Keys* keys, CipherSpecs* specs,
                   int side, void* heap, int devId, WC_RNG* rng, int tls13)
{   
    Ciphers* ed[] = {enc, dec};
    int i, ret;
    int server = side != WOLFSSL_CLIENT_END;
    int algo = specs->bulk_cipher_algorithm;
    for (i = 0; i < 2; i++) {
        if (ed[i]) {
            if (!Cipher_Inits[algo])
                return NOT_COMPILED_IN;
            ret = Cipher_Inits[algo](ed[i], heap, devId);
            if (ret) return ret;
            ret = Cipher_SetKeys[algo](ed[i], i, server, keys, specs);
            if (ret) return ret;
            ed[i]->setup = 1;
        }
    }
...
    return 0;
}
```
where i setup arrays of Cipher_Inits and Cipher_SetKeys functions like:
```c
static int rc4_init(Ciphers* c, void* heap, int devId) {...}
static int des_init(Ciphers* c, ...
    
static int (*Cipher_Inits[CIPHER_ALGO_LEN])(Ciphers* c, void* heap, int devId) = {
    0, rc4_init, 0, des_init, ... }

static int (*Cipher_SetKeys[CIPHER_ALGO_LEN])(Ciphers* c, Keys* keys, CipherSpecs* specs, WC_RNG* rng, int tls13) = {
    0, rc4_set_key, 0, des_set_key, ... }    
```
so there will be a number of small `[cipher]_init` and `[cipher]_set_key` functions introduced that do the work.
i like to always reduce the LOC count with my patches and this should still reduce LOC, even with all the small functions.

SparkiDev: Don't try to have a common function signature.
Each one is different and we have to live with that.
Pulling out a separate function for each cipher makes the code easier to read and understand, and eliminates duplication.

effbiae:
i'm quite happy with this refactor.  it reduces the codebase by 430 lines and simplifies SetKeys()

NB. i am unable to test aria -- 
```
./configure --enable-aria
...
configure: error: Error including mcapi_error.h. Please put the MagicCrypto library in /home/jack/wolfssl/wolfssl.
```
the internet doesn't know about MagicCrypto .. will the CI test this?

dgarske:
Okay to test

effbiae:
hmmm -- what happened in PRB-master-job and Visual Studio Build Test?

dgarske:
Lots of errors like this:

```
keys.c

c:\jenkins\workspace\prb-windows-test-v2\src\keys.c(2471): error C2143: syntax error : missing ';' before 'type' [C:\jenkins\workspace\PRB-windows-test-v2\wolfssl.vcxproj]

c:\jenkins\workspace\prb-windows-test-v2\src\keys.c(2471): error C2065: 'enc_malloced' : undeclared identifier [C:\jenkins\workspace\PRB-windows-test-v2\wolfssl.vcxproj]

c:\jenkins\workspace\prb-windows-test-v2\src\keys.c(2499): error C2143: syntax error : missing ';' before 'type' [C:\jenkins\workspace\PRB-windows-test-v2\wolfssl.vcxproj]

c:\jenkins\workspace\prb-windows-test-v2\src\keys.c(2499): error C2065: 'enc_malloced' : undeclared identifier [C:\jenkins\workspace\PRB-windows-test-v2\wolfssl.vcxproj]

c:\jenkins\workspace\prb-windows-test-v2\src\keys.c(2850): warning C4204: nonstandard extension used : non-constant aggregate initializer [C:\jenkins\workspace\PRB-windows-test-v2\wolfssl.vcxproj]

```

SparkiDev:
Hi @effbiae,

I've consulted with the other engineers internally and we all agree that function tables are not something we won't introduced into wolfSSL.
There can be performance impacts and debugging can become harder.

We appreciate that you are trying to make the code more compact but we don't want this at the cost of maintainability.

Please call the extracted functions directly in the each case.
Note that customers are only like to use 2 or three ciphers and the saving in code size is minimal.

Sean

effbiae:
ok, sounds good.  next question is, do you find IF_NULL(), IF_ARC4() appealing or distracting?  the general idea is

```c
#ifdef HAVE_FEATURE
#define IF_FEATURE(a,b) a
#else
#define IF_FEATURE(a,b) b
#endif
```
which chooses between a and b depending on the HAVE_ macro. it saves a few lines and can be nice if you like that sort of thing. 

otherwise, there's a bunch of `(void)var;` that i can get rid of by reducing the parameters of the _alloc and _setkey functions.

but if we're not close here, it might be best just dropping this change

dgarske:
Hi @effbiae , we appreciate your efforts to cleanup the setkey, but the end result doesn't appear to improve the quality of the code. For now I will go ahead and close the PR. If you'd like to try again we are happy to review. Thanks!

PR #9263, opened 2025-10-02
RSA API: use const pointers and clean up some comments

holtrop:
# Description

RSA API: use const pointers and clean up some comments

Note: this was mainly driven by creating the Rust wrapper and trying to avoid needing mutable references in some Rust APIs when a non-mutable reference could be used.

# Testing

Local build and CI tests.

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/wolfcrypt/src/port/st/stm32.c
+++ b/wolfcrypt/src/port/st/stm32.c
@@ -721,7 +721,7 @@ static int stm32_get_from_mp_int(uint8_t *dst, const mp_int *a, int sz)
         XMEMSET(dst, 0, offset);
 
     /* convert mp_int to array of bytes */
-    res = mp_to_unsigned_bin((mp_int*)a, dst + offset);
+    res = mp_to_unsigned_bin(a, dst + offset);
     return res;
 }
 
--- a/wolfssl/wolfcrypt/rsa.h
+++ b/wolfssl/wolfcrypt/rsa.h
@@ -352,28 +352,28 @@ WOLFSSL_API int  wc_RsaPSS_VerifyInline(byte* in, word32 inLen, byte** out,
 WOLFSSL_API int  wc_RsaPSS_VerifyInline_ex(byte* in, word32 inLen, byte** out,
                                            enum wc_HashType hash, int mgf,
                                            int saltLen, RsaKey* key);
-WOLFSSL_API int  wc_RsaPSS_Verify(byte* in, word32 inLen, byte* out,
+WOLFSSL_API int  wc_RsaPSS_Verify(const byte* in, word32 inLen, byte* out,
                                   word32 outLen, enum wc_HashType hash, int mgf,
                                   RsaKey* key);
-WOLFSSL_API int  wc_RsaPSS_Verify_ex(byte* in, word32 inLen, byte* out,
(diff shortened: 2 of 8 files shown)


wolfssl/wolfcrypt/rsa.h
```diff
@@ -352,28 +352,28 @@ WOLFSSL_API int  wc_RsaPSS_VerifyInline(byte* in, word32 inLen, byte** out,
 WOLFSSL_API int  wc_RsaPSS_VerifyInline_ex(byte* in, word32 inLen, byte** out,
                                            enum wc_HashType hash, int mgf,
                                            int saltLen, RsaKey* key);
-WOLFSSL_API int  wc_RsaPSS_Verify(byte* in, word32 inLen, byte* out,
+WOLFSSL_API int  wc_RsaPSS_Verify(const byte* in, word32 inLen, byte* out,
```
dgarske, line 355: I am not sure we can or should change public facing API's, even adding 'const`. I can guarentee this will break builds for some customers.

dgarske (changes requested):
You may need to rethink this PR. Would love feedback from @danielinux and @julek-wolfssl and @douzzer

wolfssl/wolfcrypt/rsa.h
holtrop: Hmm, I feel like I might have run into this before but forgotten the rationale. Why is it problematic to add `const`?

dgarske: It will break API compatibility. I'm fairly certain we cannot accept the changes where a public WOLFSSL_API is modified even to add const.

dgarske (changes requested):

holtrop:
retest this please (ERROR: Error cloning remote repo 'origin')

holtrop:
retest this please (hudson.remoting.RequestAbortedException)

wolfcrypt/src/port/st/stm32.c
```diff
@@ -721,7 +721,7 @@ static int stm32_get_from_mp_int(uint8_t *dst, const mp_int *a, int sz)
         XMEMSET(dst, 0, offset);
 
     /* convert mp_int to array of bytes */
-    res = mp_to_unsigned_bin((mp_int*)a, dst + offset);
+    res = mp_to_unsigned_bin(a, dst + offset);
```
julek-wolfssl, line 724: Why the cast here?

julek-wolfssl (changes requested):
In general I like adding `const` to input parameters where that promise is kept by the function. I don't think it should be causing build errors.

wolfcrypt/src/port/st/stm32.c
holtrop: I removed the cast. It was previously casting away const because mp_to_unsigned_bin() did not take a const pointer. But now it does so we do not need the cast.

dgarske (approved):
Let's opt for the improvement. There might be some customers with strict casting that issue support tickets, but that is a technical debt that we can justify. Thank you @holtrop

douzzer (approved):
excellent idea -- and we should keep going with these opportunistic `const`s for other APIs.

note that the incumbent numeric literal at doc/dox_comments/header_files/rsa.h‎:64 should be done with `strlen()`, but that can wait for another time.

PR #9274, opened 2025-10-06
Coverity: Overflowed arg and dead code

rlm2002:
# Description

549097 Overflowed integer argument - Add casts to `cnt` and result of `MIN()`.

549096 Logically dead code - Remove second NULL check for `extension`

# Testing


rlm2002:
Changes done on old branch. Issues have already been fixed.

PR #9285, opened 2025-10-09
SP: macros for variables that may be allocated

SparkiDev:
# Description

Add macros to make the code simpler around allocating from dynamic memory when WOLFSSL_SP_SMALL_STACK is defined.
Change over to using macros where it makes sense.

# Testing

Regression tested SP.

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


SparkiDev:
retest this please

Also:   hudson.remoting.Channel$CallSiteStackTrace: Remote call to wolf-linux-cloud-node-x85zdd

PR #9286, opened 2025-10-09
Rust wrapper: add wolfssl::wolfcrypt::ecc module

holtrop:
# Description

Rust wrapper: add wolfssl::wolfcrypt::ecc module

# Testing

How did you test?

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/wrapper/rust/wolfssl/src/wolfcrypt/ecc.rs
+++ b/wrapper/rust/wolfssl/src/wolfcrypt/ecc.rs
@@ -1270,10 +1270,9 @@ impl ECC {
     /// let _x963_size = ecc.export_x963(&mut x963).expect("Error with export_x963()");
     /// ```
     pub fn export_x963(&mut self, dout: &mut [u8]) -> Result<usize, i32> {
-        let dout_ptr = dout.as_ptr() as *mut u8;
         let mut out_len: u32 = dout.len() as u32;
         let rc = unsafe {
-            ws::wc_ecc_export_x963(&mut self.wc_ecc_key, dout_ptr, &mut out_len)
+            ws::wc_ecc_export_x963(&mut self.wc_ecc_key, dout.as_mut_ptr(), &mut out_len)
         };
         if rc != 0 {
             return Err(rc);
@@ -1303,10 +1302,9 @@ impl ECC {
     /// let _x963_size = ecc.export_x963_compressed(&mut x963).expect("Error with export_x963_compressed()");
     /// ```
     pub fn export_x963_compressed(&mut self, dout: &mut [u8]) -> Result<usize, i32> {
-        let dout_ptr = dout.as_ptr() as *mut u8;
         let mut out_len: u32 = dout.len() as u32;
         let rc = unsafe {
-            ws::wc_ecc_export_x963_ex(&mut self.wc_ecc_key, dout_ptr, &mut out_len, 1)
+            ws::wc_ecc_export_x963_ex(&mut self.wc_ecc_key, dout.as_mut_ptr(), &mut out_len, 1)
         };
         if rc != 0 {
             return Err(rc);


holtrop:
retest this please (Error: The operation was canceled)

wrapper/rust/wolfssl/src/wolfcrypt/ecc.rs
```diff
@@ -0,0 +1,1613 @@
...
+    /// # Parameters
+    ///
+    /// * `din`: DER-formatted buffer.
+    /// * `curve_id`: Curve ID, e.g. ECC::SECP256R1.
+    ///
+    /// # Returns
+    ///
+    /// Returns either Ok(ECCPoint) containing the ECCPoint struct instance or
+    /// Err(e) containing the wolfSSL library error code value.
+    pub fn import_der(din: &[u8], curve_id: i32) -> Result<Self, i32> {
```
gasbytes, line 55: Instead of Result<Self, i32>, this could become an enum to make it more idiomatic, something like:

```
pub enum EccError {
    BadArg,
    Memory,
    InvalidKey,
    /* this one catches all the other error codes that are not specified */
    WolfCrypt(i32), 
}
```

wrapper/rust/wolfssl/src/wolfcrypt/ecc.rs
```diff
@@ -0,0 +1,1613 @@
...
+        let curve_idx = unsafe { ws::wc_ecc_get_curve_idx(curve_id) };
+        if curve_idx < 0 {
+            return Err(curve_idx);
+        }
+        let wc_ecc_point = unsafe { ws::wc_ecc_new_point() };
+        if wc_ecc_point.is_null() {
+            return Err(0);
```
gasbytes: Since zero in wolfSSL indicates success, shouldn't this be ECC_BAD_ARG_E (-170) for consistency?

wrapper/rust/wolfssl/src/wolfcrypt/ecc.rs
```diff
@@ -0,0 +1,1613 @@
...
+    ///
+    /// ```rust
+    /// use wolfssl::wolfcrypt::random::RNG;
+    /// use wolfssl::wolfcrypt::ecc::ECC;
+    /// let mut rng = RNG::new().expect("Failed to create RNG");
+    /// let mut ecc = ECC::generate(32, &mut rng).expect("Error with generate()");
+    /// let ecc_point = ecc.make_pub_to_point(Some(&mut rng)).expect("Error with make_pub_to_point()");
+    /// ecc_point.forcezero();
+    /// ```
+    pub fn forcezero(&self) {
```
gasbytes: Since this mutates the underlying point, maybe it would be better if this became &mut self to better reflect the operation's semantics?:

```suggestion
    pub fn forcezero(&mut self) {
```

wrapper/rust/wolfssl/src/wolfcrypt/ecc.rs
```diff
@@ -0,0 +1,1613 @@
...
+    /// use wolfssl::wolfcrypt::ecc::ECC;
+    /// let mut rng = RNG::new().expect("Failed to create RNG");
+    /// let mut ecc = ECC::generate(32, &mut rng).expect("Error with generate()");
+    /// let mut x963 = [0u8; 128];
+    /// let x963_size = ecc.export_x963(&mut x963).expect("Error with export_x963()");
+    /// let x963 = &x963[0..x963_size];
+    /// let _ecc2 = ECC::import_x963(x963).expect("Error with import_x963()");
+    /// ```
+    pub fn import_x963(din: &[u8]) -> Result<ECC, i32> {
+        let din_ptr = din.as_ptr() as *const u8;
```
gasbytes: This can be simplified with:
```suggestion
        let din_ptr = din.as_ptr();
```
Since din is &[u8], as_ptr() already returns *const u8

wrapper/rust/wolfssl/src/wolfcrypt/ecc.rs
```diff
@@ -0,0 +1,1613 @@
...
+    /// ```rust
+    /// use wolfssl::wolfcrypt::random::RNG;
+    /// use wolfssl::wolfcrypt::ecc::ECC;
+    /// let mut rng = RNG::new().expect("Failed to create RNG");
+    /// let mut ecc = ECC::generate(32, &mut rng).expect("Error with generate()");
+    /// let mut x963 = [0u8; 128];
+    /// let _x963_size = ecc.export_x963(&mut x963).expect("Error with export_x963()");
+    /// ```
+    pub fn export_x963(&mut self, dout: &mut [u8]) -> Result<usize, i32> {
+        let dout_ptr = dout.as_ptr() as *mut u8;
```
gasbytes: This can be simplified with:
```suggestion
        let dout_ptr = dout.as_mut_ptr();
```

wrapper/rust/wolfssl/src/wolfcrypt/ecc.rs
holtrop: Yes, this has been discussed and is on the "to do" list across all the Rust modules. For now this matches the other merged modules but I plan to visit that (and several other things) after completing initial versions of the modules.

wrapper/rust/wolfssl/src/wolfcrypt/ecc.rs
holtrop: Yeah I think it would make sense to change. I think `MEMORY_E`  would be more applicable; it isn't related to a bad argument. Initially my thought was to indicate something that the Rust library could detect that was different from something the C library would possibly return, since Rust has that capability with its Result enum, but it's probably confusing to use 0. When we revisit errors in general in the future, perhaps it would make sense to add some Rust-specific error codes even? Could be worth a discussion then. For now I'll change to `MEMORY_E`.

wrapper/rust/wolfssl/src/wolfcrypt/ecc.rs
holtrop: Thanks, I will change that. I originally made several functions templated to take an array of any value and cast that to *const u8, but that seemed more work than it was worth in the end.

wrapper/rust/wolfssl/src/wolfcrypt/ecc.rs
holtrop: This is an interesting philosophical question... originally I had ECCPoint storing the ecc_point struct itself, not a pointer to it. But wolfCrypt doesn't have an API to initialize/free a static ecc_point like it does for many other structs, so I switched to only storing a pointer, which has other advantages as well. This meant I no longer needed a mut reference here. I was used to the Rust compiler warning that variables didn't need to be mut so expected the same thing here but it does not appear to emit the same warning for function parameters (perhaps just references?). In any case, this seems analogous to having a 'const' pointer in C that points to non-const data that is modified, and it still seems perfectly fair in that case to mark the pointer as const even if you're accessing and modifying data through it. On the other hand, the Rust view is more that you're modifying this "object" while thinking of the object opaquely not knowing that it really is just a pointer. Ok, I'll add the 'mut'. It will be good to keep this idea in mind if we switch the other structs to using heap-allocated objects as well since they may switch to pointers also. Thanks!

holtrop:
retest this please (wolfSSL error: SSL_accept failed)

dgarske:
Jenkins retest this please. Some tests were aborted

PR #9288, opened 2025-10-09
Disable WOLFSSL_OLD_OID_SUM on Python builds

mattia-moffa:
# Description

This makes sure `WOLFSSL_OLD_OID_SUM` is no longer enabled when building with `WOLFSSL_PYTHON`. After wolfSSL/osp#303 that option is no longer needed for the CPython tests to pass.

# Testing

```
./configure --enable-all
./configure --enable-python
```


dgarske:
Jenkins retest this please

PR #9294, opened 2025-10-13
Benchmark memory tracking

LinuxJedi:
This adds optional heap and stack tracking to wolfCrypt bench so that it is possible to see RAM usage. It also adds support for stack tracking in microcontrollers (tested on STM32).

# Testing

Manually tested on Linux and an STM32U585.

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/wolfcrypt/benchmark/benchmark.c
+++ b/wolfcrypt/benchmark/benchmark.c
@@ -242,7 +242,7 @@
 
 /* define the max length for each string of metric reported */
 #ifndef WC_BENCH_MAX_LINE_LEN
-#define WC_BENCH_MAX_LINE_LEN 240
+#define WC_BENCH_MAX_LINE_LEN 150
 #endif
 
 /* default units per second. See WOLFSSL_BENCHMARK_FIXED_UNITS_* to change */
@@ -262,6 +262,12 @@
 
 #if defined(HAVE_STACK_SIZE_VERBOSE)
     #define WC_BENCH_STACK_TRACKING
+#endif
+
+#if (defined(WC_BENCH_HEAP_TRACKING) || defined(WC_BENCH_STACK_TRACKING)) && \
+    (WC_BENCH_MAX_LINE_LEN < 240)
+    #undef WC_BENCH_MAX_LINE_LEN
+    #define WC_BENCH_MAX_LINE_LEN 240
 #endif
 
 #if defined(WC_BENCH_HEAP_TRACKING) || defined(WC_BENCH_STACK_TRACKING)


wolfcrypt/benchmark/benchmark.c
```diff
@@ -2519,12 +2723,6 @@ static void bench_multi_value_stats(double max, double min, double sum,
                 word[2], FLT_FMT_PREC2_ARGS(3, 3, mean),
                 word[3], FLT_FMT_PREC2_ARGS(3, 3, sd));
     }
-    printf("%s", msg);
```
dgarske, line 2525: Should these lines have been removed?

wolfssl/wolfcrypt/mem_track.h
```diff
@@ -766,6 +883,50 @@ static WC_INLINE int StackSizeCheck_reap(pthread_t threadId,
...
+            printf("stack used = %ld\n", (long)used);
+            StackSizeHWMReset();
+        }
+    }
+#endif
+
+    return (int)((size_t)status);
+}
+
+static WC_INLINE int StackSizeCheck_launch(struct func_args* args,
```
dgarske: Do these need to be a static inline? Could they just be a macro?

LinuxJedi:
> Any idea why the total allocs reported at end is so different than the total of the individual heap allocs?
> 

The total and peak are for the entire application, including everything outside of running the actual individual benchmarks. I can remove that if it is confusing.

wolfcrypt/benchmark/benchmark.c
LinuxJedi: Nope, sorry.

LinuxJedi:
Jenkins retest this please, PRB-fipsv2-regression-and-rolling-release-v3

dgarske:
> > Any idea why the total allocs reported at end is so different than the total of the individual heap allocs?
> 
> The total and peak are for the entire application, including everything outside of running the actual individual benchmarks. I can remove that if it is confusing.

Can you figure out why the XMALLOC isn't tracking some of the heap uses in wolfCrypt ... the ones with `[heap 0 bytes (0 allocs),`?

LinuxJedi:
Without small-stack enabled, most benchmarks use stack instead of heap. That being said, we were tracking from the point of the timer starting, which was after init of each test (such as `wc_AesInit` and associated output allocs). The code has now been modified to track heap/stack during init, to get a better representation, without altering the timer.

LinuxJedi:
Jenkins retest this please: Agent was removed

wolfcrypt/benchmark/benchmark.c
```diff
@@ -242,7 +242,7 @@
 
 /* define the max length for each string of metric reported */
 #ifndef WC_BENCH_MAX_LINE_LEN
-#define WC_BENCH_MAX_LINE_LEN 150
+#define WC_BENCH_MAX_LINE_LEN 240
```
dgarske: Consider only increasing this stack variable when `WC_BENCH_HEAP_TRACKING` or `WC_BENCH_STACK_TRACKING` is set? Might need to move this lower in the .c.

dgarske (changes requested):
Awesome job, just one suggestion (optional).

LinuxJedi:
Jenkins retest this please: workflow didn't appear to run?

PR #9295, opened 2025-10-13
wolfSSL_shutdown: handle non-blocking I/O

rizlik:
# Description

1. Send buffered data in case early calls returned WANT_WRITE, maintain current API behavior (returning SHUTDOWN_NOT_DONE and don't trying to read for the socket on the first invocation, try to read from the socket for bi-directional shutdown in later calls).
2. propagate WANT_READ when reading from the socket in wolfSSL_shutdown.
3. Improve test coverage.

Fixes zd#20623

### DIFF
--- a/tests/api.c
+++ b/tests/api.c
@@ -46755,25 +46755,12 @@ static int test_extra_alerts_bad_psk(void)
 }
 #endif
 
-#if defined(OPENSSL_EXTRA) && defined(HAVE_MANUAL_MEMIO_TESTS_DEPENDENCIES) && !defined(WOLFSSL_NO_TLS12)
-/*
- * Emulates wolfSSL_shutdown that goes on EAGAIN,
- * by returning on output WOLFSSL_ERROR_WANT_WRITE.*/
-static int custom_wolfSSL_shutdown(WOLFSSL *ssl, char *buf,
-        int sz, void *ctx)
-{
-    (void)ssl;
-    (void)buf;
-    (void)ctx;
-    (void)sz;
-
-    return WOLFSSL_CBIO_ERR_WANT_WRITE;
-}
-
-static int test_multiple_alerts_EAGAIN(void)
+#if defined(HAVE_MANUAL_MEMIO_TESTS_DEPENDENCIES) && !defined(WOLFSSL_NO_TLS12)
+static int test_multiple_shutdown_nonblocking(void)
 {
     EXPECT_DECLS;
     size_t size_of_last_packet = 0;
+    int dummy_recv_buffer;
 
     /* declare wolfSSL objects */
     struct test_memio_ctx test_ctx;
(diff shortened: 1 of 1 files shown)


rizlik:
retest this please:
unrecognized options: --with-libtool-sysroot

src/ssl.c
```diff
@@ -4478,6 +4478,36 @@ int wolfSSL_shutdown(WOLFSSL* ssl)
...
+            if (ssl->options.sentNotify)  {
+                if (ssl->options.closeNotify) {
+                  ssl->options.shutdownDone = 1;
+                  ret = WOLFSSL_SUCCESS;
+                }
+                else {
+                  ret = WOLFSSL_SHUTDOWN_NOT_DONE;
+                }
+                WOLFSSL_LEAVE("wolfSSL_shutdown", ret);
+                return ret;
```
julek-wolfssl: 4 space tabs

src/ssl.c
```diff
@@ -4518,6 +4554,9 @@ int wolfSSL_shutdown(WOLFSSL* ssl)
                 ret = WOLFSSL_SUCCESS;
             } else if (ret == WC_NO_ERR_TRACE(MEMORY_E)) {
                 ret = WOLFSSL_FATAL_ERROR;
+            } else if (ret == WC_NO_ERR_TRACE(WANT_READ)) {
+                ssl->error = ret;
+                ret = WOLFSSL_FATAL_ERROR;
             } else if (ssl->error == WOLFSSL_ERROR_NONE) {
                 ret = WOLFSSL_SHUTDOWN_NOT_DONE;
             } else {
```
julek-wolfssl: all `else` should be in new line

rizlik:
> In `SendAlert_ex` we check for `ssl->options.sentNotify` and if its set we return but at the end of the function. I think it would be good to move it up to the top to exit early if we're not going to be sending the alert.
> 
:+1: 

> Add a comment inside `wolfSSL_shutdown` that `SendAlert` sets `ssl->options.sentNotify`.

 Also maybe don't set it in `wolfSSL_shutdown`?

I think that's needed for QUIC and DUP codepaths that do not enter SendAlert_ex.

julek-wolfssl:
> I think that's needed for QUIC and DUP codepaths that do not enter SendAlert_ex.

I see the quic callback. Maybe it should be set when the callback returns success?

For DUP do you think it would be ok to set it on the read side when `NotifyWriteSide` returns success?

rizlik:
> > I think that's needed for QUIC and DUP codepaths that do not enter SendAlert_ex.
> 
> I see the quic callback. Maybe it should be set when the callback returns success?
> 
> For DUP do you think it would be ok to set it on the read side when `NotifyWriteSide` returns success?

I moved the sentNotify set logic inside `wolfSSL_shutdown` only.

PR #9306, opened 2025-10-15
Rust wrapper: add wolfssl::wolfcrypt::dh module

holtrop:
# Description

Rust wrapper: add wolfssl::wolfcrypt::dh module

# Testing

How did you test?

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/wrapper/rust/wolfssl/src/wolfcrypt/dh.rs
+++ b/wrapper/rust/wolfssl/src/wolfcrypt/dh.rs
@@ -40,11 +40,16 @@ pub struct DH {
 }
 
 impl DH {
+    /// ffdhe2048 named parameter group.
     pub const FFDHE_2048: i32 = ws::WC_FFDHE_2048 as i32;
-    pub const FFDHE_3072: i32 = ws::WC_FFDHE_2048 as i32;
-    pub const FFDHE_4096: i32 = ws::WC_FFDHE_2048 as i32;
-    pub const FFDHE_6144: i32 = ws::WC_FFDHE_2048 as i32;
-    pub const FFDHE_8192: i32 = ws::WC_FFDHE_2048 as i32;
+    /// ffdhe3072 named parameter group.
+    pub const FFDHE_3072: i32 = ws::WC_FFDHE_3072 as i32;
+    /// ffdhe4096 named parameter group.
+    pub const FFDHE_4096: i32 = ws::WC_FFDHE_4096 as i32;
+    /// ffdhe6144 named parameter group.
+    pub const FFDHE_6144: i32 = ws::WC_FFDHE_6144 as i32;
+    /// ffdhe8192 named parameter group.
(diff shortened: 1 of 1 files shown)


dgarske (dismissed):

wrapper/rust/wolfssl/src/wolfcrypt/dh.rs
```diff
@@ -0,0 +1,1109 @@
...
+use std::ptr::null;
+use crate::wolfcrypt::random::RNG;
+
+pub struct DH {
+    wc_dhkey: ws::DhKey,
+}
+
+impl DH {
+    pub const FFDHE_2048: i32 = ws::WC_FFDHE_2048 as i32;
+    pub const FFDHE_3072: i32 = ws::WC_FFDHE_2048 as i32;
```
gasbytes: Shouldn't this be `ws::WC_FFDHDE_3072`? Same for all the other constants matching the constant name.

wrapper/rust/wolfssl/src/wolfcrypt/dh.rs
```diff
@@ -0,0 +1,1109 @@
...
+    /// let mut private1 = [0u8; 256];
+    /// let mut private1_size = 0u32;
+    /// let mut public1 = [0u8; 256];
+    /// let mut public1_size = 0u32;
+    /// dh.generate_key_pair(&mut rng, &mut private1, &mut private1_size, &mut public1, &mut public1_size).expect("Error with generate_key_pair()");
+    /// let mut ss0 = [0u8; 256];
+    /// let ss0_size = dh.shared_secret(&mut ss0, &private0, &public1).expect("Error with shared_secret()");
+    /// let ss0 = &ss0[0..ss0_size];
+    /// ```
+    pub fn shared_secret(&mut self, dout: &mut [u8], private: &[u8], other_pub: &[u8]) -> Result<usize, i32> {
```
gasbytes, line 1086: Maybe this one can be `u32` instead of `usize`, since u32 is used throughout the whole function, though all type casting could be also changed to `usize` since it matches slice indexing.

wrapper/rust/wolfssl/src/wolfcrypt/dh.rs
holtrop: Yes, thank you. Fixed.

wrapper/rust/wolfssl/src/wolfcrypt/dh.rs
holtrop: I originally used u32 everywhere and no usize. When writing the unit tests I discovered that unfortunately Rust will complain about using a u32 when comparing to slice.len() or when using a u32 variable as a slice index, which is very frustrating. I felt it was better to use usize so that the user of our library did not need to manually convert the u32 to a usize in order to use it for len() comparison or as an argument when slicing.

I know Rust is supposed to be safer than C, but I feel like this particular aspect of not automatically converting a u32 to a usize is more annoying than helpful.

PR #9309, opened 2025-10-15
Add crypto callback support for copy/free operations (SHA-256)

night1rider:
Proposing a general Copy Callback (and probably a Free Callback) for users to be able to use if they have hardware. Ran into a situation where this is more elegant solution than trying to scheme a way to produce copies and handle contexts coming from hardware, along with freeing. The goal would to be able to have independent devCtx's. The hardware can only open up so many "Streams", so I need to keep track of the context and close them down as soon as I am done with the operation.

Looking over wc_Sha256Free we would probably also want to include a callback to free the context there as well.

The way I am currently utilizing the hardware (Very Basic Overview):
```mermaid
graph LR
    A[wc_Sha256Update<br>via CryptoCb] --> B[Start Hardware Stream]
    B --> C{devCtx != NULL?}
    C -->|Yes| D[Import Hardware Context<br/>from devCtx to Hardware]
    C -->|No| E[Fresh Stream<br/>No import needed]
    D --> F[Request Hash Update <br/>Hash/Update]
    E --> F
    F --> G[Wait for Hardware Response]
    G --> H[Free Old DevCtx then<br/>Export Hardware Context<br/>to devCtx]
    H --> I[Close Hardware Stream]
    I --> J[Return Success]
```

The way I am currently utilizing the hardware for finishing a sha256 stream:
```mermaid
graph LR
    Start[wc_Sha256Final<br/>via CryptoCb] --> OpenStream[Start hardware Stream]
    OpenStream --> CheckCtx{devCtx != NULL?}
    CheckCtx -->|Yes| Import[Import Hardware Context<br/>Restore hash state]
    CheckCtx -->|No| Init[Fresh Stream<br/>Fresh hash operation]
    Import --> Finalize[Request Finalization]
    Init --> Finalize
    Finalize --> Wait[Wait for Hardware<br/>Operation to Complete]
    Wait --> Close[Close Hardware Stream]
    Close --> FreeCtx{devCtx allocated?}
    FreeCtx -->|Yes| Free[Free devCtx]
    FreeCtx -->|No| End[Return Success<br/>with digest]
    Free --> End
```

This is the current situation:
Copy then Free:
```mermaid
graph TD
    subgraph "Current Behavior"
    A[wc_Sha256 Context A<br/>devCtx pointer 0x01] -->|wc_Sha256Copy| B{Shallow Copy}
    B -->|src| C[Context A<br/>devCtx = 0x01]
    B -->|dst| D[Context B<br/>devCtx = 0x01<br/><span style="color:red">SAME POINTER!</span>]
    C -->|wc_Sha256Final| E[Context A Freed<br/>devCTX at 0x01]
    D -->|After wc_Sha256Final: wc_Sha256Update| F[Context B tries to use<br/>devCTX at 0x01<br/><span style="color:red">ALREADY FREED!</span>]
    end
```

```mermaid
graph TD
    subgraph "Copy with Crypto Callback"
    A[wc_Sha256 Context A<br/>devCtx pointer 0x01] -->|wc_Sha256Copy| B{Copy<br/>via CryptoCb}
    B -->|src| C[Context A<br/>devCtx = 0x01]
    B -->|dst| D[Context B<br/>devCtx = 0x02<br/><span style="color:green">NEW STREAM!</span>]
    C -->|wc_Sha256Final| E[Context A Freed<br/>dev 0x01]
    D -->|Later: wc_Sha256Update| F[✓ Context B uses<br/>devCtx at 0x02<br/><span style="color:green">Independent Stream</span>]
    end
```


bigbrett:
@dgarske just some additional context/background for you...Very interested to hear your thoughts.

One tempting alternative @night1rider and I had discussed would be to modify the existing cryptoCallbacks for algorithms that DO support a "copy" operation to hold dst/src in the info struct and a flag or enum field indicating the type of operations (init/update/final/copy/free). 

I think there are pros and cons to both approaches, but we ultimately decided on proposing this one.

Cons to this approach mean maintaining an additional enum type indicating to the caller exactly which type the copy was invoked on (so they know what type they should cast the `void*` `src`/`dst` pointers to in their crypto callback copy handler). 

Pros mean we don't break existing customer crypto callbacks by modifying the `info` structs their applications are currently using (which could fail in all sorts of weird ways given that it is a union type). This was deemed more important that a potentially simpler/cleaner implementation in the other solution path.

We were also hoping to integrate this into the `CMD` crypto callback, but decided against it for the same reason -didn't want to break existing customer/port cryptoCbs

wolfcrypt/src/cryptocb.c
```diff
@@ -174,6 +174,16 @@ static const char* GetCryptoCbCmdTypeStr(int type)
 }
 #endif
 
+#ifndef NO_COPY_CB
```
dgarske: Consider having this feature off by default.

wolfssl/wolfcrypt/cryptocb.h
```diff
@@ -101,6 +101,15 @@ enum wc_CryptoCbCmdType {
 };
 #endif
 
+#ifndef NO_COPY_CB
+/* CryptoCb Copy Types - for copy operations on algorithm structures */
```
dgarske: Can't you use the existing enums like `WC_HASH_TYPE_SHA256`? This copy is only for hash, although eventually it could be for ciphers too. I see the HASH_TYPE and CIPHER_TYPE enums have conflicting values. So you should also supply wc_AlgoType. Then you'll have the same cryptocb matrix of algorithm and type.

dgarske (changes requested):

wolfssl/wolfcrypt/cryptocb.h
bigbrett: The reason was because for some algorithms we anticipated you might need both algo type and the subtype in order to deduce the type information for how to cast away the void pointer for src/dest. But by nature of it being a new cryptoCb algo type itself, you would lose that information. Not an issue if sha is only the ever "copy type" but if we add it for ciphers then it becomes a problem.


EDIT: Looks like @night1rider fixed this by having the caller propagate the necessary info to the cryptocb

wolfssl/wolfcrypt/cryptocb.h
```diff
@@ -468,6 +469,19 @@ typedef struct wc_CryptoInfo {
         void *ctx;
     } cmd;
 #endif
+#ifdef WOLFSSL_HAVE_COPY_FREE_CB
```
dgarske: Not critical, but if you push again...

```
[check-source-text] [4 of 52] [554db10759]
    configure...   real 0m7.365s  user 0m4.797s  sys 0m3.326s
    [skipping "codespell" subtest in check-source-text -- codespell is missing or too old]
overlong lines added:
wolfcrypt/src/cryptocb.c:2047  * algo: Algorithm type (enum wc_AlgoType) - WC_ALGO_TYPE_HASH, WC_ALGO_TYPE_CIPHER, etc
wolfcrypt/src/cryptocb.c:2048  * type: Specific type - for HASH: enum wc_HashType, for CIPHER: enum wc_CipherType
wolfcrypt/src/cryptocb.c:2077  * algo: Algorithm type (enum wc_AlgoType) - WC_ALGO_TYPE_HASH, WC_ALGO_TYPE_CIPHER, etc
wolfcrypt/src/cryptocb.c:2078  * type: Specific type - for HASH: enum wc_HashType, for CIPHER: enum wc_CipherType
wolfssl/wolfcrypt/cryptocb.h:474         int algo; /* enum wc_AlgoType - WC_ALGO_TYPE_HASH, WC_ALGO_TYPE_CIPHER, etc */
wolfssl/wolfcrypt/cryptocb.h:475         int type; /* For HASH: enum wc_HashType, For CIPHER: enum wc_CipherType */
wolfssl/wolfcrypt/cryptocb.h:480         int algo; /* enum wc_AlgoType - WC_ALGO_TYPE_HASH, WC_ALGO_TYPE_CIPHER, etc */
wolfssl/wolfcrypt/cryptocb.h:481         int type; /* For HASH: enum wc_HashType, For CIPHER: enum wc_CipherType */
```

tests/api.c
```diff
@@ -44805,6 +44805,77 @@ static int test_CryptoCb_Func(int thisDevId, wc_CryptoInfo* info, void* ctx)
         }
     #endif /* HAVE_ED25519 */
     }
+#ifdef WOLFSSL_HAVE_COPY_FREE_CB
```
dgarske: Maybe a better macro name? `WOLF_CRYPTO_CB_COPY` and `WOLF_CRYPTO_CB_FREE`. We already have:

```
WOLF_CRYPTO_CB_CMD
WOLF_CRYPTO_CB_FIND
WOLF_CRYPTO_CB_ONLY_ECC
WOLF_CRYPTO_CB_ONLY_RSA
```

dgarske: Could this test be in wolfcrypt/test/test.c so its available on embedded systems as well? This tests/api.c is not portable or usable on embedded targets.

night1rider: Will do

night1rider: I think this is already in `test.c` in an earlier commit? @dgarske

night1rider:
Had to rerun: [OpenWrt test / Compile container (21.02.7)]

night1rider:
Jenkins Please Retest This

PR #9311, opened 2025-10-16
Regression testing

SparkiDev:
# Description

x509.c: realloc may fail and therefore need to store result in a temporary so the old pointer is not lost.

tls.c: free the name if it is not pushed on to the stack of peer CA names. Failure to push can be from memory allocation failure.

aes.c: Don't compile XTS decrypt functions without HAVE_AES_DECRYPT.

Fix tests to have better pre-processor protection.

# Testing

Regression testing.

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/src/tls.c
+++ b/src/tls.c
@@ -7437,8 +7437,10 @@ static int TLSX_CA_Names_Parse(WOLFSSL *ssl, const byte* input,
 
         if (ret == 0) {
             CopyDecodedName(name, cert, ASN_SUBJECT);
-            if (wolfSSL_sk_X509_NAME_push(ssl->peer_ca_names, name) <= 0)
+            if (wolfSSL_sk_X509_NAME_push(ssl->peer_ca_names, name) <= 0) {
+                wolfSSL_X509_NAME_free(name);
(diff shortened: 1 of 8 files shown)


SparkiDev:
retest this please

. /host/pkgs/sdk-arm_24.60372.0/environment-setup-cortexa8hf-neon-poky-linux-gnueabi
Running configure script...
configure: error: unrecognized options: --with-libtool-sysroot
CONFIG RESULT = 1
Configuration Failed
[1;31mError: OE2: without PAA test failed[0m
[twice]

SparkiDev:
> 
>     * `check-all-return-codes` [snippet](https://github.com/wolfSSL/wolfssl/blob/c111c5baccecd9c17a348a4543074be9debcfe16/tests/api/test_tls_ext.c#L280): Determine SetDH’s return type; if it returns an error/status code, add an assertion or error-handling check after the call (e.g., AssertIntEQ(SetDH(ssl_srv), 0);).
> 
> 

SetDH has void return

dgarske (approved):

PR #9312, opened 2025-10-16
Refactor wc_rng_new to use wc_rng_new_ex, and to use WC_USE_DEVID as the devId if set at compile time

night1rider:
Found that when doing our RNG test it would fail to use my CryptoCallback. Noticed the test would use `wc_rng_new` and not `wc_rng_new_ex` and not supply my default devId which is set with `WC_USE_DEVID`.

Change is to assume that if `WC_USE_DEVID` is set that the user intends this to be the default device to use for crypto calls, so this means that if `wc_rng_new` is called it should be assumed the user intends it to use `WC_USE_DEVID` and not `INVALID_DEVID` which is our default solution. `INVALID_DEVID` can still be request if used with `wc_rng_new_ex` by the user

### DIFF
--- a/wolfcrypt/test/test.c
+++ b/wolfcrypt/test/test.c
@@ -9965,11 +9965,11 @@ EVP_TEST_END:
     #ifdef HAVE_AES_DECRYPT
     XMEMSET(dec, 0, sizeof(Aes));
     #endif
-    ret = wc_AesInit(enc, HEAP_HINT, INVALID_DEVID);
+    ret = wc_AesInit(enc, HEAP_HINT, devId);
     if (ret != 0)
         ERROR_OUT(WC_TEST_RET_ENC_EC(ret), out);
     #ifdef HAVE_AES_DECRYPT
-    ret = wc_AesInit(dec, HEAP_HINT, INVALID_DEVID);
+    ret = wc_AesInit(dec, HEAP_HINT, devId);
     if (ret != 0)
         ERROR_OUT(WC_TEST_RET_ENC_EC(ret), out);
     #endif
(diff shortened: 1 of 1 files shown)


night1rider:
@bigbrett Updated the 4 tests that where initializing RNG without the global devId. I see a bunch of other tests that use `INVALID_DEVID` for their specific _ex versions, not sure it makes sense to update those in this PR.

I do see this failing, unsure if its due to be running on an embedded target do you see a similar failure?
```
MLKEM    test passed!
DILITHIUM test passed!
XMSS     test failed!
                      error L=47668 i=0
 [fiducial line numbers: 9751 28958 48739 61407]

```

bigbrett:
@night1rider yes I'm on an embedded target. The tests wouldn't fail unless you were building with `NO_FILESYSTEM` and `NO_DEV_RANDOM`, meaning you don't have a default random entropy source as your cryptoCb is the only way to get entroyp/randomness

bigbrett:
@night1rider brings up a good point...

> @bigbrett Updated the 4 tests that where initializing RNG without the global devId. I see a bunch of other tests that use INVALID_DEVID for their specific _ex versions, not sure it makes sense to update those in this PR.

@dgarske larger question - do you think it is **ever** appropriate for our unit tests to manually specify `INVALID_DEVID` over using the global `devId`? 

IMO if user sets `WC_USE_DEVID` then this should always be respected regardless of the context (even though it perhaps only results in an issue for RNG or RSA/ECC in the `CRYPTOCB_ONLY` mode

wolfcrypt/src/random.c
```diff
@@ -1856,15 +1856,18 @@ static int _InitRng(WC_RNG* rng, byte* nonce, word32 nonceSz,
...
-        if (error) {
-            XFREE(rng, heap, DYNAMIC_TYPE_RNG);
-            rng = NULL;
-        }
+    /* Assume if WC_USE_DEVID it is intended for default usage */
+#ifdef WC_USE_DEVID
+    ret = wc_rng_new_ex(&rng, nonce, nonceSz, heap, WC_USE_DEVID);
+#else
+    ret = wc_rng_new_ex(&rng, nonce, nonceSz, heap, INVALID_DEVID);
+#endif
```
bigbrett, line 1867: can we not use `devId` here, which should be set to `INVALID_DEVID` if `WC_USE_DEVID` isn't defined, and set to the value of `WC_USE_DEVID` otherwise?

bigbrett (dismissed):

wolfcrypt/src/random.c
night1rider: @bigbrett I'm a bit confused about how we'd use `rng->devId` in the current context. In `wc_rng_new()`, we need to pass a `devId` value to `wc_rng_new_ex()` to initialize the RNG structure. Technically, `devId` is not set until the rng ctx returns from `wc_rng_new_ex().` Am I misunderstanding your comment?

bigbrett: ooops sorry brainfart I was thinking this was `test.c`

bigbrett (approved):

night1rider:
Rebased to use fixed kernel test in github actions

PR #9315, opened 2025-10-16
AES: Improve CFB and OFB and add tests

SparkiDev:
# Description

Improve performance of CFB and OFB.
Only have one implementation that is used by OFB encrypt and decrypt.

Update AES testing in unit.test.

Update benchmarking of CFB and OFb to include decrypt.

# Testing

Regression tested AES.

# Checklist

 - [x] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/wolfcrypt/src/aes.c
+++ b/wolfcrypt/src/aes.c
@@ -2903,10 +2903,6 @@ static WARN_UNUSED_RESULT int wc_AesEncrypt(
 #endif
     word32 r;
 
-    if (aes == NULL) {
-        return BAD_FUNC_ARG;
-    }
-
 #ifdef WC_DEBUG_CIPHER_LIFECYCLE
     {
         int ret = wc_debug_CipherLifecycleCheck(aes->CipherLifecycleTag, 0);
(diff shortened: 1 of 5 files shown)


SparkiDev:
Small stack respected in new test_aes.c code.

wolfcrypt/src/aes.c
```diff
@@ -4851,7 +4843,7 @@ static void AesSetKey_C(Aes* aes, const byte* key, word32 keySz, int dir)
     int wc_AesSetKey(Aes* aes, const byte* userKey, word32 keylen,
             const byte* iv, int dir)
     {
-        if (aes == NULL) {
+        if ((aes == NULL) || (userKey == NULL)) {
```
JacobBarthelmeh: I think the check on userKey == NULL might cause issues when using AES struct with hardware keys.

wolfcrypt/src/aes.c
```diff
@@ -12717,7 +12749,7 @@ int wc_AesCfb8Decrypt(Aes* aes, byte* out, const byte* in, word32 sz)
 /* Software AES - CFB Encrypt */
```
JacobBarthelmeh: Outside of your changes, but I see this comment should be AES - OFB not CFB

SparkiDev:
Removed check and removed tests for NULL.
Rebased.

wolfcrypt/src/aes.c
SparkiDev: Removed check and tests.

wolfcrypt/src/aes.c

PR #9321, opened 2025-10-20
Prevent a conversion warning

anhu:
`word16` and `unsigned char:1`

### DIFF
--- a/src/ssl_sess.c
+++ b/src/ssl_sess.c
@@ -1548,7 +1548,7 @@ int wolfSSL_SetSession(WOLFSSL* ssl, WOLFSSL_SESSION* session)
 #endif
     }
     ssl->options.resuming = 1;
-    ssl->options.haveEMS = ssl->session->haveEMS;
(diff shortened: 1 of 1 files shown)


src/ssl_sess.c
```diff
@@ -1548,7 +1548,10 @@ int wolfSSL_SetSession(WOLFSSL* ssl, WOLFSSL_SESSION* session)
 #endif
     }
     ssl->options.resuming = 1;
-    ssl->options.haveEMS = ssl->session->haveEMS;
+    if (ssl->session->haveEMS)
```
dgarske: Also can be written: `ssl->options.haveEMS = (ssl->session->haveEMS) ? 1 : 0;`. Let me know if you want to fix it or keep it as-is.

dgarske (dismissed):

src/ssl_sess.c
anhu: Good call. I like yours better.  I will make the change.

dgarske (approved):

JacobBarthelmeh (approved):

PR #9322, opened 2025-10-20
X.509 cert: crl distribution point reasons is IMPLICIT

SparkiDev:
# Description

The reasons field is IMPLICIT meaning that the value is directly under the context-specific tag. That is context-specific tag is not constructed.

Fixes zd#20681

# Testing

Used customers example certificate.

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


SparkiDev:
retest this please

fipsv2: linuxv5 unit.test failed

PR #9328, opened 2025-10-21
Rust wrapper: add wolfssl::wolfcrypt::hmac module

holtrop:
# Description

Rust wrapper: add wolfssl::wolfcrypt::hmac module

# Testing

How did you test?

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/wrapper/rust/wolfssl/src/wolfcrypt/hmac.rs
+++ b/wrapper/rust/wolfssl/src/wolfcrypt/hmac.rs
@@ -205,12 +205,12 @@ impl HMAC {
     pub fn finalize(&mut self, hash: &mut [u8]) -> Result<(), i32> {
         // Check the output buffer size since wc_HmacFinal() does not accept
         // a length parameter.
-        let typ = self.wc_hmac.macType as u32 as i32;
+        let typ = self.wc_hmac.macType as i32;
         let rc = unsafe { ws::wc_HmacSizeByType(typ) };
         if rc < 0 {
             return Err(rc);
         }
-        let expected_size = rc as u32 as usize;
+        let expected_size = rc as usize;
         if hash.len() != expected_size {
             return Err(ws::wolfCrypt_ErrorCodes_BUFFER_E);
         }


dgarske (dismissed):

wrapper/rust/wolfssl/src/wolfcrypt/hmac.rs
```diff
@@ -0,0 +1,426 @@
...
+    /// ```rust
+    /// use wolfssl::wolfcrypt::hmac::HMAC;
+    /// let key = [0x42u8; 16];
+    /// let mut hmac = HMAC::new(HMAC::TYPE_SHA256, &key).expect("Error with new()");
+    /// hmac.update(b"input").expect("Error with update()");
+    /// let hash_size = hmac.get_hmac_size().expect("Error with get_hmac_size()");
+    /// let mut hash = vec![0u8; hash_size];
+    /// hmac.finalize(&mut hash).expect("Error with finalize()");
+    /// ```
+    pub fn get_hmac_size(&mut self) -> Result<usize, i32> {
```
gasbytes: This method only reads from self from what I can tell, doesn't change it or write to it. That `&mut self` can just become `&self` instead, the mutable is not needed.

wrapper/rust/wolfssl/src/wolfcrypt/hmac.rs
```diff
@@ -0,0 +1,426 @@
...
+    /// The Rust Drop trait guarantees that this method is called when the
+    /// HMAC struct instance goes out of scope, automatically cleaning up
+    /// resources and preventing memory leaks.
+    fn drop(&mut self) {
+        unsafe { ws::wc_HmacFree(&mut self.wc_hmac); }
+    }
+}
+
+/// HMAC Key Derivation Function (HKDF) functionality.
+pub struct HKDF {
```
gasbytes: Since this struct is empty, you could maket his a module, so that the empty struct is no longer needed.

Something like this:

`pub mod hkdf { pub fn extract(...) etc...}`

I understand if you wanna keep the struct + impl scheme for consistency.

wrapper/rust/wolfssl/src/wolfcrypt/hmac.rs
```diff
@@ -0,0 +1,426 @@
...
+    /// let mut hmac = HMAC::new(HMAC::TYPE_SHA256, &key).expect("Error with new()");
+    /// hmac.update(b"input").expect("Error with update()");
+    /// let hash_size = hmac.get_hmac_size().expect("Error with get_hmac_size()");
+    /// let mut hash = vec![0u8; hash_size];
+    /// hmac.finalize(&mut hash).expect("Error with finalize()");
+    /// ```
+    pub fn finalize(&mut self, hash: &mut [u8]) -> Result<(), i32> {
+        // Check the output buffer size since wc_HmacFinal() does not accept
+        // a length parameter.
+        let typ = self.wc_hmac.macType as u32 as i32;
```
gasbytes: Can this cast be simplified like this?:

`let typ = self.wc_hmac.macType as i32;`

wrapper/rust/wolfssl/src/wolfcrypt/hmac.rs
```diff
@@ -0,0 +1,426 @@
...
+        let rc = unsafe { ws::wc_HmacSizeByType(typ) };
+        if rc < 0 {
+            return Err(rc);
+        }
+        let expected_size = rc as u32 as usize;
```
gasbytes: Can this cast be simplified like this?:

`let expected_size = rc as usize;`

gasbytes (changes requested):

wrapper/rust/wolfssl/src/wolfcrypt/hmac.rs
holtrop: You are correct. I've removed 'mut' as it was not needed here. Thank you.

wrapper/rust/wolfssl/src/wolfcrypt/hmac.rs
holtrop: Yeah that seems like a good idea. I did like the HKDF::foo naming for consistency with other structs, but to be honest it was more about the uppercase name rather than actually having a struct + impl. But I think it makes sense to move it to a separate hkdf module so I did that. I moved it to a separate file as well to stay clean.

wrapper/rust/wolfssl/src/wolfcrypt/hmac.rs
holtrop: Yes the double cast was not needed here because it looks like macType is a u8 so it will be zero extended before sign conversion to i32. Removed!

wrapper/rust/wolfssl/src/wolfcrypt/hmac.rs
holtrop: Here we do actually have a signed value first that we don't want to sign extend before conversion, however since I've already checked that rc < 0 I guess we won't be sign extending either, so I removed this double cast as well.

holtrop:
retest this please (PRB-fips-repo-and-harness-test-v3-part2 script returned exit code 255)

holtrop:
retest this please (PRB-fips-repo-and-harness-test-v3-part2 script returned exit code 255)

dgarske (approved):

JacobBarthelmeh (approved):

PR #9330, opened 2025-10-22
Dtls13: Fix handshake hangs on WANT_WRITE I/O error

rizlik:
# Description

The internal buffer index pointer wasn't incremented in case of error inside `_Dtls13HandshakeRecv`, causing an hangs. 

Fixes zd#20686

Added tests to increase coverage from non-blocking I/O.

### DIFF
--- a/tests/api/test_dtls.c
+++ b/tests/api/test_dtls.c
@@ -568,6 +568,12 @@ int test_dtls13_basic_connection_id(void)
     return EXPECT_RESULT();
 }
 
+/** Test DTLS 1.3 behavior when server hits WANT_WRITE during HRR
+ * The test sets up a DTLS 1.3 connection where the server is forced to
+ * return WANT_WRITE when sending the HelloRetryRequest. After the handshake,
+ * application data is exchanged in both directions to verify the connection
+ * works as expected.
+ */
 int test_dtls13_hrr_want_write(void)
 {
     EXPECT_DECLS;
@@ -636,6 +642,15 @@ static int test_dtls13_want_write_send_cb(WOLFSSL *ssl, char *data, int sz, void
     return test_memio_write_cb(ssl, data, sz, wwctx->text_ctx);
 }
 #endif
+/** Test DTLS 1.3 behavior when every other write returns WANT_WRITE
+ * The test sets up a DTLS 1.3 connection where both client and server
+ * alternate between WANT_WRITE and successful writes. After the handshake,
+ * application data is exchanged in both directions to verify the connection
+ * works as expected.
+ *
+ * Data exchanged after the handshake is also tested with simulated WANT_WRITE
+ * conditions to ensure the connection remains functional.
+ */
 int test_dtls13_every_write_want_write(void)
 {
     EXPECT_DECLS;


rizlik:
retest this please

src/dtls13.c
julek-wolfssl, line 1: Don't you think it would be easier to just pass in `word32* inOutIdx` to `_Dtls13HandshakeRecv`? This way we have to remember to set `processedSize` in all branches.

tests/api/test_dtls.c
```diff
@@ -568,6 +568,150 @@ int test_dtls13_basic_connection_id(void)
...
+
+    /* Resume the DTLS 1.3 handshake */
+    ExpectIntEQ(test_memio_do_handshake(ssl_c, ssl_s, 10, NULL), 0);
+
+    /* Verify post-handshake application data in both directions */
+    ExpectIntEQ(wolfSSL_write(ssl_c, msg, msgLen), msgLen);
+    ExpectIntEQ(wolfSSL_read(ssl_s, readBuf, sizeof(readBuf)), msgLen);
+    ExpectStrEQ(readBuf, msg);
+
+    XMEMSET(readBuf, 0, sizeof(readBuf));
```
julek-wolfssl, line 617: You should `XMEMSET` before both uses.

tests/api/test_dtls.c
```diff
@@ -568,6 +568,150 @@ int test_dtls13_basic_connection_id(void)
...
+                hs_complete[side] = 1;
+                continue;
+            }
+            else {
+                ExpectIntEQ(ret, -1);
+                ExpectIntEQ(wolfSSL_get_error(ssl, ret),
+                    WOLFSSL_ERROR_WANT_READ);
+            }
+        }
+    }
```
julek-wolfssl: Do you think `test_memio_do_handshake` could be re-used if you used callbacks that would return `WANT_WRITE` every second call?

rizlik: probably, but wouldn't that make the test harder to read?

src/dtls13.c
rizlik: In general I prefer the index processing outside of the logic function, even if  I've to admit that Dtls13HandshakeRecv is a very thin wrapper that could be consolidated into _Dtls13HandshakeRecv. _Dtl13HandshakeRecv is just invoked here. I'm not sure this is the right PR though. Let me know if you want this change to be included in this PR  or not.

rizlik:
retest this please 
```SSL_accept error -501, can't match cipher suite```

src/dtls13.c
julek-wolfssl: Let's leave it for now.

tests/api/test_dtls.c
julek-wolfssl: I'm just thinking about the handshake part. Since this is mostly a copy of `test_memio_do_handshake`, I think it would make it easier to maintain if you would reuse `test_memio_do_handshake` with callbacks.

Like in `test_wolfSSL_dtls_compare_stateless_read_cb_once` but you could directly call `test_memio_write_cb` and alternate between `WANT_WRITE` and the real cb.

tests/api/test_dtls.c
julek-wolfssl, line 1: Please write short descriptions above each test describing the scenario being tested.

julek-wolfssl (changes requested):

SparkiDev (approved):

PR #9343, opened 2025-10-24
Fixed issue with SiLibs AES Direct (required by DTLS v1.3)

dgarske:
# Description

Fixed issue with SiLibs AES Direct (required by DTLS v1.3). 

Fixes ZD 20695

# Testing

Tested on actual ERF32FG25 hardware with customer supplied user_settings.h.

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/wolfcrypt/benchmark/benchmark.c
+++ b/wolfcrypt/benchmark/benchmark.c
@@ -15752,13 +15752,13 @@ void bench_sphincsKeySign(byte level, byte optim)
     double current_time(int reset)
     {
         portTickType tickCount = xTaskGetTickCount();
+        (void)reset;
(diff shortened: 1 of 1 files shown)


wolfcrypt/benchmark/benchmark.c
```diff
@@ -15758,6 +15758,7 @@ void bench_sphincsKeySign(byte level, byte optim)
     #else
         return (double)tickCount / 1000;
     #endif
+        (void)reset;
```
JacobBarthelmeh: This seems off.

(void)reset is happening after the return from the function.

JacobBarthelmeh (changes requested):

PR #9352, opened 2025-10-27
Rust wrapper: add wolfssl::wolfcrypt::cmac module

holtrop:
# Description

Rust wrapper: add wolfssl::wolfcrypt::cmac module

# Testing

Unit tests.

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/wrapper/rust/wolfssl/src/wolfcrypt/cmac.rs
+++ b/wrapper/rust/wolfssl/src/wolfcrypt/cmac.rs
@@ -203,6 +203,9 @@ impl CMAC {
 
     /// Generate the final Cipher-based Message Authentication Code result.
     ///
+    /// This function consumes the `CMAC` object since no further operations
+    /// can be performed with it.
+    ///
     /// # Parameters
     ///
     /// * `dout`: Output buffer where CMAC is written.
@@ -229,7 +232,7 @@ impl CMAC {
     /// let mut finalize_out = [0u8; 16];
     /// cmac.finalize(&mut finalize_out).expect("Error with finalize()");
     /// ```
-    pub fn finalize(&mut self, dout: &mut [u8]) -> Result<(), i32> {
+    pub fn finalize(mut self, dout: &mut [u8]) -> Result<(), i32> {
         let mut dout_size = dout.len() as u32;
         let rc = unsafe {
             ws::wc_CmacFinalNoFree(&mut self.ws_cmac,


wrapper/rust/wolfssl/src/wolfcrypt/cmac.rs
```diff
@@ -0,0 +1,249 @@
...
+    /// # Parameters
+    ///
+    /// * `key`: Key to use for CMAC generation.
+    /// * `data`: CMAC input data.
+    /// * `check`: CMAC value to compare to.
+    ///
+    /// # Returns
+    ///
+    /// Returns either Ok(valid) (with valid indicating if the CMAC passed in
+    /// is correct or not) on succe or Err(e) containing the wolfSSL library
```
gasbytes: Typo.

wrapper/rust/wolfssl/src/wolfcrypt/cmac.rs
```diff
@@ -0,0 +1,249 @@
...
+    /// ];
+    /// let message = [
+    ///     0x6bu8, 0xc1, 0xbe, 0xe2, 0x2e, 0x40, 0x9f, 0x96,
+    ///     0xe9, 0x3d, 0x7e, 0x11, 0x73, 0x93, 0x17, 0x2a,
+    /// ];
+    /// let mut cmac = CMAC::new(&key).expect("Error with new()");
+    /// cmac.update(&message).expect("Error with update()");
+    /// ```
+    pub fn update(&mut self, data: &[u8]) -> Result<(), i32> {
+        let data_size = data.len() as u32;
```
gasbytes, line 194: Here a size validation of dout.len() could be added, before passing it to the `wc_CmacUpdate` function to update the hash (I think the correct size is 16 bytes for AES-CMAC).
This can be applied to the finalize function too.

wrapper/rust/wolfssl/src/wolfcrypt/cmac.rs
```diff
@@ -0,0 +1,249 @@
...
+    /// let message = [
+    ///     0x6bu8, 0xc1, 0xbe, 0xe2, 0x2e, 0x40, 0x9f, 0x96,
+    ///     0xe9, 0x3d, 0x7e, 0x11, 0x73, 0x93, 0x17, 0x2a,
+    /// ];
+    /// let mut cmac = CMAC::new(&key).expect("Error with new()");
+    /// cmac.update(&message).expect("Error with update()");
+    /// let mut finalize_out = [0u8; 16];
+    /// cmac.finalize(&mut finalize_out).expect("Error with finalize()");
+    /// ```
+    pub fn finalize(&mut self, dout: &mut [u8]) -> Result<(), i32> {
```
gasbytes: This is handled underneath by the C api, but the user could call finalize first, and then update. This can be made safer by changing `&mut self` to `mut self`, by making it consume `self` instead of a reference.
Basically takes ownership of the cmac object, making the incorrect order calls an error at compile time (e.g. if the user calls finalize or update again after this gets called once).

wrapper/rust/wolfssl/src/wolfcrypt/cmac.rs
holtrop: If you're talking about data.len() and wc_CmacUpdate, it looks like it does accept any size.

If you're talking about dout.len() and wc_CmacFinalNoFree(), it looks like wc_CmacFinalNoFree() does already check *outSz and return an error if it is out of range.

wrapper/rust/wolfssl/src/wolfcrypt/cmac.rs
holtrop: Hmm, I think that makes sense, but it would also be different from other finalize APIs for other structs, some of which do allow re-initialization. I'll think on it.

wrapper/rust/wolfssl/src/wolfcrypt/cmac.rs
holtrop: Thanks, fixed. For some reason with this keyboard sometimes I miss the 's' key...

wrapper/rust/wolfssl/src/wolfcrypt/cmac.rs
holtrop: I think it does make sense to make this change, and I'll add a TODO entry to evaluate the other modules for the same thing later.

wrapper/rust/wolfssl/src/wolfcrypt/cmac.rs
gasbytes: Yes I meant `data.len()`, and I meant to post this under the final function api. Apologies.
This check could catch errors before it gets passed to the C api, if not present and it's not respected it would return an error which needs to be interpreted eventually by the rust layer.
Adding that check prevents this, and stops before calling the C layer at runtime.

It's not necessary, but it's something that might be worth adding if not now later on.

holtrop:
retest this please (fips harness failed)

PR #9360, opened 2025-10-28
Aarch64 ASM: missing push and pop of vector regs

SparkiDev:
# Description

Generated code wasn't pushing and poping vector registers when they were 64-bit.
Generation code fixed and SHA-256 ASM code fixed.

# Testing

On Aarch64 MAC:
./autogen.sh
./configure --disable-shared --enable-all-crypto --enable-armasm=no-crypto
wolfcrypt/benchmark/benchmark -sha256

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


SparkiDev:
Assembly code generated with:
https://github.com/wolfSSL/scripts/pull/512

PR #9363, opened 2025-10-29
Improve TLS 1.3 early data handling.

julek-wolfssl:
Introduce `clientInEarlyData` to only return when in `wolfSSL_read_early_data`. This makes sure that other API don't return `ZERO_RETURN` when not in `wolfSSL_read_early_data`. Chose `APP_DATA_READY` as it won't result in a false positive return from `wolfSSL_read_early_data`.

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -22842,8 +22842,8 @@ default:
                                    exit */
                                 ssl->earlyData = no_early_data;
                                 ssl->options.processReply = doProcessInit;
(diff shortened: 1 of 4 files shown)


julek-wolfssl:
Retest this please. `test_wolfSSL_DH` failed

dgarske (approved):

douzzer (approved):

PR #9367, opened 2025-10-30
wolfDTLS_accept_stateless: Fix handling for early data

julek-wolfssl:
- wolfDTLS_accept_stateless now returns before sending any flights. This allows the user to retrieve early data.
- wolfio.c: Clearing `userSet` allows our callback to change the address while in stateless parsing

### DIFF
--- a/src/ssl.c
+++ b/src/ssl.c
@@ -11012,6 +11012,12 @@ int wolfSSL_DTLS_SetCookieSecret(WOLFSSL* ssl,
             FALL_THROUGH;
 
         case ACCEPT_FIRST_REPLY_DONE :
+            if (ssl->options.returnOnGoodCh) {
+                /* Higher level in stack wants us to return. Simulate a
+                 * WANT_WRITE to accomplish this. */
+                ssl->error = WANT_WRITE;
+                return WOLFSSL_FATAL_ERROR;
+            }
             if ( (ssl->error = SendServerHello(ssl)) != 0) {
             #ifdef WOLFSSL_CHECK_ALERT_ON_ERR
                 ProcessReplyEx(ssl, 1); /* See if an alert was sent. */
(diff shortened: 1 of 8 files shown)


src/wolfio.c
```diff
@@ -638,6 +638,18 @@ static int isDGramSock(int sfd)
     }
 }
 
+void wolfSSL_SetRecvFrom(WOLFSSL* ssl, WolfSSLRecvFrom recvFrom)
```
rizlik, line 641: Can you state the reason of this feature? also add documentation.

src/ssl.c
```diff
@@ -11312,15 +11318,19 @@ int wolfDTLS_accept_stateless(WOLFSSL* ssl)
...
     ssl->options.disableRead = disableRead;
     (void)wolfDTLS_SetChGoodCb(ssl, cb.userCb, cb.userCtx);
     if (ret == WOLFSSL_SUCCESS) {
         WOLFSSL_MSG("should not happen. maybe the user called "
                     "wolfDTLS_accept_stateless instead of wolfSSL_accept");
     }
-    else if (ssl->error == WC_NO_ERR_TRACE(WANT_READ)) {
+    else if (ssl->error == WC_NO_ERR_TRACE(WANT_READ) ||
+             ssl->error == WC_NO_ERR_TRACE(WANT_WRITE)) {
+        ssl->error = 0;
```
rizlik, line 11333: if we get a WANT_READ WANT_WRITE but we are not yet stateful, should we *dont* set ssl->error = 0?

wolfssl/internal.h
```diff
@@ -5091,6 +5091,7 @@ struct Options {
 #endif
     word16            hrrSentKeyShare:1;  /* HRR sent with key share */
 #endif
+    word16            returnOnGoodCh:1;
```
rizlik, line 5094: can't the current GoodCHcallback be used instead of a new var?

rizlik (changes requested):

src/wolfio.c
julek-wolfssl: Added primarily to test `wolfio.c` IO callbacks. I think it would be a good start to move all UDP testing connections to memio.

src/ssl.c
julek-wolfssl: I think its easier to convey that in the return of this function and leave `ssl->error` for fatal errors.

wolfssl/internal.h
julek-wolfssl: No because I don't see how it could be used to return before sending the Server's flight.

src/wolfio.c
julek-wolfssl: Added docs

julek-wolfssl:
retest this please scripts/openssl.test

rizlik (approved):

douzzer (approved):

PR #9368, opened 2025-10-31
Rust wrapper: add wolfssl::wolfcrypt::ed25519 module

holtrop:
# Description

Rust wrapper: add wolfssl::wolfcrypt::ed25519 module

# Testing

Unit tests.

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


holtrop:
retest this please (SSL_write failed)

douzzer (approved):

wrapper/rust/wolfssl/src/wolfcrypt/ed25519.rs
```diff
@@ -0,0 +1,1237 @@
...
+    /// ```rust
+    /// use wolfssl::wolfcrypt::random::RNG;
+    /// use wolfssl::wolfcrypt::ed25519::Ed25519;
+    /// let mut rng = RNG::new().expect("Error creating RNG");
+    /// let mut ed = Ed25519::generate(&mut rng).expect("Error with generate()");
+    /// let mut public = [0u8; Ed25519::PUB_KEY_SIZE];
+    /// ed.export_public(&mut public).expect("Error with export_public()");
+    /// ```
+    pub fn export_public(&self, public: &mut [u8]) -> Result<(), i32> {
+        let mut public_size = public.len() as u32;
```
gasbytes, line 212: Maybe here we could add a check that checks that the public_size is not less than Ed25519::PUB_KEY_SIZE.
So that in case it isn't, we stop at runtime at the wrapper level before calling the C code.

holtrop: The C function already performs that check. If we were to add checks like this to the Rust layer there would be many redundant checks, the size of the Rust wrapper would increase a lot, and it would be extra code to maintain that isn't adding anything that the C implementation isn't already checking.

In cases where a buffer length is not provided to the C functions, I have added checks that I can in the Rust layer. I don't see a reason to duplicate a check that is already done in the C layer though.

gasbytes: I agree, it would increase the code size significantly. LGTM.

PR #9373, opened 2025-11-03
Add missing WOLFSSL_HOSTNAME_VERIFY_ALT_NAME_ONLY guards

### LINKED ISSUE #8923: [Bug]: Possible issue with WOLFSSL_HOSTNAME_VERIFY_ALT_NAME_ONLY
### Contact Details

kcghost@gmail.com

### Version

5.8

### Description

I was digging through the code to check how `wolfSSL_check_domain_name` works internally and noticed that `WOLFSSL_HOSTNAME_VERIFY_ALT_NAME_ONLY` likely has no effect on it. I am guessing that was not intentional.

If I understand correctly `wolfSSL_check_domain_name` sets `ssl->buffers.domainName.buffer` which is checked in `ProcessPeerCerts`. `ProcessPeerCerts` calls `CheckForAltNames` and `MatchDomainName` to check against both SAN and CN. `WOLFSSL_HOSTNAME_VERIFY_ALT_NAME_ONLY` is not involved in this code path at all.

At least without  `OPENSSL_EXTRA` enabled  `WOLFSSL_HOSTNAME_VERIFY_ALT_NAME_ONLY` appears to only affect `CheckHostName`, which is ultimately called by both `wolfSSL_X509_check_host` and `wolfSSL_X509_check_ip_asc`. But neither of those functions are actually used in the normal verification process. Bit of a duplicate code problem.

Also an IP address check ends up checking both SAN and CN in all code paths. As far as I know it would be more correct to skip CN for an IP address check. Or at least that's what OpenSSL's `-verify_ip` does.

### Reproduction steps

_No response_

### Relevant log output

```shell

```

julek-wolfssl:


kcghost:
Wouldn't this print a misleading error message? The remaining error message indicates a check against CN took place, but failed. Rather than checks against CN being disabled and the SAN check failing.

SparkiDev (dismissed):

julek-wolfssl:
Thanks @kcghost, I updated the logs to match what is happening.

douzzer (approved):

PR #9379, opened 2025-11-04
Rust wrapper: add wolfssl::wolfcrypt::ed448 module

holtrop:
# Description

Rust wrapper: add wolfssl::wolfcrypt::ed448 module

# Testing

Unit tests.

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/wrapper/rust/wolfssl/src/wolfcrypt/ed448.rs
+++ b/wrapper/rust/wolfssl/src/wolfcrypt/ed448.rs
@@ -193,9 +193,8 @@ impl Ed448 {
     ///
     /// # Returns
     ///
-    /// Returns either Ok(size) containing the number of bytes written to
-    /// `public` on success or Err(e) containing the wolfSSL library error
-    /// code value.
+    /// Returns either Ok(()) on success or Err(e) containing the wolfSSL
+    /// library error code value.
     ///
     /// # Example
     ///
@@ -228,9 +227,8 @@ impl Ed448 {
     ///
     /// # Returns
     ///
-    /// Returns either Ok(size) containing the number of bytes written to
-    /// `keyout` on success or Err(e) containing the wolfSSL library error
-    /// code value.
+    /// Returns either Ok(()) on success or Err(e) containing the wolfSSL
+    /// library error code value.
     ///
     /// # Example
     ///
@@ -263,9 +261,8 @@ impl Ed448 {
     ///
     /// # Returns
     ///
-    /// Returns either Ok(size) containing the number of bytes written to
-    /// `private` on success or Err(e) containing the wolfSSL library error
-    /// code value.
+    /// Returns either Ok(()) on success or Err(e) containing the wolfSSL
+    /// library error code value.
     ///
     /// # Example
     ///
@@ -1084,7 +1081,7 @@ impl Ed448 {
         Ok(res == 1)
     }
 
-    /// Get the size of an Ed448 key (32 bytes).
+    /// Get the size of an Ed448 key (57 bytes).
     ///
     /// # Returns
     ///
@@ -1109,7 +1106,7 @@ impl Ed448 {
         Ok(rc as usize)
     }
 
-    /// Get the size of a private (including public) Ed448 key (64 bytes).
+    /// Get the size of a private (including public) Ed448 key (114 bytes).
     ///
     /// # Returns
     ///
@@ -1134,7 +1131,7 @@ impl Ed448 {
         Ok(rc as usize)
     }
 
-    /// Get the size of a public Ed448 key (32 bytes).
+    /// Get the size of a public Ed448 key (57 bytes).
     ///
     /// # Returns
     ///
@@ -1159,7 +1156,7 @@ impl Ed448 {
         Ok(rc as usize)
     }
 
-    /// Get the size of a Ed448 signature (64 bytes).
+    /// Get the size of a Ed448 signature (114 bytes).
     ///
     /// # Returns
     ///


wrapper/rust/wolfssl/src/wolfcrypt/ed25519.rs
```diff
@@ -627,7 +627,7 @@ impl Ed25519 {
     /// # Parameters
     ///
     /// * `hash`: Message digest to sign.
-    /// * `context`: Buffer containing context for which hash is being signed.
+    /// * `context`: Optional buffer containing context for which hash is being signed.
```
holtrop, line 630: I realized while doing ed448 that the context parameter to the "ph" variants of sign/verify should really be optional also, so I fixed that both there and here.

wrapper/rust/wolfssl/src/wolfcrypt/ed448.rs
```diff
@@ -0,0 +1,1193 @@
...
+            ws::wc_ed448_verify_msg_final(signature.as_ptr(), signature_size,
+                &mut res, &mut self.ws_key)
+        };
+        if rc != 0 {
+            return Err(rc);
+        }
+        Ok(res == 1)
+    }
+
+    /// Get the size of an Ed448 key (32 bytes).
```
gasbytes: Minor documentation error, ed448 keys have a size of 57 bytes.

wrapper/rust/wolfssl/src/wolfcrypt/ed448.rs
```diff
@@ -0,0 +1,1193 @@
...
+    /// ```
+    pub fn size(&self) -> Result<usize, i32> {
+        let rc = unsafe { ws::wc_ed448_size(&self.ws_key) };
+        if rc < 0 {
+            return Err(rc);
+        }
+        Ok(rc as usize)
+    }
+
+    /// Get the size of a private (including public) Ed448 key (64 bytes).
```
gasbytes: Minor documentation error, private + public in ed448 have a size in total of 114 bytes.

wrapper/rust/wolfssl/src/wolfcrypt/ed448.rs
```diff
@@ -0,0 +1,1193 @@
...
+    /// ```
+    pub fn priv_size(&self) -> Result<usize, i32> {
+        let rc = unsafe { ws::wc_ed448_priv_size(&self.ws_key) };
+        if rc < 0 {
+            return Err(rc);
+        }
+        Ok(rc as usize)
+    }
+
+    /// Get the size of a public Ed448 key (32 bytes).
```
gasbytes: Minor documentation error, ed448 keys have a size of 57 bytes.

wrapper/rust/wolfssl/src/wolfcrypt/ed448.rs
```diff
@@ -0,0 +1,1193 @@
...
+    /// ```
+    pub fn pub_size(&self) -> Result<usize, i32> {
+        let rc = unsafe { ws::wc_ed448_pub_size(&self.ws_key) };
+        if rc < 0 {
+            return Err(rc);
+        }
+        Ok(rc as usize)
+    }
+
+    /// Get the size of a Ed448 signature (64 bytes).
```
gasbytes: Minor documentation error, ed448 signatures have a size of 114 bytes.

wrapper/rust/wolfssl/src/wolfcrypt/ed448.rs
```diff
@@ -0,0 +1,1193 @@
...
+    pub fn export_public(&self, public: &mut [u8]) -> Result<(), i32> {
+        let mut public_size = public.len() as u32;
+        let rc = unsafe {
+            ws::wc_ed448_export_public(&self.ws_key, public.as_mut_ptr(),
+                &mut public_size)
+        };
+        if rc != 0 {
+            return Err(rc);
+        }
+        Ok(())
```
gasbytes, line 218: Shouldn't be this `Ok(size)`? If not, maybe it might be worth to update the documentation.

wrapper/rust/wolfssl/src/wolfcrypt/ed448.rs
```diff
@@ -0,0 +1,1193 @@
...
+    pub fn export_private(&self, keyout: &mut [u8]) -> Result<(), i32> {
+        let mut keyout_size = keyout.len() as u32;
+        let rc = unsafe {
+            ws::wc_ed448_export_private(&self.ws_key, keyout.as_mut_ptr(),
+                &mut keyout_size)
+        };
+        if rc != 0 {
+            return Err(rc);
+        }
+        Ok(())
```
gasbytes, line 252: Same as `export_public`.

wrapper/rust/wolfssl/src/wolfcrypt/ed448.rs
```diff
@@ -0,0 +1,1193 @@
...
+    pub fn export_private_only(&self, private: &mut [u8]) -> Result<(), i32> {
+        let mut private_size = private.len() as u32;
+        let rc = unsafe {
+            ws::wc_ed448_export_private_only(&self.ws_key,
+                private.as_mut_ptr(), &mut private_size)
+        };
+        if rc != 0 {
+            return Err(rc);
+        }
+        Ok(())
```
gasbytes, line 286: Same as `export_public`.

wrapper/rust/wolfssl/src/wolfcrypt/ed448.rs
holtrop: Thank you; I forgot to update these after copying from ed25519. Fixed all the size function comments now.

wrapper/rust/wolfssl/src/wolfcrypt/ed448.rs
holtrop: I originally had the function returning the size until I realized it will always be the same value so then I just changed it to return (). Updated comments, thanks.

gasbytes (approved):

SparkiDev:
Does this code differ much from ed25519.rs?

holtrop:
> Does this code differ much from ed25519.rs?

No it is very similar. Each function calls a different C API function, but the logic is pretty much the same for each. ed25519 has 'ctx' versions of sign_msg and verify_msg which are not present in ed448 because the main sign_msg and verify_msg functions take a context pointer.

SparkiDev (approved):

PR #9380, opened 2025-11-04
Improve domain and IP address matching in certificate verification

### LINKED ISSUE #9351: [Bug]: verifying SAN certs
### Contact Details

stefan.eissing@gmail.com

### Version

5.8.2

### Description

In the `curl` project, I extended our testing with a [certificate carrying '127.0.0.1' as DNS name in altnames]. This should no match a connection to the ip address, but it does in wolfSSL.

Initially, we used `wolfSSL_check_domain_name()` only. But that cannot really differentiate between DNS and SAN, apart from guessing. So, I added an additional check using `wolfSSL_X509_check_ip_asc()` and that also, falsely, verifies the certificate.

While it can be debated if a certificate carrying addresses as DNS alt names is a real attack vector, it seems the implementation could be more strict here. Maybe this can be addressed together with #8923.

Thanks!

### Reproduction steps

_No response_

### Relevant log output

```shell

```

julek-wolfssl:
- Distinguish between domain and IP address checks.
- Update curl action to test with httpd server

### DIFF
--- a/.github/workflows/curl.yml
+++ b/.github/workflows/curl.yml
@@ -52,7 +52,7 @@ jobs:
     - name: Install test dependencies
       run: |
         sudo apt-get update
-        sudo apt-get install nghttp2 libpsl5 libpsl-dev python3-impacket
+        sudo apt-get install nghttp2 libpsl5 libpsl-dev python3-impacket apache2 apache2-dev
 
     - name: Download lib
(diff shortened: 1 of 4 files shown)


julek-wolfssl:
Retest this please timeout

SparkiDev (approved):

Issue #9376, opened 2025-11-04
[Bug]:  Possible Logic Error in TLSX_SignatureAlgorithms_Parse Argument Handling


GSoJC234:
### Contact Details

thkighie1224@postech.ac.kr

### Version

v5.8.2-stable

### Description

Hello wolfSSL team,

I found a logic error in the `TLSX_SignatureAlgorithms_Parse` function.

When a client sends a signature algorithms extension whose length exceeds `WOLFSSL_MAX_SIGALGO`, the following code executes:
```
suites->hashSigAlgoSz = len;
if (suites->hashSigAlgoSz > WOLFSSL_MAX_SIGALGO) {
    WOLFSSL_MSG("TLSX SigAlgo list exceeds max, truncating");
    suites->hashSigAlgoSz = WOLFSSL_MAX_SIGALGO;
}
```

However, the function later passes `len` (instead of `suites->hashSigAlgoSz`) as the third argument to `TLSX_SignatureAlgorithms_MapPss`.
This can lead to incorrect PSS hash algorithm configuration in `TLSX_SignatureAlgorithms_MapPss`.
If a legitimate client sends the signature algorithms extension containing only (rsa_pss_sa_algo, sha512), but a malicious attacker intercepts the message and injects dummy values, appending (rsa_pss_sa_algo, sha256) as the 39th pair, the server may incorrectly enable the `sha256` algorithm in `ssl->pssAlgo`. (possible hash downgrade attack)

```
ssl->pssAlgo = 0;
for (i = 0; i < length; i += 2) {
    if (input[i] == rsa_pss_sa_algo && input[i + 1] <= sha512_mac)
        ssl->pssAlgo |= 1 << input[i + 1];
#ifdef WOLFSSL_TLS13
    if (input[i] == rsa_pss_sa_algo &&
        input[i + 1] >= pss_sha256 && input[i + 1] <= pss_sha512) {
        ssl->pssAlgo |= 1 << input[i + 1];
    }
#endif
}
```

However, I believe this behavior is not actually vulnerable or an exploitable attack vector.
There is only one code that checks `ssl->pssAlgo` in the `SendCertificateVerify` function.
```
if (ssl->hsType == DYNAMIC_TYPE_RSA) {
#ifdef WC_RSA_PSS
    if (IsAtLeastTLSv1_2(ssl) &&
        (ssl->pssAlgo & (1 << ssl->options.hashAlgo))) {
[...]

PR #9385, opened 2025-11-04
Use suites->hashSigAlgoSz when calling TLSX_SignatureAlgorithms_MapPss

anhu:
Fixes https://github.com/wolfSSL/wolfssl/issues/9376


GSoJC234:
The issue is resolved. Thanks for the fix!

PR #9391, opened 2025-11-05
Check for duplicate extensions in client hello when HAVE_TLS_EXTENSIONS is not set - fix #9377

### LINKED ISSUE #9377: [Bug]: RFC 5246 violation in DoClientHello when HAVE_TLS_EXTENSION is undefined
### Contact Details

thkighie1224@postech.ac.kr

### Version

v5.8.2-stable

### Description

Hello wolfSSL team,

I found an RFC 5246 violation in the function `DoClientHello`.

When the wolfSSL build is configured without the `HAVE_TLS_EXTENSIONS ` definition,
the following code is executed instead of calling the `TLSX_Parse` function:

```
                while (totalExtSz) {
                    word16 extId, extSz;

                    if (OPAQUE16_LEN + OPAQUE16_LEN > totalExtSz) {
                        ret = BUFFER_ERROR;
                        goto out;
                    }

                    ato16(&input[i], &extId);
                    i += OPAQUE16_LEN;
                    ato16(&input[i], &extSz);
                    i += OPAQUE16_LEN;

                    if (OPAQUE16_LEN + OPAQUE16_LEN + extSz > totalExtSz) {
                        ret = BUFFER_ERROR;
                        goto out;
                    }

                    if (extId == HELLO_EXT_SIG_ALGO) {
                        word16 hashSigAlgoSz;

                        ato16(&input[i], &hashSigAlgoSz);
                        i += OPAQUE16_LEN;

                        if (OPAQUE16_LEN + hashSigAlgoSz > extSz) {
                            ret = BUFFER_ERROR;
                            goto out;
                        }

                        if (hashSigAlgoSz % 2 != 0) {
                            ret = BUFFER_ERROR;
                            goto out;
                        }

                        ssl->clSuites->hashSigAlgoSz = hashSigAlgoSz;
                        if (ssl->clSuites->hashSigAlgoSz >
                                WOLFSSL_MAX_SIGALGO) {
                            WOLFSSL_MSG("ClientHello SigAlgo list exceeds max, "
                                                                  "truncating");
                            ssl->clSuites->hashSigAlgoSz = WOLFSSL_MAX_SIGALGO;
                        }

                        XMEMCPY(ssl->clSuites->hashSigAlgo, &input[i],
                                                  ssl->clSuites->hashSigAlgoSz);

                        i += hashSigAlgoSz;
                    }
#ifdef HAVE_EXTENDED_MASTER
                    else if (extId == HELLO_EXT_EXTMS)
                        ssl->options.haveEMS = 1;
#endif
                    else
                        i += extSz;

                    totalExtSz -= (word16)(OPAQUE16_LEN + OPAQUE16_LEN) + extSz;
                }
```

Unlike `TLSX_Parse`, which properly checks for duplicate extensions,
the above code does not perform any duplicate-extension validation.

This behavior violates RFC 5246, which explicitly states:
| There MUST NOT be more than one extension of the same type.

Although this issue only occurs when the` HAVE_TLS_EXTENSIONS` definition is undefined, it still represents a meaningful protocol compliance problem.
Accepting duplicate extensions violates RFC 5246 and could lead to unexpected behavior or interoperability issues with standards-compliant clients and servers.
Therefore, it would be advisable to ensure consistent duplicate-extension checks regardless of build configuration.

Thank you for your time and for maintaining such a great project.

Best regards,
Jaehun Lee

### Reproduction steps

1. ./configure --enable-debug --enable-usersettings --disable-examples
2. make install
3. examples/server/server -p 4433 -v 3 -l TLS_RSA_WITH_AES_256_CBC_SHA256
4. executes below code using python3

## user_settings.h
```
#undef HAVE_TLS_EXTENSIONS

#define WC_NO_HARDEN
#define HAVE_ECC
#define HAVE_ALL_CURVES
#define WOLFSSL_TLS12
#define WOLFSSL_ALT_CERT_CHAINS
#define DEBUG_WOLFSSL
#define DEBUG_WOLFSSL_VERBOSE
#define OPENSSL_EXTRA

#define HAVE_AESGCM
#define HAVE_AESCCM
#define BUILD_TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256
#define BUILD_TLS_RSA_WITH_AES_256_CBC_SHA256

#define WC_RSA_PSS
#define WOLFSSL_TLS12
#define WOLFSSL_SHA256
#define WOLFSSL_SHA384
#define WOLFSSL_SHA512
#undef WOLFSSL_MD5
#undef WOLFSSL_MD5_SHA
```

## python script
``` 
import socket
import binascii

def send_hex_to_server():
    hex_data = (
   "16 03 03 00 3f 01 00 00 3b 03 03 f4 9b 8d ef 1b"
   "fb 8c 4a 34 64 41 15 c3 93 de f7 2a 0d aa 70 12"
   "5e 67 c0 f2 f5 2a e4 6d 75 ec 50 00 00 02 00 3d"
   "01 00 00 10 00 0d 00 04 00 02 08 04 00 0d 00 04"
   "00 02 08 04"
    )
    data = binascii.unhexlify(hex_data.replace(" ", ""))
    with socket.create_connection(("127.0.0.1", 4433)) as sock:
        sock.sendall(data)
        try:
            while True:
                response = sock.recv(4096)
                if not response:
                    print("Connection closed")
                    break
                print("Received: (hex):", response.hex())
        except KeyboardInterrupt:
            print("End")

if __name__ == "__main__":
    send_hex_to_server()
```

### Relevant log output

```shell

```

holtrop:
# Description

Check for duplicate extensions in client hello when HAVE_TLS_EXTENSIONS is not set - fix #9377

# Testing

Ran reproduce steps from #9377 and verified the duplicate error is now found.
Ran unit tests for configuration `./configure --disable-sni --disable-ecc --disable-tls13 --disable-secure-renegotiation-info` (prior to this fix the `test_tls_ext_duplicate` test was failing to correctly detect the duplicate extension for this configuration).

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -38364,8 +38364,8 @@ static int DoSessionTicket(WOLFSSL* ssl, const byte* input, word32* inOutIdx,
                     ato16(&input[i], &extSz);
                     i += OPAQUE16_LEN;
 
-                    if (extId < (sizeof(extensions_seen) * 8U)) {
-                        word32 mask = 1u << extId;
+                    if (extId < (word16)(sizeof(extensions_seen) * 8U)) {
+                        word32 mask = 1U << extId;
                         if ((extensions_seen & mask) != 0U) {
                             WOLFSSL_MSG(
                                     "DoClientHello: duplicate extension found");


tests/api.c
```diff
@@ -12993,9 +12903,9 @@ static int test_tls_bad_legacy_version(void)
         0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
         0x00, 0x00, 0x00, 0x00, 0x00, 0x04, 0x13, 0x01,
         0x00, 0x9e, 0x01, 0x00,
-        /* Extensions - duplicate signature algorithms. */
+        /* Extensions */
```
SparkiDev, line 12996: Add a new test.
Don't piggyback a test that should already fail.

src/internal.c
```diff
@@ -38336,6 +38637,11 @@ static int DoSessionTicket(WOLFSSL* ssl, const byte* input, word32* inOutIdx,
                     }
 
                     ato16(&input[i], &extId);
+                    if (checkDupExt(extId, extBlock, &input[i] - extBlock) != 0u) {
```
SparkiDev: 80 character lines max

src/internal.c
```diff
@@ -37846,6 +38115,35 @@ static int DoSessionTicket(WOLFSSL* ssl, const byte* input, word32* inOutIdx,
...
+     * @param extBlock Start of extension block.
+     * @param size Max size to evaluate.
+     *
+     * @return Nonzero if the extension was previously seen.
+     */
+    static int checkDupExt(word16 extId, const byte * extBlock, word32 size)
+    {
+        word32 i = 0u;
+        word16 thisExtId, thisExtSz;
+        while (i < size) {
```
SparkiDev: You're searching all the previous extensions for this id.

More efficient would be to cache the extension ids seen and fail when it appears a second time.
We have an array of bits in the TLSX code. Each bit corresponds to an id. Some have to be mapped down to a smaller value so you can just index the bit array.

We don't even need to check all ids, just the ones we are going to process. In this case it is Signature Algorithm and EXTMS.

tests/api.c
holtrop: I think I need to clarify this:

We already had a test that was failing due to the problem described in issue 9377. The test is `test_tls_ext_duplicate`. This change (which I found while searching for "duplicate") looks like it simply reused the same buffer with a duplicate extension to test something different. I wanted to improve this test by making sure it was only testing that condition and not the duplicate extension condition. It turned out that the test passed without this change, but I think that's because the error it is testing for happens before parsing the duplicate extensions. So this change is not strictly necessary, but I thought it was an improvement to the test.

The reason we weren't seeing the `test_tls_ext_duplicate` failure is that presumably we don't have a CI configuration running the configuration option combination I was testing (`./configure --disable-sni --disable-ecc --disable-tls13 --disable-secure-renegotiation-info`). When I ran that configure locally before even making my changes I was seeing `test_tls_ext_duplicate` fail due to this issue.

So perhaps we should add a CI test configuration with those configure options?

holtrop: I realized I didn't update the comment above, so I will do that.

src/internal.c
holtrop: Wrapped 83 character line.

src/internal.c
holtrop: I see the TLSX implementation you're referring to. But, the TLS spec disallows any duplicate, not just ones we know about or process. I was trying to meet the spec and check for any duplicate, even those we don't know of.

I realize this searches all previous extensions for this ID. (It would have been great if the spec required ordering them so we could just check greater or less than the previous ID or something like that!). The number of /defined/ extensions is only in the dozens, and even then the number present will likely be much smaller than that, so the performance here for that small number of extensions did not seem worthwhile to optimize. Is there a concern beyond that that I am not considering?

SparkiDev: What if an attacker sends a very large ClientHello with lots of extensions that are uniquely id with no data?

Unsupported extensions need to be ignored in case a future extension requires a duplicate. Hopefully not ... but you never know.
OpenSSL ignores duplicate unsupported TLS extensions.

tests/api.c
SparkiDev: OK. Changing this test case to not have duplicate extension.

src/internal.c
holtrop: Ok, I will look into changing to use flag bits.

holtrop: Switched to using flag bits.

src/internal.c
```diff
@@ -38310,6 +38309,8 @@ static int DoSessionTicket(WOLFSSL* ssl, const byte* input, word32* inOutIdx,
                 /* auto populate extensions supported unless user defined */
                 if ((ret = TLSX_PopulateExtensions(ssl, 1)) != 0)
                     goto out;
+#else
+                word64 extensions_seen = 0u;
```
dgarske: Can you avoid using a word64? Many of our customers set `NO_64BIT`. Prefer a 32-bit type of array of word32. Prefer upper case unsigned `0U` although its not required here.

src/internal.c
```diff
@@ -38363,6 +38364,18 @@ static int DoSessionTicket(WOLFSSL* ssl, const byte* input, word32* inOutIdx,
                     ato16(&input[i], &extSz);
                     i += OPAQUE16_LEN;
 
+                    if (extId < (sizeof(extensions_seen) * 8u))
```
dgarske: Please add `./configure --disable-sni --disable-ecc --disable-tls13 --disable-secure-renegotiation-info` test. Just put in `.github/workflows/os-check.yml`

dgarske: Please use `if () {`.

src/internal.c
```diff
@@ -38363,6 +38364,18 @@ static int DoSessionTicket(WOLFSSL* ssl, const byte* input, word32* inOutIdx,
...
+                    {
+                        word64 mask = 1u << extId;
+                        if ((extensions_seen & mask) != 0u)
+                        {
+                            WOLFSSL_MSG("DoClientHello: duplicate extension found");
```
dgarske: Line is over 80 chars.

src/internal.c
dgarske: What if `extId` is >= 64? Are those invalid and ignored? Is that behavior you expect?

src/internal.c
```diff
@@ -38363,6 +38364,18 @@ static int DoSessionTicket(WOLFSSL* ssl, const byte* input, word32* inOutIdx,
                     ato16(&input[i], &extSz);
                     i += OPAQUE16_LEN;
 
+                    if (extId < (sizeof(extensions_seen) * 8u))
+                    {
+                        word64 mask = 1u << extId;
+                        if ((extensions_seen & mask) != 0u)
```
dgarske: Same please: `if () {`

src/internal.c
holtrop: Changed to word32 and uppercase U.

src/internal.c
holtrop: Added github workflow configure test.

holtrop: Moved curly brace.

src/internal.c
holtrop: Wrapped line.

src/internal.c
holtrop: Yes, see @SparkiDev 's earlier comments. Originally I was catching any duplicate extension but it was mentioned that we only need to check for known extension IDs that we process. Currently we only process extensions with ID < 32 in this mode (when HAVE_TLS_EXTENSIONS is not defined!).

src/internal.c
holtrop: Moved curly brace.

dgarske:
Jenkins retest this please

dgarske (approved):

src/internal.c
```diff
@@ -38363,6 +38364,17 @@ static int DoSessionTicket(WOLFSSL* ssl, const byte* input, word32* inOutIdx,
                     ato16(&input[i], &extSz);
                     i += OPAQUE16_LEN;
 
+                    if (extId < (sizeof(extensions_seen) * 8U)) {
+                        word32 mask = 1u << extId;
```
dgarske: Nit: Can you make this '1U`?

src/internal.c
```diff
@@ -38363,6 +38364,17 @@ static int DoSessionTicket(WOLFSSL* ssl, const byte* input, word32* inOutIdx,
                     ato16(&input[i], &extSz);
                     i += OPAQUE16_LEN;
 
+                    if (extId < (sizeof(extensions_seen) * 8U)) {
```
dgarske: Just to handle the explicit cast compilers so can cast this to `word16`?

`if (extId < (word16)(sizeof(extensions_seen) * 8U)) {`

dgarske (changes requested):

src/internal.c
holtrop: Updated.

src/internal.c
holtrop: Added cast.

PR #9397, opened 2025-11-06
wolfssl: preserve early-data handling across WANT_WRITE retries

rizlik:
Server-side accept (TLS 1.3/DTLS 1.3) could lose the early-data shortcut whenever sending the Finished flight first hit WANT_WRITE. The buffered data advanced acceptState past TLS13_ACCEPT_FINISHED_SENT as soon as it flushed, so the next wolfSSL_accept() call skipped the block that marks SERVER_FINISHED_COMPLETE and lets the application drain 0-RTT data. By keeping the FALL_THROUGH into TLS13_ACCEPT_FINISHED_SENT and only returning early while that handshake flag is still unset, we revisit the shortcut immediately after the buffered flight is delivered, preserving the intentional behaviour even under non-blocking I/O.

On the client, the same pattern showed up after SendTls13ClientHello() buffered due to WANT_WRITE: after flushing, the connect state is already CLIENT_HELLO_SENT so the early-data exit is no longer executed. We now fall through into the CLIENT_HELLO_SENT case and only short-circuit once per handshake, ensuring the reply-processing loop still executes on the retry.

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -42514,7 +42514,7 @@ int wolfSSL_TestAppleNativeCertValidation_AppendCA(WOLFSSL_CTX* ctx,
 #endif /* defined(__APPLE__) && defined(WOLFSSL_SYS_CA_CERTS) */
 
 /* Do not try to process error for async, non blocking io, and app_read */
-void wolfssl_i_MaybeCheckAlertOnErr(WOLFSSL* ssl, int err)
+void wolfssl_local_MaybeCheckAlertOnErr(WOLFSSL* ssl, int err)
 {
 #if defined(WOLFSSL_CHECK_ALERT_ON_ERR)
 #if defined(WOLFSSL_ASYNC_CRYPT)
--- a/src/ssl.c
+++ b/src/ssl.c
@@ -10591,7 +10591,7 @@ int wolfSSL_DTLS_SetCookieSecret(WOLFSSL* ssl,
                 #endif
                 if (ssl->options.sendVerify) {
                     if ( (ssl->error = SendCertificate(ssl)) != 0) {
-                        wolfssl_i_MaybeCheckAlertOnErr(ssl, ssl->error);
+                        wolfssl_local_MaybeCheckAlertOnErr(ssl, ssl->error);
                         WOLFSSL_ERROR(ssl->error);
                         return WOLFSSL_FATAL_ERROR;
                     }
@@ -10610,7 +10610,7 @@ int wolfSSL_DTLS_SetCookieSecret(WOLFSSL* ssl,
         #endif
             if (!ssl->options.resuming) {
                 if ( (ssl->error = SendClientKeyExchange(ssl)) != 0) {
-                    wolfssl_i_MaybeCheckAlertOnErr(ssl, ssl->error);
+                    wolfssl_local_MaybeCheckAlertOnErr(ssl, ssl->error);
 #ifdef WOLFSSL_EXTRA_ALERTS
                     if (ssl->error == WC_NO_ERR_TRACE(NO_PEER_KEY) ||
                         ssl->error == WC_NO_ERR_TRACE(PSK_KEY_ERROR)) {
@@ -10639,7 +10639,7 @@ int wolfSSL_DTLS_SetCookieSecret(WOLFSSL* ssl,
             #if !defined(NO_CERTS) && !defined(WOLFSSL_NO_CLIENT_AUTH)
                 if (ssl->options.sendVerify) {
                     if ( (ssl->error = SendCertificateVerify(ssl)) != 0) {
-                        wolfssl_i_MaybeCheckAlertOnErr(ssl, ssl->error);
+                        wolfssl_local_MaybeCheckAlertOnErr(ssl, ssl->error);
                         WOLFSSL_ERROR(ssl->error);
                         return WOLFSSL_FATAL_ERROR;
                     }
@@ -10652,7 +10652,7 @@ int wolfSSL_DTLS_SetCookieSecret(WOLFSSL* ssl,
 
         case FIRST_REPLY_THIRD :
             if ( (ssl->error = SendChangeCipher(ssl)) != 0) {
-                wolfssl_i_MaybeCheckAlertOnErr(ssl, ssl->error);
+                wolfssl_local_MaybeCheckAlertOnErr(ssl, ssl->error);
                 WOLFSSL_ERROR(ssl->error);
                 return WOLFSSL_FATAL_ERROR;
             }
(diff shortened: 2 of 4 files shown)


rizlik:
retest this please

rizlik:
retest this please

rizlik:
retest this please

rizlik:
retest this please

 RequestAbortedException

src/internal.c
```diff
@@ -42506,6 +42495,34 @@ int wolfSSL_TestAppleNativeCertValidation_AppendCA(WOLFSSL_CTX* ctx,
 
 #endif /* defined(__APPLE__) && defined(WOLFSSL_SYS_CA_CERTS) */
 
+#if defined(WOLFSSL_CHECK_ALERT_ON_ERR)
```
dgarske: Looks like this option is only enabled with the compatibility layer or --enable-harden-tls. Should it be enabled in more cases? Does the ssl.c description of this macro need updated?

```
* WOLFSSL_CHECK_ALERT_ON_ERR:
 *     Check for alerts during the handshake in the event of an error.
```

rizlik: I don't see a reason to enable it in more cases.
I also think the description is still valid.

rizlik:
retest this please

tests/api/test_tls13.c
```diff
@@ -2011,164 +2105,207 @@ int test_tls13_early_data(void)
...
+        }
 #endif
 
-            /* Test 0-RTT data */
-            wolfSSL_SetLoggingPrefix("client");
-            ExpectIntEQ(wolfSSL_write_early_data(ssl_c, msg, sizeof(msg),
-                    &written), sizeof(msg));
+        if (everyWriteWantWrite) {
+            XMEMSET(&wwrite_ctx_c, 0, sizeof(wwrite_ctx_c));
+            XMEMSET(&wwrite_ctx_s, 0, sizeof(wwrite_ctx_s));
```
julek-wolfssl: Move to loop entry. Decreases chance of bugs in future.

wolfssl/internal.h
```diff
@@ -7238,6 +7238,10 @@ WOLFSSL_LOCAL int pkcs8_encrypt(WOLFSSL_EVP_PKEY* pkey,
         word32* keySz);
 #endif /* OPENSSL_EXTRA || OPENSSL_EXTRA_X509_SMALL */
 
+#if defined(WOLFSSL_CHECK_ALERT_ON_ERR)
+WOLFSSL_LOCAL void wolfSSL_maybeCheckAlertOnErr(WOLFSSL* ssl, int err);
```
julek-wolfssl:
```suggestion
WOLFSSL_LOCAL void wolfSSL_MaybeCheckAlertOnErr(WOLFSSL* ssl, int err);
```

src/ssl.c
```diff
@@ -10592,7 +10592,7 @@ int wolfSSL_DTLS_SetCookieSecret(WOLFSSL* ssl,
                 if (ssl->options.sendVerify) {
                     if ( (ssl->error = SendCertificate(ssl)) != 0) {
                     #ifdef WOLFSSL_CHECK_ALERT_ON_ERR
-                        ProcessReplyEx(ssl, 1); /* See if an alert was sent. */
+                        wolfSSL_maybeCheckAlertOnErr(ssl, ssl->error);
                     #endif
```
julek-wolfssl: I think it would be cleaner to move `WOLFSSL_CHECK_ALERT_ON_ERR` inside of `wolfSSL_maybeCheckAlertOnErr`. When `WOLFSSL_CHECK_ALERT_ON_ERR` is not defined then `wolfSSL_maybeCheckAlertOnErr` would be a no-op.

src/internal.c
```diff
@@ -42506,6 +42495,34 @@ int wolfSSL_TestAppleNativeCertValidation_AppendCA(WOLFSSL_CTX* ctx,
...
+    if (err == WC_NO_ERR_TRACE(APP_DATA_READY)) {
+        return;
+    }
+#endif
+    if (err == WC_NO_ERR_TRACE(WANT_WRITE) ||
+            err == WC_NO_ERR_TRACE(WANT_READ)) {
+        return;
+    }
+    /* check if an alert was sent */
+    ProcessReplyEx(ssl, 1);
```
julek-wolfssl, line 42540: Should we at least log the error?

src/tls13.c
```diff
@@ -13383,8 +13383,12 @@ int wolfSSL_connect_TLSv13(WOLFSSL* ssl)
 
             ssl->options.connectState = CLIENT_HELLO_SENT;
             WOLFSSL_MSG("TLSv13 connect state: CLIENT_HELLO_SENT");
+            FALL_THROUGH;
+
+        case CLIENT_HELLO_SENT:
     #ifdef WOLFSSL_EARLY_DATA
-            if (ssl->earlyData != no_early_data) {
+            if (ssl->earlyData != no_early_data &&
+                ssl->options.handShakeState != CLIENT_HELLO_COMPLETE) {
```
julek-wolfssl, line 13404: Please fix the indentation inside this `if`. I'm talking about lines before the closing `}`.

julek-wolfssl (changes requested):

src/internal.c
rizlik: not sure, it wil be logged inside `ProcessReplyEx`, I prefer to keep the codepath cleaner.

rizlik:
retest this please:

AgentOfflineException exception:

src/tls13.c
julek-wolfssl: Not addressed @rizlik .

wolfssl/internal.h
```diff
@@ -7239,6 +7239,8 @@ WOLFSSL_LOCAL int pkcs8_encrypt(WOLFSSL_EVP_PKEY* pkey,
         word32* keySz);
 #endif /* OPENSSL_EXTRA || OPENSSL_EXTRA_X509_SMALL */
 
+WOLFSSL_LOCAL void wolfSSL_MaybeCheckAlertOnErr(WOLFSSL* ssl, int err);
```
julek-wolfssl: I would prefer to not use `wolfSSL_` prefix for internal funcs.

julek-wolfssl (changes requested):

rizlik:
retest this please.

AgentOfflineException

rizlik:
retest this please

rizlik:
retest this please

julek-wolfssl (approved):
lgtm

src/internal.c
```diff
@@ -42524,6 +42513,37 @@ int wolfSSL_TestAppleNativeCertValidation_AppendCA(WOLFSSL_CTX* ctx,
 
 #endif /* defined(__APPLE__) && defined(WOLFSSL_SYS_CA_CERTS) */
 
+/* Do not try to process error for async, non blocking io, and app_read */
+void wolfssl_i_MaybeCheckAlertOnErr(WOLFSSL* ssl, int err)
```
dgarske: Is `wolfssl_i_` mean its an "internal" function? Is that a new naming convention? I don't see the value in `_i_` and I would prefer to remove it. The WOLFSSL_LOCAL on the API def tells us its internal.

dgarske (changes requested):

dgarske (approved):

dgarske:
Jenkins retest this please. FIPS 140-3 history lost already

PR #9407, opened 2025-11-10
Rust wrapper: support optional heap and dev_id parameters

holtrop:
# Description

Rust wrapper: support optional heap and dev_id parameters

# Testing

How did you test?

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/wrapper/rust/wolfssl/src/wolfcrypt/cmac.rs
+++ b/wrapper/rust/wolfssl/src/wolfcrypt/cmac.rs
@@ -45,8 +45,6 @@ impl CMAC {
     /// * `key`: Key to use for CMAC generation.
     /// * `data`: CMAC input data.
     /// * `dout`: Output buffer where CMAC is written.
-    /// * `heap`: Optional heap hint.
-    /// * `dev_id` Optional device ID to use with crypto callbacks or async hardware.
     ///
     /// # Returns
     ///


holtrop:
retest this please

wrapper/rust/wolfssl/src/wolfcrypt/aes.rs
```diff
@@ -60,12 +60,17 @@ pub struct CBC {
...
+    ///
+    /// * `heap`: Optional heap hint.
+    /// * `dev_id` Optional device ID to use with crypto callbacks or async hardware.
+    ///
     /// # Returns
     ///
     /// A Result which is Ok(CBC) on success or an Err containing the wolfSSL
     /// library return code on failure.
-    pub fn new() -> Result<Self, i32> {
-        let ws_aes = new_ws_aes()?;
```
holtrop, line 68: I added `heap` and `dev_id` parameters to several functions. I really wish Rust had default parameter values! :(

I like avoiding too many methods, but if `heap` and `dev_id` are rarely used and it would make more sense then I could make `new()` functions that take no arguments and `new_ex()` that take `heap` and `dev_id` arguments. I don't have a feel for what common usage is.

dgarske: With rust I could image either being used, but it would be sure nice to not have to supply NULL and INVALID_DEVID for these. The devId is used for crypto callbacks or any hardware offloading. The heap_hint is used with static memory (or a custom allocator). I'm leaning towards new() and new_ex() personally, but I realize its a big refactor.

holtrop: Since I'm using Option for each parameter the caller would only have to supply `None` to use the default. But still, unless these parameters are used more often than not, it would probably be nicer to separate them. It's not a big deal and now is the time to do it; so I'll create separate `new()` and `new_ex()` now.

holtrop: Split several functions to regular / _ex variants to avoid required heap and dev_id parameters for several cases.

dgarske (approved):

wrapper/rust/wolfssl/src/wolfcrypt/cmac.rs
```diff
@@ -45,6 +45,8 @@ impl CMAC {
     /// * `key`: Key to use for CMAC generation.
     /// * `data`: CMAC input data.
     /// * `dout`: Output buffer where CMAC is written.
+    /// * `heap`: Optional heap hint.
```
gasbytes: Small documentation error, `heap` and `dev_id` are not arguments of the this function, they should appear only in the `*_ex` version of this function.

holtrop: Thanks. Fixed!

gasbytes (approved):

PR #9416, opened 2025-11-12
Fix errors when blinding private keys

julek-wolfssl:
- test_wolfSSL_inject: don't call accept on completed handshake
- Don't override errors when blinding the priv key


SparkiDev:
retest this please

SparkiDev (approved):

dgarske (approved):

PR #9421, opened 2025-11-13
ML-KEM to bytes C: not reducing all values

SparkiDev:
# Description

Call to mlkem_csubq_c was only called on first array. Fixed to do it for all.

Fixes zd#20796

# Testing

Tested that the all values are less than MLKEM_Q before encoding in mlkem_bytes_to_c.

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


SparkiDev:
retest this please

PR #9425, opened 2025-11-13
with decode enveloped data track total encrypted content size

JacobBarthelmeh:
Found when running `cd certs && ./renewcerts.sh`. In the case of multiple octets being decoded the total size was lost and there was a case where wc_PKCS7_DecodeEnvelopedData() could return 0 when the last encrypted content chunk cashed was just the padding values.


JacobBarthelmeh:
Retest this please Jenkins lots of "Agent was removed" notifications.

PR #9439, opened 2025-11-17
ML-KEM: derive secret fix

SparkiDev:
# Description

Fixes for deriving secret for ML-KEM.

# Testing

C and Intel ASM code.

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/wolfcrypt/src/wc_mlkem.c
+++ b/wolfcrypt/src/wc_mlkem.c
@@ -105,6 +105,42 @@
 
 #ifdef WOLFSSL_WC_MLKEM
 
+#ifdef DEBUG_MLKEM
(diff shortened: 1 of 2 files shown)


SparkiDev:
retest this please

wolfcrypt/src/wc_mlkem.c
```diff
@@ -105,6 +105,42 @@
 
 #ifdef WOLFSSL_WC_MLKEM
 
+#ifdef DEBUG_MLKEM
```
dgarske, line 108: These don't appear to be called from anywhere, but I'm okay having them here if you need it for future debugging because you have it wrapped in a build option. I am surprised you didn't need to add this to the .wolfssl_known_macro_extra but its not required.

PR #9443, opened 2025-11-18
TLSv1.3 certificate verify: report rsa_pss_pss_* signature algorithm when supported

holtrop:
# Description

Report rsa_pss_pss_* signature algorithm instead of rsa_pss_rsae_* when both sides support it and the key/cert in use is RSA-PSS.

Fixes ZD#20812

# Testing

Captured TLSv1.3 traffic with Wireshark and verified that the Certificate Verify message now reports rsa_pss_pss_sha256 when it previously reported rsa_pss_rsae_sha256.

# Checklist

 - [X] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/src/ssl_load.c
+++ b/src/ssl_load.c
@@ -1625,10 +1625,12 @@ static int ProcessBufferCertPublicKey(WOLFSSL_CTX* ctx, WOLFSSL* ssl,
                     ctx->minRsaKeySz, RSA_MAX_SIZE / 8, keySz, RSA_KEY_SIZE_E);
             }
     #ifdef WC_RSA_PSS
-            if (ssl)
-                ssl->ctx->useRsaPss = cert->keyOID == RSAPSSk;
-            else
+            if (ssl) {
+                ssl->useRsaPss = cert->keyOID == RSAPSSk;
+            }
+            if (ctx) {
                 ctx->useRsaPss = cert->keyOID == RSAPSSk;
+            }
     #endif
             break;
 #endif /* !NO_RSA */
--- a/src/internal.c
+++ b/src/internal.c
@@ -6983,6 +6983,9 @@ int SetSSL_CTX(WOLFSSL* ssl, WOLFSSL_CTX* ctx, int writeDup)
 #endif
 #ifndef NO_RSA
     ssl->options.minRsaKeySz = ctx->minRsaKeySz;
+  #ifdef WC_RSA_PSS
+    ssl->useRsaPss = ctx->useRsaPss;
+  #endif
 #endif
 #ifdef HAVE_ECC
     ssl->options.minEccKeySz = ctx->minEccKeySz;
--- a/src/tls13.c
+++ b/src/tls13.c
@@ -7907,7 +7907,7 @@ static WC_INLINE void EncodeSigAlg(const WOLFSSL * ssl, byte hashAlgo, byte hsTy
             /* If the private key uses the RSA-PSS OID, and the peer supports
              * the rsa_pss_pss_* signature algorithm in use, then report
              * rsa_pss_pss_* rather than rsa_pss_rsae_*. */
-            if (ssl->ctx->useRsaPss &&
+            if (ssl->useRsaPss &&
                 ((ssl->pssAlgo & (1U << hashAlgo)) != 0U) &&
                 (sha256_mac <= hashAlgo) && (hashAlgo <= sha512_mac))
             {
--- a/wolfssl/internal.h
+++ b/wolfssl/internal.h
@@ -3908,7 +3908,7 @@ struct WOLFSSL_CTX {
 #ifndef NO_RSA
     short       minRsaKeySz;      /* minimum RSA key size */
 #ifdef WC_RSA_PSS
-    word8       useRsaPss;
+    word8       useRsaPss;        /* cert supports RSA-PSS */
 #endif
 #endif
 #if defined(HAVE_ECC) || defined(HAVE_ED25519) || defined(HAVE_ED448)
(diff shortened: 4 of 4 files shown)


src/ssl_load.c
```diff
@@ -1624,6 +1624,12 @@ static int ProcessBufferCertPublicKey(WOLFSSL_CTX* ctx, WOLFSSL* ssl,
                 ret = CHECK_KEY_SZ(ssl ? ssl->options.minRsaKeySz :
                     ctx->minRsaKeySz, RSA_MAX_SIZE / 8, keySz, RSA_KEY_SIZE_E);
             }
+    #ifdef WC_RSA_PSS
+            if (ssl)
+                ssl->ctx->useRsaPss = (cert->keyOID == RSAPSSk) ? 1U : 0U;
```
SparkiDev: Both are booleans ... just assign.

src/ssl_load.c
```diff
@@ -1624,6 +1624,12 @@ static int ProcessBufferCertPublicKey(WOLFSSL_CTX* ctx, WOLFSSL* ssl,
...
+            else
+                ctx->useRsaPss = (cert->keyOID == RSAPSSk) ? 1U : 0U;
```
SparkiDev: Just assign.

src/ssl_load.c
SparkiDev: Add the field to WOLFSSL.
Don't modify the WOLFSSL_CTX when the certificate is only for the SSL.

holtrop: Everywhere that uses `ssl` in this function checks if it is non-NULL first and it looked to me that this function could be called with `ssl` NULL and only `ctx` set. I was trying to support all calls to this, but maybe I only need to support the calls where `ssl` is not NULL to set this flag?

holtrop: If I move useRsaPss flag to struct WOLFSSL then the issue is broken again and rss_pss_rsae_sha256 is reported instead of rss_pss_pss_sha256.

holtrop: Removed conditional and just assign boolean result.

src/ssl_load.c
holtrop: Removed conditional and just assign boolean result.

src/ssl_load.c
SparkiDev: If the certificate with RSA-PSS OID goes against 'ssl' and not 'ctx', then updating 'ctx' is wrong.
Did I misunderstand the code?

Are you checking the 'ssl' field after adding and are you copying the field from 'ctx' on creation of the WOLFSSL object.

holtrop:
retest this please (RequestAbortedException)

src/ssl_load.c
holtrop: I am not sure what you mean by the certificate "goes against" 'ssl' vs 'ctx'.

What I tried yesterday was (1) moving my useRsaPss flag from struct WOLFSSL_CTX to struct WOLFSSL, (2) removing the else clause from this if statement, and (3) changing "ssl->ctx->" to just "ssl->", and (4) changing "ssl->ctx->useRsaPss" to "ssl->useRsaPss" in my tls13.c change. This did not work. It looks like 'ssl' is null here when this cert is loaded so I had to set the flag in the context instead.

> Are you checking the 'ssl' field after adding and are you copying the field from 'ctx' on creation of the WOLFSSL object.

I don't know what you mean by checking the 'ssl' field after adding. I am checking that 'ssl' is not NULL before I dereference it.

Are you suggesting that I put the useRsaPss flag in both struct WOLFSSL and struct WOLFSSL_CTX? I am not sure if that is what you are suggesting or not. If it is, why have it both places when it is working to just have it in the CTX struct?

SparkiDev: Yes, put the flag useRsaPss against both WOLFSSL and WOLFSSL_CTX.
The function you added the check to is: ProcessBufferCertPublicKey()
Its job is to check the properties of the public key in the certificate based on the settings of the ctx or the ssl object. Setting the useRsaPss in here is a good idea.

ProcessBufferCertPublicKey() is called by ProcessBufferCert(). Its job is to parse a DER encode certificate and to record the relevant certificate information against either the ctx or ssl objects.
ProcessBufferCert() is called by ProcessBufferCertType(). Its job is to process a certificate. If the certificate is a CERT_TYPE this means that it is the certificate to use with the TLS handshake. The certificate is stored against either the ctx or ssl object, depending on which is not NULL, in ProcessBufferCertHandleDer().

So, either the ctx or ssl object is being given a certificate to use.
When the ctx is NULL and the ssl is not NULL, then the certificate will be set against the ssl object. The certificate will be used only for the connection that uses the ssl object.
When the ctx is not NULL, then the certificate then the certificate will be set against the ctx object. The certificate will be used by all ssl object created from the ctx object unless another call is made to set a certificate against the ssl object.

Therefore, a TLS connection will use the certificate in the ctx object unless there was one set against the ssl object.
useRsaPss is based on the certificates of the TLS connection.

Set the useRsaPss field in the ctx object when setting the certificate against the ctx.
Set the useRsaPss field in the ssl object when setting the certificate against the ssl.
Copy the useRsaPss field from the ctx object to the ssl object when creating a new ssl.
Check the useRsaPss field in the ssl object in EncodeSigAlg().

holtrop:
retest this please

org.jenkinsci.plugins.workflow.actions.ErrorAction$ErrorId: c80ad59d-806d-41dc-b681-f76263b3ea11
java.lang.NullPointerException: Cannot invoke method getEnvironment() on null object

src/ssl_load.c
SparkiDev: Looks good now.
Thanks!

PR #9445, opened 2025-11-18
extra sanity check on RSA decrypt argument

JacobBarthelmeh:
ZD20791

### DIFF
--- a/wolfcrypt/src/rsa.c
+++ b/wolfcrypt/src/rsa.c
@@ -3638,7 +3638,7 @@ static int RsaPrivateDecryptEx(const byte* in, word32 inLen, byte* out,
                     word32 j;
                     int start = (int)((size_t)pad - (size_t)key->data);
 
-                    for (j = 0; j < key->dataLen; j++) {
+                    for (j = 0; j < key->dataLen && i < outLen; j++) {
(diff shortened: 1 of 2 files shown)


dgarske (approved):

wolfcrypt/src/rsa.c
```diff
@@ -3638,7 +3638,7 @@ static int RsaPrivateDecryptEx(const byte* in, word32 inLen, byte* out,
                     word32 j;
                     int start = (int)((size_t)pad - (size_t)key->data);
 
-                    for (j = 0; j < key->dataLen; j++) {
+                    for (j = 0; j < key->dataLen && i < outLen; j++) {
```
SparkiDev: Don't want to do this if possible - constant time code.
Line 3645 check 'i' against outLen.

SparkiDev (changes requested):

wolfcrypt/src/rsa.c
JacobBarthelmeh: Thanks Sean!

dgarske (approved):

SparkiDev:
Closed in favor of: #9454

PR #9448, opened 2025-11-19
unknown extension support in wc_PKCS7_EcdsaVerify

anhu:
Allow user to prevent wc_PKCS7_EcdsaVerify from erroring out due to extensions we do not know about


anhu:
Devin's comment is invalid.  That function returns `void`.

SparkiDev (approved):

Issue #6395, opened 2023-05-10
GCC -fanalyzer finds an issue


maxgerhardt:
### Version

5.6.0-stable

### Description

Turning on the, since GCC 10 new, `-fanalyzer` [static analyzer](https://gcc.gnu.org/onlinedocs/gcc/Static-Analyzer-Options.html) switch for WolfSSL compilation makes it output a single issue it found with WolfSSL code:

https://github.com/wolfSSL/wolfssl/blob/158c0362e7b12d2455d739814aa23fa565825e2d/src/internal.c#L11552-L11570

it thinks that a dereference of `dCert` can happen through a certain codepath.

```
lib/wolfssl-5.6.0-stable/src/internal.c: In function 'CheckHostName':
lib/wolfssl-5.6.0-stable/src/internal.c:11556:34: warning: dereference of NULL 'dCert' [CWE-690] [-Wanalyzer-null-dereference]
11556 |         if (MatchDomainName(dCert->subjectCN, dCert->subjectCNLen,
      |                             ~~~~~^~~~~~~~~~~
  'CheckHostName': events 1-2
    |
    |11539 | int CheckHostName(DecodedCert* dCert, const char *domainName, size_t domainNameLen)
    |      |     ^~~~~~~~~~~~~
    |      |     |
    |      |     (1) entry to 'CheckHostName'
    |......
    |11547 |     if (CheckForAltNames(dCert, domainName, &checkCN) != 1) {
    |      |         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    |      |         |
    |      |         (2) calling 'CheckForAltNames' from 'CheckHostName'
    |
    +--> 'CheckForAltNames': events 3-9
           |
           |11479 | int CheckForAltNames(DecodedCert* dCert, const char* domain, int* checkCN)
           |      |     ^~~~~~~~~~~~~~~~
           |      |     |
           |      |     (3) entry to 'CheckForAltNames'
           |......
           |11488 |     if (dCert)
           |      |        ~
           |      |        |
           |      |        (4) following 'false' branch (when 'dCert' is NULL)...
           |......
           |11491 |     if (checkCN != NULL) {
           |      |        ~
           |      |        |
           |      |        (5) ...to here
           |      |        (6) following 'true' branch (when 'checkCN' is non-NULL)...
           |11492 |         *checkCN = (altName == NULL) ? 1 : 0;
[...]

PR #9449, opened 2025-11-19
Use only the first 64 bytes of hash output in fips-hash.sh

lealem47:
# Description

Adjust the fips-hash.sh script to only take the first 64-bytes of the sed output to account for an inconsistency in the FIPS callback for module v5.2.0.1

Also has a commit to address https://github.com/wolfSSL/wolfssl/issues/6395

# Testing

./fips-check.sh linuxv5-RC12 keep

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


lealem47:
Jenkins retest this please

PR #9453, opened 2025-11-20
Rust wrapper: add HAVE_HASHDRBG RNG functions

holtrop:
# Description

Add wrappers for:

* `wc_RNG_DRBG_Reseed()`
* `wc_RNG_TestSeed()`
* `wc_RNG_HealthTest_ex()`

# Testing

Unit tests.

# Checklist

 - [X] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/wrapper/rust/wolfssl/tests/test_random.rs
+++ b/wrapper/rust/wolfssl/tests/test_random.rs
@@ -28,6 +28,23 @@ fn test_rng_new_with_nonce_ex_and_drop() {
     let _rng = RNG::new_with_nonce_ex(&mut nonce, None, None).expect("Failed to create RNG");
 }
 
+#[test]
+#[cfg(random_hashdrbg)]
+fn test_health_test() {
+    let nonce = [99u8, 88, 77, 66];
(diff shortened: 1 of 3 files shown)


wrapper/rust/wolfssl/tests/test_random.rs
```diff
@@ -70,3 +85,10 @@ fn test_rng_generate_block_u32() {
     assert_ne!(buffer[buffer.len() - 1], 0u32);
     assert_ne!(buffer[buffer.len() - 1], 0xFFFF_FFFFu32);
 }
+
+#[test]
+fn test_rng_reseed() {
```
gasbytes, line 93: Shouldn't you add `#[cfg(random_hashdrbg)]` on top of `fn test_rng_reseed(){}`? Since the `reseed()` function method is only compiled when the `random_hashdrbg` feature is enabled, this test will fail if that feature is not available.

holtrop: Thank you for mentioning! I actually intended to add it to all of these new methods and tests!

PR #9454, opened 2025-11-21
RSA decrypt: don't write past buffer end on error

SparkiDev:
# Description

When the decrypted data is bigger than the buffer, the one extra bytes was being written to.

Fixes zd#20791

# Testing

./configure --disable-shared --enable-keygen --enable-sp --enable-sp-asm --enable-intelasm 'CC=gcc -fsanitize=address'

# Checklist

 - [x] added tests
 - [ ] updated/added doxygen


SparkiDev:
> 
>     * `no-memory-leaks` [snippet](https://github.com/wolfSSL/wolfssl/blob/804ff42d44eabd20335e35adb4f5e3663b9cb4ab/tests/api/test_rsa.c#L795): Add WC_FREE_VAR(shortPlain, NULL); after the other WC_FREE_VAR calls at the end of test_wc_RsaPublicEncryptDecrypt to release the newly allocated buffer.
> 
> 

Fixed

Issue #7356, opened 2024-03-20
[OpenSSL compatibility]: 5.6.6 and/or v5.6.6-stable-564-g3129e29a1 do not fill int pointed to by 2nd argument to SSL_CIPHER_get_bits(sc, &bitsalg);


mandree:
### Contact Details

(use Gitlab)

### Version

v5.6.6-stable-564-g3129e29a1

### Description

Apparently OpenSSL fills in the bits into the int pointed to by the 2nd argument of SSL_CIPHER_get_bits(sc, &i); in addition to returning them. I haven't found this documented for OpenSSL, but it's in use in the wild and appears to, once upon a time, have been a means to reveal weakened algorithms that only used, say 40 out of 128 bits.

### Reproduction steps

Use WolfSSL's OpenSSL compatibility layer, 
establish a TLS client connection, 
have it connect and negotiate a cipher (assuming SSL_CIPHER *sc is extant and initialized properly), then call:

int bits1;
int bits2 = -23;
bits1 = SSL_CIPHER_get_bits(sc, &bits2);

and see bits2 unchanged, whereas OpenSSL 1.1.1, 3.0 or 3.1 would fill in 128 or 256 on my typical AES-encrypting ciphers.

anhu:
Hi @mandree, 

My name is Anthony and I am a member of the wolfSSL team.  Can you please let me know what configuration flags you are using? 

`*alg_bits` is set if you have `OPENSSL_ALL` defined. If you use `configure` , add `--enable-opensslall` .  

Can you also please let us know a bit about yourself and your project?  For example, Where are you located? Are you using wolfSSL in a  personal, academic or professional project?  Are there any institutions associated with this project? What are you trying to achieve?  We love knowing about how people are using our code.  Please feel free to share whatever you comfortable with. 

Warm regards, Anthony

mandree:
Hi Anthony, this is a finding I've had with the open-source fetchmail project (https://www.fetchmail.info/ or https://gitlab.com/fetchmail/fetchmail on the legacy_6x branch) that I maintain and where wolfSSL might be easier to link to license-wise whilst still supporting TLS v1.3.

The most recent I'd tested is wolfSSL from said git commit (see initial report, above) with
a sub-directory to build and then: 

[...]

PR #9457, opened 2025-11-21
fill in alg_bits

anhu:
Fixes https://github.com/wolfSSL/wolfssl/issues/7356

### DIFF
--- a/src/ssl.c
+++ b/src/ssl.c
@@ -20799,9 +20799,11 @@ int wolfSSL_CIPHER_get_bits(const WOLFSSL_CIPHER *c, int *alg_bits)
     WOLFSSL_ENTER("wolfSSL_CIPHER_get_bits");
 
     #if defined(WOLFSSL_QT) || defined(OPENSSL_ALL)
-    (void)alg_bits;
-    if (c!= NULL)
+    if (c != NULL) {
         ret = c->bits;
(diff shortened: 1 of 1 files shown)


anhu:
I'm a little weary about this one as I think we were not filling it in on purpose.  It doesn't look like that was an accident.

src/ssl.c
```diff
@@ -20831,9 +20799,11 @@ int wolfSSL_CIPHER_get_bits(const WOLFSSL_CIPHER *c, int *alg_bits)
     WOLFSSL_ENTER("wolfSSL_CIPHER_get_bits");
 
     #if defined(WOLFSSL_QT) || defined(OPENSSL_ALL)
-    (void)alg_bits;
-    if (c!= NULL)
+    if (c != NULL) {
         ret = c->bits;
+        if (alg_bits != NULL)
+            *alg_bits = c->bits;
```
SparkiDev, line 20805: The algorithm bits is c->bits while what is returned by OpenSSL is the strength bits which may be less due to the cipher suite.
CCM_8 has a strength_bits of 64 and an algorithm bits of 128.
Integrity only cipher suites have 0 strength bits but hash length algorithm bits.

anhu:
Closing as this is doing the wrong thing.

PR #9460, opened 2025-11-21
Add option `--enable-baremetal` and `WOLFSSL_BAREMETAL` macro

dgarske:
# Description

Adds `--enable-baremetal` configure option and `WOLFSSL_BAREMETAL` macro to simplify configuration for bare-metal embedded systems. This option encapsulates commonly-used defines for bare-metal environments into a single flag.

The `WOLFSSL_BAREMETAL` macro defines:
- `SINGLE_THREADED` - No threading support
- `NO_DEV_RANDOM` - No /dev/random access
- `NO_FILESYSTEM` - No file system access
- `NO_WRITEV` - No writev() system call
- `NO_STDIO_FILESYSTEM` - No stdio-based file operations
- `WOLFSSL_NO_SOCK` - No socket support
- `WOLFSSL_NO_GETPID` - No process ID support
- `NO_ASN_TIME` - **Conditionally defined only when `WOLFCRYPT_ONLY` is also defined**. For systems without RTC, this bypasses certificate date checking.

**Important**: Users must provide their own entropy source when using this configuration since `NO_DEV_RANDOM` is defined. Implement `wc_GenerateSeed()` with platform-specific hardware RNG.

**Changes based on reviewer feedback:**
- Removed `WOLFCRYPT_ONLY` from the macro - users should enable it separately via `--enable-cryptonly` if needed
- Made `NO_ASN_TIME` conditional on `WOLFCRYPT_ONLY` to avoid forcing certificate date checking to be disabled

# Testing

Tested with autotools build system:
```bash
./autogen.sh
./configure --enable-baremetal
grep WOLFSSL_BAREMETAL wolfssl/options.h  # Verified macro is defined
# Verified NO_ASN_TIME is NOT defined with just --enable-baremetal
make clean && make
```

Tested with both flags:
```bash
./configure --enable-baremetal --enable-cryptonly
# Verified NO_ASN_TIME IS defined when WOLFCRYPT_ONLY is present
make clean && make
```

Tested with CMake build system:
```bash
mkdir build && cd build
cmake -DWOLFSSL_BAREMETAL=yes ..
make
```

Verified that `WOLFSSL_BAREMETAL` in settings.h correctly defines all component macros when enabled, and that `NO_ASN_TIME` is only defined when `WOLFCRYPT_ONLY` is also present.

# Checklist

- [ ] added tests
- [ ] updated/added doxygen


dgarske:
Jenkins retest this please

kaleb-himes:
For FIPS purposes, baremetal is a scary term that labs and CMVP won't touch. 99.9% of the time systems that would require this setting will actually have Firmware present and won't be truely and completely "baremetal".

This configure option is great for cases where FIPS doesn't matter but if we want to use this feature set on a system with primitive Firmware that isn't truely and completely "baremetal" it would be nice to have an alias that won't scare the CMVP and the CSTL's when they see it in the Module Management Manual or User Guide as we call it.


An alias like `--enable-no-os` that does exactly the same thing (just an alias for --enable-baremetal) would be nice for validation purposes?

dgarske:
Closing PR. @jackctj117 will you make sure the wolfssl/documentation user manual has a section for bare-metal and build options that are useful for that environment?

PR #9466, opened 2025-11-24
TLS 1.3, plaintext alert: ignore when expecting encrypted

SparkiDev:
# Description

In TLS 1.3, ignore valid unencrypted alerts that appear after encryption has started.
Only ignore WOLFSSL_ALERT_COUNT_MAX-1 alerts.

Fixes zd#20857

# Testing

./configure --disable-shared
make
./tests/unit.test -test_tls13_plaintext_alert

# Checklist

 - [x] added tests
 - [ ] updated/added doxygen


SparkiDev:
retest this please

rizlik:
If the goal of this PR is to protect against DoS I don't think it's a good idea:

- under TCP one can easily forge a RST packet and close the connection
- peer can still forge arbitrary packets that cause the handshake to error out (handshake, app_data) ecc

UDP is different but `DtlsShouldDrop`  ignores plaintext packet after handshake keys are computed.

SparkiDev:
DoS can be done by any message, it doesn't have to be a valid alert.
But yes that is the report.

Instead the PR is about skipping alerts that were sent by the client before it received anything from the server to indicate it should be encrypted. OpenSSL and others do this.

May make this a compile time option.

SparkiDev:
retest this please

dgarske (dismissed):

SparkiDev:
Changed to be compile time option when WOLFSSL_TLS13_IGNORE_PT_ALERT_ON_ENC is defined.
Test is testing as many alerts as is required to cause an error or one less being OK.
Test is checking that without define that an error is returned.

SparkiDev:
retest this please

src/internal.c
```diff
@@ -21726,20 +21732,36 @@ static int DoAlert(WOLFSSL* ssl, byte* input, word32* inOutIdx, int* type)
...
+#endif
     }
-    else {
-        /*
-         * A close_notify alert doesn't mean there's been an error, so we only
-         * add other types of alerts to the error queue
-         */
-        WOLFSSL_ERROR(*type);
+    else
+    {
```
julek-wolfssl:
```suggestion
    else {
```

julek-wolfssl (changes requested):
Please add `WOLFSSL_TLS13_IGNORE_PT_ALERT_ON_ENC` to `os-check.yml`. I think a good configuration would be one that explicitly enables TLS and DTLS both 1.2 and 1.3.

src/internal.c
SparkiDev: fixed

SparkiDev:
rebased and add configuration to os-check.yml

.github/workflows/os-check.yml
```diff
@@ -63,6 +63,8 @@ jobs:
           '--enable-coding=no',
           '--enable-dtls --enable-dtls13 --enable-ocspstapling --enable-ocspstapling2
            --enable-cert-setup-cb --enable-sessioncerts',
+          '--enable-dtls --enable-dtls13 --enable-tls13
+           CFLAGS=-DWOLFSSL_TLS13_IGNORE_PT_ALERT_ON_ENC',
```
julek-wolfssl: Can't use `CFLAGS` because it is used in the `configure:` parameter passed into `wolfSSL/actions-build-autotools-project`. `CPPFLAGS` needs to be used. I see one more usage of `CFLAGS` in the configurations. Could you please fix that one too?

julek-wolfssl (changes requested):

.github/workflows/os-check.yml
SparkiDev: fixed

SparkiDev:
rebased and updated os-check.yml to use CPPFLAGS

SparkiDev:
retest this please

generic:
wolf-linux-cloud-node-d3ayn7 was marked offline: Connection was broken

julek-wolfssl (approved):

SparkiDev:
retest this please

PR #9489, opened 2025-12-03
Fix AKID CA lookup

julek-wolfssl:
The `authorityCertIssuer` field refers to the Issuer field of the CA being looked up and not its Subject field.

Addresses zd#20860


julek-wolfssl:
Retest this please RequestAbortedException

dgarske (approved):

PR #9504, opened 2025-12-09
Fix uninit variables

julek-wolfssl:


dgarske (approved):
Merging with async PRB multi-test error

PR #9506, opened 2025-12-10
check if ctx and ssl are null when checking public key in certificate

rlm2002:
# Description

Adds a check for when both `ssl` and `ctx` are null, returns `BAD_FUNC_ARG` if true. Fixes CID 551515.

# Testing

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


rlm2002:
Retest this please Jenkins. PRB-generic-config-parser was not stable. Unable to create live FilePath for wolf-linux-cloud-node-gvndu9

PR #9515, opened 2025-12-10
Note that HMAC_FIPS_MIN_KEY is also salt len min for HKDF

anhu:
Fixes ZD 20873


anhu:
Please ignore build failures.  This is PR is simply adding text to a comment.

PR #9516, opened 2025-12-10
Add checking of size param and clarify usage in doc

### LINKED ISSUE #3665: wc_SrpGetPublic function output incorrect public key value (pub) when padding is needed.
Background: This function wc_SrpGetPublic takes in pointer to variable that will hold public Key.  After function executes it writes the computed public key this variable.  If public key variable is larger than resulting public key , the public key needs to be padded and done correctly by adding 0 before MSB.

It is not padding right now and expect the user of function to handles it by looking at *size output. This needs to be clearly documented otherwise, it could causes user to unpadded(incorrect) passkey.

embhorn:
# Description

The doc says that the `size` parameter of `wc_SrpGetPublic` is the size of the buffer passed in the `pub` parameter. The buffer was only cleared for the size of the public ephemeral value.

Fixes #3665

# Testing

make check

# Checklist

 - [ ] added tests
 - [x] updated/added doxygen

### DIFF
--- a/doc/dox_comments/header_files/srp.h
+++ b/doc/dox_comments/header_files/srp.h
@@ -340,6 +340,8 @@ int wc_SrpSetPrivate(Srp* srp, const byte* priv, word32 size);
     This function MUST be called after wc_SrpSetPassword or wc_SrpSetVerifier.
     The function wc_SrpSetPrivate may be called before wc_SrpGetPublic.
 
+    Caller must observe value of size upon return to know the actual size.
+
     \return 0 Success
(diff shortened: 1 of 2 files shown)


embhorn:
Retest this please

Multiple Arduino test failures
```
+ exit 1
Primary install failed: curl=0, sh=. Falling back...
/home/runner/work/_temp/c93b99ad-f6b3-4f95-8086-1712242c5803.sh: line 35: pushd: ./Arduino/sketches: No such file or directory
Error: Process completed with exit code 1.
```

JacobBarthelmeh (approved):

douzzer (approved):

PR #9536, opened 2025-12-13
 Fix AArch64 assembly syntax causing clang-21 crash

kyulee-com:
# Description

The wolfssl ARM assembly files were mixing ELF syntax (`:lo12:`) with Darwin syntax (`@PAGEOFF`) in the same instruction, which causes a clang-21 assembler crash with the error `"cannot apply another specifier to MCSpecifierExpr"`. On Darwin/Apple platforms, `@PAGEOFF` already represents the low 12 bits of the address, so the `:lo12:` prefix is redundant and incorrect. This was causing `UNREACHABLE` assertion failures in `AsmParser.cpp` when building with clang-21.
The fix removes the `:lo12:` prefix from instructions using `@PAGEOFF` on Apple platforms in both `armv8-sha3-asm.S` and `armv8-sha512-asm.S`.

# Testing

Confirmed the build didn't crash with new clang on Mac.

### DIFF
--- a/wolfcrypt/src/port/arm/armv8-sha3-asm.S
+++ b/wolfcrypt/src/port/arm/armv8-sha3-asm.S
@@ -92,7 +92,7 @@ _BlockSha3_crypto:
 	add  x1, x1, :lo12:L_SHA3_transform_crypto_r
 #else
 	adrp x1, L_SHA3_transform_crypto_r@PAGE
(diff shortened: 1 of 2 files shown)


SparkiDev:
Hi @kyulee-com,

Thanks for letting us know!

The code you changed is generated and there are many more places where this fix is required.
Closing in favour of #9537

Sean

PR #9554, opened 2025-12-18
Run rng-tools in docker

julek-wolfssl:
When running locally I noticed that sometimes the `rngd` is still running after the test fails/hangs. I suspect this may be the reason that is is more flaky recently.


julek-wolfssl:
Not the right fix

PR #9571, opened 2025-12-22
Add missing length check in sniffer for AES-GCM/AES-CCM/ARIA-GCM

mattia-moffa:
# Description

Without this check, with a malformed packet there was a possibility for integer underflow (and subsequent buffer overflow) when `AESGCM_EXP_IV_SZ` and `ssl->specs.aead_mac_size` are subtracted from the size later.

Fixes zd#20969

# Testing

```
./configure --enable-sniffer --enable-aria
```


SparkiDev:
retest this please

PR #9576, opened 2025-12-23
20251222-linuxkm-PK-initrng-optimize

douzzer:
`wolfcrypt/src/rsa.c` and `wolfssl/wolfcrypt/rsa.h`: make `RsaKey.rng` and `wc_RsaSetRNG()` available unconditionally, rather than only if `WC_RSA_BLINDING`, for use by `wc_CheckRsaKey()`.

`wolfcrypt/src/random.c` and `wolfssl/wolfcrypt/random.h`:
* add `WC_DRBG_{NOT_INIT,OK,FAILED,CONT_FAILED}` in public header file, and
* move setup for `RNG_SECURITY_STRENGTH`, `ENTROPY_SCALE_FACTOR`, `SEED_BLOCK_SZ`, `SEED_SZ`, `MAX_SEED_SZ`, and `RNG_HEALTH_TEST_CHECK_SIZE` from `random.c` to `random.h`, with public `WC_DRBG_SEED_SZ` and `WC_DRBG_MAX_SEED_SZ`.

`linuxkm/lkcapi_sha_glue.c`:
* add `drbg_init_from()` and `fork_default_rng()`, and
* use the latter to define `LKCAPI_INITRNG_FOR_SELFTEST()` opportunistically (with fallback to plain `wc_InitRng()`);

`linuxkm/lkcapi_rsa_glue.c`:
* add `km_rsa_ctx_init_rng()`,
* remove `wc_InitRng()` from `km_rsa_ctx_init()`,
* remove the `WC_RSA_BLINDING` gates around calls to `wc_RsaSetRNG()`, and
* call `km_rsa_ctx_init_rng()` before each call that needs an initialized RNG;

`linuxkm/lkcapi_dh_glue.c` and `linuxkm/lkcapi_ecdh_glue.c`: in `km_ffdhe_init()` and `km_ecdh_init()`, if `linuxkm_lkcapi_registering_now`, use `LKCAPI_INITRNG_FOR_SELFTEST()` to initialize `ctx->rng`;

`linuxkm/lkcapi_glue.c`: add notes that `lkcapi_sha_glue` inclusion and registrations must precede PK, and move declaration of `linuxkm_lkcapi_registering_now` to precede lkcapi glue inclusions.

tested with
```
wolfssl-multi-test.sh ...
--exclude-scenario '.*(v5|v6).*' '.*linuxkm.*' '.*insmod.*'
```
and
```
wolfssl-multi-test.sh ...
pr-check
```
and after backporting the changes in wolfcrypt/src and wolfssl/wolfcrypt/ to FIPS v5.2.4,
```
wolfssl-multi-test.sh ...
'.*v5.*insmod.*'
```

### DIFF
--- a/wolfcrypt/src/rsa.c
+++ b/wolfcrypt/src/rsa.c
@@ -201,11 +201,6 @@ int wc_InitRsaKey_ex(RsaKey* key, void* heap, int devId)
     (!defined(WOLFSSL_RSA_VERIFY_ONLY) && !defined(WOLFSSL_RSA_VERIFY_INLINE)))
     key->dataIsAlloc = 0;
 #endif
-    key->data = NULL;
-    key->dataLen = 0;
-#ifdef WC_RSA_BLINDING
-    key->rng = NULL;
-#endif
 
 #ifdef WOLF_CRYPTO_CB
     key->devId = devId;
@@ -741,6 +736,13 @@ int wc_CheckRsaKey(RsaKey* key)
         }
 #endif
         ret = wc_InitRng(rng);
+        if (ret != 0) {
+#if defined(WOLFSSL_SMALL_STACK) && !defined(WOLFSSL_NO_MALLOC)
+            XFREE(rng, NULL, DYNAMIC_TYPE_RNG);
+            FREE_MP_INT_SIZE(tmp, NULL, DYNAMIC_TYPE_RSA);
+#endif
+            return ret;
+        }
     }
 
     SAVE_VECTOR_REGISTERS(ret = _svr_ret;);
--- a/linuxkm/lkcapi_sha_glue.c
+++ b/linuxkm/lkcapi_sha_glue.c
@@ -1134,6 +1134,11 @@ static inline void put_drbg(struct wc_rng_inst *drbg) {
 static inline struct crypto_rng *get_crypto_default_rng(void) {
     struct crypto_rng *current_crypto_default_rng = crypto_default_rng;
 
+    if (unlikely(! current_crypto_default_rng)) {
+        pr_warn("BUG: get_default_drbg_ctx() called with NULL crypto_default_rng.");
+        return NULL;
+    }
+
     if (unlikely(! wc_linuxkm_drbg_default_instance_registered)) {
         pr_warn("BUG: get_default_drbg_ctx() called without wc_linuxkm_drbg_default_instance_registered.");
         return NULL;


wolfssl/wolfcrypt/rsa.h
```diff
@@ -214,9 +214,9 @@ struct RsaKey {
     int   type;                               /* public or private */
     int   state;
     word32 dataLen;
-#ifdef WC_RSA_BLINDING
```
philljj, line 217: Not sure it's ok to remove this WC_RSA_BLINDING guard for embedded targets that might be using RSA verify only.

philljj: For verify-only targets this will still build because `rsa.h` always includes `random.h`, which defines `struct WC_RNG {`. But rng will be a dead code member of RsaKey.

Would recommend we guard this a different way, or always require WC_RSA_BLINDING with linuxkm. We already reject `WOLFSSL_RSA_VERIFY_ONLY` and `WOLFSSL_RSA_PUBLIC_ONLY` in `linuxkm/lkcapi_rsa_glue.c`.

wolfcrypt/src/rsa.c
```diff
@@ -718,17 +721,27 @@ int wc_CheckRsaKey(RsaKey* key)
...
     NEW_MP_INT_SIZE(tmp, mp_bitsused(&key->n), NULL, DYNAMIC_TYPE_RSA);
 #ifdef MP_INT_SIZE_CHECK_NULL
     if (tmp == NULL) {
-        XFREE(rng, NULL, DYNAMIC_TYPE_RNG);
         return MEMORY_E;
     }
 #endif
 
-    ret = wc_InitRng(rng);
+    if (key->rng)
```
philljj, line 726: I don't understand need for this optimization. We're not calling wc_CheckRsaKey from linuxkm. It's used mainly in the pkcs code, and one place in wolfcrypt test.

philljj (changes requested):

wolfcrypt/src/rsa.c
philljj, line 207: This should not be guarded by WC_RSA_BLINDING if we're going to remove the guard in `rsa.h`.

wolfssl/wolfcrypt/rsa.h
douzzer: Ah, good catch!  I'll put
```
#ifndef WC_NO_RNG
```
around it, and similarly around `wc_RsaSetRNG()`.

Bigger picture, aside from super-cut-down configs with `WC_NO_RNG`, the `rng` is needed for just about every RSA op -- public key ops (pad construction), sig gen ops (padding and blinding), key gen ops of course, `_CheckProbablePrime()`, and for `_ifc_pairwise_consistency_test()` (which tests the key by generating and verifying a signature using the key under test).  Seems best to have the `rng ` in `RsaKey` unless `WC_NO_RNG`.

wolfcrypt/src/rsa.c
douzzer: This was the only `wc_InitRng()` call inside the v5 FIPS boundary that was unconditional.  Everywhere else, an rng can either be set in the key context, or passed in as an arg.  Given that WolfEntropy dynamics empirically destabilize the kernel, it seemed prudent to plan ahead in case we need to be able to do PKCS#7 in kernel mode.

I've just discovered that we added an rng dependency in curve25519.c (bb84ebfd7acff) conditional on `WOLFSSL_CURVE25519_BLINDING`, and `wc_curve25519_make_pub()` has a similar issue with a local rng -- I'll add `wc_curve25519_make_pub_ex()` that adds an `rng` arg.  That's only inside the v6  FIPS boundary but it could easily come back to bite us.

Edit: no changes needed in curve25519.c -- `wc_curve25519_make_pub_blind()` is already a public function that takes an `rng` and functions otherwse same as `wc_curve25519_make_pub()`.  In `WOLFSSL_CURVE25519_BLINDING` configs, the linuxkm can call  `wc_curve25519_make_pub_blind()` directly.

The only place in our code that's currently affected is 
src/wolfcrypto_shim.c in our fork of the WireGuard kernel module (note, WireGuard, not WolfGuard).  Not worth thinking about at present.

philljj (approved):

PR #9591, opened 2025-12-29
20251229 Coverity Dereference before Null check

rlm2002:
# Description

CID 315897 Dereference before null check - move null check to prevent dereference before null check issue

### DIFF
--- a/wolfcrypt/src/ecc.c
+++ b/wolfcrypt/src/ecc.c
@@ -2703,236 +2703,239 @@ int ecc_projective_dbl_point(ecc_point *P, ecc_point *R, mp_int* a,
 */
 int ecc_map_ex(ecc_point* P, mp_int* modulus, mp_digit mp, int ct)
 {
-   int err = MP_OKAY;
-#if !defined(WOLFSSL_SP_MATH)
-   DECL_MP_INT_SIZE_DYN(t1, mp_bitsused(modulus), MAX_ECC_BITS_USE);
-   DECL_MP_INT_SIZE_DYN(t2, mp_bitsused(modulus), MAX_ECC_BITS_USE);
-#ifdef ALT_ECC_SIZE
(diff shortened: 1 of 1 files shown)


rlm2002:
retest this please jenkins: Agent offline exception
PRB-generic-config-parser was not stable
PRB-multi-test-script was not stable
PRB-fips-repo-and-harness-test-v3-part1 was not stable

wolfcrypt/src/ecc.c
```diff
@@ -2703,236 +2703,239 @@ int ecc_projective_dbl_point(ecc_point *P, ecc_point *R, mp_int* a,
...
+    int err = MP_OKAY;
+    mp_int *x, *y, *z;
+    (void)ct;
 
-   if (P == NULL || modulus == NULL)
-       return ECC_BAD_ARG_E;
+    if (P == NULL || modulus == NULL){
+        return ECC_BAD_ARG_E;
+    }
+    {
```
dgarske, line 2713: The `DECL_MP_INT_SIZE_DYN` can be at top, this is just declaring, not allocating... that's done in NEW_MP_INT_SIZE. Then revert all the indent changes. Thanks

dgarske (changes requested):

wolfcrypt/src/ecc.c
rlm2002: I think it's reasonable to close this PR without merging then. The remaining changes wouldn't make a difference.

dgarske: I like the new argument NULL checking since this API can be built as public facing.

rlm2002: Sorry, just remembered my thought process for the change here. The current format moves `DECL_MP_INT_SIZE_DYN` below the already existing NULL check since `mp_bitsused(modulus)` is defined `((a)->used * SP_WORD_SIZE)`, which is what flagged the original Coverity issue. The additional scope and indent prevents C99 standard compilation issues.

SparkiDev (approved):

PR #9595, opened 2025-12-30
20251229-linuxkm-rng-wolfentropy

douzzer:
`linuxkm/lkcapi_sha_glue.c`: add retry loop around `wc_InitRng()`, and allow interrupt in preemptible threads, in `wc_linuxkm_drbg_init_tfm()`.

`wolfcrypt/src/random.c` and `wolfssl/wolfcrypt/settings.h`: add `WC_VERBOSE_RNG` messages, and activate by default when `WOLFSSL_KERNEL_MODE`.

`linuxkm/Makefile`: fix bash cleanup in recipe for `libwolfssl.ko` -- new trap for an event replaces previous trap rather than adding to it.

tested with `wolfssl-multi-test.sh ... check-source-text '.*insmod.*'`

### DIFF
--- a/wolfcrypt/src/random.c
+++ b/wolfcrypt/src/random.c
@@ -752,10 +752,6 @@ int wc_RNG_TestSeed(const byte* seed, word32 seedSz)
         if (ConstantCompare(seed + seedIdx,
                             seed + seedIdx + scratchSz,
                             (int)scratchSz) == 0) {
-
-    #ifdef WC_VERBOSE_RNG
-            WOLFSSL_DEBUG_PRINTF("wc_RNG_TestSeed() returning DRBG_CONT_FAILURE on duplicate word32,");
-    #endif
             ret = DRBG_CONT_FAILURE;
         }
         seedIdx += SEED_BLOCK_SZ;
@@ -985,7 +981,12 @@ static int _InitRng(WC_RNG* rng, byte* nonce, word32 nonceSz,
             if (ret != 0) {
                 WOLFSSL_MSG_EX("wc_RNG_TestSeed failed... %d", ret);
             }
+    #elif defined(WC_VERBOSE_RNG)
+            if (ret != DRBG_SUCCESS) {
+                WOLFSSL_DEBUG_PRINTF("wc_RNG_TestSeed() in _InitRng() returned err %d.", ret);
+            }
     #endif
+
             if (ret == DRBG_SUCCESS)
                 ret = Hash_DRBG_Instantiate((DRBG_internal *)rng->drbg,
             #if defined(HAVE_FIPS) || !defined(WOLFSSL_RNG_USE_FULL_SEED)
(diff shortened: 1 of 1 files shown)


douzzer:
retest this please
```
ound unhandled org.jenkinsci.plugins.workflow.support.steps.AgentOfflineException exception:
Unable to create live FilePath for wolf-linux-cloud-node-0171uz; wolf-linux-cloud-node-0171uz was marked offline: Connection was broken
```

wolfcrypt/src/random.c
```diff
@@ -525,6 +536,11 @@ static int Hash_gen(DRBG_internal* drbg, byte* out, word32 outSz, const byte* V)
     WC_FREE_VAR_EX(data, drbg->heap, DYNAMIC_TYPE_TMP_BUFFER);
 #endif
 
+    #ifdef WC_VERBOSE_RNG
+    if ((ret != DRBG_SUCCESS) && (ret != DRBG_FAILURE))
```
philljj: Why not print if ret == DRBG_FAILURE?

If intended, then maybe a small comment to explain.

wolfcrypt/src/random.c
```diff
@@ -635,6 +651,11 @@ static int Hash_DRBG_Generate(DRBG_internal* drbg, byte* out, word32 outSz)
     #endif
     }
 
+    #ifdef WC_VERBOSE_RNG
+    if ((ret != DRBG_SUCCESS) && (ret != DRBG_FAILURE))
```
philljj: Same.

wolfcrypt/src/random.c
```diff
@@ -1004,6 +1033,7 @@ static int _InitRng(WC_RNG* rng, byte* nonce, word32 nonceSz,
         ret = RNG_FAILURE_E;
     }
     else {
+    if ((ret != DRBG_SUCCESS) && (ret != DRBG_FAILURE))
```
philljj: This seems important to comment.

Is this to allow retry on `ret == DRBG_FAILURE`, because the status isn't marked as failed?

linuxkm/lkcapi_sha_glue.c
```diff
@@ -1007,19 +1011,35 @@ static int wc_linuxkm_drbg_init_tfm(struct crypto_tfm *tfm)
...
+        for (;;) {
+            u64 ts2;
+            if (wc_linuxkm_drbg_init_tfm_disable_vector_registers)
+                need_reenable_vec = (DISABLE_VECTOR_REGISTERS() == 0);
+            ret = wc_InitRng(&ctx->rngs[i].rng);
+            if (need_reenable_vec)
+                REENABLE_VECTOR_REGISTERS();
+            if (can_sleep)
+                cond_resched();
+            if (ret == 0)
```
philljj, line 1030: Why not check if `ret == 0` and break before the `if (can_sleep) cond_resched();`?

Confused by the two non-consecutive `if (can_sleep)` checks. Maybe short comment justified.

wolfcrypt/src/random.c
philljj: Also the formatting of this if is confusing. Maybe indent the if, or make it an `else if {`

wolfcrypt/src/random.c
douzzer: if it's just going to return `DRBG_FAILURE` to the caller, then there's no point printing it out because (1) the lower-level code that was remapped to `DRBG_FAILURE` already got printed before the remapping, so a `DRBG_FAILURE` message would just be spamming the kernel log, and (2) the caller will actually see the `DRBG_FAILURE` code, and is free to (and probably will) log it itself.

edit: am adding a cleaned up version of the above as a comment in the source.

wolfcrypt/src/random.c
douzzer: Same explanation.

wolfcrypt/src/random.c
douzzer: Whoops, no, that was an oopsee.  At one point I had a conditional logging event there, and I only partially reverted it.  Good catch.  Fixing.

philljj:
Retest this please.

(Found unhandled exception).

douzzer:
retest this please
```
```

philljj (approved):

PR #9599, opened 2026-01-02
Rust wrapper: add wolfssl_wolfcrypt::chacha20_poly1305 module

holtrop-wolfssl:
# Description

Rust wrapper: add wolfssl_wolfcrypt::chacha20_poly1305 module

# Testing

How did you test?

# Checklist

 - [X] added tests
 - [ ] updated/added doxygen


holtrop-wolfssl:
retest this please (org.jenkinsci.plugins.workflow.support.steps.AgentOfflineException: Unable to create live FilePath for wolf-linux-cloud-node-w6da12; wolf-linux-cloud-node-w6da12 was marked offline: Connection was broken)

wrapper/rust/wolfssl-wolfcrypt/src/chacha20_poly1305.rs
```diff
@@ -0,0 +1,242 @@
...
+        };
+        if rc != 0 {
+            return Err(rc);
+        }
+        Ok(())
+    }
+
+    /// Finalize the decrypt/encrypt operation.
+    ///
+    /// This function consumes the `ChaCha20Pol1305` instance. The
```
gasbytes: Small typo (`ChaCha20Pol1305` -> `ChaCha20Poly1305`)

wrapper/rust/wolfssl-wolfcrypt/src/chacha20_poly1305.rs
```diff
@@ -0,0 +1,242 @@
...
+This module provides a Rust wrapper for the wolfCrypt library's
+ChaCha20-Poly1305 functionality.
+*/
+
+#![cfg(chacha20_poly1305)]
+
+use crate::sys;
+use std::mem::MaybeUninit;
+
+pub struct ChaCha20Poly1305 {
```
gasbytes, line 31: Consider adding the `Drop` implementation so that in case of error without calling finalize (early error/panic) we correctly clean the previously allocated resources/buffers etc...

wrapper/rust/wolfssl-wolfcrypt/src/chacha20_poly1305.rs
holtrop-wolfssl: Thank you; fixed!

wrapper/rust/wolfssl-wolfcrypt/src/chacha20_poly1305.rs
holtrop-wolfssl: I had originally added it but then found that there is no _Free function for this struct like there is for several of our other structs. So it looked to me like there was nothing that needed to be done on drop. Are you aware of anything else that should be done?

gasbytes: If there is no _Free function for this struct, that probably means that no cleanup is needed in case of failure, my main concern was "cleanup" the state of the struct (key material, internal cipher state).
You could though derive ZeroizeOnDrop on the struct, or implement the Drop function so that it uses zeroize (I believe that's what is cqlled, the craate), so that it correctly zeroes the struct.

holtrop-wolfssl:
Retest this please (org.jenkinsci.plugins.workflow.support.steps.AgentOfflineException: Unable to create live FilePath for wolf-linux-cloud-node-km3xbh; wolf-linux-cloud-node-km3xbh was marked offline: Connection was broken)

reneleonhardt:
With even your build dependencies supporting Rust < 1.85, I'm wondering if it would be possible to dial your edition a notch back to 2021 to allow more libraries to try the brand new wolfssl-wolfcrypt? 😄

For example, you may want to be added to https://blessed.rs/crates#section-cryptography

PR #9601, opened 2026-01-05
check that we are resuming in write_early_data + minor fixes

rizlik:
# Description

1. check that we are resuming or using PSK when invoking `wolfSSL_write_early_data`, error otherwise
2. set `ssl->earlyData == expecting_early_data` only on the first `wolfSSL_write_early_data` invocation
3. merge consecutive section guarded by the same macro

### DIFF
--- a/src/tls13.c
+++ b/src/tls13.c
@@ -14990,8 +14990,9 @@ int wolfSSL_get_max_early_data(WOLFSSL* ssl)
  * sz     The size of the early data in bytes.
  * outSz  The number of early data bytes written.
  * returns BAD_FUNC_ARG when: ssl, data or outSz is NULL; sz is negative;
- * or not using TLS v1.3. SIDE ERROR when not a server. Otherwise the number of
- * early data bytes written.
+ * or not using TLS v1.3. SIDE ERROR when not a server. BAD_STATE_E if invoked
+ * without a valid session or without a valid PSK CB.
+ * Otherwise the number of early data bytes written.
  */
 int wolfSSL_write_early_data(WOLFSSL* ssl, const void* data, int sz, int* outSz)
 {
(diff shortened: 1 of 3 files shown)


src/tls13.c
```diff
@@ -14996,8 +15008,16 @@ int wolfSSL_write_early_data(WOLFSSL* ssl, const void* data, int sz, int* outSz)
...
+    if (!EarlyDataPossible(ssl)) {
+        ssl->error = BAD_STATE_E;
+        return WOLFSSL_FATAL_ERROR;
+    }
+
     if (ssl->options.handShakeState == NULL_STATE) {
-        if (ssl->error != WC_NO_ERR_TRACE(WC_PENDING_E))
+        /* avoid re-setting ssl->earlyData if we re-enter the function because
+         * of WC_PENDING_E, WANT_WRITE or WANT_READ */
+        if (ssl->error == 0)
```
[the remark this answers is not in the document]
rizlik: won't fix

src/tls13.c
```diff
@@ -14996,8 +15008,16 @@ int wolfSSL_write_early_data(WOLFSSL* ssl, const void* data, int sz, int* outSz)
     if (ssl->options.side == WOLFSSL_SERVER_END)
         return SIDE_ERROR;
 
+    /* Early data requires PSK or session resumption */
+    if (!EarlyDataPossible(ssl)) {
+        ssl->error = BAD_STATE_E;
+        return WOLFSSL_FATAL_ERROR;
+    }
```
julek-wolfssl, line 15015: Not every path in `wolfSSL_write_early_data` set `ssl->error`. Setting it effectively kills this object. Either all paths should set the error or none should (unless in bad state).

rizlik: what do you suggest? returning BAD_STATE_E directly?

julek-wolfssl: Yes I think so.

rizlik:
retest this please

PR #9607, opened 2026-01-05
Fixes for PSK Handshake without Diffie-Hellman

Frauschi:
* Make sure no KeyShare is sent from the client when DH is disabled in PSK mode
* Fix server error in Suite matching in case no KeyShare is present

Found while testing PSK-only handshakes on embedded devices where client-side KeyShare generation is costly.

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -37122,6 +37122,14 @@ static int DoSessionTicket(WOLFSSL* ssl, const byte* input, word32* inOutIdx,
         }
 #endif
 
+#if !defined(NO_PSK)
+        if(havePSK && ssl->options.noPskDheKe) {
+            /* In this case, no keyshare extension is sent from the client and
+             * we can exit the verification successfully. */
+            return 1;
+        }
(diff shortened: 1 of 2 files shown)


dgarske:
Okay to test.

Frauschi:
After checking the failing tests and digging deeper in the behavior of the `PskKeyExchangeModes` extension, I conclude that this PR in the current form does not work. 

As the `PskKeyExchangeModes` extension is also used to indicate to the server the desired PSK modes for a future SessionTicket, it is also sent when no actual PSK for the current session is provided. Hence, even when the `noPskDheKe` option is enabled, we cannot simply disable the generation of the `KeyShare` (and necessarily the `SupportedGroups`) extensions for the ClientHello. We may only do that if the ClientHello actually contains a `PreSharedKey` extension. To implement this, however, more thorough refactoring of the `TLSX_PopulateExtensions()` method would be necessary, which is probably not worth the effort at the moment.

PR #9609, opened 2026-01-05
Fix memory leak in case of handshake error

Frauschi:
Make sure a peer Dilithium key is properly freed in case the handshake fails.


SparkiDev:
Okay to test.

Frauschi:
Retest this please

SparkiDev:
retest this please

SparkiDev (approved):

PR #9610, opened 2026-01-05
Remove PQC-based buffer size increase for PreMasterSecret

Frauschi:
The size of the PreMasterSecret buffer is based on the `ENCRYPT_LEN` constant, which has been increased to 5kB for PQC support (Dilithium and Falcon, as their signatures are that large).

However, only in the TLS 1.2 case, the PreMasterSecret buffer is used to store signatures. In the TLS 1.3 path, only actual symmetric secrets are stored in that buffer, which are much smaller (the "old" size of the constant without the PQC increase).

As PQC is only allowed in TLS 1.3 and NOT in TLS 1.2, we can revert that size increase, saving around 4,5kB of dynamic memory during the handshake.

### DIFF
--- a/src/tls13.c
+++ b/src/tls13.c
@@ -10295,8 +10295,7 @@ static int DoTls13CertificateVerify(WOLFSSL* ssl, byte* input,
             args->idx += OPAQUE16_LEN;
 
             /* Signature data. */
-            if ((args->idx - args->begin) + args->sz > totalSz ||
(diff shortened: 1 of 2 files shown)


dgarske:
Okay to test

dgarske:
Jenkins retest this please.

dgarske:
Jenkins retest this please.

PR #9611, opened 2026-01-05
Fix for PSK compile option

Frauschi:
The derivation of the ResumptionSecret is only necessary in case SessionTickets are enabled.


dgarske:
Okay to test

dgarske:
Jenkins retest this please.

PR #9623, opened 2026-01-07
dtls 1.3: allow rtx interval to be less than a second

julek-wolfssl:


julek-wolfssl:
Retest this please RequestAbortedException

tests/api/test_dtls.c
```diff
@@ -2489,3 +2489,78 @@ int test_dtls_mtu_split_messages(void)
...
+
+    /* Drop everything */
+    test_memio_clear_buffer(&test_ctx, 1);
+
+    /* First timeout. This one should trigger a retransmission */
+    if (wolfSSL_dtls13_use_quick_timeout(ssl_s))
+        ExpectIntEQ(wolfSSL_dtls_got_timeout(ssl_s), WOLFSSL_SUCCESS);
+    ExpectIntEQ(wolfSSL_dtls_got_timeout(ssl_s), WOLFSSL_SUCCESS);
+    /* Save the message count to make sure no new messages are sent */
+    ExpectIntGE(c_msg_count = test_ctx.c_msg_count, 2);
```
rizlik: NIT: two instruction to make it more readable

tests/api/test_dtls.c
```diff
@@ -2489,3 +2489,78 @@ int test_dtls_mtu_split_messages(void)
...
+    ExpectIntEQ(wolfSSL_get_error(ssl_s, -1), SSL_ERROR_WANT_READ);
+
+    /* We should have SH ... FINISHED messages in the buffer */
+    ExpectIntGE(test_ctx.c_msg_count, 2);
+
+    /* Drop everything */
+    test_memio_clear_buffer(&test_ctx, 1);
+
+    /* First timeout. This one should trigger a retransmission */
+    if (wolfSSL_dtls13_use_quick_timeout(ssl_s))
```
rizlik, line 2542: wrap the result of the `wolfSSL_dtls13_use_quick_timeout` in an Expect assert

src/dtls13.c
```diff
@@ -110,7 +110,9 @@ typedef struct Dtls13RecordPlaintextHeader {
    supported. */
 #define DTLS13_UNIFIED_HEADER_SIZE 5
 #define DTLS13_MIN_CIPHERTEXT 16
-#define DTLS13_MIN_RTX_INTERVAL 1
+#ifndef DTLS13_MIN_RTX_INTERVAL
+#define DTLS13_MIN_RTX_INTERVAL 1000
```
rizlik: what do you think of using `DTLS_TIMEOUT_INIT` as default here?

tests/api/test_dtls.c
```diff
@@ -2489,3 +2489,78 @@ int test_dtls_mtu_split_messages(void)
     return TEST_SKIPPED;
 #endif
 }
+
+/* Test DTLS 1.3 minimum retransmission interval. This test calls
+ * wolfSSL_dtls_got_timeout() to simulate timeouts and verify that
+ * retransmissions are spaced at least DTLS13_MIN_RTX_INTERVAL apart.
+ */
+int test_dtls13_min_rtx_interval(void)
```
rizlik, line 2499: This test relies on timing  and may introduce flakiness. I don't see an easy way to test without timing dependency. If we want to keep it as it is, we need at least a clear comment about the risk imo.

rizlik (changes requested):

tests/api/test_dtls.c
julek-wolfssl: Why?

tests/api/test_dtls.c
julek-wolfssl: Adding comment.

src/dtls13.c
julek-wolfssl: Changing.

tests/api/test_dtls.c
julek-wolfssl: Split.

tests/api/test_dtls.c
rizlik: To skip it if the previous checks failed.

rizlik (dismissed):

tests/api/test_dtls.c
julek-wolfssl: I don't think it makes sense because it just returns `ssl->dtls13FastTimeout`. The only danger is if `ssl` is `NULL`. I'll fix `wolfSSL_dtls13_use_quick_timeout` to check this.

julek-wolfssl:
Retest this please RequestAbortedException

douzzer (approved):

Issue #9086, opened 2025-08-11
[Bug]: Questions about wolfssl verify


onepeople158:
### Version

I am using version 0.1.7 of the wolfssl Command Line Utility. Linked to wolfSSL version 5.7.6

### Description

Hello developer, I'm not sure if wolfssl has a command for CRL revocation checks. When I perform a revocation check using openssl, openssl indicates that the certificate has been revoked, but wolfssl does not display any information related to the certificate being revoked.

command:
wolfssl verify -CAfile root_cert_1.pem -CRLfile crl_revokedCert_02.pem ca1.pem

test cases:

[isRevoked.zip](https://github.com/user-attachments/files/21714002/isRevoked.zip)

### Relevant log output

```shell

```

kareem-wolfssl:
Hello @onepeople158 ,

How are you building wolfSSL?  You will need to pass in `--enable-crl` to enable CRL support.
Please use wolfSSL 5.8.2 with wolfCLU master, we have had significant changes to both since 0.1.7/5.7.6.

onepeople158:
> Hello [@onepeople158](https://github.com/onepeople158) ,
> 
> How are you building wolfSSL? You will need to pass in to enable CRL support. Please use wolfSSL 5.8.2 with wolfCLU master, we have had significant changes to both since 0.1.7/5.7.6.`--enable-crl`

Hello Developers,

Hello. I have installed the latest version of wolfSSL.

The following is the displayed result:
You are using version 0.1.8 of the wolfssl Command Line Utility.
Linked to wolfSSL version 5.8.2

I ran the command below and am not sure if wolfSSL supports CRL revocation checking:
wolfssl verify -CAfile root_cert_1.pem -CRLfile crl_revokedCert_02.pem ca1.pem

The command output was:
verifying certificate file ca1.pem
using CA file root_cert_1.pem
OK

holtrop:
Hello @onepeople158, I will take a look at this. I see your test case files in the isRevoked.zip file. Can you include the openssl verify commands you are using for comparison?

onepeople158:
> Hello [@onepeople158](https://github.com/onepeople158), I will take a look at this. I see your test case files in the isRevoked.zip file. Can you include the openssl verify commands you are using for comparison?

openssl verify -crl_check -x509_strict -CAfile root_cert_1.pem -CRLfile crl_revokedCert_02.pem ca1.pem
[...]

PR #9628, opened 2026-01-08
Fix CRL Number hex string buffer overflow in CRL parser

miyazakh:
# Description

The `crlNumber` field in the `CRL_Entry` structure stores the CRL Number as a hexadecimal string using `mp_toradix`. However, its buffer size is defined by `CRL_MAX_NUM_SZ`, which represents the maximum CRL Number size in octets, not the hexadecimal string length.

Since each octet requires two hexadecimal characters plus a null terminator, the hexadecimal string buffer size should be defined as `CRL_MAX_NUM_SZ * 2 + 1`.

[Issue#9086](https://github.com/wolfSSL/wolfssl/issues/9086)

# Testing

Add unit test

# Checklist

 - [x] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/wolfssl/internal.h
+++ b/wolfssl/internal.h
@@ -2504,7 +2504,7 @@ struct CRL_Entry {
     /* DupCRL_Entry copies data after the `verifyMutex` member. Using the mutex
      * as the marker because clang-tidy doesn't like taking the sizeof a
      * pointer. */
-    byte    crlNumber[CRL_MAX_NUM_HEX_STR_SZ];    /* CRL number extension */
+    char    crlNumber[CRL_MAX_NUM_HEX_STR_SZ];    /* CRL number extension */
     byte    issuerHash[CRL_DIGEST_SIZE];  /* issuer hash                 */
     /* byte    crlHash[CRL_DIGEST_SIZE];      raw crl data hash           */
     /* restore the hash here if needed for optimized comparisons */


miyazakh:
retest this please

tests/api.c
```diff
@@ -31518,6 +31518,58 @@ static int test_wolfSSL_CTX_LoadCRL(void)
...
+
+    cm = wolfSSL_CertManagerNew();
+    ExpectNotNull(cm);
+    ExpectIntEQ(wolfSSL_CertManagerLoadCA(cm, caCert, NULL),
+        WOLFSSL_SUCCESS);
+    ExpectIntEQ(wolfSSL_CertManagerLoadCRLFile(cm, crl_lrgcrlnum,
+                                                WOLFSSL_FILETYPE_PEM),
+        WOLFSSL_SUCCESS);
+
+    AssertTrue((f = XFOPEN(crl_lrgcrlnum, "rb")) != XBADFILE);
```
padelsbach, line 31545: When should we use `AssertTrue` vs `ExpectTrue`?

wolfssl/wolfcrypt/asn.h
```diff
@@ -2693,6 +2693,7 @@ struct RevokedCert {
 #define CRL_MAX_NUM_SZ 20 /* RFC5280 states that CRL number can be up to 20 */
 #endif                    /* octets long */
 
+#define CRL_MAX_NUM_HEX_STR_SZ (CRL_MAX_NUM_SZ * 2 + 1)
```
padelsbach, line 2696: Should we have include guards around this new definition? Eg `#ifndef CRL_MAX_NUM_HEX_STR_SZ` for consistency with the above?

wolfssl/wolfcrypt/asn.h
```diff
@@ -2693,6 +2693,7 @@ struct RevokedCert {
 #define CRL_MAX_NUM_SZ 20 /* RFC5280 states that CRL number can be up to 20 */
```
padelsbach, line 2693: After this change, it seems `CRL_MAX_NUM_SZ` is mostly used in the form `CRL_MAX_NUM_SZ * CHAR_BIT`. Consider creating a new definition such as:

`#define CRL_MAX_NUM_SZ_BITS (CRL_MAX_NUM_SZ * CHAR_BIT)`

src/crl.c
```diff
@@ -138,7 +138,7 @@ static int InitCRL_Entry(CRL_Entry* crle, DecodedCRL* dcrl, const byte* buff,
     crle->totalCerts = dcrl->totalCerts;
     crle->crlNumberSet = dcrl->crlNumberSet;
     if (crle->crlNumberSet) {
-        XMEMCPY(crle->crlNumber, dcrl->crlNumber, CRL_MAX_NUM_SZ);
+        XMEMCPY(crle->crlNumber, dcrl->crlNumber, CRL_MAX_NUM_HEX_STR_SZ);
```
padelsbach: Consider using `sizeof(crle->crlNumber)` rather than `CRL_MAX_NUM_HEX_STR_SZ` in this type of copy to couple the length with the destination size.

wolfcrypt/src/asn.c
```diff
@@ -40853,7 +40854,24 @@ static int ParseCRL_Extensions(DecodedCRL* dcrl, const byte* buf,
...
+                    /* Check CRL number size
+                     * if it exceeds CRL_MAX_NUM_SZ(octets)
+                     */
+                    if (mp_unsigned_bin_size(m) > CRL_MAX_NUM_SZ) {
+                        WOLFSSL_MSG("CRL number exceeds limitation.");
+                        ret = BUFFER_E;
+                    }
+                    /* Determine required size for hexadecimal string encoding */
+                    if (ret == MP_OKAY &&
+                        (mp_radix_size(m, MP_RADIX_HEX, &needed) != MP_OKAY)) {
```
SparkiDev: Shouldn't need to do this.
The unsigned_bin_size should be directly proportional to the hex size.

wolfssl/wolfcrypt/asn.h
SparkiDev, line 2696: Is this the string encoded size?
Why is the type of crlNumber byte if we are storing the string?

miyazakh: The crlNumber was defined as a byte, so I kept it unchanged. However, I will change it to char to store the string.

wolfcrypt/src/asn.c
miyazakh: Change to calculate the size from `unsigned_bin_size`

src/crl.c
miyazakh: Changed

wolfssl/wolfcrypt/asn.h
miyazakh: Added `CRL_MAX_NUM_SZ_BITS` definition

wolfssl/wolfcrypt/asn.h
miyazakh: I would like to leave only `CRL_MAX_NUM_SZ` configurable by the user, since CRL_MAX_NUM_HEX_STR_SZ can be calculated from `CRL_MAX_NUM_SZ`. Allowing the user to configure `CRL_MAX_NUM_HEX_STR_SZ` as well would make things more complicated.

tests/api.c
miyazakh: As far as I understand, `AssertTrue` is used to stop the test immediately when a condition fails, because continuing the test would not be meaningful. On the other hand, `ExpectTrue` is used when we want to continue testing and report as many failures as possible.

miyazakh:
retest this please

wolfssl/internal.h
```diff
@@ -2504,7 +2504,7 @@ struct CRL_Entry {
     /* DupCRL_Entry copies data after the `verifyMutex` member. Using the mutex
      * as the marker because clang-tidy doesn't like taking the sizeof a
      * pointer. */
-    byte    crlNumber[CRL_MAX_NUM_SZ];    /* CRL number extension */
+    byte    crlNumber[CRL_MAX_NUM_HEX_STR_SZ];    /* CRL number extension */
```
padelsbach: Should this become a `char` as well?

miyazakh: Yes. Thanks.

miyazakh:
retest this please

PR #9605, opened 2026-01-05
Build systems improvements

Frauschi:
Various build system fixes, mainly for CMake.

* Add `WOLFSSL_USER_SETTINGS` to CMake `options.h.in`
* Add CMake support for Dilithium
* Add `user_settings.h` support for `aes_asm.S`
* Add PKCS#11 support to CMake
* Minor ARM assembly port fixes

PR #9637, opened 2026-01-09
Increase test coverage for PQC and CMake

Frauschi:
Increase test coverage, mainly for stuff related to #9605 and #9608 

1. Add more PQC configurations to the CI tests

These are mainly `SMALL` configurations or some edge cases, which have been missing up to now but already caused some build failures (e.g. #9608).

2. More CMake setups tested

Added the functionality added in #9605 to the existing CMake CI tests. Also added a new test to verify the `user_settings.h` functionality with CMake. For that, some missing features have been added to CMake to make the example `example/configs/user_settings_all.h` config file work for the CI test.


Frauschi:
retest this please

Frauschi:
I think the remaining failing CI tests are unrelated to this PR.

dgarske:
Jenkins retest this please

Frauschi:
Jenkins retest this please

cmake/functions.cmake
```diff
@@ -942,21 +963,17 @@ function(generate_lib_src_list LIB_SOURCES)
         endif()
 
         if(BUILD_FEMATH)
-            if(BUILD_CURVE25519_SMALL)
```
douzzer, line 945: Hmmmm, why are you removing this gate?  On non-small builds, this will result in empty translation units, which are warned on pedantic builds, and errors on `-pedantic -Werror` builds.

cmake/functions.cmake
```diff
@@ -942,21 +963,17 @@ function(generate_lib_src_list LIB_SOURCES)
...
                 if(BUILD_INTELASM)
                     list(APPEND LIB_SOURCES wolfcrypt/src/fe_x25519_asm.S)
                 else()
                     list(APPEND LIB_SOURCES wolfcrypt/src/fe_operations.c)
                 endif()
-            endif()
         endif()
 
         if(BUILD_GEMATH)
-            if(BUILD_ED25519_SMALL)
```
douzzer, line 957: Same question as above.

.github/workflows/pq-all.yml
```diff
@@ -19,9 +19,14 @@ jobs:
...
+          '--enable-intelasm --enable-sp-asm --enable-all --enable-testcert --enable-acert --enable-dtls13 --enable-dtls-mtu --enable-dtls-frag-ch --enable-dtlscid --enable-quic --with-sys-crypto-policy --enable-experimental --enable-mlkem=yes,kyber,ml-kem --enable-lms --enable-xmss --enable-dilithium --enable-dual-alg-certs --disable-qt CPPFLAGS="-Wdeclaration-after-statement -DWOLFCRYPT_TEST_LINT -DNO_WOLFSSL_CIPHER_SUITE_TEST -DTEST_LIBWOLFSSL_SOURCES_INCLUSION_SEQUENCE" CC=c++',
+          '--enable-intelasm --enable-sp-asm --enable-all --enable-testcert --enable-acert --enable-dtls13 --enable-dtls-mtu --enable-dtls-frag-ch --enable-dtlscid --enable-quic --with-sys-crypto-policy --enable-experimental --enable-mlkem=yes,kyber,ml-kem --enable-lms --enable-xmss --enable-dilithium --disable-qt CPPFLAGS="-pedantic -Wdeclaration-after-statement -DWOLFCRYPT_TEST_LINT -DNO_WOLFSSL_CIPHER_SUITE_TEST -DTEST_LIBWOLFSSL_SOURCES_INCLUSION_SEQUENCE -DWOLFSSL_BLIND_PRIVATE_KEY"',
```
douzzer, line 25: Does it really add meaningful coverage to test a scenario identical to the earlier one, but without `--enable-dual-alg-certs`?

.github/workflows/pq-all.yml
```diff
@@ -19,9 +19,14 @@ jobs:
...
+          '--enable-intelasm --enable-sp-asm --enable-all --enable-testcert --enable-acert --enable-dtls13 --enable-dtls-mtu --enable-dtls-frag-ch --enable-dtlscid --enable-quic --with-sys-crypto-policy --enable-experimental --enable-mlkem=yes,kyber,ml-kem --enable-lms --enable-xmss --enable-dilithium --enable-dual-alg-certs --disable-qt CPPFLAGS="-pedantic -Wdeclaration-after-statement -DWOLFCRYPT_TEST_LINT -DNO_WOLFSSL_CIPHER_SUITE_TEST -DTEST_LIBWOLFSSL_SOURCES_INCLUSION_SEQUENCE -DWOLFSSL_BLIND_PRIVATE_KEY"',
```
douzzer, line 26: Should `-DWOLFSSL_BLIND_PRIVATE_KEY` just be added to the existing scenarios above?  Or do you think it's meaningful to test with and without it?

.github/workflows/pq-all.yml
```diff
@@ -19,9 +19,14 @@ jobs:
...
+          '--enable-intelasm --enable-sp-asm --enable-all --enable-testcert --enable-acert --enable-dtls13 --enable-dtls-mtu --enable-dtls-frag-ch --enable-dtlscid --enable-quic --with-sys-crypto-policy --enable-experimental --enable-mlkem=yes,kyber,ml-kem --enable-lms --enable-xmss --enable-dilithium --enable-dual-alg-certs --disable-qt CPPFLAGS="-pedantic -Wdeclaration-after-statement -DWOLFCRYPT_TEST_LINT -DNO_WOLFSSL_CIPHER_SUITE_TEST -DTEST_LIBWOLFSSL_SOURCES_INCLUSION_SEQUENCE -DWOLFSSL_TLSX_PQC_MLKEM_STORE_OBJ"',
```
douzzer, line 27: Ditto for `-DWOLFSSL_TLSX_PQC_MLKEM_STORE_OBJ`.

douzzer (changes requested):
Minor questions+quibbles.

Looks good otherwise.

.github/workflows/pq-all.yml
Frauschi: Yes, I think so. Both the `dual-alg-certs` option as well as the `WOLFSSL_BLIND_PRIVATE_KEY` option enable new code paths but also disable others. Hence, enabling and disabling both individually compared to the base case results in different code paths, which are not covered otherwise in the tests.

.github/workflows/pq-all.yml
Frauschi: See comment above for `dual-alg-certs`.

.github/workflows/pq-all.yml
Frauschi: This option could also be added to one of the new cases above. However, in this scenario, when the test then fails, one has to check whether the fail results from the `WOLFSSL_TLSX_PQC_MLKEM_STORE_OBJ` option or from the other one to where it would be added.

cmake/functions.cmake
Frauschi: The motivation to remove the gate was that when using the user_settings.h approach with CMake, both the `fe_low_mem.c` and the `fe_operations.c` files are required to succeed the build. In this case, pedantic builds are impossible anyway as the user_settings.h approach uses empty translation units throughout.

I re-added the gates but added logic to also build the small source files in case `WOLFSSL_USER_SETTINGS` is set. Hence, empty translation units are only caused when the user_settings.h approach is used.

cmake/functions.cmake
Frauschi: See above

Frauschi:
> Minor questions+quibbles.

I addressed the CMake gate thing in the updated commit. The options in the PQ build are beneficial in my view, see comments.

Frauschi:
Retest this please

Frauschi:
Jenkins restest this please

PR #9566, opened 2025-12-19
Added build option to allow certificate CA matching using AKID with signers SKID

dgarske:
# Description

Added build option to allow certificate CA matching using AKID with signers SKID (`WOLFSSL_ALLOW_AKID_SKID_MATCH`). Fixed issue with `cert->extAuthKeyIdSz` not being set with ASN template code.

Fixes issue with validation of ST33KTPM certificate chain where one uses UTF8STRING and other uses PRINTABLE_STRING in common name but AKID/SKID match.

# Testing

```
./configure && make
./configure CFLAGS="-DWOLFSSL_NO_ASN_STRICT" && make
./configure CFLAGS="-DWOLFSSL_AKID_NAME" && make
./configure CFLAGS="-DWOLFSSL_AKID_NAME -DWOLFSSL_NO_ASN_STRICT" && make
./configure CFLAGS="-DWOLFSSL_AKID_NAME -DWOLFSSL_NO_ASN_STRICT -DNO_SKID" && make
./configure CFLAGS="-DWOLFSSL_AKID_NAME -DWOLFSSL_NO_ASN_STRICT" && make
./configure --enable-asn=original CFLAGS="-DWOLFSSL_AKID_NAME -DWOLFSSL_NO_ASN_STRICT" && make
./configure --enable-asn=original CFLAGS="-DWOLFSSL_NO_ASN_STRICT" && make
```
Tested using https://github.com/wolfSSL/wolfssl-examples/blob/master/certmanager/certverify.c

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen
 - [ ] updated appropriate READMEs
 - [ ] Updated manual and documentation

PR #9654, opened 2026-01-13
Fix DecodeAuthKeyInternal not clearing cert->extAuthKeyIdSet

embhorn:
# Description

Prior to #9566 , cert->extAuthKeyIdSet was cleared if extAuthKeyIdSz was zero. Currently, cert->extAuthKeyIdSet is not cleared under the same condition.

Fixes zd21038

# Testing

Customer confirmed

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


dgarske (approved):
Thanks, I believe that was my error.

PR #9658, opened 2026-01-14
   cryptocb: add AES CryptoCB key import support and tests

sameehj:
Add CryptoCB-based AES key import support to enable Secure Element
    offload without exposing raw AES key material to wolfCrypt.
    
    This change introduces a new optional CryptoCB hook
    (WOLF_CRYPTO_CB_AES_SETKEY) that allows AES keys to be imported into
    external devices (e.g. Secure Elements or HSMs) and referenced via an
    opaque handle stored in aes->devCtx. When this mode is active, wolfCrypt
    stores only key metadata and routes AES-GCM operations through CryptoCB,
    bypassing software key storage and GCM table generation.
    
    Key points:
    - Add wc_CryptoCb_AesSetKey() callback for AES key import
    - Update AES SetKey paths to support key import mode with graceful
      fallback to software when CryptoCB is unavailable
    - Skip GCM H/M table generation when AES-GCM is handled by the device
    - Preserve existing software AES behavior when devId is INVALID_DEVID
    
    Testing:
    - Add unit test for CryptoCB AES SetKey behavior
    - Add end-to-end AES-GCM offload unit test that verifies:
      * SetKey, Encrypt, Decrypt, and Free are routed via CryptoCB
      * Correct ciphertext/auth tag generation
      * Correct plaintext recovery after decrypt
      * Proper lifecycle handling of device context handles
    - Tests use a mock Secure Element that internally performs software AES
      to validate routing without requiring hardware


sameehj:
retest this please

wolfssl/wolfcrypt/cryptocb.h
```diff
@@ -533,7 +540,7 @@ WOLFSSL_LOCAL void wc_CryptoCb_Init(void);
 WOLFSSL_LOCAL void wc_CryptoCb_Cleanup(void);
 WOLFSSL_LOCAL int wc_CryptoCb_GetDevIdAtIndex(int startIdx);
 WOLFSSL_API int  wc_CryptoCb_RegisterDevice(int devId, CryptoDevCallbackFunc cb, void* ctx);
-WOLFSSL_API void wc_CryptoCb_UnRegisterDevice(int devId);
+WOLFSSL_API int wc_CryptoCb_UnRegisterDevice(int devId);
```
julek-wolfssl: Why?

wolfssl/wolfcrypt/cryptocb.h
```diff
@@ -678,6 +685,30 @@ WOLFSSL_LOCAL int wc_CryptoCb_AesEcbEncrypt(Aes* aes, byte* out,
...
+ * generating GCM tables in software.
+ *
+ * \param aes     AES context
+ * \param key     Pointer to raw AES key material
+ * \param keySz   Size of key in bytes
+ *
+ * \return 0 on success
+ * \return CRYPTOCB_UNAVAILABLE if device does not support this operation
+ * \return BAD_FUNC_ARG on invalid parameters
+ */
```
julek-wolfssl: Wrong place for docs.

tests/api/test_aes.c
```diff
@@ -5222,3 +5222,735 @@ int test_wc_AesEaxDecryptAuth(void)
...
+#if defined(WOLF_CRYPTO_CB) && defined(WOLF_CRYPTO_CB_AES_SETKEY) && \
+    !defined(NO_AES) && defined(HAVE_AESGCM)
+
+#include <wolfssl/wolfcrypt/cryptocb.h>
+
+#define TEST_CRYPTOCB_AES_DEVID  7
+
+/* Test state tracking */
+static int g_ccAesSetKeyCalled = 0;
+static int g_ccAesFreeCalled = 0;
```
julek-wolfssl: What is this `g_` prefix?

wolfcrypt/src/aes.c
```diff
@@ -4785,6 +4808,37 @@ static void AesSetKey_C(Aes* aes, const byte* key, word32 keySz, int dir)
...
+                return ret;
+            }
+            /* CRYPTOCB_UNAVAILABLE: fall through to software */
+        #else
+            /* ORIGINAL: Copy key to devKey for existing CryptoCB users */
+            if (keylen > sizeof(aes->devKey)) {
+                return BAD_FUNC_ARG;
+            }
+            XMEMCPY(aes->devKey, userKey, keylen);
+        #endif
```
julek-wolfssl: Why move this up here instead of allowing `aes->devId != INVALID_DEVID` section to handle it?

wolfcrypt/src/aes.c
```diff
@@ -7494,7 +7574,20 @@ int wc_AesGcmSetKey(Aes* aes, const byte* key, word32 len)
...
+        if (aes->devCtx != NULL) {
+            /* Skip - key handled by CryptoCB */
+        }
+        else
+    #endif
+        {
+            /* ORIGINAL: Copy to devKey */
+            if (len > sizeof(aes->devKey)) {
+                return BAD_FUNC_ARG;
+            }
```
julek-wolfssl: `len` is already checked on function entry. Why this check here?

wolfcrypt/src/aes.c
```diff
@@ -13386,19 +13475,15 @@ void wc_AesFree(Aes* aes)
     if (aes->devId != INVALID_DEVID)
     #endif
     {
-        ret = wc_CryptoCb_Free(aes->devId, WC_ALGO_TYPE_CIPHER,
-                         WC_CIPHER_AES, (void*)aes);
-        /* If they want the standard free, they can call it themselves */
-        /* via their callback setting devId to INVALID_DEVID */
-        /* otherwise assume the callback handled it */
```
julek-wolfssl, line 13403: Why is this comment being removed?

julek-wolfssl:
Is it possible to setup a gh action that tests this with wolfHSM? At the very least this new config should be added to `os-check.yml`.

sameehj:
> I see "proxy-key" appearing multiple times in the PR. What does it mean? The `ORIGINAL:` comments should be cleaned up.

Dropped proxy-key terminology
and dropped 'ORIGINAL" comments

wolfssl/wolfcrypt/cryptocb.h
sameehj: Mainly to satisfy Devin's comment on the PR, but maybe we shouldn't break the API to do so.
Thank you

wolfcrypt/src/aes.c
sameehj: ack, dropped

wolfcrypt/src/aes.c
sameehj: you are right sir, readded

douzzer:
retest this please

sameehj:
retest this please

wolfcrypt/src/aes.c
```diff
@@ -7455,14 +7501,27 @@ int wc_AesGcmSetKey(Aes* aes, const byte* key, word32 len)
...
+        }
     }
     if (ret == 0) {
+#ifdef WOLF_CRYPTO_CB_AES_SETKEY
+        if (aes->devId != INVALID_DEVID && aes->devCtx != NULL) {
+            /* SE owns key - skip M0 table generation */
+        }
+        else
+#endif
+        {
```
douzzer: We just merged a fix in this area that conflicts.  Here's how I reconciled your branch while reviewing:
```
@@ -7454,6 +7500,14 @@ int wc_AesGcmSetKey(Aes* aes, const byte* key, word32 len)
     }
 #else
 #if !defined(FREESCALE_LTC_AES_GCM) && !defined(WOLFSSL_PSOC6_CRYPTO)
+
+
+#ifdef WOLF_CRYPTO_CB_AES_SETKEY
+    if ((ret == 0) && (aes->devId != INVALID_DEVID && aes->devCtx != NULL)) {
+        /* SE owns key - skip H and M table generation */
+    }
+    else
+#endif
     if (ret == 0) {
         VECTOR_REGISTERS_PUSH;
         /* AES-NI code generates its own H value, but generate it here too, to
```

wolfcrypt/src/aes.c
```diff
@@ -7490,6 +7549,7 @@ int wc_AesGcmSetKey(Aes* aes, const byte* key, word32 len)
             GenerateM0(&aes->gcm);
         }
 #endif /* GCM_TABLE || GCM_TABLE_4BIT */
+        }
```
douzzer, line 7548: This close-curly needs to be removed as part of the reconciliation.

wolfssl/wolfcrypt/aes.h
```diff
@@ -143,6 +143,7 @@ WOLFSSL_LOCAL void GHASH(Gcm* gcm, const byte* a, word32 aSz, const byte* c,
 #include <psa/crypto.h>
 #endif
 
+
```
douzzer: Don't need this whitespace change.

douzzer (changes requested):
this implementation looks really strong.

I left comments with reconciliation suggestions for the conflicts, and a comment on a stray linefeed that snuck in.  otherwise, LGTM!

sameehj:
retest this please

douzzer (approved):
`wolfssl-multi-test.sh ... pr-check`:
```
final tally for 425dc1372d with build env c2c7829175: all 51 selected checks succeeded.
```

PR #9668, opened 2026-01-14
PQ FS 2026 part1

kaleb-himes:
# Description

Replace the DRBG legacy CRNGT test with the more modern RCT/ADP tests.

CRNGT: Simply compares each block of output from the entropy source with each previous block of output to detect a "stuck entropy source" but suffers a probability failure rate of ~29.80 failures for every 4 billion checks. (keeping in mind a single false-positive means the module must be power cycled to recover from this catastrophic scenario per FIPS 140-2 ss 4.9.2) which leads to high volumes of real-world failures in long running applications (weeks or months under moderate load, days or weeks under heavy load).

RCT/ADP - The Repetition Count Test (RCT) acts as a "stuck-at" detector by failing if a single value repeats consecutively more than a calculated threshold, effectively catching total hardware failures. The Adaptive Proportion Test (ADP) functions as a "bias" detector by counting how many times a specific value appears within a sliding window of samples, identifying subtle statistical collapses that the RCT would otherwise miss.

Fixes issue reported in zd# 21032 for future modules (does not fix legacy modules)

# Testing

There is a new test (but it's off by default because it is very slow) in wolfcrypt/test/test.c with two flavors for the Re-Init test, a single threaded and a multi-threaded test.

Single threaded takes a long time but will produce the same rate of failures as the multi-threaded test if entropy pool depletion is not a concern. If entropy pool is small the single threaded option is the best way but it takes longer.

The multi threaded test completes faster by using 40 threads and 100Million iterations per thread to hit the 4 billion goal more quickly. Entropy depletion is a concern so only use this on modern large scale desktop and servers, not on embedded linux.

Examples of configuring and running the tests with and without FIPS enabled are in wolfcrypt/test/README.md along with the investigation results into this issue.

# Checklist

 - [X] added tests
 - [X] updated/added doxygen
 - [X] updated appropriate READMEs

### DIFF
--- a/wolfcrypt/src/random.c
+++ b/wolfcrypt/src/random.c
@@ -794,7 +794,7 @@ int wc_RNG_TestSeed(const byte* seed, word32 seedSz)
     int rctFailed = 0;
     int aptFailed = 0;
 
-    if (seed == NULL || seedSz == 0) {
+    if (seed == NULL || seedSz < SEED_BLOCK_SZ) {
         return BAD_FUNC_ARG;
     }
 


SparkiDev:
conflicts on bracnh

kaleb-himes:
Re-synced with master, conflicts resolved.

PR #6747, opened 2023-09-01
ci: Add CI for different archs

res0nance:
# Description

Adds CI for different architectures

This should also hopefully prevent issues such as https://github.com/wolfSSL/wolfssl/issues/5354

# Testing

CI Tested

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen
 - [ ] updated appropriate READMEs
 - [ ] Updated manual and documentation

PR #9673, opened 2026-01-15
Update from Ubuntu 22.04 to Ubuntu 24.04 for github workflows

holtrop-wolfssl:
# Description

Update from Ubuntu 22.04 to Ubuntu 24.04 for several github workflows

This does not update all of them, but the vast majority. There are a few left that need more investigation that I will get to as I can.

# Testing

How did you test?

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/.github/workflows/async.yml
+++ b/.github/workflows/async.yml
@@ -24,7 +24,7 @@ jobs:
         ]
     name: make check
     if: github.repository_owner == 'wolfssl'
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     # This should be a safe limit for the tests to run.
     timeout-minutes: 6
     steps:
--- a/.github/workflows/bind.yml
+++ b/.github/workflows/bind.yml
@@ -17,7 +17,7 @@ jobs:
     name: Build wolfSSL
     if: github.repository_owner == 'wolfssl'
     # Just to keep it the same as the testing target
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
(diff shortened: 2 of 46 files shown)


padelsbach:
Looks like `multi-arch.yml` is the last one using `runs-on: ubuntu-22.04`. What's the limitation there?

holtrop-wolfssl:
retest this please (org.jenkinsci.plugins.workflow.support.steps.AgentOfflineException: Unable to create live FilePath for wolf-linux-cloud-node-q83jgt; wolf-linux-cloud-node-q83jgt was marked offline: Connection was broken)

holtrop-wolfssl:
> Looks like `multi-arch.yml` is the last one using `runs-on: ubuntu-22.04`. What's the limitation there?

I don't remember what the issue was with that one specifically, but there are a few that required a little more investigation. That one might have triggered new sanitizer/analyzer outputs due to newer compiler versions. I will continue to investigate as I have time.

dgarske:
> > Looks like `multi-arch.yml` is the last one using `runs-on: ubuntu-22.04`. What's the limitation there?
> 
> I don't remember what the issue was with that one specifically, but there are a few that required a little more investigation. That one might have triggered new sanitizer/analyzer outputs due to newer compiler versions. I will continue to investigate as I have time.

Added by third-party in https://github.com/wolfSSL/wolfssl/pull/6747. For now I would leave this one set at 22.04. Likely due to availability of packages for the required cross compiler emulators.

PR #9681, opened 2026-01-19
Fix cert SW issues in Aes and rng

tmael:
# Description

Adds checks for the validity of the seed buffer and the seedSz parameter. When seedSz is less than SEED_BLOCK_SZ (Word32 = 4), the loop can underflow and iterate to the Word32 limit, resulting in an infinite loop and a crash.

Updates wc_AesEncryptDirect to return a negative error code when the input data buffer is NULL. Previously, the function returned 0 and exhibited undefined behavior.

Fixed in https://github.com/wolfSSL/wolfBoot-private/pull/70

# Testing

How did you test?

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/wolfcrypt/src/random.c
+++ b/wolfcrypt/src/random.c
@@ -754,7 +754,7 @@ int wc_RNG_TestSeed(const byte* seed, word32 seedSz)
     word32 seedIdx = 0;
     word32 scratchSz = 0;
 
-    if (seed == NULL || seedSz < sizeof(word32))
+    if (seed == NULL || seedSz < SEED_BLOCK_SZ)
         return BAD_FUNC_ARG;
 
     scratchSz = min(SEED_BLOCK_SZ, seedSz - SEED_BLOCK_SZ);


wolfcrypt/src/random.c
```diff
@@ -752,12 +752,18 @@ int wc_RNG_TestSeed(const byte* seed, word32 seedSz)
 
     /* Check the seed for duplicate words. */
     word32 seedIdx = 0;
-    word32 scratchSz = min(SEED_BLOCK_SZ, seedSz - SEED_BLOCK_SZ);
+    word32 scratchSz = 0;
+
+    if (seed == NULL || seedSz < sizeof(word32))
+        return BAD_FUNC_ARG;
```
rizlik, line 758: are you sure this should be sizeof(word32) instead of SEED_BLOCK_SZ?

tmael: @rizlik, you are right, it is best to use < SEED_BLOCK_SZ since the configuration can change in non-cert case

dgarske (approved):

rizlik (approved):

tmael:
Jenkins retest this please

tmael:
Jenkins retest this please

dgarske:
Jenkins retest this please: "RequestAbortedException"

PR #9687, opened 2026-01-20
Rust wrapper: add HMAC-BLAKE2[bs] wrappers

holtrop-wolfssl:
# Description

Add Rust wrapper for HMAC-BLAKE2[bs] one-shot and incremental APIs.

Also attempted to avoid a couple coverity messages about overrunning the x_key array (false positives).

# Testing

Unit tests and integration with boringtun

# Checklist

 - [X] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/wrapper/rust/wolfssl-wolfcrypt/src/blake2.rs
+++ b/wrapper/rust/wolfssl-wolfcrypt/src/blake2.rs
@@ -176,6 +176,9 @@ pub struct BLAKE2bHmac {
 
 #[cfg(blake2b_hmac)]
 impl BLAKE2bHmac {
+    /// HMAC-BLAKE2b digest size.
+    pub const DIGEST_SIZE: usize = sys::WC_BLAKE2B_DIGEST_SIZE as usize;
+
     /// Build a new BLAKE2bHmac instance.
     ///
     /// # Parameters
@@ -264,7 +267,7 @@ impl BLAKE2bHmac {
     /// let mut mac = [0u8; 64];
     /// hmac_blake2b.finalize(&key, &mut mac).expect("Error with finalize()");
     /// ```
-    pub fn finalize(&mut self, key: &[u8], mac: &mut [u8]) -> Result<(), i32> {
+    pub fn finalize(&mut self, key: &[u8], mac: &mut [u8; Self::DIGEST_SIZE]) -> Result<(), i32> {
         let rc = unsafe {
             sys::wc_Blake2bHmacFinal(&mut self.wc_blake2b,
                 key.as_ptr(), key.len(), mac.as_mut_ptr(), mac.len())
@@ -282,13 +285,14 @@ impl BLAKE2bHmac {
     ///
     /// * `data`: Input data to create MAC from.
     /// * `key`: Key to use for MAC creation.
-    /// * `out`: Buffer in which to store the computed MAC.
+    /// * `out`: Buffer in which to store the computed MAC. It must be 64 bytes
+    ///   long.
     ///
     /// # Returns
     ///
     /// Returns either Ok(()) on success or Err(e) containing the wolfSSL
     /// library error code value.
-    pub fn hmac(data: &[u8], key: &[u8], out: &mut [u8]) -> Result<(), i32> {
+    pub fn hmac(data: &[u8], key: &[u8], out: &mut [u8; Self::DIGEST_SIZE]) -> Result<(), i32> {
         let rc = unsafe {
             sys::wc_Blake2bHmac(data.as_ptr(), data.len(), key.as_ptr(),
                 key.len(), out.as_mut_ptr(), out.len())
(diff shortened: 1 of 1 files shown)


holtrop-wolfssl:
retest this please (org.jenkinsci.plugins.workflow.support.steps.AgentOfflineException: Unable to create live FilePath for wolf-linux-cloud-node-2zcwho; wolf-linux-cloud-node-2zcwho was marked offline: Connection was broken)

dgarske (dismissed):

wrapper/rust/wolfssl-wolfcrypt/src/blake2.rs
```diff
@@ -167,6 +167,141 @@ impl BLAKE2b {
...
+    ///
+    /// * `data`: Input data to create MAC from.
+    /// * `key`: Key to use for MAC creation.
+    /// * `out`: Buffer in which to store the computed MAC.
+    ///
+    /// # Returns
+    ///
+    /// Returns either Ok(()) on success or Err(e) containing the wolfSSL
+    /// library error code value.
+    #[cfg(blake2b_hmac)]
```
gasbytes: I think this is not necessary, since the whole `impl` block is gated (line 177).

wrapper/rust/wolfssl-wolfcrypt/src/blake2.rs
```diff
@@ -305,3 +440,137 @@ impl BLAKE2s {
...
+    ///
+    /// * `data`: Input data to create MAC from.
+    /// * `key`: Key to use for MAC creation.
+    /// * `out`: Buffer in which to store the computed MAC.
+    ///
+    /// # Returns
+    ///
+    /// Returns either Ok(()) on success or Err(e) containing the wolfSSL
+    /// library error code value.
+    #[cfg(blake2s_hmac)]
```
gasbytes: I think this is not necessary, since the whole `impl` block is gated (line 451).

wrapper/rust/wolfssl-wolfcrypt/src/blake2.rs
```diff
@@ -167,6 +167,141 @@ impl BLAKE2b {
...
+    /// ```rust
+    /// use wolfssl_wolfcrypt::blake2::BLAKE2bHmac;
+    /// let key = [42u8, 43, 44];
+    /// let mut hmac_blake2b = BLAKE2bHmac::new(&key).expect("Error with new()");
+    /// let data = [33u8, 34, 35];
+    /// hmac_blake2b.update(&data).expect("Error with update()");
+    /// let mut mac = [0u8; 64];
+    /// hmac_blake2b.finalize(&key, &mut mac).expect("Error with finalize()");
+    /// ```
+    pub fn finalize(&mut self, key: &[u8], mac: &mut [u8]) -> Result<(), i32> {
```
gasbytes: From the docs it says "must be 64 bytes", but there's no check to enforce that.
Maybe you can turn the mac argument to be `&mut [u8; 64]`, or at least document what happens if the length is not respected once the buffer is passed to the C api.

This applies to the `finalize` function in the HMAC-BLAKE2s module too, but with 32 bytes instead.

wrapper/rust/wolfssl-wolfcrypt/src/blake2.rs
```diff
@@ -167,6 +167,141 @@ impl BLAKE2b {
...
+    /// * `data`: Input data to create MAC from.
+    /// * `key`: Key to use for MAC creation.
+    /// * `out`: Buffer in which to store the computed MAC.
+    ///
+    /// # Returns
+    ///
+    /// Returns either Ok(()) on success or Err(e) containing the wolfSSL
+    /// library error code value.
+    #[cfg(blake2b_hmac)]
+    pub fn hmac(data: &[u8], key: &[u8], out: &mut [u8]) -> Result<(), i32> {
```
gasbytes: The out parameter doesn't document or enforce its expected size (64 bytes for BLAKE2b, 32 bytes for BLAKE2s). Maybe turn it to `&mut [u8; 64]` for compile-time enforcement, or at least document what happens if the length is not respected once the buffer is passed to the C API. This applies to both `BLAKE2bHmac::hmac()` and `BLAKE2sHmac::hmac()`.

gasbytes (changes requested):

wrapper/rust/wolfssl-wolfcrypt/src/blake2.rs
holtrop-wolfssl: Good idea; I added an exact length requirement.

wrapper/rust/wolfssl-wolfcrypt/src/blake2.rs
holtrop-wolfssl: Good idea; I added an exact length requirement.

wrapper/rust/wolfssl-wolfcrypt/src/blake2.rs
holtrop-wolfssl: Ah yes, thanks. I originally had the one-shot API in the BLAKE2b impl block instead of making a separate BLAKE2bHmac struct. Removed!

wrapper/rust/wolfssl-wolfcrypt/src/blake2.rs
holtrop-wolfssl: Removed

holtrop-wolfssl:
retest this please (org.jenkinsci.plugins.workflow.support.steps.AgentOfflineException: Unable to create live FilePath for wolf-linux-cloud-node-q02z3c; wolf-linux-cloud-node-q02z3c was marked offline: Connection was broken)

PR #9699, opened 2026-01-21
Add cipher suite filtering when downgrade is disabled

anhu:
When wolfSSL_SetVersion() is called to set a specific TLS version, the downgrade flag is now set to 0. This causes wolfSSL_parse_cipher_list() to no longer preserve cipher suites from the other TLS version group.

Previously, when using SSLv23 method and setting cipher suites for only one TLS version (e.g., TLS 1.2), the library would preserve any existing cipher suites from the other version (e.g., TLS 1.3) for OpenSSL API compatibility. With this change, if a specific version is set via wolfSSL_SetVersion(), only the cipher suites for that version are kept.

Fixes ZD 21068

### DIFF
--- a/tests/api.c
+++ b/tests/api.c
@@ -2055,6 +2055,7 @@ static int test_wolfSSL_CTX_set_cipher_list_bytes(void)
 #if defined(OPENSSL_EXTRA) && defined(WOLFSSL_TLS13) && \
     !defined(WOLFSSL_NO_TLS12) && \
     !defined(NO_WOLFSSL_CLIENT) && \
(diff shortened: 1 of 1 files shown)


anhu:
jenkins retest this please.

anhu:
Jenkins retest this please.

anhu:
jenkins retest this please.

anhu:
Jenkins, retest this please. 

testing output was expired.

anhu:
jenkins retest this please. 

anhu:
@SparkiDev , I think this is ready now.  Please merge.

PR #9707, opened 2026-01-23
Add STM32G0 hardware crypto support

danielinux:
Enables support for AES crypto accelerator in port/, on STM32G0, per ST RM0444 ch. 20.


dgarske:
Jenkins retest this please: Generic config "AgentOfflineException"

PR #9070, opened 2025-08-06
Support sha224 cryptcb

miyazakh:
# Description

Add sha224 cryptocb

# Testing

Run crypto test with `--enable-cryptocb` 

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen
 - [ ] updated appropriate READMEs
 - [ ] Updated manual and documentation

PR #9712, opened 2026-01-24
Fix Crash when using Sha224 Callback with MAX32666

night1rider:
Found that when using sha224 with callbacks on program/system would crash due to missing initialization. 

Update MAX32666 to use Copy and Free Callbacks as well, with some general improvement to copy and free logic.

Crash bug is present in v5.8.4, but not previous stable v5.8.2. Was found when adding https://github.com/wolfSSL/wolfssl/pull/9070 specifically commit 07b3695b983d2406ed02c921c02e26cf1f52bfb4 which added callback for sha224.

Also found that when commit 9c1462a9ecb5106ba7de9e257274c0ef232b86ed entered, there was no initialization for the callback structure and the MXC init call when the new wc_sha256_init_ex was added.

### DIFF
--- a/wolfcrypt/src/port/maxim/max3266x.c
+++ b/wolfcrypt/src/port/maxim/max3266x.c
@@ -204,6 +204,133 @@ int wc_MxcShaCryptoCb(wc_CryptoInfo* info)
 }
 #endif /* MAX3266X_SHA_CB */
 
+#ifdef WOLF_CRYPTO_CB_COPY
+static int wc_MxcCopyCb(wc_CryptoInfo* info)
+{
+    if (info == NULL || info->copy.src == NULL || info->copy.dst == NULL) {
+        return BAD_FUNC_ARG;
+    }
+
+    switch (info->copy.type) {
+#ifdef MAX3266X_SHA_CB
+    #ifndef NO_SHA
+        case WC_HASH_TYPE_SHA:
+            return wc_MXC_TPU_SHA_Copy(info->copy.src, info->copy.dst,
+                        sizeof(wc_Sha),
+                        &((wc_Sha*)info->copy.dst)->msg,
+                        &((wc_Sha*)info->copy.dst)->used,
+                        &((wc_Sha*)info->copy.dst)->len,
+                        ((wc_Sha*)info->copy.dst)->heap,
+                        ((wc_Sha*)info->copy.src)->heap);
+    #endif
+    #ifdef WOLFSSL_SHA224
+        case WC_HASH_TYPE_SHA224:
+            return wc_MXC_TPU_SHA_Copy(info->copy.src, info->copy.dst,
+                        sizeof(wc_Sha224),
+                        &((wc_Sha224*)info->copy.dst)->msg,
+                        &((wc_Sha224*)info->copy.dst)->used,
+                        &((wc_Sha224*)info->copy.dst)->len,
+                        ((wc_Sha224*)info->copy.dst)->heap,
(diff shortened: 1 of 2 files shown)


night1rider:
Jenkins retest this please

Jenkins disconnect from server during PRB-generic-config-parser

night1rider:
Jenkins retest this please

wolfcrypt/src/port/maxim/max3266x.c
```diff
@@ -29,6 +29,7 @@
 
 #include <stdint.h>
 #include <stdarg.h>
+#include <stdio.h>
```
JacobBarthelmeh: Is the stdio include really needed here?

wolfcrypt/src/port/maxim/max3266x.c
```diff
@@ -708,31 +844,58 @@ int wc_MXC_TPU_SHA_Copy(wc_MXC_Sha* src, wc_MXC_Sha* dst)
...
-        /* Allocate new memory for dst->msg if it points to the same location */
-        /* as src->msg */
-        dst->msg = (unsigned char*)XMALLOC(src->size, NULL,
-                                            DYNAMIC_TYPE_TMP_BUFFER);
-        if (dst->msg == NULL) {
-            return MEMORY_E; /* Handle memory allocation failure */
+
+    /* Handle case where src has no data */
+    if (src->msg == NULL || src->size == 0) {
+        /* Free dst if it has different data, then zero it */
```
JacobBarthelmeh: The comment says "free then zero" but it looks like it conditionally frees here if not null and pointers are not the same, or it zeros if NULL or the pointers are the same.

wolfcrypt/src/port/maxim/max3266x.c
```diff
@@ -708,31 +844,58 @@ int wc_MXC_TPU_SHA_Copy(wc_MXC_Sha* src, wc_MXC_Sha* dst)
...
 
 /* Free the given struct's msg buffer and then reinitialize the struct to 0 */
 /* returns void to match other wc_Sha*Free api */
 void wc_MXC_TPU_SHA_Free(wc_MXC_Sha* hash)
 {
-    if (hash == NULL) {
-        return; /* Hash Struct is Null already, dont edit potentially */
-                /* undefined memory */
+    /* Securely zero the buffer before freeing */
+    if (hash->msg != NULL) {
```
JacobBarthelmeh: Why remove the null check on hash before accessing hash->msg?

wolfcrypt/src/port/maxim/max3266x.c
night1rider: reverted

wolfcrypt/src/port/maxim/max3266x.c
night1rider: This was an oversight, fixed now.

wolfcrypt/src/port/maxim/max3266x.c
night1rider: Yes, sorry, the comment is referring to the process that happens when we use the `wc_MXC_TPU_SHA_Free` call. Inside this call it will zero out the message buffer and free the message buffer point and then set the other values as 0 for size and used.

night1rider:
Jenkins retest this please

night1rider:
Jenkins retest this please

night1rider:
rebased to pull in: https://github.com/wolfSSL/wolfssl/pull/9829

dgarske (approved):

PR #9732, opened 2026-02-02
Enable and use ML-KEM by default

Frauschi:
This PR enables the standardized PQC algorithm ML-KEM (FIPS203) by default. 

# ML-KEM enabled by default

The ML-KEM algorithm is now enabled by default when building the library. This results in the same behavior as currently passing `--enable-mlkem=yes` to configure. When a specialized configuration for ML-KEM is required, `--enable-mlkem=<desired options>` is still possible as normal.

## Algorithm changes

Prior to this PR, when ML-KEM has been enabled, all available groups that “contain” ML-KEM have been enabled to be used for the TLS key exchange. This behavior is now changed to the following:

Groups **enabled by default**: `X25519MLKEM768`, `SECP256R1MLKEM768`, `SECP384R1MLKEM1024`; these are the to-be-standardized hybrid PQ/T combinations from [draft-ietf-tls-ecdhe-mlkem](https://datatracker.ietf.org/doc/draft-ietf-tls-ecdhe-mlkem/04/). These hybrids are considered “most secure” by the community and are already widely deployed in the web, especially `X25519MLKEM768` (e.g., see [here](https://radar.cloudflare.com/adoption-and-usage)).

Currently **not** enabled by default: all **standalone** ML-KEM groups (`MLKEM512`, `MLKEM768`, `MLKEM1024` from [draft-ietf-tls-mlkem](https://datatracker.ietf.org/doc/draft-ietf-tls-mlkem/07/)), as standalone ML-KEM usage is not widely deployed presently and also considered “more dangerous”. 

Some **“extra” hybrid** PQ/T groups with different combinations of ML-KEM and ECC groups, which are [defined by the OQS project](https://github.com/open-quantum-safe/oqs-provider/blob/main/oqs-template/oqs-kem-info.md), are also **disabled** by default (`SECP256R1MLKEM512`, `SECP384R1MLKEM768`, `SECP521R1MLKEM1024`, `X25519MLKEM512`).

Next to the default enablement of these groups, the “ preference “ of these groups has also been increased. On the client side, if no specific group for the key share is set, `X25519MLKEM768` is now selected as the default one when Curve25519 is enabled. Otherwise, `SECP384R1MLKEM1024` is used (if hybrids are not disabled). On the server side, when a client presents more than one key share in its ClientHello, a hybrid one is now also preferred to a traditional-only one. 

This now also reflects the behavior of OpenSSL since version 3.5.

## Build system changes

To reflect the changes from above, changes in the build system have been performed. 

In the autoconf setup, `--enable-mlkem` is now set to `yes` by default. Furthermore, the following three new options are added:
- `--enable-pqc-hybrids` (on by default): This enables (or disables) the three to-be-standardized hybrid PQ/T combinations.
- ` --enable-tls-mlkem-standalone` (off by default): This enables the standalone usage of ML-KEM. When enabled, the three standalone groups are advertised in the SupportedGroups extension by the client. When the hybrid groups from above are disabled, then `MLKEM1024` is also selected as the default in the key share sent by the client. Otherwise, the hybrids are still ranked higher. 
- `--enable-extra-pqc-hybrids` (off by default, requires -`-enable-experimental` in addition): This enables the extra PQ/T groups from the OQS project. 

In addition to these new options, some dependencies of ML-KEM are now also enabled by default, namely SHA3 (`--enable-sha3`), SHAKE128 (`--enable-shake128`) and SHAKE256 (`--enable-shake128`). 

Furthermore, when DTLS 1.3 is enabled, then ClientHello fragmentation (`--enable-dtls-frag-ch`) is now also enabled by default, as most PQ/T and standalone PQC key shares require fragmentation.

All these options and changes have also been applied to CMake builds and to Zephyr, too.

## Runtime changes

The order of the algorithms has been changed in both the SupportedGroups extension sent by the client as well as in the `preferredGroup` array to rank the three PQ/T hybrids highest. The standalone ML-KEM groups are added based on their achieved security strength into the list, but always before their matching traditional counterparts (to prioritize the PQC groups over traditional ones in case hybrids are disabled).

For DTLS 1.3, the server now also automatically enables ClientHello fragmentation when the client sends a PQ/T or PQC key share (which requires a HelloRetryRequest and a second ClientHello).

For async crypto support, hybrid PQ/T groups are now also handled properly on both the client and server sides (only the ECC part of the hybrid combination is handled asynchronously, as ML-KEM currently lacks async support).

## Tests

In the test infrastructure, all the new options are added to be tested. Furthermore, existing tests have been adapted to incorporate the different resource usage for the ML-KEM groups (more memory usage for static memory cases, different fragmentation in case of DTLS, manually set specific groups that are required in some tests, ...).

Moreover, the general testing of PQC and the PQ/T hybrids has been extended to cover all groups and more edge-cases. 

## Other changes

- The examples have been updated to incorporate these changes.
- The two ML-KEM source files `wc_mlkem.c` and `wc_mlkem_poly.c` have been updated to fix -Wconversion warnings (now also tested in CI for ML-KEM).
- When RNG is disabled (`--disable-rng`), ML-KEM is now usable only with limited functionality. Mainly, key generation and encapsulation are only possible with a given seed. This is also handled in the unit tests.


Frauschi:
> CI is unhappy with some of the implicit casts I think:

Yeah I already talked to @SparkiDev about these and I still have to fix them (trying to fix the other failing tests first). The ML-KEM source files haven't yet been under test with these conversion checks.

dgarske:
PRB test has warning: "Your changes make libwolfssl.so size (773968) more than 40kB larger than the last release (724008)." That might be acceptable consider this change. @Frauschi will you check?

Frauschi:
> PRB test has warning: "Your changes make libwolfssl.so size (773968) more than 40kB larger than the last release (724008)." That might be acceptable consider this change. @Frauschi will you check?

I obtained some numbers on my local machine (Intel x86_64 machine running Fedora 43), which are very similar to what the Jenkins setup produces:

| Version       | Config                                                 | Size [B] | Relative | Absolute [B] |
|---------------|--------------------------------------------------------|----------|----------|--------------|
| v5.8.4-stable | `--enable-opensslextra=small`                          | 707144   | 0.9874   | -9000        |
| master        | `--enable-opensslextra=small`                          | 716144   | 1        | 0            |
| This PR       | `--enable-opensslextra=small`                          | 773464   | 1.0800   | +57320       |
|               | `--enable-opensslextra=small --enable-mlkem=yes,small` | 757032   | 1.0571   | +40888       |
|               | `--enable-opensslextra=small --disable-mlkem`          | 716152   | 1.0      | +8           |

So yeah, there are some substantial size increases when ML-KEM is enabled (even in its SMALL variant). But considering all the logic that is added, this seems acceptable to me.

wolfcrypt/src/wc_mlkem.c
```diff
@@ -582,7 +589,7 @@ int wc_MlKemKey_MakeKeyWithRandom(MlKemKey* key, const unsigned char* rand,
         /* Generate noise using PRF.
          * Alg 13: Steps 8-15: generate s and e
          */
-        ret = mlkem_get_noise(&key->prf, k, s, e, NULL, sigma);
+        ret = mlkem_get_noise(&key->prf, (int)k, s, e, NULL, sigma);
```
SparkiDev: k is int???

wolfcrypt/src/wc_mlkem.c
```diff
@@ -516,16 +522,17 @@ int wc_MlKemKey_MakeKeyWithRandom(MlKemKey* key, const unsigned char* rand,
 #ifndef WOLFSSL_MLKEM_MAKEKEY_SMALL_MEM
 #ifndef WOLFSSL_MLKEM_CACHE_A
         /* e (v) | a (m) */
-        e = (sword16*)XMALLOC((k + 1) * k * MLKEM_N * sizeof(sword16),
+        e = (sword16*)XMALLOC((size_t)(k + 1) * (size_t)k *
```
SparkiDev: Shouldn't need to cast everything.
Max k is 4. MLKEM_N is 256 (max.= 5 * 4 * 256 * 2)
Just cast the result.
Similarly below.

SparkiDev (changes requested):

wolfcrypt/src/wc_mlkem_poly.c
```diff
@@ -1805,22 +1812,23 @@ static void mlkem_ntt_add_to(sword16* r, sword16* a)
 static void mlkem_keygen_c(sword16* s, sword16* t, sword16* e, const sword16* a,
     int k)
 {
-    int i;
+    unsigned int i;
+    unsigned int uk = (unsigned int)k;
```
SparkiDev: Don't like this.
Converted then to unsigned ints unnecessarily.

wolfcrypt/src/wc_mlkem_poly.c
```diff
@@ -2209,24 +2226,25 @@ int mlkem_encapsulate_seeds(const sword16* pub, MLKEM_PRF_T* prf, sword16* u,
 static void mlkem_decapsulate_c(const sword16* s, sword16* w, sword16* u,
     const sword16* v, int k)
 {
-    int i;
+    unsigned int i;
```
SparkiDev: Same, don't like the conversion to unsigned ints.
Cast where required instead.

wolfcrypt/src/wc_mlkem_poly.c
```diff
@@ -4093,7 +4117,8 @@ static void mlkem_get_noise_x4_eta2_avx2(byte* rand, byte* seed, byte o)
     word64 state[25 * 4];
 
     for (i = 0; i < 4; i++) {
-        state[4*4 + i] = 0x1f00 + i + o;
+        state[4*4 + i] = (word64)0x1f00U + (word64)(word32)i +
```
SparkiDev: Too many casts!
Don't make it a 64-bit type until it has to be.

wolfcrypt/src/wc_mlkem_poly.c
```diff
@@ -4899,8 +4927,9 @@ static MLKEM_NOINLINE void mlkem_csubq_c(sword16* p)
  * @return  Compressed value.
  */
 #define TO_COMP_WORD_10(v, i, j, k) \
-    ((((MLKEM_V54 << 10) * (v)[(i) * MLKEM_N + (j) + (k)]) + \
-      MLKEM_V54_HALF) >> 54)
+    (word16)(((((word64)(MLKEM_V54 << 10)) * \
```
SparkiDev: You're going to screw with performance here. Minimal bit sizes - minimal casts please.

wolfcrypt/src/wc_mlkem_poly.c
```diff
@@ -4966,17 +4996,17 @@ static void mlkem_vec_compress_10_c(byte* r, sword16* v, unsigned int k)
             r[ 4] = (t[3] >> 2);
         #else
             /* Compress four polynomial values to 10 bits each. */
-            sword16 t0 = TO_COMP_WORD_10(v, i, j, 0);
-            sword16 t1 = TO_COMP_WORD_10(v, i, j, 1);
-            sword16 t2 = TO_COMP_WORD_10(v, i, j, 2);
-            sword16 t3 = TO_COMP_WORD_10(v, i, j, 3);
+            word16 t0 = TO_COMP_WORD_10(v, i, j, 0);
```
SparkiDev: Meant to be sword16.
Don't change.

wolfcrypt/src/wc_mlkem_poly.c
```diff
@@ -4986,35 +5016,35 @@ static void mlkem_vec_compress_10_c(byte* r, sword16* v, unsigned int k)
...
-            sword16 t7  = TO_COMP_WORD_10(v, i, j, 7);
-            sword16 t8  = TO_COMP_WORD_10(v, i, j, 8);
-            sword16 t9  = TO_COMP_WORD_10(v, i, j, 9);
-            sword16 t10 = TO_COMP_WORD_10(v, i, j, 10);
-            sword16 t11 = TO_COMP_WORD_10(v, i, j, 11);
-            sword16 t12 = TO_COMP_WORD_10(v, i, j, 12);
-            sword16 t13 = TO_COMP_WORD_10(v, i, j, 13);
-            sword16 t14 = TO_COMP_WORD_10(v, i, j, 14);
-            sword16 t15 = TO_COMP_WORD_10(v, i, j, 15);
+            word16 t0  = TO_COMP_WORD_10(v, i, j, 0);
```
SparkiDev: Meant to be sword16.
Don't change.

wolfcrypt/src/wc_mlkem_poly.c
```diff
@@ -5073,47 +5103,47 @@ static void mlkem_vec_compress_11_c(byte* r, sword16* v)
         /* Each 8 polynomial coefficients. */
         for (j = 0; j < MLKEM_N; j += 8) {
         #ifdef WOLFSSL_MLKEM_SMALL
-            sword16 t[8];
+            word16 t[8];
```
SparkiDev: Don't change type.

wolfcrypt/src/wc_mlkem_poly.c
```diff
@@ -5073,47 +5103,47 @@ static void mlkem_vec_compress_11_c(byte* r, sword16* v)
...
             /* Compress eight polynomial values to 11 bits each. */
-            sword16 t0 = TO_COMP_WORD_11(v, i, j, 0);
-            sword16 t1 = TO_COMP_WORD_11(v, i, j, 1);
-            sword16 t2 = TO_COMP_WORD_11(v, i, j, 2);
-            sword16 t3 = TO_COMP_WORD_11(v, i, j, 3);
-            sword16 t4 = TO_COMP_WORD_11(v, i, j, 4);
-            sword16 t5 = TO_COMP_WORD_11(v, i, j, 5);
-            sword16 t6 = TO_COMP_WORD_11(v, i, j, 6);
-            sword16 t7 = TO_COMP_WORD_11(v, i, j, 7);
+            word16 t0 = TO_COMP_WORD_11(v, i, j, 0);
```
SparkiDev: Don't change type.

wolfcrypt/src/wc_mlkem_poly.c
```diff
@@ -5425,7 +5456,8 @@ void mlkem_vec_decompress_11(sword16* v, const byte* b)
  * @return  Compressed value.
  */
 #define TO_COMP_WORD_5(p, i, j) \
-    ((((MLKEM_V27 << 5) * (p)[(i) + (j)]) + MLKEM_V27_HALF) >> 27)
+    (byte)(((((MLKEM_V27 << 5) * (word32)(word16)((p)[(i) + (j)])) + \
```
SparkiDev: Pointless casting to word32 directly after casting to word16.
Same above.

wolfcrypt/src/wc_mlkem_poly.c
```diff
@@ -5853,8 +5885,11 @@ void mlkem_from_msg(sword16* p, const byte* msg)
  * @param  [in]   j   Index of bit in byte.
  */
 #define TO_MSG_BIT(m, p, i, j) \
-    (m)[i] |= ((word32)((MLKEM_V31_2 * (p)[8 * (i) + (j)]) + \
-                        MLKEM_V31_HALF) >> 31) << (j)
+    do { \
+        word32 _t = ((((word32)(word16)((p)[8 * (i) + (j)])) * MLKEM_V31_2) + \
```
SparkiDev: Make this one statement.

Way too many casts!

wolfcrypt/src/wc_mlkem_poly.c
Frauschi: I reduced the number of casts. But I still need a cast to `word64`. But as `MLKEM_V54` is a 64 bit integer, that was a 64 bit multiplication previously anyway, wasn't it? So it shouldn't result in performance differences now.

Frauschi:
I added a commit to incorporate the feedback of @SparkiDev as well as one to add test coverage for #9777.

Frauschi:
Jenkins retest this please - logs gone

wolfcrypt/src/wc_mlkem_poly.c
```diff
@@ -1812,28 +1805,28 @@ static void mlkem_ntt_add_to(sword16* r, sword16* a)
...
          * Step 18: ... A_hat o s_hat ... */
-        mlkem_pointwise_acc_mont(t + i * MLKEM_N, a + i * uk * MLKEM_N, s, uk);
+        mlkem_pointwise_acc_mont(t + i * MLKEM_N, a + i * k * MLKEM_N, s,
+            (unsigned int)k);
         /* Convert public polynomial to Montgomery form.
          * Step 18: ... MontRed(A_hat o s_hat) ... */
         for (j = 0; j < MLKEM_N; ++j) {
-            sword32 n = t[i * MLKEM_N + j] * (sword32)MLKEM_F;
-            t[i * MLKEM_N + j] = MLKEM_MONT_RED(n);
+            sword32 n = t[(unsigned int)i * MLKEM_N + j] * (sword32)MLKEM_F;
```
SparkiDev: Why does this need to be unsigned?

Frauschi: Fixed

dgarske:
Jenkins retest this please

SparkiDev (dismissed):

Frauschi:
I squashed the last three commits regarding the ML-KEM `Wconversion` fixes after the review of those. Thanks @SparkiDev for all the feedback!

Frauschi:
@julek-wolfssl @anhu @dgarske can you please review?

wolfssl/wolfcrypt/settings.h
```diff
@@ -4512,15 +4512,20 @@ extern void uITRON4_free(void *p) ;
     #define WOLFSSL_DILITHIUM_VERIFY_NO_MALLOC
 #endif
 
-#if defined(HAVE_PQC) && defined(WOLFSSL_DTLS13) && \
-    !defined(WOLFSSL_DTLS_CH_FRAG)
-#warning "Using DTLS 1.3 + pqc without WOLFSSL_DTLS_CH_FRAG will probably" \
-         "fail.Use --enable-dtls-frag-ch to enable it."
+#if defined(HAVE_PQC) && defined(WOLFSSL_HAVE_MLKEM) && \
+    defined(WOLFSSL_DTLS13) && !defined(WOLFSSL_DTLS_CH_FRAG)
+#define WOLFSSL_DTLS_CH_FRAG
```
julek-wolfssl: `WOLFSSL_DTLS_CH_FRAG` is our hack to get PQC working with DTLS. I'm not sure that quietly enabling it is a good idea. It may clash with implementations that solve large key shares differently (eg. jumbo packets, IP fragmentation). @rizlik thoughts?

tests/api.c
```diff
@@ -29101,6 +29102,9 @@ static int test_dtls13_bad_epoch_ch(void)
     /* resend the CH */
     ExpectIntEQ(wolfSSL_dtls_got_timeout(ssl_c), WOLFSSL_SUCCESS);
 
+    /* Re-enable HRR cookie to account for potential CH fragmentation */
+    ExpectIntEQ(wolfSSL_send_hrr_cookie(ssl_s, NULL, 0), WOLFSSL_SUCCESS);
+
```
julek-wolfssl: We shouldn't promote changing settings in the middle of a connection. Maybe limiting the ciphersuites to classical only is a better approach.

src/tls.c
```diff
@@ -10198,7 +10272,7 @@ int TLSX_KeyShare_Parse_ClientHello(const WOLFSSL* ssl,
         offset += ret;
     }
 
-    if (ssl->hrr_keyshare_group != 0) {
+    if (ssl->hrr_keyshare_group != 0 && seenGroupsCnt > 0) {
```
julek-wolfssl: Is an empty key share not an error? Why success on `seenGroupsCnt == 0`?

src/dtls.c
```diff
@@ -822,8 +822,36 @@ static int SendStatelessReplyDtls13(const WOLFSSL* ssl, WolfSSL_CH* ch)
...
+            TLSX* ksExt = TLSX_Find(parsedExts, TLSX_KEY_SHARE);
+            KeyShareEntry* kse = (KeyShareEntry*)ksExt->data;
+            if (WOLFSSL_NAMED_GROUP_IS_PQC(kse->group) ||
+                WOLFSSL_NAMED_GROUP_IS_PQC_HYBRID(kse->group)) {
+                /* Allow fragmentation of the second ClientHello due to the
+                * large PQC key share. */
+                wolfSSL_dtls13_allow_ch_frag((WOLFSSL*)ssl, 1);
+            }
+        }
+#endif
```
julek-wolfssl: Again, I don't know if enabling this quietly for the user is safe. We should respect the API contract we have with users where this needs to be explicitly enabled. There are DoS considerations. @rizlik

src/dtls.c
```diff
@@ -822,8 +822,36 @@ static int SendStatelessReplyDtls13(const WOLFSSL* ssl, WolfSSL_CH* ch)
...
             goto dtls13_cleanup;
+
+#ifdef WOLFSSL_DTLS_CH_FRAG
+        {
+            /* Get the chosen group. If ret == 0 here, we are sure that the
+            * extension is present. */
+            TLSX* ksExt = TLSX_Find(parsedExts, TLSX_KEY_SHARE);
+            KeyShareEntry* kse = (KeyShareEntry*)ksExt->data;
+            if (WOLFSSL_NAMED_GROUP_IS_PQC(kse->group) ||
+                WOLFSSL_NAMED_GROUP_IS_PQC_HYBRID(kse->group)) {
```
julek-wolfssl:
```suggestion
            if (kse != NULL && (WOLFSSL_NAMED_GROUP_IS_PQC(kse->group) ||
                WOLFSSL_NAMED_GROUP_IS_PQC_HYBRID(kse->group))) {
```

wolfssl/wolfcrypt/settings.h
rizlik: I think it's OK: on the client-side we fully respect the protocol and on the server-side we are slightly less pickier compared to not having that enabled, but still on the safe side.

Even with option enabled, the server will still refuse messages that are valid per RFC, but impossible to handle statelessly: CH0 fragmented or CH1 fragmented but with cookie extension on the second fragment.

In the future we might want to allow users to drop statelessness for performance (removing both the needs of an extra round with HRR and/or interop (CH fragmented without having the cookie extension in the first fragment from other implementations). But for now, I don't see problems enabling it. We would fail interop  with library that do not respect our model regardless.

src/dtls.c
rizlik: replied above

wolfssl/wolfcrypt/settings.h
Frauschi: Ok, so we can leave the "automatic" and implicit enablement as is?

tests/api.c
Frauschi: fixed

src/dtls.c
Frauschi: fixed

src/tls.c
Frauschi: You are right that this change is indeed an error in case of TLS and stateful DTLS. 

However, in case of a stateless DTLSv1.3 server, the `ssl->hrr_keyshare_group` variable may already be set from a previous ClientHello message that triggered a HRR but never proceeded with the handshake. In this case, when a new valid ClientHello is processed with an empty key share extension, this fails the handshake although another HRR would be the correct behavior. 

This is, for example, triggered in the `test_dtls_memio_wolfio_stateless` test. 

I modified the code to only allow the case that `ssl->hrr_keyshare_group != 0` and `seenGroupsCnt == 0` when we have a stateless DTLSv1.3 server.

Frauschi:
@julek-wolfssl I addressed your feedback in a new commit. Once all is fine, I will squash them again for a clear history.

dgarske:
There is a PRB error for code size increase... "Your changes make libwolfssl.so size (778280) more than 40kB larger than the last release (724008).". I think we can assume that error is okay.

dgarske (dismissed):

wolfssl/wolfcrypt/settings.h
rizlik: yes

src/tls.c
rizlik: This is not clear to me. The point of having stateless CH parsing is that the first CH hello should have no effect on the state. If it has, handling that is the wrong fix and we should avoid that the first CH change the state altogether. 
Just one HRR is allowed in (D)TLSv1.3.

Frauschi: `ssl->hrr_keyshare_group` is a stateful variable for a specific session. In case of a stateless DTLS session, this variable may have a value != 0 from a previous incomplete handshake (e.g. one which caused a HRR, but never continued afterward), causing an error once a new valid CH is received that now also causes a HRR (which should be allowed with the current logic). Hence, the overall handling of the `ssl->hrr_keyshare_group` variable in the DTLS stateless case is invalid. I updated the PR to reflect that, exiting the if clause successfully in case of stateless DTLS.

julek-wolfssl: This should be something that was caught by `test_wolfSSL_dtls_stateless`. Please add a test case that will error out without your fix so that we don't lose statelessness in the future.

Frauschi:
@julek-wolfssl @rizlik As discussed on Slack, I created #9848 to fix the DTLS HRR group issue discussed above. Once this is merged, I will rebase this PR on top.

anhu:
@aisle-analyzer

Frauschi:
Jenkins retest this please

src/tls.c
Frauschi: Fixed via #9848

Frauschi:
I added the commit from #9848 here, too, to ensure it works with the changes here in CI. Once #9848 is merged, a simple rebase will remove that. 

With that, all the open DTLS issues should be resolved @julek-wolfssl @rizlik?

The only failure in the PRB-master-job is the already mentioned code-size increase.

src/dtls.c
```diff
@@ -823,8 +823,36 @@ static int SendStatelessReplyDtls13(const WOLFSSL* ssl, WolfSSL_CH* ch)
...
+        if (ksExt != NULL) {
+            KeyShareEntry* kse = (KeyShareEntry*)ksExt->data;
+            if (kse != NULL && (WOLFSSL_NAMED_GROUP_IS_PQC(kse->group) ||
+                                WOLFSSL_NAMED_GROUP_IS_PQC_HYBRID(kse->group))){
+                /* Allow fragmentation of the second ClientHello due to the
+                 * large PQC key share. */
+                wolfSSL_dtls13_allow_ch_frag((WOLFSSL*)ssl, 1);
+            }
+        }
+#endif
```
julek-wolfssl: Don't change settings during a connection. Call `wolfSSL_dtls13_allow_ch_frag` in `InitSSL` when PQC is enabled.

julek-wolfssl (changes requested):

src/dtls.c
Frauschi: Fixed. It is now enabled in `InitSSL()` and `InitSSL_Side()`.

julek-wolfssl (approved):

Frauschi:
Fixed conflicts to master and squashed final DTLS review commit. no other changes compared to the state that has been approved earlier.

Frauschi:
> ```
> failed: check-source-text all-check-certs-one-month quantum-safe-wolfssl-all-g++-latest-debug quantum-safe-wolfssl-all-clang++-latest all-crypto-openssl-extra-coexist-smallstack
> ```

Should all be fixed now

Frauschi:
Fixed the merge conflicts.

Frauschi:
Some of the changes of this PR have been extracted into separate PRs (#9967, #9968, and #9969) to make it into the release. This PR will then be rebased once they are merged and the release is over.

Frauschi:
Jenkins retest this please

dgarske:
Jenkins retest this please. History lost.

dgarske:
Only error is : PRB-libwolfssl-size-change #10574 failed with:
ERROR: Your changes make libwolfssl.so size (778504) more than 10kB larger than master (741592).
If we merge this does it become the new baseline? @douzzer or @night1rider

PR #9736, opened 2026-02-03
Exclude Config.cmake.in in .gitignore

### LINKED ISSUE #9731: [Bug]: .gitignore ignores cmake/Config.cmake.in on case-insensitive filesystems (wolfSSL 5.8.4)
### Version

5.8.4

### Description

In wolfSSL 5.8.4, the top-level .gitignore rule:
```
config.*
```

causes cmake/Config.cmake.in to be ignored on case-insensitive filesystems
(Windows, default macOS).

Config.cmake.in is a source-controlled CMake input file, so ignoring it can
break CMake-based builds and downstream packaging.

This issue does not reproduce on typical Linux systems with case-sensitive
filesystems, making it easy to miss.

_I may be pushing it a bit labelling this as a bug._

### Steps to Reproduce

Clone wolfSSL 5.8.4 on Windows (I used Windows) or macOS:
```
git clone --branch v5.8.4 https://github.com/wolfSSL/wolfssl.git
cd wolfssl
```

Check ignore status:
```
git check-ignore -v cmake/Config.cmake.in
```

Output shows the file is ignored due to:
```
.gitignore:config.*
```

### Proposed Fix

Add an explicit exception to .gitignore:
```
!cmake/Config.cmake.in
```

### Relevant log output

```shell

```

embhorn:
# Description

Add an explicit exception to .gitignore to exclude `Config.cmake.in`

Fixes #9731

# Testing

Confirmed with .gitignore:config.*

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -386,6 +386,7 @@ IDE/**/DerivedData
 CMakeFiles/
 CMakeCache.txt
 cmake_install.cmake
+!cmake/Config.cmake.in
 
 # GDB Settings
 \.gdbinit


embhorn:
Jenkins retest this

douzzer (approved):

PR #9743, opened 2026-02-05
20260205-fixes

douzzer:
`src/x509.c`: in `x509_aia_append_string()`, use heap pointer from supplied `*head`, and initialize `WOLFSSL_STACK`.`type` (fixes "Conditional jump or move depends on uninitialised value(s)" from 08c1397cc1);

`wolfssl/wolfcrypt/asn.h`: use `WC_BITFIELD` for `extAuthInfoListSz` and `extAuthInfoListOverflow` bitfields, for C89 compat (fixes `-Wpedantic` from 08c1397cc1).

tested with
```
wolfssl-multi-test.sh ...
check-source-text
allcryptonly-gcc-c89
quantum-safe-wolfssl-all-intelasm-sp-asm-valgrind
```

fixes #9728

### DIFF
--- a/src/x509.c
+++ b/src/x509.c
@@ -14986,53 +14986,38 @@ int wolfSSL_X509_NAME_digest(const WOLFSSL_X509_NAME *name,
 
 void wolfSSL_X509_email_free(WOLF_STACK_OF(WOLFSSL_STRING) *sk)
 {
-    WOLFSSL_STACK *curr;
-
-    while (sk != NULL) {
-        curr = sk;
-        sk = sk->next;
-
-        XFREE(curr, NULL, DYNAMIC_TYPE_OPENSSL);
-    }
+    wolfSSL_sk_pop_free(sk, NULL);
 }
 
-static WOLFSSL_STACK* x509_aia_append_string(WOLFSSL_STACK** head,
+static int x509_aia_append_string(WOLFSSL_STACK** head,
     const byte* uri, word32 uriSz)
 {
     WOLFSSL_STACK* node;
     char*          url;
 
-    node = (WOLFSSL_STACK*)XMALLOC(sizeof(WOLFSSL_STACK) + uriSz + 1,
-                                   *head ? (*head)->heap : NULL,
-                                   DYNAMIC_TYPE_OPENSSL);
-    if (node == NULL)
-        return NULL;
+    url = (char*)XMALLOC(uriSz + 1, NULL, DYNAMIC_TYPE_OPENSSL);
+    if (url == NULL)
+        return WOLFSSL_FAILURE;
 
-    url = (char*)node;
-    url += sizeof(WOLFSSL_STACK);
     XMEMCPY(url, uri, uriSz);
     url[uriSz] = '\0';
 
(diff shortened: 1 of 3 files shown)


douzzer:
retest this please
```
```

src/x509.c
```diff
@@ -15012,9 +15013,11 @@ static WOLFSSL_STACK* x509_aia_append_string(WOLFSSL_STACK** head,
     XMEMCPY(url, uri, uriSz);
     url[uriSz] = '\0';
 
+    node->heap = *head ? (*head)->heap : NULL;
     node->data.string = url;
     node->next = NULL;
     node->num = 1;
```
padelsbach: If this is not the first node, should `num` still be 1?

src/x509.c
```diff
@@ -15002,8 +15002,9 @@ static WOLFSSL_STACK* x509_aia_append_string(WOLFSSL_STACK** head,
     WOLFSSL_STACK* node;
     char*          url;
 
-    node = (WOLFSSL_STACK*)XMALLOC(sizeof(WOLFSSL_STACK) + uriSz + 1, NULL,
-            DYNAMIC_TYPE_OPENSSL);
+    node = (WOLFSSL_STACK*)XMALLOC(sizeof(WOLFSSL_STACK) + uriSz + 1,
```
padelsbach: Can we use `wolfSSL_sk_new_node` here?

src/x509.c
```diff
@@ -15012,9 +15013,11 @@ static WOLFSSL_STACK* x509_aia_append_string(WOLFSSL_STACK** head,
     XMEMCPY(url, uri, uriSz);
     url[uriSz] = '\0';
 
+    node->heap = *head ? (*head)->heap : NULL;
```
padelsbach: Maybe we should use `wolfSSL_sk_push_node` here?

wolfssl/wolfcrypt/asn.h
```diff
@@ -2075,8 +2075,8 @@ struct DecodedCert {
 #endif /* WOLFSSL_DUAL_ALG_CERTS */
 
     WOLFSSL_AIA_ENTRY extAuthInfoList[WOLFSSL_MAX_AIA_ENTRIES];
-    byte    extAuthInfoListSz:7;
-    byte    extAuthInfoListOverflow:1;
+    WC_BITFIELD extAuthInfoListSz:7;
```
padelsbach, line 2078: Thanks for fixing this

src/x509.c
douzzer: That would work, though it would frivolously zero the structure.

src/x509.c
douzzer: `wolfSSL_sk_push_node()` puts the new node at the start, not the end, so it can't be used.

but `->num` does need to be fixed up.  apparently they will _all_ need to be renumbered, to append a node, because the tail of the list needs to be num `1`.  yuck.

padelsbach: Added `wolfSSL_sk_push_back_node` to push to the back of the stack. Maybe a bit un-stack-ish, but might as well roll it into the `sk` interface.

dgarske (approved):

PR #9761, opened 2026-02-11
Implement OCSP responder

julek-wolfssl:
OCSP Responder Core API:

- Add new public API for creating and managing an OCSP responder
- Add public wrappers for internal OCSP request/response functions
- OcspRespCheck: fix check when authorized responder is loaded into CM

Header Cleanup:

- Remove circular dependency when including `#include <wolfssl/wolfcrypt/asn.h>` from wolfssl/wolfcrypt/ecc.h and wolfssl/wolfcrypt/rsa.h

OCSP Responder Example (examples/ocsp_responder/):

- Add a command-line OCSP responder for interoperability testing with OpenSSL's `openssl ocsp` client

Test Scripts (scripts/):

- ocsp-responder-openssl-interop.test: Tests wolfSSL OCSP responder with `openssl ocsp` client
- ocsp-stapling-with-wolfssl-responder.test: Tests wolfSSL OCSP responder when doing OCSP stapling

Certificate Infrastructure (certs/ocsp/):

- Add DER-format certificates and keys for OCSP testing
- Update renewcerts.sh to generate DER versions

Known Limitations (documented in src/ocsp.c header comment):

  - Single request/response per OCSP exchange only
  - Key-hash responder ID only (no name-based responder ID)
  - No singleExtensions support


julek-wolfssl:
Retest this please

julek-wolfssl:
Added windows testing coverage

examples/ocsp_responder/ocsp_responder.c
```diff
@@ -0,0 +1,1140 @@
...
+                        /* Format: YYMMDDHHMMSSZ */
+                        if (XSTRLEN(field) >= 12) {
+                            int year = (field[0] - '0') * 10 + (field[1] - '0');
+                            tm.tm_year = (year < 50) ? (100 + year) : year;
+                            tm.tm_mon = (field[2] - '0') * 10 + (field[3] - '0') - 1;
+                            tm.tm_mday = (field[4] - '0') * 10 + (field[5] - '0');
+                            tm.tm_hour = (field[6] - '0') * 10 + (field[7] - '0');
+                            tm.tm_min = (field[8] - '0') * 10 + (field[9] - '0');
+                            tm.tm_sec = (field[10] - '0') * 10 + (field[11] - '0');
+                            entry->revocationTime = XMKTIME(&tm);
```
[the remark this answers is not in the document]
julek-wolfssl: Time zone won't have an affect on the revocation dates in the tests.

src/ocsp.c
```diff
@@ -631,7 +637,8 @@ static int CheckOcspResponderChain(OcspEntry* single, DecodedCert *cert,
  * @param cert The decoded bs->cert
  * @return
  */
-int CheckOcspResponder(OcspResponse *bs, DecodedCert *cert, void* vp)
+int CheckOcspResponder(OcspResponse *bs, byte* subjectHash,
+        byte extExtKeyUsage, byte* issuerHash, void* vp)
```
rizlik, line 644: update the doc comment above the function

src/ocsp.c
```diff
@@ -2120,6 +2170,686 @@ int wolfSSL_OCSP_check_nonce(OcspRequest* req, WOLFSSL_OCSP_BASICRESP* bs)
...
+#endif
+#ifdef HAVE_ECC
+    if (ca->keyType == ECDSAk) {
+        wc_ecc_free(&ca->key.ecc);
+    }
+#endif
+
+    /* Free certificate DER if allocated */
+    if (ca->certDer != NULL) {
+        XFREE(ca->certDer, heap, DYNAMIC_TYPE_OCSP);
```
rizlik, line 2201: NIT: set ca->certDer to NULL? same for others

src/ocsp.c
```diff
@@ -2120,6 +2170,686 @@ int wolfSSL_OCSP_check_nonce(OcspRequest* req, WOLFSSL_OCSP_BASICRESP* bs)
...
+{
+    int ret = -1;
+    OcspResponderCa* ca = NULL;
+    DecodedCert* decoded = NULL;
+    word32 keyOID = 0;
+
+    WOLFSSL_ENTER("wc_OcspResponder_AddSigner");
+
+    if (responder == NULL || signerDer == NULL || signerDerSz == 0 ||
+        keyDer == NULL || keyDerSz == 0)
```
rizlik: add check that `issuerCertDer` can't be null if `issuerCertDerSz` is not 0

src/ocsp.c
```diff
@@ -2120,6 +2170,686 @@ int wolfSSL_OCSP_check_nonce(OcspRequest* req, WOLFSSL_OCSP_BASICRESP* bs)
...
+    if (decoded != NULL) {
+        wc_FreeDecodedCert(decoded);
+        XFREE(decoded, responder->heap, DYNAMIC_TYPE_OCSP);
+    }
+    if (ca != NULL)
+        FreeOcspResponderCa(ca, responder->heap);
+    return ret;
+}
+
+/* Find Auth CA by comparing cert DER */
```
rizlik: Is this stale?

src/ocsp.c
```diff
@@ -2120,6 +2170,686 @@ int wolfSSL_OCSP_check_nonce(OcspRequest* req, WOLFSSL_OCSP_BASICRESP* bs)
...
+    if (isNew) {
+        /* Add to CA's status list */
+        certStatus->next = ca->statuses;
+        ca->statuses = certStatus;
+        certStatus = NULL; /* Ownership transferred */
+    }
+    ret = 0;
+
+out:
+    if (isNew && certStatus != NULL)
```
rizlik, line 2584: NIT: you can probably drop isNew here.

wolfssl/wolfcrypt/asn.h
```diff
@@ -2642,7 +2655,31 @@ enum cert_enums {
...
+    #ifdef WOLFSSL_SHA384
+        byte sha384[WC_SHA384_DIGEST_SIZE];
+    #endif
+    #ifdef WOLFSSL_SHA512
+        byte sha512[WC_SHA512_DIGEST_SIZE];
+    #endif
+    #ifdef WOLFSSL_SM3
+        byte sm3[WC_SM3_DIGEST_SIZE];
+    #endif
+} AsnHashes;
```
rizlik, line 2678: don't we have something very similar to this for TLS?

wolfcrypt/src/asn.c
```diff
@@ -40781,6 +41475,181 @@ int EncodeOcspRequest(OcspRequest* req, byte* output, word32 size)
...
+    }
+    if (ret == 0) {
+        GetASN_GetRef(&dataASN[OCSPREQUESTASN_IDX_TBS_REQ_SERIAL],
+                &serial, &serialSz);
+        if (serialSz == 0 || serial == NULL)
+            ret = ASN_PARSE_E;
+    }
+    if (ret == 0) {
+        /* Copy serial number */
+        req->serial = (byte*)XMALLOC((size_t)serialSz, req->heap,
```
rizlik, line 41681: why don't cap `serialSz` to EXTERNAL_SERIAL_SIZE and use a static array in requests too? 
( I suppose ASN parsing will not allow arbitrary MALLOC, but probably still a good idea to cap)

rizlik (changes requested):
partial review

wolfcrypt/src/asn.c
```diff
@@ -13330,6 +13350,11 @@ static int StoreKey(DecodedCert* cert, const byte* source, word32* srcIdx,
             cert->pubKeyStored = 1;
             cert->pubKeySize   = (word32)length;
 
+#ifdef HAVE_OCSP_RESPONDER
+            cert->publicKeyForHash = cert->publicKey;
+            cert->pubKeyForHashSize = cert->pubKeySize;
+#endif
```
rizlik, line 13356: Do you think it might be better to re-compute the bitstring when needed instead of adding new fields to cert?

wolfcrypt/src/asn.c
```diff
@@ -24942,6 +25002,33 @@ int wc_ParseCert(DecodedCert* cert, int type, int verify, void* cm)
...
+
+const byte* wc_GetDecodedCertSerial(struct DecodedCert* cert, word32* serialSz)
+{
+    if (cert == NULL || serialSz == NULL) {
+        return NULL;
+    }
+    *serialSz = (word32)cert->serialSz;
+    return cert->serial;
+}
+
```
rizlik: those APIs look dangerous, are they needed? If only for test can we change the visibility? If they are needed can we use a copy instead of a reference?

rizlik (changes requested):

wolfssl/wolfcrypt/asn.h
rizlik: do these probably belongs more to ocsp.c?

examples/ocsp_responder/ocsp_responder.c
```diff
@@ -0,0 +1,1030 @@
...
+    #include <config.h>
+#endif
+
+#ifndef WOLFSSL_USER_SETTINGS
+    #include <wolfssl/options.h>
+#endif
+#include <wolfssl/wolfcrypt/settings.h>
+
+#undef TEST_OPENSSL_COEXIST /* can't use this option with this example */
+#undef OPENSSL_COEXIST /* can't use this option with this example */
```
rizlik, line 38: is this ok to have them undefined here but defined in the general config?

examples/ocsp_responder/ocsp_responder.c
```diff
@@ -0,0 +1,1134 @@
...
+#endif
+
+/* Index file entry structure */
+typedef struct IndexEntry {
+    char status;                /* V=valid, R=revoked, E=expired */
+    time_t expirationTime;
+    time_t revocationTime;
+    char serial[64];
+    char filename[256];
+    char subject[512];
```
rizlik, line 129: can we use define instead of hardcoded magic numbers?

examples/ocsp_responder/ocsp_responder.c
```diff
@@ -0,0 +1,1134 @@
...
+
+    if (isPem) {
+        ret = ConvertPemToDer(buf, bufSz, der, derSz, PRIVATEKEY_TYPE);
+        XFREE(buf, NULL, DYNAMIC_TYPE_TMP_BUFFER);
+        return ret;
+    }
+    else {
+        *der = buf;
+        *derSz = bufSz;
+        return 0;
```
rizlik, line 257: is this getting the key here? `ConvertPemToDer` extract the key but I don't see that for der certificate.

examples/ocsp_responder/ocsp_responder.c
```diff
@@ -0,0 +1,1134 @@
...
+        XMEMSET(entry, 0, sizeof(IndexEntry));
+
+        /* Parse tab-separated fields */
+        while ((field = XSTRSEP(&p, "\t")) != NULL && fieldNum < 6) {
+            switch (fieldNum) {
+                case 0: /* Status */
+                    entry->status = field[0];
+                    break;
+                case 1: /* Expiration time (YYMMDDHHMMSSZ format) */
+                    /* Parse if needed */
```
rizlik, line 318: is it needed?

examples/ocsp_responder/ocsp_responder.c
```diff
@@ -0,0 +1,1134 @@
...
+static int RecvHttpRequest(SOCKET_T fd, byte* buf, int bufSz)
+{
+    int totalLen = 0;
+    int contentLen = 0;
+    int headerSz = 0;
+
+    while (totalLen < bufSz - 1) {
+        int n = (int)recv(fd, (char*)buf + totalLen,
+                          (size_t)(bufSz - 1 - totalLen), 0);
+        if (n <= 0)
```
rizlik, line 493: we should probably manage interruptions here, probably fine for an example

examples/ocsp_responder/ocsp_responder.c
```diff
@@ -0,0 +1,1134 @@
...
+        totalLen += n;
+        buf[totalLen] = '\0';
+
+        /* Once we find end-of-headers, parse Content-Length */
+        if (headerSz == 0) {
+            const char* hdrEnd = XSTRSTR((char*)buf, "\r\n\r\n");
+            if (hdrEnd != NULL) {
+                const char* cl;
+                headerSz = (int)(hdrEnd + 4 - (char*)buf);
+                cl = XSTRSTR((char*)buf, "Content-Length:");
```
rizlik, line 504: Not sure how far we want to go into http parsing but this will match "Content-length" in any part of the buffer, in the message body as well.
Moreover it doesn't account for start of the line neither for different casing other than the two ones here.

examples/ocsp_responder/ocsp_responder.c
```diff
@@ -0,0 +1,1134 @@
...
+    int contentLen = 0;
+    int headerSz = 0;
+
+    while (totalLen < bufSz - 1) {
+        int n = (int)recv(fd, (char*)buf + totalLen,
+                          (size_t)(bufSz - 1 - totalLen), 0);
+        if (n <= 0)
+            break;
+        totalLen += n;
+        buf[totalLen] = '\0';
```
rizlik, line 496: '\0' bytes might be present in the receiving data, not sure if this can confuse other code later.

examples/ocsp_responder/ocsp_responder.c
```diff
@@ -0,0 +1,1134 @@
...
+                return -1;
+            }
+
+            /* Use Content-Length if available, otherwise use remaining data */
+            if (*bodySz == 0) {
+                *bodySz = httpReqSz - offset;
+            }
+
+            /* Ensure that the claimed body length fits in the received data */
+            if (offset + *bodySz > httpReqSz) {
```
rizlik, line 582: this can probably overflow easily

examples/ocsp_responder/ocsp_responder.c
```diff
@@ -0,0 +1,1134 @@
...
+            if (offset < 0 || offset > httpReqSz) {
+                LOG_ERROR("Invalid HTTP body offset\n");
+                *body = NULL;
+                *bodySz = 0;
+                return -1;
+            }
+
+            /* Use Content-Length if available, otherwise use remaining data */
+            if (*bodySz == 0) {
+                *bodySz = httpReqSz - offset;
```
rizlik: this is likely wrong for connection that have not "connection-close" header

examples/ocsp_responder/ocsp_responder.c
```diff
@@ -0,0 +1,1134 @@
...
+            return OCSP_INTERNAL_ERROR;
+        case ASN_NO_SIGNER_E:
+            return OCSP_UNAUTHORIZED;
+        default:
+            return OCSP_INTERNAL_ERROR;
+    }
+}
+
+/* Main OCSP responder function */
+THREAD_RETURN WOLFSSL_THREAD ocsp_responder_test(void* args)
```
rizlik, line 695: why _test?

rizlik (changes requested):

src/ocsp.c
julek-wolfssl: The entire struct is `free`d at the end so this is not necessary.

src/ocsp.c
julek-wolfssl: Removing

src/ocsp.c
julek-wolfssl: No because we don't want to free a `certStatus` that is in the list already.

wolfssl/wolfcrypt/asn.h
julek-wolfssl: No because `struct Hashes` gates some algos based on TLS versions enabled. I put it in `asn.h` because this could be useful in other areas like in cert verification.

wolfcrypt/src/asn.c
julek-wolfssl: This is how the struct is before the PR. I don't think it makes sense to make this change in this PR.

wolfcrypt/src/asn.c
julek-wolfssl: I think adding computation edge cases is not the right way to go. Storing the location of the data to be hashed keeps the parsing logic in one place. For now ECC is the only problem but maybe PQC will also introduce some custom handling in the future. This simplifies the OCSP code.

examples/ocsp_responder/ocsp_responder.c
julek-wolfssl: This is the pattern used in all the other apps in `examples/`.

examples/ocsp_responder/ocsp_responder.c
julek-wolfssl: Hard coded values are acceptable in the testing code for simplicity.

examples/ocsp_responder/ocsp_responder.c
julek-wolfssl: `ConvertPemToDer` calls `wc_PemToDer` and then sets `der` to the beginning of the DER. Neither path extracts the key.

examples/ocsp_responder/ocsp_responder.c
julek-wolfssl: For now not used in testing.

examples/ocsp_responder/ocsp_responder.c
julek-wolfssl: I'll leave it as is for simplicity for now.

examples/ocsp_responder/ocsp_responder.c
julek-wolfssl: This is enough for testing purposes.

examples/ocsp_responder/ocsp_responder.c
julek-wolfssl: This is just a safety catch. Calling code relies on the returned length.

examples/ocsp_responder/ocsp_responder.c
julek-wolfssl: Not a concern in the example.

examples/ocsp_responder/ocsp_responder.c
julek-wolfssl: Leaving as is for now.

examples/ocsp_responder/ocsp_responder.c
julek-wolfssl: Using nomenclature from other apps in `examples/`.

src/ocsp.c
julek-wolfssl: Fixing

src/ocsp.c
julek-wolfssl: Fixing

wolfcrypt/src/asn.c
julek-wolfssl: Changing

wolfcrypt/src/asn.c
rizlik: it's ok to keep the pointers but myabe it's still a good idea to use some capping. But I'm ok on relying on ASN parsing check as well

examples/ocsp_responder/ocsp_responder.c
rizlik: This is not testing though, as this is inside example folder. If you want to keep this here, can you *print* the disclaimer to the output of the program?

examples/ocsp_responder/ocsp_responder.c
rizlik: add a comment at least

rizlik (changes requested):

examples/ocsp_responder/ocsp_responder.c
julek-wolfssl: Adding a disclaimer to the top of the output.

examples/ocsp_responder/ocsp_responder.c
julek-wolfssl: Erroring out here.

wolfcrypt/src/asn.c
julek-wolfssl: Adding cap to serialSz

examples/ocsp_responder/ocsp_responder.c
```diff
@@ -0,0 +1,1138 @@
...
+
+            /* Validate that the body pointer is within the received buffer */
+            if (offset < 0 || offset > httpReqSz) {
+                LOG_ERROR("Invalid HTTP body offset\n");
+                *body = NULL;
+                *bodySz = 0;
+                return -1;
+            }
+
+            if (*bodySz == 0) {
```
rizlik, line 575: This is not right, if the "Connection-close" header is present not having the content length is ok

rizlik (changes requested):

examples/ocsp_responder/ocsp_responder.c
julek-wolfssl: Adding comment

examples/ocsp_responder/ocsp_responder.c
julek-wolfssl: Reverting. Adding comment

rizlik (approved):

examples/ocsp_responder/ocsp_responder.c
```diff
@@ -0,0 +1,1141 @@
...
+    const char* indexFile;
+    const char* readyFile;
+    int nrequests;
+    int verbose;
+    int sendCerts;
+} OcspResponderOptions;
+
+/* Usage help */
+
+/* Usage help */
```
JacobBarthelmeh, line 148: Duplicate comment here.

examples/ocsp_responder/ocsp_responder.c
```diff
@@ -0,0 +1,1141 @@
...
+#include <wolfssl/ssl.h>
+#include <wolfssl/ocsp.h>
+#include <wolfssl/test.h>
+#include <wolfssl/error-ssl.h>
+#include <wolfssl/wolfcrypt/asn.h>
+
+#include <examples/ocsp_responder/ocsp_responder.h>
+
+/* Check if we have the required features */
+#if defined(HAVE_OCSP) && defined(HAVE_OCSP_RESPONDER) && !defined(NO_FILESYSTEM)
```
JacobBarthelmeh, line 49: There are many lines in this example that are over 80 characters long.

wolfcrypt/src/asn.c
```diff
@@ -40914,12 +41835,8 @@ int CompareOcspReqResp(OcspRequest* req, OcspResponse* resp)
 
     /* match based on found status and return */
     for (single = resp->single; single; single = next) {
-    #if defined(WOLFSSL_SM2) && defined(WOLFSSL_SM3)
         ocspDigestSize = wc_HashGetDigestSize(
```
JacobBarthelmeh, line 41838: There is not a return value check on ocspDigestSize here before use. I see the hash algo ID would be checked though currently when set, before it could get to this point in the code. Would like to see a check added.

src/tls13.c
```diff
@@ -8732,7 +8732,9 @@ static int WriteCSRToBuffer(WOLFSSL* ssl, DerBuffer** certExts,
             if (tmpSz > (OPAQUE8_LEN + OPAQUE24_LEN) &&
                 certExts[extIdx] == NULL) {
                 /* csr extension is not zero */
-                extSz[extIdx] = tmpSz;
+                if (tmpSz > 0xFFFF)
```
JacobBarthelmeh, line 8735: Could use WOLFSSL_MAX_16BIT

JacobBarthelmeh (approved):

PR #9768, opened 2026-02-12
wc_CheckPrivateKey returns NOT_COMPILED_IN for certain gating flags

anhu:
Fixes ZD21187

### DIFF
--- a/wolfcrypt/src/asn.c
+++ b/wolfcrypt/src/asn.c
@@ -8565,6 +8565,9 @@ int wc_CheckPrivateKey(const byte* privKey, word32 privKeySz,
         || ks == RSAPSSk
     #endif
         ) {
+    #if defined(WOLFSSL_RSA_PUBLIC_ONLY) || defined(WOLFSSL_RSA_VERIFY_ONLY)
+        ret = NOT_COMPILED_IN;
+    #else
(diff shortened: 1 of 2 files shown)


anhu:
jenkins retest this please

wolfcrypt/src/asn.c
```diff
@@ -8565,6 +8565,9 @@ int wc_CheckPrivateKey(const byte* privKey, word32 privKeySz,
         || ks == RSAPSSk
     #endif
         ) {
+    #if defined(WOLFSSL_RSA_PUBLIC_ONLY) || defined(WOLFSSL_RSA_VERIFY_ONLY)
```
padelsbach, line 8568: Consider doing this check at the start of the function to bypass the possibility of `BAD_FUNC_ARG`. Eg:

```
#if defined(WOLFSSL_RSA_PUBLIC_ONLY) || defined(WOLFSSL_RSA_VERIFY_ONLY)
    return NOT_COMPILED_IN;
#else
   ...
```

anhu: Yeah, I've considered it and it makes the code very ugly.  I prefer my way.

padelsbach (approved):

PR #9772, opened 2026-02-13
Fix macro protection in ARM SP asm

bigbrett:
Fixes macro protection for `sp_*_cond_add_*` in the ARM SP asm code to prevent unused function warning when used with `RSA_LOW_MEM`. 

Before this PR, using RSA_LOW_MEM with WOLFSSL_SP_CORTEX_M_ASM would result in the following compiler warning:

```
wolfssl/wolfcrypt/src/sp_cortexm.c:29848:39: warning: 'sp_4096_cond_add_64' defined but not used [-Wunused-function]
29848 | WC_OMIT_FRAME_POINTER static sp_digit sp_4096_cond_add_64(sp_digit* r_p,
```

Saw it needs to be fixed in some of the other ARM asm as well.

### DIFF
--- a/wolfcrypt/src/sp_cortexm.c
+++ b/wolfcrypt/src/sp_cortexm.c
@@ -8884,6 +8884,7 @@ int sp_RsaPublic_2048(const byte* in, word32 inLen, const mp_int* em,
 }
 
 #ifndef WOLFSSL_RSA_PUBLIC_ONLY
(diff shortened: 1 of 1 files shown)


bigbrett:
@SparkiDev I believe the changes proposed in this PR actually need to be fixed in your ruby scripts. Therefore closing this PR and assigning you as an FYI.

PR #9780, opened 2026-02-16
Fix PKCS11 object leak in Pkcs11ECDH

mattia-moffa:
# Description

`Pkcs11ECDH` leaked two PKCS11 objects per ECDH operation:

- The derived secret via `C_DeriveKey` is never destroyed after its value is extracted via `C_GetAttributeValue`
- Ephemeral private keys generated by `Pkcs11EcKeyGen` (e.g. for TLS ECDHE) is not cleaned up. `Pkcs11EcKeyGen` intentionally leaves the private key alive for the subsequent `Pkcs11ECDH` call, but `Pkcs11ECDH` only destroys private keys created from raw scalar data (`if (sessionKey)`), not keys found via `Pkcs11FindEccKey`.

Reported via ZD#20700

# Testing

```
./configure --enable-pkcs11
```

### DIFF
--- a/wolfcrypt/src/wc_pkcs11.c
+++ b/wolfcrypt/src/wc_pkcs11.c
@@ -4741,28 +4741,27 @@ int wc_Pkcs11_CryptoDevCb(int devId, wc_CryptoInfo* info, void* ctx)
             ret = NOT_COMPILED_IN;
     #endif
         }
-        else
-            ret = NOT_COMPILED_IN;
-    }
-#ifdef WOLF_CRYPTO_CB_FREE
-    else if (info->algo_type == WC_ALGO_TYPE_FREE) {
+        else if (info->algo_type == WC_ALGO_TYPE_FREE) {
     #ifdef HAVE_ECC
-        if (info->free.algo == WC_ALGO_TYPE_PK &&
-            info->free.type == WC_PK_TYPE_EC_KEYGEN) {
-            ecc_key* key = (ecc_key*)info->free.obj;
-            if (key != NULL && key->devCtx != NULL) {
-                if (token->handle != NULL_PTR) {
-                    CK_OBJECT_HANDLE handle =
-                        (CK_OBJECT_HANDLE)(uintptr_t)key->devCtx;
-                    token->func->C_DestroyObject(token->handle, handle);
+            if (info->free.algo == WC_ALGO_TYPE_PK &&
+                info->free.type == WC_PK_TYPE_EC_KEYGEN) {
+                ecc_key* key = (ecc_key*)info->free.obj;
+                if (key != NULL && key->devCtx != NULL) {
+                    if (token->handle != NULL_PTR) {
+                        CK_OBJECT_HANDLE handle =
+                            (CK_OBJECT_HANDLE)(uintptr_t)key->devCtx;
+                        token->func->C_DestroyObject(token->handle, handle);
+                    }
+                    key->devCtx = NULL;
                 }
-                key->devCtx = NULL;
+                ret = 0;
             }
-            ret = 0;
-        }
     #endif
+        }
+        else {
+            ret = NOT_COMPILED_IN;
+        }
     }
-#endif /* WOLF_CRYPTO_CB_FREE */
 
     return ret;
 }


dgarske:
Jenkins retest this please. FIPS v3 no history

dgarske (dismissed):

wolfcrypt/src/wc_pkcs11.c
```diff
@@ -3085,8 +3091,23 @@ static int Pkcs11ECDH(Pkcs11Session* session, wc_CryptoInfo* info)
                                                           info->pk.ecdh.outlen);
     }
 
-    if (sessionKey)
+    if (secret != CK_INVALID_HANDLE)
+        session->func->C_DestroyObject(session->handle, secret);
+
+    if (sessionKey) {
```
SparkiDev: Please check against NULL.

wolfcrypt/src/wc_pkcs11.c
```diff
@@ -3038,6 +3039,11 @@ static int Pkcs11ECDH(Pkcs11Session* session, wc_CryptoInfo* info)
         else {
             ret = Pkcs11FindEccKey(&privateKey, CKO_PRIVATE_KEY, session,
                                           info->pk.ecdh.public_key, CKA_DERIVE);
+            if (ret == 0) {
```
SparkiDev: The private key lifecycle should be bound to the ECC object.
I don't think the private key should be destroyed here.

wolfcrypt/src/wc_pkcs11.c
```diff
@@ -3085,8 +3091,23 @@ static int Pkcs11ECDH(Pkcs11Session* session, wc_CryptoInfo* info)
                                                           info->pk.ecdh.outlen);
     }
 
-    if (sessionKey)
+    if (secret != CK_INVALID_HANDLE)
+        session->func->C_DestroyObject(session->handle, secret);
```
SparkiDev, line 3097: May want the secret held on the device!
When doing TLS on the device, you don't want the secret to leave it. Instead the secret is used to generate the TLS session keys.

wolfcrypt/src/wc_pkcs11.c
Frauschi: That would be the best option. But AFAICT, we currently don't have logic to handle that case in the `ecc_free()` method. Would be a great addition, as we will also face the same issue once ML-KEM is added to the PKCS#11 interface (private decapsulation key must also be deleted once the handshake is done on the client side — I'm currently working on that ML-KEM integration).

wolfcrypt/src/wc_pkcs11.c
Frauschi: In that case, the lifetime of the secret key object should be bound to the handshake duration, shouldn't it? Then, we can still use it in the future for potential key derivations on the external token via PKCS#11.

mattia-moffa: But do we have a way of binding it to the handshake duration from the wolfCrypt side?

Frauschi: After taking a look, no we don‘t have currently. And adding that would be a much larger effort, probably out of scope for this PR. 

Furthermore, we currently read the derived secret key into memory all the time anyway, so keeping it on the external device for longer is „unnecessary“ at the moment in my opinion. For that to change, we need full support for never exporting the key, keeping only the handle and then perform all the HKDF derivations of the key schedule on the external device via PKCS#11. What do you think @mattia-moffa @SparkiDev?

SparkiDev: Agree that it is a much larger effort.
We can delete it for now until this work is tackled.

wolfcrypt/src/wc_pkcs11.c
SparkiDev: The private key is kept on the device.
Creating a private key as ephemeral in PKCS#11 is not an idea we have supported.
Putting the private key into the session only is the closest I can come up with. That way the private key is destroyed when the session is destroyed.

mattia-moffa: AFAIK, putting the key into the session only is what we're already doing (i.e. `CKA_TOKEN` is not set on the object).

By the way, I made the private key be destroyed when the ECC object is destroyed. @SparkiDev can you take a look at the current state of the PR?

wolfcrypt/src/wc_pkcs11.c
mattia-moffa: Seems reasonable to me.

dgarske (approved):

wolfcrypt/src/wc_pkcs11.c
SparkiDev: Thanks for the change!

PR #9815, opened 2026-02-21
20260221-configure-kernel-mode-defaults

douzzer:
`configure.ac`:
* add a kernel-appropriate subset of `enable-all-crypto` to enable whenever `KERNEL_MODE_DEFAULTS` (almost all of these overlap with FIPS v5/v6 settings);
* move ocspstapling features from `enable-all-crypto` to `enable-all`.

tested with
```
wolfssl-multi-test.sh ...
check-source-text
check-configure
'.*cust-kernel.*'
```
after tweaking multi-test to remove `--enable-all-crypto` from `.*cust-kernel.*'.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1203,6 +1203,8 @@ then
         test "$enable_dtls13" = "" && enable_dtls13=yes
     fi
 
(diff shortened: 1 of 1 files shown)


douzzer:
retest this please
```
FAIL scripts/openssl.test (exit status: 1)
```

PR #9819, opened 2026-02-23
Rust wrapper: add dilithium module

holtrop-wolfssl:
# Description

Rust wrapper: add dilithium module

# Testing

Unit tests

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/wrapper/rust/wolfssl-wolfcrypt/src/dilithium.rs
+++ b/wrapper/rust/wolfssl-wolfcrypt/src/dilithium.rs
@@ -38,7 +38,7 @@ Three security parameter sets are supported, selected via
 # Examples
 
 ```rust
-#[cfg(all(dilithium, dilithium_make_key, dilithium_sign, dilithium_verify))]
+#[cfg(all(dilithium, dilithium_make_key, dilithium_sign, dilithium_verify, random))]
 {
 use wolfssl_wolfcrypt::random::RNG;
 use wolfssl_wolfcrypt::dilithium::Dilithium;
@@ -47,7 +47,7 @@ let mut key = Dilithium::generate(Dilithium::LEVEL_44, &mut rng)
     .expect("Key generation failed");
 let message = b"Hello, ML-DSA!";
 let mut sig = vec![0u8; key.sig_size().expect("sig_size failed")];
-let sig_len = key.sign_msg(message, &mut sig, Some(&mut rng))
+let sig_len = key.sign_msg(message, &mut sig, &mut rng)
     .expect("Signing failed");
 let valid = key.verify_msg(&sig[..sig_len], message)
     .expect("Verification failed");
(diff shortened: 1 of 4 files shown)


JacobBarthelmeh:
@gasbytes this PR looks good and tests are in place but I'd like you to also look it over since more familiar with Rust.

wrapper/rust/wolfssl-wolfcrypt/src/dilithium.rs
```diff
@@ -0,0 +1,1303 @@
...
+    /// #[cfg(all(dilithium, dilithium_make_key_from_seed))]
+    /// {
+    /// use wolfssl_wolfcrypt::dilithium::Dilithium;
+    /// let seed = [0x42u8; 64];
+    /// let key = Dilithium::generate_from_seed(Dilithium::LEVEL_44, &seed)
+    ///     .expect("Error with generate_from_seed()");
+    /// }
+    /// ```
+    #[cfg(dilithium_make_key_from_seed)]
+    pub fn generate_from_seed(level: u8, seed: &[u8]) -> Result<Self, i32> {
```
gasbytes, line 242: maybe here I would add a check that validates that seed is actually `DILITHIUM_PRIV_SEED_SZ`, 64 bytes.
since I don't see any validation later on in `generate_from_seed_ex`.

wrapper/rust/wolfssl-wolfcrypt/src/dilithium.rs
```diff
@@ -0,0 +1,1303 @@
...
+    /// #[cfg(dilithium)]
+    /// {
+    /// use wolfssl_wolfcrypt::dilithium::Dilithium;
+    /// let mut key = Dilithium::new().expect("Error with new()");
+    /// key.set_level(Dilithium::LEVEL_87).expect("Error with set_level()");
+    /// let level = key.get_level().expect("Error with get_level()");
+    /// assert_eq!(level, Dilithium::LEVEL_87);
+    /// }
+    /// ```
+    pub fn get_level(&mut self) -> Result<u8, i32> {
```
gasbytes, line 416: functions like this one, I don't think making `self` mutable is necessary, since the the dilithium struct doesn't actually get modified in here.
I think this happens for other functions too, like `size`, `get_size`, `sig_size` and `verify_*`.

holtrop-wolfssl: Unfortunately since the C API takes a non-const pointer that forces the Rust reference to be mutable as well. It's something I fixed in a few places when doing the original wrappers and then that caused me even more of a headache when supporting FIPS since FIPS v5 still took non-const pointers. Here at least the FIPS v5 issue would not apply, but there would still be a backward compatibility issue if anyone tries to use the Rust crate with a current wolfSSL release.

wrapper/rust/wolfssl-wolfcrypt/src/dilithium.rs
holtrop-wolfssl: When I went to fix this I realized that the C APIs only read 32 bytes. I didn't catch that on the first review of these wrappers (mostly generated by Claude!). But I fixed that now as well as added length checks as you suggested. Good idea!

holtrop-wolfssl:
retest this please (Finished: FAILURE)

PR #9824, opened 2026-02-24
Fix issues in TLS Extension size calculations

embhorn:
# Description

Three GetSize functions in src/tls.c use word16 (16-bit) variables to accumulate extension sizes in loops. With enough extensions, the value wraps past 65535 back to a small number, causing under-allocation of buffers.

Fix applied to all three functions:

* `TLSX_SNI_GetSize()`
* `TLSX_CSR_GetSize_ex()`
* `TLSX_CSR2_GetSize()`

Fixes zd21239

# Testing

Added test case `test_TLSX_SNI_GetSize_overflow`

# Checklist

 - [x] added tests
 - [ ] updated/added doxygen

Credit for finding issue to:
```
Muhammad Arya Arjuna (pelioro)
```

### DIFF
--- a/src/tls.c
+++ b/src/tls.c
@@ -2135,7 +2135,7 @@ static void TLSX_SNI_FreeAll(SNI* list, void* heap)
 }
 
 /** Tells the buffered size of the SNI objects in a list. */
-static word16 TLSX_SNI_GetSize(SNI* list)
+WOLFSSL_TEST_VIS word16 TLSX_SNI_GetSize(SNI* list)
 {
     SNI* sni;
     word32 length = OPAQUE16_LEN; /* list length */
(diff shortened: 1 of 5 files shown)


embhorn:
Jenkins retest this please

tests/api/test_tls_ext.c
```diff
@@ -545,3 +545,137 @@ int test_certificate_authorities_client_hello(void) {
...
+            return list;
+        list = list->next;
+    }
+    return NULL;
+}
+
+/* Replicate the SNI size calculation to test overflow protection.
+ * The fixed TLSX_SNI_GetSize uses word32 internally and returns 0 on
+ * overflow. This helper mirrors that logic so we can verify the behavior. */
+static word16 test_SNI_GetSize(SNI* list)
```
JacobBarthelmeh: For the test cases, we should be testing the calling of functions inside the wolfSSL library. These duplications of the functions from src/*.c are at risk of slowly drifting apart from the actual implementations. There is the visibility macro WOLFSSL_TEST_VIS for internal API that gets called directly in test cases if unable to induce the behavior from a public API.

src/tls.c
```diff
@@ -3237,15 +3241,21 @@ word16 TLSX_CSR_GetSize_ex(CertificateStatusRequest* csr, byte isRequest,
...
-            return OPAQUE8_LEN + OPAQUE24_LEN + csr->ssl->ocspCsrResp[idx].length;
+            size = OPAQUE8_LEN + OPAQUE24_LEN +
+                    csr->ssl->ocspCsrResp[idx].length;
+            if (size > WOLFSSL_MAX_16BIT)
+                return 0;
+            return (word16)size;
         }
-        return (word16)(OPAQUE8_LEN + OPAQUE24_LEN +
-                csr->responses[idx].length);
+        size = OPAQUE8_LEN + OPAQUE24_LEN + csr->responses[idx].length;
```
JacobBarthelmeh: This might be extreme, but is there a case where the word32 csr->responses[idx].length could be WOLFSSL_MAX_32BIT - OPAQUE24_LEN and then the value of "size" since adding OPAQUE8_LEN + OPAQUE24_LEN to the length would get wrapped around to 0 again? I think if adding the check we should account for that edge case scenario also. Maybe instead of the size > WOLFSSL_MAX_16BIT using something like:

```
if (WOLFSSL_MAX_16BIT - OPAQUE8_LEN - OPAQUE24_LEN < csr->responses[idx].length) {
    return 0;
}
size = OPAQUE8_LEN + OPAQUE24_LEN + csr->responses[idx].length;
``` 

Please also use brackets around these if conditions.

JacobBarthelmeh (changes requested):

JacobBarthelmeh (approved):

PR #9828, opened 2026-02-24
tls13: don't create a new suite in CertificateRequest, fallback to WOLFSSL_SUITES(sa->ssl)

rizlik:
# Description

Currently in CertificateRequest we always create a new SigAlgoHash list with all algorithms enabled.
Avoid creating a new one so we fallback to `WOLFSSL_SUITE(ssl)` that either use the one in the main context or the one create by `wolfSSL_set1_sigalgs_list`.


julek-wolfssl (dismissed):

rizlik:
retest this please

PR #9833, opened 2026-02-25
Rust wrapper: add mlkem module

holtrop-wolfssl:
# Description

Rust wrapper: add mlkem module

# Testing

Unit tests

# Checklist

 - [X] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/wrapper/rust/wolfssl-wolfcrypt/tests/test_mlkem.rs
+++ b/wrapper/rust/wolfssl-wolfcrypt/tests/test_mlkem.rs
@@ -65,7 +65,7 @@ fn test_new_ex() {
 fn test_size_queries() {
     common::setup();
     for key_type in [MlKem::TYPE_512, MlKem::TYPE_768, MlKem::TYPE_1024] {
-        let mut key = MlKem::new(key_type).expect("Error with new()");
+        let key = MlKem::new(key_type).expect("Error with new()");
         let pub_size = key.public_key_size().expect("Error with public_key_size()");
         let priv_size = key.private_key_size().expect("Error with private_key_size()");
         let ct_size = key.cipher_text_size().expect("Error with cipher_text_size()");
@@ -175,9 +175,9 @@ fn test_generate_with_random_determinism() {
     // MAKEKEY_RAND_SIZE = 64 bytes
     let rand = [0x42u8; 64];
 
-    let mut key1 = MlKem::generate_with_random(MlKem::TYPE_768, &rand)
+    let key1 = MlKem::generate_with_random(MlKem::TYPE_768, &rand)
         .expect("Error with generate_with_random() first call");
-    let mut key2 = MlKem::generate_with_random(MlKem::TYPE_768, &rand)
+    let key2 = MlKem::generate_with_random(MlKem::TYPE_768, &rand)
         .expect("Error with generate_with_random() second call");
 
     let pub_size = key1.public_key_size().expect("Error with public_key_size()");


holtrop-wolfssl:
retest this please (ERROR: wolfSSL » PRB-140-3-tests #8576 was deleted)

wrapper/rust/wolfssl-wolfcrypt/src/mlkem.rs
```diff
@@ -0,0 +1,785 @@
...
+
+    /// Generate an ML-KEM key pair from caller-supplied random bytes.
+    ///
+    /// Produces the same key pair for a given `(key_type, rand)` pair, enabling
+    /// deterministic key generation. The `rand` buffer must be exactly
+    /// [`MlKem::MAKEKEY_RAND_SIZE`] (64) bytes.
+    ///
+    /// # Parameters
+    ///
+    /// * `key_type`: Key type. One of [`MlKem::TYPE_512`], [`MlKem::TYPE_768`],
```
gasbytes, line 176: here, just like for the dilithium.rs, I think there should be a check for key_type.
either here or in generate_with_random_ex, making sure that is either 512, 768 or 1024.

wrapper/rust/wolfssl-wolfcrypt/src/mlkem.rs
```diff
@@ -0,0 +1,785 @@
...
+        &mut self,
+        ct: &mut [u8],
+        ss: &mut [u8],
+        rng: &mut RNG,
+    ) -> Result<(), i32> {
+        let expected_ct_size = self.cipher_text_size()?;
+        if ct.len() != expected_ct_size {
+            return Err(sys::wolfCrypt_ErrorCodes_BUFFER_E);
+        }
+        if ss.len() != Self::SHARED_SECRET_SIZE {
```
gasbytes, line 480: here maybe, for consistency, could they both get validated against a runtime query (like it's happening for ct.len()) or both against a rust constant.
so adding a shared_secret_size() for the ss check too or just a brief comment explaining why the constant is used directly.

wrapper/rust/wolfssl-wolfcrypt/src/mlkem.rs
```diff
@@ -0,0 +1,785 @@
...
+    pub fn decapsulate(&mut self, ss: &mut [u8], ct: &[u8]) -> Result<(), i32> {
+        if ss.len() != Self::SHARED_SECRET_SIZE {
+            return Err(sys::wolfCrypt_ErrorCodes_BUFFER_E);
+        }
+        let rc = unsafe {
+            sys::wc_MlKemKey_Decapsulate(
+                self.ws_key,
+                ss.as_mut_ptr(),
+                ct.as_ptr(),
+                ct.len() as u32,
```
gasbytes, line 616: maybe here (and other places too) I would consider a more idiomatic way of casting something like: 
u32::try_from(ct.len()).map_err(|_| ...)? (I think this right way of typing this).

in practice this is just to prevent in case ct.len() is > than u32::MAAX, the cipher text for ml-kem I don't think it can be that big, but just for better practice a better casting would be better I think.

wrapper/rust/wolfssl-wolfcrypt/src/mlkem.rs
holtrop-wolfssl: It looks like the C side already does check this parameter in wc_MlKemKey_Init(). That check is probably better anyway since it takes into account different build options the C library can be built with.

wrapper/rust/wolfssl-wolfcrypt/src/mlkem.rs
holtrop-wolfssl: I added some comments; let me know if they seem sufficient to you. The cipher text length is dependent upon the parameter set used while the shared secret length is always the same regardless of the parameter set in use.

wrapper/rust/wolfssl-wolfcrypt/src/mlkem.rs
holtrop-wolfssl: Hmm, yeah I think a whole lot of things would break if we were trying to use buffer sizes > 4GB :)

I don't really consider that a concern worth doing anything about personally. But I can change if we think it is worth the tradeoff of longer runtime and more ROM usage (there are a few places where the same thing is done).

holtrop-wolfssl: I wonder if the compiler would just elide the entire check on a 32-bit target.

holtrop-wolfssl:
retest this please (build removed)

wrapper/rust/wolfssl-wolfcrypt/src/mlkem.rs
gasbytes: > Hmm, yeah I think a whole lot of things would break if we were trying to use buffer sizes > 4GB :)
> 
> I don't really consider that a concern worth doing anything about personally. But I can change if we think it is worth the tradeoff of longer runtime and more ROM usage (there are a few places where the same thing is done).

yeah that makes sense, I looked at it a bit and ml-kem cipher text would technically max out at 1.5k bytes, so the truncation wouldn't really happen I think.
the trade off is not really worth it.
Thanks for the feedback. :)

wrapper/rust/wolfssl-wolfcrypt/src/mlkem.rs
```diff
@@ -0,0 +1,792 @@
...
+    /// Safely free the underlying wolfSSL ML-KEM key context.
+    ///
+    /// This calls `wc_MlKemKey_Delete()`. The Rust Drop trait guarantees this
+    /// is called when the `MlKem` struct goes out of scope.
+    fn drop(&mut self) {
+        unsafe {
+            sys::wc_MlKemKey_Delete(self.ws_key, core::ptr::null_mut());
+        }
+    }
+}
```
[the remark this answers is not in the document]
holtrop-wolfssl: This is completely incorrect. The second parameter to wc_MlKemKey_Delete is not the heap hint.

wrapper/rust/wolfssl-wolfcrypt/src/mlkem.rs
```diff
@@ -0,0 +1,792 @@
...
+    /// * `out`: Output buffer to receive the encoded private key.
+    ///
+    /// # Returns
+    ///
+    /// Returns either Ok(size) containing the number of bytes written or Err(e)
+    /// containing the wolfSSL library error code value.
+    ///
+    /// # Example
+    ///
+    /// ```rust
```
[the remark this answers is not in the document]
holtrop-wolfssl: Ignoring this. The C function already checks that the buffer size is exactly the required size.

wrapper/rust/wolfssl-wolfcrypt/build.rs
```diff
@@ -328,6 +328,9 @@ fn scan_cfg() -> Result<()> {
     check_cfg(&binding, "DILITHIUM_SEED_SZ", "dilithium_make_key_seed_sz");
     check_cfg(&binding, "DILITHIUM_RND_SZ", "dilithium_rnd_sz");
 
+    /* mlkem / ML-KEM */
+    check_cfg(&binding, "wc_MlKemKey_Init", "mlkem");
```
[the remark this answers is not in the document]
holtrop-wolfssl: Ignoring this. If wc_MlKemKey_Init() is defined then _New(), etc... are defined as well.

wrapper/rust/wolfssl-wolfcrypt/src/mlkem.rs
```diff
@@ -0,0 +1,792 @@
...
+        rng: &mut RNG,
+    ) -> Result<(), i32> {
+        // Verify the cipher text length is as expected based on the parameter
+        // set (key type) in use.
+        let expected_ct_size = self.cipher_text_size()?;
+        if ct.len() != expected_ct_size {
+            return Err(sys::wolfCrypt_ErrorCodes_BUFFER_E);
+        }
+        // Verify the shared secret length is as expected.
+        if ss.len() != Self::SHARED_SECRET_SIZE {
```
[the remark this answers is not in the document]
holtrop-wolfssl: I updated the comment to say the expected size is `SHARED_SECRET_SIZE` instead of `shared_secret_size()`.

wrapper/rust/wolfssl-wolfcrypt/src/mlkem.rs
```diff
@@ -0,0 +1,792 @@
...
+            return Err(sys::wolfCrypt_ErrorCodes_BUFFER_E);
+        }
+        let rc = unsafe {
+            sys::wc_MlKemKey_Decapsulate(
+                self.ws_key,
+                ss.as_mut_ptr(),
+                ct.as_ptr(),
+                ct.len() as u32,
+            )
+        };
```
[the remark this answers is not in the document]
holtrop-wolfssl: Ignoring this. The length is checked in encapsulate() because it is not passed to and checked in the C function. In contrast, for decapsulate(), it is passed in and check by the C function.

holtrop-wolfssl:
retest this please (Failed make check with enable all + openssl all)

gasbytes (approved):

PR #9837, opened 2026-02-26
Fix unused variable warning in PKCS7 without WC_RSA_PSS

jackctj117:
This pull request introduces improvements to PKCS#7 support, specifically enabling PKCS#7 functionality without requiring RSA-PSS, and adds a small code guard to prevent unused variable warnings in certain build configurations.

Enhancements to PKCS#7 support:

* Added a new build configuration in `.github/workflows/os-check.yml` to enable PKCS#7 without RSA-PSS, allowing for broader compatibility and testing scenarios.

Code maintenance and build fixes:

* Updated `wc_PKCS7_ParseSignerInfo` in `wolfcrypt/src/pkcs7.c` to explicitly mark `paramsStart` as unused when RSA-PSS is not enabled or RSA is disabled, preventing compiler warnings.

### DIFF
--- a/wolfcrypt/src/pkcs7.c
+++ b/wolfcrypt/src/pkcs7.c
@@ -5437,7 +5437,9 @@ static int wc_PKCS7_ParseSignerInfo(wc_PKCS7* pkcs7, byte* in, word32 inSz,
                 /* Only parse params when still inside the AlgorithmIdentifier;
                  * when optional params are absent, idx is already past the sequence. */
                 else if (algoContentStart + (word32)algoSeqLen > idx) {
+#if defined(WC_RSA_PSS) && !defined(NO_RSA)
                     word32 paramsStart = idx;
+#endif
                     byte paramTag;
                     int paramLen = 0;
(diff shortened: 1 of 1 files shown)


wolfcrypt/src/pkcs7.c
```diff
@@ -5445,6 +5445,9 @@ static int wc_PKCS7_ParseSignerInfo(wc_PKCS7* pkcs7, byte* in, word32 inSz,
                         ret = ASN_PARSE_E;
                     }
                     else {
+#if !defined(WC_RSA_PSS) || defined(NO_RSA)
+                        (void)paramsStart;
```
SparkiDev: Can have it without protection as it won't hurt anything or else put it in an #else after the following code.

SparkiDev (changes requested):

wolfcrypt/src/pkcs7.c
SparkiDev: Scratch that.
Don't declare variable if not used.
That is, move protection around variable declaration.

PR #9843, opened 2026-02-27
Phase 2: PQ in boundary and SHA512 DRBG

kaleb-himes:
# Description

Phase 2 of the upcoming Post Quantum Full FIPS Submission. ML-KEM, ML-DSA, LMS (Verify), XMSS (Verify) added to module boundary along with a shiny new SHA512-DRBG implementation and NIST vector tests for sanity.

Release 1-liners for this PR:

  1. SHA-512 DRBG - entire new DRBG implementation alongside SHA-256                                     
  2. FIPS wrappers - ML-KEM, ML-DSA, LMS, XMSS, SLH-DSA (dozens of new wrapper functions)                
  3. CASTs - 5+ new Conditional Algorithm Self-Tests with embedded KAT vectors                           
  4. New API surface - mu sign/verify, internal interface wrappers, runtime disable/enable functions
  5. PCT code - pairwise consistency tests in MakeKey paths  
  6. SLH-DSA - SHA2 (128, 192, 256) support in addition to SHAKE

# Testing

Many in-house FIPS custom scripts, optest app and harness.

# Checklist

 - [X] added tests
 - [X] updated/added doxygen


kaleb-himes (commented):
** NON-FIPS **

```
./wolfcrypt/benchmark/benchmark -rng -rng-sha512 -rng-init -rng-sha512-init
------------------------------------------------------------------------------
 wolfSSL version 5.8.4
------------------------------------------------------------------------------
Math: 	Multi-Precision: Wolf(SP) word-size=64 bits=4096 sp_int.c
wolfCrypt Benchmark (block bytes 1048576, min 1.0 sec each)
RNG SHA-256 DRBG           205 MiB took 1.021 seconds,  200.853 MiB/s Cycles per byte =  17.50
RNG SHA-512 DRBG           380 MiB took 1.013 seconds,  375.166 MiB/s Cycles per byte =   9.37
RNG      256 SHA256 Init/Free 146504 ops took 1.000 sec, avg 0.007 ms, 146503.616 ops/sec, 3686366718 cycles    25162.2 Cycles/op
RNG      512 SHA512 Init/Free 204287 ops took 1.000 sec, avg 0.005 ms, 204285.977 ops/sec, 3686374128 cycles    18045.1 Cycles/op
Benchmark complete
```

** FIPS **

```
./wolfcrypt/benchmark/benchmark -rng -rng-sha512 -rng-init -rng-sha512-init
------------------------------------------------------------------------------
 wolfSSL version 5.8.4
------------------------------------------------------------------------------
wolfCrypt Benchmark (block bytes 1048576, min 1.0 sec each)
RNG SHA-256 DRBG           205 MiB took 1.021 seconds,  200.763 MiB/s Cycles per byte =  17.51
RNG SHA-512 DRBG           375 MiB took 1.010 seconds,  371.333 MiB/s Cycles per byte =   9.47
RNG      256 SHA256 Init/Free 123726 ops took 1.000 sec, avg 0.008 ms, 123725.735 ops/sec, 3686363516 cycles    29794.6 Cycles/op
RNG      512 SHA512 Init/Free 172608 ops took 1.000 sec, avg 0.006 ms, 172607.506 ops/sec, 3686365894 cycles    21356.9 Cycles/op
Benchmark complete
```

kaleb-himes:
Going to add CAVP support for the perso string and additional Input also. Will keep as a draft for now.

kaleb-himes:
retest this please.

wolfcrypt/src/hash.c
```diff
@@ -498,14 +509,26 @@ int wc_HashGetDigestSize(enum wc_HashType hash_type)
         #endif
             break;
 
-        /* Not Supported */
+    #if defined(WOLFSSL_SHA3) && defined(WOLFSSL_SHAKE128)
+        case WC_HASH_TYPE_SHAKE128:
+            dig_size = 32; /* SHAKE-128: 256-bit output */
```
SparkiDev: Use WC_SHA3_256_DIGEST_SIZE.
You've used the SHA3 define for block size below.

Same for SHAKE256 - WC_SHA3_512_DIGEST_SIZE.

wolfcrypt/src/hash.c
```diff
@@ -804,26 +843,50 @@ int wc_Hash_ex(enum wc_HashType hash_type, const byte* data,
...
-            break;
-        case WC_HASH_TYPE_BLAKE2B:
-            ret = HASH_TYPE_E;
-            break;
-        case WC_HASH_TYPE_BLAKE2S:
-            ret = HASH_TYPE_E;
+    #if defined(WOLFSSL_SHA3) && defined(WOLFSSL_SHAKE128)
+        case WC_HASH_TYPE_SHAKE128:
+        {
+            wc_Shake shake;
```
SparkiDev: Make that dynamic when WOLFSSL_SMALL_STACK.

SparkiDev (dismissed):

wolfcrypt/src/random.c
```diff
@@ -803,6 +935,443 @@ static int Hash_DRBG_Uninstantiate(DRBG_internal* drbg)
...
+    }
+
+    return (compareSum == 0) ? DRBG_SUCCESS : DRBG_FAILURE;
+}
+
+#endif /* WOLFSSL_DRBG_SHA512 */
+
 
 /* FIPS 140-3 IG 10.3.A / SP800-90B Health Tests for Seed Data
  *
```
kaleb-himes, line 1438: Umm... I'm pretty sure I added a BAD_STATE_E check here @wolfSSL-Fenrir-bot, please double check your finding I think you're wrong.

kaleb-himes, line 1438: @wolfSSL-Fenrir-bot valid, I'll add the documentation in next update.

wolfcrypt/src/hash.c
kaleb-himes: Fixed.

wolfcrypt/src/hash.c
kaleb-himes: Fixed.

wolfcrypt/src/random.c
kaleb-himes: @wolfSSL-Fenrir-bot your comments are too verbose and wordy with quoting the entire file instead of just the target lines. Scrolling up and down this page is painfully slow and time consuming. Please shorten up your past commentary and re-review now after the latest updates.

kaleb-himes:
@wolfSSL-Fenrir-bot review

LinuxJedi:
@wolfSSL-Fenrir-bot review

LinuxJedi:
@wolfSSL-Fenrir-bot review

LinuxJedi:
@wolfSSL-Fenrir-bot review

LinuxJedi:
@wolfSSL-Fenrir-bot review

LinuxJedi:
@wolfSSL-Fenrir-bot review

LinuxJedi:
@wolfSSL-Fenrir-bot review

wolfcrypt/src/random.c
```diff
@@ -1778,10 +2705,182 @@ const FLASH_QUALIFIER byte outputB_data[] = {
...
+    #ifdef WOLFSSL_SMALL_STACK_CACHE
+        byte *check512 = rng->health_check_scratch_512;
+        DRBG_SHA512_internal* drbg512 = rng->drbg512_scratch;
+    #else
+        WC_DECLARE_VAR(check512, byte, RNG_HEALTH_TEST_CHECK_SIZE_SHA512, 0);
+        WC_DECLARE_VAR(drbg512, DRBG_SHA512_internal, 1, 0);
+
+        WC_ALLOC_VAR_EX(check512, byte, RNG_HEALTH_TEST_CHECK_SIZE_SHA512,
+            heap, DYNAMIC_TYPE_TMP_BUFFER, return MEMORY_E);
+        WC_ALLOC_VAR_EX(drbg512, DRBG_SHA512_internal,
```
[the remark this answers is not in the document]
kaleb-himes: Legit and fixed (Same for the dilithium comment above)

wolfcrypt/src/rsa.c
```diff
@@ -2123,14 +2221,47 @@ int wc_hash2mgf(enum wc_HashType hType)
     case WC_HASH_TYPE_SHA512_224:
```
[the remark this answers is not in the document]
kaleb-himes: Confirmed legit, fix will be applied in next update.

wolfcrypt/src/dilithium.c
```diff
@@ -10302,6 +10316,30 @@ int wc_dilithium_make_key(dilithium_key* key, WC_RNG* rng)
...
+
+        ret = wc_dilithium_sign_ctx_msg(NULL, 0, pct_msg, sizeof(pct_msg),
+            pct_sig, &pct_sigSz, key, rng);
+
+        if (ret == 0)
+            ret = wc_dilithium_verify_ctx_msg(pct_sig, pct_sigSz,
+                NULL, 0, pct_msg, sizeof(pct_msg), &pct_res, key);
+
+        if (ret == 0 && pct_res != 1)
+            ret = ML_DSA_PCT_E;
```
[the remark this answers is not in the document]
kaleb-himes: Confirmed legit, same pattern exists for ML_KEM and ECC also! Will fix all 3 in next update.

wolfcrypt/src/wc_mlkem.c
```diff
@@ -634,6 +639,35 @@ int wc_MlKemKey_MakeKey(MlKemKey* key, WC_RNG* rng)
...
+
+        if (ret == 0)
+            ret = wc_MlKemKey_Encapsulate(key, pct_ct, pct_ss1, rng);
+
+        if (ret == 0)
+            ret = wc_MlKemKey_Decapsulate(key, pct_ss2, pct_ct, ctSz);
+
+        if (ret == 0) {
+            if (XMEMCMP(pct_ss1, pct_ss2, WC_ML_KEM_SS_SZ) != 0)
+                ret = ML_KEM_PCT_E;
```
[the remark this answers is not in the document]
kaleb-himes: Same as above but for ML_KEM (Key material left over in structure could appear legitimate if caller doesn't pay attention to the return code / check the return code for 0 (success) before attempting to USE that key material. Chances are it would fail at that point but why take the risk? Free, cleanup, zeroise!

wolfcrypt/src/hash.c
kaleb-himes: I think these are resolved so marking resolved @SparkiDev but if you feel they are not please re-open.

wolfcrypt/src/hash.c
kaleb-himes: I think these are resolved so marking resolved @SparkiDev but if you feel they are not please re-open.

tests/api/test_mlkem.c
```diff
@@ -3993,22 +4003,27 @@ int test_wc_mlkem_decap_fo_reject(void)
...
+    PRIVATE_KEY_LOCK();
     ExpectIntNE(XMEMCMP(ssTampered, ss, WC_ML_KEM_SS_SZ), 0);
 
     /* Tamper at byte 0: also must be rejected. */
     XMEMCPY(ctTampered, ct, ctLen);
     ctTampered[0] ^= 0x01;
     XMEMSET(ssTampered, 0, sizeof(ssTampered));
+    PRIVATE_KEY_UNLOCK();
     ExpectIntEQ(wc_MlKemKey_Decapsulate(key, ssTampered, ctTampered, ctLen), 0);
-    ExpectIntNE(XMEMCMP(ssTampered, ss, WC_ML_KEM_SS_SZ), 0);
```
Frauschi, line 4011: Why remove this?

wolfcrypt/src/dilithium.c
```diff
@@ -10310,6 +10324,37 @@ int wc_dilithium_make_key(dilithium_key* key, WC_RNG* rng)
 #endif
     }
 
+#ifdef HAVE_FIPS
+    /* Pairwise Consistency Test (PCT) per FIPS 140-3 / ISO 19790:2012
+     * Section 7.10.3.3 (TE10.35.02): sign with new sk, verify with pk.
+     * Runs on every key generation. */
+    if (ret == 0) {
+        static const byte pct_msg[] = "wolfSSL ML-DSA PCT";
+        byte pct_sig[DILITHIUM_MAX_SIG_SIZE];
```
Frauschi: Use `WC_DECLARE_VAR()`, `WC_ALLOC_VAR()` etc. to place these on the heap with `WOLFSLL_SMALL_STACK`?

wolfcrypt/src/dilithium.c
```diff
@@ -10650,6 +10699,59 @@ int wc_dilithium_sign_ctx_hash_with_seed(const byte* ctx, byte ctxLen,
...
+            (seed == NULL)) {
+        ret = BAD_FUNC_ARG;
+    }
+    if ((ret == 0) && (muLen != DILITHIUM_MU_SZ)) {
+        ret = BAD_FUNC_ARG;
+    }
+
+    if (ret == 0) {
+        /* Build [seed||mu] buffer and call internal sign function. */
+        byte seedMu[DILITHIUM_RND_SZ + DILITHIUM_MU_SZ];
```
Frauschi, line 10745: Should we call `ForceZero()` for `seedMu` after signing?

wolfcrypt/src/dilithium.c
```diff
@@ -10650,6 +10699,59 @@ int wc_dilithium_sign_ctx_hash_with_seed(const byte* ctx, byte ctxLen,
...
+ *  sig         [out]     Buffer to write signature into.
+ *  sigLen      [in/out]  On in, size of buffer.
+ *                        On out, the length of the signature in bytes.
+ *  key         [in]      Dilithium key to use when signing.
+ *  seed        [in]      32-byte random seed (rnd).
+ *  returns BAD_FUNC_ARG when a parameter is NULL or muLen is not 64,
+ *          BUFFER_E when sigLen is too small,
+ *          0 otherwise.
+ */
+int wc_dilithium_sign_mu_with_seed(const byte* mu, word32 muLen,
```
Frauschi, line 10728: The new external_mu API for signing and verifying is currently not tested at all in the unit tests. It will be tested by the NIST test vectors, but should some unit tests be added, too?

wolfcrypt/src/hash.c
```diff
@@ -654,14 +679,26 @@ int wc_HashGetBlockSize(enum wc_HashType hash_type)
         #endif
             break;
 
-        /* Not Supported */
+    #if defined(WOLFSSL_SHA3) && defined(WOLFSSL_SHAKE128)
+        case WC_HASH_TYPE_SHAKE128:
+            block_size = WC_SHA3_128_BLOCK_SIZE;
```
Frauschi, line 691: so SHAKE128 has a SHA3_128 block size, but a SHA3_256 digest size (see above)? Is that correct? Same goes for SHAKE256

wolfcrypt/src/hash.c
```diff
@@ -804,26 +843,76 @@ int wc_Hash_ex(enum wc_HashType hash_type, const byte* data,
...
-            break;
-        case WC_HASH_TYPE_BLAKE2B:
-            ret = HASH_TYPE_E;
-            break;
-        case WC_HASH_TYPE_BLAKE2S:
-            ret = HASH_TYPE_E;
+    #if defined(WOLFSSL_SHA3) && defined(WOLFSSL_SHAKE128)
+        case WC_HASH_TYPE_SHAKE128:
+        {
+        #if defined(WOLFSSL_SMALL_STACK) && !defined(WOLFSSL_NO_MALLOC)
```
Frauschi: Usage of `WC_DECLARE_VAR()`, `WC_ALLOC_VAR()` etc. would simplify this.

Frauschi: same for SHAKE256

wolfcrypt/src/rng_bank.c
```diff
@@ -26,6 +26,33 @@
 #include <wolfssl/wolfcrypt/random.h>
 #include <wolfssl/wolfcrypt/rng_bank.h>
 
+/* Helper to get reseedCtr from the active DRBG, regardless of SHA-256/SHA-512 */
+#ifdef WOLFSSL_DRBG_SHA512
+    #define WC_RNG_BANK_RESEED_CTR(rng_ptr) \
+        (((rng_ptr)->drbgType == WC_DRBG_SHA512) \
+            ? ((struct DRBG_SHA512_internal *)(rng_ptr)->drbg512)->reseedCtr \
+            : ((struct DRBG_internal *)(rng_ptr)->drbg)->reseedCtr)
```
Frauschi, line 39: Skoll: `wolfssl/wolfcrypt/random.h` wraps the SHA-256 DRBG state (`struct DRBG_internal` and the `rng->drbg` field) in `#ifndef NO_SHA256`, and the error guard at the top of random.h now explicitly permits `NO_SHA256 + WOLFSSL_DRBG_SHA512`. The new rng_bank macros, however, only differentiate on `WOLFSSL_DRBG_SHA512` and unconditionally dereference `(rng_ptr)->drbg` / cast to `struct DRBG_internal *` in the `WOLFSSL_DRBG_SHA512` branch. In the `NO_SHA256 + WOLFSSL_DRBG_SHA512` configuration neither the struct nor the field exists, so the ternary's false branch will fail to compile. The guard should also exclude the SHA-256 branch when `NO_SHA256` is defined.

wolfcrypt/src/rsa.c
```diff
@@ -1137,6 +1137,58 @@ static int RsaMGF1(enum wc_HashType hType, byte* seed, word32 seedSz,
...
 #endif /* SHA2 Hashes */
 
+#if defined(WOLFSSL_SHA3) && \
+    (defined(WOLFSSL_SHAKE128) || defined(WOLFSSL_SHAKE256))
+/* SHAKE XOF used directly as mask generation function (not MGF1).
+ * Per FIPS 186-5, SHAKE can be used as the MGF for RSA-PSS. */
+static int RsaMGF_SHAKE(enum wc_HashType shakeType, byte* seed, word32 seedSz,
+                         byte* out, word32 outSz, void* heap)
+{
+    wc_Shake shake;
```
Frauschi: SMALL_STACK handling?

wolfcrypt/src/wc_mlkem.c
```diff
@@ -636,6 +641,42 @@ int wc_MlKemKey_MakeKey(MlKemKey* key, WC_RNG* rng)
         ret = wc_KyberKey_MakeKeyWithRandom(key, rand, sizeof(rand));
     }
 
+#ifdef HAVE_FIPS
+    /* Pairwise Consistency Test (PCT) per FIPS 140-3 / ISO 19790:2012
+     * Section 7.10.3.3: encapsulate with ek, decapsulate with dk,
+     * verify shared secrets match. */
+    if (ret == 0) {
+        byte pct_ct[WC_ML_KEM_MAX_CIPHER_TEXT_SIZE];
```
Frauschi: SMALL_STACK handling?

wolfssl/wolfcrypt/error-crypt.h
```diff
@@ -312,10 +314,14 @@ enum wolfCrypt_ErrorCodes {
...
     ALREADY_E           = -1007, /* Operation was redundant or preempted */
-
-    SEQ_OVERFLOW_E      = -1008, /* Sequence counter would overflow */
-    WC_SPAN2_LAST_E     = -1008, /* Update to indicate last used error code */
-    WC_LAST_E           = -1008, /* the last code used either here or in
+    ML_KEM_PCT_E        = -1008, /* ML-KEM Pairwise Consistency Test failure */
+    ML_DSA_PCT_E        = -1009, /* ML-DSA Pairwise Consistency Test failure */
+    DRBG_SHA512_KAT_FIPS_E = -1010, /* SHA-512 DRBG KAT failure */
+
+    SEQ_OVERFLOW_E      = -1011, /* Sequence counter would overflow */
```
Frauschi: `SEQ_OVERFLOW_E` and `WC_SPAN2_LAST_E` values are silently changed. Is this fine?

wolfssl/wolfcrypt/wc_lms.h
```diff
@@ -408,10 +457,16 @@ typedef struct LmsState {
 #endif
     /* LMS parameters. */
     const LmsParams* params;
-    /* Hash algorithm. */
+    /* Hash algorithm (SHA-256). */
     wc_Sha256 hash;
-    /* Hash algorithm for calculating K. */
+    /* Hash algorithm for calculating K (SHA-256). */
     wc_Sha256 hash_k;
+#ifdef WOLFSSL_LMS_SHAKE256
```
Frauschi: An LMS key can be either SHA or SHAKE based at a time, can't it? So maybe pack these in a union to keep the key object size smaller?

wolfcrypt/src/wc_lms_impl.c
```diff
@@ -2503,6 +2796,18 @@ static int wc_lms_verify(LmsState* state, const byte* pub, const byte* msg,
         ret = wc_lmots_calc_kc(state, pub + LMS_TYPE_LEN, msg, msgSz,
             sig_lmots, kc);
     }
+    if (ret == 0) {
+        /* Algorithm 6a. Step 3.d-e: Check LMS type in signature matches
+         * the expected type from the public key. */
+        const byte* sig_lms_type = sig + LMS_Q_LEN + LMS_TYPE_LEN +
```
Frauschi, line 2828: Skoll: 
The new RFC 8554 Step 3.d-e check indexes into `sig` at offset `LMS_Q_LEN + LMS_TYPE_LEN + params->hash_len + params->p * params->hash_len` and reads 4 bytes via `ato32`. The earlier `wc_lmots_calc_kc` call is expected to have validated sigSz, but adding a second independent offset read without an explicit size check here makes the code more fragile and harder to audit. Please either add an explicit `if (sigSz < requiredOffset + 4) return BUFFER_E;` at the top of this block, or add a short comment referencing the function that performed the earlier length validation. Also verify that the mask `params->lmsType & LMS_H_W_MASK` is correct — the encoded LMS type in an RFC 8554 signature uses the full identifier (e.g., LMS_SHA256_M32_H5 = 5, LMS_SHAKE_M32_H5 = 0x0f | LMS_SHAKE256), and masking off the hash-algorithm bits (`LMS_HASH_MASK` = 0xf000) before comparison could cause a SHAKE signature to be accepted against a SHA-256 public key (and vice versa).

wolfcrypt/src/wc_slhdsa.c
```diff
@@ -578,25 +609,751 @@ static int slhdsakey_hash_shake_4(wc_Shake* shake, const byte* data1,
...
+{
+    int ret;
+    byte address[SLHDSA_HAC_SZ];
+
+    (void)pk_seed;
+
+    /* Encode compressed address. */
+    HA_Encode_Compressed(adrs, address);
+
+    if (n == 16) {
```
Frauschi: Do we have suitable macros for these "magic numbers"?

wolfcrypt/src/random.c
```diff
@@ -1584,20 +2478,51 @@ int wc_FreeRng(WC_RNG* rng)
         wc_MemZero_Check(rng->drbg, sizeof(DRBG_internal));
     #endif
         rng->drbg = NULL;
-    }
 
     #ifdef WOLFSSL_SMALL_STACK_CACHE
-    if (rng->drbg_scratch != NULL) {
-        if (Hash_DRBG_Uninstantiate((DRBG_internal *)rng->drbg_scratch) != DRBG_SUCCESS)
+        if (rng->drbg_scratch != NULL) {
```
Frauschi: Skoll:
The SHA-256 cleanup block in the new `wc_FreeRng` only frees `rng->drbg_scratch` and `rng->health_check_scratch` when `rng->drbg != NULL`. Previously these were freed unconditionally. The allocations are always paired with `rng->drbg` so the current behavior is correct, but the pattern is brittle - any future path that nulls out `rng->drbg` early (e.g., a partial-construction failure) would leak the scratch buffers.

Frauschi (dismissed):
Some questions, some suggestions and some actual findings. Reviewed both manually and with Skoll.

tests/api/test_hash.c
```diff
@@ -182,7 +198,7 @@ static const int sizeNotSupportedHash[] = {
     WC_HASH_TYPE_MAX + 1
 };
 static const int sizeNotSupportedHashLen = (sizeof(sizeNotSupportedHash) /
-                                            sizeof(enum wc_HashType));
+                                            sizeof(int));
```
JacobBarthelmeh: It's not guaranteed that the sizeof the enum is equivalent to the sizeof int.

wolfcrypt/src/hash.c
```diff
@@ -1564,30 +1697,13 @@ int wc_HashSetFlags(wc_HashAlg* hash, enum wc_HashType type, word32 flags)
 
         /* not supported */
         case WC_HASH_TYPE_MD5_SHA:
```
JacobBarthelmeh, line 1692: I think each of these case's might have been filled out to avoid fall through warnings? Unless this is causing an issue with the build lets leave it as is.

JacobBarthelmeh: Same for later on line 1683

tests/api/test_mlkem.c
kaleb-himes: Such a loaded question but a great one!!

We removed that line because it was flaky at the byte-0 position, and the assertion isn't required there.

Quick refresher: ML-KEM decapsulation never returns an error on a bad ciphertext. By design (FIPS 203), it always returns success and hands back either the real shared secret (good ciphertext) or a pseudo-random "junk" secret (tampered ciphertext). That's "implicit rejection."

Part of the ML-KEM ciphertext is lossy-compressed, meaning some bit flips get rounded away during decompression and effectively disappear. Byte 0 sits in that compressed region, so the bit we flip there sometimes gets absorbed, decapsulation recovers the original message, and you get the real shared secret back instead of a junk one. That's not a bug, it's just how the compression works.

So at byte 0, "the two secrets must differ" isn't actually guaranteed (flaky test design), it depends on whether that bit survives decompression or not. The byte-32 case (which we kept fully intact) lives outside the compressed region, so the flip always sticks and that case still proves the rejection path produces a different secret.

What is guaranteed at byte 0 is that decapsulation still returns 0, which is the assertion we kept. No loss of real coverage, no more flaky test results.

Frauschi: Thanks for the explanation, sounds reasonable to me. I actually didn't know these details about the ML-KEM ciphertexts, so TIL. 👍🏻

wolfcrypt/src/dilithium.c
kaleb-himes: Good catch! Fixed in next update.

wolfcrypt/src/dilithium.c
kaleb-himes: Good catch! Fixed in next update.

wolfcrypt/src/dilithium.c
kaleb-himes: Oh absolutely, great feedback, thank you!

kaleb-himes: Gross... lol, working on this now and if I only tested ML-DSA-65 it will add 50KB just in raw vectors. We never just test one size (AES-128) and say the rest are good, we always test all sizes. To add tests for ML-DSA-44, 65 and 87 it's going to add ~170KB just to put the vectors in the files haha! Fun times.

Frauschi: Maybe for the general tests, we only add a small subset of the required test and build on the much larger FIPS testbed for thorough testing of all the different variants and edge-cases?

kaleb-himes: Tests added, fixed in next update.

wolfcrypt/src/hash.c
kaleb-himes: Yes, it's correct, those WC_SHA3_* macros are named for the Keccak rate parameter, not for any specific SHA-3 hash. SHAKE128 uses the 1344-bit rate (168 B block, WC_SHA3_128_BLOCK_SIZE) and produces a 256-bit default digest (WC_SHA3_256_DIGEST_SIZE); SHAKE256 uses the 1088-bit rate (136 B, same as SHA3-256) and produces a 512-bit default digest. That's per FIPS 202; the naming just looks mismatched. I've added a comment in both spots so future readers don't have to re-derive it from this feedback history! Fixed in next update.

wolfcrypt/src/hash.c
kaleb-himes: Agreed, also noticed a bit of a redundancy here, we don't need to check both WOLFSSL_SMALL_STACK and WOLFSSL_NO_MALLOC because the macros WC_DECLARE...WC_ALLOC... handle all of that (WOLFSSL_NO_MALLOC) behind the scenes. I removed the redundant checks and just trusted in the macros to get it sorted in the updates.

wolfcrypt/src/rng_bank.c
kaleb-himes: Fantastic catch. Fixed in next update!

wolfcrypt/src/rsa.c
kaleb-himes: Too easy. Fixed in next update.

wolfcrypt/src/wc_mlkem.c
kaleb-himes: Too easy, fixed next update.

wolfssl/wolfcrypt/error-crypt.h
kaleb-himes: Got it. Fixed in next update.

wolfssl/wolfcrypt/wc_lms.h
kaleb-himes: Agreed but I don't want to get backed into a C11 corner with HAVE_ANONYMOUS_INLINE_AGGREGATES so to keep it portable this is going to be a ~40 call site update give or take? Large change in next update from this item.

wolfcrypt/src/wc_lms_impl.c
kaleb-himes: Has a slight implication of making wc_hss_verify() a formal WOLFSSL_API to snake that through but I'm good with it since it's new enough code we can still be a bit flexible without worrying about severe ABI breakage. Better now than later. I will add this change in the next update.

wolfcrypt/src/wc_slhdsa.c
kaleb-himes: Oh I usually am so good about avoiding magic numbers! Thank you so much for catching this. Absolutely will fix in next update.

kaleb-himes: Aha, yeah we didn't have any "good macros" in the header yet, must be why I did that and it was probably only meant to be temporary till I added the macros then I forgot. We had some macros of the right size but semantically they were incorrectly named (seed lengths). I'll add new macros tht are SLHDSA param n specific and update the seed macros to use those, that's a better way I think.

wolfcrypt/src/random.c
kaleb-himes: Good call. I'll hoist those out of that guard to remove the brittle pattern.

tests/api/test_hash.c
kaleb-himes: You're right, maybe instead of computing this with a divide by some value we just say sizeof(sizeSupportedHash[0])? I actually see this pattern in 5 other places in this file, I'll just update all of them while I'm at it.

wolfcrypt/src/hash.c
kaleb-himes: Confirmed! That was exactly it. The explicit empty cases for WC_HASH_TYPE_MD5_SHA / MD2 / MD4 / BLAKE2B / BLAKE2S and the explicit handling of WC_HASH_TYPE_SM3 are intentional,  they avoid -Wimplicit-fallthrough warnings (and possibly clang-analyzer complaints about unhandled enum values in a switch)!

wolfcrypt/src/rng_bank.c
kaleb-himes: Fixed a few other call sites as well that were not being properly handled when one or the other DRBG was disabled at build time.

wolfcrypt/src/dilithium.c
kaleb-himes: Changes Pushed. Marking resolved.

wolfcrypt/src/dilithium.c
kaleb-himes: Changes Pushed. Marking resolved.

wolfcrypt/src/dilithium.c
kaleb-himes: Changes Pushed. Marking resolved.

wolfssl/wolfcrypt/error-crypt.h
kaleb-himes: Changes Pushed. Marking resolved.

wolfssl/wolfcrypt/wc_lms.h
kaleb-himes: Changes Pushed. Marking resolved.

wolfcrypt/src/wc_lms_impl.c
kaleb-himes: Changes Pushed. Marking resolved.

wolfcrypt/src/wc_slhdsa.c
kaleb-himes: Changes Pushed. Marking resolved.

wolfcrypt/src/random.c
kaleb-himes: Changes Pushed. Marking resolved.

tests/api/test_hash.c
kaleb-himes: Changes Pushed. Marking resolved.

wolfcrypt/src/hash.c
kaleb-himes: Changes Pushed. Marking resolved.

wolfcrypt/src/hash.c
kaleb-himes: Changes Pushed. Marking resolved.

Frauschi (approved):

Frauschi (changes requested):
The failing hostap tests seem to be related to changes in the hash code

kaleb-himes:
The failing hostap changes are unrelated. See:

https://github.com/wolfSSL/wolfssl/pull/9668 - Disables broken hostap tests (Merged Jan 14, 2026)
https://github.com/wolfSSL/wolfssl/pull/10058 - Re-introduces same broken tests (Merged Apr 23, 2026)
Master is currently failing
https://github.com/wolfSSL/wolfssl/pull/10296 - Fixes broken tests (Open PR)

Frauschi (approved):

JacobBarthelmeh (approved):

wolfcrypt/benchmark/benchmark.c
```diff
@@ -12274,6 +12739,78 @@ void bench_slhdsa(enum SlhDsaParam param)
...
+    /* Pre-hash interface: hash message, then sign the hash. */
+    PRIVATE_KEY_UNLOCK();
+    bench_stats_start(&count, &start);
+    do {
+        sigLen = WC_SLHDSA_MAX_SIG_LEN;
+        ret = wc_SlhDsaKey_SignHashDeterministic(key, ctx, 0, msg,
+            (word32)sizeof(msg), WC_HASH_TYPE_SHA256, sig, &sigLen);
+        if (ret != 0) {
+            goto exit;
+        }
```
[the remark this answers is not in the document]
kaleb-himes: Noted, will fix in 10281, not critical enough to block PR on.

src/include.am
```diff
@@ -439,7 +439,7 @@ src_libwolfssl@LIBSUFFIX@_la_SOURCES += wolfcrypt/src/fips.c \
 src_libwolfssl@LIBSUFFIX@_la_SOURCES += wolfcrypt/src/wolfcrypt_last.c
 endif BUILD_FIPS_V5
 
-if BUILD_FIPS_V6_PLUS
+if BUILD_FIPS_V6
 # FIPS 140-3 SRTP-KDF first file
 src_libwolfssl@LIBSUFFIX@_la_SOURCES += \
                wolfcrypt/src/wolfcrypt_first.c
```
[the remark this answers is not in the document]
kaleb-himes: Noted, will fix in 10281, not critical enough to block PR on.

wolfcrypt/src/error.c
```diff
@@ -668,6 +668,30 @@ const char* wc_GetErrorString(int error)
...
+        return "wolfCrypt FIPS LMS Known Answer Test Failure";
+
+    case XMSS_KAT_FIPS_E:
+        return "wolfCrypt FIPS XMSS Known Answer Test Failure";
+
+    case ML_KEM_PCT_E:
+        return "wolfcrypt ML-KEM Pairwise Consistency Test Failure";
+
+    case ML_DSA_PCT_E:
+        return "wolfcrypt ML-DSA Pairwise Consistency Test Failure";
```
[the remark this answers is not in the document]
kaleb-himes: Noted, will fix in 10281, not critical enough to block PR on.

wolfcrypt/src/hmac.c
```diff
@@ -478,6 +526,12 @@ int wc_HmacSetKey_ex(Hmac* hmac, int type, const byte* key, word32 length,
        !(type == WC_MD5 || type == WC_SHA ||
     #ifdef WOLFSSL_SM3
             type == WC_SM3 ||
+    #endif
+    #ifndef WOLFSSL_NOSHA512_224
+            type == WC_SHA512_224 ||
+    #endif
+    #ifndef WOLFSSL_NOSHA512_256
+            type == WC_SHA512_256 ||
     #endif
```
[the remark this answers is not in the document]
kaleb-himes: This is incorrect since we're actively testing enabling and disabling SHA512 and SHA256 (and cooresponding DRBG's), it is sufficient to be checking WOLFSSL_NOSHA512_224 only.

wolfcrypt/src/rsa.c
```diff
@@ -2115,22 +2217,65 @@ int wc_hash2mgf(enum wc_HashType hType)
...
+#ifdef WOLFSSL_SHAKE128
+    case WC_HASH_TYPE_SHAKE128:
+        return WC_MGF1SHAKE128;
+#else
     case WC_HASH_TYPE_SHAKE128:
+        break;
+#endif
+#ifdef WOLFSSL_SHAKE256
     case WC_HASH_TYPE_SHAKE256:
+        return WC_MGF1SHAKE256;
```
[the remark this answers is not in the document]
kaleb-himes: Co-pilot you're confused... and suggesting a bad new inconsistent pattern! For shame haha.

wc_hash2mgf() is an MGF1 helper by contract. Every other branch returns WC_MGF1<hash> SHA-1, SHA-2 family, SHA-3 family. Mapping SHAKE128 to WC_MGF1SHAKE128 follows that pattern. Making SHAKE the one exception that returns a non-MGF1 value would be surprising and inconsistent... I will NOT pursue this suggestion.

wolfcrypt/benchmark/benchmark.c
```diff
@@ -12274,6 +12739,78 @@ void bench_slhdsa(enum SlhDsaParam param)
...
     bench_stats_asym_finish(name, len, "verify", 0, count, start, ret);
 
+    /* Internal interface: sign M' directly (no M' construction). */
+    PRIVATE_KEY_UNLOCK();
+    bench_stats_start(&count, &start);
+    do {
+        sigLen = WC_SLHDSA_MAX_SIG_LEN;
+        ret = wc_SlhDsaKey_SignMsgDeterministic(key, msg,
+            (word32)sizeof(msg), sig, &sigLen);
+        if (ret != 0) {
```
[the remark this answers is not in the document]
kaleb-himes: Noted, will fix in 10281, not critical enough to block PR on.

wolfcrypt/src/hmac.c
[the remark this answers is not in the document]
kaleb-himes: Little more substance other than just "its wrong because we're testing it already to make sure this isn't the case"...

Copilot is wrong because: WC_SHA512_224 / WC_SHA512_256 are always defined (via sha512.h when WOLFSSL_SHA512 is on, or via the hmac.h enum when it's off), both already gated by WOLFSSL_NOSHA512_224/256, so the existing single guard is correct and adding #ifdef WOLFSSL_SHA512 would actually narrow the accepted-type set on no-SHA-512 builds where the identifiers still exists and could actually break builds, not "fix" any specific scenario.

PR #9848, opened 2026-03-02
Fix for DTLS1.3 HRR group handling

Frauschi:
When a server uses an HRR to negotiate the key exchange group to use, the selected group is advertised in the HRR key share extension. Furthermore, this group is also stored in the Cookie that is sent to the client. When the server receives the second CH, the group used in the key share extension MUST be the one of the HRR.

For stateless DTLS servers, the handling of this check had a bug. The key share group of the HRR is stored in the ssl->hrr_keyshare_group variable and is checked against the received key share of the second CH. However, in the stateless server case, another CH message may be received inbetween the two CH messages of the desired client, potentially overwriting the ssl->hrr_keyshare_group variable. This then causes handshake failures when the ssl->hrr_keyshare_group variable contains another group than the second CH message of the desired client.

To fix this, two changes are conducted:
1. Disable the ssl->hrr_keyshare_group check for stateless DTLS 1.3 servers. As long as the server is stateless, CHs from multiple clients may be received that individually cause HRRs with different groups. For each of these clients, the HRR group is properly stored in the cookie.
2. When a valid cookie is received from the client, the server becomes stateful. In this case, we now parse the cookie for a stored HRR group. If present, we reset the ssl->hrr_keyshare_group variable to this group to ensure the error checks succeed.

A new test is added to check for this behavior.

Identified while working on #9732.

### DIFF
--- a/src/tls.c
+++ b/src/tls.c
@@ -10216,20 +10216,6 @@ int TLSX_KeyShare_Parse_ClientHello(const WOLFSSL* ssl,
         offset += ret;
     }
 
-    if (ssl->hrr_keyshare_group != 0) {
-        /*
-         * https://datatracker.ietf.org/doc/html/rfc8446#section-4.2.8
-         *   when sending the new ClientHello, the client MUST
-         *   replace the original "key_share" extension with one containing only a
-         *   new KeyShareEntry for the group indicated in the selected_group field
-         *   of the triggering HelloRetryRequest
-         */
-        if (seenGroupsCnt != 1 || seenGroups[0] != ssl->hrr_keyshare_group) {
-            WOLFSSL_ERROR_VERBOSE(BAD_KEY_SHARE_DATA);
-            return BAD_KEY_SHARE_DATA;
-        }
-    }
-
     return 0;
 }
 
--- a/src/tls13.c
+++ b/src/tls13.c
@@ -6502,6 +6502,8 @@ static int RestartHandshakeHashWithCookie(WOLFSSL* ssl, Cookie* cookie)
         hrrIdx += 2;
         c16toa(OPAQUE16_LEN, hrr + hrrIdx);
         hrrIdx += 2;
+        /* Restore the HRR key share group from the cookie. */
+        ato16(cookieData + idx, &ssl->hrr_keyshare_group);
         hrr[hrrIdx++] = cookieData[idx++];
         hrr[hrrIdx++] = cookieData[idx++];
     }
@@ -7015,6 +7017,29 @@ int DoTls13ClientHello(WOLFSSL* ssl, const byte* input, word32* inOutIdx,
     }
 #endif
 
+#ifdef HAVE_SUPPORTED_CURVES
+    if (ssl->hrr_keyshare_group != 0) {
+        /*
+         * https://datatracker.ietf.org/doc/html/rfc8446#section-4.2.8
+         *   when sending the new ClientHello, the client MUST
+         *   replace the original "key_share" extension with one containing only
+         *   a new KeyShareEntry for the group indicated in the selected_group
+         *   field of the triggering HelloRetryRequest.
+         */
+        TLSX* extension = TLSX_Find(ssl->extensions, TLSX_KEY_SHARE);
+        if (extension != NULL) {
+            KeyShareEntry* kse = (KeyShareEntry*)extension->data;
+            /* Exactly one KeyShareEntry with the HRR group must be present. */
+            if (kse == NULL || kse->next != NULL ||
+                                        kse->group != ssl->hrr_keyshare_group) {
+                ERROR_OUT(BAD_KEY_SHARE_DATA, exit_dch);
+            }
+        }
+        else
+            ERROR_OUT(BAD_KEY_SHARE_DATA, exit_dch);
+    }
+#endif
+
 #if defined(HAVE_ECH)
     /* hash clientHelloInner to hsHashesEch independently since it can't include
      * the HRR */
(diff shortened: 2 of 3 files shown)


src/dtls.c
```diff
@@ -295,6 +295,42 @@ static int CheckDtlsCookie(const WOLFSSL* ssl, WolfSSL_CH* ch,
...
+    word16 len;
+    const byte* cookie = NULL;
+
+    if (ch->cookieExt.size < OPAQUE16_LEN + 1)
+        return BUFFER_E;
+    ato16(ch->cookieExt.elements, &len);
+    if (ch->cookieExt.size - OPAQUE16_LEN != len)
+        return BUFFER_E;
+    cookie = ch->cookieExt.elements + OPAQUE16_LEN;
+    ret = TlsCheckCookie(ssl, cookie, len);
```
rizlik: At this point, the cookie is already parsed and validated, don't re-parse it.
I suggest to refactor the function in something like get_keyshare_from_cookie(cookie, cookie_sz)

src/dtls.c
```diff
@@ -1041,6 +1077,10 @@ int DoClientHelloStateless(WOLFSSL* ssl, const byte* input, word32 helloSz,
                 e = Dtls13GetEpoch(ssl, ssl->keys.curEpoch64);
                 if (e != NULL)
                     XMEMSET(e->window, 0xFF, sizeof(e->window));
+
```
rizlik: here we already know that the ticket is good and we already parsed it in ch. 
It's probably better to just extract the keyshare instead of re-parsing the cookie

tests/api/test_dtls.c
julek-wolfssl, line 1: Please move this test case into `test_wolfSSL_dtls_stateless`.

src/dtls.c
```diff
@@ -295,6 +295,42 @@ static int CheckDtlsCookie(const WOLFSSL* ssl, WolfSSL_CH* ch,
...
+    /* Check if the cookie has a key share group */
+    if (len > cookie[0] + 4) {
+        word16 keyShareGroup = 0;
+        ato16(cookie + 3 + cookie[0], &keyShareGroup);
+
+        /* The key share group in the cookie is the group selected by the
+         * server in the HelloRetryRequest. Hence, the client must use this
+         * group in the second ClientHello.
+         */
+        ssl->hrr_keyshare_group = keyShareGroup;
```
julek-wolfssl: Let's not store state here. This assignment should be moved to the stateful cookie parsing (if we keep the variable).

src/dtls.c
Frauschi: Fixed by simplifying the logic in the new `TLSX_Cookie_RestoreHrrGroup()`

src/dtls.c
Frauschi: Fixed, see above.

tests/api/test_dtls.c
Frauschi: Moved in the same group as new `test_wolfSSL_dtls_stateless_hrr_group`

src/dtls.c
Frauschi: The cookie parsing within `TLSX_Cookie_RestoreHrrGroup()` is now called in `TLSX_Cookie_Parse()` for stateful DTLS servers (as we already know that we received a valid cookie and switched from stateless to stateful DTLS processing).

Frauschi:
@julek-wolfssl @rizlik I addressed all your comments.

src/tls.c
```diff
@@ -10217,6 +10275,18 @@ int TLSX_KeyShare_Parse_ClientHello(const WOLFSSL* ssl,
...
+         * may have already been set without further proceeding the handshake,
+         * so we never received the second ClientHello with the selected group.
+         * Hence, when we receive another ClientHello message, we do not
+         * consider that as an error here.
+         */
+        if (ssl->options.dtls && IsAtLeastTLSv1_3(ssl->version) &&
+            !ssl->options.dtlsStateful) {
+            return 0;
+        }
+#endif
```
julek-wolfssl: I would prefer to error out when `ssl->options.dtls` because `ssl->hrr_keyshare_group` should not be set before anyway right? Am I missing an edge case?

src/tls.c
```diff
@@ -7248,6 +7248,55 @@ static int TLSX_Cookie_Write(Cookie* cookie, byte* output, byte msgType,
...
 }
 
+#if defined(WOLFSSL_DTLS13) && defined (WOLFSSL_SEND_HRR_COOKIE)
+/* Extract the key share group from the cookie and store it in the
+ * ssl session for later checks.
+ *
+ * ssl    The SSL/TLS object.
+ * returns 0 on success and other values indicate failure.
+ */
+static int TLSX_Cookie_RestoreHrrGroup(WOLFSSL* ssl)
```
julek-wolfssl: Could this be restored in `RestartHandshakeHashWithCookie`? That will combine the cookie parsing logic in one place.

julek-wolfssl (changes requested):

src/tls.c
Frauschi: I removed this part of the code and instead ensured that the `ssl->hrr_keyshare_group` is never set to a value `!= 0` in the DTLS stateless case. In this case, the HRR key share group is now only stored in the cookie. Hence, no state is preserved.

This is now also checked in the `test_wolfSSL_dtls_stateless_hrr_group` test by calculating and comparing the hash over the server-side `ssl` structure using `test_wolfSSL_dtls_stateless_HashWOLFSSL()`.

src/tls.c
Frauschi: Unfortunately no. The `RestartHandshakeHashWithCookie()` method is called after parsing the extensions of the ClientHello. Hence, the incoming key share, where the check for the HRR group is placed, is performed before this function. Hence, we have to restore the HRR group ahead of the check.

src/internal.c
```diff
@@ -18566,8 +18566,12 @@ int DoHandShakeMsgType(WOLFSSL* ssl, byte* input, word32* inOutIdx,
...
                 (defined(HAVE_ED25519) && !defined(NO_ED25519_CLIENT_AUTH)) || \
                 (defined(HAVE_ED448) && !defined(NO_ED448_CLIENT_AUTH)))
-        if (ssl->options.resuming || !ssl->options.verifyPeer || \
-                     !IsAtLeastTLSv1_2(ssl) || IsAtLeastTLSv1_3(ssl->version)) {
+        if ((ssl->options.resuming || !ssl->options.verifyPeer ||
+               !IsAtLeastTLSv1_2(ssl) || IsAtLeastTLSv1_3(ssl->version))
+        #ifdef WOLFSSL_DTLS
+               && !ssl->options.dtls
+        #endif
+               ) {
```
julek-wolfssl: I'm not sure this is necessary. I think this needs to happen for dtls too.

Frauschi: Fixed in the latest commit

julek-wolfssl (approved):

PR #9849, opened 2026-03-02
Fix typo in CURVE25519_SMALL

Frauschi:
While working on zd#21262, I found that there are numerous places in the code base with the typo `CURVED25519_SMALL` instead of `CURVE25519_SMALL` (note the additional "D" before the 25519 number). Hence, for these systems, the actual small implementation of Curve25519 was never enabled, as the wrongly typed macro had no effect.

This PR fixes the typo in all places in the code base.


Frauschi:
Closing as `CURVED25519_SMALL` actually enables both `CURVE25519_SMALL` and `ED25519_SMALL`. Haven't figured that out...

PR #9851, opened 2026-03-02
Setkey/Export callbacks

night1rider:
Add `WOLF_CRYPTO_CB_SETKEY` and `WOLF_CRYPTO_CB_EXPORT_KEY` generic crypto callback utilities.

Both callbacks use a temporary key struct pattern to bridge between raw key bytes and hardware. The temporary key is always initialized with `INVALID_DEVID` so it never recurses into the callback — it serves purely as a software intermediary.

**SetKey (Import):** When a key import function (e.g. `wc_ecc_import_x963_ex2`) is called on a hardware-bound key, a temporary key is created and the raw key material is imported into it via the normal software path. The callback handler then receives both the hardware destination key and the fully-parsed temporary key. The handler can call any wolfSSL export function on the temporary key (`wc_RsaKeyToDer`, `wc_ecc_export_x963`, raw component getters, etc.) to extract the key in whatever format the hardware needs, program it, and associate it with the destination key. The temporary key is freed after the callback returns.

SetKey callbacks are wired into: `wc_AesSetKey`, `wc_HmacSetKey`, `wc_RsaPublicKeyDecodeRaw`, `wc_RsaPrivateKeyDecodeRaw`, `wc_RsaPrivateKeyDecode`, `wc_ecc_import_x963_ex2`, `wc_ecc_import_raw_private`, and `wc_ecc_import_private_key_ex`. All RSA and ECC callsites pass the actual key size in bytes via the `keySz` parameter. ECC import paths propagate custom curve parameters to the temporary key so custom curves (e.g. brainpool) work correctly through the callback.

**ExportKey:** When an export function (e.g. `wc_ecc_export_x963`) is called on a hardware-backed key, an empty temporary key is created and passed to the callback handler alongside the hardware source key. The handler extracts key material from hardware in whatever format it can and imports it into the temporary key using any wolfSSL import function. Once the callback returns, the original export function runs on the now fully-populated temporary key via the normal software path, producing the output the caller originally requested. The temporary key is freed afterward.

ExportKey callbacks are wired into `wc_RsaFlattenPublicKey`, `wc_RsaExportKey`, `wc_ecc_export_x963`, and `wc_ecc_export_ex` (which covers `wc_ecc_export_private_only`, `wc_ecc_export_public_raw`, and `wc_ecc_export_private_raw`).

Add `wc_CryptoCb_EccGetSize` and `wc_CryptoCb_EccGetSigSize` callback hooksto `wc_ecc_size()` and `wc_ecc_sig_size()`, allowing hardware-only keys (where `dp` may be NULL) to report their key and signature sizes through the crypto callback rather than requiring curve info to be present in the key struct.

Both features are enabled via `--enable-cryptocbutils=setkey,export`, `--enable-cryptocbutils`, or the corresponding `-DWOLF_CRYPTO_CB_SETKEY` and `-DWOLF_CRYPTO_CB_EXPORT_KEY` defines, with CI test configurations in `os-check.yml`. Includes test handlers in `test.c` and `api.c` with small-stack-safe allocations (`WC_DECLARE_VAR`), a `GetSetKeyTypeStr` debug helper for readable type names under `DEBUG_CRYPTOCB`, and RSA free callback support.

### DIFF
--- a/.github/workflows/os-check.yml
+++ b/.github/workflows/os-check.yml
@@ -68,6 +68,7 @@ jobs:
           '--enable-cryptocb --enable-keygen --enable-cryptocbutils=export',
           '--enable-cryptocb --enable-keygen CPPFLAGS="-DWOLF_CRYPTO_CB_EXPORT_KEY"',
           '--enable-cryptocb --enable-keygen --enable-aesgcm --enable-cryptocbutils=setkey,free,export CPPFLAGS="-DWOLF_CRYPTO_CB_AES_SETKEY"',
+          '--enable-cryptocb --enable-keygen --enable-cryptocbutils=setkey,export CPPFLAGS="-DWOLF_CRYPTO_CB_FIND"',
           'CPPFLAGS=-DWOLFSSL_NO_CLIENT_AUTH',
           'CPPFLAGS=''-DNO_WOLFSSL_CLIENT -DWOLFSSL_NO_CLIENT_AUTH''',
           'CPPFLAGS=''-DNO_WOLFSSL_SERVER -DWOLFSSL_NO_CLIENT_AUTH''',
--- a/wolfcrypt/src/cryptocb.c
+++ b/wolfcrypt/src/cryptocb.c
@@ -118,14 +118,16 @@ static const char* GetAlgoTypeStr(int algo)
 static const char* GetSetKeyTypeStr(int type)
 {
     switch (type) {
+        case WC_SETKEY_NONE:      return "None";
         case WC_SETKEY_HMAC:      return "HMAC";
         case WC_SETKEY_RSA_PUB:   return "RSA-Pub";
         case WC_SETKEY_RSA_PRIV:  return "RSA-Priv";
         case WC_SETKEY_ECC_PUB:   return "ECC-Pub";
         case WC_SETKEY_ECC_PRIV:  return "ECC-Priv";
         case WC_SETKEY_AES:       return "AES";
+        default:                  break;
     }
-    return "Unknown";
+    return NULL;
 }
 #endif /* WOLF_CRYPTO_CB_SETKEY */
 static const char* GetPkTypeStr(int pk)
@@ -2301,7 +2303,7 @@ int wc_CryptoCb_SetKey(int devId, int type, void* obj,
  * uses normal software export functions on 'out' and frees it.
  * Returns: 0 on success, CRYPTOCB_UNAVAILABLE if not handled, negative on error
  */
-int wc_CryptoCb_ExportKey(int devId, int type, void* obj, void* out)
+int wc_CryptoCb_ExportKey(int devId, int type, const void* obj, void* out)
 {
     int ret = WC_NO_ERR_TRACE(CRYPTOCB_UNAVAILABLE);
     CryptoCb* dev;
--- a/wolfcrypt/src/ecc.c
+++ b/wolfcrypt/src/ecc.c
@@ -9989,7 +9989,7 @@ int wc_ecc_export_x963(ecc_key* key, byte* out, word32* outLen)
         }
 
         ret = wc_CryptoCb_ExportKey(key->devId, WC_PK_TYPE_ECDSA_SIGN,
-                                     (void*)key, tmpKey);
+                                     key, tmpKey);
         if (ret == 0) {
             /* Call software helper (no callback recursion) */
             ret = _ecc_export_x963(tmpKey, out, outLen);
@@ -11345,7 +11345,7 @@ int wc_ecc_export_ex(ecc_key* key, byte* qx, word32* qxLen,
         }
 
         err = wc_CryptoCb_ExportKey(key->devId, WC_PK_TYPE_ECDSA_SIGN,
-                                     (void*)key, tmpKey);
+                                     key, tmpKey);
         if (err == 0) {
             /* Call software helper (no callback recursion) */
             err = _ecc_export_ex(tmpKey, qx, qxLen, qy, qyLen, d, dLen,
@@ -11431,7 +11431,7 @@ static int _ecc_import_private_key_ex(const byte* priv, word32 privSz,
     if (pub != NULL) {
     #ifndef NO_ASN
         word32 idx = 0;
-        ret = wc_ecc_import_x963_ex(pub, pubSz, key, curve_id);
+        ret = _ecc_import_x963_ex2(pub, pubSz, key, curve_id, 0);
         if (ret < 0)
             ret = wc_EccPublicKeyDecode(pub, &idx, key, pubSz);
         key->type = ECC_PRIVATEKEY;
@@ -12269,7 +12269,8 @@ int wc_ecc_size(ecc_key* key)
         return 0;
     }
 
-#ifdef WOLF_CRYPTO_CB
+#if defined(WOLF_CRYPTO_CB) && \
+    (defined(WOLF_CRYPTO_CB_SETKEY) || defined(WOLF_CRYPTO_CB_EXPORT_KEY))
     if (key->devId != INVALID_DEVID) {
         int ret;
         int keySz = 0;
@@ -12320,7 +12321,8 @@ int wc_ecc_sig_size(const ecc_key* key)
         return 0;
     }
 
-#ifdef WOLF_CRYPTO_CB
+#if defined(WOLF_CRYPTO_CB) && \
+    (defined(WOLF_CRYPTO_CB_SETKEY) || defined(WOLF_CRYPTO_CB_EXPORT_KEY))
     if (key->devId != INVALID_DEVID) {
         int ret;
         int cbKeySz = 0;
--- a/wolfcrypt/src/rsa.c
+++ b/wolfcrypt/src/rsa.c
@@ -4596,7 +4596,7 @@ int wc_RsaFlattenPublicKey(const RsaKey* key, byte* e, word32* eSz, byte* n,
         }
 
         ret = wc_CryptoCb_ExportKey(key->devId, WC_PK_TYPE_RSA,
-                                     (void*)key, tmpKey);
+                                     key, tmpKey);
         if (ret == 0) {
             /* Call software helper (no callback recursion) */
             ret = _RsaFlattenPublicKey(tmpKey, e, eSz, n, nSz);
@@ -4719,7 +4719,7 @@ int wc_RsaExportKey(const RsaKey* key,
             }
 
             ret = wc_CryptoCb_ExportKey(key->devId, WC_PK_TYPE_RSA,
-                                         (void*)key, tmpKey);
+                                         key, tmpKey);
             if (ret == 0) {
                 /* Call software helper (no callback recursion) */
                 ret = _RsaExportKey(tmpKey, e, eSz, n, nSz,
--- a/wolfssl/wolfcrypt/cryptocb.h
+++ b/wolfssl/wolfcrypt/cryptocb.h
@@ -526,7 +526,7 @@ typedef struct wc_CryptoInfo {
 #ifdef WOLF_CRYPTO_CB_EXPORT_KEY
     struct {                    /* uses wc_AlgoType=WC_ALGO_TYPE_EXPORT_KEY */
         int type;               /* enum wc_PkType (WC_PK_TYPE_RSA, etc.) */
-        void* obj;              /* Hardware key (has devCtx/id[]) */
+        const void* obj;        /* Hardware key (has devCtx/id[]) */
         void* out;              /* Software key to fill (same type as obj) */
     } export_key;
 #endif /* WOLF_CRYPTO_CB_EXPORT_KEY */
@@ -821,7 +821,7 @@ WOLFSSL_LOCAL int wc_CryptoCb_SetKey(int devId, int type, void* obj,
 #endif /* WOLF_CRYPTO_CB_SETKEY */
 #ifdef WOLF_CRYPTO_CB_EXPORT_KEY
 WOLFSSL_LOCAL int wc_CryptoCb_ExportKey(int devId, int type,
-                                         void* obj, void* out);
+                                         const void* obj, void* out);
 #endif /* WOLF_CRYPTO_CB_EXPORT_KEY */
 
 #endif /* WOLF_CRYPTO_CB */
--- a/wolfssl/wolfcrypt/types.h
+++ b/wolfssl/wolfcrypt/types.h
@@ -1570,6 +1570,8 @@ enum wc_PkType {
     WC_PK_TYPE_RSA_OAEP = 27,
     WC_PK_TYPE_EC_GET_SIZE = 28,
     WC_PK_TYPE_EC_GET_SIG_SIZE = 29,
+    #undef _WC_PK_TYPE_MAX
+    #define _WC_PK_TYPE_MAX WC_PK_TYPE_EC_GET_SIG_SIZE
     WC_PK_TYPE_MAX = _WC_PK_TYPE_MAX
 };
 
--- a/wolfcrypt/src/asn.c
+++ b/wolfcrypt/src/asn.c
@@ -8304,7 +8304,7 @@ int wc_RsaPrivateKeyDecode(const byte* input, word32* inOutIdx, RsaKey* key,
     WC_DECLARE_VAR(tmpKey, RsaKey, 1, NULL);
 #endif
 
-    if (key == NULL) {
+    if (key == NULL || input == NULL || inOutIdx == NULL) {
         return BAD_FUNC_ARG;
     }
 


night1rider:
Jenkins Retest this please

night1rider:
Rebased to fix merge conflict with master's `rsa.c`

night1rider:
Jenkins retest this please

wolfcrypt/src/ecc.c
```diff
@@ -9891,14 +9893,75 @@ int wc_ecc_export_x963(ecc_key* key, byte* out, word32* outLen)
...
+#endif
+
+    if (key == NULL || outLen == NULL) {
+        return ECC_BAD_ARG_E;
+    }
+
+    /* return length needed only */
+    if (out == NULL) {
+        word32 numlen = key->dp ? (word32)key->dp->size : MAX_ECC_BYTES;
+        *outLen = 1 + 2 * numlen;
```
bigbrett, line 9966: what is going on here? Is this just some hardcoded stuff having to do with X9.63? If so, please add a comment so its obvious to the reader what is going on.

wolfcrypt/src/ecc.c
```diff
@@ -11852,7 +12184,27 @@ static int ecc_public_key_size(ecc_key* key, word32* sz)
...
 int wc_ecc_size(ecc_key* key)
 {
-    if (key == NULL || key->dp == NULL)
+#ifdef WOLF_CRYPTO_CB
+    int ret;
+    int keySz;
+#endif
+
+    if (key == NULL)
+        return 0;
```
bigbrett: style nit: braceless ifs here and below. Just pointing it out since you fixed some in existing code in a different module. Prefer avoiding if possible.

wolfcrypt/src/ecc.c
```diff
@@ -11852,7 +12184,27 @@ static int ecc_public_key_size(ecc_key* key, word32* sz)
 WOLFSSL_ABI
 int wc_ecc_size(ecc_key* key)
 {
-    if (key == NULL || key->dp == NULL)
+#ifdef WOLF_CRYPTO_CB
+    int ret;
+    int keySz;
+#endif
```
bigbrett: nit: both `ret` and `keySz` can be scoped to the below block vs declared at the top, reducing number of ifdefs.

wolfcrypt/src/ecc.c
```diff
@@ -11882,8 +12234,27 @@ int wc_ecc_sig_size(const ecc_key* key)
 {
     int maxSigSz;
     int orderBits, keySz;
+#ifdef WOLF_CRYPTO_CB
+    int ret;
+    int cbKeySz;
+#endif
```
bigbrett: nit: both `ret` and `cbKeySz` can be scoped to the below block vs declared at the top, reducing number of ifdefs.

wolfssl/wolfcrypt/cryptocb.h
```diff
@@ -492,6 +512,24 @@ typedef struct wc_CryptoInfo {
...
+#ifdef WOLF_CRYPTO_CB_SETKEY
+    struct {                    /* uses wc_AlgoType=WC_ALGO_TYPE_SETKEY */
+        int type;               /* enum wc_SetKeyType */
+        void* obj;              /* Aes*, Hmac*, RsaKey*, ecc_key* */
+        void* key;              /* Raw key bytes or temp struct to export from */
+        word32 keySz;           /* Key size (0 when key is a struct ptr) */
+        void* aux;              /* Auxiliary data (IV, etc.) or NULL */
+        word32 auxSz;           /* Aux data size, 0 if unused */
+        int flags;              /* AES: direction (AES_ENCRYPTION/DECRYPTION) */
+    } setkey;
```
bigbrett, line 524: Clarifying question: Was the intent for the "id" field to be obtained from the wolfCrypt algorithm context pointed to in `obj`? At some point the crypto callback will need to know "which" key in HW you intend to set. This needs to be tracked somehow in the algo context regardless so imagine this was your intent?

wolfssl/wolfcrypt/cryptocb.h
```diff
@@ -492,6 +512,24 @@ typedef struct wc_CryptoInfo {
...
+#endif /* WOLF_CRYPTO_CB_SETKEY */
+#ifdef WOLF_CRYPTO_CB_EXPORT_KEY
+    struct {                    /* uses wc_AlgoType=WC_ALGO_TYPE_EXPORT_KEY */
+        int type;               /* enum wc_PkType (WC_PK_TYPE_RSA, etc.) */
+        void* obj;              /* Hardware key (has devCtx/id[]) */
+        void* out;              /* Software key to fill (same type as obj) */
+    } export_key;
```
bigbrett, line 531: same question as above re: key identifiers

wolfcrypt/src/rsa.c
```diff
@@ -4431,23 +4448,79 @@ int wc_RsaFlattenPublicKey(const RsaKey* key, byte* e, word32* eSz, byte* n,
...
     return 0;
 }
+
+/* flatten RsaKey structure into individual elements (e, n) */
+int wc_RsaFlattenPublicKey(const RsaKey* key, byte* e, word32* eSz, byte* n,
+                                                                   word32* nSz)
+{
+#if defined(WOLF_CRYPTO_CB) && defined(WOLF_CRYPTO_CB_EXPORT_KEY)
+    int ret;
+    WC_DECLARE_VAR(tmpKey, RsaKey, 1, NULL);
```
bigbrett: these could be scoped to the block below to reduce ifdef clutter

wolfcrypt/src/rsa.c
```diff
@@ -4508,6 +4591,64 @@ int wc_RsaExportKey(const RsaKey* key,
...
 }
+
+int wc_RsaExportKey(const RsaKey* key,
+                    byte* e, word32* eSz, byte* n, word32* nSz,
+                    byte* d, word32* dSz, byte* p, word32* pSz,
+                    byte* q, word32* qSz)
+{
+    int ret = WC_NO_ERR_TRACE(BAD_FUNC_ARG);
+#if defined(WOLF_CRYPTO_CB) && defined(WOLF_CRYPTO_CB_EXPORT_KEY)
+    WC_DECLARE_VAR(tmpKey, RsaKey, 1, NULL);
```
bigbrett: could be scoped to below block to reduce ifdef clutter

wolfcrypt/src/ecc.c
night1rider: This isn't new logic, it's the same length-query check from the original `wc_ecc_export_x963` that was moved to the internal `_ecc_export_x963` to avoid callback recursion. The wrapper duplicates it so callers can query the output size without entering the callback path. It shows as new code because of the function split. Added a comment to make it clearer.

wolfcrypt/src/ecc.c
night1rider: Fixed!

wolfcrypt/src/ecc.c
night1rider: Yes, I had to double check the style guide, but these can go into the other scope due to be used only there.

wolfcrypt/src/ecc.c
night1rider: addressed

wolfssl/wolfcrypt/cryptocb.h
night1rider: Yes, the intent is that the key slot identifier comes from the algorithm context in `obj`. The `_Id` and `_Label` init functions (`wc_AesInit_Id`, `wc_InitRsaKey_Id`, `wc_ecc_init_id`, etc.) store the id in the object's `id[]`/`idLen` fields. These were originally created for PKCS#11 support but are useful for any hardware backend. The `id[]` layout is platform-specific each callback implementation defines what the bytes mean for their hardware (catalog, group, slot, flags, etc.).

The callback casts `obj` to the appropriate type and reads the `id[]` field to determine which hardware key slot to target. Alternatively, a developer could also operate on `devCtx` directly beforehand to set up their own slot mapping, but if you want to use the already established API without needing any external custom APIs, the `_Id`/`_Label` init extensions are the way to go.

For example:
```c
Aes aes;
unsigned char myId[] = { /* platform-specific: catalog, group, slot, flags, ... */ };

ret = wc_AesInit_Id(&aes, myId, (int)sizeof(myId), NULL, MY_DEVID);
if (ret != 0) {
    return ret; 
}

ret = wc_AesSetKey(&aes, key, sizeof(key), iv, AES_ENCRYPTION);
if (ret != 0) {
    wc_AesFree(&aes);
    return ret;
}
```
`wc_AesSetKey` triggers the setkey callback, where the implementer casts `obj` to `Aes*` and reads `aes->id[]` to determine the target key slot.

This is how I am currently using it for the setkey path:
```c
/* SetKey CryptoCB handler — routes key imports to HSM hardware */
  int wc_CryptoCb_SetKey(int devId, wc_CryptoInfo* info)
  {
      (void)devId;

      if (info == NULL) {
          return BAD_FUNC_ARG;
      }

      switch (info->setkey.type) {
  #if !defined(NO_AES)
          case WC_SETKEY_AES:
          {
              int ret;
              Aes* aes = (Aes*)info->setkey.obj;
              ret = hsm_AesSetKey(aes,
                  (const byte*)info->setkey.key,
                  info->setkey.keySz,
                  (const byte*)info->setkey.aux);
              return ret;
          }
  #endif
  #if !defined(NO_RSA)
          case WC_SETKEY_RSA_PRIV:
          {
              int ret;
              RsaKey* rsaObj = (RsaKey*)info->setkey.obj;
              RsaKey* rsaTmp = (RsaKey*)info->setkey.key;
              ret = hsm_RsaSetKey_Priv(rsaObj, rsaTmp,
                  info->setkey.keySz);
              return ret;
          }
          case WC_SETKEY_RSA_PUB:
          {
              int ret;
              RsaKey* rsaObj = (RsaKey*)info->setkey.obj;
              RsaKey* rsaTmp = (RsaKey*)info->setkey.key;
              ret = hsm_RsaSetKey_Pub(rsaObj, rsaTmp,
                  info->setkey.keySz);
              return ret;
          }
  #endif
  #if defined(HAVE_ECC)
          case WC_SETKEY_ECC_PRIV:
          {
              int ret;
              ecc_key* eccObj = (ecc_key*)info->setkey.obj;
              ecc_key* eccTmp = (ecc_key*)info->setkey.key;
              ret = hsm_EccSetKey_Priv(eccObj, eccTmp,
                  info->setkey.keySz);
              return ret;
          }
          case WC_SETKEY_ECC_PUB:
          {
              int ret;
              ecc_key* eccObj = (ecc_key*)info->setkey.obj;
              ecc_key* eccTmp = (ecc_key*)info->setkey.key;
              ret = hsm_EccSetKey_Pub(eccObj, eccTmp,
                  info->setkey.keySz);
              return ret;
          }
  #endif
          default:
              return WC_NO_ERR_TRACE(CRYPTOCB_UNAVAILABLE);
      }
  }
```

wolfssl/wolfcrypt/cryptocb.h
night1rider: addressed earlier, similar usage for export.

wolfcrypt/src/rsa.c
night1rider: updated

wolfcrypt/src/rsa.c
night1rider: updated

night1rider:
rebased

bigbrett (dismissed):

night1rider:
Jenkins retest this please

night1rider:
rebased and force pushed

night1rider:
Jenkins retest this please

wolfcrypt/src/ecc.c
```diff
@@ -11912,8 +12246,28 @@ static int ecc_public_key_size(ecc_key* key, word32* sz)
...
+    if (key == NULL) {
         return 0;
+    }
+
+#ifdef WOLF_CRYPTO_CB
+    if (key->devId != INVALID_DEVID) {
+        int ret;
+        int keySz = 0;
+
+        ret = wc_CryptoCb_EccGetSize(key, &keySz);
```
dgarske, line 12278: ### HIGH-1: `wc_ecc_size()` and `wc_ecc_sig_size()` callback guarded too broadly — affects all WOLF_CRYPTO_CB builds
- **File:** `wolfcrypt/src/ecc.c:12247-12273` and `wolfcrypt/src/ecc.c:12295-12340`
- **Function:** `wc_ecc_size`, `wc_ecc_sig_size`
- **Category:** [api]
- **Confidence:** High

**Description:** The crypto callback lookup in `wc_ecc_size()` and `wc_ecc_sig_size()` is guarded only by `#ifdef WOLF_CRYPTO_CB`, not by the more specific `WOLF_CRYPTO_CB_SETKEY` or `WOLF_CRYPTO_CB_EXPORT_KEY`. This means ALL builds with crypto callbacks enabled (not just those opting into the new SETKEY/EXPORT features) will incur a `wc_CryptoCb_FindDevice()` lookup for every `wc_ecc_size()` and `wc_ecc_sig_size()` call when `key->devId != INVALID_DEVID`.

These are extremely hot functions called throughout ECC operations (key gen, sign, verify, import, export, certificate handling). The overhead adds up. More importantly, existing third-party callbacks that don't handle `WC_PK_TYPE_EC_GET_SIZE` / `WC_PK_TYPE_EC_GET_SIG_SIZE` may not return `CRYPTOCB_UNAVAILABLE` gracefully — they could return an unrecognized-type error, causing `wc_ecc_size()` to return 0 and breaking downstream operations silently.

**Code:**
```c
// wc_ecc_size — guarded only by WOLF_CRYPTO_CB
#ifdef WOLF_CRYPTO_CB
    if (key->devId != INVALID_DEVID) {
        int ret;
        int keySz = 0;
        ret = wc_CryptoCb_EccGetSize(key, &keySz);
            if (ret != 0) {
                return 0;  // silently returns 0 on error
            }
            return keySz;
        }
    }
#endif
```

**Recommendation:** Either guard these with a more specific define like `WOLF_CRYPTO_CB_SETKEY` or `WOLF_CRYPTO_CB_EXPORT_KEY`, or add compile-time guard documentation that the new `WC_PK_TYPE_EC_GET_SIZE`/`WC_PK_TYPE_EC_GET_SIG_SIZE` callbacks must be handled (or return CRYPTOCB_UNAVAILABLE) by all crypto callback implementations when WOLF_CRYPTO_CB is enabled. The `CRYPTO_CB_VER` bump from 2→3 helps but doesn't prevent runtime breakage.

wolfssl/wolfcrypt/types.h
```diff
@@ -1566,6 +1568,8 @@ enum wc_PkType {
     WC_PK_TYPE_RSA_PKCS = 25,
     WC_PK_TYPE_RSA_PSS = 26,
     WC_PK_TYPE_RSA_OAEP = 27,
+    WC_PK_TYPE_EC_GET_SIZE = 28,
```
dgarske, line 1571: ### MEDIUM-2: `WC_PK_TYPE_EC_GET_SIZE`/`WC_PK_TYPE_EC_GET_SIG_SIZE` exceed `WC_PK_TYPE_MAX`
- **File:** `wolfssl/wolfcrypt/types.h:1571-1573`
- **Function:** N/A (enum definition)
- **Category:** [api]
- **Confidence:** High

**Description:** The new enum values `WC_PK_TYPE_EC_GET_SIZE = 28` and `WC_PK_TYPE_EC_GET_SIG_SIZE = 29` are placed after `WC_PK_TYPE_RSA_OAEP = 27`, but `WC_PK_TYPE_MAX = _WC_PK_TYPE_MAX` resolves to at most 24 (with Dilithium/Falcon). This means these new values exceed `WC_PK_TYPE_MAX`, making the MAX misleading. While no code currently uses `WC_PK_TYPE_MAX` for array sizing (I verified), this continues and worsens a pattern that started with values 25-27.

**Code:**
```c
    WC_PK_TYPE_RSA_OAEP = 27,
    WC_PK_TYPE_EC_GET_SIZE = 28,      // New: above _WC_PK_TYPE_MAX
    WC_PK_TYPE_EC_GET_SIG_SIZE = 29,  // New: above _WC_PK_TYPE_MAX
    WC_PK_TYPE_MAX = _WC_PK_TYPE_MAX  // Could be 17, 20, or 24
```

**Recommendation:** The unconditional values 25-29 should update `_WC_PK_TYPE_MAX` after the conditional blocks, e.g.:
```suggestion
    WC_PK_TYPE_EC_GET_SIG_SIZE = 29,
    #undef _WC_PK_TYPE_MAX
    #define _WC_PK_TYPE_MAX WC_PK_TYPE_EC_GET_SIG_SIZE
    WC_PK_TYPE_MAX = _WC_PK_TYPE_MAX
```

wolfcrypt/src/rsa.c
```diff
@@ -4463,23 +4480,77 @@ int wc_RsaFlattenPublicKey(const RsaKey* key, byte* e, word32* eSz, byte* n,
...
+        }
+        XMEMSET(tmpKey, 0, sizeof(RsaKey));
+
+        ret = wc_InitRsaKey_ex(tmpKey, key->heap, INVALID_DEVID);
+        if (ret != 0) {
+            WC_FREE_VAR(tmpKey, key->heap);
+            return ret;
+        }
+
+        ret = wc_CryptoCb_ExportKey(key->devId, WC_PK_TYPE_RSA,
```
dgarske, line 4598: ### MEDIUM-3: Const-qualifying `key` parameter cast away in ExportKey calls
- **File:** `wolfcrypt/src/rsa.c:4533-4534` and `wolfcrypt/src/rsa.c:4656-4657`
- **Function:** `wc_RsaFlattenPublicKey`, `wc_RsaExportKey`
- **Category:** [convention]
- **Confidence:** High

**Description:** Both `wc_RsaFlattenPublicKey` and `wc_RsaExportKey` accept `const RsaKey* key`, but the ExportKey callback invocation casts the const away via `(void*)key`. The callback implementation then treats the key as mutable (`RsaKey* src = (RsaKey*)info->export_key.obj`), calling `wc_RsaKeyToDer(src, ...)` which takes non-const `RsaKey*`. If a callback implementation modifies the source key, this is undefined behavior.

**Code:**
```c
// In wc_RsaFlattenPublicKey (key is const RsaKey*):
ret = wc_CryptoCb_ExportKey(key->devId, WC_PK_TYPE_RSA,
                             (void*)key, tmpKey);  // cast-away-const
```

**Recommendation:** Either change `wc_CryptoCb_ExportKey`'s `obj` parameter to `const void*`, or document that the ExportKey callback MUST NOT modify the source object. The `wc_CryptoInfo` struct member `export_key.obj` should be `const void*`.

tests/api.c
```diff
@@ -28293,6 +28322,304 @@ static int test_CryptoCb_Func(int thisDevId, wc_CryptoInfo* info, void* ctx)
         }
     }
 #endif /* WOLF_CRYPTO_CB_FREE */
+#ifdef WOLF_CRYPTO_CB_SETKEY
```
dgarske, line 28338: ### MEDIUM-4: Significant code duplication between test callback implementations
- **File:** `tests/api.c:28322-28625` and `wolfcrypt/test/test.c:65838-66177`
- **Function:** `test_CryptoCb_Func`, `myCryptoDevCb`
- **Category:** [style]
- **Confidence:** High

**Description:** The SETKEY and EXPORT_KEY callback implementations in `tests/api.c` and `wolfcrypt/test/test.c` are nearly identical (~300 lines each). Every key type handler (AES, HMAC, RSA pub/priv, ECC pub/priv) is copy-pasted between the two test files. This makes maintenance error-prone — any bug fix in one copy must be manually replicated in the other.

**Recommendation:** Consider extracting common callback helper functions into a shared test utility header, or use a common `#include` file for the shared handler implementations.

wolfcrypt/src/ecc.c
```diff
@@ -11203,17 +11393,20 @@ int wc_ecc_export_private_raw(ecc_key* key, byte* qx, word32* qxLen,
...
+                                      const byte* pub, word32 pubSz,
+                                      ecc_key* key, int curve_id)
 {
     int ret;
 #ifdef WOLFSSL_CRYPTOCELL
     const CRYS_ECPKI_Domain_t* pDomain;
 #endif
-    if (key == NULL || priv == NULL)
+
+    if (key == NULL || priv == NULL) {
```
dgarske, line 11426: ### MEDIUM-5: `_ecc_import_private_key_ex` calls public `wc_ecc_import_x963_ex` — potential infinite recursion if devId not cleared
- **File:** `wolfcrypt/src/ecc.c:11415`
- **Function:** `_ecc_import_private_key_ex`
- **Category:** [bug]
- **Confidence:** Medium

**Description:** The software-only helper `_ecc_import_private_key_ex` calls `wc_ecc_import_x963_ex(pub, pubSz, key, curve_id)`, which eventually calls `wc_ecc_import_x963_ex2`. The x963 import wrapper has SETKEY callback logic that checks `key->devId != INVALID_DEVID`. Since `_ecc_import_private_key_ex` is always called with `tmpKey` that has `INVALID_DEVID`, this works correctly — the x963 import skips the callback path.

However, this relies on an implicit contract that `_ecc_import_private_key_ex` is ONLY called with `INVALID_DEVID` keys. If a future caller passes a key with a valid devId, this would trigger a recursion loop: `wc_ecc_import_private_key_ex → _ecc_import_private_key_ex → wc_ecc_import_x963_ex → wc_ecc_import_x963_ex2 (SETKEY path) → callback → wc_ecc_import_private_key_ex → ...`. The naming convention (`_` prefix) suggests internal use, but there's no compile-time enforcement.

**Recommendation:** Add an assertion or comment at the top of `_ecc_import_private_key_ex` verifying `key->devId == INVALID_DEVID`, or call `_ecc_import_x963_ex2` directly instead of going through the public wrapper.

wolfcrypt/src/aes.c
```diff
@@ -4518,6 +4518,9 @@ WC_ALL_ARGS_NOT_NULL static WARN_UNUSED_RESULT int wc_AesDecrypt(
     int wc_AesSetKey(Aes* aes, const byte* userKey, word32 keylen,
             const byte* iv, int dir)
     {
+#if defined(WOLF_CRYPTO_CB) && defined(WOLF_CRYPTO_CB_SETKEY)
```
dgarske, line 4521: ### LOW-6: Unused `cbRet` variable when WOLF_CRYPTO_CB_SETKEY not combined with WOLF_CRYPTO_CB
- **File:** `wolfcrypt/src/aes.c:4521-4523`
- **Function:** `wc_AesSetKey`
- **Category:** [style]
- **Confidence:** Medium

**Description:** The variable `cbRet` is declared under `#if defined(WOLF_CRYPTO_CB) && defined(WOLF_CRYPTO_CB_SETKEY)` but its usage is guarded by `#ifdef WOLF_CRYPTO_CB_SETKEY` alone (inside an `#ifdef WOLF_CRYPTO_CB` block). While the outer guards make the combination safe in practice, the inconsistent guard style between declaration and use could cause compiler warnings if the guards ever diverge.

**Code:**
```c
#if defined(WOLF_CRYPTO_CB) && defined(WOLF_CRYPTO_CB_SETKEY)
    int cbRet;  // Declaration
#endif
    ...
#ifdef WOLF_CRYPTO_CB
    ...
    #ifdef WOLF_CRYPTO_CB_SETKEY     // Usage
        cbRet = wc_CryptoCb_SetKey(...);
```

**Recommendation:** NIT — the code is correct as-is since `WOLF_CRYPTO_CB_SETKEY` is always inside `#ifdef WOLF_CRYPTO_CB`. No action required.

wolfcrypt/src/cryptocb.c
```diff
@@ -105,9 +105,29 @@ static const char* GetAlgoTypeStr(int algo)
...
+#ifdef WOLF_CRYPTO_CB_SETKEY
+        case WC_ALGO_TYPE_SETKEY: return "SetKey";
+#endif /* WOLF_CRYPTO_CB_SETKEY */
+#ifdef WOLF_CRYPTO_CB_EXPORT_KEY
+        case WC_ALGO_TYPE_EXPORT_KEY: return "ExportKey";
+#endif /* WOLF_CRYPTO_CB_EXPORT_KEY */
     }
     return NULL;
 }
+#ifdef WOLF_CRYPTO_CB_SETKEY
```
dgarske, line 117: ### LOW-7: `GetSetKeyTypeStr` missing default `WC_SETKEY_NONE` case
- **File:** `wolfcrypt/src/cryptocb.c:116-128`
- **Function:** `GetSetKeyTypeStr`
- **Category:** [style]
- **Confidence:** Low

**Description:** The debug helper `GetSetKeyTypeStr` handles all defined `WC_SETKEY_*` values but returns `"Unknown"` for `WC_SETKEY_NONE = 0`. While not a bug, other similar helpers in the file (like `GetPkTypeStr`) return NULL for unrecognized values. This inconsistency is minor but could confuse debuggers.

**Recommendation:** Consider adding `case WC_SETKEY_NONE: return "None";` for completeness, and returning NULL for the default case to match the pattern used by `GetPkTypeStr`.

.github/workflows/os-check.yml
```diff
@@ -84,6 +84,12 @@ jobs:
           '--disable-tls --enable-cryptocb --enable-aesgcm CPPFLAGS="-DWOLF_CRYPTO_CB_AES_SETKEY -DWOLF_CRYPTO_CB_FREE"',
           '--enable-all --enable-dilithium --enable-cryptocb --enable-cryptocbutils --enable-pkcallbacks',
           '--enable-cryptocb --enable-aesgcm CPPFLAGS="-DWOLF_CRYPTO_CB_AES_SETKEY"',
+          '--enable-cryptocb --enable-keygen --enable-cryptocbutils=setkey',
```
dgarske, line 65: ### LOW-8: New CI test configurations don't cover WOLF_CRYPTO_CB_FIND interaction
- **File:** `.github/workflows/os-check.yml:87-92`
- **Function:** N/A (CI)
- **Category:** [test]
- **Confidence:** Low

**Description:** Six new CI configurations test various `--enable-cryptocbutils` combinations but none test the interaction with `WOLF_CRYPTO_CB_FIND`. Several code paths have `#ifndef WOLF_CRYPTO_CB_FIND` guards that change behavior (e.g., skipping the `devId != INVALID_DEVID` check). Testing this combination would increase coverage of the new feature.

**Recommendation:** Consider adding at least one CI config that combines `--enable-cryptocbutils` with `WOLF_CRYPTO_CB_FIND`.

wolfcrypt/src/rsa.c
night1rider: Changed `export_key.obj` to `const void*` in the struct definition (cryptocb.h),
updated the `wc_CryptoCb_ExportKey` declaration and definition to take `const void* obj`,
and removed the `(void*)` cast-away-const at all 4 callsites in rsa.c and ecc.c.
Test callbacks left unchanged - they cast away const from `export_key.obj` since the
wolfSSL export APIs take non-const key parameters.

wolfssl/wolfcrypt/types.h
night1rider: Added `#undef _WC_PK_TYPE_MAX` / `#define _WC_PK_TYPE_MAX WC_PK_TYPE_EC_GET_SIG_SIZE`
before `WC_PK_TYPE_MAX = _WC_PK_TYPE_MAX` so MAX correctly resolves to 29.

wolfcrypt/src/ecc.c
night1rider: Changed both guards from `#ifdef WOLF_CRYPTO_CB` to
`#if defined(WOLF_CRYPTO_CB) && (defined(WOLF_CRYPTO_CB_SETKEY) || defined(WOLF_CRYPTO_CB_EXPORT_KEY))`.
These callbacks are only relevant when SETKEY or EXPORT_KEY features are enabled.

tests/api.c
night1rider: There is currently no sharing code between `tests/api.c` and `wolfcrypt/test/test.c`  they have separate include structures and their existing crypto callback implementations (`test_CryptoCb_Func` vs `myCryptoDevCb`) take fundamentally different approaches even for the pre-existing operations (e.g., api.c loads RSA keys from PEM files while test.c uses pre-initialized keys directly).

The SETKEY/EXPORT_KEY handlers are the first significant duplication between them. Introducing a shared test utility header would be a new pattern for the repo.

I can look at consolidating this in a follow-up if desired.

wolfcrypt/src/ecc.c
night1rider: Changed `_ecc_import_private_key_ex` to call `_ecc_import_x963_ex2(pub, pubSz, key, curve_id, 0)` directly instead of going through the public `wc_ecc_import_x963_ex` wrapper. This eliminates the recursion risk entirely - the internal function bypasses the SETKEY callback block.

wolfcrypt/src/cryptocb.c
night1rider: Added `case WC_SETKEY_NONE: return "None";` and changed default to return `NULL`, matching the pattern used by `GetPkTypeStr`.

.github/workflows/os-check.yml
night1rider: `'--enable-cryptocb --enable-keygen --enable-cryptocbutils=setkey,export CPPFLAGS="-DWOLF_CRYPTO_CB_FIND"'`. Also fixed a pre-existing NULL pointer dereference in `wc_RsaPrivateKeyDecode` (asn.c) where `*inOutIdx` was accessed in the SETKEY block before validating `inOutIdx != NULL`. With `WOLF_CRYPTO_CB_FIND` the SETKEY block runs unconditionally, so the bad-args test case (which passes NULL) crashed. Added early NULL check for `input` and `inOutIdx` parameters.

night1rider:
Had to rebase and force push due to merge conflict with `.github/workflows/os-check.yml`

night1rider:
rebased and forced to pull in https://github.com/wolfSSL/wolfssl/pull/10202 to fix CI issues

PR #9859, opened 2026-03-04
Fix for loop exit condition.

anhu:
size should be length.  
s includes offset, so it must be compared against length, not size.
This is because size is only what is after offset.

Fixes ZD 21271

### DIFF
--- a/src/tls.c
+++ b/src/tls.c
@@ -1910,7 +1910,7 @@ static int TLSX_ALPN_ParseAndSet(WOLFSSL *ssl, const byte *input, word16 length,
         return BUFFER_ERROR;
 
     /* validating length of entries before accepting */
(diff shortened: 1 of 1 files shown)


anhu:
I've never seen this kind of build failure before: 
```
make[2]: *** [tests/unit_test-suites.o] Abort trap: 6
make[2]: *** Deleting file `tests/unit_test-suites.o'
```

anhu:
Jenkin retest this please

Build 'wolfSSL/PRB-fips-repo-and-harness-test-v3-part2' failed with result: FAILURE

anhu:
Followed wrong procedure for opening a PR.  Will re-open with branch from my own repo.

anhu:
This has become https://github.com/wolfSSL/wolfssl/pull/9860

PR #9875, opened 2026-03-05
Treat alerts as fatal errors regardless of level in TLS1.3

Frauschi:
All alerts except for `close_notify` and `user_canceled` are now treated as `FATAL_ERROR` regardless of the level within the alert message for TLS 1.3 (in compliance with RFC 8446 §6).

Added test case to verify.

F-158

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -22036,6 +22036,12 @@ static int DoAlert(WOLFSSL* ssl, byte* input, word32* inOutIdx, int* type)
         if (level == alert_fatal) {
             ssl->options.isClosed = 1;  /* Don't send close_notify */
         }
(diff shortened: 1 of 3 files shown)


douzzer:
retest this please
(numerous FIPS scenarios aborted without logs)

douzzer (approved):

douzzer:
merging without waiting for libssh2 workflow or FIPS checks to complete -- these changes only affect the TLS layer, which libssh2 doesn't use, and the FIPS tests didn't run in Jenkins and all passed under multi-test, and again this PR only changes the TLS layer at the protocol level, with no crypto impact.

PR #9882, opened 2026-03-05
Send alert in case of decrypted all-zero message

Frauschi:
F-160


JacobBarthelmeh:
Retest this please Jenkins

@Frauschi it's possible the libssh2 test was flakey, but please investigate the failure.

douzzer:
merging without waiting for libssh2 workflow to finish (no runner available) -- change only affects TLS layer, and libssh2 only uses the crypto layer.

PR #9887, opened 2026-03-05
20260305 Coverity fixes

rlm2002:
# Description

351993 Unused value - removed assignment `tempBuf = NULL` as value is overwritten by loop and is freed before breaking from loop. `tempBuf` is not accessed after free.

315860 Out-of-bounds write - include `RIPEMD_BLOCK_SIZE` when checking value of `ripemd->buffLen` so local[ripemd->buffLen++] = 0x80;` will not attempt to write outside of bounds.

# Testing

How did you test?

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


rlm2002:
retest this please

douzzer (approved):

PR #9931, opened 2026-03-09
Fix Fenrir 376

aidangarske:
# Description

Fix copy-paste error in EncodeCertReq guard check where falconKey was checked twice instead of including dilithiumKey and sphincsKey


JacobBarthelmeh:
Retest this please Jenkins,

External test on makedistcheck failed

PR #9933, opened 2026-03-09
20260309 Coverity changes for Sunday build

rlm2002:
# Description

Resource leak - 330394: free enc and dec before returning

Resource leak - 330390: Return using ERROR_OUT instead of return ret

Use after free - 330389/330393: move `(void)` calls before freeing variables.

Wrong sizeof argument (SIZEOF_MISMATCH) - 330388: use `MAX_ECIES_TEST_SZ` instead of sizeof in `XSTRLCPY()`

# Testing

`./configure --enable-all --enable-smallstack && make check`


rlm2002:
retest this please Jenkins

douzzer (approved):

PR #9954, opened 2026-03-11
Fix potential overflows in used size calculation in generic, TI and SE050 hash functions.

### LINKED ISSUE #9951: [Bug]: Integer overflow in SE050 hash update could mess up the heap
### Contact Details

arjunaajalahla100@gmail.com (pelioro)

### Version

current branch

### Description

Yo, found a sketchy integer overflow in the SE050 hash code. Check this out at `wolfcrypt/src/port/nxp/se050_port.c` around line 275:

```c
if (se050Ctx->len < se050Ctx->used + len) {  // overflow could happen here
    tmp = XMALLOC(se050Ctx->used + len, ...);  // allocates less than expected
}
XMEMCPY(se050Ctx->msg + se050Ctx->used, data, len);  // writes too much
```

So basically if `se050Ctx->used` is like 4GB-ish (0xFFFFFFF0) and `len` is small, the addition wraps around and we allocate tiny buffer but copy full data = heap overflow city.

## Suggested fix

Just slap an overflow check before the addition:

```c
// Before line 275
if (se050Ctx->used > 0xFFFFFFFFU - len) {
    return BUFFER_E;  // overflow would happen, bail out
}
```

Or use the existing `WC_SAFE_SUM_WORD32()` macro if available.

## Bottom line

Not a critical vuln but should be patched to keep the code robust. Better safe than sorry right?

## How to trigger

Build with SE050 support:
```bash
./configure --enable-se050
# or in user_settings.h:
#define WOLFSSL_SE050_HASH
```

Need SE050 hardware connected via I2C.

## Repro steps

Theoretical repro (not really exploitable irl due to hardware limits):

1. Init hash with `se050_hash_init()`
2. Keep calling `se050_hash_update()` until `used` nears 4GB
3. Pass `len` that causes `used + len` to overflow
4. BOOM - small alloc, big write = heap corruption

- **Impact:** Low-ish - SE050 max buffer is like 200KB, so hitting 4GB is basically impossible
- **Risk:** Still worth fixing for defense-in-depth
- **Exploitable:** Nah, not really, but it's bad practice to not check for overflow

### Relevant log output

```shell

```

kareem-wolfssl:
# Description

Fixes #9951.

Thanks to Arjuna Arya for the report.

# Testing

Built in tests

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/wolfcrypt/src/port/ti/ti-hash.c
+++ b/wolfcrypt/src/port/ti/ti-hash.c
@@ -87,7 +87,8 @@ static int hashUpdate(wolfssl_TI_Hash *hash, const byte* data, word32 len)
         } else {
             p = XREALLOC(hash->msg, usedSz, NULL, DYNAMIC_TYPE_TMP_BUFFER);
         }
-        if (p == 0)return 1;
+        if (p == 0)
+            return MEMORY_E;
         hash->msg = p;
         hash->len = usedSz;
     }


kareem-wolfssl:
Retest this please.

dgarske (approved):

PR #9957, opened 2026-03-12
Fixed failing cert verification with nonblocking-OCSP and low MFL

MatanRad:
# Description

Fixes a bug that happens when using nonblocking OCSP and having a low MFL configured.
This is caused by the Certificate handshake message being fragmented and the `OCSP_WANT_READ` error being ignored, causing the library to skip over the current `pendingMsg`.

# Testing

Ran locally with a nonblocking OCSP setup and a configured MFL of 1024


kareem-wolfssl:
Hi @MatanRad,

Thanks for the bug report and PR.  As we do not have a contributor agreement on file for you, we will be treating this PR as a bug report.

PR #9964, opened 2026-03-12
SP fixes: 32-bit ARM assembly fixes

SparkiDev:
# Description

mod_exp: subtract from 32 instread of 64 as n is 32 bits 
sp_521_ecc_mulmod_fast: look up the last point in constant time when required.

# Testing

Tested SP for ARM32/Thumb2/Thumb


SparkiDev:
retest this please

PR #9976, opened 2026-03-16
autoconf: refresh bundled macros, remove AC_CANONICAL_TARGET

BrianAker:
# Description

This updates bundled autoconf-archive macros, remove AC_CANONICAL_TARGET from configure.ac which had been required because of bug in the version ax_pthread.m4 used. This keeps pthread/configure checks aligned with native host builds and pulls in newer macro behavior for compiler flag probing and compiler version detection.

To be clear, with the exception of the one line change in configure.ac, all other changes are the updated m4 from autoconf-archive ( brew gives the version as 2024.10.16 ). The autoconf-archive went through a license change sometime ago to the currently used in these files.  These m4 were included with wolfssl originally to simplify compiling, which is common for how m4 files are distributed.

# Testing

./autogen.sh && ./configure && make && make check && make dist && make distcheck


LinuxJedi:
Jenkins retest this please: multi-test bailed.

LinuxJedi:
Jenkins, retest this please. Results missing

PR #9980, opened 2026-03-16
Fixes SPHINCS else-if chain key detection

anhu:
F-751


anhu:
Jenkins retest this please.

PR #9984, opened 2026-03-16
Fix integer overflow in _wc_Hash_Grow and TI hashUpdate

### LINKED ISSUE #9955: [Bug]: Integer overflow in hash update could mess up the heap (hash.c & ti-hash.c)
### Contact Details

arjunaajalahla100@gmail.com (pelioro)

### Version

main branch

### Description

Found the same sketchy integer overflow pattern in two more files. Check these out:

**File 1:** `wolfcrypt/src/hash.c` around line 1957
```c
if (*len < *used + inSz) {  // [OVERFLOW]
    *msg = XMALLOC(*used + inSz, heap, ...);  // [SMALL ALLOC]
}
XMEMCPY(*msg + *used, in, inSz);  // [HEAP OVERFLOW]
```

**File 2:** `wolfcrypt/src/port/ti/ti-hash.c` around line 81
```c
if (hash->len < hash->used+len) {  // [OVERFLOW]
    p = XMALLOC(hash->used+len, NULL, ...);  // [SMALL ALLOC]
}
XMEMCPY(hash->msg+hash->used, data, len);  // [HEAP OVERFLOW]
```

Same deal as the SE050 issue (#9951) - if `used` is near 4GB and `len` causes addition to wrap around, we get tiny allocation but full write = heap overflow.

## How to trigger

Build with hash keep feature:
```bash
./configure --enable-hashkeep
# or in user_settings.h:
#define WOLFSSL_HASH_KEEP
```

For TI hash, need TI hardware with:
```c
#define WOLFSSL_TI_HASH
```

Same bug pattern as #9951, just in different files. Low severity but should be patched to keep the code consistent and robust. Better safe than sorry right?

## Repro steps

Theoretical repro (not really exploitable irl due to memory limits):

1. Init hash context
2. Keep calling hash update until `used` nears 4GB  
3. Pass `inSz` that causes `used + inSz` to overflow
4. Crash - small alloc, big write = heap corruption

## Real talk on severity

- **Impact:** Low - Need ~4GB hash input which will OOM on most systems before hitting overflow
- **Risk:** Still worth fixing for defense-in-depth (same pattern got fixed in SE050 #9954)
- **Exploitable:** Nah, not really, but it's the same bad pattern that was already patched once

## Suggested fix

Just add overflow check before the addition (same fix as SE050 #9954):

```c
// For hash.c:1957
if (!WC_SAFE_SUM_WORD32(*used, inSz, tmpSz)) {
    return BUFFER_E;  // overflow would happen, bail out
}

// For ti-hash.c:81  
if (!WC_SAFE_SUM_WORD32(hash->used, len, tmpSz)) {
    return BAD_FUNC_ARG;  // overflow would happen, bail out
}
```

### Relevant log output

```shell

```

Scottcjn:
## Summary

Fix integer overflow in hash buffer size computation that could lead to heap buffer overflow.

## Problem

Both `_wc_Hash_Grow()` (hash.c) and `hashUpdate()` (ti-hash.c) compute the new buffer size as `used + inSz` without overflow checking. If `used` is near `UINT32_MAX`, the addition wraps to a small value, causing:

1. Small allocation (`XMALLOC` with wrapped size)
2. Large write (`XMEMCPY` with original `inSz`) → **heap overflow**

Same pattern as the SE050 fix in #9954.

## Fix

Use `WC_SAFE_SUM_WORD32()` to detect overflow before the addition:

```c
word32 newSz;
if (!WC_SAFE_SUM_WORD32(*used, (word32)inSz, newSz)) {
    return BUFFER_E;  // overflow detected
}
```

Then use `newSz` for all allocation and length comparisons.

## Files Changed

- `wolfcrypt/src/hash.c` — `_wc_Hash_Grow()` (requires `WOLFSSL_HASH_KEEP`)
- `wolfcrypt/src/port/ti/ti-hash.c` — `hashUpdate()` (requires `WOLFSSL_TI_HASH`)

## Severity

Low — requires ~4GB of hash input to trigger, which will OOM on most systems first. But defense-in-depth is the right call, and this keeps the codebase consistent with the SE050 fix.

Fixes #9955

CLA: Previously signed (on file from PR #9932).

### DIFF
--- a/wolfcrypt/src/hash.c
+++ b/wolfcrypt/src/hash.c
@@ -1954,12 +1954,17 @@ int wc_HashGetFlags(wc_HashAlg* hash, enum wc_HashType type, word32* flags)
 int _wc_Hash_Grow(byte** msg, word32* used, word32* len, const byte* in,
                         int inSz, void* heap)
 {
-    if (*len < *used + inSz) {
+    word32 newSz;
+
+    if (!WC_SAFE_SUM_WORD32(*used, (word32)inSz, newSz)) {
+        return BUFFER_E;
+    }
+    if (*len < newSz) {
         if (*msg == NULL) {
-            *msg = (byte*)XMALLOC(*used + inSz, heap, DYNAMIC_TYPE_TMP_BUFFER);
+            *msg = (byte*)XMALLOC(newSz, heap, DYNAMIC_TYPE_TMP_BUFFER);
         }
         else {
-            byte* pt = (byte*)XREALLOC(*msg, *used + inSz, heap,
+            byte* pt = (byte*)XREALLOC(*msg, newSz, heap,
                     DYNAMIC_TYPE_TMP_BUFFER);
             if (pt == NULL) {
                 return MEMORY_E;
@@ -1969,7 +1974,7 @@ int _wc_Hash_Grow(byte** msg, word32* used, word32* len, const byte* in,
         if (*msg == NULL) {
             return MEMORY_E;
         }
-        *len = *used + inSz;
+        *len = newSz;
     }
     XMEMCPY(*msg + *used, in, inSz);
     *used += inSz;
--- a/wolfcrypt/src/port/ti/ti-hash.c
+++ b/wolfcrypt/src/port/ti/ti-hash.c
@@ -75,18 +75,22 @@ static int hashInit(wolfssl_TI_Hash *hash)
 static int hashUpdate(wolfssl_TI_Hash *hash, const byte* data, word32 len)
 {
     void *p;
+    word32 newSz;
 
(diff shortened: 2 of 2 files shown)


dgarske:
Fixed in https://github.com/wolfSSL/wolfssl/pull/9954/changes#diff-1b674ef4a0dfb6975bdc0ddf9dcddb9961e8f5c43e6111fedfb7b3fad8387ec4

Scottcjn:
Thanks @dgarske — confirmed #9954 covers the same overflow in hash.c and ti-hash.c with WC_SAFE_SUM_WORD32. Same bug, your preferred fix style. Glad the pattern got caught across all three paths.

Our other PR #9932 (POWER8 AES vcipher pipeline) is independent — still hoping that makes a future release.

PR #9996, opened 2026-03-17
linuxkm/lkcapi_aes_glue.c: fix scatterwalk_map error handling in AesG…

sameehj:
…cmCrypt_1

When scatterwalk_map fails in either the stream or non-stream path, the code jumped to cleanup without setting err, causing the function to return 0 (success) despite the failure. This could cause the kernel crypto layer to treat uninitialized data as valid ciphertext/plaintext.

- Capture the error code (PTR_ERR) into err before goto out
- Fix PTR_ERR arguments that incorrectly used assoc instead of in_map/out_map (assoc was NULL or pointed to the wrong mapping)
- Make in_map/out_map NULL assignments unconditional (previously gated behind < 6.15, but the cleanup at out: checks these pointers on all kernel versions)
- Remove bogus scatterwalk_unmap of a failed walk in the stream path on >= 6.15

# Description

Fixes zd#

# Testing

How did you test?

# Checklist

 - [NA] added tests
 - [NA] updated/added doxygen
 - [NA] updated appropriate READMEs
 - [NA] Updated manual and documentation

### DIFF
--- a/linuxkm/lkcapi_aes_glue.c
+++ b/linuxkm/lkcapi_aes_glue.c
@@ -1148,12 +1148,11 @@ static int AesGcmCrypt_1(struct aead_request *req, int decrypt_p, int rfc4106_p)
         assoc = scatterwalk_map(&assocSgWalk);
 #endif
         if (unlikely(IS_ERR(assoc))) {
+            err = (int)PTR_ERR(assoc);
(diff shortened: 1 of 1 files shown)


sameehj:
retest this please

PR #10004, opened 2026-03-18
Add custom BIO callback dispatching

julek-wolfssl:
- Route BIO_ctrl_pending, BIO_reset, and BIO_get_mem_data through the custom method's ctrlCb when set, enabling fully custom BIO types to handle these operations.
- Add test_wolfSSL_BIO_custom_method that exercises a custom BIO with all callbacks (create, destroy, read, write, puts, gets, ctrl) and verifies each callback is invoked via bitfield tracking.


julek-wolfssl:
Retest this please flaky test

dgarske (approved):

PR #10006, opened 2026-03-18
DTLS 1.3 write dup support

julek-wolfssl:
- Copy TLS 1.3 traffic secrets and DTLS 1.3 epoch/cipher state to the
  write-dup side in DupSSL so key updates can be performed.
- Delegate KeyUpdate responses from the read side to the write side via
  the shared WriteDup struct, for both peer-initiated and local key
  updates.
- Delegate DTLS 1.3 ACK sending from the read side to the write side.
- Track DTLS 1.3 KeyUpdate ACKs: write side records the in-flight
  KeyUpdate epoch/seq, read side sets keyUpdateAcked when the matching
  ACK arrives.
- Delegate post-handshake certificate authentication (CertificateRequest
  processing) from the read side to the write side, transferring
  transcript hashes, cert context, and signature parameters.
- Reset prevSent/plainSz to prevent stale values from SendData to think
  that data was already sent.
- Refactor FreeHandshakeHashes into Free_HS_Hashes for reuse.
- Move DTLS 1.3 epoch initialization earlier in InitSSL so the
  write-dup early-return path has valid epoch state.
- Add tests for write dup with all protocol versions, key update,
  post-handshake auth, and WANT_WRITE recovery.
- Add --enable-all --enable-writedup to CI os-check matrix.

### DIFF
--- a/.github/workflows/os-check.yml
+++ b/.github/workflows/os-check.yml
@@ -99,6 +99,7 @@ jobs:
           '--enable-certreq --enable-certext --enable-certgen --disable-secure-renegotiation-info CPPFLAGS="-DNO_TLS"',
           '--enable-ocsp --enable-ocsp-responder --enable-ocspstapling CPPFLAGS="-DWOLFSSL_NONBLOCK_OCSP" --enable-maxfragment',
           '--enable-all CPPFLAGS=-DWOLFSSL_HASH_KEEP',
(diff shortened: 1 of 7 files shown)


julek-wolfssl:
Retest this please flaky test.

dgarske (approved):

PR #10009, opened 2026-03-18
Add SHE (Secure Hardware Extension) support to wolfCrypt

night1rider:
Implements software-only SHE CMD_LOAD_KEY message generation (M1/M2/M3)
  and verification message computation (M4/M5) with optional crypto
  callback support for hardware offload.                                                       
   
  New files:                                                                                   
  - wolfssl/wolfcrypt/she.h: API declarations, constants, wc_SHE context
  - wolfcrypt/src/she.c: Miyaguchi-Preneel KDF, message generation,                            
    verification computation, and crypto callback integration
  - tests/api/test_she.c/h: API tests with 97% line / 100% function coverage                   
                                                                                               
  API:
  - wc_SHE_Init/Free, Init_Id, Init_Label (lifecycle)                                          
  - wc_SHE_GenerateM1M2M3 (generate M1/M2/M3 to caller buffers, callback optional)
  - wc_SHE_GenerateM4M5 (generate M4/M5 to caller buffers, independent of                      
    M1/M2/M3, callback optional)                                                               
  - wc_SHE_ImportM1M2M3 (import external M1/M2/M3 into context)                                
  - wc_SHE_GetUID (callback required, fetches UID from hardware)                               
  - wc_SHE_GetCounter (callback required, reads monotonic counter from hardware)               
  - wc_SHE_ExportKey (callback required, exports key slot as M1-M5 from hardware)              
  - wc_SHE_SetKdfConstants (callback optional, WOLFSSL_SHE_EXTENDED only)                      
  - wc_SHE_SetM2Header (callback optional, WOLFSSL_SHE_EXTENDED only)                          
  - wc_SHE_SetM4Header (callback optional, WOLFSSL_SHE_EXTENDED only)                          
  - No key material stored in the context; all inputs passed at generation time
                                                                                               
  Crypto callback integration:                                                                 
  - WC_ALGO_TYPE_SHE added to wc_AlgoType enum
  - Callback sub-types: WC_SHE_SET_UID, WC_SHE_GET_COUNTER,                                    
    WC_SHE_GENERATE_M1M2M3, WC_SHE_GENERATE_M4M5, WC_SHE_EXPORT_KEY                            
  - Per-operation structs in wc_CryptoInfo union with full parameter passthrough               
  - Dispatch functions in cryptocb.c                                                           
  - Optional features: NO_WC_SHE_GETUID, NO_WC_SHE_GETCOUNTER,                                 
    NO_WC_SHE_EXPORTKEY, NO_WC_SHE_IMPORT_M123                                                 
                                                                                               
  Build system:                                                                                
  - configure.ac: --enable-she=standard|extended                                               
  - CMakeLists.txt: WOLFSSL_SHE standard|extended|no
  - src/include.am: she.c compilation
  - .github/workflows/os-check.yml: CI configs for standard, extended,                         
    and NO_* disable flag combinations                                                         
                                                                                               
  Ported from wolfHSM wh_she_crypto.c.

### DIFF
--- a/wolfcrypt/src/wc_she.c
+++ b/wolfcrypt/src/wc_she.c
@@ -359,7 +359,7 @@ int wc_SHE_SetKdfConstants(wc_SHE* she,
 
 #endif /* WOLFSSL_SHE_EXTENDED */
 
-#if defined(WOLF_CRYPTO_CB) || !defined(NO_WC_SHE_IMPORT_M123)
+#if defined(WOLF_CRYPTO_CB) && !defined(NO_WC_SHE_IMPORT_M123)
 /* -------------------------------------------------------------------------- */
 /* Import M1/M2/M3                                                            */
 /*                                                                            */
@@ -384,7 +384,7 @@ int wc_SHE_ImportM1M2M3(wc_SHE* she,
     she->generated = 1;
     return 0;
 }
-#endif /* WOLF_CRYPTO_CB || !NO_WC_SHE_IMPORT_M123 */
+#endif /* WOLF_CRYPTO_CB && !NO_WC_SHE_IMPORT_M123 */
 
 /* -------------------------------------------------------------------------- */
 /* Portable big-endian 32-bit store                                           */
@@ -776,7 +776,7 @@ int wc_SHE_GenerateM4M5(wc_SHE* she,
 /* and frees the context.                                                      */
 /* -------------------------------------------------------------------------- */
 #ifndef NO_WC_SHE_LOADKEY
-#if defined(WOLF_CRYPTO_CB) || !defined(NO_WC_SHE_IMPORT_M123)
+#if defined(WOLF_CRYPTO_CB) && !defined(NO_WC_SHE_IMPORT_M123)
 static int wc_SHE_LoadKey_Internal(wc_SHE* she,
     const byte* m1, word32 m1Sz,
     const byte* m2, word32 m2Sz,
@@ -1080,7 +1080,7 @@ int wc_SHE_LoadKey_Verify_Label(
 }
 #endif /* WOLF_PRIVATE_KEY_ID */
 
-#endif /* WOLF_CRYPTO_CB || !NO_WC_SHE_IMPORT_M123 */
+#endif /* WOLF_CRYPTO_CB && !NO_WC_SHE_IMPORT_M123 */
 #endif /* !NO_WC_SHE_LOADKEY */
 
 /* -------------------------------------------------------------------------- */
--- a/wolfssl/wolfcrypt/wc_she.h
+++ b/wolfssl/wolfcrypt/wc_she.h
@@ -108,7 +108,7 @@ typedef struct wc_SHE {
     byte   m4pOverride;
 #endif
 
-#if defined(WOLF_CRYPTO_CB) || !defined(NO_WC_SHE_IMPORT_M123)
+#if defined(WOLF_CRYPTO_CB) && !defined(NO_WC_SHE_IMPORT_M123)
     byte   m1[WC_SHE_M1_SZ];
     byte   m2[WC_SHE_M2_SZ];
     byte   m3[WC_SHE_M3_SZ];
@@ -227,7 +227,7 @@ WOLFSSL_API int wc_SHE_SetM4Header(wc_SHE* she,
  *   m2Sz - must be WC_SHE_M2_SZ (32)
  *   m3   - 16-byte M3 message (CMAC over M1|M2)
  *   m3Sz - must be WC_SHE_M3_SZ (16) */
-#if defined(WOLF_CRYPTO_CB) || !defined(NO_WC_SHE_IMPORT_M123)
+#if defined(WOLF_CRYPTO_CB) && !defined(NO_WC_SHE_IMPORT_M123)
 WOLFSSL_API int wc_SHE_ImportM1M2M3(wc_SHE* she,
                           const byte* m1, word32 m1Sz,
                           const byte* m2, word32 m2Sz,
@@ -301,7 +301,7 @@ WOLFSSL_API int wc_SHE_GenerateM4M5(wc_SHE* she,
  *   m1-m3 - input: externally-provided SHE key update messages
  *   m4,m5 - output: verification messages returned by the HSM */
 #ifndef NO_WC_SHE_LOADKEY
-#if defined(WOLF_CRYPTO_CB) || !defined(NO_WC_SHE_IMPORT_M123)
+#if defined(WOLF_CRYPTO_CB) && !defined(NO_WC_SHE_IMPORT_M123)
 WOLFSSL_API int wc_SHE_LoadKey(
     void* heap, int devId,
     const byte* m1, word32 m1Sz,
@@ -371,7 +371,7 @@ WOLFSSL_API int wc_SHE_LoadKey_Verify_Label(
     const byte* m5Expected, word32 m5ExpectedSz);
 #endif /* WOLF_PRIVATE_KEY_ID */
 
-#endif /* WOLF_CRYPTO_CB || !NO_WC_SHE_IMPORT_M123 */
+#endif /* WOLF_CRYPTO_CB && !NO_WC_SHE_IMPORT_M123 */
 #endif /* !NO_WC_SHE_LOADKEY */
 
 /* Export a key from hardware in SHE loadable format (M1-M5).
--- a/tests/api/test_she.c
+++ b/tests/api/test_she.c
@@ -146,7 +146,7 @@ int test_wc_SHE_ImportM1M2M3(void)
 {
     EXPECT_DECLS;
 #if defined(WOLFSSL_SHE) && !defined(NO_AES) && \
-    (defined(WOLF_CRYPTO_CB) || !defined(NO_WC_SHE_IMPORT_M123))
+    (defined(WOLF_CRYPTO_CB) && !defined(NO_WC_SHE_IMPORT_M123))
     wc_SHE she;
     byte m1[WC_SHE_M1_SZ] = {0};
     byte m2[WC_SHE_M2_SZ] = {0};
--- a/wolfcrypt/test/test.c
+++ b/wolfcrypt/test/test.c
@@ -56909,7 +56909,7 @@ WOLFSSL_TEST_SUBROUTINE wc_test_ret_t she_test(void)
         goto exit_SHE_Test;
     }
 
-#if defined(WOLF_CRYPTO_CB) || !defined(NO_WC_SHE_IMPORT_M123)
+#if defined(WOLF_CRYPTO_CB) && !defined(NO_WC_SHE_IMPORT_M123)
     /* ---- Import M1/M2/M3 and generate M4/M5 (only NewKey needed) ---- */
     wc_SHE_Free(she);
     ret = wc_SHE_Init(she, HEAP_HINT, devId);
@@ -56944,7 +56944,7 @@ WOLFSSL_TEST_SUBROUTINE wc_test_ret_t she_test(void)
         ret = WC_TEST_RET_ENC_NC;
         goto exit_SHE_Test;
     }
-#endif /* WOLF_CRYPTO_CB || !NO_WC_SHE_IMPORT_M123 */
+#endif /* WOLF_CRYPTO_CB && !NO_WC_SHE_IMPORT_M123 */
 
     /* ---- One-shot M1/M2/M3 ---- */
     wc_SHE_Free(she);
@@ -57062,7 +57062,7 @@ WOLFSSL_TEST_SUBROUTINE wc_test_ret_t she_test(void)
     }
 
 #if !defined(NO_WC_SHE_LOADKEY) && !defined(NO_WC_SHE_LOADKEY_TEST) && \
-    (defined(WOLF_CRYPTO_CB) || !defined(NO_WC_SHE_IMPORT_M123))
+    (defined(WOLF_CRYPTO_CB) && !defined(NO_WC_SHE_IMPORT_M123))
     /* ---- LoadKey_Verify ---- */
     /* Override WC_TEST_SHE_LOADKEY_VERIFY to use a platform-specific variant.
      * Platforms with hardware SHE (HSM/HSE) cannot use static test vectors


.wolfssl_known_macro_extras
```diff
@@ -883,6 +888,9 @@ WOLFSSL_SE050_NO_TRNG
 WOLFSSL_SECURE_RENEGOTIATION_ON_BY_DEFAULT
 WOLFSSL_SERVER_EXAMPLE
 WOLFSSL_SETTINGS_FILE
+WOLFSSL_SH224
```
bigbrett: ?

wolfssl/wolfcrypt/wc_she.h
bigbrett, line 1: I'm wondering if `she.{c,h}` is bound to cause some annoying conflicts in automotive codebases and vendor libs. Perhaps we could namespace it to `wc_she.{c,h}` just to get ahead of that?

wolfssl/wolfcrypt/she.h
```diff
@@ -0,0 +1,346 @@
...
+                          const byte* m1, word32 m1Sz,
+                          const byte* m2, word32 m2Sz,
+                          const byte* m3, word32 m3Sz);
+#endif
+
+/* Generate M1/M2/M3 for the SHE key update protocol and write to
+ * caller-provided buffers. Callback optional -- if a callback is
+ * registered it is tried first; if it returns CRYPTOCB_UNAVAILABLE
+ * the software path runs. This allows a secure element to generate
+ * M1/M2/M3 when it holds the auth key internally.
```
bigbrett: I think it is a little confusing to mention the "Callback optional" field here, since the dispatch to cryptoCb is internal and not exposed in this module. For example, we don't say "Callback optional" in our AES API. 

Same comment for other functions in this file

wolfssl/wolfcrypt/wc_she.h
bigbrett, line 1: make sure to update the out-of-tree doxygen with your lovely API comments too, so it renders in the website docs!

wolfssl/wolfcrypt/she.h
```diff
@@ -0,0 +1,346 @@
...
+
+/* Internal: Miyaguchi-Preneel AES-128 one-way compression.
+ * H_0 = 0, H_i = E_{H_{i-1}}(M_i) XOR M_i XOR H_{i-1}.
+ * Only valid for AES-128 where key size equals block size.
+ * Exposed via WOLFSSL_TEST_VIS for testing.
+ *   aes   - caller-owned, already-initialized Aes structure
+ *   in    - input data (e.g. BaseKey || KDF_Constant, 32 bytes)
+ *   inSz  - length of input in bytes (zero-padded to block boundary)
+ *   out   - output buffer for 16-byte compressed result */
+WOLFSSL_TEST_VIS int wc_She_AesMp16(Aes* aes, const byte* in, word32 inSz,
```
bigbrett: All the other APIs use `wc_SHE` but this has `wc_She`. Unify for consistency

configure.ac
```diff
@@ -5938,6 +6116,25 @@ fi
 AS_IF([test "x$ENABLED_CMAC" = "xyes"],
       [AM_CFLAGS="$AM_CFLAGS -DWOLFSSL_CMAC -DWOLFSSL_AES_DIRECT"])
 
+# SHE (Secure Hardware Extension) key update message generation
+# --enable-she=standard: standard SHE support
+# --enable-she=extended: standard + extended overrides (custom KDF/headers)
+AC_ARG_ENABLE([she],
```
bigbrett, line 5951: do we want to auto-enable CMAC if SHE is enabled?

wolfcrypt/src/she.c
```diff
@@ -0,0 +1,725 @@
...
+int wc_SHE_Init(wc_SHE* she, void* heap, int devId)
+{
+    if (she == NULL) {
+        return BAD_FUNC_ARG;
+    }
+
+    ForceZero(she, sizeof(wc_SHE));
+    she->heap  = heap;
+    she->devId = devId;
+    /* kdfEncOverride/kdfMacOverride are zero from XMEMSET -- defaults used */
```
bigbrett: unnecessary comment

wolfcrypt/src/she.c
```diff
@@ -0,0 +1,725 @@
...
+    }
+
+    ret = wc_SHE_Init(she, heap, devId);
+    if (ret != 0) {
+        return ret;
+    }
+
+    if (len < 0 || len > WC_SHE_MAX_ID_LEN) {
+        return BUFFER_E;
+    }
```
bigbrett: consider checking before call to `wc_SHE_Init()`

wolfcrypt/src/wc_she.c
```diff
@@ -0,0 +1,725 @@
...
+
+/* -------------------------------------------------------------------------- */
+/* GetUID -- callback required                                                 */
+/*                                                                            */
+/* Dispatches to callback to fetch UID from hardware.                        */
+/* Buffer size validation is the callback's responsibility.                  */
+/* Returns CRYPTOCB_UNAVAILABLE if no callback.                              */
+/* -------------------------------------------------------------------------- */
+#if defined(WOLF_CRYPTO_CB) && !defined(NO_WC_SHE_GETUID)
+int wc_SHE_GetUID(wc_SHE* she, byte* uid, word32 uidSz,
```
bigbrett, line 265: I wonder if we should have a default global "wolfCrypt" UID just for lulz / so it could be used software-only for test implementations before hardware is brought online? Could be settable via build macro.

wolfcrypt/src/she.c
```diff
@@ -0,0 +1,725 @@
...
+/* -------------------------------------------------------------------------- */
+#if defined(WOLF_CRYPTO_CB) && !defined(NO_WC_SHE_GETUID)
+int wc_SHE_GetUID(wc_SHE* she, byte* uid, word32 uidSz,
+                   const void* ctx)
+{
+    if (she == NULL || uid == NULL) {
+        return BAD_FUNC_ARG;
+    }
+
+    return wc_CryptoCb_SheSetUid(she, uid, uidSz, ctx);
```
bigbrett: shouldn't this be "get" UID based on calling function? Though the crypto callbacks do only have a set function. What is the intent here?

wolfcrypt/src/wc_she.c
```diff
@@ -0,0 +1,725 @@
...
+#endif /* WOLF_CRYPTO_CB && !NO_WC_SHE_GETUID */
+
+/* -------------------------------------------------------------------------- */
+/* GetCounter -- callback required                                             */
+/*                                                                            */
+/* Dispatches to callback to read current counter from hardware.             */
+/* Returns CRYPTOCB_UNAVAILABLE if no callback.                              */
+/* -------------------------------------------------------------------------- */
+#if defined(WOLF_CRYPTO_CB) && !defined(NO_WC_SHE_GETCOUNTER)
+int wc_SHE_GetCounter(wc_SHE* she, word32* counter, const void* ctx)
```
bigbrett, line 301: Same comment here - I wonder if a default software impl with dumb global counter would be worthwhile just to enable testing? Could have a build macro enabling this to prevent inclusion in production builds maybe?

wolfcrypt/src/cryptocb.c
```diff
@@ -2035,6 +2035,187 @@ int wc_CryptoCb_Cmac(Cmac* cmac, const byte* key, word32 keySz,
 }
 #endif /* WOLFSSL_CMAC */
 
+#ifdef WOLFSSL_SHE
+int wc_CryptoCb_SheSetUid(wc_SHE* she, const byte* uid, word32 uidSz,
```
bigbrett: I'm confused, can you set the UID? Shouldn't this be "get" UID? Same comment is made in she.c. 

Are we setting ID in local wolfCrypt struct or are we getting it from HW?

wolfcrypt/src/wc_she.c
```diff
@@ -0,0 +1,725 @@
...
+        ret = wc_CryptoCb_SheGenerateM4M5(she, uid, uidSz,
+                  authKeyId, targetKeyId,
+                  newKey, newKeySz, counter,
+                  m4, m4Sz, m5, m5Sz);
+        if (ret != WC_NO_ERR_TRACE(CRYPTOCB_UNAVAILABLE)) {
+            return ret;
+        }
+        /* fall-through to software path */
+    }
+#endif
```
bigbrett, line 677: Need to reset `ret=0` in fallback case for good measure

wolfcrypt/src/wc_she.c
```diff
@@ -0,0 +1,725 @@
...
+    if (she == NULL) {
+        return;
+    }
+
+#if defined(WOLF_CRYPTO_CB) && defined(WOLF_CRYPTO_CB_FREE)
+    if (she->devId != INVALID_DEVID) {
+        int ret = wc_CryptoCb_Free(she->devId, WC_ALGO_TYPE_SHE,
+                                   0, 0, she);
+        if (ret != WC_NO_ERR_TRACE(CRYPTOCB_UNAVAILABLE)) {
+            return;
```
bigbrett, line 247: Ever a case where we want to ForceZero here?

bigbrett (changes requested):
Absolutely fantastic work @night1rider. A few issues to address but overall looks great.

.wolfssl_known_macro_extras
night1rider: This was a copy paste mistake I introduced, removed.

wolfssl/wolfcrypt/wc_she.h
night1rider: This makes sense since we already have files following this convention (`wc_encrypt`, `wc_lms`, `wc_mlkem`, `wc_port`, etc.) there's no issue doing it now. Names should be updated now

wolfssl/wolfcrypt/she.h
night1rider: I wanted to mention usefulness of the callbacks somewhere, especially for generating M1/M2/M3 or M4/M5 when they need to access a secure element/hardware. This can be covered in the source comments and docs instead of the public header. I removed the callback language from all five API comments in `wc_she.h`.

wolfssl/wolfcrypt/wc_she.h
night1rider: I will link a PR shortly with the doc update in the main body of the pr description

wolfssl/wolfcrypt/she.h
night1rider: I overlooked this due to it being an internal function, updated the name.

configure.ac
night1rider: Wasn't sure at first if we should initially, but why not. I set it up to auto-enabled CMAC, AES-DIRECT, and AES-CBC in `configure.ac` when SHE is enabled. Also added `#error` guards in `wc_she.c` for `NO_AES`, `HAVE_AES_CBC`, `WOLFSSL_AES_DIRECT`, and `WOLFSSL_CMAC` so users get clear build errors if dependencies are missing if they end up using a user_settings.h configure.

wolfcrypt/src/she.c
night1rider: Removed. This was missed in my initial refactor of my draft code.

wolfcrypt/src/she.c
night1rider: Makes sense. Moved the `len` bounds check above the `wc_SHE_Init()` call so we fail fast before doing unnecessary work.

wolfcrypt/src/wc_she.c
night1rider: I added `WC_SHE_SW_DEFAULT` opt-in macro for both UID and counter software defaults. When defined, `wc_SHE_GetUID` falls back to a built-in default UID (from the SHE spec test vector) and `wc_SHE_GetCounter` falls back to a simple incrementing counter. Default values can be overridden via `WC_SHE_DEFAULT_UID` and `WC_SHE_DEFAULT_COUNTER`. This is opt-in only and callbacks are still required to use these functions without it, so no one accidentally uses the software defaults. Added `she_sw_default_test()` in `test.c` that demonstrates the full `GetUID` -> `GetCounter` -> `GenerateM1M2M3` -> `GenerateM4M5` API flow. All test values are overridable via `WC_SHE_SW_TEST_*` macros to allow for a custom vector to test if we wanted for some reason.

Let me know if you had something else in mind.

wolfcrypt/src/she.c
night1rider: This was overlooked. Renamed `wc_CryptoCb_SheSetUid` to `wc_CryptoCb_SheGetUid`, `WC_SHE_SET_UID` to `WC_SHE_GET_UID`, and the union field `setUid` to `getUid` in `cryptocb.h` and `cryptocb.c`. The intent is to allow someone to use the same API after implementing the callback across different platforms. This is the same intent for GetCounter as well.

wolfcrypt/src/wc_she.c
night1rider: Addressed in my previous comment regarding the "default global "wolfCrypt" UID".

wolfcrypt/src/cryptocb.c
night1rider: I just overlooked this, addressed. We are getting the UID from hardware and some hardware supports a challenge to get the UID to prove that the hardware is real, hence the void context parameter. This is just an API to get the UID to then pass to the generate API calls. There is no setting the UID in the structure. The struct is now very basic and holds only M1/M2/M3 if they are imported and the import function is on, plus devId for callbacks, KDF constants, etc.

wolfcrypt/src/wc_she.c
night1rider: Updated, overlooked this.

wolfcrypt/src/wc_she.c
night1rider: We already have a `ForceZero` on the fall-through path. The person who implements their callback should be allowed to control how free works. If they want our software cleanup, they can clean up what they want and then return `CRYPTOCB_UNAVAILABLE` to continue to the ForceZero. If they handle everything themselves, they return success and we skip it and the callback implementer assumes the responsibility.

night1rider:
Rebased off current master

night1rider:
Added wc_SHE_LoadKey, wc_SHE_LoadKey_Id, wc_SHE_LoadKey_Label and their verify counterparts.

These wrap Init, ImportM1M2M3, GenerateM4M5, and Free into a single call for hardware crypto callback usage. Verify variants compare returned M4/M5 against expected values using ConstantCompare. 

All functions require a valid devId and can be compiled out with NO_WC_SHE_LOADKEY.

night1rider:
Jenkins retest this please

night1rider:
Rebased to fix merge conflict

CMakeLists.txt
```diff
@@ -1640,6 +1640,26 @@ if(WOLFSSL_CMAC)
     endif()
 endif()
 
+# SHE (Secure Hardware Extension) key update message generation
+# standard: core SHE support, extended: adds custom KDF/header overrides
+add_option("WOLFSSL_SHE"
+    "Enable SHE key update support (standard|extended|no)"
+    "no" "standard;extended;no")
+
+if(WOLFSSL_SHE STREQUAL "standard" OR WOLFSSL_SHE STREQUAL "extended")
```
dgarske, line 1649: ### HIGH-1: CMakeLists.txt missing required SHE dependency definitions
- **File:** `CMakeLists.txt:1649-1656`
- **Function:** N/A (build system)
- **Category:** bug
- **Confidence:** High

**Description:** The CMakeLists.txt SHE section only adds `-DWOLFSSL_SHE` but does not add `-DWOLFSSL_CMAC` or `-DWOLFSSL_AES_DIRECT`, nor does it force-enable the CMAC option. By contrast, `configure.ac` correctly adds all three flags and sets `ENABLED_CMAC=yes` and `ENABLED_AESCBC=yes`. A CMake build with `-DWOLFSSL_SHE=standard` (without separately enabling CMAC) will hit the `#error "SHE requires CMAC"` directive in `wc_she.c`.

**Code:**
```cmake
if(WOLFSSL_SHE STREQUAL "standard" OR WOLFSSL_SHE STREQUAL "extended")
    if (NOT WOLFSSL_AES)
        message(FATAL_ERROR "Cannot use SHE without AES.")
    else()
        list(APPEND WOLFSSL_DEFINITIONS
            "-DWOLFSSL_SHE")
    endif()
endif()
```

**Recommendation:** Mirror the `configure.ac` behavior: force-enable CMAC (and thus AES_DIRECT) when SHE is enabled. For example:

```suggestion
if(WOLFSSL_SHE STREQUAL "standard" OR WOLFSSL_SHE STREQUAL "extended")
    if (NOT WOLFSSL_AES)
        message(FATAL_ERROR "Cannot use SHE without AES.")
    else()
        list(APPEND WOLFSSL_DEFINITIONS
            "-DWOLFSSL_SHE")
        # SHE requires CMAC and AES-DIRECT; force-enable them
        override_cache(WOLFSSL_CMAC "yes")
    endif()
endif()
```

wolfcrypt/src/cryptocb.c
```diff
@@ -2035,6 +2035,187 @@ int wc_CryptoCb_Cmac(Cmac* cmac, const byte* key, word32 keySz,
 }
 #endif /* WOLFSSL_CMAC */
 
+#ifdef WOLFSSL_SHE
+int wc_CryptoCb_SheGetUid(wc_SHE* she, const byte* uid, word32 uidSz,
```
dgarske: ### HIGH-2: `wc_CryptoCb_SheGetUid` uid parameter is incorrectly `const`
- **File:** `wolfcrypt/src/cryptocb.c:2039`, `wolfssl/wolfcrypt/cryptocb.h:824`, `wolfssl/wolfcrypt/cryptocb.h:471`
- **Function:** `wc_CryptoCb_SheGetUid`
- **Category:** bug
- **Confidence:** High

**Description:** The `uid` parameter in `wc_CryptoCb_SheGetUid` is declared as `const byte*`, and the `getUid.uid` field in the `wc_CryptoInfo.she.op` union is also `const byte*`. However, the purpose of `wc_SHE_GetUID` (which calls this) is to **fill** the uid buffer — the header doc says "buffer to receive the 120-bit (15-byte) SHE UID". The software fallback confirms this by writing to it: `XMEMCPY(uid, wc_She_DefaultUid, ...)`. A hardware callback implementation would need to cast away `const` to write the UID, which is undefined behavior in C.

**Code:**
```c
// cryptocb.h:824
WOLFSSL_LOCAL int wc_CryptoCb_SheGetUid(wc_SHE* she, const byte* uid, ...);

// cryptocb.h:471 (inside wc_CryptoInfo)
struct {
    const byte* uid;   // <-- should be byte*
    word32      uidSz;
} getUid;
```

**Recommendation:** Change `const byte* uid` to `byte* uid` in:
1. The `wc_CryptoCb_SheGetUid` function signature (declaration and definition)
2. The `getUid.uid` field in `wc_CryptoInfo`

wolfcrypt/src/wc_she.c
```diff
@@ -0,0 +1,1114 @@
...
+        XMEMCPY(she->kdfMacC, macC, WC_SHE_KEY_SZ);
+        she->kdfMacOverride = 1;
+    }
+
+    return 0;
+}
+
+#endif /* WOLFSSL_SHE_EXTENDED */
+
+/* -------------------------------------------------------------------------- */
```
dgarske: ### MEDIUM-3: Stale section comment before `wc_SHE_ImportM1M2M3`
- **File:** `wolfcrypt/src/wc_she.c:362-363`
- **Function:** N/A
- **Category:** style
- **Confidence:** High

**Description:** There's a stale section header comment `/* GetUID */` immediately before the `wc_SHE_ImportM1M2M3` function. It appears to be a copy-paste leftover from an earlier draft.

**Code:**
```c
/* -------------------------------------------------------------------------- */
/* GetUID                                                                     */

#if defined(WOLF_CRYPTO_CB) || !defined(NO_WC_SHE_IMPORT_M123)
/* -------------------------------------------------------------------------- */
/* Import M1/M2/M3                                                            */
```

**Recommendation:** Remove the stale `/* GetUID */` comment block at lines 362-363 or replace with `/* Import M1/M2/M3 */`.

wolfcrypt/src/wc_she.c
```diff
@@ -0,0 +1,1114 @@
...
+}
+
+/* -------------------------------------------------------------------------- */
+/* wc_SHE_LoadKey                                                              */
+/*                                                                            */
+/* One-shot: Init, ImportM1M2M3, GenerateM4M5 (via callback), Free.           */
+/* Requires a valid devId (not INVALID_DEVID) since the operation dispatches   */
+/* to a hardware crypto callback.                                              */
+/* -------------------------------------------------------------------------- */
+int wc_SHE_LoadKey(
```
dgarske, line 811: ### MEDIUM-4: No test coverage for `wc_SHE_LoadKey*` and `wc_SHE_LoadKey_Verify*` functions
- **File:** `wolfcrypt/src/wc_she.c:814-975`, `tests/api/test_she.c`, `wolfcrypt/test/test.c`
- **Function:** `wc_SHE_LoadKey`, `wc_SHE_LoadKey_Id`, `wc_SHE_LoadKey_Label`, `wc_SHE_LoadKey_Verify*`
- **Category:** test
- **Confidence:** High

**Description:** The one-shot `wc_SHE_LoadKey`, `wc_SHE_LoadKey_Id`, `wc_SHE_LoadKey_Label`, `wc_SHE_LoadKey_Verify`, `wc_SHE_LoadKey_Verify_Id`, and `wc_SHE_LoadKey_Verify_Label` functions (9 public API functions total) have no test coverage in either `tests/api/test_she.c` or `wolfcrypt/test/test.c`. These are convenience wrappers around the tested primitives, but they contain their own parameter validation and internal orchestration logic (including the `wc_SHE_LoadKey_Internal` helper which calls `GenerateM4M5` with NULL uid/newKey) that deserves coverage.

**Recommendation:** Add API tests for at least `wc_SHE_LoadKey` (basic + bad args) and `wc_SHE_LoadKey_Verify` (match + mismatch) using the crypto callback test infrastructure already established in `test_wc_SHE_CryptoCb`.

configure.ac
```diff
@@ -5946,6 +5946,27 @@ fi
...
+                    [Enable SHE key update support (default: disabled)])],
+    [ ENABLED_SHE=$enableval ],
+    [ ENABLED_SHE=no ]
+    )
+
+if test "x$ENABLED_SHE" = "xstandard" || test "x$ENABLED_SHE" = "xextended"
+then
+    AM_CFLAGS="$AM_CFLAGS -DWOLFSSL_SHE -DWOLFSSL_CMAC -DWOLFSSL_AES_DIRECT"
+    ENABLED_CMAC=yes
+    ENABLED_AESCBC=yes
```
dgarske, line 5966: ### MEDIUM-5: configure.ac SHE section sets `ENABLED_AESCBC=yes` too late to take effect
- **File:** `configure.ac:5963`
- **Function:** N/A (build system)
- **Category:** bug
- **Confidence:** Medium

**Description:** The SHE configure block sets `ENABLED_AESCBC=yes` (line 5963), but the AES-CBC disable check already ran at line 3287. If a user passes `--disable-aescbc --enable-she=standard`, the `-DNO_AES_CBC` flag is already added to `AM_CFLAGS` before the SHE block runs. Setting `ENABLED_AESCBC=yes` at that point doesn't remove the already-appended flag. The `#error` in `wc_she.c` will catch this at compile time, but it would be better to fail at configure time with a clear error message.

**Code:**
```bash
# Line 3287 (runs first):
if test "$ENABLED_AESCBC" = "no"
then
    AM_CFLAGS="$AM_CFLAGS -DNO_AES_CBC"  # Already added!
fi

# Line 5959 (runs later):
if test "x$ENABLED_SHE" = "xstandard" || ...
then
    ENABLED_AESCBC=yes  # Too late, -DNO_AES_CBC already in AM_CFLAGS
fi
```

**Recommendation:** Either add a check for `NO_AES_CBC` after the SHE block or, more robustly, move the `ENABLED_AESCBC` check to after all feature blocks that might override it (similar to how some flags are consolidated at the end of configure.ac). Alternatively, add an explicit check:
```bash
if test "x$ENABLED_SHE" != "xno" && test "$ENABLED_AESCBC" = "no"
then
    AC_MSG_ERROR([SHE requires AES-CBC. Remove --disable-aescbc or disable SHE.])
fi
```

wolfcrypt/src/wc_she.c
```diff
@@ -0,0 +1,1114 @@
...
+    if (label == NULL || m1 == NULL || m2 == NULL || m3 == NULL ||
+        m4 == NULL || m5 == NULL) {
+        return BAD_FUNC_ARG;
+    }
+
+    if (devId == INVALID_DEVID) {
+        return BAD_FUNC_ARG;
+    }
+
+    if (XSTRLEN(label) == 0 || XSTRLEN(label) > WC_SHE_MAX_LABEL_LEN) {
```
dgarske: ### LOW-7: `XSTRLEN(label)` called twice in `wc_SHE_LoadKey_Label` validation
- **File:** `wolfcrypt/src/wc_she.c:947`
- **Function:** `wc_SHE_LoadKey_Label`
- **Category:** style
- **Confidence:** High

**Description:** `XSTRLEN(label)` is called twice in the same conditional expression — once for the `== 0` check and once for the `> WC_SHE_MAX_LABEL_LEN` check. This redundancy walks the string twice.

**Code:**
```c
if (XSTRLEN(label) == 0 || XSTRLEN(label) > WC_SHE_MAX_LABEL_LEN) {
    return BAD_FUNC_ARG;
}
```

**Recommendation:** Store the result in a local variable:
```suggestion
size_t len = XSTRLEN(label);
if (len == 0 || len > WC_SHE_MAX_LABEL_LEN) {
    return BAD_FUNC_ARG;
}
```

CMakeLists.txt
night1rider: Added `override_cache(WOLFSSL_CMAC "yes")` and
`override_cache(WOLFSSL_AESCBC "yes")` to the CMakeLists.txt SHE block. CMAC's own block
already handles adding `-DWOLFSSL_CMAC -DWOLFSSL_AES_DIRECT` when it's enabled, so the
override is all that's needed to cascade the dependencies correctly.

wolfcrypt/src/cryptocb.c
night1rider: Removed `const` from `uid` in three places:
- `cryptocb.c` definition: `const byte* uid` → `byte* uid`
- `cryptocb.h` declaration: `const byte* uid` → `byte* uid`
- `cryptocb.h` `wc_CryptoInfo.she.op.getUid.uid`: `const byte*` → `byte*`

wolfcrypt/src/wc_she.c
night1rider: Removed the stale `/* GetUID */` section divider comment block
(2 lines) immediately above the `#if defined(WOLF_CRYPTO_CB)` guard for `wc_SHE_ImportM1M2M3`.

wolfcrypt/src/wc_she.c
night1rider: Fixed. Added two new test functions in `tests/api/test_she.c`:
- `test_wc_SHE_LoadKey`: exercises the basic one-shot flow (generates M1/M2/M3 from
  test vectors, calls LoadKey via the crypto callback, verifies M4/M5 match expected
  values) plus bad-arg coverage (NULL inputs, INVALID_DEVID, wrong M1 size).
- `test_wc_SHE_LoadKey_Verify`: tests the matching case (expected M4/M5 equal actual)
  and the mismatch case (flipped byte in expected M4 returns SIG_VERIFY_E).

Updated `test_she_crypto_cb` to handle the LoadKey flow (uid == NULL, generated == 1)
by returning the known test-vector M4/M5. Declarations and `TEST_SHE_CB_DECLS`
registration added in `test_she.h`, guarded by `!NO_WC_SHE_LOADKEY`.

configure.ac
night1rider: Added an explicit `AC_MSG_ERROR` check at the top of the SHE enable
block in `configure.ac`: if `ENABLED_AESCBC` is `"no"` when SHE is being enabled, configure
now fails with `"SHE requires AES-CBC. Cannot use --disable-aescbc with --enable-she."`.

wolfcrypt/src/wc_she.c
night1rider: Fixed. Added a `word32 labelLen` local variable in `wc_SHE_LoadKey_Label`,
assigned once via `labelLen = (word32)XSTRLEN(label)`, and used `labelLen` in the
bounds check.

night1rider:
Had to rebase and force push due to merge conflict with `.github/workflows/os-check.yml`.

night1rider:
Jenkins retest this please

night1rider:
Rebased and fixed merge conflict of `wolfssl/wolfcrypt/types.h` caused by merging #9851

wolfcrypt/src/cryptocb.c
```diff
@@ -2120,6 +2120,187 @@ int wc_CryptoCb_Cmac(Cmac* cmac, const byte* key, word32 keySz,
 }
 #endif /* WOLFSSL_CMAC */
 
+#ifdef WOLFSSL_SHE
+int wc_CryptoCb_SheGetUid(wc_SHE* she, byte* uid, word32 uidSz,
```
dgarske, line 2124: Could this be in wc_she.c?

wolfssl/wolfcrypt/cryptocb.h
```diff
@@ -478,6 +481,65 @@ typedef struct wc_CryptoInfo {
         int type;
     } cmac;
 #endif
+#ifdef WOLFSSL_SHE
```
dgarske, line 484: Do we really have to customize cryptocb to add this SHE support? Are the existing ones not enough + the keyCtx that each has?

wolfcrypt/src/wc_she.c
```diff
@@ -0,0 +1,1113 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1335, USA
+ */
+
+/*
+ * SHE (Secure Hardware Extension) key update message generation.
```
dgarske, line 23: Where is the README.md for SHE? What hardware platform does this run on? How does someone test it?

wolfcrypt/src/cryptocb.c
night1rider: The existing convention in wolfSSL is that all `wc_CryptoCb_*` wrapper functions live in `cryptocb.c`. `wc_CryptoCb_AesCbcEncrypt`, `wc_CryptoCb_Rsa`, `wc_CryptoCb_EccSign`, `wc_CryptoCb_Sha256Hash`, `wc_CryptoCb_Cmac`, etc. live in `cryptocb.c`, not in `aes.c` / `rsa.c` / `ecc.c` / `sha256.c` / `cmac.c`.

I had SHE wrappers follow that same pattern so the callback dispatching logic stays in one file and wc_she.c does not have to pull in cryptocb.h / the dispatch plumbing.

I could look at moving them to `wc_she.c` if that is preferred, but it would be the first deviation from the current pattern.

wolfssl/wolfcrypt/cryptocb.h
night1rider: The `she` union exists to carry the arguments of the public SHE API across the callback. Each SHE API function has a distinct argument list that needs to reach the callback unchanged.

`wc_SHE_GenerateM1M2M3` takes `{wc_SHE*, uid, uidSz, authKeyId, authKey, authKeySz, targetKeyId, newKey, newKeySz, counter, flags, m1, m1Sz, m2, m2Sz, m3, m3Sz}`, and the other SHE calls each have their own shape. 

The `she` sub structs were added to mirror each SHE API call's argument list 1:1 so the callback receives exactly what the caller passed in without any repacking.

I can restructure if you have a specific layout in mind, or suggestion for how to pass the API parameters.

wolfcrypt/src/wc_she.c
night1rider: I just put the PR for the  manual chapter for SHE in [wolfSSL/documentation#263](https://github.com/wolfSSL/documentation/pull/263), and a follow up commit that adds the doxygen for the public SHE API at wolfSSL/wolfssl@85423cd152e38b77824fc5bf7f3bb4d73c454cba. I can add that doxygen commit into this PR if you would like it here instead of as a separate change.

This only implements the software M1 through M5 generation path (`wc_SHE_Init`, `wc_SHE_GenerateM1M2M3`, `wc_SHE_GenerateM4M5`, `wc_SHE_VerifyM4M5`, `wc_SHE_Free`).

To load a SHE key on a device or use `ImportM1M2M3` / `GetUID` / `GetCounter` / `LoadKey` / `LoadKey_Verify` / `ExportKey`, pair this PR with a crypto callback (our NXP S32K3 HSE port is an example) or use wolfHSM instead. `LoadKey` returns `BAD_FUNC_ARG` on `INVALID_DEVID`. `GetUID` / `GetCounter` have a `WC_SHE_SW_DEFAULT` stub, off by default and labeled example only (dummy UID, incrementing static counter).

Nothing special is required to test. Any build with `WOLFSSL_CMAC` and `WOLFSSL_SHE` enabled runs the software path via `she_test()` in `./wolfcrypt/test/testwolfcrypt`. Test vectors come from https://github.com/wolfSSL/wolfHSM/blob/main/test/wh_test_she.c so results line up with wolfHSM.

Two common uses: (1) provisioning / key manager server tooling, where a cloud backend generates M1-M5 to push to devices and a source side callback handles counter tracking or protocol header packing; (2) vendor extended SHE specs with custom KDF constants or header layouts.

dgarske (approved):

bigbrett (approved):

PR #10014, opened 2026-03-19
Add missing include in asn.h

danielinux:
wolfssl/wolfcrypt/asn.h uses WC_SHA384_DIGEST_SIZE and WC_SHA512_DIGEST_SIZE, but did not include wolfssl/wolfcrypt/sha512.h directly. This relied on transitive include order and broke builds where asn.h is  parsed before hash.h/sha512.h, such as hash.c with SHA-384 enabled.

Regression introduced in #9761


dgarske:
Merged early

PR #10017, opened 2026-03-19
Fix ssl_DecodePacketInternal chain processing

embhorn:
# Description

Problem: The chain accumulation loop summed size_t (64-bit) iov_len values into an int (32-bit) length, causing silent truncation. An undersized buffer was allocated, then full data was copied causing a heap overflow.

Fixes zd21388

# Testing

Added `test_sniffer_chain_input_overflow`

# Checklist

 - [x] added tests
 - [ ] updated/added doxygen


embhorn:
Fix confirmed by reporter

dgarske (approved):

PR #10020, opened 2026-03-19
SP int: fixes from AI review

SparkiDev:
# Description

Re-implemented wc_PKCS12_PBKDF() to not use MP. Added tests to unit.test.

sp_int.c:
  Fixes to comments.
  Added more define build options documentation to top of file.
  Fixes for builds with WOLFSSL_SP_INT_NEGATIVE defined.
Fixes for when a->used is 0 and no underflow - not actually a problem but cleaner code.
  sp_sub has different checks on a->used when values are only positive.
  sp_dic_2d  missing check for e less than zero.
  sp_to_unsigned_bin_len_ct: remove redundant check of outSz. Change i to int to handle a->used of 0 and make code tidier.

Configuration testing fixes.
Fix formatting in test.c.
Added 128-bit types word128 and sword128 for cleaner PKCS#12 code.

# Testing

Regression tested SP int

# Checklist

 - [x] added tests


SparkiDev:
retest this please

PR #10025, opened 2026-03-20
Fix DecodeObjectId unknown ext parse

embhorn:
# Description

Fixes in `DecodeObjectID`:
* Moved the bounds check inside the `y == 0` branch so that it checks `*outSz < 2` before writing out[0] and out[1]. Previously the check `y >= *outSz` only validated one slot but the first-arc code writes two.
* Changed `sizeof(decOid)` to `MAX_OID_SZ` in both `DumpOID()` and the `WC_ASN_UNKNOWN_EXT_CB` code path, so `DecodeObjectId` receives the correct element count (32) rather than the byte count (64).

Fixes zd21392

# Testing

Added `test_wc_DecodeObjectId`

# Checklist

 - [x] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/tests/api/test_asn.c
+++ b/tests/api/test_asn.c
@@ -1092,7 +1092,7 @@ int test_wc_DecodeObjectId(void)
         }
 
         /* Test 6: Buffer too small for later arcs */
-        outSz = 3; /* only room for 3 arcs, but OID has 8 */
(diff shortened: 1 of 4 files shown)


embhorn:
Jenkins retest this please

dgarske (approved):

PR #10027, opened 2026-03-20
Remove FIPS guards in GetASN_BitString length check

embhorn:
# Description

Remove FIPS guards in GetASN_BitString length check

Fixes zd21394

# Testing

Build test

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


kaleb-himes (approved):
This was only compiled out for ctaocrypt (cert 2425) but we're no longer actively supporting that very old FIPS module at this point so this is a fine change from my perspective.

I'll ask @cconlon to also review for the self-test use-case. Good from my side though.

dgarske (approved):

PR #10045, opened 2026-03-23
Fix IAR warning about volatile access

embhorn:
# Description

Read the volatile once into a local copy, then use the copy in expressions — satisfying IAR's stricter volatile access rules. "undefined behavior: the order of volatile accesses is undefined in this statement"

Fixes zd21385

# Testing

Customer confirmed clean build with IAR tools

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -41491,7 +41491,7 @@ static int DefTicketEncCb(WOLFSSL* ssl, byte key_name[WOLFSSL_TICKET_NAME_SZ],
                     {
                         RsaKey* key = (RsaKey*)ssl->hsKey;
                         volatile int lenErrMask;
-                        int lenErrMaskCopy;
+                        int mask;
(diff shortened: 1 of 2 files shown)


embhorn:
Jenkins retest this please

dgarske (changes requested):
These changes feel "hacky". Perhaps they need "volatile" added? I liked the original code better. Can we have IAR ignore that specific warning?

embhorn:
> These changes feel "hacky". Perhaps they need "volatile" added? I liked the original code better. Can we have IAR ignore that specific warning?

The issue is about multiple volatile statements on one line, and the warning is valid.
I changed the logic a bit and I think this new approach is better and more clear.

dgarske (approved):

PR #10051, opened 2026-03-23
Add bounds checks for MP integer size in SizeASN_Items

anhu:
Fixes ZD 21401

### DIFF
--- a/tests/api/test_rsa.c
+++ b/tests/api/test_rsa.c
@@ -1152,3 +1152,130 @@ int test_wc_RsaDecrypt_BoundsCheck(void)
     return EXPECT_RESULT();
 } /* END test_wc_RsaDecryptBoundsCheck */
 
+/*
+ * Test wc_RsaKeyToDer with an mp_int large enough to wrap size calculations.
+ */
+int test_wc_RsaKeyToDer_SizeOverflow(void)
+{
+    EXPECT_DECLS;
+#if !defined(NO_RSA) && defined(USE_INTEGER_HEAP_MATH) && \
+    defined(WOLFSSL_ASN_TEMPLATE) && \
+    (defined(WOLFSSL_KEY_GEN) || defined(WOLFSSL_KEY_TO_DER)) && \
+    !defined(_WIN32)
+#include <sys/mman.h>
+    RsaKey    key;
+    int       i;
+    int       derRet;
+    int       crafted_used;
+    int       top_bits;
+    mp_digit  top_digit;
+    mp_digit* crafted_dp = MAP_FAILED;
+    size_t    dp_bytes = 0;
+
+    int       orig_used  = 0;
+    int       orig_alloc = 0;
+    int       orig_sign  = 0;
+    mp_digit* orig_dp    = NULL;
+
+    mp_int* fields[8];
+
+    XMEMSET(&key, 0, sizeof(key));
+
+    /* Find 'used' count that makes (used-1)*DIGIT_BIT + top_bits = -48
+     * as signed int, causing mp_unsigned_bin_size to return -6. */
+    {
+        unsigned int target = 0xFFFFFFD0u;  /* -48 as unsigned 32-bit */
+        int found = 0;
+
+        crafted_used = 0;
+        top_bits = 0;
+        top_digit = 0;
+
+        for (top_bits = 1; top_bits < DIGIT_BIT; top_bits++) {
+            unsigned int base = target - (unsigned int)top_bits;
+            if (base % (unsigned int)DIGIT_BIT == 0) {
+                crafted_used = (int)(base / (unsigned int)DIGIT_BIT) + 1;
+                top_digit = (mp_digit)1 << (top_bits - 1);
+                found = 1;
+                break;
+            }
+        }
+        if (!found) {
+            return TEST_SKIPPED;
+        }
+    }
+
+    dp_bytes = (size_t)crafted_used * sizeof(mp_digit);
+
+    ExpectIntEQ(wc_InitRsaKey(&key, HEAP_HINT), 0);
+
+    /* Set up dummy RSA private key fields. */
+    key.type = RSA_PRIVATE;
+    fields[0] = &key.n;
+    fields[1] = &key.e;
+    fields[2] = &key.d;
+    fields[3] = &key.p;
+    fields[4] = &key.q;
+    fields[5] = &key.dP;
+    fields[6] = &key.dQ;
+    fields[7] = &key.u;
+
+    for (i = 0; i < 8; i++) {
+        if (EXPECT_SUCCESS()) {
+            ExpectIntEQ(mp_init(fields[i]), 0);
+            mp_set(fields[i], 0x42);
+        }
+    }
+
+    if (EXPECT_SUCCESS()) {
+        orig_used  = key.p.used;
+        orig_alloc = key.p.alloc;
+        orig_sign  = key.p.sign;
+        orig_dp    = key.p.dp;
+    }
+
+    /* Sparse mmap — only the page at dp[used-1] is faulted in. */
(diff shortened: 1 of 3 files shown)


dgarske:
Jenkins retest this please -history lost

anhu:
Jenkins retest this please. 

Unable to get pull request trigger.

anhu:
Jenkins retest this please

Build was aborted

anhu:
jenkins retest this please

remote hung up.

anhu:
jenkins retest this please 

ABORTED

anhu:
jenkins retest this please.

anhu:
Jenkins retest this please.

anhu:
Jenkins retest this please.

anhu:
jenkins retest this please

Not found.

anhu:
Jenkins retest this please.

anhu:
jenkins retest this please.

anhu:
jenkins retest this please.

tests/api/test_rsa.c
```diff
@@ -1152,3 +1152,130 @@ int test_wc_RsaDecrypt_BoundsCheck(void)
...
+    if (EXPECT_SUCCESS()) {
+        orig_used  = key.p.used;
+        orig_alloc = key.p.alloc;
+        orig_sign  = key.p.sign;
+        orig_dp    = key.p.dp;
+    }
+
+    /* Sparse mmap — only the page at dp[used-1] is faulted in. */
+    if (EXPECT_SUCCESS()) {
+        crafted_dp = (mp_digit*)mmap(NULL, dp_bytes,
```
SparkiDev: This seems excessive.
Can't you test it another way that works on platforms not Windows?

wolfcrypt/src/asn.c
```diff
@@ -864,8 +864,18 @@ int SizeASN_Items(const ASNItem* asn, ASNSetData *data, int count,
...
+                    return ASN_PARSE_E;
+                }
                 len = (word32)SizeASNHeader((word32)length) + (word32)length;
+                /* Check for overflow: header + length must not wrap word32. */
+                if (len < (word32)length) {
+                    return ASN_PARSE_E;
+                }
                 break;
 
             case ASN_DATA_TYPE_REPLACE_BUFFER:
```
dgarske, line 881: **🟡 [Medium] SetASN_Items MP path lacks matching bounds checks**
*💡 SUGGEST*  `bug`

The `ASN_DATA_TYPE_MP` case in `SetASN_Items` (the write path) calls `mp_unsigned_bin_size()` and `mp_leading_bit()` on lines 1097-1099 with the same pattern that was just hardened in `SizeASN_Items`, but without the new bounds checks. In normal usage `SizeASN_Items` is called first and would reject a malformed mp_int before `SetASN_Items` runs, so this is not directly exploitable. However, as defense-in-depth—and to prevent future callers from hitting the same issue—the same `length < 0` checks should be added here.

**Suggestion:**
```suggestion
case ASN_DATA_TYPE_MP:
                /* Get length in bytes. */
                length = mp_unsigned_bin_size(data[i].data.mp);
                if (length < 0) {
                    return ASN_PARSE_E;
                }
                /* Add one for leading zero to make encoding a positive num. */
                length += mp_leading_bit(data[i].data.mp) ? 1 : 0;
                if (length < 0) {
                    return ASN_PARSE_E;
                }
                /* Write out length. */
                idx += SetASNLength((word32)length, out + idx);
```

tests/api/test_rsa.c
```diff
@@ -1152,3 +1152,116 @@ int test_wc_RsaDecrypt_BoundsCheck(void)
     return EXPECT_RESULT();
 } /* END test_wc_RsaDecryptBoundsCheck */
 
+/*
+ * Test wc_RsaKeyToDer with an mp_int large enough to wrap size calculations.
+ */
+int test_wc_RsaKeyToDer_SizeOverflow(void)
+{
+    EXPECT_DECLS;
+#if !defined(NO_RSA) && defined(USE_INTEGER_HEAP_MATH) && \
```
dgarske, line 1161: **🟡 [Medium] Test only covers USE_INTEGER_HEAP_MATH; tfm.c has same overflow pattern**
*💡 SUGGEST*  `test`

The test is gated on `USE_INTEGER_HEAP_MATH` because the pointer-biasing trick relies on heap-allocated `dp`. However, `fp_count_bits` in `tfm.c` (line 4116) has the same `(a->used - 1) * DIGIT_BIT` overflow pattern as `mp_count_bits` in `integer.c` (line 262). With TFM (fastmath), `dp` is a fixed-size stack array, so the pointer-biasing trick won't work, but the underlying `SizeASN_Items` fix is still relevant for TFM builds. Consider adding a separate test path for TFM (e.g., using `mp_set_int` or another approach) or at minimum adding a comment explaining why TFM is excluded.

**Recommendation:** Consider adding a comment or separate test variant for TFM builds. Even if the pointer trick doesn't work with stack-allocated digits, verifying that `SizeASN_Items` rejects a negative `mp_unsigned_bin_size` result is valuable across all math backends.

tests/api/test_rsa.c
```diff
@@ -1152,3 +1152,116 @@ int test_wc_RsaDecrypt_BoundsCheck(void)
...
+                  - (wc_ptr_t)(crafted_used - 1) * sizeof(mp_digit));
+
+        key.p.dp    = fake_dp;
+        key.p.used  = crafted_used;
+        key.p.alloc = crafted_used;
+        key.p.sign  = 0;  /* MP_ZPOS */
+    }
+
+    /* Should return an error, not a bogus small size. */
+    derRet = wc_RsaKeyToDer(&key, NULL, 0);
```
dgarske, line 1255: **🔵 [Low] Unused variable derRet could use (void) cast or direct use in Expect**
*🔧 NIT*  `style`

The variable `derRet` is assigned from `wc_RsaKeyToDer` and immediately passed to `ExpectIntLT`. Some compilers may warn about the extra variable being unnecessary. The pattern could be simplified to avoid the intermediate variable, or the `derRet` value could be used in the `ExpectIntLT` call directly (which it already is). This is a very minor style point—the existing code is clear and readable.

**Recommendation:** No action required. The code is clear as-is. If the project convention is to avoid intermediate variables for single-use returns, consider `ExpectIntLT(wc_RsaKeyToDer(&key, NULL, 0), 0)` instead.

dgarske (commented):
## 🐺 Skoll Code Review

### Posted findings
- [Medium] **SetASN_Items MP path lacks matching bounds checks** — `wolfcrypt/src/asn.c:1097-1099`
- [Medium] **Test only covers USE_INTEGER_HEAP_MATH; tfm.c has same overflow pattern** — `tests/api/test_rsa.c:1161-1163`
- [Low] **Unused variable derRet could use (void) cast or direct use in Expect** — `tests/api/test_rsa.c:1254-1255`

wolfcrypt/src/asn.c
anhu: The suggestion removes the handling of ASN_DATA_TYPE_REPLACE_BUFFER .  This feels wrong.  Not aking the suggestion.

tests/api/test_rsa.c
anhu: Not going to bother with this.

tests/api/test_rsa.c
anhu: Will not leave revealing comment.

anhu:
jenkins retest this please

anhu:
jenkins retest this please

anhu:
Jenkins retest this please

anhu:
Good to go.  @wolfSSL-Bot , please sqash and merge at will.

douzzer (approved):

PR #9841, opened 2026-02-27
Re-enable hostap tests and remove some flaky tests

julek-wolfssl:
Set suites->setSuites to 1 in ParseCipherList function

PR #10052, opened 2026-03-23
20260323-InitSuites-ALL

douzzer:
`src/internal.c`: in `ParseCipherList()`, if the list is "ALL", add anonymous and null ciphers too.

fixes regression from 54d6555194 in #9841, detected by `scripts/openssl.test` in `super-quick-check`.

tested with
```
./configure --enable-all
WOLFSSL_OPENSSL_TEST=1  scripts/openssl.test
```

### DIFF
--- a/scripts/openssl.test
+++ b/scripts/openssl.test
@@ -369,7 +369,7 @@ do_wolfssl_client() {
 
     if [ "$client_result" != 0 ]
     then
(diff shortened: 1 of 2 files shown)


douzzer:
wrong fix
(see `-l ALL` in `scripts/openssl.test`)

PR #10070, opened 2026-03-25
Rust wrapper: add rand_core, aead, cipher trait implementations

holtrop-wolfssl:
# Description

This PR adds APIs which implement the traits in the rand_core, aead, and cipher crates.

This makes our crate more compatible with the rest of the Rust ecosystem.

# Testing

Unit tests

# Checklist

 - [X] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/wrapper/rust/wolfssl-wolfcrypt/src/chacha20_poly1305.rs
+++ b/wrapper/rust/wolfssl-wolfcrypt/src/chacha20_poly1305.rs
@@ -470,6 +470,7 @@ impl aead::AeadInPlace for XChaCha20Poly1305Aead {
         // single output buffer.  Use a stack buffer to hold both, then split
         // the tag out and copy the ciphertext back over the caller's buffer.
         const MAX_INLINE: usize = 4096;
+        debug_assert!(buffer.len() <= MAX_INLINE, "Maximum of 4096 bytes supported");
         if buffer.len() > MAX_INLINE {
             return Err(aead::Error);
         }
@@ -495,6 +496,7 @@ impl aead::AeadInPlace for XChaCha20Poly1305Aead {
         Ok(tag)
     }
 
+    // This function can decrypt a maximum of 4096 bytes.
     fn decrypt_in_place_detached(
         &self,
         nonce: &aead::Nonce<Self>,
@@ -505,6 +507,7 @@ impl aead::AeadInPlace for XChaCha20Poly1305Aead {
         // wc_XChaCha20Poly1305_Decrypt expects the auth tag appended after the
         // ciphertext.  Build a combined [ciphertext | tag] buffer on the stack.
         const MAX_INLINE: usize = 4096;
+        debug_assert!(buffer.len() <= MAX_INLINE, "Maximum of 4096 bytes supported");
         if buffer.len() > MAX_INLINE {
             return Err(aead::Error);
         }


wrapper/rust/wolfssl-wolfcrypt/src/chacha20_poly1305.rs
```diff
@@ -244,10 +243,95 @@ impl ChaCha20Poly1305 {
...
+        };
+        if rc != 0 {
+            return Err(aead::Error);
+        }
+        Ok(())
+    }
+}
+
+#[cfg(xchacha20_poly1305)]
+pub struct XChaCha20Poly1305;
```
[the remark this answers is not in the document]
holtrop-wolfssl: This is desired; no change needed.

wrapper/rust/wolfssl-wolfcrypt/src/random.rs
```diff
@@ -380,6 +380,34 @@ impl RNG {
...
+        rand_core::utils::next_word_via_fill(self)
+    }
+
+    fn try_next_u64(&mut self) -> Result<u64, Self::Error> {
+        rand_core::utils::next_word_via_fill(self)
+    }
+
+    fn try_fill_bytes(&mut self, dest: &mut [u8]) -> Result<(), Self::Error> {
+        self.generate_block(dest).expect("RNG failure");
+        Ok(())
```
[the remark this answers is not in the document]
holtrop-wolfssl: This is already documented; no change is needed.

wrapper/rust/wolfssl-wolfcrypt/src/aes.rs
```diff
@@ -2533,6 +2879,1011 @@ impl Drop for XTSStream {
...
+
+    fn unchecked_apply_keystream_inout(&mut self, mut buf: cipher::InOutBuf<'_, '_, u8>) {
+        let len = buf.len();
+        if len == 0 { return; }
+        let in_ptr = buf.get_in().as_ptr();
+        let out_ptr = buf.get_out().as_mut_ptr();
+        // SAFETY: CTR in-place is valid; raw ptrs used to avoid aliasing rules.
+        let in_slice = unsafe { core::slice::from_raw_parts(in_ptr, len) };
+        let out_slice = unsafe { core::slice::from_raw_parts_mut(out_ptr, len) };
+        self.inner.encrypt(in_slice, out_slice).expect("wc_AesCtrEncrypt failed");
```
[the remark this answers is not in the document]
holtrop-wolfssl: Fixed

wrapper/rust/wolfssl-wolfcrypt/src/chacha20_poly1305.rs
```diff
@@ -340,3 +424,107 @@ impl XChaCha20Poly1305 {
...
+        associated_data: &[u8],
+        buffer: &mut [u8],
+    ) -> Result<aead::Tag<Self>, aead::Error> {
+        // wc_XChaCha20Poly1305_Encrypt writes ciphertext + 16-byte tag into a
+        // single output buffer.  Use a stack buffer to hold both, then split
+        // the tag out and copy the ciphertext back over the caller's buffer.
+        const MAX_INLINE: usize = 4096;
+        if buffer.len() > MAX_INLINE {
+            return Err(aead::Error);
+        }
```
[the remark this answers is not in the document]
holtrop-wolfssl: Added comment to document the maximum buffer length.

wrapper/rust/wolfssl-wolfcrypt/src/aes.rs
```diff
@@ -2533,6 +2879,1011 @@ impl Drop for XTSStream {
...
+
+    fn unchecked_apply_keystream_inout(&mut self, mut buf: cipher::InOutBuf<'_, '_, u8>) {
+        let len = buf.len();
+        if len == 0 { return; }
+        let in_ptr = buf.get_in().as_ptr();
+        let out_ptr = buf.get_out().as_mut_ptr();
+        // SAFETY: OFB in-place is valid; raw ptrs used to avoid aliasing rules.
+        let in_slice = unsafe { core::slice::from_raw_parts(in_ptr, len) };
+        let out_slice = unsafe { core::slice::from_raw_parts_mut(out_ptr, len) };
+        self.inner.encrypt(in_slice, out_slice).expect("wc_AesOfbEncrypt failed");
```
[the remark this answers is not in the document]
holtrop-wolfssl: Fixed

wrapper/rust/wolfssl-wolfcrypt/src/aes.rs
```diff
@@ -2533,6 +2879,1011 @@ impl Drop for XTSStream {
...
```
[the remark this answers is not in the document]
holtrop-wolfssl: Fixed

wrapper/rust/wolfssl-wolfcrypt/src/aes.rs
```diff
@@ -2533,6 +2879,1011 @@ impl Drop for XTSStream {
...
+    fn unchecked_apply_keystream_inout(&mut self, mut buf: cipher::InOutBuf<'_, '_, u8>) {
+        let len = buf.len();
+        if len == 0 { return; }
+        // wolfCrypt AES-OFB supports in-place operation (out == in).
+        let in_ptr = buf.get_in().as_ptr();
+        let out_ptr = buf.get_out().as_mut_ptr();
+        // SAFETY: OFB in-place is valid; raw ptrs used to avoid aliasing rules.
+        let in_slice = unsafe { core::slice::from_raw_parts(in_ptr, len) };
+        let out_slice = unsafe { core::slice::from_raw_parts_mut(out_ptr, len) };
+        self.inner.encrypt(in_slice, out_slice).expect("wc_AesOfbEncrypt failed");
```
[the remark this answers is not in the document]
holtrop-wolfssl: Fixed

wrapper/rust/wolfssl-wolfcrypt/src/aes.rs
```diff
@@ -2533,6 +2879,1011 @@ impl Drop for XTSStream {
...
+
+    fn unchecked_apply_keystream_inout(&mut self, mut buf: cipher::InOutBuf<'_, '_, u8>) {
+        let len = buf.len();
+        if len == 0 { return; }
+        let in_ptr = buf.get_in().as_ptr();
+        let out_ptr = buf.get_out().as_mut_ptr();
+        // SAFETY: CTR in-place is valid; raw ptrs used to avoid aliasing rules.
+        let in_slice = unsafe { core::slice::from_raw_parts(in_ptr, len) };
+        let out_slice = unsafe { core::slice::from_raw_parts_mut(out_ptr, len) };
+        self.inner.encrypt(in_slice, out_slice).expect("wc_AesCtrEncrypt failed");
```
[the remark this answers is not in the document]
holtrop-wolfssl: Fixed

wrapper/rust/wolfssl-wolfcrypt/src/aes.rs
```diff
@@ -2533,6 +2879,1011 @@ impl Drop for XTSStream {
...
+    fn unchecked_apply_keystream_inout(&mut self, mut buf: cipher::InOutBuf<'_, '_, u8>) {
+        let len = buf.len();
+        if len == 0 { return; }
+        // wolfCrypt AES-CTR supports in-place operation (out == in).
+        let in_ptr = buf.get_in().as_ptr();
+        let out_ptr = buf.get_out().as_mut_ptr();
+        // SAFETY: CTR in-place is valid; raw ptrs used to avoid aliasing rules.
+        let in_slice = unsafe { core::slice::from_raw_parts(in_ptr, len) };
+        let out_slice = unsafe { core::slice::from_raw_parts_mut(out_ptr, len) };
+        self.inner.encrypt(in_slice, out_slice).expect("wc_AesCtrEncrypt failed");
```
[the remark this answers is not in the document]
holtrop-wolfssl: Fixed

holtrop-wolfssl:
retest this please (Calling Pipeline was cancelled)

wrapper/rust/wolfssl-wolfcrypt/src/aes.rs
```diff
@@ -1394,6 +1576,170 @@ impl Drop for GCM {
...
+    if rc != 0 {
+        return Err(aead::Error);
+    }
+    Ok(())
+}
+
+/// AES-128-GCM authenticated encryption (12-byte nonce, 16-byte tag).
+#[cfg(all(aes_gcm, feature = "aead"))]
+pub struct Aes128Gcm {
+    key: [u8; 16],
```
gasbytes, line 1644: a thing that I noticed is that the key material is stored plain inside a bytes array, but there's a missing implementation of the drop trait to zeroize it.
when out of scope, technically these struct remain in memory until type gets reused in another way.
this pretty much applies to all the new aead structs (aes*, chacha etc...)

i would either add a zeroize dependency or just implement the missing drop trait that zeroes the key bytes when exiting the scope after usage.

wrapper/rust/wolfssl-wolfcrypt/src/chacha20_poly1305.rs
```diff
@@ -340,3 +424,108 @@ impl XChaCha20Poly1305 {
...
+        nonce: &aead::Nonce<Self>,
+        associated_data: &[u8],
+        buffer: &mut [u8],
+    ) -> Result<aead::Tag<Self>, aead::Error> {
+        // wc_XChaCha20Poly1305_Encrypt writes ciphertext + 16-byte tag into a
+        // single output buffer.  Use a stack buffer to hold both, then split
+        // the tag out and copy the ciphertext back over the caller's buffer.
+        const MAX_INLINE: usize = 4096;
+        if buffer.len() > MAX_INLINE {
+            return Err(aead::Error);
```
gasbytes, line 475: This is not a blocker but let me know what you think.
I see that this is documented already saying that if a buffer bigger than 4096 gets passed Err(aead::error) gets returned.
But maybe adding an println when returning the error thattt would make the error less opaque (?).
I was thinking maybe a debugg_assert! to let know the user at runtime that they exceeded the limit in the buffer length.

wrapper/rust/wolfssl-wolfcrypt/src/aes.rs
holtrop-wolfssl: There was actually a Fenrir finding about this as well. I looked into the zeroize crate and it looks like it supports no_std and has both a ZeroizeOnDrop derive as well as supporting manually calling .zeroize() if a custom Drop impl is needed. I plan to add that to most structs.

Since I already have some follow-on changes for Fenrir after this PR, are you ok with me making this change for the next PR so that it doesn't get too messy with changes that have already been done after this? I can create a github issue to track it if you would prefer.

gasbytes: Yeah that makes sense sure, no need to create a separate github issue since you are already addressing it in the next PR.
I'll mark this as resolved.

wrapper/rust/wolfssl-wolfcrypt/src/chacha20_poly1305.rs
holtrop-wolfssl: Yeah, good idea. I don't have println! since I'm no_std, but I added a debug_assert!

wrapper/rust/wolfssl-wolfcrypt/src/aes.rs
```diff
@@ -2533,6 +2879,1011 @@ impl Drop for XTSStream {
...
+    fn new(key: &cipher::Key<Self>, iv: &cipher::Iv<Self>) -> Self {
+        let mut ctr = CTR::new().expect("wc_AesInit failed");
+        ctr.init(key.as_ref(), iv.as_ref()).expect("wc_AesSetKeyDirect failed");
+        Self { inner: ctr }
+    }
+}
+
+#[cfg(all(aes_ctr, feature = "cipher"))]
+impl StreamCipher for Aes128Ctr {
+    fn check_remaining(&self, _data_len: usize) -> Result<(), StreamCipherError> {
```
dgarske, line 3250: **🟠 [High] `check_remaining` no-op allows silent `len as u32` truncation in StreamCipher impls**
*🚫 BLOCK*  `bug`

All six `StreamCipher::check_remaining` implementations (Aes128Ctr, Aes192Ctr, Aes256Ctr, Aes128Ofb, Aes192Ofb, Aes256Ofb) unconditionally return `Ok(())` without validating the input size. The `unchecked_apply_keystream_inout` then casts `len as u32` for the wolfCrypt C API which takes `word32`. On 64-bit platforms, if `data_len > u32::MAX`, the cast silently truncates. Since the safe `apply_keystream` wrapper calls `check_remaining` first and trusts its result, a caller could pass a large buffer and only the lower 32 bits of bytes would be processed — leaving the rest as plaintext. This is a data integrity and confidentiality issue.

**Suggestion:**
```suggestion
fn check_remaining(&self, data_len: usize) -> Result<(), StreamCipherError> {
    if data_len > u32::MAX as usize {
        return Err(StreamCipherError);
    }
    Ok(())
}
```

wrapper/rust/wolfssl-wolfcrypt/src/aes.rs
```diff
@@ -414,6 +433,169 @@ impl Drop for CCM {
...
+    };
+    if rc != 0 {
+        return Err(aead::Error);
+    }
+    Ok(())
+}
+
+/// AES-128-CCM authenticated encryption (12-byte nonce, 16-byte tag).
+#[cfg(all(aes_ccm, feature = "aead"))]
+pub struct Aes128Ccm {
```
dgarske, line 499: **🟡 [Medium] AEAD key wrapper structs do not zeroize key material on Drop**
*💡 SUGGEST*  `api`

Six new structs store raw cryptographic key material in plain `[u8; N]` arrays: `Aes128Ccm.key`, `Aes256Ccm.key`, `Aes128Gcm.key`, `Aes256Gcm.key`, `ChaCha20Poly1305Aead.key`, and `XChaCha20Poly1305Aead.key`. None implement `Drop` to zeroize the key bytes. When these structs go out of scope, the key material persists in memory until overwritten. For a cryptographic wrapper crate, this is a notable gap — the `zeroize` crate is the standard Rust solution, and the companion crates (e.g., `aes-gcm`) implement `ZeroizeOnDrop`.

**Suggestion:**
```suggestion
impl Drop for Aes128Gcm {
    fn drop(&mut self) {
        self.key.iter_mut().for_each(|b| unsafe {
            core::ptr::write_volatile(b, 0);
        });
    }
}
```

wrapper/rust/wolfssl-wolfcrypt/src/aes.rs
```diff
@@ -1394,6 +1576,170 @@ impl Drop for GCM {
...
 
+// ---------------------------------------------------------------------------
+// AES-GCM aead trait implementations
+// ---------------------------------------------------------------------------
+
+/// Encrypt `buffer` in-place using AES-GCM (12-byte nonce, 16-byte tag).
+///
+/// wolfCrypt's `wc_AesGcmEncrypt` supports in-place operation (out == in).
+#[cfg(all(aes_gcm, feature = "aead"))]
+fn gcm_encrypt_in_place(
```
dgarske, line 1587: **🟡 [Medium] GCM/CCM AEAD wrappers re-initialize crypto context on every operation**
*💡 SUGGEST*  `api`

Each call to `gcm_encrypt_in_place` (and the CCM equivalent) creates a new `GCM`/`CCM` instance via `GCM::new()`, calls `gcm.init(key)`, performs the operation, then drops the context. This means every encrypt/decrypt call pays the full cost of `wc_AesInit` + `wc_AesGcmSetKey` + `wc_AesFree`. The key is already stored in the wrapper struct, but the `Aes` context is not preserved between calls. Other AEAD crates (e.g., `aes-gcm`) compute the key schedule once at construction time and reuse it. This could significantly impact performance for applications doing many small encryptions.

**Suggestion:**
```suggestion
Store a pre-initialized `GCM`/`CCM` context in the struct instead of the raw key bytes. Initialize it once in `KeyInit::new` and reuse it in every encrypt/decrypt call.
```

wrapper/rust/wolfssl-wolfcrypt/src/random.rs
```diff
@@ -380,6 +380,34 @@ impl RNG {
...
+
+    fn try_next_u32(&mut self) -> Result<u32, Self::Error> {
+        rand_core::utils::next_word_via_fill(self)
+    }
+
+    fn try_next_u64(&mut self) -> Result<u64, Self::Error> {
+        rand_core::utils::next_word_via_fill(self)
+    }
+
+    fn try_fill_bytes(&mut self, dest: &mut [u8]) -> Result<(), Self::Error> {
```
dgarske, line 401: **🟡 [Medium] `TryRng::try_fill_bytes` panics on wolfCrypt RNG failure instead of returning error**
*💡 SUGGEST*  `api`

The `TryRng` implementation uses `Error = Infallible`, which means `try_fill_bytes` cannot return an error. Instead, any wolfCrypt RNG failure causes a panic via `.expect("RNG failure")`. While this is documented in the doc comment and allows the blanket `Rng + CryptoRng` impls to kick in, it breaks the expectation of `try_fill_bytes` — callers using the `TryRng` trait specifically expect fallible behavior. The `rand_core 0.10` crate provides `rand_core::OsError` and custom error types that could propagate the wolfCrypt error code instead of panicking.

**Suggestion:**
```suggestion
Consider defining a custom error type that wraps the wolfCrypt error code and using it instead of `Infallible`. Then implement `Rng` and `CryptoRng` manually (delegating to the try_ methods with `.unwrap()`) to preserve the ergonomic non-fallible API.
```

wrapper/rust/wolfssl-wolfcrypt/src/aes.rs
```diff
@@ -2533,6 +2879,1011 @@ impl Drop for XTSStream {
     }
 }
 
+// ---------------------------------------------------------------------------
```
dgarske, line 2882: **🔵 [Low] ~1000 lines of near-identical boilerplate for AES key-size variants**
*🔧 NIT*  `style`

The cipher trait implementations for ECB (enc/dec), CTR, OFB, and CBC (enc/dec) repeat nearly identical code three times each for AES-128, AES-192, and AES-256. The only differences between variants are the key size type (`U16`/`U24`/`U32`), struct name, and inner type. This amounts to ~1000 lines of duplicated code. A declarative macro (`macro_rules!`) could reduce this to ~100 lines while being just as readable.

**Suggestion:**
```suggestion
macro_rules! impl_aes_ecb_enc {
    ($name:ident, $key_size:ty) => {
        pub struct $name { inner: ECB }
        impl cipher::KeySizeUser for $name { type KeySize = $key_size; }
        impl cipher::BlockSizeUser for $name { type BlockSize = U16; }
        // ... rest of impls
    }
}
impl_aes_ecb_enc!(Aes128EcbEnc, U16);
impl_aes_ecb_enc!(Aes192EcbEnc, U24);
impl_aes_ecb_enc!(Aes256EcbEnc, U32);
```

wrapper/rust/wolfssl-wolfcrypt/src/aes.rs
```diff
@@ -1394,6 +1576,170 @@ impl Drop for GCM {
...
+    tag: &mut [u8],
+) -> Result<(), aead::Error> {
+    let mut gcm = GCM::new().map_err(|_| aead::Error)?;
+    gcm.init(key).map_err(|_| aead::Error)?;
+    let buf_ptr = buffer.as_mut_ptr();
+    let in_ptr = buf_ptr as *const u8;
+    let rc = unsafe {
+        sys::wc_AesGcmEncrypt(
+            &mut gcm.ws_aes,
+            buf_ptr, in_ptr, buffer.len() as u32,
```
dgarske, line 1601: **🔵 [Low] AEAD GCM/CCM `buffer.len() as u32` truncates silently for buffers > 4GB**
*🔧 NIT*  `bug`

The helper functions `gcm_encrypt_in_place` and `ccm_encrypt_in_place` cast `buffer.len() as u32` and `aad.len() as u32` for the wolfCrypt C API. On 64-bit platforms, buffers larger than `u32::MAX` (4GB) would be silently truncated. While this is extremely unlikely in practice for AEAD operations, it's inconsistent with the `aead::AeadInPlace` trait contract which doesn't impose size limits. The same issue exists in `ChaCha20Poly1305Aead` (`associated_data.len() as u32`, `buffer.len() as u32`).

**Suggestion:**
```suggestion
Add a guard at the top of the helper functions:
if buffer.len() > u32::MAX as usize || aad.len() > u32::MAX as usize {
    return Err(aead::Error);
}
```

wrapper/rust/wolfssl-wolfcrypt/src/aes.rs
holtrop-wolfssl: Hi @dgarske , this PR is quite old now and most of these issues you've identified here have been fixed in a "rust-fenrir-fixes" branch here: https://github.com/holtrop-wolfssl/wolfssl/tree/rust-fenrir-fixes

Originally I was waiting for this PR to merge before creating a PR containing that branch since they were separate concepts. If you'd prefer for reviewing I can include that entire branch of fixes in this PR but the scope will be both these new features as well as many fixes.

wrapper/rust/wolfssl-wolfcrypt/src/aes.rs
holtrop-wolfssl: Zeroize functionality is also performed in my rust-fenrir-fixes branch which is waiting to enter PR.

PR #10072, opened 2026-03-25
Extend/Add label/id extension functions

night1rider:
## Summary

- Add `wc_InitCmac_Id()` / `wc_InitCmac_Label()` so CMAC can forward a hardware key-slot identifier to the underlying AES context.

- Add `wc_AesNew_Id()`/ `wc_AesNew_Label()` and `wc_NewRsaKey_Id()`/`wc_NewRsaKey_Label()` constructors following the same pattern.
- Refactor each algorithm's init into a single internal `_common` function to avoid duplication. Public API is fully backward-compatible.
- Add per-algorithm init helpers in `test.c` and `benchmark.c` that switch between plain and Id-based init when `WC_TEST_<ALG>_ID` macros are defined.
- Add test skip macros: `WC_TEST_SKIP_ECC_CHECK_KEY`, `WC_TEST_SKIP_RSA_PRIVATE_EXPORT`, `WC_TEST_SKIP_ECC_PRIVATE_EXPORT`, `WC_TEST_SKIP_ZERO_LEN_CMAC`.

### DIFF
--- a/wolfcrypt/src/cmac.c
+++ b/wolfcrypt/src/cmac.c
@@ -162,8 +162,9 @@ static int _InitCmac_common(Cmac* cmac, const byte* key, word32 keySz,
         if (ret != WC_NO_ERR_TRACE(CRYPTOCB_UNAVAILABLE)) {
             return ret;
         }
-        /* fall-through when unavailable */
+        /* fall-through when unavailable, reset ret for software path */
         ret = 0;
+        (void)ret;
     }
 #else
     (void)devId;


wolfcrypt/test/test.c
```diff
@@ -34648,10 +34902,12 @@ static wc_test_ret_t ecc_test_curve_size(WC_RNG* rng, int keySize, int testVerif
             ERROR_OUT(WC_TEST_RET_ENC_NC, done);
     }
 
+#ifndef WC_TEST_NO_ECC_CHECK_KEY
```
bigbrett: nit: naming is inconsistent with other gating macros you introduced ("skip").

bigbrett (dismissed):
Kinda hate the test/bench wrappers, but not sure what can be done about that while keeping APIs consistent, unless perhaps we just called the new variants of the functions directly and allowed them to default back to the old Init if `WOLF_PRIVATE_KEY_ID` wasn't defined? Might declutter. Food for thought.

OK as is.

wolfcrypt/test/test.c
night1rider: Updated to match

night1rider:
Jenkins retest this please

night1rider:
Rebased and force pushed

wolfcrypt/src/cmac.c
```diff
@@ -130,23 +158,55 @@ int wc_InitCmac_ex(Cmac* cmac, const byte* key, word32 keySz,
 
         ret = wc_CryptoCb_Cmac(cmac, key, keySz, NULL, 0, NULL, NULL,
                 type, unused);
-        if (ret != WC_NO_ERR_TRACE(CRYPTOCB_UNAVAILABLE))
+        if (ret != WC_NO_ERR_TRACE(CRYPTOCB_UNAVAILABLE)) {
             return ret;
-        /* fall-through when unavailable */
+        }
+        /* fall-through when unavailable, reset ret for software path */
```
dgarske, line 165: Prefer we reset 'ret = 0;` here to solve any possible future issues, but if the static analyzer is complaining then perhaps its okay. 

### MEDIUM-1: `_InitCmac_common` misleading comment — ret NOT reset after CRYPTOCB_UNAVAILABLE
- **File:** `wolfcrypt/src/cmac.c:164`
- **Function:** `_InitCmac_common`
- **Category:** bug
- **Confidence:** High

**Description:** The diff changed the comment from `"fall-through when unavailable"` to `"fall-through when unavailable, reset ret for software path"`. However, `ret` is **not** actually reset to 0 after `wc_CryptoCb_Cmac` returns `CRYPTOCB_UNAVAILABLE`. At this point `ret` holds the value `CRYPTOCB_UNAVAILABLE`. While this doesn't cause a functional bug today because the subsequent `switch` statement at line 179 unconditionally overwrites `ret` in all branches, the misleading comment could cause future maintainers to rely on `ret == 0` after this block, introducing subtle bugs.

**Code:**
```c
        ret = wc_CryptoCb_Cmac(cmac, key, keySz, NULL, 0, NULL, NULL,
                type, unused);
            return ret;
        }
        /* fall-through when unavailable, reset ret for software path */
    }
```

**Recommendation:** Either reset `ret = 0;` after the check (matching what `wc_AesCmacGenerate_ex` does at line 490), or fix the comment to remove the "reset ret" claim.

```suggestion
        if (ret != WC_NO_ERR_TRACE(CRYPTOCB_UNAVAILABLE)) {
            return ret;
        }
        /* fall-through when unavailable */
        ret = 0;
    }
```

wolfcrypt/test/test.c
```diff
@@ -25370,11 +25656,11 @@ WOLFSSL_TEST_SUBROUTINE wc_test_ret_t rsa_test(void)
     XMEMCPY(in, inStr, inLen);
 
 #if defined(WOLFSSL_SMALL_STACK) && !defined(WOLFSSL_NO_MALLOC)
-    key = wc_NewRsaKey(HEAP_HINT, devId, &ret);
+    key = test_RsaNew(HEAP_HINT, devId, &ret);
     if (key == NULL)
         ERROR_OUT(WC_TEST_RET_ENC_EC(ret), exit_rsa);
 #if defined(WOLFSSL_CERT_EXT) || defined(WOLFSSL_CERT_GEN)
-    keypub = wc_NewRsaKey(HEAP_HINT, devId, &ret);
+    keypub = test_RsaNew(HEAP_HINT, devId, &ret);
```
dgarske: ## MEDIUM-2: `test_RsaInit` / `test_RsaNew` use private key ID for public-only key (`keypub`)
- **File:** `wolfcrypt/test/test.c:25663,25678`
- **Function:** `rsa_test`
- **Category:** bug
- **Confidence:** Medium

**Description:** When `WC_TEST_RSA_PUB_ID` is defined, there's a static `testRsaPubId` array (line ~631), but no `test_RsaInit_Pub` or `test_RsaNew_Pub` helper. Both `test_RsaInit` and `test_RsaNew` unconditionally use `testRsaPrivId` when `WC_TEST_RSA_PRIV_ID` is defined. This means `keypub` (used only for public-key certificate operations like `wc_SetSubjectKeyIdFromPublicKey`) is initialized with the **private** key ID. If the hardware backend enforces key-type separation, this will cause test failures or silently use the wrong key slot.

**Code:**
```c
// Line 25663 - keypub gets private key ID
keypub = test_RsaNew(HEAP_HINT, devId, &ret);
// Line 25678 - same issue with stack allocation path
ret = test_RsaInit(keypub, HEAP_HINT, devId);
```

**Recommendation:** Add a `test_RsaInit_Pub` / `test_RsaNew_Pub` helper that uses `testRsaPubId` when `WC_TEST_RSA_PUB_ID` is defined, and use it for `keypub` initialization.

wolfcrypt/src/cmac.c
```diff
@@ -97,18 +97,23 @@ void ShiftAndXorRb(byte* out, byte* in)
...
-int wc_InitCmac_ex(Cmac* cmac, const byte* key, word32 keySz,
-                int type, void* unused, void* heap, int devId)
+#define CMAC_AES_INIT_PLAIN  0
+#ifdef WOLF_PRIVATE_KEY_ID
+#define CMAC_AES_INIT_ID     1
+#define CMAC_AES_INIT_LABEL  2
+#endif
+
+
+static int _InitCmac_common(Cmac* cmac, const byte* key, word32 keySz,
```
dgarske, line 107: ### MEDIUM-3: Removed `(void)heap;` cast in `_InitCmac_common` may cause compiler warning
- **File:** `wolfcrypt/src/cmac.c:107-247`
- **Function:** `_InitCmac_common`
- **Category:** convention
- **Confidence:** Medium

**Description:** The original `wc_InitCmac_ex` had `(void)heap;` to suppress unused-variable warnings. The refactored `_InitCmac_common` removed this cast. While `heap` is used inside the `WC_CMAC_AES` case of the switch, if a hypothetical build configuration excluded the AES direct path (`!defined(WOLFSSL_AES_DIRECT)`), `heap` would be unused, generating a `-Wunused-parameter` warning. This is fragile even though current build configurations always include AES.

**Code (from diff):**
```diff
-    (void)unused;
-    (void)heap;
...
+    (void)unused;
```

**Recommendation:** Re-add `(void)heap;` for warning suppression robustness, consistent with the existing `(void)unused;` pattern.

wolfcrypt/src/cmac.c
```diff
@@ -119,6 +124,29 @@ int wc_InitCmac_ex(Cmac* cmac, const byte* key, word32 keySz,
 #endif
     XMEMSET(cmac, 0, sizeof(Cmac));
 
+    /* Store id/label on the Cmac struct so the crypto callback can
+     * inspect them to determine the hardware key slot. */
+#ifdef WOLF_PRIVATE_KEY_ID
+    cmac->aesInitType = aesInitType;
+    if (aesInitType == CMAC_AES_INIT_ID && id != NULL &&
```
dgarske: ### LOW-1: CMAC id/label storage silently truncates while AesInit_Id receives full value
- **File:** `wolfcrypt/src/cmac.c:131-143`
- **Function:** `_InitCmac_common`
- **Category:** bug
- **Confidence:** Low

**Description:** When `aesInitType == CMAC_AES_INIT_ID` and `idLen > sizeof(cmac->id)` (i.e., > `AES_MAX_ID_LEN` = 32), the id is **not** stored in `cmac->id` (the copy is skipped), but later at line 186, `wc_AesInit_Id(&cmac->aes, id, idLen, heap, devId)` is still called with the full-length id. This creates an inconsistency: the crypto callback inspecting `cmac->id` sees an empty array, while the underlying AES object has the real id. In practice this is unlikely since ids are typically short, but the silent truncation with no error return could mask configuration issues.

**Code:**
```c
    if (aesInitType == CMAC_AES_INIT_ID && id != NULL &&
            idLen > 0 &&
            idLen <= (int)sizeof(cmac->id)) {  // skipped if idLen > 32
        XMEMCPY(cmac->id, id, (word32)idLen);
        cmac->idLen = idLen;
    }
    // ... later:
    case CMAC_AES_INIT_ID:
        ret = wc_AesInit_Id(&cmac->aes, id, idLen, heap, devId); // uses original idLen
```

**Recommendation:** Return `BAD_FUNC_ARG` when `idLen > sizeof(cmac->id)` instead of silently skipping the copy, or document that the stored copy is best-effort.

wolfcrypt/benchmark/benchmark.c
```diff
@@ -4993,6 +4993,145 @@ void bench_rng(void)
 #endif /* WC_NO_RNG */
 
 
+/* ============================================================================
```
dgarske, line 4996: ### LOW-2: Significant code duplication between benchmark.c and test.c init helpers
- **File:** `wolfcrypt/benchmark/benchmark.c:4996-5132`, `wolfcrypt/test/test.c:558-734`
- **Function:** Multiple (`bench_AesCbcInit`, `test_AesCbcInit`, etc.)
- **Category:** convention
- **Confidence:** High

**Description:** The benchmark and test helper functions (`bench_AesCbcInit` / `test_AesCbcInit`, `bench_RsaInit` / `test_RsaInit`, `bench_CmacInit` / `test_CmacInit`, `bench_EccInit_Pair` / `test_EccInit_Pair`) are nearly identical — differing only in the static array variable names (`benchAesCbcId` vs `testAesCbcId`). This is ~300 lines of duplicated logic across two files. Future changes to key-id init patterns will need to be applied twice.

**Recommendation:** Consider extracting common init-helper logic into a shared header or a common source file used by both benchmark and test code. Alternatively, accept the duplication since these files have separate build targets and the pattern is stable.

wolfcrypt/src/cmac.c
night1rider: Fixed comment and added `ret = 0;` and `(void)ret;` after the CRYPTOCB_UNAVAILABLE check,
matching the pattern in `wc_AesCmacGenerate_ex` and to show the intentional to CI checkers.

wolfcrypt/test/test.c
night1rider: Added `test_RsaInit_Pub` and `test_RsaNew_Pub` helpers that use `testRsaPubId` when `WC_TEST_RSA_PUB_ID` is defined, and updated `keypub` initialization to use them.

wolfcrypt/src/cmac.c
night1rider: Re-added `(void)heap;` to match the existing `(void)unused;` pattern.

wolfcrypt/src/cmac.c
night1rider: Now returns `BAD_FUNC_ARG` when `idLen > sizeof(cmac->id)` instead of silently skipping the copy.

wolfcrypt/benchmark/benchmark.c
night1rider: Looking to address this with a better solution in a follow up PR. Duplication would require a shared source and header between the benchmarks/test code.

night1rider:
Had to rebase and force push due to merge conflict with .github/workflows/os-check.yml.

PR #10073, opened 2026-03-25
Ensure certificates are getting into cert manager

anhu:
Fixes ZD 19760

### DIFF
--- a/src/ssl_api_cert.c
+++ b/src/ssl_api_cert.c
@@ -1536,6 +1536,14 @@ void wolfSSL_CTX_set_cert_store(WOLFSSL_CTX* ctx, WOLFSSL_X509_STORE* str)
         ctx->x509_store_pt    = str;
         /* Context has ownership and free it with context free. */
         ctx->cm->x509_store_p = ctx->x509_store_pt;
+
+#ifdef OPENSSL_EXTRA
+        /* Non-self-signed certs (intermediates) added via
+         * X509_STORE_add_cert only go into store->certs, not the
+         * CertManager. Push them into the CM now so that all
+         * verification paths can find them. */
+        X509StorePushCertsToCM(str);
+#endif
     }
 }
 
--- a/src/ssl_api_cert.c
+++ b/src/ssl_api_cert.c
@@ -1542,7 +1542,10 @@ void wolfSSL_CTX_set_cert_store(WOLFSSL_CTX* ctx, WOLFSSL_X509_STORE* str)
          * X509_STORE_add_cert only go into store->certs, not the
          * CertManager. Push them into the CM now so that all
          * verification paths can find them. */
-        X509StorePushCertsToCM(str);
+        if (X509StorePushCertsToCM(str) != WOLFSSL_SUCCESS) {
+            WOLFSSL_MSG("wolfSSL_CTX_set_cert_store: failed to push some "
+                        "certs to CertManager");
+        }
 #endif
     }
 }
--- a/tests/api/test_ossl_x509_str.c
+++ b/tests/api/test_ossl_x509_str.c
@@ -1853,7 +1853,7 @@ int test_wolfSSL_CTX_set_cert_store(void)
 {
     EXPECT_DECLS;
 #if defined(OPENSSL_EXTRA) && !defined(NO_RSA) && !defined(NO_FILESYSTEM) && \
-    !defined(NO_WOLFSSL_CLIENT)
+    !defined(NO_WOLFSSL_CLIENT) && !defined(NO_TLS)
     SSL_CTX* ctx = NULL;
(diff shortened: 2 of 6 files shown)


anhu:
This replaces https://github.com/wolfSSL/wolfssl/pull/8708

tests/api/test_ossl_x509_str.c
```diff
@@ -1074,56 +1074,6 @@ int test_X509_STORE_InvalidCa(void)
     ExpectIntEQ(X509_STORE_CTX_init(ctx, str, cert, untrusted), 1);
     ExpectIntEQ(X509_verify_cert(ctx), 1);
     ExpectIntEQ(last_errcode, X509_V_ERR_INVALID_CA);
```
dgarske, line 1076: **🟡 [Medium] Defense-in-depth assertion removed from test_X509_STORE_InvalidCa without explanation**
*💡 SUGGEST*  `test`

The PR removes the assertion `ExpectIntEQ(X509_STORE_CTX_get_error(ctx), X509_V_ERR_INVALID_CA)` and its accompanying comment: "Defense in depth: ctx->error must not be clobbered back to X509_V_OK by the later successful verification of the intermediate against the trusted root. The worst-seen error must persist." This assertion was specifically guarding against error-clobbering regressions. Its removal, along with the complete removal of `test_X509_STORE_InvalidCa_NoCallback`, reduces test coverage for INVALID_CA error handling. If these were removed because the PR's changes to cert propagation altered the verification behavior (e.g., the error no longer persists), that behavioral change should be documented and potentially re-tested differently.

**Suggestion:**
```suggestion
If the behavior change is intentional (i.e., the error is no longer expected to persist because the cert now goes through a different path), add a comment explaining why. If it's not intentional, restore the assertion. Consider adding an equivalent test to verify the new behavior.
```

src/ssl_api_cert.c
```diff
@@ -1536,6 +1536,14 @@ void wolfSSL_CTX_set_cert_store(WOLFSSL_CTX* ctx, WOLFSSL_X509_STORE* str)
         ctx->x509_store_pt    = str;
         /* Context has ownership and free it with context free. */
         ctx->cm->x509_store_p = ctx->x509_store_pt;
+
+#ifdef OPENSSL_EXTRA
+        /* Non-self-signed certs (intermediates) added via
+         * X509_STORE_add_cert only go into store->certs, not the
+         * CertManager. Push them into the CM now so that all
+         * verification paths can find them. */
+        X509StorePushCertsToCM(str);
```
dgarske, line 1545: **🟡 [Medium] Return value of X509StorePushCertsToCM silently ignored**
*💡 SUGGEST*  `convention`

`X509StorePushCertsToCM(str)` returns `WOLFSSL_FATAL_ERROR` if any certificate fails to be added to the CertManager, but the return value is completely discarded. While `wolfSSL_CTX_set_cert_store` is a `void` function (matching the OpenSSL API), silently ignoring failures could mask problems — the caller would believe all certificates were pushed into the CM when some may have been silently dropped. At minimum, a `WOLFSSL_MSG` warning on failure would help with debugging.

**Suggestion:**
```suggestion
```suggestion
#ifdef OPENSSL_EXTRA
        /* Non-self-signed certs (intermediates) added via
         * X509_STORE_add_cert only go into store->certs, not the
         * CertManager. Push them into the CM now so that all
         * verification paths can find them. */
        if (X509StorePushCertsToCM(str) != WOLFSSL_SUCCESS) {
            WOLFSSL_MSG("wolfSSL_CTX_set_cert_store: failed to push some "
                        "certs to CertManager");
        }
#endif
```
```

dgarske (changes requested):
## 🐺 Skoll Code Review

**Findings:** 3 total — 3 posted, 1 skipped

### Posted findings
- [Critical] **test_X509_STORE_InvalidCa_NoCallback removed from .c but still declared and referenced in .h — linker error** — `tests/api/test_ossl_x509_str.h:34,56`
- [Medium] **Defense-in-depth assertion removed from test_X509_STORE_InvalidCa without explanation** — `tests/api/test_ossl_x509_str.c:1076-1077`
- [Medium] **Return value of X509StorePushCertsToCM silently ignored** — `src/ssl_api_cert.c:1545`

- [Critical] test_X509_STORE_InvalidCa_NoCallback removed from .c but still declared and referenced in .h — linker error

</details>

tests/api/test_ossl_x509_str.c
anhu: fixed

anhu:
jenkins retest this please.

src/ssl_api_cert.c
anhu: I took the suggestion.

dgarske (approved):

PR #10074, opened 2026-03-25
Various security fixes and tests

anhu:
Fixes: 
zd21412
zd21413
zd21414
zd21415
zd21417
zd21418
zd21421
zd21422
zd21423
zd21426

### DIFF
--- a/wolfcrypt/src/ecc.c
+++ b/wolfcrypt/src/ecc.c
@@ -11005,8 +11005,8 @@ int wc_ecc_import_x963_ex2(const byte* in, word32 inLen, ecc_key* key,
     if (err == MP_OKAY)
         err = wc_ecc_check_key(key);
 #else
-    /* Always validate ECC imports to prevent invalid curve attacks */
-    if (err == MP_OKAY)
+    /* Validate untrusted ECC imports to prevent invalid curve attacks */
+    if ((err == MP_OKAY) && untrusted)
         err = wc_ecc_check_key(key);
 #endif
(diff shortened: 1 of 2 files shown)


anhu:
TODO: Remove references to ZD in code and tests.  They are left there to help the reviewers.  Remove once reviewers are ok with it.

wolfcrypt/src/ecc.c
```diff
@@ -11004,6 +11004,10 @@ int wc_ecc_import_x963_ex2(const byte* in, word32 inLen, ecc_key* key,
 #ifdef WOLFSSL_VALIDATE_ECC_IMPORT
     if (err == MP_OKAY)
         err = wc_ecc_check_key(key);
+#else
+    /* Always validate ECC imports to prevent invalid curve attacks */
+    if (err == MP_OKAY)
```
Frauschi: if and else blocks for the macro guard are identical now. So we can simply remove the guard

anhu: Good catch.  But we've been through this before.  we're not going to validate everything because its wasteful.  I think I need to remove this.  ...scratch that.  We need to do this. Sigh.

anhu:
Jenkins retest this please

>./wolfssl/options.h:10: error: unterminated #elif
   10 | #ifdef WOLFSSL_NO_OPTIONS_H

Seems bogus.

anhu:
Closing in favour of https://github.com/wolfSSL/wolfssl/pull/10088 which has better division of fixes.

PR #10077, opened 2026-03-26
Add ML-KEM support for PKCS#11

Frauschi:
## PKCS#11: Add ML-KEM support
                                                                         
Adds PKCS#11 integration for ML-KEM (key generation, encapsulation, decapsulation) through the crypto callback path, using the PKCS#11 3.2 `C_EncapsulateKey`/`C_DecapsulateKey` interface.              
                                                                         
### What's included:                                          

  - New PKCS#11 constants for ML-KEM: `CKK_ML_KEM`, `CKM_ML_KEM`, `CKM_ML_KEM_KEY_PAIR_GEN`, `CKA_ENCAPSULATE`/`CKA_DECAPSULATE`, `CKP_ML_KEM_{512,768,1024} `                                            
  - Key generation via `C_GenerateKeyPair`; public key bytes retrieved into the wolfSSL key object, private key handle kept in `devCtx` for direct decapsulation
  - Encapsulation and decapsulation dispatch through `wc_CryptoCb_PqcEncapsulate`/`Decapsulate` in wc_mlkem.c                   
  - `wc_MlKemKey_Init_Id` / `wc_MlKemKey_Init_Label` for private-key lookup by token ID or label                                                   
  - `wc_Pkcs11StoreKey` support for `PKCS11_KEY_TYPE_MLKEM`           
  - TLS key-share path: store the MlKemKey object (rather than private key bytes) when PKCS#11 ML-KEM is active, matching the existing WOLFSSL_MLKEM_CACHE_A behavior. Guard restricted to WOLFSSL_WC_MLKEM to avoid enabling this on the liboqs path                              
  - wc_MlKemKey_Free crypto-callback hook to destroy the ephemeral HSM private key object on free                                             
  
 ### Also fixes a pre-existing issue across ML-KEM, ML-DSA, and ECC:        
                                                            
The `WOLF_CRYPTO_CB_FREE` path in all three Key_Free functions returned early on callback success, silently skipping `ForceZero` on private key material and freeing of internal objects (PRF/hash for ML-KEM, SHAKE +  cached vectors for ML-DSA, mp_forcezero + hardware port frees for ECC).

The PKCS#11 backend worked around this by returning `CRYPTOCB_UNAVAILABLE` on success, a fragile undocumented contract. All three functions are fixed to always run software cleanup regardless of the callback result; the `CRYPTOCB_UNAVAILABLE` workaround is removed from `wc_Pkcs11_CryptoDevCb`.

### DIFF
--- a/wolfcrypt/src/wc_mlkem.c
+++ b/wolfcrypt/src/wc_mlkem.c
@@ -391,21 +391,15 @@ int wc_MlKemKey_Init_Label(MlKemKey* key, int type, const char* label,
  */
 int wc_MlKemKey_Free(MlKemKey* key)
 {
-#if defined(WOLF_CRYPTO_CB) && defined(WOLF_CRYPTO_CB_FREE)
-    int ret = 0;
-#endif
-
     if (key != NULL) {
 #if defined(WOLF_CRYPTO_CB) && defined(WOLF_CRYPTO_CB_FREE)
         if (key->devId != INVALID_DEVID) {
-            ret = wc_CryptoCb_Free(key->devId, WC_ALGO_TYPE_PK,
-                WC_PK_TYPE_PQC_KEM_KEYGEN, WC_PQC_KEM_TYPE_KYBER, (void*)key);
-            if (ret != WC_NO_ERR_TRACE(CRYPTOCB_UNAVAILABLE)) {
-                return ret;
-            }
-            /* fall-through to software cleanup */
+            (void)wc_CryptoCb_Free(key->devId, WC_ALGO_TYPE_PK,
+                                   WC_PK_TYPE_PQC_KEM_KEYGEN,
+                                   WC_PQC_KEM_TYPE_KYBER,
+                                   (void*)key);
+            /* always continue to software cleanup */
         }
-        (void)ret;
 #endif
         /* Dispose of PRF object. */
         mlkem_prf_free(&key->prf);
--- a/wolfcrypt/src/wc_pkcs11.c
+++ b/wolfcrypt/src/wc_pkcs11.c
@@ -6582,12 +6582,6 @@ int wc_Pkcs11_CryptoDevCb(int devId, wc_CryptoInfo* info, void* ctx)
                                                  (ecc_key*)info->free.obj);
                     Pkcs11CloseSession(token, &session);
                 }
-                /* Return CRYPTOCB_UNAVAILABLE so wc_ecc_free() still
-                 * performs software cleanup. This callback only releases
-                 * the HSM object. Conditional because wc_ecc_free returns
-                 * int and can propagate an HSM error to the caller. */
-                if (ret == 0)
-                    ret = CRYPTOCB_UNAVAILABLE;
             }
             else
     #endif
@@ -6601,11 +6595,6 @@ int wc_Pkcs11_CryptoDevCb(int devId, wc_CryptoInfo* info, void* ctx)
                                                    (MlDsaKey*)info->free.obj);
                     Pkcs11CloseSession(token, &session);
                 }
-                /* Always return CRYPTOCB_UNAVAILABLE so wc_dilithium_free()
-                 * performs software cleanup. This callback only releases
-                 * the HSM object. Unconditional because wc_dilithium_free
-                 * returns void and cannot propagate an error. */
-                ret = CRYPTOCB_UNAVAILABLE;
             }
             else
     #endif
@@ -6618,14 +6607,6 @@ int wc_Pkcs11_CryptoDevCb(int devId, wc_CryptoInfo* info, void* ctx)
                     ret = Pkcs11MlKemDeletePrivKey(&session,
                                                    (MlKemKey*)info->free.obj);
                     Pkcs11CloseSession(token, &session);
-                }
-                /* Return CRYPTOCB_UNAVAILABLE so wc_MlKemKey_Free() still
-                 * performs software cleanup. This callback only releases
-                 * the HSM object. Conditional because wc_MlKemKey_Free
-                 * returns int and can propagate an HSM error to the caller.
-                 */
-                if (ret == 0) {
-                    ret = CRYPTOCB_UNAVAILABLE;
                 }
             }
             else
--- a/wolfcrypt/src/dilithium.c
+++ b/wolfcrypt/src/dilithium.c
@@ -10963,22 +10963,15 @@ int wc_dilithium_get_level(dilithium_key* key, byte* level)
  */
 void wc_dilithium_free(dilithium_key* key)
 {
-#if defined(WOLF_CRYPTO_CB) && defined(WOLF_CRYPTO_CB_FREE)
-    int ret = 0;
-#endif
-
     if (key != NULL) {
 #if defined(WOLF_CRYPTO_CB) && defined(WOLF_CRYPTO_CB_FREE)
         if (key->devId != INVALID_DEVID) {
-            ret = wc_CryptoCb_Free(key->devId, WC_ALGO_TYPE_PK,
+            (void)wc_CryptoCb_Free(key->devId, WC_ALGO_TYPE_PK,
                              WC_PK_TYPE_PQC_SIG_KEYGEN,
                              WC_PQC_SIG_TYPE_DILITHIUM,
                              (void*)key);
-            if (ret != WC_NO_ERR_TRACE(CRYPTOCB_UNAVAILABLE))
-                return;
-            /* fall-through to software cleanup */
+            /* always continue to software cleanup */
         }
-        (void)ret;
 #endif
 #ifdef WOLFSSL_WC_DILITHIUM
 #ifndef WC_DILITHIUM_FIXED_ARRAY
--- a/wolfcrypt/src/ecc.c
+++ b/wolfcrypt/src/ecc.c
@@ -7930,23 +7930,19 @@ void wc_ecc_free_curve(ecc_set_type* curve, void* heap)
 WOLFSSL_ABI
 int wc_ecc_free(ecc_key* key)
 {
-#if defined(WOLF_CRYPTO_CB) && defined(WOLF_CRYPTO_CB_FREE)
-    int ret = 0;
-#endif
-
     if (key == NULL) {
         return 0;
     }
 
 #if defined(WOLF_CRYPTO_CB) && defined(WOLF_CRYPTO_CB_FREE)
     if (key->devId != INVALID_DEVID) {
-        ret = wc_CryptoCb_Free(key->devId, WC_ALGO_TYPE_PK,
+        /* Best-effort HSM resource release; errors are intentionally discarded
+         * so that software cleanup always runs and wc_ecc_free() retains its
+         * ABI guarantee of returning 0 on success. */
+        (void)wc_CryptoCb_Free(key->devId, WC_ALGO_TYPE_PK,
                          WC_PK_TYPE_EC_KEYGEN, 0, key);
-        if (ret != WC_NO_ERR_TRACE(CRYPTOCB_UNAVAILABLE))
-            return ret;
-        /* fall-through to software cleanup */
+        /* always continue to software cleanup */
     }
-    (void)ret;
 #endif
 
 #if defined(WOLFSSL_ECDSA_SET_K) || defined(WOLFSSL_ECDSA_SET_K_ONE_LOOP) || \
@@ -7960,6 +7956,7 @@ int wc_ecc_free(ecc_key* key)
         mp_free(key->sign_k);
 #ifndef WOLFSSL_NO_MALLOC
         XFREE(key->sign_k, key->heap, DYNAMIC_TYPE_ECC);
+        key->sign_k = NULL;
 #endif
     }
 #endif
@@ -8025,8 +8022,10 @@ int wc_ecc_free(ecc_key* key)
 #endif
 
 #ifdef WOLFSSL_CUSTOM_CURVES
-    if (key->deallocSet && key->dp != NULL)
+    if (key->deallocSet && key->dp != NULL) {
         wc_ecc_free_curve((ecc_set_type *)(wc_ptr_t)key->dp, key->heap);
+        key->dp = NULL;
+    }
 #endif
 
 #ifdef WOLFSSL_CHECK_MEM_ZERO


Frauschi:
Jenkins retest this please

wolfcrypt/src/wc_pkcs11.c
```diff
@@ -3979,6 +4235,506 @@ static int Pkcs11EccDeletePrivKey(Pkcs11Session* session, ecc_key* key)
...
+        { CKA_CLASS,         &pubKeyClass,  sizeof(pubKeyClass)  },
+        { CKA_ENCAPSULATE,   &ckTrue,       sizeof(ckTrue)       },
+        { CKA_KEY_TYPE,      &mlkemKeyType, sizeof(mlkemKeyType) },
+        { CKA_PARAMETER_SET, &param_set,    sizeof(param_set)    },
+        { 0,                 NULL,          0                    },
+        { 0,                 NULL,          0                    }
+    };
+    CK_ULONG                     pubTmplCnt =
+        sizeof(pubKeyTmpl) / sizeof(*pubKeyTmpl) - 2;
+    CK_ATTRIBUTE                 privKeyTmpl[] = {
```
dgarske, line 4381: **🟡 [Medium] Missing CKA_PARAMETER_SET in ML-KEM private key template for keygen**
*💡 SUGGEST*  `bug`

The `privKeyTmpl` array in `Pkcs11MlKemKeyGen` does not include `CKA_PARAMETER_SET`, while the `pubKeyTmpl` does (and the analogous ML-DSA keygen at line 4942 includes `CKA_PARAMETER_SET` in both templates). Some PKCS#11 tokens may require the parameter set on both the public and private key templates during `C_GenerateKeyPair`. This inconsistency with the ML-DSA pattern and the PKCS#11 3.2 spec guidance could cause failures on certain HSMs.

**Suggestion:**
```suggestion
CK_ATTRIBUTE                 privKeyTmpl[] = {
        { CKA_CLASS,         &privKeyClass, sizeof(privKeyClass) },
        { CKA_DECAPSULATE,   &ckTrue,       sizeof(ckTrue)       },
        { CKA_KEY_TYPE,      &mlkemKeyType, sizeof(mlkemKeyType) },
        { CKA_PARAMETER_SET, &param_set,    sizeof(param_set)    },
        { 0,                 NULL,          0                    },
        { 0,                 NULL,          0                    }
    };
    CK_ULONG                     privTmplCnt =
        sizeof(privKeyTmpl) / sizeof(*privKeyTmpl) - 2;
```

wolfcrypt/test/test.c
```diff
@@ -66105,6 +66194,21 @@ static int myCryptoDevCb(int devIdArg, wc_CryptoInfo* info, void* ctx)
                     }
                     break;
                 }
+#endif
+#ifdef WOLFSSL_HAVE_MLKEM
```
dgarske, line 66198: **🟡 [Medium] Test callback double-frees MlKemKey resources via wc_MlKemKey_Free**
*💡 SUGGEST*  `bug`

The `WC_ALGO_TYPE_FREE` handler for ML-KEM calls `wc_MlKemKey_Free(mlkem)` inside the crypto callback. The call chain is:
1. User code calls `wc_MlKemKey_Free(key)` → invokes `wc_CryptoCb_Free`
2. Callback sets `devId = INVALID_DEVID` and calls `wc_MlKemKey_Free(mlkem)` → does full software cleanup (prf_free, hash_free, ForceZero)
3. Returns to outer `wc_MlKemKey_Free` which does software cleanup AGAIN

`mlkem_prf_free` and `mlkem_hash_free` are called twice on the same object. While likely benign because `ForceZero` has been called in between, this is still incorrect and inconsistent with the dilithium callback pattern (which calls `wc_dilithium_free` which returns void, but there the same double-cleanup exists too as a pre-existing pattern). The ML-KEM callback should not call `wc_MlKemKey_Free` - it should just clean up the device-specific resources.

**Suggestion:**
```suggestion
The test callback's FREE handler should only release device resources (if any), not call wc_MlKemKey_Free. Returning 0 is sufficient since the outer wc_MlKemKey_Free will handle software cleanup:

                case WC_PK_TYPE_PQC_KEM_KEYGEN:
                {
                    if (info->free.subType == WC_PQC_KEM_TYPE_KYBER) {
                        /* Software cleanup is handled by the caller's
                         * wc_MlKemKey_Free(). Nothing device-specific to
                         * release in the test callback. */
                        ret = 0;
                    }
                    break;
                }
```

wolfcrypt/src/wc_mlkem.c
```diff
@@ -330,7 +389,19 @@ int wc_MlKemKey_Init(MlKemKey* key, int type, void* heap, int devId)
  */
 int wc_MlKemKey_Free(MlKemKey* key)
```
dgarske, line 392: **🟡 [Medium] wc_MlKemKey_Free propagates callback errors inconsistently with ecc/dilithium**
*💡 SUGGEST*  `convention`

The PR description states all three free functions are fixed to 'always run software cleanup regardless of the callback result.' The ecc and dilithium fixes cast the callback return to void: `(void)wc_CryptoCb_Free(...)`. However, `wc_MlKemKey_Free` captures and returns the error: `ret = wc_CryptoCb_Free(...)`. If the PKCS#11 `C_DestroyObject` fails (e.g., session expired), `wc_MlKemKey_Free` returns an error even though all software cleanup completed successfully. This is inconsistent — callers may interpret a non-zero return as 'the key was not properly freed' even though memory was cleaned up.

**Recommendation:** Consider whether the return value should be 0 after successful software cleanup, matching ecc/dilithium behavior. If there's a deliberate reason to propagate HSM errors from Free (the return type is int, not void), document that difference clearly.

wolfcrypt/src/wc_pkcs11.c
```diff
@@ -3979,6 +4235,506 @@ static int Pkcs11EccDeletePrivKey(Pkcs11Session* session, ecc_key* key)
 }
 #endif
 
+#ifdef HAVE_PKCS11_MLKEM
```
dgarske, line 4241: **🟡 [Medium] Pkcs11FindMlKemKey does not use label/id for key lookup**
*💡 SUGGEST*  `question`

The `Pkcs11FindMlKemKey` function searches by class, key type, and parameter set, but does not include `CKA_LABEL` or `CKA_ID` in the search template. In contrast, the callers in `Pkcs11MlKemEncapsulate` and `Pkcs11MlKemDecapsulate` have separate paths for label/id lookup using `Pkcs11FindKeyByLabel`/`Pkcs11FindKeyById`, falling through to `Pkcs11FindMlKemKey` as a last resort. If a token has multiple ML-KEM keys of the same type, `Pkcs11FindMlKemKey` may return the wrong key since it stops at `count > 0`. This is consistent with how `Pkcs11FindMldsaKey` works (same pattern), but worth documenting the limitation.

**Recommendation:** This is consistent with the existing ML-DSA pattern, but consider adding a comment documenting that this is a fallback that returns the first matching key when no label/id is available.

wolfcrypt/src/wc_mlkem.c
```diff
@@ -322,6 +329,58 @@ int wc_MlKemKey_Init(MlKemKey* key, int type, void* heap, int devId)
...
+    }
+    if (ret == 0 && id != NULL && len != 0) {
+        XMEMCPY(key->id, id, (size_t)len);
+        key->idLen = len;
+    }
+
+    return ret;
+}
+
+int wc_MlKemKey_Init_Label(MlKemKey* key, int type, const char* label,
```
dgarske, line 356: **🔵 [Low] wc_MlKemKey_Init_Label allows label length up to MLKEM_MAX_LABEL_LEN but does not NUL-terminate**
*🔧 NIT*  `api`

When `XSTRLEN(label) == MLKEM_MAX_LABEL_LEN` (32), the label fills the entire `key->label[32]` buffer with no NUL terminator. This is fine since `key->labelLen` tracks the length, and callers in wc_pkcs11.c use `key->labelLen` not strlen. However, this differs from typical C string conventions and any future code that treats `key->label` as a NUL-terminated string would read past the buffer. The existing ML-DSA code has the same pattern.

**Recommendation:** This is consistent with existing code (ML-DSA, ECC). No change needed, but a brief comment noting the label is not NUL-terminated could prevent future misuse.

wolfcrypt/src/wc_pkcs11.c
```diff
@@ -1962,6 +2192,32 @@ int wc_Pkcs11StoreKey(Pkcs11Token* token, int type, int clear, void* key)
...
+            case PKCS11_KEY_TYPE_MLKEM: {
+                MlKemKey* mlkemKey = (MlKemKey*)key;
+                CK_MECHANISM_INFO mechInfo;
+
+                ret = Pkcs11MechAvail(&session, CKM_ML_KEM, &mechInfo);
+                if (ret == 0 && ((mlkemKey->flags & MLKEM_FLAG_PRIV_SET) != 0)) {
+                    ret = Pkcs11CreateMlKemPrivateKey(&privKey, &session,
+                                                      mlkemKey, &mechInfo);
+                }
+                if (ret == 0 && ((mlkemKey->flags & MLKEM_FLAG_PUB_SET) != 0)) {
```
dgarske, line 2204: **🟠 [High] wc_Pkcs11StoreKey ML-KEM pubKey handle leaked on success**
*🚫 BLOCK*  `bug`

In the `PKCS11_KEY_TYPE_MLKEM` case of `wc_Pkcs11StoreKey`, when both private and public key creation succeed, the `pubKey` handle is never destroyed and is discarded as a local variable. The handle `pubKey` is only cleaned up if `ret != 0`, meaning on the success path the token retains the public key object but the handle is lost. Compare with the ECC path (line 2172-2183) where `pubKey` is also stored on the token intentionally (for cert validation). However, the ML-KEM code stores `pubKey` in a local scope variable `CK_OBJECT_HANDLE pubKey = NULL_PTR` that goes out of scope at the closing brace, so the handle is simply discarded. If the intent is to keep it on the token, this is fine but the handle is leaked locally. If the intent is not to keep it, it should be destroyed.

**Recommendation:** If the public key is intentionally stored on the token (for cert validation or encapsulation), add a comment. If it should be ephemeral, destroy it after creation. The ECC path keeps both keys intentionally—if ML-KEM should follow the same pattern, this is fine but the handle is still lost from the software side.

wolfcrypt/src/wc_pkcs11.c
```diff
@@ -3979,6 +4235,506 @@ static int Pkcs11EccDeletePrivKey(Pkcs11Session* session, ecc_key* key)
...
+        PKCS11_RV("C_GenerateKeyPair", rv);
+        if (rv != CKR_OK) {
+            ret = WC_HW_E;
+        }
+    }
+    if (ret == 0) {
+        ret = Pkcs11GetMlKemPublicKey(key, session, pubKey);
+    }
+
+    if (pubKey != NULL_PTR) {
```
dgarske, line 4470: **🟡 [Medium] Pkcs11MlKemKeyGen always destroys HSM public key after extracting bytes**
*💡 SUGGEST*  `question`

After key generation, `Pkcs11MlKemKeyGen` unconditionally destroys the HSM-resident public key object (`C_DestroyObject(pubKey)`) after extracting the public key bytes into the wolfSSL key object. However, in the `wc_Pkcs11StoreKey` path the public key is explicitly created on the token. This means keygen-then-encapsulate requires re-uploading the public key to the token. This is a design choice (the private key stays on token, public key is in-memory), but it's worth confirming this is intentional since it differs from the ML-DSA keygen which does the same thing. It's consistent, just somewhat surprising for a `StoreKey` workflow.

**Recommendation:** No change needed if intentional. Add a comment explaining why the public key is destroyed (bytes were extracted to the wolfSSL key object, so the HSM object is not needed).

wolfcrypt/src/wc_mlkem.c
```diff
@@ -382,6 +453,21 @@ int wc_MlKemKey_MakeKey(MlKemKey* key, WC_RNG* rng)
...
     }
 
+#ifdef WOLF_CRYPTO_CB
+    if ((ret == 0)
+    #ifndef WOLF_CRYPTO_CB_FIND
+        && (key->devId != INVALID_DEVID)
+    #endif
+    ) {
+        ret = wc_CryptoCb_MakePqcKemKey(rng, WC_PQC_KEM_TYPE_KYBER,
+                                        key->type, key);
```
dgarske, line 462: **🔵 [Low] Inconsistent braces in crypto CB dispatch for MakeKey**
*🔧 NIT*  `style`

The `if` block uses brace-less style for a multi-line `return` statement: `if (ret != WC_NO_ERR_TRACE(CRYPTOCB_UNAVAILABLE))\n    return ret;`. Other code in the same file uses braces. The decapsulate and encapsulate paths use the same pattern, so this is internally consistent within the new code, but differs from the braced style elsewhere in wc_mlkem.c.

**Recommendation:** This matches the existing PKCS#11/crypto callback pattern used throughout the codebase (e.g., dilithium, ecc). No change needed, though adding braces would improve readability.

wolfcrypt/src/wc_pkcs11.c
```diff
@@ -29,6 +29,7 @@
 
 #include <wolfssl/wolfcrypt/wc_pkcs11.h>
 #include <wolfssl/wolfcrypt/asn.h>
+#include <wolfssl/wolfcrypt/mlkem.h>
```
dgarske: **🔵 [Low] mlkem.h include added unconditionally to wc_pkcs11.c**
*💡 SUGGEST*  `convention`

The `#include <wolfssl/wolfcrypt/mlkem.h>` is added outside any `#ifdef` guard, unlike `rsa.h` which is wrapped in `#ifndef NO_RSA`. While `mlkem.h` has an internal `#ifdef WOLFSSL_HAVE_MLKEM` guard making this safe, it's inconsistent with how other algorithm headers are included in this file. Other headers like `rsa.h` are conditionally included.

**Recommendation:** Consider wrapping with `#ifdef WOLFSSL_HAVE_MLKEM` for consistency, or note that `mlkem.h` is self-guarding. Low priority.

wolfcrypt/src/wc_pkcs11.c
Frauschi: False positive.

The PKCS#11 V3.2 specification explicitly states that the `CKA_PARAMETER_SET` is omitted in the private key template for key generation [[PKCS#11 V3.2 specification, Section 6.68.3](https://docs.oasis-open.org/pkcs11/pkcs11-spec/v3.2/pkcs11-spec-v3.2.html#_Toc195693756)]:

> Note that when generating a ML-KEM private key, the parameter set is not specified in the key’s template. This is because ML-KEM private keys are only generated as part of a ML-KEM key pair, and the parameter set for the pair is specified in the template for the ML-KEM public key.

The same goes for ML-DSA private keys. However, as identified correctly, ML-DSA still had the `CKA_PARAMETER_SET` in the private key template. This is removed now, too.

Note that this behavior is explicitly supported and tested in both algorithm implementations in wolfPKCS11.

wolfcrypt/test/test.c
Frauschi: This is the same behavior as for all the other already existing algorithms in the respective free callback, so it “seems” to be the intended behavior. 
If this is not the actual intended behavior, I think it should be cleaned up for all algorithms in a separate PR.

wolfcrypt/src/wc_mlkem.c
Frauschi: fixed by applying the same behavior as for ML-DSA and ECC keys.

wolfcrypt/src/wc_pkcs11.c
Frauschi: Added comments to document the fallback.

wolfcrypt/src/wc_mlkem.c
Frauschi: Added comment to `wc_MlKemKey_Init_Label()` to indicate this behavior.

wolfcrypt/src/wc_pkcs11.c
Frauschi: This is intended behavior. Added comments to document that.

wolfcrypt/src/wc_pkcs11.c
Frauschi: Intended. Added comment to document that.

wolfcrypt/src/wc_mlkem.c
Frauschi: Same style as everywhere in the codebase for this logic.

wolfcrypt/src/wc_pkcs11.c
Frauschi: Removed the include completely, as the relevant headers are already included via the `cryptocb.h` header (properly gated there) via the `wc_pkcs11.h` file.

Frauschi:
I addressed all review comments @dgarske

dgarske (approved):

PR #10085, opened 2026-03-26
Add RFC8773(bis) cert_with_extern_psk support

Frauschi:
Implement [RFC 8773(bis)](https://datatracker.ietf.org/doc/draft-ietf-tls-8773bis/) cert_with_extern_psk for TLS 1.3.

Includes unit tests for API and handshake behavior as well as tests in the testsuite using extended examples.


Frauschi:
@wolfSSL-Fenrir-bot Findings 1 and 2 identify a real bug,
  but the recommended fix (continue instead of return) would be a spec   
  violation.                                                          
                                                                         
  RFC 8773bis §5.1 (draft-ietf-tls-8773bis-13) is unambiguous:
                                                                         
  ▎ "All of the listed PSKs MUST be external PSKs. If a resumption PSK is
   listed along with the tls_cert_with_extern_psk extension, the server  
  MUST abort the handshake with an illegal_parameter alert."             
                                                            
  There is no "skip and continue" option — the presence of any resumption
   PSK identity alongside tls_cert_with_extern_psk is a protocol
  violation requiring a hard abort. Changing to continue would allow the 
  server to accept an external PSK identity from a list that also
  contained a forbidden resumption PSK, violating the spec. The return
  PSK_KEY_ERROR is correct because PSK_KEY_ERROR maps to
  illegal_parameter in TranslateErrorToAlert (src/internal.c).

  The actual bug was that the check fired too late: only after           
  DoClientTicketCheck succeeded. If a ticket successfully decrypted
  (WOLFSSL_TICKET_RET_OK) but then failed DoClientTicketCheck (e.g.      
  cipher suite mismatch, expired), the early continue at the
  check-failure path bypassed the certWithExternOffered guard entirely. A
   client could therefore offer cert_with_extern_psk alongside a
  resumption PSK identity that happened to fail the suite/expiry check,
  and the server would proceed to evaluate subsequent external PSK
  identities without aborting — a spec violation.

  The fix moves the certWithExternOffered abort to fire immediately upon 
  successful ticket decryption, before DoClientTicketCheck and
  DoClientTicketFinalize. This ensures any recognized session ticket     
  identity triggers the mandatory abort regardless of further validity
  checks. The sess_free_cb cleanup is included in the early-exit path,
  and the move also avoids DoClientTicketFinalize polluting ssl->session
  with ticket state that will never be used.

Frauschi:
Jenkins retest this please

src/tls13.c
```diff
@@ -6192,6 +6227,26 @@ static int DoPreSharedKeys(WOLFSSL* ssl, const byte* input, word32 inputSz,
...
+            if (certWithExternOffered) {
+                if (current->sess_free_cb != NULL) {
+                    current->sess_free_cb(ssl, current->sess,
+                            &current->sess_free_cb_ctx);
+                    current->sess = NULL;
+                    XMEMSET(&current->sess_free_cb_ctx, 0,
+                            sizeof(psk_sess_free_cb_ctx));
+                }
+                WOLFSSL_ERROR_VERBOSE(PSK_KEY_ERROR);
+                return PSK_KEY_ERROR;
```
dgarske: test

tests/api/test_tls13.c
```diff
@@ -868,6 +868,196 @@ int test_tls13_apis(void)
...
+    if (key_max_len < sizeof(test_tls13_cwep_psk) || id == NULL)
+        return 0;
+    if (XSTRCMP(id, "cwep_client") != 0)
+        return 0;
+    XMEMCPY(key, test_tls13_cwep_psk, sizeof(test_tls13_cwep_psk));
+    return (unsigned int)sizeof(test_tls13_cwep_psk);
+}
+#endif
+
+int test_tls13_cert_with_extern_psk_handshake(void)
```
dgarske, line 951: test

tests/api/test_tls13.c
```diff
@@ -868,6 +868,196 @@ int test_tls13_apis(void)
...
+    WOLFSSL *ssl_s = NULL;
+    struct test_memio_ctx test_ctx;
+
+    XMEMSET(&test_ctx, 0, sizeof(test_ctx));
+    ExpectIntEQ(test_memio_setup(&test_ctx, &ctx_c, &ctx_s, &ssl_c, &ssl_s,
+        wolfTLSv1_3_client_method, wolfTLSv1_3_server_method), 0);
+
+    wolfSSL_set_verify(ssl_c, WOLFSSL_VERIFY_NONE, NULL);
+    wolfSSL_set_verify(ssl_s, WOLFSSL_VERIFY_NONE, NULL);
+#if defined(HAVE_ECC) && !defined(NO_CERTS) && !defined(NO_FILESYSTEM)
```
dgarske: test

src/tls.c
```diff
@@ -17693,6 +17872,64 @@ int TLSX_Parse(WOLFSSL* ssl, const byte* input, word16 length, byte msgType,
...
+                if (secondClientHello && !prevHasPskWithCert) {
+                    WOLFSSL_ERROR_VERBOSE(EXT_NOT_ALLOWED);
+                    return EXT_NOT_ALLOWED;
+                }
+                /* RFC8773bis: cert_with_extern_psk depends on these extensions. */
+                if (!hasPsk || !hasPskModes || !hasKeyShare || !hasSg) {
+                    WOLFSSL_ERROR_VERBOSE(EXT_MISSING);
+                    return EXT_MISSING;
+                }
+#ifdef WOLFSSL_EARLY_DATA
```
dgarske, line 17916: test

src/tls13.c
dgarske: **🟠 [High] Early return in DoPreSharedKeys skips ForceZero cleanup of sensitive key material**
*🚫 BLOCK*  `bug`

When `certWithExternOffered` is true and a session ticket is successfully decrypted (resumption PSK detected), the new code at line 6247 does `return PSK_KEY_ERROR;` instead of `goto cleanup;`. The function's cleanup label at line 6368-6370 performs `ForceZero(binderKey, ...)` and `ForceZero(binder, ...)` to scrub sensitive cryptographic material from the stack. The early return bypasses this cleanup, leaving binder key material on the stack. Every other error path in this function uses `goto cleanup`.

**Suggestion:**
```suggestion
WOLFSSL_ERROR_VERBOSE(PSK_KEY_ERROR);
                ret = PSK_KEY_ERROR;
                goto cleanup;
```

*— Skoll automated review via openclaw*

src/tls13.c
```diff
@@ -4310,8 +4310,17 @@ static int SetupPskKey(WOLFSSL* ssl, PreSharedKey* psk, int clientHello)
         }
 
         if (!clientHello) {
-            /* CLIENT: using PSK for peer authentication. */
-            ssl->options.peerAuthGood = 1;
+#ifdef WOLFSSL_CERT_WITH_EXTERN_PSK
+            if (ssl->options.certWithExternPsk) {
+                /* Certificate authentication is still required. */
+                ssl->options.peerAuthGood = 0;
```
dgarske, line 4316: **🟡 [Medium] FindPsk sets peerAuthGood=1 on server without considering certWithExternPsk**
*💡 SUGGEST*  `bug`

The `FindPsk` function (server-side PSK matching) unconditionally sets `ssl->options.peerAuthGood = 1` at line 6134 when an external PSK identity is matched. In `cert_with_extern_psk` mode, the client (the server's peer) must also authenticate via certificate. While `CheckPreSharedKeys` subsequently sets `ssl->options.sendVerify = SEND_CERT` and `verifyPeer` is preserved, `peerAuthGood` is already set to 1 before the client certificate is verified. The client-side `SetupPskKey` at line 4314 correctly gates `peerAuthGood` on `certWithExternPsk`, but the server-side `FindPsk` does not. In practice this may work because the server later validates the client cert via `DoTls13CertificateVerify`, and `peerAuthGood` is set again there. However, there's a window where `peerAuthGood=1` before the client has authenticated if `verifyPeer` is configured to not fail on missing cert.

**Suggestion:**
```suggestion
if (ret == 0) {
            /* PSK negotiation has succeeded */
            ssl->options.isPSK = 1;
#ifdef WOLFSSL_CERT_WITH_EXTERN_PSK
            if (!ssl->options.certWithExternPsk)
#endif
            {
                /* SERVER: using PSK for peer authentication. */
                ssl->options.peerAuthGood = 1;
            }
        }
```

*— Skoll automated review via openclaw*

tests/api/test_tls13.c
dgarske, line 951: **🟡 [Medium] No negative test for client receiving resumption ticket with cert_with_extern_psk**
*💡 SUGGEST*  `test`

The tests cover the happy path (successful handshake) and the HRR path, but there are no negative tests for important error scenarios introduced by this PR: (1) Server receives a resumption ticket PSK alongside `cert_with_extern_psk` extension (should fail with PSK_KEY_ERROR per line 6238-6248 in tls13.c), (2) Client receives ServerHello with `cert_with_extern_psk` but no key_share (should fail with EXT_MISSING per line 5753-5758 in tls13.c), (3) Client receives ServerHello confirming `cert_with_extern_psk` for a resumption PSK (should fail with PSK_KEY_ERROR per line 5748-5752 in tls13.c).

**Recommendation:** Add unit tests for the error paths: (1) test that resumption ticket identity with `cert_with_extern_psk` is rejected on the server, (2) test that missing key_share in ServerHello with `cert_with_extern_psk` is rejected on the client. These are critical validation paths.

*— Skoll automated review via openclaw*

tests/api/test_tls13.c
dgarske: **🟡 [Medium] test_tls13_cert_with_extern_psk_handshake may silently skip cert verification**
*💡 SUGGEST*  `test`

The handshake test loads ECC certs only under `#if defined(HAVE_ECC) && !defined(NO_CERTS) && !defined(NO_FILESYSTEM)`. If ECC is not available (e.g. RSA-only build), no server certificate is loaded, and the test may either fail the handshake (making it not a valid positive test) or succeed without actually sending a certificate. The test sets `WOLFSSL_VERIFY_NONE` on both sides, so no cert verification occurs even in the happy path. Consider also loading RSA certs as a fallback to ensure broader test coverage.

**Suggestion:**
```suggestion
#if !defined(NO_CERTS) && !defined(NO_FILESYSTEM)
#if defined(HAVE_ECC)
    ExpectTrue(wolfSSL_use_certificate_file(ssl_s, eccCertFile,
        CERT_FILETYPE) == WOLFSSL_SUCCESS);
    ExpectTrue(wolfSSL_use_PrivateKey_file(ssl_s, eccKeyFile,
        CERT_FILETYPE) == WOLFSSL_SUCCESS);
#elif !defined(NO_RSA)
    ExpectTrue(wolfSSL_use_certificate_file(ssl_s, svrCertFile,
        CERT_FILETYPE) == WOLFSSL_SUCCESS);
    ExpectTrue(wolfSSL_use_PrivateKey_file(ssl_s, svrKeyFile,
        CERT_FILETYPE) == WOLFSSL_SUCCESS);
#endif
#endif
```

*— Skoll automated review via openclaw*

src/tls.c
dgarske, line 17916: **🟡 [Medium] Early data extension validation runs only when WOLFSSL_EARLY_DATA is defined**
*💡 SUGGEST*  `question`

The RFC 8773bis validation that rejects early_data alongside cert_with_extern_psk is wrapped in `#ifdef WOLFSSL_EARLY_DATA`. If a build has `WOLFSSL_CERT_WITH_EXTERN_PSK` but not `WOLFSSL_EARLY_DATA`, a peer that sends an early_data extension would not be rejected by this check. However, this may be safe because without `WOLFSSL_EARLY_DATA` the extension wouldn't be processed at all (it would be treated as unknown). Flagging for consideration.

**Recommendation:** Verify that builds without WOLFSSL_EARLY_DATA still reject or ignore early_data extensions from peers. If unknown extensions are simply ignored (as per TLS 1.3 spec), this is safe.

*— Skoll automated review via openclaw*

src/tls13.c
Frauschi: fixed as suggested

src/tls13.c
Frauschi: fixed as suggested

tests/api/test_tls13.c
Frauschi: fixed as suggested

tests/api/test_tls13.c
Frauschi: added test

src/tls.c
Frauschi: Added a comment to document this behavior. extension is ignored if support is not compiled in

src/tls.c
```diff
@@ -17469,6 +17623,25 @@ int TLSX_Parse(WOLFSSL* ssl, const byte* input, word16 length, byte msgType,
...
+                    WOLFSSL_ERROR_VERBOSE(EXT_NOT_ALLOWED);
+                    return EXT_NOT_ALLOWED;
+                }
+
+                ret = PSK_WITH_CERT_PARSE(ssl, msgType);
+                break;
+    #endif
     #endif
 
     #ifdef WOLFSSL_EARLY_DATA
```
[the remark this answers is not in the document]
Frauschi: fixed as suggested

Frauschi:
Jenkins retest this please

src/tls.c
```diff
@@ -12373,6 +12373,98 @@ int TLSX_PreSharedKey_Use(TLSX** extensions, const byte* identity, word16 len,
...
 #endif
 
+/******************************************************************************/
+/* Certificate Authentication with External Pre-Shared Key                    */
+/******************************************************************************/
+
+#if defined(WOLFSSL_TLS13) && defined(WOLFSSL_CERT_WITH_EXTERN_PSK) && \
+    !defined(NO_PSK)
+
+static int TLSX_Cert_With_Extern_Psk_GetSize(byte msgType, word16* pSz)
```
SparkiDev: Our convention in this file is TLSX_(.*)_<Operation>. Where there is no '_' in the middle part.
That is, TLSX_CertWithExternPsk_GetSize().

src/tls.c
```diff
@@ -12373,6 +12373,98 @@ int TLSX_PreSharedKey_Use(TLSX** extensions, const byte* identity, word16 len,
...
+{
+    (void)pSz;
+    if (msgType == client_hello || msgType == server_hello) {
```
SparkiDev: This check is usually done by the caller.
Is there a good reason for doing it inside?

src/tls.c
```diff
@@ -12373,6 +12373,98 @@ int TLSX_PreSharedKey_Use(TLSX** extensions, const byte* identity, word16 len,
...
+        /* Server has not opted in - treat the extension as unknown. */
+        if (!ssl->options.certWithExternPsk)
+            return 0;
+        /* Record that the client offered the extension, leaving resp=0.
+         * CheckPreSharedKeys() is the sole writer that flips resp to 1, and
+         * only after confirming that a non-ticket PSK was matched. */
+        if (TLSX_Find(ssl->extensions, TLSX_CERT_WITH_EXTERN_PSK) == NULL) {
+            int ret = TLSX_Push(&ssl->extensions, TLSX_CERT_WITH_EXTERN_PSK,
+                NULL, ssl->heap);
+            if (ret != 0)
```
SparkiDev: No need to check ret, just return ret, unless you are going to log a message.

src/tls.c
Frauschi: Renamed accordingly.

src/tls.c
Frauschi: fixed

src/tls.c
Frauschi: no, removed the checks.

src/tls13.c
```diff
@@ -6490,6 +6605,10 @@ static int CheckPreSharedKeys(WOLFSSL* ssl, const byte* input, word32 helloSz,
     }
```
dgarske, line 6641: **🟡 [Medium] Server certWithExternPsk flag not cleared when PSK matching fails without WOLFSSL_PSK_ID_PROTECTION**
*💡 SUGGEST*  `bug`

When `WOLFSSL_PSK_ID_PROTECTION` is NOT defined and no PSK identity matches (`*usingPSK` remains 0), `CheckPreSharedKeys` returns 0 without entering the `*usingPSK != 0` block. In this case, the `TLSX_CERT_WITH_EXTERN_PSK` extension and `ssl->options.certWithExternPsk` flag inherited from the context are never cleared. The cleanup code at lines 6608-6610 only runs inside the `#ifdef WOLFSSL_PSK_ID_PROTECTION` else-branch. While analysis shows this does not cause functional issues in practice (because downstream checks gate on `pskNegotiated` first, and `certExt->resp` stays 0 so the extension is not echoed in ServerHello), it leaves stale state on the SSL object that could become problematic if future code is added that checks `certWithExternPsk` without also checking `pskNegotiated`.

**Suggestion:**
```suggestion
Add cleanup outside the WOLFSSL_PSK_ID_PROTECTION guard. For example, after the else block and before the final return 0:

#if defined(WOLFSSL_CERT_WITH_EXTERN_PSK) && !defined(WOLFSSL_PSK_ID_PROTECTION)
    if (*usingPSK == 0) {
        TLSX_Remove(&ssl->extensions, TLSX_CERT_WITH_EXTERN_PSK, ssl->heap);
        ssl->options.certWithExternPsk = 0;
    }
#endif
```

src/tls13.c
```diff
@@ -6428,6 +6489,12 @@ static int CheckPreSharedKeys(WOLFSSL* ssl, const byte* input, word32 helloSz,
 
     if (*usingPSK != 0) {
         word32 modes;
+#ifdef WOLFSSL_CERT_WITH_EXTERN_PSK
+        int usingCertWithExternPsk = 0;
+        TLSX* certExt = NULL;
+        TLSX* pskExt = NULL;
+        PreSharedKey* chosenPsk = NULL;
+#endif
     #ifdef WOLFSSL_EARLY_DATA
```
dgarske, line 6510: **🟡 [Medium] Early data acceptance not guarded against cert_with_extern_psk in CheckPreSharedKeys**
*💡 SUGGEST*  `question`

Early data processing (lines ~6498-6521) occurs BEFORE the `cert_with_extern_psk` negotiation block (line ~6531) in `CheckPreSharedKeys`. If a client offers both early data and `cert_with_extern_psk` extension, the server may accept early data and derive early data keys before the cert_with_extern_psk check. While the TLSX_Parse validation at line 17913 rejects `early_data + cert_with_extern_psk` in the ClientHello when `WOLFSSL_EARLY_DATA` is defined, the comment in `TLSX_Parse` (lines 17908-17912) notes that when `WOLFSSL_EARLY_DATA` is NOT defined, the extension is treated as unknown and ignored. In that configuration, there is no early data parser but also no rejection. This is safe because without `WOLFSSL_EARLY_DATA` the server cannot accept early data at all. However, the ordering in `CheckPreSharedKeys` means that IF the TLSX_Parse guard were ever bypassed, the server could accept early data before cert auth is established.

**Recommendation:** Consider adding a comment documenting that TLSX_Parse rejects early_data + cert_with_extern_psk in the CH validation, so the ordering here is safe. Alternatively, add an explicit early data rejection when cert_with_extern_psk is being negotiated (e.g., set `extEarlyData->resp = 0` after the cert_with_extern_psk block).

Frauschi:
Jenkins retest this please

src/tls13.c
Frauschi: fixed

src/tls13.c
Frauschi: fixed

Frauschi:
Fixed the issues from @dgarske and also incorporated some fixes Skoll found on my side.

Frauschi:
Jenkins retest this please

Frauschi:
Rebased to master and fixed merge conflicts. No other changes.

PR #10087, opened 2026-03-27
Reject negative CRL numbers when decoding

padelsbach:
# Description

Ensure negative ASN.1 integers are not allowed for CRLs (most significant bit is set in the CRL num field). Fixes an overflow which corrupts the adjacent hash field during decode.

Fixes zd21425

# Testing

How did you test?

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


padelsbach:
jenkins retest this please

Issue #10068, opened 2026-03-25
ECH (Encrypted Client Hello) code review: 11 findings


cpsource:
Code review of all ECH-specific code paths enabled by `--enable-ech` in wolfSSL 5.9.0. At the time of this review, none of the findings below have existing issues or pull requests filed against them in the wolfSSL repository. These findings are distinct from the two ECH CVEs already fixed in 5.9.0 (CVE-2026-3549, CVE-2026-3849).

### 1. `word16` truncation of `innerClientHelloLen` causes heap buffer overflow
`src/tls13.c:4727-4728`, `src/tls.c:14135`, `wolfssl/internal.h:3140`

`innerClientHelloLen` is declared as `word16` in the `WOLFSSL_ECH` struct but is computed from `word32` values (`args->length + paddingLen + hpke->Nt`). If the sum exceeds 65535, silent truncation causes an undersized allocation at line 4868. The subsequent `XMEMCPY` at line 4878 copies the actual (non-truncated) amount of data, overflowing the heap buffer. The same `word16` truncation occurs server-side in `TLSX_ECH_ExpandOuterExtensions` at `tls.c:14135`.

### 2. `GetMsgHash` error not propagated in `EchCalcAcceptance`; OOB read via integer underflow
`src/tls13.c:3910-3927`

If `GetMsgHash` returns a negative error value, `ret` stays 0 (not propagated). The negative `hashSz` is cast to `word32` at lines 3922 and 3927, producing a massive value passed to `HashRaw`, which attempts to read far beyond the stack-allocated `clientHelloInnerHash[WC_MAX_DIGEST_SIZE]` buffer.

### 3. NULL pointer dereference on `echX` in HRR path
`src/tls13.c:5652-5656`

In `DoTls13ServerHello` client path, `args->echX = TLSX_Find(...)` can return NULL, but `echX->data->confBuf` is dereferenced without a NULL check when `extMsgType == hello_retry_request`.

### 4. Unbounded pointer arithmetic for `acceptOffset`
`src/tls13.c:5656, 7711`

`confBuf - input` (client) and `confBuf - output` (server) produce `acceptOffset` with no validation that `confBuf` falls within the buffer. A stale or NULL `confBuf` produces an undefined offset passed to `HashRaw` and buffer accesses.

### 5. Potential buffer overwrite in AAD zeroing during server ECH parsing
`src/tls.c:14384-14385`

[...]

PR #10094, opened 2026-03-27
Fixes: for GH #10068

sebastian-carpenter:
# Description

Various fixes related to GH #10068.

* Increased client hello lengths in WOLFSSL_ECH struct to be word32. This wasn't necessary but seems to follow our convention better
* The error from GetMsgHash was trampled. Fixed the checks. Added a check for when it returns 0 too (not sure if this was intentionally not checked)
* Added NULL check. Not sure how important this is but it's good defense.
* Mirrored arguments over to wc_HpkeContextOpenBase since *SealBase had them.
* ech->type is changed to EHC_TYPE_INNER but on error may not be changed back to ECH_TYPE_OUTER. May cause propagating issues so I restored the value in each return.
* Optimization: verify key length is exactly equal when setting the key so it doesn't fail later.

# Testing

Broke api negative testing (with NULL parameters) into a standalone function. So far I have only written negative tests for what was in `hpke_test_single` but at some point all the other public hpke api's should be added too.

# Checklist

 - [ X] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/src/ssl_ech.c
+++ b/src/ssl_ech.c
@@ -548,7 +548,7 @@ int SetEchConfigsEx(WOLFSSL_EchConfig** outputConfigs, void* heap,
         echConfig += 2;
 
         /* hpke public_key */
-        if (hpkePubkeyLen > HPKE_Npk_MAX ||
+        if (hpkePubkeyLen == 0 || hpkePubkeyLen > HPKE_Npk_MAX ||
                 hpkePubkeyLen != wc_HpkeKemGetEncLen(workingConfig->kemId)) {
             ret = BUFFER_E;
             break;
--- a/src/ssl_ech.c
+++ b/src/ssl_ech.c
@@ -548,7 +548,8 @@ int SetEchConfigsEx(WOLFSSL_EchConfig** outputConfigs, void* heap,
         echConfig += 2;
 
         /* hpke public_key */
-        if (hpkePubkeyLen > HPKE_Npk_MAX || hpkePubkeyLen == 0) {
+        if (hpkePubkeyLen > HPKE_Npk_MAX ||
+                hpkePubkeyLen != wc_HpkeKemGetEncLen(workingConfig->kemId)) {
             ret = BUFFER_E;
             break;
         }
--- a/src/ssl_ech.c
+++ b/src/ssl_ech.c
@@ -547,9 +547,9 @@ int SetEchConfigsEx(WOLFSSL_EchConfig** outputConfigs, void* heap,
         ato16(echConfig, &hpkePubkeyLen);
         echConfig += 2;
 
-        /* hpke public_key */
-        if (hpkePubkeyLen == 0 || hpkePubkeyLen > HPKE_Npk_MAX ||
-                hpkePubkeyLen != wc_HpkeKemGetEncLen(workingConfig->kemId)) {
+        /* hpke public_key
+         * KEM support will be checked along with the ciphersuites */
+        if (hpkePubkeyLen != wc_HpkeKemGetEncLen(workingConfig->kemId)) {
             ret = BUFFER_E;
             break;
         }
--- a/src/tls.c
+++ b/src/tls.c
@@ -14091,6 +14091,9 @@ static int TLSX_ECH_ExpandOuterExtensions(WOLFSSL* ssl, WOLFSSL_ECH* ech,
 
     newInnerChLen = innerChLen - echOuterExtLen + extraSize - sessionIdLen +
                         ssl->session->sessionIDSz;
+    if (newInnerChLen > 0xFFFF) {
+        return BUFFER_E;
+    }
 
     if (!foundEchOuter && sessionIdLen == ssl->session->sessionIDSz) {
         /* no extensions + no sessionID to copy */
--- a/src/tls.c
+++ b/src/tls.c
@@ -13696,7 +13696,7 @@ static int TLSX_ECH_Write(WOLFSSL_ECH* ech, byte msgType, byte* writeBuf,
                 writeBuf_p += ech->encLen;
             }
             /* innerClientHelloLen */
-            c16toa(ech->innerClientHelloLen, writeBuf_p);
+            c16toa((word16)ech->innerClientHelloLen, writeBuf_p);
             writeBuf_p += 2;
             /* set payload offset for when we finalize */
             ech->outerClientPayload = writeBuf_p;
@@ -14155,7 +14155,7 @@ static int TLSX_ECH_ExpandOuterExtensions(WOLFSSL* ssl, WOLFSSL_ECH* ech,
     if (ret == 0) {
         XFREE(ech->innerClientHello, heap, DYNAMIC_TYPE_TMP_BUFFER);
         ech->innerClientHello = newInnerCh;
-        ech->innerClientHelloLen = (word16)newInnerChLen;
+        ech->innerClientHelloLen = newInnerChLen;
         newInnerCh = NULL;
     }
 
@@ -14269,6 +14269,7 @@ static int TLSX_ECH_Parse(WOLFSSL* ssl, const byte* readBuf, word16 size,
     word32 offset = 0;
     word16 len;
     word16 tmpVal16;
+    word16 lenCh;
 
     WOLFSSL_MSG("TLSX_ECH_Parse");
     if (ssl->options.disableECH) {
@@ -14385,7 +14386,8 @@ static int TLSX_ECH_Parse(WOLFSSL* ssl, const byte* readBuf, word16 size,
         readBuf_p += len;
         offset += len;
         /* read payload (encrypted CH) len */
-        ato16(readBuf_p, &ech->innerClientHelloLen);
+        ato16(readBuf_p, &lenCh);
+        ech->innerClientHelloLen = lenCh;
         readBuf_p += 2;
         offset += 2;
         /* Check payload is no bigger than remaining bytes. */
--- a/src/tls13.c
+++ b/src/tls13.c
@@ -3933,9 +3933,13 @@ static int EchCalcAcceptance(WOLFSSL* ssl, byte* label, word16 labelSz,
 
     if (isHrr) {
         /* the transcript hash of ClientHelloInner1 */
-        hashSz = GetMsgHash(ssl, clientHelloInnerHash);
-        if (hashSz > 0) {
+        ret = GetMsgHash(ssl, clientHelloInnerHash);
+        if (ret > 0) {
+            hashSz = ret;
             ret = 0;
+        }
+        else if (ret == 0) {
+            ret = HASH_TYPE_E;
         }
 
         /* restart ECH transcript hash, similar to RestartHandshakeHash but
@@ -3975,6 +3979,9 @@ static int EchCalcAcceptance(WOLFSSL* ssl, byte* label, word16 labelSz,
         ret = GetMsgHash(ssl, transcriptEchConf);
         if (ret > 0) {
             ret = 0;
+        }
+        else if (ret == 0) {
+            ret = HASH_TYPE_E;
         }
     }
 
@@ -8648,6 +8655,8 @@ int CreateSigData(WOLFSSL* ssl, byte* sigData, word16* sigDataSz,
     ret = GetMsgHash(ssl, &sigData[idx]);
     if (ret < 0)
         return ret;
+    if (ret == 0)
+        return HASH_TYPE_E;
 
     *sigDataSz = (word16)(idx + ret);
     ret = 0;
--- a/src/tls13.c
+++ b/src/tls13.c
@@ -4909,8 +4909,10 @@ int SendTls13ClientHello(WOLFSSL* ssl)
         args->ech->innerClientHello =
             (byte*)XMALLOC(args->ech->innerClientHelloLen - args->ech->hpke->Nt,
             ssl->heap, DYNAMIC_TYPE_TMP_BUFFER);
-        if (args->ech->innerClientHello == NULL)
+        if (args->ech->innerClientHello == NULL) {
+            args->ech->type = ECH_TYPE_OUTER;
             return MEMORY_E;
+        }
         /* set the padding bytes to 0 */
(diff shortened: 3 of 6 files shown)


src/tls13.c
```diff
@@ -4724,8 +4731,8 @@ int SendTls13ClientHello(WOLFSSL* ssl)
             args->ech->type = 0;
             /* set innerClientHelloLen to ClientHelloInner + padding + tag */
             args->ech->paddingLen = 31 - ((args->length - 1) % 32);
-            args->ech->innerClientHelloLen = (word16)(args->length +
-                args->ech->paddingLen + args->ech->hpke->Nt);
+            args->ech->innerClientHelloLen = args->length +
+                args->ech->paddingLen + args->ech->hpke->Nt;
```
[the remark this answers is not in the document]
sebastian-carpenter: Not sure how a word16 would be exceeded. This would imply that the tls record itself is larger than 2^14 which is an error itself. The other variables have already been validated as good too.

wolfcrypt/src/hpke.c
```diff
@@ -1175,7 +1175,8 @@ int wc_HpkeContextOpenBase(Hpke* hpke, HpkeBaseContext* context, byte* aad,
     int ret;
     byte nonce[HPKE_Nn_MAX];
     WC_DECLARE_VAR(aes, Aes, 1, 0);
-    if (hpke == NULL) {
+    if (hpke == NULL || context == NULL || (aad == NULL && aadSz > 0) ||
+        ciphertext == NULL || out == NULL) {
         return BAD_FUNC_ARG;
```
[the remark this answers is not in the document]
sebastian-carpenter: Added tests for the api in hpke_test_single. Will need to add the rest at some other point.

sebastian-carpenter:
Jenkins retest this please.

src/tls.c
```diff
@@ -13696,7 +13696,7 @@ static int TLSX_ECH_Write(WOLFSSL_ECH* ech, byte msgType, byte* writeBuf,
                 writeBuf_p += ech->encLen;
             }
             /* innerClientHelloLen */
-            c16toa(ech->innerClientHelloLen, writeBuf_p);
+            c16toa((word16)ech->innerClientHelloLen, writeBuf_p);
```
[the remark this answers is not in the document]
sebastian-carpenter: outer hello does this too. This is consistent with what we already do.

src/ssl_ech.c
```diff
@@ -548,7 +548,8 @@ int SetEchConfigsEx(WOLFSSL_EchConfig** outputConfigs, void* heap,
         echConfig += 2;
 
         /* hpke public_key */
-        if (hpkePubkeyLen > HPKE_Npk_MAX || hpkePubkeyLen == 0) {
+        if (hpkePubkeyLen > HPKE_Npk_MAX ||
+                hpkePubkeyLen != wc_HpkeKemGetEncLen(workingConfig->kemId)) {
             ret = BUFFER_E;
             break;
         }
```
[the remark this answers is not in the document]
sebastian-carpenter: good catch.

src/tls13.c
```diff
@@ -5694,6 +5705,9 @@ int DoTls13ServerHello(WOLFSSL* ssl, const byte* input, word32* inOutIdx,
     /* check for acceptConfirmation */
     if (ssl->echConfigs != NULL && !ssl->options.disableECH) {
         args->echX = TLSX_Find(ssl->extensions, TLSX_ECH);
+        if (args->echX == NULL || args->echX->data == NULL)
+            return WOLFSSL_FATAL_ERROR;
```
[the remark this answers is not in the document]
sebastian-carpenter: perhaps. Similar errors are done elsewhere in the code though.

sebastian-carpenter:
Jenkins retest this please.

src/tls13.c
```diff
@@ -4759,8 +4766,8 @@ int SendTls13ClientHello(WOLFSSL* ssl)
             args->ech->type = 0;
```
dgarske: **🟡 [Medium] Incomplete fix: ech->type not restored on TLSX_GetRequestSize error return**
*💡 SUGGEST*  `bug`

The PR description states: "ech->type is changed to ECH_TYPE_INNER but on error may not be changed back to ECH_TYPE_OUTER. May cause propagating issues so I restored the value in each return." The PR correctly fixes this in the lower code block (lines 4913, 4939, 4951), but misses the same pattern in the first ECH block. At line 4757, `args->ech->type` is set to `ECH_TYPE_INNER`. At lines 4762-4763, if `TLSX_GetRequestSize` fails, the function returns without restoring `type` to `ECH_TYPE_OUTER`. This is the same class of bug the PR explicitly aims to fix.

**Suggestion:**
```suggestion
ret = TLSX_GetRequestSize(ssl, client_hello, &args->length);
            if (ret != 0) {
                args->ech->type = ECH_TYPE_OUTER;
                return ret;
            }
```

*— Skoll automated review via openclaw*

dgarske: **🔵 [Low] Incomplete fix: `type = 0` not updated to `ECH_TYPE_OUTER` for consistency**
*🔧 NIT*  `convention`

The PR explicitly changed `args->ech->type = 0` to `args->ech->type = ECH_TYPE_OUTER` at line 4951, since `ECH_TYPE_OUTER` is the named constant for value 0. However, at line 4766 (in the same function, same ECH block but the size-computation section), the code still uses the literal `0`. This is functionally equivalent (`ECH_TYPE_OUTER = 0`) but inconsistent with the PR's own convention change.

**Suggestion:**
```suggestion
/* set the type to outer */
            args->ech->type = ECH_TYPE_OUTER;
```

*— Skoll automated review via openclaw*

dgarske (commented):
## 🐺 Skoll Code Review

### Posted findings
- [Medium] **Incomplete fix: ech->type not restored on TLSX_GetRequestSize error return** — `src/tls13.c:4757-4763`
- [Low] **Incomplete fix: `type = 0` not updated to `ECH_TYPE_OUTER` for consistency** — `src/tls13.c:4766`
- [Medium] **No truncation guard on c16toa after widening innerClientHelloLen to word32** — `src/tls.c:13699`
- [Medium] **Missing negative tests for changed HPKE context functions** — `wolfcrypt/test/test.c:32182`
- [Medium] **SetEchConfigsEx: wc_HpkeKemGetEncLen returns 0 for unsupported KEM, redundant with hpkePubkeyLen == 0 check** — `src/ssl_ech.c:551-553`

*Review generated by [Skoll](https://github.com/wolfSSL/skoll) via openclaw*

src/tls.c
dgarske, line 13699: **🟡 [Medium] No truncation guard on c16toa after widening innerClientHelloLen to word32**
*💡 SUGGEST*  `bug`

The PR widened `innerClientHelloLen` from `word16` to `word32` and added an explicit `(word16)` cast at line 13699: `c16toa((word16)ech->innerClientHelloLen, writeBuf_p)`. This writes a 16-bit value to the wire, but the subsequent `XMEMSET(writeBuf_p, 0, ech->innerClientHelloLen)` at line 13704 uses the full 32-bit value. If `innerClientHelloLen` were ever > 65535, the wire header would indicate a truncated length while the memset writes more bytes. While TLS message size constraints make this practically impossible today, the explicit cast silently truncates without a protective check, which could become a real issue if calling code changes.

**Suggestion:**
```suggestion
/* innerClientHelloLen */
            if (ech->innerClientHelloLen > UINT16_MAX) {
                return BUFFER_E;
            }
            c16toa((word16)ech->innerClientHelloLen, writeBuf_p);
```

*— Skoll automated review via openclaw*

wolfcrypt/test/test.c
```diff
@@ -32177,7 +32177,9 @@ WOLFSSL_TEST_SUBROUTINE wc_test_ret_t x963kdf_test(void)
         defined(HAVE_CURVE448)) && \
     defined(HAVE_AESGCM)
 
-static wc_test_ret_t hpke_test_single(Hpke* hpke)
+#if defined(HAVE_ECC) && defined(WOLFSSL_AES_128) && \
+    (!defined(NO_ECC256) || defined(HAVE_ALL_CURVES)) && !defined(NO_SHA256)
+static wc_test_ret_t hpke_test_api(Hpke* hpke)
```
dgarske: **🟡 [Medium] Missing negative tests for changed HPKE context functions**
*💡 SUGGEST*  `test`

The PR modifies NULL-argument validation in `wc_HpkeContextOpenBase` (adding the `aad == NULL && aadSz != 0` check), `wc_HpkeContextSealBase`, `wc_HpkeInitSealContext`, and `wc_HpkeInitOpenContext` (changing `> 0` to `!= 0`). However, the new `hpke_test_api` function only tests `wc_HpkeGenerateKeyPair`, `wc_HpkeSealBase`, `wc_HpkeSerializePublicKey`, and `wc_HpkeOpenBase` — not the context-based functions directly. While `wc_HpkeSealBase`/`wc_HpkeOpenBase` delegate to the context functions, their own parameter validation catches NULL arguments before reaching the context layer. The new `aad == NULL && aadSz != 0` check in `wc_HpkeContextOpenBase` is never directly exercised.

**Recommendation:** Add negative test cases for `wc_HpkeContextSealBase`, `wc_HpkeContextOpenBase`, `wc_HpkeInitSealContext`, and `wc_HpkeInitOpenContext` to directly exercise the changed validation logic. The PR description acknowledges this gap ("at some point all the other public hpke api's should be added too") — consider addressing it now rather than later.

*— Skoll automated review via openclaw*

src/ssl_ech.c
```diff
@@ -548,7 +548,8 @@ int SetEchConfigsEx(WOLFSSL_EchConfig** outputConfigs, void* heap,
         echConfig += 2;
 
         /* hpke public_key */
-        if (hpkePubkeyLen > HPKE_Npk_MAX || hpkePubkeyLen == 0) {
+        if (hpkePubkeyLen == 0 || hpkePubkeyLen > HPKE_Npk_MAX ||
```
dgarske: **🟡 [Medium] SetEchConfigsEx: wc_HpkeKemGetEncLen returns 0 for unsupported KEM, redundant with hpkePubkeyLen == 0 check**
*💡 SUGGEST*  `question`

The new check `hpkePubkeyLen != wc_HpkeKemGetEncLen(workingConfig->kemId)` validates the public key length against the expected KEM encapsulation length. When `kemId` is unsupported, `wc_HpkeKemGetEncLen` returns 0, but `hpkePubkeyLen == 0` already catches that case via short-circuit evaluation. The concern is: for a *supported* KEM where the config provides an incorrect `hpkePubkeyLen` (e.g., non-zero but wrong), this new check correctly catches it — that's the added value. However, if the KEM algorithms compiled in don't match what's in the ECH config, this returns `BUFFER_E` without a distinguishing error, making debugging harder. Consider whether a more specific error code or log message would be useful.

**Recommendation:** Consider adding a WOLFSSL_MSG before the break to indicate whether the failure was due to unsupported KEM vs. wrong key length, which aids debugging.

*— Skoll automated review via openclaw*

src/tls13.c
sebastian-carpenter: good catch, updated to enum

src/tls.c
sebastian-carpenter: Not sure this is much of an issue but decided to put a check anyway

wolfcrypt/test/test.c
sebastian-carpenter: Added extra negative tests. Also noticed that we were missing some tests for aead tags, sequences/order, etc., so I added some of those too.

src/ssl_ech.c
sebastian-carpenter: Updated it to only reject if returned kemEncLen differs from provided length. An attacker could set a 0 length and an unsupported algorithm (thus 0 == 0 and return is skipped) but this will be caught later in the function. Added error messages into the ciphersuite validation function.

src/ssl_ech.c
```diff
@@ -547,8 +547,9 @@ int SetEchConfigsEx(WOLFSSL_EchConfig** outputConfigs, void* heap,
         ato16(echConfig, &hpkePubkeyLen);
         echConfig += 2;
 
-        /* hpke public_key */
-        if (hpkePubkeyLen > HPKE_Npk_MAX || hpkePubkeyLen == 0) {
+        /* hpke public_key
+         * KEM support will be checked along with the ciphersuites */
+        if (hpkePubkeyLen != wc_HpkeKemGetEncLen(workingConfig->kemId)) {
             ret = BUFFER_E;
             break;
```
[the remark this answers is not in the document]
sebastian-carpenter: This is not an issue. It will be checked by EchConfigGetSupportedCipherSuite. This function will verify the KEM and one of the KDF/AEAD pairs is supported.

src/tls13.c
```diff
@@ -5694,6 +5710,9 @@ int DoTls13ServerHello(WOLFSSL* ssl, const byte* input, word32* inOutIdx,
     /* check for acceptConfirmation */
     if (ssl->echConfigs != NULL && !ssl->options.disableECH) {
         args->echX = TLSX_Find(ssl->extensions, TLSX_ECH);
+        if (args->echX == NULL || args->echX->data == NULL)
+            return WOLFSSL_FATAL_ERROR;
+
```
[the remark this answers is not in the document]
sebastian-carpenter: don't think this is an issue

sebastian-carpenter:
Jenkins retest this please.

src/tls.c
```diff
@@ -14155,7 +14155,7 @@ static int TLSX_ECH_ExpandOuterExtensions(WOLFSSL* ssl, WOLFSSL_ECH* ech,
     if (ret == 0) {
         XFREE(ech->innerClientHello, heap, DYNAMIC_TYPE_TMP_BUFFER);
         ech->innerClientHello = newInnerCh;
```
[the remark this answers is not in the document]
sebastian-carpenter: This follows from the previous check I added on the client side. Now both the server and client have the overflow check.

SparkiDev (approved):

JacobBarthelmeh (approved):

PR #10099, opened 2026-03-29
Fix PKCS#7 verification of signature created with OpenSSL

haxtibal:
# Description

RFC 8017 [hardcodes DER serialization samples](https://www.rfc-editor.org/rfc/rfc8017#section-9.2) of DigestInfo, where the parameter part is always NULL (05 00) for known hash algorithm. This value does thus *not* depend on SignerInfo.digestAlgorithm.parameters. Starting with 75c3030554db798f4536d370842dd16e9b2e6866, wolfSSL wrongly assumed and implemented such a dependency.

This non-conformance caused an interoperability bug with OpenSSL: A signature created with `openssl cms` could not be verified in WolfSSL. OpenSSL correctly leaves SignerInfo.digestAlgorithm.parameters absent and adds explicit NULL to DigestInfo. WolfSSL saw the absence and wrongly inferred DigestInfo would also have no explicit NULL - but it has - leading to size mismatch.

Fix it by constructing the expected DigestInfo always with NULL (05 00).

The PR further adds an optional unit test to verify externally provided DER signature samples. A new CI job uses this to verify signatures created with OpenSSL and GnuTLS.

# Testing

This was originally [found, discussed and tested](https://salsa.debian.org/python-team/packages/swugenerator/-/commit/0003563013b25d60065188be095f2a5aa8423ede) at Debian via a swugenerator autopkgtest. Pipelines in my wolfSSL fork run the newly added tests:
- [failed](https://github.com/haxtibal/wolfssl/actions/runs/23711954141) without the proposed fix
- [pass](https://github.com/haxtibal/wolfssl/actions/runs/23711983895) with the proposed fix

# Checklist

 - [x] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/wolfcrypt/src/pkcs7.c
+++ b/wolfcrypt/src/pkcs7.c
@@ -4682,8 +4682,10 @@ static int wc_PKCS7_EcdsaVerify(wc_PKCS7* pkcs7, byte* sig, int sigSz,
 #endif /* HAVE_ECC */
 
 
-/* build SignedData digest, both in PKCS#7 DigestInfo format and
- * as plain digest for CMS.
+/* Build CMS SignedData digest (RFC 5652 sec 5.4) wrapped in a DigestInfo
+ * structure (RFC 8017 sec 9.2) as for example used by RSA PKCS#1 v1.5
+ * verification. Also returns a pointer to the inner plain digest for
+ * algorithms that do not wrap digist in DigestInfo, e.g. ECDSA and RSA-PSS.
  *
  * pkcs7          - pointer to initialized PKCS7 struct
  * signedAttrib   - signed attributes
@@ -4779,9 +4781,8 @@ static int wc_PKCS7_BuildSignedDataDigest(wc_PKCS7* pkcs7, byte* signedAttrib,
         }
     }
 
-    /* Set algoID, match whatever was input to match either NULL or absent */
-    algoIdSz = SetAlgoIDEx(pkcs7->hashOID, algoId, oidHashType,
-                            0, pkcs7->hashParamsAbsent);
+    /* parameters SHALL be NULL per RFC 8017 section 9.2, not absent */
+    algoIdSz = SetAlgoID(pkcs7->hashOID, algoId, oidHashType, 0);
 
     digestStrSz = SetOctetString(hashSz, digestStr);
     digestInfoSeqSz = SetSequence(algoIdSz + digestStrSz + hashSz,


dgarske:
@haxtibal please resolve merge conflicts. For external contributions we need a signed contributor agreement. You can email support at wolfssl dot com to get that process started. Please reference the PR and tell us more about your project. Thanks, David Garske, wolfSSL

haxtibal:
> @haxtibal please resolve merge conflicts. For external contributions we need a signed contributor agreement. You can email support at wolfssl dot com to get that process started. Please reference the PR and tell us more about your project. Thanks, David Garske, wolfSSL

@dgarske Thanks, done. Let me know if something's missing.

embhorn:
@haxtibal has been accepted as a contributor

tests/api/test_pkcs7.c
```diff
@@ -5000,3 +5000,71 @@ int test_wc_PKCS7_VerifySignedData_IndefLenOOB(void)
...
+ *   configure --with-pkcs7-test-signed-data=openssl_cms.der,openssl_smime.der
+ */
+int test_wc_PKCS7_VerifySignedData_interop(void)
+{
+    EXPECT_DECLS;
+    char pathsTokenBuf[FOURK_BUF];
+    char* nextPath;
+    char* tokState = NULL;
+
+    ExpectTrue(XSTRLEN(PKCS7_TEST_SIGNED_DATA_FILES) < sizeof(pathsTokenBuf));
```
dgarske: **🔵 [Low] Missing null-termination guarantee after XSTRNCPY in interop test**
*💡 SUGGEST*  `bug`

`XSTRNCPY` maps to `strncpy`, which does NOT null-terminate the destination if the source length equals or exceeds the buffer size. The `ExpectTrue` length check on line 5057 will flag the failure in the test framework, but the `XSTRNCPY` and `XSTRTOK` calls on lines 5058-5060 are NOT guarded by `EXPECT_SUCCESS()` and always execute. If `PKCS7_TEST_SIGNED_DATA_FILES` is >= `FOURK_BUF` (4096), `pathsTokenBuf` will not be null-terminated, and `XSTRTOK` could read past the buffer. While practically unreachable (these are compile-time paths from configure), defensive null-termination is good practice.

**Suggestion:**
```suggestion
ExpectTrue(XSTRLEN(PKCS7_TEST_SIGNED_DATA_FILES) < sizeof(pathsTokenBuf));
XSTRNCPY(pathsTokenBuf, PKCS7_TEST_SIGNED_DATA_FILES, sizeof(pathsTokenBuf));
pathsTokenBuf[sizeof(pathsTokenBuf) - 1] = '\0';
```

wolfcrypt/src/pkcs7.c
```diff
@@ -4779,9 +4779,7 @@ static int wc_PKCS7_BuildSignedDataDigest(wc_PKCS7* pkcs7, byte* signedAttrib,
         }
```
dgarske, line 4781: **🟡 [Medium] Signing-path DigestInfo still uses hashParamsAbsent — potential interop asymmetry**
*💡 SUGGEST*  `question`

The PR correctly fixes the verification path (`wc_PKCS7_BuildSignedDataDigest`) to always encode DigestInfo with NULL params per RFC 8017 Section 9.2. However, the signing-path `wc_PKCS7_BuildDigestInfo` (line 2506) and related calls at lines 3158, 3191, and 3267 still use `SetAlgoIDEx` with `pkcs7->hashParamsAbsent`, meaning wolfSSL could produce DigestInfo with absent params when signing. Per RFC 8017, the DER-encoded DigestInfo should always include explicit NULL (05 00). While this is pre-existing behavior not introduced by this PR, the asymmetry could mean wolfSSL-generated signatures fail verification in strict RFC 8017 implementations. Consider whether the signing path should also be fixed for full RFC compliance.

**Recommendation:** Consider a follow-up PR to also fix the signing path to always use NULL params in DigestInfo per RFC 8017. The SignerInfo digestAlgorithm fields (lines 3158, 3267) may legitimately use absent params per CMS/RFC 5652, but the DigestInfo construction for RSA PKCS#1 v1.5 should always use NULL.

tests/api/test_pkcs7.c
```diff
@@ -5000,3 +5000,71 @@ int test_wc_PKCS7_VerifySignedData_IndefLenOOB(void)
...
+    XFREE(signedData, NULL, DYNAMIC_TYPE_TMP_BUFFER);
+    wc_PKCS7_Free(pkcs7);
+    return EXPECT_RESULT();
+}
+
+/* Verify CMS SignedData DER files created with third party tools.
+ *
+ * This test verifies signatures created with external tools can be verified by
+ * WolfSSL. Standards allow some variance when creating CMS signatures. The
+ * test is intended to catch incompatibility from subtle differences in
```
dgarske, line 5044: **⚪ [Info] Test helper comment says 'WolfSSL' instead of 'wolfSSL'**
*🔧 NIT*  `style`

The comment on line 5044 uses 'WolfSSL' instead of the project's standard capitalization 'wolfSSL'.

**Suggestion:**
```suggestion
* This test verifies signatures created with external tools can be verified by
 * wolfSSL. Standards allow some variance when creating CMS signatures.
```

tests/api/test_pkcs7.c
haxtibal: Good point, and even better

> these are compile-time paths from configure

means we can check and fail at compile time with a static assert. Opted to do that. Please tell if you prefer the literal suggestion.

wolfcrypt/src/pkcs7.c
haxtibal: Ack. And related, SignerInfo.digestAlgorithm.parameters (absent vs. NULL) should depend on the hash algo IIUC. Will do that in a separate PR.

haxtibal:
Regarding recently failed [Test interop with mbedtls](https://github.com/wolfSSL/wolfssl/actions/runs/24103192092/job/70710546447?pr=10099) and [Test interop with nss](https://github.com/wolfSSL/wolfssl/actions/runs/24072994836/job/70312705126): I don't know what the message `Exiting as fail-on-cache-miss is set` is about, but it sounds unrelated to this PR.

dgarske:
> Regarding recently failed [Test interop with mbedtls](https://github.com/wolfSSL/wolfssl/actions/runs/24103192092/job/70710546447?pr=10099) and [Test interop with nss](https://github.com/wolfSSL/wolfssl/actions/runs/24072994836/job/70312705126): I don't know what the message `Exiting as fail-on-cache-miss is set` is about, but it sounds unrelated to this PR.

This is not related to your PR. Fixed in https://github.com/wolfSSL/wolfssl/pull/10202

wolfcrypt/src/pkcs7.c
```diff
@@ -4779,9 +4779,7 @@ static int wc_PKCS7_BuildSignedDataDigest(wc_PKCS7* pkcs7, byte* signedAttrib,
...
     }
 
-    /* Set algoID, match whatever was input to match either NULL or absent */
-    algoIdSz = SetAlgoIDEx(pkcs7->hashOID, algoId, oidHashType,
-                            0, pkcs7->hashParamsAbsent);
+    algoIdSz = SetAlgoID(pkcs7->hashOID, algoId, oidHashType, 0);
```
dgarske, line 4785: Could we add a comment about why this was changed... openssl compatibility or RFC compliance?

.github/workflows/pkcs7-interop.yml
```diff
@@ -0,0 +1,55 @@
...
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install 3rd party PKCS#7 tools
+        run: sudo apt-get install -y openssl gnutls-bin
+
+      - name: Sign with 3rd party PKCS#7 tools
+        run: |
+          echo -n "pkcs7 interop test" > $RUNNER_TEMP/content.bin
+          openssl req -x509 -newkey rsa:2048 -keyout $RUNNER_TEMP/key.pem \
```
dgarske: Could/should we add testing for other algorithm combinations? I only see RSA2048-SHA256 being tested.

wolfcrypt/src/pkcs7.c
haxtibal: Primary goal is RFC compliance. OpenSSL compatibility is the consequence of RFC compliance. Hope the extended comments make it more clear.

.github/workflows/pkcs7-interop.yml
haxtibal: Added all combination from OpenSSL-cms/OpenSSL-smime/GnuTLS-p7 x RSA/ECDSA x SHA-1/256/384/512. Let me know if you miss something interesting, should be easy to add.

PR #10104, opened 2026-03-30
Cast to unsigned prior to shift to avoid UB in SLH-DSA

padelsbach:
# Description

`1 << 31` is undefined behavior in C, per spec, even though compilers may set the MSB as expected. This is possible in `slhdsakey_set_ha_from_md`.

Fixes F-1391

# Testing

How did you test?

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


douzzer:
retest this please
(another cert3389 test failure)

PR #10109, opened 2026-03-31
Fix GetLength return value check in ASN1_INTEGER functions

julek-wolfssl:
Change GetLength() return check from > 0 to >= 0 in wolfSSL_ASN1_INTEGER_get_length
and wolfSSL_ASN1_INTEGER_get0_data. GetLength returns the decoded length (≥ 0) on
success and negative error codes on failure, so checking > 0 incorrectly excluded
zero-length values, making the DER-stripping logic dead code.


julek-wolfssl:
retest this please ocsp-stapling failure

padelsbach (approved):

douzzer (approved):

PR #10113, opened 2026-03-31
Zeroize heap buffer after ML-DSA signing

Frauschi:
Ensure that the heap buffer used (among others) to store sensitive data during ML-DSA signing is zeroized before freeing the memory.

Follow-up for zd21464 and #10100


Frauschi:
Jenkins retest this please

PR #10115, opened 2026-04-01
Fix multiple bugs in OCSP implementation

julek-wolfssl:
- wolfSSL_i2d_OCSP_REQUEST_bio: save/restore pointer before i2d call
  that advances it, preventing BIO_write from wrong offset and heap
  corruption on free
- wolfSSL_d2i_OCSP_RESPONSE: remove (unsigned char) cast that truncated
  pointer advance to 8 bits, breaking responses larger than 255 bytes
- wolfSSL_OCSP_CERTID_dup: deep-copy CertStatus to prevent double-free
  when both original and duplicate are freed
- wolfSSL_i2d_OCSP_RESPONSE: add NULL check on response parameter
- wolfSSL_i2d_OCSP_REQUEST: advance *data pointer per i2d convention
- FreeOCSP: NULL-check ocsp->cm before dereferencing for heap
- Fix WOLFSSL_LEAVE strings to match actual function names in
  wc_CheckCertOcspResponse, GetOcspEntry, GetOcspStatus,
  CheckOcspResponse, CheckOcspRequest

Add test for CERTID dup (double-free confirmed under ASAN without fix)
and pointer advancement assertions for d2i_OCSP_RESPONSE callers.

Reported in: ZD21469

### DIFF
--- a/src/ocsp.c
+++ b/src/ocsp.c
@@ -89,7 +89,7 @@ int wc_CheckCertOcspResponse(WOLFSSL_OCSP *ocsp, DecodedCert *cert,
     ocspRequest = (OcspRequest*)XMALLOC(sizeof(OcspRequest), NULL,
                                                        DYNAMIC_TYPE_TMP_BUFFER);
     if (ocspRequest == NULL) {
-        WOLFSSL_LEAVE("CheckCertOCSP", MEMORY_ERROR);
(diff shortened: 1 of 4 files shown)


julek-wolfssl:
retest this please no history

julek-wolfssl:
Retest this please unrelated error

douzzer (approved):

PR #10117, opened 2026-04-01
Multiple DTLS and TLS focused fixes.

gasbytes:
dtls13.c:
- Fix wrong return value in Dtls13SendFragmentedInternal error path (return outputSz instead of recordLength)
- Fix incomplete bounds check in Dtls13SendFragmented to account for DTLS_HANDSHAKE_HEADER_SZ
- Fix wrong WOLFSSL_ENTER trace string in Dtls13EpochCopyKeys

tls13.c:
- Remove wrong (byte) cast on cookie->len passed to TlsCheckCookie
- Add missing bounds check on PSK identityLen in SetupPskKey before copying to client_identity
- Fix data race on static header array in ExpectedResumptionSecret
- Add defensive underflow check in EncryptTls13 for consistency with DecryptTls13
- Fix wrong return variable in DTLS 1.3 Finished send error path (return dtlsRet instead of ret)
- Add missing SM3 case and default in Tls13_Exporter hash switch to prevent NULL dereference
- Initialize *outSz to 0 in wolfSSL_write_early_data to match wolfSSL_read_early_data
- Add bounds check for bindersLen against helloSz in CheckPreSharedKeys
- Fix resource leak and hash state corruption in ExpectedResumptionSecret error paths
- Fix memory leak of rsaSigBuf in dual-alg RSA+RSA CertificateVerify
- Guard against word32 underflow in inputLength - HANDSHAKE_HEADER_SZ in DoTls13HandShakeMsg
- Fix swapped side parameter in DeriveFinishedSecret for server-side Finished processing
- Fix no_mac fall-through in ssl_handshake_md to return NULL instead of wrong digest
- Fix strict aliasing violation in FindPsk PSK key size check
- Remove duplicate !ssl->options.dtls check in TLS 1.3 middlebox compat condition

tests:
- Add regression tests for wolfSSL_write_early_data outSz initialization and DTLS 1.3 Finished send error propagation

### DIFF
--- a/src/dtls13.c
+++ b/src/dtls13.c
@@ -1040,7 +1040,7 @@ static int Dtls13SendFragmentedInternal(WOLFSSL* ssl)
             fragLength + DTLS_HANDSHAKE_HEADER_SZ, isEncrypted);
         if (outputSz < 0) {
             Dtls13FreeFragmentsBuffer(ssl);
-            return recordLength;
+            return outputSz;
         }
 
         ret = CheckAvailableSize(ssl, outputSz);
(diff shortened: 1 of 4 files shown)


JacobBarthelmeh:
Retest this please Jenkins -- multi-test failure, can no longer access the results.

douzzer (approved):

PR #10120, opened 2026-04-01
20260331-wolfcrypt-Wcast-qual

douzzer:
Fixes for `-Wcast-qual` hygiene in wolfCrypt.

`.github/workflows/wolfCrypt-Wconversion.yml`: Add `-Wcast-qual` to all scenarios.

`wolfssl/wolfcrypt/signature.h`, `wolfcrypt/src/signature.c`, `doc/dox_comments/header_files/signature.h`:

  remove incorrect const qualifier on the key argument in

  * `wc_SignatureVerifyHash()`
  * `wc_SignatureVerify()`
  * `wc_SignatureGenerateHash()`
  * `wc_SignatureGenerateHash_ex()`
  * `wc_SignatureGenerate()`
  * `wc_SignatureGenerate_ex()`

  This fixes UB code patterns throughout `signature.c`.  `key` is inherently
  accessed readwrite by the underlying low level crypto.  Fortunately, wolfCrypt
  has no APIs/methods to allow actual `const` MPI key objects, therefore these
  seeming breaking API changes can't actually break any users.

globally:

  * Add `const` qualifiers to all `struct` pointer members that are assigned values
    computed from `const` pointers.

  * Add `const` qualifiers to intermediate casts for accessors and read-only
    dereference constructs, as needed for `-Wcast-qual` hygiene, e.g. for a macro
    `GET_U16(a)`, use `(*(const word16*)(a))` rather than `(*(word16*)(a))`.

  * Add `const` qualifiers to internal declarations, and remove illegal casts, as
    needed for `-Wcast-qual` hygiene.

  * Add missing `const` qualifiers to all casts for argument, operand, and
    assignment type agreement, as needed for `-Wcast-qual` hygiene, e.g.
    `*data = (const byte*)dataASN->data.ref.data` rather than
    `*data = (byte*)dataASN->data.ref.data`.

`wolfssl/wolfcrypt/asn.h`, `wolfssl/wolfcrypt/asn_public.h`, `wolfcrypt/src/asn.c`, `wolfcrypt/src/asn_orig.c`:

  * Add additional lifecycle management for object members that are only sometimes locally allocated:

    `DNS_entry.nameStored`
    `DNS_entry.ipStringStored`
    `DNS_entry.ridStringStored`

`wolfssl/wolfcrypt/types.h`: add `WC_BARRIER()` macro -- a portable construct that
   prevents compiler optimizers from reordering operations across the barrier.

`wolfssl/wolfcrypt/blake2-impl.h`, `wolfcrypt/src/blake2s.c`, `wolfcrypt/src/blake2b.c`:

  * In `blake2b_init()`, `blake2b_init_key()`, `blake2s_init()`, and
    `blake2s_init_key()`, refactor `blake2b_param` initialization using `WC_BARRIER()`
    (fixes `volatile` abuse that triggered `-Wcast-qual`).

  * Remove the residual and unused `WOLFSSL_BLAKE2[BS]_INIT_EACH_FIELD` code.

`wolfcrypt/src/ecc.c` and `wolfssl/wolfcrypt/ecc.h`:

  Remove incorrect `const` qualifier on curve arg to `wc_ecc_free_curve()` (internal function).

Tested with
```
wolfssl-multi-test.sh ...
pr-check
'.*Wconversion.*'
```
with `-Wcast-qual` added to the non-FIPS `.*Wconversion.*` scenarios.

### DIFF
--- a/wolfcrypt/src/blake2b.c
+++ b/wolfcrypt/src/blake2b.c
@@ -120,37 +120,37 @@ int blake2b_init_param( blake2b_state *S, const blake2b_param *P )
 
 int blake2b_init( blake2b_state *S, const byte outlen )
 {
-  blake2b_param P;
+  volatile blake2b_param P;
 
   if ( ( !outlen ) || ( outlen > BLAKE2B_OUTBYTES ) ) return BAD_FUNC_ARG;
 
-  XMEMSET(&P, 0, sizeof(P));
+  XMEMSET((void *)(wc_ptr_t)&P, 0, sizeof(P));
   WC_BARRIER();
   P.digest_length = outlen;
   P.fanout        = 1;
   P.depth         = 1;
 
-  return blake2b_init_param(S, &P);
+  return blake2b_init_param(S, (const blake2b_param *)(wc_ptr_t)&P);
 }
 
 int blake2b_init_key( blake2b_state *S, const byte outlen, const void *key,
                       const byte keylen )
 {
   int ret = 0;
-  blake2b_param P;
+  volatile blake2b_param P;
 
   if ( ( !outlen ) || ( outlen > BLAKE2B_OUTBYTES ) ) return BAD_FUNC_ARG;
 
   if ( !key || !keylen || keylen > BLAKE2B_KEYBYTES ) return BAD_FUNC_ARG;
 
-  XMEMSET( &P, 0, sizeof( P ) );
+  XMEMSET( (void *)(wc_ptr_t)&P, 0, sizeof( P ) );
   WC_BARRIER();
   P.digest_length = outlen;
   P.key_length    = keylen;
   P.fanout        = 1;
   P.depth         = 1;
 
-  ret = blake2b_init_param(S, &P);
+  ret = blake2b_init_param(S, (const blake2b_param *)(wc_ptr_t)&P);
   if ( ret < 0 ) return ret;
 
   {
--- a/wolfcrypt/src/blake2s.c
+++ b/wolfcrypt/src/blake2s.c
@@ -117,17 +117,17 @@ int blake2s_init_param( blake2s_state *S, const blake2s_param *P )
 
 int blake2s_init( blake2s_state *S, const byte outlen )
 {
-  blake2s_param P;
+  volatile blake2s_param P;
 
   if ( ( !outlen ) || ( outlen > BLAKE2S_OUTBYTES ) ) return BAD_FUNC_ARG;
 
-  XMEMSET( &P, 0, sizeof( P ) );
+  XMEMSET( (void *)(wc_ptr_t)&P, 0, sizeof( P ) );
   WC_BARRIER();
   P.digest_length = outlen;
   P.fanout        = 1;
   P.depth         = 1;
 
-  return blake2s_init_param( S, &P );
+  return blake2s_init_param( S, (const blake2s_param *)(wc_ptr_t)&P );
 }
 
 
@@ -135,20 +135,20 @@ int blake2s_init_key( blake2s_state *S, const byte outlen, const void *key,
                       const byte keylen )
 {
   int ret = 0;
-  blake2s_param P;
+  volatile blake2s_param P;
 
   if ( ( !outlen ) || ( outlen > BLAKE2S_OUTBYTES ) ) return BAD_FUNC_ARG;
 
   if ( !key || !keylen || keylen > BLAKE2S_KEYBYTES ) return BAD_FUNC_ARG;
 
-  XMEMSET( &P, 0, sizeof( P ) );
+  XMEMSET( (void *)(wc_ptr_t)&P, 0, sizeof( P ) );
   WC_BARRIER();
   P.digest_length = outlen;
   P.key_length    = keylen;
   P.fanout        = 1;
   P.depth         = 1;
 
-  ret = blake2s_init_param( S, &P );
+  ret = blake2s_init_param( S, (const blake2s_param *)(wc_ptr_t)&P );
   if (ret < 0)
       return ret;
 


wolfcrypt/src/asn_orig.c
```diff
@@ -3182,8 +3182,9 @@ static int DecodeConstructedOtherName(DecodedCert* cert, const byte* input,
             ret = MEMORY_E;
         }
         else {
-            XMEMCPY(dnsEntry->name, &input[*idx], (size_t)strLen);
-            dnsEntry->name[strLen] = '\0';
+            XMEMCPY((void *)(wc_ptr_t)dnsEntry->name, &input[*idx],
+                    (size_t)strLen);
+            ((char *)(wc_ptr_t)dnsEntry->name)[strLen] = '\0';
```
padelsbach, line 3188: Should we set `nameStored = 1` here?

wolfcrypt/src/asn_orig.c
```diff
@@ -3272,8 +3273,9 @@ static int DecodeAltNames(const byte* input, word32 sz, DecodedCert* cert)
                 return MEMORY_E;
             }
             dnsEntry->len = strLen;
-            XMEMCPY(dnsEntry->name, &input[idx], (size_t)strLen);
-            dnsEntry->name[strLen] = '\0';
+            XMEMCPY((void *)(wc_ptr_t)dnsEntry->name, &input[idx],
+                    (size_t)strLen);
+            ((char *)(wc_ptr_t)dnsEntry->name)[strLen] = '\0';
```
padelsbach, line 3280: `nameStored = 1` here too? and several other places...

wolfcrypt/src/asn_orig.c
douzzer: Yup there's actually 7 code points in `asn_orig.c` that needed that.  Already fixed, and tested using sanitizer.

wolfcrypt/src/asn_orig.c
douzzer: See above

wolfcrypt/src/asn_orig.c
padelsbach: did you push that change? i'm not seeing it in `21c656888`

wolfcrypt/src/asn_orig.c
```diff
@@ -3426,7 +3433,7 @@ static int DecodeAltNames(const byte* input, word32 sz, DecodedCert* cert)
                 WOLFSSL_MSG("\tOut of Memory");
                 return MEMORY_E;
             }
-
+            uriEntry->nameStored = 1;
```
padelsbach, line 3436: Should `nameStored` be set prior to the malloc (and potential `return`)? We haven't actually set the name yet.

wolfcrypt/src/srp.c
```diff
@@ -389,25 +389,25 @@ int wc_SrpSetParams(Srp* srp, const byte* N,    word32 nSz,
 
     /* Set k = H(N, g) */
     r = SrpHashInit(&hash, srp->type, srp->heap);
-    if (!r) r = SrpHashUpdate(&hash, (byte*) N, nSz);
+    if (!r) r = SrpHashUpdate(&hash, (const byte*) N, nSz);
```
padelsbach, line 392: Is it required to cast as a `const` in this type of call? IME, it's fine without it, but maybe it's required with `Wcast-qual`?

wolfcrypt/src/tfm.c
```diff
@@ -4540,9 +4540,9 @@ int mp_exptmod_nct (mp_int * G, mp_int * X, mp_int * P, mp_int * Y)
 
 
 /* compare two ints (signed)*/
-int mp_cmp (mp_int * a, mp_int * b)
+int mp_cmp (const mp_int * a, const mp_int * b)
```
padelsbach, line 4543: Rather than disregard `const` here, can we change `fp_cmp`, `fp_cmd_mag` and `fp_cmd_d`) to have const params?

wolfcrypt/src/srp.c
douzzer: It _is_ required -- `N` is a `const`, so the cast also needs to be `const`.  The called function is expecting a `const` argument in that position obviously.

Those changes in srp.c were purely demand-driven from compiler warnings.

wolfssl/wolfcrypt/types.h
```diff
@@ -489,6 +489,12 @@ enum {
     static WC_INLINE WARN_UNUSED_RESULT int WC_WUR_INT(int x) { return x; }
 #endif
 
+/* XFENCE() is a no-op on some targets.  WC_BARRIER() uses C89 intrinsics as an
+ * additional portable barrier.
+ */
+#define WC_BARRIER() do { volatile byte _xfence = 0; (void)_xfence; XFENCE(); \
```
padelsbach: Why not apply this to all existing uses of `XFENCE()` ?

wolfcrypt/src/asn_orig.c
douzzer: Nope -- the incumbent pattern on all these is to set up the lengths and whatnot only after&if the allocation succeeds.

As it happens, it's harmless to set the `fooStored` bit before the allocation (and I think it is in at least one spot) because `XFREE()` of a null pointer is a harmless no-op.

wolfcrypt/src/tfm.c
douzzer: No need to bikeshed it -- TFM is attic code, like asn_orig.

wolfssl/wolfcrypt/types.h
douzzer: Because it doesn't work :slightly_smiling_face: 

Proper barrier macro def incoming -- needed to fix the ARM32 UB wrong results on blake.

wolfcrypt/src/srp.c
padelsbach: This example is fine for me on gcc 13.3.0 and `Wcast-qual`. 

```
static int hash_update(const byte *data, int sz)
{
    int sum = 0;
    int i;
    for (i = 0; i < sz; i++) {
        sum += data[i];
    }
    return sum;
}

int main(void)
{
    byte buf[] = { 0x01, 0x02, 0x03, 0x04 };
    hash_update(buf, 4);
    return 0;
}
```

Maybe there are other compilers that support `Wcast-qual` that aren't happy with it?

In any case, seems mostly harmless, just a bit verbose.

wolfcrypt/src/asn_orig.c
```diff
@@ -3529,12 +3535,13 @@ static int DecodeAltNames(const byte* input, word32 sz, DecodedCert* cert)
...
+            ((char *)(wc_ptr_t)rid->name)[strLen] = '\0';
 
             if (GenerateDNSEntryRIDString(rid, cert->heap) != 0) {
                 WOLFSSL_MSG("\tOut of Memory for registered Id string");
-                XFREE(rid->name, cert->heap, DYNAMIC_TYPE_ALTNAME);
+                XFREE((void *)(wc_ptr_t)rid->name, cert->heap,
+                      DYNAMIC_TYPE_ALTNAME);
                 XFREE(rid, cert->heap, DYNAMIC_TYPE_ALTNAME);
                 return MEMORY_E;
             }
```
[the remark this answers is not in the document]
douzzer: fixed

PR #10132, opened 2026-04-05
20260404-default_rng_bank

douzzer:
`wolfcrypt/src/rng_bank.c` and `wolfssl/wolfcrypt/rng_bank.h`: add new `wc_rng_bank_default` facility:

  * `wc_rng_bank_default_set()`
  * `wc_rng_bank_default_checkout()`
  * `wc_rng_bank_default_checkin()`
  * `wc_rng_bank_default_clear()`

`wolfssl/wolfcrypt/wc_port.h`:

  * add `wolfSSL_RefInc2()` and `wolfSSL_RefDec2()` returning the atomically determined new count in the second arg, currently only implemented for full atomic-capable targets;

  * Fix type of second arg in the fallback definition of `wolfSSL_Atomic_Ptr_CompareExchange()`.

`linuxkm/lkcapi_sha_glue.c`:

  Refactor the `_REGISTER_HASH_DRBG` / `_REGISTER_HASH_DRBG_DEFAULT` facility around
  the new `wc_rng_bank_default` facility, eliminating post-init use of
  kernel-native `crypto_default_rng`, `crypto_get_default_rng()`, and
  `crypto_put_default_rng()`, and eliminating all use on kernel 7.1+ (where these
  will become unexported kernel-native statics).  With the refactor, the
  `LINUXKM_DRBG_GET_RANDOM_BYTES` facility uses only direct native wolfCrypt
  objects and calls to fulfill requests.

tested with
```
wolfssl-multi-test.sh ...
super-quick-check
.*cust-kernel.*
linuxkm-6.15-all-cryptonly-quantum-safe-fips-dev-intelasm-insmod-crypto-fuzzer-kmemleak
linuxkm-6.15-all-cryptonly-quantum-safe-fips-dev-intelasm-insmod-crypto-fuzzer-ksan
quantum-safe-wolfssl-all-crypto-only-intelasm-sp-asm-fips-dev-linuxkm-next-insmod
.*Wconversion.*
```
(with additional fixes to WolfGuard to accommodate the changes in linux-next)

See herbertx/cryptodev@bdd2cc93bfd


linuxkm/lkcapi_sha_glue.c
```diff
@@ -1999,11 +2108,36 @@ static int wc_linuxkm_drbg_cleanup(void) {
...
+            if (ret)
+                pr_err("ERROR: wc_rng_bank_default_clear in wc_linuxkm_drbg_cleanup failed: %d", ret);
+            else {
+                ret = wc_rng_bank_fini(&default_bank);
+                if (ret)
+                    pr_err("ERROR: wc_rng_bank_fini in wc_linuxkm_drbg_cleanup failed: %d", ret);
+            }
+            default_bank_inited = 0;
+        }
+#endif /* >= 7.1.0 */
```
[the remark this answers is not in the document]
douzzer: Wrong fix.  If `wc_rng_bank_default_clear()` fails, the best strategy is to leak, rather than to blindly deallocate setting the stage for a crash on null or wild pointer dereference.

wolfcrypt/src/rng_bank.c
```diff
@@ -432,6 +432,106 @@ WOLFSSL_API int wc_rng_bank_checkout(
...
+/* The default_rng_bank facility is used by the Linux kernel module as a global
+ * resource for wc_rng_bank_checkout(),
+ * wc_local_rng_bank_checkout_for_bankref(), and wc_InitRng_BankRef(), and can
+ * be similarly used by any application, to cache DRBG seeding at application
+ * startup.
+ */
+
+static struct wc_rng_bank * volatile default_rng_bank;
+
+WOLFSSL_API int wc_rng_bank_default_set(struct wc_rng_bank *bank) {
```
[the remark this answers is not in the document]
douzzer: Fixed (and then some).

wolfcrypt/src/rng_bank.c
```diff
@@ -432,6 +432,106 @@ WOLFSSL_API int wc_rng_bank_checkout(
...
+    int ret;
+    struct wc_rng_bank *cur_default_rng_bank = default_rng_bank;
+    if (bank == NULL)
+        return BAD_FUNC_ARG;
+    if (cur_default_rng_bank == NULL)
+        return BAD_STATE_E;
+    wolfSSL_RefInc(&cur_default_rng_bank->refcount, &ret);
+    if (ret == 0)
+        *bank = cur_default_rng_bank;
+    return ret;
```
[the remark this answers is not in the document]
douzzer: Wrong fix.  Contract is clear, and use of pure atomics (no mutexes) is required for kernel compatibility.

wolfcrypt/src/rng_bank.c
```diff
@@ -432,6 +432,106 @@ WOLFSSL_API int wc_rng_bank_checkout(
...
+#endif
+        return ret;
+    }
+    else
+        return BUSY_E;
+}
+
 #ifdef WC_DRBG_BANKREF
 WOLFSSL_LOCAL int wc_local_rng_bank_checkout_for_bankref(
     struct wc_rng_bank *bank,
```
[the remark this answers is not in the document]
douzzer: fixed

douzzer:
retest this please
(Java tooling glitch)

linuxkm/lkcapi_sha_glue.c
```diff
@@ -1860,6 +1913,66 @@ static int wc_linuxkm_drbg_startup(void)
...
+
+#else /* >= 7.1.0 */
+
+#ifdef CONFIG_CRYPTO_FIPS
+    if (fips_enabled) {
+        char buf[16];
+
+        ret = crypto_del_default_rng();
+        if (ret) {
+            wc_linuxkm_rng_initing_default_bank_flag = 0;
```
philljj, line 1925: Why setting `wc_linuxkm_rng_initing_default_bank_flag` to zero both on err and success?

It could be reset to 0 immediately after `crypto_del_default_rng()` if I understand the intention.

wolfssl/wolfcrypt/wc_port.h
```diff
@@ -758,13 +758,23 @@ typedef struct wolfSSL_RefWithMutex wolfSSL_Ref;
         (void)wolfSSL_Atomic_Int_FetchAdd(&(ref)->count, 1); \
         *(err) = 0;                          \
     } while(0)
+#define wolfSSL_RefInc2(ref, new_count, err) \
```
philljj, line 761: nit: is there a better name than `RefInc2` and `RefDec2` to communicate it's `XFetch` and not `FetchX` semantics.

wolfcrypt/src/rng_bank.c
```diff
@@ -247,6 +249,126 @@ WOLFSSL_API int wc_rng_bank_free(struct wc_rng_bank **ctx) {
...
+    struct wc_rng_bank *cur_default_rng_bank = NULL;
+    int new_refcount;
+
+    if (bank == NULL)
+        return BAD_FUNC_ARG;
+
+    if ((! (bank->flags & WC_RNG_BANK_FLAG_INITED)) ||
+        (wolfSSL_RefCur(bank->refcount) < 1))
+    {
+        return BAD_STATE_E;
```
philljj, line 274: Does this deserve an error or debug log msg? Or too niche/unlikely of an error to justify.

wolfcrypt/src/rng_bank.c
```diff
@@ -247,6 +249,126 @@ WOLFSSL_API int wc_rng_bank_free(struct wc_rng_bank **ctx) {
...
+    }
+
+    wolfSSL_RefInc2(&bank->refcount, &new_refcount, &ret);
+    if (ret != 0)
+        return ret;
+#ifdef WC_VERBOSE_RNG
+        if (new_refcount < 2)
```
philljj: nit: the indentation depth here looks like it belongs to the scope of the previous `if (ret != 0)` block.

wolfcrypt/src/rng_bank.c
```diff
@@ -247,6 +249,126 @@ WOLFSSL_API int wc_rng_bank_free(struct wc_rng_bank **ctx) {
...
+            "BUG: wc_rng_bank_default_set() new_refcount %d.\n", new_refcount);
+#else
+        (void)new_refcount;
+#endif
+    if (wolfSSL_Atomic_Ptr_CompareExchange((void * volatile *)&default_rng_bank, (void **)&cur_default_rng_bank, bank))
+        return 0;
+    else {
+        wolfSSL_RefDec2(&bank->refcount, &new_refcount, &ret);
+#ifdef WC_VERBOSE_RNG
+        if (new_refcount <= 0)
```
philljj, line 292: If `new_refcount < 0`, isn't that a bad error that should not be hidden behind verbose logging?

wolfcrypt/src/rng_bank.c
```diff
@@ -247,6 +249,126 @@ WOLFSSL_API int wc_rng_bank_free(struct wc_rng_bank **ctx) {
...
+WOLFSSL_API int wc_rng_bank_default_checkout(struct wc_rng_bank **bank) {
+    int ret;
+    struct wc_rng_bank *cur_default_rng_bank = default_rng_bank;
+    if (bank == NULL)
+        return BAD_FUNC_ARG;
+    if (cur_default_rng_bank == NULL)
+        return BAD_STATE_E;
+    wolfSSL_RefInc(&cur_default_rng_bank->refcount, &ret);
+    if (ret == 0)
+        *bank = cur_default_rng_bank;
```
philljj, line 315: nit: Should it set `*bank = NULL` on error or at beginning after `bank == NULL` check?

Otherwise caller's ` struct wc_rng_bank *ctx` will be left uninitialized, might be accidental invalid pointer use.

wolfcrypt/src/rng_bank.c
```diff
@@ -247,6 +249,126 @@ WOLFSSL_API int wc_rng_bank_free(struct wc_rng_bank **ctx) {
...
+ * application shutdown to avoid races with wc_rng_bank_default_checkout(), and
+ * must be called before wc_rng_bank_fini() on a bank previously passed to
+ * wc_rng_bank_default_set().
+ */
+WOLFSSL_API int wc_rng_bank_default_clear(struct wc_rng_bank *bank) {
+    if ((bank != default_rng_bank) || (bank == NULL))
+        return BAD_FUNC_ARG;
+    if ((default_rng_bank == NULL) /* not actually possible */ ||
+        (! (default_rng_bank->flags & WC_RNG_BANK_FLAG_INITED)) ||
+        (wolfSSL_RefCur(bank->refcount) < 2))
```
philljj: potentially dumb question: Why is refcount < 2 always a bad state error?

wolfcrypt/src/rng_bank.c
```diff
@@ -712,13 +906,23 @@ WOLFSSL_API int wc_BankRef_Release(WC_RNG *rng)
...
+#endif
+
     if ((bank == NULL) ||
-        (! (bank->flags & WC_RNG_BANK_FLAG_INITED)) ||
         (rng == NULL))
     {
         return BAD_FUNC_ARG;
     }
 
+    if ((! (bank->flags & WC_RNG_BANK_FLAG_INITED)) ||
```
philljj, line 917: nit: May not be worth it, but `*bank->flags & WC_RNG_BANK_FLAG_INITED` pattern appears a lot. A short define might be more readable or less error-prone.

linuxkm/lkcapi_sha_glue.c
```diff
@@ -997,7 +997,11 @@ struct wc_swallow_the_semicolon
 #include <wolfssl/wolfcrypt/random.h>
 #include <wolfssl/wolfcrypt/rng_bank.h>
 
-static volatile int wc_linuxkm_drbg_init_tfm_disable_vector_registers = 0;
+#ifndef WC_RNG_BANK_DEFAULT_SUPPORT
+    #error LINUXKM_LKCAPI_REGISTER_HASH_DRBG requires WC_RNG_BANK_DEFAULT_SUPPORT.
+#endif
+
+static volatile int wc_linuxkm_rng_initing_default_bank_flag = 0;
```
philljj, line 1004: Is `static volatile` enough for this? No other sync mechanisms needed?

philljj (changes requested):
Overall looks good, just have some suggestions and nits.

linuxkm/lkcapi_sha_glue.c
douzzer: It needs to still be asserted when `__crypto_stdrng_get_bytes()` is called -- that's what instantiates our persistent DRBG.

wolfssl/wolfcrypt/wc_port.h
douzzer: The xfetch-vs-fetchx distinction is irrelevant because the preexisting APIs don't expose the result to the caller in the first place.  The `isZero` flag actually has ad hoc xfetch semantics -- it's set to `(__prev == 1)`.

IIRC the history is that the atomic facility gained the xfetch methods only recently, otherwise `wolfSSL_RefDec()` would have been using xfetch and a zeroness test for `isZero`.

wolfcrypt/src/rng_bank.c
douzzer: Way too nichey -- fine grained resolution around bugs like that is what `--enable-debug-trace-errcodes=backtrace` is for.

wolfcrypt/src/rng_bank.c
douzzer: fixed.  Yikes, clang-tidy should have caught that!  Good catch.

wolfcrypt/src/rng_bank.c
douzzer: This func is right in the library code and we don't want it being unconditionally chatty, or unconditionally relying on `WOLFSSL_DEBUG_PRINTF()` (it's missing on some deeply embedded targets).  In other spots in the library, refcount dementia is reported using `WOLFSSL_MSG`, which is a no-op unless `defined(DEBUG_WOLFSSL)`.  See e.g. logging around `git grep -n WOLFSSL_REFCNT_ERROR_RETURN`.

Btw `WC_VERBOSE_RNG` is on by default in the kernel, and only off if the user defines `WC_NO_VERBOSE_RNG`.

wolfcrypt/src/rng_bank.c
douzzer: No need to do anything with `*bank` if error is returned.  Basic API contract stuff.

wolfcrypt/src/rng_bank.c
douzzer: Because the refcount is initialized to 1 when the object is instantiated, and incremented to at least 2 when `wc_rng_bank_default_set()` is called on it, so if it's `< 2` when `wc_rng_bank_default_clear()` that means something went wrong -- API misuse or memory corruption.

wolfcrypt/src/rng_bank.c
douzzer: Meh, good idea for the future -- for now I didn't want to change any lines that didn't need changing.

linuxkm/lkcapi_sha_glue.c
douzzer: Yes, definitely all we need, because it's actually just the thread communicating with itself.  It's not a thread synchronization mechanism.

wolfcrypt/src/rng_bank.c
douzzer: Oh but hang on, the safer thing to do here is to NULL the `default_rng_bank` pointer even (especially?) if the refcount is bogus.  I'll tweak that.

douzzer: The best fix for this was to simply omit the refcount check in the prescreen, and leave detection+reporting of bad counts to the existing `WC_VERBOSE_RNG` code after the decrement.
```
diff --git a/wolfcrypt/src/rng_bank.c b/wolfcrypt/src/rng_bank.c
index fd53a50f6..1366a2815 100644
--- a/wolfcrypt/src/rng_bank.c
+++ b/wolfcrypt/src/rng_bank.c
@@ -342,8 +342,7 @@ WOLFSSL_API int wc_rng_bank_default_clear(struct wc_rng_bank *bank) {
     if ((bank != default_rng_bank) || (bank == NULL))
         return BAD_FUNC_ARG;
     if ((default_rng_bank == NULL) /* not actually possible */ ||
-        (! (default_rng_bank->flags & WC_RNG_BANK_FLAG_INITED)) ||
-        (wolfSSL_RefCur(bank->refcount) < 2))
+        (! (default_rng_bank->flags & WC_RNG_BANK_FLAG_INITED)))
     {
         return BAD_STATE_E;
     }
```

douzzer:
retest this please
("FAIL scripts/ocsp-stapling2.test" and Windows FIPS glitches)

douzzer:
retest this please
```
wolfSSL error: tcp bind failed: Address already in use
wolfSSL error: SSL_accept failed
pid 79853 for port 61723 exited before creating ready file.  bailing...
[...]
FAIL scripts/ocsp-stapling2.test (exit status: 1)
```

philljj (approved):

philljj:
note: last ubuntu-macos runner stuck.

PR #10136, opened 2026-04-06
Fenrir fixes

JeremiahM37:
Fixes F-1916, F-1907, F-1911, F-1915, F-1909, F-1913, F-1914
                                                                                                                                                                                                                                                                                                                           
  - Fix NULL drbg pointer dereference in wc_RNG_DRBG_Reseed on non-Intel platforms when HAVE_INTEL_RDSEED/HAVE_INTEL_RDRAND are not defined
  - Save key->heap before wc_FreeRsaKey in wc_DeleteRsaKey to prevent use-after-free on the heap hint                                                                                                                                                                                                                      
  - Add NULL parameter checks to Base64_Decode, Base64_Decode_nonCT, and DoBase64_Encode matching existing Base16_Decode pattern                                                                                                                                                                                           
  - Reject non-block-aligned input with BAD_LENGTH_E in RC2, Camellia, and DES/DES3 CBC encrypt/decrypt instead of silently truncating

### DIFF
--- a/tests/api/test_des3.c
+++ b/tests/api/test_des3.c
@@ -165,11 +165,15 @@ int test_wc_Des3_CbcEncryptDecrypt(void)
     ExpectIntEQ(wc_Des3_CbcDecrypt(&des, plain, NULL, 24),
         WC_NO_ERR_TRACE(BAD_FUNC_ARG));
 
-    /* non-block-aligned input rejected with BAD_LENGTH_E */
+#ifndef HAVE_FIPS
+    /* non-block-aligned input rejected with BAD_LENGTH_E.
+     * FIPS builds use the FIPS-certified DES3 implementation which does not
+     * have this check, so skip the test for FIPS. */
     ExpectIntEQ(wc_Des3_CbcEncrypt(&des, cipher, vector, DES_BLOCK_SIZE - 1),
         WC_NO_ERR_TRACE(BAD_LENGTH_E));
     ExpectIntEQ(wc_Des3_CbcDecrypt(&des, plain, cipher, DES_BLOCK_SIZE - 1),
         WC_NO_ERR_TRACE(BAD_LENGTH_E));
+#endif
 
     wc_Des3Free(&des);
 #endif


wolfcrypt/src/port/caam/wolfcaam_fsl_nxp.c
```diff
@@ -176,6 +178,8 @@ static int wc_CAAM_CommonHash(caam_handle_t* hndl, caam_hash_ctx_t *ctx,
             /* input not aligned */
             tmpOut = (byte*)XMALLOC(sz + CAAM_BUFFER_ALIGN, NULL,
                 DYNAMIC_TYPE_TMP_BUFFER);
+            if (tmpOut == NULL)
+                return MEMORY_E;
             alignedOut = tmpOut + (CAAM_BUFFER_ALIGN -
                 ((wc_ptr_t)tmpOut % CAAM_BUFFER_ALIGN));
```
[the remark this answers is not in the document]
JeremiahM37: False positive — tmpIn is allocated and freed within the first if (in != NULL && inSz > 0) block

wolfcrypt/src/coding.c
```diff
@@ -474,6 +480,9 @@ static int DoBase64_Encode(const byte* in, word32 inLen, byte* out,
     word32 outSz = (inLen + 3 - 1) / 3 * 4;
```
dgarske: **🔵 [Low] DoBase64_Encode NULL check placed after arithmetic on inLen**
*🔧 NIT*  `style`

The new NULL check for `in` at line 483 is placed after lines 480-481 which compute `outSz` and `addSz` from `inLen`. While this is harmless (pure arithmetic, no pointer dereference), placing the NULL check before the arithmetic would be slightly cleaner and consistent with the check order in `Base64_Decode` and `Base64_Decode_nonCT` where the NULL checks are the very first executable statements.

**Suggestion:**
```suggestion
Move the `if (in == NULL && inLen > 0)` check before the `outSz`/`addSz` computation, or leave as-is since it's functionally correct.
```

JeremiahM37: The if (in == NULL && inLen > 0) check is now placed before the outSz / addSz arithmetic setup in DoBase64_Encode.

JeremiahM37:
Jenkins retest this please

PR #10138, opened 2026-04-06
Use size_t in wolfSSL_strnstr and reject negative indices in mp_get_digit

padelsbach:
# Description

Fixes zd21539

# Testing

Manual review

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/wolfcrypt/src/wc_port.c
+++ b/wolfcrypt/src/wc_port.c
@@ -4265,9 +4265,9 @@ time_t stm32_hal_time(time_t *t1)
 
 #if (!defined(WOLFSSL_LEANPSK) && !defined(STRING_USER)) || \
     defined(USE_WOLF_STRNSTR)
-char* wolfSSL_strnstr(const char* s1, const char* s2, unsigned int n)
+char* wolfSSL_strnstr(const char* s1, const char* s2, size_t n)
 {
-    unsigned int s2_len = (unsigned int)XSTRLEN(s2);
+    size_t s2_len = XSTRLEN(s2);
 
     if (s2_len == 0)
         return (char *)(wc_ptr_t)s1;
--- a/wolfcrypt/src/wolfmath.c
+++ b/wolfcrypt/src/wolfmath.c
@@ -98,10 +98,10 @@ int mp_get_digit_count(const mp_int* a)
 
 mp_digit mp_get_digit(const mp_int* a, int n)
 {
-    if (a == NULL)
+    if (a == NULL || n < 0)
         return 0;
 
-    return (n < 0 || (unsigned int)n >= (unsigned int)a->used) ? 0 : a->dp[n];
+    return ((unsigned int)n >= (unsigned int)a->used) ? 0 : a->dp[n];
 }
 
 #if defined(HAVE_ECC) || defined(WOLFSSL_MP_COND_COPY)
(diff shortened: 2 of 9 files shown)


wolfcrypt/src/wolfmath.c
```diff
@@ -98,10 +98,10 @@ int mp_get_digit_count(const mp_int* a)
 
 mp_digit mp_get_digit(const mp_int* a, int n)
 {
-    if (a == NULL)
+    if (a == NULL || n < 0)
```
padelsbach, line 101: No logic change, just makes the code more analyzer friendly per the ZD ticket.

padelsbach:
jenkins retest this please

wolfcrypt/src/wc_port.c
```diff
@@ -4234,7 +4234,7 @@ time_t stm32_hal_time(time_t *t1)
     defined(USE_WOLF_STRNSTR)
 char* wolfSSL_strnstr(const char* s1, const char* s2, unsigned int n)
 {
-    unsigned int s2_len = (unsigned int)XSTRLEN(s2);
+    size_t s2_len = (unsigned int)XSTRLEN(s2);
```
SparkiDev: change cast of XSTRLEN to size_t or remove cast if not needed anymore.

wolfcrypt/src/wc_port.c
```diff
@@ -4267,7 +4267,7 @@ time_t stm32_hal_time(time_t *t1)
     defined(USE_WOLF_STRNSTR)
 char* wolfSSL_strnstr(const char* s1, const char* s2, unsigned int n)
```
dgarske: **🟡 [Medium] Incomplete type migration: parameter `n` still `unsigned int` while `s2_len` is now `size_t`**
*💡 SUGGEST*  `convention`

The PR changes the local `s2_len` from `unsigned int` to `size_t` (removing a potentially truncating cast from the `XSTRLEN` return value), but the function parameter `n` remains `unsigned int` in both the definition (line 4268) and the header declaration (`wolfssl/wolfcrypt/wc_port.h:1730`). This creates a mixed-width comparison on line 4275 (`n >= s2_len`) where `unsigned int` is implicitly promoted to `size_t`. While this promotion is safe for unsigned types and the change is an improvement over the old code (which truncated the `size_t` return of `XSTRLEN`), it is incomplete. On 64-bit platforms some compilers with `-Wconversion` or strict warning levels may flag the implicit widening. Additionally, all callers (in `src/conf.c`, `src/x509.c`, `src/ssl_load.c`, `wolfcrypt/src/asn.c`) currently cast their length to `(unsigned int)` or `(word32)` before passing it — an API change to `size_t n` would let them drop those casts.

**Suggestion:**
```suggestion
Consider also changing the parameter `n` to `size_t` (in both the header and definition) and updating callers to remove their `(unsigned int)` casts, or add a note explaining why this is deferred. If keeping `unsigned int n`, at minimum add a cast `(size_t)n` in the comparison for explicitness.
```

padelsbach:
addressed both comments from skoll

wolfcrypt/src/wc_port.c
padelsbach: done

padelsbach:
jenkins retest this please

PR #10141, opened 2026-04-06
TLS ECH Compliance Fixes

sebastian-carpenter:
# Description                         
                                                                           
  This PR fixes several ECH (Encrypted Client Hello) protocol compliance issues in the TLS implementation, improving correctness of ECH rejection handling, ECHConfig parsing, and inner ClientHello validation per RFC 9849.                                                                   
                                                                                                         
  ## ECH rejection handling (client, §6.1.6 / §6.1.7)                                                    
                                                                                                         
  - After a successful outer handshake where ECH was offered but rejected, the client now sends the mandatory `ech_required` fatal alert (121) and aborts with a new `ECH_REQUIRED_E` error code.        
  - Certificate verification now uses `ECHConfig.contents.public_name` instead of the SNI hostname when ECH is rejected.                                                                                
  - Responds to `CertificateRequest` with an empty `Certificate` message on ECH rejection.               
  - Session tickets and PSK resumption are now ignored/rejected during rejected-ECH handshakes.          
  - ECH processing in ClientHello2 is now gated on the outcome of ClientHello1.                          
  - Added `wolfSSL_GetEchRetryConfigs()` public API to retrieve server-supplied retry configs after ECH rejection; retry configs are stored separately from active configs and properly freed on session teardown.                                                                                    
                  
  ## ECH rejection through HRR (§6.1.4 / §6.1.5)                                                         
                  
  - If the HRR contains no ECH extension, the ECH transcript is freed and processing falls back to the outer hello.
  - ServerHello accepting ECH after an HRR rejection is now treated as an error.                         
  - HPKE state lifetime is now correctly scoped to the handshake leg that established it.                
                                                                                                         
  ## Inner ClientHello integrity (§6.1)                                                                  
                                                                                                         
  - Inner ClientHello is now built with `downgrade` disabled, preventing TLS < 1.3 from being advertised or negotiated.
  - Added `echProcessingInner` flag to correctly validate INNER/OUTER ECH extension types in their respective contexts.                                                                                 
  - The server now verifies the inner ClientHello contains an ECH extension of type `inner` after decryption, and enforces its presence in ClientHello2 when ECH was accepted in ClientHello1.         
                                                                                                         
  ## ECHConfig parsing (§4.2 / §6.2)                                                                     
                                                                                                         
  - Rewrote `SetEchConfigsEx` with a single index variable, eliminating fragile dual-pointer bounds tracking.
  - Added `EchConfigCheckExtensions`: configs with unsupported mandatory extensions (high bit set in type) are now correctly skipped.                                                                     
  - Retry configs in `EncryptedExtensions` are rejected when ECH was already accepted, and discarded on GREASE connections.                                                                               
  - `SetEchConfigsEx` now returns distinct error codes for malformed input, unsupported algorithms, and unsupported versions, rather than a generic failure. Incompatible retry configs from the server are treated as non-fatal.                                                                     
                                                                                                         
  ## `ech_outer_extensions` handling (§5.1)                                                              
                  
  - Merged two near-identical copy loops in `TLSX_ECH_CopyOuterExtensions` into one, with a proper check that `ech_outer_extensions` never references the ECH extension itself.
                                                                                                         
  ## Error codes and alerts (§11.2)                                                                      
   
  - Added `ECH_REQUIRED_E = -519` to `wolfSSL_ErrorCodes`.                                               
  - Added `ech_required = 121` to `AlertDescription`.
  - Added string representations in `AlertTypeToString` and `wolfSSL_ERR_reason_error_string`.
                                                                                                         
  Partly addresses zd#21504
                                                                                                         
  ## Tests        

  - `test_wolfSSL_Tls13_ECH_rejected_cert_valid`: cert authentication against `public_name` on rejection; expected to pass with a matching CN and fail with a mismatched one (§6.1.7).
  - `test_wolfSSL_Tls13_ECH_rejected_empty_client_cert`: empty certificate sent on rejection during mutual auth (§6.1.7).                                                                                
  - Tests for downgrade prevention in the inner ClientHello via a `WOLFSSL_TEST_ECH`-only `echInnerHelloCb` hook (§6.1).                                                                       
  - Tests for mandatory-extension skipping in ECHConfig parsing (§4.2).                                  
  - Tests for ECHConfig error handling across malformed, unsupported-algorithm, and unsupported-version inputs.                                                                          
  - HRR+ECH scenario tests, including retry config retrieval via `wolfSSL_GetEchRetryConfigs()`.
                                                                                                         
  # Checklist     
                                                                                                         
  - [X] added tests
  - [ ] updated/added doxygen

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -8773,6 +8773,10 @@ void wolfSSL_ResourceFree(WOLFSSL* ssl)
         FreeEchConfigs(ssl->echConfigs, ssl->heap);
         ssl->echConfigs = NULL;
     }
+    if (ssl->echRetryConfigs != NULL) {
+        FreeEchConfigs(ssl->echRetryConfigs, ssl->heap);
+        ssl->echRetryConfigs = NULL;
+    }
 #endif /* HAVE_ECH */
 #endif /* WOLFSSL_TLS13 */
 #ifdef WOLFSSL_HAVE_TLS_UNIQUE
@@ -15744,6 +15748,8 @@ int ProcessPeerCerts(WOLFSSL* ssl, byte* input, word32* inOutIdx,
     byte* subjectHash = NULL;
     int alreadySigner = 0;
 
+    char* domainName = NULL;
+
 #if defined(HAVE_CERTIFICATE_STATUS_REQUEST_V2)
     int addToPendingCAs = 0;
 #endif
@@ -16932,17 +16938,34 @@ int ProcessPeerCerts(WOLFSSL* ssl, byte* input, word32* inOutIdx,
                 }
             #endif
 
-                if (!ssl->options.verifyNone && ssl->buffers.domainName.buffer) {
+                domainName = (char*)ssl->buffers.domainName.buffer;
+            #if !defined(NO_WOLFSSL_CLIENT) && defined(HAVE_ECH)
+                /* RFC 9849 s6.1.7: ECH offered but rejected by the server...
+                 * verify cert is valid for ECHConfig.public_name */
+                if (ssl->options.side == WOLFSSL_CLIENT_END &&
+                        ssl->echConfigs != NULL &&
+                        !ssl->options.echAccepted) {
+                    TLSX* echX = TLSX_Find(ssl->extensions, TLSX_ECH);
+                    if (echX != NULL && echX->data != NULL) {
+                        WOLFSSL_ECH* ech = (WOLFSSL_ECH*)echX->data;
+                        if (ech->echConfig != NULL &&
+                                ech->echConfig->publicName != NULL) {
+                            domainName = ech->echConfig->publicName;
+                        }
+                    }
+                }
+            #endif
+
+                if (!ssl->options.verifyNone && domainName) {
                 #ifndef WOLFSSL_ALLOW_NO_CN_IN_SAN
                     /* Per RFC 5280 section 4.2.1.6, "Whenever such identities
                      * are to be bound into a certificate, the subject
                      * alternative name extension MUST be used." */
                     if (args->dCert->altNames) {
-                        if (CheckForAltNames(args->dCert,
-                                (char*)ssl->buffers.domainName.buffer,
-                                (ssl->buffers.domainName.buffer == NULL ? 0 :
-                                (word32)XSTRLEN(
-                                (const char *)ssl->buffers.domainName.buffer)),
+                        if (CheckForAltNames(
+                                args->dCert,
+                                domainName,
+                                (word32)XSTRLEN((const char *)domainName),
                                 NULL, 0, 0) != 1) {
                             WOLFSSL_MSG("DomainName match on alt names failed");
                             /* try to get peer key still */
@@ -16955,11 +16978,9 @@ int ProcessPeerCerts(WOLFSSL* ssl, byte* input, word32* inOutIdx,
                         if (MatchDomainName(
                                 args->dCert->subjectCN,
                                 args->dCert->subjectCNLen,
-                                (char*)ssl->buffers.domainName.buffer,
-                                (ssl->buffers.domainName.buffer == NULL ? 0 :
-                                (word32)XSTRLEN(
-                                (const char *)ssl->buffers.domainName.buffer)
-                                ), 0) == 0)
+                                domainName,
+                                (word32)XSTRLEN((const char *)domainName),
+                                0) == 0)
                     #endif
                         {
                             WOLFSSL_MSG("DomainName match failed");
@@ -16970,18 +16991,19 @@ int ProcessPeerCerts(WOLFSSL* ssl, byte* input, word32* inOutIdx,
                 #else /* WOLFSSL_ALL_NO_CN_IN_SAN */
                     /* Old behavior. */
                 #ifndef  WOLFSSL_HOSTNAME_VERIFY_ALT_NAME_ONLY
-                    if (MatchDomainName(args->dCert->subjectCN,
+                    if (MatchDomainName(
+                                args->dCert->subjectCN,
                                 args->dCert->subjectCNLen,
-                                (char*)ssl->buffers.domainName.buffer,
-                                (ssl->buffers.domainName.buffer == NULL ? 0 :
-                                (word32)XSTRLEN(ssl->buffers.domainName.buffer)), 0) == 0)
+                                domainName,
+                                (word32)XSTRLEN((const char *)domainName),
+                                0) == 0)
                 #endif
                     {
-                        if (CheckForAltNames(args->dCert,
-                                 (char*)ssl->buffers.domainName.buffer,
-                                 (ssl->buffers.domainName.buffer == NULL ? 0 :
-                                 (word32)XSTRLEN(ssl->buffers.domainName.buffer)),
-                                 NULL, 0, 0) != 1) {
+                        if (CheckForAltNames(
+                                args->dCert,
+                                domainName,
+                                (word32)XSTRLEN((const char *)domainName),
+                                NULL, 0, 0) != 1) {
                             WOLFSSL_MSG("DomainName match failed");
                             /* try to get peer key still */
                             ret = DOMAIN_NAME_MISMATCH;
@@ -22208,6 +22230,13 @@ const char* AlertTypeToString(int type)
                 static const char no_application_protocol_str[] =
                     "no_application_protocol";
                 return no_application_protocol_str;
+            }
+
+        case ech_required:
+            {
+                static const char ech_required_str[] =
+                    "ech_required";
+                return ech_required_str;
             }
 
         default:
@@ -27828,6 +27857,9 @@ const char* wolfSSL_ERR_reason_error_string(unsigned long e)
 
     case SESSION_TICKET_NONCE_OVERFLOW:
         return "Session ticket nonce overflow";
+
+    case ECH_REQUIRED_E:
+        return "ECH offered but rejected by server";
     }
 
     return "unknown error number";
--- a/src/tls.c
+++ b/src/tls.c
@@ -13796,6 +13796,8 @@ static int TLSX_ECH_CheckInnerPadding(WOLFSSL* ssl, WOLFSSL_ECH* ech)
     headerSz = ssl->options.dtls ? DTLS13_HANDSHAKE_HEADER_SZ :
                                    HANDSHAKE_HEADER_SZ;
 #else
+    (void)ssl;
+
     headerSz = HANDSHAKE_HEADER_SZ;
 #endif
 
@@ -13835,7 +13837,6 @@ static int TLSX_ECH_CheckInnerPadding(WOLFSSL* ssl, WOLFSSL_ECH* ech)
         acc |= innerCh[i];
     }
     if (acc != 0) {
-        SendAlert(ssl, alert_fatal, illegal_parameter);
         return INVALID_PARAMETER;
     }
 
@@ -13920,14 +13921,14 @@ static const byte* TLSX_ECH_FindOuterExtension(const byte* outerCh,
 
 /* If newinnerCh is NULL, validate ordering and existence of references
  *   - updates newInnerChLen with total length of selected extensions
- * If newinnerCh in not NULL, copy extensions into newInnerCh
+ * If newinnerCh is not NULL, copy extensions into newInnerCh
  *
  * outerCh          The outer ClientHello buffer.
  * outerChLen       Outer ClientHello length.
  * newInnerCh       The inner ClientHello buffer.
  * newInnerChLen    Inner ClientHello length.
  * numOuterRefs     Number of references described by OuterExtensions extension.
- * numOuterTypes    References described by OuterExtensions extension.
+ * OuterRefTypes    References described by OuterExtensions extension.
  * returns 0 on success and otherwise failure.
  */
 static int TLSX_ECH_CopyOuterExtensions(const byte* outerCh, word32 outerChLen,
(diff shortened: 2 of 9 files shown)


src/tls.c
```diff
@@ -13921,51 +13920,35 @@ static int TLSX_ECH_CopyOuterExtensions(const byte* outerCh, word32 outerChLen,
...
+                        "of order");
+            ret = INVALID_PARAMETER;
+            break;
+        }
 
+        if (newInnerCh == NULL) {
             *newInnerChLen += outerExtLen;
-
-            outerRefTypes += OPAQUE16_LEN;
         }
```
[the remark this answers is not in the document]
sebastian-carpenter: The caller is now expected to initialize to 0. This should only ever be called when expanding outer_extensions so I'm not too concerned with it

src/tls13.c
```diff
@@ -5950,6 +5987,14 @@ static int DoTls13CertificateRequest(WOLFSSL* ssl, const byte* input,
         return ret;
 #endif
 
+#if defined(HAVE_ECH)
+    /* RFC 9849 s6.1.7: ECH was offered but rejected by the server...
+     * the client MUST respond with an empty Certificate message. */
+    if (ssl->echConfigs != NULL && !ssl->options.echAccepted) {
```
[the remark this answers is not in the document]
sebastian-carpenter: seems like a good fix

src/tls.c
```diff
@@ -14438,22 +14442,37 @@ static int TLSX_ECH_Parse(WOLFSSL* ssl, const byte* readBuf, word16 size,
...
-        if (ret == 0) {
+        /* if we failed to extract/expand */
+        if (ret != 0) {
+            WOLFSSL_MSG("Failed to decrypt InnerHello");
+            if (ech->hpkeContext != NULL) {
+                /* on SH2 this is fatal */
+                SendAlert(ssl, alert_fatal, decrypt_error);
+                WOLFSSL_ERROR_VERBOSE(VERIFY_FINISHED_ERROR);
+                ret = VERIFY_FINISHED_ERROR;
```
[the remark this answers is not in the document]
sebastian-carpenter: yeah, that should be decrypt error

src/tls13.c
```diff
@@ -5134,8 +5147,15 @@ static int EchCheckAcceptance(WOLFSSL* ssl, byte* label, word16 labelSz,
...
-            ret = 0;
+            if (msgType != hello_retry_request && ssl->options.echAccepted) {
+                /* the SH has rejected ECH after the HRR has accepted it
+                 * RFC 9849, section 6.1.5 */
+                ssl->options.echAccepted = 0;
+                ret = INVALID_PARAMETER;
+            }
+            else {
+                ret = 0;
+            }
```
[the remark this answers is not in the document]
sebastian-carpenter: it is already 0 in the other case

src/tls.c
```diff
@@ -13771,6 +13771,8 @@ static int TLSX_ECH_CheckInnerPadding(WOLFSSL* ssl, WOLFSSL_ECH* ech)
     byte acc = 0;
     word32 i;
 
+    (void)ssl;
```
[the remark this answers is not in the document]
sebastian-carpenter: it's used right after that. I don't think it's an issue

src/tls.c
```diff
@@ -13814,7 +13816,6 @@ static int TLSX_ECH_CheckInnerPadding(WOLFSSL* ssl, WOLFSSL_ECH* ech)
         acc |= innerCh[i];
     }
     if (acc != 0) {
-        SendAlert(ssl, alert_fatal, illegal_parameter);
         return INVALID_PARAMETER;
```
dgarske, line 13819: **🟡 [Medium] Missing fatal alert for inner hello padding validation failure**
*💡 SUGGEST*  `convention`

The `SendAlert(ssl, alert_fatal, illegal_parameter)` call was removed from `TLSX_ECH_CheckInnerPadding` (visible in the diff removing line 13819), and the `(void)ssl` was added to suppress the unused-parameter warning. However, no replacement alert is sent by the caller chain. When the server detects non-zero padding bytes in a decrypted inner ClientHello, this is a protocol violation that RFC 9849 Section 7.1 says MUST result in an `illegal_parameter` alert. The error propagates back to TLSX_Parse and eventually terminates the handshake, but the peer receives no specific alert describing the reason. The same gap applies when `TLSX_ECH_ExpandOuterExtensions` fails after successful decryption (line 14470), as the error from that path also lacks a specific alert.

**Suggestion:**
```suggestion
Consider sending the alert in the caller at line 14473 of TLSX_ECH_Parse, where the error from CheckInnerPadding or ExpandOuterExtensions is caught:

    if (ret != 0) {
        SendAlert(ssl, alert_fatal, illegal_parameter);
        XFREE(ech->innerClientHello, ssl->heap, DYNAMIC_TYPE_TMP_BUFFER);
        ech->innerClientHello = NULL;
    }
```

dgarske (commented):
## 🐺 Skoll Code Review

### Posted findings
- [Medium] **Missing fatal alert for inner hello padding validation failure** — `src/tls.c:13819`
- [Low] **Dead null-check ternary for domainName inside non-null guard** — `src/internal.c:16878`
- [Medium] **No test for HRR ECH rejection fallback path** — `src/tls13.c:5717-5722`
- [Medium] **Retry configs not stored for real ECH clients on rejection** — `src/tls.c:14275-14289`
- [Low] **Inconsistent bit-field base type for ECH flags** — `wolfssl/internal.h:5164-5166`
- [Medium] **GREASE encrypted_extensions: error from wolfSSL_SetEchConfigs not distinguished** — `src/tls.c:14280-14289`

- [Low] Removed test_wolfSSL_SubTls13_ECH without full replacement

</details>

src/internal.c
```diff
@@ -16848,17 +16850,33 @@ int ProcessPeerCerts(WOLFSSL* ssl, byte* input, word32* inOutIdx,
...
                     /* Per RFC 5280 section 4.2.1.6, "Whenever such identities
                      * are to be bound into a certificate, the subject
                      * alternative name extension MUST be used." */
                     if (args->dCert->altNames) {
                         if (CheckForAltNames(args->dCert,
-                                (char*)ssl->buffers.domainName.buffer,
-                                (ssl->buffers.domainName.buffer == NULL ? 0 :
-                                (word32)XSTRLEN(
-                                (const char *)ssl->buffers.domainName.buffer)),
+                                domainName, (domainName == NULL ? 0 :
```
dgarske, line 16878: **🔵 [Low] Dead null-check ternary for domainName inside non-null guard**
*🔧 NIT*  `style`

The ternary `domainName == NULL ? 0 : (word32)XSTRLEN(...)` appears at lines 16878, 16892, 16907, and 16912, but these are all inside the `if (!ssl->options.verifyNone && domainName)` block at line 16871, which guarantees `domainName` is non-NULL. The NULL check in the ternary is dead code. This pattern was inherited from the original code that used `ssl->buffers.domainName.buffer` (which had the same dead check), so this is a pre-existing pattern that was migrated rather than newly introduced. Still, it could be simplified for clarity.

**Suggestion:**
```suggestion
Replace `(domainName == NULL ? 0 : (word32)XSTRLEN((const char *)domainName))` with `(word32)XSTRLEN(domainName)` at all four occurrences within the `if (domainName)` block.
```

src/tls13.c
```diff
@@ -5691,31 +5711,40 @@ int DoTls13ServerHello(WOLFSSL* ssl, const byte* input, word32* inOutIdx,
...
+    if (ssl->echConfigs != NULL && !ssl->options.disableECH &&
+            ssl->hsHashesEch != NULL) {
         args->echX = TLSX_Find(ssl->extensions, TLSX_ECH);
-        /* account for hrr extension instead of server random */
-        if (args->extMsgType == hello_retry_request) {
-            args->acceptOffset =
-                (word32)(((WOLFSSL_ECH*)args->echX->data)->confBuf - input);
-            args->acceptLabel = (byte*)echHrrAcceptConfirmationLabel;
-            args->acceptLabelSz = ECH_HRR_ACCEPT_CONFIRMATION_LABEL_SZ;
+        if (args->extMsgType == hello_retry_request &&
```
dgarske, line 5717: **🟡 [Medium] No test for HRR ECH rejection fallback path**
*💡 SUGGEST*  `test`

The new code at lines 5717-5722 handles the case where a HelloRetryRequest arrives without an ECH extension (`confBuf == NULL`), meaning the server rejected ECH during HRR. This causes the client to fall back to the outer transcript by freeing `hsHashesEch`. The existing `test_wolfSSL_Tls13_ECH_HRR` test appears to test HRR with ECH acceptance, but there is no test for HRR with ECH rejection (server sends HRR without ECH extension). This is a complex state transition that should be covered.

**Recommendation:** Add a test case where the server is configured with ECH, the client offers ECH, but an HRR is triggered and the server disables ECH processing during the HRR, causing the HRR to lack an ECH extension. Verify the handshake completes using the outer transcript and that the client correctly sends the ech_required alert.

src/tls.c
```diff
@@ -14280,16 +14265,34 @@ static int TLSX_ECH_Parse(WOLFSSL* ssl, const byte* readBuf, word16 size,
...
     if (msgType == encrypted_extensions) {
-        ret = wolfSSL_SetEchConfigs(ssl, readBuf, size);
+        /* configs should only be sent on ECH rejection (RFC9849, Section 5) */
+        if (ssl->options.echAccepted) {
+            SendAlert(ssl, alert_fatal, unsupported_extension);
+            WOLFSSL_ERROR_VERBOSE(UNSUPPORTED_EXTENSION);
+            return UNSUPPORTED_EXTENSION;
+        }
+
+        echX = TLSX_Find(ssl->extensions, TLSX_ECH);
```
dgarske, line 14275: **🟡 [Medium] Retry configs not stored for real ECH clients on rejection**
*💡 SUGGEST*  `question`

When the server rejects ECH and sends retry configs in EncryptedExtensions, the code calls `wolfSSL_SetEchConfigs(ssl, readBuf, size)` at line 14280. However, `wolfSSL_SetEchConfigs` (ssl_ech.c line 299) returns `WOLFSSL_FATAL_ERROR` if `ssl->echConfigs` is already non-NULL. For a real ECH client (not GREASE), `ssl->echConfigs` is set during connection setup, so this call will always fail. The GREASE path works because GREASE clients have `ssl->echConfigs == NULL`. This means real ECH clients never receive retry configs from the server, which may limit the ability to retry with updated configs. This appears to be a pre-existing limitation, but the new GREASE handling code at line 14284 makes it more visible since it assumes configs were successfully parsed.

**Recommendation:** Consider freeing existing `ssl->echConfigs` before calling `wolfSSL_SetEchConfigs` for real ECH clients, or document that retry configs are not stored for subsequent connection attempts. At minimum, if `wolfSSL_SetEchConfigs` fails for a non-GREASE client, the error should not propagate as a handshake failure.

wolfssl/internal.h
```diff
@@ -5163,6 +5163,7 @@ struct Options {
 #if defined(WOLFSSL_TLS13) && defined(HAVE_ECH)
     word16            echAccepted:1;
```
dgarske, line 5164: **🔵 [Low] Inconsistent bit-field base type for ECH flags**
*🔧 NIT*  `convention`

In the ECH flags block, `echAccepted` uses `word16` as its base type while `disableECH` and the new `echProcessingInner` use `byte`. All three are 1-bit bit-fields. While this works correctly (compilers handle mixed bit-field types), it's inconsistent within the same `#if` block.

**Recommendation:** Consider using the same base type for all three bit-fields for consistency. This is purely cosmetic.

src/tls.c
```diff
@@ -14280,16 +14265,34 @@ static int TLSX_ECH_Parse(WOLFSSL* ssl, const byte* readBuf, word16 size,
...
+        if (echX == NULL)
+            return BAD_FUNC_ARG;
+        ech = (WOLFSSL_ECH*)echX->data;
 
+        ret = wolfSSL_SetEchConfigs(ssl, readBuf, size);
```
dgarske, line 14280: **🟡 [Medium] GREASE encrypted_extensions: error from wolfSSL_SetEchConfigs not distinguished**
*💡 SUGGEST*  `bug`

When the server sends retry_configs in EncryptedExtensions and the client is using GREASE, the code at line 14280 calls `wolfSSL_SetEchConfigs` to syntactically validate the configs. If the configs are malformed, `ret` will be non-zero (e.g., `BUFFER_E`). At line 14284, the condition `ret == 0 && ech->state == ECH_WRITE_GREASE` is false, so the cleanup doesn't run. The function then returns the error code, which would abort the handshake. However, for a GREASE client, malformed retry configs from the server should arguably not be fatal - the GREASE client has no obligation to use them. The non-GREASE path has the same issue (pre-existing), but for GREASE specifically, aborting on a parse error seems overly strict since the configs are being discarded anyway.

**Suggestion:**
```suggestion
For the GREASE path, consider ignoring parse errors from wolfSSL_SetEchConfigs:

        ret = wolfSSL_SetEchConfigs(ssl, readBuf, size);
        if (ret == WOLFSSL_SUCCESS)
            ret = 0;

        if (ech->state == ECH_WRITE_GREASE) {
            if (ret == 0) {
                FreeEchConfigs(ssl->echConfigs, ssl->heap);
            }
            ssl->echConfigs = NULL;
            ret = 0; /* don't fail GREASE on bad retry configs */
        }
```

src/tls.c
sebastian-carpenter: The error propagates up to `TranslateErrorToAlert()` where it will then send an alert. I tried setting the condition to `acc == 0` here to force an alert and it does send it.

sebastian-carpenter:
Should I try and break this down into multiple PR's?

src/internal.c
sebastian-carpenter: removed unnecessary checks

src/tls13.c
sebastian-carpenter: Added test. Caught some bugs from it

wolfssl/internal.h
sebastian-carpenter: updated to word16

src/tls.c
sebastian-carpenter: It took a bit but I added retry config handling and improved the error propagation. Added a bunch of tests for retry configs too.

Added new API to retrieve retry configs and a field + bit to track them.

src/tls.c
sebastian-carpenter: Malformed configs MUST be fatal. That being said the function did return bad error codes. Updated it so that two cases (bad version, unsupported ciphers) don't cause the handshake to fail.

sebastian-carpenter:
Jenkins retest this please.

dgarske (changes requested):
Looks good and I want this merged soon. Please squash. If you need to split into several functionally different commits that fine. 29 is too many.

dgarske (approved):

PR #10145, opened 2026-04-07
ECC curve validation follow-up

Frauschi:
Follow-up for #10133 to fix the two open Fenrir findings:
1. Better SM2 handling for `SP_MATH` builds without SP support for SM2 curve
2. Better detection of unset public keys


SparkiDev:
retest this please

PR #10149, opened 2026-04-07
Refactor record padding handling to eliminate middle padding pattern

julek-wolfssl:
Move padSz index advancement from individual message handlers to a
single location at the end of record processing in ProcessReply.
Previously, each handler (DoFinished, DoApplicationData, DoAlert,
DoCertificate, DoServerHello, etc.) advanced the index by padSz,
which then had to be corrected when processing multiple messages
in one record. This "middle padding" pattern was error-prone.

Now curSize is reduced by padSz after decryption/verification so
it reflects content size only. Message handlers advance the index
by content size only. padSz is added once when the record is fully
processed. The correction code for multi-message records is removed.

### DIFF
--- a/scripts/multi-msg-record.test
+++ b/scripts/multi-msg-record.test
@@ -1,0 +1,37 @@
+#!/usr/bin/env bash
+#
+# multi-msg-record.test
+#
+# Wrapper around scripts/multi-msg-record.py, invoked by automake via
+# `make check`.  Verifies that the wolfSSL client correctly processes
+# TLS records that carry multiple handshake messages in a single record
+# (TLS 1.2 plaintext + renegotiation, and TLS 1.3 encrypted flight).
+#
+# The test is SKIPped (exit 77) when python3 or tlslite-ng is not
+# installed.
+#
+
+# if we can, isolate the network namespace to eliminate port collisions.
+if [[ -n "$NETWORK_UNSHARE_HELPER" ]]; then
+     if [[ -z "$NETWORK_UNSHARE_HELPER_CALLED" ]]; then
+         export NETWORK_UNSHARE_HELPER_CALLED=yes
+         exec "$NETWORK_UNSHARE_HELPER" "$0" "$@" || exit $?
+     fi
+elif [ "${AM_BWRAPPED-}" != "yes" ]; then
+    bwrap_path="$(command -v bwrap)"
+    if [ -n "$bwrap_path" ]; then
+        export AM_BWRAPPED=yes
+        exec "$bwrap_path" --unshare-net --dev-bind / / "$0" "$@"
+    fi
+    unset AM_BWRAPPED
+fi
+
+# Locate python3 – skip (exit 77) if unavailable.
+PYTHON="$(command -v python3)"
+if [ -z "$PYTHON" ]; then
+    echo "python3 not found, skipping multi-msg-record test"
+    exit 77
+fi
+
+SCRIPT_DIR="$(cd -- "$(dirname -- "$0")" && pwd)"
+exec "$PYTHON" "$SCRIPT_DIR/multi-msg-record.py" "$@"
--- a/scripts/multi-msg-record.test
+++ b/scripts/multi-msg-record.test
@@ -26,7 +26,7 @@ elif [ "${AM_BWRAPPED-}" != "yes" ]; then
     unset AM_BWRAPPED
 fi
 
-# Locate python3 – skip (exit 77) if unavailable.
+# Locate python3 - skip (exit 77) if unavailable.
 PYTHON="$(command -v python3)"
 if [ -z "$PYTHON" ]; then
     echo "python3 not found, skipping multi-msg-record test"
--- a/src/internal.c
+++ b/src/internal.c
@@ -19088,6 +19088,8 @@ static int DoHandShakeMsg(WOLFSSL* ssl, byte* input, word32* inOutIdx,
 
     /* curSize has already been reduced to content-only (padSz subtracted)
      * in ProcessReply, so curStartIdx + curSize bounds the content. */
+    if (*inOutIdx > (word32)ssl->curStartIdx + ssl->curSize)
+        return BUFFER_ERROR;
     inputLength = ssl->curStartIdx + ssl->curSize - *inOutIdx;
 
     /* If there is a pending fragmented handshake message,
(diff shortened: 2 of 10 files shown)


julek-wolfssl:
Retest this please history lost........

src/internal.c
```diff
@@ -23244,6 +23179,11 @@ static int DoProcessReplyEx(WOLFSSL* ssl, int allowSocketErr)
 
             }
 
+            /* Reduce curSize to content only, excluding padding/MAC overhead.
+             * padSz is accounted for once at the end of record processing. */
+            if (IsEncryptionOn(ssl, 0))
```
dgarske: **🟡 [Medium] Missing defensive guard before `curSize -= padSz` could allow word16 underflow**
*💡 SUGGEST*  `bug`

The new centralized padding subtraction `ssl->curSize -= (word16)ssl->keys.padSz` performs unsigned word16 arithmetic. If `padSz` exceeds `curSize` due to a malformed record or bug in MAC verification, this underflows to a very large value (e.g., curSize=50, padSz=100 → curSize=65486).

For TLS 1.3, `removeMsgInnerPadding` guarantees `padSz <= curSize` by construction (scan-based). For non-TLS-1.3, `VerifyMac` should guarantee this, but there is no explicit check at the subtraction site.

Importantly, the `MAX_PLAINTEXT_SZ` check at lines 23245/23263 catches the underflow for pre-TLS-1.3 (ETM and non-ETM paths). However, for TLS 1.3 these checks are skipped (comment at line 23261: 'TLS13 plaintext limit is checked earlier before decryption'), so if the invariant were ever violated in TLS 1.3, the underflowed curSize would propagate to message handlers unchecked.

**Suggestion:**
```suggestion
/* Reduce curSize to content only, excluding padding/MAC overhead.
             * padSz is accounted for once at the end of record processing. */
            if (IsEncryptionOn(ssl, 0)) {
                if (ssl->keys.padSz > ssl->curSize)
                    return BUFFER_E;
                ssl->curSize -= (word16)ssl->keys.padSz;
            }
```

src/internal.c
```diff
@@ -23710,53 +23653,44 @@ static int DoProcessReplyEx(WOLFSSL* ssl, int allowSocketErr)
 
             ssl->options.processReply = doProcessInit;
```
dgarske, line 23657: **🟡 [Medium] No dedicated test for multi-message encrypted record processing after refactoring**
*💡 SUGGEST*  `test`

The core change in this PR is eliminating the 'middle padding' correction pattern where padding was added per-handler and then corrected at record end. The existing test suite provides indirect coverage through standard TLS/DTLS handshakes (e.g., `test_records_span_network_boundaries`, `test_tls13_middlebox_compat_empty_session_id`), but there is no dedicated test that:

1. Constructs a single TLS 1.2 encrypted record containing multiple handshake messages
2. Verifies all messages are processed correctly
3. Verifies the index advances by exactly `content_size + padSz` at the end

This refactoring removes ~15 separate `*inOutIdx += ssl->keys.padSz` call sites and replaces them with a single centralized addition. A regression in any handler's index advancement would cause subtle protocol-level failures that might not be caught by existing tests.

**Recommendation:** Add a targeted test that sends a TLS 1.2 encrypted record with multiple handshake messages (e.g., Certificate + ServerKeyExchange in a single record) and verifies correct processing. This would directly exercise the refactored multi-message record loop and catch any regression in the centralized padding handling.

dgarske (commented):
## 🐺 Skoll Code Review

**Findings:** 4 total — 3 posted, 2 skipped

### Posted findings
- [Medium] **Missing defensive guard before `curSize -= padSz` could allow word16 underflow** — `src/internal.c:23184-23185`
- [Medium] **Callers of message handlers still pass `inputBuffer.length` as totalSz, not content boundary** — `src/internal.c:23286-23289,23306-23309,23343-23346,23360-23362`
- [Medium] **No dedicated test for multi-message encrypted record processing after refactoring** — `src/internal.c:23654-23695`

- [Low] Truncation cast of word32 padSz to word16 suppresses potential warnings
- [Medium] Callers of message handlers still pass `inputBuffer.length` as totalSz, not content boundary

</details>

src/internal.c
julek-wolfssl: Added a test in `scripts/multi-msg-record.test`.

src/internal.c
julek-wolfssl: Added.

julek-wolfssl:
Failures unrelated to PR.

JacobBarthelmeh:
Retest this please Jenkins

JacobBarthelmeh (dismissed):
This appears to be a good refactor. I need help on closer review of DTLS 1.3 changes though. @rizlik could you provide an additional review?

douzzer:
retest this please
(results lost)

scripts/multi-msg-record.test
```diff
@@ -0,0 +1,37 @@
...
+elif [ "${AM_BWRAPPED-}" != "yes" ]; then
+    bwrap_path="$(command -v bwrap)"
+    if [ -n "$bwrap_path" ]; then
+        export AM_BWRAPPED=yes
+        exec "$bwrap_path" --unshare-net --dev-bind / / "$0" "$@"
+    fi
+    unset AM_BWRAPPED
+fi
+
+# Locate python3 – skip (exit 77) if unavailable.
```
dgarske: Check for non ASCII here:

```
[check-source-text] [5 of 41] [3caf4b9da9]
    configure...   real 0m8.506s  user 0m5.226s  sys 0m4.103s
8 bit:
./scripts/multi-msg-record.test:29:# Locate python3 ¶ skip (exit 77) if unavailable.
```

dgarske (changes requested):

rizlik (changes requested):
Very nice!
Can you add a github action that install python script dependencies and run the test? My fear is that this test might be skipped because of the missing dependency and the multimessage is left untested in our CI.

scripts/multi-msg-record.test
julek-wolfssl: Fixed.

julek-wolfssl:
@rizlik added python setup to relevant workflows.

rizlik (approved):

dgarske (approved):

PR #10150, opened 2026-04-07
Guard OCSP signature params with WC_RSA_PSS ifdef

julek-wolfssl:
OCSPBASICRESPASN_IDX_SIGNATURE_PARAMS is only defined when WC_RSA_PSS
is enabled but was used unconditionally in EncodeBasicOcspResponse,
causing a build error when WC_RSA_PSS is not defined.


julek-wolfssl:
retest this please history lost

julek-wolfssl:
Failures not related to PR

SparkiDev (approved):

PR #10162, opened 2026-04-08
Use O_CLOEXEC to avoid race conditions

### LINKED ISSUE #9753: [Bug]: many open() without O_CLOEXEC
### Contact Details

socketpair@gmail.com

### Version

main branch

### Description

In multithreaded programs, O_CLOEXEC for all file descriptors is a must. I just searched for O_RDONLY and O_RDWR and found few places. 

Actually, I searched for  open("/etc/ssl/certs/ca-bundle.crt", O_RDONLY), but found other places in sources.

### Reproduction steps

_No response_

### Relevant log output

```shell

```

embhorn:
# Description

Use the "close-on-exec" flag at file creation time to harden multithread use cases against race conditions.

Fixes #9753 

# Testing

build testing

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


src/crl.c
```diff
@@ -1710,6 +1715,9 @@ static THREAD_RETURN WOLFSSL_THREAD DoMonitor(void* arg)
         SignalSetup(crl, MONITOR_SETUP_E);
         return NULL;
     }
+#ifdef FD_CLOEXEC
+    (void)fcntl(crl->mfd, F_SETFD, FD_CLOEXEC);
```
dgarske: **🔵 [Low] crl.c kqueue path: F_SETFD without read-modify-write**
*🔧 NIT*  `convention`

The kqueue code path sets `FD_CLOEXEC` directly:
```c
(void)fcntl(crl->mfd, F_SETFD, FD_CLOEXEC);
```
This overwrites any existing fd flags rather than ORing `FD_CLOEXEC` into the current flags. Elsewhere in this same PR (inotify path at line 1866-1868, benchmark.c line 1599-1601, wolfio.c line 1649-1651), the read-modify-write pattern is correctly used:
```c
int fdFlags = fcntl(fd, F_GETFD);
if (fdFlags >= 0)
    (void)fcntl(fd, F_SETFD, fdFlags | FD_CLOEXEC);
```
In practice, kqueue fds have no other fd flags, so this is harmless. But the inconsistency could mislead future developers who copy the pattern.

**Suggestion:**
```suggestion
```c
#ifdef FD_CLOEXEC
    {
        int fdFlags = fcntl(crl->mfd, F_GETFD);
        if (fdFlags >= 0)
            (void)fcntl(crl->mfd, F_SETFD, fdFlags | FD_CLOEXEC);
    }
#endif
```
```

File: src/wolfio.c [hunk 2]
```diff
@@ -1623,7 +1638,21 @@ int wolfIO_TcpBind(SOCKET_T* sockfd, word16 port)
 #ifdef HAVE_SOCKADDR
 int wolfIO_TcpAccept(SOCKET_T sockfd, SOCKADDR* peer_addr, XSOCKLENT* peer_len)
 {
-    return (int)accept(sockfd, peer_addr, peer_len);
+    int fd;
+#if !defined(USE_WINDOWS_API) && !defined(WOLFSSL_ZEPHYR) && \
```
dgarske: **🔵 [Low] wolfio.c: defined(SOCK_CLOEXEC) guard in accept4 path is tautologically true**
*🔧 NIT*  `style`

The fallback `#define SOCK_CLOEXEC 0` at line 56-57 makes `defined(SOCK_CLOEXEC)` in the `accept4` guard at line 1643 always true. The code works correctly because `_GNU_SOURCE` is defined at the top of the file ensuring `accept4` is declared, and `SOCK_CLOEXEC=0` is a no-op flag. However, the `defined(SOCK_CLOEXEC)` check is misleading — it appears to test for platform support but can never be false. A reader might incorrectly think removing the fallback define would enable the else branch.

**Suggestion:**
```suggestion
Consider removing `defined(SOCK_CLOEXEC)` from the guard since it's always true, or add a comment noting the fallback makes it tautological:
```c
#if !defined(USE_WINDOWS_API) && !defined(WOLFSSL_ZEPHYR) && \
    (defined(__linux__) || defined(__ANDROID__))
    fd = (int)accept4(sockfd, peer_addr, peer_len, SOCK_CLOEXEC);
```
```

File: wolfcrypt/src/port/af_alg/afalg_hash.c [hunk 3]
```diff
@@ -223,8 +223,13 @@ static int AfalgHashCopy(wolfssl_AFALG_Hash* src, wolfssl_AFALG_Hash* dst)
     }
 #endif
 
-    dst->rdFd = accept(src->rdFd, NULL, 0);
-    dst->alFd = accept(src->alFd, NULL, 0);
+#if defined(__linux__) && defined(SOCK_CLOEXEC)
```
dgarske: **🟡 [Medium] afalg_hash.c: accept4 used without _GNU_SOURCE but guarded by system SOCK_CLOEXEC**
*💡 SUGGEST*  `bug`

`afalg_hash.c` uses `accept4` guarded by `#if defined(__linux__) && defined(SOCK_CLOEXEC)`. Unlike `wc_afalg.c`, it does NOT define a `SOCK_CLOEXEC` fallback, so the guard correctly tests system support (if the system defines `SOCK_CLOEXEC`, `accept4` should also be declared). This is the correct pattern.

However, `afalg_hash.c` doesn't define `_GNU_SOURCE` and doesn't include `<sys/socket.h>` directly — it gets it through `wc_afalg.h`. On glibc 2.14–2.18 without `_GNU_SOURCE`, the system `SOCK_CLOEXEC` is not defined in the headers, so the code correctly falls through to `accept()`. This works, but adding `_GNU_SOURCE` (like `wolfio.c` does) would allow `accept4` to work on those older systems too.

**Recommendation:** Consider adding `_GNU_SOURCE` before system headers in `afalg_hash.c` (and transitively `wc_afalg.h`), matching the `wolfio.c` pattern, so `accept4` is available on all glibc-based Linux systems.

src/wolfio.c
```diff
@@ -24,6 +24,15 @@
 #define WOLFSSL_STRERROR_BUFFER_SIZE 256
 #endif
 
+/* Enable GNU extensions for accept4() on Linux/glibc. Must be defined
+ * before any system headers are included. Excluded for Zephyr and other
+ * embedded RTOSes whose libc layers conflict with glibc-style definitions
+ * (e.g., Zephyr's socket_select.h vs. glibc's fd_set). */
+#if (defined(__linux__) || defined(__ANDROID__)) && \
```
dgarske: **🟠 [High] wolfio.c: _GNU_SOURCE placement may conflict if libwolfssl_sources.h pulls in system headers**
*🚫 BLOCK*  `bug`

`_GNU_SOURCE` must be defined before **any** system headers are included. In `wolfio.c`, it's defined at line 33, then `libwolfssl_sources.h` is included at line 36. `libwolfssl_sources.h` includes `config.h`, then `types.h`, `error-crypt.h`, and `logging.h`. The `types.h` header includes `<stdint.h>` and potentially other system headers. On glibc, once any system header is included, defining `_GNU_SOURCE` after the fact has no effect — feature test macros are evaluated on the **first** inclusion of `<features.h>`, which is triggered by the first system header.

If `types.h` (via `<stdint.h>` → `<features.h>`) is processed before `_GNU_SOURCE` is checked, the `accept4` declaration won't be exposed when `<sys/socket.h>` is later included, even though `_GNU_SOURCE` is now defined.

Note: in practice, `config.h` from autotools may define `_GNU_SOURCE` itself (when `--enable-asyncthreads` is used), and modern glibc >= 2.19 exposes `accept4` by default. But for correctness on all supported configurations, `_GNU_SOURCE` should be defined before `libwolfssl_sources.h`.

**Suggestion:**
```suggestion
The `_GNU_SOURCE` define is already before `libwolfssl_sources.h`. The concern is whether `_GNU_SOURCE` itself can be tested before `config.h` is included. It should work because the `#if !defined(_GNU_SOURCE)` check only tests whether `_GNU_SOURCE` was defined by the build system (e.g., `-D_GNU_SOURCE` on the command line). If autotools already set it, we skip the define; if not, we set it before any headers. **However**, verify that no header included by `libwolfssl_sources.h` pulls in `<features.h>` before our `_GNU_SOURCE` takes effect. On glibc, `<stdint.h>` does include `<features.h>`, but since `_GNU_SOURCE` is defined before that include, it should be visible. This is correct order.
```

dgarske (changes requested):
## 🐺 Skoll Code Review

### Posted findings
- [Low] **crl.c kqueue path: F_SETFD without read-modify-write** — `src/crl.c:1719`
- [Low] **wolfio.c: defined(SOCK_CLOEXEC) guard in accept4 path is tautologically true** — `src/wolfio.c:1642-1643`
- [Medium] **afalg_hash.c: accept4 used without _GNU_SOURCE but guarded by system SOCK_CLOEXEC** — `wolfcrypt/src/port/af_alg/afalg_hash.c:226-232`
- [High] **wolfio.c: _GNU_SOURCE placement may conflict if libwolfssl_sources.h pulls in system headers** — `src/wolfio.c:31-36`

- [Medium] wc_afalg.c: SOCK_CLOEXEC fallback defeats accept4 guard — missing _GNU_SOURCE

</details>

SparkiDev (dismissed):

dgarske (dismissed):
@douzzer please look these over as well.

douzzer (changes requested):
Would be a lot cleaner+nicer to take the `wc_open_cloexec()` added in random.c, move it to wc_port.c, make it a WOLFSSL_LOCAL, and use it everywhere consistently.

While you're at it, it would make sense to add other syscall wrappers for calls used over and over with CLOEXEC -- probably `accept()`, probably others.

ColtonWilley:
Jenkins retest this please

douzzer:
retest this please
```
wolfSSL error: tcp bind failed: Address already in use
```

douzzer:
retest this please
(possible true positive -- ```wolfSSL error: tcp bind failed: Address already in use```
in `ocsp-stapling2.test`)

douzzer (approved):
The hardcoded fallback `SO_REUSEPORT` is absolutely atrocious but we can fix that at a later date :-)

JacobBarthelmeh (approved):

src/crl.c
```diff
@@ -1729,6 +1712,7 @@ static THREAD_RETURN WOLFSSL_THREAD DoMonitor(void* arg)
         SignalSetup(crl, MONITOR_SETUP_E);
         return NULL;
     }
+    wc_set_cloexec(crl->mfd);
```
socketpair, line 1715: No, solution is racy. Use kqueue1()

wolfcrypt/src/port/af_alg/afalg_hash.c
```diff
@@ -19,13 +19,19 @@
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1335, USA
  */
 
+#if defined(__linux__) && !defined(_GNU_SOURCE)
+    #define _GNU_SOURCE 1
```
socketpair, line 23: Move that please to the global build config (in order to use this definition for all files in the project).

wolfcrypt/src/wc_port.c
```diff
@@ -5130,6 +5136,76 @@ char* wolfSSL_strnstr(const char* s1, const char* s2, size_t n)
...
+    #define SOCK_CLOEXEC 0
+#endif
+
+void wc_set_cloexec(int fd)
+{
+#ifdef FD_CLOEXEC
+    int fdFlags;
+    if (fd < 0)
+        return;
+    fdFlags = fcntl(fd, F_GETFD);
```
socketpair, line 5159: Under linux there is a faster alternative: int result = ioctl(fd, FIOCLEX);
(If supported) See CPython sources.  Really faster you may benchmark.

wolfssl/test.h
```diff
@@ -1669,6 +1669,12 @@ static WC_INLINE void tcp_listen(SOCKET_T* sockfd, word16* port, int useAnyAddr,
         if (res < 0)
             err_sys_with_errno("setsockopt SO_REUSEADDR failed\n");
     }
+/* glibc hides SO_REUSEPORT under strict C99 feature-test visibility
+ * (no _DEFAULT_SOURCE/__USE_MISC). Fall back to the Linux numeric value
+ * so concurrent binds on the same port remain supported. */
+#if defined(__linux__) && !defined(SO_REUSEPORT)
+    #define SO_REUSEPORT 15
```
socketpair, line 1676: Are you sure the value is the same on all cpu architectures?

wolfcrypt/benchmark/benchmark.c
```diff
@@ -1574,16 +1574,30 @@ static const char* bench_result_words3[][5] = {
...
+    #ifndef PERF_FLAG_FD_CLOEXEC
+        #define PERF_FLAG_FD_CLOEXEC (1UL << 3)
+    #endif
 
     static THREAD_LS_T word64 begin_cycles;
     static THREAD_LS_T word64 total_cycles;
     static THREAD_LS_T int cycles = -1;
     static THREAD_LS_T struct perf_event_attr atr;
 
+    /* Try with PERF_FLAG_FD_CLOEXEC first; on older kernels (< 3.14) this
```
socketpair, line 1589: Do we need to support ancient kernels ? Why not to drop support for too old kernels/headers ?

PR #10172, opened 2026-04-08
Fix pkcs12 parse issue

embhorn:
# Description

When an ENCRYPTED_DATA ContentInfo is decrypted, the code allocates a heap buffer, calls DecryptContent() to decrypt in-place, then reassigns the data pointer to the decrypted buffer. However, all subsequent ASN.1 parsing operations continue using ci->dataSz (the original, larger ContentInfo size) as the bounds parameter, instead of the actual decrypted content size returned by DecryptContent().

Fixes zd21568

# Testing

Added `test_wc_PKCS12_encrypted_content_bounds`

# Checklist

 - [x] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/tests/api/test_pkcs12.c
+++ b/tests/api/test_pkcs12.c
@@ -272,6 +272,78 @@ int test_wc_d2i_PKCS12_oid_underflow(void)
     return EXPECT_RESULT();
 }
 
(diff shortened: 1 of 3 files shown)


embhorn:
Jenkins retest this please

SparkiDev (approved):

PR #10182, opened 2026-04-09
Fix TLSX_EchChangeSNI to check hostname termination

embhorn:
# Description

Add NUL to terminate hostname string

Fixes zd21576

# Testing

Added `TLSX_EchChangeSNI`

# Checklist

 - [x] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/src/tls.c
+++ b/src/tls.c
@@ -16106,6 +16106,9 @@ static int TLSX_EchChangeSNI(WOLFSSL* ssl, TLSX** pEchX,
                 hostNameSz = MAX_PUBLIC_NAME_SZ;
 
             XMEMCPY(serverName, hostName, hostNameSz);
+            /* Guarantee NUL termination after truncation so that
+             * TLSX_EchRestoreSNI's XSTRLEN cannot read past the buffer. */
+            serverName[hostNameSz - 1] = '\0';
         }
 
         /* only swap the SNI if one was found; extensions is non-NULL if an
--- a/tests/api.c
+++ b/tests/api.c
@@ -15042,6 +15042,49 @@ static int test_wolfSSL_Tls13_ECH_disable_conn(void)
 
     return EXPECT_RESULT();
 }
+/* Regression test: an inner SNI hostname >= MAX_PUBLIC_NAME_SZ (256) bytes
+ * must not cause a stack-buffer-overflow in TLSX_EchRestoreSNI.  Before the
+ * fix, the truncated copy omitted the NUL terminator and XSTRLEN read past
+ * the buffer. */
+static int test_wolfSSL_Tls13_ECH_long_SNI(void)
+{
+    EXPECT_DECLS;
+#if !defined(NO_WOLFSSL_CLIENT)
+    test_ssl_memio_ctx test_ctx;
+    /* 300 chars > MAX_PUBLIC_NAME_SZ (256) to exercise truncation */
+    char longName[300];
+
+    XMEMSET(longName, 'a', sizeof(longName) - 1);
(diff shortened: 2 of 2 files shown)


src/tls.c
```diff
@@ -16106,6 +16106,9 @@ static int TLSX_EchChangeSNI(WOLFSSL* ssl, TLSX** pEchX,
                 hostNameSz = MAX_PUBLIC_NAME_SZ;
 
             XMEMCPY(serverName, hostName, hostNameSz);
+            /* Guarantee NUL termination after truncation so that
+             * TLSX_EchRestoreSNI's XSTRLEN cannot read past the buffer. */
+            serverName[hostNameSz - 1] = '\0';
         }
 
         /* only swap the SNI if one was found; extensions is non-NULL if an
```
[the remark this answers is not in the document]
embhorn: The Fenrir finding is incorrect. It assumes hostNameSz is the string length   
  without the NUL terminator, but line 16102 defines it as:                     
                                                                                
  word32 hostNameSz = (word32)XSTRLEN(hostName) + 1;                            
                                                                                
  So hostNameSz includes the NUL. Walking through the two cases at              
  src/tls.c:16100-16111:
                                                                                
  1. Non-truncated (e.g. "example.com", hostNameSz = 12): XMEMCPY copies all 12 
  bytes including the trailing NUL, then serverName[11] = '\0' overwrites the
  byte that is already '\0'. Harmless — the m is at index 10, not 11.           
  2. Truncated (hostNameSz > MAX_PUBLIC_NAME_SZ → clamped to
  MAX_PUBLIC_NAME_SZ): XMEMCPY copies MAX_PUBLIC_NAME_SZ bytes of raw hostname  
  (no NUL), then serverName[MAX_PUBLIC_NAME_SZ - 1] = '\0' correctly
  NUL-terminates in-bounds. This is exactly the case the comment at 16109-16110 
  is guarding.                                              

  The bot's suggested "fix" using serverName[hostNameSz] = '\0' would actually  
  be a 1-byte out-of-bounds write in the truncation case (writing at index
  MAX_PUBLIC_NAME_SZ in a MAX_PUBLIC_NAME_SZ-sized buffer), unless the buffer is
   oversized — which the current code does not rely on.

embhorn:
Jenkins retest this please

SparkiDev (approved):

src/tls.c
```diff
@@ -16106,6 +16106,9 @@ static int TLSX_EchChangeSNI(WOLFSSL* ssl, TLSX** pEchX,
                 hostNameSz = MAX_PUBLIC_NAME_SZ;
 
             XMEMCPY(serverName, hostName, hostNameSz);
+            /* Guarantee NUL termination after truncation so that
+             * TLSX_EchRestoreSNI's XSTRLEN cannot read past the buffer. */
+            serverName[hostNameSz - 1] = '\0';
```
SparkiDev, line 16111: Why hostNameSz - 1?
Won't this remove the last character?

SparkiDev (changes requested):

src/tls.c
embhorn: Yes, that's exactly what is does. This is to enforce the MAX_PUBLIC_NAME_SZ and ensure the upcoming strlen call does not read past the buffer.

SparkiDev: OK hostName will always have a terminating NUL unless hostNameSz is greater than MAX_PUBLIC_NAME_SZ.

SparkiDev (approved):

PR #10169, opened 2026-04-08
Fix for peer cert verify with IP address

embhorn:
# Description

Issue with checking SAN with IP certs

Fixes zd21565

# Testing

Added test cases to `test_wolfSSL_X509_check_ip_asc`

# Checklist

 - [x] added tests
 - [ ] updated/added doxygen
 - [ ] updated appropriate READMEs
 - [ ] Updated manual and documentation

PR #10183, opened 2026-04-09
20260409-IsValidFQDN

douzzer:
`src/internal.c`: add `IsValidFQDN()`, and in `MatchDomainName()`, do pattern matching and case folding only if target string `IsValidFQDN()`.

tested with
```
wolfssl-multi-test.sh ...
check-source-text
clang-tidy-all-sp-all
```

see #10169 and ZD #21565

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -13334,6 +13334,22 @@ int MatchDomainName(const char* pattern, int patternLen, const char* str,
         return 1;
 #endif
 
+    if (leftWildcardOnly && (! wolfssl_local_IsValidFQDN(str, strLen))) {
+        /* Not a valid FQDN -- require byte-exact match, no case folding, no
+         * wildcard interpretation.  This is appropriate for an IPv4 match, for
+         * example, but also matches improvised names like "localhost", albeit
+         * case-sensitively.
+         */
+        return (((word32)patternLen == strLen) &&
+                (XMEMCMP(pattern, str, patternLen) == 0));
+    }
+
+    /* strip trailing dots if necessary (FQDN designator). */
+    if (str[strLen-1] == '.')
+        --strLen;
+    if (pattern[patternLen-1] == '.')
+        --patternLen;
+
     while (patternLen > 0) {
         /* Get the next pattern char to evaluate */
         char p = (char)XTOLOWER((unsigned char)*pattern);
--- a/tests/api/test_ossl_x509.c
+++ b/tests/api/test_ossl_x509.c
@@ -1081,10 +1081,18 @@ int test_wolfSSL_X509_check_ip_asc(void)
         ExpectIntEQ(wolfSSL_X509_check_ip_asc(cn_lit, "127.0.0.1", 0), 0);
         /* CN=*.0.0.1 with no SAN must NOT wildcard-match "127.0.0.1". */
         ExpectIntEQ(wolfSSL_X509_check_ip_asc(cn_wild, "127.0.0.1", 0), 0);
+
         /* CN-based hostname matching must still work for hostname checks
          * (sanity check that the fix didn't over-correct). */
         ExpectIntEQ(wolfSSL_X509_check_host(cn_wild, "1.0.0.1",
             XSTRLEN("1.0.0.1"), 0, NULL), 1);
+
+        /* However, when WOLFSSL_LEFT_MOST_WILDCARD_ONLY, CN-based hostname
+         * matching must not apply wildcards when the supplied hostname isn't a
+         * well-formed FQDN.
+         */
+        ExpectIntEQ(wolfSSL_X509_check_host(cn_wild, "1.0.0.1",
+            XSTRLEN("1.0.0.1"), WOLFSSL_LEFT_MOST_WILDCARD_ONLY, NULL), 0);
 
         wolfSSL_X509_free(cn_wild);
         wolfSSL_X509_free(cn_lit);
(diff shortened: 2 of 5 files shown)


src/internal.c
```diff
@@ -13272,6 +13272,76 @@ static int MatchIPv6(const char* pattern, int patternLen,
...
+
+    /* "Fully qualified" requires at least two labels, and the TLD must not
+     * be purely numeric (no such TLD exists; also rejects bare IPv4).
+     */
+    return (labelCount >= 2 && tldHasAlpha);
+}
+
 /* Match names with wildcards, each wildcard can represent a single name
    component or fragment but not multiple names, i.e.,
    *.z.com matches y.z.com but not x.y.z.com
```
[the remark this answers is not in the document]
douzzer: Not sure what Fenrir was talking about here.  In any case, to accommodate legacy use of `localhost` as a CN, the code now requires only `labelCount > 0`.

src/internal.c
```diff
@@ -13272,6 +13272,78 @@ static int MatchIPv6(const char* pattern, int patternLen,
...
+ *   - Internationalized names are valid in their ACE/punycode (xn--) form
+ */
+static int IsValidFQDN(const char* name, word32 nameSz)
+{
+    word32 i;
+    int labelLen      = 0;
+    int labelCount    = 0;
+    int tldHasAlpha   = 0;   /* tracks alpha presence in the *current* label */
+
+    if (name == NULL || nameSz <= 0)
```
dgarske, line 13295: **🟡 [Medium] Unsigned `nameSz <= 0` comparison is tautologically equivalent to `== 0`**
*💡 SUGGEST*  `bug`

`nameSz` is declared as `word32`, which is `typedef unsigned int word32` (or `unsigned long`). The comparison `nameSz <= 0` can never be true for the `< 0` branch — it is always equivalent to `nameSz == 0`. Clang may emit `-Wtautological-compare` warnings under `-Wall`, and GCC will warn with `-Wtype-limits` (part of `-Wextra`). The PR mentions testing with `clang-tidy-all-sp-all`, so this might already be caught, but it should be fixed to avoid surprise warnings in downstream builds.

**Suggestion:**
```suggestion
if (name == NULL || nameSz == 0)
        return 0;
```

src/internal.c
```diff
@@ -13272,6 +13272,78 @@ static int MatchIPv6(const char* pattern, int patternLen,
 }
 #endif /* WOLFSSL_IP_ALT_NAME && !WOLFSSL_USER_IO */
 
+/* Returns 1 if name is a syntactically valid DNS FQDN per RFC 952/1123.
+ *
+ * Rules enforced:
+ *   - Total effective length (excluding optional trailing dot) in [1, 253]
+ *   - Each label is 1-63 octets of [a-zA-Z0-9-]
+ *   - No label starts or ends with '-'
+ *   - At least two labels (single-label names are not "fully qualified")
```
dgarske, line 13281: **🔵 [Low] Header comment says 'at least two labels' but code intentionally allows single labels**
*🔧 NIT*  `style`

The function header comment states "At least two labels (single-label names are not 'fully qualified')" as a rule that is enforced, but the code at line 13344 returns `(labelCount > 0) && tldHasAlpha`, deliberately allowing single-label names (as the inline comment explains: "we allow single labels, so that ad hoc 'localhost' CN is allowed"). The header comment should match the actual behavior to avoid confusing future readers.

**Suggestion:**
```suggestion
*   - Labels must have at least one alpha character in the final label (TLD)
 *   - Single-label names are allowed (for ad-hoc names like "localhost")
```

dgarske (commented):
## 🐺 Skoll Code Review

### Posted findings
- [Medium] **Unsigned `nameSz <= 0` comparison is tautologically equivalent to `== 0`** — `src/internal.c:13295`
- [Low] **Header comment says 'at least two labels' but code intentionally allows single labels** — `src/internal.c:13281`
- [Medium] **Trailing dot stripped from `str` but not from `pattern` — asymmetric matching** — `src/internal.c:13380-13382`
- [Medium] **Underscore-containing DNS names lose case-insensitive matching** — `src/internal.c:13371-13378`
- [Medium] **No unit tests for `IsValidFQDN` or the new `MatchDomainName` early-return behavior** — `src/internal.c:13288-13345`

src/internal.c
```diff
@@ -13296,6 +13368,19 @@ int MatchDomainName(const char* pattern, int patternLen, const char* str,
...
+    if (! IsValidFQDN(str, strLen)) {
+        /* Not a valid FQDN or IPv6 address -- require byte-exact match, no case
+         * folding, no wildcard interpretation.  This is appropriate for an IPv4
+         * match, for example.
+         */
+        return (((word32)patternLen == strLen) &&
+                (XMEMCMP(pattern, str, patternLen) == 0));
+    }
+
+    /* strip trailing dot if necessary (FQDN designator). */
```
dgarske, line 13380: **🟡 [Medium] Trailing dot stripped from `str` but not from `pattern` — asymmetric matching**
*💡 SUGGEST*  `question`

After the `IsValidFQDN` check, the trailing dot is stripped from `str` but not from `pattern`. This means `str="example.com."` will match `pattern="example.com"` (good), but `str="example.com"` will NOT match `pattern="example.com."` because the pattern's trailing dot persists and exhausts the character-by-character loop after `str` is consumed. While certificates with trailing dots in SANs/CNs are unusual, RFC 5280 does not explicitly forbid them, and some CAs may produce them. This creates an asymmetry that could cause surprising match failures.

**Suggestion:**
```suggestion
/* strip trailing dot if necessary (FQDN designator). */
    if (str[strLen-1] == '.')
        --strLen;
    if (patternLen > 0 && pattern[patternLen-1] == '.')
        --patternLen;
```

src/internal.c
```diff
@@ -13296,6 +13368,19 @@ int MatchDomainName(const char* pattern, int patternLen, const char* str,
         return 1;
 #endif
 
+    if (! IsValidFQDN(str, strLen)) {
```
dgarske, line 13371: **🟡 [Medium] Underscore-containing DNS names lose case-insensitive matching**
*💡 SUGGEST*  `question`

DNS names containing underscores (e.g. `_dmarc.example.com`, `_sip._tcp.example.com`) are used in practice for SRV, DKIM, and DMARC records. While RFC 952/1123 does not allow underscores, RFC 4592/5280 and real-world certificates sometimes contain them. `IsValidFQDN` rejects such names (underscore falls through the `[a-zA-Z0-9-]` check), causing `MatchDomainName` to use the byte-exact `XMEMCMP` fallback — which means case-insensitive matching and wildcard matching are lost for these names. If a certificate has `_DMARC.EXAMPLE.COM` and the user checks `_dmarc.example.com`, the match will fail where it previously would have succeeded. This is likely acceptable given the PR's goals, but the behavioral change should be documented.

**Recommendation:** Consider whether `IsValidFQDN` should also accept underscores (expand the character set to `[a-zA-Z0-9_-]`), or document this as an intentional restriction. Underscored hostnames are uncommon in TLS certificates but not impossible.

src/internal.c
```diff
@@ -13272,6 +13272,78 @@ static int MatchIPv6(const char* pattern, int patternLen,
...
+ *   - Each label is 1-63 octets of [a-zA-Z0-9-]
+ *   - No label starts or ends with '-'
+ *   - At least two labels (single-label names are not "fully qualified")
+ *   - Final label (TLD) contains at least one letter (rejects all-numeric
+ *     strings that could be confused with IPv4 literals, and matches the
+ *     ICANN constraint that TLDs are alphabetic)
+ *   - Optional trailing dot is accepted (absolute FQDN form)
+ *   - Internationalized names are valid in their ACE/punycode (xn--) form
+ */
+static int IsValidFQDN(const char* name, word32 nameSz)
```
dgarske, line 13288: **🟡 [Medium] No unit tests for `IsValidFQDN` or the new `MatchDomainName` early-return behavior**
*💡 SUGGEST*  `test`

`IsValidFQDN` is a new 57-line function with complex validation logic covering multiple edge cases (empty labels, leading/trailing hyphens, label length limits, trailing dots, TLD alpha requirement, total length limit). There are no direct unit tests for this function. The existing `test_wolfSSL_X509_name_match*` and `test_wolfSSL_check_domain*` tests exercise `MatchDomainName` indirectly, but they don't specifically test the new FQDN-vs-non-FQDN branching behavior or the full range of `IsValidFQDN` edge cases. Key scenarios that should be tested:
- IPv4 address (`"192.168.1.1"`) → exact match path
- Valid FQDN (`"foo.example.com"`) → wildcard/case-fold path
- Trailing dot FQDN (`"example.com."`) → valid FQDN, dot stripped
- Single label with alpha (`"localhost"`) → valid FQDN path
- All-numeric TLD (`"123.456"`) → exact match path
- Label starting/ending with hyphen → exact match path
- 253-byte boundary cases
- Underscore-containing names → exact match path

**Recommendation:** Add unit tests that call `MatchDomainName` with both FQDN and non-FQDN `str` values to verify the branching behavior. If `IsValidFQDN` cannot be tested directly (it's `static`), consider adding a wrapper or testing through `MatchDomainName`'s observable behavior.

douzzer: fixed

src/internal.c
douzzer: fixed

src/internal.c
douzzer: fixed

src/internal.c
douzzer: fixed -- single labels not allowed (fall through to memcmp).

src/internal.c
douzzer: fixed

douzzer:
retest this please
(no error log)

embhorn (approved):
Looks good. I have a PR draft that will build on this.

dgarske (approved):

PR #10187, opened 2026-04-10
Fixes in TLS ECH, handle empty records, and ASN len check

embhorn:
# Description

* Fix TLSX_Parse to check dup ECH
* Add WOLFSSL_MAX_EMPTY_RECORDS
* Fix ParseCertRelative to check pathlen on self-issued certs

Fixes zd21587

# Credits

Thanks to Asaf Meizner for the report and review of the fix.

# Testing

Added tests:
* `test_tls13_duplicate_ech_extension`
* `test_tls13_empty_record_limit`
* `test_PathLenSelfIssued` and `test_PathLenNoKeyUsage`

# Checklist

 - [x] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -21862,6 +21862,23 @@ int DoApplicationData(WOLFSSL* ssl, byte* input, word32* inOutIdx, int sniff)
     }
 #endif
 
+    /* Rate-limit empty application data records to prevent DoS */
+    if (dataSz == 0) {
+        if (++ssl->options.emptyRecordCount >= WOLFSSL_MAX_EMPTY_RECORDS) {
+            WOLFSSL_MSG("Too many empty records");
+#ifdef WOLFSSL_EXTRA_ALERTS
+            if (sniff == NO_SNIFF) {
+                SendAlert(ssl, alert_fatal, unexpected_message);
+            }
+#endif
+            WOLFSSL_ERROR_VERBOSE(EMPTY_RECORD_LIMIT_E);
+            return EMPTY_RECORD_LIMIT_E;
+        }
+    }
+    else {
+        ssl->options.emptyRecordCount = 0;
+    }
+
     /* read data */
     if (dataSz) {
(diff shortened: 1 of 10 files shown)


AsafMeizner:
Reviewed all three commits. Fixes look correct.

**Fix 1: Duplicate ECH Extension (1f75deb)** 

Adding `TLSX_ECH` to the duplicate-detection gate works. Test correctly crafts a ClientHello with two ECH extensions and verifies rejection.

Note: this fixes ECH specifically, but extension types > 62 outside the hardcoded exceptions still bypass duplicate detection. A more general approach (hash set for seen types) would future-proof this, though that's a larger change.

**Fix 2: Empty Record Limit (d2a8735)** 

Counter with default 32 matching OpenSSL, reset on non-empty — correct. Both test cases (exceeded limit + counter reset) have good coverage.

Issue: `emptyRecordCount` is `byte` but `WOLFSSL_MAX_EMPTY_RECORDS` is configurable. Values > 255 wrap the counter and defeat the limit. Either widen to `word16`/`word32` or add:

```c
#if WOLFSSL_MAX_EMPTY_RECORDS > 255
    #error "WOLFSSL_MAX_EMPTY_RECORDS must be <= 255 (counter is byte)"
#endif
```

**Fix 3: pathLen for Self-Issued Certs (d6fa2b1)** 

This is the most important one. Changes are correct:
- pathLen no longer gated on `extKeyUsageSet` — handles absent KeyUsage per RFC 5280 4.2.1.3
- Self-issued certs enforce issuer's `maxPathLen` without decrementing — matches RFC 5280 6.1.4(l)
- `maxPathLen` widened from `byte` to `word16`

Tests `test_PathLenSelfIssued` and `test_PathLenNoKeyUsage` cover both attack scenarios and verify `ASN_PATHLEN_INV_E`.

Thanks for the fast turnaround - all three issues addressed with tests in under a day is impressive.

JacobBarthelmeh (approved):

PR #10189, opened 2026-04-10
ci: membrowse integration

michael-membrowse:
# Description

MemBrowse memory footprint tracking in CI (Following a discussion with wolfSSL team)

I've configured four targets in membrowse-targets.json, all on Cortex-M4 with minimal test code linked so it doesn't affect the memory tracking:
- gcc-arm-cortex-m4
- gcc-arm-cortex-m4-baremetal
- gcc-arm-cortex-m4-min-ecc
- gcc-arm-cortex-m4-tls12

It includes:
- membrowse-targets.json - defines the build targets with setup and build commands
- membrowse-report.yml - runs on every push to master and on PRs
- membrowse-comment.yml - comments on PRs with a memory usage summary
- membrowse-onboard.yml - one-time workflow to load historical reports

To get started:
1. Create an account at [membrowse.com](http://membrowse.com/)
2. Create a project to get an API key.
3. Set it as a GitHub secret called MEMBROWSE_API_KEY

The workflows will handle the rest.

Full docs are available at [docs.membrowse.com](http://docs.membrowse.com/).

Happy to help if any questions come up during integration.

### DIFF
--- a/.github/membrowse-targets.json
+++ b/.github/membrowse-targets.json
@@ -1,0 +1,42 @@
+[
+  {
+    "target_name": "gcc-arm-cortex-m4",
+    "port": "gcc-arm",
+    "board": "cortex-m4",
+    "setup_cmd": "sudo apt-get update && sudo apt-get install -y gcc-arm-none-eabi libnewlib-arm-none-eabi libstdc++-arm-none-eabi-newlib",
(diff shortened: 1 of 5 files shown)


dgarske:
Okay to test. We are in contact with contributor.

michael-membrowse:
Create a project for wolfssl on membrowse.com so the report jobs would not fail :)

dgarske (approved):

PR #10200, opened 2026-04-13
Added Unit tests to increase MC/DC coverage

danielinux:


danielinux:
overridden by #10845 .

PR #10206, opened 2026-04-13
Store the size of the allocated private key buffer for ML-KEM

Frauschi:
Follow-up for #10179: when `WOLFSSL_MLKEM_DYNAMIC_KEYS` is enabled, store the size of the allocated private key buffer for later `ForceZero()` calls. This prevents potential erroneous corruption via a wrong `k` parameter.

### DIFF
--- a/wolfssl/wolfcrypt/wc_mlkem.h
+++ b/wolfssl/wolfcrypt/wc_mlkem.h
@@ -104,6 +104,10 @@ enum {
 struct MlKemKey {
     /* Type of key: WC_ML_KEM_512, WC_ML_KEM_768, WC_ML_KEM_1024 */
     int type;
+#ifdef WOLFSSL_MLKEM_DYNAMIC_KEYS
+    /* Allocated size of priv buffer in bytes. */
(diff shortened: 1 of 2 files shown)


dgarske (dismissed):

wolfssl/wolfcrypt/wc_mlkem.h
```diff
@@ -137,6 +137,8 @@ struct MlKemKey {
     sword16* priv;
     /* Public key as a vector (dynamically allocated). */
     sword16* pub;
+    /* Allocated size of priv buffer in bytes. */
+    word32 privAllocSz;
```
SparkiDev: Move this up to after type if this makes the structure not increase in size on some platforms.

SparkiDev (dismissed):

wolfssl/wolfcrypt/wc_mlkem.h
Frauschi: Fixed. This saves 8 byte.

dgarske (approved):

PR #10207, opened 2026-04-13
Add signed-length validation to d2i, PEM, and buffer-load APIs

ColtonWilley:
Internal security review found that ~25 public API entry points accept signed
length parameters (long/int) and cast them to word32 or size_t without rejecting
negative values first. A negative length wraps to a huge unsigned value, giving
ASN.1 parsers and memcpy calls a bogus bound.

Three main areas fixed:

- d2i_* OpenSSL compat wrappers (ECDSA_SIG, RSA, X509, OCSP, etc.) — add
  len <= 0 guards before the (word32) casts
- ProcessBuffer and PemToDer shared sinks — one guard each closes ~20 buffer-load
  entry points and all PEM conversion paths respectively
- Certgen helpers (wc_SetIssuerBuffer, wc_SetSubjectBuffer, etc.) — reject
  negative derSz before forwarding to wc_SetCert_LoadDer

Also fixes an integer overflow in wolfSSL_ASN1_STRING_set where sz == INT_MAX
caused sz + 1 to wrap, bypassing the buffer size check.

8 locations were audited and confirmed already safe (existing guards catch
negatives). Those are left unchanged.

Tests cover ProcessBuffer, wc_Set*Buffer, and d2i_ECDSA_SIG (negative len + NULL deref).

### DIFF
--- a/src/crl.c
+++ b/src/crl.c
@@ -834,7 +834,7 @@ int BufferLoadCRL(WOLFSSL_CRL* crl, const byte* buff, long sz, int type,
 
     WOLFSSL_ENTER("BufferLoadCRL");
 
-    if (crl == NULL || buff == NULL || sz == 0)
+    if (crl == NULL || buff == NULL || sz <= 0)
         return BAD_FUNC_ARG;
 
     if (type == WOLFSSL_FILETYPE_PEM) {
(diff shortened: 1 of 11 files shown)


dgarske (approved):
I think there are some other functions that will need this to, but this is a great start.

PR #10209, opened 2026-04-13
Harden chain depth bounds and parser input validation

ColtonWilley:
## Summary

Harden certificate-chain depth handling and parser input validation.

- Enforce `MAX_CHAIN_DEPTH` in certificate loading (`ProcessUserChain`), OCSP
  stapling (`SendCertificateStatus`), and TLS 1.3 certificate send
  (`SendTls13Certificate`). The load-time check runs after a cert is parsed,
  so a full-depth chain followed by trailing data still loads.
- Add `idx` bounds checks to the peer-chain accessors in `ssl.c`.
- `SendCertificateStatus`: bound the OCSP chain loops by `MAX_CHAIN_DEPTH` and
  error on an over-long chain rather than silently truncating the response.
- Harden `TLSX_SNI_GetFromBuffer` length validation (require
  `listLen == extLen - OPAQUE16_LEN`, bound `sniLen`) against overreads on a
  malformed ClientHello.
- Fix an off-by-one in `TLSX_CSR_Free` (`<=` to `<`).
- Bounds-check `wc_PKCS7_DecodeAuthEnvelopedData` against truncated input
  (`encryptedContentSz`, the authTag tag read, `authTagSz`) under `NO_PKCS7_STREAM`.

## Tests

- Chain-depth boundary and chain-accessor index bounds in `tests/api.c`.
- AuthEnvelopedData truncation cases in `test_pkcs7.c`.
- Remove `test_dtls13_oversized_cert_chain`, superseded by the load-time cap.


tests/api/test_dtls.c
```diff
@@ -3035,101 +3035,6 @@ int test_dtls13_no_session_id_echo(void)
     return EXPECT_RESULT();
 }
 
-/* Test that a DTLS 1.3 handshake with an oversized certificate chain does
```
ColtonWilley: This is a regression test for a specific length check, however its no longer reachable in typical situations as you would need to load > max cert chain certs to yield such a length.

julek-wolfssl (dismissed):
Copilot has some valid points

wolfcrypt/src/pkcs7.c
```diff
@@ -14968,6 +14968,12 @@ int wc_PKCS7_DecodeAuthEnvelopedData(wc_PKCS7* pkcs7, byte* in,
                 }
             }
 
+            if (ret == 0 && encryptedContentSz > (int)(pkiMsgSz - idx)) {
+            #ifdef NO_PKCS7_STREAM
+                ret = BUFFER_E;
+            #endif
+            }
```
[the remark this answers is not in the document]
ColtonWilley: Fixed in 0ca238c5 — moved the whole check under `#ifdef NO_PKCS7_STREAM`, so the default (streaming) build carries no dead branch and it matches the surrounding `#ifdef`-per-statement style in this file. One nit on the rationale: the explicit `{ }` braces wouldn't actually trip `-Wempty-body` (that only fires on a bare `;`; gcc/clang's own fix for it is to add the braces this code already had) — but scoping the check is cleaner regardless, so the change stands.

tests/api/test_pkcs7.c
```diff
@@ -2873,6 +2873,65 @@ int test_wc_PKCS7_DecodeEnvelopedData_forgedRecipientSetLen(void)
...
+        sizeof_client_cert_der_2048), 0);
+    if (pkcs7 != NULL) {
+        pkcs7->privateKey   = (byte*)client_key_der_2048;
+        pkcs7->privateKeySz = sizeof_client_key_der_2048;
+    }
+    ExpectIntEQ(wc_PKCS7_DecodeAuthEnvelopedData(pkcs7, enveloped,
+        (word32)encSz - 32, decoded, sizeof(decoded)),
+        WC_NO_ERR_TRACE(BUFFER_E));
+    wc_PKCS7_Free(pkcs7);
+    pkcs7 = NULL;
```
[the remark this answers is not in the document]
ColtonWilley: Fixed in 0ca238c5 — the encode is now asserted `> 32`, so the `encSz - 32` / `encSz - 1` truncations below can't underflow. Note the `Expect*` macros short-circuit after a failure (each body is gated on `EXPECT_SUCCESS()`), so the `encSz <= 0` case was already covered; this tightens the assertion to also close the `0 < encSz <= 32` window. A valid RSA-2048 / AES-GCM AuthEnvelopedData encode never lands there, but worth pinning down in the test.

ColtonWilley:
> Copilot has some valid points

Agreed, I have pushed updates to address the issues.

ColtonWilley:
Jenkins retest this please

dgarske (commented):
## Skoll Code Review

### Posted findings
- [Medium] **MAX_CHAIN_DEPTH check placement can reject a valid maximum-depth chain that has trailing data** — `src/ssl_load.c:328-333`
- [Low] **Max-depth load test overshoots the limit and omits the exact-boundary success case** — `tests/api.c:3866-3915`
- [Info] **PR description lists pkcs7.c changes that are not present in the diff** — `wolfcrypt/src/pkcs7.c:14971-14975,15239-15243`
- [Info] **SendTls13Certificate depth guard is now defense-in-depth only and its direct test path was removed** — `src/tls13.c:9523-9526`

src/ssl_load.c
```diff
@@ -325,6 +325,13 @@ static int ProcessUserChain(WOLFSSL_CTX* ctx, WOLFSSL* ssl,
         while ((ret == 0) && (consumed < sz)) {
             DerBuffer* part = NULL;
 
+            /* Enforce maximum chain depth. */
```
dgarske: **🟠 [Medium] MAX_CHAIN_DEPTH check placement can reject a valid maximum-depth chain that has trailing data**

The new depth check `if (cnt >= MAX_CHAIN_DEPTH)` is placed at the TOP of the parse loop, before `DataToDerBuffer`. The loop is commonly entered one extra time after the last real certificate to consume trailing data (trailing newline/whitespace), and that extra entry is normally handled gracefully by the `ret == ASN_NO_PEM_HEADER && gotOne` branch further down. For a chain of EXACTLY `MAX_CHAIN_DEPTH` (default 9) valid certificates followed by any unconsumed trailing bytes, `cnt` equals `MAX_CHAIN_DEPTH` when the loop re-enters, so the new check fires FIRST and returns `MAX_CHAIN_ERROR` instead of letting the existing `ASN_NO_PEM_HEADER` path terminate cleanly with `ret = 0`. This regresses a valid maximum-length chain to a load failure. Note `certs/server-cert.pem` in this repo is itself a 2-cert PEM, so multi-cert PEM with trailing bytes is the norm; the boundary is real, just narrow (requires hitting exactly the max depth). The PR's own new test uses `MAX_CHAIN_DEPTH + 1` copies (overshoots) and never exercises the exact boundary.

**Fix:** Move the `cnt >= MAX_CHAIN_DEPTH` check so it only rejects when an additional real certificate is actually present (i.e., after `DataToDerBuffer` and after the `ASN_NO_PEM_HEADER && gotOne` trailing-data handling). Add a test that loads exactly `MAX_CHAIN_DEPTH` certificates (with a trailing newline) and asserts success, in addition to the over-limit case.

tests/api.c
```diff
@@ -3831,6 +3831,89 @@ static int test_wolfSSL_CTX_use_certificate_chain_buffer_format(void)
...
+    ExpectIntEQ(wolfSSL_get_chain_length(chain, count), 0);
+    ExpectIntEQ(wolfSSL_get_chain_length(chain, MAX_CHAIN_DEPTH), 0);
+    ExpectNull(wolfSSL_get_chain_cert(chain, -1));
+    ExpectNull(wolfSSL_get_chain_cert(chain, count));
+    ExpectNull(wolfSSL_get_chain_cert(chain, MAX_CHAIN_DEPTH));
+#ifdef OPENSSL_EXTRA
+    {
+        WOLFSSL_X509* x = NULL;
+        ExpectNull(x = wolfSSL_get_chain_X509(chain, -1));
+        if (x != NULL) { wolfSSL_X509_free(x); x = NULL; }
```
dgarske, line 3891: **🔵 [Low] Max-depth load test overshoots the limit and omits the exact-boundary success case**

The test sets `nCerts = MAX_CHAIN_DEPTH + 1` copies of `svrCertFile`, but `server-cert.pem` contains 2 certificates, so the buffer actually holds ~2*(MAX_CHAIN_DEPTH+1) certs. It therefore only proves that a grossly-oversized chain is rejected and does not pin the boundary. It is missing the complementary positive case: a chain of exactly `MAX_CHAIN_DEPTH` certificates must still load successfully. That gap is exactly what would let the placement issue in `ProcessUserChain` (separate finding) slip through. The `const int nCerts = MAX_CHAIN_DEPTH + 1;` naming/comment is also misleading given each file is multi-cert.

**Fix:** Add a positive boundary test (exactly MAX_CHAIN_DEPTH certs -> success) and clarify the per-file cert count in the comment.

wolfcrypt/src/pkcs7.c
```diff
@@ -14968,6 +14968,12 @@ int wc_PKCS7_DecodeAuthEnvelopedData(wc_PKCS7* pkcs7, byte* in,
                 }
             }
 
+        #ifdef NO_PKCS7_STREAM
```
dgarske, line 15025: **⚪ [Info] PR description lists pkcs7.c changes that are not present in the diff**

The PR body claims bounds checks were added to `DecodeEnvelopedData`, `DecodeEncryptedData` (two sites), `SignedData` null signature tag, and `PwriKek_KeyUnWrap` (cekLen off-by-4). The actual pkcs7.c diff only contains the two `wc_PKCS7_DecodeAuthEnvelopedData` checks (`encryptedContentSz` and `authTagSz`, both under `#ifdef NO_PKCS7_STREAM`). Either the commit is incomplete or the description is stale. The two checks that ARE present are correct: `idx` is advanced only past validated length bytes so `pkiMsgSz - idx` cannot underflow, and the `authTagSz` check is a genuine fix (the existing `authTagSz > sizeof(authTag)` check only guards the destination, not the `pkiMsg` source read at the subsequent XMEMCPY).

**Fix:** Confirm whether the other PKCS7 decoder hardening described in the PR body was intended for this commit. If not, trim the description to match the diff so reviewers/changelog stay accurate.

src/tls13.c
```diff
@@ -9520,6 +9520,10 @@ static int SendTls13Certificate(WOLFSSL* ssl)
         }
         /* Certificate Data */
         certSz = ssl->buffers.certificate->length;
+        if (ssl->buffers.certChainCnt > MAX_CHAIN_DEPTH) {
```
dgarske, line 9619: **⚪ [Info] SendTls13Certificate depth guard is now defense-in-depth only and its direct test path was removed**

With the new `ProcessUserChain` limit, an oversized chain can no longer be loaded (it fails with MAX_CHAIN_ERROR at load time), so the new `ssl->buffers.certChainCnt > MAX_CHAIN_DEPTH` guard inside `SendTls13Certificate` is effectively unreachable via the normal load path -- it is a reasonable belt-and-suspenders check, but untested. The PR simultaneously deletes `test_dtls13_oversized_cert_chain`, which was the only test that drove an oversized chain through the actual TLS1.3 certificate-send path. The deletion itself is justified (that test expected `WOLFSSL_SUCCESS` on load, which is no longer possible), and the `test_dtls13.h` macro list trailing-comma update is correct. Net effect is reduced direct coverage of the send-side guard.

**Fix:** Optional: keep the guard (cheap and defensive) but note in a comment that it is unreachable under normal loading, or add a unit test that sets ssl->buffers.certChainCnt directly to exercise the MAX_CHAIN_ERROR return.

src/ssl_load.c
ColtonWilley: Agreed, fixed in latest commit

tests/api.c
ColtonWilley: agreed, added a new test for at limit.

wolfcrypt/src/pkcs7.c
ColtonWilley: fixed in latest commit

src/tls13.c
ColtonWilley: I prefer to keep as is.

julek-wolfssl (approved):

Frauschi (approved):

douzzer (approved):

PR #10210, opened 2026-04-13
Fix dangling secure_renegotiation pointer after TLSX_FreeAll

ColtonWilley:
## Summary
- `ssl->secure_renegotiation` caches a pointer into extension data owned by the `ssl->extensions` list. Three call sites free that list via `TLSX_FreeAll` without NULLing the cached pointer, leaving it dangling: `wolfSSL_clear()`, `FreeHandshakeResources()`, and `wolfSSL_ResourceFree()`.
- After `wolfSSL_clear()`, calling `wolfSSL_SSL_get_secure_renegotiation_support()` reads the freed `SecureRenegotiation` struct. Confirmed heap-use-after-free under ASan with nginx, haproxy, and openssl-compat build profiles.
- NULL the pointer at all three sites. Add regression test covering the `wolfSSL_clear` path.

## Test plan
- [x] Existing CI passes
- [x] New `test_wolfSSL_clear_secure_renegotiation` passes
- [x] ASan build does not report heap-use-after-free on the wolfSSL_clear → wolfSSL_SSL_get_secure_renegotiation_support sequence

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -8953,10 +8953,10 @@ void wolfSSL_ResourceFree(WOLFSSL* ssl)
 #if !defined(NO_TLS)
     TLSX_FreeAll(ssl->extensions, ssl->heap);
     ssl->extensions = NULL;
-  #if defined(HAVE_SECURE_RENEGOTIATION) \
-   || defined(HAVE_SERVER_RENEGOTIATION_INFO)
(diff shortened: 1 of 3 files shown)


mattia-moffa:
Jenkins retest this please

JacobBarthelmeh (approved):

PR #10212, opened 2026-04-13
Harden X509 DER length handling in wolfSSL_X509_get_der and wolfSSL_i2d_X509

ColtonWilley:
## Summary
- **src/x509.c**: Guard `wolfSSL_X509_get_der` against `derCert->length > INT_MAX`, and reject `derSz <= 0` in `wolfSSL_i2d_X509`.
- **tests/api/test_ossl_x509_io.{c,h}**: Add API coverage for the X509 DER length guards.

## Note
The SRTP KDF work originally mentioned on this branch was absorbed upstream; this PR now contains only X509 DER length hardening.

### DIFF
--- a/src/x509.c
+++ b/src/x509.c
@@ -4404,6 +4404,10 @@ const byte* wolfSSL_X509_get_der(WOLFSSL_X509* x509, int* outSz)
     if (x509 == NULL || x509->derCert == NULL || outSz == NULL)
         return NULL;
 
+    if (x509->derCert->length > (word32)INT_MAX) {
+        return NULL;
+    }
+
     *outSz = (int)x509->derCert->length;
     return x509->derCert->buffer;
 }
@@ -8674,7 +8678,7 @@ int wolfSSL_i2d_X509(WOLFSSL_X509* x509, unsigned char** out)
     }
 
     der = wolfSSL_X509_get_der(x509, &derSz);
-    if (der == NULL) {
+    if (der == NULL || derSz <= 0) {
         WOLFSSL_LEAVE("wolfSSL_i2d_X509", MEMORY_E);
         return MEMORY_E;
     }
--- a/src/x509.c
+++ b/src/x509.c
@@ -4523,6 +4523,10 @@ const byte* wolfSSL_X509_get_der(WOLFSSL_X509* x509, int* outSz)
     if (x509 == NULL || x509->derCert == NULL || outSz == NULL)
         return NULL;
 
+    if (x509->derCert->length > (word32)INT_MAX) {
+        return NULL;
+    }
+
     *outSz = (int)x509->derCert->length;
     return x509->derCert->buffer;
 }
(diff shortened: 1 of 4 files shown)


ColtonWilley:
retest this please

src/x509.c
```diff
@@ -8835,7 +8839,7 @@ int wolfSSL_i2d_X509(WOLFSSL_X509* x509, unsigned char** out)
     }
 
     der = wolfSSL_X509_get_der(x509, &derSz);
-    if (der == NULL) {
+    if (der == NULL || derSz <= 0) {
```
dgarske, line 8842: **🔵 [Low] MEMORY_E returned for zero-length DER is semantically misleading**

The new `derSz <= 0` condition makes `wolfSSL_i2d_X509` return `MEMORY_E` for a zero-length (degenerate/corrupted) DER. Because the new overflow guard in `wolfSSL_X509_get_der` ensures `*outSz` is never negative, `derSz <= 0` is effectively `derSz == 0` (the `< 0` branch is unreachable). `MEMORY_E` is not really a memory-allocation failure here, so it is a slightly odd error code for this condition. That said, the pre-existing `der == NULL` branch already returns `MEMORY_E` for non-allocation failures, so this is consistent with the surrounding code and not worth changing on its own.

**Fix:** Optional: keep as-is for consistency with the existing `der == NULL` path, or consider `BAD_FUNC_ARG` if a more accurate error code is desired. No change required.

src/x509.c
```diff
@@ -4523,6 +4523,10 @@ const byte* wolfSSL_X509_get_der(WOLFSSL_X509* x509, int* outSz)
     if (x509 == NULL || x509->derCert == NULL || outSz == NULL)
         return NULL;
 
+    if (x509->derCert->length > (word32)INT_MAX) {
```
dgarske, line 4526: **⚪ [Info] Overflow guard is a tautology on (extremely rare) 64-bit-int platforms**

`x509->derCert->length > (word32)INT_MAX` is correct and meaningful on the standard 16/32-bit `int` targets wolfSSL supports. On a platform where `int` is 64-bit (ILP64 — not a realistically supported wolfSSL target), `(word32)INT_MAX` truncates to `0xFFFFFFFF` (UINT32_MAX), making the comparison always false and potentially triggering `-Wtype-limits`/`-Wtautological-compare` under `-Werror`. On such a platform the `(int)length` cast is always safe anyway, so behavior is still correct. Flagging only for completeness; no action expected.

**Fix:** No change needed for supported targets. If ILP64 support ever matters, the comparison would simply be a no-op there (and the cast remains safe).

dgarske (commented):
## Skoll Code Review

### Posted findings
- [Low] **MEMORY_E returned for zero-length DER is semantically misleading** — `src/x509.c:8842-8845`
- [Info] **Overflow guard is a tautology on (extremely rare) 64-bit-int platforms** — `src/x509.c:4526-4528`

src/x509.c
ColtonWilley: I think we should keep as is for consistency.

src/x509.c
ColtonWilley: Behavior is correct, I think its probably fine.

dgarske (approved):

PR #10228, opened 2026-04-15
ci: preserve membrowse push-to-master runs from cancellation

michael-membrowse:
# Description
Use unique concurrency group per commit for push events so that sequential merges to master don't cancel each other's queued runs. This ensures every master commit gets analyzed for memory tracking.


LinuxJedi:
Okay to test. We are in contact with contributor.

LinuxJedi (approved):

dgarske (approved):

PR #10232, opened 2026-04-15
20260415-confusing_globals

douzzer:
`src/ssl_load.c`, `wolfssl/ssl.h`, `tests/api.c`: rename `wolfSSL*PrivateKey_id()` to `wolfSSL*PrivateKey_Id_ex()`, and add missing `WOLF_PRIVATE_KEY_ID` gating.

`src/ssl.c`, `src/ssl_sess.c`, `src/x509.c`, `wolfssl/internal.h`: rename `wolfssl_get_ex_new_index()` to `wolfssl_local_get_ex_new_index()`.

tested with
```
wolfssl-multi-test.sh ...
super-quick-check
check-for-bad-symbols
```

Note `check-for-bad-symbols` is the old `check-for-unprefixed-symbols`, now with callouts to a newly implemented `check_libwolfssl_for_confusing_globals()` for each sub-scenario.  Currently `check_libwolfssl_for_confusing_globals()` just checks for globals (either public or hidden) that have the same name when case-folded.  There is one unavoidable exception, `wolfSSL_Free()` and `wolfSSL_free()`, the former our public name for our `free()` wrapper, the latter our public name for our API to free a `WOLFSSL *`.  Both were added in v3.3.3 (Jan 9 2015), both deriving from CyaSSL variants added in v2.0.2 (Dec 5 2011).  We're stuck with those.

### DIFF
--- a/wolfssl/ssl.h
+++ b/wolfssl/ssl.h
@@ -3769,6 +3769,8 @@ WOLFSSL_API int wolfSSL_make_eap_keys(WOLFSSL* ssl, void* key, unsigned int len,
     WOLFSSL_API int wolfSSL_CTX_use_PrivateKey_Id_ex(WOLFSSL_CTX* ctx,
                                                   const unsigned char* id, long sz,
                                                   int devId, long keySz);
+    /* backward compatibility mapping for old confusable name. */
+    #define wolfSSL_CTX_use_PrivateKey_id wolfSSL_CTX_use_PrivateKey_Id_ex
     WOLFSSL_API int wolfSSL_CTX_use_PrivateKey_Id(WOLFSSL_CTX* ctx,
                                                   const unsigned char* id, long sz,
                                                   int devId);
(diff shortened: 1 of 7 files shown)


wolfssl/ssl.h
```diff
@@ -3802,7 +3804,7 @@ WOLFSSL_API int wolfSSL_make_eap_keys(WOLFSSL* ssl, void* key, unsigned int len,
                                            const unsigned char* der, int derSz);
     WOLFSSL_API int wolfSSL_use_PrivateKey_buffer(WOLFSSL* ssl, const unsigned char* in,
                                                long sz, int format);
-    WOLFSSL_API int wolfSSL_use_PrivateKey_id(WOLFSSL* ssl, const unsigned char* id,
+    WOLFSSL_API int wolfSSL_use_PrivateKey_Id_ex(WOLFSSL* ssl, const unsigned char* id,
```
dgarske, line 3807: There is history here and I would prefer not break API compat. Perhaps @SparkiDev can recall what that history is?

dgarske (changes requested):

wolfssl/ssl.h
douzzer: Quite right, figured this out independently and just pushed the fix.  Can't break public APIs.

(I incorrectly thought the APIs were just added.)

dgarske (approved):

Frauschi (approved):

douzzer:
retest this please
(mystery failure on "140-3 tests", no specific test failure show)

PR #10018, opened 2026-03-19
Fix GetSafeContent to check length

embhorn:
# Description

Added bounds check after GetObjectId() in GetSafeContent() to verify the OID didn't consume more bytes than the ContentInfo SEQUENCE declared. Returns ASN_PARSE_E  if (localIdx - curIdx) > (word32)curSz, preventing the unsigned integer underflow in the ci->dataSz calculation.

Credit to `Muhammad Arya Arjuna (pelioro)` for reporting this issue and confirming the fix.

Fixes zd21389

# Testing

Added `test_wc_d2i_PKCS12_oid_underflow`

# Checklist

 - [x] added tests
 - [ ] updated/added doxygen
 - [ ] updated appropriate READMEs
 - [ ] Updated manual and documentation

PR #10233, opened 2026-04-15
Fix GetObjectId bounds in PKCS12 ContentInfo parsing

TristanInSec:
## Summary

- Bound `GetObjectId()` by the ContentInfo SEQUENCE end (`curIdx + curSz`) instead of the full buffer size (`(word32)size`)
- This prevents the OID TLV from being parsed past the SEQUENCE boundary in the first place, complementing the post-check added in PR #10018

## Context

PR #10018 added a post-check that catches when `GetObjectId()` advances `localIdx` past the ContentInfo SEQUENCE boundary. However, `GetObjectId()` itself still receives the full buffer size as `maxIdx`, so it parses the OID TLV beyond the SEQUENCE boundary before the post-check rejects it.

This change bounds `GetObjectId()` at the source, so it rejects an oversized OID immediately during parsing rather than after the fact. The post-check from #10018 remains as defense-in-depth.

## Test plan

- Existing `test_wc_d2i_PKCS12_oid_underflow` test (added in #10018) covers this path
- Crafted PKCS12 with OID extending past SEQUENCE boundary is rejected with `ASN_PARSE_E` (now by `GetObjectId` directly, rather than the post-check)

### DIFF
--- a/wolfcrypt/src/pkcs12.c
+++ b/wolfcrypt/src/pkcs12.c
@@ -328,7 +328,7 @@ static int GetSafeContent(WC_PKCS12* pkcs12, const byte* input,
 
             curIdx = localIdx;
             if ((ret = GetObjectId(input, &localIdx, &oid, oidIgnoreType,
-                                                           (word32)size)) < 0) {
+                                       curIdx + (word32)curSz)) < 0) {
                 WOLFSSL_LEAVE("Get object id failed", ret);
                 freeSafe(safe, pkcs12->heap);
                 return ret;


dgarske:
Hi @TristanInSec , I don’t see you setup as a contributor. Can you tell us more about yourself and your project? If you’d like to start the process for getting on file with a signed contributor agreement please email support at wolfssl dot com. 
Thanks, David Garske, wolfSSL

anhu:
@dgarske , have a look at ZD 21641 .

TristanInSec:
Hi @dgarske, I'm an independent security researcher. I identified this bounds issue while reviewing PKCS12 parsing code and wanted to contribute the fix. I'd be happy to sign the contributor agreement. Could you share the process or a link to get started?

dgarske:
Closing in favor of https://github.com/wolfSSL/wolfssl/pull/10378
Thank you @TristanInSec 
If you have any larger contributions we'll start that process, which is getting you a form that you need to sign and return via our support email

PR #10236, opened 2026-04-16
Enhance extra user data value and external cookie length max size

Roy-Carter:
# Description
Enhance configuration limits and fix max size constants to align with RFCs and large-scale deployment needs.

1. SSL_get_ex_new_index limit raised - --enable-context-extra-user-data now accepts values up to 9999 (was 99). Large platforms with high-scale operations need more than 99 ex_data indices. I've encountered it since my code uses : 
    SSL_EX_DATA_IND_DTLS_SESSION = SSL_get_ex_new_index(0, NULL, NULL, NULL, NULL);
    SSL_EX_DATA_IND_PSK          = SSL_get_ex_new_index(0, NULL, NULL, NULL, NULL); 
and on "Strong" machines in which i had 50+ cores running it which means (2x50) I failed to initialize an index for a DTLS session.

2. DTLS MAX_COOKIE_LEN raised to 254 - RFC 6347 defines cookie as opaque<0..2^8-1>, so max valid length is 255. Set to 254 to prevent buffer overflow attempts at boundary. Previous value of 32 was too restrictive for legitimate external cookie use. I've encountered it while trying to inject an external cookie which had valid length of more than 32 . 

# Testing
Build configuration tested with --enable-context-extra-user-data values: 1, 99, 100, 999, 9999
Verified configure.ac pattern matching rejects invalid inputs (0, 10000, strings)
DTLS cookie handling reviewed for buffer safety with new MAX_COOKIE_LEN

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -10376,9 +10376,10 @@ case "$ENABLED_EX_DATA" in
 no) ;;
 yes) AM_CFLAGS="$AM_CFLAGS -DHAVE_EX_DATA"
      ;;
-[[1-9]]|[[1-9]][[0-9]]) AM_CFLAGS="$AM_CFLAGS -DHAVE_EX_DATA -DMAX_EX_DATA=$ENABLED_EX_DATA"
+[[1-9]]|[[1-9]][[0-9]]|[[1-9]][[0-9]][[0-9]]|[[1-9]][[0-9]][[0-9]][[0-9]])
+       AM_CFLAGS="$AM_CFLAGS -DHAVE_EX_DATA -DMAX_EX_DATA=$ENABLED_EX_DATA"
        ;;
-*) AC_MSG_ERROR([Invalid argument to --enable-context-extra-user-data -- must be yes, no, or a number from 1 to 99])
+*) AC_MSG_ERROR([Invalid argument to --enable-context-extra-user-data -- must be yes, no, or a number from 1 to 9999])
    ;;
 esac
 
--- a/configure.ac
+++ b/configure.ac
@@ -10379,7 +10379,7 @@ yes) AM_CFLAGS="$AM_CFLAGS -DHAVE_EX_DATA"
 [[1-9]]|[[1-9]][[0-9]]|[[1-9]][[0-9]][[0-9]]|[[1-9]][[0-9]][[0-9]][[0-9]])
        AM_CFLAGS="$AM_CFLAGS -DHAVE_EX_DATA -DMAX_EX_DATA=$ENABLED_EX_DATA"
        ;;
-*) AC_MSG_ERROR([Invalid argument to --enable-context-extra-user-data -- must be yes, no, or a number from 1 to 9999 (note: each index reserves one pointer per object, so large values increase memory use)])  
+*) AC_MSG_ERROR([Invalid argument to --enable-context-extra-user-data -- must be yes, no, or a number from 1 to 9999 (note: each index reserves one pointer per object, so large values increase memory use)])
    ;;
 esac
 
--- a/configure.ac
+++ b/configure.ac
@@ -10379,7 +10379,7 @@ yes) AM_CFLAGS="$AM_CFLAGS -DHAVE_EX_DATA"
 [[1-9]]|[[1-9]][[0-9]]|[[1-9]][[0-9]][[0-9]]|[[1-9]][[0-9]][[0-9]][[0-9]])
        AM_CFLAGS="$AM_CFLAGS -DHAVE_EX_DATA -DMAX_EX_DATA=$ENABLED_EX_DATA"
        ;;
-*) AC_MSG_ERROR([Invalid argument to --enable-context-extra-user-data -- must be yes, no, or a number from 1 to 9999])
+*) AC_MSG_ERROR([Invalid argument to --enable-context-extra-user-data -- must be yes, no, or a number from 1 to 9999 (note: each index reserves one pointer per object, so large values increase memory use)])  
    ;;
 esac
 
--- a/wolfssl/internal.h
+++ b/wolfssl/internal.h
@@ -1444,8 +1444,12 @@ enum {
 
 #ifndef WOLFSSL_COOKIE_LEN
 /* Maximum size for a DTLS cookie */
-#define WOLFSSL_COOKIE_LEN 32 
+#define WOLFSSL_COOKIE_LEN 32
 #endif 
+
+#if WOLFSSL_COOKIE_LEN > 255
+#error "WOLFSSL_COOKIE_LEN must be <= 255 per RFC 6347 (opaque<0..2^8-1>)"
+#endif
 
 #if defined(WOLFSSL_TLS13) || !defined(NO_PSK)
 
--- a/wolfssl/internal.h
+++ b/wolfssl/internal.h
@@ -1445,7 +1445,7 @@ enum {
 #ifndef WOLFSSL_COOKIE_LEN
 /* Maximum size for a DTLS cookie */
 #define WOLFSSL_COOKIE_LEN 32
-#endif 
+#endif
 
 #if WOLFSSL_COOKIE_LEN > 255
 #error "WOLFSSL_COOKIE_LEN must be <= 255 per RFC 6347 (opaque<0..2^8-1>)"
--- a/wolfssl/internal.h
+++ b/wolfssl/internal.h
@@ -1569,7 +1569,7 @@ enum Misc {
     SEED_LEN     = RAN_LEN * 2, /* tls prf seed length    */
     ID_LEN       = 32,         /* session id length       */
     COOKIE_SECRET_SZ = 14,     /* dtls cookie secret size */
-    MAX_COOKIE_LEN = 32,       /* max dtls cookie size    */
+    MAX_COOKIE_LEN = 255,      /* max dtls cookie size per RFC 6347 (opaque<0..2^8-1>) */
     COOKIE_SZ    = 20,         /* use a 20 byte cookie    */
     SUITE_LEN    =  2,         /* cipher suite sz length  */
     ENUM_LEN     =  1,         /* always a byte           */
--- a/wolfssl/internal.h
+++ b/wolfssl/internal.h
@@ -1446,6 +1446,11 @@ enum {
 
 #define TLS13_TICKET_NONCE_MAX_SZ 255
 
+#ifndef WOLFSSL_COOKIE_LEN
+/* Maximum size for a DTLS cookie */
+#define WOLFSSL_COOKIE_LEN 32 
+#endif 
+
 #if (defined(HAVE_FIPS) &&                                                     \
     !(defined(FIPS_VERSION_GE) && FIPS_VERSION_GE(5,3))) &&                    \
     defined(TLS13_TICKET_NONCE_STATIC_SZ)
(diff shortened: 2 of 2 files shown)


Roy-Carter:
@julek-wolfssl  if you can please take a look , as part of the integration i've come across these 2 minor issues :)

julek-wolfssl:
@Roy-Carter is an approved contributor.

julek-wolfssl:
ok to test

wolfssl/internal.h
```diff
@@ -1569,7 +1569,7 @@ enum Misc {
     SEED_LEN     = RAN_LEN * 2, /* tls prf seed length    */
     ID_LEN       = 32,         /* session id length       */
     COOKIE_SECRET_SZ = 14,     /* dtls cookie secret size */
-    MAX_COOKIE_LEN = 32,       /* max dtls cookie size    */
+    MAX_COOKIE_LEN = 254,      /* max dtls cookie size per RFC 6347 (opaque<0..2^8-1>) more than 254 can be malformed / malicious  */
```
julek-wolfssl: Since this is an increase in the buffer used. Please add a `WOLFSSL_COOKIE_LEN` macro that is defined by default to `32` but can be increased when needed.

configure.ac
```diff
@@ -10376,9 +10376,10 @@ case "$ENABLED_EX_DATA" in
 no) ;;
 yes) AM_CFLAGS="$AM_CFLAGS -DHAVE_EX_DATA"
      ;;
-[[1-9]]|[[1-9]][[0-9]]) AM_CFLAGS="$AM_CFLAGS -DHAVE_EX_DATA -DMAX_EX_DATA=$ENABLED_EX_DATA"
+[[1-9]]|[[1-9]][[0-9]]|[[1-9]][[0-9]][[0-9]]|[[1-9]][[0-9]][[0-9]][[0-9]])
+       AM_CFLAGS="$AM_CFLAGS -DHAVE_EX_DATA -DMAX_EX_DATA=$ENABLED_EX_DATA"
        ;;
-*) AC_MSG_ERROR([Invalid argument to --enable-context-extra-user-data -- must be yes, no, or a number from 1 to 99])
+*) AC_MSG_ERROR([Invalid argument to --enable-context-extra-user-data -- must be yes, no, or a number from 1 to 9999])
```
julek-wolfssl: Do you really need 9k pointers in the external data?

julek-wolfssl (changes requested):

configure.ac
Roy-Carter: If you want I can minimze it a little but as of now we use around 500-1000 on our strongest machines , I wanted to cap it at bigger side to avoid updating it again anytime soon , and this is for the user's choice so if someone will need to he'll have the option to increase

Roy-Carter:
@julek-wolfssl can you re run the workflows ?

julek-wolfssl:
@Roy-Carter please address test failures.

julek-wolfssl (dismissed):

Roy-Carter:
@julek-wolfssl can we re-run workflows ? 
btw is there an easier option to go about testing things locally for different setups ? or is the github workflow that you run is the only option ?

wolfssl/internal.h
```diff
@@ -1442,6 +1442,11 @@ enum {
  */
 #define AEAD_SM4_CCM_LIMIT                       w64From32(0, (1 << 10) - 1)
 
+#ifndef WOLFSSL_COOKIE_LEN
+/* Maximum size for a DTLS cookie */
+#define WOLFSSL_COOKIE_LEN 32 
+#endif 
+
```
dgarske, line 1449: **🟠 [High] WOLFSSL_COOKIE_LEN defined inside wrong preprocessor guard — breaks build for DTLS without TLS13/PSK**
*🚫 BLOCK*  `bug`

The new `WOLFSSL_COOKIE_LEN` macro is placed inside the `#if defined(WOLFSSL_TLS13) || !defined(NO_PSK)` block that starts at line 1445 and ends at line 1468. However, `MAX_COOKIE_LEN = WOLFSSL_COOKIE_LEN` is referenced unconditionally inside `enum Misc` at line 1577, and the `cookie[MAX_COOKIE_LEN]` buffer at line 5290 is guarded by `#ifdef WOLFSSL_DTLS`, which is independent of `WOLFSSL_TLS13` and `NO_PSK`. A DTLS 1.2 build with `NO_PSK` defined and without `WOLFSSL_TLS13` would leave `WOLFSSL_COOKIE_LEN` undefined while `MAX_COOKIE_LEN` still tries to expand it, causing a compile error (undeclared identifier `WOLFSSL_COOKIE_LEN`). The macro must live outside the TLS13/PSK guard — either at the top of the file with other user-overridable tunables, or gated on `WOLFSSL_DTLS`.

**Suggestion:**
```suggestion
/* Move the define outside the TLS13/PSK guard, e.g. just above enum Misc */
#ifndef WOLFSSL_COOKIE_LEN
/* Maximum size for a DTLS cookie */
#define WOLFSSL_COOKIE_LEN 32
#endif

enum Misc {
    ...
    MAX_COOKIE_LEN = WOLFSSL_COOKIE_LEN, /* max dtls cookie size */
```

wolfssl/internal.h
```diff
@@ -1442,6 +1442,11 @@ enum {
...
 #if defined(WOLFSSL_TLS13) || !defined(NO_PSK)
```
dgarske, line 1455: **🟡 [Medium] PR description vs. code mismatch — default MAX_COOKIE_LEN not actually raised**
*💡 SUGGEST*  `question`

The PR description states `DTLS MAX_COOKIE_LEN raised to 254 - RFC 6347 defines cookie as opaque<0..2^8-1>, so max valid length is 255. Set to 254 to prevent buffer overflow attempts at boundary`. However the default for `WOLFSSL_COOKIE_LEN` is 32 and `MAX_COOKIE_LEN` remains 32 unless the user explicitly overrides it. This silently leaves stock builds with the same restriction the PR claims to remove. Either (a) raise the default (as the description promises) or (b) update the PR description and clearly document the new override knob in the relevant header/docs so users know how to opt in.

**Suggestion:**
```suggestion
/* If the intent is truly to raise the cap, bump the default */
#ifndef WOLFSSL_COOKIE_LEN
/* Maximum size for a DTLS cookie (RFC 6347 opaque<0..2^8-1>) */
#define WOLFSSL_COOKIE_LEN 255
#endif
```

wolfssl/internal.h
```diff
@@ -1569,7 +1574,7 @@ enum Misc {
     SEED_LEN     = RAN_LEN * 2, /* tls prf seed length    */
     ID_LEN       = 32,         /* session id length       */
     COOKIE_SECRET_SZ = 14,     /* dtls cookie secret size */
-    MAX_COOKIE_LEN = 32,       /* max dtls cookie size    */
+    MAX_COOKIE_LEN = WOLFSSL_COOKIE_LEN, /* max dtls cookie size */
     COOKIE_SZ    = 20,         /* use a 20 byte cookie    */
     SUITE_LEN    =  2,         /* cipher suite sz length  */
     ENUM_LEN     =  1,         /* always a byte           */
```
dgarske, line 1584: **🟡 [Medium] cookieSz is a `byte` but WOLFSSL_COOKIE_LEN is unbounded — silent truncation risk when override exceeds 255**
*💡 SUGGEST*  `bug`

`cookieSz` is declared as `byte` (uint8_t). Any user who defines `-DWOLFSSL_COOKIE_LEN=300` (or similar >255) will get a cookie buffer of 300 bytes but a size field that can only represent 0..255. In `src/internal.c:38156` the check `peerCookieSz > MAX_COOKIE_LEN` is compared against a `word32`-ish value, but downstream assignments `cookieSz = ...` will silently truncate. Since WOLFSSL_COOKIE_LEN is now a user-overridable macro, a compile-time guard should reject values > 255 (the protocol maximum per RFC 6347 anyway).

**Suggestion:**
```suggestion
#ifndef WOLFSSL_COOKIE_LEN
#define WOLFSSL_COOKIE_LEN 32
#endif
#if WOLFSSL_COOKIE_LEN > 255
#error "WOLFSSL_COOKIE_LEN must be <= 255 per RFC 6347 (opaque<0..2^8-1>)"
#endif
```

wolfssl/internal.h
dgarske, line 1455: **🔵 [Low] Trailing whitespace in new #define and #endif**
*🔧 NIT*  `style`

Both new lines have trailing whitespace: `#define WOLFSSL_COOKIE_LEN 32 ` and `#endif `. wolfSSL's style discourages trailing spaces; some CI checks may flag this.

**Suggestion:**
```suggestion
#define WOLFSSL_COOKIE_LEN 32
#endif
```

configure.ac
dgarske: **🔵 [Low] Error message wording — 'a number from 1 to 9999' is accurate but consider clarifying the memory tradeoff**
*🔧 NIT*  `style`

The error message only lists the accepted range. It would be user-friendly to note (either here or in the docs) that very large values have a memory cost because the ex_data array is fixed size. Not a blocker.

**Suggestion:**
```suggestion
AC_MSG_ERROR([Invalid argument to --enable-context-extra-user-data -- must be yes, no, or a number from 1 to 9999 (note: each index reserves one pointer per object, so large values increase memory use)])
```

dgarske (changes requested):
## 🐺 Skoll Code Review

### Posted findings
- [High] **WOLFSSL_COOKIE_LEN defined inside wrong preprocessor guard — breaks build for DTLS without TLS13/PSK** — `wolfssl/internal.h:1449-1452`
- [Medium] **PR description vs. code mismatch — default MAX_COOKIE_LEN not actually raised** — `wolfssl/internal.h:1451`
- [Medium] **cookieSz is a `byte` but WOLFSSL_COOKIE_LEN is unbounded — silent truncation risk when override exceeds 255** — `wolfssl/internal.h:5290-5291`
- [Medium] **No new tests for the expanded configure range or overridable cookie length** — `configure.ac:10376-10382, wolfssl/internal.h:1449-1452`
- [Low] **Trailing whitespace in new #define and #endif** — `wolfssl/internal.h:1451-1452`
- [Low] **Error message wording — 'a number from 1 to 9999' is accurate but consider clarifying the memory tradeoff** — `configure.ac:10382`

- [Medium] MAX_EX_DATA=9999 yields very large fixed arrays (~80 KB per object) — document memory cost

</details>

wolfssl/internal.h
Roy-Carter: fixed

Roy-Carter:
@dgarske can you re run the workflows ? fixed your notes

dgarske (changes requested):
```
trailing whitespace:
./wolfssl/internal.h:1448:#endif·
./configure.ac:10428:*) AC_MSG_ERROR([Invalid argument to --enable-context-extra-user-data -- must be yes, no, or a number from 1 to 9999 (note: each index reserves one pointer per object, so large values increase memory use)])··
```

Roy-Carter:
@dgarske fixed the trailing whitespace , can we run the workflow ?

Roy-Carter:
Hey @dgarske i see tests are failing on http requests and curl setups , do you have a way to look into it and re run tests after ?

dgarske:
> Hey @dgarske i see tests are failing on http requests and curl setups , do you have a way to look into it and re run tests after ?

The errors appear to be network related. I retried the failing tests. Your PR code is probably fine

Roy-Carter:
@julek-wolfssl  is this PR fine ? @dgarske  had some notes but I fixed them already

dgarske (approved):

PR #10240, opened 2026-04-16
wolfssl/wolfcrypt/types.h: avoid error w/ clang defining GCC >= 5

berolinux:
# Description

clang defines __GNUC__ for compatibility with gcc. The version of gcc it identifies itself as can be configured at clang build time, or by passing -fgnuc-version=x.y

While it currently defaults to setting __GNUC__ to 4, there are use cases for using "clang -fgnuc-version=15.2" or similar (typically working around constructs like
```
#if __GNUC__ < 10
// slow, unoptimized code path that works with ancient gcc
#else
// optimized code path that makes use of intrinsics available
// in modern gcc (and clang)
#endif
```

Without this patch, wolfssl's types.h errors out on being built with clang claiming to be a newer version of gcc - resulting in a duplicate definition of PRAGMA.

Don't add the gcc pragma definitions if clang is used.

# Testing

```
CFLAGS="-O2 -fgnuc-version=15.1" CXXFLAGS="-O2 -fgnuc-version=15.1" cmake .....
make
```

### DIFF
--- a/wolfssl/wolfcrypt/types.h
+++ b/wolfssl/wolfcrypt/types.h
@@ -2088,7 +2088,7 @@ WOLFSSL_API word32 CheckRunTimeSettings(void);
     #define WC_MP_TO_RADIX
 #endif
 
-#if defined(__GNUC__) && __GNUC__ > 5
+#if defined(__GNUC__) && __GNUC__ > 5 && !defined(__clang__)
     #define PRAGMA_GCC_DIAG_PUSH _Pragma("GCC diagnostic push")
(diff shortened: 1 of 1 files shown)


anhu:
Hi @berolinux , 

Thank you for noticing the need for this extra guarding condition. Your change, while welcome,  is small and not worth the effort on your part to sign a contributor agreement.  As such we will be  treating it as a bug report and I will be adding the fix  to another PR. 

This message will now close this pull request. 

Warm regards, Anthony

PR #10248, opened 2026-04-17
Rust wrapper: add digest and signature crate trait implementations

holtrop-wolfssl:
# Description

Rust wrapper: add digest and signature crate trait implementations

# Testing

Unit/CI tests

# Checklist

 - [X] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/wrapper/rust/wolfssl-wolfcrypt/build.rs
+++ b/wrapper/rust/wolfssl-wolfcrypt/build.rs
@@ -433,6 +433,22 @@ fn scan_cfg() -> Result<()> {
     check_cfg(&binding, "wc_RNG_DRBG_Reseed", "random_hashdrbg");
     check_cfg(&binding, "wc_InitRng", "random");
 
+    // When WOLFSSL_NO_MALLOC is set without WOLFSSL_STATIC_MEMORY, the
+    // WC_RNG struct contains an inline `drbg_data` field and wolfCrypt sets
+    // `rng->drbg = &rng->drbg_data` — a self-referential pointer.  Rust
+    // moves values by memcpy, which would silently invalidate that pointer.
+    // Detect this configuration and refuse to build.
+    if binding.contains("drbg_data") {
+        eprintln!(
+            "error: wolfSSL appears to be built with WOLFSSL_NO_MALLOC \
+             (without WOLFSSL_STATIC_MEMORY). This embeds a self-referential \
+             pointer inside WC_RNG (drbg -> drbg_data) that is incompatible \
+             with Rust move semantics. Please rebuild wolfSSL without \
+             WOLFSSL_NO_MALLOC, or enable WOLFSSL_STATIC_MEMORY."
+        );
+        std::process::exit(1);
+    }
+
     /* rsa */
     check_cfg(&binding, "wc_InitRsaKey", "rsa");
     check_cfg(&binding, "wc_RsaDirect", "rsa_direct");
--- a/wrapper/rust/wolfssl-wolfcrypt/src/lib.rs
+++ b/wrapper/rust/wolfssl-wolfcrypt/src/lib.rs
@@ -65,7 +65,7 @@ pub mod rsa;
 pub mod rsa_pkcs1v15;
 pub mod sha;
 #[cfg(feature = "digest")]
-mod sha_digest;
+pub mod sha_digest;
 
 /// Convert a buffer length to `u32`, returning `BUFFER_E` if it overflows.
 pub(crate) fn buffer_len_to_u32(len: usize) -> Result<u32, i32> {
--- a/wrapper/rust/wolfssl-wolfcrypt/src/lib.rs
+++ b/wrapper/rust/wolfssl-wolfcrypt/src/lib.rs
@@ -48,6 +48,8 @@ pub mod curve25519;
 pub mod dh;
 pub mod dilithium;
 pub mod ecc;
+#[cfg(feature = "signature")]
+pub mod ecdsa;
 pub mod ed25519;
 pub mod ed448;
 pub mod fips;
@@ -59,6 +61,8 @@ pub mod mlkem;
 pub mod prf;
 pub mod random;
 pub mod rsa;
+#[cfg(feature = "signature")]
+pub mod rsa_pkcs1v15;
 pub mod sha;
 #[cfg(feature = "digest")]
 mod sha_digest;
--- a/wrapper/rust/wolfssl-wolfcrypt/src/lib.rs
+++ b/wrapper/rust/wolfssl-wolfcrypt/src/lib.rs
@@ -60,6 +60,8 @@ pub mod prf;
 pub mod random;
 pub mod rsa;
 pub mod sha;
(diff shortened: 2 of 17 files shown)


wrapper/rust/wolfssl-wolfcrypt/build.rs
```diff
@@ -433,6 +433,22 @@ fn scan_cfg() -> Result<()> {
...
+    if binding.contains("drbg_data") {
+        eprintln!(
+            "error: wolfSSL appears to be built with WOLFSSL_NO_MALLOC \
+             (without WOLFSSL_STATIC_MEMORY). This embeds a self-referential \
+             pointer inside WC_RNG (drbg -> drbg_data) that is incompatible \
+             with Rust move semantics. Please rebuild wolfSSL without \
+             WOLFSSL_NO_MALLOC, or enable WOLFSSL_STATIC_MEMORY."
+        );
+        std::process::exit(1);
+    }
```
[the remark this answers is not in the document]
holtrop-wolfssl: This check is ok. There are only two uses of "drbg_data" in the code base for this purpose.

wrapper/rust/wolfssl-wolfcrypt/src/rsa_pkcs1v15.rs
```diff
@@ -0,0 +1,346 @@
...
+            panic!("wc_RsaFlattenPublicKey failed: {rc}");
+        }
+        VerifyingKey {
+            n,
+            e,
+            e_len: e_len as u8,
+            _hash: PhantomData,
+        }
+    }
+}
```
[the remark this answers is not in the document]
holtrop-wolfssl: Fixed

wrapper/rust/wolfssl-wolfcrypt/src/lib.rs
```diff
@@ -59,7 +61,11 @@ pub mod mlkem;
 pub mod prf;
 pub mod random;
 pub mod rsa;
+#[cfg(feature = "signature")]
+pub mod rsa_pkcs1v15;
 pub mod sha;
+#[cfg(feature = "digest")]
+mod sha_digest;
```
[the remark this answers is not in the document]
holtrop-wolfssl: This was not necessary for functionality, but marked pub for consistency.

wrapper/rust/wolfssl-wolfcrypt/build.rs
```diff
@@ -433,6 +433,22 @@ fn scan_cfg() -> Result<()> {
     check_cfg(&binding, "wc_RNG_DRBG_Reseed", "random_hashdrbg");
     check_cfg(&binding, "wc_InitRng", "random");
 
+    // When WOLFSSL_NO_MALLOC is set without WOLFSSL_STATIC_MEMORY, the
+    // WC_RNG struct contains an inline `drbg_data` field and wolfCrypt sets
+    // `rng->drbg = &rng->drbg_data` — a self-referential pointer.  Rust
+    // moves values by memcpy, which would silently invalidate that pointer.
+    // Detect this configuration and refuse to build.
+    if binding.contains("drbg_data") {
```
gasbytes, line 441: I think this hard-fails the build for the entire crate even when the user has not enabled random/signature/rand_core and doesn't touch WC_RNG at all.

maybe I'd consider gating this on cfg!(feature = "random")

also, not sure about this but, binding.contains("drbg_data") is a raw substring match on the generated bindings any unrelated future symbol containing drbg_data would trip this check.

probably matching on the struct field in WC_RNG specifically (e.g. a regex anchored to the struct) would be more robust.

wrapper/rust/wolfssl-wolfcrypt/src/rsa_pkcs1v15.rs
```diff
@@ -0,0 +1,349 @@
...
+                &self.inner.wc_rsakey,
+                e.as_mut_ptr(), &mut e_len,
+                n.as_mut_ptr(), &mut n_len,
+            )
+        };
+        if rc != 0 {
+            panic!("wc_RsaFlattenPublicKey failed: {rc}");
+        }
+        if (n_len as usize) != N || e_len == 0 || (e_len as usize) > MAX_E_LEN {
+            panic!("wc_RsaFlattenPublicKey failed: e_len: {e_len}, n_len: {n_len}");
```
gasbytes: "wc_RsaFlattenPublicKey returned unexpected lengths: e_len: {e_len}, n_len: {{n_len}" would be a more appropriate message I think, since at this point wc_RsaFlattenPublicKey returned zero.

wrapper/rust/wolfssl-wolfcrypt/src/rsa_pkcs1v15.rs
```diff
@@ -0,0 +1,349 @@
...
+            )
+        };
+        if rc != 0 || sig_len as usize != N {
+            return Err(Error::new());
+        }
+        Ok(Signature(sig))
+    }
+}
+
+const MAX_E_LEN: usize = 8;
```
gasbytes, line 201: maybe a documentaation comment on this would be worth adding, since exponents that are longer than 8 bytes are rejected, saying that bad_func_argument I think would be the returned value in case of failure.

gasbytes (changes requested):

wrapper/rust/wolfssl-wolfcrypt/build.rs
holtrop-wolfssl: You're right, it would break the build even without the traits enabled. I considered this ok though since any use of RNG that moves the struct memory (which Rust compiler can do at will!) will cause a problem in this regard, regardless of whether any of these features are enabled. If we only stored a pointer to the RNG instead of the C struct itself then that probably would not be an issue. Perhaps an improvement for the future.

As far as the name "drbg_data", there are no other occurrences of that in the repository. I'm hesitant to try to come up with a regular expression that anchors the "drbg_data" match to within the RNG struct since Rust grammar is definitely not a "regular" language to match against with a regular expression.

wrapper/rust/wolfssl-wolfcrypt/src/rsa_pkcs1v15.rs
holtrop-wolfssl: Changed as suggested

wrapper/rust/wolfssl-wolfcrypt/src/rsa_pkcs1v15.rs
holtrop-wolfssl: This isn't a pub const so I'm not sure a documentation comment would show up, but I added one!

wrapper/rust/wolfssl-wolfcrypt/build.rs
gasbytes: that makes sense, thanks, marking as solved.

gasbytes:
Jenkins retest this please (lost logs).

gasbytes (approved):

JacobBarthelmeh (approved):

PR #10249, opened 2026-04-17
Add negative-count and NULL checks to group-setting and shared-cipher APIs

ColtonWilley:
- `wolfSSL_CTX_set_groups` / `wolfSSL_set_groups`: add `count < 0` to
  entry guard — negative count passes the upper-bound check and truncates
  to a garbage `numGroups` byte, causing OOB read during SSL init
- `wolfSSL_CTX_set1_groups` / `wolfSSL_set1_groups`: widen `count == 0`
  to `count <= 0` and add NULL `groups` check
- `wolfSSL_get_shared_ciphers`: add NULL checks for `buf` and
  `wolfSSL_get_cipher_name_iana` return value

### DIFF
--- a/src/ssl.c
+++ b/src/ssl.c
@@ -1274,10 +1274,12 @@ const char* wolfSSL_get_shared_ciphers(WOLFSSL* ssl, char* buf, int len)
 {
     const char* cipher;
 
-    if (ssl == NULL || len <= 0)
+    if (ssl == NULL || buf == NULL || len <= 0)
         return NULL;
 
     cipher = wolfSSL_get_cipher_name_iana(ssl);
+    if (cipher == NULL)
+        return NULL;
     len = (int)min((word32)len, (word32)(XSTRLEN(cipher) + 1));
     XMEMCPY(buf, cipher, (size_t)len);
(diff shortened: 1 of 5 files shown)


src/ssl.c
```diff
@@ -3360,8 +3362,8 @@ int  wolfSSL_set1_groups(WOLFSSL* ssl, int* groups, int count)
     int i;
     int _groups[WOLFSSL_MAX_GROUP_COUNT];
     WOLFSSL_ENTER("wolfSSL_CTX_set1_groups");
-    if (count <= 0) {
-        WOLFSSL_MSG("Group count is not positive");
+    if (groups == NULL || count <= 0) {
```
dgarske, line 3365: Not there is a pre-existing typo above. Would you mind fixing it? `WOLFSSL_ENTER("wolfSSL_CTX_set1_groups");` -> `WOLFSSL_ENTER("wolfSSL_set1_groups");`

dgarske (changes requested):

dgarske (approved):

dgarske (approved):

PR #10259, opened 2026-04-19
TLS ECH keylogging

sebastian-carpenter:
# Description

Adds keylog support for TLS 1.3 ECH by exposing the HPKE shared secret and ECHConfig as new `ECH_SECRET` / `ECH_CONFIG` labels via the existing `tls13SecretCb` and OpenSSL `tls13KeyLogCb` callbacks. Enables Wireshark decryption of ECH-protected ClientHellos.

https://www.ietf.org/archive/id/draft-ietf-tls-keylogfile-05.html#name-secret-labels-for-ech

## Changes
* New `Tls13Secret` enum values, wired into keylog label tables.
* `hpke.c`: capture HPKE shared secret when `echSecret` is allocated; new `wc_HpkeInit/FreeEchSecret` helpers.
* `tls.c`: `EchWriteKeyLog()` emits both labels; `TLSX_FinalizeEch`/`TLSX_ExtractEch` now take `WOLFSSL*` and log after HPKE setup.
* Gated on `HAVE_SECRET_CALLBACK` + `HAVE_ECH`; no overhead when unused.
* Fixed when inner random is installed

# Testing

* Extended the existing `test_wolfSSL_Tls13_Key_Logging_test` to run an ECH handshake and assert the keylog file contains `ECH_SECRET` and `ECH_CONFIG` (and also `EXPORTER_SECRET`, which was previously unchecked).
* Added testing to the same `test_wolfSSL_Tls13_Key_Logging_test` and `test_wolfSSL_Tls13_Key_Logging_reject` to verify the correct random is used.

# Checklist

 - [X] added tests
 - [ ] updated/added doxygen


src/tls.c
```diff
@@ -14332,16 +14379,44 @@ static int TLSX_ExtractEch(WOLFSSL_ECH* ech, WOLFSSL_EchConfig* echConfig,
...
+            echConfigRaw = (byte*)XMALLOC(echConfigRawSz, ssl->heap,
+                DYNAMIC_TYPE_TMP_BUFFER);
+            if (echConfigRaw == NULL)
+                ret = MEMORY_E;
+        }
+        if (ret == 0)
+            ret = GetEchConfig(echConfig, echConfigRaw, &echConfigRawSz);
+        if (ret == 0) {
+            ret = EchWriteKeyLog(ssl, ech->hpke->echSecret, ech->hpke->Nsecret,
+                    echConfigRaw, echConfigRawSz);
```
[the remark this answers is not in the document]
sebastian-carpenter: then I would need to rederive the shared secret if there is an HRR.

tests/api.c
```diff
@@ -14176,11 +14182,21 @@ static int test_wolfSSL_Tls13_Key_Logging_test(void)
     test_ssl_cbf server_cbf;
     test_ssl_cbf client_cbf;
     XFILE fp = XBADFILE;
+#ifdef HAVE_ECH
+    const int label_count = 7;
+#else
+    const int label_count = 5;
+#endif
```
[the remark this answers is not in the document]
sebastian-carpenter: fixed

sebastian-carpenter:
Jenkins retest this please.

dgarske (approved):

src/tls.c
```diff
@@ -13813,6 +13748,42 @@ static int TLSX_ECH_GetSize(WOLFSSL_ECH* ech, byte msgType)
     return (int)size;
 }
 
+#ifdef HAVE_SECRET_CALLBACK
+/* log ECH_SECRET and ECH_CONFIG
+ * returns 0 on success, TLS13_SECRET_CB_E otherwise */
+static int EchWriteKeyLog(WOLFSSL* ssl, const byte* secret, word32 secretSz,
```
dgarske, line 13958: ### MEDIUM-1: Server-side ECH keylog failure misreported as decrypt error
- **File:** `src/tls.c:14283-14289 (TLSX_ExtractEch) and 14525-14534 (TLSX_ECH_Parse)`
- **Function:** `TLSX_ExtractEch / TLSX_ECH_Parse`
- **Action:** SUGGEST
- **Tag:** bug
- **Confidence:** Medium

**Description:** When `EchWriteKeyLog` fails (user keylog callback returns non-zero), `TLSX_ExtractEch` returns `TLS13_SECRET_CB_E`. Up in `TLSX_ECH_Parse`, the non-zero `ret` from `TLSX_ExtractEch` is treated as a *decryption failure* — it triggers `SendAlert(ssl, alert_fatal, decrypt_error)` and `WOLFSSL_ERROR_VERBOSE(DECRYPT_ERROR)`. Because the AEAD open and inner-CH decode actually already succeeded by the time `EchWriteKeyLog` runs, a buggy keylog callback now causes the server to send a misleading `decrypt_error` alert instead of a more accurate internal-error/secret-cb error. This is a behavioral regression introduced by inserting the keylog after the decrypt step.

**Code:**
```
#ifdef HAVE_SECRET_CALLBACK
    if (ret == 0 && ech->hpke->echSecret != NULL) {
        ret = EchWriteKeyLog(ssl, ech->hpke->echSecret, ech->hpke->Nsecret,
                info + TLS_INFO_CONST_STRING_SZ + 1, rawConfigLen);
    }
    wc_HpkeFreeEchSecret(ech->hpke);
#endif /* HAVE_SECRET_CALLBACK */
...
    /* if we failed to extract/expand */
    if (ret != 0) {
        WOLFSSL_MSG("ECH rejected");
        if (ssl->options.echAccepted == 1) {
            SendAlert(ssl, alert_fatal, decrypt_error);
            WOLFSSL_ERROR_VERBOSE(DECRYPT_ERROR);
            ret = DECRYPT_ERROR;
        }
```

**Recommendation:** Decide on intended semantics: most key logging callbacks should not break the handshake. Consider returning success from EchWriteKeyLog (after WOLFSSL_ERROR_VERBOSE) so the partial-log failure doesn't fail the connection, matching the spirit of keylogging being a debug-only feature.

tests/api.c
```diff
@@ -14544,12 +14544,32 @@ static int test_wolfSSL_Tls12_Key_Logging_test(void)
 
 #if defined(WOLFSSL_TLS13) && defined(OPENSSL_EXTRA) && \
     defined(HAVE_SECRET_CALLBACK)
+#ifdef HAVE_ECH
```
dgarske, line 14555: ### MEDIUM-5: Server-side ECH keylog not exercised by tests
- **File:** `tests/api.c:14545-14672`
- **Function:** `test_wolfSSL_Tls13_Key_Logging_test`
- **Action:** SUGGEST
- **Tag:** test
- **Confidence:** High

**Description:** The new `EchWriteKeyLog` call inside `TLSX_ExtractEch` (server side, src/tls.c:14283) is reachable in production but never exercised by the test — the test sets the keylog callback only on the client side (`client_cbf.ctx_ready`), not the server. So the server-side path through `wc_HpkeSetupBaseReceiver` -> echSecret copy -> `EchWriteKeyLog` is untested. Adding an analogous server-side ctx_ready that also sets the keylog callback and verifying a second log appears would close this gap.

**Code:**
```
    client_cbf.ctx_ready = &test_wolfSSL_Tls13_Key_Logging_client_ctx_ready;
    client_cbf.ssl_ready = &test_wolfSSL_Tls13_Key_Logging_client_ssl_ready;
#ifdef HAVE_ECH
    server_cbf.ctx_ready = &test_ech_server_ctx_ready;
    server_cbf.ssl_ready = &test_ech_server_ssl_ready;
#endif
```

**Recommendation:** Augment the test to cover the server-side `EchWriteKeyLog` path via TLSX_ExtractEch; otherwise rely on hand-testing for that branch.

dgarske (dismissed):

src/tls.c
sebastian-carpenter: Fixed. Wrapped acceptance logic so callback failure causes an internal failure instead. Added test for keylog failure as well.

tests/api.c
sebastian-carpenter: added server-side testing to test

douzzer (approved):

PR #10262, opened 2026-04-20
20260420-test-fixes

douzzer:
`wolfcrypt/test/test.c`: add missing FIPS gating for backward-incompatible `NULL` arg tests in `hkdf_test()` and `srtpkdf_test()`.  (fixes #10238 / 3175c3387f9f08d7c3bc627f58913056e2145cbf).

`linuxkm/module_hooks.c`: in `wolfssl_init()` `FIPS_OPTEST_FULL_RUN_AT_MODULE_INIT` code path (currently unused), add missing declaration for `i` (fixes F-3026).

fix `examples/pem/` and `scripts/pem.test`:

`examples/pem/pem.c`:
* improve error messages,
* add `wc_SetSeed_Cb()` if `WC_RNG_SEED_CB`, and
* add `wolfCrypt_Init()` and `wolfCrypt_Cleanup()`.

`scripts/pem.test`:
* fix exit code to unmask script failure,
* add configured feature detection,
* improve error messages and handling,
* add configuration gating around subtests, and
* comment out currently failing subtests.

tested with
```
wolfssl-multi-test.sh ...
super-quick-check
linuxkm-fips-v6-insmod-cust-kernel-2
all-crypto-openssl-extra-coexist-fips-v6
fips-140-3-v6-optest-acvp-sp-asm
```

### DIFF
--- a/examples/pem/pem.c
+++ b/examples/pem/pem.c
@@ -39,6 +39,8 @@
 
 #if defined(WOLFSSL_PEM_TO_DER) && !defined(NO_FILESYSTEM)
 
+static const char *progname;
+
 /* Increment allocated data by this much. */
 #define DATA_INC_LEN        256
 /* Maximum block size of a cipher. */
@@ -554,15 +556,20 @@ static int EncryptDer(unsigned char* in, word32 in_len, char* password,
         if (ret == WC_NO_ERR_TRACE(LENGTH_ONLY_E)) {
             ret = 0;
         }
-        else if (ret == 0) {
-            ret = 1;
+        else {
+            fprintf(stderr,
+                    "%s: wc_CreateEncryptedPKCS8Key() with enc_alg_id %d: "
+                    "unexpected retval: %s.\n",
+                    progname, enc_alg_id, wc_GetErrorString(ret));
+            if (ret == 0)
+                ret = 1;
         }
     }
     if (ret == 0) {
         /* Allocate memory for encrypted DER data. */
         *enc = (unsigned char*)XMALLOC(*enc_len, NULL, DYNAMIC_TYPE_TMP_BUFFER);
         if (*enc == NULL) {
-            ret = 1;
+            ret = MEMORY_E;
         }
     }
     if (ret == 0) {
@@ -748,6 +755,12 @@ int main(int argc, char* argv[])
 #ifdef DEBUG_WOLFSSL
     int log = 0;
 #endif
+
+    progname = strrchr(argv[0], '/');
+    if (progname)
+        ++progname;
+    else
+        progname = argv[0];
 
     memset(&info, 0, sizeof(info));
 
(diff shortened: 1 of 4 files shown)


examples/pem/pem.c
```diff
@@ -554,15 +556,20 @@ static int EncryptDer(unsigned char* in, word32 in_len, char* password,
...
+            if (ret == 0)
+                ret = 1;
         }
     }
     if (ret == 0) {
         /* Allocate memory for encrypted DER data. */
         *enc = (unsigned char*)XMALLOC(*enc_len, NULL, DYNAMIC_TYPE_TMP_BUFFER);
         if (*enc == NULL) {
-            ret = 1;
+            ret = MEMORY_E;
```
SparkiDev, line 572: This is application code not library.
Using the errors from wolfSSL isn't needed.

scripts/pem.test
```diff
@@ -19,21 +19,126 @@ CR=$'\n'
...
+    exit 77
+fi
+
+SRC_DIR="$(dirname "$0")/.."
+if [ ! -d "${SRC_DIR}/certs" ]; then
+    echo "certs not found at ${SRC_DIR}/certs -- skipping pem.test."
+    exit 77
+fi
+
+if grep -q -E '^#define HAVE_FIPS$' wolfssl/options.h; then
```
SparkiDev, line 40: Does this work for user_options.h?

scripts/pem.test
```diff
@@ -135,10 +240,12 @@ test_fail() {
 # Use asn1 to check DER produced is valid.
 check_der() {
     $ASN1_EXE $tmp_der_file >$tmp_file 2>&1
-    if [ "$?" != "0" ]; then
+    local ret=$?
+    if [ "$ret" != "0" ]; then
         echo
         echo "  DER result bad"
         test_fail
+        return $ret
```
SparkiDev, line 248: We are collecting test failures for a final summary.
Do we need to return $ret here.

SparkiDev (changes requested):

scripts/pem.test
douzzer: that's harmless -- the existing callers are ignoring it.  it didn't sit right with me that `check_der` was returning success when its one and only action failed for unexplained reasons.

scripts/pem.test
douzzer: nope, but it would be pretty unusual for examples/ to get built in a `user_settings.h` config.

but feel free to follow up this PR with one that fixes the broken tests and considers `user_settings.h`.

examples/pem/pem.c
douzzer: sure but it's harmless, and with `--enable-debug-trace-errcodes CFLAGS=-DWOLFSSL_DEBUG_TRACE_ERROR_CODES_ALWAYS` it even allows the failed alloc to be instrumented right in the app!  yes bikeshedding...

SparkiDev (approved):

PR #10272, opened 2026-04-21
Fix support on CHERI RISC-V architecture

wbeasley-thegoodpenguin:
# Description

Whilst integrating wolfssl into the meta-cheri yocto layer I encountered issues during compilation and at runtime.

This PR is made up of 3 commits:

Commit 1:
At runtime this was causing a CHERI bounds exception. It seem `XorTable_Multi` was using hardcoded masks `0xf0`/`0x0f` which assume 64-bit cache lines. This meant that it was attempting to stride beyond the 256-entry table.

Commit 2: 
I was get a CHERI tag violation fault. On CHERI when you cast a pointer to a `size_t` this removes the hardware validity tag. When it is cast back to a pointer its tag is no longer valid, and leads to a violation fault. My fix here needs input, as it invalidates constant-time, but I left it as is in the interest of getting it working on CHERI.
Update: A fix has now been put in place to handle this by conditionally copying the digit data vs the pointers

Commit 3: 
I tripped a static_assert `WOLFSSL_ASSERT_SIZEOF_GE(ssl->async->args, *args);`. On CHERI pointers are capabilities and are larger (16-bytes in this case). I made the buffer larger here to accommodate that. Also CHERI pointers need to be 16-byte aligned so swapped it from a `word32` to a `max_align_t`. 

Commit 4:
I switched to building with `--enable-all` and found more cases where things need to be made bigger.

Commit 5:
Similar issue to Commit 2. Although opted for using the existing mp_cond_copy function for CHERI purecap

# Testing

Built with `--enable-all --enable-dtls --enable-dtls13 --enable-dtlscid --enable-dtls-frag-ch`
Ran wolfcrypttest - no issues
Ran wolfssl-unittest - Got `Failed/Skipped/Passed/All: 0/103/1102/1205`

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/wolfcrypt/src/sakke.c
+++ b/wolfcrypt/src/sakke.c
@@ -2649,12 +2649,19 @@ static int sakke_modexp_loop(SakkeKey* key, mp_int* b, mp_int* e, mp_proj* r,
             err = sakke_proj_mul_qx1(c[0], by, prime, mp, c[j^1], t1, t2);
 #else
             err = sakke_proj_mul_qx1(c[0], by, prime, mp, c[2], t1, t2);
+#ifdef WC_NO_PTR_INT_CAST
+            err = mp_cond_copy(c[2]->x, j,   c[0]->x);
+            err = mp_cond_copy(c[2]->x, j^1, c[1]->x);
+            err = mp_cond_copy(c[2]->y, j,   c[0]->y);
+            err = mp_cond_copy(c[2]->y, j^1, c[1]->y);
+#else
             mp_copy(c[2]->x,
             (mp_int*) ( ((wc_ptr_t)c[0]->x & wc_off_on_addr[j]) +
                         ((wc_ptr_t)c[1]->x & wc_off_on_addr[j^1]) ) );
             mp_copy(c[2]->y,
             (mp_int*) ( ((wc_ptr_t)c[0]->y & wc_off_on_addr[j]) +
                         ((wc_ptr_t)c[1]->y & wc_off_on_addr[j^1]) ) );
+#endif
 #endif
         }
     }
--- a/wolfcrypt/src/sp_int.c
+++ b/wolfcrypt/src/sp_int.c
@@ -5115,12 +5115,28 @@ static WC_INLINE sp_int_digit sp_div_word(sp_int_digit hi, sp_int_digit lo,
     (defined(HAVE_ECC) && defined(HAVE_COMP_KEY)) || defined(OPENSSL_EXTRA) || \
     (defined(WOLFSSL_SP_MATH_ALL) && !defined(WOLFSSL_RSA_PUBLIC_ONLY))
 #ifndef WC_NO_CACHE_RESISTANT
+#ifdef WC_NO_PTR_INT_CAST
+static void _sp_cond_copy(const sp_int* a, int copy, sp_int* r, sp_size_t used)
+{
+    sp_int_digit mask = (sp_int_digit)0 - (sp_int_digit)copy;
+    unsigned int i;
+
+    for (i = 0; i < (unsigned int)used; i++) {
+        r->dp[i] ^= (r->dp[i] ^ a->dp[i]) & mask;
+    }
+    r->used ^= (r->used ^ a->used) & (sp_size_t)mask;
+#ifdef WOLFSSL_SP_INT_NEGATIVE
+    r->sign ^= (r->sign ^ a->sign) & (sp_sign_t)mask;
+#endif
+}
+#else
     /* Mask of address for constant time operations. */
     const size_t sp_off_on_addr[2] =
     {
         (size_t) 0,
         (size_t)-1
     };
+#endif
 #endif
 #endif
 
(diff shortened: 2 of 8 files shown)


pzalewski-thegoodpenguin:
@wbeasley-thegoodpenguin see https://github.com/The-Capable-Hub/meta-cheri/blob/pz/refactoring-2/kas/riscv64-multilib.yml#L42 for python on CHERI (but in integer mode)

wolfcrypt/src/sp_int.c
```diff
@@ -5121,6 +5121,14 @@ static WC_INLINE sp_int_digit sp_div_word(sp_int_digit hi, sp_int_digit lo,
         (size_t) 0,
         (size_t)-1
     };
+/* Constant time access here wont work on CHERI so fallback to basic for now */
```
LinuxJedi, line 5124: Can you fix "wont". This fails our codespell test.

wolfssl/internal.h
```diff
@@ -5761,15 +5761,15 @@ typedef struct BuildMsgArgs {
...
+    #define MAX_ASYNC_ARGS 24
     typedef void (*FreeArgsCb)(struct WOLFSSL* ssl, void* pArgs);
 
     struct WOLFSSL_ASYNC {
 #if defined(WOLFSSL_ASYNC_CRYPT) && !defined(WOLFSSL_NO_TLS12)
         BuildMsgArgs  buildArgs; /* holder for current BuildMessage args */
 #endif
         FreeArgsCb    freeArgs; /* function pointer to cleanup args */
-        word32        args[MAX_ASYNC_ARGS]; /* holder for current args */
+        max_align_t args[MAX_ASYNC_ARGS * sizeof(word32) / sizeof(max_align_t)];
```
LinuxJedi, line 5772: This change breaks things on several platforms. We need to support C89 standard and this appears to be C99 code. If we need to use `max_align_t` then this should definitely be under a CHERI specific guard.

LinuxJedi (changes requested):
Hi @wbeasley-thegoodpenguin,

Many thanks for this contribution. I've added some initial review comments based on the CI output. In the mean time, can you please email support@wolfssl.com so that we can file a CLA for you?

@SparkiDev can you please look at commit 2 here? I suspect there has to be a way to keep constant time here for the CHERI platform, but this is beyond me.

LinuxJedi:
Many thanks for resolving these issues. I think we should keep `MAX_ASYNC_ARGS` at the new value for all platforms because Scv13Args with WOLFSSL_DUAL_ALG_CERTS would blow the the current default of 18 by 16 bytes.

I've pinged Sean internally just to check over the constant time things. But in general I'm happy.

wolfcrypt/src/sp_int.c
```diff
@@ -5121,6 +5121,14 @@ static WC_INLINE sp_int_digit sp_div_word(sp_int_digit hi, sp_int_digit lo,
         (size_t) 0,
         (size_t)-1
     };
+/* Constant time access here will not work on CHERI so fallback to basic for now */
+#ifdef __CHERI_PURE_CAPABILITY__
+    #define SP_CT_ADDR(t, idx) ((t)[(idx)])
```
SparkiDev: The purpose of the code you had to disable is to make sure both address are loaded into cache even though only one is accessed.

Is there a way to preload t[0] and t[1] with CHERI and then use only one of the addresses?

SparkiDev (dismissed):

wolfcrypt/src/sp_int.c
wbeasley-thegoodpenguin: I dont think there is a guaranteed way to port this pattern to CHERI. There are ways of loading the capabilities (there are load/store capability instructions) but they are not constant-time guaranteed because that would be memory dependent? 

I wonder if changing things around here to do the XOR'ing on the digit data like in mp_cond_copy would work here?

dgarske:
@wbeasley-thegoodpenguin has been approved as a contributor and is on file for any future work. Thank you

wbeasley-thegoodpenguin:
I've pushed new changes that enable building with `--enable-all --enable-dtls --enable-dtls13 --enable-dtlscid --enable-dtls-frag-ch` then ran the unittests

LinuxJedi (dismissed):
After discussions with @wbeasley-thegoodpenguin and @SparkiDev, I believe this is ready now.

wolfcrypt/src/sakke.c
```diff
@@ -2649,12 +2649,19 @@ static int sakke_modexp_loop(SakkeKey* key, mp_int* b, mp_int* e, mp_proj* r,
             err = sakke_proj_mul_qx1(c[0], by, prime, mp, c[j^1], t1, t2);
 #else
             err = sakke_proj_mul_qx1(c[0], by, prime, mp, c[2], t1, t2);
+#ifdef __CHERI_PURE_CAPABILITY__
```
dgarske: I would prefer a macro that is generic name addressing the core issue not with CHERI in the name...  like `WC_NO_PTR_INT_CAST` or `WC_NO_PTR_MASKING`. Same with others too. Thanks

dgarske (changes requested):

wolfcrypt/src/sakke.c
wbeasley-thegoodpenguin: Ok I have switched this to `WC_NO_PTR_INT_CAST` and added it to `settings.h` - Are there any other places that need updating (just thinking for documentation)

dgarske (approved):

LinuxJedi (approved):

PR #10278, opened 2026-04-22
Add wolfCrypt support for lpc55s69 hw crypto

twcook86:
# Description

Fixes zd#

# Testing

How did you test?

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


philljj:
Retest this please.

(failure logs not found).

philljj:
multi-test is flagging these macros:
```
unrecognized macros used:
WOLFSSL_NXP_CASPER_ECC_MUL2ADD
WOLFSSL_NXP_CASPER_ECC_MULMOD
WOLFSSL_NXP_LPC55S69_NO_HWACCEL
WOLFSSL_NXP_LPC55S69_WITH_HWACCEL
```

and these comments and whitespace:
```
trailing whitespace:
./wolfcrypt/src/port/nxp/hashcrypt_port.c:202:        aes_handle.keySize = kHASHCRYPT_Aes128;·
./wolfcrypt/src/port/nxp/hashcrypt_port.c:204:        aes_handle.keySize = kHASHCRYPT_Aes192;·
./wolfcrypt/src/port/nxp/hashcrypt_port.c:208:        return BAD_FUNC_ARG;·
./wolfcrypt/src/port/nxp/casper_port.c:89:#define CASPER_MAX_ECC_SIZE_BYTES (72)·
./wolfcrypt/src/port/nxp/casper_port.c:159:}····
./wolfcrypt/src/port/nxp/casper_port.c:250:}····
C++-style comments:
./wolfssl/wolfcrypt/settings.h:2138:    // #define WOLFSSL_NXP_CASPER_RSA_PUB_EXPTMOD
./wolfcrypt/src/port/nxp/hashcrypt_port.c:70:    // if (wolfSSL_CryptHwMutexLock() != 0)
./wolfcrypt/src/port/nxp/hashcrypt_port.c:71:    //     return BAD_MUTEX_E;
```

philljj:
note: this is to support wolfboot:
- https://github.com/wolfSSL/wolfBoot/pull/757

wolfcrypt/src/port/nxp/casper_port.c
```diff
@@ -0,0 +1,253 @@
+/* casper_port.c
```
philljj, line 1: This is pretty cool and imo warrants a separate readme for the lpc55s69 in `wolfcrypt/src/port/nxp/README.md`, like how we have `README_SE050.md`.

Probably some of the material in the wolfboot targets MD (https://github.com/wolfSSL/wolfBoot/blob/master/docs/Targets.md#lpc55s69-hardware-acceleration) should be documented here as well.

Something that lists what all we're accelerating for this this target (symm, hash, and asym mul/expt ops). Could also list the hardware limitations as well.

wolfcrypt/test/test.c
```diff
@@ -12559,6 +12574,7 @@ static wc_test_ret_t EVP_test(const WOLFSSL_EVP_CIPHER* type, const byte* key,
         if (XMEMCMP(plain, msg3, WC_AES_BLOCK_SIZE * 4))
             ERROR_OUT(WC_TEST_RET_ENC_NC, out);
     #endif /* HAVE_AES_DECRYPT */
+    #endif
```
philljj: nitpicky comment, but I recommend commenting all `#endif` lines like
```
#endif /* WOLFSSL_NXP_HASHCRYPT_AES */
```
Otherwise these sources become a define gate soup X-D

wolfcrypt/test/test.c
```diff
@@ -4964,12 +4964,14 @@ WOLFSSL_TEST_SUBROUTINE wc_test_ret_t sha_test(void)
     ret = wc_InitSha_ex(&sha, HEAP_HINT, devId);
     if (ret != 0)
         ERROR_OUT(WC_TEST_RET_ENC_EC(ret), exit);
+#ifndef NO_WOLFSSL_SHA256_INTERLEAVE
```
philljj, line 4947: Why is this define needed? Is this due to a quirk of the lpc55s69 sha accel?

philljj (dismissed):

wolfcrypt/test/test.c
twcook86: Yes, the hashcrypt engine can't swap context in the middle of a hash.  Have to finish one hash before starting the next.

dgarske: This means the HASH hardware won't work with TLS and you may need to cache the updates in some cases until final. Did you test with TLS? You might also consider automatically setting `NO_WOLFSSL_SHA256_INTERLEAVE` in the settings.h for this platform. Look at wc_Pic32HashUpdate .

dgarske (changes requested):

wolfcrypt/test/test.c
twcook86: @dgarske I turned on tls 1.3 kdf, but that passed the test.  Which test should I turn on to see a failure?

dgarske: You would need to run a full TLS v1.3 client or server. Perhaps that's not required for this port. Its just worth noting. 

With a TLS session you will have multiple active hashing sessions at the same time (update called for different wc_Sha256 contexts) and possibly even different hashing algorithms.

wolfcrypt/src/port/nxp/casper_port.c
twcook86: done

wolfcrypt/test/test.c
twcook86: done

wolfcrypt/test/test.c
twcook86: done. notes added in port README.md

ColtonWilley:
Jenkins retest this please

wolfssl/wolfcrypt/port/nxp/hashcrypt_port.h
```diff
@@ -0,0 +1,28 @@
+/* hashcrypt_port.h
```
dgarske, line 1: I don't see these new files added to the include.am? If they aren't then they won't be part of the "make dist". See `wolfssl/wolfcrypt/include.am`

wolfssl/wolfcrypt/settings.h
```diff
@@ -2132,6 +2132,19 @@ extern void uITRON4_free(void *p) ;
     #endif
 #endif /* FREESCALE_USE_LTC */
 
+#if defined(WOLFSSL_NXP_LPC55S69_WITH_HWACCEL)
```
dgarske: Is it possible to use a less specific macro? I suspect this Casper and HashCrypt will exist on other flavors too? Like is there a family of LPCS44SXX ?

wolfssl/wolfcrypt/settings.h
```diff
@@ -2132,6 +2132,19 @@ extern void uITRON4_free(void *p) ;
...
+    #define WOLFSSL_NXP_RNG_1
+    #define WOLFSSL_NXP_HASHCRYPT
+    #define WOLFSSL_NXP_HASHCRYPT_AES
+    #define WOLFSSL_NXP_HASHCRYPT_SHA
```
dgarske, line 2140: Not sure the level of SHA and SHA256 granularity is required. I'd prefer one WOLFSSL_NXP_HASHCRYPT macro and use the NO_SHA and NO_SHA256 to gate the other details.

wolfssl/wolfcrypt/settings.h
```diff
@@ -2132,6 +2132,19 @@ extern void uITRON4_free(void *p) ;
...
+    #define WOLFSSL_NXP_HASHCRYPT_SHA256
+    #define WOLFSSL_NXP_CASPER
+    #define WOLFSSL_NXP_CASPER_RSA_PUB_EXPTMOD
+    #define NO_WOLFSSL_SHA256_INTERLEAVE
```
dgarske, line 2143: Consider this being forced for `WOLFSSL_NXP_HASHCRYPT` outside this group... not specifically for the LPC55S69 and if not already defined.

wolfssl/wolfcrypt/settings.h
```diff
@@ -2132,6 +2132,19 @@ extern void uITRON4_free(void *p) ;
...
+#elif defined(WOLFSSL_NXP_LPC55S69_NO_HWACCEL)
+    #define WOLFSSL_NXP_RNG_1
```
dgarske: What's the _1? Are there multiple RNG's? Do we really need both a force disable `WOLFSSL_NXP_LPC55S69_NO_HWACCEL` and enable `WOLFSSL_NXP_LPC55S69_WITH_HWACCEL`? Typically its off by default anyways

wolfcrypt/src/port/nxp/casper_port.c
dgarske, line 1: Same these need to be in the include.am... otherwise won't be in "make dist". See `wolfcrypt/src/include.am`

wolfcrypt/src/port/nxp/casper_port.c
```diff
@@ -0,0 +1,253 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1335, USA
+ */
+
+#ifdef HAVE_CONFIG_H
```
dgarske: Replace top section... config.h and settings.h with `wolfssl/wolfcrypt/libwolfssl_sources.h`

wolfcrypt/src/port/nxp/hashcrypt_port.c
```diff
@@ -0,0 +1,398 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1335, USA
+ */
+
+
+#ifdef HAVE_CONFIG_H
```
dgarske: Replace top section... config.h and settings.h with `wolfssl/wolfcrypt/libwolfssl_sources.h`

dgarske (changes requested):

wolfssl/wolfcrypt/settings.h
twcook86: Well, this is what I was running up against...

Everything I've done applies to the LPC55S6x and LPC55S2x, but not necessarily the LPC55S0x and LPC55S1x.  For example, the 6x and 2x do not support hash "reload" (thus we need NO_WOLFSSL_SHA256_INTERLEAVE), but the 0x and 1x DO support hash reload.

I could change it to 6X, I suppose.

wolfssl/wolfcrypt/settings.h
twcook86: My thinking was that, with the no-interleave limitation in the hashcrypt engine, user could turn off its use for SHA, but keep it on for AES.  (The code is also already gated on NO_SHA and NO_SHA256.)  But, I could combine the SHA and SHA256 into one.  Or I could just use WOLFSSL_NXP_HASHCRYPT for all.  I'm fine either way.

wolfssl/wolfcrypt/settings.h
twcook86: See above comment about the 0x and 1x chip variants.

wolfssl/wolfcrypt/settings.h
twcook86: Yes, there is a regular "rng" in the nxp sdk, and a "rng_1" in the sdk.  This chip uses the rng_1.  There's also a "rnga" and a "trng" in the sdk.

Regarding explicit WOLFSSL_NXP_LPC55S69_WITH_HWACCEL and WOLFSSL_NXP_LPC55S69_NO_HWACCEL switches, the only reason they're both used is because I turn on the RNG_1 use for both in settings.h.  Daniele said we should have it on regardless.  I suppose I could pass in WOLFSSL_NXP_RNG_1 from wolfBOOT, but I wanted to keep all the switches in one place.

dgarske: Please see if you can reduce the number of options and also make them less specific. Stick to vendor or family only not specific chip. At some point someone wants to use a slightly different model in same family and the macro names are confusing.

wolfssl/wolfcrypt/settings.h
dgarske: I am okay with separate ones for AES and SHA of course. I was only referring to the SHA256/SHA granularity.

wolfssl/wolfcrypt/settings.h
dgarske: Find the best option and change it. If its 6X great. Thanks

dgarske (approved):

PR #10290, opened 2026-04-23
Fix emNET support and add tests

LinuxJedi:
The emNET `wolfSSL_LastError` branches were incorrect. The second one was never hit and would never compile. The first one inverts error codes that should not be inverted.

This fixes that code and adds a test with a shim layer to test emNET calls without using emNET.

ZD #21673


LinuxJedi:
Jenkins retest this please. No results

LinuxJedi:
Jenkins retest this please, I think wrong commit id tested

dgarske (approved):
Nice!

PR #10291, opened 2026-04-23
Add negative tests for AEAD, PKCS7, PSS, DSA, DRBG, and PQ key

JeremiahM37:
Addresses F-2195, F-2204, F-2198, F-2199, F-3072, F-3073, F-2640, F-2643, F-3074.

- AEAD bad auth tag: streaming AES-GCM decrypt final and XChaCha20-Poly1305 one-shot decrypt.
- PKCS7 SignedData with tampered signedAttribs (non-messageDigest bytes) for RSA and ECDSA signers — exercises `wc_PKCS7_RsaVerify` recovered-digest `XMEMCMP` and `wc_PKCS7_EcdsaVerify` `res==1` guard.
- RSA PSS: constructs a signature whose recovered EM has a wrong `0xbc` terminator; asserts `BAD_PADDING_E` specifically.
- DRBG: sets `reseedCtr = WC_RESEED_INTERVAL - 1` and verifies the boundary check fires (counter resets after next generate).
- ML-KEM: flips a byte inside `H(ek)` of an encoded private key blob; asserts `MLKEM_PUB_HASH_E`.
- SLH-DSA: tampers a seed in `key->sk` after import; asserts `wc_SlhDsaKey_CheckKey` returns `WC_KEY_MISMATCH_E`.
- DSA: tampered hash produces `answer == 0` via the final `mp_cmp(r, v)` check.

### DIFF
--- a/tests/api/test_aes.c
+++ b/tests/api/test_aes.c
@@ -4400,6 +4400,70 @@ int test_wc_AesGcmStream_ReinitAfterFinal(void)
     return EXPECT_RESULT();
 } /* END test_wc_AesGcmStream_ReinitAfterFinal */
 
+int test_wc_AesGcmStream_BadAuthTag(void)
+{
+    EXPECT_DECLS;
+#if !defined(NO_AES) && defined(HAVE_AESGCM) && defined(HAVE_AES_DECRYPT) && \
+    defined(WOLFSSL_AES_128) && defined(WOLFSSL_AESGCM_STREAM)
+    static const byte key[AES_128_KEY_SIZE] = {
+        0xfe,0xff,0xe9,0x92, 0x86,0x65,0x73,0x1c,
+        0x6d,0x6a,0x8f,0x94, 0x67,0x30,0x83,0x08
+    };
+    static const byte iv[GCM_NONCE_MID_SZ] = {
+        0xca,0xfe,0xba,0xbe, 0xfa,0xce,0xdb,0xad,
+        0xde,0xca,0xf8,0x88
+    };
+    static const byte aad[20] = {
+        0xfe,0xed,0xfa,0xce, 0xde,0xad,0xbe,0xef,
+        0xfe,0xed,0xfa,0xce, 0xde,0xad,0xbe,0xef,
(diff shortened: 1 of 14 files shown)


JeremiahM37:
Jenkins retest this please

tests/api/test_aes.c
```diff
@@ -4400,6 +4392,72 @@ int test_wc_AesGcmStream_ReinitAfterFinal(void)
...
+    ExpectIntEQ(wc_AesGcmDecryptUpdate(dec, pt, ct, sizeof(ct),
+        aad, sizeof(aad)), 0);
+    ExpectIntEQ(wc_AesGcmDecryptFinal(dec, tag, sizeof(tag)),
+        WC_NO_ERR_TRACE(AES_GCM_AUTH_E));
+    wc_AesFree(dec);
+
+    tag[0] ^= 0x01;
+    ExpectIntEQ(wc_AesInit(dec, NULL, INVALID_DEVID), 0);
+    ExpectIntEQ(wc_AesGcmDecryptInit(dec, key, sizeof(key), iv, sizeof(iv)), 0);
+    {
```
dgarske: Let's not have empty braces. Did all these new negative tests not uncover any core code fixes? Can you make sure they are actually running???

dgarske (changes requested):

tests/api/test_aes.c
JeremiahM37: Braces fixed. Fenrir flags these missing tests for things that could fail and not be caught in tests. Each test asserts a specific error code (AES_GCM_AUTH_E, MAC_CMP_FAILED_E, etc.) that no existing test asserts, so a deletion of the corresponding ConstantCompare/integrity check would now be caught. Verified they run and pass.

dgarske (dismissed):
Again, very surprised this didn't propigate any errors. I sent you a note to double check things (force a failure to make sure its actually reported). For now marking approved, but holding merge since it did not change/fix any core code...

JeremiahM37:
Jenkins retest this please

SparkiDev (approved):

PR #10293, opened 2026-04-23
Remove liboqs for ML-KEM and ML-DSA, update for Falcon

Frauschi:
Remove liboqs integrations for ML-KEM and ML-DSA, building only on our own implementations. The liboqs integration for these two algorithms has been formally deprecated for a long time. Integration for SPHINCS+ via liboqs will be independently removed in #10261.

In this process, the two configuration macros `WOLFSSL_HAVE_MLKEM` and `HAVE_DILITHIUM` are now the sole option required to enable the desired algorithm. The additional options `WOLFSSL_WC_MLKEM` and `WOLFSSL_WC_DILITHIUM` have been removed, as they are redundant now. 

The current Falcon (future FN-DSA) integration is updated to work with the latest liboqs release (version ) and both OIDs as well as code points are updated to the latest oqs-provider (version ).  Furthermore, the private key DER encoding is updated to be aligned with oqs-provider. Configuration-wise, a new `--enable-falcon` is introduced which has to go together with the existing `--with-liboqs` (which itself requires `--enable-experimental`) to enable this functionality. Same goes for CMake.

This will need so updates when #9843 and #10261 land, but put it up now to get CI tests working.


Frauschi:
The Jenkins related failure for liboqs is addressed in https://github.com/wolfSSL/testing/pull/947

Frauschi:
Jenkins retest this please

Frauschi:
Jenkins retest this please

douzzer (changes requested):
As an intermediary step, this PR needs `HAVE_SPHINCS` added to `.wolfssl_known_macro_extras`.

A bigger issue is that, for the asm, the feature macro refactor needs to be implemented in the generator scripts, e.g. `scripts/mldsa/x86_64/mldsa.rb` and `sha3/x86_64/sha3_x4_avx2.rb`.

Frauschi:
Opened https://github.com/wolfSSL/scripts/pull/574 for the asembly generation script fixes.

douzzer (approved):

PR #10323, opened 2026-04-27
Speedup DTLS max fragment size calculation for MTU limits

julek-wolfssl:
Refactor wolfssl_local_GetRecordSize to compute the record size directly
from cipher specs instead of calling BuildMessage. Add a unit test that
compares the new calculation to BuildMessage's size-only output across
every registered cipher suite and supported (D)TLS version.


src/internal.c
```diff
@@ -42252,30 +42252,100 @@ int wolfSSL_AsyncPush(WOLFSSL* ssl, WC_ASYNC_DEV* asyncDev)
...
+                + ssl->specs.aead_mac_size;
+    #ifdef WOLFSSL_DTLS13
+        if (ssl->options.dtls && sz < Dtls13MinimumRecordLength(ssl))
+            sz = Dtls13MinimumRecordLength(ssl);
+    #endif
+        return sz;
+    }
+#endif
+
+    /* TLS 1.2 / TLS 1.1 / DTLS 1.2 path. Mirror BuildMessage's size
```
rizlik, line 42336: what about TLS 1.3 and DTLS  1.3?

rizlik (changes requested):

julek-wolfssl:
retest this please..

tests/api/test_tls.c
```diff
@@ -1197,7 +1197,8 @@ int test_record_size_matches_build_message(void)
                 ExpectIntEQ(test_memio_setup_ex(&test_ctx, &ctx_c, &ctx_s,
                         &ssl_c, &ssl_s, versions[v].client, versions[v].server,
                         (byte*)ca_sm2_der, (int)sizeof_ca_sm2_der,
-                        (byte*)server_sm2_der, (int)sizeof_server_sm2_der,
+                        (byte*)server_sm2_cert_der,
+                        (int)sizeof_server_sm2_cert_der,
```
rizlik, line 1201: Consider using path or load from path as it's less noiser for the version control system.

rizlik (commented):
I've doubts on this refactor.
I'm trying to understand what makes the `sizeOnly=1` path in `BuildMessage` expensive enough to justify the duplication. 
For non-SCR paths the overhead looks modest — `XMEMSET(BuildMsgArgs)` and the `state-machine traversal`.
That said, you measured this and added the safety net test, so I'm comfortable if you think the current shape is the right trade-off

julek-wolfssl:
@rizlik closing in favor of https://github.com/wolfSSL/wolfssl/pull/10476

PR #10326, opened 2026-04-27
Add maximum_name_length to TLS ECH padding

sebastian-carpenter:
# Description

Improve TLS ECH ClientHelloInner padding to follow RFC 9849 §6.1.3, which uses the ECHConfig maximum_name_length to produce a stable padded length across SNIs.

  - Add WOLFSSL_EchConfig.maxNameLen and honor it when emitting/parsing ECHConfigs (previously hard-coded to 0 on emit and ignored on parse).
  - Add new public API wolfSSL_CTX_GenerateEchConfigEx() that takes a maxNameLen; the existing wolfSSL_CTX_GenerateEchConfig() becomes a thin wrapper passing 0.
  - Update SendTls13ClientHello() to compute padding from maxNameLen and the inner SNI length (with the spec's 9-byte fallback when no SNI is present), then round up to a 32-byte boundary.

Fixes zd#21504

# Testing

Update callback in `tests/api.c` to use new `*Ex` function with a longer maximum name length. Kept some instances of the old `wolfSSL_CTX_GenerateEchConfig` to test with no max name length (i.e., shorter length).

# Checklist

 - [X] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/wolfssl/internal.h
+++ b/wolfssl/internal.h
@@ -3147,6 +3147,7 @@ typedef struct WOLFSSL_EchConfig {
     byte configId;
     byte numCipherSuites;
     byte receiverPubkey[HPKE_Npk_MAX];
+    byte maxNameLen;
 } WOLFSSL_EchConfig;
 
 typedef struct WOLFSSL_ECH {
--- a/src/ssl_ech.c
+++ b/src/ssl_ech.c
@@ -33,6 +33,15 @@
 int wolfSSL_CTX_GenerateEchConfig(WOLFSSL_CTX* ctx, const char* publicName,
     word16 kemId, word16 kdfId, word16 aeadId)
 {
+    return wolfSSL_CTX_GenerateEchConfigEx(ctx, publicName, kemId, kdfId,
+            aeadId, 0);
+}
+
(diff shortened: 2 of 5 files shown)


sebastian-carpenter:
Jenkins retest this please.

wolfssl/internal.h
```diff
@@ -3137,6 +3137,7 @@ typedef struct WOLFSSL_EchConfig {
     byte configId;
     byte numCipherSuites;
     byte receiverPubkey[HPKE_Npk_MAX];
+    byte maxNameLen;
```
dgarske, line 3150: `byte`? up to 255? Is that really the right limit?

dgarske (changes requested):

wolfssl/internal.h
sebastian-carpenter: Yeah.... The RFC has it as a [uint8 maximum_name_length;](https://www.rfc-editor.org/rfc/rfc9849.html#name-encrypted-clienthello-confi).

sebastian-carpenter:
Jenkins retest this please.

dgarske (approved):

PR #10332, opened 2026-04-27
tls.c: send missing_extension alert on TLS 1.3 SNI absence

jackctj117:
This pull request updates the alert handling logic in the `TLSX_SNI_VerifyParse` function to improve compliance with TLS 1.3 requirements. Specifically, it changes the type of alert sent when the SNI (Server Name Indication) extension is absent or does not match, sending a `missing_extension` alert for TLS 1.3 and above, while maintaining the previous behavior for earlier versions.

Alert handling improvements for TLS 1.3:

* Updated `TLSX_SNI_VerifyParse` in `src/tls.c` to send a `missing_extension` alert instead of `handshake_failure` when the SNI extension is missing or has no match, but only for TLS 1.3 and above. For earlier versions, the function still sends `handshake_failure`. [[1]](diffhunk://#diff-bb132d62ad3f3a5eac6006e2a383ad909e30072360265a49a962ac765962be34L2598-R2601) [[2]](diffhunk://#diff-bb132d62ad3f3a5eac6006e2a383ad909e30072360265a49a962ac765962be34L2609-R2615)

### DIFF
--- a/src/tls.c
+++ b/src/tls.c
@@ -2593,7 +2593,10 @@ static int TLSX_SNI_VerifyParse(WOLFSSL* ssl,  byte isRequest)
                         continue;
                 }
 
-                SendAlert(ssl, alert_fatal, handshake_failure);
+                SendAlert(ssl, alert_fatal,
+                          IsAtLeastTLSv1_3(ssl->version)
+                              ? missing_extension
+                              : handshake_failure);
                 WOLFSSL_ERROR_VERBOSE(SNI_ABSENT_ERROR);
                 return SNI_ABSENT_ERROR;
             }
@@ -2604,7 +2607,10 @@ static int TLSX_SNI_VerifyParse(WOLFSSL* ssl,  byte isRequest)
                 if (ssl_sni->status != WOLFSSL_SNI_NO_MATCH)
                     continue;
 
-                SendAlert(ssl, alert_fatal, handshake_failure);
+                SendAlert(ssl, alert_fatal,
+                          IsAtLeastTLSv1_3(ssl->version)
+                              ? missing_extension
+                              : handshake_failure);
                 WOLFSSL_ERROR_VERBOSE(SNI_ABSENT_ERROR);
                 return SNI_ABSENT_ERROR;
             }


ColtonWilley:
Jenkins retest this please.

dgarske:
Jenkins retest this please. PRB-fsanitize-addr-v3 -> FAIL: scripts/resume.test
The triage shows this was a timeout

src/tls.c
```diff
@@ -2606,7 +2609,10 @@ static int TLSX_SNI_VerifyParse(WOLFSSL* ssl,  byte isRequest)
                 if (ssl_sni->status != WOLFSSL_SNI_NO_MATCH)
                     continue;
 
-                SendAlert(ssl, alert_fatal, handshake_failure);
+                SendAlert(ssl, alert_fatal,
+                          IsAtLeastTLSv1_3(ssl->version)
+                              ? missing_extension
+                              : handshake_failure);
```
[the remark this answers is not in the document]
jackctj117: Confirmed via end-to-end testing that
  the suggested swap regresses the SSL-level "client omits SNI" path this PR is fixing,
  because the 2nd branch is reachable from both absent and present-but-no-match states with
  no local signal to distinguish them — filing a follow-up for the proper fix instead.

src/tls.c
```diff
@@ -2595,7 +2595,10 @@ static int TLSX_SNI_VerifyParse(WOLFSSL* ssl,  byte isRequest)
                         continue;
                 }
 
-                SendAlert(ssl, alert_fatal, handshake_failure);
+                SendAlert(ssl, alert_fatal,
+                          IsAtLeastTLSv1_3(ssl->version)
+                              ? missing_extension
+                              : handshake_failure);
```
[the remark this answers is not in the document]
jackctj117: Two copies of a one-line ternary doesn't warrant a helper, and a correct fix for Comment 1 will make the two branches diverge anyway.

src/tls.c
```diff
@@ -2593,7 +2593,10 @@ static int TLSX_SNI_VerifyParse(WOLFSSL* ssl,  byte isRequest)
                         continue;
                 }
 
-                SendAlert(ssl, alert_fatal, handshake_failure);
+                SendAlert(ssl, alert_fatal,
+                          IsAtLeastTLSv1_3(ssl->version)
+                              ? missing_extension
```
dgarske, line 2598: @SparkiDev do we really want to reply with this specific error code when the SNI is not found? Sometimes we mitigate possible leaks by sending a generic error despite the "spec". Curious what you think Sean

dgarske: @SparkiDev please review this comment. If its okay then approve and merge. thanks

SparkiDev: I don't have a problem with sending the specific alert as I don't see how the attacker could use this information.

SparkiDev (approved):

PR #10340, opened 2026-04-28
harden falcon key handling

JeremiahM37:
Fixes ~F-3316~, F-3317, F-3320, ~F-3321~.

- ~wc_sphincs_check_key: actually validate level/optim/keySet flags and compare embedded PK against key->p (was always returning 0)~
- wc_falcon_check_key: reject unset level and partially-populated keys before comparison (was passing on default-init key)
- GetKeyOID falcon: replace dead else-if chain so level 5 keys are detected (was only ever trying level 1)
- ~GetKeyOID sphincs: replace dead else-if chain so all six variants are detected (was only ever trying fast-L1)~

### DIFF
--- a/wolfcrypt/src/asn.c
+++ b/wolfcrypt/src/asn.c
@@ -9530,8 +9530,8 @@ int wc_GetKeyOID(byte* key, word32 keySz, const byte** curveOID, word32* oidSz,
             return MEMORY_E;
 
         if (wc_falcon_init(falcon) == 0) {
-            tmpIdx = 0;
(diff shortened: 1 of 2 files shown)


Frauschi:
Please note that the liboqs-based SPHINCS implementation is about to be dropped completely in #10261, so no changes for this are required.

JeremiahM37:
Jenkins retest this please

Frauschi (approved):

dgarske (approved):

PR #10341, opened 2026-04-28
Remove the amount of macOS based tests in CI

Frauschi:
Reduce the number of tests running on macOS in `os-check.yml` to the minimum required number to cover all mac os specific features. These currently slow down our CI testing dramatically. All other platform-agnostic configs and setups are only tested on Linux, which is much faster in GitHub CI.


Frauschi:
Jenkins retest this please

julek-wolfssl (approved):

douzzer (approved):

PR #10292, opened 2026-04-23
Remove deprecated liblms and libxmss

Frauschi:
Remove all support and references to both liblms and libxmss, which have been used to integrate support of LMS and XMSS stateful hash-bash PQC signature algorithms to wolfssl before we had our own implementations.

As we have our own implementations now and both integrations have already been formally deprecated for a long time, its time to remove the dependencies and simplify both the code and the SBOM. 

This will need some rework when #9843 lands (most probably before this one), but put it up now to ensure CI tests are working.

PR #805, opened 2017-03-17
Fix RNG issue with Intel RD and cleanup to remove old ARC4 support

dgarske:
Fix RNG issue with Intel RDRAND and RDSEED accelerations not being used because HAVE_HASHDRBG was always being defined if !WOLFSSL_FORCE_RC4_DRBG.  Added new --enable-intelrand option to indicate use of RDRAND preference for RNG source (if RDRAND not supported by CPU then HASHDRBG will be used). The --enable-intelasm option enables the RDSEED support for seeding HASHDRBG if CPU supports it. Allow use of seed as RNG source if --disable-hashdbrg (shows build warning). Cleanup to remove old ARC4 RNG support. Fixed random_test return code with !HAVE_HASHDRBG. Cleanup of ./configure --help alignment.

PR #414, opened 2016-05-10
Dtls multi core

JacobBarthelmeh:
Adds the option to use --enable-sessionexport with DTLS and send a session after completing change cipher spec in handshake.

PR #410, opened 2016-05-04
update version to 3.9.1

kaleb-himes:

PR #10343, opened 2026-04-28
feat: SBOM generation and OmniBOR build provenance (CRA compliance)

MarkAtwood:
## Summary

Adds two complementary supply chain transparency targets to the wolfSSL autotools build:

- **`make sbom`** — generates a Software Bill of Materials (SPDX 2.3 + CycloneDX 1.6) for EU Cyber Resilience Act (CRA) compliance
- **`make bomsh`** — generates an OmniBOR artifact dependency graph (cryptographic source-to-binary traceability) using the [Bomsh](https://github.com/omnibor/bomsh) project
- When both are run, `make bomsh` enriches the SPDX document with a `PERSISTENT-ID gitoid` ExternalRef, bridging component identity and build provenance in a single file

## make sbom

Produces three files:

| File | Format |
|---|---|
| `wolfssl-<version>.cdx.json` | CycloneDX 1.6 JSON |
| `wolfssl-<version>.spdx.json` | SPDX 2.3 JSON |
| `wolfssl-<version>.spdx` | SPDX 2.3 tag-value (validated by pyspdxtools) |

SBOM contents: name, version, supplier, license (parsed from `LICENSING` at generation time), copyright, SHA-256 of installed library, CPE, PURL, download location, build configuration defines, and optional third-party dependency versions (liboqs, libz, libxmss, liblms).

Implementation: `scripts/gen-sbom` (Python 3, stdlib only) stages a `make install` into a temp directory, hashes the installed library, generates both formats, removes staging. `configure.ac` detects `python3`, `pyspdxtools`, and `git` via `AC_PATH_PROG`. `make sbom` fails with a clear error if either required tool is missing.

## make bomsh

Runs a full clean rebuild under `bomtrace3` (patched strace — userspace only, no kernel modifications) to produce an OmniBOR artifact graph in `omnibor/`. If `bomsh_sbom.py` is available and `wolfssl-<version>.spdx.json` exists, annotates the SPDX with a `PERSISTENT-ID gitoid` ExternalRef.

`configure.ac` detects `bomtrace3`, `bomsh_create_bom.py`, and `bomsh_sbom.py` via `AC_PATH_PROG`. The raw logfile and conf file are written to the build directory (not `/tmp/`) to avoid concurrent-build collisions. All generated files removed by `make clean`.

## Install targets

```sh
make install-sbom     # installs SBOM files to $(datadir)/doc/wolfssl/
make install-bomsh    # installs omnibor/ and enriched SPDX to $(datadir)/doc/wolfssl/
```

## Documentation

- `doc/SBOM.md` — unified reference covering both targets, combined workflow, all output files, implementation notes
- `doc/CRA.md` — product-integrator guide: how to incorporate wolfSSL's SBOM artefacts into a downstream product SBOM (SPDX `ExternalDocumentRef`, CycloneDX component reference), commercial license guidance, OmniBOR gitoid meaning, auditor handoff checklist, links to OpenSSF CRA and SBOM Everywhere SIG guidance
- `INSTALL`: sections 21 and 22
- `README.md`: brief SBOM/CRA and OmniBOR/Bomsh sections
- `doc/include.am`: `SBOM.md` and `CRA.md` added to `dist_doc_DATA`

## Test plan

- [ ] `./autogen.sh && ./configure && make && make sbom` — verify three SBOM files produced and SPDX validates
- [ ] `make install-sbom` / `make uninstall-sbom` — verify files install and remove cleanly
- [ ] `make clean` — verify all generated files removed
- [ ] With `bomtrace3` and `bomsh_create_bom.py` in PATH: `make sbom && make bomsh` — verify `omnibor/` produced and `omnibor.wolfssl-<ver>.spdx.json` contains `PERSISTENT-ID gitoid` ExternalRef
- [ ] Without `bomtrace3` in PATH: `make bomsh` — verify clear error message, non-zero exit
- [ ] Without prior `make sbom`: `make bomsh` — verify OmniBOR graph produced, NOTE printed about missing SPDX, no failure


Frauschi:
Note that we are in the process of removing liblms and libxmss (see #10292), which will simplify the dependency analysis in the python script.

MarkAtwood:
Three CI failures to address:

**1. `bomsh integration (linux)` — upstream repo restructured**
The install step checks for `.devcontainer/bomtrace3/` in the omnibor/bomsh repo, which no longer exists. Hit our own error string: `bomsh repo layout changed; please update CI`. Check the current bomsh repo layout and fix the install step.

**2. `SBOM integration (linux)` — `liboqs-dev` not in apt**
`E: Unable to locate package liboqs-dev` — that package isn't in Ubuntu 24.04's default repos. Either build liboqs from source in CI, use a PPA, or make that step conditional.

**3. `Windows Build Test (x64)` — `tcp bind failed`**
Build succeeded (0 errors), runtime fails with `wolfSSL error: tcp bind failed`. Looks like port contention on the Windows runner. Please check if this also fails on master — if so it's pre-existing and not our bug, and we can ignore it.

PRB-master-job failure appears to be a reporting artifact (no test results found), not a real failure.

Frauschi:
> 3. `Windows Build Test (x64)` — `tcp bind failed`

FYI, this will be resolved with https://github.com/wolfSSL/wolfssl/pull/10387

sameehj:
retest this please

MarkAtwood (commented):
## Pre-merge self-review

Ran a structured pre-merge review across `scripts/gen-sbom`, `scripts/test_gen_sbom.py`, `.github/workflows/sbom.yml`, `Makefile.am`, and the docs. Five commits result, each listing its closed findings in the trailer:

```
test(sbom): expand coverage; CI hard-requires pcpp
fix(sbom): correctness fixes in gen-sbom
build(sbom): Makefile.am hygiene
ci(sbom): pin deps, harden token, partial-clone bomsh
docs(sbom): correct NTIA/SPDX claims; fix broken anchor
```

### Real defects fixed (`fix(sbom)`)

1. **`parse_options_h` URL corruption.** `re.split(r'/\*|//', raw)[0]` truncated `PACKAGE_URL "https://..."` at the first `//` inside the URL, emitting `PACKAGE_URL = "https:` in the SBOM `wolfssl:build:*` properties. Replaced with a string-literal-aware state machine.
2. **`detect_license` missed the canonical FSF preamble.** `or\s+(any\s+)?later` failed on `or (at your option) any later version` (the FSF GPLv2 preamble used in millions of upstream COPYING files); silently mis-detected as `GPL-X.0-only`. Broadened with a sentence-bounded filler match.
3. **Empty `--lib` accepted.** A misconfigured build pointing `--lib` at `/dev/null` or an empty `.a` silently emitted the well-known empty-file SHA-256 (`e3b0c44…b855`) as the wolfSSL component checksum; both SPDX and CDX validators accepted it. Hard-fail gate on size 0 with an actionable error; `--srcs` zero-byte placeholders get a soft warning (cross-compile setups legitimately use them).

### Coverage gaps closed (`test(sbom)`)

+42 tests covering `detect_license` (had **zero** tests), `sha256_file`'s chunked-read path (the loop body wasn't exercised by the only existing empty-file vector), all four SBOM document generators (`generate_cdx` / `generate_spdx` / `cdx_dep_component` / `spdx_dep_package`), and the LicenseRef-without-license-text CLI rejection. Independent oracles throughout — SPDX licence list, `hashlib` one-shot SHA-256, CDX 1.6 / SPDX 2.3 schema field semantics.

`TestParseUserSettings.setUp` previously called `self.skipTest` on missing `pcpp`. The unit job didn't install `pcpp`, so 9 standalone-path tests silently skipped on every PR run. Now hard-fails with an actionable install hint, and the unit job installs `pcpp==1.30.*`.

### Build hygiene (`build(sbom)`)

Five Makefile.am improvements — none changes user-facing behaviour for any correct invocation; each closes a specific failure mode:

- bomsh artifact glob quoted for spaces in `abs_builddir`
- `SOURCE_DATE_EPOCH` only exported when `git log -1 --format=%ct` returned non-empty (silent shallow-clone case)
- `--license-override` / `--license-text` only passed when set (decouples Makefile from gen-sbom's argparse default)
- `install-bomsh` switched from a two-tar pipeline (which masked source-side errors due to POSIX sh's pipeline exit-status semantics) to `cp -R src/. dst/`
- `uninstall-hook` depends on `uninstall-sbom` + `uninstall-bomsh` instead of duplicating their bodies

### CI hardening (`ci(sbom)`)

`pcpp==1.30.*` pinned, `permissions: contents: read` at workflow level (defence-in-depth — workflow does no API writes), `git clone --filter=blob:none` for the bomsh fetch (saves ~10s per run; pinned SHA makes blob filter safe).

### Doc corrections (`docs(sbom)`)

- `README.md` previously said "Both produce SPDX 2.3 + CycloneDX 1.6 JSON validated against NTIA minimum elements". Neither path runs an NTIA-minimum-elements checker. Only `make sbom` runs `pyspdxtools` (SPDX-spec, not NTIA-specific). Corrected.
- `doc/CRA.md` named `ntia-conformance-checker` as enforcing SPDX §10.1 `hasExtractedLicensingInfos`; it does not (that is pyspdxtools' job). Corrected with explicit text on what `ntia-conformance-checker` actually validates so a reader cannot be misled into running a malformed LicenseRef SBOM through the wrong tool and getting a false pass.
- `doc/SBOM.md` anchor link `#3-make-bomsh` resolved to the wrong section; fixed to the GFM auto-anchor for §3.1.

### Re the Copilot inline review

The Copilot review raised 8 concerns. 7 are addressed: the NOASSERTION CDX wording in SBOM.md §2.4, `make bomsh` → `make sbom` rebuild ordering, hardcoded library paths replaced with the `WOLFSSL_LIB_DSO_BASENAMES` glob, `scripts/gen-sbom` added to EXTRA_DIST, trap-based staging cleanup, and removing the hardcoded `licenseListVersion`.

The remaining concern — `install-bomsh` using a non-`INSTALL_DATA` copy mechanism — is unchanged in intent (this round uses `cp -R` rather than the previous `tar` pipeline for clean error propagation but did not adopt `find ... | xargs $(INSTALL_DATA)`). The OmniBOR tree is a directory of content-addressed binary objects; using `$(INSTALL_DATA)` per-file is the more automakey approach but adds complexity for a directory whose layout is meant to be preserved verbatim. Open to revisiting if maintainers prefer.

### Acknowledgement

@sameehj — the standalone `--user-settings` / `--srcs` entry point, SPDX 2.3 LicenseRef compliance, library-discovery glob, and CI workflow shape this PR from "works on my machine" into a production-ready, embedded-customer-friendly tool. The test infrastructure you built was solid. Thanks.

cc @Frauschi for visibility — `make clean` removes both the doxygen artefacts and `BOMSH_OMNIBORDIR` via `doc/include.am`'s `clean-local:`.

sameehj:
retest this please

sameehj:
retest this please

dgarske (commented):
## Skoll Code Review

### Posted findings
- [Medium] **make sbom does not thread --dep-version; liboqs/zlib version becomes NOASSERTION without pkg-config** — `Makefile.am:409-475 (sbom target)`
- [Low] **Redundant/dead branch logic in FIPS product bucket computation** — `scripts/gen-advisory:product_model (FIPS block)`
- [Low] **Duplicate pkg-config invocation per dependency** — `scripts/gen-sbom:cdx_dep_component / spdx_dep_package`
- [Info] **Executable-bit inconsistency between gen-sbom (755) and gen-advisory (644)** — `scripts/gen-advisory:1`

Makefile.am
```diff
@@ -351,3 +351,304 @@ merge-clean:
...
+	echo "SBOM: hashing $$sbom_lib"; \
+	if test -z "$${SOURCE_DATE_EPOCH:-}" && test -n "$(GIT)" && \
+	   $(GIT) -C "$(srcdir)" rev-parse --git-dir >/dev/null 2>&1; then \
+	    sde=`$(GIT) -C "$(srcdir)" log -1 --format=%ct 2>/dev/null`; \
+	    if test -n "$$sde"; then \
+	        SOURCE_DATE_EPOCH="$$sde"; \
+	        export SOURCE_DATE_EPOCH; \
+	    fi; \
+	fi; \
+	$(PYTHON3) $(srcdir)/scripts/gen-sbom \
```
dgarske, line 574: **🟠 [Medium] make sbom does not thread --dep-version; liboqs/zlib version becomes NOASSERTION without pkg-config**

The autotools `sbom:` recipe passes `--dep-libz $(ENABLED_LIBZ)` / `--dep-liboqs $(ENABLED_LIBOQS)` but never passes `--dep-version`. `gen-sbom` then resolves the linked dependency version solely via `pkg-config --modversion`; if pkg-config is absent or the dep's `.pc` file is missing on the build host, the SPDX records `versionInfo: NOASSERTION` and CycloneDX omits the version/purl. The `--dep-version` override flag (and the embedded entry point) explicitly exists for exactly this case, but the autotools path offers no make-variable to feed it through. For a `--with-liboqs` build packaged on a host without `liboqs.pc`, the SBOM silently loses the version a CVE scanner keys on. This is documented in doc/SBOM.md §2.4 as intended, so this is a confirm-intent note rather than a bug.

**Fix:** Consider exposing a `SBOM_DEP_VERSIONS` make variable that forwards to `--dep-version`, so autotools packagers on hosts lacking pkg-config can still record dep versions instead of NOASSERTION. If the documented NOASSERTION fallback is acceptable for all autotools consumers, no change needed.

scripts/gen-sbom
```diff
@@ -0,0 +1,1184 @@
...
+        seen.add(name)
+        entries.append((name, gitoid_blob_sha256(path)))
+    entries.sort()
+    h = hashlib.sha256()
+    for name, oid in entries:
+        h.update(f'{name}\x00{oid}\n'.encode())
+    return h.hexdigest()
+
+
+def cdx_dep_component(name, pkg_version, key, dep_version_overrides=None):
```
dgarske, line 741: **🔵 [Low] Duplicate pkg-config invocation per dependency**

Each enabled dependency resolves its version twice: `generate_cdx` -> `cdx_dep_component` -> `dep_version()` and `generate_spdx` -> `spdx_dep_package` -> `dep_version()`. With no `--dep-version` override, each call shells out to `pkg-config --modversion`, so a default `--with-liboqs --with-libz` build runs pkg-config four times instead of two. Harmless and fast, but the version could be resolved once per dep and passed into both generators for clarity and to guarantee CDX/SPDX never disagree if pkg-config output were ever non-deterministic.

**Fix:** Optional cleanup: resolve each dep version a single time and share it between the CDX and SPDX emitters.

scripts/gen-advisory
```diff
@@ -0,0 +1,848 @@
+#!/usr/bin/env python3
```
dgarske: **⚪ [Info] Executable-bit inconsistency between gen-sbom (755) and gen-advisory (644)**

`scripts/gen-sbom` is added with mode 100755 while `scripts/gen-advisory` is added with mode 100644, even though both carry a `#!/usr/bin/env python3` shebang and are sibling generators with parallel structure. Both are invoked via `$(PYTHON3) $(srcdir)/scripts/...` so the bit is not functionally required, but the inconsistent mode is a minor convention wart that a reviewer or packager may notice.

**Fix:** Make the executable bit consistent across the two generator scripts (either both 755 or both 644).

MarkAtwood:
Good catch on the --dep-version gap. Yes, let's expose a SBOM_DEP_VERSIONS make variable so autotools packagers without pkg-config can still record dep versions.

@sameehj can you pick this up along with the other Skoll findings?

sameehj:
retest this please

MarkAtwood (commented):
## Code Review

### Blockers

**1. `gen-advisory`: `resolved` → `known_affected` mapping inversion**

`_STATE_TO_BUCKET` maps `resolved` to `known_affected`. This is a semantic inversion — patched products get reported as actively vulnerable in the CSAF output. Downstream scanners will tell operators to patch what's already fixed. Should be `'resolved': 'fixed'`. Same gap for `resolved_with_pedigree` which is absent from the mapping entirely and falls through to the `known_affected` default.

**2. `gen-advisory`: unmapped states default to `known_affected`**

```python
bucket = _STATE_TO_BUCKET.get(state, 'known_affected')
```

`false_positive` and any other unmapped CycloneDX analysis state silently becomes "actively vulnerable." Either map them correctly (e.g. `false_positive` → `known_not_affected`) or raise an error on unknown states rather than defaulting to the worst-case bucket.

**3. `Makefile.am`: `uninstall-advisory` missing from `uninstall-hook`**

```make
uninstall-hook: uninstall-sbom uninstall-bomsh
```

`uninstall-advisory` is absent. `make install-advisory && make uninstall` leaves advisory files behind.

**4. `Makefile.am`: `install-advisory` glob loop breaks on empty match**

```make
@for f in $(ADVISORY_OUT_DIR)/*.json; do \
    $(INSTALL_DATA) "$$f" $(DESTDIR)$(advisorydir)/; \
done
```

On POSIX sh (without `nullglob`), when the glob matches nothing the loop body executes with the literal glob string, calling `$(INSTALL_DATA)` on a nonexistent file. Add `[ -f "$$f" ] || continue` as the first statement in the loop body.

---

### High

**5. CVSS scoring errors in CVE records**

Both CVE records use `privilegesRequired: LOW`. Neither vulnerability requires system account access — obtaining a TLS cert (5501) or sending network traffic (5778) are `PR:N` scenarios. CVE-2026-5778 also uses `userInteraction: PASSIVE` for a sniffer crash triggered by network traffic with no human involvement — should be `UI:N`. These errors understate the severity scores.

**6. `gen-sbom` TOCTOU in `gitoid_blob_sha256`**

`os.path.getsize(path)` is called before `open(path, 'rb')`. If the file changes between the two calls, the gitoid header encodes the wrong size. Fix: use `os.fstat(f.fileno()).st_size` after opening the file descriptor (same approach git itself uses).

**7. `bomsh_verify.py` path traversal via unsanitized gitoid**

The gitoid string extracted from the SPDX document is used directly in `os.path.join(omnibor_objects_dir, gid[:2], gid[2:])` without validating that it is a 40-character hex string. A crafted SPDX file with `gitoid:blob:sha1:../../../etc/shadow` passes the prefix check and enables path traversal. Impact is read-only (`os.path.isfile`), but still an info-disclosure risk when SPDX comes from untrusted build artifacts. Fix: `re.fullmatch(r'[0-9a-f]{40}', gid)` before path construction.

**8. `EXTRA_DIST` hardcodes CVE filenames**

```make
EXTRA_DIST += advisories/records/CVE-2026-5501.json \
              advisories/records/CVE-2026-5778.json
```

Every new CVE record requires a manual `Makefile.am` edit or it's silently missing from `make dist` tarballs. The failure is silent — `make dist` succeeds, downstream `make advisory` fails. Use a dist-hook glob instead.

---

### Medium

**9. CI actions pinned to tags, not SHAs**

All `uses:` lines reference floating tags (`actions/checkout@v4`, etc.). For a supply-chain provenance workflow this is the most visible inconsistency — the SBOM workflow's own build inputs are mutable. The bomsh clone correctly uses a full commit SHA; actions and the liboqs/strace clones should follow the same pattern.

**10. CSAF `category: 'self'` with a generic landing page URL**

```python
'references': [{'summary': '...', 'url': WOLFSSL_ADVISORIES_URL, 'category': 'self'}]
```

CSAF 2.0 mandatory test 6.1.20 requires the `self` reference to point to the canonical location of this specific document, not a generic advisories page. This will fail conformance validation when the document is published.

**11. VEX overlay schema doesn't require `state`**

The `overlayEntry` object has no `required` array, so the core VEX determination field (`state`) is optional. An empty overlay entry `{}` validates silently against the schema. Add `"required": ["state"]` to the `overlayEntry` definition.

sameehj:
retest this please

sameehj:
> ## Code Review
> ### Blockers
> **1. `gen-advisory`: `resolved` → `known_affected` mapping inversion**
> 
> `_STATE_TO_BUCKET` maps `resolved` to `known_affected`. This is a semantic inversion — patched products get reported as actively vulnerable in the CSAF output. Downstream scanners will tell operators to patch what's already fixed. Should be `'resolved': 'fixed'`. Same gap for `resolved_with_pedigree` which is absent from the mapping entirely and falls through to the `known_affected` default.
> 
Won't change, by design. The bucket applies to the CVE's vulnerable version ranges, which stay known_affected permanently; the patched release is emitted separately into fixed in the per-product loop. Mapping resolved → fixed would mislabel the old vulnerable ranges as patched. Added resolved_with_pedigree to the explicit map (the real gap). Documented in-code.

> **2. `gen-advisory`: unmapped states default to `known_affected`**
> 
> ```python
> bucket = _STATE_TO_BUCKET.get(state, 'known_affected')
> ```
> 
> `false_positive` and any other unmapped CycloneDX analysis state silently becomes "actively vulnerable." Either map them correctly (e.g. `false_positive` → `known_not_affected`) or raise an error on unknown states rather than defaulting to the worst-case bucket.
> 
Fixed

> **3. `Makefile.am`: `uninstall-advisory` missing from `uninstall-hook`**
> 
> ```makefile
> uninstall-hook: uninstall-sbom uninstall-bomsh
> ```
> 
> `uninstall-advisory` is absent. `make install-advisory && make uninstall` leaves advisory files behind.
> 
Fixed

> **4. `Makefile.am`: `install-advisory` glob loop breaks on empty match**
> 
> ```makefile
> @for f in $(ADVISORY_OUT_DIR)/*.json; do \
>     $(INSTALL_DATA) "$$f" $(DESTDIR)$(advisorydir)/; \
> done
> ```
> 
> On POSIX sh (without `nullglob`), when the glob matches nothing the loop body executes with the literal glob string, calling `$(INSTALL_DATA)` on a nonexistent file. Add `[ -f "$$f" ] || continue` as the first statement in the loop body.
> 
Fixed
> ### High
> **5. CVSS scoring errors in CVE records**
> 
> Both CVE records use `privilegesRequired: LOW`. Neither vulnerability requires system account access — obtaining a TLS cert (5501) or sending network traffic (5778) are `PR:N` scenarios. [CVE-2026-5778](https://github.com/advisories/GHSA-2jvp-h4w4-2vxh) also uses `userInteraction: PASSIVE` for a sniffer crash triggered by network traffic with no human involvement — should be `UI:N`. These errors understate the severity scores.
> 
Fixed
> **6. `gen-sbom` TOCTOU in `gitoid_blob_sha256`**
> 
> `os.path.getsize(path)` is called before `open(path, 'rb')`. If the file changes between the two calls, the gitoid header encodes the wrong size. Fix: use `os.fstat(f.fileno()).st_size` after opening the file descriptor (same approach git itself uses).
> 
Fixed
> **7. `bomsh_verify.py` path traversal via unsanitized gitoid**
> 
> The gitoid string extracted from the SPDX document is used directly in `os.path.join(omnibor_objects_dir, gid[:2], gid[2:])` without validating that it is a 40-character hex string. A crafted SPDX file with `gitoid:blob:sha1:../../../etc/shadow` passes the prefix check and enables path traversal. Impact is read-only (`os.path.isfile`), but still an info-disclosure risk when SPDX comes from untrusted build artifacts. Fix: `re.fullmatch(r'[0-9a-f]{40}', gid)` before path construction.
> 
Fixed
> **8. `EXTRA_DIST` hardcodes CVE filenames**
> 
> ```makefile
> EXTRA_DIST += advisories/records/CVE-2026-5501.json \
>               advisories/records/CVE-2026-5778.json
> ```
> 
> Every new CVE record requires a manual `Makefile.am` edit or it's silently missing from `make dist` tarballs. The failure is silent — `make dist` succeeds, downstream `make advisory` fails. Use a dist-hook glob instead.
> 
Fixed
> ### Medium
> **9. CI actions pinned to tags, not SHAs**
> 
> All `uses:` lines reference floating tags (`actions/checkout@v4`, etc.). For a supply-chain provenance workflow this is the most visible inconsistency — the SBOM workflow's own build inputs are mutable. The bomsh clone correctly uses a full commit SHA; actions and the liboqs/strace clones should follow the same pattern.
> 
Fixed
> **10. CSAF `category: 'self'` with a generic landing page URL**
> 
> ```python
> 'references': [{'summary': '...', 'url': WOLFSSL_ADVISORIES_URL, 'category': 'self'}]
> ```
> 
> CSAF 2.0 mandatory test 6.1.20 requires the `self` reference to point to the canonical location of this specific document, not a generic advisories page. This will fail conformance validation when the document is published.
> 
Fixed
> **11. VEX overlay schema doesn't require `state`**
> 
> The `overlayEntry` object has no `required` array, so the core VEX determination field (`state`) is optional. An empty overlay entry `{}` validates silently against the schema. Add `"required": ["state"]` to the `overlayEntry` definition.
Fixed

sameehj:
retest this please

MarkAtwood:
Holding merge until after the current release — wolfSSL/wolfssl is in code freeze. Will pick this up once master reopens.

Makefile.am
```diff
@@ -426,3 +426,323 @@ merge-clean:
...
+	    echo "ERROR: 'bomsh_create_bom.py' not found in PATH. Cannot process OmniBOR data."; \
+	    echo "       Install from: https://github.com/omnibor/bomsh"; \
+	    echo ""; \
+	    exit 1; \
+	fi
+	$(MAKE) clean
+	@printf 'raw_logfile=%s\n' '$(BOMSH_RAWLOG_BASE)' > '$(BOMSH_CONF)'
+	$(BOMTRACE3) -c '$(BOMSH_CONF)' $(MAKE)
+	$(BOMSH_CREATE_BOM) -r '$(BOMSH_RAWLOG)' -b '$(BOMSH_OMNIBORDIR)'
+	@bomsh_artifact=""; \
```
dgarske: **🟠 [Medium] bomsh recipe shell block lacks set -e; a failed nested make sbom is masked**

The `bomsh:` recipe's multi-statement shell block beginning `@bomsh_artifact=""; \` (line 698) is a single `/bin/sh -c` invocation joined by line-continuations, but unlike the sibling `sbom:` recipe (`@set -e; \` at line 507) and `advisory:` recipe (`@set -e; \` at line 610) it does NOT enable `set -e`. As a result, if the nested `$(MAKE) sbom SBOM_LIB_OVERRIDE="$$bomsh_artifact"` (line 712) fails, the shell does not abort: it falls through to the `if test -z "$(BOMSH_SBOM)"` check and, when `bomsh_sbom.py` is present, runs `$(BOMSH_SBOM)` to enrich whatever `$(SBOM_SPDX)` is on disk. The recipe's exit status is that of the LAST command, so a failed SBOM regeneration/validation can be swallowed and `make bomsh` reports success. This is precisely the case where the `make sbom` path writes the `.spdx.json` but the subsequent `pyspdxtools` validation step fails — the enrichment then proceeds against an unvalidated document. The happy-path CI job does not exercise this, so the regression would land silently.

**Fix:** Add `set -e` to the start of the `@bomsh_artifact` block (matching `sbom:`/`advisory:`). The two intentional skip paths already use explicit `exit 0`, so they remain unaffected, while a genuine `make sbom` failure will correctly abort `make bomsh`.

Makefile.am
```diff
@@ -426,3 +426,323 @@ merge-clean:
...
+#         --out-dir advisories/out
+#
+# Inputs  (tracked in git): advisories/records/*.json + advisories/vex-overlay.json
+# Outputs (build artifacts): advisories/out/*.{csaf,cdx}.json
+ADVISORY_RECORDS_DIR = $(srcdir)/advisories/records
+ADVISORY_OVERLAY     = $(srcdir)/advisories/vex-overlay.json
+ADVISORY_OUT_DIR     = $(abs_builddir)/advisories/out
+advisorydir          = $(datadir)/doc/$(PACKAGE)/advisories
+
+.PHONY: advisory install-advisory uninstall-advisory
```
dgarske, line 624: **🔵 [Low] make advisory / install-advisory targets undocumented in INSTALL and README**

This PR adds a full `advisory` / `install-advisory` / `uninstall-advisory` make target trio plus `scripts/gen-advisory`, but the user-facing docs only document `make sbom` (INSTALL section 19) and `make bomsh` (INSTALL section 20). The new `make advisory` workflow, its prerequisites (`python3`), and its output layout (`advisories/out/*.{csaf,cdx}.json`) are not mentioned in INSTALL or README, even though the SBOM/bomsh features got dedicated sections. The behavior is only discoverable from the Makefile comments and the CI workflow. This is a discoverability/consistency gap, not a functional defect.

**Fix:** Add a short INSTALL section (and optionally a README blurb) documenting `make advisory` / `make install-advisory`, its `python3` prerequisite, and the generated CSAF + CycloneDX-VEX outputs, consistent with the existing SBOM/bomsh sections.

dgarske (commented):
## Skoll Code Review

### Posted findings
- [Medium] **bomsh recipe shell block lacks set -e; a failed nested make sbom is masked** — `Makefile.am:698-720`
- [Low] **make advisory / install-advisory targets undocumented in INSTALL and README** — `Makefile.am:598-644`

sameehj:
retest this please

Makefile.am
sameehj: Done.

Makefile.am
sameehj: done

MarkAtwood:
@cconlon — one focused ask: please review **this PR (#10343) only**. It's the keystone of the SBOM / EU CRA work and the one piece with real logic to look at — the `gen-sbom` script plus the `make sbom` / `make bomsh` plumbing. All prior review findings are resolved; dgarske's recent Skoll pass left one open item (a `set -e` in the bomsh recipe) which I'll fold in.

The rest of the rollout does **not** need your eyes — flagging only so you know it's out there, not asking you to review it:
- Product `make sbom` PRs (wolfSSH #1050, wolfTPM #536, wolfMQTT #557, wolfHSM #414, wolfBoot #805, wolfProvider #410, wolfsentry #85, wolfCLU #258, wolfscep #25, wolfEngine #227) — mechanical, per the RUNBOOK pattern, going to each product's maintainer + Skoll.
- cmake target #10752 — same.

Happy to walk through #10343 whenever.

dgarske (dismissed):

dgarske (dismissed):
@sameehj or @MarkAtwood please resolve merge conflicts.

MarkAtwood:
@dgarske Merge conflicts resolved — rebased the branch onto current master (the only content conflict was the codespell ignore_words_list; kept both additions). Verified on the rebased head: full build, make sbom producing all three 5.9.2 artifacts, and both script test suites (162 + 57 tests) pass. Ready for re-review.

scripts/gen-advisory
```diff
@@ -0,0 +1,900 @@
...
+            for fx in prod['fixed']:
+                pid, leaf = _leaf_fixed(pname, fx, prod['model_numbers'])
+                _register(pname, pid, leaf)
+                status_buckets.setdefault('fixed', []).append(pid)
+            if prod['remediation'] and affected_pids:
+                url = adv['references'][0] if adv['references'] else None
+                key = (prod['remediation_category'], prod['remediation'], url)
+                remediations.setdefault(key, []).extend(affected_pids)
+
+        vuln = {
```
rlm2002, line 538: **🟠 [Medium] Empty English CVE description produces an invalid CSAF note (text minLength 1) with exit 0**

`parse_record` sets `description = ''` when no `lang == 'en*'` description is present in the CVE record. `generate_csaf` then unconditionally emits `{'category': 'description', 'text': adv['description'], 'title': 'Vulnerability description'}` and, for a single-record advisory, also uses `advs[0]['description']` as the document summary note text. CSAF 2.0 requires `/vulnerabilities[]/notes[]/text` (and `/document/notes[]/text`) to be non-empty (schema minLength 1). A record missing an English description therefore yields a document that the strict CSAF schema rejects — but gen-advisory still returns exit 0 and writes the file, contradicting the module's stated 'fail-rather-than-emit-garbage' contract (and the CI conformance gate would only catch it if such a record were exercised). The committed fixtures all have descriptions, so this is latent.

**Fix:** Validate that a non-empty English description was parsed (either in `parse_record` or before emitting), and fail loudly rather than write a note with empty text.

Makefile.am
```diff
@@ -438,3 +438,324 @@ merge-clean:
...
+	$(PYTHON3) $(srcdir)/scripts/gen-sbom \
+	    --name $(PACKAGE) \
+	    --version $(PACKAGE_VERSION) \
+	    --license-file $(srcdir)/LICENSING \
+	    $(if $(SBOM_LICENSE_OVERRIDE),--license-override '$(SBOM_LICENSE_OVERRIDE)') \
+	    $(if $(SBOM_LICENSE_TEXT),--license-text '$(SBOM_LICENSE_TEXT)') \
+	    $(if $(SBOM_DOCUMENT_NAMESPACE),--document-namespace '$(SBOM_DOCUMENT_NAMESPACE)') \
+	    --options-h $(abs_builddir)/wolfssl/options.h \
+	    --lib "$$sbom_lib" \
+	    --dep-libz $(ENABLED_LIBZ) \
```
rlm2002: **🔵 [Low] Unquoted $(ENABLED_LIBZ)/$(ENABLED_LIBOQS) can break `make sbom` if ever empty**

The recipe passes `--dep-libz $(ENABLED_LIBZ)` and `--dep-liboqs $(ENABLED_LIBOQS)` unquoted. `gen-sbom` declares these as value-taking options (`--dep-libz`/`--dep-liboqs`). If either substituted variable is ever empty (e.g. a configure path that AC_SUBSTs the var without assigning it), the shell collapses the token and argparse consumes the following `--dep-liboqs`/`--cdx-out` as the option's value, producing either a wrong dependency flag or an `expected one argument` failure. In the standard wolfSSL configure these are always 'yes'/'no', so this is defensive hardening rather than an active bug, but the failure mode is non-obvious.

**Fix:** Quote the substituted values so an empty variable becomes an explicit empty argument (`--dep-libz ""`, which `.lower() == 'yes'` correctly rejects) instead of swallowing the next flag.

scripts/csaf_validate.mjs
```diff
@@ -0,0 +1,83 @@
+// CSAF 2.0 conformance gate for documents emitted by scripts/gen-advisory.
```
rlm2002, line 1: **🔵 [Low] csaf_validate.mjs has no direct unit coverage**

The new CSAF conformance runner is only exercised end-to-end in `.github/workflows/advisory.yml` against generated documents. Its own logic — `summarize()` treating `isValid === false || errors.length > 0` as failure, the gate-vs-advisory split, and the exit-code contract (0 all-pass, 1 any-invalid, 2 usage) — is untested in isolation. A regression in `summarize` (e.g. mis-reading the validator result shape after a `@secvisogram/csaf-validator-lib` bump) would silently turn the gate into a no-op that still exits 0. The Python generators got thorough unit suites; the JS gate did not.

**Fix:** Add a small Node test that feeds a deliberately-invalid CSAF doc through the runner and asserts non-zero exit, plus a valid doc asserting exit 0, so the gate cannot silently degrade.

scripts/gen-advisory
```diff
@@ -0,0 +1,900 @@
...
+    'exploitable': 'known_affected',
+    'in_triage': 'under_investigation',
+    'resolved': 'known_affected',
+    'resolved_with_pedigree': 'known_affected',
+    'false_positive': 'known_not_affected',
+    'not_affected': 'known_not_affected',
+}
+
+
+def _bucket_for(state):
```
rlm2002, line 138: **🔵 [Low] No unit coverage for `_bucket_for` unknown-state hard-fail**

`_bucket_for` deliberately `sys.exit(...)`s on an unrecognized analysis state (a good fail-loud choice that prevents an unknown determination from silently defaulting to `known_affected`). The overlay JSON Schema validates the enum in CI, but the schema check and the Python emitter are separate gates; the test suite exercises the known states (`exploitable`, `not_affected`) but never asserts that an out-of-vocabulary state raises SystemExit. A future edit that changed the KeyError default to a silent fallback would pass every existing test.

**Fix:** Add a small regression test asserting `_bucket_for` (and, by extension, `product_model` with a bad overlay `state`) exits non-zero, pinning the fail-loud contract.

rlm2002 (commented):
## Skoll Code Review

**Findings:** 10 total — 6 posted, 4 skipped

### Posted findings
- [Medium] **make bomsh fails late (after full traced rebuild) when python3/pyspdxtools are missing** — `Makefile.am:bomsh target (recipe ~ lines added for bomsh:)`
- [Medium] **Empty English CVE description produces an invalid CSAF note (text minLength 1) with exit 0** — `scripts/gen-advisory:generate_csaf / parse_record`
- [Low] **SBOM/advisory CI jobs are gated to the wolfssl org, so fork PRs get no test signal** — `.github/workflows/sbom.yml, .github/workflows/advisory.yml (every job)`
- [Low] **Unquoted $(ENABLED_LIBZ)/$(ENABLED_LIBOQS) can break make sbom if ever empty** — `Makefile.am:sbom recipe (--dep-libz/--dep-liboqs)`
- [Low] **csaf_validate.mjs has no direct unit coverage** — `scripts/csaf_validate.mjs:1-83`
- [Low] **No unit coverage for _bucket_for unknown-state hard-fail** — `scripts/gen-advisory:_bucket_for; scripts/test_gen_advisory.py`

### Skipped findings
- [Info] ~~make clean leaves the empty advisories/out directory behind~~
- [Info] ~~BOMSH_OMNIBORDIR cleanup placed in doc/include.am clean-local~~
- [Low] ~~bomsh_verify object-store integrity always hashes with SHA-1 but accepts SHA-256-length blob paths~~
- [Info] ~~load_cve_record fetches from cveawg.mitre.org while comments reference cve.org~~

scripts/gen-advisory
```diff
@@ -0,0 +1,900 @@
...
+
+    state = ov.get('state', 'exploitable')
+    bucket = _bucket_for(state)
+
+    # ---- mainline product(s) from the CVE record's affected[] ----
+    for a in adv['affected']:
+        product = a['product']
+        affected_ranges = []
+        for v in a['versions']:
+            if v.get('status') == 'affected' or a['default_status'] == 'affected':
```
rlm2002: ***product_model mislabels versions when defaultStatus == "affected"***

```
for v in a['versions']:
    if v.get('status') == 'affected' or a['default_status'] == 'affected':
        affected_ranges.append({...})   # treated as a VULNERABLE range
```
The per-version decision should fall back to defaultStatus only when the entry has no explicit status (i.e. `v.get('status', a['default_status']) == 'affected'`). As written, when `defaultStatus == "affected"`, every listed version entry is added to `affected_ranges`, including entries explicitly marked status: "unaffected". That is the standard CVE-5.x "affected-by-default with unaffected/fixed exceptions" pattern, and here it would emit a fixed release as a `known_affected` range in CSAF and an affected range in the CycloneDX VEX — the opposite of the truth.

scripts/gen-advisory
MarkAtwood: Fixed. `parse_record` now hard-fails on a missing or blank English description (a0f3a090), so an invalid CSAF note can no longer reach the emitter with exit 0. Added unit tests for both cases (no `en` description, whitespace-only) in da65bf506b; verified they fail against the pre-guard script.

Makefile.am
MarkAtwood: Fixed in a0f3a090: both values are now quoted (`--dep-libz "$(ENABLED_LIBZ)"`, `--dep-liboqs "$(ENABLED_LIBOQS)"`, Makefile.am:582-583), so an empty substitution becomes an explicit empty argument instead of swallowing the next flag.

scripts/csaf_validate.mjs
MarkAtwood: Fixed in a0f3a090: added scripts/test_csaf_validate.mjs, a runner self-test that pins the exit-code contract — usage error exits 2, a deliberately-invalid CSAF doc exits 1, and (when handed a generated doc) a valid one exits 0. Wired into the advisory.yml csaf-conformance job ahead of the real gate, so a validator-lib bump that breaks summarize() fails CI instead of silently turning the gate into a no-op.

scripts/gen-advisory
MarkAtwood: Fixed in a0f3a090: `test_unknown_state_hard_fails` asserts `_bucket_for` raises SystemExit on an out-of-vocabulary state, and `test_product_model_propagates_unknown_state_failure` pins the same contract through product_model with a bad overlay state. A future edit that silently defaulted the KeyError would now fail both.

scripts/gen-advisory
MarkAtwood: Fixed in a0f3a090: the per-version decision is now `v.get('status', a['default_status']) == 'affected'` — defaultStatus only applies to entries with no explicit status, with a comment in gen-advisory explaining exactly this trap. Two regression tests cover it: `test_default_status_affected_honours_explicit_unaffected` (explicit unaffected entry stays out of affected_ranges under defaultStatus=affected) and `test_default_status_affected_covers_unspecified_entries`.

MarkAtwood:
retest this please

MarkAtwood:
rerun tests

douzzer:
Retest this please
(lost logs)

MarkAtwood:
retest this please

SparkiDev (approved):

SparkiDev (dismissed):
weird control chars, hard tabs, CRs, trailing whitespace:
./scripts/sbom.am:116:»@test -n "$(PYTHON3)" || { \
./scripts/sbom.am:117:»    echo "ERROR: 'python3' not found in PATH. Cannot generate SBOM."; \
./scripts/sbom.am:118:»    exit 1; }
./scripts/sbom.am:119:»@test -n "$(PYSPDXTOOLS)" || { \
./scripts/sbom.am:120:»    echo "ERROR: 'pyspdxtools' not found (pip install spdx-tools)."; \
./scripts/sbom.am:121:»    exit 1; }
./scripts/sbom.am:122:»@test -f "$(SBOM_GEN)" || { \
./scripts/sbom.am:123:»    echo "ERROR: gen-sbom not found. Vendor scripts/gen-sbom, or re-run:"; \
./scripts/sbom.am:124:»    echo "       make sbom WOLFSSL_DIR=/path/to/wolfssl"; \
./scripts/sbom.am:125:»    exit 1; }
[...and more...]
sort: /tmp/workspace/PRB-multi-test-script/wolfssl/.wolfssl_known_macro_extras:1083: disorder: _ABI64

rlm2002 (dismissed):
Please resolve merge conflicts.

MarkAtwood:
merge conflicts are resolved

MarkAtwood:
conflicts resolved and there are no whitespace/tabs/ctrl/non-ASCII

doc/CRA.md
LinuxJedi, line 1: This file should probably be in the docs repo, not here.

doc/SBOM.md
LinuxJedi, line 1: Likewise this file

LinuxJedi (commented):
Only commenting rather than reviewing as I'm missing a whole middle part that I couldn't find in the comments of the PR.

My main question is: why is this all living here and not in other repos? I presume there is some kind of standard that dictates it, but I can't find it mentioned in the PR description.

MarkAtwood:
**"Why here and not other repos"**: it is in the other repos. This PR is the base implementation; the per-product targets are separate PRs:

- wolfSSL/wolfTPM#536 (merged)
- wolfSSL/wolfMQTT#557 (open)
- wolfSSL/wolfssh#1050 (open)
- wolfSSL/wolfsentry#85 (open)
- wolfSSL/wolfssl#10752 (open, cmake path)
- wolfBoot (landed directly in its Makefile)

`scripts/gen-sbom` lives here because every other product already carries wolfssl as a submodule or sibling checkout, so they invoke one generator instead of each maintaining a copy that drifts. wolfBoot does exactly this through `lib/wolfssl`. Those product PRs are mostly a `sbom.am` include plus configure glue, so the logic stays in one place.

As for why it's a build target at all rather than a standalone tool: wolfssl is compile-time configured, so what's actually in the shipped library depends on the `./configure` flags. The generator reads the build's own `options.h` and hashes a staged `make install`. Anything running outside the build tree would be guessing at the contents of the binary it's describing.

The standard is the EU Cyber Resilience Act (Regulation (EU) 2024/2847), which requires manufacturers to produce an SBOM covering at least the top-level dependencies of the product. Main obligations apply from December 2027. That's the driver.

**"Missing a whole middle part"**: true. It also contains `scripts/gen-advisory`, the CSAF/VEX overlay tooling, `.github/workflows/advisory.yml`, and the `advisories/` tree, which is the CRA vulnerability-reporting half.

All that stuff will be moving to wolfGlass, but this PR has to land first so we have SBOM generation that will work for customers.

MarkAtwood:
@LinuxJedi the background you're missing is written up in wolfSSL/wolfGlass, which is public: https://github.com/wolfSSL/wolfGlass/blob/master/docs/PLAN.md

Section 3 covers why the tools currently live in the wolfssl repo and why that's a problem. Section 4 covers what stays central versus what gets vendored per product. Section 19 phase 3 is the wolfssl migration specifically: wolfGlass becomes canonical for the generator and `sbom.am`, wolfssl keeps a vendored copy through a sync target, and there's no build-time dependency on wolfGlass.

`central/gen-advisory` is already in wolfGlass, so the move is underway rather than hypothetical.

SparkiDev (approved):

PR #10344, opened 2026-04-28
20260416-linuxkm-fips-rodata-canonify

douzzer:
`linuxkm/`: implement support for stabilization of `.rodata_wolfcrypt` segment in `WC_SYM_RELOC_TABLES` (FIPS) kernel module builds:

`linuxkm/`: refactor `wc_reloc_table_segments.reloc_tab_*` as `wc_reloc_table_segments.text_reloc_tab.*` (using the new `struct wc_reloc_table_fenceposts` and `WC_RELOC_TABLE_FENCEPOSTS_INITIALIZER`), and add `wc_reloc_table_segments.rodata_reloc_tab`.

`linuxkm/Makefile`: update the `GENERATE_RELOC_TAB` recipe to generate both `wc_linuxkm_pie_text_reloc_tab[]` and `wc_linuxkm_pie_rodata_reloc_tab[]`.

`linuxkm/linuxkm-fips-hash-wrapper.sh`: add handling for `wc_linuxkm_pie_rodata_reloc_tab`.

`linuxkm/linuxkm-fips-hash.c`: add handling for `rodata_reloc_tab.*`.

`linuxkm/linuxkm_memory.c`:
* refactor `find_reloc_tab_offset()` to be segment-agnostic and tolerate empty reloc tabs.
* refactor `wc_reloc_normalize_segment()`:
  * to be segment-agnostic,
  * identify the src segment dynamically,
  * return `BAD_FUNC_ARG` where previously returning literal `-1`,
  * use `seg_in_out_len` arg to accommodate size skew between input and output (not currently used), and
  * rename working vars for better mnemonicitude.
* update `wc_fips_generate_hash()` to
  * handle `seg_map->rodata_reloc_tab`,
  * use new calling convention for `wc_reloc_normalize_segment()`, and
  * add `wc_reloc_normalize_segment()` loop for `.rodata_wolfcrypt`.

`linuxkm/linuxkm_memory.h` and `linuxkm/linuxkm_wc_port.h`: rename `WOLFSSL_TEXT_SEGMENT_CANONICALIZER*` to `WOLFSSL_SEGMENT_CANONICALIZER*`, with backward-compat provisions.

`linuxkm/module_hooks.c`:
* add `wc_linuxkm_normalize_relocations_noresize()` backward-compat wrapper.
* `wolfssl_init()`: add `.rodata_wolfcrypt` relocation handling alongside existing `.text_wolfcrypt` handling, and update for new `wc_reloc_normalize_segment()` calling convention.
* add `seg_map.rodata_reloc_tab` initialization.
* update `wc_linuxkm_normalize_relocations()` to be segment-agnostic and use new `wc_reloc_normalize_segment()` calling convention.

misc ARM32-kernel-specific fixes:

`wolfcrypt/src/ge_operations.c` and `wolfssl/wolfcrypt/ge_operations.h`: when `ge_tobytes_nct` and `ge_tobytes` have identical definitions, map the former to the latter using a macro and omit the latter definition, to avoid problematic `R_ARM_THM_JUMP11` tail call.

`linuxkm/Kbuild`: define `NO_PIE_FLAG` to `1`, not empty, to satisfy gnu make criteria for `ifdef`.

misc kernel fixes:

`wolfcrypt/src/wc_lms_impl.c`: work around false-positive `-Wmaybe-uninitialized` in `wc_lms_treehash_update()`.

`linuxkm/lkcapi_*.c` and `linuxkm/module_hooks.c`: add missing linefeed characters in format args to `pr_*()`, for proper line flushing.

**Note, on targets with no relocations in .rodata_wolfcrypt, the new code is backward-compatible (generates the same HMAC value) so interoperates with old fips_test.c.**

tested with
```
wolfssl-multi-test.sh ...

linuxkm-all-cryptonly-pie-arm32-fips-dev-LKCAPI-build
quantum-safe-wolfssl-all-crypto-only-intelasm-fips-dev-linuxkm-next-insmod
linuxkm-6.12-cryptonly-intelasm-fips-v6-dyn-hash-LKCAPI-insmod
'.*cust-kernel-2.*'
```

Also tested on actual ARM32 hardware.

### DIFF
--- a/linuxkm/module_hooks.c
+++ b/linuxkm/module_hooks.c
@@ -621,7 +621,7 @@ static int wolfssl_init(void)
 #ifdef HAVE_FIPS
     /* The compiled-in verifycore must be the right length, else the module
      * geometry will change when the correct value is passed in, destabilizing
-     * wc_linuxkm_pie_reloc_tab.  It also must be the right length for the
+     * wc_linuxkm_pie_text_reloc_tab.  It also must be the right length for the
      * module-update-fips-hash recipe (in-place overwrite) to work, and for
      * updateFipsHash() (WOLFCRYPT_FIPS_CORE_DYNAMIC_HASH_VALUE) to be safe from
      * overruns.
@@ -1118,17 +1118,17 @@ MODULE_VERSION(LIBWOLFSSL_VERSION_STRING);
 
 #ifdef WC_SYM_RELOC_TABLES
 
-extern const struct wc_reloc_table_ent wc_linuxkm_pie_reloc_tab[];
-extern const unsigned int wc_linuxkm_pie_reloc_tab_length;
+extern const struct wc_reloc_table_ent wc_linuxkm_pie_text_reloc_tab[];
+extern const unsigned int wc_linuxkm_pie_text_reloc_tab_length;
 
 static const struct wc_reloc_table_segments seg_map = {
     .start = 0, .end = 0,
     .text_start = (size_t)(uintptr_t)__wc_text_start,
     .text_end = (size_t)(uintptr_t)__wc_text_end,
-    .reloc_tab_start = (size_t)(uintptr_t)wc_linuxkm_pie_reloc_tab,
-    .reloc_tab_end = 0,
-    .reloc_tab_len_start = (size_t)(uintptr_t)&wc_linuxkm_pie_reloc_tab_length,
-    .reloc_tab_len_end = 0,
+    .text_reloc_tab.start = (size_t)(uintptr_t)wc_linuxkm_pie_text_reloc_tab,
+    .text_reloc_tab.end = 0,
+    .text_reloc_tab.len_start = (size_t)(uintptr_t)&wc_linuxkm_pie_text_reloc_tab_length,
+    .text_reloc_tab.len_end = 0,
 #ifdef HAVE_FIPS
 #ifdef WC_USE_PIE_FENCEPOSTS_FOR_FIPS
     .fips_text_start = (size_t)(uintptr_t)__wc_text_start,
--- a/linuxkm/module_hooks.c
+++ b/linuxkm/module_hooks.c
@@ -1115,7 +1115,7 @@ static void wolfssl_exit(void)
         pr_err("ERROR: wc_RunAllCast_fips() failed at shutdown with return value %d\n", ret);
     }
     else
-        pr_info("wolfCrypt FIPS re-self-test succeeded at unload: all algorithms re-verified.");
+        pr_info("wolfCrypt FIPS re-self-test succeeded at unload: all algorithms re-verified.\n");
 #endif
 
     (void)libwolfssl_cleanup();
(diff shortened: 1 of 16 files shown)


linuxkm/module_hooks.c
```diff
@@ -711,39 +711,64 @@ static int wolfssl_init(void)
...
                 free(canon_buf);
                 return -ECANCELED;
             }
-            stabilized_text_hash = hash_span(canon_buf, canon_buf + progress, stabilized_text_hash);
+            stabilized_text_hash = hash_span(canon_buf, canon_buf + text_in_out_len, stabilized_text_hash);
             text_p += progress;
         }
 
+        /* note verifyCore is hashed along with the rest of .rodata_wolfcrypt. */
+        while (rodata_p < (const u8 *)__wc_rodata_end) {
```
douzzer, line 753: fixed.  good catch.

jackctj117 (dismissed):

jackctj117:
Ran a skoll review and these caught my eye:
### HIGH-1: Text canonicalization error is silently masked by rodata loop
- **File:** `linuxkm/linuxkm_memory.c:844-905`
- **Function:** `wc_fips_generate_hash`
- **Category:** High
- **Confidence:** When the text-canonicalization while-loop sets `ret` to a non-zero error and `break`s (either from `progress <= 0` or a failing `hmac_update`), control falls straight into the new rodata while-loop. Inside the rodata loop, `ret = hmac_update(hmac_ctx, buf, ...)` is unconditionally overwritten on every iteration. If the rodata loop subsequently completes without error, the prior text-loop error code is lost and `ret == 0`, so the function will go on to call `hmac_final()` and `Base16_Encode()` and return success — producing a hash that does not include the failed/partial text region. This is a regression introduced by adding the rodata loop; previously the text path was followed by a `goto out` (in the non-WC_SYM_RELOC_TABLES branch). Must short-circuit on text-loop failure before entering the rodata loop.

**Description:**         while (text_p < (const byte *)seg_map->fips_text_end) {

### LOW-3: stabilized_rodata_hash is computed but never reported
- **File:** `linuxkm/module_hooks.c:719,770,781-786`
- **Function:** `wolfssl_init`
- **Category:** High
- **Confidence:** `stabilized_rodata_hash` is declared, initialized, and updated each iteration of the new rodata canonicalization loop, but the subsequent `pr_info("wolfCrypt segment hashes (spans): ...")` only logs `stabilized_text_hash`. Either the rodata stabilized hash should be added to the log line (parallel to text), or it should not be computed. Given the rodata canonicalization was added specifically to validate the new rodata reloc tab, logging it seems the intent.

**Description:**         unsigned int stabilized_rodata_hash = 1;

douzzer:
Both items noted by Skoll are now fixed here, and in wolfssl/fips#386, and in `WCv5.2.4-KRNL-CHKIN-r7`.

douzzer:
Note `PRB-master-job` failure is unrelated to this PR -- new test broken on `master`:
```
configure: error: unrecognized options: --enable-falcon
```
because #10293 can't be merged yet.

lealem47 (approved):

PR #10351, opened 2026-04-29
CRYPTOCB_ONLY: add test infra + SHA256 + AES

rizlik:
This PR adds `wc_swdev`, a software CryptoCb device used by the test programs to exercise  `WOLF_CRYPTO_CB_ONLY_* ` builds without requiring external hardware. 
  
 
  
  `swdev`  mirrors the parent  libwolfSSL source set and build flags, while undefining the `WOLF_CRYPTO_CB_ONLY_*` gates internally so it
  can provide the software implementation behind the CryptoCb dispatch layer.

The PR also add `CRYPTO_CB_ONLY_AES` and `CRYPTO_CB_ONLY_SHA256 support`

The big caveat here is that the structures can't be really simplified in a general way, because they are used by the crypto callback layer as a mean to transfer state and keys from wolfcrypt to the cryptocb device.
Single crypto device implementation might choose to selectively remove unused field from struct if not needed, but the right solution would be to provide a feature agnostic layer between wolfcrypt and the cryptocb handler.

### DIFF
--- a/examples/client/client.c
+++ b/examples/client/client.c
@@ -51,6 +51,10 @@ static const char *wolfsentry_config_path = NULL;
 #include <wolfssl/test.h>
 #include <wolfssl/error-ssl.h>
 
+#ifdef WOLFSSL_SWDEV
+    #include "tests/swdev/swdev_loader.h"
+#endif
+
 #ifdef USE_FLAT_TEST_H
     #include "client.h"
 #else
@@ -5056,6 +5060,12 @@ exit:
         wolfSSL_Debugging_ON();
 #endif
         wolfSSL_Init();
+#ifdef WOLFSSL_SWDEV
+        if (wc_SwDev_Init() != 0) {
+            fprintf(stderr, "wc_SwDev_Init failed\n");
+            return EXIT_FAILURE;
+        }
+#endif
         ChangeToWolfRoot();
 
 #if !defined(NO_WOLFSSL_CLIENT) && !defined(NO_TLS)
@@ -5066,6 +5076,9 @@ exit:
 #endif
 #else
         fprintf(stderr, "Client not compiled in!\n");
+#endif
+#ifdef WOLFSSL_SWDEV
+        wc_SwDev_Cleanup();
 #endif
         wolfSSL_Cleanup();
 
--- a/examples/server/server.c
+++ b/examples/server/server.c
@@ -67,6 +67,10 @@ static const char *wolfsentry_config_path = NULL;
 #include <wolfssl/test.h>
 #include <wolfssl/error-ssl.h>
 
+#ifdef WOLFSSL_SWDEV
+    #include "tests/swdev/swdev_loader.h"
+#endif
+
 #ifdef USE_FLAT_TEST_H
     #include "server.h"
 #else
@@ -4256,6 +4260,12 @@ exit:
 #ifdef WC_RNG_SEED_CB
         wc_SetSeed_Cb(WC_GENERATE_SEED_DEFAULT);
 #endif
+#ifdef WOLFSSL_SWDEV
+        if (wc_SwDev_Init() != 0) {
+            fprintf(stderr, "wc_SwDev_Init failed\n");
+            return EXIT_FAILURE;
+        }
+#endif
         ChangeToWolfRoot();
 
 #if !defined(NO_WOLFSSL_SERVER) && !defined(NO_TLS)
@@ -4268,6 +4278,9 @@ exit:
         fprintf(stderr, "Server not compiled in!\n");
 #endif
 
+#ifdef WOLFSSL_SWDEV
+        wc_SwDev_Cleanup();
+#endif
         wolfSSL_Cleanup();
         FreeTcpReady(&ready);
 
--- a/tests/api.c
+++ b/tests/api.c
@@ -40799,7 +40799,11 @@ TEST_CASE testCases[] = {
 
 static void TestSetup(void)
 {
-/* Stub, for now. Add common test setup code here. */
+#ifdef WOLFSSL_SWDEV
+    /* Re-arm the swdev cryptocb registration in case the previous test
+     * tore down wolfCrypt and wiped gCryptoDev. */
+    (void)wc_SwDev_Init();
+#endif
 }
 
 static void TestCleanup(void)
@@ -40808,6 +40812,9 @@ static void TestCleanup(void)
     /* Clear any errors added to the error queue during the test run. */
     wolfSSL_ERR_clear_error();
 #endif /* OPENSSL_EXTRA || DEBUG_WOLFSSL_VERBOSE */
+#ifdef WOLFSSL_SWDEV
+    wc_SwDev_Cleanup();
+#endif
 }
 
 void ApiTest_StopOnFail(void)
--- a/tests/api.c
+++ b/tests/api.c
@@ -30609,6 +30609,7 @@ static int test_SSL_CIPHER_get_xxx(void)
 }
 
 #if defined(WOLF_CRYPTO_CB) && defined(HAVE_IO_TESTS_DEPENDENCIES) && \
+    !defined(WOLF_CRYPTO_CB_ONLY_SHA256) && \
     !defined(WOLF_CRYPTO_CB_ONLY_ECC) && !defined(WOLF_CRYPTO_CB_ONLY_RSA)
 
 static int load_pem_key_file_as_der(const char* privKeyFile, DerBuffer** pDer,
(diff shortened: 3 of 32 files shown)


tests/swdev/swdev_loader.c
```diff
@@ -0,0 +1,57 @@
...
+    int ret;
+
+    /* always re-register: cryptocb table is wiped by wolfCrypt_Cleanup */
+    ret = wc_CryptoCb_RegisterDevice(WC_SWDEV_ID, wc_SwDev_Callback, NULL);
+    if (ret != 0)
+        return ret;
+
+#ifdef WOLF_CRYPTO_CB_FIND
+    wc_CryptoCb_SetDeviceFindCb(wc_SwDev_FindCb);
+#endif
```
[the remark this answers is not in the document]
rizlik: refcounted Init/Clenaup

tests/swdev/swdev_loader.c
```diff
@@ -0,0 +1,57 @@
...
+    if (!swdev_registered)
+        return;
+
+#ifdef WOLF_CRYPTO_CB_FIND
+    wc_CryptoCb_SetDeviceFindCb(NULL);
+#endif
+
+    wc_CryptoCb_UnRegisterDevice(WC_SWDEV_ID);
+    swdev_registered = 0;
+}
```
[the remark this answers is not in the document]
rizlik: fixed by refcounting

examples/server/server.c
```diff
@@ -4258,6 +4262,12 @@ THREAD_RETURN WOLFSSL_THREAD server_test(void* args)
         wolfSSL_Init();
 #ifdef WC_RNG_SEED_CB
         wc_SetSeed_Cb(WC_GENERATE_SEED_DEFAULT);
+#endif
+#ifdef WOLFSSL_SWDEV
+        if (wc_SwDev_Init() != 0) {
+            fprintf(stderr, "wc_SwDev_Init failed\n");
+            return EXIT_FAILURE;
+        }
```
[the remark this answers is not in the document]
rizlik: fixed with refcounting

examples/client/client.c
```diff
@@ -5054,6 +5058,12 @@ THREAD_RETURN WOLFSSL_THREAD client_test(void* args)
         wolfSSL_Debugging_ON();
 #endif
         wolfSSL_Init();
+#ifdef WOLFSSL_SWDEV
+        if (wc_SwDev_Init() != 0) {
+            fprintf(stderr, "wc_SwDev_Init failed\n");
+            return EXIT_FAILURE;
+        }
+#endif
```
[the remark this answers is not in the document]
rizlik: fixed by refcounting

tests/api.c
```diff
@@ -37707,7 +37715,9 @@ TEST_CASE testCases[] = {
 
 static void TestSetup(void)
 {
-/* Stub, for now. Add common test setup code here. */
+#ifdef WOLFSSL_SWDEV
+    (void)wc_SwDev_Init();
+#endif
```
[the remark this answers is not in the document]
rizlik: removed

wolfcrypt/test/test.c
```diff
@@ -68062,14 +68085,16 @@ static wc_test_ret_t ecc_onlycb_test(myCryptoDevCtx *ctx)
     }
     XFREE(pub, HEAP_HINT, DYNAMIC_TYPE_TMP_BUFFER);
     XFREE(out, HEAP_HINT, DYNAMIC_TYPE_TMP_BUFFER);
-    #if !defined(WOLFCRYPT_ONLY) && defined(OPENSSL_EXTRA)
+    #if !defined(WOLFCRYPT_ONLY) && defined(OPENSSL_EXTRA) && \
+        defined(USE_CERT_BUFFERS_256)
     if (check) {
         FREE(check, HEAP_HINT, DYNAMIC_TYPE_TMP_BUFFER);
```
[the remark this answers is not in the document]
rizlik: not in this pr but fixed

danielinux (dismissed):

tests/swdev/swdev.c
```diff
@@ -0,0 +1,516 @@
+/* tests/swdev/swdev.c -- wc_swdev callback. */
```
dgarske: New files are missing standard license header.

tests/swdev/swdev.c
```diff
@@ -0,0 +1,516 @@
...
+
```
dgarske, line 23: What about HAVE_CONFIG_H include? Or does that not apply for this custom Makefile test area?

tests/swdev/swdev.c
```diff
@@ -0,0 +1,516 @@
...
+#include "swdev.h"
+
+#include <wolfssl/wolfcrypt/settings.h>
```
dgarske, line 26: And no include wolfssl/options.h?

dgarske (changes requested):
LGTM! Love the work here

tests/swdev/swdev.c
rizlik: Good catch: swdev do not support libraries compiled without autotools. 
I'll work to make it compatible with a main library compiled using a custom `user_settings.h` too.

rizlik:
CI failure do not look related

tests/swdev/swdev.c
```diff
@@ -0,0 +1,545 @@
+/* swdev.c
```
dgarske, line 1: Not seeing where the new files in tests/swdev/* are being added to include.am for "make dist"?

dgarske (changes requested):

tests/swdev/swdev.c
rizlik: they were only conditionally added, fixed. thanks

dgarske:
Jenkins retest this please

tests/swdev/swdev.c
dgarske, line 1: Please add a README.md (and to EXTRA_DIST) in tests/swdev. Here is a proposed README.md.

[README.md](https://github.com/user-attachments/files/27534955/README.md)

dgarske (changes requested):

bigbrett (dismissed):
Nice work. I'm good with this and glad you figured out a way to eek out the necessary test coverage!!

Not sure if it matters or if you want to ensure compat with these features, but currently this can't be used with `--enable-opensslextra` and `--enable-lms` due to linkage errors in `sha256.c`.

Also missing some macro protection in `sha256.c` for the intel speedups, but those are clearly incompatible features. 

Only thing I don't like is the "shadow" concept and the brittleness of needing to add fields to the copy when they are added in the header, but not sure of a great way around it. 

Also want to see docs for it but I see @dgarske requested as well

rizlik:
> Not sure if it matters or if you want to ensure compat with these features, but currently this can't be used with `--enable-opensslextra` and `--enable-lms` due to linkage errors in `sha256.c`.

Unfortunately that's true and the reason is that openssl exposes the raw SHA transformation function. This is not a supported operation of our crypto callback interface.
It's not easy to re-use the normal SHA update operation as building block to create the internal raw one, as we can't rely on the `sha256->digest`  to be used by the hardware device (it can save state internally).

I'll check if a more clean error is easy to add at compilation time.

LMS instead can be built without the raw transformation function with `WC_LMS_FULL_HASH`.
`--enable-lms --disable-sha224 -DWC_LMS_FULL_HASH` 

> 
> Also missing some macro protection in `sha256.c` for the intel speedups, but those are clearly incompatible features.
> 

Yes, I'll try to put up some error out for this case.

> Only thing I don't like is the "shadow" concept and the brittleness of needing to add fields to the copy when they are added in the header, but not sure of a great way around it.

There is no double definition of the struct, the copy in the shadow it's to avoid managing devId overwrite/restore and be more surgically touching only the fields needed.

The real problem here is that we can't remove software-only fields (or any fields) from the wolfcrypt without breaking swdev. This stem of cryptocb relying on the internal wolfcrypt structure to convey state and key.  This is a know limitation.

> 
> Also want to see docs for it but I see @dgarske requested as well

Coming!

ColtonWilley:
Jenkins retest this please

tests/swdev/README.md
```diff
@@ -0,0 +1,161 @@
+# wc_swdev: Software CryptoCb Device for Tests
```
dgarske, line 1: Did you decide if you want this directory and its files in the "make dist"? I don't see them in the include.am.

dgarske (approved):

tests/swdev/README.md
rizlik: they are here https://github.com/rizlik/wolfssl/blob/cbaf9c154a670493ed2aaaad48add93d076009e0/wolfcrypt/test/include.am#L67-L73

PR #10353, opened 2026-04-29
DTLS 1.3 client-only minimum: WOLFSSL_DTLS_ONLY + autoconf cascade

julek-wolfssl:
* configure.ac: --enable-dtls13 auto-enables --enable-dtls and TLS 1.3,
  with a targeted error if either is explicitly --disabled.
* src/internal.c, src/wolfio.c, wolfssl/wolfio.h: new WOLFSSL_DTLS_ONLY
  compile-time flag elides EmbedReceive / EmbedSend default callbacks
  and the GetSEQIncrement helper.
* wolfcrypt/src/aes.c: add HAVE_AES_DECRYPT to the inv_col_mul
  definition gate to match its only caller; without it the function is
  emitted dead under WOLFSSL_AES_DIRECT && NO_AES_DECRYPT and
  -Werror=unused-function fails the build.
* .github/workflows/os-check.yml: matrix entry for a minimal DTLS 1.3
  client-only build.


julek-wolfssl:
retest this please.

dgarske (approved):

PR #10367, opened 2026-04-30
fix: reorder wc_curve25519_make_pub/generic to input-before-output

MarkAtwood:
Refs: ZD-21733

## The Problem

Four raw-bytes Curve25519 functions had output parameters before input parameters:

```c
// Before (wrong):
int wc_curve25519_make_pub(int public_size, byte* pub,
                           int private_size, const byte* priv);
```

Every other wolfCrypt raw-bytes function uses input-before-output ordering:

```c
wc_curve25519_import_private(const byte* priv, word32 privSz, curve25519_key* key)
wc_ed25519_make_public(ed25519_key* key, unsigned char* pubKey, word32 pubKeySz)
wc_curve25519_make_priv(WC_RNG* rng, int keysize, byte* priv)
```

The same wrong ordering was shared by all four related raw-bytes functions: `wc_curve25519_make_pub`, `wc_curve25519_make_pub_blind`, `wc_curve25519_generic`, and `wc_curve25519_generic_blind`.

## Why It Was Not Caught

Both size parameters are `int` and both must equal `CURVE25519_KEYSIZE` (32). Transposing the arguments compiles without warning and passes the runtime size check. The clamping check (RFC 7748 §5) provides a partial accidental guard — an arbitrary 32-byte buffer will usually fail — but is not reliable. There was no test that deliberately passed arguments in the wrong order to verify the error path.

Additionally, the original author got the argument order wrong in six places in the test suite when the function was first introduced (2020-08-05), then had to push a follow-up commit the next day titled *"fix order of args to wc_curve25519_make_pub()"*. The signature was not corrected at that point; only the test was patched.

## How Fixed

Reordered parameters to private-before-public on all four functions:

```c
// After (correct):
int wc_curve25519_make_pub(int private_size, const byte* priv,
                           int public_size, byte* pub);
```

Updated all five internal call sites in `wolfcrypt/src/curve25519.c`, the test suite, and the Doxygen doc comments and code examples.

## Risk to External Callers

This is a breaking API change for any external caller that uses these four functions directly. Mitigating factors:

- These are low-level raw-bytes functions; the typical application uses `wc_curve25519_make_key()` instead
- All in-tree callers (5 sites, all in `wolfcrypt/src/curve25519.c`) have been updated atomically
- External callers with the wrong argument order were producing wrong output silently; the fix makes their code fail to compile rather than fail silently

If desired, a deprecated wrapper preserving the old name and signature can be added alongside.

### DIFF
--- a/doc/dox_comments/header_files/curve25519.h
+++ b/doc/dox_comments/header_files/curve25519.h
@@ -780,10 +780,10 @@ int wc_curve25519_size(curve25519_key* key);
     \return ECC_BAD_ARG_E If the key sizes are invalid
     \return BAD_FUNC_ARG If any input parameters are NULL
 
-    \param public_size Size of the public key buffer (must be 32)
-    \param pub Pointer to buffer to store the public key
     \param private_size Size of the private key (must be 32)
     \param priv Pointer to buffer containing the private key
+    \param public_size Size of the public key buffer (must be 32)
+    \param pub Pointer to buffer to store the public key
 
     _Example_
     \code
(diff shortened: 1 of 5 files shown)


dgarske:
Jenkins retest this please

dgarske (changes requested):
This is a breaking API change right?? If so we can't accept.

douzzer:
can't merge -- breaking API change.

PR #10380, opened 2026-05-01
Add crypto callbacks for LMS and XMSS

padelsbach:
# Description

Add struct fields and callbacks (`MakeKey`, `Sign`, `Verify` and `SigsLeft`) into existing LMS and XMSS code. Add `wc_LmsKey_InitId ` and `_InitLabel` for PKCS11 compat. Added unit tests.

# Testing

New unit tests

# Checklist

 - [x] added tests
 - [ ] updated/added doxygen


wolfcrypt/src/wc_xmss.c
```diff
@@ -36,6 +36,92 @@
...
+    if (key->params == NULL)
+        return BAD_FUNC_ARG;
+    needSz = (word32)key->params->n;
+    if (*hashSz < needSz)
+        return BAD_FUNC_ARG;
+
+    switch (key->params->hash) {
+    #ifdef WC_XMSS_SHA256
+        case WC_HASH_TYPE_SHA256:
+            ret = wc_Hash(WC_HASH_TYPE_SHA256, msg, msgSz, hash, needSz);
```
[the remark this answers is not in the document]
padelsbach: fixed

doc/LMS_XMSS_CryptoCb.md
Frauschi, line 1: I think we can remove this file. I only added it to the patch I originally sent you as a reference and easy starting point.

wolfssl/wolfcrypt/cryptocb.h
```diff
@@ -344,6 +350,41 @@ typedef struct wc_CryptoInfo {
...
+        #if defined(WOLFSSL_HAVE_LMS) || defined(WOLFSSL_HAVE_XMSS)
+            struct {
+                WC_RNG*     rng;
+                void*       key;
+                int         type; /* enum wc_PqcStatefulSignatureType */
+            } pqc_stateful_sig_kg;
+            struct {
+                /* Raw message. Backends following the PKCS#11 v3.2
+                 * CKM_HSS / CKM_XMSS convention of operating on a
+                 * pre-computed digest can call wc_LmsKey_HashMsg /
```
Frauschi: Nit, but I'm unsure if we want to have a specific PKCS#11 reference here.

wolfssl/wolfcrypt/cryptocb.h
```diff
@@ -711,6 +752,23 @@ WOLFSSL_LOCAL int wc_CryptoCb_Ed25519Verify(const byte* sig, word32 sigLen,
     const byte* context, byte contextLen);
 #endif /* HAVE_ED25519 */
 
+#if defined(WOLFSSL_HAVE_LMS) || defined(WOLFSSL_HAVE_XMSS)
+WOLFSSL_LOCAL int wc_CryptoCb_PqcStatefulSigGetDevId(int type, void* key);
+
+WOLFSSL_LOCAL int wc_CryptoCb_PqcStatefulSigKeyGen(int type, void* key,
+    WC_RNG* rng);
+/* The raw message is forwarded to the callback. Backends that follow the
+ * PKCS#11 v3.2 CKM_HSS / CKM_XMSS convention (digest input) can call
```
Frauschi: Same as above regarding the mention of PKCS#11 V3.2

wolfcrypt/src/wc_lms.c
```diff
@@ -36,6 +37,77 @@
...
+ * @param [in]      msgSz   Length of msg in bytes.
+ * @param [out]     hash    Buffer receiving the digest.
+ * @param [in,out]  hashSz  On entry, size of hash buffer. On success,
+ *                          the digest length.
+ * @return  0 on success.
+ * @return  BAD_FUNC_ARG when an argument is NULL or the buffer is too
+ *          small for the digest.
+ * @return  NOT_COMPILED_IN when the param set's hash family is disabled.
+ */
+int wc_LmsKey_HashMsg(const LmsKey* key, const byte* msg, word32 msgSz,
```
Frauschi: Should we move this method further down in the file to where the other public API methods are located? At the top of the file, only static internal functions are placed otherwise.

wolfcrypt/src/wc_lms.c
```diff
@@ -36,6 +37,77 @@
...
+
+    if ((key == NULL) || (msg == NULL) || (hash == NULL) || (hashSz == NULL))
+        return BAD_FUNC_ARG;
+    if (key->params == NULL)
+        return BAD_FUNC_ARG;
+    needSz = (word32)key->params->hash_len;
+    if (*hashSz < needSz)
+        return BAD_FUNC_ARG;
+
+    switch (key->params->lmsType & 0xF000) {
```
Frauschi: Use `LMS_HASH_MASK` instead of the magic `0xF000` literal.

wolfcrypt/src/wc_lms.c
```diff
@@ -36,6 +37,77 @@
...
+            ret = wc_Sha256Hash(msg, msgSz, full);
+            if (ret == 0)
+                XMEMCPY(hash, full, needSz);
+            break;
+        }
+    #ifdef WOLFSSL_LMS_SHAKE256
+        case LMS_SHAKE256:      /* SHAKE256 with 32-byte output */
+        case LMS_SHAKE256_192:  /* SHAKE256 with 24-byte output */ {
+            wc_Shake shake;
+            ret = wc_InitShake256(&shake, NULL, INVALID_DEVID);
```
Frauschi: You should also be able to use `wc_Shake256Hash()` here.

wolfcrypt/src/wc_lms.c
```diff
@@ -1313,6 +1487,27 @@ int wc_LmsKey_SigsLeft(LmsKey* key)
...
     /* NULL keys have no signatures remaining. */
     if (key != NULL) {
+#ifdef WOLF_CRYPTO_CB
+        if (key->devId != INVALID_DEVID) {
+            word32 sigsLeft = 0;
+            int cbRet = wc_CryptoCb_PqcStatefulSigSigsLeft(
+                WC_PQC_STATEFUL_SIG_TYPE_LMS, key, &sigsLeft);
+            if (cbRet == 0) {
+                /* Clamp to int range; callers treat 0 as "exhausted". */
+                return (sigsLeft > (word32)0x7FFFFFFF)
```
Frauschi: `wc_LmsKey_SigsLeft()` returns only `1` ("there are signatures left") or `0` ("all signatures exhausted") instead of the actual amount of signatures left. Hence, we have to follow this contract here, too, even when we receive an actual amount from the callback.

wolfcrypt/src/wc_lms.c
```diff
@@ -1505,6 +1734,9 @@ int wc_LmsKey_Verify(LmsKey* key, const byte* sig, word32 sigSz,
     if ((key == NULL) || (sig == NULL) || (msg == NULL)) {
         ret = BAD_FUNC_ARG;
     }
+    if ((ret == 0) && (msgSz <= 0)) {
```
Frauschi: Per LMS specification, a signature for an empty zero length message is not prohibited. We already prohibit a zero length for signing (as signing a zero length message makes no sense and a valuable signature would be lost), but I'm not sure if we really also have to prohibit that on the verification side. But I'm open for discussion on this.

wolfcrypt/src/wc_lms.c
```diff
@@ -36,6 +37,77 @@
     #include <wolfcrypt/src/misc.c>
 #endif
 
+#ifdef WOLF_CRYPTO_CB
+    #include <wolfssl/wolfcrypt/cryptocb.h>
+#endif
+
+/* Compute the digest of msg using the hash function dictated by the LMS
+ * parameter set. Crypto-callback / HSM backends that follow PKCS#11 v3.2
```
Frauschi: I think we should drop (or weaken) the PKCS#11 reference here and be more general in the comment on why this method is available.

wolfcrypt/src/wc_xmss.c
```diff
@@ -36,6 +37,99 @@
     #include <wolfcrypt/src/misc.c>
 #endif
 
+#ifdef WOLF_CRYPTO_CB
+    #include <wolfssl/wolfcrypt/cryptocb.h>
+#endif
+
+/* Compute the digest of msg using the hash function dictated by the XMSS
+ * parameter set. Crypto-callback / HSM backends that follow PKCS#11 v3.2
```
Frauschi: Similar to LMS, I think we should drop/weaken the PKCS#11 reference here and be more general about the reasoning of this function.

wolfcrypt/src/wc_xmss.c
```diff
@@ -36,6 +37,99 @@
...
+    needSz = (word32)key->params->n;
+    if (*hashSz < needSz)
+        return BAD_FUNC_ARG;
+
+    switch (key->params->hash) {
+    #ifdef WC_XMSS_SHA256
+        case WC_HASH_TYPE_SHA256: {
+            /* SHA2_*_192 variants set n=24, but wc_Hash rejects an output
+             * smaller than WC_SHA256_DIGEST_SIZE. Hash to a full buffer and
+             * copy the requested prefix, mirroring LMS_SHA256_192. */
```
Frauschi: Nit, but I'd drop the LMS reference here.

wolfcrypt/src/wc_xmss.c
```diff
@@ -36,6 +37,99 @@
...
+                    ret = wc_Shake128_Final(&shake, hash, needSz);
+                wc_Shake128_Free(&shake);
+            }
+            break;
+        }
+    #endif
+    #ifdef WC_XMSS_SHAKE256
+        case WC_HASH_TYPE_SHAKE256: {
+            wc_Shake shake;
+            ret = wc_InitShake256(&shake, NULL, INVALID_DEVID);
```
Frauschi: `wc_Shake256Hash()` simplifies this.

wolfcrypt/src/wc_xmss.c
```diff
@@ -36,6 +37,99 @@
...
+    #endif
+    #ifdef WC_XMSS_SHA512
+        case WC_HASH_TYPE_SHA512:
+            ret = wc_Hash(WC_HASH_TYPE_SHA512, msg, msgSz, hash, needSz);
+            break;
+    #endif
+    #ifdef WC_XMSS_SHAKE128
+        case WC_HASH_TYPE_SHAKE128: {
+            wc_Shake shake;
+            ret = wc_InitShake128(&shake, NULL, INVALID_DEVID);
```
Frauschi: `wc_Shake128Hash()` simplifies this.

wolfcrypt/src/wc_xmss.c
```diff
@@ -36,6 +37,99 @@
...
+ * @param [in]      msgSz   Length of msg in bytes.
+ * @param [out]     hash    Buffer receiving the digest.
+ * @param [in,out]  hashSz  On entry, size of hash buffer. On success,
+ *                          the digest length.
+ * @return  0 on success.
+ * @return  BAD_FUNC_ARG when an argument is NULL or the buffer is too
+ *          small for the digest.
+ * @return  NOT_COMPILED_IN when the param set's hash family is disabled.
+ */
+int wc_XmssKey_HashMsg(const XmssKey* key, const byte* msg, word32 msgSz,
```
Frauschi: similar to LMS, I would move this new method further below to the other public API methods. Here in the beginning are only local functions.

padelsbach:
jenkins retest this please

wolfcrypt/src/wc_xmss.c
```diff
@@ -1066,16 +1290,33 @@ int wc_XmssKey_MakeKey(XmssKey* key, WC_RNG* rng)
...
     if ((ret == 0) && (key->write_private_key == NULL)) {
         WOLFSSL_MSG("error: XmssKey write callback is not set");
         ret = BAD_FUNC_ARG;
     }
-    /* Ensure read/write callback context available. */
-    if ((ret == 0) && (key->context == NULL)) {
-        WOLFSSL_MSG("error: XmssKey context is not set");
-        ret = BAD_FUNC_ARG;
-    }
+    /* Callback context is opaque to wolfCrypt and may legitimately be NULL
```
Frauschi: I think this comment is unnecessary.

wolfcrypt/src/wc_xmss.c
```diff
@@ -1191,17 +1432,23 @@ int wc_XmssKey_Reload(XmssKey* key)
...
             (key->read_private_key == NULL))) {
         WOLFSSL_MSG("error: XmssKey write/read callbacks are not set");
         ret = BAD_FUNC_ARG;
     }
-    /* Ensure read and write callback context is available. */
-    if ((ret == 0) && (key->context == NULL)) {
-        WOLFSSL_MSG("error: XmssKey context is not set");
-        ret = BAD_FUNC_ARG;
-    }
+    /* Callback context is opaque; NULL is allowed. */
```
Frauschi: Unnecessary comment.

wolfcrypt/src/wc_xmss.c
```diff
@@ -1313,17 +1560,27 @@ int wc_XmssKey_Sign(XmssKey* key, byte* sig, word32* sigLen, const byte* msg,
...
             (key->read_private_key == NULL))) {
         WOLFSSL_MSG("error: XmssKey write/read callbacks are not set");
         ret = BAD_FUNC_ARG;
     }
-    /* Check read/write callback context available. */
-    if ((ret == 0) && (key->context == NULL)) {
-        WOLFSSL_MSG("error: XmssKey context is not set");
-        ret = BAD_FUNC_ARG;
-    }
+    /* Callback context is opaque; NULL is allowed. */
```
Frauschi: Unnecessary comment.

wolfcrypt/src/wc_xmss.c
```diff
@@ -1342,14 +1599,36 @@ int wc_XmssKey_Sign(XmssKey* key, byte* sig, word32* sigLen, const byte* msg,
...
+        return 0;
+
+#ifdef WOLF_CRYPTO_CB
+    if (key->devId != INVALID_DEVID) {
+        word32 sigsLeft = 0;
+        int cbRet = wc_CryptoCb_PqcStatefulSigSigsLeft(
+            WC_PQC_STATEFUL_SIG_TYPE_XMSS, key, &sigsLeft);
+        if (cbRet == 0) {
+            /* Clamp to int range; callers treat 0 as "exhausted". */
+            return (sigsLeft > (word32)0x7FFFFFFF)
```
Frauschi: Similar to LMS, we only return `1` or `0` here, not the actual amount of signatures left.

wolfcrypt/src/wc_xmss.c
```diff
@@ -1606,6 +1897,9 @@ int wc_XmssKey_Verify(XmssKey* key, const byte* sig, word32 sigLen,
     if ((key == NULL) || (sig == NULL) || (m == NULL)) {
         ret = BAD_FUNC_ARG;
     }
+    if ((ret == 0) && (mLen <= 0)) {
```
Frauschi: Similar to LMS, should we allow `mLen == 0` for verify? Not prohibited by the spec.

wolfcrypt/src/wc_lms.c
```diff
@@ -969,16 +1109,34 @@ int wc_LmsKey_MakeKey(LmsKey* key, WC_RNG* rng)
...
     if ((ret == 0) && (key->write_private_key == NULL)) {
         WOLFSSL_MSG("error: LmsKey write callback is not set");
         ret = BAD_FUNC_ARG;
     }
-    /* Check callback context set. */
-    if ((ret == 0) && (key->context == NULL)) {
-        WOLFSSL_MSG("error: LmsKey context is not set");
-        ret = BAD_FUNC_ARG;
-    }
+    /* Callback context is opaque to wolfCrypt and may legitimately be NULL
```
Frauschi: Comment is unnecessary I think.

wolfssl/wolfcrypt/wc_xmss.h
```diff
@@ -418,12 +446,20 @@ WOLFSSL_API void wc_XmssKey_Free(XmssKey* key);
...
+    void* heap, int devId);
 WOLFSSL_API int  wc_XmssKey_ExportPubRaw(const XmssKey* key, byte* out,
     word32* outLen);
 WOLFSSL_API int  wc_XmssKey_ImportPubRaw(XmssKey* key, const byte* in,
     word32 inLen);
 WOLFSSL_API int  wc_XmssKey_Verify(XmssKey* key, const byte* sig, word32 sigSz,
     const byte* msg, int msgSz);
+/* Compute the digest of a message with the hash function dictated by the
+ * XMSS parameter set. Useful for crypto-callback / HSM backends that follow
+ * the PKCS#11 v3.2 CKM_XMSS / CKM_XMSSMT convention of taking a
```
Frauschi: Again, I would remove/weaken the PKCS#11 reference here and be more general.

Frauschi (changes requested):
A lot of minor things, mainly as the WiP patch I sent you was in a very rough state tbh (sorry for that).

doc/LMS_XMSS_CryptoCb.md
padelsbach: removed

wolfcrypt/src/wc_lms.c
padelsbach: done

wolfcrypt/src/wc_lms.c
padelsbach: done

wolfcrypt/src/wc_lms.c
padelsbach: done

wolfcrypt/src/wc_lms.c
padelsbach: done

wolfcrypt/src/wc_lms.c
padelsbach: Not sure there is a real world use case for this, but I removed the check for spec compliance.

wolfcrypt/src/wc_xmss.c
padelsbach: done

wolfcrypt/src/wc_xmss.c
padelsbach: done

wolfcrypt/src/wc_xmss.c
padelsbach: done

wolfcrypt/src/wc_xmss.c
padelsbach: done

wolfcrypt/src/wc_xmss.c
padelsbach: removed

wolfcrypt/src/wc_xmss.c
padelsbach: removed

wolfcrypt/src/wc_xmss.c
padelsbach: removed

wolfcrypt/src/wc_xmss.c
padelsbach: done

wolfcrypt/src/wc_xmss.c
padelsbach: allowed len of 0

wolfcrypt/src/wc_lms.c
padelsbach: removed

wolfcrypt/src/wc_lms.c
padelsbach: IIUC, this method is available for the PKCS#11 case specifically, where it takes the pre-computed hash. So it's a convenience method for any crypto-cb users that implement that. It's not required for the wolfHSM use case, so we can just remove the method(s), but I believe they were part of your original patch.

Let me know what specifically you'd like to see here.

padelsbach:
jenkins retest this please

Frauschi:
Ok, I have updates regarding the PKCS#11 "pre-hash" thing. It turns out that this is actually an error in their specification, which is already corrected in the current working drafts for the next version. The data passed to the sign and verify operations is indeed the whole message, not any pre-hashed digest! 

After taking a more in-depth look into the algorithms, a simple pre-hash as proposed in the two helper methods of this PR also does not work and breaks spec compliance. Within both algorithms, the message is hashed together with the current state of the private key, which is not available in a CryptoCb / PKCS#11 setup on the host. Hence, this would have never worked properly anyway.

In total, this greatly simplifies our efforts, as the “normal” callbacks can directly be used for PKCS#11 as well. Hence, the two new helper methods for pre-hashing (`wc_LmsKey_HashMsg()` and wc_`XmssKey_HashMsg()`) can be removed completely. Furthermore, all the PKCS#11-related comments are unnecessary as a result.

wolfcrypt/src/wc_lms.c
Frauschi: Please see my new comment in the PR below, we can actually remove these methods completely.

padelsbach: Wow, good find. Removed.

wolfssl/wolfcrypt/wc_xmss.h
padelsbach: Removed

wolfcrypt/src/wc_lms.c
```diff
@@ -1371,13 +1496,42 @@ int wc_LmsKey_ExportPub(LmsKey* keyDst, const LmsKey* keySrc)
...
+ * wc_LmsKey_ExportPub_ex.
+ *
+ * @param [out] keyDst  LMS key to copy into.
+ * @param [in]  keySrc  LMS key to copy.
+ * @return  0 on success.
+ * @return  BAD_FUNC_ARG when keyDst or keySrc is NULL.
+ */
+int wc_LmsKey_ExportPub(LmsKey* keyDst, const LmsKey* keySrc)
+{
+    return wc_LmsKey_ExportPub_ex(keyDst, keySrc, NULL, INVALID_DEVID);
```
Frauschi: Should we pass `keySrc->heap` here like we do for XMSS?

wolfcrypt/src/wc_xmss.c
Frauschi: Still `<=`

padelsbach: removed

wolfcrypt/src/wc_lms.c
padelsbach: done

Frauschi (approved):

dgarske (approved):
Very nice!

PR #10387, opened 2026-05-04
Fix flaky tcp bind on Windows test runs

Frauschi:
Windows test code pre-picked a random port via `GetRandomPort()` (returning a value in [49152, 65535]) before calling `bind()`, with no check that the port was free and no retry on collision. Under load this occasionally collided with an already-bound port and aborted the test with "tcp bind failed", producing intermittent Jenkins failures (e.g. PRB windows-test-v2 in the OCSP responder test).

The Unix path already does the right thing: bind to port 0 (OS-assigned ephemeral) and read the port back via `getsockname()`. The same primitives exist in Winsock 1.1, so drop the `USE_WINDOWS_API` guard around the `getsockname` block in `tcp_listen()`/`udp_accept()` and remove the per-caller `GetRandomPort()` workarounds in the OCSP responder, server example, and the api.c / test_ossl_bio.c test sites. `socklen_t` is already typedef'd as int on Windows in test.h.

`GetRandomPort()` itself is left in place since it is a static inline in a shipped public test header.

### DIFF
--- a/examples/ocsp_responder/ocsp_responder.c
+++ b/examples/ocsp_responder/ocsp_responder.c
@@ -943,13 +943,6 @@ THREAD_RETURN WOLFSSL_THREAD ocsp_responder_test(void* args)
         }
     }
 
-#ifdef USE_WINDOWS_API
-    if (opts.port == 0) {
-        /* Generate random port for testing */
-        opts.port = GetRandomPort();
-    }
-#endif /* USE_WINDOWS_API */
-
     /* Create and listen on server socket */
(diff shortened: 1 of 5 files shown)


julek-wolfssl (approved):
Checks out
> For TCP/IP, if the port is specified as zero, the service provider assigns a unique port to the application from the dynamic client port range. On Windows Vista and later, the dynamic client port range is a value between 49152 and 65535.

https://learn.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-bind

dgarske (approved):

PR #10394, opened 2026-05-04
Add RSA/DH SP non-blocking support for C/Small 2048/3072/4096

dgarske:
## Summary

Extends `WOLFSSL_SP_NONBLOCK` (already covering ECC and Curve25519) to RSA and Diffie-Hellman so a bare-metal loop or async-driven TLS handshake never blocks for more than ~1 ms / 100 MHz on a single big-int op. Targets the `WOLFSSL_SP_SMALL` C32/C64 backend; assembly variants are unchanged.

The chunked state machine yields per Montgomery op and per inner bit-extract step (mirrors TFM's `fp_exptmod_nb` design). A 2048-bit RSA private op yields ~10240 times = `2 * 2048 + ~13`.

## What's added

- `sp_RsaPublic_<bits>_nb` and `sp_RsaPrivate_<bits>_nb` — `D`-only path (`RSA_LOW_MEM` / `SP_RSA_PRIVATE_EXP_D`); CRT private is unsupported in non-block mode (configure-time `#error`).
- `sp_DhExp_<bits>_nb` and `sp_ModExp_<bits>_nb` — byte-buffer base for nb-friendly TLS use.
- New `WC_DH_NONBLOCK` API: `DhNb` context + `wc_DhSetNonBlock(DhKey*, DhNb*)`, parallel to `WC_RSA_NONBLOCK` / `WC_ECC_NONBLOCK`.
- `wc_RsaFunctionAsync` / `wc_DhAgree_Sync` dispatch to the SP non-block state machine in their compute path. `wc_AsyncSimulate` already translates per-yield `FP_WOULDBLOCK` into `WC_PENDING_E`, so the TLS state machine drives completion via the standard async-event loop.
- TLS layer per-SSL nb-context allocation in `AllocKey` / `FreeKey` and the TLS 1.3 keyShare paths, gated on `ssl->devId != INVALID_DEVID` (mirrors the ECC / Curve25519 hooks).

## Configure

```
./configure --enable-rsa=nonblock --enable-dh=nonblock \
            --enable-ecc=nonblock --enable-curve25519=nonblock \
            --enable-sp=yes,nonblock \
            CPPFLAGS="-DWOLFSSL_PUBLIC_MP -DWOLFSSL_DEBUG_NONBLOCK -DRSA_LOW_MEM"
```

`--enable-rsa=nonblock` and `--enable-dh=nonblock` auto-enable `--enable-asynccrypt --enable-asynccrypt-sw` and define `RSA_LOW_MEM`. Build-time `#error` checks in `rsa.h` / `dh.h` enforce the required `WOLFSSL_SP_SMALL`, `WOLFSSL_SP_NO_MALLOC`, and `!WOLFSSL_SP_FAST_MODEXP` companion flags.

## Tests

- `wolfcrypt/test/testwolfcrypt`: `RSA non-block sign: 10249 times`, verify 94 times, inline verify 94 times; `DH non-block agree: 1 times` (drives via `wc_AsyncWait`); both `RSA test passed!` and `DH test passed!`.
- `make check`: 5 PASS / 3 SKIP / 0 FAIL.
- TLS 1.2 handshake (`TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384`) verified end-to-end with `examples/server` + `examples/client`; server's RSA private signature is driven through ~10437 SP non-block yields by the SW-shim translation in `wc_AsyncSimulate`.

## CI

`.github/workflows/os-check.yml` extends the existing nonblock matrix entry to add `--enable-rsa=nonblock --enable-dh=nonblock`, plus a companion entry that forces `-DSP_WORD_SIZE=32` so both `sp_c32.c` and `sp_c64.c` are exercised on every push.


SparkiDev (dismissed):

SparkiDev:
Jenkins: retest this please
Aborted sanitize

SparkiDev (approved):

PR #10400, opened 2026-05-05
Fix Dilithium signing when WC_DILITHIUM_CACHE_MATRIX_A is enabled

### LINKED ISSUE #10383: [Bug]: Memory leak in Dilithium signing when WC_DILITHIUM_CACHE_MATRIX_A is enabled
### Contact Details

arjunaajalahla100@gmail.com

### Version

main branch

### Description

Found a memory leak in the Dilithium/ML-DSA implementation when the `WC_DILITHIUM_CACHE_MATRIX_A` compile option is enabled and `WC_DILITHIUM_FIXED_ARRAY` is disabled.

In the signing function around line 8321 of `wolfcrypt/src/dilithium.c`, there's this code:

```c
if ((ret == 0) && (key->a == NULL)) {
    a = (sword32*)XMALLOC(params->aSz, key->heap, DYNAMIC_TYPE_DILITHIUM);  // Line 8321
    if (a == NULL) {
        ret = MEMORY_E;
    }
}
#endif
if (ret == 0) {
    a = key->a;  // Line 8328 - overwrites the freshly allocated pointer
}
```

The problem is that it allocates memory into the local variable `a` instead of storing it in `key->a`. Then on line 8328, `a` gets overwritten with `key->a` (which is still NULL at this point), so the allocated memory is lost and never freed.

The correct pattern is used elsewhere in the same file (e.g., line 7765-7766 in `dilithium_make_key`):

```c
if (key->a == NULL) {
    key->a = (sword32*)XMALLOC(params->aSz, key->heap, DYNAMIC_TYPE_DILITHIUM);
    if (key->a == NULL) {
        ret = MEMORY_E;
    }
}
```

## Fix

Change line 8321 from:
```c
a = (sword32*)XMALLOC(params->aSz, key->heap, DYNAMIC_TYPE_DILITHIUM);
```

To:
```c
key->a = (sword32*)XMALLOC(params->aSz, key->heap, DYNAMIC_TYPE_DILITHIUM);
```

And remove line 8328 (`a = key->a;`) since it's no longer needed.

### Reproduction steps

1. Configure wolfSSL with `WC_DILITHIUM_CACHE_MATRIX_A` defined and `WC_DILITHIUM_FIXED_ARRAY` undefined
2. Build and run the Dilithium signing operation
3. The memory leak will occur each time the signing function is called when `key->a` is NULL

### Relevant log output

```shell

```

embhorn:
# Description

`dilithium_sign_with_seed_mu()` allocated matrix A into a local variable a, then immediately overwrote it with a = key->a

Fixes #10383 

# Testing

Added `dilithium_sign_cache_alloc_test`

`WC_DILITHIUM_CACHE_MATRIX_A` is enabled in Debian RPM tests

# Checklist

 - [x] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/wolfcrypt/src/dilithium.c
+++ b/wolfcrypt/src/dilithium.c
@@ -8333,13 +8333,13 @@ static int dilithium_sign_with_seed_mu(dilithium_key* key,
 #ifdef WC_DILITHIUM_CACHE_MATRIX_A
 #ifndef WC_DILITHIUM_FIXED_ARRAY
     if ((ret == 0) && (key->a == NULL)) {
-        key->a = (sword32*)XMALLOC(params->aSz, key->heap,
+        key->a = (sword32*)XMALLOC((size_t)params->aSz, key->heap,
             DYNAMIC_TYPE_DILITHIUM);
         if (key->a == NULL) {
             ret = MEMORY_E;
         }
         else {
-            XMEMSET(key->a, 0, params->aSz);
+            XMEMSET(key->a, 0, (size_t)params->aSz);
         }
     }
 #endif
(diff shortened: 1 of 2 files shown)


Frauschi (changes requested):
As ML-DSA is now added to Wconversion testing, you will need type casts on the `XMALLOC()` and `XMEMSET()` calls for the `params->aSz` argument.

Frauschi (approved):

dgarske (approved):

PR #10401, opened 2026-05-05
plumb caller heap into Cmac before cryptocb fires

night1rider:
`_InitCmac_` common `XMEMSET`s the Cmac struct then immediately invokes the cryptocb, so the heap arg passed to `wc_InitCmac_*` was lost before any hardware callback could read it. Stash heap in `cmac->aes.heap` (or a new `NO_AES` only `cmac->heap` fallback) right after the memset.


night1rider:
Jenkins retest this please

SparkiDev (approved):

PR #10402, opened 2026-05-05
Rust wrapper: ensure memory safety for C RNG struct

holtrop-wolfssl:
# Description

This PR includes several Fenrir fixes and some new functionality I found missing during my integration tests with boringtun.

See the commits for individual descriptions of each change.

# Testing

Unit/CI tests.

# Checklist

 - [X] added tests
 - [ ] updated/added doxygen

# Change Log

```
Rust wrapper: gate RSA doc tests that use set_shared_rng() on alloc feature
Rust wrapper: check slice lengths in ECC::import_unsigned
Rust wrapper: test more empty slices in test_ecc
Rust wrapper: use Rc instead of Arc for RNG references
Rust wrapper: fix RSA::set_rng() doc test to actually call set_rng()
Rust wrapper: drop mut from rng in ecc doc tests
Rust wrapper: avoid double wc_ecc_init_ex() call
Rust wrapper: test for null-terminated C-style strings in ECC import_raw APIs
Rust wrapper: document wolfSSL version requirement
Rust wrapper: remove outdated set_rng safety contract comments
Rust wrapper: add blake2_digest module
Rust wrapper: add blake2_mac module
Rust wrapper: implement Clone for HMAC types
Rust wrapper: add Aes192Ccm and Aes192Gcm
Rust wrapper: store pointer to C ECC key struct instead of instance
Rust wrapper: ensure memory safety for C RNG struct
```

### DIFF
--- a/wrapper/rust/wolfssl-wolfcrypt/src/ecc.rs
+++ b/wrapper/rust/wolfssl-wolfcrypt/src/ecc.rs
@@ -59,11 +59,11 @@ impl ECCPoint {
     /// {
     /// use wolfssl_wolfcrypt::random::RNG;
     /// use wolfssl_wolfcrypt::ecc::{ECC,ECCPoint};
-    /// let mut rng = RNG::new().expect("Failed to create RNG");
+    /// let rng = RNG::new().expect("Failed to create RNG");
     /// let curve_id = ECC::SECP256R1;
     /// let curve_size = ECC::get_curve_size_from_id(curve_id).expect("Error with get_curve_size_from_id()");
-    /// let mut ecc = ECC::generate_ex(curve_size, &mut rng, curve_id, None, None).expect("Error with generate()");
-    /// let ecc_point = ecc.make_pub_to_point(Some(&mut rng), None).expect("Error with make_pub_to_point()");
+    /// let mut ecc = ECC::generate_ex(curve_size, &rng, curve_id, None, None).expect("Error with generate()");
+    /// let ecc_point = ecc.make_pub_to_point(Some(&rng), None).expect("Error with make_pub_to_point()");
     /// let mut der = [0u8; 128];
     /// let size = ecc_point.export_der(&mut der, curve_id).expect("Error with export_der()");
     /// ECCPoint::import_der(&der[0..size], curve_id, None).expect("Error with import_der()");
@@ -117,11 +117,11 @@ impl ECCPoint {
     /// {
     /// use wolfssl_wolfcrypt::random::RNG;
     /// use wolfssl_wolfcrypt::ecc::{ECC,ECCPoint};
-    /// let mut rng = RNG::new().expect("Failed to create RNG");
+    /// let rng = RNG::new().expect("Failed to create RNG");
     /// let curve_id = ECC::SECP256R1;
     /// let curve_size = ECC::get_curve_size_from_id(curve_id).expect("Error with get_curve_size_from_id()");
-    /// let mut ecc = ECC::generate_ex(curve_size, &mut rng, curve_id, None, None).expect("Error with generate()");
-    /// let ecc_point = ecc.make_pub_to_point(Some(&mut rng), None).expect("Error with make_pub_to_point()");
+    /// let mut ecc = ECC::generate_ex(curve_size, &rng, curve_id, None, None).expect("Error with generate()");
+    /// let ecc_point = ecc.make_pub_to_point(Some(&rng), None).expect("Error with make_pub_to_point()");
     /// let mut der = [0u8; 128];
     /// let size = ecc_point.export_der(&mut der, curve_id).expect("Error with export_der()");
     /// ECCPoint::import_der_ex(&der[0..size], curve_id, 0, None).expect("Error with import_der_ex()");
@@ -172,11 +172,11 @@ impl ECCPoint {
     /// {
     /// use wolfssl_wolfcrypt::random::RNG;
     /// use wolfssl_wolfcrypt::ecc::{ECC,ECCPoint};
-    /// let mut rng = RNG::new().expect("Failed to create RNG");
+    /// let rng = RNG::new().expect("Failed to create RNG");
     /// let curve_id = ECC::SECP256R1;
     /// let curve_size = ECC::get_curve_size_from_id(curve_id).expect("Error with get_curve_size_from_id()");
-    /// let mut ecc = ECC::generate_ex(curve_size, &mut rng, curve_id, None, None).expect("Error with generate()");
-    /// let ecc_point = ecc.make_pub_to_point(Some(&mut rng), None).expect("Error with make_pub_to_point()");
+    /// let mut ecc = ECC::generate_ex(curve_size, &rng, curve_id, None, None).expect("Error with generate()");
+    /// let ecc_point = ecc.make_pub_to_point(Some(&rng), None).expect("Error with make_pub_to_point()");
     /// let mut der = [0u8; 128];
     /// let size = ecc_point.export_der(&mut der, curve_id).expect("Error with export_der()");
     /// assert!(size > 0 && size <= der.len());
@@ -219,11 +219,11 @@ impl ECCPoint {
     /// {
     /// use wolfssl_wolfcrypt::random::RNG;
     /// use wolfssl_wolfcrypt::ecc::{ECC,ECCPoint};
-    /// let mut rng = RNG::new().expect("Failed to create RNG");
+    /// let rng = RNG::new().expect("Failed to create RNG");
     /// let curve_id = ECC::SECP256R1;
     /// let curve_size = ECC::get_curve_size_from_id(curve_id).expect("Error with get_curve_size_from_id()");
-    /// let mut ecc = ECC::generate_ex(curve_size, &mut rng, curve_id, None, None).expect("Error with generate()");
-    /// let ecc_point = ecc.make_pub_to_point(Some(&mut rng), None).expect("Error with make_pub_to_point()");
+    /// let mut ecc = ECC::generate_ex(curve_size, &rng, curve_id, None, None).expect("Error with generate()");
+    /// let ecc_point = ecc.make_pub_to_point(Some(&rng), None).expect("Error with make_pub_to_point()");
     /// let mut der = [0u8; 128];
     /// let size = ecc_point.export_der_compressed(&mut der, curve_id).expect("Error with export_der_compressed()");
     /// }
@@ -255,9 +255,9 @@ impl ECCPoint {
     /// {
     /// use wolfssl_wolfcrypt::random::RNG;
     /// use wolfssl_wolfcrypt::ecc::ECC;
-    /// let mut rng = RNG::new().expect("Failed to create RNG");
-    /// let mut ecc = ECC::generate(32, &mut rng, None, None).expect("Error with generate()");
-    /// let mut ecc_point = ecc.make_pub_to_point(Some(&mut rng), None).expect("Error with make_pub_to_point()");
+    /// let rng = RNG::new().expect("Failed to create RNG");
+    /// let mut ecc = ECC::generate(32, &rng, None, None).expect("Error with generate()");
+    /// let mut ecc_point = ecc.make_pub_to_point(Some(&rng), None).expect("Error with make_pub_to_point()");
     /// ecc_point.forcezero();
     /// }
     /// ```
@@ -438,8 +438,8 @@ impl ECC {
     /// {
     /// use wolfssl_wolfcrypt::random::RNG;
     /// use wolfssl_wolfcrypt::ecc::ECC;
-    /// let mut rng = RNG::new().expect("Failed to create RNG");
-    /// let mut ecc = ECC::generate(32, &mut rng, None, None).expect("Error with generate()");
+    /// let rng = RNG::new().expect("Failed to create RNG");
+    /// let mut ecc = ECC::generate(32, &rng, None, None).expect("Error with generate()");
     /// ecc.check().expect("Error with check()");
     /// }
     /// ```
@@ -486,10 +486,10 @@ impl ECC {
     /// {
     /// use wolfssl_wolfcrypt::random::RNG;
     /// use wolfssl_wolfcrypt::ecc::ECC;
-    /// let mut rng = RNG::new().expect("Failed to create RNG");
+    /// let rng = RNG::new().expect("Failed to create RNG");
     /// let curve_id = ECC::SECP256R1;
     /// let curve_size = ECC::get_curve_size_from_id(curve_id).expect("Error with get_curve_size_from_id()");
-    /// let mut ecc = ECC::generate_ex(curve_size, &mut rng, curve_id, None, None).expect("Error with generate_ex()");
+    /// let mut ecc = ECC::generate_ex(curve_size, &rng, curve_id, None, None).expect("Error with generate_ex()");
     /// ecc.check().expect("Error with check()");
     /// }
     /// ```
@@ -537,10 +537,10 @@ impl ECC {
     /// {
     /// use wolfssl_wolfcrypt::random::RNG;
     /// use wolfssl_wolfcrypt::ecc::ECC;
-    /// let mut rng = RNG::new().expect("Failed to create RNG");
+    /// let rng = RNG::new().expect("Failed to create RNG");
     /// let curve_id = ECC::SECP256R1;
     /// let curve_size = ECC::get_curve_size_from_id(curve_id).expect("Error with get_curve_size_from_id()");
-    /// let mut ecc = ECC::generate_ex2(curve_size, &mut rng, curve_id, ECC::FLAG_COFACTOR, None, None).expect("Error with generate_ex2()");
+    /// let mut ecc = ECC::generate_ex2(curve_size, &rng, curve_id, ECC::FLAG_COFACTOR, None, None).expect("Error with generate_ex2()");
     /// ecc.check().expect("Error with check()");
     /// }
     /// ```
@@ -582,10 +582,10 @@ impl ECC {
     /// {
     /// use wolfssl_wolfcrypt::random::RNG;
     /// use wolfssl_wolfcrypt::ecc::ECC;
-    /// let mut rng = RNG::new().expect("Failed to create RNG");
+    /// let rng = RNG::new().expect("Failed to create RNG");
     /// let curve_id = ECC::SECP256R1;
     /// let curve_size = ECC::get_curve_size_from_id(curve_id).expect("Error with get_curve_size_from_id()");
-    /// let mut ecc = ECC::generate_ex(curve_size, &mut rng, curve_id, None, None).expect("Error with generate()");
+    /// let mut ecc = ECC::generate_ex(curve_size, &rng, curve_id, None, None).expect("Error with generate()");
     /// ecc.check().expect("Error with check()");
     /// }
     /// ```
@@ -619,7 +619,7 @@ impl ECC {
     /// use wolfssl_wolfcrypt::random::RNG;
     /// use wolfssl_wolfcrypt::ecc::ECC;
     /// use std::fs;
-    /// let mut rng = RNG::new().expect("Failed to create RNG");
+    /// let rng = RNG::new().expect("Failed to create RNG");
     /// let key_path = "../../../certs/ecc-client-key.der";
     /// let der: Vec<u8> = fs::read(key_path).expect("Error reading key file");
     /// let mut ecc = ECC::import_der(&der, None, None).expect("Error with import_der()");
@@ -667,13 +667,13 @@ impl ECC {
     /// use wolfssl_wolfcrypt::random::RNG;
     /// use wolfssl_wolfcrypt::ecc::ECC;
     /// use std::fs;
-    /// let mut rng = RNG::new().expect("Failed to create RNG");
+    /// let rng = RNG::new().expect("Failed to create RNG");
     /// let key_path = "../../../certs/ecc-client-key.der";
     /// let der: Vec<u8> = fs::read(key_path).expect("Error reading key file");
     /// let mut ecc = ECC::import_der(&der, None, None).expect("Error with import_der()");
     /// let hash = [0x42u8; 32];
     /// let mut signature = [0u8; 128];
-    /// let signature_length = ecc.sign_hash(&hash, &mut signature, &mut rng).expect("Error with sign_hash()");
+    /// let signature_length = ecc.sign_hash(&hash, &mut signature, &rng).expect("Error with sign_hash()");
     /// assert!(signature_length > 0 && signature_length <= signature.len());
     /// let signature = &mut signature[0..signature_length];
     /// let key_path = "../../../certs/ecc-client-keyPub.der";
@@ -726,11 +726,11 @@ impl ECC {
     /// {
     /// use wolfssl_wolfcrypt::random::RNG;
     /// use wolfssl_wolfcrypt::ecc::ECC;
-    /// let mut rng = RNG::new().expect("Failed to create RNG");
-    /// let mut ecc = ECC::generate(32, &mut rng, None, None).expect("Error with generate()");
+    /// let rng = RNG::new().expect("Failed to create RNG");
+    /// let mut ecc = ECC::generate(32, &rng, None, None).expect("Error with generate()");
     /// let hash = [0x42u8; 32];
     /// let mut signature = [0u8; 128];
-    /// let signature_length = ecc.sign_hash(&hash, &mut signature, &mut rng).expect("Error with sign_hash()");
+    /// let signature_length = ecc.sign_hash(&hash, &mut signature, &rng).expect("Error with sign_hash()");
     /// let signature = &signature[0..signature_length];
     /// let mut d = [0u8; 32];
     /// let d_size = ecc.export_private(&mut d).expect("Error with export_private()");
@@ -791,13 +791,13 @@ impl ECC {
     /// {
     /// use wolfssl_wolfcrypt::random::RNG;
     /// use wolfssl_wolfcrypt::ecc::ECC;
-    /// let mut rng = RNG::new().expect("Failed to create RNG");
+    /// let rng = RNG::new().expect("Failed to create RNG");
     /// let curve_id = ECC::SECP256R1;
     /// let curve_size = ECC::get_curve_size_from_id(curve_id).expect("Error with get_curve_size_from_id()");
-    /// let mut ecc = ECC::generate_ex(curve_size, &mut rng, curve_id, None, None).expect("Error with generate_ex()");
+    /// let mut ecc = ECC::generate_ex(curve_size, &rng, curve_id, None, None).expect("Error with generate_ex()");
     /// let hash = [0x42u8; 32];
     /// let mut signature = [0u8; 128];
-    /// let signature_length = ecc.sign_hash(&hash, &mut signature, &mut rng).expect("Error with sign_hash()");
+    /// let signature_length = ecc.sign_hash(&hash, &mut signature, &rng).expect("Error with sign_hash()");
     /// let signature = &signature[0..signature_length];
     /// let mut d = [0u8; 32];
     /// let d_size = ecc.export_private(&mut d).expect("Error with export_private()");
@@ -969,10 +969,10 @@ impl ECC {
     /// {
     /// use wolfssl_wolfcrypt::random::RNG;
     /// use wolfssl_wolfcrypt::ecc::ECC;
-    /// let mut rng = RNG::new().expect("Failed to create RNG");
+    /// let rng = RNG::new().expect("Failed to create RNG");
     /// let curve_id = ECC::SECP256R1;
     /// let curve_size = ECC::get_curve_size_from_id(curve_id).expect("Error with get_curve_size_from_id()");
-    /// let mut ecc = ECC::generate_ex(curve_size, &mut rng, curve_id, None, None).expect("Error with generate()");
+    /// let mut ecc = ECC::generate_ex(curve_size, &rng, curve_id, None, None).expect("Error with generate()");
     /// let mut qx = [0u8; 32];
     /// let mut qx_len = 0u32;
     /// let mut qy = [0u8; 32];
@@ -1026,8 +1026,8 @@ impl ECC {
     /// {
     /// use wolfssl_wolfcrypt::random::RNG;
     /// use wolfssl_wolfcrypt::ecc::ECC;
-    /// let mut rng = RNG::new().expect("Failed to create RNG");
-    /// let mut ecc = ECC::generate(32, &mut rng, None, None).expect("Error with generate()");
+    /// let rng = RNG::new().expect("Failed to create RNG");
+    /// let mut ecc = ECC::generate(32, &rng, None, None).expect("Error with generate()");
     /// let mut x963 = [0u8; 128];
     /// let x963_size = ecc.export_x963(&mut x963).expect("Error with export_x963()");
     /// let x963 = &x963[0..x963_size];
@@ -1080,10 +1080,10 @@ impl ECC {
     /// {
     /// use wolfssl_wolfcrypt::random::RNG;
     /// use wolfssl_wolfcrypt::ecc::ECC;
-    /// let mut rng = RNG::new().expect("Failed to create RNG");
+    /// let rng = RNG::new().expect("Failed to create RNG");
     /// let curve_id = ECC::SECP256R1;
     /// let curve_size = ECC::get_curve_size_from_id(curve_id).expect("Error with get_curve_size_from_id()");
-    /// let mut ecc = ECC::generate_ex(curve_size, &mut rng, curve_id, None, None).expect("Error with generate_ex()");
+    /// let mut ecc = ECC::generate_ex(curve_size, &rng, curve_id, None, None).expect("Error with generate_ex()");
     /// let mut x963 = [0u8; 128];
     /// let x963_size = ecc.export_x963(&mut x963).expect("Error with export_x963()");
     /// let x963 = &x963[0..x963_size];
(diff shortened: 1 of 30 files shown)


wrapper/rust/wolfssl-wolfcrypt/src/ecc.rs
```diff
@@ -392,6 +396,20 @@ impl ECC {
...
+    fn new_ecc_key(heap: *mut core::ffi::c_void, dev_id: i32) -> Result<*mut sys::ecc_key, i32> {
+        let key = unsafe { sys::wc_ecc_key_new(heap) };
+        if key.is_null() {
+            return Err(sys::wolfCrypt_ErrorCodes_MEMORY_E);
+        }
+        let rc = unsafe { sys::wc_ecc_init_ex(key, heap, dev_id) };
+        if rc != 0 {
+            unsafe { sys::wc_ecc_key_free(key); }
+            return Err(rc);
+        }
```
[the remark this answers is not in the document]
holtrop-wolfssl: The second wc_ecc_init_ex() call is needed to pass dev_id. The wc_ecc_init_ex() called by wc_ecc_key_new() does not pass the user's dev_id value.

wrapper/rust/wolfssl-wolfcrypt/src/ecc.rs
```diff
@@ -1749,8 +1683,14 @@ impl ECC {
...
-    ///   instance. The `RNG` struct should not be moved in memory after
-    ///   calling this method.
+    ///   instance.
+    ///
+    /// # Safety contract
+    ///
+    /// The caller must ensure that the `RNG` instance is not dropped before
+    /// this `ECC` instance. The `ECC` struct holds an internal pointer to the
+    /// `RNG`'s underlying `WC_RNG` context, and dropping the `RNG` first
+    /// would result in a dangling pointer.
```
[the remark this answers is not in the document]
holtrop-wolfssl: Removed outdated comment.

wrapper/rust/wolfssl-wolfcrypt/src/rsa.rs
```diff
@@ -1165,8 +1185,14 @@ impl RSA {
...
-    ///   instance. The `RNG` struct should not be moved in memory after
-    ///   calling this method.
+    ///   instance.
+    ///
+    /// # Safety contract
+    ///
+    /// The caller must ensure that the `RNG` instance is not dropped before
+    /// this `RSA` instance. The `RSA` struct holds an internal pointer to the
+    /// `RNG`'s underlying `WC_RNG` context, and dropping the `RNG` first
+    /// would result in a dangling pointer.
```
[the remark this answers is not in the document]
holtrop-wolfssl: Removed outdated comment.

wrapper/rust/wolfssl-wolfcrypt/src/rsa.rs
```diff
@@ -1177,42 +1196,73 @@ impl RSA {
...
     /// let key_path = "../../../certs/client-key.der";
     /// let der: Vec<u8> = fs::read(key_path).expect("Error reading key file");
     /// let mut rsa = RSA::new_from_der(&der).expect("Error with new_from_der()");
-    /// rsa.set_rng(&mut rng).expect("Error with set_rng()");
+    /// rsa.set_shared_rng(std::sync::Arc::clone(&rng)).expect("Error with set_shared_rng()");
     /// let mut plain_out: [u8; 512] = [0; 512];
     /// let dec_len = rsa.private_decrypt(&enc[0..enc_len], &mut plain_out).expect("Error with private_decrypt()");
     /// assert!(dec_len as usize == plain.len());
     /// assert_eq!(plain_out[0..dec_len], *plain);
     /// }
```
MarkAtwood, line 1222: All doc examples for `set_rng` and `set_shared_rng` now use `Arc` and are gated `#[cfg(all(random, feature = "alloc"))]`. A `no_alloc` caller has no example showing how to use the owned `set_rng` path (which is their only option). Please add a doc example for `set_rng` that does not require `alloc`.

wrapper/rust/wolfssl-wolfcrypt/tests/test_ecc.rs
```diff
@@ -365,6 +367,34 @@ fn test_ecc_import() {
...
+    assert!(ECC::import_raw(qx, qy_no_nul, d, curve_name, None, None).is_err());
+    assert!(ECC::import_raw(qx, qy, d_no_nul, curve_name, None, None).is_err());
+    assert!(ECC::import_raw(qx, qy, d, curve_name_no_nul, None, None).is_err());
+    assert!(ECC::import_raw(empty, qy, d, curve_name, None, None).is_err());
+    assert!(ECC::import_raw(qx, qy, d, empty, None, None).is_err());
+
+    assert!(ECC::import_raw_ex(qx_no_nul, qy, d, ECC::SECP256R1, None, None).is_err());
+    assert!(ECC::import_raw_ex(qx, qy_no_nul, d, ECC::SECP256R1, None, None).is_err());
+    assert!(ECC::import_raw_ex(qx, qy, d_no_nul, ECC::SECP256R1, None, None).is_err());
+    assert!(ECC::import_raw_ex(qx, qy, empty, ECC::SECP256R1, None, None).is_err());
```
MarkAtwood, line 428: `import_raw` is tested with `empty` for both `qx` (line 389) and the curve-name slot (line 390), but `import_raw_ex` is only tested with `empty` for `d` (line 395). Missing: `import_raw_ex(empty, qy, d, ...)` and `import_raw_ex(qx, empty, d, ...)` — both presumably call the same underlying C code so they would pass, but the coverage is asymmetric with the `import_raw` checks above.

wrapper/rust/wolfssl-wolfcrypt/src/ecc.rs
```diff
@@ -421,24 +439,17 @@ impl ECC {
...
-    pub fn generate(size: i32, rng: &mut RNG, heap: Option<*mut core::ffi::c_void>, dev_id: Option<i32>) -> Result<Self, i32> {
-        let mut wc_ecc_key: MaybeUninit<sys::ecc_key> = MaybeUninit::uninit();
-        let heap = match heap {
-            Some(heap) => heap,
-            None => core::ptr::null_mut(),
-        };
-        let dev_id = match dev_id {
-            Some(dev_id) => dev_id,
-            None => sys::INVALID_DEVID,
+    pub fn generate(size: i32, rng: &RNG, heap: Option<*mut core::ffi::c_void>, dev_id: Option<i32>) -> Result<Self, i32> {
```
MarkAtwood, line 447: Many doc examples in `ecc.rs` still declare `let mut rng = RNG::new()` and pass `&mut rng` to `generate()`, `generate_ex()`, `generate_ex2()`, `sign_hash()`, `make_pub()`, `make_pub_to_point()`, etc. Those methods now take `&RNG` (shared reference), so `mut` on the binding and `&mut` at the call site are both unnecessary and will produce compiler warnings when doc-tests are run. Please drop the `mut` qualifier (e.g. `let rng = RNG::new()`) and change `&mut rng` to `&rng` throughout the untouched doc examples.

wrapper/rust/wolfssl-wolfcrypt/src/ecc.rs
```diff
@@ -1765,22 +1706,51 @@ impl ECC {
...
             return Err(rc);
         }
+        self.rng = Some(RngHandle::Owned(rng));
         Ok(())
     }
 
+    /// Bind a shared `RNG` to this key. Available when the `alloc` feature
+    /// is enabled.
+    #[cfg(all(random, feature = "alloc"))]
+    pub fn set_shared_rng(&mut self, rng: alloc::sync::Arc<RNG>) -> Result<(), i32> {
```
MarkAtwood: The doc comment for `set_shared_rng` doesn't mention the `!Sync` constraint on `RNG`. Because `RNG: !Sync`, `Arc<RNG>` is also `!Sync`, so it cannot be shared between threads without a `Mutex`. Callers expecting to use the same RNG from multiple threads will need `Arc<Mutex<RNG>>`, but that type cannot be passed to `set_shared_rng`. The intended use case (sharing one RNG among multiple keys on the **same** thread) should be stated in the doc comment so callers aren't surprised when they try to move an `Arc<RNG>` to another thread.

wrapper/rust/wolfssl-wolfcrypt/src/ecc.rs
```diff
@@ -392,6 +396,20 @@ impl ECC {
     pub const FLAG_COFACTOR: i32 = sys::WC_ECC_FLAG_COFACTOR as i32;
     pub const FLAG_DEC_SIGN: i32 = sys::WC_ECC_FLAG_DEC_SIGN as i32;
 
+    /// Allocate and initialize a new `sys::ecc_key` on the C heap.
```
gasbytes, line 399: from my understanding I see that wc_ecc_key_new already runs wc_ecc_init_ex(key, heap, INVALID_DEVID) internally (https://github.com/wolfSSL/wolfssl/blob/master/wolfcrypt/src/ecc.c#L6182), I think the options is to either drop the wc_ecc_key_new drop and use xmalloc  wc_ecc_init_ex directly, or skip the second init

gasbytes (changes requested):

wrapper/rust/wolfssl-wolfcrypt/src/ecc.rs
holtrop-wolfssl: Thanks. Fixed!

wrapper/rust/wolfssl-wolfcrypt/src/ecc.rs
holtrop-wolfssl: Fixed

wrapper/rust/wolfssl-wolfcrypt/src/rsa.rs
holtrop-wolfssl: Fixed

wrapper/rust/wolfssl-wolfcrypt/src/ecc.rs
holtrop-wolfssl: Changed from Arc to Rc to reinforce that WC_RNG is not thread-safe internally.

wrapper/rust/wolfssl-wolfcrypt/tests/test_ecc.rs
holtrop-wolfssl: Added

gasbytes (dismissed):

holtrop-wolfssl:
retest this please (build removed)

holtrop-wolfssl:
retest this please (Finished: FAILURE)

holtrop-wolfssl:
retest this please (fips repo failures)

holtrop-wolfssl:
retest this please (fips repo failures)

holtrop-wolfssl:
retest this please (Build 'wolfSSL/PRB-fips-repo-and-harness-test-v3-part2' failed with result: FAILURE)

MarkAtwood (changes requested):
Reviewed the seven logical chunks (RNG memory-safety refactor, ECC heap allocation + double-init fix + empty-slice tests, AES-192 additions, BLAKE2 modules, HMAC `Clone`, doc/version changes) with FFI checks against the wolfCrypt C sources.

**Headline:** no P0 (UB / wrong-crypto) regressions introduced. The three architectural changes — `RngHandle::Owned(RNG) | Shared(Rc<RNG>)`, C-heap allocation of `ecc_key`, manual `wc_HmacCopy`-based `Clone` — are sound and address real previous issues. Drop ordering, double-`set_rng` ordering, and `MaybeUninit`/`assume_init` flow all check out.

That said, there is one concrete correctness blocker, one doc-test breakage, and one version-bump item I'd want addressed before merge.

---

## Blocker

### `ECC::import_unsigned` has no length validation — empty slice triggers OOB read in C
`wrapper/rust/wolfssl-wolfcrypt/src/ecc.rs:987-1004`

```rust
pub fn import_unsigned(qx: &[u8], qy: &[u8], d: &[u8], curve_id: i32, ...) -> Result<Self, i32> {
    ...
    let rc = unsafe {
        sys::wc_ecc_import_unsigned(ecc.wc_ecc_key, qx.as_ptr(),
            qy.as_ptr(), d.as_ptr(), curve_id)
    };
```

The C path `wc_ecc_import_unsigned → _ecc_import_raw_private` (`wolfcrypt/src/ecc.c:11898-11904`) calls:

```c
err = mp_read_unsigned_bin(key->pubkey.x, (const byte*)qx, (word32)key->dp->size);
```

It reads `key->dp->size` bytes (32 for P-256, 48 for P-384, 66 for P-521) **regardless of the Rust slice length**. The `sp_read_unsigned_bin` guard at `wolfcrypt/src/sp_int.c:18177` only catches `(in == NULL) && (inSz > 0)`. An empty Rust slice's `as_ptr()` returns a non-null dangling pointer, so the guard is bypassed and the C side reads invalid memory → UB at the FFI boundary.

This is pre-existing, but commit `c6d14218d` ("test more empty slices in test_ecc") explicitly added empty-slice rejection tests for `import_raw` and `import_raw_ex` — which *are* validated in Rust via the trailing-NUL + `is_empty` check at `ecc.rs:866-867,923-924` — while conspicuously skipping `import_unsigned`. The natural reviewer expectation given the commit's title is that `import_unsigned` got the same treatment.

Same audit applies to `priv_buf` in `import_private_key` / `import_private_key_ex` (`ecc.rs:746-766, 813-833`).

**Suggested fix:** validate each slice length against `wc_ecc_get_curve_size_from_id(curve_id)` in the Rust wrapper and return `BAD_FUNC_ARG` on short input. Add a regression test mirroring the `c6d14218d` pattern.

---

## Should fix in this PR

### Five RSA doc tests miss `feature = "alloc"` in their cfg gates
`wrapper/rust/wolfssl-wolfcrypt/src/rsa.rs`

These doc-test blocks use `std::rc::Rc::new(...)` + `rsa.set_shared_rng(...)`:

| Lines | Gate |
|---|---|
| `30-57` (module-level example) | `#[cfg(random)]` |
| `871-901` (`pss_sign`) | `#[cfg(all(random, rsa_pss))]` |
| `934-963` (`pss_check_padding`) | `#[cfg(all(random, rsa_pss, rsa_const_api))]` |
| `1000-1029` (`pss_verify`) | same |
| `1071-1100` (`pss_verify_check`) | same |

But `set_shared_rng` is `#[cfg(all(random, feature = "alloc"))]`. `cargo test --doc --features random` (without `--features alloc`) will fail to compile these.

The correctly-gated pattern is visible at `rsa.rs:148-175` and `rsa.rs:1284-1310`, both of which include `feature = "alloc"`. Either add `feature = "alloc"` to the five gates above, or rewrite to use owned `RNG` via `set_rng` (the approach `4ace46f2e` used for the `set_rng` doc test at `:1197-1223`).

### Bump crate version and document the API breaks
`wrapper/rust/wolfssl-wolfcrypt/Cargo.toml`, `README.md` / PR description

Per the [Cargo SemVer reference](https://doc.rust-lang.org/cargo/reference/semver.html), both of these are MAJOR-bump changes:

1. **Public function signature change**: `set_rng(&mut self, &mut RNG)` → `set_rng(&mut self, RNG)` on `ECC`, `RSA`, `Curve25519Key`. Callers passing `&mut rng` get a hard compile error.
2. **Cargo feature rename**: `std = []` → `alloc = []`. Anyone with `features = ["std"]` gets a hard build failure. (Listed verbatim under MAJOR in the Cargo reference.)

Given the crate is narrow-audience right now, `1.3.0` with an explicit CHANGELOG / PR-description note is pragmatic and honest. Bumping to `2.0.0` would be the strict-SemVer answer. Silently going to `1.3.0` without flagging the breaks would be the worst option.

Suggested CHANGELOG bullets:

```
## 1.3.0 — 2026-05-NN

### Breaking changes
- `ECC::set_rng`, `RSA::set_rng`, `Curve25519Key::set_rng` now take owned
  `RNG` instead of `&mut RNG`. The previous borrow-based API was unsound:
  the C library cached the raw `*WC_RNG` past the borrow's lifetime. New
  `set_shared_rng(Rc<RNG>)` is also available behind `feature = "alloc"`.
- Cargo feature `std` has been renamed to `alloc`. Update any
  `features = ["std"]` in your Cargo.toml.
- `&mut RNG` parameters on `generate*` / `make_pub*` / `sign_*` are now `&RNG`.
  (Source-compatible via reborrow coercion.)

### Added
- `AesGcm` / `AesCcm` variants for 192-bit keys.
- `blake2_digest` module (typed BLAKE2b-{256,384,512} and BLAKE2s-{128,192,256}).
- `blake2_mac` module (typed BLAKE2b/2s MACs).
- `Clone` impl for HMAC types (deep-copy via `wc_HmacCopy`).
- `set_shared_rng(Rc<RNG>)` for sharing an RNG across multiple consumers.

### Fixed
- Avoid double `wc_ecc_init_ex` call (was leaking async device context under
  WOLFSSL_ASYNC_CRYPT && WC_ASYNC_ENABLE_ECC).
- ECC key is now C-heap-allocated to remain valid across Rust moves under
  non-default math configurations.
```

Also: minimum wolfSSL version is correctly documented as 5.9.0 in the README. `wc_HmacCopy` is the binding constraint (introduced in v5.9.0); `wc_rng_new_ex` is older (v5.7.0).

---

## Advisory (track for follow-up, not blocking)

### Inline vs. heap inconsistency between ECC and RSA/DH/Curve25519

Commit `113b6b7ec`'s rationale ("internal pointers breaking if Rust moves the ECC struct with some build configurations") implies certain wolfCrypt configs put self-pointers in C structs. ECC was migrated to a C-heap pointer; `RSA.wc_rsakey`, `DH.wc_dhkey`, and `Curve25519Key.wc_key` remain inline. In the default bindgen config these are move-safe, but if the same configs that affected ECC (non-SP integer math, `WOLFSSL_ASYNC_CRYPT`, FIPS, hardware-CB) also affect the other key types, they carry the same latent UB this PR just fixed for ECC. Worth a follow-up audit and/or a `// Safety:` comment in each consumer struct recording the move-safety assumption.

---

## Nits (not blocking)

- `src/ecc.rs:405` — the comment "`wc_ecc_key_new()` always initializes the key with INVALID_DEVID" is wrong on CAAM builds (`ecc.c:6174-6180` initializes with `WOLFSSL_CAAM_DEVID`). The subsequent `(*key).devId = dev_id;` under `#[cfg(wolf_crypto_cb)]` will then clobber the CAAM default if the caller passed `dev_id: None`.
- `src/random.rs:67-70` — the `unsafe impl Send for RNG` safety comment claims the pointer is non-null but it isn't asserted. A `debug_assert!(!wc_rng.is_null())` after the success branch at `:137` and `:197` would document and enforce the invariant.
- `src/random.rs:60-62` — DRBG state is mutated through `&self`. Sound because the state lives behind an opaque `*mut WC_RNG`, but `UnsafeCell<*mut WC_RNG>` would make the interior-mutability intent explicit and defeat any future compiler improvement that might re-derive `Sync` or hoist reads.
- `src/ecc.rs:866-867, 923-924` — the null-terminator check accepts buffers with embedded NULs (only the last byte is checked). `XSTRLEN` / `mp_read_radix` on the C side stop at the first NUL, so embedded NULs cause silent truncation. Accepting `&CStr` would be cleaner.
- `src/rsa.rs:1208-1212` — the `set_rng` doc test (fixed in `4ace46f2e`) does call `set_rng`, but the immediately-following `public_encrypt` uses a separate `enc_rng` argument, so the bound RNG isn't exercised here. Only the second `set_rng` at `:1217` + `private_decrypt` at `:1219` actually depends on the bound RNG (under `WC_RSA_BLINDING`). The first `set_rng` call is dead in this example.
- `src/ecdsa.rs:177-185` and many `tests/test_*.rs` sites — stale `let mut rng = ...` + `&mut rng` reborrows left from before the `&RNG` migration. Compiles via reborrow coercion, just noisy. Same cleanup as commit `1a441c910`, just not applied everywhere.
- `tests/test_hmac_mac.rs:43-62` — the Clone test would be stronger with divergent post-clone updates (both branches feed `suffix`, so the test verifies "clone preserves state" + "clone doesn't poison original" but not "diverged updates produce different MACs"). The existing test catches the most likely shallow-copy bug, so this is a nice-to-have.

---

## What the PR does right (worth calling out)

- **RNG ownership refactor:** the old `&'a mut WC_RNG` borrow was unsound (C cached the raw pointer past the borrow lifetime). The new design heap-allocates `WC_RNG` via `wc_rng_new_ex` (pointer stable across moves), correctly picks `Rc` over `Arc` (WC_RNG is not internally synchronized), splits `set_rng(RNG)` from `set_shared_rng(Rc<RNG>)`, and gets Drop ordering right — consumer free runs before RNG free, so `wc_FreeRsaKey`/`wc_ecc_key_free` can still safely read `key->rng`. Double-`set_rng` ordering is also correct (C pointer replaced before old `RngHandle` drops).
- **ECC heap allocation:** switching to `*mut sys::ecc_key` via `wc_ecc_key_new` is the right fix for move-unsafety under configs where `mp_int.dp` points back into the same struct. `Pin` isn't needed because the Rust handle is a pointer to a never-moving C-heap object.
- **Double-init fix:** real bug — `wc_ecc_key_new` already calls `wc_ecc_init_ex` internally; the explicit second call `XMEMSET`s the struct, blowing away heap context and leaking async device contexts in `WOLFSSL_ASYNC_CRYPT && WC_ASYNC_ENABLE_ECC` builds.
- **HMAC `Clone`:** uses a *manual* impl delegating to `wc_HmacCopy` instead of bitwise `#[derive(Clone)]` — the latter would be unsound for hash configs with internal pointers. `MaybeUninit::assume_init` is correctly placed after the error check.

holtrop-wolfssl:
> Reviewed the seven logical chunks (RNG memory-safety refactor, ECC heap allocation + double-init fix + empty-slice tests, AES-192 additions, BLAKE2 modules, HMAC `Clone`, doc/version changes) with FFI checks against the wolfCrypt C sources.
> 
> **Headline:** no P0 (UB / wrong-crypto) regressions introduced. The three architectural changes — `RngHandle::Owned(RNG) | Shared(Rc<RNG>)`, C-heap allocation of `ecc_key`, manual `wc_HmacCopy`-based `Clone` — are sound and address real previous issues. Drop ordering, double-`set_rng` ordering, and `MaybeUninit`/`assume_init` flow all check out.
> 
> That said, there is one concrete correctness blocker, one doc-test breakage, and one version-bump item I'd want addressed before merge.
> 
> ## Blocker
> ### `ECC::import_unsigned` has no length validation — empty slice triggers OOB read in C
> `wrapper/rust/wolfssl-wolfcrypt/src/ecc.rs:987-1004`
> 
> ```rust
> pub fn import_unsigned(qx: &[u8], qy: &[u8], d: &[u8], curve_id: i32, ...) -> Result<Self, i32> {
>     ...
>     let rc = unsafe {
>         sys::wc_ecc_import_unsigned(ecc.wc_ecc_key, qx.as_ptr(),
>             qy.as_ptr(), d.as_ptr(), curve_id)
>     };
> ```
> 
> The C path `wc_ecc_import_unsigned → _ecc_import_raw_private` (`wolfcrypt/src/ecc.c:11898-11904`) calls:
> 
> ```c
> err = mp_read_unsigned_bin(key->pubkey.x, (const byte*)qx, (word32)key->dp->size);
> ```
> 
> It reads `key->dp->size` bytes (32 for P-256, 48 for P-384, 66 for P-521) **regardless of the Rust slice length**. The `sp_read_unsigned_bin` guard at `wolfcrypt/src/sp_int.c:18177` only catches `(in == NULL) && (inSz > 0)`. An empty Rust slice's `as_ptr()` returns a non-null dangling pointer, so the guard is bypassed and the C side reads invalid memory → UB at the FFI boundary.
> 
> This is pre-existing, but commit `c6d14218d` ("test more empty slices in test_ecc") explicitly added empty-slice rejection tests for `import_raw` and `import_raw_ex` — which _are_ validated in Rust via the trailing-NUL + `is_empty` check at `ecc.rs:866-867,923-924` — while conspicuously skipping `import_unsigned`. The natural reviewer expectation given the commit's title is that `import_unsigned` got the same treatment.
> 
> Same audit applies to `priv_buf` in `import_private_key` / `import_private_key_ex` (`ecc.rs:746-766, 813-833`).
> 
> **Suggested fix:** validate each slice length against `wc_ecc_get_curve_size_from_id(curve_id)` in the Rust wrapper and return `BAD_FUNC_ARG` on short input. Add a regression test mirroring the `c6d14218d` pattern.
> 
> ## Should fix in this PR
> ### Five RSA doc tests miss `feature = "alloc"` in their cfg gates
> `wrapper/rust/wolfssl-wolfcrypt/src/rsa.rs`
> 
> These doc-test blocks use `std::rc::Rc::new(...)` + `rsa.set_shared_rng(...)`:
> 
> Lines	Gate
> `30-57` (module-level example)	`#[cfg(random)]`
> `871-901` (`pss_sign`)	`#[cfg(all(random, rsa_pss))]`
> `934-963` (`pss_check_padding`)	`#[cfg(all(random, rsa_pss, rsa_const_api))]`
> `1000-1029` (`pss_verify`)	same
> `1071-1100` (`pss_verify_check`)	same
> But `set_shared_rng` is `#[cfg(all(random, feature = "alloc"))]`. `cargo test --doc --features random` (without `--features alloc`) will fail to compile these.
> 
> The correctly-gated pattern is visible at `rsa.rs:148-175` and `rsa.rs:1284-1310`, both of which include `feature = "alloc"`. Either add `feature = "alloc"` to the five gates above, or rewrite to use owned `RNG` via `set_rng` (the approach `4ace46f2e` used for the `set_rng` doc test at `:1197-1223`).
> 
> ### Bump crate version and document the API breaks
> `wrapper/rust/wolfssl-wolfcrypt/Cargo.toml`, `README.md` / PR description
> 
> Per the [Cargo SemVer reference](https://doc.rust-lang.org/cargo/reference/semver.html), both of these are MAJOR-bump changes:
> 
> 1. **Public function signature change**: `set_rng(&mut self, &mut RNG)` → `set_rng(&mut self, RNG)` on `ECC`, `RSA`, `Curve25519Key`. Callers passing `&mut rng` get a hard compile error.
> 2. **Cargo feature rename**: `std = []` → `alloc = []`. Anyone with `features = ["std"]` gets a hard build failure. (Listed verbatim under MAJOR in the Cargo reference.)
> 
> Given the crate is narrow-audience right now, `1.3.0` with an explicit CHANGELOG / PR-description note is pragmatic and honest. Bumping to `2.0.0` would be the strict-SemVer answer. Silently going to `1.3.0` without flagging the breaks would be the worst option.
> 
> Suggested CHANGELOG bullets:
> 
> ```
> ## 1.3.0 — 2026-05-NN
> 
> ### Breaking changes
> - `ECC::set_rng`, `RSA::set_rng`, `Curve25519Key::set_rng` now take owned
>   `RNG` instead of `&mut RNG`. The previous borrow-based API was unsound:
>   the C library cached the raw `*WC_RNG` past the borrow's lifetime. New
>   `set_shared_rng(Rc<RNG>)` is also available behind `feature = "alloc"`.
> - Cargo feature `std` has been renamed to `alloc`. Update any
>   `features = ["std"]` in your Cargo.toml.
> - `&mut RNG` parameters on `generate*` / `make_pub*` / `sign_*` are now `&RNG`.
>   (Source-compatible via reborrow coercion.)
> 
> ### Added
> - `AesGcm` / `AesCcm` variants for 192-bit keys.
> - `blake2_digest` module (typed BLAKE2b-{256,384,512} and BLAKE2s-{128,192,256}).
> - `blake2_mac` module (typed BLAKE2b/2s MACs).
> - `Clone` impl for HMAC types (deep-copy via `wc_HmacCopy`).
> - `set_shared_rng(Rc<RNG>)` for sharing an RNG across multiple consumers.
> 
> ### Fixed
> - Avoid double `wc_ecc_init_ex` call (was leaking async device context under
>   WOLFSSL_ASYNC_CRYPT && WC_ASYNC_ENABLE_ECC).
> - ECC key is now C-heap-allocated to remain valid across Rust moves under
>   non-default math configurations.
> ```
> 
> Also: minimum wolfSSL version is correctly documented as 5.9.0 in the README. `wc_HmacCopy` is the binding constraint (introduced in v5.9.0); `wc_rng_new_ex` is older (v5.7.0).
> 
> ## Advisory (track for follow-up, not blocking)
> ### Inline vs. heap inconsistency between ECC and RSA/DH/Curve25519
> Commit `113b6b7ec`'s rationale ("internal pointers breaking if Rust moves the ECC struct with some build configurations") implies certain wolfCrypt configs put self-pointers in C structs. ECC was migrated to a C-heap pointer; `RSA.wc_rsakey`, `DH.wc_dhkey`, and `Curve25519Key.wc_key` remain inline. In the default bindgen config these are move-safe, but if the same configs that affected ECC (non-SP integer math, `WOLFSSL_ASYNC_CRYPT`, FIPS, hardware-CB) also affect the other key types, they carry the same latent UB this PR just fixed for ECC. Worth a follow-up audit and/or a `// Safety:` comment in each consumer struct recording the move-safety assumption.
> 
> ## Nits (not blocking)
> * `src/ecc.rs:405` — the comment "`wc_ecc_key_new()` always initializes the key with INVALID_DEVID" is wrong on CAAM builds (`ecc.c:6174-6180` initializes with `WOLFSSL_CAAM_DEVID`). The subsequent `(*key).devId = dev_id;` under `#[cfg(wolf_crypto_cb)]` will then clobber the CAAM default if the caller passed `dev_id: None`.
> * `src/random.rs:67-70` — the `unsafe impl Send for RNG` safety comment claims the pointer is non-null but it isn't asserted. A `debug_assert!(!wc_rng.is_null())` after the success branch at `:137` and `:197` would document and enforce the invariant.
> * `src/random.rs:60-62` — DRBG state is mutated through `&self`. Sound because the state lives behind an opaque `*mut WC_RNG`, but `UnsafeCell<*mut WC_RNG>` would make the interior-mutability intent explicit and defeat any future compiler improvement that might re-derive `Sync` or hoist reads.
> * `src/ecc.rs:866-867, 923-924` — the null-terminator check accepts buffers with embedded NULs (only the last byte is checked). `XSTRLEN` / `mp_read_radix` on the C side stop at the first NUL, so embedded NULs cause silent truncation. Accepting `&CStr` would be cleaner.
> * `src/rsa.rs:1208-1212` — the `set_rng` doc test (fixed in `4ace46f2e`) does call `set_rng`, but the immediately-following `public_encrypt` uses a separate `enc_rng` argument, so the bound RNG isn't exercised here. Only the second `set_rng` at `:1217` + `private_decrypt` at `:1219` actually depends on the bound RNG (under `WC_RSA_BLINDING`). The first `set_rng` call is dead in this example.
> * `src/ecdsa.rs:177-185` and many `tests/test_*.rs` sites — stale `let mut rng = ...` + `&mut rng` reborrows left from before the `&RNG` migration. Compiles via reborrow coercion, just noisy. Same cleanup as commit `1a441c910`, just not applied everywhere.
> * `tests/test_hmac_mac.rs:43-62` — the Clone test would be stronger with divergent post-clone updates (both branches feed `suffix`, so the test verifies "clone preserves state" + "clone doesn't poison original" but not "diverged updates produce different MACs"). The existing test catches the most likely shallow-copy bug, so this is a nice-to-have.
> 
> ## What the PR does right (worth calling out)
> * **RNG ownership refactor:** the old `&'a mut WC_RNG` borrow was unsound (C cached the raw pointer past the borrow lifetime). The new design heap-allocates `WC_RNG` via `wc_rng_new_ex` (pointer stable across moves), correctly picks `Rc` over `Arc` (WC_RNG is not internally synchronized), splits `set_rng(RNG)` from `set_shared_rng(Rc<RNG>)`, and gets Drop ordering right — consumer free runs before RNG free, so `wc_FreeRsaKey`/`wc_ecc_key_free` can still safely read `key->rng`. Double-`set_rng` ordering is also correct (C pointer replaced before old `RngHandle` drops).
> * **ECC heap allocation:** switching to `*mut sys::ecc_key` via `wc_ecc_key_new` is the right fix for move-unsafety under configs where `mp_int.dp` points back into the same struct. `Pin` isn't needed because the Rust handle is a pointer to a never-moving C-heap object.
> * **Double-init fix:** real bug — `wc_ecc_key_new` already calls `wc_ecc_init_ex` internally; the explicit second call `XMEMSET`s the struct, blowing away heap context and leaking async device contexts in `WOLFSSL_ASYNC_CRYPT && WC_ASYNC_ENABLE_ECC` builds.
> * **HMAC `Clone`:** uses a _manual_ impl delegating to `wc_HmacCopy` instead of bitwise `#[derive(Clone)]` — the latter would be unsound for hash configs with internal pointers. `MaybeUninit::assume_init` is correctly placed after the error check.

I added the slice length checks to `ECC::import_unsigned`. I added the gating on the `alloc` feature for the RSA doc tests that called `set_shared_rng()`.

The version will be bumped later outside of this PR when preparing for the crate release.

holtrop-wolfssl:
retest this please (wolfSSL/PRB-fips-repo-and-harness-test-v3-part2' failed with result: FAILURE)

MarkAtwood (approved):
Re-reviewed at head `8c11a8e`. All three of my prior items are resolved to my satisfaction:

1. **Blocker — `ECC::import_unsigned` length validation** — fixed in `735d00d`. Validates `qx`/`qy`/`d` against `get_curve_size_from_id(curve_id)` and returns `BAD_FUNC_ARG` on short input. Regression test `test_ecc_import_unsigned_short_slices` covers the six short/empty permutations. Good.

   Correction to my prior review: I also flagged `import_private_key` / `import_private_key_ex`. Re-reading `wc_ecc_import_private_key_ex` (`wolfcrypt/src/ecc.c:11222`), that function threads the caller-supplied `privSz` through to `mp_read_unsigned_bin` — it does **not** read a hardcoded `key->dp->size` like `_ecc_import_raw_private` does. So those two functions are not vulnerable to the same OOB-read pattern, and scoping the fix to `import_unsigned` only was correct. My ask was overbroad — sorry for the noise.

2. **Five RSA doc tests missing `feature = "alloc"`** — fixed in `8c11a8e`. Diff hits all five gates I enumerated (`30`, `869`, `932`, `998`, `1069`) exactly as requested.

3. **Crate version bump + CHANGELOG** — accepting the deferral. Just please make sure the release commit captures both breaking changes:
   - `set_rng(&mut self, &mut RNG)` → `set_rng(&mut self, RNG)` on `ECC`, `RSA`, `Curve25519Key`
   - Cargo feature rename `std` → `alloc`

The advisory items (inline-vs-heap audit for `RSA`/`DH`/`Curve25519Key`, etc.) and the nits from my prior review remain as suggested follow-ups; none are merge-blocking.

Approving.

gasbytes (approved):

JacobBarthelmeh (approved):

PR #10405, opened 2026-05-06
ML-KEM: fix comments, API signatures, minor issues

SparkiDev:
# Description

More checks for public or private key not set.
wc_MlKemKey_Free clears key->flags
wc_MlKemKey_DecodePrivateKey now checks the public key is valid. wc_MlKemKey_EncodePrivateKey doesn't need calculate hash of public key as encoding the public key will do this.
EncodePrivateKey/EncodePublicKey now return BAD_STATE_E when flags not set.
mlkem_kdf, mlkem_check_public, mlkem_xof_absorb pointer parameters are now const.
Now all mlkem_redistribute_*_rand_avx2 functions are WOLFSSL_LOCAL. Changed Kyber uses to MlKem.

# Testing

Tested ML-KEM still works.

### DIFF
--- a/wolfcrypt/src/wc_mlkem.c
+++ b/wolfcrypt/src/wc_mlkem.c
@@ -28,30 +28,30 @@
  *   post-quantum-cryptography-standardization/round-3-submissions
  */
 
-/* Possible Kyber options:
+/* Possible ML-KEM options:
  *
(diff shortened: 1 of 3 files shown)


Frauschi (approved):

SparkiDev:
Re-based for the Python test fixes.

dgarske (approved):

PR #10408, opened 2026-05-06
Add UDP support to NetX sockets for DTLS sessions

hmohide:
# Description

Add UDP support to NetX sockets for DTLS sessions

Fixes zd#

src/internal.c
src/wolfio.c
wolfssl/internal.h
wolfssl/wolfio.h

# Testing

Yes, verified the DTLS communication with multiple session

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


dgarske:
Okay to test. Contributor agreement on file. Thank you @hmohide

night1rider:
#10414 Should fix the zephyr tests if you rebase to pull in the updated workflow

hmohide:
@dgarskeI think above failure not triggered by the changes? Do i need to make some change? Please guide

dgarske:
Hi @hmohide these would all be solved by a rebase against latest master and force push. Please go ahead and do that. Thanks

dgarske (approved):

src/wolfio.c
```diff
@@ -2628,15 +2628,136 @@ int NetX_Send(WOLFSSL* ssl, char *buf, int sz, void *ctx)
...
+    UINT  status;
+
+    (void)ssl;
+
+    if (nxCtx == NULL || nxCtx->nxUdpSocket == NULL) {
+        WOLFSSL_MSG("NetX Recv NULL parameters");
+        return WOLFSSL_CBIO_ERR_GENERAL;
+    }
+
+    if (nxCtx->nxPacket == NULL) {
```
dgarske: **🔴 [High] DTLS NetX receive maps timeouts to fatal error, breaking retransmission**

`NetX_ReceiveFrom` was copied from the TCP `NetX_Receive` and maps *every* non-`NX_SUCCESS` status from `nx_udp_socket_receive()` (including `NX_NO_PACKET`, i.e. a normal receive timeout) to `WOLFSSL_CBIO_ERR_GENERAL`. For TLS-over-TCP that is fine, but DTLS relies on the receive callback returning `WOLFSSL_CBIO_ERR_TIMEOUT` (or `WOLFSSL_CBIO_ERR_WANT_READ`) so the DTLS state machine can retransmit the last handshake flight on packet loss. The Embed DTLS paths in this same file (e.g. lines 2857-2861, 2935-2939) deliberately distinguish timeout from a hard error. As written, any lost handshake packet over a real/lossy link turns a recoverable timeout into a fatal `SOCKET_ERROR_E`, aborting the handshake. With `NX_WAIT_FOREVER` the callback instead blocks forever and wolfSSL never gets a chance to drive its own retransmit timer. Either way DTLS retransmission is non-functional. This passes a clean-link smoke test (matching the PR's testing note) but fails the core DTLS reliability requirement.

**Fix:** Map `NX_NO_PACKET` (and any other timeout/would-block status) to `WOLFSSL_CBIO_ERR_TIMEOUT`/`WOLFSSL_CBIO_ERR_WANT_READ`, and drive `nxWait` from the current DTLS timeout (`wolfSSL_dtls_get_current_timeout`) rather than a fixed wait option, so retransmission works on a lossy link.

src/wolfio.c
```diff
@@ -2628,15 +2628,136 @@ int NetX_Send(WOLFSSL* ssl, char *buf, int sz, void *ctx)
...
+        ssl->nxCtx.nxTcpSocket  = nxsocket;
+        ssl->nxCtx.nxWait       = waitoption;
+    }
+}
+
+#ifdef WOLFSSL_DTLS 
+/* The NetX receive callback for DTLS
+ *  return :  bytes read, or error
+ */
+int NetX_ReceiveFrom(WOLFSSL *ssl, char *buf, int sz, void *ctx)
```
dgarske, line 2668: **🟠 [Medium] DTLS receive does not capture peer address (single fixed-peer limitation)**

`NetX_ReceiveFrom` reads the datagram but never extracts the sender's IP/port (e.g. via `nx_udp_source_extract`). The send path (`NetX_SendTo`) always uses the statically configured `nxdIp`/`nxPort`. This means a DTLS *server* cannot reply to whichever client contacted it, and an `ssl` can only ever communicate with a single pre-known peer set via `wolfSSL_SetIO_NetX_Dtls`. The PR title mentions "multiple session" support; if server-side or learn-the-peer behavior is intended, the receive callback needs to record the source address and the send path needs to use it. If client-only with a fixed peer is the intended scope, that should be documented.

**Fix:** Clarify intended scope (client-only fixed peer vs. server). If server/multi-peer is intended, extract and store the source address on receive and use it in the send path.

dgarske, line 2668: **🟠 [Medium] New public API and DTLS IO callbacks have no test coverage**

The PR adds three new functions and a new public API (`wolfSSL_SetIO_NetX_Dtls`) with non-trivial logic (packet draining, IPv4/IPv6 branch, NULL-parameter validation) but no tests, and the PR checklist leaves 'added tests' unchecked. NetX is hardware/RTOS-specific so full integration tests are hard in CI, but at minimum the NULL-parameter error paths and the `copied == left` drain logic are unit-testable with a mocked context. Without any coverage, regressions in this path will go unnoticed.

**Fix:** Add at least minimal coverage for the NULL-parameter branches and the packet-drain/offset logic, or document why the NetX paths are excluded from CI.

wolfssl/wolfio.h
```diff
@@ -790,9 +790,16 @@ WOLFSSL_API void wolfSSL_SetIOWriteFlags(WOLFSSL* ssl, int flags);
 #ifdef HAVE_NETX
     WOLFSSL_LOCAL int NetX_Receive(WOLFSSL *ssl, char *buf, int sz, void *ctx);
     WOLFSSL_LOCAL int NetX_Send(WOLFSSL *ssl, char *buf, int sz, void *ctx);
-
     WOLFSSL_API void wolfSSL_SetIO_NetX(WOLFSSL* ssl, NX_TCP_SOCKET* nxsocket,
                                       ULONG waitoption);
+#ifdef WOLFSSL_DTLS 
+    WOLFSSL_LOCAL int NetX_ReceiveFrom(WOLFSSL *ssl, char *buf, int sz, void *ctx);
+    WOLFSSL_LOCAL int NetX_SendTo(WOLFSSL *ssl, char *buf, int sz, void *ctx);
+    WOLFSSL_API void wolfSSL_SetIO_NetX_Dtls(WOLFSSL* ssl, NX_UDP_SOCKET* nxsocket,
```
dgarske: **🟠 [Medium] New public API lacks doxygen; existing doxygen references renamed nxSocket field**

The new public API `wolfSSL_SetIO_NetX_Dtls` (WOLFSSL_API) has no doxygen entry, and the PR checklist 'updated/added doxygen' is unchecked. Additionally, the existing doxygen for `wolfSSL_SetIO_NetX` in `doc/dox_comments/header_files/wolfio.h` (and the -ja translation) documents the struct member as `nxSocket`, which this PR renamed to `nxTcpSocket`, so the published docs are now stale. (doc/ is outside the src/*.c review scope but is a direct consequence of the rename.)

**Fix:** Add a doxygen block for `wolfSSL_SetIO_NetX_Dtls` and update the `wolfSSL_SetIO_NetX` doxygen to reflect the `nxTcpSocket` rename.

src/wolfio.c
```diff
@@ -2628,15 +2628,136 @@ int NetX_Send(WOLFSSL* ssl, char *buf, int sz, void *ctx)
...
+    }
+
+    status = nx_packet_data_append(packet, buf, sz, pool, nxCtx->nxWait);
+    if (status != NX_SUCCESS) {
+        nx_packet_release(packet);
+        WOLFSSL_MSG("NetX Send data append error");
+        return WOLFSSL_CBIO_ERR_GENERAL;
+    }
+
+    if(nxCtx->nxdIp->nxd_ip_version == NX_IP_VERSION_V4)
```
dgarske: **🟠 [Medium] DTLS NetX path requires NetX Duo APIs without a guard**

The new DTLS code uses NetX Duo-only types/APIs: `NXD_ADDRESS`, `nxd_ip_version`/`NX_IP_VERSION_V4`, and `nxd_udp_socket_send`. Classic (non-Duo) NetX does not provide these. The code is gated only on `HAVE_NETX` + `WOLFSSL_DTLS`, so a classic-NetX build with DTLS enabled would fail to compile. The TCP path remains compatible with classic NetX, so this asymmetry is easy to hit.

**Fix:** Either gate the DTLS NetX path on a NetX Duo feature macro and document the dependency, or provide a classic-NetX fallback using `nx_udp_socket_send` only.

dgarske: **🔵 [Low] NetX_SendTo uses Allman braces and missing space after 'if' (style mismatch)**

The IPv4/IPv6 branch uses Allman-style braces (`{` on its own line) and `if(...)`/`}else` without the surrounding space, which does not match the K&R style used throughout wolfio.c (and the rest of this very function). The duplicated send-and-error-check block in both branches could also be collapsed.

**Fix:** Reformat to K&R braces with a space after `if`/`else` to match the file, and deduplicate the error-handling block.

wolfssl/internal.h
```diff
@@ -5711,10 +5711,15 @@ typedef struct DtlsMsg {
 
     /* NETX I/O Callback default */
     typedef struct NetX_Ctx {
-        NX_TCP_SOCKET* nxSocket;    /* send/recv socket handle */
+        NX_TCP_SOCKET* nxTcpSocket; /* send/recv tcp socket handle */
         NX_PACKET*     nxPacket;    /* incoming packet handle for short reads */
         ULONG          nxOffset;    /* offset already read from nxPacket */
         ULONG          nxWait;      /* wait option flag */
+#ifdef WOLFSSL_DTLS 
+        NX_UDP_SOCKET* nxUdpSocket; /* send/recv udp socket handle */
```
dgarske, line 5786: **🔵 [Low] Incorrect/misleading struct member comments (nxPort, nxUdpSocket)**

`nxPort` is commented `/* Port number for udp recv*/` but it is actually the *destination* port used only by `NetX_SendTo`. `nxUdpSocket` is commented `/* send/recv udp socket handle */` which is accurate, but `nxdIp`/`nxPort` are send-only. Also missing the space before `*/`. The comments should reflect actual usage to avoid confusing future maintainers (and `nxPort`/`nxdIp` would ideally be stored by value per the API finding).

**Fix:** Fix the `nxPort` comment (send, not recv) and tidy the `*/` spacing.

dgarske (commented):
## Skoll Code Review

**Findings:** 9 total — 9 posted, 0 skipped

### Posted findings
- [High] **DTLS NetX receive maps timeouts to fatal error, breaking retransmission** — `src/wolfio.c:2659-2665`
- [Medium] **wolfSSL_SetIO_NetX_Dtls stores caller pointers dereferenced much later (lifetime hazard)** — `src/wolfio.c:2751-2759, wolfssl/internal.h:5719-5721`
- [Medium] **DTLS receive does not capture peer address (single fixed-peer limitation)** — `src/wolfio.c:2644-2694`
- [Medium] **New public API and DTLS IO callbacks have no test coverage** — `src/wolfio.c:2644-2759`
- [Medium] **New public API lacks doxygen; existing doxygen references renamed nxSocket field** — `wolfssl/wolfio.h:798-801`
- [Medium] **DTLS NetX path requires NetX Duo APIs without a guard** — `src/wolfio.c:2729-2745`
- [Low] **NetX_SendTo uses Allman braces and missing space after 'if' (style mismatch)** — `src/wolfio.c:2729-2745`
- [Low] **Trailing whitespace will fail wolfSSL whitespace CI check** — `src/wolfio.c:2640,2708,2736,2744,2754; wolfssl/internal.h:5718; wolfssl/wolfio.h:795,798-801`
- [Low] **Incorrect/misleading struct member comments (nxPort, nxUdpSocket)** — `wolfssl/internal.h:5719-5721`

dgarske (changes requested):
@hmohide  Our internal AI review found several items to consider: https://github.com/wolfSSL/wolfssl/pull/10408#pullrequestreview-4469917435

src/wolfio.c
hmohide: addressed findings

src/wolfio.c
hmohide: addressed findings

hmohide: addressed findings

wolfssl/wolfio.h
hmohide: addressed findings

src/wolfio.c
hmohide: addressed findings

hmohide: addressed findings

wolfssl/internal.h
hmohide: addressed findings

hmohide:
@dgarske i have updated the review comments

src/wolfio.c
```diff
@@ -2628,15 +2628,158 @@ int NetX_Send(WOLFSSL* ssl, char *buf, int sz, void *ctx)
...
+        }
+        else {
+            waitOption = NX_NO_WAIT; /* non-blocking */
+        }
+
+        status = nx_udp_socket_receive(nxCtx->nxUdpSocket, &nxCtx->nxPacket,
+                                       waitOption);
+        if (status != NX_SUCCESS) {
+            if (status == NX_NO_PACKET) {
+                /* Normal receive timeout — allow DTLS to retransmit */
```
dgarske:
```
  8 bit:
  ./src/wolfio.c:2681:                /* Normal receive timeout ¶ allow DTLS to retransmit */
```

wolfssl/internal.h
```diff
@@ -5711,10 +5711,16 @@ typedef struct DtlsMsg {
 
     /* NETX I/O Callback default */
     typedef struct NetX_Ctx {
-        NX_TCP_SOCKET* nxSocket;    /* send/recv socket handle */
+        NX_TCP_SOCKET* nxTcpSocket; /* send/recv tcp socket handle */
         NX_PACKET*     nxPacket;    /* incoming packet handle for short reads */
         ULONG          nxOffset;    /* offset already read from nxPacket */
         ULONG          nxWait;      /* wait option flag */
+/* WOLFSSL_NETX_DUO: requires ThreadX NetX Duo (NXD_ADDRESS, nxd_udp_socket_send) */
+#if defined(WOLFSSL_DTLS) && defined(WOLFSSL_NETX_DUO)
```
dgarske, line 5785:
```
  unrecognized macros used:
  WOLFSSL_NETX_DUO
```

dgarske (changes requested):

src/wolfio.c
hmohide: updated the comment

wolfssl/internal.h
hmohide: added the macro to file

dgarske:
Jenkins retest this please "Build 'wolfSSL/PRB-fips-repo-and-harness-test-v3-part2' failed with result: FAILURE"

dgarske (commented):
## Skoll Code Review

### Posted findings
- [Medium] **NetX_ReceiveFrom does not validate the datagram source peer** — `src/wolfio.c:2645-2718`
- [Medium] **DTLS 1.3 quick-timeout can divide to zero and fall back to nxWait, disabling retransmission** — `src/wolfio.c:2662-2675`
- [Low] **Zero-length datagram returns 0 instead of being ignored** — `src/wolfio.c:2692-2717`
- [Low] **Redundant repeated wolfSSL_dtls_get_using_nonblock() calls** — `src/wolfio.c:2662-2685`

### Skipped findings
- [Medium] ~~No test coverage for new NetX DTLS UDP callbacks~~

src/wolfio.c
```diff
@@ -2628,15 +2628,158 @@ int NetX_Send(WOLFSSL* ssl, char *buf, int sz, void *ctx)
...
+        ssl->nxCtx.nxWait       = waitoption;
+    }
+}
+
+/* WOLFSSL_NETX_DUO: requires ThreadX NetX Duo (NXD_ADDRESS, nxd_udp_socket_send) */
+#if defined(WOLFSSL_DTLS) && defined(WOLFSSL_NETX_DUO)
+/* The NetX receive callback for DTLS
+ *  return :  bytes read, or error
+ */
+int NetX_ReceiveFrom(WOLFSSL *ssl, char *buf, int sz, void *ctx)
```
dgarske, line 2668: **🟠 [Medium] NetX_ReceiveFrom does not validate the datagram source peer**

The new DTLS receive callback hands every datagram delivered to the UDP socket straight to the record layer without checking the sender's address/port. The existing UDP DTLS callback `EmbedReceiveFrom` (same file, ~lines 857-918) deliberately compares the received peer against `dtlsCtx->peer` via `sockAddrEqual()` and silently drops/ignores packets from an unexpected peer (and rate-limits invalid peers toward a timeout). `NetX_ReceiveFrom` has no equivalent: after `nx_udp_socket_receive()` it never inspects the source (e.g. via `nxd_udp_packet_info_extract`). The PR description specifically claims support for "multiple session", which is exactly the scenario where unvalidated source filtering can cause datagrams from one peer to be fed into another session's record layer (and makes off-path datagram injection easier). If the intended deployment model is strictly one bound NetX UDP socket per session, this may be acceptable, but that assumption should be confirmed/documented.

**Fix:** Add source-peer validation to NetX_ReceiveFrom, or document that the API requires a dedicated UDP socket per DTLS session and is not safe to share across peers.

src/wolfio.c
```diff
@@ -2628,15 +2628,158 @@ int NetX_Send(WOLFSSL* ssl, char *buf, int sz, void *ctx)
...
+
+    if (nxCtx == NULL || nxCtx->nxUdpSocket == NULL) {
+        WOLFSSL_MSG("NetX Recv NULL parameters");
+        return WOLFSSL_CBIO_ERR_GENERAL;
+    }
+
+    if (nxCtx->nxPacket == NULL) {
+        /* Compute wait ticks from the current DTLS retransmit timeout so the
+         * DTLS state machine can drive retransmission on packet loss. */
+        if (!wolfSSL_dtls_get_using_nonblock(ssl)) {
```
dgarske: **🟠 [Medium] DTLS 1.3 quick-timeout can divide to zero and fall back to nxWait, disabling retransmission**

The wait time is derived from the current DTLS timeout (seconds), but for DTLS 1.3 with quick timeout it does `dtls_timeout = dtls_timeout / 4` using integer division and no lower bound. With the default `DTLS_TIMEOUT_INIT` of 1 second, `1 / 4 == 0`, so the subsequent `(dtls_timeout > 0)` test is false and the code falls back to `nxCtx->nxWait`. If the application configured `nxWait` as `NX_WAIT_FOREVER` (the value used in the new API's own doxygen example), `nx_udp_socket_receive()` will block indefinitely and the DTLS retransmit timer never fires on packet loss. The canonical pattern in `wolfssl/test.h` guards this with `if (timeout > 4 && wolfSSL_dtls13_use_quick_timeout(ssl)) timeout /= 4;` precisely to avoid collapsing to zero.

**Fix:** Guard the /4 so it cannot collapse the timeout to zero, and avoid falling back to a potentially infinite nxWait inside the blocking-receive timeout computation.

src/wolfio.c
```diff
@@ -2628,15 +2628,158 @@ int NetX_Send(WOLFSSL* ssl, char *buf, int sz, void *ctx)
...
+                return wolfSSL_dtls_get_using_nonblock(ssl)
+                       ? WOLFSSL_CBIO_ERR_WANT_READ
+                       : WOLFSSL_CBIO_ERR_TIMEOUT;
+            }
+            WOLFSSL_MSG("NetX Recv receive error");
+            return WOLFSSL_CBIO_ERR_GENERAL;
+        }
+    }
+
+    if (nxCtx->nxPacket) {
```
dgarske: **🔵 [Low] Zero-length datagram returns 0 instead of being ignored**

When a 0-byte UDP datagram is received, `nx_packet_length_get` yields `total == 0`, `left == 0`, `nx_packet_data_extract_offset` copies `copied == 0`, the packet is drained, and the function returns `0`. The companion `EmbedReceiveFrom` explicitly special-cases this (`WOLFSSL_MSG("Ignoring 0-length datagram"); continue;`) rather than returning 0, because a 0 return from the receive callback is ambiguous to the record layer (it is the conventional 'connection closed' indication for stream sockets). For DTLS this is an unusual edge case, but the divergence from the established Embed behavior is worth handling explicitly.

**Fix:** Explicitly discard zero-length datagrams instead of returning 0 from the receive callback.

src/wolfio.c
dgarske: **🔵 [Low] Redundant repeated wolfSSL_dtls_get_using_nonblock() calls**

`wolfSSL_dtls_get_using_nonblock(ssl)` is invoked up to three times in the same receive path (once to choose the wait option at line 2662, and again at line 2683 to choose the return code). The value cannot change between these calls within a single invocation; caching it in a local would be slightly clearer and avoids repeated function-call overhead. Minor, take it or leave it.

**Fix:** Cache the non-block flag in a single local (declared at the top of the function per wolfSSL C89 style) and reuse it.

doc/dox_comments/header_files/wolfio.h
```diff
@@ -450,14 +450,14 @@ void wolfSSL_SetIOWriteFlags(WOLFSSL* ssl, int flags);
...
     struct within the WOLFSSL structure.
 
     \return none No returns.
 
     \param ssl a pointer to a WOLFSSL structure, created using wolfSSL_new().
     \param nxSocket a pointer to type NX_TCP_SOCKET that is set to the
-    nxSocket member of the nxCTX structure.
+    nxTcpSocket member of the nxCtx structure.
     \param waitOption a ULONG type that is set to the nxWait member of
     the nxCtx structure.
```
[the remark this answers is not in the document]
hmohide: corrected

doc/dox_comments/header_files-ja/wolfio.h
```diff
@@ -404,12 +404,12 @@ void wolfSSL_SetIOWriteFlags(WOLFSSL* ssl, int flags);
...
 
-    \brief この関数は、WOLFSSL構造体内のnxCtx構造体のnxSocketおよびnxWaitメンバーを設定します。
+    \brief この関数は、WOLFSSL構造体内のnxCtx構造体のnxTcpSocketおよびnxWaitメンバーを設定します。
 
     \return none 戻り値なし。
 
     \param ssl wolfSSL_new()を使用して作成されたWOLFSSL構造体へのポインタ。
-    \param nxSocket nxCTX構造体のnxSocketメンバーに設定されるNX_TCP_SOCKET型へのポインタ。
+    \param nxSocket nxCtx構造体のnxTcpSocketメンバーに設定されるNX_TCP_SOCKET型へのポインタ。
     \param waitOption nxCtx構造体のnxWaitメンバーに設定されるULONG型。
```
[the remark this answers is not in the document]
hmohide: corrected

src/wolfio.c
hmohide: added the src & port validation

src/wolfio.c
hmohide: added the zero check for dtls timeout

src/wolfio.c
hmohide: added the handling for 0 length datagram

src/wolfio.c
hmohide: corrected

src/wolfio.c
```diff
@@ -2628,15 +2628,158 @@ int NetX_Send(WOLFSSL* ssl, char *buf, int sz, void *ctx)
...
+            return WOLFSSL_CBIO_ERR_GENERAL;
+        }
+
+        left = total - nxCtx->nxOffset;
+        status = nx_packet_data_extract_offset(nxCtx->nxPacket, nxCtx->nxOffset,
+                                               buf, sz, &copied);
+        if (status != NX_SUCCESS) {
+            WOLFSSL_MSG("NetX Recv data extract offset error");
+            return WOLFSSL_CBIO_ERR_GENERAL;
+        }
```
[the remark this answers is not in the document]
hmohide: corrected

src/wolfio.c
```diff
@@ -2628,15 +2628,158 @@ int NetX_Send(WOLFSSL* ssl, char *buf, int sz, void *ctx)
...
+                IsAtLeastTLSv1_3(ssl->version))
+                dtls_timeout = dtls_timeout / 4;
+        #endif
+            waitOption = (dtls_timeout > 0)
+                         ? (ULONG)dtls_timeout * NX_IP_PERIODIC_RATE
+                         : nxCtx->nxWait;
+        }
+        else {
+            waitOption = NX_NO_WAIT; /* non-blocking */
+        }
```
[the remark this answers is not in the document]
hmohide: corrected

dgarske:
Ah the "Build 'wolfSSL/PRB-fips-repo-and-harness-test-v3-part2' failed with result: FAILURE" test won't pass since you opened this PR on your fork's master branch. That's a known issue with our CI. I won't make you re-open the PR.

dgarske (commented):
## Skoll Code Review

### Posted findings
- [Medium] **Invalid-peer / 0-length datagram receive loop has no bound and re-arms full wait each iteration** — `src/wolfio.c:2700-2790`
- [Medium] **Strict pre-set peer filtering prevents DTLS server use (peer unknown until first ClientHello)** — `src/wolfio.c:2746-2751`
- [Low] **Infinite-loop callback has no trailing return; may warn 'control reaches end of non-void function'** — `src/wolfio.c:2790-2793`
- [Info] **Two stray blank lines added in generated doxygen header** — `doc/dox_comments/header_files/wolfio.h:517-518`

### Skipped findings
- [Low] ~~New NetX Duo DTLS callbacks have no test coverage~~

src/wolfio.c
```diff
@@ -2628,15 +2628,233 @@ int NetX_Send(WOLFSSL* ssl, char *buf, int sz, void *ctx)
...
+    doDtlsTimeout = ssl->options.handShakeState != HANDSHAKE_DONE;
+#ifdef WOLFSSL_DTLS13
+    if (ssl->options.dtls && IsAtLeastTLSv1_3(ssl->version)) {
+        doDtlsTimeout =
+            doDtlsTimeout || ssl->dtls13Rtx.rtxRecords != NULL ||
+            (ssl->dtls13FastTimeout && ssl->dtls13Rtx.seenRecords != NULL);
+    }
+#endif /* WOLFSSL_DTLS13 */
+
+    do {
```
dgarske, line 2709: **🟠 [Medium] Invalid-peer / 0-length datagram receive loop has no bound and re-arms full wait each iteration**

The `do { ... } while (1)` loop in `NetX_ReceiveFrom` uses `continue` to retry whenever a packet arrives from a non-matching peer (line 2750) or when a 0-length datagram is received (line 2764). On every `continue`, control re-enters the `if (nxCtx->nxPacket == NULL)` block and recomputes `waitOption` from the *full* DTLS retransmit timeout, then blocks again in `nx_udp_socket_receive`. There is no cap on how many junk packets can be ignored and no accounting of elapsed time. In blocking mode, a steady stream of mismatched-peer or 0-length datagrams (noisy network or adversarial traffic) keeps resetting the wait, so the call can block far longer than the DTLS retransmit timer intended and `WOLFSSL_CBIO_ERR_TIMEOUT` is never returned to the state machine -- retransmission stalls and the handshake can hang. The reference `EmbedReceiveFrom` explicitly guards against this with `invalidPeerPackets++ > DTLS_RECEIVEFROM_MAX_INVALID_PEER`, returning `WOLFSSL_CBIO_ERR_TIMEOUT`/`WOLFSSL_CBIO_ERR_WANT_READ` once the bound is exceeded (src/wolfio.c:881-887). The new callback omits any equivalent.

**Fix:** Bound the ignore loop (count or elapsed-time based) and return WOLFSSL_CBIO_ERR_TIMEOUT / WANT_READ so DTLS retransmission timing is preserved under noisy/adversarial traffic, mirroring EmbedReceiveFrom's DTLS_RECEIVEFROM_MAX_INVALID_PEER handling.

src/wolfio.c
```diff
@@ -2628,15 +2628,233 @@ int NetX_Send(WOLFSSL* ssl, char *buf, int sz, void *ctx)
...
+            }
+
+            status = nxd_udp_source_extract(nxCtx->nxPacket, &srcIp, &srcPort);
+            if (status != NX_SUCCESS) {
+                NetX_ResetPacket(nxCtx);
+                WOLFSSL_MSG("NetX Recv source extract error");
+                return WOLFSSL_CBIO_ERR_GENERAL;
+            }
+
+            if ((USHORT)srcPort != nxCtx->nxPort ||
```
dgarske: **🟠 [Medium] Strict pre-set peer filtering prevents DTLS server use (peer unknown until first ClientHello)**

`NetX_ReceiveFrom` always rejects datagrams whose source does not match `nxCtx->nxPort`/`nxCtx->nxdIp`, which are populated only by `wolfSSL_SetIO_NetX_Dtls`. A DTLS server does not know the client's address until the first ClientHello arrives, so it cannot pre-set a peer; with `nxdIp`/`nxPort` left zero-initialized, every incoming packet matches neither and is dropped (`NetX_ResetPacket` + `continue`), so the handshake never starts. By contrast `EmbedReceiveFrom` only filters once a peer is set (`dtlsCtx->userSet` / `peer.sz > 0`, src/wolfio.c:860-870) and otherwise accepts and can learn the source. There is also no API to update the stored peer from a received packet. This effectively limits the new NetX Duo DTLS path to clients (or scenarios with a statically known peer). If server support is intended, this is a functional gap; if client-only is intended, it should be documented.

**Fix:** Clarify intended scope. If DTLS servers should be supported, accept the first packet when no peer is configured and capture the source as the peer; otherwise document the client/known-peer-only restriction in the doxygen for wolfSSL_SetIO_NetX_Dtls.

src/wolfio.c
```diff
@@ -2628,15 +2628,233 @@ int NetX_Send(WOLFSSL* ssl, char *buf, int sz, void *ctx)
...
+
+        nxCtx->nxOffset += copied;
+
+        if (copied == left) {
+            WOLFSSL_MSG("NetX Recv Drained packet");
+            NetX_ResetPacket(nxCtx);
+        }
+
+        return copied;
+    } while (1);
```
dgarske, line 2835: **🔵 [Low] Infinite-loop callback has no trailing return; may warn 'control reaches end of non-void function'**

`NetX_ReceiveFrom` is declared `int` but its only exit points are `return` statements inside a `do { ... } while (1)` loop; there is no statement after the loop (just a `/* unreachable */` comment). GCC/Clang recognize the infinite loop and stay quiet, but several embedded toolchains commonly paired with NetX/ThreadX (ARM Compiler 5/6, IAR) emit a 'control may reach end of non-void function' / 'statement is unreachable' diagnostic, which becomes an error under wolfSSL's `-Werror`-style warning-clean builds. The existing TCP `NetX_Receive` avoids this by ending with a plain `return copied;`.

**Fix:** Add a trailing `return` after the loop (or restructure to a single exit) to keep the build warning-clean across embedded toolchains.

doc/dox_comments/header_files/wolfio.h
```diff
@@ -481,6 +481,42 @@ void wolfSSL_SetIOWriteFlags(WOLFSSL* ssl, int flags);
...
+    wolfSSL_SetIO_NetX_Dtls(ssl, &udpsocket, peeraddr, peerport, wait);
+    \endcode
+
+    \sa wolfSSL_SetIO_NetX
+    \sa NetX_SendTo
+    \sa NetX_ReceiveFrom
+*/
+void wolfSSL_SetIO_NetX_Dtls(WOLFSSL* ssl, NX_UDP_SOCKET* nxsocket,
+                              NXD_ADDRESS nxdip, USHORT nxport,
+                              ULONG waitoption);
```
dgarske, line 517: **⚪ [Info] Two stray blank lines added in generated doxygen header**

The new doxygen block for `wolfSSL_SetIO_NetX_Dtls` is followed by two consecutive blank lines before the next `/*!` block (same in the -ja copy). Cosmetic only and consistent enough to ignore, but worth tidying to a single blank line to match the surrounding file.

**Fix:** Remove the extra blank line for consistency.

src/wolfio.c
hmohide: added the DTLS_RECEIVEFROM_MAX_INVALID_PEER

src/wolfio.c
hmohide: Added handling the DTLS server use

src/wolfio.c
hmohide: added the trailing return

doc/dox_comments/header_files/wolfio.h
hmohide: updated

dgarske:
Jenkins retest this please "Build 'wolfSSL/PRB-fips-repo-and-harness-test-v3-part2' failed with result: FAILURE"

src/wolfio.c
```diff
@@ -2628,15 +2628,264 @@ int NetX_Send(WOLFSSL* ssl, char *buf, int sz, void *ctx)
...
+    doDtlsTimeout = ssl->options.handShakeState != HANDSHAKE_DONE;
+#ifdef WOLFSSL_DTLS13
+    if (ssl->options.dtls && IsAtLeastTLSv1_3(ssl->version)) {
+        doDtlsTimeout =
+            doDtlsTimeout || ssl->dtls13Rtx.rtxRecords != NULL ||
+            (ssl->dtls13FastTimeout && ssl->dtls13Rtx.seenRecords != NULL);
+    }
+#endif /* WOLFSSL_DTLS13 */
+
+    do {
```
dgarske, line 2709: **🟠 [Medium] NetX_ReceiveFrom does not track cumulative elapsed time across loop iterations**

Unlike the reference implementation `EmbedReceiveFrom` (src/wolfio.c:755-814), which decrements `dtls_timeout` by the elapsed time on each loop iteration (`dtls_timeout -= (LowResTimer() - start)`) so the *total* blocking time during a handshake is bounded by the DTLS retransmit timeout, `NetX_ReceiveFrom` recomputes the full `waitOption` from `wolfSSL_dtls_get_current_timeout(ssl)` on every iteration of the `do { } while(1)` loop. When an invalid-peer datagram (line 2752-2768) or a 0-length datagram (line 2784-2796) is received and the code `continue`s, the next `nx_udp_socket_receive` call again waits the full timeout. Under a flood of invalid/empty packets during the handshake, the effective time before the callback returns `WOLFSSL_CBIO_ERR_TIMEOUT` can grow to roughly `maxInvalidPeerPackets * dtls_timeout` (up to ~10x), delaying DTLS retransmission. The 10-packet cap bounds it (so it is not unbounded), but the retransmit timer fidelity degrades under attack. Consider tracking start time and decrementing the remaining wait, matching EmbedReceiveFrom.

**Fix:** Bound the cumulative receive time across loop iterations to the DTLS retransmit timeout rather than restarting the full wait each iteration.

src/internal.c
```diff
@@ -7907,9 +7913,13 @@ int InitSSL(WOLFSSL* ssl, WOLFSSL_CTX* ctx, int writeDup)
     if (wc_InitRwLock(&ssl->buffers.dtlsCtx.peerLock) != 0)
         return BAD_MUTEX_E;
 #endif
-
+#ifdef HAVE_NETX
```
dgarske, line 7963: **🔵 [Low] DTLS+NetX IOCB ctx switch means wolfSSL_dtls_set_peer() no longer affects NetX I/O**

For DTLS builds with `HAVE_NETX`, `IOCB_ReadCtx`/`IOCB_WriteCtx` now point to `&ssl->nxCtx` instead of `&ssl->buffers.dtlsCtx`. This is correct and necessary for the new NetX DTLS callbacks (which expect a `NetX_Ctx*`), and it also fixes a latent type-confusion (previously the NetX callbacks would have received a `dtlsCtx` pointer). However, as a consequence, the standard DTLS peer-setting API `wolfSSL_dtls_set_peer()` (which writes into `ssl->buffers.dtlsCtx.peer`) no longer influences the NetX UDP send/receive path; the peer must instead be set via the new `wolfSSL_SetIO_NetX_Dtls()` (or auto-locked on first received datagram). This is a reasonable design but is an easy-to-miss behavioral gap for users porting existing DTLS code to NetX. Worth documenting explicitly.

**Fix:** Document in the wolfSSL_SetIO_NetX_Dtls dox comment that the peer must be configured through this API (or auto-locked from the first datagram) and that wolfSSL_dtls_set_peer() does not apply to the NetX UDP path.

src/wolfio.c
```diff
@@ -2628,15 +2628,264 @@ int NetX_Send(WOLFSSL* ssl, char *buf, int sz, void *ctx)
...
+
+        nxCtx->nxOffset += copied;
+
+        if (copied == left) {
+            WOLFSSL_MSG("NetX Recv Drained packet");
+            NetX_ResetPacket(nxCtx);
+        }
+
+        return copied;
+    } while (1);
```
dgarske, line 2835: **🔵 [Low] Unreachable return after infinite do-while loop**

The `do { ... } while (1)` loop only exits via the `return` statements inside it; the trailing `return (int)copied;` at line 2823 is unreachable dead code. It is harmless (and arguably defensive against 'control reaches end of non-void function' warnings on some compilers), but it can also trigger -Wunreachable-code in stricter CI configurations. Minor cleanup.

**Fix:** Remove the unreachable trailing return, or convert the loop to a `for(;;)`/`while` with a single tail return if a compiler warns about the non-void path.

dgarske (commented):
## Skoll Code Review

### Posted findings
- [Medium] **NetX_ReceiveFrom does not track cumulative elapsed time across loop iterations** — `src/wolfio.c:2706-2774`
- [Low] **DTLS+NetX IOCB ctx switch means wolfSSL_dtls_set_peer() no longer affects NetX I/O** — `src/internal.c:7916-7922`
- [Low] **Unreachable return after infinite do-while loop** — `src/wolfio.c:2821-2823`
- [Low] **New code uses if (ssl) instead of if (ssl != NULL)** — `src/wolfio.c:2877-2887`

### Skipped findings
- [Low] ~~No test coverage for new NetX DTLS/UDP callbacks~~

src/wolfio.c
```diff
@@ -2628,15 +2628,264 @@ int NetX_Send(WOLFSSL* ssl, char *buf, int sz, void *ctx)
...
+        nx_packet_release(packet);
+        WOLFSSL_MSG("NetX Send socket send error");
+        return WOLFSSL_CBIO_ERR_GENERAL;
+    }
+
+    return sz;
+}
+
+/* like set_fd, but for default NetX Duo UDP context */
+void wolfSSL_SetIO_NetX_Dtls(WOLFSSL* ssl, NX_UDP_SOCKET* nxsocket,
```
dgarske, line 2892: **🔵 [Low] New code uses if (ssl) instead of if (ssl != NULL)**

Per the wolfSSL C coding style, NULL checks in new code should be written `if (ssl != NULL)` rather than the truthiness form `if (ssl)`. The new `wolfSSL_SetIO_NetX_Dtls` (and the rewritten `wolfSSL_SetIO_NetX` at line 2632) use `if (ssl)`. This matches the adjacent pre-existing style, so it is purely stylistic, but the new function is new code subject to the explicit-comparison convention.

**Fix:** Use explicit `if (ssl != NULL)` in the new function to match wolfSSL conventions for new code.

dgarske (changes requested):
https://github.com/wolfSSL/wolfssl/pull/10408#pullrequestreview-4526857829

hmohide:
@dgarske Earlier i saw this PR was approved by you. but i still getting the AI agent review comments. do i need to fix?

dgarske:
> @dgarske Earlier i saw this PR was approved by you. but i still getting the AI agent review comments. do i need to fix?

Yes take a look at them. I ran our internal code scanner tool and it found those. Note you are welcome to push back on the findings if you don't think they are needed.

dgarske:
> > @dgarske Earlier i saw this PR was approved by you. but i still getting the AI agent review comments. do i need to fix?
> 
> Yes take a look at them. I ran our internal code scanner tool and it found those. Note you are welcome to push back on the findings if you don't think they are needed.

Oh also please make sure you rebase to latest master. We made some CI optimization updates that you need to pull in. Thanks!

hmohide:
> > > @dgarske Earlier i saw this PR was approved by you. but i still getting the AI agent review comments. do i need to fix?
> > 
> > 
> > Yes take a look at them. I ran our internal code scanner tool and it found those. Note you are welcome to push back on the findings if you don't think they are needed.
> 
> Oh also please make sure you rebase to latest master. We made some CI optimization updates that you need to pull in. Thanks!

i have rebased the branch

hmohide:
@dgarske i have rebased. can you please review and approve

dgarske:

dgarske:
The CI "Build 'wolfSSL/PRB-fips-repo-and-harness-test-v3-part2' failed with result: FAILURE" is failing only because you have this PR opened on master. We can accept with that known CI issue. We are in a release cycle, so this won't be merged for a few days.

hmohide:
@dgarske any update on this PR. this one of the critical PR required for our relaese

src/wolfio.c
```diff
@@ -2628,15 +2628,264 @@ int NetX_Send(WOLFSSL* ssl, char *buf, int sz, void *ctx)
...
+    return left->nxd_ip_address.v6[0] == right->nxd_ip_address.v6[0] &&
+           left->nxd_ip_address.v6[1] == right->nxd_ip_address.v6[1] &&
+           left->nxd_ip_address.v6[2] == right->nxd_ip_address.v6[2] &&
+           left->nxd_ip_address.v6[3] == right->nxd_ip_address.v6[3];
+}
+
+/* The NetX receive callback for DTLS
+ *  return :  bytes read, or error
+ */
+int NetX_ReceiveFrom(WOLFSSL *ssl, char *buf, int sz, void *ctx)
```
dgarske, line 2668: **🟠 [Medium] NetX_ReceiveFrom does not reject negative sz before using it as an unsigned length**

The new DTLS receive callback passes the signed `int sz` directly into `nx_packet_data_extract_offset(nxCtx->nxPacket, nxCtx->nxOffset, buf, sz, &copied)` where the 4th parameter is an unsigned `ULONG buffer_length`. If `sz` were ever negative it would sign-extend to a huge length and permit a copy past the end of `buf`. The DTLS sibling this function mirrors, `EmbedReceiveFrom`, defensively guards this at its top (`if (sz < 0) return WOLFSSL_CBIO_ERR_GENERAL;`, wolfio.c:696-697), but `NetX_ReceiveFrom` omits the check. In practice wolfSSL passes a positive MTU-sized length, so this is hard to reach, but the guard is cheap and matches the pattern of the callback being copied.

**Fix:** Add `if (sz < 0) return WOLFSSL_CBIO_ERR_GENERAL;` early in NetX_ReceiveFrom (and consider the same in NetX_SendTo, where `sz` feeds nx_packet_data_append), matching EmbedReceiveFrom.

dgarske (commented):
## Skoll Code Review

### Posted findings
- [Medium] **NetX_ReceiveFrom does not reject negative sz before using it as an unsigned length** — `src/wolfio.c:2668-2806`
- [Low] **Unreachable return at end of NetX_ReceiveFrom (dead code)** — `src/wolfio.c:2821-2823`
- [Low] **Post-handshake invalid-peer/0-length flood is not rate-limited in blocking mode** — `src/wolfio.c:2759-2795`

### Skipped findings
- [Low] ~~No test coverage for new DTLS NetX Duo I/O callbacks~~

src/wolfio.c
dgarske, line 2835: **🔵 [Low] Unreachable return at end of NetX_ReceiveFrom (dead code)**

The `do { ... } while (1);` loop has no `break`; every path inside either returns or `continue`s. The trailing `return (int)copied;` after the loop is therefore unreachable dead code. Some compilers with -Wunreachable-code (e.g. MSVC C4702 under /WX) flag this, though this block only compiles under HAVE_NETX + WOLFSSL_NETX_DUO. The sibling `EmbedReceiveFrom` uses `break;` followed by a single `return recvd;`, which is cleaner.

**Fix:** Remove the unreachable trailing `return (int)copied;`, or restructure the loop to `break` out and return once, mirroring EmbedReceiveFrom.

src/wolfio.c
```diff
@@ -2628,15 +2628,264 @@ int NetX_Send(WOLFSSL* ssl, char *buf, int sz, void *ctx)
...
+            }
+
+            if (nxCtx->nxPort != 0 &&
+                (nxCtx->nxdIp.nxd_ip_version == NX_IP_VERSION_V4 ||
+                 nxCtx->nxdIp.nxd_ip_version == NX_IP_VERSION_V6)) {
+                if ((USHORT)srcPort != nxCtx->nxPort ||
+                    !NetX_PeerAddrEqual(&srcIp, &nxCtx->nxdIp)) {
+                    WOLFSSL_MSG("NetX Recv ignored packet from invalid peer");
+                    NetX_ResetPacket(nxCtx);
+                    if (doDtlsTimeout) {
```
dgarske, line 2766: **🔵 [Low] Post-handshake invalid-peer/0-length flood is not rate-limited in blocking mode**

The `invalidPeerPackets` counter that bounds discarded packets is only incremented when `doDtlsTimeout` is true (i.e., during a handshake/RTX window). After the handshake completes on a blocking socket (waitOption = nxCtx->nxWait, potentially NX_WAIT_FOREVER), a continuous stream of spoofed-source or 0-length datagrams causes the loop to `continue` indefinitely without returning to the application. This exactly mirrors the existing `EmbedReceiveFrom` behavior (its counter is also gated by `doDtlsTimeout`), so it is consistent with the repo, but worth confirming it is the intended semantics for NetX as well.

**Fix:** Confirm this matches the intended non-blocking/blocking contract for NetX (it matches EmbedReceiveFrom). No change required if intentional; otherwise consider bounding the discard loop even when doDtlsTimeout is false.

dgarske:
> @dgarske any update on this PR. this one of the critical PR required for our relaese

Just a few minor items to review here -> https://github.com/wolfSSL/wolfssl/pull/10408#pullrequestreview-4612671103
You are welcome to push back / reject items that are not required/needed.

src/wolfio.c
hmohide: removed

src/wolfio.c
hmohide: added the check

src/wolfio.c
hmohide: this is intentional and added the comments

dgarske:
Okay to test. Thank you @hmohide

dgarske:
Jenkins retest this please "ERROR: Error fetching remote repo 'origin'"

dgarske (approved):

src/wolfio.c
```diff
@@ -2628,15 +2628,276 @@ int NetX_Send(WOLFSSL* ssl, char *buf, int sz, void *ctx)
...
+                if (invalidPeerPackets > maxInvalidPeerPackets) {
+                    return usingNonblock
+                           ? WOLFSSL_CBIO_ERR_WANT_READ
+                           : WOLFSSL_CBIO_ERR_TIMEOUT;
+                }
+            }
+            continue;
+        }
+
+        if (nxCtx->nxOffset > total) {
```
dgarske, line 2809: **🟠 [Medium] NetX_ReceiveFrom retains leftover datagram bytes, diverging from DTLS datagram semantics**

When the caller's `sz` is smaller than the received datagram, the code copies `sz` bytes, advances `nxCtx->nxOffset`, keeps the packet, and returns `copied` without draining. The remainder of that datagram is then returned on the *next* call. This mirrors the TCP `NetX_Receive` byte-stream behavior, but for UDP/DTLS the standard reference `EmbedReceiveFrom` (recvfrom) delivers one datagram per call and discards any bytes beyond `sz` (datagram truncation semantics). Feeding the tail of an over-sized datagram back as a subsequent read could mis-frame DTLS records if the record layer ever requests fewer bytes than the datagram holds. In typical use wolfSSL passes a buffer large enough for a full record so this is an edge case, but the semantic difference from `EmbedReceiveFrom` is worth confirming.

**Fix:** Document/verify the leftover-retention choice for DTLS or switch to draining the packet each call so a single datagram is never split across reads.

src/wolfio.c
```diff
@@ -2628,15 +2628,276 @@ int NetX_Send(WOLFSSL* ssl, char *buf, int sz, void *ctx)
...
+    }
+
+    status = nx_packet_data_append(packet, buf, sz, pool, nxCtx->nxWait);
+    if (status != NX_SUCCESS) {
+        nx_packet_release(packet);
+        WOLFSSL_MSG("NetX Send data append error");
+        return WOLFSSL_CBIO_ERR_GENERAL;
+    }
+
+    if (nxCtx->nxdIp.nxd_ip_version == NX_IP_VERSION_V4) {
```
dgarske, line 2873: **🔵 [Low] NetX_SendTo silently uses IPv6 path when destination address version is unset**

The send path branches on `nxCtx->nxdIp.nxd_ip_version == NX_IP_VERSION_V4`, and the `else` falls through to `nxd_udp_socket_send()`. If a send is attempted before a peer address has been established (e.g. a server that sends before receiving a ClientHello, or an application that forgot to call `wolfSSL_SetIO_NetX_Dtls` with a valid peer), `nxdIp.nxd_ip_version` is 0 (zero-initialized), so the IPv6 branch runs with an all-zero NXD_ADDRESS. This fails gracefully (returns WOLFSSL_CBIO_ERR_GENERAL) rather than crashing, but the failure is opaque. In normal DTLS flows the peer is always locked before a send, so this is edge-case robustness only.

**Fix:** Optionally validate that `nxdIp.nxd_ip_version` is V4 or V6 before selecting the send path, to make misuse produce a clear diagnostic.

src/wolfio.c
```diff
@@ -2628,15 +2628,276 @@ int NetX_Send(WOLFSSL* ssl, char *buf, int sz, void *ctx)
...
+        nx_packet_release(packet);
+        WOLFSSL_MSG("NetX Send socket send error");
+        return WOLFSSL_CBIO_ERR_GENERAL;
+    }
+
+    return sz;
+}
+
+/* like set_fd, but for default NetX Duo UDP context */
+void wolfSSL_SetIO_NetX_Dtls(WOLFSSL* ssl, NX_UDP_SOCKET* nxsocket,
```
dgarske, line 2892: **🔵 [Low] wolfSSL_SetIO_NetX_Dtls does not reset any buffered packet/offset state**

The setter assigns `nxUdpSocket`, `nxdIp`, `nxPort`, and `nxWait` but leaves `nxPacket`/`nxOffset` untouched. On a freshly allocated WOLFSSL the nxCtx is zero-initialized so this is harmless in the normal one-shot setup case. However, if the API is ever called to re-point IO mid-session while a partially-drained packet is buffered, that stale packet/offset would be applied to the new socket. This matches the existing `wolfSSL_SetIO_NetX` (TCP) behavior, so it is a consistency observation rather than a new defect.

**Fix:** Leave as-is for parity with wolfSSL_SetIO_NetX, or reset nxPacket/nxOffset if mid-session re-binding is intended to be supported.

dgarske (commented):
## Skoll Code Review

### Posted findings
- [Medium] **NetX_ReceiveFrom retains leftover datagram bytes, diverging from DTLS datagram semantics** — `src/wolfio.c:2809-2832`
- [Low] **NetX_SendTo silently uses IPv6 path when destination address version is unset** — `src/wolfio.c:2870-2878`
- [Low] **wolfSSL_SetIO_NetX_Dtls does not reset any buffered packet/offset state** — `src/wolfio.c:2889-2899`

### Skipped findings
- [Medium] ~~No test coverage for new NetX Duo DTLS UDP callbacks~~

dgarske (changes requested):
https://github.com/wolfSSL/wolfssl/pull/10408#pullrequestreview-4647045554

src/wolfio.c
hmohide: Pushing back on this one. The review itself notes: "In normal DTLS flows the peer is always locked before a send" and "fails gracefully (returns WOLFSSL_CBIO_ERR_GENERAL) rather than crashing". Adding an explicit version check would add defensive code for a misuse scenario that already fails safely with a general error. The caller is responsible for setting a valid peer via wolfSSL_SetIO_NetX_Dtls before any send — no change needed.

src/wolfio.c
hmohide: Pushing back on this one. The review itself acknowledges: "This matches the existing wolfSSL_SetIO_NetX (TCP) behavior, so it is a consistency observation rather than a new defect." Diverging from the TCP setter behavior here would create an inconsistency without a clear benefit. Mid-session re-binding is not a supported use case for this API. No change needed.

src/wolfio.c
hmohide: fixed the issue

dgarske:
Jenkins retest this please.

dgarske (approved):

dgarske:
Our PRB CI won't pass the FIPS test because you did this from "master" however I know the changes do not impact FIPS. So I am going to accept as-is. Thank you for being an approved contributor.

PR #10416, opened 2026-05-06
fix: guard wc_Ed448PublicKeyToDer ed448_export_public call for FIPS<7

jackctj117:
This pull request introduces a conditional compilation change to the `wc_Ed448PublicKeyToDer` function in `asn.c` to improve compatibility with FIPS builds. Specifically, it adds a preprocessor check to handle the function call to `wc_ed448_export_public` differently based on the FIPS version.

Compatibility fix for FIPS builds:

* In `wolfcrypt/src/asn.c`, the call to `wc_ed448_export_public` is now conditionally compiled: for FIPS version 3 less than 7.0.0, the `key` parameter is explicitly cast to non-const, addressing compatibility or API differences in those builds.

### DIFF
--- a/wolfcrypt/src/asn.c
+++ b/wolfcrypt/src/asn.c
@@ -12931,7 +12931,11 @@ int wc_Ed448PublicKeyToDer(const ed448_key* key, byte* output, word32 inLen,
         return BAD_FUNC_ARG;
     }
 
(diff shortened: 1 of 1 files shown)


dgarske (approved):

dgarske:
Fixes issue introduced with https://github.com/wolfSSL/wolfssl/pull/10372

PR #10418, opened 2026-05-06
20260506 Coverity

rlm2002:
# Description

560796: Dereference before null check - remove check as `extension->data` would not be NULL at this point
560760: Logically dead code - remove dead length check
560126: Dereference after null check - introduce `defaultSz` variable to handle `s == NULL`
560794: Dereference after null check - remove check as `ssl->ctx` and `ssl->ctx->method` would not be NULL in normal operation at this point

Also adds unit test for `NULL` parameters in `EVP_DigestFinal()` with SHAKE

# Testing

# Checklist

 - [x] added tests
 - [ ] updated/added doxygen


rlm2002:
Retest this please Jenkins: history lost

dgarske (approved):

PR #10426, opened 2026-05-07
protocol correctness, OpenSSL-compat hardening, and sensitive-memory zeroization

JeremiahM37:
Fixes F-1837, F-1838, F-2086, F-2883, F-2884, F-3218, F-3239, F-3240, F-3454, F-3457, F-3491, F-3677, F-3679, F-3685, F-3688.

- des: fix sizeof(Des3) -> sizeof(Des) (F-2883/2884).
- x509: correct NOT_BEFORE_FIELD reason string (F-3218).
- ocsp: pendingCAs fallback uses parent hash, not seed (F-3677).
- dtls13: reject seq diff >= 2^32 in window update (F-2086).
- dtls13: rtx ctx match accounts for handshake header + length (F-3454).
- tls13: illegal_parameter alert for non-zero CertReq ctx (F-3457).
- pkcs7: return WOLFSSL_FAILURE on alloc fail (F-3679).
- rand: implement RAND_load_file (file I/O + DRBG reseed) (F-3685).
- rand: probe global RNG in RAND_status (F-3688).
- dtls13: zero epoch keys on free; ForceZero in slot reuse (F-3491).
- sniffer: zero keylog stack secrets on return (F-1837).
- sniffer: zero SecretNode before free (F-1838).
- tls13: zero Exporter scratch on all returns (F-3239).
- keys: zero shaOutput in master derive (F-3240).

### DIFF
--- a/src/ssl.c
+++ b/src/ssl.c
@@ -20129,19 +20129,45 @@ int wolfSSL_RAND_poll(void)
     int wolfSSL_RAND_status(void)
     {
         int ret = WOLFSSL_SUCCESS;
+        int useGlobalRng = 1;
     #ifndef WOLFSSL_NO_OPENSSL_RAND_CB
         if (wolfSSL_RAND_InitMutex() == 0 &&
                 wc_LockMutex(&gRandMethodMutex) == 0) {
-            if (gRandMethods && gRandMethods->status)
+            if (gRandMethods && gRandMethods->status) {
                 ret = gRandMethods->status();
+                useGlobalRng = 0;
+            }
             wc_UnLockMutex(&gRandMethodMutex);
         }
         else {
             ret = WOLFSSL_FAILURE;
+            useGlobalRng = 0;
         }
-    #else
-        /* wolfCrypt provides enough seed internally, so return success */
     #endif
+
+        /* Drive the global RNG so init / DRBG state failures (mutex
+         * acquisition, reseed required, corrupted state) surface to the
+         * caller. DRBG output is deterministic between reseeds, so this
+         * does not directly probe the entropy source. */
+    #ifdef HAVE_GLOBAL_RNG
+        if (useGlobalRng) {
+            if (wolfSSL_RAND_Init() != WOLFSSL_SUCCESS) {
+                ret = WOLFSSL_FAILURE;
+            }
+            else if (wc_LockMutex(&globalRNGMutex) != 0) {
+                ret = WOLFSSL_FAILURE;
+            }
+            else {
+                byte b = 0;
+                int genRet = wc_RNG_GenerateBlock(&globalRNG, &b, 1);
+                wc_UnLockMutex(&globalRNGMutex);
+                ForceZero(&b, 1);
+                if (genRet != 0)
+                    ret = WOLFSSL_FAILURE;
+            }
+        }
+    #endif
+        (void)useGlobalRng;
         return ret;
     }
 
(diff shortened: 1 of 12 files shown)


src/ssl.c
```diff
@@ -20130,17 +20130,40 @@ int wolfSSL_RAND_poll(void)
...
-            if (gRandMethods && gRandMethods->status)
+            if (gRandMethods && gRandMethods->status) {
                 ret = gRandMethods->status();
+                customMethodChecked = 1;
+            }
             wc_UnLockMutex(&gRandMethodMutex);
         }
         else {
-            ret = WOLFSSL_FAILURE;
+            return WOLFSSL_FAILURE;
```
SparkiDev: Only one exit point.

src/ssl.c
```diff
@@ -20175,12 +20198,96 @@ void wolfSSL_RAND_screen(void)
 
 int wolfSSL_RAND_load_file(const char* fname, long len)
 {
+#if !defined(NO_FILESYSTEM) && defined(HAVE_HASHDRBG)
+    XFILE  f;
+    long   maxBytes;
+    long   readSoFar = 0;
+    int    ret = 0;
+#ifndef WOLFSSL_SMALL_STACK
+    unsigned char buf[256];
```
SparkiDev: 256 is a magic number - use a #define.

src/ssl.c
```diff
@@ -20175,12 +20198,96 @@ void wolfSSL_RAND_screen(void)
...
+    WOLFSSL_ENTER("wolfSSL_RAND_load_file");
+
+    if (fname == NULL)
+        return WOLFSSL_FATAL_ERROR;
+
+    /* OpenSSL semantics: RAND_load_file(file, -1) reads up to an
+     * implementation-defined maximum. Match OpenSSL's RAND_LOAD_BUF_SIZE
+     * (1 MB) so callers passing -1 to ingest a multi-KB seed file aren't
+     * silently truncated. */
+    maxBytes = (len < 0) ? (1L << 20) : len;
```
SparkiDev: 1L << 20 is a magic number - use a #define.

src/ssl.c
```diff
@@ -20175,12 +20198,96 @@ void wolfSSL_RAND_screen(void)
...
+
+cleanup:
+    XFCLOSE(f);
+    ForceZero(buf, 256);
+#ifdef WOLFSSL_CHECK_MEM_ZERO
+    /* Balance the wc_MemZero_Add above for both stack/heap paths. */
+    wc_MemZero_Check(buf, 256);
+#endif
+#ifdef WOLFSSL_SMALL_STACK
+    XFREE(buf, NULL, DYNAMIC_TYPE_TMP_BUFFER);
```
SparkiDev, line 20285: Freeing 'buf' has the MemZero check called.
Only do the check if !defined(WOLFSSL_SMALL_STACK)

SparkiDev (changes requested):

src/ssl.c
JeremiahM37: Fixed

src/ssl.c
JeremiahM37: Added WOLFSSL_RAND_LOAD_FILE_BUF_SZ

src/ssl.c
JeremiahM37: Added WOLFSSL_RAND_LOAD_FILE_MAX_BYTES

src/ssl.c
JeremiahM37: Fixed

JeremiahM37:
Jenkins retest this please

SparkiDev (approved):

PR #10435, opened 2026-05-08
Improved handling for ClientHello default key share group

Frauschi:
Decouples the speculative key share group for the ClientHello from `preferredGroup[0]`. The new macro `WOLFSSL_KEY_SHARE_DEFAULT_GROUP` prefers widely deployed groups (PQ/T hybrids with X25519 or SECP256R1, then SECP256R1/X25519/SECP384R1, then FFDHE 2048/3072) to reduce the chance of a HelloRetryRequest, and falls back to `preferredGroup[0]` for configurations not covered explicitly. Users can override the default via user_settings.h or a manually passed `-DWOLFSSL_KEY_SHARE_DEFAULT_GROUP=xxx` via autoconf.

Furthermore, an empty key share is now sent when the user's group list does not intersect `preferredGroup[]`, keeping TLS 1.3 negotiation alive instead of allowing a silent TLS 1.2 downgrade or handshake failure due to a missing key share extension.

Finally, some refactoring in the key share code and the `preferredGroup` list code happened to address some latent issues or remove dead code.

### DIFF
--- a/src/tls.c
+++ b/src/tls.c
@@ -11164,10 +11164,14 @@ int TLSX_KeyShare_Empty(WOLFSSL* ssl)
     return ret;
 }
 
+/* Compile-time gating must stay aligned with TLSX_PopulateSupportedGroups().
+ * Runtime-only conditions in that function (TLS 1.3 version check, FFDHE
+ * key-size bounds, session-resumption short-circuit, downgrade-aware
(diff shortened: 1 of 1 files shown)


Frauschi:
Jenkins retest this please

dgarske (approved):

SparkiDev (approved):

PR #10453, opened 2026-05-11
Fix mem_track.h compile failure on multi-threaded non-Linux builds

LinuxJedi:
The memLock mutex and #include <pthread.h> in mem_track.h were declared under #ifdef DO_MEM_LIST (Linux/macOS/Zephyr only), but referenced under the broader guard

    !defined(SINGLE_THREADED) && \
    (defined(DO_MEM_LIST) || defined(DO_MEM_STATS))

Since DO_MEM_STATS is defined whenever WOLFSSL_TRACK_MEMORY + USE_WOLFSSL_MEMORY are set without WOLFSSL_STATIC_MEMORY, any non-Linux/Mac/Zephyr multi-threaded build failed to compile with implicit pthread_mutex_lock declarations and undeclared memLock.

Replace the raw pthread mutex with wolfSSL's portable mutex API (wc_InitMutex / wc_LockMutex / wc_UnLockMutex / wc_FreeMutex) so locking works on every platform wolfSSL already ports to. InitMemoryTracker now calls wc_InitMutex before
wolfSSL_SetAllocators installs TrackMalloc, guarded by a memLockInit flag for idempotency. CleanupMemoryTracker calls wc_FreeMutex after restoring the default allocators so no in-flight allocation races a freed mutex. The four mutex guards in TrackMalloc/TrackFree and the two in InitMemoryTracker/ ShowMemoryTracker are unified on the same condition as the memLock declaration itself.

ZD #21763

### DIFF
--- a/.github/workflows/freertos-mem-track.yml
+++ b/.github/workflows/freertos-mem-track.yml
@@ -1,0 +1,35 @@
+name: FreeRTOS mem_track.h compile regression
+
+# START OF COMMON SECTION
+on:
+  push:
+    branches: [ 'master', 'main', 'release/**' ]
+  pull_request:
+    branches: [ '*' ]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+# END OF COMMON SECTION
+
+# Regression test for the memLock / pthread.h guard misalignment in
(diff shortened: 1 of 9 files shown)


LinuxJedi:
Jenkins retest this please, FIPS test fixed

SparkiDev (approved):

PR #10474, opened 2026-05-13
Add timeout to membrowse CI tests

Frauschi:
These tests often hang in the package installation process, which is quickly resolved with a retry. Without a manual timeout, the default 6 hour timeout is used, which is far too long. The new 10 minute timeout is more than enough for a successful test run.

Also moves the package installation to the `install-apt-deps` action to make that more robust.


Frauschi:
Following #10475, the second commit moves the membrowse workflows also to the `install-apt-deps` action for more-robust package installation.

SparkiDev (approved):

PR #10404, opened 2026-05-06
test.c: Improved testing

SparkiDev:
# Description

Top-level test extraction into typed sub-functions

- Digest tests (MD5, SHA-1, SHA-224, SHA-256, SHA-384, SHA-512,
SHA-512/224,   SHA-512/256, SHA-3 224/256/384/512, SHAKE128/256): each
test category (KAT, large-hash, copy-cleanup, unaligned-memory,
interleave, copy-Sha3 heap-allocation pattern) is now its own static
wc_test_ret_t function rather than inline blocks inside one giant
top-level test.
- rsa_test: extracted rsa_pkcs1_test, rsa_cert_parse_test,
rsa_pub_key_decode_test, rsa_certreq_test.
- ecc_test: extracted ecc_test_all_curves, ecc_test_all_deterministic_k
from the dispatcher; ecc_encrypt_combos_test from ecc_encrypt_test;
ecc_buffers_encrypt_test, ecc_buffers_sign_verify_test,
ecc_buffers_pub_key_decode_test from ecc_test_buffers.
- dh_test: extracted dh_keyagree_test, dh_check_priv_key_test,
dh_agree_ct_test, dh_key_import_export_test, dh_set_check_key_test.
- Curve25519 / Ed25519 / Curve448 / Ed448: extracted KAT, key-agreement,
ASN, and rare-sig sub-tests from each top-level dispatcher.
- Cipher tests: extracted streaming sub-tests from aesofb_test,
aescfb_test, aes_cts_test (aes_cts_192/256_stream_test).

Coverage additions

- Added state-copy testing inside the SHA-3 KAT loops (matching the existing pattern in SHA-1/224/256/etc. KAT loops).
- Added a SHA-224 large-hash test (parity with the other digest families).
- Added copy-cleanup tests for SHA-512/224 and SHA-512/256.
- Added Unaligned memory access testing to all *_large_hash_test functions (MD5, SHA-1/224/256/384, SHA3-224/256/384/512), matching the pattern that previously only existed for SHA-512/224/256.

# Testing

/configure --disable-shared --enable-all --enable-aescts

PR #10480, opened 2026-05-13
20260513-fixes

douzzer:
wolfcrypt/test/test.c: reapply lost fixes from 03cee6f2bf to aes_cbc_iv_state_test() (lost in #10404 / df5b2b6cb1).

.wolfssl_known_macro_extras: add __CHERI_PURE_CAPABILITY__.


dgarske (approved):

wolfcrypt/test/test.c
```diff
@@ -15738,7 +15738,7 @@ static wc_test_ret_t aes_cbc_iv_state_test(Aes* enc, Aes* dec, byte* cipher,
      * the wrong ciphertext block into aes->reg between calls (e.g. the
      * first block instead of the last) will fail the second KAT. */
     {
-        WOLFSSL_SMALL_STACK_STATIC const byte msg4[] = {
+        static const byte msg4[] = {
```
dgarske, line 15741: If this keeps getting lost perhaps better to fix with proper macros around it or a (void) at end? I recall its related to a possibly unused scenario?

douzzer: actually I think we can globally replace `WOLFSSL_SMALL_STACK_STATIC` with `static` in test.c -- would that fit the bill?

dgarske: That doesn't seem like the right path. WOLFSSL_SMALL_STACK_STATIC was introduced for a reason. Is there a reason that `WOLFSSL_SMALL_STACK_STATIC` does map to -> `static`?

douzzer:
retest this please

PR #10486, opened 2026-05-14
Add error when RPK is used with DANE stub

embhorn:
# Description

DANE support was stubbed for the compatibility layer. Ensure an actual connection errors if it is used.

Thanks to Peter Samarin for reporting the issue and reviewing the fix.

Fixes zd21825

# Testing

Added test cases in `test_tls13_rpk_handshake`

# Checklist

 - [x] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/doc/dox_comments/header_files/ssl.h
+++ b/doc/dox_comments/header_files/ssl.h
@@ -16096,6 +16096,78 @@ int wolfSSL_get_negotiated_client_cert_type(WOLFSSL* ssl, int* tp);
 int wolfSSL_get_negotiated_server_cert_type(WOLFSSL* ssl, int* tp);
 
 /*!
+ \ingroup Setup
+ \brief  Pin a DER-encoded SubjectPublicKeyInfo that the peer is expected to
+ present as a Raw Public Key (RFC 7250), establishing out-of-band trust on the
+ WOLFSSL_CTX object. An unauthenticated RPK peer is otherwise rejected: when the
+ peer is being authenticated (any verify mode other than WOLFSSL_VERIFY_NONE)
+ the handshake fails closed unless the presented key matches a pin (or a verify
+ callback accepts it). May be called more than once to pin several keys, up to
+ WOLFSSL_MAX_RPK_PINS. The key is stored as its SHA-256 digest, so this API
+ requires SHA-256. Pins are append-only for the lifetime of the CTX.
+
+ \return WOLFSSL_SUCCESS on success
+ \return BAD_FUNC_ARG if ctx or spki is NULL, or spkiSz is 0
+ \return BUFFER_E if the pin table is already full (WOLFSSL_MAX_RPK_PINS)
+ \return other negative error code on a hashing failure
+
+ \param ctx     WOLFSSL_CTX object pointer
+ \param spki    DER-encoded SubjectPublicKeyInfo the peer is expected to present
+ \param spkiSz  length of spki in bytes
+    _Example_
+ \code
+  int ret;
+  WOLFSSL_CTX* ctx;
+  const unsigned char* spki; /* DER SubjectPublicKeyInfo */
+  unsigned int spkiSz;
+  ...
+
+  ret = wolfSSL_CTX_set_expected_rpk(ctx, spki, spkiSz);
+ \endcode
+ \sa wolfSSL_set_expected_rpk
+ \sa wolfSSL_set_client_cert_type
+ \sa wolfSSL_set_server_cert_type
+ */
+int wolfSSL_CTX_set_expected_rpk(WOLFSSL_CTX* ctx, const unsigned char* spki,
+    unsigned int spkiSz);
+
+/*!
+ \ingroup Setup
+ \brief  Pin a DER-encoded SubjectPublicKeyInfo that the peer is expected to
+ present as a Raw Public Key (RFC 7250), establishing out-of-band trust on the
+ WOLFSSL object. An unauthenticated RPK peer is otherwise rejected: when the
+ peer is being authenticated (any verify mode other than WOLFSSL_VERIFY_NONE)
+ the handshake fails closed unless the presented key matches a pin (or a verify
+ callback accepts it). May be called more than once to pin several keys, up to
+ WOLFSSL_MAX_RPK_PINS. The key is stored as its SHA-256 digest, so this API
+ requires SHA-256. Pins are append-only for the lifetime of the object.
+
+ \return WOLFSSL_SUCCESS on success
+ \return BAD_FUNC_ARG if ssl or spki is NULL, or spkiSz is 0
+ \return BUFFER_E if the pin table is already full (WOLFSSL_MAX_RPK_PINS)
+ \return other negative error code on a hashing failure
+
+ \param ssl     WOLFSSL object pointer
+ \param spki    DER-encoded SubjectPublicKeyInfo the peer is expected to present
+ \param spkiSz  length of spki in bytes
+    _Example_
+ \code
+  int ret;
+  WOLFSSL* ssl;
+  const unsigned char* spki; /* DER SubjectPublicKeyInfo */
+  unsigned int spkiSz;
+  ...
+
+  ret = wolfSSL_set_expected_rpk(ssl, spki, spkiSz);
+ \endcode
+ \sa wolfSSL_CTX_set_expected_rpk
+ \sa wolfSSL_set_client_cert_type
+ \sa wolfSSL_set_server_cert_type
+ */
+int wolfSSL_set_expected_rpk(WOLFSSL* ssl, const unsigned char* spki,
+    unsigned int spkiSz);
+
+/*!
 
 \brief Enable use of ConnectionID extensions for the SSL object. See RFC 9146
 and RFC 9147
--- a/doc/dox_comments/header_files/ssl.h
+++ b/doc/dox_comments/header_files/ssl.h
@@ -16104,7 +16104,8 @@ int wolfSSL_get_negotiated_server_cert_type(WOLFSSL* ssl, int* tp);
  the handshake fails closed unless the presented key matches a pin (or a verify
  callback accepts it). May be called more than once to pin several keys, up to
  WOLFSSL_MAX_RPK_PINS. The key is stored as its SHA-256 digest, so this API
- requires SHA-256. Pins are append-only for the lifetime of the CTX.
+ requires SHA-256. Pins are append-only - there is no per-entry remove, but
+ wolfSSL_CTX_clear_expected_rpk() empties the table so it can be repopulated.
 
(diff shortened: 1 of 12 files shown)


doc/dox_comments/header_files/ssl.h
```diff
@@ -16012,6 +16012,78 @@ int wolfSSL_get_negotiated_client_cert_type(WOLFSSL* ssl, int* tp);
...
+ \return BUFFER_E if the pin table is already full (WOLFSSL_MAX_RPK_PINS)
+ \return other negative error code on a hashing failure
+
+ \param ctx     WOLFSSL_CTX object pointer
+ \param spki    DER-encoded SubjectPublicKeyInfo the peer is expected to present
+ \param spkiSz  length of spki in bytes
+    _Example_
+ \code
+  int ret;
+  WOLFSSL_CTX* ctx;
```
dgarske, line 16121: **🟠 [Medium] Nested '*/' inside doxygen \code block truncates the comment**

Both new doxygen blocks embed an inline C comment inside the `\code` example: `const unsigned char* spki; /* DER SubjectPublicKeyInfo */`. The enclosing doxygen block opens with `/*!` and the inner `*/` is the first comment terminator the lexer sees, so it closes the doxygen comment early. Everything after it (the rest of the `\code`, the `\endcode`, and the `\sa` cross-references) falls outside the comment and the trailing `*/` at line 16047/16083 becomes stray text. A grep over the entire `doc/dox_comments/header_files/` tree shows these are the ONLY two inline `; /* ... */` comments in the directory, so this also deviates from the established convention (other examples never put a block comment inside `\code`). Result: the generated documentation for these two new public APIs is likely truncated/mis-associated.

**Fix:** Remove the inline `/* */` from both `\code` examples (or describe the buffer in the surrounding prose instead) so the doxygen comment is not terminated early.

src/internal.c
```diff
@@ -16818,6 +16855,63 @@ int ProcessPeerCerts(WOLFSSL* ssl, byte* input, word32* inOutIdx,
...
+                         * the actual key status even under WOLFSSL_VERIFY_NONE
+                         * rather than leaving it at WOLFSSL_X509_V_OK. */
+                        if (!rpkTrusted &&
+                                ssl->peerVerifyRet == WOLFSSL_X509_V_OK) {
+                            ssl->peerVerifyRet = (unsigned long)
+                                WOLFSSL_X509_V_ERR_RPK_UNTRUSTED;
+                        }
+                    #endif /* OPENSSL_EXTRA || OPENSSL_EXTRA_X509_SMALL */
+
+                        /* Fail closed when the peer is being authenticated: an
```
dgarske, line 17803: **🟠 [Medium] RPK now fails closed for all HAVE_RPK builds — notable behavior change**

This PR converts Raw Public Key (RFC 7250) handshakes from "always complete, validate out-of-band afterward" to "fail closed unless pinned or accepted by a verify callback" whenever the peer is being authenticated (`!ssl->options.verifyNone`). This applies to every `HAVE_RPK` build, including builds without `OPENSSL_EXTRA` (which previously had no untrusted-RPK handling) and `NO_SHA256` builds (where the new pinning API is unavailable, so every RPK peer fails closed unless a verify callback accepts it). The change is correct and the intended fix for the dead DANE stub, and it is documented extensively in `wolfssl/ssl.h`. Flagging only so the change in default behavior is consciously acknowledged: existing applications that relied on RPK handshakes completing and then calling `wolfSSL_get_verify_result()` will now see handshakes abort with `ASN_NO_SIGNER_E` unless they pin the key or install a verify callback.

**Fix:** Confirm this default change is intended for all HAVE_RPK configs and ensure release notes / migration docs call it out prominently (the ssl.h comment is good; a CHANGELOG entry would help downstreams who don't read the header).

src/ssl_api_cert.c
```diff
@@ -430,6 +430,82 @@ int wolfSSL_get_negotiated_server_cert_type(WOLFSSL* ssl, int* tp)
...
+ *
+ * @param [in] cfg     RPK configuration to add the pin to.
+ * @param [in] spki    DER-encoded SubjectPublicKeyInfo.
+ * @param [in] spkiSz  Length of spki in bytes.
+ * @return  0 on success.
+ * @return  BAD_FUNC_ARG when cfg or spki is NULL, or spkiSz is 0.
+ * @return  BUFFER_E when no more pins can be stored.
+ * @return  negative hashing error on failure.
+ */
+static int rpk_add_expected(RpkConfig* cfg, const unsigned char* spki,
```
dgarske, line 446: **🔵 [Low] Append-only pin table has no clear/replace API**

The new public pinning APIs are strictly append-only for the lifetime of the WOLFSSL/WOLFSSL_CTX (the diff and the ssl.h header comment acknowledge this). A long-lived CTX that re-pins across peer key rotations will monotonically consume the `WOLFSSL_MAX_RPK_PINS`-entry table and eventually return `BUFFER_E` with no way to reset. This is a reasonable v1 design and it is documented, but it is a usability sharp edge for a new public API: there is no `wolfSSL_CTX_clear_expected_rpk()` companion. Worth a brief mention so it is a conscious choice.

**Fix:** Either accept as a documented limitation (current state) or add a clear API in a follow-up PR. No change required to merge.

src/ssl_api_cert.c
```diff
@@ -430,6 +430,82 @@ int wolfSSL_get_negotiated_server_cert_type(WOLFSSL* ssl, int* tp)
...
+ *
+ * @param [in] ctx     SSL/TLS CTX object.
+ * @param [in] spki    DER-encoded SubjectPublicKeyInfo the peer will present.
+ * @param [in] spkiSz  Length of spki in bytes.
+ * @return  WOLFSSL_SUCCESS on success.
+ * @return  BAD_FUNC_ARG when ctx or spki is NULL, or spkiSz is 0.
+ * @return  BUFFER_E when the pin table is full (WOLFSSL_MAX_RPK_PINS reached).
+ * @return  Other negative error code on hashing failure.
+ */
+int wolfSSL_CTX_set_expected_rpk(WOLFSSL_CTX* ctx, const unsigned char* spki,
```
dgarske, line 475: **🔵 [Low] CTX-level pinning API NULL/arg-error path not directly tested**

`test_tls13_rpk_trust` exercises the argument-error paths (NULL handle, NULL spki, zero length, table-full BUFFER_E) only for the SSL-level `wolfSSL_set_expected_rpk`. The CTX-level `wolfSSL_CTX_set_expected_rpk` is covered only on its success path (via `test_rpk_pin_peer` in `test_rpk_memio_setup`). Its `if (ctx == NULL) return BAD_FUNC_ARG;` branch and the BUFFER_E overflow are never asserted. Both functions share `rpk_add_expected`, so the gap is small, but the CTX wrapper's own NULL guard is untested.

**Fix:** Add a couple of CTX-level argument-error assertions to `test_tls13_rpk_trust` mirroring the SSL-level ones for symmetric coverage.

tests/api.c
```diff
@@ -28058,6 +28058,21 @@ static int error_test(void)
             }
         }
     }
+
+#if defined(OPENSSL_EXTRA)
```
dgarske: **🔵 [Low] RPK reason-string regression check omitted for OPENSSL_EXTRA_X509_SMALL-only builds**

The new explicit check that `wolfSSL_ERR_reason_error_string(WOLFSSL_X509_V_ERR_RPK_UNTRUSTED)` returns "raw public key not trusted" is guarded by `#if defined(OPENSSL_EXTRA)`. However the reason string itself is compiled under the broader guard `OPENSSL_EXTRA || OPENSSL_EXTRA_X509_SMALL || HAVE_WEBSERVER || HAVE_MEMCACHED` (internal.c:27810). In a build that defines OPENSSL_EXTRA_X509_SMALL but not OPENSSL_EXTRA, the new string is present but never exercised by the test. Minor coverage gap; the chosen guard is also consistent with surrounding test code, so this is just an observation.

**Fix:** Broaden the guard to match the string's compilation condition if you want the RPK string covered in X509_SMALL builds; otherwise leave as-is.

wolfssl/openssl/x509.h
```diff
@@ -210,6 +210,10 @@
 #define X509_V_ERR_CA_CERT_MISSING_KEY_USAGE           92
 #define X509_V_ERR_EXTENSIONS_REQUIRE_VERSION_3        93
 #define X509_V_ERR_EC_KEY_EXPLICIT_PARAMS              94
+/* wolfSSL-specific value: this header's numbering already diverges from
```
dgarske: **🔵 [Low] x509.h comment may understate OpenSSL compatibility of value 95**

The comment states the value 95 is "a wolfSSL-specific value ... rather than any particular OpenSSL number." OpenSSL 3.2+ defines `X509_V_ERR_RPK_UNTRUSTED` as 95 in its own RPK support, so the chosen value actually appears to align with OpenSSL rather than being arbitrary. If that alignment was intentional (which is good for compatibility), the comment is slightly misleading; if coincidental, it's still worth noting the match exists. Low confidence since I did not cross-check against a vendored OpenSSL header in this tree.

**Fix:** Verify OpenSSL's assigned value for X509_V_ERR_RPK_UNTRUSTED; if it is 95, reword the comment to note the value is intentionally OpenSSL-compatible.

dgarske (commented):
## Skoll Code Review

### Posted findings
- [Medium] **Nested '*/' inside doxygen \code block truncates the comment** — `doc/dox_comments/header_files/ssl.h:16038,16074`
- [Medium] **RPK now fails closed for all HAVE_RPK builds — notable behavior change** — `src/internal.c:16884-16911`
- [Low] **Append-only pin table has no clear/replace API** — `src/ssl_api_cert.c:446-507`
- [Low] **CTX-level pinning API NULL/arg-error path not directly tested** — `src/ssl_api_cert.c:475-485`
- [Low] **RPK reason-string regression check omitted for OPENSSL_EXTRA_X509_SMALL-only builds** — `tests/api.c:28062-28075`
- [Low] **DANE removal leaves useDANE field with no remaining reader/writer** — `wolfssl/internal.h:5301-5303`
- [Low] **x509.h comment may understate OpenSSL compatibility of value 95** — `wolfssl/openssl/x509.h:213-215`

dgarske (changes requested):
Please review these skoll reports -> https://github.com/wolfSSL/wolfssl/pull/10486#pullrequestreview-4487509348

dgarske (commented):
## Skoll Code Review

### Posted findings
- [Low] **useDANE bitfield is now dead after removing its only reader** — `src/internal.c:16933-16940 (field at wolfssl/internal.h:5301-5303)`

dgarske (dismissed):

embhorn:
retest this please

dgarske (approved):

Frauschi (approved):

PR #10495, opened 2026-05-18
Add PIC32MZ emulator tests

LinuxJedi:
Adds PIC32MZ EC / EF emulators with hardware acceleration. Also fix a stack pointer free bug in the PIC32 hardware acceleration code.


JacobBarthelmeh (approved):
New PIC32 test ran for 19m. Just making note of that, wish it was a faster test but the additional coverage for pic32 builds is nice to have. Thanks!

PR #10505, opened 2026-05-20
Fix hashType comparison in SLH-DSA

Frauschi:
Fix for `Werror=type-limit` errors in wolfssl-multi-test.


douzzer (approved):
reproduced and tested with `cross-armv7m-armasm-thumb-fips-140-3-dev-sp-asm-all-crypto-only`

PR #10507, opened 2026-05-20
CI Optimizations

dgarske:
# Description

- Skip CI for draft PRs and redundant master-push re-runs; membrowse nightly.
- Add smoke test (8 configs, AddressSanitizer, CFLAGS=-Werror, post-merge tree, fail-fast on conflicts).
- Add wait-for-smoke composite action for downstream CI gating.
- Add check-source-text + bash -n + shellcheck workflow (script in make dist).
- Cache apt-get update in install-apt-deps composite on cache hit.
- Add check-headers workflow: every public wolfssl/*.h compiles standalone.
- Remove no-tls.yml (its single config is already covered by os-check.yml).
 
# Testing

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


Makefile.am
```diff
@@ -149,6 +149,7 @@ EXTRA_DIST+= wolfssl64.sln
 EXTRA_DIST+= valgrind-error.sh
 EXTRA_DIST+= valgrind-bash.supp
 EXTRA_DIST+= fips-hash.sh
+EXTRA_DIST+= .github/scripts/check-source-text.sh
```
kareem-wolfssl, line 152: Do you not need to add .github/scripts/check-headers.sh here as well?

kareem-wolfssl (changes requested):

Makefile.am
dgarske: All items in .github are not part of the dist bundle. So its on purpose. Thanks. Okay as is.

kareem-wolfssl (approved):

PR #10511, opened 2026-05-21
Fix minor compile error with `WOLFSSL_DEBUG_MEMORY`

dgarske:
# Description

Fix minor compile error with `WOLFSSL_DEBUG_MEMORY`
Fixes ZD21780

# Testing

How did you test?

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/wolfcrypt/src/memory.c
+++ b/wolfcrypt/src/memory.c
@@ -1184,14 +1184,13 @@ void wolfSSL_Free(void *ptr, void* heap, int type)
 #endif
 {
     int i;
-    wc_Memory* pt = NULL;
 
     if (ptr) {
         /* check for testing heap hint was set */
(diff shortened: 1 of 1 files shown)


wolfcrypt/src/memory.c
```diff
@@ -1316,14 +1316,14 @@ void wolfSSL_Free(void *ptr, void* heap, int type)
                 }
             }
         #endif
+            (void)pt;
```
SparkiDev: Shouldn't be needed.
Allocation is preprocessor unconditional.
Should be using the allocated memory after that!

SparkiDev (dismissed):

JacobBarthelmeh (approved):

PR #10516, opened 2026-05-22
Finalize ML-DSA renaming

Frauschi:
Final part of the ML-DSA renaming, following #10436 and #10497. This updates all the call sites within the library to the new `wc_MlDsaKey` API and renames further internal identifiers.

The ML-DSA tests are migrated to the new API as well, but an additional set of tests ensures that the legacy shim compatibility layer for the old names works.


Frauschi:
Jenkins retest this please

dgarske (approved):

PR #10518, opened 2026-05-22
Missing compatibility layer functionality 

Roy-Carter:
# Description
Implement needed as part of migrating from OpenSSL -> WolfSSL
wolfSSL_SSL_CIPHER_find - find cipher by 2 bytes in wired like openssl
wolfSSL_sk_SSL_CIPHER_delete - remove cipher at given index
SSL_clear_chain_certs

# Testing

created unitests .

# Checklist

 - [*] added tests
 - [*] updated/added doxygen


Roy-Carter:
@julek-wolfssl adding PR for the 3  functions i've seen problems with (missing functionality) lmk what you think

Roy-Carter:
Hey @julek-wolfssl before I fix anything here since I wasn't 100% sure this is correct implementation I'd like if you can verify on your end if it's fine or not before I continue editing this

dgarske (changes requested):
@Roy-Carter looks like issue with `--enable-asynccrypt` "FAIL: scripts/unit.test"

src/ssl.c
```diff
@@ -10997,6 +10992,49 @@ const WOLFSSL_CIPHER* wolfSSL_get_cipher_by_value(word16 value)
...
+    defined(WOLFSSL_HAPROXY) || defined(OPENSSL_EXTRA) || defined(HAVE_LIGHTY)
+/* Locate a cipher in the SSL's cipher list by 2-byte wire-format suite id.
+ *
+ * Mirrors OpenSSL's SSL_CIPHER_find(ssl, ptr). The two bytes pointed to by
+ * ptr are the on-the-wire cipher suite identifier (ptr[0] is the high byte,
+ * ptr[1] is the low byte). Lookup is restricted to ciphers in ssl's cipher
+ * list, matching OpenSSL semantics.
+ *
+ * Returned pointer references storage owned by the SSL object's internal
+ * cipher list; callers must not free it. It remains valid until SSL_free.
```
julek-wolfssl: This entire comment except maybe the first line is redundant.

src/ssl_sk.c
```diff
@@ -1205,6 +1205,54 @@ void wolfSSL_sk_SSL_CIPHER_free(WOLF_STACK_OF(WOLFSSL_CIPHER)* sk)
...
+    /* Walk to the node so we can capture its inline cipher value before the
+     * pop_node call frees the underlying memory. */
+    node = sk;
+    {
+        int i;
+        for (i = 0; i < idx && node != NULL; i++)
+            node = node->next;
+    }
+    if (node == NULL)
+        return NULL;
```
julek-wolfssl: Just use `wolfSSL_sk_SSL_CIPHER_value`.

julek-wolfssl (changes requested):
I would like a test to check that the connection still completes after a call to `SSL_clear_chain_certs`.

Roy-Carter:
> I would like a test to check that the connection still completes after a call to `SSL_clear_chain_certs`.

Regarding this without the PR fix of https://github.com/wolfSSL/wolfssl/pull/10517 there's a segfault in clear , so just for the sake of the test + fix i've added the two bugs I encountered as part of the  test because  wolfSSL_add0_chain_cert never increments certChainCnt (CTX version does), and clear_chain_certs never resets it thus TLS 1.3 send walks a freed/NULL
  chain. so maybe worth just closing the other one and sticking to this one for the addition of clearing also.

src/ssl.c
```diff
@@ -10997,6 +10997,44 @@ const WOLFSSL_CIPHER* wolfSSL_get_cipher_by_value(word16 value)
...
+
+    for (node = sk; node != NULL; node = node->next) {
+        if (node->data.cipher.cipherSuite0 == ptr[0] &&
+            node->data.cipher.cipherSuite  == ptr[1]) {
+            return &node->data.cipher;
+        }
+    }
+
+    return NULL;
+}
```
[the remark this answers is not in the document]
julek-wolfssl: @Roy-Carter this appears to be correct. Please address this.

julek-wolfssl (changes requested):
Please address the copilot comment and fix the CI. Thanks.

Roy-Carter:
@julek-wolfssl can we re run workflow see if its working now ?

julek-wolfssl (changes requested):
@Roy-Carter just some simple build errors to fix.

src/ssl_sk.c
```diff
@@ -1205,6 +1205,44 @@ void wolfSSL_sk_SSL_CIPHER_free(WOLF_STACK_OF(WOLFSSL_CIPHER)* sk)
...
 }
+
+/* Remove the cipher at the given index from the stack.
+ *
+ * @param [in,out] sk   Stack of ciphers.
+ * @param [in]     idx  Index of cipher to remove.
+ * @return  Heap copy of removed cipher on success.
+ * @return  NULL on failure.
+ */
+WOLFSSL_CIPHER* wolfSSL_sk_SSL_CIPHER_delete(
```
dgarske, line 1225: **🟠 [Medium] wolfSSL_sk_SSL_CIPHER_delete returns heap-allocated copy that caller must free, diverging from OpenSSL contract**

Because wolfSSL stores `WOLFSSL_CIPHER` inline in each stack node (no stable element pointer to hand back), this implementation `XMALLOC`s a copy of the removed cipher and returns it. The function is wired to the OpenSSL name via `#define sk_SSL_CIPHER_delete wolfSSL_sk_SSL_CIPHER_delete`. In OpenSSL, `sk_SSL_CIPHER_delete()` returns a pointer to a static/library-owned `SSL_CIPHER` that the caller must NOT free. Real-world code ported from OpenSSL routinely discards the return value of `sk_SSL_CIPHER_delete()`; against this implementation every such call leaks one `WOLFSSL_CIPHER`. The doxygen says "Heap copy of removed cipher on success" but never states the caller must `XFREE(..., DYNAMIC_TYPE_OPENSSL)` it. The unit test does free it, masking the leak from CI. This is an ownership-contract mismatch that will silently leak in migrated applications.

**Fix:** At minimum, document the caller-must-free ownership clearly in the doxygen (the current comment does not). Consider whether returning a heap copy is the right contract for an OpenSSL drop-in, since callers that mirror OpenSSL usage will leak. If callers in the migration target only need the side effect (removal), document that the return value must be freed and add a note to the ChangeLog.

src/ssl.c
```diff
@@ -10997,6 +10997,70 @@ const WOLFSSL_CIPHER* wolfSSL_get_cipher_by_value(word16 value)
     return cipher;
 }
 
+#if defined(OPENSSL_ALL) || defined(WOLFSSL_NGINX) || \
```
dgarske, line 6204: **🔵 [Low] HAVE_LIGHTY guard on wolfSSL_SSL_CIPHER_find is broader than its dependency wolfSSL_get_ciphers_compat**

`wolfSSL_SSL_CIPHER_find` is guarded by `OPENSSL_ALL || WOLFSSL_NGINX || WOLFSSL_HAPROXY || OPENSSL_EXTRA || HAVE_LIGHTY` and calls `wolfSSL_get_ciphers_compat()`, which is guarded by `OPENSSL_EXTRA || OPENSSL_ALL || WOLFSSL_NGINX || WOLFSSL_HAPROXY` (ssl.c:16544) -- it does NOT include `HAVE_LIGHTY`. In configure-driven builds this is harmless because `--enable-lighty` also defines `OPENSSL_EXTRA` and `OPENSSL_ALL` (configure.ac:8945-8955), so the dependency is always present. But a user who manually defines only `HAVE_LIGHTY` (without OPENSSL_EXTRA/OPENSSL_ALL) would compile `wolfSSL_SSL_CIPHER_find` while `wolfSSL_get_ciphers_compat` is excluded, producing an undefined-reference link error.

**Fix:** Align the `wolfSSL_SSL_CIPHER_find` preprocessor guard (in src/ssl.c and the matching wolfssl/ssl.h declaration) with `wolfSSL_get_ciphers_compat`'s guard to avoid a potential link error on manually-configured HAVE_LIGHTY-only builds.

src/ssl_load.c
```diff
@@ -5220,6 +5220,8 @@ int wolfSSL_add0_chain_cert(WOLFSSL* ssl, WOLFSSL_X509* x509)
         if (ret == 1) {
             /* We now own cert chain. */
             ssl->buffers.weOwnCertChain = 1;
+            /* Account for the certificate just added to the chain. */
```
dgarske, line 5484: **🔵 [Low] certChainCnt++ in add0_chain_cert is unconditional while the CTX equivalent guards it with WOLFSSL_TLS13**

The new `ssl->buffers.certChainCnt++;` is unconditional, whereas the analogous CTX-side helper `wolfssl_ctx_add_to_chain` (ssl_load.c:4964-4967) increments `ctx->certChainCnt++` only inside `#ifdef WOLFSSL_TLS13`. The unconditional SSL-side increment is actually the more correct choice -- it keeps `certChainCnt` in sync with the number of certs in `ssl->buffers.certChain` for all paths (TLS 1.3 send gating at tls13.c:9660, the per-cert extension sizing at tls13.c:9651, and the OCSP-multi cap at internal.c:26141), and the field exists unconditionally in the struct. Flagging only for awareness: the two sibling code paths now treat the same counter differently. Confirm this asymmetry is intentional (SSL path always counts, CTX path counts only under TLS 1.3).

**Fix:** Keep the unconditional SSL-side increment (it is correct), but consider removing the `#ifdef WOLFSSL_TLS13` guard around the CTX-side `ctx->certChainCnt++` so the two paths are consistent. No functional change required for this PR.

src/ssl_sk.c
```diff
@@ -1205,6 +1205,44 @@ void wolfSSL_sk_SSL_CIPHER_free(WOLF_STACK_OF(WOLFSSL_CIPHER)* sk)
...
+    WOLF_STACK_OF(WOLFSSL_CIPHER)* sk, int idx)
+{
+    WOLFSSL_CIPHER* ret = NULL;
+    WOLFSSL_CIPHER* cipher;
+
+    WOLFSSL_ENTER("wolfSSL_sk_SSL_CIPHER_delete");
+
+    if (sk == NULL || idx < 0)
```
dgarske: **🔵 [Low] Delete relies on sk_value bounds while removal relies on node-num; mismatch on empty-placeholder cipher stack**

Bounds checking in `wolfSSL_sk_SSL_CIPHER_delete` uses `wolfSSL_sk_SSL_CIPHER_value()` (which walks the node list and ignores `sk->num`), while the actual removal uses `wolfSSL_sk_pop_node()` (which early-returns when `sk->num == 0`). For a normal non-empty stack these agree. But for the wolfSSL "empty" cipher-stack representation -- a single placeholder head node with `num == 0` -- `wolfSSL_sk_SSL_CIPHER_value(sk, 0)` returns `&head.data.cipher` (non-NULL stale data), so the function would `XMALLOC` and return a bogus copy while `wolfSSL_sk_pop_node` removes nothing (num stays 0). Current callers do not hit this (get_ciphers_compat returns NULL for empty stacks, and the test uses a non-empty dup), so it is latent rather than active.

**Fix:** Add an explicit upper-bound check against `wolfSSL_sk_SSL_CIPHER_num(sk)` so the function behaves correctly even for a zero-count placeholder stack, keeping the bounds notion consistent with the removal path.

dgarske (commented):
## Skoll Code Review

### Posted findings
- [Medium] **wolfSSL_sk_SSL_CIPHER_delete returns heap-allocated copy that caller must free, diverging from OpenSSL contract** — `src/ssl_sk.c:1216-1245`
- [Medium] **Test coverage gap: only head-index delete is exercised; SSL_CIPHER_find untested under OPENSSL_EXTRA/NGINX-only builds** — `tests/api.c:16265-16282, tests/api.c:16295-16366`
- [Low] **HAVE_LIGHTY guard on wolfSSL_SSL_CIPHER_find is broader than its dependency wolfSSL_get_ciphers_compat** — `src/ssl.c:11000-11062`
- [Low] **certChainCnt++ in add0_chain_cert is unconditional while the CTX equivalent guards it with WOLFSSL_TLS13** — `src/ssl_load.c:5223-5224`
- [Low] **Delete relies on sk_value bounds while removal relies on node-num; mismatch on empty-placeholder cipher stack** — `src/ssl_sk.c:1224-1242`

Roy-Carter:
can we re-run workflow ? @dgarske @julek-wolfssl

julek-wolfssl (dismissed):

dgarske (changes requested):
Please resolve merge conflicts. Sorry for the delay on this.

dgarske (approved):

philljj (approved):

PR #10519, opened 2026-05-22
CI Fix: fix flaky ECH test

sebastian-carpenter:
# Description

`test_wolfSSL_Tls13_ECH_ch2_decrypt_error` flakes out after a couple thousand iterations. Previously this test was seeking the first `0xFE0D` bytes (TLSXT_ECH) and would then tamper with the 14th byte after that marker. Every couple thousand iterations `0xFE0D` would show up before the extension marker and cause an incorrect failure.

Updated the test to iterate along the extensions list and find TLSXT_ECH.

# Testing

Reproduced the error three times within 10000 iterations. Tested the fix with 20000 iterations.

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/tests/api.c
+++ b/tests/api.c
@@ -16027,6 +16027,57 @@ static int test_wolfSSL_Tls13_ECH_long_SNI(void)
     return EXPECT_RESULT();
 }
 
+static int ech_seek_extensions(byte* buf, word16* innerExtLen)
+{
+    word16 idx;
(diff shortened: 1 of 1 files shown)


sebastian-carpenter:
Jenkins retest this please.

JacobBarthelmeh (approved):

PR #10520, opened 2026-05-22
Implementing missing macros to openssl compatibility layer 

Roy-Carter:
# Description
Implementing as part of migration to wolfssl
SSL_set_ciphersuites
X509_add1_ext_i2d
sk_GENERAL_NAME_new_null
EC_KEY_oct2key

# Testing

unitests to each functionality added

# Checklist

 - [ *] added tests
 - [ *] updated/added doxygen


Roy-Carter:
@julek-wolfssl second part of the functions i've implemented , split the PR's for clarity of checks .

Roy-Carter:
@dgarske errors seemed related to master diff. both tests worked for me locally , I rebased , can we re-run workflow?

Roy-Carter:
@julek-wolfssl @dgarske I've rebased to master , seems related to new changes added .

dgarske:
Jenkins okay to test

julek-wolfssl (changes requested):
@Roy-Carter please address copilot review. Thanks.

Roy-Carter:
@julek-wolfssl can we re-run workflow ?

julek-wolfssl (dismissed):

julek-wolfssl:
```
$ ./autogen.sh && ./configure --enable-all && make check
...
FAILURES:
   871: test_wolfSSL_X509_add_ext
   929: test_wolfSSL_X509_add1_ext_i2d_flags

 End API Tests
 Failed/Skipped/Passed/All: 2/182/1432/1616
```

Roy-Carter:
can we re-run workflow @julek-wolfssl 

also writing here about our ticket , I think it might've been closed / I lost perms to write on it even though I did send emails there it seems as if zendesk link is unavailable

tests/api.c
```diff
@@ -2522,6 +2522,27 @@ static int test_wolfSSL_set_cipher_list_exclusions(void)
...
+    EXPECT_DECLS;
+#if defined(OPENSSL_EXTRA) && defined(WOLFSSL_TLS13) && \
+    !defined(NO_WOLFSSL_CLIENT) && defined(HAVE_AESGCM)
+    WOLFSSL_CTX* ctx = NULL;
+    WOLFSSL*     ssl = NULL;
+
+    ExpectNotNull(ctx = SSL_CTX_new(wolfSSLv23_client_method()));
+    ExpectNotNull(ssl = SSL_new(ctx));
+
+    ExpectIntEQ(SSL_set_ciphersuites(ssl, "TLS_AES_128_GCM_SHA256"), 1);
```
dgarske: With ` --disable-errorstrings` 

```
   1309: test_wolfSSL_SSL_set_ciphersuites                  :
ERROR - tests/api.c line 2537 failed with:
    expected: wolfSSL_set_cipher_list(ssl, "TLS_AES_128_GCM_SHA256") == 1
    result:   0 != 1
 failed (  0.00002)
ERROR - tests/api.c line 36346 failed with:
    expected: Test failed
    result:   ret 0
```

dgarske (changes requested):

julek-wolfssl:
retest this please.

julek-wolfssl:
retest this please.

src/pk_ec.c
```diff
@@ -4716,6 +4716,64 @@ int wolfSSL_EC_KEY_set_public_key(WOLFSSL_EC_KEY *key,
...
+ * Return code compliant with OpenSSL.
+ *
+ * @param [in, out] key  EC key (must already have a group set).
+ * @param [in]      buf  Octet-encoded public point.
+ * @param [in]      len  Length of @buf in bytes.
+ * @param [in]      ctx  BN context. May be NULL.
+ * @return  1 on success.
+ * @return  0 on failure.
+ */
+int wolfSSL_EC_KEY_oct2key(WOLFSSL_EC_KEY *key, const unsigned char *buf,
```
dgarske, line 4768: **🟠 [Medium] wolfSSL_EC_KEY_oct2key does not record the point conversion form**

OpenSSL's `EC_KEY_oct2key()` saves the encoding form on the key (`key->conv_form = (point_conversion_form_t)(buf[0] & ~0x01)`), so that a later `i2o_ECPublicKey()` / `EC_KEY_key2buf()` re-emits the same form. The wolfSSL implementation never touches `key->form`, so a compressed input silently round-trips back as uncompressed.

Verified empirically on this branch:
```
compressed enc len=33 first=0x03
oct2key(compressed)=1 conv_form after=4 (OpenSSL sets 2)
```
This matters for applications that hash or compare the re-encoded public key (e.g. JWK/COSE key thumbprints, TLS raw public keys), because they will get a different byte string than OpenSSL would.

**Fix:** Call `wolfSSL_EC_KEY_set_conv_form(key, buf[0] & ~0x01)` on success, and document the behaviour in the function's doxygen block.

src/x509.c
```diff
@@ -3585,6 +3598,211 @@ WOLFSSL_X509_EXTENSION *wolfSSL_X509V3_EXT_i2d(int nid, int crit,
     return NULL;
 }
 
+/* Return 1 if an extension of type @nid is already present in the in-memory
```
dgarske: **🟠 [Medium] wolfssl_x509_ext_is_set() silently reports 'absent' for every unrecognized NID**

The `default:` arm returns 0 for any NID outside the six recognized ones. `wolfSSL_X509V3_EXT_i2d()` also accepts `WC_NID_issuer_alt_name` and `WC_NID_info_access`, and with `WOLFSSL_CUSTOM_OID` `wolfSSL_X509_add_ext()` will happily store arbitrary NIDs in `x509->custom_exts[]`. For those NIDs the presence-dependent operations quietly do the wrong thing rather than reporting that they are unsupported:

- `X509V3_ADD_DEFAULT` never detects a duplicate, so repeated calls keep appending custom extensions until `NUM_CUSTOM_EXT` is exhausted (then `wolfssl_x509_add_custom_ext()` fails with a confusing 'Bad value for customExtCount').
- `X509V3_ADD_KEEP_EXISTING` always adds instead of keeping.
- `X509V3_ADD_DELETE` / `X509V3_ADD_REPLACE_EXISTING` always fail with 'No extension to delete', which is indistinguishable from the genuinely-absent case.

None of these NIDs are covered by the new tests.

**Fix:** Have `wolfssl_x509_ext_is_set()` report 'unknown NID' distinctly and make `wolfSSL_X509_add1_ext_i2d()` fail the presence-dependent operations (DEFAULT / KEEP_EXISTING / REPLACE_EXISTING / DELETE) for those NIDs, or at minimum document the limitation in the doxygen block alongside the extKeyUsage note. Add a test for `NID_info_access` / a custom OID.

src/x509.c
```diff
@@ -1474,6 +1474,17 @@ int wolfSSL_X509_add_ext(WOLFSSL_X509 *x509, WOLFSSL_X509_EXTENSION *ext,
                 return WOLFSSL_FAILURE;
             }
         }
+        else if (ext && ext->ext_sk != NULL) {
```
dgarske, line 1625: **🟠 [Medium] wolfSSL_X509_add_ext() now fails for extKeyUsage object stacks that previously returned success**

`wolfSSL_X509_add_ext(x509, ext, -1)` with an extKeyUsage extension carrying an `ext_sk` object stack used to return `WOLFSSL_SUCCESS` (a silent no-op); it now returns `WOLFSSL_FAILURE`. Turning a silent no-op into an explicit failure is the right call, but it is a visible behaviour change for existing OPENSSL_ALL applications that build an EKU extension with `X509V3_EXT_i2d(NID_ext_key_usage, ...)` and then `X509_add_ext()` it - those calls will now start failing.

Blast radius checked: `wolfSSL_X509_set_ext()` (the source of extensions returned by `X509_get_ext()`) only sets `ext_sk` for `AUTH_INFO_OID` and `ALT_NAMES_OID` and always populates `ext->value.data`, so the common 'copy extensions from one cert to another' pattern still takes the first branch and is unaffected. `tests/api/test_ossl_x509_ext.c:391` (`No Data - no change` -> SUCCESS) also still passes because that ext has neither data nor `ext_sk`.

The underlying gap remains: wolfSSL cannot consume an EKU object stack at all. The OIDs in the stack could be mapped to the `EXTKEYUSE_*` bits the same way `ParseExtKeyUsageStr()` maps names.

**Fix:** Either implement the OID -> `EXTKEYUSE_*` mapping so extKeyUsage actually works through `X509_add1_ext_i2d()`, or call out the behaviour change in the PR description / release notes so downstream users of `X509_add_ext()` are not surprised.

src/x509.c
```diff
@@ -3585,6 +3598,211 @@ WOLFSSL_X509_EXTENSION *wolfSSL_X509V3_EXT_i2d(int nid, int crit,
...
+
+    if (x == NULL) {
+        WOLFSSL_MSG("Bad parameter");
+        return WOLFSSL_FAILURE;
+    }
+
+    op = flags & WOLFSSL_X509V3_ADD_OP_MASK;
+    exists = wolfssl_x509_ext_is_set(x, nid);
+
+    switch (op) {
```
dgarske, line 3898: **🔵 [Low] X509V3_ADD_SILENT has no observable effect**

`WOLFSSL_X509V3_ADD_SILENT` only gates `WOLFSSL_MSG()` calls, which are already compiled out unless `DEBUG_WOLFSSL` is defined, and wolfSSL does not push to an OpenSSL-style error queue here. The three `if ((flags & WOLFSSL_X509V3_ADD_SILENT) == 0)` wrappers therefore add branching and indentation for no behavioural difference. The doxygen block already states that SILENT does not change the result, so a single note would be clearer than three dead conditionals.

**Fix:** Either drop the three SILENT conditionals and note once that the flag is accepted-but-inert, or keep them and add a brief comment explaining they are placeholders for a future error-queue implementation.

dgarske (commented):
## Skoll Code Review

**Findings:** 13 total — 4 posted, 9 skipped

### Posted findings
- [Medium] **wolfSSL_EC_KEY_oct2key does not record the point conversion form** — `src/pk_ec.c:4731-4775`
- [Medium] **wolfssl_x509_ext_is_set() silently reports 'absent' for every unrecognized NID** — `src/x509.c:3601-3622`
- [Medium] **wolfSSL_X509_add_ext() now fails for extKeyUsage object stacks that previously returned success** — `src/x509.c:1477-1488`
- [Low] **X509V3_ADD_SILENT has no observable effect** — `src/x509.c:3740-3778`

### Skipped findings
- [High] ~~X509V3_ADD_REPLACE destroys the existing extension when the subsequent add fails~~
- [High] ~~basicConstraints pathLenConstraint is taken from ASN1_INTEGER-length (DER size), not the integer value~~
- [High] ~~Removing subjectAltName leaves the altName iterator permanently exhausted~~
- [Medium] ~~basicConstraints test does not verify the pathlen round-trip~~
- [Medium] ~~EC_KEY_oct2key test only covers the uncompressed form and hardcodes the P-256 size~~
- [Medium] ~~X509V3_ADD_APPEND does not append for most NIDs~~
- [Low] ~~Flag test computes dnsName/dnsLen but never asserts on them for APPEND and KEEP_EXISTING~~
- [Low] ~~SSL_set_ciphersuites alias accepts TLS 1.2 cipher names~~
- [Low] ~~test_san_first_dns() truncates into a static buffer while reporting the untruncated length~~

dgarske (changes requested):
https://github.com/wolfSSL/wolfssl/pull/10520#pullrequestreview-4848742438

julek-wolfssl:
Retest this please

julek-wolfssl:
Retest this please

julek-wolfssl:
retest this please

julek-wolfssl:
@dgarske addressed review

julek-wolfssl:
Not sure why this got closed. Re-opened as https://github.com/wolfSSL/wolfssl/pull/11375 to make it easier for me to track.

PR #10527, opened 2026-05-25
Allow --enable-writedup when DTLS is disabled

mattia-moffa:
# Description

Adds some include guards so that the writedup test still compiles when DTLS is disabled.

# Testing

```
./configure --enable-writedup
make
make check
```


mattia-moffa:
Jenkins retest this please

JacobBarthelmeh (approved):

PR #10530, opened 2026-05-26
RISC-V ASM unaligned read/writes: alternative assembly

### LINKED ISSUE #10525: [Bug]: Several Issues with respect to alignment for RISC-V assembly routines
### Contact Details

alex@minres.com

### Version

5.9.1 (commit 6a3eb6f)

### Description

RISC-V systems may, but do not need to support unaligned loads and stores. Concretely,  there are systems that expect `sw / lw`  and `sd / ld` to be 4 and 8 byte aligned, respectively. Misaligned loads and stores can then raise an exception.

I ran into a few occasions where alignment was not given.

The first issue I ran into was that there are no alignment requirements for the `secret` and `masterSecret` byte Arrays in the `Arrays` struct. These happened to lay at 1 byte alignment for me and were propagated to some RISC-V assembly, which expected 4 or 8 byte alignment.
I managed to work around this locally by enabling `WOLFSSL_USE_ALIGN` and prepending the corresponding define:

```
<internal.h>
typedef struct Arrays {
...
#ifdef WOLFSSL_TLS13
    ALIGN8 byte            secret[SECRET_LEN];
#endif
#ifdef HAVE_KEYING_MATERIAL
    byte            exporterSecret[WC_MAX_DIGEST_SIZE];
#endif
    ALIGN8 byte            masterSecret[SECRET_LEN];
...
} Arrays;

```

The one I cannot solve on my own is from the following call stack (breakpoint on the faulting instruction):

Image

In this very case `createSigData` propagates an address in `ret = GetMsgHash(ssl, &sigData[idx]);` with idx equal to 92. `sigData` is 4 byte aligned, an offset of 92 is then only 2 byte aligned. This then faults when it propagates the call stack down to this [store instruction](https://github.com/wolfSSL/wolfssl/blob/8597d9da83579273df53744340022063b66144e7/wolfcrypt/src/port/riscv/riscv-64-sha256.c#L856).

I have tested the standalone wolfcrypt API for RISC-V assembly successfully, but I encounter these issues in the TLS1.3 application I am trying to run. My application runs fine when not enabling the RISC-V assembly.

This is not an exhaustive list but shows what I encountered so far, basically stepping through the build-up of the TLS connection. I encountered the first error when keys were generated, the second one when validating certificates.

Let me know what more I can provide to help you get this resolved.

-Alex

SparkiDev:
# Description

Not all RISC-V chips allow unaligned reads and writes with basic assembly instructions like lw/sw.
Add alternative assembly that is turned on with:
WOLFSSL_RISCV_ASM_NO_UNALIGNED.

Fixes #10525

# Testing

./configure --disable-shared LDFLAGS=--static --host=riscv64 CC=riscv64-linux-gnu-gcc --enable-riscv-asm CFLAGS=-DWOLFSSL_RISCV_ASM_NO_UNALIGNED


SparkiDev:
Jenkins: retest this please

EAlexJ:
I would like to push back on this implementation a bit.

Currently, all instructions that could be unaligned get emulated.

I think it makes sense to check if the pointers are actually unaligned and only then use the more costly emulation.
I suggest something like this (with the check guarded behind `WOLFSSL_RISCV_ASM_NO_UNALIGNED`):

```
Sha512Transform(wc_Sha512* sha512, const byte* data,
    word32 blocks)
{
if((uintptr_t) data % 8) // modulo 8 as the routine uses UNALIGNED_[LS]D
    <emulate using the UNALIGNED_* macro>
else
    <normal execution>
}
```
There should only be a negligible impact on performance in case the data is aligned. This also only introduces a small size overhead, the check should be a few instructions at most.

EAlexJ:
On a separate issue: In my opinion readability would be increased if you just used the `UNALIGNED_*` Macro unconditionally and only check once in the actual definition of the macro if `WOLFSSL_RISCV_ASM_NO_UNALIGNED` is defined and then emit the emulation. 
This is then similar to how `REV8` and others are handled. It would then also make sense to name the macro differently.

EDIT: This is probably obsolete in case the suggestion above gets adopted

wolfcrypt/src/port/riscv/riscv-64-aes.c
EAlexJ, line 1: I think you missed some in
```
int wc_AesSetKey(Aes* aes, const byte* key, word32 keyLen, const byte* iv,
    int dir)
{
```

The pointer to key is not necessarily aligned (Found it out the hard way)

EAlexJ, line 1: Also, why did you choose to use ALIGN16, is ALIGN8 not already sufficient?

SparkiDev: Because they are 16 byte buffers in AES, I made it align on 16 bytes.
Looking into it, the vector instructions are 64 bit loads and stores so changing to ALIGN8 will be fine.

SparkiDev: Are you saying then that none of the fields of Aes will be aligned?
If so, then I will need to change the access to the fields: key, reg and tmp.

EAlexJ: That I do not know, the best solution for me is to align the buffers at its source, so I went into the ssl struct and corrected the placement of `byte* key`.
This was sufficient, so in this very case only key seems to not be aligned by default

SparkiDev:
I've modified the macros to check the alignment and choose the sequence to use.
Let me know if that is better.

Thanks,
Sean

wolfssl/wolfcrypt/port/riscv/riscv-64-asm.h
EAlexJ, line 1: I see the following pattern several times:

```
addi    t, p, o          
andi    t, t, <alignment mask>           
bnez    t, <label>
----
t: scratch register
p: register holding the base address
o: offset
```
This checks the alignment of p+o, but since o is always a valid offset (is it not?) it is sufficient to check p only.
One instruction can then be saved by doing:
```          
andi    t, p, <alignment mask>           
bnez    t, <label>
```

EAlexJ, line 1: One issue I see is that there is a lot of double checking for alignments:
The bulk variants call the underlying N times, but since the check for the alignment is done in the underlying only, it also gets checked N times.

In case the data is actually aligned there should only be one check.

I'd argue that most of the time the data is (or at the very least should be) aligned, so I'd try to keep the overhead for this case as small as possible.

If the data is unaligned, performance will take a hit even on hardware that supports misaligned loads and stores, e.g. when accesses cross cache-line boundaries.

wolfcrypt/src/port/riscv/riscv-64-aes.c
SparkiDev: OK. The answer is to require alignment.

wolfssl/wolfcrypt/port/riscv/riscv-64-asm.h
SparkiDev: Made it that it will do many at once after the check to reduce the overhead.
Will test for largest alignment first and use the smallest number of instructions.

SparkiDev: fixed

dgarske (approved):
Tests pass. Thank you

Issue #60, opened 2015-03-30
Missing strnlen() on MingW


gvanem:
MingW doesn't have `strnlen()` which is needed in sniffer.c + snifftest.c. To solve it, I added this to my `config.h`:
```
    extern __inline__ size_t strnlen (const char *str, size_t maxsize)
    {
      size_t n;
      for (n = 0; n < maxsize && *str; n++, str++)
        ;
      return (n);
    }
```
Maybe the missing functions (strnlen, strsep) could be added to a more suitable place?

ejohnstown:
I replaced the call to `strnlen()` with `strlen()` and did my size check separately. `strnlen()` isn't very portable, as you found, so we shouldn't use it.

PR #10535, opened 2026-05-27
crypto-policies: add MinProtocol, Ciphersuites, and key=value config format

dom-omg:
## Summary

Extends the Fedora system crypto-policy support (added in #8205) to parse config files in the same key=value format as OpenSSL's `opensslcnf.config` backend, and to **actually call wolfSSL API to enforce every policy parameter** read from the file.

### What was missing (fedora-crypto-policies work item #60)

The original implementation only parsed a single-line `@SECLEVEL=N:cipher_string` format and applied the cipher list. It did not:
- Parse an explicit `MinProtocol = TLSv1.2` key
- Parse a `Ciphersuites = ...` key for TLS 1.3 suites
- Use the same config format as `opensslcnf.config` so that `fedora-crypto-policies` can generate a `wolfssl.config` with the same tooling

### Changes

**`wolfssl/internal.h`**
- Add `explicitMinDowngrade` and `explicitMinDtlsDowngrade` fields to `SystemCryptoPolicy` to store values from `MinProtocol` / `DTLS.MinProtocol` keys.
- Add `cipherSuites[]` field for TLS 1.3 `Ciphersuites` key.
- Add `MAX_WOLFSSL_CRYPTO_POLICY_FILE_SIZE = 4096` for multi-line configs.

**`src/ssl.c`**
- `crypto_policy_parse()`: detects legacy `@SECLEVEL=` format (fully backward-compatible) **or** new key=value format.
- New key=value parser handles: `MinProtocol`, `DTLS.MinProtocol`, `CipherString`, `Ciphersuites`. Unknown keys silently ignored.
- `wolfSSL_crypto_policy_enable()`: allocates a temporary buffer with `XMALLOC` so multi-line files up to 4 KiB are accepted.
- `wolfSSL_CTX_new_ex()`: also applies `Ciphersuites` (TLS 1.3) when set.
- New internal accessors: `wolfSSL_crypto_policy_get_min_downgrade()`, `wolfSSL_crypto_policy_get_min_dtls_downgrade()`, `wolfSSL_crypto_policy_get_ciphersuites()`.

**`src/internal.c`**
- `wolfSSL_crypto_policy_init_ctx()`: after deriving `minDowngrade` from `@SECLEVEL`, overrides it with the explicit `MinProtocol` value if one was provided.

**Example files** — updated to new format:
```ini
[system_default_sect]
MinProtocol = TLSv1.2
[...]

PR #10541, opened 2026-05-27
crypto-policies: granular allowlist parser for Fedora system policies

### LINKED ISSUE #9802: Full support of Fedora crypto-policies
### Version

Latest

### Description

https://gitlab.com/redhat-crypto/fedora-crypto-policies/-/issues/60

Is the discussion. Because of lack of full support, the package is missing in Fedora.

max-qlab:
## Summary

Adds a granular allowlist back-end for the system-wide crypto-policy file consumed by `wolfSSL_crypto_policy_enable*()`, in parallel with the existing single-line `@SECLEVEL=N:...` parser. Routing is done by header sniff (`version =` / `override-mode =` / `[section]`) so existing legacy deployments keep working unchanged.

This is the wolfSSL-side counterpart of the format that `fedora-crypto-policies` emits via the `WolfSSLGenerator` Python class (drop-in next to the existing GnuTLS back-end). The companion Python generator is being submitted in parallel against fedora-crypto-policies.

Closes the file-format objection from @asosedkin on fedora-crypto-policies [#60](https://gitlab.com/redhat-crypto/fedora-crypto-policies/-/issues/60) (PR #8205's single-line OpenSSL cipher string judged insufficient) and unblocks wolfSSL [#9802](https://github.com/wolfSSL/wolfssl/issues/9802).

## What changed

| File | Change |
|---|---|
| `src/crypto_policy_granular.{c,h}` | New module: parser + applier (≈720 lines) |
| `src/ssl.c` | `wolfSSL_crypto_policy_enable*()` sniffs the header and routes; `wolfSSL_CTX_new_ex()` calls the applier when granular; `wolfSSL_CTX_SetMinVersion()` guard widened to allow the applier through |
| `src/include.am` | Compile the new file under the existing `!BUILD_CRYPTONLY` guard |
| `tests/api.c` | `test_wolfSSL_crypto_policy_granular`: enable / suite-count / membership / monotonicity / DTLS / format-version / override-mode invariants |
| `examples/crypto_policies/{legacy,default,future}/wolfssl-allowlist.txt` | Real outputs of the Fedora generator, used as test fixtures |
| `examples/crypto_policies/default/wolfssl-allowlist-dtls.txt` | DTLS-only slice (hand-crafted) |
| `examples/crypto_policies/README.md` | Format specification |

The granular applier drives, on every fresh `WOLFSSL_CTX`:

1. `wolfSSL_CTX_SetMinVersion` (from the lowest `enabled-version`)
2. `wolfSSL_CTX_set_cipher_list` (from the `cipher × kx × mac × version` cross-product against the known TLS suite table)
3. `wolfSSL_CTX_UseSupportedCurve` (per `enabled-group`)
4. `wolfSSL_CTX_set1_sigalgs_list` (from the mapped `enabled-sig` set)
5. `wolfSSL_CTX_SetMin{Rsa,Dh,Ecc}Key_Sz` (from `min-rsa-bits` / `min-dh-bits`)

Steps 1 and 4 are best-effort: a build that lacks TLS 1.0 support or rejects an `rsa_pss_*` sigalg the policy lists logs and continues. Cipher list and key-size floors are authoritative.

## Test plan

`make check` on three configurations, all green:

- `--enable-opensslextra --enable-tls13 --enable-arc4 --enable-debug --enable-supportedcurves --enable-ed25519 --enable-ed448 --enable-curve25519 --enable-curve448`: **743 passed / 0 failed / 1507 total**
- Same flags without `--enable-debug` (production build): **738 passed / 0 failed / 1507 total**
- `--disable-dtls` variant: **688 passed / 0 failed / 1507 total** (the DTLS section of our test auto-skips via `#ifdef WOLFSSL_DTLS`)

Zero regression vs `master` baseline. The new `test_wolfSSL_crypto_policy_granular` exercises:

- All three real Fedora fixtures (LEGACY / DEFAULT / FUTURE) parse, apply, and yield a valid `WOLFSSL_CTX` with the expected security level and suite contents.
- Suite-count monotonicity: LEGACY > FUTURE (FUTURE excludes AES-128 so it must derive strictly fewer suites).
- DTLS-only fixture: a pure DTLS 1.2 policy yields a usable DTLS CTX.
- `version = 2` is rejected outright.
- `override-mode = blocklist` is rejected.

## Scope

What this PR does **not** do (deferred):

- Extend the mapping tables beyond the Fedora-relevant primitive set (post-quantum sigalgs like `MLDSA*`, additional cipher suites, `KEM-ECDH`, `HMAC-SHA1`, etc. are parsed and silently ignored when the build does not implement them — explicit support can land per primitive).
- Extend `wolfSSL_crypto_policy_init_ctx()` to derive its min-key floors from the granular struct rather than from `secLevel`. Today the granular path mirrors `security-level` into the legacy `crypto_policy.secLevel` so the existing min-key-size scaffolding in `wolfSSL_crypto_policy_init_ctx()` continues to behave; an explicit pass-through can land in a follow-up.

The exact section/key names (`enabled-version`, `enabled-cipher`, …) are deliberately close to the GnuTLS back-end so the crypto-policies maintainers can adjust them cheaply. That naming call is theirs, on fedora-crypto-policies work item #60.

## Authors

- Maxime Noiseux ([@max-qlab](https://github.com/max-qlab))
- Dominik Blain (Co-Authored-By in each commit; LinkedIn thread with @anthonyhu started this work)


philljj:
Thank you @max-qlab, this is very cool! I'm reviewing now.

philljj:
Hi @max-qlab and @dom-omg, could you email `support@wolfssl.com` for a contributor agreement?

Best,
Jordan

max-qlab:
Hi @philljj, @Copilot — sorry about the noisy initial submission. New commit `0b95179` addresses every piece of feedback we received, and we kept the changes scoped to the granular back-end so the rest of the PR review surface doesn't shift under you.

**What changed**

1. **CI:** `src/crypto_policy_granular.c` now uses `#include <wolfssl/wolfcrypt/libwolfssl_sources.h>` instead of the standalone `config.h` + `settings.h` pair. That was the source of the 75× `make check linux` red on the `TEST_LIBWOLFSSL_SOURCES_INCLUSION_SEQUENCE` matrix (first seen on `CPPFLAGS=-DNO_VERIFY_OID`) and the `-Werror=pedantic` empty-translation-unit error when the module is gated out.

2. **DTLS-only allowlist now derives a non-empty cipher list.** New helper `wcp_protocol_family_enabled()` treats `TLS1.x` and `DTLS1.x` as suite-equivalent at the same minor version, in both directions. Addresses https://github.com/wolfSSL/wolfssl/pull/10541#discussion_r3312977857.

3. **`min_version` is now scoped to the CTX's protocol family.** Added `is_dtls` argument; `apply_granular()` derives it from `ctx->method->version.major == DTLS_MAJOR`. Prevents the silent-drop on `SetMinVersion(WOLFSSL_TLSV1_2)` against a DTLS CTX. Addresses https://github.com/wolfSSL/wolfssl/pull/10541#discussion_r3312977903.

4. **Empty derived cipher list now hard-fails** (`return WOLFSSL_FAILURE`) instead of skipping `set_cipher_list` and returning success. Allowlist is now authoritative — the CTX is torn down by `wolfSSL_CTX_new_ex` rather than silently keeping its default suites. Addresses https://github.com/wolfSSL/wolfssl/pull/10541#discussion_r3312977926.

5. **`wolfSSL_crypto_policy_get_ciphers()` no longer returns an empty string in granular mode.** Both `enable()` paths mirror the derived cipher list into `crypto_policy.str` after a successful parse. Addresses https://github.com/wolfSSL/wolfssl/pull/10541#discussion_r3312977962 and https://github.com/wolfSSL/wolfssl/pull/10541#discussion_r3312977985.

6. **`crypto_policy_cipher_count()` uses `wolfSSL_sk_SSL_CIPHER_num()`** instead of looping until NULL. Addresses https://github.com/wolfSSL/wolfssl/pull/10541#discussion_r3312978030.

**Test extension**

The DTLS block of `test_wolfSSL_crypto_policy_granular()` now asserts (a) the global cipher list is non-empty for a DTLS-only fixture, (b) the DTLS CTX survives `apply_granular`, (c) the derived list round-trips through `wolfSSL_CTX_set_cipher_list()` on a fresh DTLS CTX. We deliberately don't assert on `wolfSSL_get_ciphers_compat()` for a DTLS SSL — its `sslCipherMinMaxCheck()` compares TLS minor numbers (e.g. `TLSv1_2_MINOR = 3`) with DTLS minor numbers (e.g. `DTLSv1_2_MINOR = 0xfd`), which looks like a pre-existing TLS-vs-DTLS quirk unrelated to this PR. Happy to file a separate issue if that's useful.

**Local verification**

Built and tested on Ubuntu 24.04 + gcc 13 with five configurations:

| # | configure flags | result |
|---|-----------------|--------|
| 1 | `CPPFLAGS=-DNO_VERIFY_OID` (the original CI failure) | build clean, `make check` 5/5 |
| 2 | `--enable-all --enable-dtls --enable-dtls13 --with-sys-crypto-policy=…allowlist.txt` | build clean, `make check` 17/17 |
| 3 | bare `./configure` | build clean, `make check` 5/5 |
| 4 | `--enable-all` + `-pedantic -Werror -fsanitize=undefined -O1` | build clean (a pre-existing UBSan finding in `src/ssl_crypto.c` DES code is unrelated) |
| 5 | `--enable-dtls --enable-dtls13` + DTLS-only allowlist | granular test passes |

**On the contributor agreement** — Max and Dom (Dominik Blain, co-author on every commit) will email `support@wolfssl.com` shortly with the paperwork on our side. Let us know if there's a preferred format.

Thanks again for the careful read — the DTLS issues in particular are the kind of thing you only catch when someone who's read the code points at the right line.

— @max-qlab and @dom-omg

philljj (changes requested):
It looks like non-ASCII chars are sneaking in, which is causing the `check-source-text` failure:
```
[check-source-text] [1 of 1] [0b95179dac]
...
8 bit:
./tests/api.c:30067: * created afterwards must reflect the policy's primitives ¶ cipher
./tests/api.c:30158:     * applier silently kept the CTX's default suites ¶ i.e. the
./tests/api.c:30166:     *      success ¶ empty derive now returns WOLFSSL_FAILURE and
./src/ssl.c:332: * per-setter guards (SetMinVersion, SetMinRsaKey_Sz, ¶) must let our
./src/ssl.c:5547:        /* Mirror the derived cipher list ¶ see comment in
./src/crypto_policy_granular.c:17: * the `crypto_policy_applying` flag in src/ssl.c ¶ that is the only
./src/crypto_policy_granular.c:23: * upgrading; the intersection of "policy-enabled ¶ build-supported"
./src/crypto_policy_granular.c:259:     * may add directives that change the *meaning* of existing keys ¶
./src/crypto_policy_granular.c:410: * enable every TLS 1.2 row that survives the other constraints ¶ and
./src/crypto_policy_granular.c:549:         * family ¶ the right floor. */
[...and more...]
```

Please remove the non-ASCII characters. You might need to instruct Claude to use ASCII-only for code and comments.

Thank you!

philljj (changes requested):
Still a few non-ascii chars added:
```
[check-source-text] [1 of 1] [7d2b407b9c]
...
8 bit:
./src/ssl.c:332: * per-setter guards (SetMinVersion, SetMinRsaKey_Sz, ¶) must let our
./src/ssl.c:5547:        /* Mirror the derived cipher list ¶ see comment in
```

max-qlab:
Hi @philljj — apologies, you were right, my previous ASCII pass missed the comments I touched in `src/ssl.c` and the format-spec README. New commit `48d2e28` cleans up the remaining nine lines:

| File | Line | Was | Now |
|---|---|---|---|
| `src/ssl.c` | 332 | `…` | `...` |
| `src/ssl.c` | 5547 | `—` | ` -- ` |
| `examples/crypto_policies/README.md` | 36 | `—` | ` -- ` |
| `examples/crypto_policies/README.md` | 87 | `…` | `...` |
| `examples/crypto_policies/README.md` | 92 | `∩` | `intersection of A and B` |
| `examples/crypto_policies/README.md` | 104 | `× × ×` | `x x x` |
| `examples/crypto_policies/README.md` | 113, 132, 136 | `—` | ` -- ` |

Comments and prose only — no code, fixtures, or test logic touched (9 insertions / 9 deletions).

Verified locally on the full PR file set:

```
$ grep -rnP '[^\x00-\x7F]' \
    src/ssl.c src/crypto_policy_granular.c \
    src/crypto_policy_granular.h src/include.am \
    examples/crypto_policies/ tests/api.c
# no matches
```

Sorry for the back-and-forth on the check-source-text gate.

— @max-qlab

examples/crypto_policies/README.md
```diff
@@ -0,0 +1,140 @@
...
+   (ECC floor derived from RSA-equivalent strength).
+
+Steps 1, 3 and 4 are best-effort: if a build lacks the primitive (no
+TLS 1.0 support, no `rsa_pss_*`), the applier logs and continues
+rather than tearing down the CTX -- the remaining steps still enforce
+the policy.
+
+## The five fixtures shipped here
+
+`legacy/`, `default/`, `future/`, `fips/`, `bsi/` are unmodified
```
philljj, line 118: I see only legacy, default, future. Where are `fips/`, `bsi/`?

Probably legacy, default, future are sufficient.

examples/crypto_policies/README.md
```diff
@@ -0,0 +1,140 @@
...
+outputs of the Fedora crypto-policies generator. They are checked into
+this tree so the wolfSSL unit tests can exercise the parser end-to-end
+against the same files a Fedora install would produce. Regenerate
+with:
+
+```sh
+python3 build-crypto-policies.py --flat --policy DEFAULT policies out
```
philljj, line 125: Imo we should include `build-crypto-policies.py` in this PR as well.

src/ssl.c
```diff
@@ -620,22 +630,40 @@ WOLFSSL_CTX* wolfSSL_CTX_new_ex(WOLFSSL_METHOD* method, void* heap)
...
-
-        if (list != NULL && *list != '\0') {
-            if (AllocateCtxSuites(ctx) != 0) {
-                WOLFSSL_MSG("allocate ctx suites failed");
+        if (crypto_policy_gran_enabled) {
+            /* Granular allowlist: drive the wolfSSL public API directly.
+             * The per-setter policy guards are temporarily disarmed so
+             * our own apply step can install the policy values. */
+            int ret;
+            crypto_policy_applying = 1;
```
philljj, line 638: This `crypto_policy_applying` flag feels complicated and dangerous. I recommend remove this mechanism, and just allow calling `wolfSSL_CTX_SetMinVersion` even if the crypto policy is enabled (remove the `if (crypto_policy.enabled` check).

src/crypto_policy_granular.c
```diff
@@ -0,0 +1,697 @@
...
+                           "build: %d (continuing)", min_ver, rc);
+        }
+    }
+
+    /* 2. Cipher list. An allowlist is authoritative: if the
+     * intersection of policy-enabled cipher suites and the suite table
+     * is empty, the CTX would silently keep its default cipher list
+     * and the policy would not actually constrain anything. Refuse
+     * outright in that case. */
+    rc = wolfSSL_crypto_policy_derive_cipher_list(p, buf, sizeof(buf));
```
philljj, line 603: If previously `wolfSSL_crypto_policy_derive_cipher_list()` was already called to fill out `crypto_policy.str`, then couldn't we reuse that `crypto_policy.str`, and avoid calling it again here?

src/ssl.c
```diff
@@ -5390,23 +5421,84 @@ int wolfSSL_crypto_policy_enable(const char * policy_file)
         return WOLFSSL_BAD_FILE;
     }
 
-    if (sz <= 0 || sz > MAX_WOLFSSL_CRYPTO_POLICY_SIZE) {
+    /* Granular allowlist files can exceed MAX_WOLFSSL_CRYPTO_POLICY_SIZE
+     * (the legacy single-line cap). Allocate a heap buffer for the sniff
+     * pass; we fall back to the legacy in-place buffer when the file
+     * fits and turns out to be legacy format. */
+    if (sz <= 0 || sz > (long)(1L << 20)) {
```
philljj, line 5428: `1L << 20` seems really big for an allow list file. Could we not set a smaller upper limit based on these?

```
#define WOLF_CP_MAX_TOKENS      64                                              
#define WOLF_CP_MAX_TOKEN_LEN   48                                              
#define WOLF_CP_MAX_LINE       256
```

src/ssl.c
```diff
@@ -319,9 +319,19 @@ int wc_OBJ_sn2nid(const char *sn)
 
 
 #if defined(WOLFSSL_SYS_CRYPTO_POLICY)
+#include "crypto_policy_granular.h"
 /* The system wide crypto-policy. Configured by wolfSSL_crypto_policy_enable.
  * */
 static struct SystemCryptoPolicy crypto_policy;
```
philljj, line 325: I appreciate you are trying to preserve the legacy `crypto_policy`, but I think these two structs (`SystemCryptoPolicy`, and `WolfGranularPolicy`) should be consolidated into one common struct.

src/crypto_policy_granular.c
```diff
@@ -0,0 +1,697 @@
+/* crypto_policy_granular.c
```
philljj, line 1: These two new files need a wolfssl copyright header.

src/crypto_policy_granular.c
```diff
@@ -0,0 +1,697 @@
...
+    if (buf == NULL) {
+        return 0;
+    }
+
+    while (*p != '\0') {
+        const char *line_start = p;
+        size_t      n = 0;
+        const char *cursor;
+
+        while (*p != '\0' && *p != '\n') {
```
philljj, line 111: Can this be replaced with `strchr*p, '\n'); like below?
```
const char *nl = strchr(p, '\n');
```

philljj (changes requested):
This is my first pass reviewing.

I really like the allow-list style config this is shooting for, e.g. similar to `/etc/crypto-policies/back-ends/gnutls.config`. I also like the parsing logic isolated to `crypto_policy_granular.c`, and think this is a good foundation to build on.

I appreciate this is trying to preserve the legacy implementation (and this companion PR is simultaneously trying to expand it: https://github.com/wolfSSL/wolfssl/pull/10535), however I think some of this will need to be simplified and consolidated. E.g. 2 redundant structs for legacy and granular config seems too much complexity. Please feel free to change the legacy system, because it was insufficient.

If we are going to support both config styles,  I think the only thing that should be different is the initial config loading, and beyond that it should be common code applying a common crypto policy struct to new CTX's.

t184256:
Hello, crypto-policies maintainer here.

> The allowlist format mirrors the GnuTLS back-end that crypto-policies already endorses as granular: explicit primitive names, one directive per primitive, grouped by category.

> The exact section/key names (enabled-version, enabled-cipher, …) are deliberately close to the GnuTLS back-end so the crypto-policies maintainers can adjust them cheaply. That naming call is theirs

> The vocabulary is owned by crypto-policies; the wolfSSL-side mapping tables live in `src/crypto_policy_granular.c`.

I'm afraid there might be a misunderstanding here =)

I appreciate the willingness to accommodate my requests, but I really didn't mean to force any specific naming or config file structure onto you. Doing so would save me maybe tens of minutes of my time and require disproportional effort on your side. E.g., whatever mapping tables you end up with, that wasn't my intention, and it's perfectly fine for any kind of `'TLS13-AES256-GCM-SHA384' = ('TLS1.3', 'AES-256-GCM')` mappings to live in crypto-policiies.

Similarly, I do not insist on the config file to be superficially or totally similar to GnuTLS allowlisting format.

What I requested is configurability. I want to be able to express what's enabled, fine-grained. Maintaining some name mapping on c-p sige --- that's totally expected; if your config requires specifying ciphersuites by their names --- perfectly fine, if the form most convenient to you is happens to be an XML file --- I can live with that. In general, it'd be wonderful if you just designed a granular config format most comfortable to you and your users, as free from weird unorthogonal legacy cruft like `SECLEVEL` being the only knob for key sizes that also mucks with protocols, ciphers and signature algorithms all at the same time. I'd then rather use the well-tested and documented config format you'd find easy to maintain than force you to carry a special codepath just for me.

Make it simple. I'll adapt.

philljj:
@t184256 thank you for the feedback.

I think it's in our interest to follow an accepted standard here, rather than create yet another.

It seems like `gnutls.config` style is preferred over `opensslcnf.config`. Also, personally the allowlist ini config in this PR is pretty and very readable.

t184256:
I can use any format, I can use the proposed one as well. =) It's just... this MR feels like it catered a bit too much to my preferences, resulting in a ton of work and maintenance burden, so I'd like to stress that I'm perfectly willing to host some of the complexity in the crypto-policies generator if it makes the wolfssl side of things any simpler.

philljj:
Will be done in a separate PR.

PR #10542, opened 2026-05-27
CI Addition (ECH): Check only the public SNI is visible

sebastian-carpenter:
# Description

Added a test to check that the SNI in the outer ClientHello is always the public SNI. Other checks exist to verify the inner SNI is correct. Also added additional testing for the ECH rejection path as well.

Corrected `TLSX_EchChangeSNI` to always swap the SNI when it is processing the outer hello.

# Testing

Added `test_wolfSSL_Tls13_ECH_wire_sni` which forces a HelloRetryRequest with either ech rejection or acceptance.

Added interop testing for the ECH rejection path.

# Checklist

 - [X] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/tests/api.c
+++ b/tests/api.c
@@ -14024,6 +14024,63 @@ static word16 echCbTestKemID = 0;
 static word16 echCbTestKdfID = 0;
 static word16 echCbTestAeadID = 0;
 
+/* return the index of the first extension
+ * extLen will be updated with the length of the extensions field */
+static int ech_seek_extensions(byte* buf, word16* extLen)
+{
+    word16 idx;
+    byte sessionIdLen;
+    word16 cipherSuitesLen;
+    byte compressionLen;
+
+    idx = OPAQUE16_LEN + RAN_LEN;
+
+    sessionIdLen = buf[idx++];
+    idx += sessionIdLen;
+
+    ato16(buf + idx, &cipherSuitesLen);
+    idx += OPAQUE16_LEN + cipherSuitesLen;
+
+    compressionLen = buf[idx++];
+    idx += compressionLen;
+
+    ato16(buf + idx, extLen);
+    idx += OPAQUE16_LEN;
+
+    return idx;
+}
+
+/* locate a particular extension:
+ *   idx_p is updated with the location of that extension
+ *     -> idx_p should start just after the handshake header
+ *   0 returned on success, error otherwise */
+static int ech_find_extension(byte* buf, word16* idx_p, word16 extType)
+{
+    word16 idx;
+    word16 extIdx;
+    word16 extLen;
+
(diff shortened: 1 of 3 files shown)


tests/api.c
```diff
@@ -15914,6 +15914,94 @@ static int test_wolfSSL_Tls13_ECH_GREASE(void)
     return EXPECT_RESULT();
 }
 
+/* The public name must be visible and the private name must not be visible */
+static int test_wolfSSL_Tls13_ECH_wire_sni_ex(int accept)
+{
+    EXPECT_DECLS;
+    test_ssl_memio_ctx test_ctx;
+    WOLFSSL_CTX* tempCtx = NULL;
```
SparkiDev: Reduce the scope - put declarations down in block where used.

tests/api.c
```diff
@@ -15914,6 +15914,94 @@ static int test_wolfSSL_Tls13_ECH_GREASE(void)
...
+    ExpectIntEQ(test_ctx.s_ssl->options.serverState,
+        SERVER_HELLO_RETRY_REQUEST_COMPLETE);
+
+    /* client reads HRR from c_buff and writes CH2 into s_buff */
+    ExpectIntEQ(wolfSSL_connect(test_ctx.c_ssl), WOLFSSL_FATAL_ERROR);
+    ExpectIntEQ(wolfSSL_get_error(test_ctx.c_ssl, WOLFSSL_FATAL_ERROR),
+        WOLFSSL_ERROR_WANT_READ);
+
+    /* CH2 wire bytes: same property must hold */
+    ExpectNotNull(mymemmem(test_ctx.s_buff, test_ctx.s_len,
```
SparkiDev: Can we expect echCbTestPublicName to be at a specific location?

SparkiDev (changes requested):

tests/api.c
sebastian-carpenter: fixed, moved them down to !accept block.

tests/api.c
sebastian-carpenter: Yes! Removed memmem check and replaced it with `ech_find_extension` to only perform the check at the SNI extension.

Once more extensions are added to ECH this test should be refactored to check that those are swapped out as well.

sebastian-carpenter:
Rebased to get `ech_find_extension`. 'check sent SNI is correct' comments mark the major change locations for checking the SNI.

Added a sanity check at the end of the test to make sure ECH is suceeding or failing as expected.

sebastian-carpenter:
Jenkins retest this please

sebastian-carpenter:
Moving this into #10568 since it only contains testing now. #10568 was also the source for this wire_sni test anyway.

PR #10551, opened 2026-05-28
Add DTLS throughput benchmark tool and optimize send path

julek-wolfssl:
Add examples/benchmark/dtls_bench, a DTLS throughput benchmark that completes a handshake and then measures bulk-send throughput. It supports DTLS 1.2 and 1.3, selectable cipher suites, an end-to-end mode, and a -z sink mode that discards records on the server after the handshake to isolate the sender's record-layer cost. The socket is set up with wolfSSL_set_dtls_fd_connected.

Optimize the send path exercised by the benchmark:

- wolfio (EmbedSendTo): cache the per-descriptor socket-type probe (getsockopt SO_TYPE) in WOLFSSL_DTLS_CTX instead of running it on every send, removing a syscall from the record send path. The cache is invalidated whenever rfd/wfd is reassigned.

- internal (BuildMessage): for AEAD suites whose explicit nonce is the 8-byte record sequence number, write the sequence number directly as nonce_explicit instead of drawing it from the RNG. This covers AES-GCM (RFC 5288 sec 3), AES-CCM (RFC 6655 sec 3), SM4-GCM/CCM (RFC 8998 sec 3), and Camellia-/ARIA-GCM which inherit the RFC 5288 construction; ChaCha20 uses an implicit nonce and is excluded. A new read-only PeekSEQ() helper reads the sequence number without advancing the per-direction counter, leaving the single mandated increment to writeAeadAuthData().

Also ignore the built dtls_bench binary in .gitignore.


src/wolfio.c
```diff
@@ -642,19 +642,28 @@ static int PeerIsIpv6(const SOCKADDR_S *peer, XSOCKLENT len)
...
             &length) == 0 && type != SOCK_DGRAM) {
-        return 0;
+        dtlsCtx->isDGram = 0;
     }
     else {
-        return 1;
+        dtlsCtx->isDGram = 1;
     }
+    dtlsCtx->isDGramCached = 1;
+    return dtlsCtx->isDGram;
```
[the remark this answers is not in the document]
julek-wolfssl: Not important as its only used for sending.

julek-wolfssl:
retest this please.

julek-wolfssl:
The jenkins failures are not related to this PR.

examples/benchmark/dtls_bench.c
```diff
@@ -0,0 +1,766 @@
...
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <stdint.h>
+#include <errno.h>
+#include <time.h>
+#include <unistd.h>
+#include <getopt.h>
+#include <sys/socket.h>
```
douzzer, line 80: MinGW doesn't like ungated sys/socket.h:
```
[cross-mingw-all-crypto] [20 of 55] [9c858a9190]
    configure...   real 0m19.295s  user 0m0.004s  sys 0m0.003s
    build...1e7d632a0a (<juliusz@wolfssl.com> 2026-05-27 14:32:21 +0000 68) #include <sys/socket.h>
examples/benchmark/dtls_bench.c:68:10: fatal error: sys/socket.h: No such file or directory
   68 | #include <sys/socket.h>
      |          ^~~~~~~~~~~~~~
compilation terminated.
```

repro with something like:
```
failed config: 'EXTRA_CPPFLAGS=-Werror' '--srcdir' '.' '--disable-jobserver' '--enable-option-checking=fatal' 'FILECMD=file' 'MANIFEST_TOOL=/bin/false' 'DLLTOOL=/bin/false' '--host=x86_64-w64-mingw32' '--disable-shared' '--enable-static' '--enable-all-crypto' '--enable-dtls' '--enable-opensslextra' '--disable-siphash' 'CFLAGS=-DTEST_ALWAYS_RUN_TO_END' 'CPPFLAGS=-DNO_WOLFSSL_CIPHER_SUITE_TEST -DWOLFSSL_OLD_PRIME_CHECK -pedantic -Wdeclaration-after-statement -Wnull-dereference -DTEST_LIBWOLFSSL_SOURCES_INCLUSION_SEQUENCE'
```

douzzer (dismissed):

examples/benchmark/dtls_bench.c
julek-wolfssl: Should be fixed now

src/wolfio.c
```diff
@@ -642,19 +642,28 @@ static int PeerIsIpv6(const SOCKADDR_S *peer, XSOCKLENT len)
...
+ * fixed for its lifetime, so the getsockopt(SO_TYPE) probe runs once per
+ * connection and is cached in dtlsCtx; the cache is invalidated wherever
+ * rfd/wfd is reassigned. */
+static int isDGramSock(WOLFSSL_DTLS_CTX* dtlsCtx, int sfd)
 {
     int type = 0;
     /* optvalue 'type' is of size int */
     XSOCKLENT length = (XSOCKLENT)sizeof(type);
 
+    if (dtlsCtx->isDGramCached)
```
dgarske: **🟠 [Medium] isDGram cache write is an unsynchronized data race under WOLFSSL_RW_THREADED**

Previously `isDGramSock()` was a pure read of the socket type. It now lazily writes `dtlsCtx->isDGram` and `dtlsCtx->isDGramCached` (bit-fields that share a storage unit with `connected`, and with `processingPendingRecord` when `WOLFSSL_DTLS_CID` is set). In a `WOLFSSL_RW_THREADED` build, `EmbedSendTo` (write thread) calls this on every send and `EmbedReceiveFrom` (read thread) calls it on the `recvd == 0` path, so two threads can perform the first-time write concurrently. Writing adjacent bit-fields is a read-modify-write of the shared byte and is not atomic, so this is a data race (UB per the C memory model) on `dtlsCtx`. In practice it is benign because both threads write the same socket-type value and `connected` is not modified concurrently, but it can be reported by ThreadSanitizer CI and is a new race introduced by moving shared-state mutation into the hot I/O path. The rest of `dtlsCtx` peer state is already protected by `peerLock` precisely because these callbacks run concurrently.

**Fix:** Confirm intent for threaded builds; prefer computing the cache eagerly at fd assignment to remove the race from the data path entirely.

dgarske (commented):
## Skoll Code Review

### Posted findings
- [Medium] **isDGram cache write is an unsynchronized data race under WOLFSSL_RW_THREADED** — `src/wolfio.c:655-666`
- [Low] **Single isDGram cache shared between rfd and wfd of potentially different socket types** — `src/wolfio.c:649-667`
- [Low] **Confirm discarded explicit-nonce value is intentional across FIPS and epoch-order paths** — `src/internal.c:24809-24831`
- [Low] **now_sec ignores clock_gettime failure, can return uninitialized time** — `examples/benchmark/dtls_bench.c:96-101`
- [Low] **-z (sink send) silently ignored when combined with -s (server)** — `examples/benchmark/dtls_bench.c:parse_args`

src/wolfio.c
```diff
@@ -642,19 +642,28 @@ static int PeerIsIpv6(const SOCKADDR_S *peer, XSOCKLENT len)
 }
 #endif /* !WOLFSSL_IPV6 */
 
-static int isDGramSock(int sfd)
+/* Return non-zero iff sfd is a SOCK_DGRAM socket. A descriptor's type is
+ * fixed for its lifetime, so the getsockopt(SO_TYPE) probe runs once per
+ * connection and is cached in dtlsCtx; the cache is invalidated wherever
+ * rfd/wfd is reassigned. */
+static int isDGramSock(WOLFSSL_DTLS_CTX* dtlsCtx, int sfd)
```
dgarske: **🔵 [Low] Single isDGram cache shared between rfd and wfd of potentially different socket types**

The cache stores one `isDGram` flag per `dtlsCtx`, but it is consulted for both the read descriptor (`dtlsCtx->rfd`, via `EmbedReceiveFrom`) and the write descriptor (`dtlsCtx->wfd`, via `EmbedSendTo`). If an application sets `rfd` and `wfd` to different sockets of different types (legal via `wolfSSL_set_read_fd`/`wolfSSL_set_write_fd`), the first probe wins and the cached value is then incorrectly applied to the other descriptor. This is unusual for DTLS (rfd == wfd normally) and the cache is invalidated on each fd reassignment, so the common path is fine, but the original code re-probed the exact `sfd` passed in each time, whereas the cache is now keyed on the context rather than the descriptor.

**Fix:** Confirm the rfd==wfd (or same-type) assumption holds for all supported DTLS usages; document it next to the cache fields.

src/internal.c
```diff
@@ -24774,9 +24806,29 @@ int BuildMessage(WOLFSSL* ssl, byte* output, int outSz, const byte* input,
                     args->iv = args->staticIvBuffer;
                 }
 
-                ret = wc_RNG_GenerateBlock(ssl->rng, args->iv, args->ivSz);
-                if (ret != 0)
-                    goto exit_buildmsg;
+#ifdef HAVE_AEAD
```
dgarske, line 24870: **🔵 [Low] Confirm discarded explicit-nonce value is intentional across FIPS and epoch-order paths**

The new code writes the record sequence number into the explicit nonce via `PeekSEQ(ssl, epochOrder, args->iv)` instead of `wc_RNG_GenerateBlock`. In every standard configuration this written value is subsequently overwritten before transmission: the non-FIPS/FIPSv2+ path copies the cipher-generated IV over it (`XMEMCPY(out, ssl->encrypt.nonce + AESGCM_IMP_IV_SZ, ...)` at line ~20741), and the legacy-FIPS path overwrites `args->iv` with `ssl->keys.aead_exp_IV` (line ~24839). So the real effect of this change is removing the per-record RNG draw; the seq value itself is intentionally discarded. Note also that the actual AAD seq is written with `CUR_ORDER` inside `writeAeadAuthData` (line ~20695) while `PeekSEQ` here uses `epochOrder`, so the two would differ for DTLS SCR PREV_ORDER retransmits, but this is harmless precisely because the `PeekSEQ`-written bytes are discarded. This is correct but subtle; a future reader could mistakenly assume the on-wire explicit nonce equals `PeekSEQ`.

**Fix:** Keep the change; add a short clarifying comment to prevent future misreading and confirm the RNG-removal is the intended win.

examples/benchmark/dtls_bench.c
```diff
@@ -0,0 +1,788 @@
...
+#define DEFAULT_HOST       "127.0.0.1"
+#define DEFAULT_PORT       11111
+#define DEFAULT_DURATION   10
+#define DEFAULT_MTU        1400
+#define DEFAULT_RECORD     1300
+#define DEFAULT_VERSION    13
+#define DEFAULT_SOCKBUF    (8 * 1024 * 1024)
+
+/* Upper bound on per-record overhead the user must leave under MTU.
+ * DTLS 1.2 worst case: 13 (hdr) + 8 (explicit nonce) + 16 (AEAD tag) = 37.
```
dgarske, line 97: **🔵 [Low] now_sec ignores clock_gettime failure, can return uninitialized time**

`clock_gettime()`'s return value is cast to void and `ts` is not initialized, so on the (rare) failure of `CLOCK_MONOTONIC` the function returns garbage from the stack, corrupting all throughput/duration math. This is benchmark-only code so impact is limited, but it is trivial to harden.

**Fix:** Initialize ts and check the return value, returning a sentinel on failure.

examples/benchmark/dtls_bench.c
```diff
@@ -0,0 +1,788 @@
...
+                    strcmp(optarg, "list") == 0) {
+                    c->listCipher = 1;
+                }
+                else {
+                    c->cipherList = optarg;
+                }
+                break;
+            case 'b': c->sockBuf     = atoi(optarg); break;
+            case 'n': c->plainUdp    = 1; break;
+            case 'z': c->sinkSend    = 1; break;
```
dgarske, line 223: **🔵 [Low] -z (sink send) silently ignored when combined with -s (server)**

The help text documents `-z` as client-only, but `parse_args` performs no validation when `-z` is passed together with `-s`. In server mode `sinkSend` is parsed and then never used, so a user invoking `-s -z` gets no warning that the flag had no effect. Minor usability nit in example code.

**Fix:** Emit a warning (or error) when -z is combined with -s so the no-op is visible to the user.

src/wolfio.c
julek-wolfssl: Addressed.

src/wolfio.c
julek-wolfssl: Addressed.

src/internal.c
julek-wolfssl: Addressed.

examples/benchmark/dtls_bench.c
julek-wolfssl: Addressed.

examples/benchmark/dtls_bench.c
julek-wolfssl: Addressed.

src/internal.c
```diff
@@ -24834,9 +24867,33 @@ int BuildMessage(WOLFSSL* ssl, byte* output, int outSz, const byte* input,
...
```
dgarske, line 24870: **🟠 [Medium] Explicit-nonce 'placeholder' invariant not guaranteed on ATOMIC_USER AEAD path**

The new code writes `PeekSEQ(ssl, epochOrder, args->iv)` for `cipher_type == aead && ivSz == AESGCM_EXP_IV_SZ`, and the comment asserts the bytes are placeholders that "the encrypt paths below overwrite the explicit nonce before transmission." I verified this holds for the internal `Encrypt()` paths (AES-GCM/CCM line 20784, ARIA line 20828, SM4 line 20931) and the FIPS<2 copy at line 24902. However, when an application registers an ATOMIC_USER record-layer callback (`ssl->ctx->MacEncryptCb`, line 24992), `BuildMessage` calls that callback and `goto exit_buildmsg` BEFORE `Encrypt()` runs — so the explicit nonce placed by `BuildMessage` (now the record sequence number instead of a random value) is what the user callback frames/transmits. The `MacEncryptCb` block at line 24992 is not gated on `cipher_type`, so it is reachable for AEAD suites. Using the sequence number as the explicit nonce is RFC 5288-compliant and uniqueness is preserved, so this is NOT a correctness/interop break, but it is a behavioral change (random -> seq) on that niche path. Also note the comment says the overwrite happens 'below' in the encrypt path, whereas for FIPS<2 it actually happens just above in BuildMessage itself.

**Fix:** Confirm that ATOMIC_USER + AEAD is either unsupported or that emitting the sequence number as the explicit nonce there is acceptable (it is RFC 5288-compliant). Optionally tighten the comment to note the FIPS<2 case overwrites args->iv in BuildMessage itself, and that the only path where the value reaches the wire (atomic user) still produces a valid nonce.

examples/benchmark/dtls_bench.c
```diff
@@ -0,0 +1,796 @@
...
+    if (wolfSSL_dtls_set_peer(ssl, &addr, sizeof(addr)) != WOLFSSL_SUCCESS) {
+        fprintf(stderr, "dtls_set_peer failed\n"); goto cleanup;
+    }
+    /* Socket was connect()-ed above; tell wolfSSL so EmbedSendTo can skip
+     * its per-send getsockopt(SO_TYPE) probe. */
+    if (wolfSSL_set_dtls_fd_connected(ssl, fd) != WOLFSSL_SUCCESS) {
+        fprintf(stderr, "set_dtls_fd_connected failed\n"); goto cleanup;
+    }
+    if (set_mtu(ssl, c->mtu) != 0) goto cleanup;
+
```
dgarske, line 727: **🟠 [Medium] DTLS benchmark client aborts on transient send errors that udp_client tolerates**

The plain-UDP baseline (`udp_client`) deliberately retries on `ENOBUFS`/`EAGAIN`/`EWOULDBLOCK` (the qdisc/tx-queue-full case on Linux UDP sends, which can occur even on blocking sockets). The DTLS client's write loop treats any `wolfSSL_write` return != recordSz as fatal and `goto cleanup`, so a transient `ENOBUFS` surfaced through `EmbedSendTo`/wolfSSL as `WANT_WRITE` (or a CBIO error) ends the benchmark early with an error instead of retrying. In the default (non -z) DTLS mode over loopback the sender can outrun the kernel, making this reachable. The `-z` sink mode avoids the kernel and is unaffected.

**Fix:** Mirror udp_client's transient-error handling so the non-sink DTLS benchmark does not abort on a recoverable backpressure event.

examples/benchmark/dtls_bench.c
```diff
@@ -0,0 +1,796 @@
...
+                    c->listCipher = 1;
+                }
+                else {
+                    c->cipherList = optarg;
+                }
+                break;
+            case 'b': c->sockBuf     = atoi(optarg); break;
+            case 'n': c->plainUdp    = 1; break;
+            case 'z': c->sinkSend    = 1; break;
+            case '?':
```
dgarske, line 224: **🔵 [Low] Help flag -? returns failure exit code**

`-?` is an explicit help request but is handled by the same `case '?': default:` arm as unknown/invalid options, which prints usage to stderr and returns -1, causing `main` to return exit code 1. A user asking for help conventionally expects success (exit 0) and output on stdout.

**Fix:** Optionally distinguish an explicit `-?`/help request (print usage to stdout, return a success sentinel handled like the cipher-list help that returns 0) from genuine getopt errors.

examples/benchmark/dtls_bench.c
```diff
@@ -0,0 +1,796 @@
...
+        "                  no kernel send).\n"
+        "  -?              show this help\n",
+        prog, DEFAULT_HOST, DEFAULT_PORT,
+        DEFAULT_DURATION, DEFAULT_RECORD, DEFAULT_MTU,
+        DEFAULT_VERSION, DEFAULT_SOCKBUF);
+}
+
+static void list_ciphers(void)
+{
+    char  buf[8192];
```
dgarske: **🔵 [Low] Large stack buffer and unvalidated numeric option parsing in benchmark**

`list_ciphers` declares an 8192-byte stack buffer (`char buf[8192]`), exceeding the wolfSSL convention of using the WOLFSSL_SMALL_STACK pattern for stack buffers larger than ~100 bytes. Separately, `-p` (port) and `-b` (sockbuf) are parsed with `atoi` and not range-validated (unlike -d/-r/-m/-v), so non-numeric or out-of-range values silently become 0/garbage (e.g. `-p abc` -> port 0). These are example-only quality items, not library defects.

**Fix:** Consider a smaller/heap buffer for the cipher list and add simple bounds checks for -p and -b for consistency with the other numeric options. Low priority for a benchmark tool.

dgarske (commented):
## Skoll Code Review

### Posted findings
- [Medium] **Explicit-nonce 'placeholder' invariant not guaranteed on ATOMIC_USER AEAD path** — `src/internal.c:24870-24896`
- [Medium] **DTLS benchmark client aborts on transient send errors that udp_client tolerates** — `examples/benchmark/dtls_bench.c:710-722`
- [Low] **No regression test for seq-as-nonce path or cached socket-type fields** — `src/internal.c:24885-24888, src/wolfio.c:649-662`
- [Low] **Help flag -? returns failure exit code** — `examples/benchmark/dtls_bench.c:266-269, 786-789`
- [Low] **Large stack buffer and unvalidated numeric option parsing in benchmark** — `examples/benchmark/dtls_bench.c:140-156, 186-227`

src/internal.c
julek-wolfssl: addressed

examples/benchmark/dtls_bench.c
julek-wolfssl: addressed

examples/benchmark/dtls_bench.c
julek-wolfssl: addressed

examples/benchmark/dtls_bench.c
julek-wolfssl: addressed

dgarske (approved):

PR #10568, opened 2026-06-01
Enhancement: Public extension manager for ECH

sebastian-carpenter:
# Description

Reworks ECH public-vs-private extension handling. The old path swapped a single SNI *string* in place on the live extension list - this was fragile and SNI-only. This PR generalizes it to a public-extension list on the ECH struct that's swapped by extension type, so the outer ClientHello can carry arbitrary public extensions.

ZD#21931

## Public extension manager
- Add `TLSX* extensions` to `WOLFSSL_ECH` holding the public extensions (e.g. the public-name SNI), populated in `TLSX_ECH_Use` from `echConfig->publicName`.
- Replace `TLSX_EchChangeSNI`/`TLSX_EchRestoreSNI` with:
  - `TLSX_EchShouldHideInner` - true when ECH is the outer type.
  - `TLSX_EchSwapExtensions` - swaps matching extension types between `ssl->extensions` and `ech->extensions`; unmatched public exts are prepended; `popCount` reverses leading nodes to preserve OuterExtensions ordering.
  - `TLSX_EchReplaceExtensions` - accept: free the public list; reject: swap public exts into `ssl->extensions`.
- `TLSX_GetSizeWithEch`/`TLSX_WriteWithEch` install the public exts before size/write and swap back after.
- `ForceZero` the hpke; free `ech->extensions` in `TLSX_ECH_Free`.

## SNI changes
- Remove the old SNI-state machinery: drop the `EchStateSNI` enum and the `sniState`/`privateName` fields, plus the sniState transitions in `DoTls13HandShakeMsgType`.
- `TLSX_SNI_Parse`: drop the sniState-based private-name save/restore. On outer-CH parse, match against any `echConfig->publicName` and write a matched public SNI to `ech->extensions`, not `ssl->extensions`; simplify the `matched`/`cacheOnly` and response-flag logic.
- Inner SNI is now optional - removed the client-side "inner SNI must be present" requirement enforced by `TLSX_EchChangeSNI`.
- `SendTls13ClientHello`: read padding length via `TLSX_SNI_GetRequest` instead of `ech->privateName`.
- Call `TLSX_EchReplaceExtensions` at the accept/reject points (`EchCheckAcceptance`, `DoTls13ServerHello` reject fallback, `DoTls13ClientHello`). This installs the correct SNI for use later in the handshake.

## Padding
- If the client ssl had no SNI then it would take the generic no SNI path. Updated padding calculation to fallback to ctx's SNI if it is available.

# Testing
- Add `test_wolfSSL_Tls13_ECH_wire_sni`: drive the handshake manually and assert the public name is present and the private name absent in raw CH1/CH2 bytes, across accept/reject.
- Update `test_wolfSSL_Tls13_ECH_no_private_name`: no-inner-SNI now succeeds against a permissive server (`ACCEPTED`); rejected only with `ABORT_ON_ABSENCE`.
- Add `b64MandatoryFirst` config case; `test_ech_server_sni_callback` rejects a wrong public name; add a double-public-SNI scenario to `bad_configs_ex`.
- Remove obsolete `test_wolfSSL_Tls13_ECH_long_SNI` (its overflow target - the deleted swap helpers - no longer exist) and relocate the over-long-SNI `BAD_LENGTH_E` check into `test_wolfSSL_UseSNI_params`.
- Testing for the new `TLSX_EchSwapExtensions` function: `test_TLSX_EchSwapExtensions`. This emulates the swap between ssl->extensions and ech->extensions with various configs: 'match', 'no match', 'match and non-match', 'match and no-match + out of order'. This required qualifying the function with `WOLFSSL_TEST_VIS`.
- SNI test vectors. Enumerated many possible SNI configurations / possibilities for ECH connections (`test_wolfSSL_Tls13_ECH_sni_parse`).

Added testing from #10542:
- More efficient *_wire_sni test
  * Updated further to test with SNI installed on CTX and to strictly check for Hostname SNI only.
- Interop testing for when ECH is rejected

# Follow-ups
- Lays groundwork for managing multiple public extensions and multiple SNI's in the outer ClientHello.

# Checklist

 - [X] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/.github/scripts/openssl-ech.sh
+++ b/.github/scripts/openssl-ech.sh
@@ -11,7 +11,7 @@ cleanup() {
 trap cleanup EXIT
 
 usage() {
-    echo "Usage: $0 <client|server> [--suite <KEM,KDF,AEAD>] [--pqc <group>] [--hrr] [--workspace <path>]"
+    echo "Usage: $0 <client|server> [--suite <KEM,KDF,AEAD>] [--pqc <group>] [--hrr] [--reject] [--workspace <path>]"
     exit 1
 }
 
@@ -22,6 +22,7 @@ MODE=""
 SUITE=""
 PQC=""
 FORCE_HRR=0
+REJECT=0
 
 WORKSPACE=${GITHUB_WORKSPACE:-"."}
 
@@ -49,6 +50,10 @@ while [ $# -gt 0 ]; do
             ;;
         --hrr)
             FORCE_HRR=1
+            shift
+            ;;
+        --reject)
+            REJECT=1
             shift
             ;;
         --workspace)
@@ -84,9 +89,16 @@ WOLFSSL_CLIENT=${WOLFSSL_CLIENT:-"$WORKSPACE/examples/client/client"}
 WOLFSSL_SERVER=${WOLFSSL_SERVER:-"$WORKSPACE/examples/server/server"}
 CERT_DIR=${CERT_DIR:-"$WORKSPACE/certs"}
 
+# correct ECH config, but it's old, ECH will be rejected
+REJECT_ECH_CONFIG="AD7+DQA6rAAgACCATZdDlHed6GlDeiYsu3r7sdWUkLVHZuTa3lbOf+hIbAAEAAEAAQALZXhhbXBsZS5jb20AAA=="
+
 TMP_LOG="$WORKSPACE/tmp_file.log"
+# Will need to look into validating the name against the cert for the OSSL cli.
+# This is fine, but should be upgraded to use a second cert in the future.
 PRIV_NAME="ech-private-name.com"
-PUB_NAME="ech-public-name.com"
+# example.com is taken from the server certificate,
+# echConfigs needs to authenticate against the cert with this name to succeed
+PUB_NAME="example.com"
 MAX_WAIT=50
 
 # --------------------------------------------------------------------------
@@ -128,6 +140,8 @@ openssl_server(){
 
     # parse ECH config from file
     ech_config=$(sed -n '/BEGIN ECHCONFIG/,/END ECHCONFIG/{/BEGIN ECHCONFIG\|END ECHCONFIG/d;p}' "$ech_file" | tr -d '\n')
+    # reject overrides the config the client connects with
+    [ "$REJECT" -ne 0 ] && ech_config="$REJECT_ECH_CONFIG"
     echo "parsed ech config: $ech_config" &>> "$TMP_LOG"
 
     # start OpenSSL ECH server with ephemeral port; line-buffer so the
@@ -158,17 +172,24 @@ openssl_server(){
     done
     echo "parsed port: $port" &>> "$TMP_LOG"
 
+    rm -f "$ech_file"
+
     # test with wolfssl client
+    # in reject mode the client is expected to error out, so tolerate a
+    #   nonzero exit
     $WOLFSSL_CLIENT -v 4 \
         -p "$port" \
         -S "$PRIV_NAME" \
         --ech "$ech_config" \
         $wolfssl_extra \
-        &>> "$TMP_LOG"
+        &>> "$TMP_LOG" || [ "$REJECT" -ne 0 ]
 
-    rm -f "$ech_file"
-
-    grep -q "ech_success=1" "$TMP_LOG"
+    if [ "$REJECT" -ne 0 ]; then
+        grep -q "ECH offered but rejected by server" "$TMP_LOG" && \
+            grep -q "ech_success=0" "$TMP_LOG"
+    else
+        grep -q "ech_success=1" "$TMP_LOG"
+    fi
 }
 
 # --------------------------------------------------------------------------
@@ -246,9 +267,13 @@ openssl_client(){
             exit 1
         fi
     done
+    # reject overrides the config the client connects with
+    [ "$REJECT" -ne 0 ] && ech_config="$REJECT_ECH_CONFIG"
     echo "parsed ech config: $ech_config" &>> "$TMP_LOG"
 
     # test with OpenSSL s_client using ECH
+    # in reject mode the s_client is expected to error out, so tolerate a
+    #   nonzero exit
     echo "wolfssl" | $OPENSSL s_client \
         -tls1_3 \
         -connect "localhost:$port" \
@@ -258,9 +283,13 @@ openssl_client(){
         -servername "$PRIV_NAME" \
         -ech_config_list "$ech_config" \
         $openssl_groups \
-        &>> "$TMP_LOG"
+        &>> "$TMP_LOG" || [ "$REJECT" -ne 0 ]
 
-    grep -q "ECH: success: 1" "$TMP_LOG"
+    if [ "$REJECT" -ne 0 ]; then
+        grep -q "ECH: Got 1 retry-configs" "$TMP_LOG"
+    else
+        grep -q "ECH: success: 1" "$TMP_LOG"
+    fi
 }
 
 rm -f "$TMP_LOG"
--- a/src/tls.c
+++ b/src/tls.c
@@ -13915,14 +13915,20 @@ static int TLSX_ECH_Use(WOLFSSL_EchConfig* echConfig, TLSX** extensions,
     /* setup the ephemeralKey */
     if (ret == 0)
         ret = wc_HpkeGenerateKeyPair(ech->hpke, &ech->ephemeralKey, rng);
+    /* use the chosen config's public name for the outer SNI */
     if (ret == 0) {
-        ret = TLSX_Push(extensions, TLSX_ECH, ech, heap);
-        if (ret != 0) {
-            wc_HpkeFreeKey(ech->hpke, ech->hpke->kem, ech->ephemeralKey,
-                ech->hpke->heap);
-        }
+        ret = TLSX_UseSNI(&ech->extensions, WOLFSSL_SNI_HOST_NAME,
+            echConfig->publicName, (word16)XSTRLEN(echConfig->publicName),
+            heap);
+        if (ret == WOLFSSL_SUCCESS)
+            ret = 0;
     }
+    if (ret == 0)
+        ret = TLSX_Push(extensions, TLSX_ECH, ech, heap);
     if (ret != 0) {
+        TLSX_FreeAll(ech->extensions, heap);
+        wc_HpkeFreeKey(ech->hpke, ech->hpke->kem, ech->ephemeralKey,
+            ech->hpke->heap);
         XFREE(ech->hpke, heap, DYNAMIC_TYPE_TMP_BUFFER);
         XFREE(ech, heap, DYNAMIC_TYPE_TMP_BUFFER);
     }
@@ -14974,9 +14980,8 @@ static void TLSX_ECH_Free(WOLFSSL_ECH* ech, void* heap)
 {
     XFREE(ech->innerClientHello, heap, DYNAMIC_TYPE_TMP_BUFFER);
     if (ech->hpke != NULL) {
-        if (ech->ephemeralKey != NULL)
-            wc_HpkeFreeKey(ech->hpke, ech->hpke->kem, ech->ephemeralKey,
-                ech->hpke->heap);
+        wc_HpkeFreeKey(ech->hpke, ech->hpke->kem, ech->ephemeralKey,
+            ech->hpke->heap);
         /* wc_HpkeFreeEchSecret is intentionally not here, free it in
          * TLSX_ExtractEch / TLSX_FinalizeEch */
         XFREE(ech->hpke, heap, DYNAMIC_TYPE_TMP_BUFFER);
@@ -14985,8 +14990,7 @@ static void TLSX_ECH_Free(WOLFSSL_ECH* ech, void* heap)
         ForceZero(ech->hpkeContext, sizeof(HpkeBaseContext));
         XFREE(ech->hpkeContext, heap, DYNAMIC_TYPE_TMP_BUFFER);
     }
-    if (ech->privateName != NULL)
-        XFREE((char*)ech->privateName, heap, DYNAMIC_TYPE_TMP_BUFFER);
+    TLSX_FreeAll(ech->extensions, heap);
 
     XFREE(ech, heap, DYNAMIC_TYPE_TMP_BUFFER);
     (void)heap;
@@ -16733,95 +16737,95 @@ int TLSX_PopulateExtensions(WOLFSSL* ssl, byte isServer)
 #if defined(WOLFSSL_TLS13) || !defined(NO_WOLFSSL_CLIENT)
 
 #if defined(WOLFSSL_TLS13) && defined(HAVE_ECH)
-static int TLSX_EchChangeSNI(WOLFSSL* ssl, TLSX** pEchX,
-                             char* serverName, TLSX** pServerNameX,
-                             TLSX*** pExtensions)
+/* Returns 1 if the extensions should be hidden for this write */
+static int TLSX_EchShouldHideInner(WOLFSSL* ssl, WOLFSSL_ECH* ech)
 {
-    int ret = 0;
-    TLSX* echX = NULL;
-    TLSX* serverNameX = NULL;
-    TLSX** extensions = NULL;
-
-    /* calculate the rest of the extensions length with inner ech */
-    if (ssl->extensions)
-        echX = TLSX_Find(ssl->extensions, TLSX_ECH);
-
-    if (echX == NULL && ssl->ctx && ssl->ctx->extensions)
-        /* if not NULL the semaphore will stop it from being counted */
-        echX = TLSX_Find(ssl->ctx->extensions, TLSX_ECH);
-
-    /* if type is outer and this is a real ECH then change sni to public name */
-    if (echX != NULL && ssl->echConfigs != NULL &&
-        ((WOLFSSL_ECH*)echX->data)->type == ECH_TYPE_OUTER) {
-        if (ssl->extensions) {
-            serverNameX = TLSX_Find(ssl->extensions, TLSX_SERVER_NAME);
-
-            if (serverNameX != NULL)
-                extensions = &ssl->extensions;
-        }
-
-        if (serverNameX == NULL && ssl->ctx && ssl->ctx->extensions) {
-            serverNameX = TLSX_Find(ssl->ctx->extensions, TLSX_SERVER_NAME);
-            if (serverNameX != NULL)
-                extensions = &ssl->ctx->extensions;
-        }
-
-        /* ECH requires an inner SNI to be present for ClientHelloInner.
-         * Without it, fail instead of mutating extension lists. */
-        if (serverNameX == NULL) {
-            ret = BAD_FUNC_ARG;
-        }
-
-        /* store the inner server name */
-        if (ret == 0 && serverNameX != NULL) {
-            char* hostName = ((SNI*)serverNameX->data)->data.host_name;
-            word32 hostNameSz = (word32)XSTRLEN(hostName) + 1;
-
-            if (hostNameSz > WOLFSSL_HOST_NAME_MAX)
-                ret = BAD_LENGTH_E;
-            else
-                XMEMCPY(serverName, hostName, hostNameSz);
-        }
-
-        /* only swap the SNI if one was found; extensions is non-NULL if an
-         * SNI entry was found on ssl->extensions or ctx->extensions */
-        if (ret == 0 && extensions != NULL) {
-            /* remove the inner server name */
-            TLSX_Remove(extensions, TLSX_SERVER_NAME, ssl->heap);
-
-            /* set the public name as the server name */
-            if ((ret = TLSX_UseSNI(extensions, WOLFSSL_SNI_HOST_NAME,
-                    ((WOLFSSL_ECH*)echX->data)->echConfig->publicName,
-                    XSTRLEN(((WOLFSSL_ECH*)echX->data)->echConfig->publicName),
-                    ssl->heap)) == WOLFSSL_SUCCESS)
-                ret = 0;
-        }
-    }
-    *pServerNameX = serverNameX;
-    *pExtensions = extensions;
-    *pEchX = echX;
-    return ret;
+    if (ech == NULL || ech->type != ECH_TYPE_OUTER)
+        return 0;
+    return ssl->options.echAccepted || ech->innerCount == 0;
 }
 
-static int TLSX_EchRestoreSNI(WOLFSSL* ssl, char* serverName,
-                              TLSX* serverNameX, TLSX** extensions)
+/* Swap matching extension types between *sslExts and *echExts.
+ *   Non-matched extensions in *echExts are prepended to *sslExts
+ *   popCount is the number of leading extensions to move from
+ *     *sslExts to *echExts
+ *
+ * Ordering is kept in mind for OuterExtensions. This is why the leading
+ * popCount extensions are 'reversed' off the list.
+ *
+ * Returns a count of extensions prepended to sslExts. */
+static word16 TLSX_EchSwapExtensions(TLSX** sslExts, TLSX** echExts,
+    word16 popCount)
 {
-    int ret = 0;
+    TLSX* chunk = NULL;
+    TLSX* node;
+    TLSX* outer;
+    TLSX* inner;
+    TLSX** outerLink;
+    TLSX** innerLink;
+    word16 appended = 0;
 
-    /* always remove the publicName SNI we injected, regardless of whether
-     * there was a prior inner SNI to restore */
-    if (extensions != NULL)
-        TLSX_Remove(extensions, TLSX_SERVER_NAME, ssl->heap);
-
-    if (serverNameX != NULL) {
-        /* restore the inner server name */
-        ret = TLSX_UseSNI(extensions, WOLFSSL_SNI_HOST_NAME,
-            serverName, XSTRLEN(serverName), ssl->heap);
-
-        if (ret == WOLFSSL_SUCCESS)
-            ret = 0;
+    /* unhook popCount nodes off *sslExts head into chunk.
+     * Head-prepend undoes the reversal caused by appending onto sslExts. */
+    while (popCount > 0 && *sslExts != NULL) {
+        node = *sslExts;
+        *sslExts = node->next;
+        node->next = chunk;
+        chunk = node;
+        popCount--;
     }
-    return ret;
+
+    outerLink = echExts;
+    while (*outerLink != NULL) {
+        innerLink = sslExts;
+        outer = *outerLink;
+
+        while (*innerLink != NULL && (*innerLink)->type != outer->type)
+            innerLink = &(*innerLink)->next;
+
+        if (*innerLink != NULL) {
+            inner = *innerLink;
+
+            *innerLink  = outer;
+            *outerLink  = inner;
+            node        = outer->next;
+            outer->next = inner->next;
+            inner->next = node;
+
+            outerLink = &inner->next;
+        }
+        else {
+            *outerLink  = outer->next;
+            outer->next = *sslExts;
+            *sslExts    = outer;
+            appended++;
+        }
+    }
+
+    /* outerLink is at the tail of *echExts; append the chunk */
+    *outerLink = chunk;
+
+    return appended;
+}
+
+/* If ECH is accepted, delete ech->extensions
+ * If rejected, replace matching ssl->extensions with ech->extensions, appending
+ *   to head if necessary */
+void TLSX_EchReplaceExtensions(WOLFSSL* ssl, byte accepted)
+{
+    TLSX* echX;
+    WOLFSSL_ECH* ech;
+
+    echX = TLSX_Find(ssl->extensions, TLSX_ECH);
+    if (echX == NULL || echX->data == NULL)
+        return;
+    ech = (WOLFSSL_ECH*)echX->data;
+
+    if (!accepted)
+        (void)TLSX_EchSwapExtensions(&ssl->extensions, &ech->extensions, 0);
+
+    TLSX_FreeAll(ech->extensions, ssl->heap);
+    ech->extensions = NULL;
 }
 
 /* Returns 1 if the extension may be encoded into ech_outer_extensions,
@@ -16931,39 +16935,43 @@ static int TLSX_ECH_BuildOuterExtensions(WOLFSSL* ssl, const byte* semaphore,
 static int TLSX_GetSizeWithEch(WOLFSSL* ssl, byte* semaphore, byte msgType,
     word16* pLength)
 {
-    int ret = 0, r = 0;
+    int ret = 0;
     TLSX* echX = NULL;
-    TLSX* serverNameX = NULL;
-    TLSX** extensions = NULL;
     WOLFSSL_ECH* ech = NULL;
     word16 count = 0;
-    WC_DECLARE_VAR(serverName, char, WOLFSSL_HOST_NAME_MAX, 0);
+    word16 appended = 0;
+    byte installed = 0;
 
-    WC_ALLOC_VAR_EX(serverName, char, WOLFSSL_HOST_NAME_MAX, NULL,
-                    DYNAMIC_TYPE_TMP_BUFFER, return MEMORY_E);
+    if (ssl->extensions)
+        echX = TLSX_Find(ssl->extensions, TLSX_ECH);
+    if (echX == NULL && ssl->ctx && ssl->ctx->extensions)
+        echX = TLSX_Find(ssl->ctx->extensions, TLSX_ECH);
+    if (echX != NULL)
+        ech = (WOLFSSL_ECH*)echX->data;
 
-    r = TLSX_EchChangeSNI(ssl, &echX, serverName, &serverNameX, &extensions);
+    if (TLSX_EchShouldHideInner(ssl, ech)) {
+        appended = TLSX_EchSwapExtensions(&ssl->extensions,
+            &ech->extensions, 0);
+        installed = 1;
+    }
 
     if (echX != NULL)
         ech = (WOLFSSL_ECH*)echX->data;
 
     /* if encoding, then count encoded form of inner ClientHello.
      * `semaphore` is in/out so encodable extensions will later be ignored */
-    if (r == 0 && ech != NULL && ech->type == ECH_TYPE_INNER &&
-            ech->writeEncoded) {
+    if (ech != NULL && ech->type == ECH_TYPE_INNER && ech->writeEncoded) {
         ret = TLSX_ECH_BuildOuterExtensions(ssl, semaphore, msgType,
             NULL, pLength, &count, semaphore);
     }
-    if (r == 0 && ret == 0 && ssl->extensions)
+    if (ret == 0 && ssl->extensions)
         ret = TLSX_GetSize(ssl->extensions, semaphore, msgType, pLength);
-    if (r == 0 && ret == 0 && ssl->ctx && ssl->ctx->extensions)
+    if (ret == 0 && ssl->ctx && ssl->ctx->extensions)
         ret = TLSX_GetSize(ssl->ctx->extensions, semaphore, msgType, pLength);
-    if (r == 0)
-        r = TLSX_EchRestoreSNI(ssl, serverName, serverNameX, extensions);
 
-    WC_FREE_VAR_EX(serverName, ssl->heap, DYNAMIC_TYPE_TMP_BUFFER);
-    if (ret == 0 && r != 0)
-        ret = r;
+    if (installed)
+        (void)TLSX_EchSwapExtensions(&ssl->extensions, &ech->extensions,
+            appended);
     return ret;
 }
 #endif
@@ -17093,19 +17101,26 @@ int TLSX_GetRequestSize(WOLFSSL* ssl, byte msgType, word32* pLength)
 static int TLSX_WriteWithEch(WOLFSSL* ssl, byte* output, byte* semaphore,
     byte msgType, word16* pOffset)
 {
-    int r = 0, ret = 0;
+    int ret = 0;
     TLSX* echX = NULL;
-    TLSX* serverNameX = NULL;
-    TLSX** extensions = NULL;
     WOLFSSL_ECH* ech = NULL;
-    WC_DECLARE_VAR(serverName, char, WOLFSSL_HOST_NAME_MAX, 0);
+    word16 appended = 0;
+    byte installed = 0;
 
-    WC_ALLOC_VAR_EX(serverName, char, WOLFSSL_HOST_NAME_MAX, NULL,
-                    DYNAMIC_TYPE_TMP_BUFFER, return MEMORY_E);
-    r = TLSX_EchChangeSNI(ssl, &echX, serverName, &serverNameX, &extensions);
-    ret = r;
-    if (ret == 0 && echX != NULL) {
+    if (ssl->extensions)
+        echX = TLSX_Find(ssl->extensions, TLSX_ECH);
+    if (echX == NULL && ssl->ctx && ssl->ctx->extensions)
+        echX = TLSX_Find(ssl->ctx->extensions, TLSX_ECH);
+    if (echX != NULL)
         ech = (WOLFSSL_ECH*)echX->data;
+
+    if (TLSX_EchShouldHideInner(ssl, ech)) {
+        appended = TLSX_EchSwapExtensions(&ssl->extensions,
+            &ech->extensions, 0);
+        installed = 1;
+    }
+
+    if (echX != NULL) {
         /* turn ech on so it doesn't write, then write it last */
         TURN_ON(semaphore, TLSX_ToSemaphore(echX->type));
     }
@@ -17157,7 +17172,7 @@ static int TLSX_WriteWithEch(WOLFSSL* ssl, byte* output, byte* semaphore,
         /* turn off and write it last */
         TURN_OFF(semaphore, TLSX_ToSemaphore(echX->type));
 
-        if (ret == 0 && ssl->extensions) {
+        if (ssl->extensions) {
             ret = TLSX_Write(ssl->extensions, output + *pOffset, semaphore,
                 msgType, pOffset);
         }
@@ -17168,12 +17183,9 @@ static int TLSX_WriteWithEch(WOLFSSL* ssl, byte* output, byte* semaphore,
         }
     }
 
-    if (r == 0)
-        r = TLSX_EchRestoreSNI(ssl, serverName, serverNameX, extensions);
-    WC_FREE_VAR_EX(serverName, ssl->heap, DYNAMIC_TYPE_TMP_BUFFER);
-
-    if (ret == 0 && r != 0)
-        ret = r;
+    if (installed)
+        (void)TLSX_EchSwapExtensions(&ssl->extensions, &ech->extensions,
+            appended);
     return ret;
 }
 #endif
--- a/src/tls.c
+++ b/src/tls.c
@@ -1737,14 +1737,23 @@ int TLSX_HandleUnsupportedExtension(WOLFSSL* ssl)
 #endif
 
 #if !defined(NO_WOLFSSL_SERVER) || defined(WOLFSSL_TLS13)
+static void TLSX_SetResponseInList(TLSX* list, TLSX_Type type);
+/** Mark an extension to be sent back to the client.
+ *  Operates on a list instead of the ssl.
+ *      (Should only be used on ssl->extensions or ech->extensions) */
+static void TLSX_SetResponseInList(TLSX* list, TLSX_Type type)
+{
+    TLSX *extension = TLSX_Find(list, type);
+
+    if (extension)
+        extension->resp = 1;
+}
+
 void TLSX_SetResponse(WOLFSSL* ssl, TLSX_Type type);
 /** Mark an extension to be sent back to the client. */
 void TLSX_SetResponse(WOLFSSL* ssl, TLSX_Type type)
 {
-    TLSX *extension = TLSX_Find(ssl->extensions, type);
-
-    if (extension)
-        extension->resp = 1;
+    TLSX_SetResponseInList(ssl->extensions, type);
 }
 #endif
 
@@ -2384,10 +2393,10 @@ static int TLSX_SNI_Parse(WOLFSSL* ssl, const byte* input, word16 length,
     word16 size = 0;
     word16 offset = 0;
     int cacheOnly = 0;
-    SNI *sni = NULL;
-    const char *hostName = NULL;
+    int checkPublic = 0;
+    SNI* sni = NULL;
     byte type;
-    byte matched;
+    byte matched = 0;
 #if defined(WOLFSSL_TLS13) && defined(HAVE_ECH)
     TLSX* echX = NULL;
     WOLFSSL_ECH* ech = NULL;
@@ -2424,7 +2433,7 @@ static int TLSX_SNI_Parse(WOLFSSL* ssl, const byte* input, word16 length,
 
 #ifndef NO_WOLFSSL_SERVER
 #if defined(WOLFSSL_TLS13) && defined(HAVE_ECH)
-    if (!ssl->options.disableECH) {
+    if (!ssl->options.disableECH && !ssl->options.echProcessingInner) {
         echX = TLSX_Find(ssl->extensions, TLSX_ECH);
         if (echX != NULL) {
             ech = (WOLFSSL_ECH*)(echX->data);
@@ -2448,8 +2457,20 @@ static int TLSX_SNI_Parse(WOLFSSL* ssl, const byte* input, word16 length,
             WOLFSSL_MSG("Forcing SSL object to store SNI parameter");
         }
         else {
-            /* Skipping, SNI not enabled at server side. */
-            return 0;
+        #if defined(WOLFSSL_TLS13) && defined(HAVE_ECH)
+            if (ech == NULL)
+        #endif
+            {
+                /* Skipping, SNI not enabled at server side. */
+                return 0;
+            }
+
+        #if defined(WOLFSSL_TLS13) && defined(HAVE_ECH)
+            /* No server SNI configured but ECH is active:
+             * the outer SNI still needs to be matched against the echConfig
+             * publicName and recorded on ech->extensions. */
+            checkPublic = 1;
+        #endif
         }
     }
 
@@ -2477,7 +2498,8 @@ static int TLSX_SNI_Parse(WOLFSSL* ssl, const byte* input, word16 length,
     if (offset + size != length || size == 0)
         return BUFFER_ERROR;
 
-    if (!cacheOnly && !(sni = TLSX_SNI_Find((SNI*)extension->data, type)))
+    if (!cacheOnly && !checkPublic &&
+            !(sni = TLSX_SNI_Find((SNI*)extension->data, type)))
         return 0; /* not using this type of SNI. */
 
 #if defined(WOLFSSL_TLS13)
@@ -2488,34 +2510,42 @@ static int TLSX_SNI_Parse(WOLFSSL* ssl, const byte* input, word16 length,
         return 0;
 #endif
 
-    hostName = (sni != NULL) ? sni->data.host_name : NULL;
-    matched = (hostName != NULL &&
-        XSTRLEN(hostName) == size &&
-        XSTRNCMP(hostName, (const char*)input + offset, size) == 0);
-
 #if defined(WOLFSSL_TLS13) && defined(HAVE_ECH)
     /* While parsing the outer CH accept a match against any
      * echConfig publicName */
-    if (!matched && ech != NULL && !ssl->options.echProcessingInner) {
+    if (ech != NULL) {
         workingConfig = ech->echConfig;
         while (workingConfig != NULL) {
-            matched = XSTRLEN(workingConfig->publicName) == size &&
-                XSTRNCMP(workingConfig->publicName,
-                (const char*)input + offset, size) == 0;
-            if (matched)
+            if (XSTRLEN(workingConfig->publicName) == size &&
+                    XSTRNCMP(workingConfig->publicName,
+                    (const char*)input + offset, size) == 0) {
+                matched = 1;
                 break;
+            }
             workingConfig = workingConfig->next;
         }
 
         /* If a publicName is matched then this SNI is not something that should
-         * be forcibly cached */
+         * be forcibly cached. This allows an SNI response to be given for the
+         * public name */
         if (matched)
             cacheOnly = 0;
     }
-#endif
-
     if (!matched)
-        matched = cacheOnly;
+#endif
+    {
+        const char* hostName;
+        hostName = (sni != NULL) ? sni->data.host_name : NULL;
+        matched = cacheOnly || (hostName != NULL &&
+            XSTRLEN(hostName) == size &&
+            XSTRNCMP(hostName, (const char*)input + offset, size) == 0);
+    }
+
+    /* No server SNI configured and the outer name did not match a publicName:
+     * stay permissive and record nothing. If ECH is accepted, the absent
+     * publicName match is caught after the outer parse. */
+    if (!matched && checkPublic)
+        return 0;
 
     if (matched ||
             (sni != NULL && (sni->options & WOLFSSL_SNI_ANSWER_ON_MISMATCH))) {
@@ -2523,6 +2553,7 @@ static int TLSX_SNI_Parse(WOLFSSL* ssl, const byte* input, word16 length,
         int r;
         TLSX** writeList = &ssl->extensions;
 #if defined(WOLFSSL_TLS13) && defined(HAVE_ECH)
+        /* install onto ech->extensions if the public name was matched */
         if (workingConfig != NULL)
             writeList = &ech->extensions;
 #endif
@@ -2547,12 +2578,8 @@ static int TLSX_SNI_Parse(WOLFSSL* ssl, const byte* input, word16 length,
 
         TLSX_SNI_SetStatus(*writeList, type, (byte)matchStat);
 
-        if (!cacheOnly) {
-            extension = TLSX_Find(*writeList, TLSX_SERVER_NAME);
-
-            if (extension)
-                extension->resp = 1;
-        }
+        if (!cacheOnly)
+            TLSX_SetResponseInList(*writeList, TLSX_SERVER_NAME);
     }
     else if ((sni == NULL) ||
             !(sni->options & WOLFSSL_SNI_CONTINUE_ON_MISMATCH)) {
@@ -2672,8 +2699,8 @@ int TLSX_UseSNI(TLSX** extensions, byte type, const void* data, word16 size,
     return WOLFSSL_SUCCESS;
 }
 
-#ifndef NO_WOLFSSL_SERVER
-
+/* client-side needs this function when ECH is enabled */
+#if !defined(NO_WOLFSSL_SERVER) || defined(HAVE_ECH)
 /** Tells the SNI requested by the client. */
 word16 TLSX_SNI_GetRequest(TLSX* extensions, byte type, void** data,
         byte ignoreStatus)
@@ -2693,7 +2720,9 @@ word16 TLSX_SNI_GetRequest(TLSX* extensions, byte type, void** data,
 
     return 0;
 }
+#endif
 
+#ifndef NO_WOLFSSL_SERVER
 /** Sets the options for a SNI object. */
 void TLSX_SNI_SetOptions(TLSX* extensions, byte type, byte options)
 {
@@ -13810,11 +13839,9 @@ static int TLSX_GreaseECH_Use(TLSX** extensions, void* heap, WC_RNG* rng)
 
     ech = (WOLFSSL_ECH*)XMALLOC(sizeof(WOLFSSL_ECH), heap,
         DYNAMIC_TYPE_TMP_BUFFER);
-
     if (ech == NULL)
         return MEMORY_E;
-
-    ForceZero(ech, sizeof(WOLFSSL_ECH));
+    XMEMSET(ech, 0, sizeof(WOLFSSL_ECH));
 
     ech->state = ECH_WRITE_GREASE;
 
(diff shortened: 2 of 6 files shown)


src/tls.c
```diff
@@ -14716,15 +14703,13 @@ static void TLSX_ECH_Free(WOLFSSL_ECH* ech, void* heap)
 {
     XFREE(ech->innerClientHello, heap, DYNAMIC_TYPE_TMP_BUFFER);
     if (ech->hpke != NULL) {
-        if (ech->ephemeralKey != NULL)
-            wc_HpkeFreeKey(ech->hpke, ech->hpke->kem, ech->ephemeralKey,
-                ech->hpke->heap);
+        wc_HpkeFreeKey(ech->hpke, ech->hpke->kem, ech->ephemeralKey,
+            ech->hpke->heap);
         XFREE(ech->hpke, heap, DYNAMIC_TYPE_TMP_BUFFER);
```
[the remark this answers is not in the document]
sebastian-carpenter: `wc_HpkeFreeKey` does do the NULL check

src/tls.c
```diff
@@ -13732,19 +13712,26 @@ static int TLSX_ECH_Use(WOLFSSL_EchConfig* echConfig, TLSX** extensions,
...
     }
+    if (ret == 0)
+        ret = TLSX_Push(extensions, TLSX_ECH, ech, heap);
     if (ret != 0) {
+        TLSX_FreeAll(ech->extensions, heap);
+        wc_HpkeFreeKey(ech->hpke, ech->hpke->kem, ech->ephemeralKey,
+            ech->hpke->heap);
         XFREE(ech->hpke, heap, DYNAMIC_TYPE_TMP_BUFFER);
         XFREE(ech, heap, DYNAMIC_TYPE_TMP_BUFFER);
     }
```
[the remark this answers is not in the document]
sebastian-carpenter: `wc_HpkeFreeKey` does do the NULL check

tests/api.c
```diff
@@ -14608,34 +14692,31 @@ static int test_wolfSSL_Tls13_ECH_disable_conn_ex(int enableServer,
...
     test_ssl_memio_ctx test_ctx;
-    /* 300 chars > MAX_PUBLIC_NAME_SZ (256) to exercise truncation */
-    char longName[300];
-
-    XMEMSET(longName, 'a', sizeof(longName) - 1);
-    longName[sizeof(longName) - 1] = '\0';
+    WOLFSSL_CTX* tempCtx = NULL;
+    byte badConfig[128];
+    word32 badConfigLen = sizeof(badConfig);
+    const char* expectedSni =
```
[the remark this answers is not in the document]
sebastian-carpenter: not really an issue. 512 is too big anyway

src/tls.c
```diff
@@ -13833,19 +13822,26 @@ static int TLSX_ECH_Use(WOLFSSL_EchConfig* echConfig, TLSX** extensions,
...
+    /* use the chosen config's public name for the outer SNI */
     if (ret == 0) {
-        ret = TLSX_Push(extensions, TLSX_ECH, ech, heap);
-        if (ret != 0) {
-            wc_HpkeFreeKey(ech->hpke, ech->hpke->kem, ech->ephemeralKey,
-                ech->hpke->heap);
-        }
+        ret = TLSX_UseSNI(&ech->extensions, WOLFSSL_SNI_HOST_NAME,
+            echConfig->publicName, (word16)XSTRLEN(echConfig->publicName),
+            heap);
```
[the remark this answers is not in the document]
sebastian-carpenter: sni is enabled with --enable-ech. ECH can't be used without SNI anyway since there must be a public name in the outer hello.

src/tls13.c
```diff
@@ -4839,8 +4841,9 @@ int SendTls13ClientHello(WOLFSSL* ssl)
                 return ret;
 
             /* calculate padding (RFC 9849, section 6.1.3) */
-            if (args->ech->privateName != NULL) {
-                word16 nameLen = (word16)XSTRLEN(args->ech->privateName);
+            nameLen = TLSX_SNI_GetRequest(ssl->extensions,
+                WOLFSSL_SNI_HOST_NAME, &hostName, 1);
+            if (hostName != NULL) {
```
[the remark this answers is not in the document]
sebastian-carpenter: SNI is a necessity for ECH

sebastian-carpenter:
Jenkins retest this please.

src/tls.c
```diff
@@ -16656,95 +16662,95 @@ int TLSX_PopulateExtensions(WOLFSSL* ssl, byte isServer)
...
+/* Swap matching extension types between *sslExts and *echExts.
+ *   Non-matched extensions in *echExts are prepended to *sslExts
+ *   popCount is the number of leading extensions to move from
+ *     *sslExts to *echExts
+ *
+ * Ordering is kept in mind for OuterExtensions. This is why the leading
+ * popCount extensions are 'reversed' off the list.
+ *
+ * Returns a count of extensions prepended to sslExts. */
+static word16 TLSX_EchSwapExtensions(TLSX** sslExts, TLSX** echExts,
```
dgarske: **🟠 [Medium] Multi-public-extension swap/reversal path is untested**

`TLSX_EchSwapExtensions` contains non-trivial pointer manipulation specifically designed for multiple public extensions: the `popCount` head-pop/reverse loop and the prepend-and-reverse ordering logic exist to preserve OuterExtensions ordering when more than one public extension is present. The PR body explicitly states this lays "groundwork for managing multiple public extensions and multiple SNI's". However, in the current code `ech->extensions` only ever holds a single SNI (populated in `TLSX_ECH_Use` and `TLSX_SNI_Parse`), so the multi-node `popCount > 1` reversal path and the multi-prepend ordering are never exercised by any test. The new `test_wolfSSL_Tls13_ECH_wire_sni` only validates the single-SNI case. Since this is the most intricate part of the change, a regression here would be silent until the future multi-ext work lands.

**Fix:** Add a small unit test for the multi-extension swap/reverse ordering to lock in the reversal behavior the PR is building toward.

src/tls13.c
```diff
@@ -4839,8 +4841,9 @@ int SendTls13ClientHello(WOLFSSL* ssl)
                 return ret;
 
             /* calculate padding (RFC 9849, section 6.1.3) */
-            if (args->ech->privateName != NULL) {
-                word16 nameLen = (word16)XSTRLEN(args->ech->privateName);
+            nameLen = TLSX_SNI_GetRequest(ssl->extensions,
```
dgarske, line 4848: **🔵 [Low] ECH padding length only considers ssl-extensions, not ctx-level inner SNI**

The replacement padding calculation reads the inner SNI via `TLSX_SNI_GetRequest(ssl->extensions, ...)`. This only inspects `ssl->extensions`, whereas a client may configure its SNI at the CTX level via `wolfSSL_CTX_UseSNI` (which lands in `ssl->ctx->extensions`). In that configuration `hostName` will be NULL here and the code takes the `paddingLen = maxNameLen + 9` branch as if no inner name were present. The effect is benign (more padding, never a buffer issue, and the actual outer SNI on the wire is still correct due to the semaphore override in `TLSX_GetSize`/`TLSX_Write`), but the padding no longer reflects the true inner name length, slightly weakening the length-hiding intent of RFC 9849 6.1.3 for ctx-level-SNI clients. The old code used `ech->privateName`. Worth confirming this is intentional / acceptable.

**Fix:** Confirm ctx-level inner SNI is out of scope, or add a ctx->extensions fallback for the padding length calculation.

src/tls.c
```diff
@@ -2375,13 +2375,13 @@ static int TLSX_SNI_Parse(WOLFSSL* ssl, const byte* input, word16 length,
     word16 offset = 0;
     int cacheOnly = 0;
     SNI *sni = NULL;
+    const char *hostName = NULL;
```
dgarske: **⚪ [Info] Pointer declaration style inconsistent with repo convention**

The newly introduced declaration uses `const char *hostName = NULL;` (asterisk bound to the name), whereas the surrounding wolfSSL code and the original removed line used `const char* hostName` (asterisk bound to the type). Minor consistency nit only.

**Fix:** Match the `const char* name` pointer style used elsewhere in the file.

src/tls.c
```diff
@@ -16656,95 +16662,95 @@ int TLSX_PopulateExtensions(WOLFSSL* ssl, byte isServer)
 #if defined(WOLFSSL_TLS13) || !defined(NO_WOLFSSL_CLIENT)
 
 #if defined(WOLFSSL_TLS13) && defined(HAVE_ECH)
-static int TLSX_EchChangeSNI(WOLFSSL* ssl, TLSX** pEchX,
-                             char* serverName, TLSX** pServerNameX,
-                             TLSX*** pExtensions)
+/* Returns 1 if the extensions should be hidden for this write */
```
dgarske, line 16763: **⚪ [Info] TLSX_EchShouldHideInner can be a single boolean expression**

The helper is a two-branch return of a simple predicate. It can be collapsed to a single expression for readability without changing behavior. Purely cosmetic.

**Fix:** Optionally simplify to a single return; take it or leave it.

dgarske (commented):
## Skoll Code Review

### Posted findings
- [Medium] **Multi-public-extension swap/reversal path is untested** — `src/tls.c:16682-16734`
- [Low] **ECH padding length only considers ssl-extensions, not ctx-level inner SNI** — `src/tls13.c:4844-4855`
- [Info] **Pointer declaration style inconsistent with repo convention** — `src/tls.c:2378`
- [Info] **TLSX_EchShouldHideInner can be a single boolean expression** — `src/tls.c:16665-16671`

src/tls.c
sebastian-carpenter: Added testing (more than a little...)

src/tls13.c
sebastian-carpenter: Updated. Now checks the ctx if SNI is not found on the ssl

src/tls.c
sebastian-carpenter: fixed

src/tls.c
sebastian-carpenter: updated

dgarske (commented):
## Skoll Code Review

### Posted findings
- [Low] **TLSX_FreeAll echList tracks only a single ECH extension per list** — `src/tls.c:15190-15216`
- [Low] **BAD_STATE_E on restore-swap leaves ssl-extensions inconsistent without repair** — `src/tls.c:16895-16901, 17111-17117`

src/tls.c
```diff
@@ -15199,6 +15190,7 @@ void TLSX_FreeAll(TLSX* list, void* heap)
 #if defined(WOLFSSL_TLS13) && defined(HAVE_ECH)
```
dgarske, line 15279: **🔵 [Low] TLSX_FreeAll echList tracks only a single ECH extension per list**

The new deferred-free logic stores `ech->extensions` into a single local `echList` when an `TLSX_ECH` node is encountered, and only consumes it once the main `list` is exhausted. If a list ever contained two `TLSX_ECH` nodes, the second assignment to `echList` would overwrite the first, leaking the first ECH's `extensions` sublist. In practice this is unreachable because `TLSX_Push`/`TLSX_Find` guarantee a single ECH extension per list, so this is a latent robustness gap rather than an active leak. The approach itself (iterative free to avoid a `TLSX_ECH_Free` -> `TLSX_FreeAll` recursive chain) is good and respects the no-recursion convention.

**Fix:** Optional. Add a brief comment noting the single-ECH-per-list invariant the logic relies on, or make echList accumulate. Not required for merge.

src/tls.c
```diff
@@ -16864,39 +16863,42 @@ static int TLSX_ECH_BuildOuterExtensions(WOLFSSL* ssl, const byte* semaphore,
...
-    if (r == 0 && ret == 0 && ssl->ctx && ssl->ctx->extensions)
+    if (ret == 0 && ssl->ctx && ssl->ctx->extensions)
         ret = TLSX_GetSize(ssl->ctx->extensions, semaphore, msgType, pLength);
-    if (r == 0)
-        r = TLSX_EchRestoreSNI(ssl, serverName, serverNameX, extensions);
 
-    WC_FREE_VAR_EX(serverName, ssl->heap, DYNAMIC_TYPE_TMP_BUFFER);
-    if (ret == 0 && r != 0)
-        ret = r;
+    if (installed) {
```
dgarske: **🔵 [Low] BAD_STATE_E on restore-swap leaves ssl-extensions inconsistent without repair**

After computing size/writing, the restore swap `TLSX_EchSwapExtensions(&ssl->extensions, &ech->extensions, prepended)` is expected to return 0; if it returns non-zero, `ret` is set to `BAD_STATE_E`. I verified this is unreachable for valid inputs (the displaced inner extensions always re-match the swapped-in public extensions on the way back, since `ech->extensions` cannot contain duplicate types), so the check is purely defensive. The note is only that, were it ever to trigger, `ssl->extensions` would be left in a mutated/inconsistent state (a stranded public extension or a missing inner extension) and the function returns an error without restoring the list. Because the error aborts the handshake and full teardown frees everything, there is no leak — just a non-obvious failure mode.

**Fix:** Optional clarifying comment. No code change required.

src/tls.c
sebastian-carpenter: Should be impossible to have a duplicate, but I decided to deal with that case and reworked the free mechanism. Now the list will either be appended to or be set to echList depending on if ECH is at the end of list.

src/tls.c
sebastian-carpenter: Added WOLFSSL_MSG("Bad restore with TLSX_EchSwapExtensions");

src/tls.c
```diff
@@ -2516,42 +2488,46 @@ static int TLSX_SNI_Parse(WOLFSSL* ssl, const byte* input, word16 length,
...
+         * be forcibly cached */
+        if (matched)
+            cacheOnly = 0;
     }
 #endif
 
+    if (!matched)
+        matched = cacheOnly;
+
     if (matched ||
```
dgarske, line 2550: **🔴 [High] Servers with WOLFSSL_SNI_ABORT_ON_ABSENCE now abort every ECH handshake at the outer ClientHello, even when a valid…**

The PR redirects a publicName-matched outer SNI into `ech->extensions` (`writeList = &ech->extensions` when `workingConfig != NULL`), and `TLSX_SNI_SetStatus(*writeList, ...)` records `WOLFSSL_SNI_REAL_MATCH` only on that private copy. The configured SNI node in `ssl->extensions`/`ctx->extensions` is left at `WOLFSSL_SNI_NO_MATCH`. `TLSX_SNI_VerifyParse` (called unconditionally at the end of every `TLSX_Parse`, src/tls.c:18685) only consults `ssl->extensions`/`ctx->extensions`, so when the server has `WOLFSSL_SNI_ABORT_ON_ABSENCE` set it returns `SNI_ABSENT_ERROR` while parsing the OUTER ClientHello -- before the inner hello (which may carry a perfectly valid private SNI) is ever processed. In the old code the publicName match was written into `ssl->extensions` via `TLSX_UseSNI(&ssl->extensions, ...)` with `REAL_MATCH`, so the outer VerifyParse passed and the inner SNI was then evaluated. EMPIRICALLY VERIFIED on this branch (--enable-ech build): adding `wolfSSL_UseSNI(c_ssl, WOLFSSL_SNI_HOST_NAME, echCbTestPrivateName, ...)` to the ABORT_ON_ABSENCE case of `test_wolfSSL_Tls13_ECH_no_private_name` still fails the handshake with `error = -402, No Server Name Indication extension Error`, i.e. the abort fires on CH1's outer parse regardless of the inner SNI. This affects both the accept path (server can never accept ECH when ABORT_ON_ABSENCE is set) and the reject path (server aborts instead of continuing the outer handshake with retry configs).

**Fix:** Make the outer-CH ABORT_ON_ABSENCE check aware of the public-SNI match now stored in ech->extensions (or defer SNI_VERIFY_PARSE enforcement to the inner CH when ECH decryption succeeded, or mirror the match status onto the ssl-level SNI for the outer pass). Add a regression test: server with ABORT_ON_ABSENCE + ECH, client with private inner SNI -> handshake must succeed with ECH ACCEPTED.

tests/api.c
```diff
@@ -14297,14 +14622,19 @@ static int test_wolfSSL_Tls13_ECH_no_private_name(void)
 
     ExpectIntEQ(test_ssl_memio_setup(&test_ctx), TEST_SUCCESS);
 
+    /* SNI is permissive by default, force a failure when SNI is absent */
```
dgarske: **🟠 [Medium] test_wolfSSL_Tls13_ECH_no_private_name ABORT_ON_ABSENCE case encodes the outer-parse abort; missing test for…**

The comment says 'SNI is permissive by default, force a failure when SNI is absent', implying the failure is caused by the missing inner SNI. In reality the `SNI_ABSENT_ERROR` triggers during the OUTER ClientHello parse (the inner hello is never consulted), so the same failure occurs even when an inner SNI is present and valid (verified empirically). The test as written passes for the wrong reason and locks in the regressed behavior described in the BLOCK finding. There is no test covering the combination ABORT_ON_ABSENCE + inner SNI present, which is exactly the case that regressed.

**Fix:** After fixing the BLOCK finding, add the ABORT_ON_ABSENCE + inner-SNI-present scenario asserting handshake success and ECH ACCEPTED on both sides, and fix the misleading comment on the absent-SNI case.

src/tls.c
```diff
@@ -2569,10 +2545,16 @@ static int TLSX_SNI_Parse(WOLFSSL* ssl, const byte* input, word16 length,
             matchStat = WOLFSSL_SNI_FAKE_MATCH;
         }
 
-        TLSX_SNI_SetStatus(ssl->extensions, type, (byte)matchStat);
+        TLSX_SNI_SetStatus(*writeList, type, (byte)matchStat);
+
+        if (!cacheOnly) {
```
dgarske: **🔵 [Low] Inline reimplementation of TLSX_SetResponse in TLSX_SNI_Parse**

The code comments 'This is just a reimplementation of TLSX_SetResponse to accept a TLSX* instead' and then open-codes the find-and-set. A small list-based helper (e.g. `TLSX_SetResponseInList(TLSX* list, TLSX_Type type)`) that `TLSX_SetResponse` also calls would remove the duplication and keep response-flag semantics in one place for future public-extension types (the PR's stated follow-up direction).

**Fix:** Extract a shared helper; not blocking.

src/tls13.c
```diff
@@ -5300,6 +5307,12 @@ static int EchCheckAcceptance(WOLFSSL* ssl, byte* label, word16 labelSz,
         ssl->hsHashes = tmpHashes;
     }
 
+    /* Skip only when the HRR signals ECH acceptance
```
dgarske, line 5317: **🔵 [Low] EchCheckAcceptance calls TLSX_EchReplaceExtensions even when EchCalcAcceptance failed**

The new replace call at the end of `EchCheckAcceptance` runs even when `EchCalcAcceptance` returned an error (the `if (ret == 0)` block above was skipped, `echAccepted` may be stale). It is harmless because the handshake aborts on the non-zero return, but swapping the public extensions into `ssl->extensions` on a hard-error path is unnecessary state mutation and makes the function's failure behavior harder to reason about.

**Fix:** Gate the replace on ret == 0 for clarity; behavior is unchanged either way since the connection aborts on error.

.github/scripts/openssl-ech.sh
```diff
@@ -84,9 +89,12 @@ WOLFSSL_CLIENT=${WOLFSSL_CLIENT:-"$WORKSPACE/examples/client/client"}
 WOLFSSL_SERVER=${WOLFSSL_SERVER:-"$WORKSPACE/examples/server/server"}
 CERT_DIR=${CERT_DIR:-"$WORKSPACE/certs"}
 
+# correct ECH config, but it's old, ECH will be rejected
+REJECT_ECH_CONFIG="AD7+DQA6rAAgACCATZdDlHed6GlDeiYsu3r7sdWUkLVHZuTa3lbOf+hIbAAEAAEAAQALZXhhbXBsZS5jb20AAA=="
+
 TMP_LOG="$WORKSPACE/tmp_file.log"
 PRIV_NAME="ech-private-name.com"
-PUB_NAME="ech-public-name.com"
+PUB_NAME="example.com"
```
dgarske, line 101: **⚪ [Info] PUB_NAME changed to example.com for all interop scenarios, not just the new reject tests**

`PUB_NAME` was changed from "ech-public-name.com" to "example.com" so it matches the public name embedded in the hard-coded `REJECT_ECH_CONFIG`. This silently changes the public name used by every pre-existing interop scenario (suite tests, HRR, PQC), slightly reducing name diversity in the non-reject coverage. If intentional, a brief comment tying PUB_NAME to REJECT_ECH_CONFIG would help; otherwise the reject path could use its own name variable. The `set -e` at the top of the script does make the chained `grep -q` assertions effective, so the new reject verification logic is sound.

**Fix:** Add a comment documenting the coupling between PUB_NAME and REJECT_ECH_CONFIG, or scope the example.com name to the reject paths.

dgarske (commented):
## Skoll Code Review

**Findings:** 6 total — 6 posted, 0 skipped

### Posted findings
- [High] **Servers with WOLFSSL_SNI_ABORT_ON_ABSENCE now abort every ECH handshake at the outer ClientHello, even when a valid inner SNI is present** — `src/tls.c:2520-2557`
- [Medium] **test_wolfSSL_Tls13_ECH_no_private_name ABORT_ON_ABSENCE case encodes the outer-parse abort; missing test for ABORT_ON_ABSENCE with inner SNI present** — `tests/api.c:14625-14639`
- [Low] **BAD_STATE_E restore-failure paths log with WOLFSSL_MSG only, no WOLFSSL_ERROR_VERBOSE** — `src/tls.c:16900-16907, src/tls.c:17117-17124`
- [Low] **Inline reimplementation of TLSX_SetResponse in TLSX_SNI_Parse** — `src/tls.c:2550-2557`
- [Low] **EchCheckAcceptance calls TLSX_EchReplaceExtensions even when EchCalcAcceptance failed** — `src/tls13.c:5310-5314`
- [Info] **PUB_NAME changed to example.com for all interop scenarios, not just the new reject tests** — `.github/scripts/openssl-ech.sh:97-99`

dgarske (changes requested):
See these skoll review items -> https://github.com/wolfSSL/wolfssl/pull/10568#pullrequestreview-4479728910

src/tls.c
sebastian-carpenter: Addressed. Moved the block that replaces the ech extensions to happen before the SNI_VERIFY.

Added WOLFSSL_SNI_ABORT_ON_ABSENCE to the wire_sni test which is expected to pass, thus testing the desired accept case. Correct connections should no longer fail.

tests/api.c
sebastian-carpenter: added sni option to wire_sni test.

src/tls.c
sebastian-carpenter: wrote the given helper function, removed inline code and replaced.

src/tls13.c
sebastian-carpenter: added the `ret == 0` guard

.github/scripts/openssl-ech.sh
sebastian-carpenter: documented. Also explained why PRIV_NAME is what it is

src/tls.c
```diff
@@ -18666,6 +18709,60 @@ WOLFSSL_TEST_VIS int TLSX_Parse(WOLFSSL* ssl, const byte* input, word16 length,
     }
 #endif
 
+#if defined(WOLFSSL_TLS13) && defined(HAVE_ECH)
+    /* Reconcile ECH inner/outer extensions before verifying SNI so the verify
+     * pass sees the authoritative list */
+    if (ret == 0 && msgType == client_hello && isRequest &&
```
dgarske, line 18825: **🟠 [Medium] ECH reconciliation block in TLSX_Parse omits the TLS1.3 version guard used by sibling SNI logic**

The new ECH accept/reject reconciliation block runs whenever `msgType == client_hello && isRequest && !echProcessingInner && ssl->ctx->echConfigs != NULL && !ssl->options.disableECH`. Unlike the parallel logic this PR added in `TLSX_SNI_Parse` (line ~2517, which explicitly checks `IsAtLeastTLSv1_3(ssl->version)` with the comment "A TLS 1.2 hello never gets an inner pass..."), this block has no version guard. It is currently benign only because the server's `TLSX_ECH` is created exclusively under `IsAtLeastTLSv1_3` in `TLSX_PopulateExtensions` (so `ech` resolves to NULL for a downgraded TLS 1.2 ClientHello and the block no-ops). Relying on that implicit invariant is more fragile than the explicit check used a few thousand lines earlier in the same file. Adding the guard would make the two ECH code paths consistent and self-documenting.

**Fix:** Add `IsAtLeastTLSv1_3(ssl->version)` to the guard so the reconciliation block matches the defensive version check the PR added to `TLSX_SNI_Parse`.

dgarske (commented):
## Skoll Code Review

### Posted findings
- [Medium] **ECH reconciliation block in TLSX_Parse omits the TLS1.3 version guard used by sibling SNI logic** — `src/tls.c:18715-18717`
- [Low] **Duplicated client/server invocation block in openssl-ech.sh reject branches** — `.github/scripts/openssl-ech.sh:175-197,275-307`

.github/scripts/openssl-ech.sh
```diff
@@ -158,17 +172,29 @@ openssl_server(){
...
-        -S "$PRIV_NAME" \
-        --ech "$ech_config" \
-        $wolfssl_extra \
-        &>> "$TMP_LOG"
-
     rm -f "$ech_file"
 
-    grep -q "ech_success=1" "$TMP_LOG"
+    # test with wolfssl client
+    if [ "$REJECT" -ne 0 ]; then
```
dgarske, line 187: **🔵 [Low] Duplicated client/server invocation block in openssl-ech.sh reject branches**

Both `openssl_server()` and `openssl_client()` now wrap the test invocation in an `if [ "$REJECT" -ne 0 ]; then ... else ... fi` where the only differences between the two branches are `|| true` and the expected grep string. The full multi-line `$WOLFSSL_CLIENT` / `$OPENSSL s_client` command is copy-pasted in each branch, so a future flag change must be made in two places. This is CI-only scripting (not src/*.c), so it is purely a maintainability nit.

**Fix:** Optionally de-duplicate the invocation to a single command and branch only on the success assertion. Not blocking.

dgarske (changes requested):
https://github.com/wolfSSL/wolfssl/pull/10568#pullrequestreview-4487553608
Make sure you rebase to latest master to pull in CI optimizations.

dgarske (commented):
## Skoll Multi-Scan Review

### Posted findings
- [Medium] [review] **Server now hard-rejects an accepted ECH handshake when the outer ClientHello SNI did not match a publicName** — `src/tls.c:18724-18741`
- [Medium] [review] **New UNKNOWN_SNI_HOST_NAME_E error path is not exercised by tests** — `src/tls.c:18727-18734`
- [Low] [review] **Duplicated install/restore swap boilerplate in TLSX_GetSizeWithEch and TLSX_WriteWithEch** — `src/tls.c:16911-16935, 17076-17152`
- [Info] [review-security] **CTX-level private SNI suppression in ECH outer hello relies on implicit semaphore ordering and is untested** — `src/tls.c:16896-16937 (TLSX_GetSizeWithEch), src/tls.c:17062-17153 (TLSX_WriteWithEch)`

src/tls.c
```diff
@@ -18666,6 +18709,60 @@ WOLFSSL_TEST_VIS int TLSX_Parse(WOLFSSL* ssl, const byte* input, word16 length,
...
+            !ssl->options.echProcessingInner &&
+            ssl->ctx->echConfigs != NULL && !ssl->options.disableECH) {
+        TLSX* echX = TLSX_Find(ssl->extensions, TLSX_ECH);
+        WOLFSSL_ECH* ech = NULL;
+        if (echX != NULL)
+            ech = (WOLFSSL_ECH*)echX->data;
+
+        if (ech != NULL) {
+            if (ech->state == ECH_WRITE_NONE && ech->innerClientHello != NULL) {
```
dgarske, line 18834: **🟠 [Medium] Server now hard-rejects an accepted ECH handshake when the outer ClientHello SNI did not match a publicName** · _Behavior Change / Interop_

The reconciliation block introduces a new, stricter requirement: when the server successfully decrypts the inner ClientHello on the first flight (`state == ECH_WRITE_NONE && innerClientHello != NULL` and `serverState < SERVER_HELLO_RETRY_REQUEST_COMPLETE`), it returns `UNKNOWN_SNI_HOST_NAME_E` unless `ech->extensions` carries a `TLSX_SERVER_NAME` (i.e. the outer SNI matched a configured `echConfig->publicName`). The old `EchStateSNI` machinery did not gate ECH acceptance on the outer SNI matching the public name. A peer that omits the outer SNI entirely, or sends an outer SNI that does not match any `publicName`, will now be rejected even though the inner decrypted successfully and ECH could be accepted. wolfSSL's own client always sends the public name (set in `TLSX_ECH_Use`), so wolfSSL<->wolfSSL and the tests pass, but this is stricter than the ECH spec requires (the public_name is a SHOULD on the client) and may affect interop with third-party clients.

**Fix:** Confirm this hard-fail-on-missing-public-name is the intended policy. If interop with clients that send a different/absent outer SNI must be preserved, consider downgrading to a permissive path (accept ECH and route on the inner SNI) rather than returning UNKNOWN_SNI_HOST_NAME_E.

src/tls.c
```diff
@@ -18666,6 +18709,60 @@ WOLFSSL_TEST_VIS int TLSX_Parse(WOLFSSL* ssl, const byte* input, word16 length,
...
+                /* The outer ClientHello must have carried the echConfig
+                 * publicName as its SNI */
+                if (ssl->options.serverState <
```
dgarske: **🟠 [Medium] New UNKNOWN_SNI_HOST_NAME_E error path is not exercised by tests** · _Missing Tests_

The PR adds a new server-side rejection path that returns `UNKNOWN_SNI_HOST_NAME_E` when the outer ClientHello does not carry a public-name SNI on the first flight, yet none of the added/updated tests drive a ClientHello whose outer SNI is missing or does not match a `publicName` while the inner still decrypts. All new ECH tests (`test_wolfSSL_Tls13_ECH_wire_sni`, the `bad_configs_ex` scenarios) rely on the wolfSSL client which always emits the public-name SNI, so this branch is never taken. Error paths introduced by a PR should be covered.

**Fix:** Add a test that produces an outer ClientHello with no public-name SNI (or a non-matching outer SNI) while the inner decrypts, and assert the handshake fails with UNKNOWN_SNI_HOST_NAME_E. This both documents and locks in the intended behavior of finding #1.

src/tls.c
```diff
@@ -16864,39 +16896,43 @@ static int TLSX_ECH_BuildOuterExtensions(WOLFSSL* ssl, const byte* semaphore,
...
-    if (r == 0 && ret == 0 && ssl->ctx && ssl->ctx->extensions)
+    if (ret == 0 && ssl->ctx && ssl->ctx->extensions)
         ret = TLSX_GetSize(ssl->ctx->extensions, semaphore, msgType, pLength);
-    if (r == 0)
-        r = TLSX_EchRestoreSNI(ssl, serverName, serverNameX, extensions);
 
-    WC_FREE_VAR_EX(serverName, ssl->heap, DYNAMIC_TYPE_TMP_BUFFER);
-    if (ret == 0 && r != 0)
-        ret = r;
+    if (installed) {
```
dgarske: **🔵 [Low] Duplicated install/restore swap boilerplate in TLSX_GetSizeWithEch and TLSX_WriteWithEch** · _Code Duplication / Maintainability_

Both functions now repeat the same pattern: find ECH, `if (TLSX_EchShouldHideInner(ech)) { prepended = TLSX_EchSwapExtensions(...); installed = 1; }` before the work, then `if (installed) { prepended = TLSX_EchSwapExtensions(..., prepended); if (ret == 0 && prepended != 0) ret = BAD_STATE_E; }` after. This install/restore-with-BAD_STATE_E-check boilerplate is identical in both call sites and is easy to get subtly out of sync when one is edited.

**Fix:** Consider small helpers (e.g. TLSX_EchInstallPublic/TLSX_EchRestorePublic) to share the install/restore-and-verify logic between the size and write paths. Optional cleanup; behavior is correct as written.

src/tls.c
sebastian-carpenter: This is stricter but probably for the better. A client that sends a different public name would stick-out (bad! I don't want to encourage that). The option of omitting the SNI is an interesting idea but is another potential method of fingerprinting..

If someone needs it to be more lenient in the future then an API could be added at that time.

src/tls.c
sebastian-carpenter: In retrospect this guard never made sense to add anyways. Removed the two instances I previously added.

.github/scripts/openssl-ech.sh
sebastian-carpenter: deduplicated

src/tls.c
```diff
@@ -18702,6 +18752,60 @@ WOLFSSL_TEST_VIS int TLSX_Parse(WOLFSSL* ssl, const byte* input, word16 length,
...
+     * pass sees the authoritative list */
+    if (ret == 0 && msgType == client_hello && isRequest &&
+            !ssl->options.echProcessingInner &&
+            ssl->ctx->echConfigs != NULL && !ssl->options.disableECH) {
+        TLSX* echX = TLSX_Find(ssl->extensions, TLSX_ECH);
+        WOLFSSL_ECH* ech = NULL;
+        if (echX != NULL)
+            ech = (WOLFSSL_ECH*)echX->data;
+
+        if (ech != NULL) {
```
dgarske, line 18833: **🟠 [Medium] ECH outer ClientHello reconciliation may falsely reject valid handshakes (new public-name SNI gate)** · _Logic_

Both the review and review-security modes flagged the new outer-CH reconciliation gate at the end of TLSX_Parse, which returns INVALID_PARAMETER when ECH is accepted (ech->state == ECH_WRITE_NONE && ech->innerClientHello != NULL, serverState < SERVER_HELLO_RETRY_REQUEST_COMPLETE) but no TLSX_SERVER_NAME was recorded on ech->extensions. The review mode identifies a concrete bug (Medium/SUGGEST): when the server itself configures wolfSSL_UseSNI() with a host name equal to the ECH public name, TLSX_SNI_Parse sets matched=1 first and skips the publicName block (workingConfig stays NULL), so the SNI is routed to ssl->extensions instead of ech->extensions; the reconciliation then finds no SNI on ech->extensions and aborts an otherwise valid ECH-accepted handshake. This combination is plausible because the public name is typically a real fronting host with a matching cert. The review-security mode rates the same gate Info (an intentional, tested tightening per RFC 9849 via test_wolfSSL_Tls13_ECH_outer_sni_mismatch) and notes there is no security/DoS impact, since only the offending client's own handshake fails and it is not attacker-amplifiable; it flags that spec-compliant peers omitting the outer SNI or using an empty public_name will now be rejected. The two severity views differ (Medium bug vs Info awareness); the stricter Medium is retained.

**Fix:** Relax the reconciliation check to also accept a public-name SNI recorded on ssl->extensions, or make TLSX_SNI_Parse always route a publicName-matching SNI to ech->extensions even when a server-configured SNI already matched (run the publicName check even when matched is true). Confirm the stricter outer-SNI requirement is the intended policy for all supported ECH peers (including those that omit the outer SNI or use an empty public_name); if broader interop leniency is later desired, skip the hard abort when no server-side SNI policy is configured. Add a regression test for a server that configures wolfSSL_UseSNI() with a host name equal to the ECH public name and accepts ECH.

dgarske (commented):
## Skoll Multi-Scan Review

### Posted findings
- [Medium] [review+review-security] **ECH outer ClientHello reconciliation may falsely reject valid handshakes (new public-name SNI gate)** — `src/tls.c:18766-18784`
- [Low] [review] **Redundant TLSX_Find(TLSX_ECH) lookups in TLSX_SNI_Parse** — `src/tls.c:2456-2458, 2513-2519`
- [Info] [review] **ForceZero used to zero-initialize a non-secret Hpke struct** — `src/tls.c:13890`

src/tls.c
```diff
@@ -2454,8 +2449,20 @@ static int TLSX_SNI_Parse(WOLFSSL* ssl, const byte* input, word16 length,
             WOLFSSL_MSG("Forcing SSL object to store SNI parameter");
         }
         else {
-            /* Skipping, SNI not enabled at server side. */
-            return 0;
+#if defined(WOLFSSL_TLS13) && defined(HAVE_ECH)
+            /* No server SNI configured: when ECH is active the outer SNI still
+             * needs to be parsed so it can be matched against the echConfig
+             * publicName and recorded on ech->extensions. */
+            if (!ssl->options.disableECH && !ssl->options.echProcessingInner &&
```
dgarske: **🔵 [Low] Redundant TLSX_Find(TLSX_ECH) lookups in TLSX_SNI_Parse** · _Style_

The ECH extension is looked up twice on the same ssl->extensions list within one call: once to set checkPublic (TLSX_Find(ssl->extensions, TLSX_ECH) != NULL) and again later to populate echX/ech for the publicName comparison. The earlier lookup discards the result. Caching the echX/ech pointer once would avoid the duplicate list walk and make the two ECH-dependent branches share state.

**Fix:** Resolve echX/ech once near the top of the server block and reuse it for both the checkPublic decision and the publicName comparison. Optional cleanup; not a correctness issue.

src/tls.c
```diff
@@ -13879,19 +13887,26 @@ static int TLSX_ECH_Use(WOLFSSL_EchConfig* echConfig, TLSX** extensions,
         XFREE(ech, heap, DYNAMIC_TYPE_TMP_BUFFER);
         return MEMORY_E;
     }
+    ForceZero(ech->hpke, sizeof(Hpke));
```
dgarske: **⚪ [Info] ForceZero used to zero-initialize a non-secret Hpke struct** · _Convention_

ForceZero(ech->hpke, sizeof(Hpke)) is added to zero the freshly-XMALLOC'd Hpke struct before wc_HpkeInit. ForceZero is intended for wiping secrets so the compiler cannot elide it; using it for plain initialization is unusual (XMEMSET is the conventional init). This is harmless and consistent with the surrounding ForceZero(ech, sizeof(WOLFSSL_ECH)), so it is cosmetic. The unconditional wc_HpkeFreeKey() call (NULL ephemeralKey) introduced alongside it is safe because wc_ecc_key_free/wc_curve25519_free both null-check.

**Fix:** Use XMEMSET(ech->hpke, 0, sizeof(Hpke)) for initialization, reserving ForceZero for secret wiping. Take it or leave it; no functional impact.

src/tls.c
sebastian-carpenter: UNKNOWN_SNI_HOST_NAME was actually the wrong thing to report in this case, changed it to INVALID_PARAMETER so it sends illegal_parameter (as the RFC desires: section 7.1).

~Added a test: `test_wolfSSL_Tls13_ECH_outer_sni_mismatch`~
Added a test: `test_wolfSSL_Tls13_ECH_sni_parse`

src/tls.c
sebastian-carpenter: added helpers `TLSX_EchConcealExtensions`, `TLSX_EchExposeExtensions`

src/tls.c
```diff
@@ -16702,95 +16732,126 @@ int TLSX_PopulateExtensions(WOLFSSL* ssl, byte isServer)
...
+        for (node = *sslExts; node != NULL; node = node->next)
+            len++;
+        sslTail = sslExts;
+        while (len > popCount) {
+            sslTail = &(*sslTail)->next;
+            len--;
+        }
+        chunk = *sslTail;
+        *sslTail = NULL;
+    }
```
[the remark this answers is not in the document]
sebastian-carpenter: unnecessary. popCount cannot exceed the current length

.github/scripts/openssl-ech.sh
```diff
@@ -158,17 +172,24 @@ openssl_server(){
...
 
-    rm -f "$ech_file"
-
-    grep -q "ech_success=1" "$TMP_LOG"
+    if [ "$REJECT" -ne 0 ]; then
+        grep -q "ECH offered but rejected by server" "$TMP_LOG"
+        grep -q "ech_success=0" "$TMP_LOG"
+    else
+        grep -q "ech_success=1" "$TMP_LOG"
+    fi
```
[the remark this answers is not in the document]
sebastian-carpenter: set -e is being used. Any error is fatal

dgarske:
This PR has many changes too late in release process. There are a few minor good things that @sebastian-carpenter is going to pull out into a smaller PR to see if it can go into the release.

src/tls.c
```diff
@@ -16702,95 +16732,126 @@ int TLSX_PopulateExtensions(WOLFSSL* ssl, byte isServer)
...
-static int TLSX_EchRestoreSNI(WOLFSSL* ssl, char* serverName,
-                              TLSX* serverNameX, TLSX** extensions)
+/* If ECH is accepted, delete ech->extensions
+ * If rejected, replace matching ssl->extensions with ech->extensions,
+ *   appending to the tail if necessary */
+void TLSX_EchReplaceExtensions(WOLFSSL* ssl, byte accepted)
 {
-    int ret = 0;
+    TLSX* echX;
+    WOLFSSL_ECH* ech;
```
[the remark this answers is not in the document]
sebastian-carpenter: WOLFSSL_LOCAL doesn't seem to be used in the source files, only the headers

src/tls.c
```diff
@@ -14943,9 +14958,8 @@ static void TLSX_ECH_Free(WOLFSSL_ECH* ech, void* heap)
 {
     XFREE(ech->innerClientHello, heap, DYNAMIC_TYPE_TMP_BUFFER);
     if (ech->hpke != NULL) {
-        if (ech->ephemeralKey != NULL)
-            wc_HpkeFreeKey(ech->hpke, ech->hpke->kem, ech->ephemeralKey,
-                ech->hpke->heap);
+        wc_HpkeFreeKey(ech->hpke, ech->hpke->kem, ech->ephemeralKey,
+            ech->hpke->heap);
```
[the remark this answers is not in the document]
sebastian-carpenter: null dereference shouldn't be an issue here

src/tls.c
sebastian-carpenter: Switched order of check in SNI_PARSE. Now the public name is matched first so it is installed on ECH, this prevents this error from happening.

src/tls.c
sebastian-carpenter: Yeah, not sure why I moved this. Put it back at the top.

src/tls.c
sebastian-carpenter: changed offending locations to memset (in this case though it was removed because HpkeInit does memset anyway). Messed around a bit more with free'ing the HPKE key since it could free uninitialized data otherwise.

src/tls.c
```diff
@@ -1737,15 +1737,24 @@ int TLSX_HandleUnsupportedExtension(WOLFSSL* ssl)
 #endif
 
 #if !defined(NO_WOLFSSL_SERVER) || defined(WOLFSSL_TLS13)
-void TLSX_SetResponse(WOLFSSL* ssl, TLSX_Type type);
-/** Mark an extension to be sent back to the client. */
-void TLSX_SetResponse(WOLFSSL* ssl, TLSX_Type type)
+void TLSX_SetResponseInList(TLSX* list, TLSX_Type type);
```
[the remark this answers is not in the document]
sebastian-carpenter: changed to static

src/tls.c
```diff
@@ -1737,15 +1737,24 @@ int TLSX_HandleUnsupportedExtension(WOLFSSL* ssl)
...
+/** Mark an extension to be sent back to the client.
+ *  Operates on a list instead of the ssl.
+ *      (Should only be used on ssl->extensions or ech->extensions) */
+void TLSX_SetResponseInList(TLSX* list, TLSX_Type type)
 {
-    TLSX *extension = TLSX_Find(ssl->extensions, type);
+    TLSX *extension = TLSX_Find(list, type);
```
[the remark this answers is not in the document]
sebastian-carpenter: changed to static

tests/api.c
```diff
@@ -341,6 +341,13 @@ int testDevId = INVALID_DEVID;
     #define MESSAGE_TYPE_CAST void*
 #endif
 
+#ifdef HAVE_ECH
+    #define echPublicName  "example.com"
+    #define echPrivateName "ech-private-name.com"
+    #define echOtherName   "mismatch.io"
+    #define ECH_CONFIG_LEN 256
```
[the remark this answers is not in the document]
sebastian-carpenter: mimicking the cert macro names in wolfssl/test.h

sebastian-carpenter:
Jenkins retest this please.

dgarske:
Jenkins retest this please "StreamCorruptedException"

tests/api.c
```diff
@@ -14246,14 +14540,69 @@ static int test_wolfSSL_SubTls13_ECH(void)
...
 static word32 echCbTestConfigsLen;
-static const char* echCbTestPublicName = "example.com";
-static const char* echCbTestPrivateName = "ech-private-name.com";
 static word16 echCbTestKemID = 0;
 static word16 echCbTestKdfID = 0;
 static word16 echCbTestAeadID = 0;
 
+/* return the index of the first extension
+ * extLen will be updated with the length of the extensions field */
+static int ech_seek_extensions(byte* buf, word16* extLen)
```
dgarske, line 14687: **🔵 [Low] Test helper ech_seek_extensions dropped its malformed-buffer bounds checks**

The relocated `ech_seek_extensions`/`ech_find_extension` helpers lost the defensive bounds checks that the deleted copies had (`if (cipherSuitesLen > MAX_RECORD_SIZE) return BAD_FUNC_ARG;`, the `*innerExtLen > MAX_RECORD_SIZE` guard, and the per-extension `len > MAX_RECORD_SIZE || wrap` check in the walk loop). These parse raw ClientHello bytes by advancing `idx` with lengths read straight from the buffer. In the current tests the buffers are self-generated valid ClientHellos so this is safe, but if a future test feeds a truncated/hostile buffer the helper can read out of bounds instead of returning `BAD_FUNC_ARG`.

**Fix:** Restore the MAX_RECORD_SIZE sanity checks (and the length/wrap check in ech_find_extension's loop) so the helpers fail closed on malformed input; low priority since inputs are currently trusted test-generated buffers.

src/tls.c
```diff
@@ -13892,8 +13898,16 @@ static int TLSX_ECH_Use(WOLFSSL_EchConfig* echConfig, TLSX** extensions,
     if (ret == 0)
         ret = wc_HpkeGenerateKeyPair(ech->hpke, &ech->ephemeralKey, rng);
     if (ret == 0) {
-        ret = TLSX_Push(extensions, TLSX_ECH, ech, heap);
+        /* use the chosen config's public name for the outer SNI */
+        ret = TLSX_UseSNI(&ech->extensions, WOLFSSL_SNI_HOST_NAME,
+            echConfig->publicName, (word16)XSTRLEN(echConfig->publicName),
+            heap);
+        if (ret == WOLFSSL_SUCCESS)
```
dgarske: **🔵 [Low] Dead defensive branch for WOLFSSL_FAILURE from TLSX_UseSNI**

`TLSX_UseSNI` returns only `WOLFSSL_SUCCESS`, `BAD_FUNC_ARG`, `BAD_LENGTH_E`, or `MEMORY_E` (see src/tls.c:2653-2698); it never returns `WOLFSSL_FAILURE`. The `else if (ret == WC_NO_ERR_TRACE(WOLFSSL_FAILURE)) ret = BAD_STATE_E;` branch is therefore unreachable. It is harmless (the generic `if (ret != 0)` cleanup handles the real error returns), but it is dead code that can mislead readers into thinking WOLFSSL_FAILURE is a possible result.

**Fix:** Remove the unreachable WOLFSSL_FAILURE else-if, or convert to `if (ret != WOLFSSL_SUCCESS)` and rely on the existing `ret != 0` cleanup.

src/tls.c
```diff
@@ -16709,95 +16738,126 @@ int TLSX_PopulateExtensions(WOLFSSL* ssl, byte isServer)
...
+    TLSX* node;
+    TLSX* outer;
+    TLSX* inner;
+    TLSX** outerLink;
+    TLSX** innerLink;
+    TLSX** sslTail;
+    word16 len = 0;
+    word16 appended = 0;
+
+    if (popCount > 0) {
```
dgarske: **⚪ [Info] TLSX_EchSwapExtensions: popCount larger than the ssl list length would relocate the entire list (latent, not currently…** · _Logic_

The new `TLSX_EchSwapExtensions` computes the pop point from `popCount` with `while (len > popCount) { sslTail = &(*sslTail)->next; len--; }`. If `popCount` were ever greater than or equal to the number of nodes in `*sslExts`, the loop never advances, `sslTail` stays at the head, and `chunk = *sslTail` captures the ENTIRE `ssl->extensions` list (then `*sslTail = NULL` empties it), silently moving all of `ssl->extensions` onto `ech->extensions`. Today this cannot be triggered: the only callers (`TLSX_EchExposeExtensions` via `TLSX_GetSizeWithEch`/`TLSX_WriteWithEch`) pass `popCount = appended`, where `appended` was produced by the paired conceal call and the list only grows (never shrinks) between conceal and expose, so `popCount <= len` always holds. This is a defense-in-depth/robustness note on newly introduced code, not an exploitable condition. Because the function is now exported via `WOLFSSL_TEST_VIS`/`WOLFSSL_API_PREFIX_MAP`, a future or external caller could violate the invariant.

**Fix:** Clamp defensively, e.g. only pop when `popCount < len` (or add `if (popCount >= len) popCount = len ? len : 0;` / an assert), so an out-of-range `popCount` cannot relocate the whole extension list. Optional hardening; no functional change for the current callers.

dgarske (commented):
## Skoll Multi-Scan Review

### Posted findings
- [Low] [review] **Test helper ech_seek_extensions dropped its malformed-buffer bounds checks** — `tests/api.c:14551-14606`
- [Low] [review] **Dead defensive branch for WOLFSSL_FAILURE from TLSX_UseSNI** — `src/tls.c:13905-13908`
- [Info] [review-security] **TLSX_EchSwapExtensions: popCount larger than the ssl list length would relocate the entire list (latent, not currently reachable)** — `src/tls.c:16770-16780`

### Skipped findings
- [Low] ~~workingConfig pointer reused as an implicit 'public name matched' flag~~

dgarske (dismissed):
https://github.com/wolfSSL/wolfssl/pull/10568#pullrequestreview-4649011650

src/tls.c
```diff
@@ -16733,94 +16759,146 @@ int TLSX_PopulateExtensions(WOLFSSL* ssl, byte isServer)
...
+/* Swap matching extension types between *sslExts and *echExts.
+ *   Non-matched extensions in *echExts are appended to the tail of *sslExts
+ *   popCount is the number of trailing extensions to move from
+ *     *sslExts back to *echExts (the undo of a prior append)
+ *
+ * Extensions are stored in reverse wire order, so non-matched extensions are
+ * appended to the tail rather than the head; this avoids displacing the leading
+ * extension (e.g. pre_shared_key, which must stay last on the wire).
+ *
+ * Returns a count of extensions appended to sslExts on success,
```
dgarske: **🔵 [Low] Misleading return-value doc comment on TLSX_EchSwapExtensions** · _style_

The header comment states "Returns a count of extensions appended to sslExts on success, error otherwise." but the function actually `return 0;` on success and reports the appended count via the out-parameter `*appended`. All three callers (`TLSX_EchConcealExtensions`, `TLSX_EchExposeExtensions`, `TLSX_EchReplaceExtensions`) correctly test for `ret == 0` as success, so there is no functional bug, but the comment could mislead a future maintainer into treating a positive return as a valid count rather than an error.

**Fix:** Correct the comment to describe the 0/error return contract and the `*appended` out-parameter.

src/tls.c
```diff
@@ -2454,8 +2457,19 @@ static int TLSX_SNI_Parse(WOLFSSL* ssl, const byte* input, word16 length,
             WOLFSSL_MSG("Forcing SSL object to store SNI parameter");
         }
         else {
-            /* Skipping, SNI not enabled at server side. */
-            return 0;
+        #if defined(WOLFSSL_TLS13) && defined(HAVE_ECH)
```
dgarske, line 2460: **🔵 [Low] Convoluted checkPublic flag with nested #if in TLSX_SNI_Parse** · _style_

The "no server SNI configured" branch introduces `checkPublic` and then uses a nested `#if`/double-`if` (`if (ech != NULL) checkPublic = 1; if (!checkPublic) { return 0; }`) to decide whether to keep parsing. The logic is correct but harder to read than necessary; when ECH is compiled out it collapses to an unconditional `return 0`, and when compiled in it is equivalent to `if (ech == NULL) return 0;` with `checkPublic = 1` otherwise.

**Fix:** Optionally simplify for readability; behavior is already correct and tested.

src/tls.c
```diff
@@ -16733,94 +16759,146 @@ int TLSX_PopulateExtensions(WOLFSSL* ssl, byte isServer)
...
-    /* if type is outer and this is a real ECH then change sni to public name */
-    if (echX != NULL && ssl->echConfigs != NULL &&
-        ((WOLFSSL_ECH*)echX->data)->type == ECH_TYPE_OUTER) {
-        if (ssl->extensions) {
-            serverNameX = TLSX_Find(ssl->extensions, TLSX_SERVER_NAME);
+    *appended = 0;
 
-            if (serverNameX != NULL)
-                extensions = &ssl->extensions;
+    if (popCount > 0) {
```
dgarske: **⚪ [Info] Undocumented restore invariant: TLSX_EchSwapExtensions restore fails if ssl list was originally empty** · _question_

The restore path guards with `if (popCount >= len) return BAD_FUNC_ARG;` (popCount must be strictly less than the current ssl list length). During expose, popCount == appended and len == M + appended, where M is the number of ssl extensions present before conceal. The check therefore fails only when M == 0, i.e. `ssl->extensions` was empty at conceal time and every ech (public) extension was appended, which would surface as `BAD_STATE_E` from `TLSX_EchExposeExtensions`. This is not reachable for a real ClientHello (which always carries key_share/supported_versions/etc.), and `TLSX_EchReplaceExtensions` uses popCount==0 so it is unaffected, but the implicit "ssl list must have >= 1 non-appended node" invariant is undocumented and depends on caller context.

**Fix:** Add a comment noting the invariant; no functional change required since the empty-ssl case does not occur in the handshake flow.

dgarske (commented):
## Skoll Multi-Scan Review

### Posted findings
- [Low] [review] **Misleading return-value doc comment on TLSX_EchSwapExtensions** — `src/tls.c:16778`
- [Low] [review] **Convoluted checkPublic flag with nested #if in TLSX_SNI_Parse** — `src/tls.c:2460-2473`
- [Info] [review] **Undocumented restore invariant: TLSX_EchSwapExtensions restore fails if ssl list was originally empty** — `src/tls.c:16797-16805, 16861-16876`

dgarske:
Jenkins retest this please

src/tls.c
```diff
@@ -18733,6 +18815,67 @@ WOLFSSL_TEST_VIS int TLSX_Parse(WOLFSSL* ssl, const byte* input, word16 length,
...
+            !ssl->options.echProcessingInner &&
+            ssl->ctx->echConfigs != NULL && !ssl->options.disableECH) {
+        TLSX* echX = TLSX_Find(ssl->extensions, TLSX_ECH);
+        WOLFSSL_ECH* ech = NULL;
+        if (echX != NULL)
+            ech = (WOLFSSL_ECH*)echX->data;
+
+        if (ech != NULL) {
+            if (ech->state == ECH_WRITE_NONE && ech->innerClientHello != NULL) {
+                /* TODO: This check will require some additional work. Disabling
```
dgarske: Can we not leave TODO's? Either do it or cleanup. Thanks

src/tls.c
```diff
@@ -16733,94 +16758,149 @@ int TLSX_PopulateExtensions(WOLFSSL* ssl, byte isServer)
...
+static int TLSX_EchExposeExtensions(WOLFSSL* ssl, WOLFSSL_ECH* ech,
+    word16 appended, int installed)
+{
+    int ret = 0;
+
+    if (installed) {
+        ret = TLSX_EchSwapExtensions(&ssl->extensions, &ech->extensions,
+            &appended);
+        if (ret == 0 && appended != 0) {
+            WOLFSSL_MSG("Bad restore with TLSX_EchSwapExtensions");
```
dgarske, line 16876: In TLSX_GetSizeWithEch/TLSX_WriteWithEch, if TLSX_EchExposeExtensions fails (either TLSX_EchSwapExtensions returns BAD_FUNC_ARG, or the post-restore appended != 0 yields BAD_STATE_E), the error is propagated but ssl->extensions and ech->extensions are left half-swapped -- the public SNI stays installed on ssl->extensions and the private extensions remain parked on ech->extensions. The round-trip is proven balanced by test_TLSX_EchSwapExtensions, so this is a defensive path that should not trigger in practice, but a corrupted extension list is a worse failure mode than a clean abort, and there is no assertion/cleanup guarding it.

  Code:
```
    if (installed) {
        ret = TLSX_EchSwapExtensions(&ssl->extensions, &ech->extensions, &appended);
        if (ret == 0 && appended != 0) {
            WOLFSSL_MSG("Bad restore with TLSX_EchSwapExtensions");
            ret = BAD_STATE_E;
        }
    }
```
  Recommendation: Add a brief comment that this restore is expected to be infallible given the conceal/expose invariant, and that a non-zero return indicates unrecoverable list corruption so the connection must be torn down. Confirm callers abort the handshake (do not reuse the ssl) if expose returns non-zero.

dgarske (changes requested):

src/tls.c
sebastian-carpenter: added comment

src/tls.c
sebastian-carpenter: removed the TODO's and their associated '#if 0'

sebastian-carpenter:
Jenkins retest this please.

src/tls.c
```diff
@@ -2516,42 +2510,55 @@ static int TLSX_SNI_Parse(WOLFSSL* ssl, const byte* input, word16 length,
...
-        matched = cacheOnly || (hostName != NULL &&
-            XSTRLEN(hostName) == size &&
-            XSTRNCMP(hostName, (const char*)input + offset, size) == 0);
-    }
-
 #if defined(WOLFSSL_TLS13) && defined(HAVE_ECH)
-    if (!matched && ech != NULL && ech->sniState == ECH_OUTER_SNI) {
+    /* While parsing the outer CH accept a match against any
+     * echConfig publicName */
+    if (ech != NULL) {
```
dgarske, line 2516: **🔵 [Low] workingConfig overloaded as both loop cursor and public-name-matched flag** · _style_

In the reworked SNI parse, the local `workingConfig` serves double duty: it iterates the echConfig list, and later `if (workingConfig != NULL) writeList = &ech->extensions;` reuses it as a boolean meaning "the outer name matched a publicName". This only works because the `while (workingConfig != NULL)` loop `break`s (leaving it non-NULL) exactly when a public name matched, and runs to NULL otherwise. The logic is correct, but a reader must reconstruct that invariant to trust the `writeList` selection. A dedicated `int matchedPublic` flag set alongside `matched = 1` would make the intent explicit and decouple it from the loop's terminal state.

**Fix:** Introduce a dedicated flag, e.g. set `matchedPublic = 1;` next to `matched = 1;` inside the loop, and use `if (matchedPublic) writeList = &ech->extensions;` so the write-list choice does not depend on the loop cursor's residual value. Optional readability cleanup; behavior is correct as written.

.github/scripts/openssl-ech.sh
```diff
@@ -158,17 +172,24 @@ openssl_server(){
...
+    rm -f "$ech_file"
+
     # test with wolfssl client
+    # in reject mode the client is expected to error out, so tolerate a
+    #   nonzero exit
     $WOLFSSL_CLIENT -v 4 \
         -p "$port" \
         -S "$PRIV_NAME" \
         --ech "$ech_config" \
         $wolfssl_extra \
```
dgarske, line 184: **⚪ [Info] Reject-mode CI check greps for ech_success=0, a token not emitted anywhere in the tree** · _test_

The new reject path asserts success with `grep -q "ECH offered but rejected by server" && grep -q "ech_success=0"`. The error string exists (src/internal.c: "ECH offered but rejected by server"), but the `ech_success=\<N>` token is not printed by any C source in the repo (the pre-existing accept check already relied on `ech_success=1`, so it must be produced by the CLI build used in CI rather than the tree). The reject check is therefore consistent with the existing accept check and will fail loudly rather than pass silently if the string ever changes, so this is only a robustness note: the pass condition is coupled to exact CLI log text that is not visible/greppable in-repo.

**Fix:** Consider keying the reject assertion on a wolfSSL_GetEchStatus-derived status string that is defined in-tree, or add a brief comment noting where `ech_success=` originates, to keep the CI contract discoverable. Optional; no functional issue.

dgarske (commented):
## Skoll Multi-Scan Review**Overall recommendation:** `COMMENT`

### Posted findings
- [Low] [review] **workingConfig overloaded as both loop cursor and public-name-matched flag** — `src/tls.c:2516-2558`
- [Info] [review] **Reject-mode CI check greps for ech_success=0, a token not emitted anywhere in the tree** — `.github/scripts/openssl-ech.sh:184-192`

### Skipped findings
- [Info] ~~WOLFSSL_ECH initialization downgraded from ForceZero to XMEMSET~~
- [Info] ~~ech-extensions list order is not a perfect round-trip identity after conceal/expose (latent, multi-extension only)~~
- [Low] ~~New TLSX_EchExposeExtensions failure leaves ssl/ech extension lists in swapped state with no recovery~~
- [Low] ~~Outer-ClientHello public-name SNI verification intentionally disabled (#if 0)~~
- [Low] ~~ECH-accept path returns from TLSX_Parse before SNI/TCA verify of the outer ClientHello~~

dgarske (changes requested):
https://github.com/wolfSSL/wolfssl/pull/10568#pullrequestreview-4404720189

src/tls.c
sebastian-carpenter: An extra variable seems like unnecessary decoupling. The workingConfig doing double duty is already commented so it should be clear enough.

One variable also helps reduce what needs to be checked when making changes.

.github/scripts/openssl-ech.sh
sebastian-carpenter: that's a good idea but I'll hold off for now. I'm already editing the client example in a local branch so I can update it there.

dgarske (approved):

PR #10572, opened 2026-06-02
Support RFC 9802 LMS and XMSS in X.509 certificate and CSR generation

Frauschi:
Extends wolfCrypt's certificate generation to issue HSS/LMS and XMSS/XMSS^MT certificates and PKCS#10 requests, building on the existing RFC 9802 *verification* support so the library can now both produce and consume stateful hash-based-signature certificates.

### Changes
- **New key-type selectors** `LMS_TYPE` / `XMSS_TYPE` / `XMSSMT_TYPE` for `wc_MakeCert_ex`, `wc_SignCert_ex`, and `wc_MakeCertReq_ex`.
- **New SPKI encoders** `wc_LmsKey_PublicKeyToDer` and `wc_XmssKey_PublicKeyToDer` to emit RFC 9802 SubjectPublicKeyInfo.
- Runtime signature-buffer sizing and `sigType`/key consistency checks for the new algorithms.
- Generation is **ASN.1-template only** (`WOLFSSL_ASN_TEMPLATE`), matching where the verification path already lives. Non-template builds reject these key types with `ALGO_ID_E`.

### Tests
- New in-process tests generate self-signed LMS/XMSS roots, CSRs, and a CA→ECC-leaf chain, then verify each through the parsing/verification path.
- These replace the previously checked-in (patched Bouncy Castle) DER fixtures; only the stock RFC 9802-aligned LMS interop anchor is retained.

No change to default builds — the feature is gated behind the existing LMS/XMSS and certificate-generation options.


Frauschi:
Jenkins retest this please

wolfcrypt/src/asn.c
```diff
@@ -29492,24 +29946,51 @@ static int SignCert(int requestSz, int sType, byte* buf, word32 buffSz,
...
     #endif
-        heap = eccKey->heap;
     #else
         return NOT_COMPILED_IN;
     #endif /* HAVE_ECC */
     }
+    heap = GetSigningKeyHeap(rsaKey, eccKey, ed25519Key, ed448Key, mldsaKey,
+        slhDsaKey, lmsKey, xmssKey);
+
+    /* The signatureAlgorithm OID is written from sType while the signature is
```
dgarske, line 29964: **🟠 [Medium] CheckSigTypeForKey now rejects the wc_InitCert default sigType for non-RSA keys**

`SignCert`, `wc_MakeSigWithBitStr`, `wc_SignCRL_ex`, and `wc_SignCert_cb` now call `CheckSigTypeForKey`, which returns `ALGO_ID_E` when `sType` does not match the signing key's algorithm family. `wc_InitCert`/`wc_InitCert_ex` default `cert->sigType` to `CTC_SHA256wRSA` (confirmed at asn.c:26216 / wc_InitCert_ex). Any existing caller that signs with a non-RSA key via `wc_SignCert`/`wc_SignCert_ex` but forgets to set `cert.sigType` previously succeeded (emitting a cert whose signatureAlgorithm OID contradicted the key) and will now fail with `ALGO_ID_E`. This is a beneficial hardening and all new tests set `sigType` explicitly, but it is a behavior change that could surface as a 'regression' for buggy third-party callers.

**Fix:** Confirm all in-tree cert/CSR/CRL generation tests and examples set `cert.sigType` for non-RSA keys (wolfCrypt internal tests do), and call out the new strict sigType/key validation in the release notes so downstream callers that relied on the previously-permissive behavior update accordingly.

wolfcrypt/src/asn.c
```diff
@@ -29492,24 +29946,51 @@ static int SignCert(int requestSz, int sType, byte* buf, word32 buffSz,
...
+        return ret;
+    }
+
+    /* Size the signature buffer from the key in use. */
+    maxSigSz = GetSignatureBufferSz(rsaKey, eccKey, ed25519Key, ed448Key,
+        falconKey, mldsaKey, slhDsaKey, lmsKey, xmssKey);
+    if (maxSigSz <= 0)
+        return (maxSigSz < 0) ? maxSigSz : ALGO_ID_E;
 
 #ifndef WOLFSSL_NO_MALLOC
```
dgarske, line 29980: **🔵 [Low] NO_MALLOC sign paths pass MAX_ENCODED_CLASSIC_SIG_SZ capacity against a WOLFSSL_MAX_SIG_SZ buffer**

Under `WOLFSSL_NO_MALLOC`, `certSignCtx->sig` is the fixed array `byte sig[WOLFSSL_MAX_SIG_SZ]` (types.h:2491). `wc_SignCert_cb` and `wc_SignCRL_ex` pass `MAX_ENCODED_CLASSIC_SIG_SZ` as the buffer capacity to `MakeSignatureCb`/`MakeSignature`. With the default `WOLFSSL_MAX_SIG_SZ == MAX_ENCODED_SIG_SZ >= MAX_ENCODED_CLASSIC_SIG_SZ` this is safe. However, the new `WOLFSSL_MAX_SIG_SZ` override is documented (types.h:2470) as a way to 'shrink builds that only sign with classic/compact algorithms'; if a user sets it below `MAX_ENCODED_CLASSIC_SIG_SZ`, the claimed capacity would exceed the real fixed buffer and a full-modulus RSA signature could overflow it. The coupling is implicit.

**Fix:** Either clamp the capacity passed to MakeSignature*/MakeSignatureCb to `min(MAX_ENCODED_CLASSIC_SIG_SZ, WOLFSSL_MAX_SIG_SZ)` under WOLFSSL_NO_MALLOC, or document that `WOLFSSL_MAX_SIG_SZ` must not be reduced below `MAX_ENCODED_CLASSIC_SIG_SZ` (the max RSA modulus).

tests/api/test_lms_xmss.c
```diff
@@ -644,323 +650,607 @@ static int rfc9802_xmss_import_negative(void)
...
+    /* Negative: corrupt the leaf's signature (last byte of the DER, in the
+     * signatureValue) and confirm verification fails even with the CA loaded.
+     * This proves the CA's hash-based signature is cryptographically checked,
+     * not accepted on issuer-name chaining alone. */
+    if (EXPECT_SUCCESS() && leafSz > 0) {
+        byte saved = leafDer[leafSz - 1];
+        leafDer[leafSz - 1] ^= 0xFF;
+        ExpectNotNull(cm = wolfSSL_CertManagerNew());
+        ExpectIntEQ(wolfSSL_CertManagerLoadCABuffer(cm, caDer, (long)caSz,
+            WOLFSSL_FILETYPE_ASN1), WOLFSSL_SUCCESS);
```
dgarske, line 921: **🔵 [Low] XMSS test key-persistence callbacks use raw stdio instead of XFILE abstraction**

The new XMSS generation tests define key persistence callbacks using raw `fopen`/`fwrite`/`fread`/`fclose` and `FILE*`, whereas the same file's `rfc9802_load_file` helper uses the `XFILE`/`XFOPEN`/`XFREAD`/`XFCLOSE` abstraction. Mixing raw stdio is inconsistent with the file's own conventions and will not honor platforms that remap file I/O through the X-macros.

**Fix:** Use XFILE/XFOPEN/XFWRITE/XFREAD/XFCLOSE for consistency with rfc9802_load_file and portability across NO_FILESYSTEM-style platform remaps.

tests/api/test_lms_xmss.c
```diff
@@ -644,323 +650,607 @@ static int rfc9802_xmss_import_negative(void)
...
+    if (cm != NULL)
         wolfSSL_CertManagerFree(cm);
-        cm = NULL;
+    return EXPECT_RESULT();
+}
+
+#ifdef WOLFSSL_CERT_REQ
+/* Parse a generated CSR and confirm its proof-of-possession signature. */
+static int rfc9802_gen_verify_csr(const byte* der, int derSz)
+{
```
dgarske, line 766: **🔵 [Low] Uninitialized derSz in rfc9802_gen_roundtrip**

`int derSz;` is declared uninitialized. It is only read inside `ExpectIntEQ(rfc9802_gen_verify_selfsigned(der, derSz), ...)` blocks that the Expect framework short-circuits once a prior expectation fails, so in practice it is never read while uninitialized. Still, `caSz`/`leafSz` in the sibling `rfc9802_gen_chain` are initialized to 0; initializing `derSz` too avoids a possible -Wmaybe-uninitialized under -Werror and matches the surrounding convention.

**Fix:** Initialize `derSz` to 0 for consistency and to silence potential uninitialized-use warnings.

dgarske (commented):
## Skoll Code Review

### Posted findings
- [Medium] **CheckSigTypeForKey now rejects the wc_InitCert default sigType for non-RSA keys** — `wolfcrypt/src/asn.c:29956-29964`
- [Low] **NO_MALLOC sign paths pass MAX_ENCODED_CLASSIC_SIG_SZ capacity against a WOLFSSL_MAX_SIG_SZ buffer** — `wolfcrypt/src/asn.c:30360-30370, 36899-36912`
- [Low] **XMSS test key-persistence callbacks use raw stdio instead of XFILE abstraction** — `tests/api/test_lms_xmss.c:921-944`
- [Low] **Uninitialized derSz in rfc9802_gen_roundtrip** — `tests/api/test_lms_xmss.c:766-790`
- [Info] **OID-occurrence count assertion assumes the HBS OID never appears inside the signature value** — `tests/api/test_lms_xmss.c:660-676, 1027-1033`

tests/api/test_lms_xmss.c
```diff
@@ -644,323 +650,607 @@ static int rfc9802_xmss_import_negative(void)
...
     }
-#endif
-#if (!defined(WOLFSSL_XMSS_MAX_HEIGHT) || (WOLFSSL_XMSS_MAX_HEIGHT >= 20)) && \
-    (!defined(WOLFSSL_XMSS_MIN_HEIGHT) || (WOLFSSL_XMSS_MIN_HEIGHT <= 40))
-    for (i = 0; i < sizeof(xmssmtFiles) / sizeof(xmssmtFiles[0]); i++) {
-        ExpectIntEQ(rfc9802_verify_one_cert(xmssmtFiles[i],
-            XMSSMTk, CTC_XMSSMT), TEST_SUCCESS);
+
+    if (EXPECT_SUCCESS() && derSz > 0) {
+        n = rfc9802_collect_hbs_oid_offsets(der, (word32)derSz, oidLast, off, 8);
```
dgarske, line 1126: **⚪ [Info] OID-occurrence count assertion assumes the HBS OID never appears inside the signature value**

`rfc9802_gen_xmss_oid_tamper` asserts exactly 3 occurrences of the 8-byte `1.3.6.1.5.5.7.6.\<lastByte>` OID body in the generated DER (TBS-sig, SPKI-key, outer-sig). Because the XMSS signatureValue is large and freshly random per run, the same 8-byte sequence could in principle appear inside the signature, making `ExpectIntEQ(n, 3)` fail. The probability is ~2^-64 per cert (effectively never), so this is informational only.

**Fix:** Optionally bound the scan to the TBS/algId region (before the signatureValue BIT STRING) to make the count structurally exact rather than statistically exact. Not required given the negligible probability.

wolfcrypt/src/asn.c
Frauschi: Internally, all call-sites properly set `cert.sigType`. For external users, this has been a latent bug anyway as the resulting certificates couldn't be used with any proper verification library. Hence, I think this does not need to be called out explicitly in the release notes.

wolfcrypt/src/asn.c
Frauschi: Fixed by clamping properly

tests/api/test_lms_xmss.c
Frauschi: fixed

tests/api/test_lms_xmss.c
Frauschi: fixed

Frauschi:
Jenkins retest this please

Frauschi:
The Arduino failure is unrelated and fixed in #10644.

wolfssl/wolfcrypt/types.h
```diff
@@ -2375,29 +2375,39 @@ enum Max_ASN {
...
-#ifdef WOLFSSL_HAVE_SLHDSA
-    /* Largest raw SLH-DSA signature (SHAKE-256f) is 49856 bytes; round up
-     * to leave headroom for ASN.1 wrapping (BIT STRING tag + length). */
-    MAX_ENCODED_SIG_SZ  = 51200,
-#elif defined(HAVE_FALCON) || defined(WOLFSSL_HAVE_MLDSA)
-    MAX_ENCODED_SIG_SZ  = 5120,
-#elif !defined(NO_RSA)
+    /* Max classic sig (RSA/DSA/ECC); separate from MAX_ENCODED_SIG_SZ so
+     * PKCS#1/verify buffers stay small when PQC is enabled for verify-only. */
+#if !defined(NO_RSA)
```
dgarske, line 2380: ### MEDIUM-1: MAX_ENCODED_CLASSIC_SIG_SZ ladder omits Ed25519/Ed448; SendCertificateVerify classic signature buffer undersized for Ed448
- **File:** `wolfssl/wolfcrypt/types.h:2380-2397; src/internal.c:35154-35159`
- **Function:** `SendCertificateVerify / enum Max_ASN`
- **Category:** Regression
- **Action:** SUGGEST
- **Tag:** bug
- **Confidence:** High

**Description:** The diff splits the old MAX_ENCODED_SIG_SZ into a classic tier (MAX_ENCODED_CLASSIC_SIG_SZ) and a PQC-inclusive MAX_ENCODED_SIG_SZ, and changes SendCertificateVerify to allocate ssl->buffers.sig.buffer with MAX_ENCODED_CLASSIC_SIG_SZ. For ECC/EdDSA the actual signature is written directly into this buffer (EccSign/Ed25519Sign/Ed448Sign at internal.c:35296-35340), and the inline comment claims it 'comfortably holds an ECC/EdDSA signature'. But the ladder has branches only for RSA, ECC (140), and CURVE448 (114), falling through to 64 otherwise -- there is NO branch for HAVE_ED448 (114-byte signature) or HAVE_ED25519 (64). In a build with HAVE_ED448 but without RSA, ECC, and CURVE448, the value is 64, smaller than the 114-byte Ed448 signature. The change WORSENS PQC builds: previously MAX_ENCODED_SIG_SZ was 5120 (Falcon/ML-DSA) or 51200 (SLH-DSA) and safely held the 114-byte Ed448 signature; now it drops back to 64. Both reviewers agree the PQC+Ed448 combination is newly broken by this change. SEVERITY VIEWS DIFFER: the `review` mode rates this Medium and characterizes it as a potential heap write past the allocation (out-of-bounds write) for an Ed448-only config, since the Ed448 signature is copied into the under-sized buffer. The `review-security` mode rates it Low and argues it is NOT a memory-safety overflow because wc_ed448_sign_msg checks *outLen < ED448_SIG_SIZE (ed448.c:461) and returns BUFFER_E first, making it a functional/availability regression that breaks Ed448 certificate-verify signing rather than corrupting memory. The stricter (Medium) severity is retained pending confirmation of the write path. Either way the comment is inaccurate for Ed448 and the configuration regresses.

**Code:**
```
#elif defined(HAVE_ECC)
    MAX_ENCODED_CLASSIC_SIG_SZ = 140,
#elif defined(HAVE_CURVE448)
    MAX_ENCODED_CLASSIC_SIG_SZ = 114,
#else
    MAX_ENCODED_CLASSIC_SIG_SZ =  64,   /* too small for Ed448 (114) */
#endif
...
/* src/internal.c SendCertificateVerify */
ssl->buffers.sig.length = MAX_ENCODED_CLASSIC_SIG_SZ;
ssl->buffers.sig.buffer = (byte*)XMALLOC(MAX_ENCODED_CLASSIC_SIG_SZ, ssl->heap, DYNAMIC_TYPE_SIGNATURE);
...
ret = Ed448Sign(ssl, ssl->hsHashes->messages, ssl->hsHashes->length,
    ssl->buffers.sig.buffer, (word32*)&ssl->buffers.sig.length, key, ...);
```

**Recommendation:** Add explicit HAVE_ED448 (114) and HAVE_ED25519 (64) tiers to the MAX_ENCODED_CLASSIC_SIG_SZ ladder so it is never below the largest enabled classic signature -- e.g. take the max of the RSA/ECC/CURVE448 base and a per-EdDSA floor (>=114 when HAVE_ED448, >=64 when HAVE_ED25519). This restores pre-diff safety for PQC+Ed448 builds, closes the latent Ed448-only undersize, and makes the buffer match the comment's stated intent. Also correct the SendCertificateVerify comment.

wolfcrypt/src/asn.c
```diff
@@ -29820,22 +30346,40 @@ int wc_SignCert_cb(int requestSz, int sType, byte* buf, word32 buffSz,
     return NOT_COMPILED_IN;
 #endif
 
+    /* The callback produces the signature for keyType while the cert's
```
dgarske, line 30349: ### MEDIUM-2: New signature-type vs key checks on wc_SignCert_cb and wc_SignCRL_ex lack negative-path tests
- **File:** `wolfcrypt/src/asn.c:30349-30359; wolfcrypt/src/asn.c:36914-36921`
- **Function:** `wc_SignCert_cb / wc_SignCRL_ex`
- **Category:** Test Coverage
- **Action:** SUGGEST
- **Tag:** test
- **Confidence:** Medium

**Description:** The diff adds new family/OID-consistency validation: wc_SignCert_cb now returns ALGO_ID_E when keyType==RSA_TYPE && !IsRsaSigType(sType) (and the ECC analogue), and wc_SignCRL_ex now calls CheckSigTypeForKey(...). The new tests (test_rfc9802_lms_x509_gen / test_rfc9802_xmss_x509_gen) exercise the LMS/XMSS branches of CheckSigTypeForKey via wc_SignCert_ex, but the RSA/ECC mismatch branches reached through wc_SignCert_cb and wc_SignCRL_ex are not directly covered. A regression that weakened or removed these specific checks would not be caught.

**Code:**
```
#ifndef NO_RSA
    if (keyType == RSA_TYPE && !IsRsaSigType(sType))
        return ALGO_ID_E;
#endif
#ifdef HAVE_ECC
    if (keyType == ECC_TYPE && !IsEccSigType(sType))
        return ALGO_ID_E;
#endif
```

**Recommendation:** Add a small negative test that calls wc_SignCert_cb with an RSA keyType but an ECDSA sType (and vice versa), and wc_SignCRL_ex with a mismatched sType, asserting ALGO_ID_E. This locks in the new consistency guarantee on those paths.

src/internal.c
```diff
@@ -14249,8 +14249,10 @@ int CopyDecodedToX509(WOLFSSL_X509* x509, DecodedCert* dCert)
 #endif
     }
 
-    if (dCert->signature != NULL && dCert->sigLength != 0 &&
-            dCert->sigLength <= MAX_ENCODED_SIG_SZ) {
+    /* Store a copy of the signature for later retrieval. The buffer is sized
```
dgarske, line 14252: ### INFO-3: Removal of MAX_ENCODED_SIG_SZ ceiling on X.509 signature copy (reviewed; bounded and safe)
- **File:** `src/internal.c:14252-14263, 14599-14611`
- **Function:** `CopyDecodedToX509 / CopyDecodedAcertToX509`
- **Category:** Security
- **Confidence:** High

**Description:** The dCert->sigLength <= MAX_ENCODED_SIG_SZ (and acert equivalent) guard was removed so large LMS/XMSS signatures are retained. BEFORE, an oversized signature was silently dropped (buffer left NULL); AFTER, it is always copied. This is safe: the allocation XMALLOC(dCert->sigLength, ...) is bounded by the parsed cert DER already resident in memory (no amplification), the XMEMCPY length equals the allocation size (no overflow), and the only in-tree consumer, wolfSSL_X509_get_signature (src/x509.c:4399-4405), validates *bufSz < x509->sig.length before copying. No exploitable path was found. Recorded for reviewer awareness because it relaxes a long-standing input bound on attacker-supplied certificate signatures; downstream OpenSSL-compat consumers that assume a bounded x509->sig.length should be confirmed to length-check.

**Code:**
```
-    if (dCert->signature != NULL && dCert->sigLength != 0 &&
-            dCert->sigLength <= MAX_ENCODED_SIG_SZ) {
+    if (dCert->signature != NULL && dCert->sigLength != 0) {
         x509->sig.buffer = (byte*)XMALLOC(
                           dCert->sigLength, x509->heap, DYNAMIC_TYPE_SIGNATURE);
```

**Recommendation:** No change required for memory safety. Optionally retain a generous sanity ceiling (e.g. derived from the maximum supported LMS/XMSS parameter set) to bound worst-case allocations from malformed certs, and audit any external consumers of x509->sig.length for fixed-size copies.

dgarske (changes requested):

wolfcrypt/src/asn.c
Frauschi: Added a test

wolfssl/wolfcrypt/types.h
Frauschi: Good catch (although the analysis and recommendation of Skoll are actually quite bad). The problem has been an incorrect `HAVE_CURVE448` instead of `HAVE_ED448`. Fixed now.

dgarske (approved):

douzzer (approved):
passed `pr-check`, except for expected-to-fail `all-no-tls`.

PR #10580, opened 2026-06-02
aes.c: extend GCM H-skip guard to cover WOLF_CRYPTO_CB_SETKEY

MarkAtwood:
# aes.c: extend GCM H-skip guard to cover WOLF_CRYPTO_CB_SETKEY

## What

One-line fix at `wolfcrypt/src/aes.c:7855`: extend the existing
"SE owns key, skip H and M-table generation" guard from
`#ifdef WOLF_CRYPTO_CB_AES_SETKEY` to
`#if defined(WOLF_CRYPTO_CB_AES_SETKEY) || defined(WOLF_CRYPTO_CB_SETKEY)`.

```diff
-#ifdef WOLF_CRYPTO_CB_AES_SETKEY
+#if defined(WOLF_CRYPTO_CB_AES_SETKEY) || defined(WOLF_CRYPTO_CB_SETKEY)
     if ((ret == 0) && (aes->devId != INVALID_DEVID && aes->devCtx != NULL)) {
         /* SE owns key - skip H and M table generation */
     }
     else
 #endif
```

## Why

`WOLF_CRYPTO_CB_AES_SETKEY` is the older, AES-specific cryptocb
mechanism. `WOLF_CRYPTO_CB_SETKEY` is the newer generic SetKey
dispatch — a single hook handles AES, HMAC, ECC, and any future
algorithm via `WC_SETKEY_AES`, `WC_SETKEY_HMAC`, etc. tags
(`cryptocb.c:2615`, `cryptocb.h::wc_SetKeyType`).

For AES specifically, both mechanisms have identical semantics: when
the registered callback successfully imports the key into a secure
element, `wc_AesSetKey` returns 0 without populating the software key
schedule (`aes->rounds`, `aes->keylen`, expanded key bytes). The SE
owns the key; software state is intentionally left empty.

`wc_AesGcmSetKey` then needs to generate the GCM hash subkey `H` and
(when enabled) the M0 table. The existing path at `aes.c:7861-7900`
does this by calling `wc_AesEncrypt(aes, iv_zeros, aes->gcm.H)`. But
`wc_AesEncrypt` checks the rounds count at `aes.c:3148` and bails
with `KEYUSAGE_E` (-226) if `aes->rounds == 0`. The H/M-skip guard at
line 7855 exists precisely to avoid this dead-end.

The bug: that guard only checks the AES-specific macro. Any port using
the generic `WOLF_CRYPTO_CB_SETKEY` mechanism — the wolfSSL-recommended
direction for new ports — falls through to the broken path and gets
`-226` on every `wc_AesGcmSetKey` call.

## Discovery

I hit this while bringing up a Caliptra (CHIPS Alliance hardware Root
of Trust) cryptocb port that uses `WOLF_CRYPTO_CB_SETKEY`. The Caliptra
port is on a separate branch and PR; this one-line fix is the
prerequisite that should land first.

## Compatibility

- No behavior change for builds that don't define either macro.
- No behavior change for builds that only define
  `WOLF_CRYPTO_CB_AES_SETKEY` (the existing condition still triggers
  via the left operand of the `||`).
- New behavior only when `WOLF_CRYPTO_CB_SETKEY` is defined AND a
  cryptocb is registered AND `devId != INVALID_DEVID` AND `devCtx !=
  NULL` AND the callback returned success — i.e., exactly the case
  that's currently broken.

## Testing

Three configurations verified, all `0 FAIL`:

| Config | Result |
|---|---|
| `./configure && make check` | 5 PASS, 4 SKIP, 0 FAIL |
| `./configure --enable-cryptocb CFLAGS=-DWOLF_CRYPTO_CB_SETKEY && make check` | 5 PASS, 4 SKIP, 0 FAIL |
| `./configure --enable-all && make check` | 17 PASS, 5 SKIP, 0 FAIL |

The 4-5 SKIPs across configurations are all network/external-dep
(openssl interop, google.com TLS, etc.) — standard for offline CI.

A full end-to-end demonstration that this enables HW-offloaded GCM via
`WOLF_CRYPTO_CB_SETKEY` is in the Caliptra port branch (separate PR);
that branch with this fix applied passes 12 caliptra subtests
including AES-GCM round-trip, KAT, and authentication-failure
detection through the SetKey-dispatched path.

### DIFF
--- a/wolfcrypt/src/aes.c
+++ b/wolfcrypt/src/aes.c
@@ -8653,7 +8653,7 @@ int wc_AesGcmSetKey(Aes* aes, const byte* key, word32 len)
 #if !defined(FREESCALE_LTC_AES_GCM) && !defined(WOLFSSL_PSOC6_CRYPTO)
 
 
-#ifdef WOLF_CRYPTO_CB_AES_SETKEY
+#if defined(WOLF_CRYPTO_CB_AES_SETKEY) || defined(WOLF_CRYPTO_CB_SETKEY)
     if ((ret == 0) && (aes->devId != INVALID_DEVID && aes->devCtx != NULL)) {
         /* SE owns key - skip H and M table generation */
     }


MarkAtwood:
Discovered while bringing up the Caliptra cryptocb port; that work is in draft PR #10579 (held until this lands).

wolfcrypt/src/aes.c
```diff
@@ -7852,7 +7852,7 @@ int wc_AesGcmSetKey(Aes* aes, const byte* key, word32 len)
 #if !defined(FREESCALE_LTC_AES_GCM) && !defined(WOLFSSL_PSOC6_CRYPTO)
 
 
-#ifdef WOLF_CRYPTO_CB_AES_SETKEY
+#if defined(WOLF_CRYPTO_CB_AES_SETKEY) || defined(WOLF_CRYPTO_CB_SETKEY)
     if ((ret == 0) && (aes->devId != INVALID_DEVID && aes->devCtx != NULL)) {
         /* SE owns key - skip H and M table generation */
     }
```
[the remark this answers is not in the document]
MarkAtwood: Out of scope — this PR extends an existing guard to cover an additional define. The WOLF_CRYPTO_CB_AES_SETKEY path it mirrors also has no dedicated test. A CryptoCb SetKey test would be good but belongs in a separate PR.

wolfcrypt/src/aes.c
```diff
@@ -7852,7 +7852,7 @@ int wc_AesGcmSetKey(Aes* aes, const byte* key, word32 len)
 #if !defined(FREESCALE_LTC_AES_GCM) && !defined(WOLFSSL_PSOC6_CRYPTO)
 
 
-#ifdef WOLF_CRYPTO_CB_AES_SETKEY
+#if defined(WOLF_CRYPTO_CB_AES_SETKEY) || defined(WOLF_CRYPTO_CB_SETKEY)
```
dgarske, line 8656: Extend the guard at line 7911 with the same `|| defined(WOLF_CRYPTO_CB_SETKEY)` condition so the two paired guards stay symmetric. Leaving them out of sync means the SE-owns-key path still leaks the key into `aes->devKey`, defeating the key-isolation property that is the entire reason the H-skip exists.

dgarske, line 8656: Recommendation: Clear `aes->devCtx` (or the explicit ownership flag suggested in the related finding) on the `CRYPTOCB_UNAVAILABLE` fall-through in both `wc_AesSetKey` (aes.c:4627) and `wc_AesSetKeyLocal` (aes.c:5088). A wrong-but-successful GCM tag is far worse than the -226 this PR is removing; the skip must only happen when this specific key install was actually claimed by the device.

dgarske, line 8656: Recommendation: Add a mock guarded by `WOLF_CRYPTO_CB_SETKEY && !WOLF_CRYPTO_CB_AES_SETKEY` whose `WC_SETKEY_AES` handler sets `aes->devCtx` and does not populate the software schedule, then assert `wc_AesGcmSetKey` returns 0 and a GCM round-trip succeeds through the callback. Add a negative test too: callback sets `devCtx`, then returns `CRYPTOCB_UNAVAILABLE` for the GCM operation - it must not silently produce a zero-H tag. Add a matching CI line such as `--enable-cryptocb --enable-aesgcm --enable-cryptocbutils=setkey` to `.github/workflows/os-check.yml`.

dgarske (changes requested):

philljj:
Retest this please.

no logs

wolfcrypt/src/aes.c
philljj: I think this is still unresolved?

philljj (changes requested):

PR #10582, opened 2026-06-03
Fenrir 2026-06-02: TLS/DTLS correctness, resumption & renegotiation safety fixes

julek-wolfssl:
## Summary

Addresses ten Fenrir-tracked findings (TLS/DTLS correctness, session-resumption safety, renegotiation safety, and key hygiene). Each fix is its own commit, tagged with its `F-<NUMBER>`.

## Fixes

- **F-3888** — Add negative tests for tampered RSA signatures/hashes in `test_wolfSSL_RSA_verify` and `test_wolfSSL_RSA_padding_add_PKCS1_PSS` (PKCS#1 v1.5 and PSS), guarding the `XMEMCMP` signature-acceptance check.
- **F-4867** — `DoTls13EncryptedExtensions` now rejects trailing bytes after the extensions block (length equality, RFC 8446 §4.3.1).
- **F-4868** — `DoTls13CertificateRequest` now rejects trailing bytes after the extensions block (RFC 8446 §4.3.2).
- **F-5633** — Zeroize DTLS 1.3 ChaCha record-number protection keys before freeing (`FreeCiphers`, plus the `wc_Chacha_SetKey` failure path).
- **F-5807** — Client enforces extended_master_secret consistency on resumption; abort on mismatch (RFC 7627 §5.3).
- **F-5810** — Server requires `renegotiation_info` on a renegotiation ClientHello and aborts if absent; also rejects an SCSV received during renegotiation (RFC 5746 §3.5/§3.7).
- **F-5811** — Client verifies the resumed cipher suite matches the cached session and aborts on mismatch (RFC 5246 §7.4.1.3).
- **F-5813** — `BuildMessage` refuses to emit a TLS 1.2 record once the write sequence number reaches its maximum, instead of silently wrapping (new `SEQUENCE_NUMBER_E`, RFC 5246 §6.1).
- **F-5818** — Invalidate the cached session on a fatal alert (receipt and transmission) so it cannot be resumed (RFC 5246 §7.2.2).
- **F-4144** — Honor `WOLFSSL_OP_NO_RENEGOTIATION`: refuse peer-initiated renegotiation with a `no_renegotiation` *warning* while keeping the connection established, on both the client (HelloRequest) and server (renegotiation ClientHello) paths.

## Tests

New memio regression tests in `tests/api/test_tls_ext.c`:
- `test_scr_no_renegotiation_option` — server refuses renegotiation, stays established, and still passes application data.
- `test_helloRequest_no_renegotiation_option` — client refuses a server HelloRequest, stays established, and still passes application data.

Both directions of post-refusal application data exchange are asserted.

## Verification

- `./commit-tests.sh` (current / `--disable-fastmath` / full configs) — all pass, 0 failures.
- `./configure --enable-all && make check` — pass, 0 failures.
- `make check` on a secure-renegotiation config (exercising the new SCR tests) — pass, 0 failures.

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -3343,6 +3343,10 @@ void FreeCiphers(WOLFSSL* ssl)
     ssl->dtlsRecordNumberDecrypt.aes = NULL;
 #endif /* BUILD_AES */
 #ifdef HAVE_CHACHA
+    if (ssl->dtlsRecordNumberEncrypt.chacha)
+        ForceZero(ssl->dtlsRecordNumberEncrypt.chacha, sizeof(ChaCha));
+    if (ssl->dtlsRecordNumberDecrypt.chacha)
+        ForceZero(ssl->dtlsRecordNumberDecrypt.chacha, sizeof(ChaCha));
     XFREE(ssl->dtlsRecordNumberEncrypt.chacha, ssl->heap, DYNAMIC_TYPE_CIPHER);
     XFREE(ssl->dtlsRecordNumberDecrypt.chacha, ssl->heap, DYNAMIC_TYPE_CIPHER);
     ssl->dtlsRecordNumberEncrypt.chacha = NULL;
--- a/src/internal.c
+++ b/src/internal.c
@@ -32263,6 +32263,22 @@ static void MakePSKPreMasterSecret(Arrays* arrays, byte use_psk_key)
         }
         else {
             if (DSH_CheckSessionId(ssl)) {
+                /* RFC 7627 5.3: resumed session EMS state must match the
+                 * ServerHello; abort on mismatch. Stateless (session-ticket)
+                 * resumption - e.g. EAP-FAST, whose PAC is a TLS ticket - binds
+                 * the EMS state in the ticket and need not re-advertise the
+                 * extension, so this applies only to session-ID resumption. */
+                if (
+            #ifdef HAVE_SESSION_TICKET
+                    ssl->session->ticketLen == 0 &&
+            #endif
+                    ssl->session->haveEMS != ssl->options.haveEMS) {
+                    WOLFSSL_MSG("Resumed session EMS state does not match "
+                                "ServerHello EMS state");
+                    SendAlert(ssl, alert_fatal, handshake_failure);
+                    WOLFSSL_ERROR_VERBOSE(EXT_MASTER_SECRET_NEEDED_E);
+                    return EXT_MASTER_SECRET_NEEDED_E;
+                }
                 if (SetCipherSpecs(ssl) == 0) {
                     if (!HaveUniqueSessionObj(ssl)) {
                         WOLFSSL_MSG("Unable to have unique session object");
--- a/src/internal.c
+++ b/src/internal.c
@@ -32362,14 +32362,26 @@ static void MakePSKPreMasterSecret(Arrays* arrays, byte use_psk_key)
                 /* RFC 5246 Section 7.4.1.3: on resumption the ServerHello
                  * reuses the previously negotiated cipher suite. Reject a
                  * server that resumes the session but selects a different
-                 * suite. Skipped for ticket resumption (suite is bound in the
-                 * ticket), consistent with the EMS check above. */
-                if (
-            #ifdef HAVE_SESSION_TICKET
-                    ssl->session->ticketLen == 0 &&
-            #endif
-                    (ssl->options.cipherSuite0 != ssl->session->cipherSuite0 ||
-                        ssl->options.cipherSuite != ssl->session->cipherSuite)) {
+                 * suite. Unlike the EMS check above this also covers ticket
+                 * resumption: the ticket is opaque to the client, so it cannot
+                 * rely on the suite being bound inside the ticket and must
+                 * enforce the match against the suite retained in the cached
+                 * session (SetupSession stores it for ticket sessions too).
+                 * Otherwise a server could resume a ticket under a different,
+                 * weaker suite the client offered and the downgrade would go
+                 * undetected.
+                 *
+                 * Skip only when the client retained no suite for the session
+                 * (cipherSuite0/cipherSuite both zero): then there is nothing to
+                 * compare against. This is the case for EAP-FAST, whose PAC is a
+                 * TLS ticket whose keys are supplied through the session-secret
+                 * callback and which never populates the cached suite. (0,0) is
+                 * TLS_NULL_WITH_NULL_NULL, never a negotiated suite, so it
+                 * unambiguously means "no retained suite". */
+                if ((ssl->session->cipherSuite0 != 0 ||
+                            ssl->session->cipherSuite != 0) &&
+                        (ssl->options.cipherSuite0 != ssl->session->cipherSuite0 ||
+                         ssl->options.cipherSuite != ssl->session->cipherSuite)) {
                     WOLFSSL_MSG("Resumed session cipher suite does not match "
                                 "ServerHello cipher suite");
                     SendAlert(ssl, alert_fatal, illegal_parameter);
--- a/src/internal.c
+++ b/src/internal.c
@@ -24777,11 +24777,11 @@ int BuildMessage(WOLFSSL* ssl, byte* output, int outSz, const byte* input,
 #endif
 
 #ifndef WOLFSSL_NO_TLS12
-    /* RFC 5246 6.1: record sequence numbers MUST NOT wrap. Refuse to emit a
-     * record once the write sequence number has reached its maximum value
-     * (2^64-1); reusing sequence number 0 with the same keys would break the
-     * record protection. The caller must renegotiate or close the connection
-     * instead. DTLS sequence numbers are epoch-scoped and handled elsewhere. */
+    /* RFC 5246 6.1: sequence numbers MUST NOT wrap. GetSEQIncrement post-
+     * increments, so refuse at hi == lo == 0xFFFFFFFF (2^64-1): that last legal
+     * value is deliberately sacrificed to avoid wrapping to 0 and reusing
+     * sequence number 0. The caller must renegotiate or close. DTLS sequence
+     * numbers are epoch-scoped and handled elsewhere. */
     if (!sizeOnly && !ssl->options.dtls &&
             ssl->keys.sequence_number_hi == 0xFFFFFFFFU &&
             ssl->keys.sequence_number_lo == 0xFFFFFFFFU) {
--- a/src/internal.c
+++ b/src/internal.c
@@ -18051,6 +18051,15 @@ static int DoHelloRequest(WOLFSSL* ssl, word32 size)
     }
 #ifdef HAVE_SECURE_RENEGOTIATION
     else if (ssl->secure_renegotiation && ssl->secure_renegotiation->enabled) {
+        /* WOLFSSL_OP_NO_RENEGOTIATION: caller opted into rejecting
+         * peer-initiated renegotiation. Respond with a no_renegotiation
+         * warning alert instead of starting a secure renegotiation. */
+        if (ssl->options.mask & WOLFSSL_OP_NO_RENEGOTIATION) {
+            WOLFSSL_MSG("Rejecting HelloRequest: WOLFSSL_OP_NO_RENEGOTIATION");
+            WOLFSSL_LEAVE("DoHelloRequest", 0);
+            WOLFSSL_END(WC_FUNC_HELLO_REQUEST_DO);
+            return SendAlert(ssl, alert_warning, no_renegotiation);
+        }
         ssl->secure_renegotiation->startScr = 1;
         WOLFSSL_LEAVE("DoHelloRequest", 0);
         WOLFSSL_END(WC_FUNC_HELLO_REQUEST_DO);
@@ -18783,6 +18792,17 @@ int DoHandShakeMsgType(WOLFSSL* ssl, byte* input, word32* inOutIdx,
             ssl->secure_renegotiation &&
             ssl->secure_renegotiation->enabled)
     {
+        /* WOLFSSL_OP_NO_RENEGOTIATION: caller opted into rejecting
+         * peer-initiated renegotiation. RFC 5246 7.2.2: no_renegotiation is a
+         * warning-level alert, so refuse the renegotiation but keep the
+         * established connection rather than aborting it. Skip the ClientHello
+         * body and leave handshake state untouched, mirroring the client-side
+         * HelloRequest refusal in DoHelloRequest(). */
+        if (ssl->options.mask & WOLFSSL_OP_NO_RENEGOTIATION) {
+            WOLFSSL_MSG("Refusing renegotiation: WOLFSSL_OP_NO_RENEGOTIATION");
+            *inOutIdx = expectedIdx;
+            return SendAlert(ssl, alert_warning, no_renegotiation);
+        }
         WOLFSSL_MSG("Reset handshake state");
         XMEMSET(&ssl->msgsReceived, 0, sizeof(MsgsReceived));
         ssl->options.serverState = NULL_STATE;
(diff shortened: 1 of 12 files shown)


julek-wolfssl:
Retest this please..

douzzer:
analysis from Opus 4.8-High:

F-5811 / F-5807 skip ticket resumption. Both the suite-match and EMS checks are gated behind `ssl->session->ticketLen == 0`. For EMS this is well-justified (EAP-FAST PAC, stateless binding — the comment is right). For the cipher-suite check I'm less convinced: if the client retains `session->cipherSuite0`/`cipherSuite` for ticket-based sessions, then skipping the comparison leaves a downgrade gap symmetric to the session-ID hole F-5811 closes — a server could resume a ticket under a different (weaker) suite and the client wouldn't catch it. The "suite is bound in the ticket" rationale only constrains an honest server; the ticket is opaque to the client, so it can't independently enforce the binding except by comparing against its own retained `session->cipherSuite`.

Worth confirming whether ticket-resumed suite is validated elsewhere on the client path. If not, this is a candidate for a precise follow-up (apply the same `cipherSuite0`/`cipherSuite` comparison to the ticket path, leaving only the EMS skip ticket-gated).

julek-wolfssl:
@douzzer addressed

src/internal.c
```diff
@@ -32259,6 +32342,53 @@ static void MakePSKPreMasterSecret(Arrays* arrays, byte use_psk_key)
         }
         else {
             if (DSH_CheckSessionId(ssl)) {
```
dgarske, line 32407: **🟠 [Medium] Client-side EMS resumption check (server-drops-EMS direction) has no test**

F-5807 adds a NEW client-side check in `CompleteServerHello`: if the resumed session used EMS but the ServerHello drops it (or vice-versa), abort with `EXT_MASTER_SECRET_NEEDED_E`. The only EMS-resumption test, `test_tls_ems_resumption_downgrade` (tests/api/test_tls_ext.c:108), exercises the *opposite* direction — the client drops EMS and the *server* catches it via the pre-existing check at src/internal.c:38341 (it asserts `wolfSSL_get_error(ssl_s, 0) == EXT_MASTER_SECRET_NEEDED_E`). The newly introduced client-side branch (a malicious/test server resuming a session-ID session while omitting EMS) is never executed by any test, so a regression in this branch would go undetected.

**Fix:** Add a client-direction EMS-downgrade regression test to lock in the new branch.

src/internal.c
```diff
@@ -32259,6 +32342,53 @@ static void MakePSKPreMasterSecret(Arrays* arrays, byte use_psk_key)
...
+                /* RFC 7627 5.3: resumed session EMS state must match the
+                 * ServerHello; abort on mismatch. Stateless (session-ticket)
+                 * resumption - e.g. EAP-FAST, whose PAC is a TLS ticket - binds
+                 * the EMS state in the ticket and need not re-advertise the
+                 * extension, so this applies only to session-ID resumption. */
+                if (
```
dgarske: **🟠 [Medium] EMS resumption check excludes ALL ticket resumptions, asymmetric with the suite check**

The EMS consistency check is skipped for any session with `ticketLen > 0`, citing EAP-FAST (whose PAC is a ticket). But the very next check (F-5811, cipher suite) explicitly DOES cover tickets and instead carves out EAP-FAST more narrowly — by skipping only when the cached suite is (0,0). For a generic (non-EAP-FAST) TLS session-ticket resumption the client still has `session->haveEMS` populated by `SetupSession`, so it could enforce RFC 7627 5.3 just like the suite check does. As written, a MITM/server can drop EMS on a generic ticket resumption undetected on the client. The EAP-FAST scenario is more precisely detectable via `ssl->sessionSecretCb != NULL` (see src/internal.c:32283), which would let normal ticket resumptions keep the EMS guard. This is a deliberate, documented scoping (a net improvement over no check at all), so flagging as a question rather than a bug.

**Fix:** Confirm whether generic ticket resumptions are intentionally exempt from EMS enforcement, or tighten the carve-out to EAP-FAST.

dgarske (commented):
## Skoll Code Review

**Findings:** 8 total — 3 posted, 5 skipped

### Posted findings
- [Medium] **Client-side EMS resumption check (server-drops-EMS direction) has no test** — `src/internal.c:32344-32360`
- [Medium] **EMS resumption check excludes ALL ticket resumptions, asymmetric with the suite check** — `src/internal.c:32350-32360`
- [Low] **Trailing-byte rejection returns BUFFER_ERROR without a TLS alert** — `src/tls13.c:6033, src/tls13.c:6171-6173`

### Skipped findings
- [Low] ~~Fatal-alert session invalidation (F-5818) is untested on both send and receive paths~~
- [Low] ~~TLS 1.3 send path evicts session only on alert_fatal, asymmetric with receive path~~
- [Medium] ~~No negative tests for TLS 1.3 trailing-byte rejection (EncryptedExtensions / CertificateRequest)~~
- [Low] ~~New brace-only block introduced solely to declare a variable in test~~
- [Medium] ~~Most correctness fixes ship without dedicated regression tests~~

src/internal.c
julek-wolfssl: Addressed.

src/internal.c
julek-wolfssl: Addressed.

dgarske (approved):

dgarske:
Jenkins retest this please: "Build 'PRB-valgrind-check-v3' failed with result: FAILURE" -> "[cd wolfssl && ../testing/Jenkins/PRB-valgrind-check-v3/valgrind-check.sh --enable-opensslall --enable-sniffer CPPFLAGS="-DWOLFSSL_OLD_PRIME_CHECK"]"

```
 Success -- All results as expected.
 End Cipher Suite Tests
unit_test: Success for all configured tests.
==23307== 
==23307== HEAP SUMMARY:
==23307==     in use at exit: 0 bytes in 0 blocks
==23307==   total heap usage: 2,363,029 allocs, 2,363,029 frees, 861,792,597 bytes allocated
==23307== 
==23307== All heap blocks were freed -- no leaks are possible
==23307== 
==23307== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
Exiting with status: 3
Script ran for 646 seconds.
script returned exit code 3
```

douzzer:
retest this please
(valgrind PRB failed, no logs)

dgarske (commented):
## Skoll Code Review

### Posted findings
- [High] **Client EMS/cipher-suite resumption checks fire on unconfirmed ticket resumption, breaking RFC 5077 ticket-decline fallback to full handshake** — `src/internal.c:32344-32395`
- [Medium] **DoAlert invalidates the cached session before the alert is validated; forged TLS 1.3 plaintext alert evicts the session** — `src/internal.c:22596-22604`
- [Medium] **test_tls12_resume_ticket_wrong_suite missing WOLFSSL_NO_DEF_TICKET_ENC_CB gate** — `tests/api/test_tls.c:889-896`
- [Low] **Null-pointer checks in FreeCiphers use truthiness instead of explicit NULL comparison** — `src/internal.c:3346-3349`
- [Low] **BuildMessage refuses the last RFC-legal sequence number (2^64-1)** — `src/internal.c:24717-24728`

### Skipped findings
- [Medium] ~~No regression tests for F-4867, F-4868, F-5810, F-5813, F-5818, F-5633~~
- [Low] ~~SecureRenegotiation struct: new member inserted mid-struct and existing byte fields converted to bitfields~~

src/internal.c
```diff
@@ -32259,6 +32342,57 @@ static void MakePSKPreMasterSecret(Arrays* arrays, byte use_psk_key)
         }
         else {
             if (DSH_CheckSessionId(ssl)) {
```
dgarske, line 32407: **🔴 [High] Client EMS/cipher-suite resumption checks fire on unconfirmed ticket resumption, breaking RFC 5077 ticket-decline…**

The new F-5807 EMS check and F-5811 cipher-suite check run inside `if (DSH_CheckSessionId(ssl))` in `CompleteServerHello`, i.e. at ServerHello-parse time. For session-ID resumption this is sound: `DSH_CheckSessionId` clause 3 requires the server to echo the cached session ID, which is the definitive TLS 1.2 resumption signal. But clause 2 returns true merely because `ssl->session->ticketLen > 0` -- at ServerHello time the client CANNOT yet know whether the server accepted the ticket (RFC 5077 Section 3.4: with an empty session ID, 'the client ... determines if the server is resuming a session by the subsequent handshake messages'). wolfSSL's ticket client sends an empty session ID (`idSz = 0`, src/internal.c:31614), and the pre-existing code handled the decline case gracefully via the 'Not resuming as thought' path (src/internal.c:18913-18949). The new checks abort at ServerHello before that fallback can run.

Empirically verified against this branch (build with --enable-session-ticket): handshake 1 establishes a ticket session on ECDHE-RSA-AES128-GCM-SHA256; handshake 2 resumes against a fresh stateless server ctx (session cache off, new default ticket key -> ticket decline) configured with ECDHE-RSA-AES256-GCM-SHA384, which the client also offers. Pre-change (checks removed): full handshake succeeds, `wolfSSL_session_reused()==0`, AES256 suite negotiated. With this PR: the client aborts with `MATCH_SUITE_ERROR` (-501) and sends a fatal illegal_parameter alert. This is exactly the stale-ticket-against-reconfigured-server / heterogeneous-server-farm scenario where ticket…

**Fix:** Rework the ticket-path enforcement so it only triggers on confirmed resumption; add a regression test for ticket-decline (fresh server ctx with session cache off and a different cipher suite) asserting the full-handshake fallback still succeeds.

src/internal.c
```diff
@@ -22547,6 +22593,15 @@ static int DoAlert(WOLFSSL* ssl, byte* input, word32* inOutIdx, int* type)
                 code != close_notify && code != user_canceled) {
             ssl->options.isClosed = 1;
         }
+#ifndef NO_SESSION_CACHE
```
dgarske: **🟠 [Medium] DoAlert invalidates the cached session before the alert is validated; forged TLS 1.3 plaintext alert evicts the session**

The new F-5818 invalidation runs immediately after the alert level/code bytes are read, but BEFORE the existing check at src/internal.c:22618-22636 that rejects a plaintext alert received while encryption is on in TLS 1.3 ('Unexpected message when encryption is on and alert not encrypted' -> PARSE_ERROR). That code path exists precisely because unauthenticated plaintext alerts reach DoAlert in TLS 1.3 (see the WOLFSSL_TLS13_IGNORE_PT_ALERT_ON_ENC machinery). So an off-path attacker who injects a plaintext fatal alert record gets the cached session/ticket evicted by `InvalidateSessionOnFatalAlert` even though the function then declares the alert invalid. The pre-existing behavior already tears down the connection on such a forged record (PARSE_ERROR), but the PR adds persistent state mutation (session-cache eviction -> forced full handshake on reconnect) keyed off unvalidated input. Note that simply moving the DoAlert call later is not sufficient on its own: the PARSE_ERROR path calls `SendAlert(ssl, alert_fatal, unexpected_message)`, which now also invalidates via the new SendAlert hook (src/internal.c:27496-27503), so the eviction-on-forged-input would still occur through that route. Both paths need a conscious decision. RFC 8446 6.2 does not require session invalidation (the PR comment itself calls the TLS 1.3 eviction defense-in-depth), so suppressing eviction for the plaintext-alert-under-encryption case loses nothing.

**Fix:** Only invalidate on alerts that pass the plaintext-under-encryption validation; decide explicitly whether the internally generated unexpected_message teardown alert should evict (arguably yes, since the connection is genuinely terminating, but document it).

tests/api/test_tls.c
```diff
@@ -876,6 +876,99 @@ int test_tls12_etm_failed_resumption(void)
...
+ * session's cipher suite. The ticket is opaque to the client, so the client
+ * cannot rely on the suite being bound inside it and must compare the
+ * ServerHello suite against the suite retained in the cached session (F-5811
+ * does this for session-ID resumption; it must hold for tickets too). This
+ * test establishes a ticket-based session, rewrites the cached session's suite
+ * to emulate a server that resumes the ticket under a different suite, and
+ * asserts the client aborts the resumption with MATCH_SUITE_ERROR. The same
+ * server CTX is reused for the second handshake so its ticket key persists. */
+int test_tls12_resume_ticket_wrong_suite(void)
+{
```
dgarske, line 889: **🟠 [Medium] test_tls12_resume_ticket_wrong_suite missing WOLFSSL_NO_DEF_TICKET_ENC_CB gate**

The test's #if guard requires HAVE_SESSION_TICKET but not `!defined(WOLFSSL_NO_DEF_TICKET_ENC_CB)`. The server side relies on the default ticket encryption callback to issue a ticket (no callback is registered by test_memio_setup), so in a build with WOLFSSL_NO_DEF_TICKET_ENC_CB the server issues no ticket and `ExpectIntGT(sess->ticketLen, 0)` fails -- the test FAILS instead of being skipped. The sibling test added in the same PR, `test_tls_ems_resumption_server_downgrade`, correctly gates its ticket variant on `defined(HAVE_SESSION_TICKET) && !defined(WOLFSSL_NO_DEF_TICKET_ENC_CB)` (tests/api/test_tls_ext.c:284-286); this test should match.

**Fix:** Add !defined(WOLFSSL_NO_DEF_TICKET_ENC_CB) to the guard so the test skips rather than fails in builds without the default ticket encryption callback.

src/internal.c
```diff
@@ -3343,6 +3343,10 @@ void FreeCiphers(WOLFSSL* ssl)
     ssl->dtlsRecordNumberDecrypt.aes = NULL;
 #endif /* BUILD_AES */
 #ifdef HAVE_CHACHA
+    if (ssl->dtlsRecordNumberEncrypt.chacha)
```
dgarske: **🔵 [Low] Null-pointer checks in FreeCiphers use truthiness instead of explicit NULL comparison**

The new guards use `if (ssl->dtlsRecordNumberEncrypt.chacha)` rather than the project's preferred explicit form `if (ptr != NULL)` for new code. The adjacent pre-existing poly1305 block uses the same truthy style, so this matches local context; flagging only because it is new code.

**Fix:** Use explicit `!= NULL` comparisons; otherwise the ForceZero-before-XFREE fix itself is correct (verified the matching failure-path fix in Dtls13InitChaChaCipher as well).

src/internal.c
```diff
@@ -24658,6 +24714,19 @@ int BuildMessage(WOLFSSL* ssl, byte* output, int outSz, const byte* input,
 #endif
 
 #ifndef WOLFSSL_NO_TLS12
+    /* RFC 5246 6.1: record sequence numbers MUST NOT wrap. Refuse to emit a
```
dgarske: **🔵 [Low] BuildMessage refuses the last RFC-legal sequence number (2^64-1)**

WriteSEQ (src/internal.c:9569-9574) uses the current counter value as the record's sequence number and post-increments, so when hi==lo==0xFFFFFFFF the record about to be built would use sequence number 2^64-1 -- which RFC 5246 6.1 still permits ('may not exceed 2^64-1'); only the subsequent wrap to 0 is forbidden. The new check therefore refuses one record earlier than strictly required. This is a reasonable conservative choice (it avoids the silent post-increment wrap entirely and costs nothing in practice), but the comment says 'has reached its maximum value (2^64-1)' as if that value itself were unusable. Also verified the placement is correct: it sits after the TLS 1.3 dispatch to BuildTls13Message, excludes DTLS and sizeOnly, and is deterministic across async-crypt re-entry.

**Fix:** Either keep the behavior and adjust the comment to state that the last sequence number is deliberately sacrificed to avoid the post-increment wrap, or detect the wrap at increment time in WriteSEQ.

dgarske (changes requested):
Please review skoll reports: https://github.com/wolfSSL/wolfssl/pull/10582#pullrequestreview-4478990945

src/internal.c
julek-wolfssl: addressed

src/internal.c
julek-wolfssl: addressed

tests/api/test_tls.c
julek-wolfssl: addressed

src/internal.c
julek-wolfssl: addressed

src/internal.c
julek-wolfssl: addressed

dgarske (approved):

PR #10268, opened 2026-04-21
zephyr: changes needed for Zephyr 4.3 default TLS support

ColtonWilley:
Follow-up to #7731. Zephyr 4.3's TLS socket integration uses three additional
wolfSSL features plus an extension to the native_sim time-source gates
introduced in #7731.

native_sim timer gates (src/internal.c, wolfcrypt/src/wc_port.c):
  Extend !CONFIG_BOARD_NATIVE_POSIX gates in LowResTimer() and z_time() RTC
  path to also cover CONFIG_BOARD_NATIVE_SIM. Zephyr 4.3 renamed the simulator
  board. Behavior on native_posix is unchanged.

New Kconfig options (zephyr/Kconfig, zephyr/user_settings.h):
  CONFIG_WOLFSSL_SESSION_EXPORT   -> HAVE_EXT_CACHE
  CONFIG_WOLFSSL_KEEP_PEER_CERT   -> KEEP_PEER_CERT
  CONFIG_WOLFSSL_ALWAYS_VERIFY_CB -> WOLFSSL_ALWAYS_VERIFY_CB

All default-off. Customers opt in the same way as CONFIG_WOLFSSL_DTLS / ALPN / PSK.

.wolfssl_known_macro_extras: register HAVE_EXT_CACHE (new use in zephyr/user_settings.h).

Companion OSP patch for Zephyr 4.3 will follow in wolfSSL/osp and depends on this PR.

PR #10583, opened 2026-06-03
Fixes for Zephyr secure sockets integration

Frauschi:
Follow-up for #10268 with some more changes required for the updated patch set for Zephyr 4.3 secure sockets integration (for https://github.com/wolfSSL/osp/pull/330 with an iteration of the patch currently under review by @ColtonWilley).


Frauschi:
Jenkins retest this please

zephyr/samples/wolfssl_tls_sock/prj-no-malloc.conf
```diff
@@ -3,22 +3,12 @@ CONFIG_MAIN_STACK_SIZE=655360
 CONFIG_ENTROPY_GENERATOR=y
 CONFIG_INIT_STACKS=y
 
-# General config
-CONFIG_NEWLIB_LIBC=y
-
-# Pthreads
```
ColtonWilley, line 9: This is not for using Zephyr native TLS interface swap, this is for general app integration. Im not sure if we should be changing these defaults unless we are absolutely sure.

Frauschi: You are right that these changes are not directly related to the `secure_sockets` integration. However, they are required to make the sample compile for Zephyr 4.3 in the `no-malloc` config (as these are not present in newer Zephyr versions after 3.7). For older versions, these configs are still enabled via the `zephyr_legacy.conf` file. So we are safe to remove them here.

douzzer (approved):

PR #10584, opened 2026-06-03
Support hardware PUF on lpc55s69

twcook86:
This adds hardware puf support to wolfCrypt...

1) Provides a new HWPUF api (WOLFSSL_HWPUF), based on crypto callbacks, to support a common set of puf features across multiple chips
2) Supports lpc55s69 hw puf device (WOLFSSL_NXP_HWPUF)

An effort has been made to be as compatible with the sw puf (WOLFSSL_PUF) implementation as possible, but the two fundamentally diverge in certain respects.

Not all features of the hw puf are supported yet.  Specifically, this initial implementation supports...
1) keys of size 16, 24, 32 bytes for use with aes
2) ability to pass keys directly on the hw bus to the aes engine, thus never being exposed in ram

Note: this pr goes with https://github.com/wolfSSL/wolfBoot/pull/787

# Testing
Tested with new test functionality added to wolfcrypt/test/test.c

# Checklist

 - [x] added tests
 - [x] updated/added doxygen


wolfcrypt/src/puf.c
```diff
@@ -464,6 +464,18 @@ int wc_PufInit(wc_PufCtx* ctx)
     return 0;
 }
 
+int wc_PufDeinit(wc_PufCtx* ctx)
```
dgarske: Shouldn't this entire file not be compiled when `WOLFSSL_NXP_PUF` is defined? Maybe add a !WOLFSSL_NXP_PUF around it? Also did you consider extending the crypto callbacks to support PUF? So have an init_ex with devId/heap and support registering a PUF handler for the NXP?

dgarske (changes requested):

wolfcrypt/src/hwpuf.c
```diff
@@ -0,0 +1,223 @@
...
+    #include <wolfcrypt/src/misc.c>
+#endif
+
+/* The various supported device ports...  One must be defined. */
+#ifdef WOLFSSL_NXP_HWPUF
+    #include <wolfssl/wolfcrypt/port/nxp/hwpuf_port.h>
+#endif
+
+
+WOLFSSL_API int wc_HWPUF_Register(wc_HWPUF* hwpuf, void* heap, int devId)
```
dgarske, line 43: **🔴 [High] wc_HWPUF_Register reads uninitialized hwpuf-registered before zeroing**

`wc_HWPUF_Register` checks `if (hwpuf->registered) return HWPUF_REGISTER_E;` at line 48 BEFORE it `ForceZero(hwpuf, sizeof(wc_HWPUF))` at line 51. The documented usage (doxygen `hwpuf.h` example and `hwpuf_test()` in test.c line 23344/23358) passes a stack `wc_HWPUF hwpuf;` that is NOT zero-initialized by the caller. Reading `hwpuf->registered` therefore reads indeterminate stack memory: this is undefined behavior and can spuriously return `HWPUF_REGISTER_E`, causing registration (and thus all HWPUF use) to fail non-deterministically. Tools like valgrind/MSAN will flag it. The double-register guard cannot reliably work on a freshly-declared struct because Register is also the function that initializes the struct.

**Fix:** Either remove the `hwpuf->registered` pre-zero check (it cannot work reliably on an uninitialized struct), or explicitly document that the caller must `XMEMSET`/`wc_HWPUF` zero the context before calling Register and update the test/doxygen examples accordingly. Add a test that registers, unregisters, and re-registers to validate the intended double-register semantics.

wolfcrypt/src/port/nxp/hwpuf_port.c
```diff
@@ -0,0 +1,351 @@
...
+
+typedef struct nxp_hwpuf_ctx {
+    word32 keyMask; /* unique per reset */
+} nxp_hwpuf_ctx;
+
+static nxp_hwpuf_ctx ctx;
+static puf_config_t conf;
+
+
+static int getACFromPFR(byte *ac)
```
dgarske: **🟠 [Medium] Unused static function getACFromPFR (dead code, uses memset)**

`getACFromPFR` is defined but never called anywhere in the tree (grep confirms only the definition). A file-scope `static` function that is never referenced triggers `-Wunused-function`, which breaks any `-Werror` build of the NXP port. It also uses bare `memset` instead of the wolfSSL `XMEMSET` wrapper, against repo convention.

**Fix:** Remove the unused helper, or actually use it (it appears intended to load the activation code from PFR). If kept, replace `memset` with `XMEMSET`.

wolfcrypt/src/port/nxp/hwpuf_port.c
```diff
@@ -0,0 +1,351 @@
...
+
+    ret = PUF_SetIntrinsicKey(PUF, (puf_key_index_register_t)keyIdx, keySz,
+                              keyCode, keyCodeSz);
+    if (ret != kStatus_Success)
+        return HWPUF_GENERATE_KEY_E;
+
+    return 0;
+}
+
+static int nxp_hwpuf_SetKey(wc_HWPUF* hwpuf, byte keyIdx,
```
dgarske: **🟠 [Medium] nxp_hwpuf_SetKey has multiple unused parameters**

`nxp_hwpuf_SetKey` is a stub returning `CRYPTOCB_UNAVAILABLE`, but its parameters `keyIdx`, `key`, `keySz`, `keyCode`, `keyCodeSz` are never referenced. This produces `-Wunused-parameter` warnings (breaks `-Werror` NXP builds). wolfSSL convention is to add explicit `(void)param;` casts for unimplemented stubs.

**Fix:** Add `(void)` casts for the unused parameters, matching the pattern used elsewhere in the codebase for not-yet-implemented stubs.

wolfcrypt/src/port/nxp/hwpuf_port.c
```diff
@@ -0,0 +1,351 @@
...
+
+    ret = PUF_Zeroize(PUF);
+    PUF_Deinit(PUF, &conf);
+    if (ret != kStatus_Success) {
+        return HWPUF_ZEROIZE_E;
+    }
+    return 0;
+}
+
+static int nxp_hwpuf_CryptoDevCb(int devId, wc_CryptoInfo* info, void* ctx)
```
dgarske: **🟠 [Medium] Callback parameter ctx shadows file-scope global ctx and is unused**

The callback signature `nxp_hwpuf_CryptoDevCb(int devId, wc_CryptoInfo* info, void* ctx)` names its third parameter `ctx`, which shadows the file-scope `static nxp_hwpuf_ctx ctx;` (line 50) that the PUF code relies on (e.g. `ctx.keyMask`). This triggers `-Wshadow` and is a maintenance hazard: any future code in this function that intends to touch the global key-mask context would silently hit the (unused) void* parameter instead. The parameter is also unused, triggering `-Wunused-parameter`.

**Fix:** Rename the parameter (e.g. `devCtx`) to avoid shadowing the global `ctx`, and add `(void)` for the unused parameter.

wolfcrypt/src/port/nxp/hwpuf_port.c
```diff
@@ -0,0 +1,351 @@
...
+
+    if (hwpuf == NULL)
+        return BAD_FUNC_ARG;
+
+    PUF_GetDefaultConfig(&conf);
+    if (PUF_Init(PUF, &conf) != kStatus_Success) {
+        PUF_Deinit(PUF, &conf);
+        return HWPUF_INIT_E;
+    }
+    ctx.keyMask = RNG->RANDOM_NUMBER;
```
dgarske, line 98: **🟠 [Medium] RNG hardware register read without ensuring RNG is initialized**

`ctx.keyMask = RNG->RANDOM_NUMBER;` reads the RNG peripheral's random-number register directly during PUF init, but nothing in this file initializes the RNG peripheral (no `RNG_Init`/`RNG_GetRandomData`). If the RNG clock/peripheral is not already enabled by the application, this read can return 0 or a stale/constant value. `keyMask` is then used as the mask in `PUF_GetHwKey(..., ctx.keyMask)`; a predictable/zero mask weakens the masking of keys delivered on the hardware bus. The return value is also not checked for readiness.

**Fix:** Initialize the RNG (and confirm a fresh value is available) before reading `RANDOM_NUMBER`, or document that the application must initialize the RNG before `wc_HWPUF_Init`. Confirm the masking value is non-trivial.

wolfcrypt/test/test.c
```diff
@@ -23324,6 +23337,171 @@ WOLFSSL_TEST_SUBROUTINE wc_test_ret_t puf_test(void)
...
+    /* ---- Test 7: generate a key and send directly to hw bus ---- */
+    ret = wc_HWPUF_GenerateKey(&hwpuf, 0, 32, keyCode32, sizeof(keyCode32));
+    if (ret != 0)
+        return WC_TEST_RET_ENC_EC(ret);
+    /* get key from keyCode */
+    ret = wc_HWPUF_GetKey(&hwpuf, keyCode32, sizeof(keyCode32), key32_2, sizeof(key32_2));
+    if (ret != 0)
+        return WC_TEST_RET_ENC_EC(ret);
+    { /* key1 should be zeroed */
+        int idx;
```
dgarske: **🔵 [Low] Signed/unsigned comparison in hwpuf_test loop**

`int idx; ... for (idx = 0; idx < sizeof(key32_2); ++idx)` compares a signed `int` against the `size_t` result of `sizeof`, producing a `-Wsign-compare` warning. wolfSSL test builds typically run with `-Wall -Wextra` (and frequently `-Werror`).

**Fix:** Use an unsigned index type (e.g. `word32`) or cast `sizeof` to match, to avoid the sign-compare warning.

wolfssl/wolfcrypt/settings.h
```diff
@@ -2204,6 +2204,9 @@
     #define WOLFSSL_NXP_CASPER_RSA_PUB_EXPTMOD
     #define NO_WOLFSSL_SHA256_INTERLEAVE
 #endif
+#if defined(WOLFSSL_NXP_HWPUF) && !defined(WOLF_CRYPTO_CB)
```
dgarske: **🔵 [Low] WOLFSSL_NXP_HWPUF does not imply WOLFSSL_HWPUF**

Defining `WOLFSSL_NXP_HWPUF` auto-enables `WOLF_CRYPTO_CB`, but does NOT enable `WOLFSSL_HWPUF`. Every HWPUF source file is gated on `WOLFSSL_HWPUF` (`hwpuf.c`), or on `WOLFSSL_HWPUF && WOLFSSL_NXP_HWPUF` (`hwpuf_port.c`, `hwpuf_port.h`). So a user who defines only `WOLFSSL_NXP_HWPUF` gets `WOLF_CRYPTO_CB` turned on but no HWPUF code compiled at all - a silent no-op that is easy to misdiagnose.

**Fix:** Have `WOLFSSL_NXP_HWPUF` also imply `WOLFSSL_HWPUF` so selecting the port is sufficient, consistent with how the port is the concrete implementation of the generic feature.

wolfcrypt/src/port/nxp/hwpuf_port.c
```diff
@@ -0,0 +1,351 @@
...
+}
+
+static int keyCodeCheck(byte* keyCode, word32* keytype,
+                        word32* keyidx, word32* keysize)
+{
+    *keytype = keyCode[0];
+    *keyidx = keyCode[1];
+    *keysize = keyCode[3] == 0 ? 512 : 8 * keyCode[3] ;
+
+    if (*keytype >= 2)
```
dgarske: **🔵 [Low] Inconsistent key-index bound: literal 16 vs kPUF_KeyIndexMax**

`keyCodeCheck` validates `*keyidx >= 16` with a magic literal and `*keytype >= 2` with another magic literal, while `nxp_hwpuf_GenerateKey` validates the same index domain with the named SDK constant `kPUF_KeyIndexMax` (line 169). Using two different spellings of the same bound is inconsistent and brittle if the SDK ever changes the max index.

**Fix:** Use the named SDK constant (`kPUF_KeyIndexMax`) and a named constant for the key-type bound rather than bare literals.

wolfssl/wolfcrypt/cryptocb.h
```diff
@@ -575,6 +578,42 @@ typedef struct wc_CryptoInfo {
         } op;
     } she;
 #endif
+#ifdef WOLFSSL_HWPUF
+    struct {
```
dgarske, line 582: **🔵 [Low] wc_CryptoInfo.hwpuf.hwpuf typed as void* loses type safety**

The `hwpuf` member is declared `void* hwpuf;` even though `cryptocb.h` already `#include`s `hwpuf.h` under `WOLFSSL_HWPUF` (so `wc_HWPUF` is a complete type at this point). Other algo blocks in this union use concrete pointer types. Using `void*` discards compile-time type checking on every callsite that assigns/reads it.

**Fix:** Type the member as `wc_HWPUF*` for type safety, consistent with other algorithm structs in the union.

wolfssl/wolfcrypt/error-crypt.h
```diff
@@ -328,8 +328,18 @@ enum wolfCrypt_ErrorCodes {
     DRBG_SHA512_KAT_FIPS_E = -1017, /* SHA-512 DRBG KAT failure */
     SLH_DSA_KAT_FIPS_E  = -1018, /* SLH-DSA CAST KAT failure */
 
-    WC_SPAN2_LAST_E     = -1018, /* Update to indicate last used error code */
-    WC_LAST_E           = -1018, /* the last code used either here or in
+    HWPUF_REGISTER_E    = -1019, /* HWPUF registration failed */
+    HWPUF_INIT_E        = -1020, /* HWPUF initialization failed */
```
dgarske, line 332: **🔵 [Low] HWPUF_DEINIT_E error code defined but never returned**

`HWPUF_DEINIT_E` (-1021) is added to the error enum and given a string in `error.c`, but it is never returned by any code path - `wc_HWPUF_Deinit`/`nxp_hwpuf_Deinit` always return 0 (or the cryptocb passthrough), never `HWPUF_DEINIT_E`. It is harmless but dead, and it consumes a reserved error code.

**Fix:** Either return `HWPUF_DEINIT_E` when the underlying deinit fails, or remove the unused error code to avoid an orphaned reserved value.

wolfcrypt/test/test.c
```diff
@@ -23324,6 +23337,171 @@ WOLFSSL_TEST_SUBROUTINE wc_test_ret_t puf_test(void)
 }
 #endif /* WOLFSSL_PUF */
 
+#ifdef WOLFSSL_HWPUF
+WOLFSSL_TEST_SUBROUTINE wc_test_ret_t hwpuf_test(void)
```
dgarske, line 23341: **🔵 [Low] No test coverage for SetKey path or double-register**

`hwpuf_test` exercises Init/Enroll/Start/GenerateKey/GetKey/Zeroize and several bad-arg cases, but there is no test for `wc_HWPUF_SetKey` (which currently returns CRYPTOCB_UNAVAILABLE in the NXP port) and no test for the registered-state guard (e.g. calling `wc_HWPUF_Register` twice, or `wc_HWPUF_Init` before Register expecting `HWPUF_REGISTER_E`). Given the uninitialized-read concern in Register, a double-register/registered-guard test would be valuable.

**Fix:** Add tests for the SetKey return contract and for the register/registered state machine.

wolfcrypt/test/test.c
```diff
@@ -23324,6 +23337,171 @@ WOLFSSL_TEST_SUBROUTINE wc_test_ret_t puf_test(void)
...
+    /* generate a 32-byte key and get a keyCode */
+    ret = wc_HWPUF_GenerateKey(&hwpuf, 1, 32, keyCode32, sizeof(keyCode32));
+    if (ret != 0)
+        return WC_TEST_RET_ENC_EC(ret);
+    /* get key from keyCode */
+    ret = wc_HWPUF_GetKey(&hwpuf, keyCode32, sizeof(keyCode32), key32_1, sizeof(key32_1));
+    if (ret != 0)
+        return WC_TEST_RET_ENC_EC(ret);
+
+    /* ---- Test 5: restart and derive the same 3 keys ---- */
```
dgarske, line 23407: **⚪ [Info] hwpuf_test step numbering skips 'Test 6'**

The test step comments jump from `Test 5` directly to `Test 7` (no `Test 6`). Purely cosmetic, but it suggests a step was removed and may confuse a reader looking for a missing case.

**Fix:** Renumber the step comments to be contiguous.

dgarske (commented):
## Skoll Code Review

**Findings:** 12 total — 12 posted, 0 skipped
**12 finding(s) posted as inline comments** (see file-level comments below)

### Posted findings
- [High] **wc_HWPUF_Register reads uninitialized hwpuf-registered before zeroing** — `wolfcrypt/src/hwpuf.c:42-65`
- [Medium] **Unused static function getACFromPFR (dead code, uses memset)** — `wolfcrypt/src/port/nxp/hwpuf_port.c:54-65`
- [Medium] **nxp_hwpuf_SetKey has multiple unused parameters** — `wolfcrypt/src/port/nxp/hwpuf_port.c:185-195`
- [Medium] **Callback parameter ctx shadows file-scope global ctx and is unused** — `wolfcrypt/src/port/nxp/hwpuf_port.c:258`
- [Medium] **RNG hardware register read without ensuring RNG is initialized** — `wolfcrypt/src/port/nxp/hwpuf_port.c:96`
- [Low] **Signed/unsigned comparison in hwpuf_test loop** — `wolfcrypt/test/test.c:23443-23444`
- [Low] **WOLFSSL_NXP_HWPUF does not imply WOLFSSL_HWPUF** — `wolfssl/wolfcrypt/settings.h:2207-2209`
- [Low] **Inconsistent key-index bound: literal 16 vs kPUF_KeyIndexMax** — `wolfcrypt/src/port/nxp/hwpuf_port.c:74-79`
- [Low] **wc_CryptoInfo.hwpuf.hwpuf typed as void* loses type safety** — `wolfssl/wolfcrypt/cryptocb.h:582`
- [Low] **HWPUF_DEINIT_E error code defined but never returned** — `wolfssl/wolfcrypt/error-crypt.h:332`
- [Low] **No test coverage for SetKey path or double-register** — `wolfcrypt/test/test.c:23341-23502`
- [Info] **hwpuf_test step numbering skips 'Test 6'** — `wolfcrypt/test/test.c:23407-23434`

dgarske:

wolfcrypt/test/test.c
```diff
@@ -23324,6 +23337,176 @@ WOLFSSL_TEST_SUBROUTINE wc_test_ret_t puf_test(void)
...
+        return WC_TEST_RET_ENC_EC(ret);
+    { /* key1 should be zeroed */
+        word32 idx;
+        for (idx = 0; idx < sizeof(key32_2); ++idx) {
+            if (key32_2[idx])
+                return WC_TEST_RET_ENC_NC;
+        }
+    }
+
+    /* ---- Test 7: Bad argument checks ---- */
```
dgarske, line 23450: **🟠 [Medium] Bad-arg / error-path test coverage gaps for new HWPUF API** · _test_

Test 7 exercises NULL/bad args for `wc_HWPUF_Init`, `Deinit`, `Enroll`, `Zeroize`, and a couple of GenerateKey/GetKey cases, but several new public entry points have no NULL/error-path coverage: `wc_HWPUF_Start(NULL,...)`, `wc_HWPUF_GetKey(NULL,...)`, `wc_HWPUF_GenerateKey(NULL,...)`, `wc_HWPUF_Unregister(NULL)`, and `wc_HWPUF_Register(NULL,...)`. The state-machine error returns are also untested: `wc_HWPUF_Start`/`Enroll` before `Init` (expect `HWPUF_INIT_E`), `wc_HWPUF_GenerateKey`/`GetKey` before `Start` (expect `HWPUF_START_E`), and `Enroll` when already enrolled (expect `HWPUF_ENROLL_E`). These are the gating branches the wrapper adds on top of the port, so they are worth pinning.

**Fix:** Extend hwpuf_test to cover the remaining NULL-argument paths and the wrapper's state-machine error returns (Start before Init -> HWPUF_INIT_E, GetKey before Start -> HWPUF_START_E, double Enroll -> HWPUF_ENROLL_E).

wolfcrypt/test/test.c
```diff
@@ -23324,6 +23337,176 @@ WOLFSSL_TEST_SUBROUTINE wc_test_ret_t puf_test(void)
 }
 #endif /* WOLFSSL_PUF */
 
+#ifdef WOLFSSL_HWPUF
```
dgarske, line 23340: **🔵 [Low] Large stack frame in hwpuf_test on embedded target (no WOLFSSL_SMALL_STACK)** · _Resource Leak_

`activationCode[HWPUF_ACTIVATION_CODE_SIZE]` is 1192 bytes, and together with the three 52-byte key-code buffers and six key buffers the function consumes roughly 1.4-1.5 KB of stack. This test only builds/runs on the LPC55S69 (`WOLFSSL_HWPUF` + `WOLFSSL_NXP_HWPUF`), an embedded MCU with a small stack, so it risks stack exhaustion on the very target it is meant to run on. wolfSSL convention for stack buffers larger than ~100 bytes is the `WOLFSSL_SMALL_STACK` (heap) pattern. (Both reviewers flagged this; review rated it Low, review-security rated it Info -- stricter Low retained.)

**Fix:** Allocate `activationCode` (and optionally the key/keycode buffers) from the heap under the `WOLFSSL_SMALL_STACK` pattern, or document the stack requirement, to avoid a ~1.5KB stack frame on the MCU.

wolfcrypt/src/port/nxp/hwpuf_port.c
```diff
@@ -0,0 +1,335 @@
...
+    if (keyidx == kPUF_KeyIndex_00) {
+        /* keyslot 0 means send to aes engine */
+        ret = PUF_GetHwKey(PUF, keyCode, keyCodeSz, kPUF_KeySlot0,
+                           ctx.keyMask);
+        if (ret != kStatus_Success)
+            return HWPUF_GET_KEY_E;
+        if (key)
+            XMEMSET(key, 0, keySz); /* no key to return, zero out */
+    }
+    else {
```
dgarske, line 221: **🔵 [Low] Retrieved key buffer not zeroized on PUF_GetKey failure path** · _Zeroization_

In the non-hw-bus branch of `nxp_hwpuf_GetKey`, `PUF_GetKey()` writes derived key material into the caller's `key` buffer. If the vendor call returns a non-success status after partially populating `key`, the function returns `HWPUF_GET_KEY_E` without zeroizing `key`. Sensitive key material may be left in the output buffer on an error return. The wolfSSL convention (and the project rules) is to `ForceZero()` secret buffers on every return path, including error paths. The success path is fine; the gap is only on the error path.

**Fix:** On the `PUF_GetKey() != kStatus_Success` path, call `ForceZero(key, keySz)` (when `key != NULL`) before returning `HWPUF_GET_KEY_E`.

wolfcrypt/src/port/nxp/hwpuf_port.c
```diff
@@ -0,0 +1,335 @@
...
+    if (ret == kStatus_EnrollNotAllowed) {
+        /* power cycle and try again */
+        (void)PUF_PowerCycle(PUF, &conf);
+        ret = PUF_Enroll(PUF, actCode, actCodeSz);
+    }
+    if (ret != kStatus_Success) {
+        return HWPUF_ENROLL_E;
+    }
+
+    /* wipe ctx if enroll succeeded (re-enroll will render ctx moot) */
```
dgarske, line 133: **🔵 [Low] ctx keyMask wiped with XMEMSET instead of ForceZero in Enroll** · _Zeroization_

`nxp_hwpuf_Enroll` clears the file-static `ctx` (which holds `keyMask`, the random mask used by `PUF_GetHwKey` to obscure key material on the HW bus) using `XMEMSET(&ctx, 0, sizeof(ctx))`. The companion `nxp_hwpuf_Zeroize` correctly uses `ForceZero(&ctx, sizeof(ctx))` for the same structure. A plain `XMEMSET` can be elided by the optimizer for data the compiler believes is dead, so a sensitive mask value could persist. This is an inconsistency with the rest of the file and with the project rule to zero secrets with `ForceZero()`.

**Fix:** Use `ForceZero(&ctx, sizeof(ctx));` here for consistency with `nxp_hwpuf_Zeroize` and to guarantee the wipe is not optimized away.

wolfcrypt/src/port/nxp/hwpuf_port.c
```diff
@@ -0,0 +1,335 @@
...
+        return BAD_FUNC_ARG;
+
+    ret = keyCodeCheck(keyCode, &keytype, &keyidx, &keysize);
+    if (ret != kStatus_Success)
+        return BAD_FUNC_ARG;
+
+    kcSz = PUF_GET_KEY_CODE_SIZE_FOR_KEY_SIZE(keysize);
+    if (kcSz != keyCodeSz)
+        return BAD_FUNC_ARG;
+    if (keyidx != kPUF_KeyIndex_00 && (key == NULL || keysize != keySz))
```
dgarske, line 208: **🔵 [Low] hw-bus GetKey path zeroes caller buffer using unvalidated keySz** · _Logic_

For the normal (non-hw-bus) path, `nxp_hwpuf_GetKey` validates `key != NULL && keysize == keySz` before writing, where `keysize` is derived from the validated key code. For the hw-bus path (`keyidx == kPUF_KeyIndex_00`), that validation is intentionally skipped, but the code then executes `XMEMSET(key, 0, keySz)` using the caller-supplied `keySz` without bounding it against the derived `keysize`. If a caller passes a `key` buffer smaller than the `keySz` it reports (a caller-side mistake, not network input), this is an out-of-bounds write. The non-hw path enforces an equality check that prevents this; the hw path has no equivalent guard, an asymmetry worth closing for robustness.

**Fix:** For the hw-bus path, validate `keySz` against the derived `keysize` (e.g. require `keySz == keysize` when `key != NULL`) before the `XMEMSET`, matching the guarantee the non-hw-bus path provides.

wolfcrypt/src/cryptocb.c
```diff
@@ -346,6 +369,12 @@ void wc_CryptoCb_InfoString(wc_CryptoInfo* info)
         printf("Crypto CB: %s %s (%d)\n", GetAlgoTypeStr(info->algo_type),
                GetKdfTypeStr(info->kdf.type), info->kdf.type);
     }
+#endif
+#ifdef WOLFSSL_HWPUF
+    else if (info->algo_type == WC_ALGO_TYPE_HWPUF) {
```
dgarske, line 374: **🔵 [Low] GetHwpufTypeStr may return NULL passed to printf %s** · _Logic_

`GetHwpufTypeStr()` returns `NULL` for any `type` not in the handled set (notably `WC_HWPUF_TYPE_NONE == 0`). In `wc_CryptoCb_InfoString` the result is passed directly to `printf("... %s ...", GetHwpufTypeStr(info->hwpuf.type), ...)`. Passing `NULL` to `%s` is undefined behavior (some libc print "(null)", others crash). This is only reachable under `DEBUG_CRYPTOCB` with an unset/invalid hwpuf type. It mirrors the pre-existing `GetKdfTypeStr` pattern, so it is low priority, but the new code re-introduces the same fragility.

**Fix:** Have `GetHwpufTypeStr` return a non-NULL fallback (e.g. "UNKNOWN") for unhandled types, or guard the `printf` against a NULL return.

wolfssl/wolfcrypt/error-crypt.h
```diff
@@ -328,8 +328,16 @@ enum wolfCrypt_ErrorCodes {
...
```
dgarske, line 332: **⚪ [Info] Missing space before '=' in HWPUF_GENERATE_KEY_E enumerator** · _style_

`HWPUF_GENERATE_KEY_E= -1023` is missing the space before `=` that every other enumerator in the block uses, and it breaks the value-column alignment of the surrounding `HWPUF_*_E` entries.

**Fix:** Add the missing space and align the `=` with the neighboring enumerators: `HWPUF_GENERATE_KEY_E = -1023,`.

wolfcrypt/src/port/nxp/hwpuf_port.c
```diff
@@ -0,0 +1,335 @@
...
+    word32 kcSz;
+
+    WOLFSSL_ENTER("nxp_hwpuf_GetKey");
+
+    if (hwpuf == NULL)
+        return BAD_FUNC_ARG;
+    if (keyCode == NULL || keyCodeSz < PUF_MIN_KEY_CODE_SIZE)
+        return BAD_FUNC_ARG;
+
+    ret = keyCodeCheck(keyCode, &keytype, &keyidx, &keysize);
```
dgarske, line 201: **⚪ [Info] keyCodeCheck custom return compared against NXP kStatus_Success** · _style_

`keyCodeCheck` returns its own codes 0/1/2/3, but the caller checks `if (ret != kStatus_Success)`, mixing a local convention with the NXP status enum. It happens to work only because `kStatus_Success` is `0`. This is misleading to readers who would expect `kStatus_Success` only to be paired with NXP SDK calls.

**Fix:** Compare keyCodeCheck's result against literal 0 (or give it named return codes) rather than the NXP kStatus_Success enum.

wolfcrypt/test/test.c
```diff
@@ -23324,6 +23337,176 @@ WOLFSSL_TEST_SUBROUTINE wc_test_ret_t puf_test(void)
...
+
+    /* ---- Test 6: generate a key and send directly to hw bus ---- */
+    ret = wc_HWPUF_GenerateKey(&hwpuf, 0, 32, keyCode32, sizeof(keyCode32));
+    if (ret != 0)
+        return WC_TEST_RET_ENC_EC(ret);
+    /* get key from keyCode */
+    ret = wc_HWPUF_GetKey(&hwpuf, keyCode32, sizeof(keyCode32), key32_2, sizeof(key32_2));
+    if (ret != 0)
+        return WC_TEST_RET_ENC_EC(ret);
+    { /* key1 should be zeroed */
```
dgarske, line 23442: **⚪ [Info] New brace-scope created in test just to declare a loop index** · _convention_

A `{ word32 idx; for (...) ... }` block is introduced solely to declare `idx`. wolfSSL C89 style declares variables at the top of the function or an existing block rather than opening a new scope just to declare. `idx` could be declared with the other locals at the top of `hwpuf_test`.

**Fix:** Move the loop-index declaration to the function's top-of-block declarations to match wolfSSL C89 style and drop the extra brace scope.

wolfcrypt/test/test.c
```diff
@@ -23324,6 +23337,176 @@ WOLFSSL_TEST_SUBROUTINE wc_test_ret_t puf_test(void)
...
+    /* null hwpuf */
+    if (wc_HWPUF_Init(NULL) != WC_NO_ERR_TRACE(BAD_FUNC_ARG))
+        return WC_TEST_RET_ENC_NC;
+    if (wc_HWPUF_Deinit(NULL) != WC_NO_ERR_TRACE(BAD_FUNC_ARG))
+        return WC_TEST_RET_ENC_NC;
+    if (wc_HWPUF_Enroll(NULL, NULL, 0) != WC_NO_ERR_TRACE(BAD_FUNC_ARG))
+        return WC_TEST_RET_ENC_NC;
+    if (wc_HWPUF_Zeroize(NULL) != WC_NO_ERR_TRACE(BAD_FUNC_ARG))
+        return WC_TEST_RET_ENC_NC;
+    /* out of bounds key index */
```
dgarske, line 23460: **⚪ [Info] Misleading test comment references wrong variable** · _Logic_

In Test 6 (hw-bus key), the comment says `/* key1 should be zeroed */` but the loop actually validates `key32_2`. There is no `key1` variable. Cosmetic only; no behavioral impact.

**Fix:** Update the comment to reference `key32_2` (the buffer actually checked).

wolfssl/wolfcrypt/cryptocb.h
```diff
@@ -575,6 +578,35 @@ typedef struct wc_CryptoInfo {
...
+        wc_HWPUF*   hwpuf;      /* wc_HWPUF* context */
+        int         type;       /* enum wc_HwpufType - discriminator */
+        const void* ctx;        /* read-only caller context */
```
dgarske, line 585: **⚪ [Info] wc_CryptoInfo.hwpuf.ctx field declared but never populated** · _style_

The hwpuf cryptoinfo struct declares `const void* ctx; /* read-only caller context */`, but none of the `wc_CryptoCb_Hwpuf*` setters in cryptocb.c ever assign `cryptoInfo.hwpuf.ctx`, and the NXP port reads `devCtx` (the registered callback ctx), not `info->hwpuf.ctx`. The field is currently dead. Either wire it up (set it from a caller-supplied context) or drop it to avoid implying a contract that is not honored.

**Fix:** Drop the unused `ctx` member, or populate it in the wc_CryptoCb_Hwpuf* wrappers if caller context is intended to be passed through, so the struct reflects real behavior.

dgarske (commented):
## Skoll Multi-Scan Review

**Findings:** 14 total — 14 posted, 0 skipped
**11 finding(s) posted as inline comments** (see file-level comments below)
**3 finding(s) not tied to a diff line** (full detail below)

### Posted findings
- [Medium] [review] **Bad-arg / error-path test coverage gaps for new HWPUF API** — `wolfcrypt/test/test.c:23450-23482`
- [Low] [review+review-security] **Large stack frame in hwpuf_test on embedded target (no WOLFSSL_SMALL_STACK)** — `wolfcrypt/test/test.c:23340-23360`
- [Low] [review-security] **Retrieved key buffer not zeroized on PUF_GetKey failure path** — `wolfcrypt/src/port/nxp/hwpuf_port.c:221-225`
- [Low] [review-security] **ctx keyMask wiped with XMEMSET instead of ForceZero in Enroll** — `wolfcrypt/src/port/nxp/hwpuf_port.c:133-134`
- [Low] [review-security] **hw-bus GetKey path zeroes caller buffer using unvalidated keySz** — `wolfcrypt/src/port/nxp/hwpuf_port.c:208-220`
- [Low] [review-security] **GetHwpufTypeStr may return NULL passed to printf %s** — `wolfcrypt/src/cryptocb.c:250-271,373-378`
- [Info] [review] **Missing space before '=' in HWPUF_GENERATE_KEY_E enumerator** — `wolfssl/wolfcrypt/error-crypt.h:332`
- [Info] [review] **keyCodeCheck custom return compared against NXP kStatus_Success** — `wolfcrypt/src/port/nxp/hwpuf_port.c:63-78,201-203`
- [Info] [review] **New brace-scope created in test just to declare a loop index** — `wolfcrypt/test/test.c:23442-23448`
- [Info] [review-security] **Misleading test comment references wrong variable** — `wolfcrypt/test/test.c:23460-23467`
- [Info] [review] **wc_CryptoInfo.hwpuf.ctx field declared but never populated** — `wolfssl/wolfcrypt/cryptocb.h:585`

#### Singleton global state conflicts with per-context wc_HWPUF* API
**File:** `wolfcrypt/src/hwpuf.c:41; wolfcrypt/src/port/nxp/hwpuf_port.c:56-57,80`
**Function:** `wc_HWPUF_Register / wc_HWPUF_Unregister / nxp_hwpuf_*`
**Severity:** Medium
**Category:** api

The public API is shaped around a per-instance `wc_HWPUF*` context, but registration/device state is held in file-scope statics: `hwpuf_registered` in `hwpuf.c`, and `ctx`, `conf`, `nxp_rng_initialized` in the NXP port. This makes the implementation effectively a singleton. Two consequences: (1) `wc_HWPUF_Register` on a second context returns `HWPUF_REGISTER_E` even though it is a distinct object; (2) `wc_HWPUF_Unregister(ctxB)` while `ctxA` is the registered context still passes the `if (!hwpuf_registered)` guard (the global is 1) and calls `nxp_hwpuf_UnregisterDevice(ctxB)`, unregistering `ctxB->devId` -- which for a never-registered/zeroed context is `0`/garbage, potentially tearing down the wrong device. The PUF peripheral is genuinely a singleton so static `ctx`/`conf` is defensible, but the pointer-based API invites misuse.

**Recommendation:** Either document the singleton restriction explicitly in the public header, or guard Unregister against tearing down a device for a context that was never the registered one (e.g. match a stored devId/pointer before unregistering).

_Referenced code: `wolfcrypt/src/hwpuf.c:41; wolfcrypt/src/port/nxp/hwpuf_port.c:56-57,80` (8 lines)_

---

#### Public key-code-size macro hardcoded, decoupled from NXP SDK formula
**File:** `wolfssl/wolfcrypt/hwpuf.h:37-38; wolfcrypt/src/port/nxp/hwpuf_port.c:175,205`
**Function:** `HWPUF_KEY_SIZE_TO_KEY_CODE_SIZE / nxp_hwpuf_GenerateKey / nxp_hwpuf_GetKey`
**Severity:** Low
**Category:** question

The public macro `HWPUF_KEY_SIZE_TO_KEY_CODE_SIZE(keysz)` always expands to the literal `52` (ignoring its argument), and callers size their key-code buffers from it. The port, however, validates buffer sizes against the NXP SDK macro `PUF_GET_KEY_CODE_SIZE_FOR_KEY_SIZE(keySz)` (`kcSz != keyCodeSz` -> BAD_FUNC_ARG). If a future NXP SDK changes its key-code size formula, the public macro and the port's validation will silently diverge and every GenerateKey/GetKey call will fail with BAD_FUNC_ARG. Deriving the public macro from the SDK macro (or adding a compile-time assertion that they agree) would make the coupling explicit.

**Recommendation:** Define HWPUF_KEY_SIZE_TO_KEY_CODE_SIZE in terms of PUF_GET_KEY_CODE_SIZE_FOR_KEY_SIZE where possible, or add a static assert that the two agree for 16/24/32-byte keys so they cannot drift apart.

_Referenced code: `wolfssl/wolfcrypt/hwpuf.h:37-38; wolfcrypt/src/port/nxp/hwpuf_port.c:175,205` (5 lines)_

---

#### Function-like macros missing parentheses around parameter
**File:** `wolfssl/wolfcrypt/port/nxp/hwpuf_port.h:32-33; wolfssl/wolfcrypt/hwpuf.h:38`
**Function:** `HWPUF_KEY_SIZE_IS_VALID / HWPUF_KEY_SIZE_TO_KEY_CODE_SIZE`
**Severity:** Info
**Category:** style

`HWPUF_KEY_SIZE_IS_VALID(keysz)` expands `keysz` without wrapping it in parentheses (`(keysz == 16 || keysz == 24 || keysz == 32)`); a caller passing an expression rather than a plain variable could hit operator-precedence surprises. (Current callers pass simple vars so it is harmless today.) Separately, `HWPUF_KEY_SIZE_TO_KEY_CODE_SIZE(keysz)` discards its argument entirely (always 52), which is intentional but worth a comment so readers do not assume it scales with key size.

**Recommendation:** Parenthesize the macro parameter in HWPUF_KEY_SIZE_IS_VALID: `((keysz) == 16 || (keysz) == 24 || (keysz) == 32)`; keep a comment noting the key-code size is fixed at 52.

_Referenced code: `wolfssl/wolfcrypt/port/nxp/hwpuf_port.h:32-33; wolfssl/wolfcrypt/hwpuf.h:38-41` (4 lines)_

---

PR #10594, opened 2026-06-04
Fix a few TLS compliance issues

kareem-wolfssl:
# Description

Partially fixes zd#21873

# Testing

Built in tests

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


src/internal.c
```diff
@@ -31648,16 +31648,43 @@ static void MakePSKPreMasterSecret(Arrays* arrays, byte use_psk_key)
...
+             * full handshake (avoids leaking a stale ticket and saves a
+             * round-trip). Expiry is measured against ssl->session->timeout
+             * (the session's own lifetime) so this stays consistent with
+             * wolfSSL_SetSession(), which gates resumption on the same field;
+             * keying off ssl->timeout instead could contradict a decision
+             * SetSession() already made when the two values differ.
+             * If bornOn is 0 or the secret callback is set, it is assumed that
+             * the session is being externally managed and this check is
+             * skipped.  This is needed for hostap. */
+            if (ssl->session->bornOn != 0 &&
```
dgarske: **🟠 [Medium] Expired-ticket branch drops SessionTicket extension entirely; client may no longer obtain a fresh ticket**

When the stored ticket is detected as expired the code sets `ssl->options.resuming = 0` and skips `TLSX_UseSessionTicket()` entirely, so no SessionTicket extension is emitted in the ClientHello. Previously the (stale) ticket extension was sent, which still advertised ticket support and prompted the server to issue a NEW ticket during the resulting full handshake. The client-side empty/real SessionTicket extension is added in only two places: `wolfSSL_UseSessionTicket()`/`wolfSSL_CTX_UseSessionTicket()` (ssl_api_ext.c) and this resumption block. In a configuration where the extension is added only via this path (e.g. an app that loaded a ticket-bearing session with `wolfSSL_set_session()` without separately enabling the ticket extension on the new object/context), the expired branch now yields a ClientHello with no ticket extension at all, so the server will not issue a replacement ticket and the client permanently loses ticket-based resumption. Most apps enable tickets at the CTX level, where the empty extension persists and this does not bite, hence Medium/conditional. Separately, the comment's claim that suppressing the ticket 'saves a round-trip' is inaccurate: whether or not an expired ticket is sent, the server falls back to the same full handshake with the same number of round-trips.

**Fix:** Consider emitting an empty SessionTicket extension (mirroring the server-side blank-ticket behavior in tls.c TLSX_SessionTicket_Parse) in the expired branch so the client still requests a new ticket while not sending the stale one. Also correct the 'saves a round-trip' wording in the comment.

src/ssl.c
```diff
@@ -4068,6 +4068,17 @@ int wolfSSL_export_keying_material(WOLFSSL *ssl,
         return WOLFSSL_FAILURE;
     }
 
+    /* RFC 8446 Section 7.5 / RFC 5705: keying-material exporters derive from
+     * exporter_master_secret, which exists only after the handshake is
+     * complete. Refuse the export until the handshake has completed so that
+     * a premature call cannot derive material from an uninitialised
+     * exporterSecret buffer. */
+    if (ssl->options.handShakeDone == 0 ||
```
dgarske: **🔵 [Low] Redundant dual condition in keying-material handshake-complete gate**

The gate checks both `handShakeDone == 0` and `handShakeState != HANDSHAKE_DONE`. These two flags are set together at every handshake-completion site (e.g. internal.c:8258-8259, 18136-18137, 18155-18156; tls13.c:11966, 12223) and `handShakeState` is never advanced past HANDSHAKE_DONE (the post-DONE state SERVER_FINISHED_ACKED lives in `serverState`, not `handShakeState`). The established idiom elsewhere (ssl.c:1025, internal.c:22140) uses `handShakeDone == 0` alone. The double check is harmless and slightly more defensive, but inconsistent with the rest of the codebase. This is informational only — not a correctness problem.

**Fix:** Optionally simplify to the single `handShakeDone == 0` check to match existing usage, or keep as-is if the extra belt-and-suspenders is intentional.

dgarske (commented):
## Skoll Code Review

### Posted findings
- [Medium] **Expired-ticket branch drops SessionTicket extension entirely; client may no longer obtain a fresh ticket** — `src/internal.c:31665-31687`
- [Medium] **No tests added for the three new compliance behaviors** — `tests/api.c`
- [Low] **Redundant dual condition in keying-material handshake-complete gate** — `src/ssl.c:4076-4080`

### Skipped findings
- [Low] ~~Preprocessor directives indented to code column instead of repo convention~~

dgarske (changes requested):
See skoll review https://github.com/wolfSSL/wolfssl/pull/10594#pullrequestreview-4480784807

kareem-wolfssl:
Retest this please

src/internal.c
```diff
@@ -3731,6 +3731,15 @@ static word16 InitSuites_Tls13(Suites* suites, word16 idx, int tls1_3,
...
 #ifdef HAVE_NULL_CIPHER
+    /* RFC 9150 integrity-only (zero-confidentiality) TLS 1.3 suites.
+     * These provide authentication and integrity but no confidentiality, so
+     * they are NOT advertised in the default cipher preference list. A caller
+     * that genuinely needs them (e.g. constrained/IoT deployments) must opt in
+     * explicitly with a cipher list, for example:
+     *     wolfSSL_set_cipher_list(ssl, "TLS13-SHA256-SHA256");
+     * Define WOLFSSL_TLS13_NULL_CIPHER_IN_DEFAULT to restore the legacy
+     * behaviour of including them in the default list. */
+    #ifdef WOLFSSL_TLS13_NULL_CIPHER_IN_DEFAULT
```
dgarske: **🟠 [Medium] New build macro not added to the src/internal.c macro documentation block**

`WOLFSSL_TLS13_NULL_CIPHER_IN_DEFAULT` was added to `.wolfssl_known_macro_extras` but not to the macro reference block at the top of src/internal.c, which documents every comparable knob with a one-line description and a default - including a "TLS 1.3 Internals:" section (lines 101-109) holding `WOLFSSL_TLS13_IGNORE_PT_ALERT_ON_ENC`, `WOLFSSL_TLS13_IGNORE_AEAD_LIMITS`, etc., and `WOLFSSL_NO_TICKET_EXPIRE` at line 99. That block is the discoverable place a user looks for build-time behaviour switches. The same block would also be the right place to note that `WOLFSSL_NO_TICKET_EXPIRE` now gates a client-side check too (its description and the HandleTlsResumption doxygen at src/internal.c:39352-39359 both describe only the server-side TLS 1.2 check). There is likewise no configure.ac option for the new macro, unlike `--enable-nullcipher` itself.

**Fix:** Add the macro to the "TLS 1.3 Internals" section of the src/internal.c macro block, and extend the WOLFSSL_NO_TICKET_EXPIRE line / HandleTlsResumption doxygen to mention that it now also disables the client-side ClientHello check.

src/internal.c
```diff
@@ -3731,6 +3731,15 @@ static word16 InitSuites_Tls13(Suites* suites, word16 idx, int tls1_3,
 #endif
 
 #ifdef HAVE_NULL_CIPHER
```
dgarske, line 4004: **🟠 [Medium] "eNULL" directive now behaves asymmetrically between TLS 1.2 and TLS 1.3 NULL suites**

`ParseCipherList()` still sets `haveNull = 1` for the "eNULL" keyword and passes it to `InitSuites()` (src/internal.c:30406-30407, 30703). After this change that mask still generates the TLS 1.2 NULL-cipher suites (e.g. TLS_ECDHE_ECDSA_WITH_NULL_SHA, src/internal.c:4368-4392) but no longer generates the RFC 9150 TLS 1.3 suites - so an application that explicitly asks for "ALL:eNULL" gets zero-confidentiality suites for TLS 1.2 and not for TLS 1.3. The new tests/api.c block encodes exactly this asymmetry. Meanwhile the exclusion side stays symmetric: `CipherSuiteExcluded()` (src/internal.c:30126-30133) still drops the TLS 1.3 suites for "!eNULL". An explicit opt-in keyword doing different things for different protocol versions is easy to misread as a bug.

**Fix:** Decide whether the gate is about "defaults" or about "NULL suites in general" and make "eNULL" consistent across versions; if the asymmetry is intentional, call it out in the release notes since it is a user-visible cipher-string semantics change.

src/internal.c
```diff
@@ -32682,16 +32692,41 @@ static void MakePSKPreMasterSecret(Arrays* arrays, byte use_psk_key)
...
             SessionTicket* ticket;
 
-            ticket = TLSX_SessionTicket_Create(0, ssl->session->ticket,
+#if !defined(WOLFSSL_NO_TICKET_EXPIRE) && !defined(NO_ASN_TIME)
+            /* RFC 5077 Section 3.3 / RFC 8446 Section 4.6.1: a client SHOULD
+             * NOT use a ticket whose lifetime has expired. Drop the expired
+             * ticket and fall back to a full handshake. Skip the check when
+             * bornOn is 0 or a secret callback is set (session is managed
+             * externally, e.g. hostap). */
+            if (ssl->session->bornOn != 0 &&
```
dgarske: **🟠 [Medium] Expired ticket is not deleted from the session, contrary to the cited RFC**

The new branch sets `ssl->options.resuming = 0` and sends an empty ticket extension, but leaves `ssl->session->ticket` / `ssl->session->ticketLen` populated and the expired session in the cache. RFC 5077 section 3.3 - which the comment cites - says the client "SHOULD delete the ticket and associated state when the time expires", and the server-side equivalent in this file does evict the entry (see the `EvictSessionFromCache()` call on the timed-out path in src/ssl_sess.c:1387-1400). Keeping the stale ticket means the secret material stays resident, and every subsequent connect on the same session object re-runs the expiry check instead of simply having no ticket. It also leaves `ssl->session->ticketLen > 0` visible to callers of `wolfSSL_get_session()` / `wolfSSL_SESSION_get_ticket*` for a ticket wolfSSL has already decided is unusable.

**Fix:** Drop the stale ticket bytes (zeroizing them) when the expiry check fires, so the client state matches the RFC's "delete the ticket and associated state".

dgarske: **🔵 [Low] Expiry comparison differs from the server-side check and treats timeout 0 as immediately expired**

Two small consistency points. (1) The new check uses `>=` while the server-side check 6700 lines below uses `>` (`if (LowResTimer() > session->bornOn + ssl->timeout)`, src/internal.c:39447) for the same notion of expiry; the session-cache helpers use `>=` (src/ssl_sess.c:1387, 1597), so the new code matches the majority, but the two in-file checks now disagree by one second. (2) The code defensively guards `bornOn != 0` but not `timeout != 0`; a TLS 1.2 server may legally send `ticket_lifetime_hint = 0` ("lifetime unspecified"), which `DoSessionTicket()` copies straight into `ssl->timeout` (src/internal.c:36757) and hence into `session->timeout`, making `bornOn + 0 <= now` always true. In practice `wolfSSL_SetSession()` already rejects such a session at src/ssl_sess.c:1597 so the new branch is not the first thing to break, but a `timeout != 0` guard alongside the existing `bornOn != 0` guard would be consistent with the defensive intent.

**Fix:** Add the `timeout != 0` guard for symmetry with the `bornOn != 0` guard, and consider aligning the `>` / `>=` choice with src/internal.c:39447.

src/internal.c
```diff
@@ -32682,16 +32692,41 @@ static void MakePSKPreMasterSecret(Arrays* arrays, byte use_psk_key)
...
+                 * client still requests a new ticket from the server without
+                 * sending the stale one. */
+                ret = TLSX_UseSessionTicket(&ssl->extensions, NULL, ssl->heap);
+                if (ret != WOLFSSL_SUCCESS)
+                    return ret;
+            }
+            else
+#endif
+            {
+                ticket = TLSX_SessionTicket_Create(0, ssl->session->ticket,
```
dgarske, line 34309: **🔵 [Low] Continuation line left misaligned and over 80 columns after re-indent**

The `TLSX_SessionTicket_Create()` call was indented by four spaces when it moved into the new else-block, but its continuation line was not, so the arguments no longer line up under the opening parenthesis, and the line is now 81 columns (the only >80-column line introduced in the changed regions of src/internal.c, src/ssl.c and tests/api.c).

**Fix:** Re-align the continuation under the opening parenthesis and keep the line within 80 columns.

dgarske (commented):
## Skoll Code Review

**Findings:** 13 total — 6 posted, 7 skipped

### Posted findings
- [Medium] **New build macro not added to the src/internal.c macro documentation block** — `src/internal.c:3742`
- [Medium] **"eNULL" directive now behaves asymmetrically between TLS 1.2 and TLS 1.3 NULL suites** — `src/internal.c:3733-3757`
- [Medium] **Expired ticket is not deleted from the session, contrary to the cited RFC** — `src/internal.c:32701-32715`
- [Low] **Expiry comparison differs from the server-side check and treats timeout 0 as immediately expired** — `src/internal.c:32701-32706`
- [Low] **Continuation line left misaligned and over 80 columns after re-indent** — `src/internal.c:32719-32720`

#### Removal of integrity-only .conf cases drops end-to-end coverage that would still pass
**File:** `tests/test-tls13.conf:213-231, tests/test-dtls13.conf:270-292`
**Function:** `n/a`
**Severity:** Medium

Eight entries (four TLS 1.3, four DTLS 1.3) that ran real client/server processes over TLS13-SHA256-SHA256 / TLS13-SHA384-SHA384 were deleted. These do not appear to be broken by the source change: `IsValidCipherSuite()` in tests/suites.c:128-169 validates a conf `-l` entry by calling `wolfSSL_CTX_set_cipher_list()` with the literal name, and the explicit-name path in `ParseCipherList()` is untouched by the InitSuites_Tls13 gating - which is precisely what the new tests/api.c:2548-2552 assertion (`"TLS13-SHA256-SHA256"` -> suite present) proves. Both peers pass `-l`, so the handshake should still negotiate. After the removal the documented opt-in path is exercised only by in-process memio unit tests (tests/api/test_tls_ext.c:420, tests/api/test_dtls13.c:131,783); the cross-process client/server coverage, including the DTLS 1.3 record path, is gone.

**Recommendation:** Restore the conf entries (they exercise the opt-in path the PR says remains supported), or state in the PR why they can no longer pass.

_Referenced code: `tests/test-tls13.conf:213-231, tests/test-dtls13.conf:270-276` (7 lines)_

---

### Skipped findings
- [Medium] ~~Default TLS 1.3 suite list can become empty in NULL-cipher-only builds~~
- [Medium] ~~Exporter gate is stricter than the TLS 1.3 key schedule requires (0.5-RTT server window)~~
- [Medium] ~~New public-API precondition in wolfSSL_export_keying_material has no test~~
- [Medium] ~~New BUFFER_E - decode_error alert mapping is untested~~
- [Medium] ~~BUFFER_E also signals local send-side failures, so decode_error can blame the peer for our own error~~
- [Low] ~~Comment cites RFC 8446 4.6.1 in a code path unreachable for TLS 1.3~~
- [Low] ~~Test opens a block solely to declare a variable and cites a non-existent flag~~

tests/test-dtls13.conf
```diff
@@ -270,23 +270,3 @@
 -u
 -v 4
 -l TLS_AES_128_GCM_SHA256
-
```
dgarske, line 273: Can you explain why these were removed?

tests/test-tls13.conf
```diff
@@ -213,19 +213,3 @@
 # client TLSv1.3 Send Ticket explicitly
 -v 4
 -l TLS13-AES128-GCM-SHA256
-
```
dgarske, line 216: Can you explain why these were removed?

.wolfssl_known_macro_extras
```diff
@@ -1031,6 +1031,7 @@ WOLFSSL_TI_CURRTIME
 WOLFSSL_TLS13_DRAFT
 WOLFSSL_TLS13_IGNORE_AEAD_LIMITS
 WOLFSSL_TLS13_IGNORE_PT_ALERT_ON_ENC
+WOLFSSL_TLS13_NULL_CIPHER_IN_DEFAULT
```
dgarske, line 1116: Document this at top of tls13.c

dgarske (changes requested):
https://github.com/wolfSSL/wolfssl/pull/10594#pullrequestreview-4848988576

JacobBarthelmeh:
@kareem-wolfssl please go through and address comments. Assign to wolfSSL-bot and un-assign yourself when ready for review/merge.

src/internal.c
```diff
@@ -33870,16 +33891,52 @@ static void MakePSKPreMasterSecret(Arrays* arrays, byte use_psk_key)
...
+            #endif
+                LowResTimer() >=
+                    (ssl->session->bornOn + ssl->session->timeout)) {
+                WOLFSSL_MSG("Stored session ticket expired; full handshake");
+                ssl->options.resuming = 0;
+                /* RFC 5077 Section 3.3: delete the ticket and associated
+                 * state. */
+                ForceZero(ssl->session->ticket, ssl->session->ticketLen);
+                if (ssl->session->ticketLenAlloc > 0) {
+                    XFREE(ssl->session->ticket, NULL,
```
JacobBarthelmeh: Expired-ticket cleanup frees and zeroes ssl->session->ticket without HaveUniqueSessionObj(), corrupting a shared / application-owned WOLFSSL_SESSION. I'm not sure that deleting the session ticket here is the thing to do, maybe just setting not resuming so that it is cleaned up elsewhere.

src/internal.c
```diff
@@ -3932,19 +3937,35 @@ static word16 InitSuites_Tls13(Suites* suites, word16 idx, int tls1_3,
...
+     * These provide authentication and integrity but no confidentiality, so
+     * they are NOT advertised in the default cipher preference list. A caller
+     * that genuinely needs them (e.g. constrained/IoT deployments) must opt in
+     * explicitly with a cipher list, either by suite name:
+     *     wolfSSL_set_cipher_list(ssl, "TLS13-SHA256-SHA256");
+     * or with the "eNULL" keyword. Define
+     * WOLFSSL_TLS13_NULL_CIPHER_IN_DEFAULT to restore the legacy behaviour of
+     * including them in the default list. */
+    #ifndef WOLFSSL_TLS13_NULL_CIPHER_IN_DEFAULT
+    if (haveNull == SUITES_NULL_EXPLICIT)
```
JacobBarthelmeh, line 4015: I think the other InitSuites() location will need altered to account for this new flag now too.

For example this call flow would erase the NULL cipher suites set since the value '1' is being passed in with internal InitSuites() calls.

```
wolfSSL_CTX_set_cipher_list(ctx, "TLS13-SHA256-SHA256");
wolfSSL_CTX_set_options(ctx, WOLFSSL_OP_NO_SSLv3);
ssl = wolfSSL_new(ctx);
```

src/internal.c
JacobBarthelmeh: If keeping the XFREE here use the same heap hint that was used when allocating the session ticket.

src/ssl.c
```diff
@@ -3340,6 +3340,16 @@ int wolfSSL_export_keying_material(WOLFSSL *ssl,
         return WOLFSSL_FAILURE;
     }
 
+    /* RFC 8446 Section 7.5 / RFC 5705: keying-material exporters derive from
+     * exporter_master_secret, which exists only after the handshake is
+     * complete. Refuse the export until the handshake has completed so that
+     * a premature call cannot derive material from an uninitialised
+     * exporterSecret buffer. */
+    if (ssl->options.handShakeDone == 0) {
```
JacobBarthelmeh, line 3354: This gate here on handShakeDone flag might break keying-material export from the OpenSSL SSL_CB_HANDSHAKE_DONE info callback (callback fires before handShakeDone = 1)

tests/api.c
```diff
@@ -29588,8 +29632,15 @@ static int test_export_keying_material_cb(WOLFSSL_CTX *ctx, WOLFSSL *ssl)
 
 static int test_export_keying_material_ssl_cb(WOLFSSL* ssl)
 {
+    EXPECT_DECLS;
+    byte ekm[32] = {0};
+
     wolfSSL_KeepArrays(ssl);
-    return TEST_SUCCESS;
+    /* Export must be refused until the handshake has completed. */
+    ExpectIntEQ(wolfSSL_export_keying_material(ssl, ekm, sizeof(ekm),
```
JacobBarthelmeh, line 30996: I think this test is going to pass regardless of the changes to src/ssl.c. The expanded test coverage should test that the fail case is due to the export being called before handshake is complete not due to other errors that would be have been returned prior to these code changes.

JacobBarthelmeh (changes requested):

kareem-wolfssl:
Retest this please

kareem-wolfssl:
Retest this please

kareem-wolfssl:
Retest this please

kareem-wolfssl:
Retest this please

JacobBarthelmeh (dismissed):

src/ssl.c
```diff
@@ -3348,6 +3348,16 @@ int wolfSSL_export_keying_material(WOLFSSL *ssl,
...
```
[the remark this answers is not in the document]
kareem-wolfssl: At the point that handShakeDone is set, every secret, including the exporter master secret, is fully derived.  Material exported after a failed final flush is well-defined and correctly keyed.  There's no uninitialized-data issue, which is what this PR's gate exists to prevent.
"Invoke the handshake-done callback after setting it"  This is already being done.

JacobBarthelmeh (approved):

dgarske (approved):

PR #10595, opened 2026-06-04
f5381 enforce trailerField==1 in DecodeRsaPssParams

miyazakh:
# asn: enforce trailerField == trailerFieldBC(1) in DecodeRsaPssParams

## Summary

- **Bug fix**: Validate that the `trailerField` parameter in RSASSA-PSS-params
  equals `trailerFieldBC(1)` as mandated by RFC 8017 §A.2.3, in both the ASN
  template and non-template paths of `DecodeRsaPssParams`.
- **Test**: Extend `test_wc_DecodeRsaPssParams` with three trailerField-specific
  cases covering the valid value and two invalid values.

## Problem

RFC 8017 §A.2.3 defines:

```
TrailerField ::= INTEGER { trailerFieldBC(1) }
```

and states:

> The value of the `trailerField` field **SHALL** be `trailerFieldBC(1)`.

`DecodeRsaPssParams` parsed the `trailerField` element syntactically but
never validated its integer value, so any value was silently accepted.

### Non-template path (`!WOLFSSL_ASN_TEMPLATE`)

```c
/* before */
ret = GetInteger16Bit(params, &idx, sz);
if (ret > 0) {      /* accepts 1, 2, 3, …, 65535 */
    ret = 0;
}
else if (ret != 0) {
    WOLFSSL_MSG("DecodeRsaPssParams: fail at trailer_value");
}
/* ret == 0 (value 0) also falls through silently */
```

### Template path (`WOLFSSL_ASN_TEMPLATE`)

`RSAPSSPARAMSASN_IDX_TRAILERINT` appeared in the enum but
`GetASN_Int16Bit` was never called for it, so `GetASN_Items` validated
only the ASN.1 type tag while the integer value went completely unread.

### Impact

- Any positive trailerField value (2, 3, …) and value 0 were accepted without
  error in both paths.
- PSS signature verification always assumes the standard 0xbc trailer
  regardless; a non-conformant `trailerField` in an AlgorithmIdentifier is
  therefore accepted by the parser while an RFC-compliant peer would reject it.
- This is a **conformance / input-validation defect**, not a signature-forgery
  vulnerability.

## Fix

Both paths now reject `trailerField != 1` with `ASN_PARSE_E` when
`WOLFSSL_NO_ASN_STRICT` is not defined (the default). When
`WOLFSSL_NO_ASN_STRICT` is defined the previous permissive behaviour is
preserved for backward compatibility.

### Template path (`wolfcrypt/src/asn.c`)

```c
#ifndef WOLFSSL_NO_ASN_STRICT
    word16 trailerVal = 1;
#endif
    ...
#ifndef WOLFSSL_NO_ASN_STRICT
    GetASN_Int16Bit(&dataASN[RSAPSSPARAMSASN_IDX_TRAILERINT], &trailerVal);
#endif
    ret = GetASN_Items(...);
    ...
#ifndef WOLFSSL_NO_ASN_STRICT
    /* RFC 8017 A.2.3: trailerField SHALL be trailerFieldBC(1). */
    if ((ret == 0) && (dataASN[RSAPSSPARAMSASN_IDX_TRAILERINT].tag != 0)) {
        if (trailerVal != 1) {
            WOLFSSL_MSG("DecodeRsaPssParams: trailerField must be 1");
            ret = ASN_PARSE_E;
        }
    }
#endif
```

### Non-template path (`wolfcrypt/src/asn.c`)

```c
ret = GetInteger16Bit(params, &idx, sz);
#ifndef WOLFSSL_NO_ASN_STRICT
    /* RFC 8017 A.2.3: trailerField SHALL be trailerFieldBC(1). */
    if (ret == 1) {
        ret = 0;
    }
    else {
        WOLFSSL_MSG("DecodeRsaPssParams: trailerField must be 1");
        if (ret >= 0)
            ret = ASN_PARSE_E;
    }
#else
    if (ret > 0) { ret = 0; }
    else if (ret != 0) { WOLFSSL_MSG("...fail at trailer_value"); }
#endif
```

## Test

Three cases added to `test_wc_DecodeRsaPssParams` in
`tests/api/test_asn.c`. All use a minimal DER-encoded RSASSA-PSS-params
containing only the `trailerField` component:
`SEQUENCE { [3] CONSTRUCTED { INTEGER <value> } }`.

| Test | DER (hex) | trailerField value | Guard | Expected |
|---|---|---|---|---|
| 9 | `30 05 a3 03 02 01 01` | 1 (trailerFieldBC) | none — all modes | `0` (success) |
| 10 | `30 05 a3 03 02 01 02` | 2 | `!WOLFSSL_NO_ASN_STRICT` | `ASN_PARSE_E` |
| 11 | `30 05 a3 03 02 01 00` | 0 | `!WOLFSSL_NO_ASN_STRICT` | `ASN_PARSE_E` |

Tests 10 and 11 exercise both the template and non-template paths with
the same DER input; the path is selected at build time.

### How to build and run

**Template path (default), strict mode — all three tests active:**

```bash
./configure --enable-rsapss --enable-all \
    CFLAGS="-DNO_WOLFSSL_CIPHER_SUITE_TEST"
git clean -dfx && make
./tests/unit.test --list | grep DecodeRsaPssParams
./tests/unit.test -<N>
```

**Non-template path, strict mode:**

```bash
./configure --enable-rsapss --enable-all --enable-asn=original \
    CFLAGS="-DNO_WOLFSSL_CIPHER_SUITE_TEST"
git clean -dfx && make
```

**Non-strict mode — only Test 9 active, Tests 10/11 compiled out:**

```bash
./configure --enable-rsapss --enable-all \
    CFLAGS="-DNO_WOLFSSL_CIPHER_SUITE_TEST -DWOLFSSL_NO_ASN_STRICT"
git clean -dfx && make
```

## Files changed

| File | Change |
|---|---|
| `wolfcrypt/src/asn.c` | Template path: capture and validate `trailerField` value; non-template path: enforce `trailerField == 1`; both guarded by `!WOLFSSL_NO_ASN_STRICT` |
| `tests/api/test_asn.c` | Add Tests 9–11 to `test_wc_DecodeRsaPssParams` |

# Checklist

 - [x] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/tests/api/test_asn.c
+++ b/tests/api/test_asn.c
@@ -1097,6 +1097,47 @@ int test_wc_DecodeRsaPssParams(void)
             &hash, &mgf, &saltLen), WC_NO_ERR_TRACE(ASN_PARSE_E));
     }
 
+    /* --- Test 9: trailerField = 1 (trailerFieldBC) => valid in all modes --- */
+    /* SEQUENCE { [3] CONSTRUCTED { INTEGER 1 } } = 30 05 a3 03 02 01 01 */
+    {
+        static const byte trailerValid[] = {
+            0x30, 0x05, 0xa3, 0x03, 0x02, 0x01, 0x01
+        };
+        hash    = WC_HASH_TYPE_NONE;
+        mgf     = 0;
+        saltLen = 0;
+        ExpectIntEQ(wc_DecodeRsaPssParams(trailerValid,
+            (word32)sizeof(trailerValid), &hash, &mgf, &saltLen), 0);
+        ExpectIntEQ((int)hash, (int)WC_HASH_TYPE_SHA);
+        ExpectIntEQ(mgf, WC_MGF1SHA1);
+        ExpectIntEQ(saltLen, 20);
+    }
+
+#ifndef WOLFSSL_NO_ASN_STRICT
+    /* --- Test 10: trailerField = 2 => ASN_PARSE_E (strict mode) --- */
+    /* RFC 8017 A.2.3: trailerField SHALL be trailerFieldBC(1). */
+    /* SEQUENCE { [3] CONSTRUCTED { INTEGER 2 } } = 30 05 a3 03 02 01 02 */
+    {
+        static const byte trailerTwo[] = {
+            0x30, 0x05, 0xa3, 0x03, 0x02, 0x01, 0x02
+        };
+        ExpectIntEQ(wc_DecodeRsaPssParams(trailerTwo,
+            (word32)sizeof(trailerTwo), &hash, &mgf, &saltLen),
+            WC_NO_ERR_TRACE(ASN_PARSE_E));
+    }
+
+    /* --- Test 11: trailerField = 0 => ASN_PARSE_E (strict mode) --- */
+    /* SEQUENCE { [3] CONSTRUCTED { INTEGER 0 } } = 30 05 a3 03 02 01 00 */
+    {
+        static const byte trailerZero[] = {
+            0x30, 0x05, 0xa3, 0x03, 0x02, 0x01, 0x00
+        };
+        ExpectIntEQ(wc_DecodeRsaPssParams(trailerZero,
+            (word32)sizeof(trailerZero), &hash, &mgf, &saltLen),
+            WC_NO_ERR_TRACE(ASN_PARSE_E));
+    }
+#endif /* !WOLFSSL_NO_ASN_STRICT */
+
 #endif /* WC_RSA_PSS && !NO_RSA && !NO_ASN */
     return EXPECT_RESULT();
 }
--- a/tests/api/test_asn.c
+++ b/tests/api/test_asn.c
@@ -1136,6 +1136,20 @@ int test_wc_DecodeRsaPssParams(void)
             (word32)sizeof(trailerZero), &hash, &mgf, &saltLen),
             WC_NO_ERR_TRACE(ASN_PARSE_E));
     }
+
+    /* --- Test 12: trailerField = 256 (multi-byte INTEGER) => ASN_PARSE_E ---
+     * Exercises the 2-byte integer branch in GetInteger16Bit (non-template)
+     * and the len==2 case of ASN_DATA_TYPE_WORD16 (template path).
+     * SEQUENCE { [3] CONSTRUCTED { INTEGER 256 } } = 30 06 a3 04 02 02 01 00
+     */
+    {
+        static const byte trailerMultiByte[] = {
+            0x30, 0x06, 0xa3, 0x04, 0x02, 0x02, 0x01, 0x00
+        };
+        ExpectIntEQ(wc_DecodeRsaPssParams(trailerMultiByte,
+            (word32)sizeof(trailerMultiByte), &hash, &mgf, &saltLen),
+            WC_NO_ERR_TRACE(ASN_PARSE_E));
+    }
 #endif /* !WOLFSSL_NO_ASN_STRICT */
 
 #endif /* WC_RSA_PSS && !NO_RSA && !NO_ASN */
--- a/wolfcrypt/src/asn.c
+++ b/wolfcrypt/src/asn.c
@@ -8074,6 +8074,9 @@ static int DecodeRsaPssParams(const byte* params, word32 sz,
 {
     DECL_ASNGETDATA(dataASN, rsaPssParamsASN_Length);
     int ret = 0;
+#ifndef WOLFSSL_NO_ASN_STRICT
+    word16 trailerVal = 1;
+#endif
     word16 sLen = 20;
 
     /* Default values. */
@@ -8089,6 +8092,10 @@ static int DecodeRsaPssParams(const byte* params, word32 sz,
         GetASN_OID(&dataASN[RSAPSSPARAMSASN_IDX_MGFHOID], oidHashType);
         /* Place the salt length into 16-bit var sLen. */
         GetASN_Int16Bit(&dataASN[RSAPSSPARAMSASN_IDX_SALTLENINT], &sLen);
+#ifndef WOLFSSL_NO_ASN_STRICT
+        /* Capture trailerField value for RFC 8017 A.2.3 validation. */
+        GetASN_Int16Bit(&dataASN[RSAPSSPARAMSASN_IDX_TRAILERINT], &trailerVal);
+#endif
         /* Decode the algorithm identifier. */
         ret = GetASN_Items(rsaPssParamsASN, dataASN, rsaPssParamsASN_Length, 1,
             params, &inOutIdx, sz);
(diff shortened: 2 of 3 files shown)


wolfcrypt/src/asn.c
```diff
@@ -8251,12 +8267,24 @@ static int DecodeRsaPssParams(const byte* params, word32 sz,
...
+#ifndef WOLFSSL_NO_ASN_STRICT
+                /* RFC 8017 A.2.3: trailerField SHALL be trailerFieldBC(1). */
+                if (ret == 1) {
+                    ret = 0;
+                }
+                else {
+                    WOLFSSL_MSG("DecodeRsaPssParams: trailerField must be 1");
+                    if (ret >= 0)
+                        ret = ASN_PARSE_E;
+                }
```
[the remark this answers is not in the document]
miyazakh: fixed

dgarske (dismissed):

dgarske (commented):
## Skoll Code Review

### Posted findings
- [Medium] **No test for multi-byte trailerField value (exercises 2-byte parse branch)** — `tests/api/test_asn.c:1116-1138`
- [Medium] **Stricter trailerField default tightens cert/PKCS7 parsing for all callers** — `wolfcrypt/src/asn.c:8115-8123,8270-8289`
- [Low] **Behavioral tightening: default builds now reject non-conformant trailerField in cert/sig parsing** — `wolfcrypt/src/asn.c:8115-8123,8270-8289`
- [Low] **Literal 1 for trailerFieldBC could be a named constant** — `wolfcrypt/src/asn.c:8118,8272`

### Skipped findings
- [Low] ~~Tests 10/11 do not reset hash/mgf/saltLen before invocation~~

tests/api/test_asn.c
```diff
@@ -1097,6 +1097,47 @@ int test_wc_DecodeRsaPssParams(void)
...
+        mgf     = 0;
+        saltLen = 0;
+        ExpectIntEQ(wc_DecodeRsaPssParams(trailerValid,
+            (word32)sizeof(trailerValid), &hash, &mgf, &saltLen), 0);
+        ExpectIntEQ((int)hash, (int)WC_HASH_TYPE_SHA);
+        ExpectIntEQ(mgf, WC_MGF1SHA1);
+        ExpectIntEQ(saltLen, 20);
+    }
+
+#ifndef WOLFSSL_NO_ASN_STRICT
```
dgarske, line 1116: **🟠 [Medium] No test for multi-byte trailerField value (exercises 2-byte parse branch)** · _test_

Tests 9-11 cover single-byte trailerField values 1, 2 and 0. None exercises a multi-byte INTEGER value (e.g. 256, encoded `02 02 01 00`). This matters because the new fix takes different code paths for multi-byte integers: in the non-template path `GetInteger16Bit` uses its `input[idx] == 2` (2-byte length) branch, and in the template path the newly-added `GetASN_Int16Bit` is what now constrains the value to 16 bits. A value such as 256 should still be rejected with `ASN_PARSE_E` in strict mode, but that path is currently unverified. Adding it would also guard the boundary where a 3-byte trailer (>65535) is rejected by `GetASN_Int16Bit`/`GetInteger16Bit` rather than by the explicit `trailerVal != 1` check.

**Fix:** Add a strict-mode case with a multi-byte trailerField, e.g. SEQUENCE { [3] { INTEGER 256 } } = `30 06 a3 04 02 02 01 00`, expecting `ASN_PARSE_E`, to cover the 2-byte integer parse branch in both paths.

wolfcrypt/src/asn.c
```diff
@@ -8105,6 +8112,15 @@ static int DecodeRsaPssParams(const byte* params, word32 sz,
         word32 oid = dataASN[RSAPSSPARAMSASN_IDX_MGFHOID].data.oid.sum;
         ret = RsaPssHashOidToMgf1(oid, mgf);
     }
+#ifndef WOLFSSL_NO_ASN_STRICT
+    /* RFC 8017 A.2.3: trailerField SHALL be trailerFieldBC(1). */
```
dgarske, line 8116: **🟠 [Medium] Stricter trailerField default tightens cert/PKCS7 parsing for all callers**

`DecodeRsaPssParams` is called from several certificate and PKCS7 parsing paths (`pkcs7.c:5673`, `asn.c:8903`, `asn.c:11409`, `asn.c:13452`, `asn.c:13458`, `asn.c:16920`, `asn_orig.c:298`). All of them propagate the new `ASN_PARSE_E`, e.g. `pkcs7.c` returns `ASN_PARSE_E` and fails the parse when `wc_DecodeRsaPssParams` is non-zero. The previous behavior silently accepted any positive `trailerField` (and value 0), so this PR changes the default (`!WOLFSSL_NO_ASN_STRICT`) behavior: a certificate or CMS structure carrying a non-conformant `trailerField` that parsed before will now be rejected. This is the intended fix per RFC 8017 A.2.3 and is gated behind `WOLFSSL_NO_ASN_STRICT`, but it is a behavioral/interoperability change that should be surfaced in release notes/changelog so downstream users who hit non-conformant peers know about the `WOLFSSL_NO_ASN_STRICT` escape hatch.

**Fix:** Confirm this default-on stricter behavior is intended for cert/PKCS7 parsing and note the change (and the `WOLFSSL_NO_ASN_STRICT` opt-out) in the changelog/release notes.

wolfcrypt/src/asn.c
```diff
@@ -8074,6 +8074,9 @@ static int DecodeRsaPssParams(const byte* params, word32 sz,
 {
     DECL_ASNGETDATA(dataASN, rsaPssParamsASN_Length);
     int ret = 0;
+#ifndef WOLFSSL_NO_ASN_STRICT
```
dgarske, line 8077: **🔵 [Low] Behavioral tightening: default builds now reject non-conformant trailerField in cert/sig parsing** · _question_

DecodeRsaPssParams is reached from five cert/signature-parsing call sites (asn.c:8903, 11409, 13452, 13458, 16920) and from pkcs7.c:5673. In default builds (WOLFSSL_NO_ASN_STRICT not defined), inputs with trailerField != 1 (including value 0, which the non-template path previously accepted, and any value the template path previously ignored entirely) are now rejected with ASN_PARSE_E. This is the intended RFC 8017 A.2.3 conformance fix and is correctly behind the WOLFSSL_NO_ASN_STRICT escape hatch, so previously-permissive behavior remains available. Flagging only to document the blast radius: a previously-accepted (malformed) RSA-PSS AlgorithmIdentifier in a cert or PKCS#7 structure will now fail to parse by default. Real-world encoders always emit trailerFieldBC(1), so practical risk is low.

**Fix:** No code change required. Confirm this stricter default is intended for the cert/PKCS#7 verification paths (it appears so per the PR description) and that it is mentioned in release notes/ChangeLog so downstreams relying on lax parsing know to define WOLFSSL_NO_ASN_STRICT.

dgarske:
Jenkins retest this please

dgarske (approved):

PR #10597, opened 2026-06-04
SP gen: FP_ECC init mutex improvement

SparkiDev:
# Description

F-1379
Better handling of the lazy mutex initialization to use atomics where available.

Improved atomic code when no system support:
 - add types
 - used types in functions

Add --no-ec to unit.test to not run wolfCrypt tests.

# Testing

./configure --disable-shared --enable-sp --enable-fpecc CFLAGS=-DWOLFSSL_TEST_NO_MUTEX_INITIALIZER --disable-threadlocal

### DIFF
--- a/tests/unit.c
+++ b/tests/unit.c
@@ -43,6 +43,7 @@
 
 int allTesting = 1;
 int apiTesting = 1;
+int wolfCryptTesting = 1;
 int myoptind = 0;
 char* myoptarg = NULL;
 int unit_test(int argc, char** argv);
@@ -202,6 +203,9 @@ int unit_test(int argc, char** argv)
             ApiTest_PrintTestCases();
             goto exit;
         }
+        else if (XSTRCMP(argv[1], "--no-wc") == 0) {
+            wolfCryptTesting = 0;
+        }
         else if (XSTRCMP(argv[1], "--api") == 0) {
             allTesting = 0;
         }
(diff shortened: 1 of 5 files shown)


wolfcrypt/src/sp_arm32.c
```diff
@@ -75866,19 +75867,33 @@ static int sp_256_ecc_mulmod_8(sp_point_256* r, const sp_point_256* g,
...
+                if (err == 0) {
+                    WOLFSSL_ATOMIC_STORE(initCacheMutex_256, 1);
+                }
+            }
+
+            WOLFSSL_ATOMIC_STORE(lock_token_256, 0);
         }
-        #endif
-        if (wc_LockMutex(&sp_cache_256_lock) != 0) {
+    #endif
```
douzzer, line 75900: You only need one atomic to get the semantics you need here, namely, `initCacheMutex_256` -- loop a locked compare-exchange on it expecting `0`.  On success, call `wc_InitMutex()` then store `2` to `initCacheMutex_256`.  On failure, check if the actual value (saved to `orig` which should actually be called `expected_then_actual` or somesuch) is `2`.  If so, the waited thread finished its `wc_InitMutex()`, and this thread can now safely jump forward to `wc_LockMutex()`.  If not, `WC_RELAX_LONG_LOOP()` then try again (optionally assert that it's `1`).

douzzer (changes requested):

wolfcrypt/src/sp_arm32.c
SparkiDev: AI agrees!
Fixed.

douzzer (approved):

PR #10598, opened 2026-06-04
Create individual crypto callbacks for hkdf extract and hkdf expand

twcook86:
# Description

Fixes zd#21919

# Testing

built-in test.c expanded

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/wolfcrypt/src/cryptocb.c
+++ b/wolfcrypt/src/cryptocb.c
@@ -2532,6 +2532,7 @@ int wc_CryptoCb_Hkdf(int hashType, const byte* inKey, word32 inKeySz,
     return wc_CryptoCb_TranslateErrorCode(ret);
 }
 
+/* NOTE: size of 'out' must be the digest size per hashType */
 int wc_CryptoCb_Hkdf_Extract(int hashType, const byte* salt, word32 saltSz,
                         const byte* inKey, word32 inKeySz, byte* out, int devId)
 {
--- a/wolfcrypt/src/cryptocb.c
+++ b/wolfcrypt/src/cryptocb.c
@@ -230,6 +230,10 @@ static const char* GetKdfTypeStr(int type)
     switch (type) {
         case WC_KDF_TYPE_HKDF:
             return "HKDF";
+        case WC_KDF_TYPE_HKDF_EXTRACT:
+            return "HKDF Extract";
+        case WC_KDF_TYPE_HKDF_EXPAND:
+            return "HKDF Expand";
         case WC_KDF_TYPE_TWOSTEP_CMAC:
             return "TWOSTEP_CMAC";
     }
@@ -2521,6 +2525,64 @@ int wc_CryptoCb_Hkdf(int hashType, const byte* inKey, word32 inKeySz,
         cryptoInfo.kdf.hkdf.infoSz   = infoSz;
         cryptoInfo.kdf.hkdf.out      = out;
         cryptoInfo.kdf.hkdf.outSz    = outSz;
+
+        ret = dev->cb(dev->devId, &cryptoInfo, dev->ctx);
+    }
+
+    return wc_CryptoCb_TranslateErrorCode(ret);
+}
+
+int wc_CryptoCb_Hkdf_Extract(int hashType, const byte* salt, word32 saltSz,
+                        const byte* inKey, word32 inKeySz, byte* out, int devId)
+{
+    int ret = WC_NO_ERR_TRACE(CRYPTOCB_UNAVAILABLE);
+    CryptoCb* dev;
+
+    /* Find registered callback device */
+    dev = wc_CryptoCb_FindDevice(devId, WC_ALGO_TYPE_KDF);
+
+    if (dev && dev->cb) {
+        wc_CryptoInfo cryptoInfo;
+        XMEMSET(&cryptoInfo, 0, sizeof(cryptoInfo));
+
+        cryptoInfo.algo_type                 = WC_ALGO_TYPE_KDF;
+        cryptoInfo.kdf.type                  = WC_KDF_TYPE_HKDF_EXTRACT;
+        cryptoInfo.kdf.hkdf_extract.hashType = hashType;
+        cryptoInfo.kdf.hkdf_extract.salt     = salt;
+        cryptoInfo.kdf.hkdf_extract.saltSz   = saltSz;
+        cryptoInfo.kdf.hkdf_extract.inKey    = inKey;
+        cryptoInfo.kdf.hkdf_extract.inKeySz  = inKeySz;
+        cryptoInfo.kdf.hkdf_extract.out      = out;
+
+        ret = dev->cb(dev->devId, &cryptoInfo, dev->ctx);
+    }
+
+    return wc_CryptoCb_TranslateErrorCode(ret);
+}
+
+int wc_CryptoCb_Hkdf_Expand(int hashType, const byte* inKey, word32 inKeySz,
+                     const byte* info, word32 infoSz, byte* out, word32 outSz,
+                     int devId)
+{
+    int       ret = WC_NO_ERR_TRACE(CRYPTOCB_UNAVAILABLE);
+    CryptoCb* dev;
+
+    /* Find registered callback device */
+    dev = wc_CryptoCb_FindDevice(devId, WC_ALGO_TYPE_KDF);
+
+    if (dev && dev->cb) {
+        wc_CryptoInfo cryptoInfo;
+        XMEMSET(&cryptoInfo, 0, sizeof(cryptoInfo));
+
+        cryptoInfo.algo_type                = WC_ALGO_TYPE_KDF;
+        cryptoInfo.kdf.type                 = WC_KDF_TYPE_HKDF_EXPAND;
+        cryptoInfo.kdf.hkdf_expand.hashType = hashType;
+        cryptoInfo.kdf.hkdf_expand.inKey    = inKey;
+        cryptoInfo.kdf.hkdf_expand.inKeySz  = inKeySz;
+        cryptoInfo.kdf.hkdf_expand.info     = info;
+        cryptoInfo.kdf.hkdf_expand.infoSz   = infoSz;
+        cryptoInfo.kdf.hkdf_expand.out      = out;
+        cryptoInfo.kdf.hkdf_expand.outSz    = outSz;
 
         ret = dev->cb(dev->devId, &cryptoInfo, dev->ctx);
     }
--- a/wolfcrypt/src/hmac.c
+++ b/wolfcrypt/src/hmac.c
@@ -1832,35 +1832,35 @@ int wolfSSL_GetHmacMaxSize(void)
         word32 hashSz;
         byte   n = 0x1;
 
+        if (inKey == NULL || out == NULL)
+            return BAD_FUNC_ARG;
+
+        ret = wc_HmacSizeByType(type);
+        if (ret < 0)
+            return ret;
+        hashSz = (word32)ret;
+
+        /* RFC 5869 states that the length of output keying material in
+         * octets must be L <= 255*HashLen or N = ceil(L/HashLen) */
+        if (outSz/hashSz + ((outSz % hashSz) != 0) > 255)
+            return BAD_FUNC_ARG;
+
 #ifdef WOLF_CRYPTO_CB
         /* Try crypto callback first for complete operation */
         if (devId != INVALID_DEVID) {
-             ret = wc_CryptoCb_Hkdf_Expand(type, inKey, inKeySz, info, infoSz,
+            ret = wc_CryptoCb_Hkdf_Expand(type, inKey, inKeySz, info, infoSz,
                                            out, outSz, devId);
             if (ret != WC_NO_ERR_TRACE(CRYPTOCB_UNAVAILABLE))
                 return ret;
         }
 #endif
 
-        ret = wc_HmacSizeByType(type);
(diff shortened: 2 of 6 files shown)


wolfssl/wolfcrypt/types.h
```diff
@@ -1440,8 +1440,10 @@ enum wc_AlgoType {
 enum wc_KdfType {
     WC_KDF_TYPE_NONE = 0,
     WC_KDF_TYPE_HKDF = 1,
-    WC_KDF_TYPE_TWOSTEP_CMAC = 2 /* NIST SP 800-56C two-step cmac kdf. */
-    /* Future: WC_KDF_TYPE_PBKDF2 = 3, WC_KDF_TYPE_SCRYPT = 4, etc. */
+    WC_KDF_TYPE_HKDF_EXTRACT = 2,
+    WC_KDF_TYPE_HKDF_EXPAND = 3,
```
bigbrett: In general, I would try to not modify existing public enum values once they are declared, just in case they are used by customers.

In this case I think it is fine as I believe it is ONLY used internally. But something to keep in mind going forward

bigbrett (dismissed):
Would prefer you append-only to public structs, but its fine as is since it is only used by CryptoCbs and those aren't ABI restricted

wolfcrypt/src/hmac.c
```diff
@@ -1758,6 +1758,16 @@ int wolfSSL_GetHmacMaxSize(void)
...
+#ifdef WOLF_CRYPTO_CB
+        /* Try crypto callback first */
+        if (devId != INVALID_DEVID) {
+             ret = wc_CryptoCb_Hkdf_Extract(type, salt, saltSz, inKey, inKeySz,
+                                            out, devId);
+            if (ret != WC_NO_ERR_TRACE(CRYPTOCB_UNAVAILABLE))
+                return ret;
+        }
+#endif
+
```
billphipps, line 1770: This should be moved to the wc_Hkdf_Extract public function so ANY potential invocation will trigger this check/offload.  This means that if someone invokes the one-shot, it will try to offload the one-shot.  If the one-shot is not offloaded, then the each of the extract/expand will be subsequently attempted to be offloaded since the SW version of the one-shot invokes the Expand/Extract directly.  This allows a customer to invoke the one-shot separately since it may optimize better.

wolfcrypt/src/cryptocb.c
```diff
@@ -2527,6 +2531,64 @@ int wc_CryptoCb_Hkdf(int hashType, const byte* inKey, word32 inKeySz,
 
     return wc_CryptoCb_TranslateErrorCode(ret);
 }
+
+int wc_CryptoCb_Hkdf_Extract(int hashType, const byte* salt, word32 saltSz,
+                        const byte* inKey, word32 inKeySz, byte* out, int devId)
```
billphipps, line 2537: Are all of these parameters necessary?  Please confirm.  Fill in NULL/0 as necessary in the shared cryptoInfo structure below.

wolfcrypt/src/cryptocb.c
```diff
@@ -2527,6 +2531,64 @@ int wc_CryptoCb_Hkdf(int hashType, const byte* inKey, word32 inKeySz,
...
+
+        ret = dev->cb(dev->devId, &cryptoInfo, dev->ctx);
+    }
+
+    return wc_CryptoCb_TranslateErrorCode(ret);
+}
+
+int wc_CryptoCb_Hkdf_Expand(int hashType, const byte* inKey, word32 inKeySz,
+                     const byte* info, word32 infoSz, byte* out, word32 outSz,
+                     int devId)
```
billphipps, line 2566: Samme comment about if all of these are necessary

wolfcrypt/test/test.c
billphipps, line 1: Consider looking at the test/swdev series of tests to support testing of the cryptocb plumbing.

billphipps (changes requested):
Please move the check for cryptocb into the hdkf functions, not in the side that is using it.  Check on unit testing the interface as well.  Looks like a great start!

wolfcrypt/src/cryptocb.c
twcook86: salt is optional, so we just pass salt and saltSz through to the cb, just like the one-shot wc_HKDF_ex() does.

wolfcrypt/src/cryptocb.c
twcook86: info is optional, so we just pass info and infoSz through to the cb, just like the one-shot wc_HKDF_ex() does.

wolfcrypt/src/hmac.c
twcook86: It already works that way as written...

```
wc_HKDF_Extract_ex() {
    if (extract cb)
        call it;
    else {
       do sw extract;
    }
}

wc_HKDF_Expand_ex() {
    if (expand cb)
        call it;
    else {
       do sw expand;
    }
}

wc_HKDF_ex() {
    if (one-shot cb)
        call it;
    else {
        wc_HKDF_Extract_ex();
        wc_HKDF_Expand_ex();
    }
}
```

wolfssl/wolfcrypt/types.h
twcook86: understood.  fixed anyway.

ColtonWilley:
Jenkins retest this please

wolfcrypt/test/test.c
twcook86: I wasn't aware of the swdev cb testing.  There's no hkdf nor kdf support in there yet.  Perhaps we should scope a separate pr to support kdf/hkdf in swdev testing?

billphipps (dismissed):
LGTM!  Thank you!

wolfcrypt/test/test.c
```diff
@@ -34846,6 +34847,105 @@ WOLFSSL_TEST_SUBROUTINE wc_test_ret_t hkdf_test(void)
...
+    if (ret != 0)
+        return WC_TEST_RET_ENC_EC(ret);
+
+    if (XMEMCMP(okm1, res4, (unsigned long)L) != 0)
+        return WC_TEST_RET_ENC_NC;
+#endif /* HAVE_FIPS */
+#endif /* !NO_SHA256 */
 #endif /* !NO_SHA || !NO_SHA256 */
 
 #if !defined(NO_SHA256) && !defined(HAVE_SELFTEST) && \
```
dgarske, line 34951: **🔵 [Low] New inKey/out NULL validation in wc_HKDF_Expand_ex lacks a negative test**

The PR adds a new boundary check `if (inKey == NULL || out == NULL) return BAD_FUNC_ARG;` to `wc_HKDF_Expand_ex` (hmac.c:1835-1836). The expanded test adds bad-arg coverage only for `wc_HKDF_Extract` (NULL out, and NULL inKey with non-zero size at lines 34953-34960) but nothing exercises the newly added `wc_HKDF_Expand` NULL-inKey / NULL-out path. Adding a matching negative test would lock in the new validation and prevent regressions.

**Fix:** Add a bad-arg test for wc_HKDF_Expand covering NULL inKey and NULL out, alongside the existing wc_HKDF_Extract bad-arg tests.

wolfcrypt/src/hmac.c
```diff
@@ -1758,6 +1758,16 @@ int wolfSSL_GetHmacMaxSize(void)
             return BAD_FUNC_ARG;
         }
 
+#ifdef WOLF_CRYPTO_CB
+        /* Try crypto callback first */
+        if (devId != INVALID_DEVID) {
+             ret = wc_CryptoCb_Hkdf_Extract(type, salt, saltSz, inKey, inKeySz,
```
dgarske: **⚪ [Info] Misaligned indentation in new Extract callback block**

In the new crypto-callback block in `wc_HKDF_Extract_ex`, the `ret = wc_CryptoCb_Hkdf_Extract(...)` line is indented 13 spaces while the surrounding `if`/`return` are at 12, leaving an extra leading space. The sibling block in `wc_HKDF_Expand_ex` (line 1851) is correctly aligned at 12. This mirrors a pre-existing quirk in `wc_HKDF_ex`, so it's harmless, but since these are new lines it's worth cleaning up.

**Fix:** Remove the extra leading space on the `ret =` line so it aligns at 12 spaces, matching the Expand block.

dgarske (commented):
## Skoll Code Review

### Posted findings
- [Low] **New inKey/out NULL validation in wc_HKDF_Expand_ex lacks a negative test** — `wolfcrypt/test/test.c:34951-34962`
- [Info] **Misaligned indentation in new Extract callback block** — `wolfcrypt/src/hmac.c:1764`

### Skipped findings
- [Medium] ~~Existing KDF crypto-cb devices now receive new HKDF_EXTRACT/HKDF_EXPAND callbacks~~

twcook86:
jenkins retest this please

wolfssl/wolfcrypt/cryptocb.h
```diff
@@ -645,6 +645,23 @@ typedef struct wc_CryptoInfo {
                 byte*       out; /* Output key material */
                 word32      outSz;
             } hkdf;
+            struct {                  /* HKDF extract */
```
dgarske, line 648: **🔵 [Low] hkdf_extract cryptoInfo struct has no output-size field for callback implementers**

The new `hkdf_extract` struct (and `wc_CryptoCb_Hkdf_Extract`) pass `out` with no size field, since HKDF-Extract output is always one hash length per RFC 5869. That is correct, but unlike `hkdf` and `twostep_cmac` (which carry `outSz`), a hardware callback implementer receives no explicit length and must independently know to write exactly `hashType`'s digest length into `out`. The salt-NULL semantics (software path treats NULL salt as `hashSz` zero bytes) are also implicit — the callback receives `salt=NULL, saltSz=0` and must replicate that behavior. The in-tree test device (`myCryptoDevCb`) hides this because it re-dispatches to the software `wc_HKDF_Extract_ex`, which handles both cases. A real device backend could get the buffer length or NULL-salt handling wrong.

**Fix:** Add a short comment to the `hkdf_extract` struct (and/or the `wc_CryptoCb_Hkdf_Extract` doxygen) documenting that `out` must be the digest length and that a NULL salt must be treated as a hashLen zero block, so hardware callback authors implement it correctly.

wolfssl/wolfcrypt/cryptocb.h
```diff
@@ -894,6 +911,11 @@ WOLFSSL_LOCAL int wc_CryptoCb_Hkdf(int hashType, const byte* inKey,
                                    word32 saltSz, const byte* info,
                                    word32 infoSz, byte* out, word32 outSz,
                                    int devId);
+WOLFSSL_LOCAL int wc_CryptoCb_Hkdf_Extract(int hashType, const byte* salt,
```
dgarske, line 914: **⚪ [Info] New cryptocb prototypes guarded by HAVE_HKDF only while definitions also require !NO_HMAC**

The new prototypes are declared under `#ifdef HAVE_HKDF`, while their definitions in cryptocb.c are under `#if defined(HAVE_HKDF) && !defined(NO_HMAC)`. If `HAVE_HKDF` were ever defined together with `NO_HMAC`, the header would declare symbols the .c file does not define. In practice HKDF requires HMAC so this combination should not occur, and the PR is consistent with the pre-existing `wc_CryptoCb_Hkdf` prototype which has the same guard mismatch — so this is not a regression introduced here, just an inherited inconsistency carried forward.

**Fix:** Optionally align the header guard with the .c guard (`#if defined(HAVE_HKDF) && !defined(NO_HMAC)`) for all three HKDF prototypes; low priority since the combination is not a supported build.

wolfcrypt/src/hmac.c
```diff
@@ -1754,19 +1754,27 @@ int wolfSSL_GetHmacMaxSize(void)
         const  byte* localSalt;  /* either points to user input or tmp */
         word32 hashSz;
 
-        if (out == NULL || (inKey == NULL && inKeySz > 0)) {
+        if (out == NULL || (inKey == NULL && inKeySz > 0))
```
dgarske: **🔵 [Low] Drive-by removal of braces from existing if-statements in hmac.c**

The PR reformats several pre-existing single-statement `if` blocks from braced to brace-less form (e.g. `if (out == NULL ...) { return BAD_FUNC_ARG; }` becomes `if (out == NULL ...) return BAD_FUNC_ARG;`, and similarly for the `wc_HmacSizeByType` checks). This is unrelated to the feature being added and enlarges the diff/blame noise. wolfSSL code predominantly keeps braces on `if` bodies, and the surrounding function originally used them. The functional behavior is unchanged and correct; this is purely a style/diff-hygiene note.

**Fix:** Optional: revert the unnecessary brace/style changes to keep the diff focused on the new functionality.

---
> **Note:** Referenced line (`wolfcrypt/src/hmac.c:1757-1772, 1833-1844, 1947-1949`) is outside the diff hunk. Comment anchored to nearest changed region.

dgarske (commented):
## Skoll Code Review

### Posted findings
- [Low] **Drive-by removal of braces from existing if-statements in hmac.c** — `wolfcrypt/src/hmac.c:1757-1772, 1833-1844, 1947-1949`
- [Low] **hkdf_extract cryptoInfo struct has no output-size field for callback implementers** — `wolfssl/wolfcrypt/cryptocb.h:648-655`
- [Info] **New cryptocb prototypes guarded by HAVE_HKDF only while definitions also require !NO_HMAC** — `wolfssl/wolfcrypt/cryptocb.h:914-918`

dgarske (approved):

PR #10603, opened 2026-06-05
PQC enforce modulus and eta range checks in DecodePrivateKey

aidangarske:
# Description

- In https://github.com/wolfSSL/wolfProvider/pull/399 I did inter-op test with WP and openssl. There wherre 2 issues with mlkem and mldsa 
- 1. mlkem specifically needed A modulus check needed to be added to reject non reduced priv key vectors on decode
- 2. mldsa needs to reject priv key data with out of range s1/s2 
- Added neg test vectors to fix 
- Tested in WP 2602 interop pass from openssl test suite
- skoll, copilot, fenrir addressed


Frauschi (dismissed):
LGTM, but @SparkiDev should give the final go on this.

ColtonWilley:
Jenkins retest this please

SparkiDev (approved):

SparkiDev:
Private values should come from a trusted source but having the check won't hurt.

PR #10604, opened 2026-06-05
Expose wc_CryptoCb_GetDevice and add CryptoCb API test coverage

AlexLanzano:
# Description

Exposes `wc_CryptoCb_GetDevice()` as a public API and adds unit test coverage for it.
This is necessary to if you want to check if a device ID has already been registered.

# Testing

Added a unit test in `test_wc_CryptoCb()` (`tests/api.c`) that exercises the lookup behavior:

```
./configure --enable-cryptocb --enable-all
make -j tests/unit.test
./tests/unit.test -test_wc_CryptoCb
# => test_wc_CryptoCb : passed
#    Failed/Skipped/Passed/All: 0/0/1/1
```

### DIFF
--- a/tests/api.c
+++ b/tests/api.c
@@ -29227,13 +29227,14 @@ static int test_wc_CryptoCb(void)
     (!defined(NO_RSA) || defined(HAVE_ECC) || defined(HAVE_ED25519))
     int tlsVer;
 #endif
-    /* TODO: Add crypto callback API tests */
-
-    /* Test wc_CryptoCb_GetDevice() lookup behavior. */
+    /* Exercise the exposed CryptoCb device-management API:
+     * wc_CryptoCb_RegisterDevice / UnRegisterDevice / GetDevice /
+     * DefaultDevID (and InfoString when DEBUG_CRYPTOCB is enabled). */
     {
         int    getDevId  = 1234;
         int    getDevCtx = 0;
         CryptoCb* dev     = NULL;
+        int    i, n, rc;
 
         /* Unregistered devId is not found. */
         ExpectNull(wc_CryptoCb_GetDevice(getDevId));
@@ -29250,9 +29251,50 @@ static int test_wc_CryptoCb(void)
         /* A different, unregistered devId is still not found. */
         ExpectNull(wc_CryptoCb_GetDevice(getDevId + 1));
 
+        /* Re-registering an already-registered devId is rejected. */
+        ExpectIntEQ(wc_CryptoCb_RegisterDevice(getDevId, NULL, &getDevCtx),
+            WC_NO_ERR_TRACE(ALREADY_E));
+
+        /* wc_CryptoCb_DefaultDevID behavior depends on the build config. */
+    #ifdef WC_NO_DEFAULT_DEVID
+        ExpectIntEQ(wc_CryptoCb_DefaultDevID(), INVALID_DEVID);
+    #elif !defined(WOLFSSL_CAAM_DEVID) && !defined(HAVE_ARIA) && \
+          !defined(WC_USE_DEVID)
+        /* "first available" mode: a device is registered, so one is returned. */
+        ExpectIntNE(wc_CryptoCb_DefaultDevID(), INVALID_DEVID);
+    #endif
+
         /* After unregistering, the device is no longer found. */
         wc_CryptoCb_UnRegisterDevice(getDevId);
         ExpectNull(wc_CryptoCb_GetDevice(getDevId));
+
+        /* Unregistering is a harmless no-op for unknown or invalid devIds. */
+        wc_CryptoCb_UnRegisterDevice(getDevId);       /* already removed */
+        wc_CryptoCb_UnRegisterDevice(INVALID_DEVID);  /* never a real devId */
+        ExpectNull(wc_CryptoCb_GetDevice(getDevId));
+
+        /* The device table is finite: once full, registration returns
+         * BUFFER_E. Register unique devIds until that happens, then clean up.
+         * The cap (256) just guards against an unexpectedly large table. */
+        rc = 0;
+        for (i = 0; rc == 0 && i < 256; i++) {
+            rc = wc_CryptoCb_RegisterDevice(0x5000 + i, NULL, NULL);
+        }
+        ExpectIntEQ(rc, WC_NO_ERR_TRACE(BUFFER_E));
+        /* Every id except the final failed attempt registered; unregister them. */
+        for (n = 0; n + 1 < i; n++) {
+            wc_CryptoCb_UnRegisterDevice(0x5000 + n);
+        }
+
+    #ifdef DEBUG_CRYPTOCB
+        /* wc_CryptoCb_InfoString smoke test: must not dereference bad memory. */
+        {
+            wc_CryptoInfo info;
+            XMEMSET(&info, 0, sizeof(info));
+            info.algo_type = WC_ALGO_TYPE_NONE;
+            wc_CryptoCb_InfoString(&info);
+        }
+    #endif
     }
 
 #ifdef HAVE_IO_TESTS_DEPENDENCIES
--- a/tests/api.c
+++ b/tests/api.c
@@ -29223,13 +29223,39 @@ static int test_wc_CryptoCb(void)
     (!defined(WOLF_CRYPTO_CB_ONLY_SHA256) && !defined(WOLF_CRYPTO_CB_ONLY_AES) && \
      !defined(WOLF_CRYPTO_CB_ONLY_ECC) && !defined(WOLF_CRYPTO_CB_ONLY_RSA) && \
      !defined(WOLF_CRYPTO_CB_ONLY_SHA512))
+#if defined(HAVE_IO_TESTS_DEPENDENCIES) && \
+    (!defined(NO_RSA) || defined(HAVE_ECC) || defined(HAVE_ED25519))
+    int tlsVer;
+#endif
     /* TODO: Add crypto callback API tests */
 
-#ifdef HAVE_IO_TESTS_DEPENDENCIES
-    #if !defined(NO_RSA) || defined(HAVE_ECC) || defined(HAVE_ED25519)
-    int tlsVer;
-    #endif
+    /* Test wc_CryptoCb_GetDevice() lookup behavior. */
+    {
+        int    getDevId  = 1234;
+        int    getDevCtx = 0;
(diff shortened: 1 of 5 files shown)


dgarske:
@AlexLanzano please fix merge conflict and review the CI failures. Thanks

dgarske:
Jenkins retest this please: "ERROR: Build 'wolfSSL/PRB-fips-repo-and-harness-test-v3-part2' failed with result: FAILURE"

wolfcrypt/src/cryptocb.c
```diff
@@ -451,12 +446,14 @@ void wc_CryptoCb_SetDeviceFindCb(CryptoDevCallbackFind cb)
 int wc_CryptoCb_RegisterDevice(int devId, CryptoDevCallbackFunc cb, void* ctx)
```
dgarske, line 457: **🟠 [Medium] Public API behavior change: re-registering a devId now fails with ALREADY_E instead of updating in place**

Beyond exposing `wc_CryptoCb_GetDevice`, this PR silently changes the semantics of the long-standing public API `wc_CryptoCb_RegisterDevice`. Previously, calling it with an already-registered `devId` located the existing slot and overwrote its `cb`/`ctx` (an in-place update / idempotent re-register). Now the first thing the function does is `if (wc_CryptoCb_GetDevice(devId) != NULL) return ALREADY_E;`, so any caller that re-registers a `devId` to swap its callback or context will now fail instead of updating. No in-tree caller appears to break (verified: `maxq10xx_port_init` and the `wc_port.c` registrations run once per `wolfCrypt_Init` because the init body is guarded by `wc_local_InitUp`/refcount, and the Renesas port allocates a fresh `gdevId++` each time), but external integrations that relied on update-on-re-register will silently regress. Also note a minor edge case: passing `devId == INVALID_DEVID` now returns `ALREADY_E` (because the free-slot probe matches), which is a confusing error code for that misuse versus `BAD_FUNC_ARG`. This semantic change is not mentioned in the PR description (which frames the change as only exposing `GetDevice`). Please confirm the change is intentional and call it out explicitly.

**Fix:** Confirm intent and, if kept, treat as an intentional behavior change: document the new ALREADY_E return and consider noting it in release notes since it affects a widely-used public API.

wolfssl/wolfcrypt/cryptocb.h
```diff
@@ -672,12 +672,19 @@ typedef struct wc_CryptoInfo {
...
+    void* ctx;
+} CryptoCb;
+
 WOLFSSL_LOCAL void wc_CryptoCb_Init(void);
 WOLFSSL_LOCAL void wc_CryptoCb_Cleanup(void);
 WOLFSSL_LOCAL int wc_CryptoCb_GetDevIdAtIndex(int startIdx);
 WOLFSSL_API int  wc_CryptoCb_RegisterDevice(int devId, CryptoDevCallbackFunc cb, void* ctx);
 WOLFSSL_API void wc_CryptoCb_UnRegisterDevice(int devId);
 WOLFSSL_API int wc_CryptoCb_DefaultDevID(void);
+WOLFSSL_API CryptoCb* wc_CryptoCb_GetDevice(int devId);
```
dgarske: **🟠 [Medium] Doxygen not updated for new ALREADY_E return and missing docs for new public API wc_CryptoCb_GetDevice**

The PR adds a new `WOLFSSL_API` (`wc_CryptoCb_GetDevice`) but does not add a corresponding doxygen block in `doc/dox_comments/header_files/cryptocb.h`, where every other public CryptoCb API is documented. In addition, `wc_CryptoCb_RegisterDevice`'s doxygen (`doc/dox_comments/header_files/cryptocb.h:1-99`) still lists only `\return CRYPTOCB_UNAVAILABLE / 0 / negative` and does not mention the newly added `ALREADY_E` rejection path. wolfSSL convention is to document public APIs in dox_comments and keep return-value lists accurate.

**Fix:** Add/update doxygen in doc/dox_comments/header_files/cryptocb.h for both the new GetDevice API and the new ALREADY_E return of RegisterDevice.

wolfssl/wolfcrypt/cryptocb.h
```diff
@@ -672,12 +672,19 @@ typedef struct wc_CryptoInfo {
 
 typedef int (*CryptoDevCallbackFunc)(int devId, struct wc_CryptoInfo* info, void* ctx);
 
+typedef struct CryptoCb {
```
dgarske: **🟠 [Medium] Exposing internal CryptoCb struct and returning a mutable pointer to global table state**

To expose `wc_CryptoCb_GetDevice`, the previously file-private `struct CryptoCb` is now published in the public header and the accessor returns a non-const `CryptoCb*` that points directly into the static global `gCryptoDev[]` table. This has two consequences. (1) ABI/encapsulation: the struct layout (`devId`, `cb`, `ctx`) is now a public commitment, and the PR's stated goal -- 'check if a device ID has already been registered' -- only needs a boolean/int, not a live pointer into internal state. Callers can now mutate `dev->cb`/`dev->devId`/`dev->ctx` directly, bypassing the register/unregister API and the `WOLF_CRYPTO_CB_CMD` register/unregister callback notifications. (2) Thread-safety: the module is explicitly not mutex-protected (`/* TODO: Consider linked list with mutex */` at cryptocb.c:82), so a returned pointer can be concurrently cleared/overwritten by another thread calling UnRegister/Register -- a TOCTOU/data-race hazard that was previously contained to internal synchronous use and is now part of the public surface.

**Fix:** Narrow the exposed surface (boolean check or const pointer) and document the aliasing/thread-safety caveats; if a mutable pointer truly must be public, state that the table is not mutex-protected.

tests/api.c
```diff
@@ -29223,13 +29223,81 @@ static int test_wc_CryptoCb(void)
...
+
+        /* Unregistering is a harmless no-op for unknown or invalid devIds. */
+        wc_CryptoCb_UnRegisterDevice(getDevId);       /* already removed */
+        wc_CryptoCb_UnRegisterDevice(INVALID_DEVID);  /* never a real devId */
+        ExpectNull(wc_CryptoCb_GetDevice(getDevId));
+
+        /* The device table is finite: once full, registration returns
+         * BUFFER_E. Register unique devIds until that happens, then clean up.
+         * The cap (256) just guards against an unexpectedly large table. */
+        rc = 0;
```
dgarske, line 29280: **🔵 [Low] Table-full test leaks one registered device if the 256 safety cap is reached**

The fill-the-table loop and its cleanup assume the loop terminates because `wc_CryptoCb_RegisterDevice` returned `BUFFER_E` (i.e., the last attempt -- index `i-1` -- failed and was not registered, so cleanup unregisters `0x5000`..`0x5000+(i-2)`). With the default `MAX_CRYPTO_DEVID_CALLBACKS` (8) this is correct. But if a build sets `MAX_CRYPTO_DEVID_CALLBACKS >= 256` (free slots >= 256), the loop instead exits because `i` hit the 256 cap with `rc == 0`: the assertion `ExpectIntEQ(rc, BUFFER_E)` fails AND the final successfully-registered device (`0x5000 + 255`) is never unregistered, leaking global crypto-device state into subsequent tests. The comment acknowledges the cap 'just guards against an unexpectedly large table' but the cleanup off-by-one only holds on the BUFFER_E path.

**Fix:** Adjust the cleanup loop so it cannot leave a device registered when the 256 cap path is taken.

dgarske (commented):
## Skoll Code Review

### Posted findings
- [Medium] **Public API behavior change: re-registering a devId now fails with ALREADY_E instead of updating in place** — `wolfcrypt/src/cryptocb.c:446-458`
- [Medium] **Doxygen not updated for new ALREADY_E return and missing docs for new public API wc_CryptoCb_GetDevice** — `wolfssl/wolfcrypt/cryptocb.h:687`
- [Medium] **Exposing internal CryptoCb struct and returning a mutable pointer to global table state** — `wolfssl/wolfcrypt/cryptocb.h:675-687`
- [Low] **Table-full test leaks one registered device if the 256 safety cap is reached** — `tests/api.c:29279-29287`

wolfcrypt/src/cryptocb.c
```diff
@@ -451,12 +446,14 @@ void wc_CryptoCb_SetDeviceFindCb(CryptoDevCallbackFind cb)
...
-    if (dev == NULL)
-        dev = wc_CryptoCb_GetDevice(INVALID_DEVID);
+    /* Reject re-registration of an already-registered device ID. */
+    if (wc_CryptoCb_GetDevice(devId) != NULL)
+        return ALREADY_E;
 
+    /* find a free slot */
+    dev = wc_CryptoCb_GetDevice(INVALID_DEVID);
     if (dev == NULL)
         return BUFFER_E; /* out of devices */
```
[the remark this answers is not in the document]
AlexLanzano: We shouldnt silently overwrite existing registrations

tests/api.c
AlexLanzano: Fixed

wolfcrypt/src/cryptocb.c
AlexLanzano: This is intended. I updated the docs to mention this behavior

wolfssl/wolfcrypt/cryptocb.h
AlexLanzano: Fixed. Created a new API function wc_CryptoCb_IsDeviceRegistered which returns a 1 or 0. Moved wc_CryptoCb_GetDevice back to being a static function

dgarske (approved):

bigbrett (approved):

wolfcrypt/src/cryptocb.c
```diff
@@ -451,12 +457,18 @@ void wc_CryptoCb_SetDeviceFindCb(CryptoDevCallbackFind cb)
...
-    CryptoCb* dev = wc_CryptoCb_GetDevice(devId);
-    if (dev == NULL)
-        dev = wc_CryptoCb_GetDevice(INVALID_DEVID);
+    /* INVALID_DEVID marks a free slot and cannot be registered as a device. */
+    if (devId == INVALID_DEVID)
+        return BAD_FUNC_ARG;
+
+    /* Reject re-registration of an already-registered device ID. */
+    if (wc_CryptoCb_GetDevice(devId) != NULL)
+        return ALREADY_E;
```
dgarske, line 468: This is API breaking change! Many of my wolfTPM tests are failing now... ALREADY_E . I am fixing in https://github.com/wolfSSL/wolfTPM/pull/546

PR #10606, opened 2026-06-05
fix tls_bench DTLS mode failures

miyazakh:
# Fix tls_bench DTLS (`-u`) mode failures

## Summary

Running `tls_bench -u` (DTLS mode) failed immediately with the following errors:

```
Running TLS Benchmarks...
ERROR: failed to listen
error creating ctx
Client Error: -125 (out of memory error)
```

And after partial fixes, a third error appeared:

```
error on client write
Client Error: -439 (DTLS trying to send too much in single datagram error)
```

This PR fixes three independent bugs in `examples/benchmark/tls_bench.c`.

---

### Root cause

When DTLS mode is active, `SetupSocketAndListen()` correctly created a `SOCK_DGRAM`
(UDP) socket instead of a `SOCK_STREAM` (TCP) socket. However, `setsockopt()`,
`bind()`, and `listen()` were all inside the TCP-only `else` branch, so:

- The UDP socket was never bound to the server address.
- `listen()` was called unconditionally after the socket-type branch, which always
  fails on a UDP socket (`EOPNOTSUPP`), producing `ERROR: failed to listen`.

### Fix

- Move `setsockopt(SO_REUSEADDR)` and `bind()` outside the TCP-only block so they
  apply to both TCP and UDP sockets.
- Guard the `listen()` call with `#ifdef WOLFSSL_DTLS if (!doDTLS)` so it is only
  called for TCP connections.

---

### Root cause

When iterating over the default cipher list in DTLS mode, TLS 1.3 cipher suites
(names starting with `"TLS13"`) were encountered. The client/server setup code
attempted to use `wolfDTLSv1_3_client/server_method()` inside an
`#ifdef WOLFSSL_DTLS13` guard. When `WOLFSSL_DTLS13` is not compiled in, no
method was assigned, leaving `ctx == NULL`. The subsequent NULL check returned
`MEMORY_E` (-125), producing:

```
error creating ctx
Client Error: -125 (out of memory error)
```

### Fix

Add an early-continue in the cipher iteration loop: when DTLS mode is active and
`WOLFSSL_DTLS13` is not compiled in, silently skip any TLS 1.3 cipher suite.

```c
#if defined(WOLFSSL_DTLS) && !defined(WOLFSSL_DTLS13)
        if (doDTLS && XSTRNCMP(cipher, "TLS13", 5) == 0) {
            /* DTLS 1.3 not compiled in; skip TLS 1.3 ciphers */
            cipher = (next_cipher != NULL) ? (next_cipher + 1) : NULL;
            continue;
        }
#endif
```

---

### Root cause

`TEST_DTLS_PACKET_SIZE` was defined as 8092 bytes (referencing `MAX_UDP_SIZE` in
`internal.h`). However, wolfSSL's `SendData()` checks that the total DTLS record
size fits within `MAX_MTU` (default: **1400 bytes**). Since 8092 bytes of payload
produces a record far larger than 1400 bytes, `SendData()` returned
`DTLS_SIZE_ERROR` (-439) on every write:

```
error on client write
Client Error: -439 (DTLS trying to send too much in single datagram error)
```

`MAX_UDP_SIZE` (8092) is the maximum UDP datagram size wolfSSL is prepared to
receive, not the send-side MTU limit. The two constants serve different purposes
and were conflated.

### Fix

Change `TEST_DTLS_PACKET_SIZE` for non-embedded builds from `8092` to `1200`,
which fits within `MAX_MTU (1400) - MAX_MSG_EXTRA (~102)`.  A comment documents
the two compile-time options that allow larger packets:

- `WOLFSSL_DTLS_MTU` + `wolfSSL_dtls_set_mtu()` — set a per-session MTU
- `WOLFSSL_NO_DTLS_SIZE_CHECK` — auto-fragment writes that exceed the MTU

The automatic fallback already present in tls_bench (`argTestPacketSize` is
reduced to `TEST_DTLS_PACKET_SIZE` when DTLS mode is detected and no `-p` flag
is given) now produces a value that actually works with the default MTU.

---

## Improvement: startup message now reflects the active protocol

The banner printed at the start of the run was always `"Running TLS Benchmarks..."`,
even when DTLS mode (`-u`) was selected.  It now reads `"Running DTLS Benchmarks..."`
when DTLS is active and `"Running TLS Benchmarks..."` otherwise.

---

## Files changed

- `examples/benchmark/tls_bench.c`

---

## Testing

Verified on Linux (x86-64) with a wolfSSL build that has `WOLFSSL_DTLS` and
`WOLFSSL_DTLS13` enabled.

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/examples/benchmark/tls_bench.c
+++ b/examples/benchmark/tls_bench.c
@@ -126,15 +126,10 @@ Or
 #endif
 
 #ifdef WOLFSSL_DTLS
-    #ifdef BENCH_EMBEDDED
-        /* WOLFSSL_MAX_MTU in internal.h */
-        #define TEST_DTLS_PACKET_SIZE   (1500)
-    #else
-        /* Must fit in one DTLS record: WOLFSSL_MAX_MTU (1400) - MAX_MSG_EXTRA (~102).
-         * Define WOLFSSL_DTLS_MTU and call wolfSSL_dtls_set_mtu() to use larger
-         * packets; or build with WOLFSSL_NO_DTLS_SIZE_CHECK to auto-fragment. */
-        #define TEST_DTLS_PACKET_SIZE   (1200)
-    #endif
+    /* Must fit in one DTLS record: WOLFSSL_MAX_MTU (1400) - MAX_MSG_EXTRA (~102).
+     * Define WOLFSSL_DTLS_MTU and call wolfSSL_dtls_set_mtu() to use larger
+     * packets; or build with WOLFSSL_NO_DTLS_SIZE_CHECK to auto-fragment. */
+    #define TEST_DTLS_PACKET_SIZE   (1200)
 #endif
 
 /* In memory transfer buffer maximum size */
--- a/examples/benchmark/tls_bench.c
+++ b/examples/benchmark/tls_bench.c
@@ -130,8 +130,10 @@ Or
         /* WOLFSSL_MAX_MTU in internal.h */
         #define TEST_DTLS_PACKET_SIZE   (1500)
     #else
-        /* MAX_UDP_SIZE in interna.h */
-        #define TEST_DTLS_PACKET_SIZE   (8092)
+        /* Must fit in one DTLS record: WOLFSSL_MAX_MTU (1400) - MAX_MSG_EXTRA (~102).
+         * Define WOLFSSL_DTLS_MTU and call wolfSSL_dtls_set_mtu() to use larger
+         * packets; or build with WOLFSSL_NO_DTLS_SIZE_CHECK to auto-fragment. */
+        #define TEST_DTLS_PACKET_SIZE   (1200)
     #endif
 #endif
 
@@ -1304,25 +1306,29 @@ static int SetupSocketAndListen(int* listenFd, word32 port, int doDTLS)
             fprintf(stderr, "ERROR: failed to create the socket\n");
             return -1;
         }
-
-        /* allow reuse */
-        if (setsockopt(*listenFd, SOL_SOCKET, SO_REUSEADDR,
-                    &optval, sizeof(optval)) == -1) {
-            fprintf(stderr, "setsockopt SO_REUSEADDR failed\n");
-            return -1;
-        }
-
-        /* Listen for the client. */
-        if (bind(*listenFd, (struct sockaddr*)&servAddr,
-                                                    sizeof(servAddr)) == -1) {
-            fprintf(stderr, "ERROR: failed to bind\n");
-            return -1;
-        }
     }
 
-    if (listen(*listenFd, 5) != 0) {
-        fprintf(stderr, "ERROR: failed to listen\n");
+    /* allow reuse */
+    if (setsockopt(*listenFd, SOL_SOCKET, SO_REUSEADDR,
+                &optval, sizeof(optval)) == -1) {
+        fprintf(stderr, "setsockopt SO_REUSEADDR failed\n");
         return -1;
+    }
+
+    if (bind(*listenFd, (struct sockaddr*)&servAddr,
+                                                sizeof(servAddr)) == -1) {
+        fprintf(stderr, "ERROR: failed to bind\n");
+        return -1;
+    }
+
+#ifdef WOLFSSL_DTLS
+    if (!doDTLS)
+#endif
+    {
+        if (listen(*listenFd, 5) != 0) {
+            fprintf(stderr, "ERROR: failed to listen\n");
+            return -1;
+        }
     }
 
 #ifdef BENCH_USE_NONBLOCK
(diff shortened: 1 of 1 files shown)


miyazakh:
retest this please

miyazakh:
retest this please

examples/benchmark/tls_bench.c
```diff
@@ -130,8 +130,10 @@ Or
...
         #define TEST_DTLS_PACKET_SIZE   (1500)
     #else
-        /* MAX_UDP_SIZE in interna.h */
-        #define TEST_DTLS_PACKET_SIZE   (8092)
+        /* Must fit in one DTLS record: WOLFSSL_MAX_MTU (1400) - MAX_MSG_EXTRA (~102).
+         * Define WOLFSSL_DTLS_MTU and call wolfSSL_dtls_set_mtu() to use larger
+         * packets; or build with WOLFSSL_NO_DTLS_SIZE_CHECK to auto-fragment. */
+        #define TEST_DTLS_PACKET_SIZE   (1200)
     #endif
 #endif
```
dgarske, line 133: **🟠 [Medium] Embedded branch TEST_DTLS_PACKET_SIZE (1500) may still exceed WOLFSSL_MAX_MTU; fix and comment look incomplete**

This PR fixes Bug 3 by lowering the non-embedded `TEST_DTLS_PACKET_SIZE` from 8092 to 1200 so a DTLS record fits within `WOLFSSL_MAX_MTU` (confirmed = 1400 in `wolfssl/internal.h`) minus `MAX_MSG_EXTRA` (~102). However, the adjacent `BENCH_EMBEDDED` branch still defines `TEST_DTLS_PACKET_SIZE (1500)` and its comment claims `WOLFSSL_MAX_MTU` is 1500. `WOLFSSL_MAX_MTU` actually defaults to 1400. Unless an embedded config overrides `WOLFSSL_MAX_MTU` to be larger, a `BENCH_EMBEDDED` DTLS build would hit the exact same `DTLS_SIZE_ERROR (-439)` this PR is fixing, because the auto-fallback sets `argTestPacketSize = TEST_DTLS_PACKET_SIZE = 1500`, which exceeds 1400. This branch is not modified by the PR, so it is technically pre-existing, but the PR fixes only one of the two arms of the same `#ifdef` and leaves a stale/contradictory comment that a reviewer comparing the branches will trip over.

**Fix:** Confirm whether embedded configs raise WOLFSSL_MAX_MTU above 1500; if not, apply the same fit-in-MTU value (e.g. 1200) to the BENCH_EMBEDDED branch and correct its comment, so both arms of the #ifdef are consistent with the actual default MTU.

examples/benchmark/tls_bench.c
```diff
@@ -1304,27 +1306,31 @@ static int SetupSocketAndListen(int* listenFd, word32 port, int doDTLS)
...
 
-    if (listen(*listenFd, 5) != 0) {
-        fprintf(stderr, "ERROR: failed to listen\n");
+    if (bind(*listenFd, (struct sockaddr*)&servAddr,
+                                                sizeof(servAddr)) == -1) {
+        fprintf(stderr, "ERROR: failed to bind\n");
         return -1;
     }
 
+#ifdef WOLFSSL_DTLS
```
dgarske, line 1319: **🔵 [Low] No automated test exercises the fixed tls_bench DTLS (-u) path**

The PR fixes three independent DTLS-mode failures (UDP bind/listen, NULL ctx for TLS1.3 ciphers without DTLS13, oversized packet size), all reportedly verified manually. A grep shows `tls_bench` is referenced only in `examples/benchmark/include.am` (build), with no shell/CI script invoking it, so these DTLS regressions could silently return. This is a pre-existing coverage gap (the example has no test harness), not something the PR introduces, but the DTLS path is now functional and worth guarding.

**Fix:** Consider adding a small smoke test (e.g. a script that runs `tls_bench -u` for a short runtime and checks exit status) so the DTLS server/client path is exercised in CI and these fixes don't silently regress.

dgarske (commented):
## Skoll Code Review

### Posted findings
- [Medium] **Embedded branch TEST_DTLS_PACKET_SIZE (1500) may still exceed WOLFSSL_MAX_MTU; fix and comment look incomplete** — `examples/benchmark/tls_bench.c:128-138`
- [Low] **No automated test exercises the fixed tls_bench DTLS (-u) path** — `examples/benchmark/tls_bench.c:1273-1341`

examples/benchmark/tls_bench.c
miyazakh: DTLS functionality is already covered by `scripts/dtls.test` and `scripts/dtlscid.test`. A `tls_bench -u` smoke test would only verify example-tool integration, so we are not adding it since example binaries are not part of CI coverage.

dgarske (approved):

dgarske (approved):

PR #10607, opened 2026-06-05
Add EVP_PKEY encoded public key get/set compatibility functions

julek-wolfssl:
### Description

Adds the OpenSSL 3.0 encoded-public-key compatibility functions to the EVP layer and hardens the X25519/X448 setter against partial failure.

**`EVP_PKEY_set1_encoded_public_key` / `EVP_PKEY_get1_encoded_public_key`**

Implements both functions in the OpenSSL compatibility layer. The new OpenSSL 3.0 names and the deprecated `EVP_PKEY_{set1,get1}_tls_encodedpoint` names map to single shared implementations.

Supported key types:
- **EC**: uncompressed octet point (`0x04 || X || Y`), reusing `i2o_ECPublicKey` / `o2i_ECPublicKey`. `set1` also syncs the internal wolfCrypt key (`SetECKeyInternal`) so the key is usable by `EVP_PKEY_derive`, and refreshes the cached DER.
- **X25519 / X448**: raw little-endian public key (RFC 7748).

**Failure-atomicity fix for X25519/X448 `set1`**

The replacement curve25519/curve448 key is now built in a temporary and the existing key is only freed once the import succeeds, so a failed `set1` leaves the original `EVP_PKEY` intact (matching the EC branch).

### Testing

Adds `test_wolfSSL_EVP_PKEY_encoded_public_key` covering NULL handling, EC encode/decode round-trip and ECDH agreement, X25519/X448 round-trips, deprecated-name parity, and a wrong-length `set1` leaving the existing key usable (for both X25519 and X448).

### DIFF
--- a/tests/api/test_evp_pkey.c
+++ b/tests/api/test_evp_pkey.c
@@ -2864,3 +2864,242 @@ int test_wolfSSL_EVP_PKEY_x448(void)
     return EXPECT_RESULT();
 }
 
+int test_wolfSSL_EVP_PKEY_encoded_public_key(void)
+{
+    EXPECT_DECLS;
+#if defined(OPENSSL_EXTRA) && (defined(HAVE_ECC) || defined(HAVE_CURVE25519) || \
+    defined(HAVE_CURVE448))
+    /* Type-independent bad-argument handling. The deprecated tls_encodedpoint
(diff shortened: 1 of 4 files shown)


julek-wolfssl:
retest this please history lost.....

dgarske (approved):

PR #10625, opened 2026-06-05
Add SSL_CTX_add_client_custom_ext (OpenSSL-compat client custom extensions)

julek-wolfssl:
Implements the OpenSSL-compatible legacy client custom extension API
(`SSL_CTX_add_client_custom_ext`) for TLS 1.2 and below.

Application-defined extensions carry arbitrary IANA types, which cannot live
in the TLSX list (it keys every extension on a fixed 72-bit semaphore index
that an arbitrary type would overrun), so they are kept in a separate list on
the `WOLFSSL_CTX` and processed alongside the unknown-extension handling,
mirroring OpenSSL's custext.

### Commits

1. **Add `SSL_CTX_add_client_custom_ext`** — registers a `{add,free,parse}`
   method set with validation. ClientHello runs each `add_cb` and serializes
   the wire bytes; ServerHello dispatches the unknown-type case to `parse_cb`.
   Gated to non-TLS-1.3 so TLS 1.3 keeps RFC 8446 `unsupported_extension`
   behavior. Lists/buffers freed in `SSL_CtxResourceFree`/`SSL_ResourceFree`;
   `openssl/ssl.h` maps the API and `custom_ext_*_cb` typedefs.

2. **Reject unsolicited/duplicate exts and ignore on resumption** — tracks the
   custom types emitted in the ClientHello and rejects a server echo of an
   unsent type (`unsupported_extension`), suppresses parsing on a resumed
   handshake (`SSL_EXT_IGNORE_ON_RESUMPTION`), and scans for duplicate custom
   types (`DUPLICATE_TLS_EXT_E`), matching `extensions_cust.c`/`extensions.c`.

3. **Confirmed-resumption gate and flexible client methods** — gates the
   resumption ignore on the server-confirmed signal (resumption attempted AND
   the server echoed our session ID), and always offers the extension in the
   ClientHello regardless of the method's max version, so
   `wolfSSLv23_client_method()`/`wolfTLS_client_method()` offer it when
   negotiating TLS 1.2. The negotiated-version restriction remains enforced on
   the parse side.

Adds unit tests covering registration validation, TLS 1.2/1.3 and
flexible-method handshakes (add/free callbacks balance), ServerHello parse
dispatch, and unsolicited/duplicate/resumption handling. Builds with and
without `OPENSSL_EXTRA`; full API suite passes.


src/tls.c
```diff
@@ -16790,6 +17133,27 @@ int TLSX_GetRequestSize(WOLFSSL* ssl, byte msgType, word32* pLength)
...
+     * TLSX_WriteRequest. */
+    if (msgType == client_hello) {
+        word16 customSz = 0;
+        ret = TLSX_CustomExt_BuildRequest(ssl, &customSz);
+        if (ret != 0)
+            return ret;
+        if ((word32)length + customSz > (WOLFSSL_MAX_16BIT - OPAQUE16_LEN)) {
+            WOLFSSL_MSG("TLSX_GetRequestSize extensions exceed word16");
+            return BUFFER_E;
+        }
```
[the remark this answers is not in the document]
julek-wolfssl: Object will be torn down immediately after. Not adding complexity.

julek-wolfssl:
retest this please no history

dgarske (dismissed):

src/tls.c
```diff
@@ -16968,6 +16976,347 @@ static int TLSX_GetSizeWithEch(WOLFSSL* ssl, byte* semaphore, byte msgType,
...
+    XFREE(ssl->customExtSent, ssl->heap, DYNAMIC_TYPE_TLSX);
+    ssl->customExtSent = NULL;
+    ssl->customExtSentCnt = 0;
+
+    if (ssl->ctx == NULL || ssl->ctx->customExt == NULL)
+        return 0;
+
+    for (meth = ssl->ctx->customExt; meth != NULL && ret == 0;
+            meth = meth->next) {
+        const unsigned char* out = NULL;
```
SparkiDev: Extract this to a new function

SparkiDev (changes requested):

julek-wolfssl:
retest this please timeout cloning

src/tls.c
julek-wolfssl: refactored

SparkiDev (approved):

PR #10626, opened 2026-06-05
DTLS bugfixes

### LINKED ISSUE #10609: [Bug]: CID peer address update lacks a strict newer sequence-number gate
### Contact Details

lxd_dong@bupt.edu.cn

### Version

5.9.1

### Description

# CID peer address update lacks a strict newer sequence-number gate

## Summary

Verification result: `confirmed_partial`.

RFC 9146 Section 6 requires a DTLS CID receiver to replace its peer address only when all peer-address-update conditions are met. One of those conditions is that the datagram triggering the address update is "newer" than the newest datagram already received, considering both epoch and sequence number.

wolfSSL implements an important part of this requirement: a pending peer address is promoted only after the next encrypted DTLS record has been successfully deprotected. However, the promotion path does not also check whether the triggering record is newer than the newest datagram already received. A focused runtime test confirms that a valid but not-newer DTLS 1.2 CID record can still promote `pendingPeer` to the active peer address.

Scope: this finding applies when DTLS CID support is enabled and an application uses `wolfSSL_dtls_set_pending_peer()` for address migration. It is not a claim that wolfSSL's default socket receive path unconditionally accepts arbitrary source-address changes.

## Standard Requirement

Official standard: https://www.rfc-editor.org/rfc/rfc9146#section-6

Section: RFC 9146 Section 6, Peer Address Update

```text
When a record with a CID is received that has a source address
different from the one currently associated with the DTLS connection,
the receiver MUST NOT replace the address it uses for sending records
to its peer with the source address specified in the received
datagram, unless the following three conditions are met:

*  The received datagram has been cryptographically verified using
   the DTLS record layer processing procedures.

*  The received datagram is "newer" (in terms of both epoch and
   sequence number) than the newest datagram received.
```

The "newer" condition is separate from ordinary DTLS replay-window acceptance. A datagram can be a valid DTLS record and still be ineligible to trigger a peer address update.

## Relevant Source Code

`wolfssl-master/src/ssl.c:1480`

```c
int wolfSSL_dtls_set_pending_peer(WOLFSSL* ssl, void* peer, unsigned int peerSz)
{
    ...
    else {
        ret = SockAddrSet(&ssl->buffers.dtlsCtx.pendingPeer, peer, peerSz,
                ssl->heap);
    }
    if (ret == WOLFSSL_SUCCESS)
        ssl->buffers.dtlsCtx.processingPendingRecord = 0;
    ...
}
```

`wolfSSL_dtls_set_pending_peer()` stores a candidate address in `ssl->buffers.dtlsCtx.pendingPeer`; it does not store the current record epoch/sequence number or compare it with a newest-datagram state.

`wolfssl-master/src/internal.c:22899`

```c
static void dtlsProcessPendingPeer(WOLFSSL* ssl, int deprotected)
{
    if (ssl->buffers.dtlsCtx.pendingPeer.sa != NULL) {
        if (!deprotected) {
            ...
        }
        else {
            /* Pending peer present and record deprotected. Update the peer. */
            (void)wolfSSL_dtls_set_peer(ssl,
                    ssl->buffers.dtlsCtx.pendingPeer.sa,
                    ssl->buffers.dtlsCtx.pendingPeer.sz);
            ssl->buffers.dtlsCtx.processingPendingRecord = 0;
            dtlsClearPeer(&ssl->buffers.dtlsCtx.pendingPeer);
        }
    }
    ...
}
```

After deprotection, the pending peer is promoted directly with `wolfSSL_dtls_set_peer()`.

`wolfssl-master/src/internal.c:23607`

```c
if (IsDtlsNotSctpMode(ssl)) {
    DtlsUpdateWindow(ssl);
}
#ifdef WOLFSSL_DTLS_CID
/* Update the peer if we were able to de-protect the message */
if (IsEncryptionOn(ssl, 0))
    dtlsProcessPendingPeer(ssl, 1);
#endif
```

This confirms that wolfSSL has a cryptographic/deprotection gate before pending-peer promotion.

`wolfssl-master/src/internal.c:19333`

```c
if (ssl->keys.curEpoch == peerSeq->nextEpoch) {
    next_hi = peerSeq->nextSeq_hi;
    next_lo = peerSeq->nextSeq_lo;
    window = peerSeq->window;
}
else if (ssl->keys.curEpoch == peerSeq->nextEpoch - 1) {
    next_hi = peerSeq->prevSeq_hi;
    next_lo = peerSeq->prevSeq_lo;
    window = peerSeq->prevWindow;
}
else {
    return 0;
}
```

The DTLS 1.2 replay-window path can accept records from the current epoch, and also previous-epoch records when they fall within the allowed window.

## Implementation Behavior

The implementation flow is:

1. The application calls `wolfSSL_dtls_set_pending_peer()` after observing a candidate source-address change.
2. wolfSSL stores that candidate in `pendingPeer`.
3. During record processing, `dtlsProcessPendingPeer(ssl, 0)` marks that a pending peer is tied to the next encrypted record.
4. After successful deprotection, `dtlsProcessPendingPeer(ssl, 1)` promotes `pendingPeer` by calling `wolfSSL_dtls_set_peer()`.

What is missing is an address-update-specific comparison between the triggering record's epoch/sequence number and the newest datagram already received. wolfSSL relies on the normal DTLS record acceptance path, but replay-window acceptance is weaker than RFC 9146's "newer than newest" update condition.

## Inconsistency Reason

RFC 9146 requires peer-address replacement to be gated by a datagram that is both cryptographically verified and newer than the newest datagram already received.

wolfSSL satisfies the cryptographic-verification part for the pending-peer path: `pendingPeer` is promoted only after the record is successfully deprotected. The missing part is the strict newer-datagram gate. A reordered but valid record, with a sequence number older than a record already received, can still pass DTLS replay-window checks and be delivered to the application. Under RFC 9146, that record must not trigger peer-address replacement.

Therefore this is a partial inconsistency: CID parsing and deprotection gating exist, but pending-peer promotion does not enforce the separate RFC 9146 newer-than-newest condition.

## Runtime Evidence

Focused test added:

- `wolfssl-master/tests/api/test_dtls.c:525`
- `wolfssl-master/tests/api/test_dtls.h:28`
- `wolfssl-master/tests/api/test_dtls.h:88`

The test `test_wolfSSL_dtls_cid_pending_peer_not_newer()` creates this sequence:

1. Establish a DTLS 1.2 connection with CID enabled.
2. Generate two valid client-to-server application records: `old0` first, then `new1`.
3. Deliver `new1` to the server first; the server successfully reads it.
4. Set `pendingPeer` to `"peer2"`.
5. Deliver the older `old0` record; the server successfully reads it.
6. Assert that `wolfSSL_dtls_get_peer()` still reports no promoted peer, because `old0` is not newer than `new1`.

Build command:

```powershell
cmake -S .\wolfssl-master `
  -B .\test-wolfssl-dtls\rfc9146\001-022\runtime-id017-build `
  -G Ninja `
  -DCMAKE_BUILD_TYPE=Debug `
  -DCMAKE_C_COMPILER=gcc `
  -DWOLFSSL_DTLS=yes `
  -DWOLFSSL_TLS13=yes `
  -DWOLFSSL_DTLS13=yes `
  -DWOLFSSL_DTLS_CID=yes `
  -DWOLFSSL_EXAMPLES=yes `
  -DWOLFSSL_CRYPT_TESTS=no

cmake --build .\test-wolfssl-dtls\rfc9146\001-022\runtime-id017-build --target unit_test -j 4
```

Runtime command:

```powershell
$build = (Resolve-Path ..\test-wolfssl-dtls\rfc9146\001-022\runtime-id017-build).Path
$env:PATH = $build + ';' + $env:PATH
& "$build\tests\unit.test.exe" --api -test_wolfSSL_dtls_cid_pending_peer_not_newer
```

Observed result:

```text
starting unit tests...
 Begin API Tests
  Begin Group: dtls
   1409: test_wolfSSL_dtls_cid_pending_peer_not_newer       :
ERROR - D:/project/SpecTrace/wolfssl-master/tests/api/test_dtls.c line 579 failed with:
    expected: wolfSSL_dtls_get_peer(ssl_s, peer, &peerSz) == 0
    result:   1 != 0

FAILURES:
   1409: test_wolfSSL_dtls_cid_pending_peer_not_newer

 End API Tests
 Failed/Skipped/Passed/All: 1/0/0/1
```

The failure means `wolfSSL_dtls_get_peer()` returned success after the older `old0` record was processed. In other words, the older but valid record promoted `pendingPeer` to the active peer address, which RFC 9146's newer-than-newest gate should prevent.

Control test:

```powershell
& "$build\tests\unit.test.exe" --api -test_wolfSSL_dtls_set_pending_peer
```

Observed result:

```text
test_wolfSSL_dtls_set_pending_peer : passed
Failed/Skipped/Passed/All: 0/0/1/1
```

The existing positive pending-peer test passes in the same build, so the failing focused test is not caused by a broken test harness or failed CID setup.

## Impact

In deployments that enable DTLS CID and use `wolfSSL_dtls_set_pending_peer()` for address migration, a valid but not-newer record can update the active peer address. This can let reordered traffic trigger an address change that RFC 9146 intended to prevent.

The practical reachability depends on how the application detects source-address changes and when it calls `wolfSSL_dtls_set_pending_peer()`, but the runtime test demonstrates that the library's pending-peer promotion path itself does not enforce the RFC 9146 newer-than-newest requirement.

## Fix Direction

Track the newest received datagram for the CID-protected DTLS connection, including at least epoch and sequence number.

Before promoting `pendingPeer` in `dtlsProcessPendingPeer(ssl, 1)`, compare the triggering record's epoch/sequence number with that newest-datagram state. Only call `wolfSSL_dtls_set_peer()` when the record has been cryptographically verified and is strictly newer in the RFC 9146 sense.

The normal replay-window logic should continue to decide whether a record is acceptable DTLS traffic. The new gate should only decide whether that accepted record is eligible to trigger peer-address replacement.

### Relevant log output

```shell

```

### LINKED ISSUE #10610: [Bug]: CMake DTLS CID Option Incorrectly Requires DTLS 1.3
### Contact Details

lxd_dong@bupt.edu.cn

### Version

5.9.1

### Description

# CMake DTLS CID Option Incorrectly Requires DTLS 1.3

## Summary
This finding is a CMake build-configuration mismatch, not a DTLS record-layer data-path violation.

wolfSSL does implement DTLS 1.2 Connection ID support: it defines the RFC 9146 `tls12_cid(25)` content type, exposes CID APIs, negotiates the `connection_id` extension, emits DTLS 1.2 CID records, validates received CIDs, and has a dedicated DTLS 1.2 CID test. However, the CMake option `WOLFSSL_DTLS_CID` rejects configurations unless `WOLFSSL_DTLS13` is also enabled. That prevents a CMake user from enabling DTLS 1.2 CID alone, even though RFC 9146 is specifically a DTLS 1.2 CID specification and the source tree contains a DTLS 1.2 implementation path.

## Standard Requirement
Official standard: https://www.rfc-editor.org/rfc/rfc9146

RFC 9146 defines Connection Identifiers for DTLS 1.2. Section 10.3 registers the new TLS content type:

```text
Value: 25
Description: tls12_cid
DTLS-OK: Y
Reference: RFC 9146
Comment: The tls12_cid content type is only applicable to DTLS 1.2.
```

The important constraint is that `tls12_cid(25)` is a DTLS 1.2 content type. DTLS 1.3 CID uses a different record-header mechanism and should not rely on the DTLS 1.2 `tls12_cid` content type.

## Relevant Source Code
`CMakeLists.txt` requires DTLS 1.3 when enabling DTLS CID:

```cmake
CMakeLists.txt:423
if(WOLFSSL_DTLS_CID)
    if(NOT WOLFSSL_DTLS13)
        message(FATAL_ERROR "CID are supported only for DTLSv1.3")
    endif()
    list(APPEND WOLFSSL_DEFINITIONS "-DWOLFSSL_DTLS_CID")
endif()
```

The autotools configure path does not impose the same DTLS 1.3 dependency:

```m4
configure.ac:5723
AC_ARG_ENABLE([dtlscid],
    [AS_HELP_STRING([--enable-dtlscid],[Enable wolfSSL DTLS ConnectionID (default: disabled)])],
    [ ENABLED_DTLS_CID=$enableval ],
    [ ENABLED_DTLS_CID=no ]
    )
if test "x$ENABLED_DTLS_CID" = "xyes"
then
  AM_CFLAGS="$AM_CFLAGS -DWOLFSSL_DTLS_CID"
fi
```

The DTLS 1.2 record-layer constant is present:

```c
wolfssl/internal.h:6614
enum ContentType {
    no_type            = 0,
    change_cipher_spec = 20,
    alert              = 21,
    handshake          = 22,
    application_data   = 23,
    dtls12_cid         = 25,
#ifdef WOLFSSL_DTLS13
    ack                = 26,
#endif
};
```

The DTLS 1.2 send path uses `dtls12_cid` when a transmit CID is configured:

```c
src/internal.c:24488
#if defined(WOLFSSL_DTLS) && defined(WOLFSSL_DTLS_CID)
            if (ssl->options.dtls && DtlsGetCidTxSize(ssl) > 0)
                args->type = dtls12_cid;
#endif
```

The DTLS 1.2 receive path validates the received CID:

```c
src/internal.c:12205
#ifdef WOLFSSL_DTLS_CID
    if (rh->type == dtls12_cid) {
        byte* ourCid = NULL;
        if (ssl->buffers.inputBuffer.length - *inOutIdx <
                (word32)cidSz + LENGTH_SZ)
            return LENGTH_ERROR;
        if (wolfSSL_dtls_cid_get0_rx(ssl, &ourCid) != WOLFSSL_SUCCESS)
            return DTLS_CID_ERROR;
        if (XMEMCMP(ssl->buffers.inputBuffer.buffer + *inOutIdx, ourCid, cidSz)
                != 0)
            return DTLS_CID_ERROR;
        *inOutIdx += cidSz;
    }
#endif
```

There is a dedicated DTLS 1.2 CID test:

```c
tests/api/test_dtls.c:37
int test_dtls12_basic_connection_id(void)
{
#if defined(HAVE_MANUAL_MEMIO_TESTS_DEPENDENCIES) && defined(WOLFSSL_DTLS_CID)
    ...
    ExpectIntEQ(test_memio_setup(&test_ctx, &ctx_c, &ctx_s, &ssl_c,
        &ssl_s, wolfDTLSv1_2_client_method, wolfDTLSv1_2_server_method),
        0);

    ExpectIntEQ(wolfSSL_dtls_cid_use(ssl_c), 1);
    ExpectIntEQ(wolfSSL_dtls_cid_set(ssl_c, server_cid,
            sizeof(server_cid)), 1);
    ExpectIntEQ(wolfSSL_dtls_cid_use(ssl_s), 1);
    ExpectIntEQ(wolfSSL_dtls_cid_set(ssl_s, client_cid,
            sizeof(client_cid)), 1);
```

DTLS 1.3 CID uses a different unified-header CID bit path:

```c
src/dtls13.c:1163
static int Dtls13AddCID(WOLFSSL* ssl, byte* flags, byte* out, word16* idx)
{
    ...
    *flags |= DTLS13_CID_BIT;
    ...
}
```

## Implementation Behavior
The DTLS 1.2 CID implementation exists in the source tree and is not just a placeholder. The API path enables the `connection_id` extension, the send path changes the outer record type to `dtls12_cid`, the receive path checks the CID bytes, and the test suite includes a DTLS 1.2 handshake/data-flow test using `wolfDTLSv1_2_client_method` and `wolfDTLSv1_2_server_method`.

The problem is specifically the CMake enablement path: `WOLFSSL_DTLS_CID=yes` is rejected unless `WOLFSSL_DTLS13=yes`. This does not match the protocol feature's DTLS 1.2 applicability and is inconsistent with the autotools `--enable-dtlscid` path.

## Inconsistency Reason
RFC 9146 defines a DTLS 1.2 CID content type. wolfSSL's record-layer implementation supports that DTLS 1.2 content type, and DTLS 1.3 has a separate unified-header CID mechanism.

The inconsistent part is that the CMake configuration says CID is supported only for DTLS 1.3. That statement is too narrow and blocks a valid DTLS 1.2 CID-only build configuration.

This should be classified as `confirmed_partial` because:

- Implemented: DTLS 1.2 CID data-plane support exists.
- Implemented: DTLS 1.3 CID does not appear to misuse `tls12_cid(25)`.
- Missing/inconsistent: CMake cannot enable the DTLS 1.2 CID feature unless DTLS 1.3 is also enabled.

## Runtime Evidence
Focused verification script: `test-wolfssl-dtls/rfc9146/001-022/verify_wolfssl_dtls_cid_001_022.py`

Log file: `test-wolfssl-dtls/rfc9146/001-022/verify_wolfssl_dtls_cid_001_022.log`

The verification confirms:

- CMake rejects `WOLFSSL_DTLS_CID` without `WOLFSSL_DTLS13`.
- `dtls12_cid = 25` exists.
- The DTLS 1.2 sender sets `args->type = dtls12_cid`.
- The sender writes the CID and inner content type.
- The checked default build does not expose `WOLFSSL_DTLS_CID` because the option is disabled.

## Impact
CMake users cannot build a DTLS 1.2-only CID configuration through the documented CMake option, despite the source code containing DTLS 1.2 CID support. This can lead to unnecessary DTLS 1.3 enablement or to the false impression that wolfSSL has no DTLS 1.2 CID implementation.

## Fix Direction
Adjust the CMake option logic so that `WOLFSSL_DTLS_CID` requires DTLS support, not DTLS 1.3 specifically. For example, allow `WOLFSSL_DTLS_CID` when `WOLFSSL_DTLS` is enabled, and keep DTLS 1.3-specific CID checks inside DTLS 1.3-specific code paths.

If the maintainers intentionally only support the CMake CID option together with DTLS 1.3, the option text and fatal error should be renamed or clarified so it does not contradict the existing DTLS 1.2 CID implementation.

### Reproduction steps

_No response_

### Relevant log output

```shell

```

### LINKED ISSUE #10612: DTLSHandshake Body Lacks Dynamic CID Message Branches
# DTLSHandshake Body Lacks Dynamic CID Message Branches

## Summary

RFC 9147 extends the DTLS 1.3 handshake body selection with two DTLS-specific dynamic Connection ID messages: `request_connection_id` and `new_connection_id`.

wolfSSL implements the ordinary TLS 1.3 / DTLS 1.3 handshake body dispatch path. It parses DTLS handshake headers, passes the body to the TLS 1.3 handshake dispatcher, and handles common message types such as `ClientHello`, `ServerHello`, `CertificateRequest`, `Certificate`, `Finished`, `NewSessionTicket`, and `KeyUpdate`.

The dynamic CID alternatives are not implemented. There is no `request_connection_id(9)` or `new_connection_id(10)` handshake type, no `RequestConnectionId` or `NewConnectionId` body parser/serializer, and no branch in the handshake dispatcher. This confirms ID 028 as **partially satisfied**.

## Standard Requirement

RFC 9147 Section 5.7, `Handshake Protocol`, defines the `DTLSHandshake` body as a selection over `msg_type`:

```text
select (msg_type) {
    case client_hello:          ClientHello;
    case server_hello:          ServerHello;
    case end_of_early_data:     EndOfEarlyData;
    case encrypted_extensions:  EncryptedExtensions;
    case certificate_request:   CertificateRequest;
    case certificate:           Certificate;
    case certificate_verify:    CertificateVerify;
    case finished:              Finished;
    case new_session_ticket:    NewSessionTicket;
    case key_update:            KeyUpdate;
    case request_connection_id: RequestConnectionId;
    case new_connection_id:     NewConnectionId;
} body;
```

Appendix A.2 lists the DTLS 1.3 handshake message type additions:

```text
request_connection_id(9), /* New */
new_connection_id(10),    /* New */
```

Appendix A.4 defines the corresponding CID message structures:

```text
struct { uint8 num_cids; } RequestConnectionId;
```

```text
struct {
    ConnectionId cids<0..2^16-1>;
    ConnectionIdUsage usage;
} NewConnectionId;
```

The requirement is therefore not only a generic handshake parser. A complete DTLS 1.3 `body` implementation needs dispatch entries and message handling for the dynamic CID body alternatives.

### Handshake Type Table Omits Dynamic CID Types

In `D:\project\wolfssl-master\wolfssl\internal.h`, `enum HandShakeType` includes ordinary TLS and DTLS handshake values, including `key_update`, but it does not include values 9 or 10 for `request_connection_id` and `new_connection_id`:

```c
enum HandShakeType {
    hello_request        =   0,
    client_hello         =   1,
    server_hello         =   2,
    hello_verify_request =   3,    /* DTLS addition */
    session_ticket       =   4,
    end_of_early_data    =   5,
    hello_retry_request  =   6,
    encrypted_extensions =   8,
    certificate          =  11,
    server_key_exchange  =  12,
    certificate_request  =  13,
    server_hello_done    =  14,
    certificate_verify   =  15,
    client_key_exchange  =  16,
    finished             =  20,
    certificate_status   =  22,
    key_update           =  24,
    ...
};
```

Because the symbolic message types are missing, the product code has no named entry point for the RFC 9147 dynamic CID handshake bodies.

### DTLS 1.3 Handshake Framing Exists

In `D:\project\wolfssl-master\src\dtls13.c`, `_Dtls13HandshakeRecv` parses the DTLS handshake header and then forwards the message type and body to the TLS 1.3 handshake dispatcher:

```c
ret = GetDtlsHandShakeHeader(ssl, input, &idx, &handshakeType,
    &messageLength, &fragOff, &fragLength, size);
```

```c
ret = DoTls13HandShakeMsgType(ssl, input, &idx, handshakeType,
    messageLength, size);
```

This confirms the implemented part of the requirement: DTLS handshake message framing and body dispatch infrastructure exists.

### TLS 1.3 / DTLS 1.3 Dispatcher Handles Ordinary Bodies Only

In `D:\project\wolfssl-master\src\tls13.c`, `DoTls13HandShakeMsgType` dispatches ordinary TLS 1.3 body cases:

```c
case server_hello:
    WOLFSSL_MSG("processing server hello");
    ret = DoTls13ServerHello(ssl, input, inOutIdx, size, &type);
    break;

case encrypted_extensions:
    WOLFSSL_MSG("processing encrypted extensions");
    ret = DoTls13EncryptedExtensions(ssl, input, inOutIdx, size);
    break;

case certificate_request:
    WOLFSSL_MSG("processing certificate request");
    ret = DoTls13CertificateRequest(ssl, input, inOutIdx, size);
    break;

case session_ticket:
    WOLFSSL_MSG("processing new session ticket");
    ret = DoTls13NewSessionTicket(ssl, input, inOutIdx, size);
    break;
```

The same function also handles `client_hello`, `certificate`, `certificate_verify`, `finished`, and `key_update`:

```c
case key_update:
    WOLFSSL_MSG("processing key update");
    ret = DoTls13KeyUpdate(ssl, input, inOutIdx, size);
    break;
```

The default path rejects unknown handshake types:

```c
default:
    WOLFSSL_MSG("Unknown handshake message type");
    ret = UNKNOWN_HANDSHAKE_TYPE;
    break;
```

There is no `case request_connection_id` and no `case new_connection_id`.

### Static CID Support Exists, But It Is Not Dynamic CID Body Handling

wolfSSL has static DTLS CID support. In `D:\project\wolfssl-master\wolfssl\internal.h`, CID state stores one current TX CID and one current RX CID:

```c
typedef struct CIDInfo {
    ConnectionID* tx;
    ConnectionID* rx;
    byte negotiated : 1;
} CIDInfo;
```

In `D:\project\wolfssl-master\src\dtls13.c`, record-layer code can add and parse the current negotiated CID:

```c
static int Dtls13AddCID(WOLFSSL* ssl, byte* flags, byte* out, word16* idx)
```

```c
static int Dtls13UnifiedHeaderParseCID(WOLFSSL* ssl, byte flags,
    const byte* input, word16 inputSize, word16* idx)
```

However, this is record-layer handling of a current negotiated CID. It is not an implementation of `RequestConnectionId`, `NewConnectionId`, `ConnectionIdUsage`, `cid_immediate`, or `cid_spare`.

The public CID setter also explicitly rejects changing the RX CID after one has already been configured:

```c
if (cidInfo->rx != NULL) {
    WOLFSSL_MSG("wolfSSL doesn't support changing the CID during a "
                "connection");
    return WOLFSSL_FAILURE;
}
```

That behavior is consistent with static CID support, not RFC 9147 dynamic CID body handling.

### Compiled C Harness

I added and compiled a focused C source-check harness:

```text
D:\project\SpecTrace\test-wolfssl-dtls\rfc9147\001-050\repro_dynamic_cid_body_028_source_check.c
```

Build command run from `D:\project`:

```text
clang D:\project\SpecTrace\test-wolfssl-dtls\rfc9147\001-050\repro_dynamic_cid_body_028_source_check.c -o D:\project\SpecTrace\test-wolfssl-dtls\rfc9147\001-050\repro_dynamic_cid_body_028_source_check.exe
```

The executable was run and its output was saved here:

```text
D:\project\SpecTrace\test-wolfssl-dtls\rfc9147\001-050\repro_dynamic_cid_body_028_source_check.log
```

Observed output:

```text
PASS HandshakeType enum has ordinary TLS 1.3 body types
PASS HandshakeType enum lacks request_connection_id
PASS HandshakeType enum lacks new_connection_id
PASS DTLS handshake receive path forwards body to TLS13 dispatcher
PASS TLS13 dispatcher handles ordinary body cases
PASS TLS13 dispatcher has no request_connection_id branch
PASS TLS13 dispatcher has no new_connection_id branch
PASS Static DTLS CID record support exists
RESULT confirmed: ordinary DTLSHandshake body handling exists, dynamic CID body branches are absent
```

The harness is compiled and run as a native executable. It directly reads the local wolfSSL product source and verifies both implemented behavior and missing dynamic CID body branches.

### Executable Source Probe

I also ran a focused executable source probe against the local wolfSSL tree:

```text
D:\project\SpecTrace\test-wolfssl-dtls\rfc9147\001-050\focused_dynamic_cid_body_028_probe.py
```

Saved output:

```text
D:\project\SpecTrace\test-wolfssl-dtls\rfc9147\001-050\focused_dynamic_cid_body_028_probe.log
```

Observed result:

```text
HandshakeType enum exists: PASS
HandshakeType enum includes normal TLS 1.3 messages: PASS
HandshakeType enum lacks request_connection_id(9): PASS
HandshakeType enum lacks new_connection_id(10): PASS
DTLS 1.3 handshake receive path exists: PASS
TLS 1.3 handshake dispatcher handles ordinary body cases: PASS
TLS 1.3 dispatcher has no dynamic CID body branch: PASS
Repository has no RequestConnectionId implementation: PASS
Repository has no NewConnectionId implementation: PASS
Repository has no ConnectionIdUsage implementation: PASS
Static DTLS CID support exists: PASS
CID state is current tx/rx, not a dynamic CID body queue: PASS
Existing API rejects changing CID during a connection: PASS
RESULT: confirmed partial satisfaction for RFC9147 DTLSHandshake.body: normal body cases exist, request_connection_id/new_connection_id branches are absent
```

This turn did not rerun the full wolfSSL `unit.test.exe`, but it did compile and run the focused C harness above with `clang`.

## Inconsistency

| RFC 9147 `DTLSHandshake.body` component | wolfSSL behavior |
|---|---|
| Ordinary body alternatives such as `ClientHello`, `ServerHello`, `CertificateRequest`, `Certificate`, `Finished`, `NewSessionTicket`, and `KeyUpdate` | Implemented in `DoTls13HandShakeMsgType` |
| DTLS handshake header parsing and body forwarding | Implemented in `_Dtls13HandshakeRecv` |
| `request_connection_id: RequestConnectionId` | Missing handshake type, parser, serializer, and dispatcher branch |
| `new_connection_id: NewConnectionId` | Missing handshake type, parser, serializer, and dispatcher branch |
| Static negotiated CID in DTLS records | Implemented |
| Dynamic CID body/state-machine behavior | Not implemented |

The issue is therefore accurately classified as **partial**: the generic and ordinary handshake body machinery exists, but the RFC 9147 dynamic CID body alternatives are absent.

## Root Cause

wolfSSL's DTLS 1.3 implementation reuses the TLS 1.3 handshake dispatcher and adds DTLS handshake framing around it. The implementation also has static CID record-layer support.

RFC 9147's dynamic CID messages require additional handshake type values, body parsing, body serialization, state validation, CID list management, and future-record CID selection. Those pieces are not present in the source tree. Unknown dynamic CID handshake message types would fall through to `UNKNOWN_HANDSHAKE_TYPE`.

## Impact

A peer that sends RFC 9147 dynamic CID handshake messages cannot interoperate with wolfSSL on this path. Ordinary DTLS 1.3 handshakes and static CID record handling may still work, but dynamic CID update messages will not be parsed or processed.

For ID 028 specifically, this means the `body` selection in RFC 9147 is not complete in wolfSSL: common body alternatives are implemented, while `request_connection_id` and `new_connection_id` are not.

## Suggested Fix

To fully satisfy the RFC 9147 body selection, wolfSSL would need to add:

| Required change | Expected effect |
|---|---|
| `request_connection_id(9)` and `new_connection_id(10)` in `enum HandShakeType` | The DTLS 1.3 message type table matches RFC 9147 |
| `RequestConnectionId` parser and serializer | `request_connection_id` body can be received and sent |
| `NewConnectionId` parser and serializer | `new_connection_id` body can be received and sent |
| `ConnectionIdUsage` support | `cid_immediate` and `cid_spare` semantics can be represented |
| Dispatcher branches in `DoTls13HandShakeMsgType` | Dynamic CID messages are routed to implementation code |
| State-machine and ACK/retransmission tests | Dynamic CID messages are covered as DTLS 1.3 handshake messages |

### LINKED ISSUE #10613: DTLS 1.3 Dynamic Connection ID Update Semantics Are Missing
# DTLS 1.3 Dynamic Connection ID Update Semantics Are Missing

## Summary

RFC 9147 defines dynamic DTLS 1.3 Connection ID update messages. A peer can send `NewConnectionId` with `ConnectionIdUsage = cid_immediate` or `cid_spare`, and endpoints should maintain receiver-provided CIDs in the order they were provided.

wolfSSL has static DTLS CID support: it can negotiate a CID, store one current TX/RX CID pair, add the current CID to DTLS 1.3 unified record headers, and validate the current RX CID. It does not implement the RFC 9147 dynamic CID update messages or the required CID list/queue semantics.

This confirms IDs 031, 032, and 033 as **not satisfied**.

## Standard Requirement

RFC 9147 Section 9, `Connection ID Updates`, defines the dynamic CID messages:

```text
enum { cid_immediate(0), cid_spare(1), (255) } ConnectionIdUsage;
opaque ConnectionId<0..2^8-1>;
struct {
    ConnectionId cids<0..2^16-1>;
    ConnectionIdUsage usage;
} NewConnectionId;
```

It defines the `cid_immediate` behavior:

```text
If usage is set to "cid_immediate", then one of the new CIDs MUST be used immediately for all future records.
```

It defines spare CID handling:

```text
If it is set to "cid_spare", then either an existing or new CID MAY be used.
```

It also defines receiver-provided CID ordering and excess spare-CID handling:

```text
Endpoints SHOULD use receiver-provided CIDs in the order they were provided.
```

```text
Implementations which receive more spare CIDs than they wish to maintain MAY simply discard any extra CIDs.
```

Appendix A.2 and A.4 list the corresponding handshake message types and structures:

```text
request_connection_id(9), /* New */
new_connection_id(10),    /* New */
```

```text
struct { uint8 num_cids; } RequestConnectionId;
```

These requirements depend on parsing and serializing `RequestConnectionId` / `NewConnectionId`, storing multiple receiver-provided CIDs, and selecting future record CIDs according to `ConnectionIdUsage`.

### Handshake Types Do Not Include Dynamic CID Messages

In `wolfssl/internal.h`, `enum HandShakeType` includes normal TLS/DTLS handshake types through `key_update`, but not `request_connection_id(9)` or `new_connection_id(10)`:

```c
enum HandShakeType {
    hello_request        =   0,
    client_hello         =   1,
    server_hello         =   2,
    hello_verify_request =   3,    /* DTLS addition */
    session_ticket       =   4,
    end_of_early_data    =   5,
    hello_retry_request  =   6,
    encrypted_extensions =   8,
    certificate          =  11,
    ...
    key_update           =  24,
    ...
};
```

This means the product handshake type table lacks the RFC 9147 dynamic CID handshake message identifiers.

### TLS 1.3 / DTLS 1.3 Dispatch Has No Dynamic CID Branch

In `src/tls13.c`, the handshake validation/dispatch logic has `case key_update`, but no `case request_connection_id` or `case new_connection_id` branch:

```c
case key_update:
    /* Valid on both sides. */
    ...
    break;
```

The focused probe also searched the product source and tests for `RequestConnectionId`, `NewConnectionId`, `ConnectionIdUsage`, `cid_immediate`, and `cid_spare`; no dynamic CID message implementation was found.

### CID State Stores Only Current TX/RX CID

In `wolfssl/internal.h`, `CIDInfo` stores only one current TX CID, one current RX CID, and a negotiated flag:

```c
typedef struct CIDInfo {
    ConnectionID* tx;
    ConnectionID* rx;
    byte negotiated : 1;
} CIDInfo;
```

There is no spare-CID list, no receiver-provided CID queue, no usage metadata, and no outstanding `NewConnectionId` / `RequestConnectionId` state.

### Existing API Explicitly Rejects In-Connection CID Changes

In `src/dtls.c`, the extension parse path rejects changing the CID on rehandshake:

```c
/* For now we don't support changing the CID on a rehandshake */
if (cidSz != info->tx->length ||
        XMEMCMP(info->tx->id, input + OPAQUE8_LEN, cidSz) != 0)
    return DTLS_CID_ERROR;
```

The public setter also rejects changing the RX CID during a connection:

```c
if (cidInfo->rx != NULL) {
    WOLFSSL_MSG("wolfSSL doesn't support changing the CID during a "
                "connection");
    return WOLFSSL_FAILURE;
}
```

These checks are the opposite of dynamic CID update support.

### DTLS 1.3 Record Layer Supports Static Current CID

In `src/dtls13.c`, `Dtls13AddCID` can add the current negotiated TX CID to the unified header:

```c
static int Dtls13AddCID(WOLFSSL* ssl, byte* flags, byte* out, word16* idx)
```

and `Dtls13UnifiedHeaderParseCID` validates the current RX CID:

```c
static int Dtls13UnifiedHeaderParseCID(WOLFSSL* ssl, byte flags,
    const byte* input, word16 inputSize, word16* idx)
```

This confirms that wolfSSL implements static CID record handling, but not the dynamic RFC 9147 update messages required for IDs 031-033.

### Compiled C Harness

I added and compiled a focused C source-check harness:

```text
D:\project\SpecTrace\test-wolfssl-dtls\rfc9147\001-050\repro_dynamic_cid_031_033_source_check.c
```

Build command run from `D:\project`:

```text
clang D:\project\SpecTrace\test-wolfssl-dtls\rfc9147\001-050\repro_dynamic_cid_031_033_source_check.c -o D:\project\SpecTrace\test-wolfssl-dtls\rfc9147\001-050\repro_dynamic_cid_031_033_source_check.exe
```

The executable was run and its output was saved here:

```text
D:\project\SpecTrace\test-wolfssl-dtls\rfc9147\001-050\repro_dynamic_cid_031_033_source_check.log
```

Observed output:

```text
PASS CIDInfo stores only current tx/rx CID pointers
PASS Handshake enum lacks request_connection_id
PASS Handshake enum lacks new_connection_id
PASS No RequestConnectionId implementation in checked protocol files
PASS No NewConnectionId implementation in checked protocol files
PASS No ConnectionIdUsage/cid_immediate/cid_spare symbols
PASS Existing API rejects changing CID during a connection
PASS DTLS extension parser rejects changing CID on rehandshake
PASS Static DTLS 1.3 CID record support exists
RESULT confirmed: static CID support exists, dynamic RFC9147 CID update messages and queues are absent
```

The harness is compiled and run as a native executable. It directly reads the local wolfSSL product source and verifies that static CID support exists while the dynamic CID update messages and queue/usage symbols are absent.

### Executable Source Probe

I also ran a focused executable Python probe against the local source tree:

```text
D:\project\SpecTrace\test-wolfssl-dtls\rfc9147\001-050\focused_dynamic_cid_031_033_probe.py
```

Saved log:

```text
D:\project\SpecTrace\test-wolfssl-dtls\rfc9147\001-050\focused_dynamic_cid_031_033_probe.log
```

Observed output:

```text
CIDInfo stores only tx/rx current CIDs: PASS
Handshake enum has no request_connection_id: PASS
Handshake enum has no new_connection_id: PASS
TLS13 handshake dispatcher has no RequestConnectionId/NewConnectionId branch: PASS
No ConnectionIdUsage symbols: PASS
No cid_immediate symbol: PASS
No cid_spare symbol: PASS
No too_many_cids_requested alert symbol: PASS
DTLS CID API rejects changing CID during a connection: PASS
DTLS extension parser rejects changing CID on rehandshake: PASS
DTLS 1.3 record layer can add current negotiated CID: PASS
DTLS 1.3 record layer validates current RX CID: PASS
RESULT: confirmed static CID support exists, dynamic RFC9147 CID update messages and queues are absent
```

Existing saved wolfSSL unit-test evidence in this directory shows that static CID tests pass:

```text
D:\project\SpecTrace\test-wolfssl-dtls\rfc9147\001-050\phase2_wolfssl_builtin_dtls13_tests.log
```

Relevant excerpt:

```text
test_dtls13_basic_connection_id : passed
test_wolfSSL_dtls_cid_parse     : passed
```

This turn did not rerun the full wolfSSL `unit.test.exe`, but it did compile and run the focused C harness above with `clang`.

## Inconsistency

| RFC 9147 requirement component | wolfSSL behavior |
|---|---|
| `NewConnectionId` message with `ConnectionIdUsage` | No handshake type, parser, serializer, or dispatcher found |
| `cid_immediate` must switch future records to one of the new CIDs immediately | No `cid_immediate` symbol or CID switch state exists |
| `cid_spare` supports existing/new spare CID use | No spare CID list or usage metadata exists |
| Receiver-provided CIDs should be used in provided order | No receiver-provided CID queue exists |
| Extra spare CIDs may be discarded by local policy | No dynamic spare-CID receive path exists |
| Static negotiated CID in record headers | Implemented |

## Root Cause

wolfSSL implements static DTLS CID negotiation and current-CID record-layer processing, but not RFC 9147 dynamic CID update messages.

The product state model has only current TX/RX CID pointers. It lacks `RequestConnectionId`, `NewConnectionId`, `ConnectionIdUsage`, spare CID storage, receiver-provided CID ordering, and future-record CID migration logic.

## Impact

Peers that rely on RFC 9147 dynamic CID update behavior will not interoperate with wolfSSL for these paths:

| ID | Missing behavior | Impact |
|---:|---|---|
| 031 | `cid_immediate` immediate switch | Future records cannot be migrated to a newly provided CID as required |
| 032 | spare CID list management | Spare CIDs cannot be retained, selected, or discarded by policy |
| 033 | receiver-provided CID queue | CIDs cannot be used in the order provided by the peer |

This is a protocol feature-completeness gap, not evidence of memory corruption.

## Suggested Fix

To satisfy RFC 9147, wolfSSL would need to add:

| Required change | Expected effect |
|---|---|
| `request_connection_id(9)` and `new_connection_id(10)` handshake types | Dynamic CID messages can be parsed and serialized |
| `ConnectionIdUsage` parser and serializer | `cid_immediate` and `cid_spare` semantics can be represented |
| Receiver-provided CID queue | CIDs can be used in peer-provided order |
| Spare CID list management | Extra spare CIDs can be retained or discarded by policy |
| Record-layer CID switch logic | `cid_immediate` can affect all future records |
| Regression tests | Static CID and dynamic CID update behavior are separated and covered |

### LINKED ISSUE #10621: NewConnectionId cid_immediate Usage Is Not Implemented
# NewConnectionId cid_immediate Usage Is Not Implemented

## Summary

This is a confirmed unsatisfied DTLS 1.3 compliance finding in wolfSSL. RFC 9147 Section 9 defines `NewConnectionId` with `usage = cid_immediate`, which requires an endpoint to switch to one of the newly provided CIDs immediately for future records. The audited wolfSSL tree supports static CID negotiation and record-header CID use, but it does not implement `NewConnectionId` or any immediate CID-switch state machine.

## Standard Requirement

Official standard: https://www.rfc-editor.org/rfc/rfc9147

Local standard text: `document/dtls/RFC9147.txt:2227`

Section: RFC 9147 Section 9, Connection ID Updates

```text
enum {
    cid_immediate(0), cid_spare(1), (255)
} ConnectionIdUsage;
```

```text
usage:  Indicates whether the new CIDs should be used immediately or
   are spare.  If usage is set to "cid_immediate", then one of the
   new CIDs MUST be used immediately for all future records.
```

The requirement is an active state transition. After receiving `NewConnectionId(usage = cid_immediate)`, the endpoint must select one of the newly supplied CIDs and use it for all future outbound records.

## Relevant Source Code

`wolfssl-master/wolfssl/internal.h:6753`

```c
enum HandShakeType {
    hello_request        =   0,
    client_hello         =   1,
    server_hello         =   2,
    hello_verify_request =   3,
    session_ticket       =   4,
    end_of_early_data    =   5,
    hello_retry_request  =   6,
    encrypted_extensions =   8,
    certificate          =  11,
```

The RFC 9147 handshake type `new_connection_id = 10` is absent, so wolfSSL has no named message type for the message that carries `cid_immediate`.

`wolfssl-master/src/tls13.c:13527`

```c
if ((ret = SanityCheckTls13MsgReceived(ssl, type)) != 0) {
    WOLFSSL_MSG("Sanity Check on handshake message type received failed");
    if (ret == WC_NO_ERR_TRACE(VERSION_ERROR))
        SendAlert(ssl, alert_fatal, wolfssl_alert_protocol_version);
    else
        SendAlert(ssl, alert_fatal, unexpected_message);
    return ret;
}
```

`wolfssl-master/src/tls13.c:13488`

```c
default:
    WOLFSSL_MSG("Unknown message type");
    WOLFSSL_ERROR_VERBOSE(SANITY_MSG_E);
    return SANITY_MSG_E;
```

An unknown handshake type fails the TLS 1.3 sanity check before any dynamic CID handler can run.

`wolfssl-master/src/dtls13.c:1178`

```c
cidSz = DtlsGetCidTxSize(ssl);

/* no cid */
if (cidSz == 0)
    return 0;
*flags |= DTLS13_CID_BIT;
ret = wolfSSL_dtls_cid_get_tx(ssl, out + *idx, cidSz);
```

Outgoing records use the currently configured transmit CID. There is no code path here that switches to a CID received in a `NewConnectionId` message.

`wolfssl-master/src/dtls.c:1382`

```c
if (cidInfo->rx != NULL) {
    WOLFSSL_MSG("wolfSSL doesn't support changing the CID during a "
                "connection");
    return WOLFSSL_FAILURE;
}
```

The public CID API is also static in practice and rejects changing the configured receive CID during a connection.

## Implementation Behavior

wolfSSL can put an already configured CID into outgoing DTLS 1.3 records. It can also validate an incoming record's CID against the configured receive CID. However, there is no `NewConnectionId` parser, no `ConnectionIdUsage` enum, and no branch that applies `cid_immediate` by replacing the transmit CID used for future records.

## Inconsistency Reason

RFC 9147 requires immediate CID activation when `NewConnectionId.usage` is `cid_immediate`. wolfSSL has no implementation path that can receive that message, read the `usage` value, select a new CID from the message body, and switch future records to that CID. Static CID header support is therefore insufficient to satisfy this requirement.

## Runtime Evidence

Focused verification script: `test-wolfssl-dtls/rfc9147/151-187/verify_wolfssl_dtls13_151_187.ps1`

Log: `test-wolfssl-dtls/rfc9147/151-187/verify_wolfssl_dtls13_151_187.log`

Command executed:

```text
powershell -NoProfile -ExecutionPolicy Bypass -File test-wolfssl-dtls/rfc9147/151-187/verify_wolfssl_dtls13_151_187.ps1
```

Key runtime output:

```text
== Dynamic CID message symbol scan ==
ABSENT NewConnectionId
ABSENT ConnectionIdUsage
ABSENT cid_immediate
ABSENT cid_spare
ABSENT num_cids

== Static CID positive-control scans ==
PASS DTLS 1.3 unified header adds CID
wolfssl-master/src/dtls13.c:1183:    *flags |= DTLS13_CID_BIT;
PASS DTLS 1.3 unified header checks CID
wolfssl-master/src/dtls13.c:1204:        if (!DtlsCIDCheck(ssl, input + *idx, inputSize - *idx)) {
PASS CID API rejects in-connection CID changes
wolfssl-master/src/dtls.c:1383:        WOLFSSL_MSG("wolfSSL doesn't support changing the CID during a "

== Dispatcher checks ==
PASS dispatcher unknown-message fallback returns UNKNOWN_HANDSHAKE_TYPE
ABSENT case new_connection_id

== Decision ==
PASS dynamic CID update support is absent in wolfssl-master: the static CID extension/header paths are present, but no RequestConnectionId/NewConnectionId message types, ConnectionIdUsage values, num_cids handling, or cid_spare/cid_immediate state machine were found.
```

## Impact

A peer that sends `NewConnectionId(usage = cid_immediate)` cannot force wolfSSL to switch to a fresh CID for future records. This breaks the dynamic CID rotation behavior RFC 9147 expects for path changes, privacy refresh, and peer-directed CID management.

## Fix Direction

Add `new_connection_id = 10` to the DTLS 1.3 handshake type handling, parse `NewConnectionId`, define `ConnectionIdUsage`, and implement the `cid_immediate` transition by replacing the transmit CID used by the record layer for all subsequent records. Add tests that inject `NewConnectionId(usage = cid_immediate)` and assert that the next outbound DTLS 1.3 records use one of the newly received CIDs.

mattia-moffa:
# Description

- 4b7e8cc: Previously, when DTLS CID was enabled and an app used `wolfSSL_dtls_set_pending_peer()`, `dtlsProcessPendingPeer()` promoted the pending peer without checking that the record is newer than the newest already received. RFC 9146 § 6 requires to check this (second condition in the list). Without the check, a reordered or replayed CID record could cause an erroneous migration of the active peer. Fixes #10609
- d71c3b7: Allow `WOLFSSL_DTLS_CID` with `WOLFSSL_DTLS` (not just `WOLFSSL_DTLS13`). Fixes #10610
- f495200: implement receive-side handling of DTLS 1.3 new_connection_id and request_connection_id messages.
  - Fixes #10612;
  - Fixes #10613;
  - Fixes #10621.

# Testing

```
./configure --enable-all && make && make check
```

# Checklist

 - [x] added tests

### DIFF
--- a/src/dtls.c
+++ b/src/dtls.c
@@ -1417,6 +1417,9 @@ int DtlsCidReplaceTx(WOLFSSL* ssl, const byte* cid, byte size)
     if (ssl == NULL || cid == NULL || size == 0)
         return BAD_FUNC_ARG;
 
+    if (size > DTLS_CID_MAX_SIZE)
+        return LENGTH_ERROR;
+
     cidInfo = DtlsCidGetInfo(ssl);
     if (cidInfo == NULL)
         return BAD_STATE_E;
--- a/src/dtls.c
+++ b/src/dtls.c
@@ -1426,6 +1426,7 @@ int DtlsCidReplaceTx(WOLFSSL* ssl, const byte* cid, byte size)
         return MEMORY_ERROR;
     XFREE(cidInfo->tx, ssl->heap, DYNAMIC_TYPE_TLSX);
     cidInfo->tx = newCid;
+    ssl->recordSzOverhead = 0;
     return 0;
 }
 
--- a/src/dtls.c
+++ b/src/dtls.c
@@ -1407,6 +1407,28 @@ int wolfSSL_dtls_cid_set(WOLFSSL* ssl, unsigned char* cid, unsigned int size)
     return WOLFSSL_SUCCESS;
 }
 
+/* Replace the CID we use when sending records, after the peer provided a new
+ * one in a NewConnectionId message (RFC 9147 Section 9). */
+int DtlsCidReplaceTx(WOLFSSL* ssl, const byte* cid, byte size)
+{
+    CIDInfo* cidInfo;
+    ConnectionID* newCid;
+
+    if (ssl == NULL || cid == NULL || size == 0)
+        return BAD_FUNC_ARG;
+
+    cidInfo = DtlsCidGetInfo(ssl);
+    if (cidInfo == NULL)
+        return BAD_STATE_E;
+
+    newCid = DtlsCidNew(cid, size, ssl->heap);
+    if (newCid == NULL)
+        return MEMORY_ERROR;
+    XFREE(cidInfo->tx, ssl->heap, DYNAMIC_TYPE_TLSX);
+    cidInfo->tx = newCid;
+    return 0;
+}
+
 int wolfSSL_dtls_cid_get_rx_size(WOLFSSL* ssl, unsigned int* size)
 {
     return DtlsCidGetSize(ssl, size, 1);
--- a/tests/api/test_dtls.c
+++ b/tests/api/test_dtls.c
@@ -680,6 +680,319 @@ int test_wolfSSL_dtls_set_pending_peer_not_newest(void)
     return EXPECT_RESULT();
 }
 
+#if defined(HAVE_MANUAL_MEMIO_TESTS_DEPENDENCIES) && \
+    defined(WOLFSSL_DTLS13) && defined(WOLFSSL_DTLS_CID)
+/* Set up a DTLS 1.3 connection, optionally negotiating CID, and drain all
+ * post-handshake messages (session tickets, ACKs) so both buffers are empty. */
+static int test_dtls13_cid_setup(struct test_memio_ctx* test_ctx,
+        WOLFSSL_CTX** ctx_c, WOLFSSL_CTX** ctx_s, WOLFSSL** ssl_c,
+        WOLFSSL** ssl_s, int useCid)
+{
+    EXPECT_DECLS;
+    unsigned char client_cid[] = { 9, 8, 7, 6, 5, 4, 3, 2, 1, 0 };
+    unsigned char server_cid[] = { 0, 1, 2, 3, 4, 5, 6, 7, 8, 9 };
+    byte readBuf[16];
+    int i;
+
+    XMEMSET(test_ctx, 0, sizeof(*test_ctx));
+
+    ExpectIntEQ(test_memio_setup(test_ctx, ctx_c, ctx_s, ssl_c, ssl_s,
+            wolfDTLSv1_3_client_method, wolfDTLSv1_3_server_method), 0);
+
+    if (useCid) {
+        ExpectIntEQ(wolfSSL_dtls_cid_use(*ssl_c), 1);
+        ExpectIntEQ(wolfSSL_dtls_cid_set(*ssl_c, server_cid,
+                sizeof(server_cid)), 1);
+        ExpectIntEQ(wolfSSL_dtls_cid_use(*ssl_s), 1);
+        ExpectIntEQ(wolfSSL_dtls_cid_set(*ssl_s, client_cid,
+                sizeof(client_cid)), 1);
+    }
+
+    ExpectIntEQ(test_memio_do_handshake(*ssl_c, *ssl_s, 10, NULL), 0);
+
+    for (i = 0; i < 5 && EXPECT_SUCCESS() &&
+            (test_ctx->c_len > 0 || test_ctx->s_len > 0); i++) {
+        if (test_ctx->c_len > 0)
+            ExpectIntEQ(wolfSSL_read(*ssl_c, readBuf, sizeof(readBuf)), -1);
+        if (test_ctx->s_len > 0)
+            ExpectIntEQ(wolfSSL_read(*ssl_s, readBuf, sizeof(readBuf)), -1);
+    }
+    ExpectIntEQ(test_ctx->c_len, 0);
+    ExpectIntEQ(test_ctx->s_len, 0);
+
+    return EXPECT_RESULT();
+}
+
+/* Build a post-handshake handshake message of type hsType with the given body
+ * and encrypt it with the client's current traffic keys. The message sequence
+ * number is the one the server expects next. */
+static int test_dtls13_build_post_hs_msg(WOLFSSL* ssl_c, WOLFSSL* ssl_s,
+        byte hsType, const byte* body, word16 bodyLen, byte* rec, int* recSz)
+{
+    EXPECT_DECLS;
+    byte msg[64];
+    size_t idx = 0;
+
+    ExpectIntLE(DTLS_HANDSHAKE_HEADER_SZ + bodyLen, sizeof(msg));
+
+    msg[idx++] = hsType;
+    c32to24(bodyLen, msg + idx);
+    idx += 3;
+    c16toa(ssl_s->keys.dtls_expected_peer_handshake_number, msg + idx);
+    idx += 2;
+    /* fragment_offset */
+    c32to24(0, msg + idx);
+    idx += 3;
+    /* fragment_length */
+    c32to24(bodyLen, msg + idx);
+    idx += 3;
+    XMEMCPY(msg + idx, body, bodyLen);
+    idx += bodyLen;
+
+    ExpectIntGT(*recSz = BuildTls13Message(ssl_c, rec, *recSz, msg, (int)idx,
+            handshake, 0, 0, 0), 0);
+
+    return EXPECT_RESULT();
+}
+
+/* Inject a crafted post-handshake message into the server, expect it to fail
+ * with expErr and to alert the client with a fatal alert of type expAlert. */
+static int test_dtls13_post_hs_cid_msg_err(byte hsType, const byte* body,
+        word16 bodyLen, int useCid, int expErr, int expAlert)
+{
+    EXPECT_DECLS;
+    WOLFSSL_CTX *ctx_c = NULL, *ctx_s = NULL;
+    WOLFSSL *ssl_c = NULL, *ssl_s = NULL;
+    struct test_memio_ctx test_ctx;
+    WOLFSSL_ALERT_HISTORY h;
+    byte rec[256];
+    int recSz = (int)sizeof(rec);
+    byte readBuf[16];
+
+    ExpectIntEQ(test_dtls13_cid_setup(&test_ctx, &ctx_c, &ctx_s, &ssl_c,
+            &ssl_s, useCid), TEST_SUCCESS);
+    ExpectIntEQ(test_dtls13_build_post_hs_msg(ssl_c, ssl_s, hsType, body,
+            bodyLen, rec, &recSz), TEST_SUCCESS);
+    ExpectIntEQ(test_memio_inject_message(&test_ctx, 0, (const char*)rec,
+            recSz), 0);
+
+    ExpectIntEQ(wolfSSL_read(ssl_s, readBuf, sizeof(readBuf)), -1);
+    ExpectIntEQ(wolfSSL_get_error(ssl_s, -1), expErr);
+
+    /* the server must send a fatal alert */
+    ExpectIntEQ(wolfSSL_read(ssl_c, readBuf, sizeof(readBuf)), -1);
+    XMEMSET(&h, 0, sizeof(h));
+    ExpectIntEQ(wolfSSL_get_alert_history(ssl_c, &h), WOLFSSL_SUCCESS);
+    ExpectIntEQ(h.last_rx.level, alert_fatal);
+    ExpectIntEQ(h.last_rx.code, expAlert);
+
+    wolfSSL_free(ssl_s);
+    wolfSSL_free(ssl_c);
+    wolfSSL_CTX_free(ctx_s);
+    wolfSSL_CTX_free(ctx_c);
+
+    return EXPECT_RESULT();
+}
+#endif /* HAVE_MANUAL_MEMIO_TESTS_DEPENDENCIES && WOLFSSL_DTLS13 &&
+        * WOLFSSL_DTLS_CID */
+
+int test_dtls13_new_connection_id(void)
+{
+    EXPECT_DECLS;
+#if defined(HAVE_MANUAL_MEMIO_TESTS_DEPENDENCIES) && \
+    defined(WOLFSSL_DTLS13) && defined(WOLFSSL_DTLS_CID)
+    WOLFSSL_CTX *ctx_c = NULL, *ctx_s = NULL;
+    WOLFSSL *ssl_c = NULL, *ssl_s = NULL;
+    struct test_memio_ctx test_ctx;
+    unsigned char server_cid[] = { 0, 1, 2, 3, 4, 5, 6, 7, 8, 9 };
+    /* one 3-byte CID, usage cid_spare */
+    const byte spareBody[] = { 0x00, 0x04, 0x03, 0xaa, 0xbb, 0xcc, 0x01 };
+    /* one zero-length CID followed by a 3-byte CID, usage cid_immediate */
+    const byte immBody[] = { 0x00, 0x05, 0x00, 0x03, 0x11, 0x22, 0x33, 0x00 };
+    const byte newCid[] = { 0x11, 0x22, 0x33 };
+    unsigned char cidBuf[DTLS_CID_MAX_SIZE];
+    unsigned int cidSz = 0;
+    byte rec[256];
+    int recSz = (int)sizeof(rec);
+    byte readBuf[16];
+    const byte* parsedCid = NULL;
+
+    /* usage cid_spare: the new CID may be discarded, ours must not change */
+    ExpectIntEQ(test_dtls13_cid_setup(&test_ctx, &ctx_c, &ctx_s, &ssl_c,
+            &ssl_s, 1), TEST_SUCCESS);
+    ExpectIntEQ(test_dtls13_build_post_hs_msg(ssl_c, ssl_s, new_connection_id,
+            spareBody, sizeof(spareBody), rec, &recSz), TEST_SUCCESS);
+    ExpectIntEQ(test_memio_inject_message(&test_ctx, 0, (const char*)rec,
+            recSz), 0);
+    ExpectIntEQ(wolfSSL_read(ssl_s, readBuf, sizeof(readBuf)), -1);
+    ExpectIntEQ(wolfSSL_get_error(ssl_s, -1), WOLFSSL_ERROR_WANT_READ);
+    /* the server ACKed the message */
+    ExpectIntGT(test_ctx.c_len, 0);
+    ExpectIntEQ(wolfSSL_dtls_cid_get_tx_size(ssl_s, &cidSz), 1);
+    ExpectIntEQ(cidSz, sizeof(server_cid));
+    ExpectIntEQ(wolfSSL_dtls_cid_get_tx(ssl_s, cidBuf, sizeof(cidBuf)), 1);
+    ExpectBufEQ(cidBuf, server_cid, sizeof(server_cid));
+
+    /* duplicate delivery: the retransmitted record must be dropped */
+    ExpectIntEQ(test_memio_inject_message(&test_ctx, 0, (const char*)rec,
+            recSz), 0);
+    ExpectIntEQ(wolfSSL_read(ssl_s, readBuf, sizeof(readBuf)), -1);
+    ExpectIntEQ(wolfSSL_get_error(ssl_s, -1), WOLFSSL_ERROR_WANT_READ);
+
+    /* connection must still work in both directions */
+    ExpectIntEQ(wolfSSL_read(ssl_c, readBuf, sizeof(readBuf)), -1);
+    ExpectIntEQ(wolfSSL_get_error(ssl_c, -1), WOLFSSL_ERROR_WANT_READ);
+    ExpectIntEQ(wolfSSL_write(ssl_c, "test", 5), 5);
+    ExpectIntEQ(wolfSSL_read(ssl_s, readBuf, sizeof(readBuf)), 5);
+    ExpectStrEQ(readBuf, "test");
+    ExpectIntEQ(wolfSSL_write(ssl_s, "back", 5), 5);
+    ExpectIntEQ(wolfSSL_read(ssl_c, readBuf, sizeof(readBuf)), 5);
+    ExpectStrEQ(readBuf, "back");
+
+    wolfSSL_free(ssl_s);
+    wolfSSL_free(ssl_c);
+    wolfSSL_CTX_free(ctx_s);
+    wolfSSL_CTX_free(ctx_c);
+    ssl_c = ssl_s = NULL;
+    ctx_c = ctx_s = NULL;
+
+    /* usage cid_immediate: the first non-empty CID must be used for all
+     * records sent from now on, including the ACK of this message */
+    ExpectIntEQ(test_dtls13_cid_setup(&test_ctx, &ctx_c, &ctx_s, &ssl_c,
+            &ssl_s, 1), TEST_SUCCESS);
+    recSz = (int)sizeof(rec);
+    ExpectIntEQ(test_dtls13_build_post_hs_msg(ssl_c, ssl_s, new_connection_id,
+            immBody, sizeof(immBody), rec, &recSz), TEST_SUCCESS);
+    ExpectIntEQ(test_memio_inject_message(&test_ctx, 0, (const char*)rec,
+            recSz), 0);
+    ExpectIntEQ(wolfSSL_read(ssl_s, readBuf, sizeof(readBuf)), -1);
+    ExpectIntEQ(wolfSSL_get_error(ssl_s, -1), WOLFSSL_ERROR_WANT_READ);
+    cidSz = 0;
+    ExpectIntEQ(wolfSSL_dtls_cid_get_tx_size(ssl_s, &cidSz), 1);
+    ExpectIntEQ(cidSz, sizeof(newCid));
+    ExpectIntEQ(wolfSSL_dtls_cid_get_tx(ssl_s, cidBuf, sizeof(cidBuf)), 1);
+    ExpectBufEQ(cidBuf, newCid, sizeof(newCid));
+    /* the ACK record already carries the new CID */
+    ExpectIntGT(test_ctx.c_len, 0);
+    ExpectNotNull(parsedCid = wolfSSL_dtls_cid_parse(test_ctx.c_buff,
+            (unsigned int)test_ctx.c_len, sizeof(newCid)));
+    if (parsedCid != NULL)
+        ExpectBufEQ(parsedCid, newCid, sizeof(newCid));
+
+    wolfSSL_free(ssl_s);
+    wolfSSL_free(ssl_c);
+    wolfSSL_CTX_free(ctx_s);
+    wolfSSL_CTX_free(ctx_c);
+#endif
+    return EXPECT_RESULT();
+}
+
+int test_dtls13_new_connection_id_not_negotiated(void)
+{
+    EXPECT_DECLS;
+#if defined(HAVE_MANUAL_MEMIO_TESTS_DEPENDENCIES) && \
+    defined(WOLFSSL_DTLS13) && defined(WOLFSSL_DTLS_CID)
+    /* one 3-byte CID, usage cid_spare */
+    const byte body[] = { 0x00, 0x04, 0x03, 0xaa, 0xbb, 0xcc, 0x01 };
+
+    /* RFC 9147 Section 9: receiving a CID message when CIDs were not
+     * negotiated must be answered with an unexpected_message alert */
+    ExpectIntEQ(test_dtls13_post_hs_cid_msg_err(new_connection_id, body,
+            sizeof(body), 0, WC_NO_ERR_TRACE(SANITY_MSG_E),
+            unexpected_message), TEST_SUCCESS);
+    ExpectIntEQ(test_dtls13_post_hs_cid_msg_err(request_connection_id,
+            (const byte*)"\x02", 1, 0, WC_NO_ERR_TRACE(SANITY_MSG_E),
+            unexpected_message), TEST_SUCCESS);
+#endif
+    return EXPECT_RESULT();
+}
+
+int test_dtls13_request_connection_id(void)
+{
+    EXPECT_DECLS;
+#if defined(HAVE_MANUAL_MEMIO_TESTS_DEPENDENCIES) && \
+    defined(WOLFSSL_DTLS13) && defined(WOLFSSL_DTLS_CID)
+    WOLFSSL_CTX *ctx_c = NULL, *ctx_s = NULL;
+    WOLFSSL *ssl_c = NULL, *ssl_s = NULL;
+    struct test_memio_ctx test_ctx;
+    const byte body[] = { 0x02 }; /* num_cids */
+    byte rec[256];
+    int recSz = (int)sizeof(rec);
+    byte readBuf[16];
+
+    /* the request is ACKed but ignored: we never send NewConnectionId */
+    ExpectIntEQ(test_dtls13_cid_setup(&test_ctx, &ctx_c, &ctx_s, &ssl_c,
+            &ssl_s, 1), TEST_SUCCESS);
+    ExpectIntEQ(test_dtls13_build_post_hs_msg(ssl_c, ssl_s,
+            request_connection_id, body, sizeof(body), rec, &recSz),
+            TEST_SUCCESS);
+    ExpectIntEQ(test_memio_inject_message(&test_ctx, 0, (const char*)rec,
+            recSz), 0);
+    ExpectIntEQ(wolfSSL_read(ssl_s, readBuf, sizeof(readBuf)), -1);
+    ExpectIntEQ(wolfSSL_get_error(ssl_s, -1), WOLFSSL_ERROR_WANT_READ);
+    ExpectIntGT(test_ctx.c_len, 0);
+    /* nothing but the ACK reaches the client */
+    ExpectIntEQ(wolfSSL_read(ssl_c, readBuf, sizeof(readBuf)), -1);
+    ExpectIntEQ(wolfSSL_get_error(ssl_c, -1), WOLFSSL_ERROR_WANT_READ);
+    ExpectIntEQ(test_ctx.c_len, 0);
+    ExpectIntEQ(test_ctx.s_len, 0);
+
+    /* connection must still work in both directions */
+    ExpectIntEQ(wolfSSL_write(ssl_c, "test", 5), 5);
+    ExpectIntEQ(wolfSSL_read(ssl_s, readBuf, sizeof(readBuf)), 5);
+    ExpectStrEQ(readBuf, "test");
+    ExpectIntEQ(wolfSSL_write(ssl_s, "back", 5), 5);
+    ExpectIntEQ(wolfSSL_read(ssl_c, readBuf, sizeof(readBuf)), 5);
+    ExpectStrEQ(readBuf, "back");
+
+    wolfSSL_free(ssl_s);
+    wolfSSL_free(ssl_c);
+    wolfSSL_CTX_free(ctx_s);
+    wolfSSL_CTX_free(ctx_c);
+#endif
+    return EXPECT_RESULT();
+}
+
+int test_dtls13_cid_msg_malformed(void)
+{
+    EXPECT_DECLS;
+#if defined(HAVE_MANUAL_MEMIO_TESTS_DEPENDENCIES) && \
+    defined(WOLFSSL_DTLS13) && defined(WOLFSSL_DTLS_CID)
+    /* cids length exceeding the body */
+    const byte truncated[] = { 0x00, 0x10, 0x03, 0x11, 0x22, 0x33, 0x01 };
+    /* CID length exceeding the cids vector */
+    const byte innerOverrun[] = { 0x00, 0x03, 0x05, 0x11, 0x22, 0x01 };
+    /* RequestConnectionId must be exactly one byte */
+    const byte reqTooLong[] = { 0x02, 0x00 };
+    /* invalid usage */
+    const byte badUsage[] = { 0x00, 0x04, 0x03, 0x11, 0x22, 0x33, 0x02 };
+    /* cid_immediate with no CID to use */
+    const byte immEmpty[] = { 0x00, 0x00, 0x00 };
+    const byte immZeroLenCid[] = { 0x00, 0x01, 0x00, 0x00 };
+
+    ExpectIntEQ(test_dtls13_post_hs_cid_msg_err(new_connection_id, truncated,
+            sizeof(truncated), 1, WC_NO_ERR_TRACE(BUFFER_ERROR),
+            decode_error), TEST_SUCCESS);
+    ExpectIntEQ(test_dtls13_post_hs_cid_msg_err(new_connection_id,
+            innerOverrun, sizeof(innerOverrun), 1,
+            WC_NO_ERR_TRACE(BUFFER_ERROR), decode_error), TEST_SUCCESS);
+    ExpectIntEQ(test_dtls13_post_hs_cid_msg_err(request_connection_id,
+            reqTooLong, sizeof(reqTooLong), 1, WC_NO_ERR_TRACE(BUFFER_ERROR),
+            decode_error), TEST_SUCCESS);
+    ExpectIntEQ(test_dtls13_post_hs_cid_msg_err(new_connection_id, badUsage,
+            sizeof(badUsage), 1, WC_NO_ERR_TRACE(INVALID_PARAMETER),
+            illegal_parameter), TEST_SUCCESS);
+    ExpectIntEQ(test_dtls13_post_hs_cid_msg_err(new_connection_id, immEmpty,
+            sizeof(immEmpty), 1, WC_NO_ERR_TRACE(INVALID_PARAMETER),
+            illegal_parameter), TEST_SUCCESS);
+    ExpectIntEQ(test_dtls13_post_hs_cid_msg_err(new_connection_id,
+            immZeroLenCid, sizeof(immZeroLenCid), 1,
+            WC_NO_ERR_TRACE(INVALID_PARAMETER), illegal_parameter),
+            TEST_SUCCESS);
+#endif
+    return EXPECT_RESULT();
+}
+
 
 int test_dtls_version_checking(void)
 {
--- a/tests/api/test_dtls.c
+++ b/tests/api/test_dtls.c
@@ -686,11 +686,10 @@ int test_wolfSSL_dtls_set_pending_peer_not_newest(void)
  * post-handshake messages (session tickets, ACKs) so both buffers are empty. */
 static int test_dtls13_cid_setup(struct test_memio_ctx* test_ctx,
         WOLFSSL_CTX** ctx_c, WOLFSSL_CTX** ctx_s, WOLFSSL** ssl_c,
-        WOLFSSL** ssl_s, int useCid)
+        WOLFSSL** ssl_s, const byte* serverCid, word32 serverCidSz)
 {
     EXPECT_DECLS;
     unsigned char client_cid[] = { 9, 8, 7, 6, 5, 4, 3, 2, 1, 0 };
-    unsigned char server_cid[] = { 0, 1, 2, 3, 4, 5, 6, 7, 8, 9 };
     byte readBuf[16];
     int i;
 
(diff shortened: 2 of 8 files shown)


src/dtls.c
```diff
@@ -1407,6 +1407,28 @@ int wolfSSL_dtls_cid_set(WOLFSSL* ssl, unsigned char* cid, unsigned int size)
...
+
+    cidInfo = DtlsCidGetInfo(ssl);
+    if (cidInfo == NULL)
+        return BAD_STATE_E;
+
+    newCid = DtlsCidNew(cid, size, ssl->heap);
+    if (newCid == NULL)
+        return MEMORY_ERROR;
+    XFREE(cidInfo->tx, ssl->heap, DYNAMIC_TYPE_TLSX);
+    cidInfo->tx = newCid;
```
Frauschi, line 1431: **TX CID swapped here without invalidating the cached record-size overhead**

`cidInfo->tx = newCid;` replaces the TX CID with one of a different length, but `DtlsCidReplaceTx` never resets `ssl->recordSzOverhead` afterward. That cache's key includes the CID length (see the comment on `wolfssl_local_GetRecordSize`, `src/internal.c:43180`), and every other site that changes record framing clears it: `TLSX_ConnectionID_Parse` (`dtls.c:1327`), `SetKeysSide` (`keys.c:3506`), and `ssl.c:7904`. A `NewConnectionId` doesn't go through `SetKeysSide`, so nothing else invalidates it.

Impact: after a peer sends `NewConnectionId` with a CID **longer** than the negotiated one, `wolfssl_local_GetRecordSize` returns a too-small size for large payloads → `CheckAvailableSize` under-allocates → `BuildTls13Message` hits `args->sz > outSz` (`tls13.c:3388`) and returns `BUFFER_E`, so `wolfSSL_write` fails with `BUILD_MSG_ERROR`. The current test only downsizes (10→3), which leaves the cache stale-high (over-allocation, still works) and masks the defect.

**Fix:** set `ssl->recordSzOverhead = 0;` right after `cidInfo->tx = newCid;`, mirroring `TLSX_ConnectionID_Parse`. Add the regression test the `TLSX_ConnectionID_Parse` comment already asks for.

tests/api/test_dtls.c
```diff
@@ -562,6 +562,437 @@ int test_wolfSSL_dtls_set_pending_peer(void)
...
+    ExpectStrEQ(readBuf, "back");
+
+    wolfSSL_free(ssl_s);
+    wolfSSL_free(ssl_c);
+    wolfSSL_CTX_free(ctx_s);
+    wolfSSL_CTX_free(ctx_c);
+    ssl_c = ssl_s = NULL;
+    ctx_c = ctx_s = NULL;
+
+    /* usage cid_immediate: the first non-empty CID must be used for all
```
Frauschi, line 877: **`cid_immediate` coverage can't catch the record-size cache bug**

This case only downsizes the CID (negotiated 10 → new 3) and never does a `wolfSSL_write` afterward that exercises `wolfssl_local_GetRecordSize` with the changed CID. A downsize leaves the cache stale-**high** (over-allocation, still functional), so the test passes even with the missing `recordSzOverhead` invalidation.

**Add coverage:**

1. Prime the cache before the CID change (a `wolfSSL_write`) and assert `wolfssl_local_GetRecordSize` equals `BuildMessage(sizeOnly=1)` after the change — the `TLSX_ConnectionID_Parse` comment requests exactly this.
2. Use a `NewConnectionId` whose CID is **larger** than the negotiated one, then verify a subsequent `wolfSSL_write` still succeeds.

This both fixes the coverage gap and would have caught the BLOCK finding above.

Frauschi (changes requested):

src/dtls.c
mattia-moffa: Thanks Tobias, I had no idea that the record size is cached. I also updated the test to explicitly check for this.

Frauschi:
Jenkins retest this please

src/dtls.c
```diff
@@ -1407,6 +1407,29 @@ int wolfSSL_dtls_cid_set(WOLFSSL* ssl, unsigned char* cid, unsigned int size)
...
 
+/* Replace the CID we use when sending records, after the peer provided a new
+ * one in a NewConnectionId message (RFC 9147 Section 9). */
+int DtlsCidReplaceTx(WOLFSSL* ssl, const byte* cid, byte size)
+{
+    CIDInfo* cidInfo;
+    ConnectionID* newCid;
+
+    if (ssl == NULL || cid == NULL || size == 0)
+        return BAD_FUNC_ARG;
```
Frauschi, line 1418: `DtlsCidReplaceTx()` accepts a peer-supplied TX CID with no size check:

`size` is never bounded here: any value up to 255 is stored as TX CID. This is asymmetric with `wolfSSL_dtls_cid_set()` just above, which rejects `size > DTLS_CID_MAX_SIZE` (default 10) with `LENGTH_ERROR`.

Two concerns:
- A remote peer can force every outgoing record to carry an oversized (up to 255-byte) CID.
- Latent breakage: the DTLS 1.2 AAD path (`writeAeadAuthData()`) hard-fails when `cidSz > DTLS_CID_MAX_SIZE`. Not reachable today (`NewConnectionId` is 1.3-only), but reusing this helper on the 1.2 send path would silently break sending.

Suggestion: enforce the bound here (and have `DoDtls13NewConnectionId` treat oversize as a protocol error):

```
if (size > DTLS_CID_MAX_SIZE)
    return LENGTH_ERROR;
```
or add a comment explaining why the TX CID is intentionally decoupled from `DTLS_CID_MAX_SIZE`.

rizlik (dismissed):

Frauschi:
the tiny-tls13 failures are unrelated and will be fixed by #10835

Frauschi (approved):

mattia-moffa:
Jenkins retest this please

douzzer (approved):

PR #10631, opened 2026-06-08
bound sessionID read in sniffer ProcessServerHello

netliomax25-code:
ProcessServerHello copies a fixed ID_LEN bytes of session id out of the captured ServerHello when the session-id length byte is non-zero, but the bounds check before it only ensures that length-byte value plus the cipher suite and compression fields are present. A ServerHello whose session id is shorter than 32 bytes, in a record with no trailing data, makes that copy read past the end of the frame buffer. ProcessClientHello already guards the identical copy with an ID_LEN check against the remaining bytes, so add the same check on the server side before the copy.

### DIFF
--- a/src/sniffer.c
+++ b/src/sniffer.c
@@ -3733,6 +3733,10 @@ static int ProcessServerHello(int msgSz, const byte* input, int* sslBytes,
         return WOLFSSL_FATAL_ERROR;
     }
     if (b) {
+        if (ID_LEN > *sslBytes) {
+            SetError(SERVER_HELLO_INPUT_STR, error, session, FATAL_ERROR_STATE);
+            return WOLFSSL_FATAL_ERROR;
+        }
     #ifdef WOLFSSL_TLS13
         XMEMCPY(session->sslServer->session->sessionID, input, ID_LEN);
         session->sslServer->session->sessionIDSz = ID_LEN;
--- a/src/sniffer.c
+++ b/src/sniffer.c
@@ -3733,16 +3733,20 @@ static int ProcessServerHello(int msgSz, const byte* input, int* sslBytes,
         return WOLFSSL_FATAL_ERROR;
     }
     if (b) {
-        if (ID_LEN > *sslBytes) {
+        /* session id length is attacker controlled; reject anything that
+         * would overflow the fixed-size sessionID buffer. Reading b bytes is
+         * already bounded by the check above, so copy the advertised length
+         * instead of a fixed ID_LEN and keep short non-zero ids intact. */
+        if (b > ID_LEN) {
             SetError(SERVER_HELLO_INPUT_STR, error, session, FATAL_ERROR_STATE);
             return WOLFSSL_FATAL_ERROR;
         }
     #ifdef WOLFSSL_TLS13
-        XMEMCPY(session->sslServer->session->sessionID, input, ID_LEN);
-        session->sslServer->session->sessionIDSz = ID_LEN;
+        XMEMCPY(session->sslServer->session->sessionID, input, b);
+        session->sslServer->session->sessionIDSz = b;
     #endif
-        XMEMCPY(session->sslServer->arrays->sessionID, input, ID_LEN);
-        session->sslServer->arrays->sessionIDSz = ID_LEN;
+        XMEMCPY(session->sslServer->arrays->sessionID, input, b);
+        session->sslServer->arrays->sessionIDSz = b;
         session->sslServer->options.haveSessionId = 1;
     }
     input     += b;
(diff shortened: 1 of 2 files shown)


parasol-aser:
**src/sniffer.c:ProcessServerHello**
The new `ID_LEN > *sslBytes` check rejects complete ServerHello messages that advertise a short nonzero session ID and have no or few extension bytes. A server or pcap using a short nonzero session ID can now make the sniffer mark the flow fatal and remove the session, so affected sniffer deployments lose visibility for that TLS flow; this is flow-level availability/evasion, not a crash or secret disclosure.
Fix: Validate the encoded length (`b <= ID_LEN`) and copy/set the session ID size from `b`; if the sniffer intentionally supports only 32-byte nonzero IDs, reject that invariant based on `b` and add focused pcap coverage for short IDs.

netliomax25-code:
Good catch, you're right that the first version rejected legitimate short session IDs.

1. Issue: the `ID_LEN > *sslBytes` check fataled any ServerHello advertising a non-zero session id shorter than 29 bytes with no trailing data, which removes the flow and drops sniffer visibility for it.
2. Fix: copy the advertised length `b` and set `sessionIDSz = b`, guarded by `b > ID_LEN` so the fixed-size sessionID buffer can't overflow. Reading `b` bytes is already covered by the existing `(b + SUITE_LEN + ENUM_LEN) > *sslBytes` check, so the original out-of-bounds read stays closed.
3. Resume path is unchanged: the match still requires `sessionIDSz == ID_LEN` on both sides, and for the usual 32-byte id `b == ID_LEN`, so it still resumes.

Added `test_sniffer_serverhello_short_sessionid` in tests/api.c that drives a 1-byte-session-id ServerHello through `ssl_DecodePacket`; it fails against the old check and passes with this change. The existing sniffer pcap tests (static-rsa, tls13-ecc, tls13-ecc-resume) still pass.

dgarske:
Hi @netliomax25-code , can you tell us more about your project and use of the sniffer tool? Is this commercial or accedemic? What tools did you use to find these issues? We do require a signed contributor agreement. If that's something that you'd like to request please email support at wolfssl dot com and include details like your region and your organizational affiliation. We encourage reports like this and thank you for submitting your two PR's. Are you planning to submit further work? Thanks, David Garske, wolfSSL

netliomax25-code:
Hi David, thanks for taking a look.
I'm with Digiscrypt Technologies (digiscrypt.com), a software engineering and open-source security studio based in Bengaluru, India. Part of our work is auditing and hardening widely-used open-source libraries, which is how we came across the sniffer issues — so commercial affiliation, though the contributions themselves are just us trying to get fixes upstream.
On tooling — these were found with an in-house analysis platform we've built. It combines deep static analysis of parsing and memory-handling paths with targeted validation of the findings, and we manually review and reproduce every issue before reporting, so what reaches you is confirmed, not scanner noise. The sniffer's ServerHello/ClientHello handling came up during a sweep of packet-processing code in widely-deployed TLS implementations. The platform is currently internal-only — we plan to bring it to market down the line, but until then we're not sharing internals. Happy to walk through how any specific finding was confirmed, though.
Yes, we're planning to submit further work as we find things worth fixing. I'll email [support@wolfssl.com](mailto:support@wolfssl.com) about the contributor agreement with our details.
Thanks again for the quick review on  PRs.

netliomax25-code:
any update?

dgarske:
> any update?

Have you emailed our support alias yet? I can't find it. If you did already please let me know the zendesk ticket number. Thanks.

netliomax25-code:
Hi David - apologies for the delay on this. I'd flagged it with our team
last week, but it only actually went out today; it's now in with your
support alias as ticket #22029. Sorry for the back-and-forth. Happy to
provide anything else you need to get the contributor agreement moving -
thanks for your patience.

netliomax25-code:
gentle ping

src/sniffer.c
```diff
@@ -3733,12 +3733,20 @@ static int ProcessServerHello(int msgSz, const byte* input, int* sslBytes,
...
+         * already bounded by the check above, so copy the advertised length
+         * instead of a fixed ID_LEN and keep short non-zero ids intact. */
+        if (b > ID_LEN) {
+            SetError(SERVER_HELLO_INPUT_STR, error, session, FATAL_ERROR_STATE);
+            return WOLFSSL_FATAL_ERROR;
+        }
     #ifdef WOLFSSL_TLS13
-        XMEMCPY(session->sslServer->session->sessionID, input, ID_LEN);
-        session->sslServer->session->sessionIDSz = ID_LEN;
+        XMEMCPY(session->sslServer->session->sessionID, input, b);
```
dgarske, line 3745: The previous code would also never overflow the sessionID field. Closing this PR. Thanks

PR #10632, opened 2026-06-08
Fenrir fixes

Frauschi:
### Summary

This PR addresses a set of Fenrir-reported issues across wolfCrypt and the TLS layer. Each fix is accompanied by a regression test where applicable.

#### Cipher key-state validation
  - **F-5378** — `wc_Arc4Process()` now refuses to run unless a key was configured via `wc_Arc4SetKey()`, preventing ARC4 from silently copying plaintext to ciphertext with an all-zero keystream.
  - **F-5379** — `wc_Des3_CbcEncrypt()` and `wc_Des3_CbcDecrypt()` now refuse to run unless a key was configured via `wc_Des3_SetKey()`, returning `MISSING_KEY` instead of performing DES3 operations with uninitialized or zeroed key material.

#### Sensitive-memory zeroization
- **F-5437** — `eccToPKCS8()` now zeroizes the temporary buffer holding the plaintext ECC private key on every exit path via a consolidated `goto exit` cleanup.
- **F-5438** — ML-DSA key generation and signing now `ForceZero()` the private vectors (s1, s2, t, y, …) before freeing, while skipping the public matrix A that need not be cleared.

#### Post-quantum / hardware ports
- **F-4443** — Fixed `wolfSSL_liboqsGetRandomData()` to advance the output buffer between chunks and guard against `size_t`→`word32` truncation, avoiding overwritten output and a potential infinite loop on large requests.
- **F-4441 / F-4442** — Corrected the Cavium Nitrox port to write the result count into the right struct field in `NitroxCheckRequests()` and to pass the output-length argument with the expected `Uint16*` type in the RSA decrypt/verify paths, using a dedicated `Uint16` temporary so the `ntohs()` conversion stays correct on big-endian hosts.

#### TLS supported_groups / ec_point_formats handling
- **F-4891** — Hardened `supported_groups` parsing so an empty named-group list is rejected as malformed and an all-unsupported-groups list still records a (now empty) restriction instead of being treated as absent.
- **F-4892** — The TLS 1.2 server now enforces RFC 8422 §5.1.2 by aborting with an `illegal_parameter` alert when the client's `ec_point_formats` extension omits the uncompressed format and an ECC cipher suite is actually negotiated; the check runs after suite selection (`ssl->specs.kea`) so a client that negotiates a non-ECC suite is not rejected.

### Testing

Added regression tests for ARC4, DES3, supported_groups parsing, and the `ec_point_formats` enforcement — covering both the abort path when an ECC suite is negotiated (driven through a full handshake via memio) and the complementary case where a non-ECC (DHE_RSA) suite is negotiated and the handshake proceeds.


wolfssl/wolfcrypt/des3.h
```diff
@@ -101,6 +101,7 @@ struct Des3 {
     word32 key[3][DES_KS_SIZE];
     word32 reg[DES_BLOCK_SIZE / sizeof(word32)];      /* for CBC mode */
     word32 tmp[DES_BLOCK_SIZE / sizeof(word32)];      /* same         */
+    byte   keySet;    /* set to 1 once a key has been configured */
```
dgarske: Please use WC_BITFIELD and :1 and put at bottom of struct. Best compat.

wolfssl/wolfcrypt/arc4.h
```diff
@@ -46,6 +46,7 @@ enum {
 typedef struct Arc4 {
     byte x;
     byte y;
+    byte keySet;    /* set to 1 once a key has been configured */
```
dgarske: Same. Use WC_BITFIELD and :1 and put at bottom. Thank you

dgarske (changes requested):

wolfcrypt/src/wc_mldsa.c
```diff
@@ -8158,6 +8163,10 @@ static int mldsa_make_key_from_seed(wc_MlDsaKey* key, const byte* seed)
         key->pubKeySet = 1;
```
dgarske, line 8163: **⚪ [Info] ML-DSA small-mem key-gen zeroizes the public expansion buffer too**

In the second (small-mem) key-generation variant the cleanup does `ForceZero(s1, allocSz)`, zeroizing the entire allocation including the trailing rejection-sampling / A-expansion working region (`MLDSA_REJ_NTT_POLY_H_SIZE + MLDSA_POLY_SIZE`), which is derived from the public seed. This is harmless (over-zeroization of public data, fully within the allocation) but is inconsistent with the stated PR intent ('skipping the public matrix A that need not be cleared') and with the other two paths - keygen variant 1 zeroes only `s1Sz + 2*s2Sz`, and `mldsa_sign_with_seed_mu` zeroes `allocSz - aSz`. Unlike those paths this variant has no discrete `aSz` to subtract (A is expanded on the fly), so the over-zeroization is minor; flagging only for consistency with the comment/intent.

**Fix:** No functional change required; optionally align the zeroized size (or the comment) with the 'skip public material' intent used in the other two paths.

dgarske (commented):
## Skoll Code Review

**Findings:** 4 total — 1 posted, 3 skipped

### Posted findings
- [Info] **ML-DSA small-mem key-gen zeroizes the public expansion buffer too** — `wolfcrypt/src/wc_mldsa.c:8163-8168`

### Skipped findings
- [Medium] ~~Des3 struct: keySet inserted mid-struct (ABI / FIPS boundary concern)~~
- [Medium] ~~supported_groups restriction not exercised end-to-end~~
- [Low] ~~Arc4 struct: keySet inserted before state[] instead of at end~~

wolfssl/wolfcrypt/des3.h
Frauschi: fixed

wolfssl/wolfcrypt/arc4.h
Frauschi: fixed

wolfcrypt/src/wc_mldsa.c
Frauschi: This is intentional for SMALL_MEM builds, as the sensitive data is split over the buffer. The overhead of simply zeroing the whole buffer is negligible compared to two distinct ForceZero calls. Added an explanation in a comment.

Frauschi:
Jenkins retest this please

Frauschi:
The Arduino failure is unrelated and fixed in #10644.

tests/api/test_tls_ext.c
```diff
@@ -1033,3 +1033,170 @@ int test_TLSX_SRTP_msg_type_validation(void)
...
+int test_TLSX_SupportedCurve_empty_or_unsupported(void)
+{
+    EXPECT_DECLS;
+#if !defined(NO_WOLFSSL_CLIENT) && !defined(NO_TLS) && \
+    defined(HAVE_SUPPORTED_CURVES) && !defined(WOLFSSL_NO_TLS12)
+    /* This exercises the server's parsing of a received ClientHello: the
+     * relevant code path (TLSX_SupportedCurve_Parse) is selected by the
+     * message type passed to TLSX_Parse (client_hello => isRequest), not by
+     * the side of the WOLFSSL object. A client-side WOLFSSL is used purely as
+     * the parse vehicle because creating a server-side WOLFSSL would require a
```
dgarske, line 1098: **🟠 [Medium] Supported-curve parse test lacks server/TLS1.3 guard; fails in --disable-server non-TLS1.3 builds**

The new test calls `TLSX_Parse(..., client_hello, ...)` and asserts the empty-list path returns `BUFFER_ERROR` and the unsupported-only path records a `TLSX_SUPPORTED_GROUPS` node. Both depend on `TLSX_SupportedCurve_Parse` actually running. In src/tls.c the dispatch macro `EC_PARSE` is the real function only when `!defined(NO_WOLFSSL_SERVER) || (defined(WOLFSSL_TLS13) && !defined(WOLFSSL_NO_SERVER_GROUPS_EXT))`; otherwise `#define EC_PARSE(a,b,c,d,e) 0`. The test's `#if` gate is `!defined(NO_WOLFSSL_CLIENT) && !defined(NO_TLS) && defined(HAVE_SUPPORTED_CURVES) && !defined(WOLFSSL_NO_TLS12)` and does NOT require server or TLS1.3. In a `--disable-server` build without TLS 1.3, `EC_PARSE` becomes a no-op returning 0, so `TLSX_Parse(emptyList)` returns 0 (not `BUFFER_ERROR`) and no node is added — both `ExpectIntEQ(..., BUFFER_ERROR)` and `ExpectNotNull(TLSX_Find(...))` then fail at runtime. The sibling test `test_TLSX_PointFormat_uncompressed_required` correctly guards on `!defined(NO_WOLFSSL_SERVER)` (its comment even explains PF_PARSE is a no-op without server), so the gating is inconsistent.

**Fix:** Add a guard so the TLS1.2 portion of the test only runs when `TLSX_SupportedCurve_Parse` (EC_PARSE) is the real function — mirror the EC_PARSE condition (`!defined(NO_WOLFSSL_SERVER) || (defined(WOLFSSL_TLS13) && !defined(WOLFSSL_NO_SERVER_GROUPS_EXT))`), matching the approach used by the point-format sibling test.

src/internal.c
```diff
@@ -38894,6 +38894,13 @@ static int AddPSKtoPreMasterSecret(WOLFSSL* ssl)
                 }
 
 #ifdef HAVE_TLS_EXTENSIONS
```
dgarske, line 38904: **🔵 [Low] peerNoUncompPF reset skipped when ClientHello has no extensions block**

The defensive reset `ssl->options.peerNoUncompPF = 0;` is placed inside the `if ((i - begin) < helloSz)` extensions-present block (and inside the `TLSX_SupportExtensions(ssl)` block), immediately before `TLSX_Parse`. If a subsequent ClientHello on the same WOLFSSL object carries no extensions block at all, the reset is not executed and a stale `peerNoUncompPF=1` from an earlier handshake persists; if an ECC suite is then negotiated, the new RFC 8422 check would fire a spurious `illegal_parameter` abort. In practice this is not reachable for secure renegotiation (RFC 5746 requires the renegotiation_info extension, so the extensions block is always present and the reset always runs), so the stated motivation in the comment is satisfied. It is only theoretically reachable with unsafe legacy renegotiation enabled and a no-extensions ClientHello that still negotiates ECC. Flagged for robustness only.

**Fix:** Optionally move the reset earlier to guarantee it runs for every ClientHello regardless of whether an extensions block is present. Low priority since secure renegotiation always carries an extensions block.

dgarske (commented):
## Skoll Code Review

### Posted findings
- [Medium] **Supported-curve parse test lacks server/TLS1.3 guard; fails in --disable-server non-TLS1.3 builds** — `tests/api/test_tls_ext.c:1058-1100`
- [Low] **peerNoUncompPF reset skipped when ClientHello has no extensions block** — `src/internal.c:38896-38907`

tests/api/test_tls_ext.c
Frauschi: fixed by extending the guards

src/internal.c
Frauschi: fixed

dgarske (approved):

PR #10634, opened 2026-06-08
Add DTLS secondary cookie secret for app-driven rotation

### LINKED ISSUE #10617: [Bug]: DTLS 1.3 Cookie Secret Rotation Window and Cookie Lifetime Policy Are Application-Managed
### Contact Details

lxd_dong@bupt.edu.cn

### Version

5.9.1

### Description

# DTLS 1.3 Cookie Secret Rotation Window and Cookie Lifetime Policy Are Application-Managed

## Summary

This report covers IDs 101, 103, and 104.

wolfSSL implements the main DTLS 1.3 HelloRetryRequest cookie mechanism: the server can send a cookie extension, the cookie is HMAC-protected with a server secret, and an invalid cookie is rejected. However, the audited implementation stores and verifies cookies with one active `tls13CookieSecret` only. It does not provide a built-in previous/current secret transition window, key identifier, two-secret verification path, or cookie timestamp/expiration policy.

The finding is therefore confirmed as **partially satisfied**.

## Standard Requirement

Relevant sections: RFC 9147 Section 5.1, `Denial-of-Service Countermeasures`, and Section 5.3, `ClientHello Format`.

Relevant original English text from RFC 9147:

```text
The server can defend against this attack by changing the secret value frequently, thus invalidating those cookies.
```

```text
If the server wishes to allow legitimate clients to handshake through the transition (e.g., a client received a cookie with Secret 1 and then sent the second ClientHello after the server has changed to Secret 2), the server can have a limited window during which it accepts both secrets.
```

```text
[RFC7296] suggests adding a key identifier to cookies to detect this case. An alternative approach is simply to try verifying with both secrets.
```

```text
It is RECOMMENDED that servers implement a key rotation scheme that allows the server to manage keys with overlapping lifetimes.
```

```text
Alternatively, the server can store timestamps in the cookie and reject cookies that were generated outside a certain interval of time.
```

```text
If a server receives a ClientHello with an invalid cookie, it MUST terminate the handshake with an "illegal_parameter" alert.
```

The rotation-window and timestamp language is not an unconditional "MUST" for every implementation strategy. It is a recommended or alternative server-side policy for managing cookie validity during secret changes. This is why the result is partial rather than fully unsatisfied.

### Single HRR Cookie Secret Field

`wolfssl-master\wolfssl\internal.h:5024`

```c
buffer          tls13CookieSecret;     /* HRR cookie secret */
```

The audited structure has one HRR cookie secret field. No companion previous-secret field such as `tls13CookieSecretPrev`, `oldCookieSecret`, or similar was found.

### Cookie Generation Uses the Single Secret

`wolfssl-master\src\tls13.c:3632`

```c
int CreateCookieExt(const WOLFSSL* ssl, byte* hash, word16 hashSz,
                    TLSX** exts, byte cipherSuite0, byte cipherSuite)
```

`wolfssl-master\src\tls13.c:3648`

```c
if (ssl->buffers.tls13CookieSecret.buffer == NULL ||
        ssl->buffers.tls13CookieSecret.length == 0) {
    WOLFSSL_MSG("Missing DTLS 1.3 cookie secret");
    return COOKIE_ERROR;
}
```

`wolfssl-master\src\tls13.c:3692`

```c
ret = wc_HmacSetKey(cookieHmac, cookieType,
                    ssl->buffers.tls13CookieSecret.buffer,
                    ssl->buffers.tls13CookieSecret.length);
```

`CreateCookieExt` constructs cookie data from the ClientHello hash and selected negotiation values, then protects it with HMAC under `tls13CookieSecret`. In DTLS mode it also binds the cookie to the peer address.

### Cookie Verification Uses the Same Single Secret

`wolfssl-master\src\tls13.c:6844`

```c
int TlsCheckCookie(const WOLFSSL* ssl, const byte* cookie, word16 cookieSz)
```

`wolfssl-master\src\tls13.c:6852`

```c
if (ssl->buffers.tls13CookieSecret.buffer == NULL ||
        ssl->buffers.tls13CookieSecret.length == 0) {
    WOLFSSL_MSG("Missing DTLS 1.3 cookie secret");
    return COOKIE_ERROR;
}
```

`wolfssl-master\src\tls13.c:6883`

```c
ret = wc_HmacSetKey(cookieHmac, cookieType,
                    ssl->buffers.tls13CookieSecret.buffer,
                    ssl->buffers.tls13CookieSecret.length);
```

`wolfssl-master\src\tls13.c:6908`

```c
if (ConstantCompare(cookie + cookieSz, mac, macSz) != 0) {
    WOLFSSL_ERROR_VERBOSE(HRR_COOKIE_ERROR);
    return HRR_COOKIE_ERROR;
}
```

This verifies the received cookie with the current secret. There is no second verification attempt with a previous secret, no key identifier branch, and no timestamp/expiration check in this function.

### DTLS ClientHello Path Calls the Cookie Verifier

`wolfssl-master\src\dtls.c:263`

```c
static int CheckDtlsCookie(const WOLFSSL* ssl, WolfSSL_CH* ch,
                           byte isTls13, byte* cookieGood)
```

`wolfssl-master\src\dtls.c:278`

```c
ret = TlsCheckCookie(ssl, ch->cookieExt.elements + OPAQUE16_LEN,
        (word16)(ch->cookieExt.size - OPAQUE16_LEN));
```

The DTLS 1.3 stateless ClientHello path routes the cookie extension to `TlsCheckCookie`.

### Public API Replaces the Single Secret

`wolfssl-master\src\tls13.c:14484`

```c
int wolfSSL_send_hrr_cookie(WOLFSSL* ssl, const unsigned char* secret,
                            unsigned int secretSz)
```

`wolfssl-master\src\tls13.c:14512`

```c
if (ssl->buffers.tls13CookieSecret.buffer != NULL) {
    ForceZero(ssl->buffers.tls13CookieSecret.buffer,
              ssl->buffers.tls13CookieSecret.length);
    XFREE(ssl->buffers.tls13CookieSecret.buffer,
          ssl->heap, DYNAMIC_TYPE_COOKIE_PWD);
}
```

`wolfssl-master\src\tls13.c:14544`

```c
XMEMCPY(ssl->buffers.tls13CookieSecret.buffer, secret, secretSz);
```

`wolfssl-master\src\tls13.c:14546`

```c
ssl->options.sendCookie = 1;
```

This API allows an application to configure or regenerate the HRR cookie secret. When the secret length changes, the old buffer is zeroed and freed before the new one is installed. The code does not preserve the old secret for an overlap window.

## Existing wolfSSL Test Coverage

`wolfssl-master\tests\api\test_tls13.c:5468`

```c
/* Test that a corrupted HRR cookie HMAC is rejected with HRR_COOKIE_ERROR. */
int test_tls13_hrr_bad_cookie(void)
```

`wolfssl-master\tests\api\test_tls13.c:5507`

```c
ssl_s->buffers.tls13CookieSecret.buffer[
    ssl_s->buffers.tls13CookieSecret.length - 1] ^= 0xFF;
```

`wolfssl-master\tests\api\test_tls13.c:5516`

```c
WC_NO_ERR_TRACE(HRR_COOKIE_ERROR));
```

wolfSSL has a unit test proving that a cookie fails when the active server-side secret no longer verifies it. That supports the current-secret integrity path, but it also demonstrates the absence of an automatic old-secret acceptance window in the tested behavior.

## Runtime Evidence

Compiled source-behavior harness:

`test-wolfssl-dtls\rfc9147\101-150\repro_cookie_secret_window_101_103_104_source_check.c`

Build command:

```powershell
& 'D:\LLVM\bin\clang.exe' 'test-wolfssl-dtls\rfc9147\101-150\repro_cookie_secret_window_101_103_104_source_check.c' -o 'test-wolfssl-dtls\rfc9147\101-150\repro_cookie_secret_window_101_103_104_source_check.exe'
```

Run command:

```powershell
& 'test-wolfssl-dtls\rfc9147\101-150\repro_cookie_secret_window_101_103_104_source_check.exe' *> 'test-wolfssl-dtls\rfc9147\101-150\repro_cookie_secret_window_101_103_104_source_check.log'
```

Observed result:

```text
Conclusion: PASS - source behavior confirms IDs 101/103/104 are partially satisfied: single-secret HMAC cookies exist, but no built-in dual-secret window or cookie timestamp/expiration policy was found.
```

Important selected assertions from the same log:

```text
PASS single HRR cookie secret field                             contains "buffer          tls13CookieSecret;"
PASS TlsCheckCookie verifies with the same single secret        contains "ssl->buffers.tls13CookieSecret.buffer"
PASS TlsCheckCookie rejects bad MAC                             contains "return HRR_COOKIE_ERROR;"
PASS no previous cookie secret field                            does not contain "tls13CookieSecretPrev"
PASS no old cookie secret field                                 does not contain "oldCookieSecret"
PASS no both-secret verification wording                        does not contain "both secrets"
PASS CreateCookieExt has no timestamp                           does not contain "timestamp"
PASS TlsCheckCookie has no expiration                           does not contain "expiration"
```

This is a compiled and executed source-behavior check. It is not a full packet-level interoperability test because this local checkout does not contain a prebuilt wolfSSL library or test executable to link and run a complete DTLS 1.3 handshake scenario.

## Inconsistency Reason

Implemented behavior:

| Requirement area | wolfSSL behavior |
|---|---|
| DTLS 1.3 HRR cookie extension | Implemented through `CreateCookieExt` and the DTLS ClientHello cookie path. |
| Cookie integrity | Implemented with HMAC over cookie data using `tls13CookieSecret`. |
| Invalid cookie rejection | Implemented through `HRR_COOKIE_ERROR`. |
| Application-controlled secret | Implemented through `wolfSSL_send_hrr_cookie`. |

Missing or application-managed behavior:

| RFC 9147 policy area | wolfSSL audited behavior |
|---|---|
| Frequently changing cookie secret | Application can call `wolfSSL_send_hrr_cookie`, but wolfSSL does not provide a built-in rotation scheduler. |
| Limited transition window accepting both old and new secrets | No previous-secret storage and no two-secret verification loop were found. |
| Key identifier in cookie | No key-id field or verifier branch was found in the HRR cookie path. |
| Cookie timestamp and expiration interval | No timestamp/lifetime/expiration field or check was found in `CreateCookieExt` or `TlsCheckCookie`. |

Therefore, IDs 101, 103, and 104 are real partial-consistency findings: wolfSSL implements the basic cookie mechanism, but not the built-in overlapping-lifetime or timestamp-expiration policy described by RFC 9147 as a recommended/alternative server-side defense.

## Impact

Deployments using wolfSSL's default single active HRR cookie secret can reject stale cookies after a secret change, but legitimate clients that received a cookie before rotation may fail during the transition unless the application manages rotation carefully. Because the cookie does not carry a timestamp or key identifier in the audited path, built-in age-based rejection and old/new secret selection are not available.

## Suggested Fix Direction

Add an internal cookie-secret manager for DTLS 1.3 HRR cookies. A complete fix should include:

1. Current and previous cookie secrets with explicit activation and expiration times.
2. Cookie encoding that optionally includes a key identifier or timestamp.
3. Verification that accepts cookies generated by either valid secret during the overlap window.
4. Rejection of cookies outside the configured lifetime.
5. Unit tests for current-secret success, previous-secret transition success, expired-cookie rejection, and malformed key-id/timestamp rejection.

### Reproduction steps

_No response_

### Relevant log output

```shell

```

julek-wolfssl:
### Description

Add verify-only secondary cookie secrets for both DTLS 1.2 (HelloVerifyRequest) and DTLS 1.3 (HelloRetryRequest). A secondary secret is used only during cookie verification, and only if the primary secret fails to verify the cookie; it is never used to issue cookies.

This lets an application rotate the cookie secret on a stateless DTLS server without rejecting clients that received a cookie under the previous secret: after rotating the primary, the application installs the old secret as the secondary so in-flight cookies still verify for an overlap window (RFC 9147 Section 5.1).

### Why two APIs

The DTLS 1.2 and 1.3 cookie mechanisms keep separate secrets and separate verify routines, set their primary secret through different APIs (`wolfSSL_DTLS_SetCookieSecret` for 1.2, `wolfSSL_send_hrr_cookie` for 1.3), and live under different build guards (a DTLS-1.2-only build has none of the TLS 1.3 / HRR machinery). So each version gets its own sibling API:

- `wolfSSL_set_hrr_cookie_secret_secondary()` — DTLS 1.3. Per-connection and DTLS-only; returns `BAD_FUNC_ARG` for a non-DTLS or non-TLS-1.3 object, since TLS 1.3 over a reliable transport does not verify cookies statelessly across the HelloRetryRequest exchange.
- `wolfSSL_DTLS_SetCookieSecretSecondary()` — DTLS 1.2, the counterpart of the above. `NULL`/0 clears it. A server handling both versions sets both secondary secrets.

### Implementation

- New per-SSL buffers `tls13CookieSecretSecondary` and `dtlsCookieSecretSecondary`, freed on SSL teardown (the 1.3 buffer is also freed in `wolfSSL_disable_hrr_cookie()`).
- `TlsCheckCookie()` factors its HMAC compute/compare into `TlsCheckCookieMac()` and tries the primary secret, then the secondary on mismatch.
- `CreateDtls12Cookie()` now takes the secret as parameters; `CheckDtlsCookie()` computes with the primary and retries with the secondary on mismatch.
- Issuance (`CreateCookieExt`, `SendHelloVerifyRequest`) always uses the primary.

### Testing

Tests cover, for both versions: a handshake where the cookie is accepted via the secondary after a rotation; a negative case where the signing secret has been rotated out of both slots; secondary clear behavior; primary still wins on the fast path; replacing the secondary activates the new value; secondary equal to primary is harmless; and issuance uses the primary, not the verify-only secondary. The 1.3 API additionally has NULL/side/version argument-validation coverage.

### DIFF
--- a/doc/dox_comments/header_files/ssl.h
+++ b/doc/dox_comments/header_files/ssl.h
@@ -8964,7 +8964,7 @@ int   wolfSSL_DTLS_SetCookieSecret(WOLFSSL* ssl,
     WOLFSSL* ssl = wolfSSL_new(ctx);
     const byte* oldSecret;
     word32 oldSecretSz; // size of oldSecret
-    …
+    ...
     if(wolfSSL_DTLS_SetCookieSecretSecondary(ssl, oldSecret, oldSecretSz) != 0){
     	// Code block for failure to set secondary DTLS cookie secret
     } else {
--- a/doc/dox_comments/header_files/ssl.h
+++ b/doc/dox_comments/header_files/ssl.h
@@ -8939,6 +8939,45 @@ int   wolfSSL_DTLS_SetCookieSecret(WOLFSSL* ssl,
                                                word32 secretSz);
 
 /*!
+    \brief This function sets a secondary DTLS 1.2 cookie secret used only when
+    verifying a received HelloVerifyRequest cookie, and only if the primary
+    secret (set with wolfSSL_DTLS_SetCookieSecret()) fails to verify it.  This
+    lets an application rotate the cookie secret on a stateless server without
+    rejecting clients whose cookie was issued under the previous secret: install
+    the new secret as the primary and the previous secret here for an overlap
+    window.  The secondary secret is never used to issue cookies.  It is the
+    DTLS 1.2 counterpart of wolfSSL_set_hrr_cookie_secret_secondary().
+
+    \return 0 returned if the function executed without an error.
+    \return BAD_FUNC_ARG returned if ssl is NULL.
+    \return MEMORY_ERROR returned if there was a problem allocating
+    memory for the secondary cookie secret.
+
+    \param ssl a pointer to a WOLFSSL structure, created using wolfSSL_new().
+    \param secret a constant byte pointer representing the secret buffer.
+    Passing NULL (or a secretSz of 0) clears any previously set secondary secret.
+    \param secretSz the size of the buffer.
+
+    _Example_
+    \code
+    WOLFSSL* ssl = wolfSSL_new(ctx);
+    const byte* oldSecret;
+    word32 oldSecretSz; // size of oldSecret
+    …
+    if(wolfSSL_DTLS_SetCookieSecretSecondary(ssl, oldSecret, oldSecretSz) != 0){
+    	// Code block for failure to set secondary DTLS cookie secret
+    } else {
+    	// Success! Secondary cookie secret is set.
+    }
+    \endcode
+
+    \sa wolfSSL_DTLS_SetCookieSecret
+*/
+int   wolfSSL_DTLS_SetCookieSecretSecondary(WOLFSSL* ssl,
+                                               const byte* secret,
+                                               word32 secretSz);
+
+/*!
     \brief This function retrieves the random number.
 
     \return rng upon success.
@@ -14007,6 +14046,51 @@ int  wolfSSL_connect(WOLFSSL* ssl);
     \sa wolfSSL_disable_hrr_cookie
 */
 int  wolfSSL_send_hrr_cookie(WOLFSSL* ssl,
+    const unsigned char* secret, unsigned int secretSz);
+
+/*!
+    \ingroup Setup
+
+    \brief This function sets a secondary HelloRetryRequest cookie secret on a
+    DTLS 1.3 server.  It is used only when verifying a received cookie, and only
+    if the primary secret (set with wolfSSL_send_hrr_cookie()) fails to verify
+    it.  This lets an application rotate the cookie secret on a stateless DTLS
+    1.3 server without rejecting clients whose cookie was issued under the
+    previous secret: install the new secret as the primary and the previous
+    secret here for an overlap window.  The secondary secret is never used to
+    issue cookies.  This API is DTLS only; TLS 1.3 over a reliable transport
+    does not verify cookies statelessly across the HelloRetryRequest exchange.
+
+    \param [in,out] ssl a pointer to a WOLFSSL structure, created using wolfSSL_new().
+    \param [in] secret a pointer to a buffer holding the secondary secret.
+    Passing NULL (or a secretSz of 0) clears any previously set secondary secret.
+    \param [in] secretSz Size of the secret in bytes.
+
+    \return BAD_FUNC_ARG if ssl is NULL, not using TLS v1.3, or not using DTLS.
+    \return SIDE_ERROR if called with a client.
+    \return WOLFSSL_SUCCESS if successful.
+    \return MEMORY_ERROR if allocating dynamic memory for storing secret failed.
+
+    _Example_
+    \code
+    int ret;
+    WOLFSSL* ssl;
+    char newSecret[32];
+    char oldSecret[32];
+    ...
+    // rotate: new secret becomes primary, previous secret stays valid
+    wolfSSL_send_hrr_cookie(ssl, newSecret, sizeof(newSecret));
+    ret = wolfSSL_set_hrr_cookie_secret_secondary(ssl, oldSecret,
+        sizeof(oldSecret));
+    if (ret != WOLFSSL_SUCCESS) {
+        // failed to set the secondary cookie secret
+    }
+    \endcode
+
+    \sa wolfSSL_send_hrr_cookie
+    \sa wolfSSL_disable_hrr_cookie
+*/
+int  wolfSSL_set_hrr_cookie_secret_secondary(WOLFSSL* ssl,
     const unsigned char* secret, unsigned int secretSz);
 
 /*!
--- a/doc/dox_comments/header_files/ssl.h
+++ b/doc/dox_comments/header_files/ssl.h
@@ -8955,7 +8955,8 @@ int   wolfSSL_DTLS_SetCookieSecret(WOLFSSL* ssl,
 
     \param ssl a pointer to a WOLFSSL structure, created using wolfSSL_new().
     \param secret a constant byte pointer representing the secret buffer.
-    Passing NULL (or a secretSz of 0) clears any previously set secondary secret.
+    Passing NULL (or a secretSz of 0) clears any previously set secondary
+    secret.
     \param secretSz the size of the buffer.
 
     _Example_
@@ -14061,9 +14062,11 @@ int  wolfSSL_send_hrr_cookie(WOLFSSL* ssl,
     issue cookies.  This API is DTLS only; TLS 1.3 over a reliable transport
     does not verify cookies statelessly across the HelloRetryRequest exchange.
 
-    \param [in,out] ssl a pointer to a WOLFSSL structure, created using wolfSSL_new().
+    \param [in,out] ssl a pointer to a WOLFSSL structure, created using
+    wolfSSL_new().
     \param [in] secret a pointer to a buffer holding the secondary secret.
-    Passing NULL (or a secretSz of 0) clears any previously set secondary secret.
+    Passing NULL (or a secretSz of 0) clears any previously set secondary
+    secret.
     \param [in] secretSz Size of the secret in bytes.
 
     \return BAD_FUNC_ARG if ssl is NULL, not using TLS v1.3, or not using DTLS.
--- a/doc/dox_comments/header_files/ssl.h
+++ b/doc/dox_comments/header_files/ssl.h
@@ -9407,7 +9407,7 @@ int   wolfSSL_DTLS_SetCookieSecret(WOLFSSL* ssl,
     WOLFSSL* ssl = wolfSSL_new(ctx);
     const byte* oldSecret;
     word32 oldSecretSz; // size of oldSecret
-    …
+    ...
     if(wolfSSL_DTLS_SetCookieSecretSecondary(ssl, oldSecret, oldSecretSz) != 0){
     	// Code block for failure to set secondary DTLS cookie secret
     } else {
--- a/doc/dox_comments/header_files/ssl.h
+++ b/doc/dox_comments/header_files/ssl.h
@@ -9382,6 +9382,45 @@ int   wolfSSL_DTLS_SetCookieSecret(WOLFSSL* ssl,
                                                word32 secretSz);
 
 /*!
+    \brief This function sets a secondary DTLS 1.2 cookie secret used only when
+    verifying a received HelloVerifyRequest cookie, and only if the primary
+    secret (set with wolfSSL_DTLS_SetCookieSecret()) fails to verify it.  This
+    lets an application rotate the cookie secret on a stateless server without
+    rejecting clients whose cookie was issued under the previous secret: install
+    the new secret as the primary and the previous secret here for an overlap
+    window.  The secondary secret is never used to issue cookies.  It is the
+    DTLS 1.2 counterpart of wolfSSL_set_hrr_cookie_secret_secondary().
+
+    \return 0 returned if the function executed without an error.
+    \return BAD_FUNC_ARG returned if ssl is NULL.
+    \return MEMORY_ERROR returned if there was a problem allocating
+    memory for the secondary cookie secret.
+
+    \param ssl a pointer to a WOLFSSL structure, created using wolfSSL_new().
+    \param secret a constant byte pointer representing the secret buffer.
+    Passing NULL (or a secretSz of 0) clears any previously set secondary secret.
+    \param secretSz the size of the buffer.
+
+    _Example_
+    \code
+    WOLFSSL* ssl = wolfSSL_new(ctx);
+    const byte* oldSecret;
+    word32 oldSecretSz; // size of oldSecret
+    …
+    if(wolfSSL_DTLS_SetCookieSecretSecondary(ssl, oldSecret, oldSecretSz) != 0){
+    	// Code block for failure to set secondary DTLS cookie secret
+    } else {
+    	// Success! Secondary cookie secret is set.
+    }
+    \endcode
+
+    \sa wolfSSL_DTLS_SetCookieSecret
+*/
+int   wolfSSL_DTLS_SetCookieSecretSecondary(WOLFSSL* ssl,
+                                               const byte* secret,
+                                               word32 secretSz);
+
+/*!
     \brief This function retrieves the random number.
 
     \return rng upon success.
@@ -14464,6 +14503,51 @@ int  wolfSSL_connect(WOLFSSL* ssl);
     \sa wolfSSL_disable_hrr_cookie
 */
 int  wolfSSL_send_hrr_cookie(WOLFSSL* ssl,
+    const unsigned char* secret, unsigned int secretSz);
+
+/*!
+    \ingroup Setup
+
+    \brief This function sets a secondary HelloRetryRequest cookie secret on a
+    DTLS 1.3 server.  It is used only when verifying a received cookie, and only
+    if the primary secret (set with wolfSSL_send_hrr_cookie()) fails to verify
+    it.  This lets an application rotate the cookie secret on a stateless DTLS
+    1.3 server without rejecting clients whose cookie was issued under the
+    previous secret: install the new secret as the primary and the previous
+    secret here for an overlap window.  The secondary secret is never used to
+    issue cookies.  This API is DTLS only; TLS 1.3 over a reliable transport
+    does not verify cookies statelessly across the HelloRetryRequest exchange.
+
+    \param [in,out] ssl a pointer to a WOLFSSL structure, created using wolfSSL_new().
+    \param [in] secret a pointer to a buffer holding the secondary secret.
+    Passing NULL (or a secretSz of 0) clears any previously set secondary secret.
+    \param [in] secretSz Size of the secret in bytes.
+
+    \return BAD_FUNC_ARG if ssl is NULL, not using TLS v1.3, or not using DTLS.
+    \return SIDE_ERROR if called with a client.
+    \return WOLFSSL_SUCCESS if successful.
+    \return MEMORY_ERROR if allocating dynamic memory for storing secret failed.
+
+    _Example_
+    \code
+    int ret;
+    WOLFSSL* ssl;
+    char newSecret[32];
+    char oldSecret[32];
+    ...
+    // rotate: new secret becomes primary, previous secret stays valid
+    wolfSSL_send_hrr_cookie(ssl, newSecret, sizeof(newSecret));
+    ret = wolfSSL_set_hrr_cookie_secret_secondary(ssl, oldSecret,
+        sizeof(oldSecret));
+    if (ret != WOLFSSL_SUCCESS) {
+        // failed to set the secondary cookie secret
+    }
+    \endcode
+
+    \sa wolfSSL_send_hrr_cookie
+    \sa wolfSSL_disable_hrr_cookie
+*/
+int  wolfSSL_set_hrr_cookie_secret_secondary(WOLFSSL* ssl,
     const unsigned char* secret, unsigned int secretSz);
 
 /*!
--- a/doc/dox_comments/header_files/ssl.h
+++ b/doc/dox_comments/header_files/ssl.h
@@ -9398,7 +9398,8 @@ int   wolfSSL_DTLS_SetCookieSecret(WOLFSSL* ssl,
 
     \param ssl a pointer to a WOLFSSL structure, created using wolfSSL_new().
     \param secret a constant byte pointer representing the secret buffer.
-    Passing NULL (or a secretSz of 0) clears any previously set secondary secret.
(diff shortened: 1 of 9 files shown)


julek-wolfssl:
retest this please no history

Frauschi (dismissed):

src/ssl_api_dtls.c
```diff
@@ -1455,6 +1455,76 @@ int wolfSSL_DTLS_SetCookieSecret(WOLFSSL* ssl,
...
+    if (ssl->buffers.dtlsCookieSecretSecondary.buffer != NULL) {
+        ForceZero(ssl->buffers.dtlsCookieSecretSecondary.buffer,
+                  ssl->buffers.dtlsCookieSecretSecondary.length);
+        XFREE(ssl->buffers.dtlsCookieSecretSecondary.buffer,
+              ssl->heap, DYNAMIC_TYPE_COOKIE_PWD);
+        ssl->buffers.dtlsCookieSecretSecondary.buffer = NULL;
+        ssl->buffers.dtlsCookieSecretSecondary.length = 0;
+    }
+
+    /* A NULL/empty secret just clears the secondary secret. */
```
dgarske, line 1827: **🔵 [Low] Secondary DTLS 1.2 setter treats (non-NULL secret, size 0) as a silent clear, diverging from the primary setter** · _api / Logic_

The primary setter wolfSSL_DTLS_SetCookieSecret() returns BAD_FUNC_ARG when called with a non-NULL secret but secretSz == 0 ("can't have a new secret without a size"), whereas the new secondary setter treats the same (secret != NULL, secretSz == 0) call as a silent request to clear the secondary secret and returns 0. A caller who passes a valid secret pointer but mistakenly computes a 0 length silently drops the secondary secret instead of getting an error. The behavior is documented in the header ("NULL or a secretSz of 0 clears any previously set secondary secret") and is covered by tests, so this is a deliberate, documented contract rather than a bug; the DTLS 1.3 sibling wolfSSL_set_hrr_cookie_secret_secondary() shares the same clear-on-0 semantics. Both scan modes flagged this: review rated it Low/SUGGEST (api foot-gun), review-security rated it Info/Logic with no security impact (a zero-length secret would never be usable for issuance anyway). Stricter Low severity retained.

**Fix:** Optional: return BAD_FUNC_ARG for the (secret != NULL, secretSz == 0) case to mirror wolfSSL_DTLS_SetCookieSecret(), reserving the clear behavior for secret == NULL only. If the current clear-on-0 semantics are intended, no change is needed since it is documented and tested; optionally note the intentional difference in the doxygen \param text.

doc/dox_comments/header_files/ssl.h
```diff
@@ -8933,6 +8933,46 @@ int   wolfSSL_DTLS_SetCookieSecret(WOLFSSL* ssl,
...
+    Passing NULL (or a secretSz of 0) clears any previously set secondary
+    secret.
+    \param secretSz the size of the buffer.
+
+    _Example_
+    \code
+    WOLFSSL* ssl = wolfSSL_new(ctx);
+    const byte* oldSecret;
+    word32 oldSecretSz; // size of oldSecret
+    …
```
dgarske: **🔵 [Low] New Doxygen example introduces a non-ASCII ellipsis character** · _style_

The newly added \code example for wolfSSL_DTLS_SetCookieSecretSecondary() uses a Unicode horizontal-ellipsis character (U+2026) on line 8962 instead of an ASCII '...'. The sibling block added for wolfSSL_set_hrr_cookie_secret_secondary() (line ~14075) correctly uses ASCII '...', so this is an internal inconsistency introduced by the same PR. wolfSSL committed sources/comments are expected to be 7-bit ASCII.

**Fix:** Replace the U+2026 ellipsis with ASCII '...' to match the sibling block and keep the added content 7-bit ASCII.

src/tls13.c
```diff
@@ -6935,19 +6935,65 @@ static int CheckPreSharedKeys(WOLFSSL* ssl, const byte* input, word32 helloSz,
...
+ *
+ * ssl        SSL/TLS object.
+ * cookie     The cookie data - hash and MAC.
+ * dataSz     Length of the cookie data preceding the trailing MAC.
+ * secret     Secret to key the HMAC with.
+ * secretSz   Length of the secret in bytes.
+ * cookieType Digest type to use for the HMAC.
+ * macSz      Length of the MAC in bytes.
+ * returns 0 on match, HRR_COOKIE_ERROR on mismatch, otherwise a negative error.
+ */
```
dgarske, line 7045: **⚪ [Info] Orphaned/misattributed doc comment left above TlsCheckCookieMac** · _Logic_

The refactor moved the HMAC compute/compare into the new TlsCheckCookieMac(), but the original doxygen block describing TlsCheckCookie (ssl/cookie/cookieSz params, "returns Length of the hash on success") was left in place immediately above the new TlsCheckCookieMac comment. The result is two comment blocks stacked before TlsCheckCookieMac, and the public TlsCheckCookie function now has no doc comment of its own. Purely cosmetic; no behavioral impact.

**Fix:** Remove the stale comment block or move it back to directly precede TlsCheckCookie.

dgarske (commented):
## Skoll Multi-Scan Review

### Posted findings
- [Low] [review+review-security] **Secondary DTLS 1.2 setter treats (non-NULL secret, size 0) as a silent clear, diverging from the primary setter** — `src/ssl_api_dtls.c:1501-1505`
- [Low] [review] **New Doxygen example introduces a non-ASCII ellipsis character** — `doc/dox_comments/header_files/ssl.h:8962`
- [Info] [review-security] **Orphaned/misattributed doc comment left above TlsCheckCookieMac** — `src/tls13.c:6931-6950`

dgarske (changes requested):
Please resolve merge conflicts and skoll findings:
https://github.com/wolfSSL/wolfssl/pull/10634#pullrequestreview-4649719866

src/ssl_api_dtls.c
julek-wolfssl: addressed

doc/dox_comments/header_files/ssl.h
julek-wolfssl: addressed

src/tls13.c
julek-wolfssl: addressed

Frauschi:
Jenkins retest this please.

Frauschi (dismissed):

src/dtls.c
```diff
@@ -288,13 +285,30 @@ static int CheckDtlsCookie(const WOLFSSL* ssl, WolfSSL_CH* ch,
         if (ch->cookie.size != DTLS_COOKIE_SZ)
             return 0;
         if (!ch->dtls12cookieSet) {
-            ret = CreateDtls12Cookie(ssl, ch, ch->dtls12cookie);
+            ret = CreateDtls12Cookie(ssl, ch,
```
SparkiDev, line 308: An empty cookie returns COOKIE_ERROR.
This error is returned and the secondary isn't checked.

SparkiDev (dismissed):

src/dtls.c
julek-wolfssl: The primary cookie buffer must always be present. The secondary cookie buffer is only meant to allow for an overlap where two secrets are accepted.

src/tls13.c
```diff
@@ -14955,11 +15003,99 @@ int wolfSSL_disable_hrr_cookie(WOLFSSL* ssl)
...
+int wolfSSL_set_hrr_cookie_secret_secondary(WOLFSSL* ssl,
+    const unsigned char* secret, unsigned int secretSz)
+{
+    int ret;
+
+    if (ssl == NULL || !IsAtLeastTLSv1_3(ssl->version))
+        return BAD_FUNC_ARG;
+#ifndef NO_WOLFSSL_SERVER
+    if (ssl->options.side == WOLFSSL_CLIENT_END)
```
dgarske, line 15810: **🟠 [Medium] DTLS-only restriction leaves the new secondary branch in TlsCheckCookie() unreachable for TLS 1.3 over TCP**

The setter rejects any non-DTLS object with BAD_FUNC_ARG, but the verify code the PR modified is shared. `TlsCheckCookie()` is reached for plain TLS 1.3 too: at src/tls13.c:7889 the guard is `ssl->options.sendCookie && (ssl->options.serverState == SERVER_HELLO_RETRY_REQUEST_COMPLETE || (WOLFSSL_DTLS13 && ssl->options.dtls))`, so a TCP TLS 1.3 server that called `wolfSSL_send_hrr_cookie()` runs RestartHandshakeHashWithCookie() -> TlsCheckCookie() on CH2. That makes the whole `tls13CookieSecretSecondary` branch at src/tls13.c:7177-7185 dead code in a non-DTLS build, while still being compiled in (it is guarded only by WOLFSSL_SEND_HRR_COOKIE, not WOLFSSL_DTLS13). Separately, BAD_FUNC_ARG is now overloaded for three distinct conditions (NULL ssl, not TLS 1.3, not DTLS), so a caller cannot tell a programming error from an unsupported-transport rejection, and an application that shares one setup routine across TLS and DTLS listeners has to special-case the transport to avoid a spurious failure.

**Fix:** Decide whether the restriction is real. If it stays, tighten the compile guard on the secondary branch in TlsCheckCookie() so TLS-only builds do not carry dead code, and consider a distinct return code (or at least a clearer doc note) so callers can distinguish 'not DTLS' from a bad argument.

src/tls13.c
```diff
@@ -14955,11 +15003,99 @@ int wolfSSL_disable_hrr_cookie(WOLFSSL* ssl)
         ssl->buffers.tls13CookieSecret.length = 0;
     }
 
+    if (ssl->buffers.tls13CookieSecretSecondary.buffer != NULL) {
```
dgarske, line 15768: **🟠 [Medium] New free path in wolfSSL_disable_hrr_cookie() for the secondary secret is untested**

The PR adds a ForceZero + XFREE + reset block for `tls13CookieSecretSecondary` inside `wolfSSL_disable_hrr_cookie()`, but none of the 15 new tests calls `wolfSSL_disable_hrr_cookie()`. I confirmed this by running the new group: `./tests/unit.test --api -~cookie_secret` exercises every other new path but never touches this one. The test suite otherwise covers the equivalent free/replace paths well (the `_replaced` and `_cleared` tests are explicitly framed as double-free guards), so this is the one new memory path with no regression coverage. A future refactor that, say, moved the free without clearing the pointer would go unnoticed here until an ASAN/valgrind run on an unrelated test.

**Fix:** Add a case that sets the secondary, calls wolfSSL_disable_hrr_cookie(), sets it again, and disables again, so the new free path is covered under ASAN/WOLFSSL_CHECK_MEM_ZERO CI runs.

src/ssl_api_dtls.c
```diff
@@ -1542,6 +1542,76 @@ int wolfDTLS_SetChGoodCb(WOLFSSL* ssl, ClientHelloGoodCb cb, void* user_ctx)
...
+    if (secret == NULL || secretSz == 0) {
+        WOLFSSL_LEAVE("wolfSSL_DTLS_SetCookieSecretSecondary", 0);
+        return 0;
+    }
+
+    {
```
dgarske: **🔵 [Low] Bare block opened solely to declare a variable**

The allocation is wrapped in an anonymous `{ ... }` block that exists only so `byte* newSecret` can be declared there. wolfSSL C89 style puts declarations at the top of the function or of an existing block and does not open a new scope just to declare. The DTLS 1.3 sibling in src/tls13.c:15071 gets this right by declaring `newSecret` inside the existing `else` block, so the two functions added by the same PR are stylistically inconsistent with each other.

**Fix:** Hoist `byte* newSecret;` to the top of the function and drop the anonymous block.

src/internal.c
```diff
@@ -8969,6 +8969,12 @@ static void FreeSSL_DtlsResources(WOLFSSL* ssl)
     }
     XFREE(ssl->buffers.dtlsCookieSecret.buffer, ssl->heap,
           DYNAMIC_TYPE_COOKIE_PWD);
+    if (ssl->buffers.dtlsCookieSecretSecondary.buffer != NULL) {
```
dgarske, line 9445: **🔵 [Low] FreeSSL_DtlsResources() leaves dtlsCookieSecretSecondary dangling while the same PR adds defensive resets for the TLS…**

In `wolfSSL_ResourceFree()` the PR deliberately adds `buffer = NULL; length = 0;` after the XFREE for both `tls13CookieSecret` and `tls13CookieSecretSecondary` (src/internal.c:9268-9277) — a defensive change worth having, since `wolfSSL_ResourceFree` is a WOLFSSL_TEST_VIS symbol that ports call directly (wolfssl/internal.h:6794 notes "Micrium uses"). The new DTLS 1.2 free added a few lines earlier in the same PR does not do this, so `dtlsCookieSecretSecondary.buffer` is left pointing at freed memory with a non-zero length. It is not a live bug today (FreeSSL_DtlsResources has exactly one caller, src/internal.c:9280), but the PR is internally inconsistent about it.

**Fix:** Add the same NULL/length reset for dtlsCookieSecretSecondary (and, while there, for the pre-existing dtlsCookieSecret) so all four cookie-secret free sites behave identically.

dgarske (commented):
## Skoll Code Review

**Findings:** 11 total — 4 posted, 7 skipped

### Posted findings
- [Medium] **DTLS-only restriction leaves the new secondary branch in TlsCheckCookie() unreachable for TLS 1.3 over TCP** — `src/tls13.c:15048-15053`
- [Medium] **New free path in wolfSSL_disable_hrr_cookie() for the secondary secret is untested** — `src/tls13.c:15006-15013`
- [Low] **Bare block opened solely to declare a variable** — `src/ssl_api_dtls.c:1594-1609`
- [Low] **FreeSSL_DtlsResources() leaves dtlsCookieSecretSecondary dangling while the same PR adds defensive resets for the TLS 1.3 secrets** — `src/internal.c:8972-8977`

### Skipped findings
- [Medium] ~~Non-NULL secret with secretSz==0 silently clears the secondary instead of returning BAD_FUNC_ARG~~
- [Medium] ~~DTLS 1.2 and DTLS 1.3 verify paths disagree when the primary secret is absent~~
- [Low] ~~No coverage for a secondary secret whose length differs from the primary~~
- [Low] ~~Spurious HRR_COOKIE_ERROR trace emitted on the primary miss even when the secondary verifies~~
- [Low] ~~New tests inserted between a section-marker comment and the function it labels~~
- [Low] ~~Reject assertion duplicated instead of using the helper added ten lines later~~
- [Low] ~~Cookie verification cost doubles on the failure path when a secondary secret is configured~~

dgarske (changes requested):
https://github.com/wolfSSL/wolfssl/pull/10634#pullrequestreview-4848804467

dgarske:
Jenkins retest this please

dgarske (approved):

JacobBarthelmeh (approved):

PR #10639, opened 2026-06-08
Fix cipher property NIDs for SSL_get_current_cipher and add PSK kx mapping

julek-wolfssl:
### Description

The cipher property helpers (`SSL_CIPHER_get_kx_nid` / `get_auth_nid` /
`get_cipher_nid` / `get_digest_nid` / `is_aead`) parse the cipher name that
`GetCipherSegment()` looks up via `cipher->offset`. That `offset` is only
populated when the cipher is obtained through `wolfSSL_get_ciphers_compat()`
(`SSL_get_ciphers()`). When the cipher comes from `SSL_get_current_cipher()`,
`offset` is left at `0`, so these helpers parsed `cipher_names[0]` (a TLS 1.3
suite) instead of the negotiated cipher — e.g. returning `NID_kx_any` for a
plain PSK suite while `SSL_CIPHER_get_name()` (which uses the suite bytes)
reported the correct name.

### Changes

- `GetCipherSegment()`: resolve the `cipher_names` entry from the
  always-populated suite bytes (`cipherSuite0` / `cipherSuite`), falling back to
  `cipher->offset` when no match is found.
- `wolfSSL_CIPHER_get_kx_nid()`: add the missing plain `"PSK"` -> `NID_kx_psk`
  entry to the kx lookup table so PSK suites report `NID_kx_psk` instead of
  `NID_undef`.
- Add a regression test that drives the `SSL_get_current_cipher()` path for
  `TLS_PSK_WITH_AES_128_GCM_SHA256` and checks all five property helpers.

### Testing

Added an API regression test exercising the `SSL_get_current_cipher()` path.

ZD 21233

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -28749,6 +28749,29 @@ const char* GetCipherSegment(const WOLFSSL_CIPHER* cipher, char n[][MAX_SEGMENT_
 
     offset = cipher->offset;
 
+    /* The offset field is only populated when the cipher is obtained through
+     * wolfSSL_get_ciphers_compat() (e.g. SSL_get_ciphers()). When the cipher
+     * comes from wolfSSL_get_current_cipher() the offset is left at its default
+     * value of 0, which would make the parsing below operate on
+     * cipher_names[0] (a TLS 1.3 suite) rather than on the actual cipher. The
+     * suite bytes are always populated, so resolve the offset from them. */
+    {
+        int idx;
+        int namesSz = GetCipherNamesSize();
+        for (idx = 0; idx < namesSz; idx++) {
+            if (cipher_names[idx].cipherSuite0 == cipher->cipherSuite0 &&
+                cipher_names[idx].cipherSuite  == cipher->cipherSuite
+            #ifndef NO_CIPHER_SUITE_ALIASES
+                && (!(cipher_names[idx].flags &
+                      WOLFSSL_CIPHER_SUITE_FLAG_NAMEALIAS))
+            #endif
+                ) {
+                offset = (unsigned long)idx;
+                break;
+            }
+        }
+    }
+
     if (offset >= (unsigned long)GetCipherNamesSize())
         return NULL;
 
--- a/src/internal.c
+++ b/src/internal.c
@@ -28749,26 +28749,17 @@ const char* GetCipherSegment(const WOLFSSL_CIPHER* cipher, char n[][MAX_SEGMENT_
 
     offset = cipher->offset;
 
-    /* The offset field is only populated when the cipher is obtained through
-     * wolfSSL_get_ciphers_compat() (e.g. SSL_get_ciphers()). When the cipher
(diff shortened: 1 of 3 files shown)


src/internal.c
```diff
@@ -28749,6 +28749,20 @@ const char* GetCipherSegment(const WOLFSSL_CIPHER* cipher, char n[][MAX_SEGMENT_
...
+            && (!(cipher_names[i].flags & WOLFSSL_CIPHER_SUITE_FLAG_NAMEALIAS))
+        #endif
+            ) {
+            offset = (unsigned long)i;
+            break;
+        }
+    }
+
     if (offset >= (unsigned long)GetCipherNamesSize())
         return NULL;
```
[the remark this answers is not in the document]
julek-wolfssl: Not a critical section. Simple and always correct is better.

src/internal.c
```diff
@@ -28749,6 +28749,20 @@ const char* GetCipherSegment(const WOLFSSL_CIPHER* cipher, char n[][MAX_SEGMENT_
 
     offset = cipher->offset;
 
+    /* offset is not set via wolfSSL_get_current_cipher(), so resolve it from
```
dgarske, line 28752: **🔵 [Low] GetCipherSegment re-searches cipher_names unconditionally, duplicating existing lookup**

The new loop runs on every call, including for ciphers obtained via `wolfSSL_get_ciphers_compat()` where `cipher->offset` (and `cipher->in_stack`) are already set correctly. For those the search is redundant work that re-derives the same index. Because `wolfSSL_CIPHER_get_*_nid()`/`is_aead()` are commonly called once per cipher while iterating a full stack (as in `test_SSL_CIPHER_get_xxx`), this turns an O(1) offset lookup into an O(n) scan over `cipher_names` (~150 entries) per call, i.e. O(n^2) over a stack. This is a minor efficiency/cleanliness point, not a correctness bug: canonical (non-alias) entries always precede their `SUITE_ALIAS` entries, so the suite-byte scan returns the same index the stack path already stored, so no behavior changes for the existing path. Also note this loop duplicates the search already implemented in `GetCipherNameInternal()` and in `wolfSSL_get_ciphers_compat()`.

**Fix:** Optional. Consider guarding the scan on `!cipher->in_stack` (or `cipher->offset == 0`) to preserve the existing O(1) fast path for stack ciphers; behavior is unchanged either way. Not required for merge.

dgarske (commented):
## Skoll Code Review

### Posted findings
- [Low] **GetCipherSegment re-searches cipher_names unconditionally, duplicating existing lookup** — `src/internal.c:28752-28764`

### Skipped findings
- [Info] ~~Preprocessor/operator style inconsistent with adjacent code~~

src/internal.c
julek-wolfssl: Copilot found this earlier. I feel that the added complexity is not worth sacrificing the simplicity of the current solution.

dgarske (approved):

PR #10643, opened 2026-06-09
fix extension length underflow in sniffer hello parsers

netliomax25-code:
1. ProcessServerHello and ProcessClientHello walk the extension list with a word16 `len` taken from the 2-byte extensions length, but the per-extension guard only checks `extLen > *sslBytes` (bytes left in the capture), not the remaining `len`.
2. With trailing capture data past the extensions block, an `extLen` above the remaining `len` but still under `*sslBytes` passes that guard, so `len -= extLen + EXT_TYPE_SZ + LENGTH_SZ` wraps and the loop keeps reading extension headers past the end of the frame.
3. Bounded `extLen` against the remaining `len` at both sites so the subtraction cannot underflow.

Validated against an ASAN build of the loop: a ServerHello declaring an extensions length of 4 in a frame with 6 captured bytes and one extension of extLen 2 reads 2 bytes past the buffer before the change, and returns an input error cleanly after it.

### DIFF
--- a/src/sniffer.c
+++ b/src/sniffer.c
@@ -3812,8 +3812,11 @@ static int ProcessServerHello(int msgSz, const byte* input, int* sslBytes,
             input     += LENGTH_SZ;
             *sslBytes -= LENGTH_SZ;
 
-            /* make sure can read through individual extension */
-            if (extLen > *sslBytes) {
+            /* make sure can read through individual extension, and that it
+             * fits within the remaining declared extensions length so the
+             * len subtraction below cannot wrap */
+            if (extLen > *sslBytes ||
+                    extLen > len - (EXT_TYPE_SZ + LENGTH_SZ)) {
                 SetError(SERVER_HELLO_INPUT_STR, error, session,
                          FATAL_ERROR_STATE);
                 return WOLFSSL_FATAL_ERROR;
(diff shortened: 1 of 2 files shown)


netliomax25-code:
Added a regression test (`test_sniffer_serverhello_ext_len_underflow` in tests/api.c) so this doesn't go in without coverage.

1. Issue: ProcessServerHello/ProcessClientHello check `extLen` only against the bytes left in the capture, not the remaining declared `len`. With trailing data past the extensions block, an `extLen` above the remaining `len` but under `*sslBytes` passes, then `len -= extLen + EXT_TYPE_SZ + LENGTH_SZ` wraps and the loop keeps reading extension headers past the frame.
2. Fix: bound `extLen` against `len - (EXT_TYPE_SZ + LENGTH_SZ)` at both sites so the subtraction can't underflow.
3. Test: drives a SYN / SYN-ACK / ClientHello / ServerHello sequence through `ssl_DecodePacket`, where the ServerHello declares an extensions length of 4 but a single extension of extLen 16 with trailing bytes inside the handshake. The malformed hello is copied into a tightly sized heap buffer so the over-read is visible.

Validation: built with `--enable-sniffer`. Under an ASAN build the test trips a heap-buffer-overflow READ in the hello parser before this change and returns an input error cleanly after it. The existing static-rsa sniffer pcap still decodes.

netliomax25-code:
any update?

src/sniffer.c
```diff
@@ -3812,8 +3812,11 @@ static int ProcessServerHello(int msgSz, const byte* input, int* sslBytes,
             input     += LENGTH_SZ;
             *sslBytes -= LENGTH_SZ;
 
-            /* make sure can read through individual extension */
-            if (extLen > *sslBytes) {
+            /* make sure can read through individual extension, and that it
```
dgarske, line 3815: The sniffer does not need this type of check. Thanks for the report. Closing PR.

Issue #10359, opened 2026-04-30
DecodeSubjInfoAcc() rejects RFC 5280-compliant certs without id-ad-caRepository


jayharper:
## Summary

`DecodeSubjInfoAcc()` in `wolfcrypt/src/asn.c` returns `ASN_PARSE_E` when a Subject Information Access (SIA) extension does not contain an `id-ad-caRepository` (OID 1.3.6.1.5.5.7.48.5) access description. This is overly strict and rejects certificates that are structurally valid per RFC 5280.

RFC 5280 section 4.2.2.2 defines SIA as a `SEQUENCE OF AccessDescription` where each entry may use any access method OID — `id-ad-caRepository` is only one possibility. The current behavior appears to enforce the Federal PKI `fpki-x509-cert-profile-common` profile (section 5.3), which does require `caRepository`, but this is a profile-specific constraint and should not be applied as a generic parsing requirement.

## Real-world impact

ISO 15118-20 (Plug & Charge / EV charging) PKI certificates include SIA extensions with custom access-method OIDs defined under arc `2.40.246.14.20.0`. These certificates are RFC 5280-compliant but are currently rejected by wolfSSL with `ASN_PARSE_E`, preventing interoperability with ISO 15118-20 charging infrastructure.

Discovered while parsing contract certificate chains from a Keysight EV simulator on production SECC equipment running the wolfSSL-backed ISO 15118 stack.

## Reproduction

Any certificate whose SIA extension contains only non-`caRepository` access methods will trigger the rejection. Minimal repro:

1. Generate or obtain a certificate with an SIA extension containing only a custom access-method OID (e.g., any OID under `2.40.246.14.20.0` for ISO 15118-20, or any other non-`caRepository` AccessDescription).
2. Attempt to parse with `wc_ParseCert()` / `ParseCertRelative()`.
3. Parsing fails with `ASN_PARSE_E` at `wolfcrypt/src/asn.c:22423` (v5.8.2) 
   with `WOLFSSL_MSG("SubjectInfoAccess missing an URL.")` in the debug log.

## Expected behavior

The SIA extension should be parsed in full. If `id-ad-caRepository` entries are present they should be extracted as today; if none are present, the cert should still be accepted (a debug log message is appropriate, but not a hard failure).

## Environment

- wolfSSL version: v5.8.2
- Platform: Ubuntu 20.04, cross-compiled for 32-bit ARM (arm-linux-gnueabihf)
[...]

PR #10649, opened 2026-06-10
CI: optimize os-check.yml runner usage

dgarske:
## Why

`os-check.yml` is the largest GitHub-hosted runner consumer in the repo (~100 jobs per run, with macOS billed at 10x and Windows at 2x). This trims billable minutes without dropping any test coverage.

## Changes

- **Per-config ccache keys (Linux).** `make_check_linux` previously passed the default `config-hash: shared`, so all ~76 matrix entries computed the same cache key. With `actions/cache`, only the first job's save persists, so every later run restored an unrelated config's cache and mostly missed. Each entry now gets its own cache lineage via `config-hash: ${{ strategy.job-index }}` (`max-size 150M` to stay under the 10GB/repo cache cap; LRU handles the tail).
- **ccache on macOS.** The `ccache-setup` composite is now cross-platform (`brew install ccache`; PATH uses `$(brew --prefix)/opt/ccache/libexec`, which differs by runner arch). Wired into `make_check_macos` with post-build stats. Highest ROI since macOS bills at 10x.
- **Skip docs-only runs.** `paths-ignore: ['**/*.md', 'doc/**']` on both the `push` and `pull_request` triggers so documentation-only changes don't spin up the full matrix. The list is intentionally narrow (no cert/test-data extensions). Safe because no os-check job is a required status check.

## Not changed

- macOS and Linux matrices are left intact - analysis found no strictly-safe removals (each entry exercises a distinct compile-time `#ifdef` configuration where the absence of a define is itself the coverage).
- Windows still runs on every non-draft PR and on `release/**` pushes; personal-branch pushes never trigger this workflow.

## Verification

- YAML validated; rebased cleanly onto latest `master`.
- ccache effectiveness shows up after two consecutive runs on the same branch: the "ccache stats (post-build)" step hit rate should jump from near-0 to high, and macOS jobs now report ccache stats too.

### DIFF
--- a/.github/actions/ccache-setup/action.yml
+++ b/.github/actions/ccache-setup/action.yml
@@ -1,9 +1,10 @@
 name: 'Set up ccache'
 description: >
-  Install ccache (on Ubuntu), restore the ccache directory from a previous
-  run, and prepend the ccache compiler-symlink dir to PATH. Subsequent
-  gcc/cc/g++/c++ invocations are transparently intercepted by ccache, so
-  no other workflow step needs to change. macOS is not supported yet.
+  Install ccache (Ubuntu via apt, macOS via brew), restore the ccache
+  directory from a previous run, and prepend the ccache compiler-symlink
+  dir to PATH. Subsequent gcc/cc/g++/c++/clang invocations are
+  transparently intercepted by ccache, so no other workflow step needs to
+  change.
 
 inputs:
   workflow-id:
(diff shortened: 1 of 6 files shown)


dgarske:
Jenkins retest this please: "Build [PRB-fsanitize-addr-v3 #11416] completed: FAILURE" "Retry match found: 'java.io.InterruptedIOException'"

dgarske:
Jenkins retest this please: "Build [wolfSSL » PRB-fips-repo-and-harness-test-v3-part2 #10359] completed: FAILURE"

```CAVP_SHA384
cJSON_Compare against expected file 140-2-known-tests-expected/SHA2_384_47314-response.json
FAIL-ingish...
FAILED!```

julek-wolfssl (changes requested):
I will post a PR soon with hopefully even better optimizations. Let's not merge this just yet.

julek-wolfssl:
https://github.com/wolfSSL/wolfssl/pull/10667 has incorporated these changes and more.

PR #10655, opened 2026-06-10
add sp_hex_iszero for catching edge case of negative zero value

JacobBarthelmeh:
Fix for oss-fuzz report when handling a sp_int that has 'used' greater than 0 but has a magnitude of 0.

### DIFF
--- a/wolfcrypt/src/sp_int.c
+++ b/wolfcrypt/src/sp_int.c
@@ -18739,6 +18739,27 @@ int sp_read_radix(sp_int* a, const char* in, int radix)
 
 #if (defined(WOLFSSL_SP_MATH_ALL) && !defined(WOLFSSL_RSA_VERIFY_ONLY)) || \
     defined(WC_MP_TO_RADIX)
+/* Determine whether the magnitude of a is zero by scanning the digits.
+ *
+ * A non-normalized value (used >= 1 with all-zero digits, e.g. negative zero)
+ * is recognized as zero so that sp_radix_size() and sp_tohex() size and write
+ * the same canonical zero.
+ */
+static int sp_hex_iszero(const sp_int* a)
+{
+    int zero = 1;
+    unsigned int i;
+
+    for (i = 0; i < a->used; i++) {
+        if (a->dp[i] != 0) {
+            zero = 0;
+            break;
+        }
+    }
+
+    return zero;
+}
+
 /* Put the big-endian, hex string encoding of a into str.
  *
  * Assumes str is large enough for result.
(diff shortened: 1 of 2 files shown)


wolfcrypt/src/sp_int.c
```diff
@@ -18739,6 +18739,27 @@ int sp_read_radix(sp_int* a, const char* in, int radix)
 
 #if (defined(WOLFSSL_SP_MATH_ALL) && !defined(WOLFSSL_RSA_VERIFY_ONLY)) || \
     defined(WC_MP_TO_RADIX)
+/* Determine whether the magnitude of a is zero by scanning the digits.
+ *
+ * A non-normalized value (used >= 1 with all-zero digits, e.g. negative zero)
+ * is recognized as zero so that sp_radix_size() and sp_tohex() size and write
+ * the same canonical zero.
+ */
+static int sp_hex_iszero(const sp_int* a)
```
Frauschi, line 18772: **Suggestion (test coverage):** This fixes an oss-fuzz OOB read for a non-normalized zero (`used >= 1` with all-zero digits, e.g. negative zero), but adds no regression test. The existing `mp_tohex`/`mp_radix_size` tests in `wolfcrypt/test/test.c` only cover NULL params and normal values, so a future refactor could silently reintroduce the crash.

Consider adding a targeted test that constructs a non-normalized zero (set `a->used` to a non-zero value with all `a->dp[i] == 0`, optionally `sign == MP_NEG`) and asserts that `mp_radix_size()` and `mp_tohex()` agree on the canonical zero output. That locks in the fix.

SparkiDev (dismissed):

wolfcrypt/src/sp_int.c
JacobBarthelmeh: Added

JacobBarthelmeh:
Did a force push to pick up recent tests and wolfssl/master code. Added in a test case.

JacobBarthelmeh:
Retest this please Jenkins

philljj (changes requested):
Please review the buffer underflow fenrir item

wolfcrypt/src/sp_int.c
```diff
@@ -18785,7 +18806,7 @@ int sp_tohex(const sp_int* a, char* str)
 
     if (err == MP_OKAY) {
         /* Quick out if number is zero. */
-        if (sp_iszero(a) == MP_YES) {
+        if (sp_hex_iszero(a)) {
```
JacobBarthelmeh, line 18809: fixed and added a test case

philljj:
Tests good with skoll and multi-test sp-asm scenarios.

philljj (approved):

PR #10662, opened 2026-06-11
Fixes for Zephyr 4.4

Frauschi:
Minor fixes to make wolfssl usable as a module in Zephyr version 4.4.

Also add Zephyr 4.4 to the CI testing.


Frauschi:
Jenkins retest this please

.github/scripts/zephyr-4.x/zephyr-test.sh
```diff
@@ -70,9 +70,12 @@ select_docker_image() {
     local minor="${ver#*.}"
     minor="${minor%%.*}"
 
-    if [[ "$major" -ge 4 && "$minor" -ge 2 ]]; then
-        # Zephyr 4.2+ needs SDK 0.17.x (v0.28.7 image)
-        echo "${GHCR}:v0.28.7"
+    if [[ "$major" -ge 4 && "$minor" -ge 4 ]]; then
```
dgarske, line 73: **🟠 [Medium] Docker image selection ladder mis-routes future Zephyr 5.x (minor 4) to the oldest 4.x SDK image**

The new branch tests `major -ge 4 && minor -ge 4` and the existing branch tests `major -ge 4 && minor -ge 2`, treating major and minor independently. For a future `v5.0.0` or `v5.1.0` (major=5, minor=0/1), both compound conditions fail (`minor` is < 2), so control falls through to the `major -ge 4` branch and selects `v0.27.4` — the image documented as being for Zephyr 4.0-4.1 with SDK 0.17.0 — instead of the newest image. The comment claims the branch covers "Zephyr 4.4+" but the code does not implement ">= 4.4". This anti-pattern pre-existed with the 4.2 rung, but the PR adds a third rung to the same broken ladder, widening the mis-routing window (5.0-5.3 would now skip two newer images instead of one). Since the script otherwise anticipates future versions (`-ge` everywhere), the ladder should compare a combined version number.

**Fix:** Combine major/minor into a single comparable number (e.g. `major*100 + minor`) so each rung genuinely means ">= X.Y". This keeps the latest-image branch correct when Zephyr 5.0 appears.

zephyr/samples/wolfssl_tls_sock/prj.conf
```diff
@@ -2,7 +2,7 @@
 CONFIG_MAIN_STACK_SIZE=16384
 CONFIG_ENTROPY_GENERATOR=y
 CONFIG_INIT_STACKS=y
-CONFIG_COMMON_LIBC_MALLOC_ARENA_SIZE=8192
+CONFIG_COMMON_LIBC_MALLOC_ARENA_SIZE=131072
```
dgarske, line 5: **🟠 [Medium] 16x malloc arena increase in tls_sock prj.conf applies to all boards and duplicates the existing native_sim override**

`CONFIG_COMMON_LIBC_MALLOC_ARENA_SIZE` is raised from 8192 to 131072 in the base prj.conf. Two observations: (1) `zephyr/samples/wolfssl_tls_sock/boards/native_sim.conf` already sets `CONFIG_COMMON_LIBC_MALLOC_ARENA_SIZE=131072`, so for the native_sim target (the one actually executed in CI) this change is a no-op and the board override becomes redundant. (2) The effective change is therefore only for hardware boards such as `frdm_rw612/rw612`, which now statically reserve a 128 KB malloc arena where 8 KB previously sufficed for Zephyr 4.1/4.3 builds. The PR description does not explain why Zephyr 4.4 needs 16x more heap on embedded targets (ML-KEM allocations? picolibc malloc behavior change in SDK 1.x?). If only 4.4 or only certain boards need the larger arena, scoping it (board conf or a version overlay) would keep the sample usable on smaller parts; samples are copied verbatim by users onto far more RAM-constrained boards than the rw612's 1.2 MB SRAM.

**Fix:** Document why the sample needs a 128 KB arena (which allocation fails at 8 KB on 4.4), confirm the size is required on hardware boards and not just native_sim, and remove the now-duplicated override in boards/native_sim.conf.

.github/scripts/zephyr-4.x/zephyr-test.sh
```diff
@@ -70,9 +70,12 @@ select_docker_image() {
...
+        # Zephyr 4.4+ needs SDK 1.x (v0.29.2 image)
+        echo "${GHCR}:v0.29.2"
+    elif [[ "$major" -ge 4 && "$minor" -ge 2 ]]; then
```
dgarske, line 76: **🔵 [Low] Silent Docker image bump v0.28.7 - v0.28.8 for existing Zephyr 4.2/4.3 rows, with stale comment left behind**

Besides adding the 4.4 branch, the PR also changes the image used for the existing Zephyr 4.2-4.3 rows from `v0.28.7` to `v0.28.8`. That is a behavior change for already-passing CI cells and is not mentioned in the PR description ("Minor fixes to make wolfssl usable as a module in Zephyr 4.4"). It is exercised by the v4.3.0 matrix rows so any breakage would surface, but the motivation (image deprecation? SDK patch fix?) should be recorded. Additionally, the comment on the 4.0-4.1 branch still reads "v0.26.18/v0.28.7 picolibc is incompatible", referencing an image tag the script no longer uses anywhere, which will read as stale.

**Fix:** Mention the 4.2/4.3 image bump and its reason in the commit message or a comment, and refresh the v0.28.7 reference in the 4.0-4.1 branch comment so the rationale stays accurate.

dgarske (commented):
## Skoll Code Review

**Findings:** 5 total — 3 posted, 2 skipped

### Posted findings
- [Medium] **Docker image selection ladder mis-routes future Zephyr 5.x (minor  4) to the oldest 4.x SDK image** — `.github/scripts/zephyr-4.x/zephyr-test.sh:73-82`
- [Medium] **16x malloc arena increase in tls_sock prj.conf applies to all boards and duplicates the existing native_sim override** — `zephyr/samples/wolfssl_tls_sock/prj.conf:5`
- [Low] **Silent Docker image bump v0.28.7 - v0.28.8 for existing Zephyr 4.2/4.3 rows, with stale comment left behind** — `.github/scripts/zephyr-4.x/zephyr-test.sh:76-80`

### Skipped findings
- [Medium] ~~Undocumented and asymmetric Zephyr 4.4 guard: wolfio.h excludes posix/sys/socket.h while test.h still includes it (directly and transitively)~~
- [Medium] ~~DTLS / HAVE_SOCKADDR configuration on Zephyr 4.4 is not exercised by the new CI coverage~~

zephyr/samples/wolfssl_tls_sock/prj.conf
Frauschi: The old value actually made the sample fail on _any_ target except for the `native_sim`, which already had the 128 KB heap. In the sample, 48 KB of heap are used for the thread stack alone. On top of that, a lot of heap memory is required for the WOLFSSL_CTX and WOFLSSL objects as well as for crypto operations during the TLS handshake. 128 KB of heap results in a sound margin for all this, and the user can always override it when running the example on a target that has problems with this heap size.

dgarske (approved):

PR #10664, opened 2026-06-11
the examples pem.c --padding flag does not need gated behind WOLFSSL_DER_TO_PEM

JacobBarthelmeh:
Reproduce with : `./configure --enable-enckeys --enable-des3 && make check`

Issue was reported in the nightly CDT enable tests.

### DIFF
--- a/examples/pem/pem.c
+++ b/examples/pem/pem.c
@@ -657,11 +657,11 @@ const char* usage[] = {
     "  -o --offset      offset into file where data to convert starts",
 #if defined(WOLFSSL_ENCRYPTED_KEYS) && !defined(NO_PWDBASED)
     "  -p --pass        password to use with encrypted keys",
+    "  --padding        Remove padding on decrypted data",
 #endif
 #ifdef WOLFSSL_DER_TO_PEM
     "  -d --der         input is DER and output is PEM",
 #if defined(WOLFSSL_ENCRYPTED_KEYS) && !defined(NO_PWDBASED)
-    "  --padding        Remove padding on decrypted data",
     "  -e --encrypt     DER key is to be encrypted",
     "  -v --pbe-ver     PBE version to use when encrypting key (see below)",
     "  -p --pbe         PBE to use when encrypting key (see below)",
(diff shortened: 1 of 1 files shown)


JacobBarthelmeh:
Retest this please Jenkins. FIPS 140-2 failing test.

dgarske (approved):

examples/pem/pem.c
```diff
@@ -840,17 +840,19 @@ int main(int argc, char* argv[])
             info.passwd_userdata = argv[0];
         }
 #endif
+#if defined(WOLFSSL_ENCRYPTED_KEYS) && !defined(NO_PWDBASED)
+        /* Remove padding leftover from decryption. */
+        else if (strcmp(argv[0], "--padding") == 0) {
+            padding = 1;
+        }
+#endif /* WOLFSSL_ENCRYPTED_KEYS && !NO_PWDBASED */
```
dgarske, line 846: **🔵 [Low] Adjacent duplicate preprocessor guard for --padding option**

The fix correctly relocates the `--padding` parsing branch out of the `WOLFSSL_DER_TO_PEM` block, but it does so by opening a brand-new `#if defined(WOLFSSL_ENCRYPTED_KEYS) && !defined(NO_PWDBASED)` block (lines 843-848) immediately after the existing identical guard that handles `-p`/`--pass` (lines 828-842). The two adjacent blocks share the exact same condition, so the `--padding` `else if` could simply be appended inside the first block. Functionally identical; purely cosmetic.

**Fix:** Optionally fold the `--padding` branch into the preceding encrypted-keys `#if` block to avoid two back-to-back identical guards. Not blocking - current form compiles and behaves correctly in every configuration.

dgarske (commented):
## Skoll Code Review

### Posted findings
- [Low] **Adjacent duplicate preprocessor guard for --padding option** — `examples/pem/pem.c:828-848`

examples/pem/pem.c
dgarske: This would have been cleaner, but I am okay merged as-is.

dgarske (changes requested):

examples/pem/pem.c
JacobBarthelmeh: updated it, thanks

dgarske (approved):

PR #10665, opened 2026-06-11
Zeroize some private data before releasing memory

anhu:
Fixes ZD21954.

### DIFF
--- a/wolfcrypt/src/wc_mlkem.c
+++ b/wolfcrypt/src/wc_mlkem.c
@@ -965,11 +965,24 @@ int wc_MlKemKey_MakeKeyWithRandom(MlKemKey* key, const unsigned char* rand,
 #endif
     }
 
+    /* Zeroize the secret seed material in rho||sigma (sigma) before return. */
+    ForceZero(buf, sizeof(buf));
+#ifdef WC_MLKEM_FAULT_HARDEN
+    ForceZero(sigma, sizeof(sigma));
+#endif
+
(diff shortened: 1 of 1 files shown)


dgarske (approved):

dgarske (commented):
## Skoll Code Review

### Posted findings
- [Low] **Comment implies rho is secret seed material** — `wolfcrypt/src/wc_mlkem.c:968`

wolfcrypt/src/wc_mlkem.c
```diff
@@ -965,11 +965,24 @@ int wc_MlKemKey_MakeKeyWithRandom(MlKemKey* key, const unsigned char* rand,
 #endif
     }
 
+    /* Zeroize the secret seed material in rho||sigma (sigma) before return. */
```
dgarske, line 968: **🔵 [Low] Comment implies rho is secret seed material**

The comment `/* Zeroize the secret seed material in rho||sigma (sigma) before return. */` reads as if the whole `rho||sigma` buffer is secret. In fact `rho` is the *public* seed (it is copied to `key->pubSeed` at line 922 and is transmitted as part of the public key); only `sigma` is the secret seed. Clearing the full `buf` is correct and harmless, but the wording could mislead a future reader into thinking `rho` must be kept secret. The change itself is correct.

**Fix:** Optionally reword the comment to clarify that sigma is the secret part and rho is public. No functional change needed.

dgarske (approved):

Frauschi (approved):

PR #10676, opened 2026-06-12
20260610-linuxkm-fenrir

douzzer:
Fixes for numerous "low" Fenrir issues in linuxkm/.  See commit messages for the details.

tested with
```
wolfssl-multi-test.sh ...
--exclude-scenario '.*linuxkm-defaults.*'
'.*linuxkm.*'
check-source-text
```

### DIFF
--- a/linuxkm/lkcapi_aes_glue.c
+++ b/linuxkm/lkcapi_aes_glue.c
@@ -3140,7 +3140,7 @@ static int linuxkm_test_aescbc(void)
     req = skcipher_request_alloc(tfm, GFP_KERNEL);
     if (! req) {
         ret = -ENOMEM;
-        pr_err("error: allocating AES skcipher request %s failed\n",
+        pr_err("error: allocating AES skcipher request %s failed.\n",
                WOLFKM_AESCBC_DRIVER);
         goto test_cbc_end;
     }
@@ -3342,7 +3342,7 @@ static int linuxkm_test_aescfb(void)
     req = skcipher_request_alloc(tfm, GFP_KERNEL);
     if (! req) {
         ret = -ENOMEM;
-        pr_err("error: allocating AES skcipher request %s failed\n",
+        pr_err("error: allocating AES skcipher request %s failed.\n",
                WOLFKM_AESCFB_DRIVER);
         goto test_cfb_end;
     }
@@ -3608,8 +3608,8 @@ static int linuxkm_test_aesgcm(void)
     req = aead_request_alloc(tfm, GFP_KERNEL);
     if (! req) {
         ret = -ENOMEM;
-        pr_err("error: allocating AES aead request %s failed: %ld\n",
-               WOLFKM_AESCBC_DRIVER, PTR_ERR(req));
+        pr_err("error: allocating AES aead request %s failed.\n",
+               WOLFKM_AESCBC_DRIVER);
         goto test_gcm_end;
     }
 
@@ -4705,8 +4705,8 @@ static int aes_xts_256_test(void)
     req = skcipher_request_alloc(tfm, GFP_KERNEL);
     if (! req) {
         ret = -ENOMEM;
-        pr_err("error: allocating AES skcipher request %s failed: %d\n",
-               WOLFKM_AESXTS_DRIVER, ret);
+        pr_err("error: allocating AES skcipher request %s failed.\n",
+               WOLFKM_AESXTS_DRIVER);
         goto test_xts_end;
     }
 
--- a/linuxkm/lkcapi_aes_glue.c
+++ b/linuxkm/lkcapi_aes_glue.c
@@ -3616,16 +3616,14 @@ static int linuxkm_test_aesgcm(void)
     src = malloc(sizeof(struct scatterlist) * 2);
 
     if (! src) {
-        pr_err("error: malloc src failed: %ld\n",
-               PTR_ERR(src));
+        pr_err("error: malloc src failed,\n");
         goto test_gcm_end;
     }
 
     dst = malloc(sizeof(struct scatterlist) * 2);
 
     if (! dst) {
-        pr_err("error: malloc dst failed: %ld\n",
-               PTR_ERR(dst));
+        pr_err("error: malloc dst failed.\n");
         goto test_gcm_end;
     }
 
--- a/linuxkm/lkcapi_aes_glue.c
+++ b/linuxkm/lkcapi_aes_glue.c
@@ -3639,7 +3639,7 @@ static int linuxkm_test_aesgcm(void)
     src = malloc(sizeof(struct scatterlist) * 2);
 
     if (! src) {
-        pr_err("error: malloc src failed,\n");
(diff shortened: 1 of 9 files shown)


linuxkm/module_hooks.c
```diff
@@ -472,6 +472,7 @@ static WC_INLINE int IntelRDseed64_r(word64* rnd)
         WC_SANITIZE_ENABLE();
         buf = 0;
     }
+    wc_ForceZero(&buf, 0, sizeof buf);
```
dgarske: **🔴 [High] wc_ForceZero() called with 3 arguments (compile error / broken entropy zeroization fix)** · _Logic_

Both the review and review-security modes independently flag this. The newly added stack-entropy scrub call invokes `wc_ForceZero(&buf, 0, sizeof buf)` with THREE arguments, but `wc_ForceZero` is declared as `void wc_ForceZero(void *mem, size_t len)` (wolfssl/wolfcrypt/memory.h:335; defined wolfcrypt/src/memory.c:1681) and takes only TWO. The stray middle `0` looks like a `memset(&buf, 0, sizeof buf)` signature carried over by mistake. This is a C constraint violation that GCC reports as `error: too many arguments to function 'wc_ForceZero'`, and under the kernel build's -Werror it fails to compile. The same file already shows the correct 2-argument form at line 523: `wc_ForceZero(&rndTmp, sizeof(rndTmp));`. The function `IntelRDseed64_r` is compiled only under `WC_LINUXKM_RDSEED_IN_GLUE_LAYER` (guard at module_hooks.c:413-528) with `HAVE_INTEL_RDSEED`/`HAVE_AMD_RDSEED` (e.g. FIPS v5 RDSEED-in-glue builds), which is why the PR's tested configs (which took the `WC_LINUXKM_WOLFENTROPY_IN_GLUE_LAYER` branch at line 405) did not surface the break. Net effect: (1) the module will not build in the affected RDSEED-glue configs, and (2) the F-3294 entropy-zeroization hardening it implements is malformed. No 3-argument `wc_ForceZero`/`ForceZero` macro exists anywhere in the tree.

**Fix:** Drop the stray `0` argument so the call matches the 2-arg prototype: `wc_ForceZero(&buf, sizeof buf);`. Ideally add a CI config that defines WC_LINUXKM_RDSEED_IN_GLUE_LAYER (HAVE_INTEL_RDSEED / HAVE_AMD_RDSEED) so this path is compiled.

linuxkm/lkcapi_aes_glue.c
```diff
@@ -3608,24 +3631,24 @@ static int linuxkm_test_aesgcm(void)
...
         goto test_gcm_end;
     }
 
     src = malloc(sizeof(struct scatterlist) * 2);
 
     if (! src) {
-        pr_err("error: malloc src failed: %ld\n",
-               PTR_ERR(src));
+        pr_err("error: malloc src failed,\n");
+        ret = MEMORY_E;
```
dgarske, line 3643: **🔵 [Low] Trailing comma instead of period in error message** · _style_

The reworked malloc-failure message reads `pr_err("error: malloc src failed,\n");` with a trailing comma, whereas the sibling messages in the same function use a period (e.g. `"error: malloc dst failed.\n"`). Purely cosmetic; the removal of the bogus `PTR_ERR(src)` (which is meaningless for a NULL malloc return) and the added `ret = MEMORY_E;` are correct.

**Fix:** Change the trailing comma to a period for consistency: `pr_err("error: malloc src failed.\n");`. Not blocking.

linuxkm/lkcapi_rsa_glue.c
```diff
@@ -793,11 +797,12 @@ static int km_direct_rsa_dec(struct akcipher_request *req)
     }
 
     if (req->dst_len != (unsigned int)ctx->key_len) {
```
dgarske, line 799: **🔵 [Low] km_direct_rsa_dec no longer reports required dst_len on -EINVAL path** · _Logic_

The diff moved `req->dst_len = ctx->key_len;` inside the `-EOVERFLOW` branch. BEFORE, `req->dst_len` was set to `ctx->key_len` for both the `-EOVERFLOW` (dst_len too small) and `-EINVAL` (dst_len == 0 or dst_len > key_len) branches. AFTER, on the `-EINVAL` branch `req->dst_len` is left unchanged. This is a behavioral change but is benign and arguably more correct: the kernel akcipher convention only requires the driver to report the needed size on `-EOVERFLOW` so the caller can retry; `-EINVAL` is a hard error. No out-of-bounds or security impact -- flagged only as a behavior-delta note.

**Fix:** No change required if intentional. If callers were relying on the previous behavior of always receiving key_len in req->dst_len, restore the unconditional assignment; otherwise document the deliberate change.

linuxkm/x86_vector_register_glue.c
```diff
@@ -72,7 +72,7 @@ WARN_UNUSED_RESULT int allocate_wolfcrypt_linuxkm_fpu_states(void)
     if (! wc_linuxkm_fpu_states) {
```
dgarske, line 72: **⚪ [Info] x86_vector_register_glue.c allocation-failure log size argument changed** · _Logic_

The error message argument was changed from `nr_cpu_ids * sizeof(struct fpu_state *)` to `nr_cpu_ids * sizeof(wc_linuxkm_fpu_states[0])`, making the logged byte count track the actual element type of the allocation. This is a purely cosmetic logging accuracy improvement with no functional or security impact; noted for completeness.

**Fix:** No action needed; the change is correct.

dgarske (commented):
## Skoll Multi-Scan Review

### Posted findings
- [High] [review+review-security] **wc_ForceZero() called with 3 arguments (compile error / broken entropy zeroization fix)** — `linuxkm/module_hooks.c:475`
- [Low] [review] **Trailing comma instead of period in error message** — `linuxkm/lkcapi_aes_glue.c:3643`
- [Low] [review-security] **km_direct_rsa_dec no longer reports required dst_len on -EINVAL path** — `linuxkm/lkcapi_rsa_glue.c:799`
- [Info] [review-security] **x86_vector_register_glue.c allocation-failure log size argument changed** — `linuxkm/x86_vector_register_glue.c:72`

linuxkm/lkcapi_rsa_glue.c
douzzer: This was indeed intentional for API compliance.

linuxkm/lkcapi_aes_glue.c
douzzer: Fixed.  Typo obviously.

linuxkm/module_hooks.c
douzzer: Caught locally too.  Fixed, to commit&push in a minute.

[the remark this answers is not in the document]
douzzer: Fixed.

dgarske (approved):

douzzer:
retest this please
```
FAIL: scripts/ocsp-stapling-with-wolfssl-responder.test
```

PR #10677, opened 2026-06-12
Add RealTek AmebaPro2 (RTL8735B) HUK crypto-callback port

dgarske:
## Summary

Adds a wolfCrypt port that binds keys to the RealTek RTL8735B (AmebaPro2) silicon Hardware Unique Key (HUK) via the crypto-callback (CryptoCb) framework. A 256-bit "seed" is run through the AmebaPro2 HAL secure HKDF key-ladder against the HUK to derive a per-purpose working key in a secure key-storage slot, and the working key never enters software.

It is a pure crypto-callback device: no new wolfSSL public API and no new public struct fields. AES reads its seed from the standard `aes->devKey`, and ECDSA reads a port-defined `wc_Rtl8735b_EccKey` (the HUK-wrapped scalar plus seed) attached via the standard `ecc_key->devCtx`.

The bulk of the change is the new port under `wolfcrypt/src/port/realtek/`, but it does make a few minimal, `#if`-gated edits to shared files (see "Shared-file changes" below); it is not entirely self-contained.

## Features

- AES-GCM (full payload), AES-ECB, AES-CBC, and AES-CTR under the HUK-derived slot key. GCM/ECB use the HAL secure-key engine; CBC/CTR chain in software over the single-block ECB secure-key op (the HAL has no CBC/CTR secure-key variant), so the key still never leaves hardware. CBC handles in-place (`in == out`) and multi-call chaining; CTR maintains the partial-block keystream state across calls.
- HMAC-SHA256 over the HUK-derived secure-key slot (the working key never enters software). The message is accumulated and MAC'd one-shot at final; the accumulation buffer is carried correctly across `wc_HmacCopy`/`wc_HmacFree` via the cryptocb copy/free ops (requires `WOLF_CRYPTO_CB_COPY` and `WOLF_CRYPTO_CB_FREE`).
- HUK-bound ECDSA P-256 sign: the wrapped scalar is AES-GCM decrypt-verified under the HUK into a short-lived buffer, signed, and scrubbed. Because the wrap is authenticated (nonce + tag), a tampered, corrupted, or wrong-device blob fails at unwrap (`AES_GCM_AUTH_E`) rather than yielding a garbage scalar. The blob is device-bound (only unwraps on the silicon whose HUK produced the slot key). Requires `HAVE_AESGCM`.
- Optional HW ECDSA engine (`hal_ecdsa`) path for sign and verify, usable both HUK-bound and as a general P-256 offload for any `ecc_key` on `WC_HUK_DEVID`.
- The secure TRNG exposed as the wolfCrypt RNG seed source, so an application seeds its DRBG just by using the HUK device id.
- Unaligned caller buffers are tolerated (the HAL needs 32-byte-aligned DMA buffers; the port bounces through aligned temporaries). A non-12-byte GCM IV is a hard error (no silent software fallback that would key off the seed).

## Build options

- `WOLFSSL_RTL8735B_HUK` + `WOLF_CRYPTO_CB` enable the device (set in `user_settings.h`). Slot/device IDs are overridable (`WC_HUK_DEVID`, `WC_RTL8735B_*`). HMAC additionally needs `WOLF_CRYPTO_CB_COPY`/`WOLF_CRYPTO_CB_FREE`, and the wrapped-scalar ECDSA path needs `HAVE_AESGCM`.
- `--enable-rtl8735b` builds a host compile-and-self-test gate against a HAL shim (no vendor SDK needed): it sets `WOLFSSL_RTL8735B_HOST_TEST`, exercises the crypto-callback dispatch and build wiring, and runs helper KATs (see "Testing"). Note: it also enables AES-ECB build-wide (`HAVE_AES_ECB`) and the cryptocb copy/free ops, which is called out in the configure output.

## Shared-file changes

Minimal and `#if`-gated so they are inert for every other build:

- `wolfssl/wolfcrypt/wc_port.h`: add `WOLFSSL_RTL8735B_HUK` to the list of ports that auto-enable `WOLFSSL_CRYPT_HW_MUTEX`.
- `wolfcrypt/src/wc_port.c`: add a `WOLFSSL_MUTEX_INITIALIZER` (static-init) arm to `wolfSSL_CryptHwMutexUnLock` (needed on RTOS targets that statically initialize the mutex).
- `wolfcrypt/test/test.c`: a guarded call to `wc_Rtl8735b_HukSelfTest()` under `WOLFSSL_RTL8735B_HOST_TEST`.
- `.github/workflows/codespell.yml`: add "ameba" to the ignore-words list.
- Build wiring: `configure.ac`, `wolfcrypt/src/include.am`, `wolfssl/wolfcrypt/include.am`, `zephyr/CMakeLists.txt`.

Reviewers with STM32/MAX3266X configs that define `WOLFSSL_MUTEX_INITIALIZER` may want to confirm the `wolfSSL_CryptHwMutexUnLock` arm behaves as expected for those.

## Testing

- Host: `--enable-rtl8735b` builds clean and `testwolfcrypt` runs `wc_Rtl8735b_HukSelfTest()` (KATs for the silicon-independent helpers: CTR counter increment/rollover, big-endian/little-endian word conversion, HMAC accumulator growth/overflow/cap and the copy deep-copy + devId/heap carry-over, and bounce-buffer alignment). CI: `.github/workflows/rtl8735b.yml` builds several algo/guard and `WOLFSSL_SMALL_STACK` combinations and runs the self-test. Default build unaffected.
- RTL8735B silicon (RealTek FreeRTOS SDK and Zephyr): all of AES-GCM/ECB/CBC/CTR (including unaligned buffers, in-place and multi-call CBC, non-12-byte-IV reject), HMAC-SHA256, and HUK-bound ECDSA P-256 (GCM-wrapped scalar, HW-engine sign, and HW verify with good/tampered signatures) PASS.
- HUK-in-TLS: an in-memory TLS 1.2 ECDHE-ECDSA handshake on the RTL8735B, with the server certificate signature produced on the HUK engine via a PK sign callback (`wolfSSL_CTX_SetEccSignCb`), completes and exchanges data.

## Documentation

- `wolfcrypt/src/port/realtek/README.md`: hardware blocks used, enabling, API (AES plus the `devCtx` ECDSA usage), config macros, notes/limitations, on-target benchmarks, and optimization notes (`sp_cortexm.c`, Thumb-2 asm).
- `wolfssl/wolfcrypt/port/realtek/rtl8735b.h`: the port API and the `wc_Rtl8735b_EccKey` context struct.
- Companion SDK-integration examples in `wolfssl-examples` under `rtl8735b/`: a HUK crypto demo and the HUK-in-TLS handshake.


wolfssl/wolfcrypt/port/realtek/rtl8735b.h
```diff
@@ -0,0 +1,194 @@
...
+#endif
+
+/* Hook called once per ECDSA wait-loop iteration. The wait runs while the global
+ * crypto mutex is held, so on an RTOS a busy spin blocks other crypto threads;
+ * define this to a cooperative yield (e.g. taskYIELD() on FreeRTOS, k_yield() on
+ * Zephyr) so the core is released while the engine finishes. Default: no-op. */
+#ifndef WC_RTL8735B_ECDSA_YIELD
+    #define WC_RTL8735B_ECDSA_YIELD()     do {} while (0)
+#endif
+
```
night1rider, line 132: Should we add a global XYIELD macro to wolfSSL?

Looking at this reminds me of a similar situation I ran into while working on another hardware port. I ended up making a port-specific RTOS yield for IRQ waiting, but honestly, this seems like something that should be handled by a global wolfSSL macro instead of creating a new per-port macro every time. We already abstract things like XMALLOC and XMEMCPY, so XYIELD would fit perfectly into that architecture.

If we introduce a global XYIELD() in settings.h, port-specific macros like the one in this PR can simply map to it:
```
#ifndef WC_RTL8735B_ECDSA_YIELD
    #define WC_RTL8735B_ECDSA_YIELD()  XYIELD()
#endif
```
This approach gives the library a unified, global yield out of the box that automatically hooks into the correct RTOS, while still allowing individual ports to override it for specific algorithms if they really need to.

For context, the central implementation in settings.h could look something like this:
```
/* Global Yield Macro (XYIELD) */
#ifndef XYIELD
    #if defined(FREERTOS) || defined(FREERTOS_WOLFSSL)
        #define XYIELD() taskYIELD()
    #elif defined(WOLFSSL_ZEPHYR)
        #define XYIELD() k_yield()
    #elif defined(THREADX)
        #define XYIELD() tx_thread_relinquish()
    #else
        /* Default fallback for bare metal */
        #define XYIELD() do {} while (0)
    #endif
#endif
```

What do you think about doing it this way so we can reuse it across ports?

wolfcrypt/src/port/realtek/rtl8735b.c
```diff
@@ -0,0 +1,1767 @@
...
+    word32 newCap;
+    word32 dbl;
+
+    if (inSz == 0) {
+        return 0;
+    }
+    if (in == NULL) {
+        return BAD_FUNC_ARG;
+    }
+    ctx = (Rtl8735bHmacCtx*)hmac->devCtx;
```
night1rider: The port hangs its accumulation buffer on the top level `hmac->devCtx`. I don't think there's a dedicated cryptocb free op for that. `wc_HmacFree` only reclaims it by faking a finalize callback, and `wc_HmacCopy` seems to shallow copies it:

```
XMEMCPY(dst, src, sizeof(*dst));   /* dst->devCtx aliases src->devCtx */
```

So I think copying an Hmac after an Update, then freeing, double frees or corrupts the shared buffer.

Since the `devId` is passed down to the sha structs (`_InitHmac` -> `HmacKeyInitHash` -> `wc_InitHash`), I think you could instead store the buffer on the hash context `devCtx` (`hmac->hash.sha256.devCtx`) and create sha copy and free callbacks (`WC_ALGO_TYPE_COPY`/`FREE`) that deep copy and then free `hash.devCtx`. The real `wc_CryptoCb_Copy`/`wc_CryptoCb_Free` hooks already fire for it (`sha256.c:2976`/`:2638`), so that fixes the copy alias and gives a real free op. (Needs `WOLF_CRYPTO_CB_COPY`/`FREE` defined.) This is only worth doing if you think it's an issue.

wolfssl/wolfcrypt/port/realtek/rtl8735b.h
```diff
@@ -0,0 +1,194 @@
...
+ * are enabled in one build. Override before include if it collides. */
+#ifndef WC_HUK_DEVID
+    #define WC_HUK_DEVID                    810
+#endif
+
+/* Secure key-storage slot indices for the key ladder (KEY_STG_* values from
+ * rtl8735b_crypto_ctrl.h, NOT the hkdf_key_storage_e enum in rtl8735b_hkdf.h):
+ * the HUK source is HUK1 (0xC); slots 0..7 are general. The PRK and the derived
+ * working key each take a general slot -- the working-key slot is the one
+ * AES/HMAC *_sk_init reference. All overridable from user_settings. */
```
night1rider, line 66: Would it make sense to have a runtime selectable slot, if that is possible on this hardware? You could use the `_Id`/`_Label` init functions (`wc_AesInit_Id`, `wc_HmacInit_Id`/`wc_HmacInit_Label`, under `WOLF_PRIVATE_KEY_ID`) to pass the desired slot into the object as its own `id`/`label` field, and a SetKey callback would read that `id`/`label` to know where to derive the key. The AES/HMAC context then already carries which slot to use for encryption/MAC, so you would not hold the seed on the key context or re-derive on every op. Thoughts?

wolfcrypt/src/port/realtek/rtl8735b.c
```diff
@@ -0,0 +1,1774 @@
...
+#ifdef WOLFSSL_SMALL_STACK
+    XFREE(t, NULL, DYNAMIC_TYPE_TMP_BUFFER);
+#endif
+    return ret;
+}
+#endif /* HAVE_ECC_VERIFY */
+
+/* Unwrap the HUK-wrapped private scalar into scalar[] (scalarSz bytes). Derives
+ * the HUK slot key from the seed and ECB-sk decrypts the blob in place. */
+static int Rtl8735bHuk_UnwrapScalar(const wc_Rtl8735b_EccKey* hk,
```
night1rider: The wrapped ECDSA scalar is unwrapped with plain ECB, so a tampered, corrupted, or wrong device blob just decrypts to a garbage scalar. Is there a reason not to use the GCM port? This way a bad blob fails at unwrap instead of relying on the resulting signature failing verification downstream? I realize this would add an IV and tag to the blob format and require a change on the provisioning side.

wolfcrypt/src/port/realtek/rtl8735b.c
```diff
@@ -0,0 +1,1774 @@
...
+
+/* The HAL crypto engine DMAs its buffers on 32-byte (cache line) boundaries.
+ * Unaligned caller buffers are bounced through an aligned heap temp: over-
+ * allocate by 31 and round the usable pointer up; keep the raw pointer for XFREE. */
+#define WC_RTL8735B_IS_ALIGNED32(p) ((((wc_ptr_t)(p)) & 31u) == 0)
+#define WC_RTL8735B_ALIGN_UP32(p) \
+    ((byte*)((((wc_ptr_t)(p)) + 31u) & ~(wc_ptr_t)31u))
+/* Largest length we will bounce: leaves headroom so (sz + 31) cannot wrap.
+ * Anything larger is rejected (defensive; not reachable in device RAM). */
+#define WC_RTL8735B_BOUNCE_MAX  (0xFFFFFFFFUL - 32u)
```
night1rider, line 85: The port aligns DMA buffers to the 32 byte cache line but does no cache clean or invalidate around the HAL crypto calls. Is cache coherency a concern on this part, or does the HAL handle it?

night1rider (commented):
I went through the port and left notes below covering one likely issue, a couple of questions, and some suggestions. The issue is a potential HMAC devCtx double free on copy, the questions are `ECB` vs `GCM` for the wrapped scalar and whether cache coherency matters on this part, and the suggestions are adding a generic `XYEILD` to wolfSSL and a runtime selectable keyslot (if it makes sense).

wolfcrypt/src/port/realtek/rtl8735b.c
```diff
@@ -0,0 +1,1856 @@
...
+        }
+        XFREE(ctx, sha->heap, DYNAMIC_TYPE_HMAC);
+        sha->devCtx = NULL;
+    }
+}
+
+/* Deep-copy the accumulation buffer from src's SHA-256 devCtx to dst's (the
+ * WC_ALGO_TYPE_COPY handler). wc_Sha256Copy has already shallow-copied
+ * dst->devCtx = src->devCtx (an alias); replace it with an owned copy. */
+static int Rtl8735bHuk_HmacCopySha(wc_Sha256* src, wc_Sha256* dst)
```
night1rider, line 775: `wc_HmacCopy` overwrites the destination and zeroes its inner hash, and then `wc_Sha256Copy` calls the copy callback and returns as soon as `Rtl8735bHuk_HmacCopySha` returns 0. So the copy keeps the deep-copied `devCtx`, but its `hash.sha256.devId` is left at 0 instead of `WC_HUK_DEVID`.

`wc_HmacCopy` clobbers the destination, then zeroes its inner hash (`wolfcrypt/src/hmac.c`)
```
int wc_HmacCopy(Hmac* src, Hmac* dst) {
    ...
    XMEMCPY(dst, src, sizeof(*dst));              /* dst becomes a clone of src */
    XMEMSET(&dst->hash, 0, sizeof(wc_HmacHash));  /* inner hash zeroed: devId -> 0 */
    ret = HmacKeyCopyHash(src->macType, &src->hash, &dst->hash); /* -> wc_Sha256Copy */
    ...
}
```

And then because `wc_Sha256Copy` returns early on the handler's `0`, skipping the struct copy that would carry `devId` (`wolfcrypt/src/sha256.c`)

```
#if defined(WOLF_CRYPTO_CB) && defined(WOLF_CRYPTO_CB_COPY)
    if (src->devId != INVALID_DEVID) {
        ret = wc_CryptoCb_Copy(src->devId, WC_ALGO_TYPE_HASH,
                               WC_HASH_TYPE_SHA256, (void*)src, (void*)dst);
        if (ret != WC_NO_ERR_TRACE(CRYPTOCB_UNAVAILABLE))
            return ret;         /* handler returned 0 -> everything below is skipped */
    }
    ret = 0;
#endif
    wc_Sha256Free(dst);
    XMEMCPY(dst, src, sizeof(wc_Sha256));
```

Then, on free, `devId == 0` misroutes and the buffer is never released (`wolfcrypt/src/sha256.c`)

`wc_CryptoCb_Free(0, ...)` finds no device and returns `CRYPTOCB_UNAVAILABLE`, so the normal free path runs and the accumulation buffer is leaked.

The handler should carry over the fields the rest of wolfSSL relies on, so every later copy/free routes back through the port's handlers:

```
    dst->devId = src->devId;   /* route later copy/free back to the HUK handlers */
    dst->heap  = src->heap;    /* set before the XMALLOCs so alloc and free use one pool */
    d = (Rtl8735bHmacCtx*)XMALLOC(sizeof(Rtl8735bHmacCtx), dst->heap,
                                  DYNAMIC_TYPE_HMAC);
    ...
    dst->devCtx = d;
    return 0;
```

night1rider (dismissed):

dgarske:

night1rider (approved):

PR #10678, opened 2026-06-12
Fix: prevent public SNI from being installed on ctx

sebastian-carpenter:
# Description

Reworked the changeSNI logic so it never installs the public SNI onto the shared ctx.

# Testing

Added a regression test to check the plaintext SNI is always the public_name. Added matrix to verify ctx SNI swap works.
* test_wolfSSL_Tls13_ECH_wire_sni

# Checklist

 - [X] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/src/tls.c
+++ b/src/tls.c
@@ -16737,9 +16737,9 @@ static int TLSX_EchChangeSNI(WOLFSSL* ssl, TLSX** pEchX,
         }
 
         if (serverNameX == NULL && ssl->ctx && ssl->ctx->extensions) {
+            /* do not set extensions to ctx->extensions otherwise the shared
+             * list would be mutated */
             serverNameX = TLSX_Find(ssl->ctx->extensions, TLSX_SERVER_NAME);
-            if (serverNameX != NULL)
-                extensions = &ssl->ctx->extensions;
         }
 
         /* ECH requires an inner SNI to be present for ClientHelloInner.
@@ -16759,14 +16759,21 @@ static int TLSX_EchChangeSNI(WOLFSSL* ssl, TLSX** pEchX,
                 XMEMCPY(serverName, hostName, hostNameSz);
         }
 
-        /* only swap the SNI if one was found; extensions is non-NULL if an
-         * SNI entry was found on ssl->extensions or ctx->extensions */
-        if (ret == 0 && extensions != NULL) {
-            /* remove the inner server name */
-            TLSX_Remove(extensions, TLSX_SERVER_NAME, ssl->heap);
+        /* only swap the SNI if one was found, but only install onto
+         * ssl->extensions */
+        if (ret == 0) {
+            if (extensions != NULL) {
+                /* remove the inner server name */
+                TLSX_Remove(extensions, TLSX_SERVER_NAME, ssl->heap);
+            }
+            else {
+                /* inner SNI stays on the ctx; nothing to restore on ssl */
+                serverNameX = NULL;
+                extensions = &ssl->extensions;
+            }
 
             /* set the public name as the server name */
-            if ((ret = TLSX_UseSNI(extensions, WOLFSSL_SNI_HOST_NAME,
+            if ((ret = TLSX_UseSNI(&ssl->extensions, WOLFSSL_SNI_HOST_NAME,
                     ((WOLFSSL_ECH*)echX->data)->echConfig->publicName,
                     XSTRLEN(((WOLFSSL_ECH*)echX->data)->echConfig->publicName),
                     ssl->heap)) == WOLFSSL_SUCCESS)
--- a/tests/api.c
+++ b/tests/api.c
@@ -14061,6 +14061,63 @@ static word16 echCbTestKemID = 0;
 static word16 echCbTestKdfID = 0;
 static word16 echCbTestAeadID = 0;
 
+/* return the index of the first extension
+ * extLen will be updated with the length of the extensions field */
+static int ech_seek_extensions(byte* buf, word16* extLen)
+{
+    word16 idx;
+    byte sessionIdLen;
+    word16 cipherSuitesLen;
+    byte compressionLen;
+
+    idx = OPAQUE16_LEN + RAN_LEN;
+
+    sessionIdLen = buf[idx++];
+    idx += sessionIdLen;
+
+    ato16(buf + idx, &cipherSuitesLen);
+    idx += OPAQUE16_LEN + cipherSuitesLen;
+
+    compressionLen = buf[idx++];
+    idx += compressionLen;
+
+    ato16(buf + idx, extLen);
+    idx += OPAQUE16_LEN;
+
+    return idx;
+}
+
+/* locate a particular extension:
+ *   idx_p is updated with the location of that extension
+ *     -> idx_p should start just after the handshake header
+ *   0 returned on success, error otherwise */
+static int ech_find_extension(byte* buf, word16* idx_p, word16 extType)
+{
+    word16 idx;
+    word16 extIdx;
+    word16 extLen;
+
+    extIdx = ech_seek_extensions(buf + *idx_p, &extLen) + *idx_p;
+    idx = extIdx;
+
+    while (idx - extIdx < extLen) {
+        word16 type;
+        word16 len;
+
+        ato16(buf + idx, &type);
+        if (type == extType) {
+            *idx_p = idx;
+            return 0;
+        }
+
+        idx += OPAQUE16_LEN;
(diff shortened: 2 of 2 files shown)


tests/api.c
```diff
@@ -14999,6 +15056,141 @@ static int test_wolfSSL_Tls13_ECH_disable_conn(void)
...
+
+    /* check sent SNI is correct */
+    idx = RECORD_HEADER_SZ + HANDSHAKE_HEADER_SZ;
+    ExpectIntEQ(ech_find_extension(test_ctx.s_buff, &idx, TLSXT_SERVER_NAME),
+        0);
+    ExpectIntEQ(test_ctx.s_buff[idx + 6], WOLFSSL_SNI_HOST_NAME);
+    ato16(test_ctx.s_buff + idx + 7, &nameLen);
+    ExpectIntEQ(nameLen, publicLen);
+    ExpectIntEQ(XMEMCMP(test_ctx.s_buff + idx + 9, echCbTestPublicName,
+        publicLen), 0);
```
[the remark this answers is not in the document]
sebastian-carpenter: We know that the name will appear at the SNI extension. Having both public and private SNI's visible shouldn't be possible.

I don't expect a malformed extension to be given so I'm not worried about an OOB read.

tests/api.c
```diff
@@ -14999,6 +15056,141 @@ static int test_wolfSSL_Tls13_ECH_disable_conn(void)
...
```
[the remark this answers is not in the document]
sebastian-carpenter: see above

sebastian-carpenter:
Jenkins retest this please.

Frauschi (approved):

dgarske (commented):
## Skoll Multi-Scan Review

### Posted findings
- [Medium] [review] **useCtx test does not directly assert the shared ctx SNI is left unmutated** — `tests/api.c:15101-15157`
- [Low] [review+review-security] **Redundant inner-SNI copy into serverName buffer in the ctx-only branch (harmless dead work)** — `src/tls.c:16751-16773`
- [Low] [review] **TLSX_UseSNI hardcodes &ssl-extensions instead of the extensions local** — `src/tls.c:16764-16780`

tests/api.c
```diff
@@ -14999,6 +15056,141 @@ static int test_wolfSSL_Tls13_ECH_disable_conn(void)
...
+        ExpectIntEQ(wolfSSL_CTX_GenerateEchConfig(tempCtx, echCbTestPublicName,
+            0, 0, 0), WOLFSSL_SUCCESS);
+        ExpectIntEQ(wolfSSL_CTX_GetEchConfigs(tempCtx, badConfig,
+            &badConfigLen), WOLFSSL_SUCCESS);
+        wolfSSL_CTX_free(tempCtx);
+        ExpectIntEQ(wolfSSL_SetEchConfigs(test_ctx.c_ssl, badConfig,
+            badConfigLen), WOLFSSL_SUCCESS);
+    }
+
+    /* install the private (inner) SNI on the shared ctx or the per-connection
```
dgarske, line 15101: **🟠 [Medium] useCtx test does not directly assert the shared ctx SNI is left unmutated** · _test_

The `useCtx` matrix arm is intended to prove the fix (public SNI is never installed on the shared `ctx`). However, the test only inspects the outer ClientHello on the wire. The pre-fix code mutated `ssl->ctx->extensions` transiently and then restored it inside the same size/write cycle, so for a single sequential connection the outer SNI on the wire was still the public name. That means these wire assertions would also PASS against the old buggy code, so this arm does not actually guard against re-introducing the ctx-mutation regression -- it only confirms the new else-branch produces correct wire output. To make it a true regression test, additionally assert that `test_ctx.c_ctx`'s SERVER_NAME extension still holds `echCbTestPrivateName` after CH1/CH2 (e.g. via `TLSX_Find(c_ctx->extensions, TLSX_SERVER_NAME)` or a public SNI getter), confirming the shared list was not swapped to the public name.

**Fix:** Add a direct assertion that the shared client CTX SNI is unchanged (still the private name) after the ClientHello is produced, so the useCtx arm actually protects against the fixed regression rather than only exercising the new branch. For example, after the CH1 wire check verify TLSX_Find(test_ctx.c_ctx->extensions, TLSX_SERVER_NAME) resolves to echCbTestPrivateName.

src/tls.c
```diff
@@ -16759,14 +16759,21 @@ static int TLSX_EchChangeSNI(WOLFSSL* ssl, TLSX** pEchX,
...
+        if (ret == 0) {
+            if (extensions != NULL) {
+                /* remove the inner server name */
+                TLSX_Remove(extensions, TLSX_SERVER_NAME, ssl->heap);
+            }
+            else {
+                /* inner SNI stays on the ctx; nothing to restore on ssl */
+                serverNameX = NULL;
+                extensions = &ssl->extensions;
+            }
```
dgarske, line 16773: **🔵 [Low] Redundant inner-SNI copy into serverName buffer in the ctx-only branch (harmless dead work)** · _Logic_

When the inner SNI is found only on the shared `ssl->ctx->extensions`, the 'store the inner server name' step still runs (serverNameX is non-NULL at that point), copying the private name into `serverName` and enforcing `hostNameSz > WOLFSSL_HOST_NAME_MAX -> BAD_LENGTH_E`. Immediately afterward, the new `else` branch sets `serverNameX = NULL`, which causes `TLSX_EchRestoreSNI` to skip restoring from `serverName` entirely (the inner SNI legitimately stays on the untouched ctx list), so the copied buffer is unused on the ctx path. This is not a memory-safety defect: the copy is bounded (rejected with BAD_LENGTH_E before the copy) and the destination is a WOLFSSL_HOST_NAME_MAX allocation. One notable side effect flagged by the review mode: a ctx inner SNI at/above WOLFSSL_HOST_NAME_MAX would abort the handshake with BAD_LENGTH_E even though its value is never used on this path (only reachable with a pathologically long >= ~256 byte inner SNI, so not a practical defect). Severity views differ across modes -- review: Low, review-security: Info; the stricter Low is kept. The core change itself is correct: the BEFORE code mutated the shared ctx->extensions (removed the private SNI and installed the public name on it), which the AFTER code correctly avoids by installing the public name only on ssl->extensions and relying on the ssl-before-ctx semaphore dedup in TLSX_GetSize/TLSX_Write so the public name masks the private SNI still on ctx.

**Fix:** No action strictly required. Optionally, scope the inner-name copy/length-check to the ssl branch where the restore actually needs it (i.e., skip the serverName copy when the SNI originates from ctx), or add a comment noting the copy is intentionally unused on the ctx path. This makes the intent explicit and avoids an unnecessary BAD_LENGTH_E abort and dead work.

src/tls.c
```diff
@@ -16759,14 +16759,21 @@ static int TLSX_EchChangeSNI(WOLFSSL* ssl, TLSX** pEchX,
...
         }
 
-        /* only swap the SNI if one was found; extensions is non-NULL if an
-         * SNI entry was found on ssl->extensions or ctx->extensions */
-        if (ret == 0 && extensions != NULL) {
-            /* remove the inner server name */
-            TLSX_Remove(extensions, TLSX_SERVER_NAME, ssl->heap);
+        /* only swap the SNI if one was found, but only install onto
+         * ssl->extensions */
+        if (ret == 0) {
```
dgarske, line 16764: **🔵 [Low] TLSX_UseSNI hardcodes &ssl-extensions instead of the extensions local** · _style_

Inside the `if (ret == 0)` block, the code removes the inner SNI via `TLSX_Remove(extensions, ...)` (using the `extensions` local) but then adds the public SNI via `TLSX_UseSNI(&ssl->extensions, ...)` (hardcoded). In both branches `extensions` is guaranteed to equal `&ssl->extensions` at that point (the else-branch sets `extensions = &ssl->extensions`), so the behavior is correct, but mixing `extensions` and `&ssl->extensions` in the same block reads inconsistently and invites confusion if the code is later modified. Using `extensions` throughout would make the intent (always operate on ssl-level) explicit.

**Fix:** Use the `extensions` local in the TLSX_UseSNI call for consistency with the adjacent TLSX_Remove, since both refer to &ssl->extensions.

sebastian-carpenter: Will not change. &ssl->extensions is a necessary change since it explicitly states the list to mount onto. Using `extensions` instead encourages the issue.

tests/api.c
sebastian-carpenter: Ported over to #10568

sebastian-carpenter:
This logic is all reworked by #10568, so this PR is no longer necessary. Novel developments have been moved over.

PR #10685, opened 2026-06-15
CI: offload ccache/apt/buildx caches off the GitHub Actions cache

julek-wolfssl:
## Problem

The 10 GB, LRU-evicted, PR-scoped GitHub Actions cache was being thrashed: the
docker simulator buildx layers (~6 GiB), plus per-PR ccache and apt-archive
writes whose keys never hit, kept evicting the shared ccache, while the apt
mirror timed out often enough to break PR CI.

This moves the heavy caches to ghcr (free, separate pool) and makes PR runs
read-only against the Actions cache.

### apt dependencies from prebuilt ghcr `.deb` bundles
- `ci-deps-image.yml` resolves each package list under `.github/ci-deps/` into
  its `.deb` closure and publishes `ghcr.io/<owner>/wolfssl-ci-debs:<tag>` in
  two tiers: `<ver>-minimal` (make-check family) and `<ver>-full` (interop
  superset), for ubuntu-22.04 and 24.04.
- `install-apt-deps` gains a `ghcr-debs-tag` input: pull the bundle and install
  offline (`--no-download`) so the apt mirror is never on the PR critical path.
  Any failure (bundle missing/not public/incomplete) falls through to the
  existing apt path, so it is always safe to set.

### sim-test buildx layers to a shared ghcr registry cache
- The 7 docker simulator workflows switch from `cache-to: type=gha` to
  `ghcr.io/wolfssl/wolfssl-sim-cache:<scope>`. `cache-from` reads on every run
  (anonymous); `cache-to` writes only on the weekend cron and manual
  `workflow_dispatch`. Per-distinct-image tags and de-duplicated writers keep
  parallel matrix jobs from racing on one ref.

### ccache: PRs read, the schedule writes
- `ccache-setup` gains `read-only`: PR runs restore the shared master-scoped
  cache but never upload; schedule/push runs refresh it. Wired across os-check
  (linux + macOS), pq-all, smoke-test and the 12 small make-check workflows.
- `parallel-make-check.py` gains `--build-only` (compile every config, skip the
  test phase) so weekday-morning seed crons warm the cache PR runs consume.

### Misc
- Artifact retention capped at 7 days on the failure-log/result uploads that
  previously defaulted to 90.

## ⚠️ One-time setup

After their first publish, make the ghcr packages `wolfssl-ci-debs` and
`wolfssl-sim-cache` **public** so anonymous pulls work from PR (including fork)
runs. Until then everything falls back cleanly.


.github/workflows/atecc608-sim.yml
```diff
@@ -7,12 +7,21 @@ on:
...
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 # END OF COMMON SECTION
 
+permissions:
+  contents: read
+  packages: write
+
```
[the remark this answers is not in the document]
julek-wolfssl: Fork PRs (the untrusted ones) already get a read-only GITHUB_TOKEN from GitHub regardless of the permissions: block. Only same-repo PRs get write — and those come only from collaborators who already have repo write access (trusted).

.github/workflows/os-check.yml
```diff
@@ -81,8 +89,11 @@ jobs:
       - name: Pin ccache directory
         run: echo "CCACHE_DIR=$HOME/.cache/ccache" >> "$GITHUB_ENV"
 
+      # PRs restore the cache the weekday seed writes but never save it, so
+      # PR runs add no per-shard ccache entries to the Actions cache. The
+      # seed (schedule) saves below.
       - name: Restore ccache
```
dgarske, line 95: **🟠 [Medium] release/** pushes no longer refresh the os-check linux ccache**

Previously the linux shard used `actions/cache@v4`, which saves its contribution at post-job on every triggering event including `push` to `release/**`. The PR switches to `actions/cache/restore@v4` plus a save step gated to `if: github.event_name == 'schedule'`. As a result, pushes to `release/**` now restore but never save, so release-branch CI relies entirely on the broad restore-keys fallback to the master seed and never refreshes a release-scoped lineage. This is almost certainly the intended "seed writes, everything else reads" model, but it is a behavior change for release-branch runs worth confirming.

**Fix:** Confirm release-branch runs are meant to consume only the master seed. If release branches should keep their own warm cache, widen the save condition (e.g. `github.event_name == 'schedule' || github.event_name == 'push'`).

.github/ci-deps/packages-ubuntu-24.04-minimal.txt
```diff
@@ -0,0 +1,20 @@
+# make-check family apt packages for ubuntu-24.04 (the '-minimal'
```
dgarske, line 1: **🔵 [Low] No CI guard keeping consumer 'packages' a subset of the ghcr bundle lists**

The offline install runs `apt-get install --no-download \<packages>` against the bundle; if even one requested package is absent from the bundle tag, the whole offline install fails and silently falls back to apt (only a `::notice::`), defeating the optimization without any visible failure. All current consumers were verified to be subsets of their bundle lists, but the lists are maintained by hand (the file headers ask maintainers to keep them sorted/in sync). A new workflow that adds a package and forgets to add it to the corresponding `.github/ci-deps/packages-*.txt` would degrade to apt undetected.

**Fix:** Add a lightweight CI check (a script in ci-deps-image.yml or a standalone lint job) that parses each workflow's `packages` + `ghcr-debs-tag` and asserts membership in the named list file, failing if any requested package is missing from the bundle.

.github/actions/ccache-setup/action.yml
```diff
@@ -52,6 +66,16 @@ runs:
         restore-keys: |
           ccache-${{ inputs.workflow-id }}-${{ runner.os }}-${{ runner.arch }}-${{ inputs.config-hash }}-
           ccache-${{ inputs.workflow-id }}-${{ runner.os }}-${{ runner.arch }}-
+    # read-only=true: restore the shared cache but never upload (PR runs).
+    - name: Restore ccache (read-only)
```
dgarske, line 70: **🔵 [Low] Dead run_id/run_attempt suffix in the read-only ccache restore key**

In the read-only branch the primary `key` includes `${{ github.run_id }}-${{ github.run_attempt }}`, which is unique to the current run and was never used to save any cache (the read-only path never saves). It therefore can never produce an exact hit and always falls through to `restore-keys`. This works, but the run_id/run_attempt suffix is dead and slightly misleading in a restore-only step; the primary key could simply be the config-hash restore key.

**Fix:** Drop the run_id/run_attempt from the read-only primary key (or leave as-is for symmetry with the save path, with a one-line comment that it is intentionally never an exact hit).

.github/actions/install-apt-deps/action.yml
```diff
@@ -20,11 +20,60 @@ inputs:
...
+    # image must be PUBLIC so anonymous `docker pull` works (including from
+    # fork PRs whose GITHUB_TOKEN cannot read private packages).
+    - name: Install from ghcr .deb bundle (offline)
+      id: ghcr
+      if: inputs.ghcr-debs-tag != ''
+      shell: bash
+      run: |
+        set -u
+        command -v docker >/dev/null 2>&1 || { echo "::notice::docker unavailable; using apt"; exit 0; }
+        OWNER=$(echo "${{ github.repository_owner }}" | tr '[:upper:]' '[:lower:]')
```
dgarske: **🔵 [Low] Inconsistent ghcr owner reference between install-apt-deps and sim workflows**

`install-apt-deps` derives the bundle image owner from `${{ github.repository_owner }}` (lowercased), so a fork PR looks up `ghcr.io/\<fork-owner>/wolfssl-ci-debs`, which won't exist, and always falls back to apt. The sim workflows instead hardcode `ghcr.io/wolfssl/wolfssl-sim-cache:...` so fork PRs read the upstream cache. Both behaviors are safe (apt fallback works; sim cache-from failures are non-fatal), but the two approaches differ. Using `github.repository_owner` for apt-debs means forks never benefit from the bundle even though the public upstream image exists.

**Fix:** Decide on one convention. If forks should also use the public bundle, hardcode the `wolfssl` owner (matching the sim workflows); otherwise leave as-is and add a brief comment noting forks intentionally fall back to apt.

.github/workflows/os-check.yml
dgarske, line 95: **🟠 [Medium] Scheduled ccache reseed (CCACHE_RECACHE) does not cover the os-check linux shards**

The new `ccache-setup` composite forces clean recompiles on `schedule` via `CCACHE_RECACHE=1` specifically so the shared cache "can't drift indefinitely" (per its own description). The `make_check_macos` half of os-check uses that composite and therefore reseeds cleanly. But `make_check_linux` manages ccache directly (its own `actions/cache/restore` + `Save ccache (seed only)` steps) and never sets `CCACHE_RECACHE`, so its weekday seed restores the previous master cache and only layers deltas onto it. The two halves of the same workflow now behave inconsistently, and the linux shared cache retains the very drift the reseed mechanism was added to prevent. Practical impact is bounded by ccache's own LRU (`CCACHE_MAXSIZE: 500M`), so this is not a correctness blocker.

**Fix:** Add an equivalent `CCACHE_RECACHE=1` step (gated to `github.event_name == 'schedule'`) to the os-check linux job before the build, or document why incremental seeding is acceptable for linux while macOS reseeds cleanly.

---
> **Note:** Referenced line (`.github/workflows/os-check.yml:95-106,401-406`) is outside the diff hunk. Comment anchored to nearest changed region.

dgarske (commented):
## Skoll Code Review

### Posted findings
- [Medium] **Scheduled ccache reseed (CCACHE_RECACHE) does not cover the os-check linux shards** — `.github/workflows/os-check.yml:95-106,401-406`
- [Medium] **release/** pushes no longer refresh the os-check linux ccache** — `.github/workflows/os-check.yml:95-96,401-406`
- [Low] **No CI guard keeping consumer 'packages' a subset of the ghcr bundle lists** — `.github/ci-deps/packages-ubuntu-24.04-minimal.txt:1-20`
- [Low] **Dead run_id/run_attempt suffix in the read-only ccache restore key** — `.github/actions/ccache-setup/action.yml:70-78`
- [Low] **Inconsistent ghcr owner reference between install-apt-deps and sim workflows** — `.github/actions/install-apt-deps/action.yml:50-51`

.github/workflows/os-check.yml
julek-wolfssl: Intended. The model is "schedule seeds, everything else reads": PRs and release/** pushes restore the master-seeded ccache but never save — only the weekday cron writes. Release branches consume the master seed rather than keeping a separate lineage, so this behavior change is by design. Leaving as-is.

.github/ci-deps/packages-ubuntu-24.04-minimal.txt
julek-wolfssl: Deferring to a follow-up. A package missing from the bundle is non-fatal — the offline --no-download install just fails and falls back to apt (slower, still correct), so there's no silent breakage. A real guard means parsing every workflow's install-apt-deps packages + ghcr-debs-tag and asserting subset membership, which is more than this PR should take on. Tracking it separately.

.github/actions/ccache-setup/action.yml
julek-wolfssl: Addressed.

.github/actions/install-apt-deps/action.yml
julek-wolfssl: Addressed.

.github/workflows/os-check.yml
julek-wolfssl: Addressed.

dgarske (approved):

PR #10688, opened 2026-06-15
20260615-linuxkm-fenrir-fixes

douzzer:
fixes for F-5956 F-5957 F-5958 F-1236 F-3291 F-1431

tested with
```
wolfssl-multi-test.sh ...
check-source-text
quantum-safe-wolfssl-all-crypto-only-noasm-linuxkm-mainline-clang-tidy
quantum-safe-wolfssl-all-crypto-only-intelasm-sp-asm-sp-linuxkm-mainline-clang-tidy
quantum-safe-wolfssl-all-crypto-only-intelasm-sp-asm-fips-dev-linuxkm-next-clang-tidy
quantum-safe-wolfssl-all-crypto-only-intelasm-sp-asm-sp-linuxkm-6.12-insmod
quantum-safe-wolfssl-all-crypto-only-intelasm-fips-dev-linuxkm-next-insmod
'.*cust-kernel-.*'
```

### DIFF
--- a/linuxkm/lkcapi_sha_glue.c
+++ b/linuxkm/lkcapi_sha_glue.c
@@ -1835,8 +1835,8 @@ static int wc_linuxkm_drbg_startup(void)
             u8 buf1[16], buf2[17];
             int i, j;
 
-            memset(buf1, 0, sizeof buf1);
-            memset(buf2, 0, sizeof buf2);
+            XMEMSET(buf1, 0, sizeof buf1);
+            XMEMSET(buf2, 0, sizeof buf2);
 
             ret = crypto_rng_generate(tfm, NULL, 0, buf1, (unsigned int)sizeof buf1);
             if (! ret)
@@ -1858,7 +1858,7 @@ static int wc_linuxkm_drbg_startup(void)
                  */
                 for (i = 1; i <= (int)sizeof buf2; ++i) {
                     for (j = 0; j < 20; ++j) {
-                        memset(buf2, 0, (size_t)i);
+                        XMEMSET(buf2, 0, (size_t)i);
                         ret = crypto_rng_generate(tfm, NULL, 0, buf2, (unsigned int)i);
                         if (ret)
                             break;
--- a/linuxkm/lkcapi_sha_glue.c
+++ b/linuxkm/lkcapi_sha_glue.c
@@ -1230,10 +1230,16 @@ static int wc_linuxkm_drbg_generate(struct wc_rng_bank *ctx,
         if (ret == 0)
             continue;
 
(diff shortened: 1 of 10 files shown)


linuxkm/lkcapi_sha_glue.c
```diff
@@ -1230,10 +1230,16 @@ static int wc_linuxkm_drbg_generate(struct wc_rng_bank *ctx,
...
 
-        if (unlikely(ret == WC_NO_ERR_TRACE(RNG_FAILURE_E)) && (! retried)) {
-            if (slen > 0)
+        if (unlikely(ret == WC_NO_ERR_TRACE(RNG_FAILURE_E))) {
+            if (slen > 0) {
+                ret = -EINVAL;
                 break;
+            }
 
+            if (retried) {
```
dgarske, line 1239: **🔵 [Low] DRBG retry-exhaustion path no longer logs the second RNG failure**

The refactor of the `RNG_FAILURE_E` handling is functionally correct and improves the returned error code (now `-EINVAL` instead of the raw wolfSSL `RNG_FAILURE_E` value). However, there is a small loss of observability: in the previous code, a second `RNG_FAILURE_E` (after a successful reinit and `continue`) fell through to the `else` branch and emitted `pr_warn_once("ERROR: ... wc_RNG_GenerateBlock returned %d")`. In the new code that case is caught by `if (retried) { ret = -EINVAL; break; }`, which breaks silently with no log. The reinit-failure path still logs, but a generate failure that recurs after a successful reinit is now silent. This is cosmetic only and may be intentional.

**Fix:** Optionally add a pr_warn_once() in the `if (retried)` branch so a persistent RNG failure after reinit is still visible in the kernel log. Not blocking.

dgarske (commented):
## Skoll Code Review

### Posted findings
- [Low] **DRBG retry-exhaustion path no longer logs the second RNG failure** — `linuxkm/lkcapi_sha_glue.c:1239-1242`

philljj (approved):

linuxkm/lkcapi_sha_glue.c
douzzer: No need to bikeshed this -- `retried` is only true if the `RNG_FAILURE_E` code has already been evaluated, and if it's evaluated it will always log a successful reinit, and always log the first failure.  If `wc_RNG_GenerateBlock()` fails again, that second failure will indeed be silent, but that's OK.  The successful recovery path always logs, and the failure paths use `pr_warn_once()` as those are unrecoverable errors likely to spam the kernel log to death without the throttling (experience suggests...).  Separate observation: those errors should be `pr_err_once()`.  But not worth holding up the PR for it -- problem is preexisting.

JacobBarthelmeh (approved):

PR #10689, opened 2026-06-15
.github: bump JavaScript actions to Node.js 24 runtimes

julek-wolfssl:
GitHub Actions now emits "Node.js 20 actions are deprecated" warnings: actions are forced to Node.js 24 by default starting 2026-06-16, and Node.js 20 is removed from the runners on 2026-09-16. This updates every JavaScript action referenced by the workflows and the local composite actions to the lowest release that runs on Node.js 24:

| Action | From | To | Note |
|--------|------|----|------|
| `actions/checkout` | v4 | v5 | |
| `actions/checkout` (SHA pin) | v4.1.7 | v5 | |
| `actions/upload-artifact` | v4 | v6 | v5 still Node.js 20 |
| `actions/download-artifact` | v4 | v7 | v5/v6 still Node.js 20 |
| `actions/cache[/restore\|/save]` | v4 | v5 | |
| `actions/setup-python` | v5 | v6 | |
| `actions/github-script` | v7 | v8 | |
| `docker/setup-buildx-action` | v3 | v4 | |
| `docker/build-push-action` | v5 | v7 | v6 still Node.js 20 |
| `docker/login-action` | v3 | v4 | |
| `microsoft/setup-msbuild` | v2 | v3 | |
| `open-watcom/setup-watcom` | v0 | v1 | |

Actions already running on Node.js 24 (`jwlawson/actions-setup-cmake`, `shogo82148/actions-setup-perl`, `msys2/setup-msys2`, `dorny/paths-filter`) are left unchanged.

These bumps are runtime-only; no workflow uses an input or output removed by the new majors, and v4-format artifacts remain compatible across the upload v6 / download v7 backends.

### DIFF
--- a/.github/workflows/membrowse-zephyr.yml
+++ b/.github/workflows/membrowse-zephyr.yml
@@ -48,7 +48,7 @@ jobs:
 
       - name: Download Zephyr build artifact
         id: download
-        uses: actions/download-artifact@v4
+        uses: actions/download-artifact@v7
         with:
           name: ${{ matrix.artifact }}
           path: zephyr-artifacts/${{ matrix.target_name }}
--- a/.github/actions/ccache-setup/action.yml
+++ b/.github/actions/ccache-setup/action.yml
@@ -42,7 +42,7 @@ runs:
         fi
 
     - name: Restore + save ccache
-      uses: actions/cache@v4
+      uses: actions/cache@v5
       with:
         path: ~/.ccache
         # Unique per run+attempt+config so each job persists its own
--- a/.github/actions/install-apt-deps/action.yml
+++ b/.github/actions/install-apt-deps/action.yml
@@ -37,7 +37,7 @@ runs:
     - name: Restore apt cache
       if: inputs.cache == 'true'
       id: apt-cache
-      uses: actions/cache/restore@v4
+      uses: actions/cache/restore@v5
       with:
         path: ~/apt-cache
         key: ${{ steps.cache-key.outputs.key }}
@@ -100,7 +100,7 @@ runs:
 
     - name: Save apt cache
       if: inputs.cache == 'true' && steps.apt-cache.outputs.cache-hit != 'true'
-      uses: actions/cache/save@v4
+      uses: actions/cache/save@v5
       with:
         path: ~/apt-cache
         key: ${{ steps.cache-key.outputs.key }}
--- a/.github/workflows/ada.yml
+++ b/.github/workflows/ada.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-24.04
 
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v5
 
     - name: Install alire
       uses: alire-project/setup-alire@v5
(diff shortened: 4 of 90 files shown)


.github/workflows/membrowse-zephyr.yml
```diff
@@ -48,7 +48,7 @@ jobs:
 
       - name: Download Zephyr build artifact
         id: download
-        uses: actions/download-artifact@v4
+        uses: actions/download-artifact@v7
```
dgarske, line 51: **⚪ [Info] Cross-workflow artifact pairing now spans upload-artifact v6 and download-artifact v7**

`membrowse-zephyr.yml` runs on `workflow_run` after `zephyr-4.x.yml` and downloads artifacts (`membrowse-zephyr-native_sim` / `membrowse-zephyr-frdm_rw612`) produced in a *different* workflow run. After this PR the producer uses `actions/upload-artifact@v6` (zephyr-4.x.yml:94) and the consumer uses `actions/download-artifact@v7` (membrowse-zephyr.yml:51), so this is the PR's only cross-major upload/download pairing that also crosses a workflow boundary. The PR body asserts v4-format artifacts remain compatible across the upload-v6 / download-v7 backends, and the download already passes `run-id` + `github-token` (required for cross-run downloads), so this should work. Because a backend incompatibility here would silently degrade rather than hard-fail (the step has `continue-on-error: true` and the downstream `Verify artifact present` step only emits a `::warning::` when the ELF/linker files are missing), it is worth an explicit confirmation that a v7 download can read a v6-produced artifact. The same cross-major pairing also occurs entirely within single workflow runs in bind/curl/hostap-vm/openssl-ech, where a mismatch would surface more loudly.

**Fix:** Confirm (ideally with one CI run on this branch) that `download-artifact@v7` successfully retrieves a `membrowse-zephyr-*` artifact uploaded by `upload-artifact@v6` across the workflow_run boundary, since the `continue-on-error: true` + warning-only verification path would otherwise mask a regression as a skipped Membrowse report.

dgarske (commented):
## Skoll Code Review

### Posted findings
- [Low] **Verify the new SHA-pinned actions/checkout commit actually resolves to a v5 release tag** — `.github/workflows/wolfboot-integration.yml:29,130,205,331,439`
- [Info] **Cross-workflow artifact pairing now spans upload-artifact v6 and download-artifact v7** — `.github/workflows/membrowse-zephyr.yml:51`

JacobBarthelmeh (approved):

.github/workflows/membrowse-zephyr.yml
julek-wolfssl: Thanks — both findings verified, no changes needed:

1. **SHA pin → v5**: `93cb6efe18208431cddfb8368fd83d5badbf9bfd` is the commit
   `actions/checkout`'s `v5` tag currently resolves to (released as v5.0.1),
   matching the `# v5` comment.

2. **v6→v7 artifact compatibility**: `upload-artifact@v6` and
   `download-artifact@v7` both depend on `@actions/artifact ^5.0.0` (same
   Artifact Service v4 backend), so a v7 download reads a v6-produced
   artifact — including across the `workflow_run` boundary in
   membrowse-zephyr. The same v6/v7 pairing also runs within single
   workflows (bind/curl/hostap-vm/openssl-ech), where a mismatch would
   fail loudly rather than silently.

dgarske (approved):

PR #10693, opened 2026-06-15
Fix use-after-free possibility in GetCRLInfo

padelsbach:
# Description

Requires CRL and `WOLFSSL_SMALL_STACK`

Fixes one of many issues in 21992

# Testing

Added a unit test

# Checklist

 - [x] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/wolfssl/ssl.h
+++ b/wolfssl/ssl.h
@@ -4021,6 +4021,9 @@ typedef struct CrlInfo {
     word32 nextDateMaxLen;
     byte nextDateFormat;
     byte crlNumberSet:1;
+    byte issuerHashData[SIGNER_DIGEST_SIZE];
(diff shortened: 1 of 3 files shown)


wolfssl/ssl.h
```diff
@@ -4021,6 +4021,9 @@ typedef struct CrlInfo {
     word32 nextDateMaxLen;
     byte nextDateFormat;
     byte crlNumberSet:1;
+    byte issuerHashData[SIGNER_DIGEST_SIZE];
```
padelsbach, line 4024: Added these buffers alongside the existing pointers rather than changing the pointers to arrays so that this change does not break the API/ABI

dgarske (approved):

PR #10696, opened 2026-06-16
20260615-linuxkm-fixes

douzzer:
fixes for false positives on linuxkm CONFIG_FORTIFY_SOURCE builds on gcc-16:

`linuxkm/linuxkm_memory.c`: use packed-struct intermediates rather than `memcpy()`s for `wc_get_unaligned()` and `wc_put_unaligned()`.

`linuxkm/linuxkm_wc_port.h`: on old FIPS, retrofit nonnull attribute to `GHASH()` arg 1, so that it unconditionally writes out the hash.

`wolfcrypt/src/aes.c` and `wolfssl/wolfcrypt/aes.h`: in `GHASH()`, add `nonnull` attribute to arg 1, and remove runtime nullness check for arg 1 in the implementations.

`linuxkm/lkcapi_sha_glue.c`: refactor error code handling in `wc_linuxkm_drbg_generate()` (followup to 3c9996efe0 in #10688).

`wolfcrypt/src/aes.c`: fix performance regressions on GMAC and AES-CFB decrypt:
* add `WC_VAES_MIN_BLOCKS`, `WC_VAES_ECB_MIN_BLOCKS`, and `WC_VAES_GCM_MIN_BLOCKS`, and check against them before using AVX512/VAES implementations.
* in `AesCfbDecrypt_C()`, enlarge the `tmp[]` buffer and parameterize its size with newly added `WC_AES_CFB_DEC_BUF_BLOCKS`.

src/include.am: remove wolfcrypt/src/aes_x`86_64_`asm.S from AESNI source lists in FIPS v2/v5/v6 sections.

in all FIPS-relevant C sources, add a "#define _WC_BUILDING_foo" first (where foo is a stylization of the filename), before including `libwolfssl_sources.h`, to allow future file-specific suppressions or other settings without altering FIPS sources.

`wolfcrypt/src/aes.c`, `wolfcrypt/src/port/`, `wolfssl/wolfcrypt/aes.h`, `wolfcrypt/src/pkcs7.c`, `wolfcrypt/test/test.c`:
* implement `wc_local_AesGcmCheckTagSz()` with pedantic checks for valid authtag size.  SP 800-38D restrictions are now uniformly imposed, unless `WC_AES_GCM_ALLOW_NONSTANDARD_TAG_LENGTH` is defined (not allow with FIPS).
* refactor tag size checks in `wc_AesGcmEncrypt()`, `wc_AesGcmDecrypt()`, `wc_AesGcmEncryptFinal()`, `wc_AesGcmDecryptFinal()`, and `wc_PKCS7_DecodeAuthEnvelopedData()`.
* in `test.c`, update `aesgcm_non12iv_test()` to skip tag sizes expected to fail.

`wolfcrypt/test/test.c`:
* in `slhdsa_keygen_kat()` and `slhdsa_id_label_test()`, pass devId to wc_SlhDsaKey_Init*()
* in `cryptocb_test()`, inhibit the callback verification check for SLHDSA if FIPS (no crypto callbacks in FIPS-wrapped calls).

`linuxkm/x86_vector_register_glue.c`, `linuxkm/linuxkm_wc_port.h`, `wolfssl/wolfcrypt/memory.h`: add `WC_SVR_FLAG_FUZZ`, implement support for `DEBUG_VECTOR_REGISTER_ACCESS_FUZZING` directly in the save/restore implementations, and properly reflect existing save state there and in the `_FUZZING` variants of `SAVE_VECTOR_REGISTERS2()`.

`wolfcrypt/src/sha3.c`: refactor `WC_C_DYNAMIC_FALLBACK` using `SAVE_VECTOR_REGISTERS2()`.

clean up setup code for kernel modules:

`configure.ac`:
* remove -`DWC_SHA3_NO_ASM` from `ENABLED_LINUXKM` `AM_CFLAGS`.
* refactor initial setup for `KERNEL_MODE_DEFAULTS`, adding generic `--enable-kernel-settings` while retaining legacy `--enable-linuxkm-defaults`.
* rename $`DEF_SP_MATH` to $`DEF_SP_MATH_ALL`.
* remove redundant and unneeded setup for `KERNEL_MODE_DEFAULTS` and `ENABLED_LINUXKM` (leverage existing setup in `settings.h`).
* move some still-needed `KERNEL_MODE_DEFAULTS` and `ENABLED_LINUXKM` setup from `configure.ac` to `settings.h`.
* set up -`DWOLFSSL_KERNEL_MODE_DEFAULTS`, so that `settings.h` can pivot on it.

`wolfssl/wolfcrypt/settings.h`:
* revise `WOLFSSL_LINUXKM` section of `settings.h` to require `WOLFSSL_MIN_AUTH_TAG_SZ` at least 8 for old FIPS and 12 for new FIPS.  still force down to 4 bytes if crypto fuzzer is enabled, otherwise force down to 8 to support legacy IPsec ESP.
* in the `WOLFSSL_LINUXKM` section, don't set `WC_MLKEM_NO_ASM`, and disable `DEBUG_VECTOR_REGISTER_ACCESS_FUZZING` in ML-KEM, ML-DSA, and SLH-DSA -- intelasm works right, but fuzzing doesn't (yet).

* add hard compile-time assert in `settings.h` for FIPS v7+ asserting that `WOLFSSL_MIN_AUTH_TAG_SZ` meets SP 800-38D Rev 1 requirements.

`linuxkm/linuxkm_wc_port.h`, `linuxkm/module_hooks.c`: add linuxkm-pie support for `CheckOcspResponder()` (`WOLFSSL_NO_OCSP_ISSUER_CHECK` is no longer implied by `KERNEL_MODE_DEFAULTS`).

`linuxkm/linuxkm_wc_port.h`, `linuxkm/module_hooks.c`: on kernel >= 7.2, remove indirect symbol support for strncpy and add backward-compat implementation `wc_linuxkm_strncpy()`.

replace several nonconformant uses of `__FUNCTION__` with __func__ (`linuxkm/linuxkm_memory.c`, `wolfcrypt/src/random.c`, `wolfcrypt/test/test.c`, `wolfssl/wolfcrypt/mem_track.h`, `wolfssl/wolfcrypt/memory.h`, `wolfssl/wolfcrypt/settings.h`).

wolfcrypt/src/port/riscv/riscv-64-`aes.c`: in `GHASH()`, remove runtime nullness check for arg 1 (matching nonnull attribute to arg 1 added to prototype in earlier commit).

`wolfcrypt/src/wc_slhdsa.c`:
* disable asm accelerations if `WC_SHA3_NO_ASM` is set.
* fix an uninited-data warning in `slhdsakey_wots_pkgen_chain_c()`.

linuxkm-related loose ends:
* `wolfssl/ocsp.h`: gate out the `CheckOcspResponder()` prototype if `defined(CheckOcspResponder)` (for linuxkm-pie).
* `wolfcrypt/src/wc_mldsa.c`: add support for `WC_MLDSA_NO_ASM`.
* .wolfssl_known_macro_extras: add new macros.

tweaks for linuxkm targeting clang-built kernels:

`linuxkm/`: when logging `PTR_ERR()`, cast it to `int`, and use "%d" as the format.  Globally, `#define PTR_ERR(x) ((int)PTR_ERR(x))` in `linuxkm_wc_port.h` to fix clang warnings on kernel headers.

`linuxkm/lkcapi_aes_glue.c`: add casts in `linuxkm_test_aesgcm()` to mollify clang.

`linuxkm/linuxkm_wc_port.h`, `linuxkm/module_hooks.c`:
* add `__clang__` compat code to allow including clang `stdatomic.h` while masking out kernel-incompatible `__CLANG_STDINT_H`.
* add clang-specific suppressions for kernel headers (`-Wshorten-64-to-32`, `-Wframe-address`).

`linuxkm/lkcapi_sha_glue.c`:
* in `wc__get_random_bytes()`, add bounds-checking for `len`.
* in `wc_extract_crng_user()`, fix type conflicts.

`wolfssl/wolfcrypt/wc_port.h` and `wolfssl/wolfcrypt/types.h`:
* move the old-FIPS compatibility mapping from `INLINE` to `WC_INLINE` from `types.h` to `wc_port.h`.
* activate `stdatomic.h` for clang kernel module builds.

`linuxkm/Kbuild`:
* add clang-specific flags.
* add gcc gate around gcc-specific flags.
* allow override value for `MAX_STACK_FRAME_SIZE`.

`wolfcrypt/src/asn.c`: add casts in `GetFormattedTime_ex()` to mollify clang build of linuxkm.

tested with
```
wolfssl-multi-test.sh ...
pr-check
'.*linuxkm.*'
```


linuxkm/lkcapi_sha_glue.c
```diff
@@ -1248,20 +1244,19 @@ static int wc_linuxkm_drbg_generate(struct wc_rng_bank *ctx,
...
+                pr_warn_ratelimited("WARNING: reinitialized DRBG #%d after RNG_FAILURE_E from wc_RNG_GenerateBlock().\n", raw_smp_processor_id());
                 continue;
             }
             else {
-                pr_warn_once("ERROR: reinitialization of DRBG #%d after RNG_FAILURE_E failed with ret %d.\n", raw_smp_processor_id(), ret);
-                ret = -EINVAL;
+                pr_err_ratelimited("ERROR: reinitialization of DRBG #%d after RNG_FAILURE_E failed with ret %d.\n", raw_smp_processor_id(), ret);
                 break;
             }
         }
```
douzzer, line 1254: Good catch.  Fixed.

douzzer:
retest this please
(unrelated FIPS glitch)

philljj:
Nice, this fixes `linuxkm-aesni-insmod-kmemleak` gcc-16 false positive failure on master.

linuxkm/linuxkm_wc_port.h
```diff
@@ -779,6 +779,25 @@
...
+         * to mitigate with a grafted-on attribute.
+         */
+        #if FIPS_VERSION3_LT(6,0,0)
+            struct Aes;
+            WOLFSSL_LOCAL void __attribute__((nonnull(1))) GHASH(struct Aes *aes, const unsigned char* a,
+                                            unsigned int aSz, const unsigned char* c,
+                                            unsigned int cSz, unsigned char* s, unsigned int sSz);
+        #else
+            struct Gcm;
+            WOLFSSL_LOCAL void __attribute__((nonnull(1))) GHASH(struct Gcm *gcm, const unsigned char* a,
```
philljj, line 822: dumb question, is this redundant with the change in `aes.h`?

Also can `WC_ARG_NOT_NULL`  be used here?

linuxkm/lkcapi_sha_glue.c
```diff
@@ -1231,15 +1231,11 @@ static int wc_linuxkm_drbg_generate(struct wc_rng_bank *ctx,
             continue;
 
         if (unlikely(ret == WC_NO_ERR_TRACE(RNG_FAILURE_E))) {
-            if (slen > 0) {
-                ret = -EINVAL;
+            if (slen > 0)
```
philljj, line 1234: nit: could condense this to `if (slen > 0 || retried) break;` now.

Of if they are worth being separate checks, maybe a short 1 line comment explaining what slen > 0 vs retried means in error conditions.

philljj (changes requested):
Tests good, just some nits in review.

linuxkm/linuxkm_wc_port.h
douzzer: Not on old FIPS -- that's what it's for.

And `WC_ARG_NOT_NULL` doesn't exist yet when `linuxkm_wc_port.h` is evaluated, alas.  That's why it's also using `unsigned char *` instead of `byte *`, etc.

linuxkm/lkcapi_sha_glue.c
douzzer: Need to be separate so that instrumentation (`--enable-debug-trace-errcodes`) shows what failed.

IIRC, the `slen` situation isn't retryable because the whole point of a gen paired with a seed is to have the seed deterministically determine what's gen'd, which it wouldn't if a reinit were done midstream.  This could be bikeshedded to work right on reinit but clearly isn't worth the extra trouble.

linuxkm/linuxkm_wc_port.h
```diff
@@ -1748,6 +1811,23 @@
...
+            if (*src == 0) {
+                *dst = 0;
+                /* don't bother zero-filling dst. */
+                break;
+            }
+            *dst++ = *src++;
+        }
+        return dstart;
+    }
+    #define strncpy wc_linuxkm_strncpy
```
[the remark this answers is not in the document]
douzzer: Deliberate -- as explained in the implementation,
```
                /* don't bother zero-filling dst. */
```

wolfcrypt/src/pkcs7.c
```diff
@@ -15262,10 +15262,12 @@ int wc_PKCS7_DecodeAuthEnvelopedData(wc_PKCS7* pkcs7, byte* in,
             }
             if (ret == 0 &&
                     (encOID == AES128GCMb || encOID == AES192GCMb ||
-                     encOID == AES256GCMb) &&
-                    authTagSz < WOLFSSL_MIN_AUTH_TAG_SZ) {
-                WOLFSSL_MSG("AuthEnvelopedData GCM authTag too small");
-                ret = ASN_PARSE_E;
+                     encOID == AES256GCMb)) {
+                ret = wc_local_AesGcmCheckTagSz(authTagSz);
```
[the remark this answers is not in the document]
douzzer: Fixed

douzzer:
retest this please
(resume test failed in "CAVP self test")

douzzer:
retest this please
(Jenkins tooling glitch on PRB-python-port)

douzzer:
retest this please
(fips-ready test aborted by Jenkins)

philljj (approved):
Looks good, tests good on freebsd, debian, fedora.

PR #10704, opened 2026-06-16
Length Width Hardening

aidangarske:
# Description

50-55 - ZD21992

### DIFF
--- a/wolfcrypt/src/chacha20_poly1305.c
+++ b/wolfcrypt/src/chacha20_poly1305.c
@@ -411,6 +411,14 @@ static WC_INLINE int wc_XChaCha20Poly1305_crypt_oneshot(
         goto out;
     }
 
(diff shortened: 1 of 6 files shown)


Frauschi (dismissed):

wolfcrypt/src/chacha20_poly1305.c
```diff
@@ -411,6 +411,11 @@ static WC_INLINE int wc_XChaCha20Poly1305_crypt_oneshot(
         goto out;
     }
 
+    if ((word32)ad_len != ad_len) {
```
dgarske: Is this checking for negative or size_t vs word32 overflow? This syntax is hard to understand. Both sides of the expression are "ad_len"

dgarske (changes requested):

aidangarske:
Jenkins retest this please

dgarske (approved):

PR #10709, opened 2026-06-16
20260616-aes-fixes

douzzer:
`wolfcrypt/src/aes.c`: catch and error on total length overflow in `wc_AesGcmEncryptUpdate()`, `wc_AesGcmDecryptUpdate()`, `wc_AesCcmEncrypt()`, and `wc_AesCcmEncrypt()`.

tested with
```
wolfssl-multi-test.sh ...
super-quick-check
```

Incidentally resolves Fenrir finding F-6151 (the added `r == 0` guard in AesEncrypt/AesDecrypt makes an unkeyed AES call fail closed).

### DIFF
--- a/wolfcrypt/src/aes.c
+++ b/wolfcrypt/src/aes.c
@@ -12808,6 +12808,18 @@ int wc_AesGcmEncryptUpdate(Aes* aes, byte* out, const byte* in, word32 sz,
         ret = MISSING_IV;
     }
 
+    /* Prevent overflow of aes->cSz and ->aSz.  Per NIST SP 800-38D section
+     * 5.2.1.1, the maximum allowed ciphertext limit is 2^32 - 2 blocks, but we
+     * currently pass around the cunulative sizes in bytes as word32s, so we
+     * can't currently support the maximum allowed.
+     */
+    if ((ret == 0) &&
+        (((aes->cSz > 0xffffffff - sz)) ||
+         ((aes->aSz > 0xffffffff - authInSz))))
+    {
+        ret = AES_GCM_OVERFLOW_E;
+    }
+
     if ((ret == 0) && aes->ctrSet && (aes->aSz == 0) && (aes->cSz == 0)) {
         aes->invokeCtr[0]++;
         if (aes->invokeCtr[0] == 0) {
@@ -12948,6 +12960,18 @@ int wc_AesGcmDecryptUpdate(Aes* aes, byte* out, const byte* in, word32 sz,
     /* Check IV has been set. */
     if ((ret == 0) && (!aes->nonceSet)) {
         ret = MISSING_IV;
+    }
+
+    /* Prevent overflow of aes->cSz and ->aSz.  Per NIST SP 800-38D section
+     * 5.2.1.1, the maximum allowed ciphertext limit is 2^32 - 2 blocks, but we
+     * currently pass around the cunulative sizes in bytes as word32s, so we
+     * can't currently support the maximum allowed.
+     */
+    if ((ret == 0) &&
+        (((aes->cSz > 0xffffffff - sz)) ||
+         ((aes->aSz > 0xffffffff - authInSz))))
+    {
+        ret = AES_GCM_OVERFLOW_E;
     }
 
     if (ret == 0) {
@@ -13586,6 +13610,17 @@ int wc_AesCcmEncrypt(Aes* aes, byte* out, const byte* in, word32 inSz,
         return BAD_FUNC_ARG;
     }
 
+    lenSz = (byte)(WC_AES_BLOCK_SIZE - 1U - nonceSz);
+
+    /* With a large nonce, B[] runs out of room to represent inSz, and beyond
+     * that, the counter itself can wrap.
+     */
+    if ((lenSz < sizeof(inSz)) &&
+        (inSz >= ((word32)1 << (lenSz * 8))))
+    {
+        return AES_CCM_OVERFLOW_E;
+    }
+
 #ifdef WOLF_CRYPTO_CB
     #ifndef WOLF_CRYPTO_CB_FIND
     if (aes->devId != INVALID_DEVID)
@@ -13602,7 +13637,7 @@ int wc_AesCcmEncrypt(Aes* aes, byte* out, const byte* in, word32 inSz,
 
     XMEMSET(A, 0, sizeof(A));
     XMEMCPY(B+1, nonce, nonceSz);
-    lenSz = (byte)(WC_AES_BLOCK_SIZE - 1U - nonceSz);
+
     B[0] = (byte)((authInSz > 0 ? 64 : 0)
                   + (8 * (((byte)authTagSz - 2) / 2))
                   + (lenSz - 1));
@@ -13739,6 +13774,17 @@ int  wc_AesCcmDecrypt(Aes* aes, byte* out, const byte* in, word32 inSz,
         return BAD_FUNC_ARG;
     }
 
+    lenSz = (byte)(WC_AES_BLOCK_SIZE - 1U - nonceSz);
+
+    /* With a large nonce, B[] runs out of room to represent inSz, and beyond
+     * that, the counter itself can wrap.
+     */
+    if ((lenSz < sizeof(inSz)) &&
+        (inSz >= ((word32)1 << (lenSz * 8))))
+    {
+        return AES_CCM_OVERFLOW_E;
+    }
+
 #ifdef WOLF_CRYPTO_CB
     #ifndef WOLF_CRYPTO_CB_FIND
     if (aes->devId != INVALID_DEVID)
@@ -13757,7 +13803,6 @@ int  wc_AesCcmDecrypt(Aes* aes, byte* out, const byte* in, word32 inSz,
     oSz = inSz;
     XMEMSET(A, 0, sizeof A);
     XMEMCPY(B+1, nonce, nonceSz);
-    lenSz = (byte)(WC_AES_BLOCK_SIZE - 1U - nonceSz);
 
     B[0] = (byte)(lenSz - 1U);
     for (i = 0; i < lenSz; i++)
--- a/wolfcrypt/src/aes.c
+++ b/wolfcrypt/src/aes.c
@@ -12814,8 +12814,8 @@ int wc_AesGcmEncryptUpdate(Aes* aes, byte* out, const byte* in, word32 sz,
      * can't currently support the maximum allowed.
      */
     if ((ret == 0) &&
-        ((aes->cSz > 0xffffffff - sz) ||
-         (aes->aSz > 0xffffffff - authInSz)))
+        ((aes->cSz > WOLFSSL_MAX_32BIT - sz) ||
+         (aes->aSz > WOLFSSL_MAX_32BIT - authInSz)))
     {
         ret = AES_GCM_OVERFLOW_E;
     }
@@ -12968,8 +12968,8 @@ int wc_AesGcmDecryptUpdate(Aes* aes, byte* out, const byte* in, word32 sz,
      * can't currently support the maximum allowed.
      */
     if ((ret == 0) &&
-        ((aes->cSz > 0xffffffff - sz) ||
-         (aes->aSz > 0xffffffff - authInSz)))
+        ((aes->cSz > WOLFSSL_MAX_32BIT - sz) ||
+         (aes->aSz > WOLFSSL_MAX_32BIT - authInSz)))
     {
         ret = AES_GCM_OVERFLOW_E;
     }
--- a/wolfcrypt/src/aes.c
+++ b/wolfcrypt/src/aes.c
@@ -12810,12 +12810,12 @@ int wc_AesGcmEncryptUpdate(Aes* aes, byte* out, const byte* in, word32 sz,
 
     /* Prevent overflow of aes->cSz and ->aSz.  Per NIST SP 800-38D section
      * 5.2.1.1, the maximum allowed ciphertext limit is 2^32 - 2 blocks, but we
-     * currently pass around the cunulative sizes in bytes as word32s, so we
+     * currently pass around the cumulative sizes in bytes as word32s, so we
      * can't currently support the maximum allowed.
      */
     if ((ret == 0) &&
-        (((aes->cSz > 0xffffffff - sz)) ||
-         ((aes->aSz > 0xffffffff - authInSz))))
+        ((aes->cSz > 0xffffffff - sz) ||
+         (aes->aSz > 0xffffffff - authInSz)))
     {
         ret = AES_GCM_OVERFLOW_E;
     }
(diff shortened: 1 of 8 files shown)


wolfcrypt/src/aes.c
```diff
@@ -13586,6 +13610,17 @@ int wc_AesCcmEncrypt(Aes* aes, byte* out, const byte* in, word32 inSz,
         return BAD_FUNC_ARG;
     }
 
+    lenSz = (byte)(WC_AES_BLOCK_SIZE - 1U - nonceSz);
+
+    /* With a large nonce, B[] runs out of room to represent inSz, and beyond
+     * that, the counter itself can wrap.
+     */
+    if ((lenSz < sizeof(inSz)) &&
+        (inSz >= (1U << (lenSz * 8))))
```
douzzer: fixed

wolfcrypt/src/aes.c
```diff
@@ -12808,6 +12808,18 @@ int wc_AesGcmEncryptUpdate(Aes* aes, byte* out, const byte* in, word32 sz,
         ret = MISSING_IV;
     }
 
+    /* Prevent overflow of aes->cSz and ->aSz.  Per NIST SP 800-38D section
+     * 5.2.1.1, the maximum allowed ciphertext limit is 2^32 - 2 blocks, but we
+     * currently pass around the cunulative sizes in bytes as word32s, so we
```
Frauschi: typo: cu**n**ulative -> cu**m**ulative

wolfcrypt/src/aes.c
```diff
@@ -12808,6 +12808,18 @@ int wc_AesGcmEncryptUpdate(Aes* aes, byte* out, const byte* in, word32 sz,
...
+     * can't currently support the maximum allowed.
+     */
+    if ((ret == 0) &&
+        (((aes->cSz > 0xffffffff - sz)) ||
```
Frauschi: Redundant double parentheses, also below

Frauschi (dismissed):
Should we add test coverage for the `OVERFLOW_E` cases?

wolfcrypt/src/aes.c
```diff
@@ -12808,6 +12808,18 @@ int wc_AesGcmEncryptUpdate(Aes* aes, byte* out, const byte* in, word32 sz,
         ret = MISSING_IV;
     }
 
+    /* Prevent overflow of aes->cSz and ->aSz.  Per NIST SP 800-38D section
+     * 5.2.1.1, the maximum allowed ciphertext limit is 2^32 - 2 blocks, but we
```
dgarske, line 12812: **🔵 [Low] Comment typo: 'cunulative' should be 'cumulative'**

The new comment block in both GCM Update functions contains the misspelling 'cunulative' (should be 'cumulative').

**Fix:** Fix the typo in both copies of the comment.

wolfcrypt/src/aes.c
```diff
@@ -12808,6 +12808,18 @@ int wc_AesGcmEncryptUpdate(Aes* aes, byte* out, const byte* in, word32 sz,
...
+     * currently pass around the cunulative sizes in bytes as word32s, so we
+     * can't currently support the maximum allowed.
+     */
+    if ((ret == 0) &&
```
dgarske, line 12816: **🔵 [Low] Brace-on-own-line and redundant double parentheses deviate from file style**

The surrounding file consistently uses K&R brace style (`if (...) {` on the same line), but the new GCM checks place the opening brace on its own line. Additionally the condition has redundant inner parentheses: `(((aes->cSz > 0xffffffff - sz)))`. The CCM checks have the same own-line brace style. This is purely cosmetic and matches neither the immediate neighbors (e.g. the `invokeCtr` block just below uses same-line braces) nor the rest of aes.c.

**Fix:** Use same-line braces to match the file and drop the redundant extra parentheses.

wolfcrypt/src/aes.c
dgarske: **🔵 [Low] Bare 0xffffffff literal; prefer suffixed/named constant**

The overflow bound is written as the bare literal `0xffffffff`. It is functionally correct (an unsigned hex literal that fits word32 on all supported platforms, including 16-bit-int targets where it promotes to unsigned long), but the wolfSSL convention favors an explicit unsigned suffix (`0xFFFFFFFFU`) or a named maximum to make the word32-max intent obvious and avoid any signedness ambiguity in the subtraction.

**Fix:** Add a `U` suffix (or use a named WORD32_MAX-style constant) for clarity and consistency.

dgarske (commented):
## Skoll Code Review

### Posted findings
- [Low] **Comment typo: 'cunulative' should be 'cumulative'** — `wolfcrypt/src/aes.c:12813,12967`
- [Low] **Brace-on-own-line and redundant double parentheses deviate from file style** — `wolfcrypt/src/aes.c:12816-12821,12970-12975`
- [Low] **Bare 0xffffffff literal; prefer suffixed/named constant** — `wolfcrypt/src/aes.c:12817-12818,12971-12972`
- [Low] **Reused AES_CCM_OVERFLOW_E error string is about 'invocation counter', not input length** — `wolfcrypt/src/aes.c:13621,13785`

### Skipped findings
- [Medium] ~~No tests added for new AES-GCM/CCM overflow error paths~~

wolfcrypt/src/aes.c
```diff
@@ -13586,6 +13610,17 @@ int wc_AesCcmEncrypt(Aes* aes, byte* out, const byte* in, word32 inSz,
         return BAD_FUNC_ARG;
     }
 
+    lenSz = (byte)(WC_AES_BLOCK_SIZE - 1U - nonceSz);
+
+    /* With a large nonce, B[] runs out of room to represent inSz, and beyond
+     * that, the counter itself can wrap.
+     */
+    if ((lenSz < sizeof(inSz)) &&
```
dgarske, line 13642: **🔵 [Low] Reused AES_CCM_OVERFLOW_E error string is about 'invocation counter', not input length**

The new CCM checks return `AES_CCM_OVERFLOW_E`, whose human-readable string (error.c:503-504) and header comment (error-crypt.h:255) describe it as 'AES-CCM invocation counter overflow'. The new condition is actually an input-length-vs-nonce-size limitation (the message length doesn't fit the CCM L field), not an invocation-counter overflow. Reusing the closest existing code is reasonable and avoids adding a new error number, but the diagnostic string may mislead someone debugging the failure. (The GCM reuse of AES_GCM_OVERFLOW_E for cumulative-size overflow is a closer semantic fit.)

**Fix:** Consider broadening the error string (e.g. 'AES-CCM length/counter overflow') so it covers the input-size case, or document the dual meaning. Not blocking.

wolfcrypt/src/aes.c
douzzer: fixed

wolfcrypt/src/aes.c
douzzer: fixed

wolfcrypt/src/aes.c
douzzer: fixed

wolfcrypt/src/aes.c
douzzer: fixed

wolfcrypt/src/aes.c
douzzer: fixed.  used `WOLFSSL_MAX_32BIT`, a macro I wasn't previously familiar with.

wolfcrypt/src/aes.c
douzzer: The error string does need updating.  Have made a note and will fix in a follow-up PR, or if this one needs another spin for a substantive reason, will fix it here.

dgarske (commented):
## Skoll Multi-Scan Review

**Modes:** `review`**Overall recommendation:** `COMMENT`

### Posted findings
- [Low] [review] **GCM aSz (AAD) overflow branch is not exercised by tests** — `wolfcrypt/test/test.c:19152-19161,19179-19188`
- [Low] [review] **Raw sizeof in mixed-signedness comparison deviates from function's wordSz convention** — `wolfcrypt/src/aes.c:13618,13782`
- [Info] [review] **Test code pointer style and line length deviate from wolfSSL conventions** — `wolfcrypt/test/test.c:21128-21130`

wolfcrypt/test/test.c
```diff
@@ -19148,6 +19148,18 @@ static wc_test_ret_t aesgcm_stream_test(Aes* enc)
     ret = wc_AesGcmEncryptUpdate(enc, resultC, p, sizeof(p), a, sizeof(a));
     if (ret != 0)
         ERROR_OUT(WC_TEST_RET_ENC_EC(ret), out);
+
+#if !defined(HAVE_FIPS) || FIPS_VERSION3_GE(7, 0, 0)
+    ret = wc_AesGcmEncryptUpdate(enc, resultC, p, 0xffffffff,
```
dgarske: **🔵 [Low] GCM aSz (AAD) overflow branch is not exercised by tests** · _test_

The new overflow guard in `wc_AesGcmEncryptUpdate`/`wc_AesGcmDecryptUpdate` has two independent branches: `aes->cSz > 0xffffffff - sz` (ciphertext) and `aes->aSz > 0xffffffff - authInSz` (AAD). The new tests only drive the ciphertext branch by passing `sz = 0xffffffff` with `authInSz = 0`. The AAD-accumulation branch (`aes->aSz` overflow) is never tested, so a regression in that condition would go undetected.

**Fix:** Add a test that triggers the aSz overflow branch so both conditions of the new guard are covered. Add a symmetric case that passes a large authInSz (e.g. authIn buffer with authInSz = 0xffffffff and sz = 0) after a prior update that left aes->aSz > 0, asserting AES_GCM_OVERFLOW_E.

wolfcrypt/src/aes.c
dgarske, line 13642: **🔵 [Low] Raw sizeof in mixed-signedness comparison deviates from function's wordSz convention** · _convention_

The new check `(lenSz < sizeof(inSz))` compares `lenSz` (a `byte`, integer-promoted to signed `int`) against `sizeof(inSz)` (`size_t`, unsigned). Both functions already define `const word32 wordSz = (word32)sizeof(word32);` precisely to avoid comparing against a raw `sizeof` (used in the loop a few lines below). Using the raw `sizeof` here is inconsistent with that local convention; depending on warning flags it can also surface as a signed/unsigned comparison note. The logic itself is correct (lenSz is small and non-negative).

**Fix:** Use the existing `wordSz` (with an explicit cast on `lenSz`) to keep the comparison consistent with the rest of the function and avoid mixed-signedness: `if (((word32)lenSz < wordSz) && (inSz >= ((word32)1 << (lenSz * 8)))) { return AES_CCM_OVERFLOW_E; }`

wolfcrypt/test/test.c
```diff
@@ -21095,6 +21119,100 @@ static wc_test_ret_t aesccm_128_badarg_test(Aes* enc)
         ERROR_OUT(WC_TEST_RET_ENC_EC(ret), out);
 #endif
 
+#if !defined(BENCH_EMBEDDED) && !defined(WOLFSSL_NO_MALLOC)
+    /* Test 64k - 1 with full nonce -- must work. */
+    {
+        wc_static_assert(sizeof(iv) >= 13);
+        word32 clear_size = ((word32)1 << 16);
+        word32 cipher_size = ((word32)1 << 16);
+        byte *large_alloc = (byte *)XMALLOC(clear_size + cipher_size, HEAP_HINT, DYNAMIC_TYPE_TMP_BUFFER);
```
dgarske: **⚪ [Info] Test code pointer style and line length deviate from wolfSSL conventions** · _style_

The new allocation in the CCM large-buffer test uses `byte *large_alloc` / `byte *clear` / `byte *cipher` and `(byte *)` casts rather than the wolfSSL `byte* x` placement, and line 21128 (`byte *large_alloc = (byte *)XMALLOC(...)`) exceeds the project's ~80-column width. The scoped `{ ... }` block is justified (it bounds the malloc lifetime and frees on every path before `goto out`), so only the pointer style and line width are at issue. New-code-only nit; pre-existing mixed styles in test.c are out of scope.

**Fix:** Adopt `byte*` pointer placement and wrap the long XMALLOC line to stay within the column limit for new code.

dgarske (changes requested):
https://github.com/wolfSSL/wolfssl/pull/10709#pullrequestreview-4518340575

wolfcrypt/src/aes.c
```diff
@@ -12808,6 +12808,18 @@ int wc_AesGcmEncryptUpdate(Aes* aes, byte* out, const byte* in, word32 sz,
...
     }
 
+    /* Prevent overflow of aes->cSz and ->aSz.  Per NIST SP 800-38D section
+     * 5.2.1.1, the maximum allowed ciphertext limit is 2^32 - 2 blocks, but we
+     * currently pass around the cumulative sizes in bytes as word32s, so we
+     * can't currently support the maximum allowed.
+     */
+    if ((ret == 0) &&
+        ((aes->cSz > 0xffffffff - sz) ||
+         (aes->aSz > 0xffffffff - authInSz)))
```
JacobBarthelmeh: It this is for WOLFSSL_MAX_32BIT values I think we should use the macro available.

wolfcrypt/test/test.c
douzzer: fixed

wolfcrypt/src/aes.c
douzzer: Not something to fix.  The same new code pattern is used at 10 different code points, with widely varying local context.  `wordSz` is only available at 2 of them.  It is more useful to keep the checks identical, than to fork 2 of the 10 to leverage the (pointless) `wordSz` stack allocation.

wolfcrypt/test/test.c
douzzer: Fixed the line length.

The incumbent code in `aesccm_128_badarg_test()` uses the form I'm using here (e.g. casts by `(const byte *)`).

wolfcrypt/src/aes.c
douzzer: Definitely the right idea!  Fixing.  This could protect us from build failures on weird targets in the future.

wolfcrypt/src/aes.c
douzzer: fixed.

dgarske (approved):

douzzer:
Note latest push passed `super-quick-check` here.

JacobBarthelmeh:
Retest this please Jenkins. Generic config failed. `--enable-linuxkm-defaults --enable-all --enable-crypttests --enable-examples --disable-asm --enable-stacksize=verbose CFLAGS="-DWOLFSSL_TEST_MAX_RELATIVE_STACK_BYTES=8192" ` failed ocsp-stappling during `make check`

douzzer:
retest this please
```FAIL: scripts/ocsp-stapling2.test```

JacobBarthelmeh (approved):

PR #10710, opened 2026-06-16
OIDs represented as word32 instead of word16

aidankeefe2022:
# Description
This fixes two issues in how OIDs are represented in wolfSSL.
Breaking change.  Word16 cannot represent many real OID arcs -- e.g. 1.2.840.113549.1.1.1 (RSA encryption). This was widened to word 32 in three public symbols:

- Two callback types: wc_UnknownExtCallback and wc_UnknownExtCallbackEx
- One function: wc_EncodeObjectId

The callback types appear as fields in WOLFSSL_CERT_MANAGER, wc_PKCS7, and DecodedCert.

New internal functions were added to handle this change EncodeObjectId_ex and DecodeObjectId_ex both handle
word32 OIDs. The non _ex functions were left due to them being called by FIPS code and that code expects word16.

Non-breaking bug fix. DER encoding packs the first two arcs into a single value (40 x first + second), so decoding assumed the second arc was 0-39. That assumption holds only when the first arc is 0 or 1; when the first arc is 2, the second arc is unbounded and base-128 encodes to a value above 127. Decoding now special-cases this.

Example -- 2.100.3:
encodes -> 0x81, 0x34, 0x03
before:  decoded as 4.20.3   (0x81,0x34 unpacked to 180; 180/40=4 rem 20)
after:   decodes as 2.100.3

# Testing
Added tests for DecodeObjectId and EncodeObjectId covering word32 OIDs, plus a test for an OID whose first arc is 2 and second arc is 100, verifying correct decoding of the byte form.

# Checklist

- [x] added tests
- [x] updated/added doxygen
- [x] updated manual and documentation

### DIFF
--- a/tests/api/test_asn.c
+++ b/tests/api/test_asn.c
@@ -2690,21 +2690,25 @@ int test_wc_DecodeObjectId_ex(void)
     (defined(HAVE_OID_DECODING) || defined(WOLFSSL_ASN_PRINT))
     {
         word32 i;
+
+        /* Tests multi byte encoding for arc 1 and 2
+         * (only possible when arc 1 is 2 and arc 2 is greater than 48) */
         static const word32 oid_dot_2[] = {
             2, 100, 4, 6
         };
 
+        /* Tests multi byte encoding for arc 1 and 2
+         * (only possible when arc 1 is 2 and arc 2 is greater than 48) */
         static const byte oid_start_with_2[] = {
             0x81, 0x34, 0x04, 0x06
         };
-
 
         /* OID 1.3.132.0.6 (secp112r1)
          * DER encoding: 2b 81 04 00 06
          * First byte 0x2b = 43 => arc0 = 43/40 = 1, arc1 = 43%40 = 3
          * Remaining arcs: 132 0 6
          */
-        static const byte oid_sha256rsa[] = {
+        static const byte oid_secp112r1[] = {
             0x2B, 0x81, 0x04, 0x00, 0x06
         };
 
@@ -2718,7 +2722,7 @@ int test_wc_DecodeObjectId_ex(void)
         word32 trueOutSz = sizeof(oid_dot_form) / sizeof(word32);
         /* Test 1: Normal decode */
         outSz = MAX_OID_SZ;
-        ExpectIntEQ(DecodeObjectId_ex(oid_sha256rsa, sizeof(oid_sha256rsa),
+        ExpectIntEQ(DecodeObjectId_ex(oid_secp112r1, sizeof(oid_secp112r1),
                                    out, &outSz), 0);
         ExpectIntEQ((int)outSz, trueOutSz);
         for (i = 0; i < ((outSz <= trueOutSz) ? outSz : trueOutSz); i++) {
@@ -2727,22 +2731,22 @@ int test_wc_DecodeObjectId_ex(void)
 
         /* Test 2: NULL args */
         outSz = MAX_OID_SZ;
-        ExpectIntEQ(DecodeObjectId_ex(NULL, sizeof(oid_sha256rsa), out, &outSz),
+        ExpectIntEQ(DecodeObjectId_ex(NULL, sizeof(oid_secp112r1), out, &outSz),
                     WC_NO_ERR_TRACE(BAD_FUNC_ARG));
-        ExpectIntEQ(DecodeObjectId_ex(oid_sha256rsa, sizeof(oid_sha256rsa),
+        ExpectIntEQ(DecodeObjectId_ex(oid_secp112r1, sizeof(oid_secp112r1),
                                    out, NULL),
                     WC_NO_ERR_TRACE(BAD_FUNC_ARG));
 
         /* Test 3 (Bug 1): outSz=1 must return BUFFER_E, not OOB write.
          * The first OID byte decodes into two arcs, so outSz must be >= 2. */
         outSz = 1;
-        ExpectIntEQ(DecodeObjectId_ex(oid_sha256rsa, sizeof(oid_sha256rsa),
+        ExpectIntEQ(DecodeObjectId_ex(oid_secp112r1, sizeof(oid_secp112r1),
                                    out, &outSz),
                     WC_NO_ERR_TRACE(BUFFER_E));
 
         /* Test 4: outSz=0 must also return BUFFER_E */
         outSz = 0;
-        ExpectIntEQ(DecodeObjectId_ex(oid_sha256rsa, sizeof(oid_sha256rsa),
+        ExpectIntEQ(DecodeObjectId_ex(oid_secp112r1, sizeof(oid_secp112r1),
                                    out, &outSz),
                     WC_NO_ERR_TRACE(BUFFER_E));
 
@@ -2759,11 +2763,11 @@ int test_wc_DecodeObjectId_ex(void)
 
         /* Test 6: Buffer too small for later arcs */
         outSz = 3; /* only room for 3 arcs, but OID has 7 */
-        ExpectIntEQ(DecodeObjectId_ex(oid_sha256rsa, sizeof(oid_sha256rsa),
+        ExpectIntEQ(DecodeObjectId_ex(oid_secp112r1, sizeof(oid_secp112r1),
                                    out, &outSz),
                     WC_NO_ERR_TRACE(BUFFER_E));
 
-    /* Test 7: first Arc is 2 */
+        /* Test 7: first Arc is 2 */
         {
             word32 trueOutSz2 = sizeof(oid_dot_2) / sizeof(word32);
             outSz = MAX_OID_SZ;
@@ -2777,8 +2781,7 @@ int test_wc_DecodeObjectId_ex(void)
             }
         }
 
-        /* Test 8: an OID with an arc that exceeds word16.
-         */
+        /* Test 8: an OID with an arc that exceeds word16. */
         {
             static const byte oid_large_arc[] = {
                 0x2a, 0x86, 0x48, 0x86, 0xf7, 0x0d, 0x01, 0x01, 0x0b
@@ -2858,8 +2861,7 @@ int test_wc_EncodeObjectId(void)
         ExpectIntEQ(wc_EncodeObjectId(oid_small, oid_small_cnt, out, &outSz),
                     WC_NO_ERR_TRACE(BUFFER_E));
 
-        /* Test 5 : arc greater that SHRT_MAX
-         */
+        /* Test 5 : arc greater that SHRT_MAX */
         {
             static const word32 oid_large[] = {
                 1U, 2U, 840U, 113549U, 1U, 1U, 11U
--- a/tests/api/test_asn.c
+++ b/tests/api/test_asn.c
@@ -2682,7 +2682,7 @@ int test_wc_DecodeObjectId_FIPS16(void)
     return EXPECT_RESULT();
 }
 
-int test_wc_DecodeObjectId_ex(void)
+int test_wc_DecodeObjectId32(void)
 {
     EXPECT_DECLS;
 
@@ -2722,7 +2722,7 @@ int test_wc_DecodeObjectId_ex(void)
         word32 trueOutSz = sizeof(oid_dot_form) / sizeof(word32);
         /* Test 1: Normal decode */
         outSz = MAX_OID_SZ;
-        ExpectIntEQ(DecodeObjectId_ex(oid_secp112r1, sizeof(oid_secp112r1),
+        ExpectIntEQ(DecodeObjectId32(oid_secp112r1, sizeof(oid_secp112r1),
                                    out, &outSz), 0);
         ExpectIntEQ((int)outSz, trueOutSz);
         for (i = 0; i < ((outSz <= trueOutSz) ? outSz : trueOutSz); i++) {
@@ -2731,22 +2731,22 @@ int test_wc_DecodeObjectId_ex(void)
 
         /* Test 2: NULL args */
         outSz = MAX_OID_SZ;
-        ExpectIntEQ(DecodeObjectId_ex(NULL, sizeof(oid_secp112r1), out, &outSz),
+        ExpectIntEQ(DecodeObjectId32(NULL, sizeof(oid_secp112r1), out, &outSz),
                     WC_NO_ERR_TRACE(BAD_FUNC_ARG));
-        ExpectIntEQ(DecodeObjectId_ex(oid_secp112r1, sizeof(oid_secp112r1),
+        ExpectIntEQ(DecodeObjectId32(oid_secp112r1, sizeof(oid_secp112r1),
                                    out, NULL),
                     WC_NO_ERR_TRACE(BAD_FUNC_ARG));
 
         /* Test 3 (Bug 1): outSz=1 must return BUFFER_E, not OOB write.
          * The first OID byte decodes into two arcs, so outSz must be >= 2. */
         outSz = 1;
-        ExpectIntEQ(DecodeObjectId_ex(oid_secp112r1, sizeof(oid_secp112r1),
+        ExpectIntEQ(DecodeObjectId32(oid_secp112r1, sizeof(oid_secp112r1),
                                    out, &outSz),
                     WC_NO_ERR_TRACE(BUFFER_E));
 
         /* Test 4: outSz=0 must also return BUFFER_E */
         outSz = 0;
-        ExpectIntEQ(DecodeObjectId_ex(oid_secp112r1, sizeof(oid_secp112r1),
+        ExpectIntEQ(DecodeObjectId32(oid_secp112r1, sizeof(oid_secp112r1),
                                    out, &outSz),
                     WC_NO_ERR_TRACE(BUFFER_E));
 
@@ -2754,7 +2754,7 @@ int test_wc_DecodeObjectId_ex(void)
         {
             static const byte oid_one_byte[] = { 0x2a }; /* 1.2 */
             outSz = 2;
-            ExpectIntEQ(DecodeObjectId_ex(oid_one_byte, sizeof(oid_one_byte),
+            ExpectIntEQ(DecodeObjectId32(oid_one_byte, sizeof(oid_one_byte),
                                        out, &outSz), 0);
             ExpectIntEQ((int)outSz, 2);
             ExpectIntEQ(out[0], 1);
@@ -2763,7 +2763,7 @@ int test_wc_DecodeObjectId_ex(void)
 
         /* Test 6: Buffer too small for later arcs */
         outSz = 3; /* only room for 3 arcs, but OID has 7 */
-        ExpectIntEQ(DecodeObjectId_ex(oid_secp112r1, sizeof(oid_secp112r1),
+        ExpectIntEQ(DecodeObjectId32(oid_secp112r1, sizeof(oid_secp112r1),
                                    out, &outSz),
                     WC_NO_ERR_TRACE(BUFFER_E));
 
@@ -2771,7 +2771,7 @@ int test_wc_DecodeObjectId_ex(void)
         {
             word32 trueOutSz2 = sizeof(oid_dot_2) / sizeof(word32);
             outSz = MAX_OID_SZ;
-            ExpectIntEQ(DecodeObjectId_ex(oid_start_with_2,
+            ExpectIntEQ(DecodeObjectId32(oid_start_with_2,
                         sizeof(oid_start_with_2),
                         out, &outSz), 0);
             ExpectIntEQ((int)outSz, trueOutSz2);
@@ -2792,7 +2792,7 @@ int test_wc_DecodeObjectId_ex(void)
             word32 trueOutSz3 = sizeof(oid_dot_large_arc) / sizeof(word32);
 
             outSz = MAX_OID_SZ;
-            ExpectIntEQ(DecodeObjectId_ex(oid_large_arc, sizeof(oid_large_arc),
+            ExpectIntEQ(DecodeObjectId32(oid_large_arc, sizeof(oid_large_arc),
                                        out, &outSz), 0);
             ExpectIntEQ((int)outSz, (int)trueOutSz3);
             for (i = 0; i < ((outSz <= trueOutSz3) ? outSz : trueOutSz3); i++) {
@@ -2809,7 +2809,7 @@ int test_wc_DecodeObjectId_ex(void)
                 0x2a, 0xFF, 0xFF, 0xFF, 0xFF, 0x7F
             };
             outSz = MAX_OID_SZ;
-            ExpectIntEQ(DecodeObjectId_ex(oid_overflow_arc,
+            ExpectIntEQ(DecodeObjectId32(oid_overflow_arc,
                         sizeof(oid_overflow_arc), out, &outSz),
                         WC_NO_ERR_TRACE(ASN_OBJECT_ID_E));
         }
@@ -2824,13 +2824,13 @@ int test_wc_EncodeObjectId(void)
     EXPECT_DECLS;
 #if defined(HAVE_OID_ENCODING) && !defined(NO_ASN)
     {
-        /* 1.3.132.0.6 (secp112r1) -- every arc fits in word16, so this
-         * encodes identically in both build configs. */
-        static const word32 oid_small[] = { 1U, 3U, 132U, 0U, 6U };
+        /* wc_EncodeObjectId() takes word16 arcs, so only OIDs whose arcs all
+         * fit in a word16 can be encoded with it. 1.3.132.0.6 (secp112r1). */
+        static const word16 oid_small[] = { 1U, 3U, 132U, 0U, 6U };
         static const byte oid_small_der[] = {
             0x2b, 0x81, 0x04, 0x00, 0x06
         };
-        const word32 oid_small_cnt = sizeof(oid_small) / sizeof(word32);
+        const word32 oid_small_cnt = sizeof(oid_small) / sizeof(word16);
         byte   out[MAX_OID_SZ];
         word32 outSz;
         word32 i;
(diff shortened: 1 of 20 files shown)


JacobBarthelmeh:
For some reason I had just envisioned an API argument change. It makes sense though for some of the internals of the function to change too, this looks good but it is more risk than I had envisioned.

aidankeefe2022:
Jenkins retest this please

wolfcrypt/src/asn.c
```diff
@@ -7299,6 +7400,71 @@ int DecodeObjectId(const byte* in, word32 inSz, word16* out, word32* outSz)
...
+    word32 t = 0;
+    int cnt = 0;
+
+    /* check args */
+    if (in == NULL || outSz == NULL) {
+        return BAD_FUNC_ARG;
+    }
+
+    /* decode bytes */
+    while (inSz--) {
```
Frauschi, line 7764: **🔵 [Low] DecodeObjectId_ex still caps arcs at 28 bits, so word32 arcs >= 2^28 are rejected**
*💡 SUGGEST*  `question`

The stated goal of the PR is to represent OID arcs that don't fit in a word16 without truncation. `DecodeObjectId_ex` outputs word32, but it inherits the `if (cnt == 4) return ASN_OBJECT_ID_E;` guard from the word16 function, which limits any single arc to 4 base-128 bytes = 28 bits (max 268,435,455). Arcs that need the full word32 range (e.g. the large sub-arcs under 2.25.<uuid>) still fail to decode. So the widening is partial: it fixes the 2^16..2^28 range (e.g. 113549) but not the full word32 range implied by the type. This is a carry-over of pre-existing behavior, not a regression, but it undercuts the 'word32 without truncation' framing.

**Suggestion:**
```suggestion
/* allow up to 5 bytes (35 bits worth) and cap the accumulator at word32 */
        if (cnt == 5)
            return ASN_OBJECT_ID_E;
```

wolfcrypt/src/asn.c
```diff
@@ -7164,12 +7164,106 @@ static int CheckCurve(word32 oid)
...
+}
+
+int EncodeObjectId_ex(const word32* in, word32 inSz,
+        byte* out, word32* outSz)
+{
+    int i, x, len;
+    word32 d, t;
+
+    /* check args */
+    if (in == NULL || outSz == NULL || inSz <= 0) {
```
Frauschi: **🔵 [Low] EncodeObjectId_ex reads in[1] when inSz == 1 (out-of-bounds read)**
*🔧 NIT*  `bug`

The arg check is `inSz <= 0` (i.e. only rejects inSz == 0, since inSz is unsigned), then unconditionally computes `d = in[0]*40 + in[1]`. With inSz == 1 this reads in[1] out of bounds before the encode loop (which then does nothing). This mirrors the pre-existing behavior in the word16 `EncodeObjectId`, so it is a faithful copy rather than a new defect, but the new function now backs the public `wc_EncodeObjectId`. A valid OID always has at least 2 arcs, so the guard should reflect that.

**Suggestion:**
```suggestion
if (in == NULL || outSz == NULL || inSz < 2) {
        return BAD_FUNC_ARG;
    }
```

Frauschi (commented):
## 🐺 Skoll Code Review

**Findings:** 6 total — 5 posted, 4 skipped

### Posted findings
- [Low] **DecodeObjectId_ex still caps arcs at 28 bits, so word32 arcs >= 2^28 are rejected** — `wolfcrypt/src/asn.c:7431-7433`
- [Low] **Japanese doxygen for the unknown-ext callback still shows word16** — `doc/dox_comments/header_files-ja/asn_public.h:1937,1951`
- [Low] **EncodeObjectId_ex reads in[1] when inSz == 1 (out-of-bounds read)** — `wolfcrypt/src/asn.c:7180-7185`
- [Info] **printf uses %d for word32 OID arc values** — `wolfcrypt/src/asn.c:7555-7557,37819`
- [Info] **Leftover #undef LARGE_ARC_EXPECTED for a macro that is never defined** — `tests/api/test_asn.c:1748,1826`

- [Medium] Public API wc_EncodeObjectId signature changed in place instead of adding an _ex variant
- [Low] Japanese doxygen for the unknown-ext callback still shows word16
- [Info] printf uses %d for word32 OID arc values
- [Info] Leftover #undef LARGE_ARC_EXPECTED for a macro that is never defined

</details>

wolfcrypt/src/asn.c
aidankeefe2022: fixed

wolfcrypt/src/asn.c
aidankeefe2022: fixed

Frauschi:
Jenkins retest this please

wolfssl/version.h
```diff
@@ -28,8 +28,8 @@
 extern "C" {
 #endif
 
-#define LIBWOLFSSL_VERSION_STRING "5.9.2"
-#define LIBWOLFSSL_VERSION_HEX 0x05009002
+#define LIBWOLFSSL_VERSION_STRING "5.9.1"
```
Frauschi: **🟠 [High] Version downgraded 5.9.2 -> 5.9.1 (regression)**
*🚫 BLOCK*  `bug`

This PR changes `LIBWOLFSSL_VERSION_STRING` from `"5.9.2"` to `"5.9.1"` and `LIBWOLFSSL_VERSION_HEX` from `0x05009002` to `0x05009001`. `git show master:wolfssl/version.h` confirms the base branch is at 5.9.2 (commit `0fe21bbcd prepare for release 5.9.2`), so this is a downgrade introduced by the PR — almost certainly an accidental revert from a bad rebase/merge. Shipping this would make the library report an older version than the release it is based on, break version-gated feature checks in downstream code, and confuse packaging.

**Suggestion:**
```suggestion
#define LIBWOLFSSL_VERSION_STRING "5.9.2"
#define LIBWOLFSSL_VERSION_HEX 0x05009002
```

tests/api/test_asn.c
```diff
@@ -1704,63 +1704,196 @@ int test_SerialNumber0_RootCA(void)
...
+            }
+        }
+
+        /* Test 8: an OID with an arc that exceeds word16. Tests that wrong
+         * but unchangeable behavior is working correctly,
+         *
+         * word16 version is used in FIPS build
+         */
+        {
+            static const byte oid_large_arc[] = {
```
Frauschi, line 2665: **🟡 [Medium] FIPS16 truncation test does not exercise the truncation path**
*💡 SUGGEST*  `test`

Test 8 claims to test "an OID with an arc that exceeds word16" and "wrong but unchangeable" truncation behavior, but the DER input `{0x2a,0x86,0x48,0x82,0xf7,0x0d,...}` decodes arc 4 to the *word32* value 48013 (0x82,0xf7,0x0d = 48013), which already fits in a word16. The word16 `DecodeObjectId` therefore never truncates anything here — the internal accumulator `t` stays at 48013. The comparison `out[3] == (word16)113549` passes only because `(word16)113549` also equals 48013 by coincidence, not because truncation occurred. To actually exercise the truncation path the input bytes should encode 113549 (`0x86,0xf7,0x0d`), which decodes internally to 113549 and is then truncated to 48013 on assignment to the word16 output.

**Suggestion:**
```suggestion
static const byte oid_large_arc[] = {
    0x2a, 0x86, 0x48, 0x86, 0xf7, 0x0d, 0x01, 0x01, 0x0b
};
```

Frauschi (changes requested):
## 🐺 Skoll Code Review

**Findings:** 5 total — 2 posted, 3 skipped

### Posted findings
- [High] **Version downgraded 5.9.2 -> 5.9.1 (regression)** — `wolfssl/version.h:31-32`
- [Medium] **FIPS16 truncation test does not exercise the truncation path** — `tests/api/test_asn.c:1806-1823`

- [Medium] wc_EncodeObjectId public signature change is a breaking ABI/API change
- [Low] Large duplicate block between EncodeObjectId_ex and EncodeObjectId
- [Info] Misaligned indentation of zero-value else branch

</details>

tests/api/test_asn.c
aidankeefe2022: fixed

wolfssl/version.h
aidankeefe2022: fixed

Frauschi:
Jenkins retest this please

Frauschi (dismissed):
LGTM from my side. But @JacobBarthelmeh should confirm the API breakage.

Frauschi:
And a PR for wolfCLU is also required to fix the build failures and the CI tests. These should be merged simultaneously then.

aidankeefe2022:
Once https://github.com/wolfSSL/wolfCLU/pull/253 is merged this can go in

aidankeefe2022:
retest this please

philljj:
Retest this please.

windows test:
```
  connecting to 127.0.0.1:53652
  [0] Incorrect return -308

  Failure Reason Given is -1
```

tests/api/test_asn.c
```diff
@@ -2007,63 +2007,196 @@ int test_SerialNumber0_RootCA(void)
...
+            0x81, 0x34, 0x04, 0x06
+        };
+
+
+        /* OID 1.3.132.0.6 (secp112r1)
+         * DER encoding: 2b 81 04 00 06
+         * First byte 0x2b = 43 => arc0 = 43/40 = 1, arc1 = 43%40 = 3
+         * Remaining arcs: 132 0 6
          */
         static const byte oid_sha256rsa[] = {
```
philljj: The variable name is `oid_sha256rsa`, but the comment says `OID 1.3.132.0.6 (secp112r1)`

tests/api/test_asn.c
```diff
@@ -2007,63 +2007,196 @@ int test_SerialNumber0_RootCA(void)
...
+    {
+        word32 i;
+        static const word16 oid_dot_2[] = {
+            2, 100, 4, 6
+        };
+
+        static const byte oid_start_with_2[] = {
+            0x81, 0x34, 0x04, 0x06
+        };
+
```
philljj, line 2581: double blank line.

Also, I think these variables deserve a one line comment explaining what is being tested, and what they are.

tests/api/test_asn.c
```diff
@@ -2007,63 +2007,196 @@ int test_SerialNumber0_RootCA(void)
...
-         * Remaining arcs: 840, 113549, 1, 1, 11
+        word32 i;
+        static const word32 oid_dot_2[] = {
+            2, 100, 4, 6
+        };
+
+        static const byte oid_start_with_2[] = {
+            0x81, 0x34, 0x04, 0x06
+        };
+
```
philljj, line 2707: double blank line

tests/api/test_asn.c
```diff
@@ -2072,15 +2205,156 @@ int test_wc_DecodeObjectId(void)
 
         /* Test 6: Buffer too small for later arcs */
         outSz = 3; /* only room for 3 arcs, but OID has 7 */
-        ExpectIntEQ(DecodeObjectId(oid_sha256rsa, sizeof(oid_sha256rsa),
+        ExpectIntEQ(DecodeObjectId_ex(oid_sha256rsa, sizeof(oid_sha256rsa),
                                    out, &outSz),
                     WC_NO_ERR_TRACE(BUFFER_E));
+
+    /* Test 7: first Arc is 2 */
```
philljj: indentation for `/* Test 7:` is off

tests/api/test_asn.c
```diff
@@ -2072,15 +2205,156 @@ int test_wc_DecodeObjectId(void)
...
+                        sizeof(oid_start_with_2),
+                        out, &outSz), 0);
+            ExpectIntEQ((int)outSz, trueOutSz2);
+            for (i = 0; i < ((outSz <= trueOutSz2) ?
+                        outSz : trueOutSz2); i++) {
+                ExpectIntEQ(out[i], oid_dot_2[i]);
+            }
+        }
+
+        /* Test 8: an OID with an arc that exceeds word16.
```
philljj: comment can be 1 line

tests/api/test_asn.c
```diff
@@ -2072,15 +2205,156 @@ int test_wc_DecodeObjectId(void)
...
+                    WC_NO_ERR_TRACE(BAD_FUNC_ARG));
+        ExpectIntEQ(wc_EncodeObjectId(oid_small, oid_small_cnt, out, NULL),
+                    WC_NO_ERR_TRACE(BAD_FUNC_ARG));
+
+        /* Test 4: output buffer too small */
+        outSz = 1;
+        ExpectIntEQ(wc_EncodeObjectId(oid_small, oid_small_cnt, out, &outSz),
+                    WC_NO_ERR_TRACE(BUFFER_E));
+
+        /* Test 5 : arc greater that SHRT_MAX
```
philljj: comment can be 1 line

wolfcrypt/src/asn.c
```diff
@@ -7389,18 +7389,119 @@ static int CheckCurve(word32 oid)
  * @return  BAD_FUNC_ARG when in or outSz is NULL.
  * @return  BUFFER_E when buffer too small.
  */
-int wc_EncodeObjectId(const word16* in, word32 inSz, byte* out, word32* outSz)
+int wc_EncodeObjectId(const word32* in, word32 inSz,
```
philljj: No reason to break these args across 2 lines.

wolfcrypt/src/asn.c
```diff
@@ -7389,18 +7389,119 @@ static int CheckCurve(word32 oid)
...
+        byte* out, word32* outSz)
 {
-    return EncodeObjectId(in, inSz, out, outSz);
+    return EncodeObjectId_ex(in, inSz, out, outSz);
+}
+
+int EncodeObjectId_ex(const word32* in, word32 inSz,
```
philljj: No reason to break args across 2 lines, they will in 78 columns.

wolfcrypt/src/asn.c
```diff
@@ -7389,18 +7389,119 @@ static int CheckCurve(word32 oid)
...
+ * @param [in]      in     Dotted form of OID.
+ * @param [in]      inSz   Count of numbers in dotted form.
+ * @param [in]      out    Buffer to hold OID.
+ * @param [in, out] outSz  On in, size of buffer.
+ *                         On out, number of bytes in buffer.
+ * @return  0 on success
+ * @return  BAD_FUNC_ARG when in or outSz is NULL.
+ * @return  BUFFER_E when buffer too small.
+ */
+int EncodeObjectId(const word16* in, word32 inSz,
```
philljj: No reason to break function args across 2 lines.

wolfcrypt/src/asn.c
```diff
@@ -7389,18 +7389,119 @@ static int CheckCurve(word32 oid)
...
+    *outSz = (word32)len;
+
+    return 0;
 }
 
-int EncodeObjectId(const word16* in, word32 inSz, byte* out, word32* outSz)
+/* Encode dotted form of OID into byte array version.
+ *
+ * Uses word16 for OIDs and can cause OIDs to be truncated.
+ * Do not use this for new code unless necessary. (use wc_EncodeObjectId
```
philljj: very good comment describing change

wolfcrypt/src/asn.c
```diff
@@ -7481,7 +7582,8 @@ int EncodeObjectId(const word16* in, word32 inSz, byte* out, word32* outSz)
  * @return  BAD_FUNC_ARG when in or outSz is NULL.
  * @return  BUFFER_E when dotted form buffer too small.
  */
-int DecodeObjectId(const byte* in, word32 inSz, word16* out, word32* outSz)
+int DecodeObjectId(const byte* in, word32 inSz,
```
philljj: same, no need to break across 2 lines

wolfcrypt/src/asn.c
```diff
@@ -7524,6 +7635,72 @@ int DecodeObjectId(const byte* in, word32 inSz, word16* out, word32* outSz)
...
+ * @param [in]      in     Byte array containing OID.
+ * @param [in]      inSz   Size of OID in bytes.
+ * @param [in]      out    Array to hold dotted form of OID.
+ * @param [in, out] outSz  On in, number of elements in array.
+ *                         On out, count of numbers in dotted form.
+ * @return  0 on success
+ * @return  BAD_FUNC_ARG when in or outSz is NULL.
+ * @return  BUFFER_E when dotted form buffer too small.
+ */
+int DecodeObjectId_ex(const byte* in, word32 inSz,
```
philljj: no need to break across 2 lines

wolfcrypt/src/asn.c
```diff
@@ -7389,18 +7389,119 @@ static int CheckCurve(word32 oid)
...
+
+    /* guard against word32 overflow of the combined first arc:
+     * ((in[0] * 40) + in[1]) is computed below in both the length and
+     * encode passes, so rejecting it here covers both. */
+    if (in[1] > 0xFFFFFFFFU - (in[0] * 40)) {
+        return BAD_FUNC_ARG;
+    }
+
+    /* compute length of encoded OID */
+    d = ((word32)in[0] * 40) + in[1];
```
philljj: dumb question: are these (word32) casts necessary when `in` and `d` and related are already word32 now?

wolfcrypt/src/asn.c
```diff
@@ -7389,18 +7389,119 @@ static int CheckCurve(word32 oid)
...
+    }
+
+    if (out) {
+        /* verify length */
+        if ((int)*outSz < len) {
+            return BUFFER_E; /* buffer provided is not large enough */
+        }
+
+        /* calc first byte */
+        d = ((word32)in[0] * 40) + in[1];
```
philljj: same question as above.

wolfcrypt/src/wc_pkcs11.c
```diff
@@ -1383,7 +1383,7 @@ static int Pkcs11EccSetParams(ecc_key* key, CK_ATTRIBUTE* tmpl, int idx)
         unsigned char* derParams = tmpl[idx].pValue;
     #if defined(HAVE_OID_ENCODING)
         word32 oidSz = ECC_MAX_OID_LEN - 2;
-        ret = wc_EncodeObjectId(key->dp->oid, key->dp->oidSz, derParams+2, &oidSz);
+        ret = EncodeObjectId(key->dp->oid, key->dp->oidSz, derParams+2, &oidSz);
```
philljj, line 1386: `EncodeObjectId` is being used here, I guess to preserve the original `word16` behavior?

But this function is not public:
```
#ifdef HAVE_OID_ENCODING
    WOLFSSL_API int wc_EncodeObjectId(const word32* in, word32 inSz,
        byte* out, word32* outSz);
    WOLFSSL_LOCAL int EncodeObjectId_ex(const word32* in, word32 inSz,
        byte* out, word32* outSz);
    /* Unchangeable due to being called from FIPS code expecting word16 in */
    WOLFSSL_LOCAL int EncodeObjectId(const word16* in, word32 inSz,
        byte* out, word32* outSz);
#endif
```

Consumers of our library will not be able to make this switch to `EncodeObjectId()` if they want. This spooks me for causing disruptions with integrations with our library.

Suggestion: instead of removing the old word16 public `wc_EncodeObjectId`, why not keep it but add a separate word32 `wc_EncodeObjectId_ex`? It might make fips interop and other things more pleasant.

philljj (changes requested):

wolfcrypt/src/wc_pkcs11.c
aidankeefe2022: The reason this is being kept is because fips code that uses this api expects an array of word16 and changing any types breaks fix integration tests. Fips code that uses this api also is lucky that all possible OIDs it encodes fix in word16 arcs so there is no bugs yet! Other than that the API break is real but I spoke around and looks like the best course of action.

In the referenced code key->dp->oid is ecc_oid_t which is word16 and something that we cannot touch with out breaking FIPs integration

wolfcrypt/src/asn.c
aidankeefe2022: No left over from old code

wolfcrypt/src/asn.c
aidankeefe2022: no its a left over from the old code

wolfcrypt/src/asn.c
aidankeefe2022: fixed

wolfcrypt/src/asn.c
aidankeefe2022: fixed

wolfcrypt/src/asn.c
aidankeefe2022: Thank you!

wolfcrypt/src/asn.c
aidankeefe2022: fixed

wolfcrypt/src/asn.c
aidankeefe2022: fixed

wolfcrypt/src/asn.c
aidankeefe2022: fixed

tests/api/test_asn.c
aidankeefe2022: fixed

tests/api/test_asn.c
aidankeefe2022: fixed

wolfcrypt/src/asn.c
philljj: `in` was `const word16* in` in the old code though, so the `(word32)` cast was necessary.

tests/api/test_asn.c
aidankeefe2022: fixed

tests/api/test_asn.c
aidankeefe2022: fixed

wolfcrypt/src/asn.c
aidankeefe2022: Sorry that was not a clear reply haha. This cast is a leftover from the old word16 version and not needed in the word32 version (I am removing it) but still needed for the word16 version as you said!

tests/api/test_asn.c
aidankeefe2022: fixed

tests/api/test_asn.c
aidankeefe2022: Added comment!

wolfcrypt/src/wc_pkcs11.c
ejohnstown: My two bits, wc_EncodeObjectId() has been published in v5.9.2-stable as taking a pointer to 16-bit words. That API/ABI is in the world now. Changing it is breaking things for everyone, not just FIPS users. We should be adding wc_EncodeObjectId32(). (I would reserve "_ex" for adding parameters, personally.)

aidankeefe2022: My main concern with keeping both word16 and word32 is that having parallel correct and incorrect functionality is confusing (IMO) and that our current word16 implementation can cause real bugs. The original plan was for this to be a breaking change in 5.9.2 but it got pushed back. I am open to adding the parallel behavior for sure but that's why I have hesitated on doing so!

aidankeefe2022:
There is a new set of changes to update new asn.c code to the new word32 OID

aidankeefe2022:
retest this please

wolfcrypt/src/wc_pkcs11.c
aidankeefe2022: Adding the 32 bit as a new function!

aidankeefe2022:
retest this please

wolfssl/wolfcrypt/asn.h
```diff
@@ -1850,13 +1850,12 @@ typedef struct WOLFSSL_AIA_ENTRY {
...
 } WOLFSSL_AIA_ENTRY;
 #endif /* WOLFSSL_AIA_ENTRY_DEFINED */
-
 #ifdef WC_ASN_UNKNOWN_EXT_CB
-typedef int (*wc_UnknownExtCallback)(const word16* oid, word32 oidSz, int crit,
-                                     const unsigned char* der, word32 derSz);
-typedef int (*wc_UnknownExtCallbackEx)(const word16* oid, word32 oidSz,
-                                       int crit, const unsigned char* der,
-                                       word32 derSz, void *ctx);
+    typedef int (*wc_UnknownExtCallback)(const word32* oid, word32 oidSz,
```
philljj: This could potentially break existing `wc_UnknownExtCallback` users?

wolfcrypt/src/asn.c
```diff
@@ -39006,64 +39195,6 @@ int wc_Asn1_SetOidToNameCb(Asn1* asn1, Asn1OidToNameCb nameCb)
...
- * @param [in]      in     Byte array containing OID.
- * @param [in]      inSz   Size of OID in bytes.
- * @param [in]      out    Array to hold dotted form of OID.
- * @param [in, out] outSz  On in, number of elements in array.
- *                         On out, count of numbers in dotted form.
- * @return  0 on success
- * @return  BAD_FUNC_ARG when in or outSz is NULL.
- * @return  BUFFER_E when dotted form buffer too small.
- */
-static int EncodedDottedForm(const byte* in, word32 inSz, word32* out,
```
philljj, line 39808: There are still MCDC tests referencing this deleted function:
```
grep -r "EncodedDottedForm" *
tests/unit-mcdc/test_asn_certgen_whitebox.c: *   EncodedDottedForm       :~38862  in==NULL || outSz==NULL
tests/unit-mcdc/test_asn_certgen_whitebox.c:    WB_NOTE("EncodedDottedForm(): NULL-arg OR [~38862] (via PrintObjectIdNum path, "
tests/unit-mcdc/test_asn_certgen_whitebox.c:        ret = EncodedDottedForm(NULL, 3, dotted, &num);
tests/unit-mcdc/test_asn_certgen_whitebox.c:        ret = EncodedDottedForm(oidBytes, sizeof(oidBytes), dotted, NULL);
tests/unit-mcdc/test_asn_certgen_whitebox.c:        ret = EncodedDottedForm(oidBytes, sizeof(oidBytes), dotted, &num);
```
We should preserve this function's name, or update the tests, so MCDC coverage is uninterrupted.

philljj (changes requested):

wolfcrypt/src/asn.c
aidankeefe2022: I'll update these tests! This was an identical copy of DecodeObjectId with a different name and incorrect documentation: It actually does the opposite of encoding a dot form OID into a byte array.

wolfssl/wolfcrypt/asn.h
aidankeefe2022: Yes I think it is worth it to break. Users will get errors but the update is easy enough to change to 32 bit.

wolfcrypt/src/asn.c
aidankeefe2022: The Asn1Printing functionality uses the new 32 bit DecodeOidId32 function so there is no static helper for this task anymore; removing the tests.

philljj: Ok that's fair.

wolfssl/wolfcrypt/asn.h
ejohnstown: I don't think this is a good enough reason to break the API/ABI. (This is spoken as someone who has done it, and we're still paying the price.) This will break some users.

aidankeefe2022:
retest this please

aidankeefe2022:
retest this please

aidankeefe2022:
retest this please

aidankeefe2022:
retest this please

aidankeefe2022:
retest this please

aidankeefe2022:
merge conflit

PR #10718, opened 2026-06-17
Aes key wrap with padding

night1rider:
Adding AES Key Wrapping with Padding: https://datatracker.ietf.org/doc/html/rfc5649, and callback functionality for it.


night1rider:
jenkins retest this please

night1rider:
jenkins retest this please

dgarske (changes requested):
Please resolve merge conflicts. Thanks

dgarske:
Jenkins retest this please. scan-build java.io.IOException: Unexpected EOF

tests/api/test_aes.c
```diff
@@ -10431,6 +10859,271 @@ int test_wc_AesSivEncryptDecrypt(void)
...
+ *----------------------------------------------------------------------------*/
+
+#if defined(WOLF_CRYPTO_CB) && defined(HAVE_AES_KEYWRAP) && \
+    !defined(NO_AES) && defined(WOLFSSL_AES_128)
+
+#include <wolfssl/wolfcrypt/cryptocb.h>
+
+/* Test CryptoCB device IDs (must be unique across test_aes.c); 7/8/9 are used
+ * by the SetKey/AES-GCM/TLS13 tests above. */
+#define TEST_CRYPTOCB_KEYWRAP_DEVID  10
```
dgarske: Recommendation: Use unused IDs (13/14), and correct the comment — 7/8/9/10/11/12 are all taken and the referenced tests are below, not above. Better still, hoist a single central list of test devIds so the uniqueness rule is enforceable rather than restated per test.

wolfcrypt/test/test.c
```diff
@@ -23371,8 +23380,587 @@ WOLFSSL_TEST_SUBROUTINE wc_test_ret_t aeskeywrap_test(void)
...
+        forgeBlk[0] = 0xa6; forgeBlk[1] = 0x59; forgeBlk[2] = 0x59;
+        forgeBlk[3] = 0xa6;
+        forgeBlk[8] = 0x11; /* arbitrary recovered plaintext octet */
+        fr = wc_AesInit(faes, HEAP_HINT, devId);
+        if (fr == 0)
+            fr = wc_AesSetKey(faes, test_kwp[0].kek, test_kwp[0].kekLen, NULL,
+                              AES_ENCRYPTION);
+        if (fr == 0)
+            fr = wc_AesEncryptDirect(faes, forged, forgeBlk);
+        wc_AesFree(faes);
```
dgarske: In both forged-ciphertext blocks, `faes` comes from a bare `XMALLOC` with no `XMEMSET`, and `wc_AesFree(faes)` is called unconditionally — including on the path where `wc_AesInit()` failed and therefore never ran its own zeroing. `wc_AesFree()`

wolfcrypt/src/cryptocb.c
```diff
@@ -2270,6 +2271,89 @@ int wc_CryptoCb_AesEcbDecrypt(Aes* aes, byte* out,
...
+        cryptoInfo.cipher.enc = 1;
+        cryptoInfo.cipher.aeskeywrap.aes = aes;
+        cryptoInfo.cipher.aeskeywrap.in = in;
+        cryptoInfo.cipher.aeskeywrap.inSz = inSz;
+        cryptoInfo.cipher.aeskeywrap.out = out;
+        cryptoInfo.cipher.aeskeywrap.outSz = outSz;
+        cryptoInfo.cipher.aeskeywrap.iv = iv;
+        cryptoInfo.cipher.aeskeywrap.pad = pad;
+
+        ret = dev->cb(dev->devId, &cryptoInfo, dev->ctx);
```
dgarske, line 2307: Recommendation: Reject `outResSz > outSz` in both dispatchers

dgarske (changes requested):

tests/api/test_aes.c
night1rider: Addressed. Switched to 13 and 14 and rewrote the comment to list what is actually taken (7 SetKey, 8 AES-GCM, 9 TLS13, 10 AES-CFB, 11 AES-OFB) and that those tests sit below.

wolfcrypt/test/test.c
night1rider: Addressed in both blocks. `wc_AesFree(faes)` now only runs inside the `wc_AesInit` success branch, so an `Aes` that was never initialized never reaches it. The raw malloc is still released with `XFREE` on the failure path.

wolfcrypt/src/cryptocb.c
night1rider: Addressed. `wc_CryptoCb_AesKeyWrap` and `wc_CryptoCb_AesKeyUnWrap` both return `BUFFER_E` now when the device reports an `outResSz` larger than the caller's `outSz`.

wolfcrypt/src/aes.c
```diff
@@ -17358,6 +17464,301 @@ int wc_AesKeyUnWrap(const byte* key, word32 keySz, const byte* in, word32 inSz,
...
+         * block, so scan it at fixed offsets, requiring zero where off >= MLI. */
+        for (j = 0; j < KEYWRAP_BLOCK_SIZE; j++) {
+            word32 off = lastBlk + j;
+            fail |= ctMaskWord32GTE(off, mli)          /* off >= MLI */
+                    & (word32)ctMaskNotEq((int)out[off], 0);
+        }
+#else
+        /* No word64: compare in int range.  MLI with its high bit set (>= 2^31
+         * > 8*n) is forced to fail so the int compares see valid values. */
+        {
```
dgarske: I don't like empty braces. Let's fix that. Are we testing the non 64-bit case too?

wolfcrypt/src/aes.c
```diff
@@ -17358,6 +17464,301 @@ int wc_AesKeyUnWrap(const byte* key, word32 keySz, const byte* in, word32 inSz,
...
+        word32 j;
+
+        /* check 1: MSB(32,A) == constant */
+        fail = (word32)ctMaskNotEq(ConstantCompare(a, expConst,
+                                           (int)sizeof(kwpAivConst)), 0);
+
+#ifdef WORD64_AVAILABLE
+        /* check 2: 8*(n-1) < MLI <= 8*n */
+        fail |= ~(ctMaskWord32GTE(mli, lastBlk + 1)   /* MLI >= 8*(n-1)+1 */
+                & ctMaskWord32GTE(dataSz, mli));       /* 8*n >= MLI       */
```
dgarske, line 17726: NIT: Let's make the comments line up

dgarske (changes requested):

wolfcrypt/src/aes.c
night1rider: Addressed

For testing added an `aeskeywrap-padding-no-64bit os-check` entry that builds with `-m32  -DNO_64BIT` and runs the key wrap tests.

wolfcrypt/src/aes.c
night1rider: Addressed

night1rider:
rebased on master to fix some CI

dgarske (dismissed):

dgarske (approved):

PR #10728, opened 2026-06-18
Intel x86/x64 assembly fixes

SparkiDev:
# Description

Allow x86 to build with assembly for enable all - disable assembly when x86 and not assembly code available. Add file fe_operations.c when assembly and x86. x86: fix ECB decrypt to use corect offsets for parameters
     fix AES-NI and AVX1 assembly code
     fix sp_int with assembly to compile
     minor optimizations of AES-GCM
x64: Don't emit move instruction if source and destination are the same reg
     Use xor instead of mov 0
     minor optimizations of AES-GCM for AES-NI

# Testing

./autogen.sh
./configure --disable-shared LDFLAGS=--static --host=x86 CC=i686-linux-gnu-gcc --enable-intelasm --enable-all --disable-crl-monitor --enable-mldsa
and
./configure --disable-shared --enable-intelasm --enable-all --enable-mldsa

Change cpuid.c to have AVX2/AVX1 disabled.

### DIFF
--- a/src/include.am
+++ b/src/include.am
@@ -1885,7 +1885,9 @@ src_libwolfssl@LIBSUFFIX@_la_SOURCES += wolfcrypt/src/ge_operations.c
 if BUILD_CURVE25519_INTELASM
 if !BUILD_X86_ASM
 src_libwolfssl@LIBSUFFIX@_la_SOURCES += wolfcrypt/src/fe_x25519_asm.S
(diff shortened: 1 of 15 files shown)


SparkiDev:
Code generated from PR:
https://github.com/wolfSSL/scripts/pull/592

dgarske (approved):

PR #10731, opened 2026-06-18
CI: only save dependency caches on master, restore on PRs

julek-wolfssl:
GitHub Actions caches are branch-scoped: an entry written by a `pull_request`
run lives under `refs/pull/<N>/merge` and is invisible to other PRs. The
haproxy, mbedtls, nss, ntp, threadx and hostap-vm workflows used combined
`actions/cache` with fixed keys, so every PR re-saved its own copy of the same
dependency, yielding one duplicate cache entry per PR.

**`CI: only save dependency caches on master, restore on PRs`**

- Split each into `actions/cache/restore` (always) plus `actions/cache/save`
  gated to `refs/heads/master`, and add a daily schedule so a master run
  reseeds the single shared entry that all PRs restore.
- mbedtls/nss save in their build job only; the test jobs restore-only.
- Disable the `setup-msys2` package cache: the action only toggles caching
  on/off and cannot save on master while restoring on PRs.

**`CI: factor threadx NetXDuo version into NETXDUO_REF env var`**

- The `v6.4.3_rel` version was repeated in the cache path, cache key, download
  URL and extract command. Define it once as a workflow-level env var and
  reference it everywhere.

### DIFF
--- a/.github/workflows/haproxy.yml
+++ b/.github/workflows/haproxy.yml
@@ -7,6 +7,9 @@ on:
   pull_request:
     types: [opened, synchronize, reopened, ready_for_review]
     branches: [ '*' ]
+  # Daily run on master reseeds the shared cache (see save step below).
+  schedule:
+    - cron: '0 4 * * *'
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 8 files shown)


dgarske:
Jenkins restest this please: "ERROR: Build 'PRB-generic-config-parser' failed with result: FAILURE" -> "DTLS" -> "FAIL scripts/unit.test (exit status: 1)"

dgarske (approved):

PR #10738, opened 2026-06-18
Add WOLFSSL_ASYNC_CERT_YIELD: per-certificate non-blocking yield

dgarske:
Adds an opt-in build option, `WOLFSSL_ASYNC_CERT_YIELD`, that makes `ProcessPeerCerts()` return `WC_PENDING_E` to the caller after each certificate in the peer's chain (and after the leaf) is verified, instead of verifying the whole chain in one `wolfSSL_connect()` / `wolfSSL_accept()` call. This lets a cooperative, single-threaded scheduler regain control between certificates to service a watchdog. Requires `WOLFSSL_ASYNC_CRYPT`; off by default. Independent of `WC_ECC_NONBLOCK` (which subdivides each verify further).

ZD 21071


JacobBarthelmeh (approved):
Ran the additions to the example async applications.

```
bash-3.2$ ./async_server --cert-chain
Async server mode: secp256r1, TLS 1.3
Waiting for a connection...
Negotiated cipher: TLS13-AES128-GCM-SHA256
Negotiated group: SECP256R1
Client connected successfully
Client: GET / HTTP/1.1
Host: 127.0.0.1
User-Agent: wolfSSL-async
Connection: close


WANT_READ/WRITE count: 47462
WC_PENDING_E count: 35870
```

The feature looks well contained to async builds and useful, thanks!

PR #10745, opened 2026-06-22
Enable support for mandatory PSKs

Frauschi:
Add new APIs, `wolfSSL_CTX_require_psk()` / `wolfSSL_require_psk()`, to require that an external Pre-Shared Key is negotiated for a (D)TLS 1.3 handshake to succeed. When set on an endpoint that has a PSK callback registered, a handshake that completes without negotiating an external PSK is aborted with the new error `PSK_MISSING_ERROR` instead of falling back to a certificate handshake.

This is relevant for external PSKs only (not session tickets) to make sure the PSKs are used as an additional security factor (together with (EC)DHE key exchange). Session-ticket resumption is exempt, and to preserve forward secrecy, a pure `psk_ke` handshake is rejected with `PSK_KEY_ERROR`. When used with `WOLFSSL_CERT_WITH_EXTERN_PSK`, it also makes sure that peers are properly authenticated with both the PSK and via certificates.

The new APIs sit alongside the existing `wolfSSL_[CTX_]no_dhe_psk()` / `wolfSSL_[CTX_]only_dhe_psk()` PSK options and do not depend on certificate support, so the feature is also usable in NO_CERTS (PSK-only) builds. They apply to (D)TLS 1.3 only and return `BAD_FUNC_ARG` for a non-TLS-1.3 context.

In (D)TLS 1.2 the use of a PSK is encoded directly in the negotiated cipher suite, so a mandatory PSK is enforced simply by restricting the cipher suite list to (preferably (EC)DHE-)PSK suites.

Fixes zd#22012.

### Testing

Added unit tests for the new option.

### Checklist

 - [x] added tests
 - [x] updated/added doxygen


Frauschi:
Jenkins retest this please

douzzer (approved):

PR #10747, opened 2026-06-22
Reject status_request_v2 ocsp-multi staples that bundle multiple SingleResponses

gasbytes:
# Description

Reject status_request_v2 ocsp-multi staples that bundle multiple SingleResponses
So a revoked certificate's non-first single can no longer be accepted

# Testing

Added regression test + generated blob with the certificate chain to reproduce the behaviour.
config used: ./configure --enable-all

# Checklist

 - [x] added tests
 - [ ] updated/added doxygen


src/internal.c
```diff
@@ -17953,6 +17953,17 @@ static int DoCertificateStatus(WOLFSSL* ssl, byte* input, word32* inOutIdx,
...
                         ret = BAD_CERTIFICATE_STATUS_ERROR;
 
+                    /* Bundling more than one SingleResponse inside a single
+                     * stapled BasicOCSPResponse is not supported. The per-cert
+                     * status checked above comes from the first SingleResponse
+                     * in wire order, which is not necessarily the one matching
+                     * the stapled certificate (CompareOcspReqResp() below is
+                     * what locates the match). Rather than risk acting on the
+                     * wrong SingleResponse, reject any response that carries
+                     * multiple singles. */
```
julek-wolfssl: Just say its not supported. Comment too long.

julek-wolfssl (changes requested):

julek-wolfssl (dismissed):

dgarske (commented):
## Skoll Multi-Scan Review

**Modes:** `review-security`**Overall recommendation:** `APPROVE`

### Posted findings
- [Low] [review-security] **New test's method array can be an empty initializer under TLS1.3-only + ocspstapling2 builds** — `tests/api/test_ocsp.c:1270-1281`
- [Info] [review-security] **Fix rejects all multi-SingleResponse BasicOCSPResponses (conservative vs. RFC 6960)** — `src/internal.c:17956-17959`

tests/api/test_ocsp.c
```diff
@@ -1218,11 +1218,140 @@ int test_ocsp_tls_cert_cb(void)
...
+/*
+ * status_request_v2 OCSP-multi must reject a stapled leaf response
+ * whose MATCHING SingleResponse is CERT_REVOKED, even when an unrelated
+ * CERT_GOOD SingleResponse appears first in the same BasicOCSPResponse.
+ */
+int test_ocsp_status_request_v2_multi_revoked_single(void)
+{
+    EXPECT_DECLS;
+    size_t i;
+    struct {
```
dgarske, line 1273: **🔵 [Low] New test's method array can be an empty initializer under TLS1.3-only + ocspstapling2 builds** · _Logic_

The new test function's `params[]` array is populated only inside `#if !defined(WOLFSSL_NO_TLS12)` (TLSv1_2 and, optionally, DTLSv1_2). The function body itself is compiled whenever the outer feature guard holds AND `HAVE_CERTIFICATE_STATUS_REQUEST_V2` is defined (inner `#ifdef` at line 1221). In a TLS 1.3-only build that still enables status_request_v2 stapling (`--enable-tls13 --disable-tls12 --enable-ocspstapling2`, i.e. `WOLFSSL_NO_TLS12` defined and `HAVE_CERTIFICATE_STATUS_REQUEST_V2` defined), every array element is `#ifdef`-ed out, leaving `struct {...} params[] = { };`. An empty brace initializer for an array of unknown size is a zero-length array that GCC/Clang flag under `-Wpedantic` ("ISO C forbids empty initializer braces"); with `-Werror` this breaks the build. Note the pre-existing `test_ocsp_tls_cert_cb` avoids this because its `params[]` always includes `WOLFSSL_TLS13` entries; this new test has no TLS1.3 fallback entry, so it introduces the empty-array possibility. Runtime behavior is otherwise benign (the loop simply does not execute).

**Fix:** Guard the whole function body with `#if !defined(WOLFSSL_NO_TLS12)` (the fix under test lives in TLS1.2-only code anyway) and provide a `TEST_SKIPPED` stub for the `WOLFSSL_NO_TLS12` case, or add a trailing sentinel entry / `#if` so `params[]` is never empty. This keeps the test compiling cleanly under a TLS1.3-only + ocspstapling2 configuration.

src/internal.c
```diff
@@ -17953,6 +17953,11 @@ static int DoCertificateStatus(WOLFSSL* ssl, byte* input, word32* inOutIdx,
                     ||  (response->single->status->status != CERT_GOOD))
                         ret = BAD_CERTIFICATE_STATUS_ERROR;
 
+                    /* Bundling more than one SingleResponse inside a single
```
dgarske, line 17956: **⚪ [Info] Fix rejects all multi-SingleResponse BasicOCSPResponses (conservative vs. RFC 6960)** · _Logic_

The added check `if (ret == 0 && response->single->next != NULL) ret = BAD_CERTIFICATE_STATUS_ERROR;` rejects any stapled BasicOCSPResponse that carries more than one SingleResponse. RFC 6960 permits a BasicOCSPResponse to contain multiple SingleResponses, so a responder that legitimately bundles statuses into one stapled response (uncommon under RFC 6961 status_request_v2, where one response per certificate is expected) would now be rejected. This is the intended, security-hardening behavior of the PR and matches wolfSSL's one-single-per-stapled-response model; it is safer than the alternative of validating only the request-matching single's status. Noted only so the interop trade-off is explicit. The chosen placement (before `CompareOcspReqResp` reorders `response->single`) is correct and is what makes the fix effective.

**Fix:** No change required. Optionally document the one-SingleResponse-per-staple expectation in the OCSP stapling API docs so integrators using bundled responders understand the rejection.

dgarske (changes requested):
https://github.com/wolfSSL/wolfssl/pull/10747#pullrequestreview-4656975322

gasbytes:

tests/api/test_ocsp.c
```diff
@@ -1218,11 +1218,138 @@ int test_ocsp_tls_cert_cb(void)
...
+ * BasicOCSPResponse that bundles a benign CERT_GOOD single FIRST and the
+ * server1 leaf cert's CERT_REVOKED single SECOND. */
+static int test_ocsp_revoked_multi_single_status_cb(WOLFSSL* ssl, void* ioCtx)
+{
+    byte* leaf_resp = NULL;
+    byte* int_resp = NULL;
+    byte* root_resp = NULL;
+    int ret = WOLFSSL_OCSP_STATUS_CB_ALERT_FATAL;
+    (void)ioCtx;
+    leaf_resp = (byte*)XMALLOC(sizeof(resp_server1_revoked_good_first), NULL, 0);
```
dgarske: Why is the XMALLOC using 0 in these for the DYNAMIC_TYPE?

dgarske (changes requested):

tests/api/test_ocsp.c
gasbytes: Right, it should be *_TMP_BUFFER, fixing it right now.
Thanks.

tests/api/test_ocsp.c
```diff
@@ -1218,11 +1218,141 @@ int test_ocsp_tls_cert_cb(void)
...
+        if (wolfSSL_set_tlsext_status_ocsp_resp_multi(ssl, int_resp,
+                sizeof(resp_intermediate1_cert), 1) == WOLFSSL_SUCCESS)
+            int_resp = NULL;
+        if (wolfSSL_set_tlsext_status_ocsp_resp_multi(ssl, root_resp,
+                sizeof(resp_root_ca_cert), 2) == WOLFSSL_SUCCESS)
+            root_resp = NULL;
+        if (leaf_resp == NULL && int_resp == NULL && root_resp == NULL)
+            ret = WOLFSSL_OCSP_STATUS_CB_OK;
+    }
+    XFREE(leaf_resp, NULL, 0);
```
dgarske: These also need to be DYNAMIC_TYPE_TMP_BUFFER

dgarske (changes requested):

dgarske (approved):

PR #10749, opened 2026-06-22
Add fips-hash-offline.sh

holtrop-wolfssl:
# Description

This script calculates the FIPS integrity hash for an already-linked binary offline at build time.

Tested with fips/optest-140-3/static-building/optest-run-static-test.sh

# Testing

How did you test?

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/fips-hash-offline.sh
+++ b/fips-hash-offline.sh
@@ -1,0 +1,170 @@
+#!/bin/bash
+
+# fips-hash-offline.sh
+#
+# Copyright (C) 2006-2026 wolfSSL Inc.
+#
+# This file is part of wolfSSL.
+#
+# wolfSSL is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
+# wolfSSL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1335, USA
+
+# This script computes the wolfCrypt FIPS in-core integrity hash at compile
+# time directly from an already linked binary, and patches the result into the
+# verifyCore[] array inside that binary.
+#
+# This reproduces, at build time and against the final ELF image, exactly what
+# DoInCoreCheck() in wolfcrypt/src/fips_test.c computes at run time:
+#
+#   HMAC-SHA256( coreKey,
+#                .text  bytes in [wolfCrypt_FIPS_first, wolfCrypt_FIPS_last)
+#                ||
+#                .rodata bytes in [wolfCrypt_FIPS_ro_start, wolfCrypt_FIPS_ro_end)
+#                with the verifyCore[] bytes skipped )
+#
+# Because verifyCore[] is excluded from the digest, overwriting it with the
+# computed value does not change the digest, so a single pass over the linked
+# binary is sufficient -- no rebuild and no scraping of the module's reported
+# hash is required.
+#
+# Assumptions (match the optest user-space static FIPS build):
+#   - single fixed-address (ET_EXEC) image, not stripped
+#   - measured .text/.rodata bracketed by the wolfCrypt_FIPS_first/last and
+#     wolfCrypt_FIPS_ro_start/ro_end symbols
+#   - no text-segment canonicalizer and no relocation-table indirection in
+#     effect (i.e. the plain #else path of DoInCoreCheck())
+#
+# Usage: fips-hash-offline.sh [path-to-binary]
+#   default binary: wolfcrypt/test/testwolfcrypt
+
+set -u
+
+BIN="${1:-wolfcrypt/test/testwolfcrypt}"
+
+die() { echo "fips-hash-offline: error: $*" >&2; exit 1; }
+
+[ -f "$BIN" ] || die "binary not found: $BIN"
+command -v awk     >/dev/null 2>&1 || die "awk not found"
+command -v dd      >/dev/null 2>&1 || die "dd not found"
+command -v readelf >/dev/null 2>&1 || die "readelf not found"
+command -v openssl >/dev/null 2>&1 || die "openssl not found"
+command -v xxd     >/dev/null 2>&1 || die "xxd not found"
+
+# Symbol lookup from the ELF symbol table.
+# readelf -s columns: Num: Value Size Type Bind Vis Ndx Name
+#   -> Value ($2) is hexadecimal, Size ($3) is decimal, Ndx ($7), Name ($8).
+sym_val()  { readelf -sW "$BIN" | awk -v n="$1" '$8==n && $7!="UND"{print $2; exit}'; }
+sym_size() { readelf -sW "$BIN" | awk -v n="$1" '$8==n && $7!="UND"{print $3; exit}'; }
+
+# Map a virtual address to a file offset using the containing PROGBITS section.
+# This avoids assuming any particular load base or section padding.
+vaddr_to_off() {
+    local v="$1"
+    local name type addr off size a o s
+    while read -r name type addr off size; do
+        [ "$type" = "PROGBITS" ] || continue
+        a=$((0x$addr)); o=$((0x$off)); s=$((0x$size))
+        if [ "$v" -ge "$a" ] && [ "$v" -lt $((a + s)) ]; then
+            echo $((o + v - a))
+            return 0
+        fi
+    done < <(readelf -SW "$BIN" | sed 's/\[[ 0-9]*\]//' | awk '/PROGBITS/{print $1, $2, $3, $4, $5}')
+    return 1
+}
+
+# Extract byte_count bytes starting at file_offset (both in bytes) to stdout.
+# Uses GNU dd byte-granular skip/count with a large block size for speed.
+extract() {
+    dd if="$BIN" bs=1M iflag=skip_bytes,count_bytes skip="$1" count="$2" 2>/dev/null
+}
+
+# --- gather the FIPS boundary and verifyCore/coreKey symbols ---
+FIRST_H=$(sym_val wolfCrypt_FIPS_first)
+LAST_H=$(sym_val wolfCrypt_FIPS_last)
+ROSTART_H=$(sym_val wolfCrypt_FIPS_ro_start)
+ROEND_H=$(sym_val wolfCrypt_FIPS_ro_end)
+VC_H=$(sym_val verifyCore)
+VCSZ=$(sym_size verifyCore)
+KEY_H=$(sym_val coreKey)
+KEYSZ=$(sym_size coreKey)
+
+[ -n "$FIRST_H" ] && [ -n "$LAST_H" ] && [ -n "$ROSTART_H" ] && [ -n "$ROEND_H" ] \
+    && [ -n "$VC_H" ] && [ -n "$VCSZ" ] && [ -n "$KEY_H" ] && [ -n "$KEYSZ" ] \
+    || die "missing FIPS boundary symbols (is $BIN a non-stripped static FIPS build?)"
+
+first=$((0x$FIRST_H))
+last=$((0x$LAST_H))
+rostart=$((0x$ROSTART_H))
+roend=$((0x$ROEND_H))
+vc=$((0x$VC_H))
+keyaddr=$((0x$KEY_H))
+
+[ "$last"  -gt "$first"   ] || die "wolfCrypt_FIPS_last <= wolfCrypt_FIPS_first"
+[ "$roend" -gt "$rostart" ] || die "wolfCrypt_FIPS_ro_end <= wolfCrypt_FIPS_ro_start"
+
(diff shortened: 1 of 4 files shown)


fips-hash-offline.sh
```diff
@@ -0,0 +1,170 @@
...
+# the same verifyCore exclusion as DoInCoreCheck().
+NEWHASH=$(
+    {
+        extract "$codeoff" $((last - first))
+        if [ "$vc_in_ro" -eq 1 ]; then
+            extract "$roff" $((vc - rostart))
+            extract "$aoff" $((roend - vc - VCSZ))
+        else
+            extract "$roff" $((roend - rostart))
+        fi
```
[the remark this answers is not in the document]
holtrop-wolfssl: verifyCore can't end at exactly wolfCrypt_FIPS_ro_end; this is not a concern.

fips-hash-offline.sh
```diff
@@ -0,0 +1,170 @@
...
+[ "$last"  -gt "$first"   ] || die "wolfCrypt_FIPS_last <= wolfCrypt_FIPS_first"
+[ "$roend" -gt "$rostart" ] || die "wolfCrypt_FIPS_ro_end <= wolfCrypt_FIPS_ro_start"
+
+# Select the digest from the size of verifyCore[] (= digest_bytes*2 + 1).
+digest_bytes=$(( (VCSZ - 1) / 2 ))
+case "$digest_bytes" in
+    32) ALG=sha256 ;;
+    48) ALG=sha384 ;;
+    *)  die "unexpected verifyCore size ($VCSZ); cannot determine digest" ;;
+esac
```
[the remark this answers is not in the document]
holtrop-wolfssl: This is not a concern.

fips-hash-offline.sh
```diff
@@ -0,0 +1,170 @@
...
+
+vc_in_ro=0
+if [ "$vc" -ge "$rostart" ] && [ "$vc" -lt "$roend" ]; then
+    vc_in_ro=1
+    aoff=$(vaddr_to_off $((vc + VCSZ))) || die "cannot map verifyCore tail"
+fi
+
+# Compute the digest, streaming the measured bytes in the same order and with
+# the same verifyCore exclusion as DoInCoreCheck().
+NEWHASH=$(
```
dgarske, line 143: **🟠 [Medium] Hash pipeline can silently emit a wrong-but-correct-length hash (no pipefail, dd errors hidden)**

The script runs `set -u` but not `set -o pipefail`, and every `extract()` (dd) call redirects stderr to `/dev/null`. In the NEWHASH subshell the measured bytes are streamed through three `extract` calls into `openssl dgst`. If one of the `extract` calls fails or returns short (e.g. a bad offset, an unexpected non-GNU `dd`, a transient I/O error), the failure is swallowed and `openssl` still produces a valid-length HMAC over whatever bytes it received. The only guard is `[ "${#NEWHASH}" -eq $((digest_bytes*2)) ]`, which passes for ANY complete digest regardless of what was hashed. The subsequent "Confirm the patch landed" check (lines 161-163) only re-reads the bytes it just wrote and compares them to NEWHASH — it does NOT validate that NEWHASH is the correct integrity hash. Net effect: a silently incorrect verifyCore can be baked in and the script reports success; the error only surfaces later as a runtime FIPS in-core integrity failure.

**Fix:** Add `set -o pipefail` and consider failing hard if any `extract` returns fewer bytes than requested, so a partial read cannot yield a plausible-but-wrong hash.

fips-hash-offline.sh
```diff
@@ -0,0 +1,170 @@
...
+)
+[ "${#NEWHASH}" -eq $((digest_bytes * 2)) ] || die "hash computation failed"
+
+# Overwrite the first digest_bytes*2 ASCII characters of verifyCore[] in place.
+# The trailing NUL terminator (byte digest_bytes*2) is left untouched.
+vcoff=$(vaddr_to_off "$vc") || die "cannot map verifyCore address to file offset"
+printf '%s' "$NEWHASH" | dd of="$BIN" bs=1M oflag=seek_bytes conv=notrunc seek="$vcoff" 2>/dev/null \
+    || die "failed to write verifyCore"
+
+# Confirm the patch landed.
```
dgarske, line 163: **🟠 [Medium] Offline hash is never validated against the module's own self-test**

Unlike fips-hash.sh, which obtains the hash by executing testwolfcrypt (so the value is by construction what the module computes), the offline script recomputes the digest independently under several documented assumptions (ET_EXEC, no text-segment canonicalizer, no relocation-table indirection, verifyCore excluded only from rodata). If any of those assumptions does not hold for a given build, the script will happily compute and write a WRONG hash, and its own final check (lines 161-163) confirms only that the bytes were written, giving false confidence. The failure then only appears at runtime as a FIPS integrity error. For a FIPS-critical helper this false-success mode is worth closing.

**Fix:** Either run the patched binary to confirm the FIPS self-test passes when it is runnable, or make the header comment explicitly warn that a successful run does NOT prove the hash is correct if the stated assumptions are violated.

fips-hash-offline.sh
```diff
@@ -0,0 +1,170 @@
+#!/bin/bash
```
dgarske, line 1: **🟠 [Medium] New 170-line FIPS helper script has no automated test**

The PR adds a substantial new script that patches a FIPS integrity value into a linked binary, but adds no repository test or CI hook; the PR checklist item 'added tests' is unchecked and testing is described only as a one-off manual optest run. Because the script's whole value is producing a byte-exact match to DoInCoreCheck(), a silent regression (e.g. an ordering change, a readelf column-format change across toolchain versions, or a dd-flag portability issue) would not be caught. A smoke test that builds the static FIPS optest target, runs fips-hash-offline.sh, and then runs the binary's self-test would catch these.

**Fix:** Add at least one automated smoke test exercising the script end-to-end against the optest static build it targets.

fips-hash-offline.sh
```diff
@@ -0,0 +1,170 @@
...
+            echo $((o + v - a))
+            return 0
+        fi
+    done < <(readelf -SW "$BIN" | sed 's/\[[ 0-9]*\]//' | awk '/PROGBITS/{print $1, $2, $3, $4, $5}')
+    return 1
+}
+
+# Extract byte_count bytes starting at file_offset (both in bytes) to stdout.
+# Uses GNU dd byte-granular skip/count with a large block size for speed.
+extract() {
```
dgarske, line 90: **🔵 [Low] Relies on GNU dd byte-granular flags; command -v check does not detect them**

The script depends on GNU coreutils `dd` (`iflag=skip_bytes,count_bytes`, `oflag=seek_bytes`) and bash process substitution `< <(...)`. The preflight only runs `command -v dd`, which succeeds on BSD/macOS where those iflag/oflag options do not exist. Combined with `2>/dev/null`, a non-GNU dd would produce empty output rather than a clear diagnostic (it would eventually fail the length check, but with a confusing message). This is acceptable given the documented Linux/optest scope, but a clearer guard or a documented requirement would help.

**Fix:** Add a one-time probe that GNU-style dd flags are supported, or state the GNU coreutils requirement in the header comment.

fips-hash-offline.sh
```diff
@@ -0,0 +1,170 @@
...
+command -v awk     >/dev/null 2>&1 || die "awk not found"
+command -v dd      >/dev/null 2>&1 || die "dd not found"
+command -v readelf >/dev/null 2>&1 || die "readelf not found"
+command -v openssl >/dev/null 2>&1 || die "openssl not found"
+command -v xxd     >/dev/null 2>&1 || die "xxd not found"
+
+# Symbol lookup from the ELF symbol table.
+# readelf -s columns: Num: Value Size Type Bind Vis Ndx Name
+#   -> Value ($2) is hexadecimal, Size ($3) is decimal, Ndx ($7), Name ($8).
+sym_val()  { readelf -sW "$BIN" | awk -v n="$1" '$8==n && $7!="UND"{print $2; exit}'; }
```
dgarske: **🔵 [Low] readelf re-invoked many times over the same binary**

`sym_val` and `sym_size` each run `readelf -sW` (~10 invocations for 8 symbols across two accessors), and `vaddr_to_off` runs `readelf -SW` on every call (~5+ times). The symbol table and section table could each be captured once into a variable and reused. Purely a minor efficiency/readability point for a build helper, not a correctness issue.

**Fix:** Cache the readelf symbol/section output once and reuse; optional cleanup.

---
> **Note:** Referenced line (`fips-hash-offline.sh:67-84,93-100,125-157`) is outside the diff hunk. Comment anchored to nearest changed region.

dgarske (commented):
## Skoll Code Review

### Posted findings
- [Medium] **Hash pipeline can silently emit a wrong-but-correct-length hash (no pipefail, dd errors hidden)** — `fips-hash-offline.sh:141-163`
- [Medium] **Offline hash is never validated against the module's own self-test** — `fips-hash-offline.sh:41-46,161-170`
- [Medium] **New 170-line FIPS helper script has no automated test** — `fips-hash-offline.sh:1-170`
- [Low] **Relies on GNU dd byte-granular flags; command -v check does not detect them** — `fips-hash-offline.sh:58-62,88-90,158`
- [Low] **readelf re-invoked many times over the same binary** — `fips-hash-offline.sh:67-84,93-100,125-157`

fips-hash-offline.sh
```diff
@@ -0,0 +1,170 @@
...
+#   - single fixed-address (ET_EXEC) image, not stripped
+#   - measured .text/.rodata bracketed by the wolfCrypt_FIPS_first/last and
+#     wolfCrypt_FIPS_ro_start/ro_end symbols
+#   - no text-segment canonicalizer and no relocation-table indirection in
+#     effect (i.e. the plain #else path of DoInCoreCheck())
+#
+# Usage: fips-hash-offline.sh [path-to-binary]
+#   default binary: wolfcrypt/test/testwolfcrypt
+
+set -u
```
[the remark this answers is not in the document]
holtrop-wolfssl: fixed

fips-hash-offline.sh
```diff
@@ -0,0 +1,170 @@
...
+        fi
+    done < <(readelf -SW "$BIN" | sed 's/\[[ 0-9]*\]//' | awk '/PROGBITS/{print $1, $2, $3, $4, $5}')
+    return 1
+}
+
+# Extract byte_count bytes starting at file_offset (both in bytes) to stdout.
+# Uses GNU dd byte-granular skip/count with a large block size for speed.
+extract() {
+    dd if="$BIN" bs=1M iflag=skip_bytes,count_bytes skip="$1" count="$2" 2>/dev/null
+}
```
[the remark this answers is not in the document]
holtrop-wolfssl: fixed

fips-hash-offline.sh
holtrop-wolfssl: added set -o pipefail

fips-hash-offline.sh
holtrop-wolfssl: out of scope for this PR

fips-hash-offline.sh
holtrop-wolfssl: Changed to only use POSIX-compatible dd flags

fips-hash-offline.sh
holtrop-wolfssl: This is not worth changing.

fips-hash-offline.sh
holtrop-wolfssl: out of scope for this PR

fips-hash-offline.sh
```diff
@@ -0,0 +1,170 @@
...
+# Usage: fips-hash-offline.sh [path-to-binary]
+#   default binary: wolfcrypt/test/testwolfcrypt
+
+set -u
+
+BIN="${1:-wolfcrypt/test/testwolfcrypt}"
+
+die() { echo "fips-hash-offline: error: $*" >&2; exit 1; }
+
+[ -f "$BIN" ] || die "binary not found: $BIN"
```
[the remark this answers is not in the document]
holtrop-wolfssl: fixed

fips-hash-offline.sh
```diff
@@ -0,0 +1,172 @@
...
+command -v dd      >/dev/null 2>&1 || die "dd not found"
+command -v readelf >/dev/null 2>&1 || die "readelf not found"
+command -v openssl >/dev/null 2>&1 || die "openssl not found"
+command -v xxd     >/dev/null 2>&1 || die "xxd not found"
+
+# Symbol lookup from the ELF symbol table.
+# readelf -s columns: Num: Value Size Type Bind Vis Ndx Name
+#   -> Value ($2) is hexadecimal, Size ($3) is decimal, Ndx ($7), Name ($8).
+sym_val()  { readelf -sW "$BIN" | awk -v n="$1" '$8==n && $7!="UND"{print $2; exit}'; }
+sym_size() { readelf -sW "$BIN" | awk -v n="$1" '$8==n && $7!="UND"{print $3; exit}'; }
```
douzzer: These aren't valid portable `awk` scripts, and independent of that, they crash the script with `SIGPIPE`.  This worked for me:
```
sym_val()  { readelf -sW "$BIN" | awk -v n="$1" '{if (!printed && $8==n && $7!="UND"){print $2; printed=1; }}'; }
sym_size() { readelf -sW "$BIN" | awk -v n="$1" '{if (!printed && $8==n && $7!="UND"){print $3; printed=1; }}'; }
```

douzzer (changes requested):
This is great!  Worked for me on `src/.libs/libwolfssl.so.45.0.0` once I fixed `sym_val()` and `sym_size()`.

fips-hash-offline.sh
holtrop-wolfssl: Thanks, copied those in.

douzzer (approved):

PR #10755, opened 2026-06-23
Reject certificates with trailing data.  Avoid copying trailing data to WOLFSSL_X509 structs.

kareem-wolfssl:
# Description

Fixes zd#22010

# Testing

Provided reproducer, built-in tests

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


kareem-wolfssl:
Retest this please

tests/api.c
```diff
@@ -11894,6 +11894,10 @@ static int test_wc_CheckCertSigPubKey(void)
     ExpectNotNull(cert_der = (byte*)malloc(cert_dersz));
     ExpectIntGE(ret = wc_CertPemToDer(cert_buf, (int)cert_sz, cert_der,
         (int)cert_dersz, CERT_TYPE), 0);
+    /* Use the actual DER length, not the (larger) PEM buffer size, otherwise
```
dgarske, line 12829: **🟠 [Medium] No positive test that a certificate with trailing data is rejected** · _test_

The only test change adjusts an existing test to stop passing the (larger) PEM buffer size as the DER length so it no longer trips the new check. There is no new test that exercises the actual new behavior: appending arbitrary bytes after a certificate's outer SEQUENCE and asserting the parse now fails with ASN_PARSE_E (and that it is accepted under WOLFSSL_NO_ASN_STRICT). There is also no test for the derCert-bounding fix in CopyDecodedToX509 (i.e. that wolfSSL_i2d_X509 / wolfSSL_X509_get_der no longer re-emit trailing bytes on a non-strict build). The PR checklist has 'added tests' unchecked. review-security independently flagged missing regression coverage for the OCSP multi-cert and d2i-trailing-bytes paths.

**Fix:** Add a positive test: take a known-good DER cert, append N junk bytes, and assert wc_ParseCert(...) returns ASN_PARSE_E in a default build. Add a companion assertion that under WOLFSSL_NO_ASN_STRICT the parse succeeds but wolfSSL_i2d_X509/wolfSSL_X509_get_der return only the certificate length (srcIdx), proving the derCert bounding. Optionally cover the TRUSTED_CERT_TYPE (trailing aux trust data allowed and trimmed) and OCSP multi-cert paths.

wolfcrypt/src/asn.c
```diff
@@ -22229,6 +22229,35 @@ static int DecodeCertInternal(DecodedCert* cert, int verify, int* criticalExt,
         }
     }
 
+#ifndef WOLFSSL_NO_ASN_STRICT
```
dgarske: **🔵 [Low] Strict trailing-data rejection absent in non-WOLFSSL_ASN_TEMPLATE builds (incomplete hardening)** · _Security_

The trailing-data rejection is placed inside DecodeCertInternal, which is compiled only under #ifdef WOLFSSL_ASN_TEMPLATE. The legacy (non-template) ParseCertRelative path (DecodeToKey + GetSigAlg + GetSignature, asn.c:23498+) has no equivalent srcIdx == maxIdx check, so builds without the ASN template still silently accept certificates with arbitrary appended bytes at parse time. The CopyDecodedToX509 / CopyDecodedAcertToX509 srcIdx bounding does prevent that trailing data from being re-emitted by i2d/get_der/digest on both build types, so the storage/canonicalization risk is mitigated; however, the parse-time rejection that is the primary point of the PR does not apply to non-template builds. WOLFSSL_ASN_TEMPLATE is the default, so most builds are covered; this is a defense-in-depth / configuration-asymmetry gap rather than an exploitable flaw.

**Fix:** If parity is desired, add an equivalent end-of-certificate check to the non-template ParseCertRelative path (after GetSignature, verify cert->srcIdx == cert->maxIdx unless allowTrailing/isCSR). Otherwise explicitly document that trailing-data rejection requires WOLFSSL_ASN_TEMPLATE, with the derCert bounding as the fallback protection for legacy-parser builds.

wolfcrypt/src/asn.c
```diff
@@ -22229,6 +22229,35 @@ static int DecodeCertInternal(DecodedCert* cert, int verify, int* criticalExt,
...
+     * wolfSSL_X509_digest (which operate on the stored source buffer of length
+     * maxIdx).
+     *
+     * cert->srcIdx points just past the certificate's outer SEQUENCE after the
+     * x509CertASN parse above. Only enforce this on a full parse; the
+     * pubkey-only paths (stopAtPubKey/stopAfterPubKey) intentionally parse the
+     * whole template but their callers may pass larger buffers. The TRUSTED
+     * CERTIFICATE format legitimately carries auxiliary trust data after the
+     * certificate, so allow it when cert->allowTrailing is set. */
+    if ((ret == 0) && (!done) && (!stopAtPubKey) && (!stopAfterPubKey) &&
```
dgarske: **🔵 [Low] Redundant stopAtPubKey/stopAfterPubKey conditions in trailing-data guard** · _style_

The guard tests `(!done) && (!stopAtPubKey) && (!stopAfterPubKey)`. Whenever ret == 0 and either stopAtPubKey or stopAfterPubKey is set, `done` is already 1 (set at lines 22066 and 22155), so `(!done)` already implies both of the other two conditions. The extra terms are harmless and arguably self-documenting, but strictly redundant.

**Fix:** Optionally drop the redundant `(!stopAtPubKey) && (!stopAfterPubKey)` terms, or keep them purely for documentation. No functional impact.

wolfssl/wolfcrypt/asn.h
```diff
@@ -2072,6 +2072,10 @@ struct DecodedCert {
 
     /* Option Bits */
     WC_BITFIELD subjectCNStored:1;      /* have we saved a copy we own */
+    WC_BITFIELD allowTrailing:1;        /* permit data after the cert's outer
```
dgarske: **🔵 [Low] New bitfield inserted mid-struct in DecodedCert rather than appended** · _convention_

`allowTrailing:1` is inserted between the existing `subjectCNStored` and `extSubjKeyIdSet` bitfields. Per the project's C style guidance, new struct members should generally be added at the end to avoid changing layout/offsets of subsequent members. In practice this is a packed bitfield added within the existing 'Option Bits' cluster (following the local grouping convention and, since it fits the existing storage unit, unlikely to change sizeof or shift non-bitfield members), so ABI impact is minimal. Flagged only for consistency with the append-at-end guidance.

**Fix:** Acceptable as-is given it joins the existing option-bit group; if strict append-at-end is preferred, move the new bitfield to the end of the option-bit run.

dgarske (commented):
## Skoll Multi-Scan Review

### Posted findings
- [Medium] [review] **No positive test that a certificate with trailing data is rejected** — `tests/api.c:11897`
- [Low] [review+review-security] **Strict trailing-data rejection absent in non-WOLFSSL_ASN_TEMPLATE builds (incomplete hardening)** — `wolfcrypt/src/asn.c:22232-22259`
- [Low] [review] **Redundant stopAtPubKey/stopAfterPubKey conditions in trailing-data guard** — `wolfcrypt/src/asn.c:22249`
- [Low] [review] **New bitfield inserted mid-struct in DecodedCert rather than appended** — `wolfssl/wolfcrypt/asn.h:2075`

#### OCSP responses embedding a certificate chain (1 cert) are now rejected, breaking OCSP validation
**File:** `wolfcrypt/src/asn.c:22249 (manifests in OcspCheckCert, asn.c:35008-35014)`
**Function:** `DecodeCertInternal / OcspCheckCert`
**Severity:** Medium
**Category:** Regression

The new trailing-data check `if ((ret == 0) && (!done) && (!stopAtPubKey) && (!stopAfterPubKey) && (!cert->allowTrailing) && (cert->srcIdx != cert->maxIdx)...) ret = ASN_PARSE_E;` runs on every full `ParseCertRelative(CERT_TYPE)`. In `OcspCheckCert` the embedded responder certificate is parsed via `InitDecodedCert(cert, resp->cert, resp->certSz, heap)`, where `resp->cert` points at the first Certificate but `resp->certSz` spans the ENTIRE `SEQUENCE OF Certificate` (from GetASN_GetRef on OCSPBASICRESPASN_IDX_CERTS_SEQ). RFC 6960 permits the certs field to carry a chain, and the code comment at asn.c:35225 confirms wolfSSL intentionally parses only the first embedded cert ('TODO: support more than one certificate'). BEFORE this PR the trailing certs were ignored and OCSP validation succeeded; AFTER, cert->srcIdx (end of first cert) != cert->maxIdx (end of all certs) -> ASN_PARSE_E, so OcspCheckCert/OcspResponseDecode fail. This is a functional/availability regression for delegated OCSP responders that include a cert chain, and OCSP hard-fail configs will drop otherwise-valid handshakes. `allowTrailing` is only set for TRUSTED_CERT_TYPE, so this CERT_TYPE parse is not exempted.

**Recommendation:** Exempt the OCSP responder-cert parse from the trailing-data check (e.g. set cert->allowTrailing = 1 before ParseCertRelative in OcspCheckCert, since only the first cert is intentionally parsed), OR size the DecodedCert to exactly the first Certificate (parse the leading SEQUENCE length and pass that as inSz). Add a regression test with a multi-certificate OCSP certs field.

_Referenced code: `wolfcrypt/src/asn.c:22249 (manifests in OcspCheckCert, asn.c:35008-35014)` (3 lines)_

---

#### wolfSSL_d2i_X509 / d2i_X509_REQ now reject DER buffers with trailing data (OpenSSL d2i contract change)
**File:** `wolfcrypt/src/asn.c:22249 (d2i path: src/x509.c:4228)`
**Function:** `DecodeCertInternal / d2i_X509orX509REQ`
**Severity:** Medium
**Category:** Regression

The new strict trailing-data rejection in DecodeCertInternal fires whenever a full parse ends with cert->srcIdx != cert->maxIdx, i.e. whenever the buffer is longer than one encoded certificate. This reaches the OpenSSL-compat path wolfSSL_d2i_X509 -> d2i_X509orX509REQ, which calls ParseCertRelative(cert, CERT_TYPE, ...) with len = the full remaining buffer. OpenSSL's d2i_X509 is a streaming decoder: it parses one object and advances *pp past the consumed bytes; the wrapper even does `*in += newX509->derCert->length;` (src/x509.c:4201-4203), clearly intended to allow walking a concatenation of DER objects. With this PR any multi-object or over-sized buffer now returns ASN_PARSE_E and d2i_X509/d2i_X509_REQ return NULL, so the pointer-advance iteration path is effectively unreachable in a default (strict) build. Severity views differ (review: Medium/api-compat; review-security: Low/regression); the stricter Medium is retained. Note the paired CopyDecodedToX509 bounding would make the pointer-advance idiom correct under WOLFSSL_NO_ASN_STRICT, but in strict builds the parse fails first. This appears intended by the security fix (zd#22010), but its interaction with the *in += derCert->length iteration idiom should be a conscious, documented decision.

**Recommendation:** Decide whether strict rejection is intended for the OpenSSL-compat d2i surface (wolfSSL_d2i_X509 / wolfSSL_X509_d2i / wolfSSL_X509_load_certificate_buffer DER). If OpenSSL streaming semantics must be preserved, trim the DecodedCert to the leading SEQUENCE length before parsing (or set cert->allowTrailing) so a single cert is parsed and the pointer advanced by its exact length, while still rejecting trailing data on trust-load paths. If strict rejection is deliberate here, document the behavior change in the API/ChangeLog for OPENSSL_EXTRA users and add a test for a cert-plus-trailing-bytes buffer.

_Referenced code: `wolfcrypt/src/asn.c:22249 (d2i path: src/x509.c:4228)` (5 lines)_

---

### Skipped findings
- [Info] ~~New trailing-data rejection changes wc_ParseCert public-API contract for over-long buffers~~

dgarske (changes requested):
https://github.com/wolfSSL/wolfssl/pull/10755#pullrequestreview-4656987604

SparkiDev (dismissed):

src/internal.c
```diff
@@ -14730,9 +14730,19 @@ int CopyDecodedToX509(WOLFSSL_X509* x509, DecodedCert* dCert)
...
-                                                                         == 0) {
-            XMEMCPY(x509->derCert->buffer, dCert->source, dCert->maxIdx);
+        /* Store only the certificate itself, bounded by the end of its outer
+         * SEQUENCE (dCert->srcIdx), never any trailing bytes that may follow in
+         * the source buffer. This keeps wolfSSL_i2d_X509 / wolfSSL_X509_get_der
+         * / wolfSSL_X509_digest canonical - they operate on derCert - even on
+         * builds/paths that do not reject trailing data (e.g.
+         * WOLFSSL_NO_ASN_STRICT). It removes only bytes after the certificate,
+         * so the signed certificate bytes are preserved verbatim. */
+        word32 derCertSz = dCert->maxIdx;
```
dgarske: **🔴 [High] CopyDecodedToX509() can now store a truncated derCert when ParseCertRelative() failed (non-template builds)**

The new bound `derCertSz = min(dCert->maxIdx, dCert->srcIdx)` assumes `srcIdx` always points at the end of the certificate. That holds only for a *successful* parse. In `--enable-asn=original` builds (exercised by .github/workflows/os-check.yml:149,151) `GetCertHeader()` (wolfcrypt/src/asn_orig.c:1535-1545) rebinds `cert->maxIdx` to the end of the certificate's outer SEQUENCE *early*, while `cert->srcIdx` advances incrementally through the rest of `ParseCertRelative()`. If parsing aborts part way -- e.g. `DecodeCertExtensions()` returns an error other than `ASN_CRIT_EXT_E` (wolfcrypt/src/asn.c:24018-24023) or `GetSigAlg()` returns `ASN_SIG_OID_E` (wolfcrypt/src/asn.c:24038-24046) -- `srcIdx` is left mid-certificate with `0 < srcIdx < maxIdx`. `src/ssl_p7p12.c:2028-2033` (`wolfSSL_PKCS12_parse()`) calls `CopyDecodedToX509()` *unconditionally* after a failed `ParseCertRelative()`: it only logs `WOLFSSL_MSG("Issue with parsing certificate")` with no `else`. Before this PR the full certificate (`dCert->maxIdx` bytes) was stored; now `x509->derCert` is silently truncated mid-certificate, so `wolfSSL_i2d_X509()` / `wolfSSL_X509_get_der()` / `wolfSSL_X509_digest()` emit an unparseable DER fragment while `wolfSSL_PKCS12_parse()` still returns success. In the default template build `GetASN_Items()` only writes `*inOutIdx` on success (wolfcrypt/src/asn.c, tail of `GetASN_Items`), so `srcIdx` is either 0 or the end of the cert and the bug does not appear -- which is why it is easy to miss.

**Fix:** Do not derive the DER bound from the mutable `srcIdx`. Either (a) add an explicit `certEnd` field to `DecodedCert` that is only set once the certificate has been fully parsed, or (b) skip the bounding entirely unless the parse succeeded. Separately, `src/ssl_p7p12.c:2030` should stop on the parse failure rather than fall through to `CopyDecodedToX509()`. Add a regression test under `--enable-asn=original` that copies a partially-parsed DecodedCert and asserts `derCert->length` equals the full certificate length.

wolfcrypt/src/asn.c
```diff
@@ -23761,13 +23791,30 @@ int ParseCertRelative(DecodedCert* cert, int type, int verify, void* cm,
         return BAD_FUNC_ARG;
     }
 
+    /* TRUSTED CERTIFICATE blobs (RFC/OpenSSL "TRUSTED CERTIFICATE") carry
```
dgarske, line 24181: **🟠 [Medium] New TRUSTED_CERT_TYPE / allowTrailing branch in ParseCertRelative() has no test coverage**

The `if (type == TRUSTED_CERT_TYPE) { cert->allowTrailing = 1; type = CERT_TYPE; }` branch is the only thing keeping `PEM_read_bio_X509_AUX()` working now that trailing data is rejected by default, yet nothing in the test suite exercises it with a real TRUSTED CERTIFICATE blob. The two existing `PEM_read_bio_X509_AUX()` tests (tests/api.c:13480 and tests/api/test_ossl_pem.c:976-979) both feed a plain "CERTIFICATE" PEM (`svrCertFile` / `cliCertFile`) with no auxiliary trust info, so they would still pass even if the exemption were removed. A suitable fixture already exists and is used elsewhere in api.c: `./certs/test/ossl-trusted-cert.pem` (tests/api.c:11616). The new `test_cert_trailing_data()` does not cover this path either.

**Fix:** Add a strict-build test that loads `./certs/test/ossl-trusted-cert.pem` via `PEM_read_bio_X509_AUX()` (and/or `wc_ParseCert(..., TRUSTED_CERT_TYPE, ...)`) and asserts both that the parse succeeds and that `wolfSSL_X509_get_der()` returns only the certificate portion.

src/x509.c
```diff
@@ -6183,17 +6183,18 @@ static WOLFSSL_X509* loadX509orX509REQFromBuffer(
...
             InitDecodedCert(cert, der->buffer, der->length, NULL);
-            ret = ParseCertRelative(cert, parseType, 0, NULL, NULL);
+            /* For TRUSTED_CERT_TYPE the DER buffer holds the certificate
+             * followed by auxiliary trust info. ParseCertRelative() recognizes
+             * the type: it parses only the certificate, permits the trailing
+             * aux data, and treats it as CERT_TYPE for verification. The DER is
+             * trimmed to the certificate below. */
+            ret = ParseCertRelative(cert, type, 0, NULL, NULL);
             if (ret == 0) {
                 /* For TRUSTED_CERT_TYPE, truncate the DER buffer to exclude
```
dgarske: **🔵 [Low] TRUSTED_CERT_TYPE maxIdx truncation in loadX509orX509REQFromBuffer() is now dead code**

`CopyDecodedToX509()` now computes `min(dCert->maxIdx, dCert->srcIdx)` itself (src/internal.c:14740-14743), so the explicit `cert->maxIdx = cert->srcIdx;` here can no longer change the outcome: it is the same clamp applied one call earlier. The result is two places encoding the same rule, and a reader has to check both to know which one governs. This is also the one caller that mutates a `DecodedCert` field after parsing purely to influence a downstream copy, which is fragile.

**Fix:** Delete the `cert->maxIdx = cert->srcIdx;` block and keep the single bounding rule inside `CopyDecodedToX509()`, or keep this one and drop the duplication -- but not both.

wolfcrypt/src/asn.c
```diff
@@ -22541,6 +22541,31 @@ static int DecodeCertInternal(DecodedCert* cert, int verify, int* criticalExt,
         }
     }
 
+#ifndef WOLFSSL_NO_ASN_STRICT
```
dgarske: **🔵 [Low] Template trailing-data check runs after full extension decoding**

`cert->srcIdx` is final immediately after the `GetASN_Items(x509CertASN, ...)` call at line 22300, but the trailing check is placed at line 22561 -- after `GetCertName()` for issuer and subject, `GetCertKey()`, and the full `DecodeCertExtensions()` pass. Input that is going to be rejected anyway is therefore fully parsed first, which costs work on attacker-supplied data and makes the check's dependency on `srcIdx` less obvious than it needs to be. It also means the check has to guard on `!done`, which would be unnecessary if it ran before the `stopAtPubKey`/`stopAfterPubKey` early exits diverge.

**Fix:** Consider moving the check up to just after the `x509CertASN` parse (still skipping it for `stopAtPubKey`/`stopAfterPubKey`) so malformed input is rejected before the expensive name/key/extension decoding.

tests/api.c
```diff
@@ -12223,6 +12223,10 @@ static int test_wc_CheckCertSigPubKey(void)
     ExpectNotNull(cert_der = (byte*)malloc(cert_dersz));
     ExpectIntGE(ret = wc_CertPemToDer(cert_buf, (int)cert_sz, cert_der,
```
dgarske: **🔵 [Low] test_wc_CheckCertSigPubKey: ExpectIntGE allows a zero-length DER to silently keep the PEM size**

The fix guards the assignment with `if (ret > 0)`, but the assertion above it is `ExpectIntGE(ret = wc_CertPemToDer(...), 0)` -- so `ret == 0` is treated as success while leaving `cert_dersz` at the (much larger) PEM buffer size. In that case the subsequent `wc_ParseCert()` now fails with `ASN_PARSE_E` for a confusing reason rather than reporting the real problem (empty conversion). The same applies if an earlier `Expect*` already failed: `ExpectIntGE` short-circuits, `ret` stays 0, and `cert_dersz` keeps the PEM size. Tightening the assertion makes the intent explicit and the failure message accurate.

**Fix:** Change `ExpectIntGE(..., 0)` to `ExpectIntGT(..., 0)` so a zero-length conversion fails at the conversion step rather than surfacing later as a trailing-data parse error.

src/internal.c
dgarske: **🔵 [Low] Duplicated DER-bounding block in CopyDecodedToX509() and CopyDecodedAcertToX509()**

The two blocks are byte-for-byte identical apart from the `dCert`/`dAcert` variable name, including the alloc/copy/`MEMORY_E` handling. `DecodedCert` and `DecodedAcert` do not share a base type, but a tiny static helper taking `(DerBuffer** out, const byte* src, word32 maxIdx, word32 srcIdx, void* heap)` would keep the rule in one place so a future fix (see the truncation finding above) does not have to be applied twice.

**Fix:** Factor the shared bounding + AllocDer + XMEMCPY into one static helper in src/internal.c and call it from both places.

---
> **Note:** Referenced line (`src/internal.c:14740-14749, 14949-14958`) is outside the diff hunk. Comment anchored to nearest changed region.

dgarske (commented):
## Skoll Code Review

**Findings:** 11 total — 8 posted, 3 skipped

### Posted findings
- [High] **CopyDecodedToX509() can now store a truncated derCert when ParseCertRelative() failed (non-template builds)** — `src/internal.c:14740-14746`
- [Medium] **New TRUSTED_CERT_TYPE / allowTrailing branch in ParseCertRelative() has no test coverage** — `wolfcrypt/src/asn.c:23794-23803`
- [Low] **TRUSTED_CERT_TYPE maxIdx truncation in loadX509orX509REQFromBuffer() is now dead code** — `src/x509.c:6199-6204`
- [Low] **Duplicated DER-bounding block in CopyDecodedToX509() and CopyDecodedAcertToX509()** — `src/internal.c:14740-14749, 14949-14958`
- [Low] **Template trailing-data check runs after full extension decoding** — `wolfcrypt/src/asn.c:22544-22567`
- [Low] **test_wc_CheckCertSigPubKey: ExpectIntGE allows a zero-length DER to silently keep the PEM size** — `tests/api.c:12224-12229`

#### Raw Public Key certificates bypass both new trailing-data checks
**File:** `wolfcrypt/src/asn.c:22273-22277, 23832-23836`
**Function:** `DecodeCertInternal / ParseCertRelative`
**Severity:** Medium

Both new checks are unreachable when `HAVE_RPK` is enabled and the input is recognised as an RFC 7250 Raw Public Key. In the template path `DecodeCertInternal()` returns at wolfcrypt/src/asn.c:22273-22274 (`if (ret == 0) { return ret; }`) immediately after the RPK `GetASN_Items()`, well before the trailing check at 22561. In the non-template path `ParseCertRelative()` returns at 23833-23835 (`if (cert->isRPK) { return ret; }`) before the check at 24071. So an RPK blob with arbitrary appended bytes is still accepted on a strict build, while the equivalent X.509 input is rejected -- an inconsistency in what the PR advertises as "reject certificates with trailing data". Because `CopyDecodedToX509()` now also trims `derCert` to `srcIdx` for RPK peers (src/internal.c:15373, 15383), `ssl->peerCert.derCert` and the raw wire buffer used by `RpkIsTrusted()` (src/internal.c:17810-17812) can diverge for such input.

**Recommendation:** Either extend the trailing-data check to the RPK early-return paths, or add a comment at both RPK return sites explaining why RPK is deliberately exempt so a future reader does not assume the check is universal.

_Referenced code: `wolfcrypt/src/asn.c:22273-22277, 23832-23836` (14 lines)_

---

#### New test function name does not follow the api.c naming convention
**File:** `tests/api.c:12276, 37114`
**Function:** `test_cert_trailing_data`
**Severity:** Low

Test functions in tests/api.c are named after the API under test -- `test_wc_*` for wolfCrypt entry points and `test_wolfSSL_*` for the SSL/compat layer (e.g. the immediately adjacent `test_wc_CheckCertSigPubKey` and `test_wolfSSL_X509_ext_d2i`). `test_cert_trailing_data` matches neither, which makes it harder to locate when scanning the `testCases[]` table. Also, the `XMEMSET(&decoded, 0, sizeof(decoded));` calls before each `wc_InitDecodedCert()` are redundant -- `InitDecodedCert_ex()` already zeroes the whole struct.

**Recommendation:** Rename to `test_wc_ParseCert_trailing_data` (or similar `test_wc_*` form) and drop the redundant `XMEMSET()` calls.

_Referenced code: `tests/api.c:12276, 37114` (6 lines)_

---

### Skipped findings
- [High] ~~d2i_X509()/wolfSSL_X509_load_certificate_buffer() now reject any buffer longer than the certificate, breaking OpenSSL compat contract~~
- [Medium] ~~Behavior/compat change to public wc_ParseCert()/ParseCertRelative() is undocumented~~
- [Medium] ~~derCert bounding and the strict-rejection paths are only partially covered by the new test~~

dgarske (changes requested):
https://github.com/wolfSSL/wolfssl/pull/10755#pullrequestreview-4849047163

kareem-wolfssl:
Retest this please

dgarske (approved):

JacobBarthelmeh (approved):

PR #10758, opened 2026-06-23
Bound waited example servers with timeout to fix flaky CI test hangs

julek-wolfssl:
### Problem

Several test scripts background the example server and then `wait` for it to
exit with no timeout. When the server occasionally fails to exit — a timing
race seen under heavy parallel CI load — the script blocks until the CI job's
`timeout-minutes`, cancelling the whole `trackmemory` run. This was seen
consistently on the `all-wolfentropy` config via
`ocsp-stapling_tls13multi.test` (cases 6 and 7).

### Fix

Wrap the waited example servers in `timeout -s KILL 2m` (the same pattern
`scripts/dtls.test` already uses), so a stuck server is killed and the test
fails fast instead of hanging the entire job.

Scripts updated:

- `ocsp-stapling_tls13multi.test`
- `ocsp-stapling.test`
- `ocsp-stapling2.test`
- `ocsp-stapling-with-wolfssl-responder.test`
- `crl-revoked.test`
- `tls13.test`
- `resume.test`
- `pkcallbacks.test`
- `dtlscid.test`

### DIFF
--- a/scripts/crl-revoked.test
+++ b/scripts/crl-revoked.test
@@ -1,5 +1,12 @@
 #!/usr/bin/env bash
 
+# timeout(1) is GNU coreutils and absent on macOS; where it's missing, run the
+# command unbounded (the flaky hang this guards against is Linux-only CI).
+if ! command -v timeout >/dev/null 2>&1; then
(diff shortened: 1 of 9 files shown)


julek-wolfssl:
retest this please flaky test

Frauschi (approved):

douzzer (approved):

PR #10759, opened 2026-06-23
PKCS#7: add ML-DSA support

Frauschi:
### Summary

Add ML-DSA as a signature algorithm for PKCS#7/CMS `SignedData`, for both signing and verification, following RFC 9882 ("Use of ML-DSA in the Cryptographic Message Syntax").

Unlike RSA/ECDSA, ML-DSA is used in CMS "pure" mode: the signature is computed over the complete message - the `DER SET OF` signed attributes, or the `eContent` directly when no signed attributes are present - rather than over a pre-computed `DigestInfo`, with an empty `context` string and absent `signatureAlgorithm` parameters.

### Changes

`wolfcrypt/src/pkcs7.c` - core implementation
- New helpers `wc_PKCS7_MlDsaLevelFromOID`, `wc_PKCS7_BuildPureSigMessage`, `wc_PKCS7_MlDsaSign`, `wc_PKCS7_MlDsaVerify`, wired into the existing per-algorithm switch sites (`GetSignSize`, `SignedDataGetEncAlgoId`, `SetPublicKeyOID`, `CheckPublicKeyDer`) and the sign/verify dispatchers.
- `wc_MlDsaKey` is always heap-allocated (due to its size), matching the ML-DSA handling in `asn.c`; the accompanying `DecodedCert` uses the `WC_DECLARE_VAR`/`WC_ALLOC_VAR_EX` macros for `WOLFSSL_SMALL_STACK` handling.
- Signer public key is cached in `wc_PKCS7` only for RSA/ECC (consumed by the raw-sign callback paths), with the buffer-size bound retained; large PQC keys are not stored.

`wolfcrypt/src/hash.c` - map the `SHAKE128`/`SHAKE256` OIDs to their hash types in `wc_OidGetHash()` (RFC 8702 digest algorithms).

`certs/mldsa/ `- add expanded-only PKCS#8 DER private keys (mldsa44/65/87-key.der) matching the existing self-signed ML-DSA certificates.

`wolfcrypt/test/test.c` - `pkcs7signed_mldsa_test()`: round-trip encode→verify across every enabled ML-DSA level with SHA-512 and (where available) SHAKE digests, asserting:
- recovered content matches,
- SHAKE digest AlgorithmIdentifiers are encoded with absent parameters per RFC 8702 (both the digestAlgorithms SET and SignerInfo.digestAlgorithm occurrences),
- a tampered signature is rejected with `SIG_VERIFY_E` (negative path).

### Configuration

Gated behind existing ML-DSA macros - no new public config option. Active when `WOLFSSL_HAVE_MLDSA` is enabled with ASN.1 support; signing and verification compile independently based on the ML-DSA backend's private-/public-key availability (`WOLFSSL_MLDSA_NO_SIGN` / `WOLFSSL_MLDSA_NO_VERIFY`). No change to default builds.

### Compatibility

- No public API changes; no changes to existing struct members.
- RSA/ECDSA/RSA-PSS SignedData paths are unchanged.

### Testing

- `./configure --enable-all --enable-debug && make && ./wolfcrypt/test/testwolfcrypt` - `PKCS7signed test passed!`, full suite green.
- Verified clean compilation under degenerate ML-DSA feature combinations (`NO_SIGN`, `NO_VERIFY`, both, and `NO_MAKE_KEY`+`NO_SIGN`+`NO_VERIFY`) with the project's `-Werror` flag set.


wolfcrypt/src/pkcs7.c
```diff
@@ -2596,6 +2734,173 @@ static int wc_PKCS7_BuildDigestInfo(wc_PKCS7* pkcs7, byte* flatSignedAttribs,
...
+        }
+
+        XMEMCPY(msg, attribSet, attribSetSz);
+        XMEMCPY(msg + attribSetSz, attribs, attribsSz);
+
+        *outMsg = msg;
+        *outMsgSz = attribSetSz + attribsSz;
+        *outAlloc = 1;
+    }
+    else {
```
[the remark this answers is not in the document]
Frauschi: fixed

wolfcrypt/src/pkcs7.c
```diff
@@ -1033,6 +1048,36 @@ static int wc_PKCS7_RecipientListVersionsAllZero(wc_PKCS7* pkcs7)
     return 1;
 }
 
+#if defined(WOLFSSL_MLDSA_PUBLIC_KEY) || defined(WC_PKCS7_MLDSA_SIGN)
```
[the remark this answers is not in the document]
Frauschi: fixed

dgarske (commented):
## Skoll Multi-Scan Review

### Posted findings
- [High] [review] **ML-DSA verify references wc_PKCS7_CertMatchesSignerInfo which is not compiled in RSA-less/ECC-less builds** — `wolfcrypt/src/pkcs7.c:5396`
- [Low] [review+review-security] **RSA raw-sign callback output buffer tightened to modulus size with no post-write bounds check** — `wolfcrypt/src/pkcs7.c:2973-2978`
- [Low] [review] **InitWithCert no longer size-checks non-RSA/ECC public keys** — `wolfcrypt/src/pkcs7.c:1315-1340`
- [Info] [review-security] **wc_OidGetHash SHAKE128/SHAKE256 mapping affects OCSP/CRL/PKCS12/RSA-PSS beyond the stated PKCS#7 scope** — `wolfcrypt/src/hash.c:356-369`
- [Info] [review] **ML-DSA verify masks hard internal errors from key decode as SIG_VERIFY_E** — `wolfcrypt/src/pkcs7.c:5416-5449`

wolfcrypt/src/pkcs7.c
```diff
@@ -4908,6 +5319,145 @@ static int wc_PKCS7_EcdsaVerify(wc_PKCS7* pkcs7, byte* sig, int sigSz,
...
+        if (ret < 0) {
+            WOLFSSL_MSG("ASN ML-DSA cert parse error");
+            FreeDecodedCert(dCert);
+            wc_MlDsaKey_Free(key);
+            continue;
+        }
+
+        /* Only try the certificate identified by the SignerInfo sid. */
+        if (pkcs7->signerInfo != NULL && pkcs7->signerInfo->sid != NULL &&
+                !wc_PKCS7_CertMatchesSignerInfo(pkcs7, dCert)) {
```
dgarske, line 5571: **🔴 [High] ML-DSA verify references wc_PKCS7_CertMatchesSignerInfo which is not compiled in RSA-less/ECC-less builds** · _bug_

The new `wc_PKCS7_MlDsaVerify()` is gated only on ML-DSA macros (`WC_PKCS7_HAVE_MLDSA && !WOLFSSL_MLDSA_NO_VERIFY && WOLFSSL_MLDSA_PUBLIC_KEY`) and calls `wc_PKCS7_CertMatchesSignerInfo(pkcs7, dCert)` at line 5396. That helper is defined under `#if !defined(NO_RSA) || defined(HAVE_ECC)` (line 4771), which contains no PQC/ML-DSA condition. In a PQC-only PKCS#7 configuration (`NO_RSA` defined and `HAVE_ECC` undefined, but ML-DSA verify enabled) the helper is never compiled while the ML-DSA verifier still references it, producing an implicit-declaration / undefined-symbol build failure (a hard error under the project's `-Werror`). This config is newly meaningful because ML-DSA is the first signature algorithm that lets PKCS#7 SignedData verify work without RSA or ECC. The PR verified several degenerate ML-DSA combinations but all under `--enable-all` (RSA+ECC present), so this combination was not exercised.

**Fix:** Widen the `wc_PKCS7_CertMatchesSignerInfo` definition guard to also include the ML-DSA verify condition (`|| defined(WC_PKCS7_MLDSA_VERIFY)`), and add a build of the PQC-only combination (ML-DSA verify with `NO_RSA` and no ECC) to the compile matrix so this class of gating gap is caught.

wolfcrypt/src/pkcs7.c
```diff
@@ -2666,7 +2972,7 @@ static int wc_PKCS7_SignedDataBuildSignature(wc_PKCS7* pkcs7,
                 /* user signing plain digest, build DigestInfo themselves */
                 ret = pkcs7->rsaSignRawDigestCb(pkcs7,
```
dgarske, line 3067: **🔵 [Low] RSA raw-sign callback output buffer tightened to modulus size with no post-write bounds check** · _Security_

Both the review and review-security modes flagged the same change. BEFORE: `esd->encContentDigest` was a fixed `byte[MAX_ENCRYPTED_KEY_SZ]` stack array, so the `rsaSignRawDigestCb` callback always had MAX_ENCRYPTED_KEY_SZ of writable space regardless of key size. AFTER: the buffer is heap-allocated and, on the reserve path (RSA with `hashBuf == NULL`), right-sized to exactly the RSA modulus length returned by `wc_PKCS7_GetSignSize()`, and the callback is handed `esd->encContentDigestBufSz` (= modulus size) as its capacity. Unlike the ECDSA raw-callback path (lines 3002-3005: `if (ret > 0 && (word32)ret > esd->encContentDigestBufSz) ret = BUFFER_E;`), the RSA raw-callback path performs no post-write length validation. A correct PKCS#1 v1.5 signature is exactly the modulus size, so a spec-conforming callback fits precisely and cannot overflow; the concern is that the defensive margin present in the prior code (MAX_ENCRYPTED_KEY_SZ) is gone for a caller-supplied callback that writes more than the modulus size. This is only reachable with `HAVE_PKCS7_RSA_RAW_SIGN_CALLBACK` compiled in and an integrator-supplied callback that violates its own size contract; it is not attacker-controlled data. The subsequent fixed-size invariant check (line 4046) would report `BUFFER_E`, but only after the write has occurred. Both modes rate this Low; the review mode framed it as a SUGGEST/behavioral question about consumer assumptions.

**Fix:** Mirror the ECDSA raw-callback guard for the RSA raw-callback path: after the call, add `if (ret > 0 && (word32)ret > esd->encContentDigestBufSz) ret = BUFFER_E;`. This restores the defensive check for a path whose buffer this PR intentionally tightened. Also confirm no in-tree/consumer raw-sign callback relies on the buffer being MAX_ENCRYPTED_KEY_SZ rather than exactly the RSA output size.

wolfcrypt/src/pkcs7.c
```diff
@@ -1237,16 +1312,32 @@ int wc_PKCS7_InitWithCert(wc_PKCS7* pkcs7, byte* derCert, word32 derCertSz)
             return ret;
         }
 
-        if (dCert->pubKeySize > (MAX_RSA_INT_SZ + MAX_RSA_E_SZ) ||
-            dCert->serialSz > MAX_SN_SZ) {
-            WOLFSSL_MSG("Invalid size in certificate");
+        if (dCert->serialSz > MAX_SN_SZ) {
```
dgarske, line 1332: **🔵 [Low] InitWithCert no longer size-checks non-RSA/ECC public keys** · _behavioral-change_

The previous code rejected any certificate whose `dCert->pubKeySize > (MAX_RSA_INT_SZ + MAX_RSA_E_SZ)` regardless of key type. The rewrite moves that bound inside the `RSAk/RSAPSSk/ECDSAk` branch and takes the `else { pkcs7->publicKeySz = 0; }` path for everything else. This is correct and necessary for large ML-DSA keys, but it also silently drops the sanity bound for any other/unsupported key OID (e.g. Ed25519, Falcon, or an unknown OID): such a cert now passes `InitWithCert` (with `publicKeySz = 0`) instead of being rejected here. `wc_PKCS7_CheckPublicKeyDer` has no `default` case, so it also returns success (ret stays 0) for those OIDs, so the error is only surfaced later as `BAD_FUNC_ARG` from the sign/verify switch. No memory-safety impact (nothing is copied into the fixed buffer for the else branch), but it is a behavioral change in where unsupported-key certs are rejected.

**Fix:** Confirm this deferred-rejection behavior for unsupported key types is intended. Optionally reject clearly-unsupported keyOIDs explicitly in `wc_PKCS7_CheckPublicKeyDer` (add a `default: ret = ASN_SIG_KEY_E;`) so unsupported certs fail at init time as before, rather than only at sign/verify.

wolfcrypt/src/hash.c
```diff
@@ -353,6 +353,20 @@ enum wc_HashType wc_OidGetHash(int oid)
             hash_type = WC_ERR_TRACE(WC_HASH_TYPE_NONE);
         #endif
             break;
+        case SHAKE128h:
```
dgarske, line 356: **⚪ [Info] wc_OidGetHash SHAKE128/SHAKE256 mapping affects OCSP/CRL/PKCS12/RSA-PSS beyond the stated PKCS#7 scope** · _Blast Radius_

BEFORE: `wc_OidGetHash(SHAKE128h/SHAKE256h)` returned `WC_HASH_TYPE_NONE`, so any consumer that fed a SHAKE digest OID through this function failed (NONE -> `wc_HashGetDigestSize` error). AFTER: they resolve to `WC_HASH_TYPE_SHAKE128`/`WC_HASH_TYPE_SHAKE256` (fixed 32/64-byte output via the existing wc_Hash SHAKE support). `wc_OidGetHash` is shared by many subsystems beyond PKCS#7: `asn.c` OCSP/CRL digest sizing (lines 33991/34054/35543/35730/35917), `pkcs12.c` MAC (line 577), and RSA-PSS MGF resolution in `pk_rsa.c`. This PR consequently broadens where SHAKE digest OIDs are silently accepted across all of those paths, not only in the new CMS code. Since SHAKE is used here with a defined fixed output length (RFC 8702) and `wc_Hash*` fully supports SHAKE, this is most likely intended and safe, but the behavioral expansion is wider than the PR summary (which scopes it to PKCS#7/CMS) implies and should be confirmed against the OCSP/CRL/PKCS12 consumers.

**Fix:** Confirm the OCSP/CRL (`asn.c`), PKCS12 MAC (`pkcs12.c`), and RSA-PSS MGF (`pk_rsa.c`) consumers behave acceptably when a SHAKE digest OID now resolves to a valid hash type (previously rejected). No code change needed if the expanded acceptance is intended; otherwise gate the SHAKE mapping to the CMS use case.

wolfcrypt/src/pkcs7.c
```diff
@@ -4908,6 +5319,145 @@ static int wc_PKCS7_EcdsaVerify(wc_PKCS7* pkcs7, byte* sig, int sigSz,
...
+            wc_MlDsaKey_Free(key);
+            continue;
+        }
+
+        ret = wc_MlDsaKey_SetParams(key, level);
+        if (ret == 0) {
+            ret = wc_MlDsaKey_PublicKeyDecode(key, dCert->publicKey,
+                                              dCert->pubKeySize, &idx);
+        }
+        if (ret < 0) {
```
dgarske, line 5591: **⚪ [Info] ML-DSA verify masks hard internal errors from key decode as SIG_VERIFY_E** · _convention_

On a `wc_MlDsaKey_SetParams`/`wc_MlDsaKey_PublicKeyDecode` failure the loop `continue`s and, if no cert verifies, returns `SIG_VERIFY_E` even for hard failures such as `MEMORY_E`. This is explicitly documented in the code and intentionally mirrors the RSA/ECC verify helpers, so it is consistent with existing behavior. Noting only for awareness: a genuine `MEMORY_E` during public-key decode is reported to the caller as a signature-verification failure rather than an allocation error. The `wc_MlDsaKey_Init` failure path (line 5370) is handled distinctly and does return the hard error, so the treatment is asymmetric but matches the established pattern.

**Fix:** No change needed; behavior is deliberate and consistent with the RSA/ECC verifiers. Left as an observation only.

dgarske (changes requested):
https://github.com/wolfSSL/wolfssl/pull/10759#pullrequestreview-4621654238

wolfcrypt/src/pkcs7.c
Frauschi: Oh wow, what a catch... Fixed by widening the guard. Also, PKCS#7 with ML-DSA only is now possible in configure.ac and tested in CI. 

(Why haven't my Skoll flagged this earlier?! 😔)

wolfcrypt/src/pkcs7.c
Frauschi: fixed

wolfcrypt/src/pkcs7.c
Frauschi: This is intentional, as nothing is copied for non-RSA/ECC public keys. And adding the default in wc_PKCS7_CheckPublicKeyDer() would break ML-DSA sign-only builds.

wolfcrypt/src/hash.c
Frauschi: This is intended.

wolfcrypt/src/pkcs7.c
Frauschi: Intended

Frauschi:
Jenkins retest this please

wolfcrypt/src/pkcs7.c
```diff
@@ -1033,6 +1048,37 @@ static int wc_PKCS7_RecipientListVersionsAllZero(wc_PKCS7* pkcs7)
...
+    if (pkcs7->hashOID == SHAKE256h || pkcs7->hashOID == SHAKE128h) {
+        return 1;
+    }
+#endif
+    return pkcs7->hashParamsAbsent;
+}
+
 /* Verify RSA/ECC key is correctly formatted, used as sanity check after
  * import of key/cert.
  *
```
dgarske, line 1096: **🟠 [Medium] scratch set-but-unused in wc_PKCS7_CheckPublicKeyDer for ML-DSA-sign-only + no-RSA/no-ECC build**

This PR relaxes the PKCS#7 build guards (configure.ac / settings.h) to permit PKCS#7 with ML-DSA as the *only* algorithm. `word32 scratch = 0;` at line 1094 is only ever read inside the `switch(keyOID)` cases, all of which are conditionally compiled: RSA (`!NO_RSA`), ECC (`HAVE_ECC`), and the new ML-DSA case (`WC_PKCS7_HAVE_MLDSA && WOLFSSL_MLDSA_PUBLIC_KEY`). Per dilithium.h:317-320, `WOLFSSL_MLDSA_PUBLIC_KEY` is undefined when both `WOLFSSL_MLDSA_NO_VERIFY` and `WOLFSSL_MLDSA_NO_MAKE_KEY` are set (i.e. an ML-DSA sign-only build). In the combination `--enable-pkcs7 --enable-mldsa(sign-only) --disable-rsa --disable-ecc`, the switch has zero cases compiled, so `scratch` is set but never read, which trips `-Wunused-but-set-variable` under the project's `-Werror`. Before this PR this combination was impossible (configure required RSA or ECC for PKCS#7), so the PR newly exposes it. The added `pkcs7-mldsa-only` CI job uses full `--enable-mldsa` (verify enabled), so it does not exercise this sub-variant. The PR description claims '-Werror clean under degenerate ML-DSA feature combinations', so this gap is worth closing.

**Fix:** Add `(void)scratch;` after the declaration (or guard the declaration), and ideally add a CI matrix entry for an ML-DSA sign-only PKCS#7 build (e.g. `--enable-mldsa` with verify disabled + `--disable-rsa --disable-ecc`) to cover the degenerate path.

---
> **Note:** Referenced line (`wolfcrypt/src/pkcs7.c:1094`) is outside the diff hunk. Comment anchored to nearest changed region.

dgarske (commented):
## Skoll Multi-Scan Review

### Posted findings
- [Medium] [review] **scratch set-but-unused in wc_PKCS7_CheckPublicKeyDer for ML-DSA-sign-only + no-RSA/no-ECC build** — `wolfcrypt/src/pkcs7.c:1094`

#### Redundant content digest computed on ML-DSA verify path
**File:** `wolfcrypt/src/pkcs7.c:5799`
**Function:** `wc_PKCS7_SignedDataVerifySignature`
**Severity:** Low

For the ML-DSA case, `wc_PKCS7_BuildSignedDataDigest` (line 5799) still computes `pkcs7Digest`/`plainDigest`, but the ML-DSA branch ignores them and rebuilds the pure-mode message via `wc_PKCS7_BuildPureSigMessage`. This is harmless (and the digest is genuinely needed for the earlier `wc_PKCS7_VerifyContentMessageDigest` when signed attributes are present), so it is not a correctness issue - just a small amount of unused work for the no-signed-attributes ML-DSA case. Noting only for awareness; no change required.

**Recommendation:** Optional: no action needed. Left as-is keeps the shared verify flow simple and preserves the messageDigest binding check.

_Referenced code: `wolfcrypt/src/pkcs7.c:5799-5803` (5 lines)_

---

#### Detached ML-DSA SignedData over empty content is rejected (functional gap, not a security defect)
**File:** `wolfcrypt/src/pkcs7.c:4305-4310, 3447-3453`
**Function:** `wc_PKCS7_EncodeSignedData / PKCS7_EncodeSigned`
**Severity:** Info
**Category:** Logic

`wc_PKCS7_EncodeSignedData` pre-computes a content hash and calls `PKCS7_EncodeSigned` with a non-NULL `hashBuf` for three cases, including `(pkcs7->detached && pkcs7->contentSz == 0)` (the SCEP empty-content case). The PR adds a guard in `PKCS7_EncodeSigned` (line 3447) that rejects any ML-DSA signer when `hashBuf != NULL` with `BAD_FUNC_ARG`. As a result, a detached ML-DSA SignedData over empty content (`pkcs7->detached` set, `contentSz == 0`) is unconditionally rejected, whereas RSA/ECDSA support it. This is a new-feature limitation (ML-DSA was previously unsupported in PKCS#7, so there is no regression) and is a clean, safe error return rather than a memory-safety issue. Detached/non-empty and embedded ML-DSA both work correctly. Worth documenting the limitation, but it carries no security impact.

**Recommendation:** No security action required. Optionally document that detached-over-empty-content SignedData is unsupported for ML-DSA, or special-case the `detached && contentSz==0` path for ML-DSA so the pure-mode signature is computed over the (empty) eContent instead of taking the pre-hash branch.

_Referenced code: `wolfcrypt/src/pkcs7.c:4305-4310, 3447-3453` (13 lines)_

---

wolfcrypt/src/pkcs7.c
Frauschi: fixed

Frauschi:
Jenkins retest this please

wolfcrypt/src/pkcs7.c
```diff
@@ -2699,12 +3022,24 @@ static int wc_PKCS7_SignedDataBuildSignature(wc_PKCS7* pkcs7,
...
-        return ret;
+#ifdef WC_PKCS7_HAVE_MLDSA
+    /* ML-DSA signs the full message in pure mode (RFC 9882); it consumes no
+     * DigestInfo or content digest, so skip building one. */
+    if (pkcs7->publicKeyOID == ML_DSA_44k ||
+        pkcs7->publicKeyOID == ML_DSA_65k ||
+        pkcs7->publicKeyOID == ML_DSA_87k) {
+        skipDigestInfo = 1;
+    }
+#endif
```
dgarske, line 3033: **🔵 [Low] Unused DigestInfo buffer allocated for ML-DSA signing** · _efficiency_

`digestInfo` (MAX_PKCS7_DIGEST_SZ) is unconditionally allocated via WC_ALLOC_VAR_EX before the `skipDigestInfo` check. For ML-DSA (`skipDigestInfo == 1`) the buffer is never written or read, yet is still allocated and freed on every sign. Minor heap churn / pool pressure on constrained targets that build ML-DSA-only PKCS#7.

**Fix:** Defer the digestInfo allocation until after the skipDigestInfo determination, allocating only when !skipDigestInfo; skip the allocation for the ML-DSA (pure-mode) path.

wolfcrypt/src/pkcs7.c
```diff
@@ -5461,6 +6046,31 @@ static int wc_PKCS7_SignedDataVerifySignature(wc_PKCS7* pkcs7, byte* sig,
...
+                                               signedAttribSz, &msg, &msgSz,
+                                               &msgAlloc);
+            if (ret == 0) {
+                ret = wc_PKCS7_MlDsaVerify(pkcs7, sig, (int)sigSz, msg, msgSz);
+                if (msgAlloc) {
+                    XFREE(msg, pkcs7->heap, DYNAMIC_TYPE_TMP_BUFFER);
+                }
+            }
+            break;
+        }
```
dgarske, line 6088: **🔵 [Low] Content digest recomputed then discarded on ML-DSA verify** · _efficiency_

On the ML-DSA verify branch, `wc_PKCS7_BuildSignedDataDigest()` is still invoked (line 5945) to build `pkcs7Digest`/`plainDigest`, but ML-DSA verification uses the full message rebuilt by `wc_PKCS7_BuildPureSigMessage` and never consumes those digests. This performs an extra full hash of the attributes/content per verify. The digest is also used by the no-cert store path, so it cannot simply be removed, but for the common (has-cert) ML-DSA case it is wasted work.

**Fix:** Consider skipping the digest build for ML-DSA when a certificate is present in the bundle (i.e. when the store-for-later path will not run). Optional micro-optimization; not required for correctness.

dgarske (commented):
## Skoll Multi-Scan Review

**Findings:** 6 total — 3 posted, 3 skipped

### Posted findings
- [Low] [review] **Unused DigestInfo buffer allocated for ML-DSA signing** — `wolfcrypt/src/pkcs7.c:3015-3033`
- [Low] [review] **Content digest recomputed then discarded on ML-DSA verify** — `wolfcrypt/src/pkcs7.c:5945-6071`

#### ML-DSA 'no certificates in bundle' verify path stores an unusable digest
**File:** `wolfcrypt/src/pkcs7.c:5954-6015`
**Function:** `wc_PKCS7_SignedDataVerifySignature`
**Severity:** Medium
**Category:** correctness

When a SignedData bundle contains no certificates, the verifier stores `pkcs7->signature`, `pkcs7->plainDigest`, and `pkcs7->pkcs7Digest` and returns `PKCS7_SIGNEEDS_CHECK` so the caller can verify externally with their own key. This design assumes the signature is over a digest. ML-DSA is used in 'pure' mode (RFC 9882) and signs the complete message (the re-tagged SET OF signed attributes, or the eContent), not a digest. For an ML-DSA signer with no cert in the bundle, the stored `plainDigest`/`pkcs7Digest` is insufficient to verify the signature -- the caller has no way to reconstruct the exact octet string that was signed. This runs before the `switch (pkcs7->publicKeyOID)` dispatch, so it applies to ML-DSA too. It fails closed (no false-accept), but silently hands back data that cannot complete an ML-DSA verification.

**Recommendation:** For ML_DSA_44k/65k/87k, either return a clear 'unsupported' error in the no-cert case, or store the reconstructed pure-mode message (from wc_PKCS7_BuildPureSigMessage) instead of a digest so an external verifier can use it. Document the limitation or special-case ML-DSA in the no-cert path so callers do not receive verification material they cannot use.

_Referenced code: `wolfcrypt/src/pkcs7.c:5954-5958` (5 lines)_

---

### Skipped findings
- [Medium] ~~No test coverage for ML-DSA no-certificate external-verify path~~
- [Low] ~~configure.ac PKCS7 check does not account for WOLFSSL_MLDSA_NO_ASN1~~
- [Info] ~~wc_OidGetHash() SHAKE OID mapping widens accepted digests beyond PKCS#7 (blast radius)~~

wolfcrypt/src/pkcs7.c
Frauschi: This is won't fix. The buffer is small (< 100 bytes) and adding a fix would increase code complexity.

wolfcrypt/src/pkcs7.c
Frauschi: Also won't fix. This micro-optimization would increase the complexity for only a small win.

Frauschi:
I fixed the latest _"ML-DSA 'no certificates in bundle' verify path"_ issue.

dgarske (approved):

PR #10760, opened 2026-06-23
PKCS#7 improvements

Frauschi:
### Summary

Server-side PKCS#7/CMS encode improvements so downstream EST/SCEP enrollment code can drive the existing encoder through the public API instead of hand-rolling DER, plus a related multi-certificate decode bug fix found along the way. All changes are gated under the existing `HAVE_PKCS7` - no new build options and no new public functions.

### Changes

1. Allow degenerate (certs-only) `SignedData` encode

Relaxes the `hashOID != 0` requirement in `PKCS7_EncodeSigned()` when `sidType == DEGENERATE_SID`. A caller can now produce a degenerate, certs-only `SignedData` (no signer, no signed attributes, no eContent - the form used by EST `/cacerts` and SCEP `GetCACert` responses) entirely through the existing public API:

```C
wc_PKCS7_InitWithCert(pkcs7, cert, certSz);          
/* wc_PKCS7_AddCertificate() */
wc_PKCS7_SetSignerIdentifierType(pkcs7, DEGENERATE_SID);
pkcs7->detached   = 1;
pkcs7->contentOID = DATA;
wc_PKCS7_EncodeSignedData(pkcs7, out, outSz);
```

The output round-trips through `wc_PKCS7_VerifySignedData()` (which repopulates `pkcs7->cert[]`).

2. Size the signed-attribute array to the actual count

The `SignerInfo` attribute working array is now sized to the real attribute count instead of a fixed `[7]` array. An inline buffer (sized `MAX_SIGNED_ATTRIBS_SZ`, the historical footprint) covers the common case with no heap allocation - important for no-malloc/static-memory builds - and a heap buffer is used only when the count exceeds it. The default-attribute count comes from a single helper (`wc_PKCS7_GetDefaultSignedAttribCount()`) so the sizing matches the emission logic exactly, and the canned-attribute write is bound-checked against the array capacity so any future drift fails with `BUFFER_E` rather than overflowing.

This also fixes a latent overflow where the backing array was hardcoded `[7]` while the bound check used `MAX_SIGNED_ATTRIBS_SZ`. Profiles such as SCEP `CertRep` (RFC 8894), which carry several user attributes alongside the CMS auto-defaults, now encode without spurious `BUFFER_E`. `MAX_SIGNED_ATTRIBS_SZ` is retained for source compatibility but no longer caps the count (it only sizes the embedded inline buffer).

3. Document the decoded-attribute value shape

Documents the stable, guaranteed shape of `PKCS7DecodedAttrib.value` (the contents of the `SET OF AttributeValue`, with the outer `SET` tag stripped) so callers can rely on it. Documentation only - no behavior change.

4. Multi-certificate decode fix (non-streaming)

Bounds the additional-certificate loop in `wc_PKCS7_VerifySignedData()` against the absolute end of the certificate set (`idx + length`) rather than the relative length. In `NO_PKCS7_STREAM` builds the previous bound stopped short and dropped trailing certificates - and, with a large eContent preceding the set, could drop all but the first certificate, causing verification to fail when the signer cert was among those dropped. Streaming builds were unaffected.

### Public API

No additions or changes. `MAX_SIGNED_ATTRIBS_SZ` is retained for source compatibility but no longer caps the attribute count (it now sizes the inline encode-time working buffer).

### Tests

Added coverage in `pkcs7signed_test`:
- Degenerate (certs-only) encode via the public API, round-tripped through verify.
- Nine signed attributes (six user + three auto-defaults), exceeding the inline capacity to exercise the heap-fallback path.
- Decoded-attribute value shape for `PrintableString` and `OCTET STRING`.
- A multi-certificate decode regression with large content that triggers the bound bug under `NO_PKCS7_STREAM`.

Config-sensitive cases are guarded (`MAX_PKCS7_CERTS >= 3`; the nine-attribute case requires heap or an enlarged inline array). The multi-cert test only validates the bound fix under `NO_PKCS7_STREAM` (documented in the test); in streaming builds it is a general multi-cert round-trip check.

### Verification

- Builds clean (`-Werror`) on the default/streaming configuration.
- `wolfcrypt/test/testwolfcrypt` passes (`PKCS7signed test passed!`).


Frauschi:
Jenkins retest this please

wolfcrypt/test/test.c
```diff
@@ -67438,6 +67438,395 @@ static wc_test_ret_t pkcs7signed_run_SingleShotVectors(
...
+    encSz = wc_PKCS7_EncodeSignedData(pkcs7, out, outSz);
+    if (encSz <= 0)
+        ERROR_OUT(WC_TEST_RET_ENC_EC(encSz), out);
+    wc_PKCS7_Free(pkcs7);
+    pkcs7 = NULL;
+
+    pkcs7 = wc_PKCS7_New(HEAP_HINT, devId);
+    if (pkcs7 == NULL)
+        ERROR_OUT(WC_TEST_RET_ENC_ERRNO, out);
+    ret = wc_PKCS7_VerifySignedData(pkcs7, out, (word32)encSz);
```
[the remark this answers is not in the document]
Frauschi: fixed

wolfcrypt/src/pkcs7.c
```diff
@@ -3266,6 +3325,53 @@ static int PKCS7_EncodeSigned(wc_PKCS7* pkcs7,
         }
         signerInfoSz += esd->digEncAlgoIdSz;
 
+        /* Point the working attribute array at the inline buffer, sized for the
+         * actual attribute count: the user-supplied attributes plus the exact
+         * number of CMS auto-defaults that will be emitted for this
+         * pkcs7->defaultSignedAttribs selection. Only fall back to a heap
+         * allocation when more attributes are needed than fit inline, so the
+         * common case stays allocation-free. */
+        {
```
dgarske: **🔵 [Low] New brace-scoped block introduced to declare local temporaries in PKCS7_EncodeSigned** · _convention_

The attribute-array sizing logic is wrapped in a fresh `{ ... }` scope solely so `defaultAttribCap` and `neededCap` can be declared mid-function. wolfSSL's C89 house style favors declaring locals at the top of the function or an existing block rather than opening a new brace scope. This is borderline since the block also scopes the associated allocation logic (not purely a declaration), so it is defensible; noting it only for consistency with the surrounding style. No functional impact.

**Fix:** Take it or leave it; acceptable as-is because the block scopes real logic, not just declarations.

dgarske (commented):
## Skoll Multi-Scan Review

### Posted findings
- [Low] [review] **New brace-scoped block introduced to declare local temporaries in PKCS7_EncodeSigned** — `wolfcrypt/src/pkcs7.c:3334-3373`

### Skipped findings
- [Low] ~~Default-attribute count is a second source of truth that must stay hand-synced with the emission logic~~

dgarske (changes requested):
https://github.com/wolfSSL/wolfssl/pull/10760#pullrequestreview-4621576995

wolfcrypt/src/pkcs7.c
Frauschi: fixed

Frauschi:
Jenkins retest this please

Frauschi:
Jenkins retest this please

dgarske (approved):

PR #10762, opened 2026-06-23
AES reject-no-key guard and RSA-PSS hash-binding test (F-6151, F-6157)

julek-wolfssl:
Two independent hardening fixes with regression tests.

### AES: reject mode operations when no key has been set (F-6151)
`wc_AesInit` zeroes the `Aes` struct, leaving `keylen`/`rounds` at 0. The CTR/CBC/GCM/ECB/CFB/OFB entry points never verified a key had been installed via `wc_AesSetKey`, so calling them right after `wc_AesInit` ran the software cipher against an all-zero key schedule and returned success. For CTR this exposes a reconstructable keystream; for GCM the hash subkey H is also zero, making the tag forgeable.

Adds the `keylen == 0` guard (already used by `wc_AesXtsEncrypt`) to the software mode paths, placed after the crypto-callback fall-through so device-managed keys are unaffected. New `aes_no_key_set_test()` (run from `aes_test()`) confirms every mode rejects use before a key is set.

### RSA-PSS: test signature-to-message hash binding (F-6157)
`wc_RsaPSS_CheckPadding_ex2` binds a PSS signature to the message with a single comparison of the recomputed hash against the encoded hash H. Every positive PSS test passed the correct digest and the negative tests failed earlier on a wrong salt length, so removing the comparison left all default tests passing even though any well-formed PSS structure would verify against any message.

After each successful sign/verify round-trip in `rsa_pss_test`, flips one bit of the message digest, keeps the correct salt length, and requires `BAD_PADDING_E` — making the hash comparison the deciding check.


Frauschi:
@julek-wolfssl please resolve conflicts with master

wolfcrypt/src/aes.c
```diff
@@ -15184,6 +15249,10 @@ int wc_AesEcbEncrypt(Aes* aes, byte* out, const byte* in, word32 sz)
 {
```
Frauschi, line 16481: **🟡 [Medium] AES-CCM not covered by the keyInstalled guard (same zero-key exposure as GCM)**
*💡 SUGGEST*  `test`

The PR adds a `keyInstalled == 0` guard to the software CBC/CTR/GCM/ECB/CFB/OFB entry points to stop them running against the all-zero key schedule left by `wc_AesInit`. The stated motivation for GCM is that with a zero key the hash subkey H is also zero, making the tag forgeable. The software `wc_AesCcmEncrypt`/`wc_AesCcmDecrypt` are AEAD modes with the exact same failure mode (CBC-MAC/CTR over an all-zero key schedule via `wc_AesEncrypt`), yet they receive no guard and are not exercised by `aes_no_key_set_test`. This leaves an inconsistent gap in the hardening: CCM before `wc_AesSetKey`/`wc_AesCcmSetNonce` still runs with the zero key. Consider extending the same guard and the test to CCM, or documenting why CCM is intentionally out of scope.

**Suggestion:**
```suggestion
Add `if (aes->keyInstalled == 0) return BAD_FUNC_ARG;` to the software CCM enc/dec paths (after the crypto-cb fall-through, mirroring GCM) and add CCM cases to aes_no_key_set_test guarded by HAVE_AESCCM.
```

wolfssl/wolfcrypt/aes.h
```diff
@@ -293,6 +293,12 @@ struct Aes {
 #endif
     int     keylen;
 
+    /* Set to 1 once a key has been installed (wc_AesSetKey/SetKeyDirect/
+     * GcmSetKey). Checked by the mode APIs so they fail instead of running
+     * with the all-zero key schedule left by wc_AesInit. Distinct from the
+     * Cavium-only keySet field below. */
+    WC_BITFIELD keyInstalled:1;
```
Frauschi: **🔵 [Low] New Aes struct member inserted mid-struct rather than appended at the end**
*💡 SUGGEST*  `convention`

`WC_BITFIELD keyInstalled:1;` is inserted between `int keylen;` and `ALIGN16 word32 reg[...]`. wolfSSL convention for public structs is to append new members at the end to avoid changing offsets of existing members (binary/ABI compatibility across library versions). In practice the following `reg` member is `ALIGN16`, so the 1-bit field most likely lands in pre-existing alignment padding and offsets are unchanged — but relying on that is fragile and the placement still deviates from the documented convention. Recommend placing new members at the end of the struct unless there is a specific reason (grouping with the nearby `keySet`/bitfields) that should be called out.

**Suggestion:**
```suggestion
Move the keyInstalled bitfield to the end of struct Aes (or immediately adjacent to the existing keySet/bitfield group) to preserve member layout.
```

Frauschi (commented):
## 🐺 Skoll Code Review

**Findings:** 3 total — 3 posted, 1 skipped

### Posted findings
- [Medium] **AES-CCM not covered by the keyInstalled guard (same zero-key exposure as GCM)** — `wolfcrypt/src/aes.c:14403`
- [Low] **New Aes struct member inserted mid-struct rather than appended at the end** — `wolfssl/wolfcrypt/aes.h:300`
- [Info] **sz==0 early-return precedes keyInstalled check in some mode paths (inconsistent no-key result)** — `wolfcrypt/src/aes.c:7001, 15568, 15656, 16113`

- [Info] sz==0 early-return precedes keyInstalled check in some mode paths (inconsistent no-key result)

</details>

wolfssl/wolfcrypt/aes.h
julek-wolfssl: this is wrong

wolfcrypt/src/aes.c
julek-wolfssl: addressing

wolfcrypt/src/aes.c
```diff
@@ -10806,6 +10863,12 @@ int wc_AesGcmEncrypt(Aes* aes, byte* out, const byte* in, word32 sz,
     }
 #endif
 
+    /* Software/HW key schedule (and hash subkey H) required from here on. */
+    if (aes->keyInstalled == 0) {
```
Frauschi: **🟠 [High] keyInstalled guard breaks AES modes for ports that supply their own wc_AesSetKey**
*🚫 BLOCK*  `bug`

The new `if (aes->keyInstalled == 0) return BAD_FUNC_ARG;` guards were added to the generic software/HW AES mode functions in `aes.c`, but `keyInstalled = 1` is set **only** by aes.c's own key-schedule path (`wc_AesSetKey`/`wc_AesSetKeyLocal`) and the RISC-V port. Every port that ships its **own compile-time `wc_AesSetKey`** leaves `keyInstalled = 0`, so any AES mode that falls through to a guarded `aes.c` path now returns `BAD_FUNC_ARG` after a *correct* key setup — a functional regression.

**Confirmed by tracing PR head (`d20beba91`):**
- silabs `wc_AesSetKey` (`port/silabs/silabs_aes.c:39`) never sets `keyInstalled`; `wc_AesGcmSetKey` → `wc_AesSetKey` (`aes.c:8376`) → silabs, so the flag stays `0`.
- In `wc_AesGcmEncrypt` the guard (`aes.c:10867`) sits **before** the silabs dispatch (`aes.c:10909`), so silabs GCM encrypt/decrypt always fail (GCM decrypt: guard `11662` → silabs `11704`).

**Scope-limiting factor:** the guard is placed **after** the `wc_CryptoCb_*` dispatch in every mode function, so crypto-callback / `devId` ports return before reaching it and are unaffected.

**Blast radius** — ports with their own `wc_AesSetKey`, none of which set `keyInstalled`:

| Port | Broken modes (fall through to guarded aes.c) |
|------|------|
| **silabs** | **GCM** enc/dec (dispatch after guard), CTR, CFB, OFB |
| **af_alg** | **CBC**, CTR, CFB, OFB, CCM (implements only GCM+ECB; SW key schedule *is* expanded, L147-168) |
| **caam** | **GCM**, CFB, OFB |
| **kcapi** | **CTR**, CCM, CFB, OFB, ECB (implements only CBC+GCM) |
| **devcrypto** | CCM, CFB, OFB |
| **ti** | ECB, CFB, OFB |

**Not affected:** RISC-V (PR sets the flag), Renesas FSPSM and other crypto-cb/`devId` ports (guard is after the cb dispatch), PSA (own key object + own CBC/CTR).

Highest-impact regressions: **silabs GCM** (primary TLS AEAD), **af_alg CBC**, **caam GCM**, **kcapi CTR**.

**Recommendation:** Set `aes->keyInstalled = 1` in every static-dispatch port's `wc_AesSetKey` (silabs, af_alg, devcrypto, caam, ti — grep `wolfcrypt/src -rn 'int wc_AesSetKey'` for the full list), OR restructure the guard so it only applies where the aes.c software key schedule is actually the key store. Add a regression build/test for at least one HW port (e.g. silabs) with GCM/CTR enabled.

Frauschi (changes requested):
## 🐺 Skoll Code Review

### Posted findings
- [High] **keyInstalled guard breaks AES modes for ports that supply their own wc_AesSetKey** — `wolfcrypt/src/aes.c:10867`

wolfcrypt/src/aes.c
julek-wolfssl: addressed.

wolfcrypt/src/aes.c
```diff
@@ -5578,6 +5591,7 @@ static void AesSetKey_C(Aes* aes, const byte* key, word32 keySz, int dir)
          * reads it as the source of truth for the configured key size. */
```
Frauschi, line 5803: **🟡 [Medium] WOLFSSL_SECO_CAAM key-set path returns success without setting keyInstalled**
*💡 SUGGEST*  `bug`

Every other success return in the AES key-set paths now sets `aes->keyInstalled = 1`, and the new mode guards (`wc_AesGcmEncrypt`, `wc_AesCbcEncrypt`, `wc_AesCcmEncrypt`, etc.) reject the call when `keyInstalled == 0`. The `WOLFSSL_SECO_CAAM` branch, however, wraps the key into a black key and `return 0;` without setting `keyInstalled`. For `WOLFSSL_SECO_CAAM` builds, `wc_AesGcmSetKey` calls `wc_AesSetKey` (which returns 0 here) and then returns; `keyInstalled` stays 0. If any subsequent SECO AES mode operation reaches a guarded software/HW path in `aes.c` (the software `wc_AesGcmEncrypt` at line 10830 has no SECO early-return branch), it will now wrongly return `BAD_FUNC_ARG`, breaking AES for that platform. This is the one success path in the whole diff that was missed.

**Suggestion:**
```suggestion
aes->blackKey = keyIdOut;
        aes->keyInstalled = 1;
        return 0;
```

wolfcrypt/test/test.c
```diff
@@ -16241,12 +16241,160 @@ static wc_test_ret_t aes_ecb_direct_test(void)
...
+
+    XMEMSET(plain, 0, sizeof(plain));
+    XMEMSET(cipher, 0, sizeof(cipher));
+#if defined(HAVE_AESGCM) || defined(HAVE_AESCCM)
+    XMEMSET(iv, 0, sizeof(iv));
+    XMEMSET(tag, 0, sizeof(tag));
+#endif
+
+#if defined(WOLFSSL_SMALL_STACK) && !defined(WOLFSSL_NO_MALLOC)
+    aes = wc_AesNew(HEAP_HINT, devId, &ret);
```
Frauschi: **🟡 [Medium] aes_no_key_set_test may fail under WOLF_CRYPTO_CB / device builds**
*💡 SUGGEST*  `test`

The test creates the Aes context with the global `devId` and asserts every mode API returns `BAD_FUNC_ARG` before a key is set. The new `keyInstalled` guards are deliberately placed AFTER the crypto-callback dispatch (per the PR design), so on a build where `devId != INVALID_DEVID` (e.g. a registered `WOLF_CRYPTO_CB` test device), `wc_AesGcmEncrypt`/`wc_AesCbcEncrypt`/etc. route to the callback first. The callback runs against the zero key context and may return 0 (success) or a device-specific error rather than `BAD_FUNC_ARG`, causing the test to ERROR_OUT. The test only excludes `HAVE_FIPS`/`HAVE_SELFTEST`, not crypto-cb/offload builds.

**Suggestion:** Use INVALID_DEVID for this specific negative test (the guard is a pure-software hardening), e.g. `wc_AesInit(aes, HEAP_HINT, INVALID_DEVID)`, or gate the test body with `#if !defined(WOLF_CRYPTO_CB)` in addition to the existing FIPS/SELFTEST guard.

Frauschi (commented):
## 🐺 Skoll Code Review

### Posted findings
- [Medium] **WOLFSSL_SECO_CAAM key-set path returns success without setting keyInstalled** — `wolfcrypt/src/aes.c:5560-5566`
- [Medium] **aes_no_key_set_test may fail under WOLF_CRYPTO_CB / device builds** — `wolfcrypt/test/test.c:16277-16281`

- [Low] New keyInstalled bitfield inserted mid-struct rather than appended
- [Info] Duplicated keyInstalled guard block repeated across many mode entry points

</details>

Frauschi (changes requested):
## 🐺 Skoll Code Review

**Findings:** 3 total — 1 posted, 3 skipped

### Posted findings
- [High] **FREESCALE_MMCAU / FREESCALE_LTC wc_AesSetKeyLocal never sets keyInstalled, breaking guarded modes** — `wolfcrypt/src/aes.c:5062 (FREESCALE_MMCAU), wolfcrypt/src/aes.c:4600 (FREESCALE_LTC)`

- [Medium] New keyInstalled bitfield inserted mid-struct breaks ABI and packs poorly
- [Low] Inconsistent sz==0 handling between generic and PIC32 CBC guards
- [High] FREESCALE_MMCAU / FREESCALE_LTC wc_AesSetKeyLocal never sets keyInstalled, breaking guarded modes

</details>

julek-wolfssl:
retest this please

Frauschi (changes requested):
## 🐺 Skoll Code Review

### Posted findings
- [High] **FREESCALE_MMCAU and FREESCALE_LTC key-schedule paths never set keyInstalled, breaking their software AES modes** — `wolfcrypt/src/aes.c:5089, wolfcrypt/src/aes.c:4627`

- [Medium] New keyInstalled bitfield inserted mid-struct instead of at end of struct Aes

</details>

Frauschi (dismissed):
LGTM

SparkiDev (dismissed):

Frauschi (dismissed):

wolfssl/wolfcrypt/aes.h
```diff
@@ -294,6 +294,12 @@ struct Aes {
 #endif
     int     keylen;
```
dgarske, line 356: **🟠 [Medium] keyInstalled bitfield inserted mid-struct in the public struct Aes: ABI break, and +4 bytes on builds where ALIGN16 is…**

The field is inserted between `int keylen;` (aes.h:295) and `ALIGN16 word32 reg[]` (aes.h:303) rather than appended, shifting the offset of `reg`/`tmp`/`gcm`/`asyncDev` and everything after them. `struct Aes` is a public caller-allocated type (apps declare `Aes aes;` on the stack), so this breaks binary compatibility with a prebuilt libwolfssl. The assumption that it lands free in ALIGN16 padding only holds for some configs: `WOLFSSL_ALIGN(x)` is a null expansion unless `WOLFSSL_USE_ALIGN` is defined (types.h:1802-1813), which is only auto-defined for AESNI/ARMASM/USE_INTEL_SPEEDUP/Xilinx/no-unaligned-ARM (types.h:1760-1780). On a plain portable-C build (typical embedded, RISC-V, generic Cortex-M) the new `byte:1` forces 3 bytes of padding and grows `sizeof(Aes)` by 4 — measured 280->284 / 248->252 on a gcc x86_64 layout replica. This also contradicts the project convention of appending new members at the end. Separately, `struct Aes` already has a bitfield cluster at aes.h:445-447 (`gcmKeySet:1; nonceSet:1; ctrSet:1;`) that the new bit should have joined so it shares an allocation unit instead of consuming a fresh one.

**Fix:** Move the bit-field to the end of struct Aes (or into the existing gcmKeySet/nonceSet/ctrSet cluster) so no existing member offset changes and the bit packs into trailing padding.

wolfcrypt/src/aes.c
```diff
@@ -7014,6 +7053,10 @@ int wc_AesCbcEncrypt(Aes* aes, byte* out, const byte* in, word32 sz)
 #endif
 
 #if defined(WOLFSSL_RISCV_ASM)
```
dgarske: **🟠 [Medium] RISC-V CBC guard is placed above the DCP and crypto-cb dispatches, contradicting the PR's own stated placement rule**

The PR's stated invariant is that the guard sits after every crypto-cb / hardware-offload dispatch so device-managed keys are unaffected. The RISC-V copies (7056 and 7297) sit inside the `#if defined(WOLFSSL_RISCV_ASM)` block, which is ABOVE the WOLFSSL_IMXRT_DCP dispatch (7067/7308) and the WOLF_CRYPTO_CB dispatch (7073/7314). Today this is masked by a pre-existing defect — the RISC-V block returns unconditionally (7064/7305 `return 0;`), making the dispatches below it dead code on RISC-V — so 7056 and 7086 are mutually exclusive rather than duplicated. But the guard is on the wrong side of the dispatch and will silently break device-managed keys the moment that unconditional early return is corrected. Note also that on RISC-V the guard buys nothing today: `wc_CryptoCb_AesSetKey` sets keyInstalled=1 (aes.c:5510) without ever populating `aes->key`, so a device-keyed context passes and `AES_CBC_encrypt_RISCV64` runs on the all-zero schedule anyway.

**Fix:** Move the RISC-V asm block below the DCP and WOLF_CRYPTO_CB dispatches (or move the dispatches above it) so there is a single guard placed after all offload dispatches, instead of two on opposite sides.

wolfcrypt/src/aes.c
```diff
@@ -7000,6 +7037,8 @@ int wc_AesCbcEncrypt(Aes* aes, byte* out, const byte* in, word32 sz)
         }
 
         if (sz == 0) {
```
dgarske, line 7275: **🟠 [Medium] sz == 0 handling is now inconsistent across modes: CBC returns 0, CFB/CFB1/CFB8/OFB/CTR/PIC32MZ-CBC return BAD_FUNC_ARG**

The PR added an explicit comment in the generic CBC functions explaining that `if (sz == 0) return 0;` must stay ABOVE the guard ("A missing key is only reported when there is work"), but the opposite ordering was used everywhere else: the guard is BEFORE the sz==0 check in AesCfbEncrypt_C (16166 vs 16170), AesCfbDecrypt_C (16253 vs 16257), wc_AesFeedbackCFB8 (16485 vs 16489), wc_AesFeedbackCFB1 (16549 vs 16553), AesOfbCrypt_C (16711 vs 16715) and the PIC32MZ CBC arm (6891 vs 6896). `wc_AesCtrEncrypt` has no sz==0 check at all, so it also returns BAD_FUNC_ARG. Net effect: the *same public API* `wc_AesCbcEncrypt` returns 0 for an unkeyed sz==0 call on a generic build but BAD_FUNC_ARG on PIC32MZ. Real callers pass 0: `wolfSSL_EVP_Cipher` (evp.c:8846) forwards len==0 with non-NULL src/dst straight to wc_AesCfbEncrypt/wc_AesOfbEncrypt/wc_AesCtrEncrypt, and `AesCtsUpdate` (aes.c:20813) calls wc_AesCbcEncrypt with blocks==0. Combined with the WOLF_CRYPTO_CB_SETKEY finding above (which leaves keyInstalled==0 on a legitimately keyed context), and given wc_AesCfb1/8 have no crypto-cb dispatch, these zero-length calls now fail.

**Fix:** Pick one policy and apply it everywhere. Simplest is to move the guard below `if (sz == 0) return 0;` in the five CFB/OFB helpers and the PIC32MZ CBC arm, and add an early sz==0 return to wc_AesCtrEncrypt.

---
> **Note:** Referenced line (`wolfcrypt/src/aes.c:7039-7043, 7264-7268, 6891-6897, 6926-6932, 16166-16172, 16253-16259, 16485-16491, 16549-16555, 16711-16717, 7812`) is outside the diff hunk. Comment anchored to nearest changed region.

wolfcrypt/src/aes.c
```diff
@@ -10966,6 +11035,12 @@ int wc_AesGcmEncrypt(Aes* aes, byte* out, const byte* in, word32 sz,
     }
 #endif
 
+    /* Software/HW key schedule (and hash subkey H) required from here on. */
+    if (aes->keyInstalled == 0) {
```
dgarske: **🟠 [Medium] New guards return BAD_FUNC_ARG where wolfCrypt already defines MISSING_KEY for exactly this condition**

error-crypt.h:279 defines `MISSING_KEY = -278, /* Key was not set */`, and the AES-GCM streaming API already uses it for precisely this case (aes.c:14172-14174, 14259-14261, 14334, 14409). All ~24 new guards return `BAD_FUNC_ARG` instead. The result is that `wc_AesGcmEncrypt()` and `wc_AesGcmEncryptUpdate()` now report the same condition with two different codes within the same API family, and `BAD_FUNC_ARG` is indistinguishable from a genuine NULL/length error, making the new failure hard to diagnose in the field. The PR is consistent with `wc_AesXtsEncrypt`'s pre-existing `keylen == 0` -> `BAD_FUNC_ARG` (aes.c:17688-17691), so the codebase now has three different no-key behaviours; worth a deliberate decision rather than an accident.

**Fix:** Prefer MISSING_KEY (it exists precisely for this), or state in the PR description why BAD_FUNC_ARG was chosen for API compatibility and accept the divergence from the streaming GCM API.

---
> **Note:** Referenced line (`wolfcrypt/src/aes.c:11039-11042 vs 14171-14174`) is outside the diff hunk. Comment anchored to nearest changed region.

dgarske (commented):
## Skoll Code Review

**Findings:** 20 total — 8 posted, 12 skipped
**4 finding(s) not tied to a diff line** (full detail below)

### Posted findings
- [Medium] **keyInstalled bitfield inserted mid-struct in the public struct Aes: ABI break, and +4 bytes on builds where ALIGN16 is a null expansion** — `wolfssl/wolfcrypt/aes.h:295-303`
- [Medium] **sz == 0 handling is now inconsistent across modes: CBC returns 0, CFB/CFB1/CFB8/OFB/CTR/PIC32MZ-CBC return BAD_FUNC_ARG** — `wolfcrypt/src/aes.c:7039-7043, 7264-7268, 6891-6897, 6926-6932, 16166-16172, 16253-16259, 16485-16491, 16549-16555, 16711-16717, 7812`
- [Medium] **New guards return BAD_FUNC_ARG where wolfCrypt already defines MISSING_KEY for exactly this condition** — `wolfcrypt/src/aes.c:11039-11042 vs 14171-14174`
- [Medium] **RISC-V CBC guard is placed above the DCP and crypto-cb dispatches, contradicting the PR's own stated placement rule** — `wolfcrypt/src/aes.c:7053-7089, 7294-7330`

#### wc_AesEncryptDirect / wc_AesDecryptDirect and wc_AesKeyWrap_ex left unguarded — the most direct zero-key bypass
**File:** `wolfcrypt/src/aes.c:6089-6120, 16862-16900, 16964-17037`
**Function:** `wc_AesEncryptDirect / wc_AesDecryptDirect / wc_AesKeyWrap_ex / wc_AesKeyUnWrap_ex`
**Severity:** Medium

`wc_AesEncryptDirect`/`wc_AesDecryptDirect` are public one-block APIs that run `wc_AesEncrypt`/`wc_AesDecrypt` straight against the software schedule with no key check at all. `wc_AesKeyWrap_ex`/`wc_AesKeyUnWrap_ex` take a caller-owned `Aes` and loop on `wc_AesEncryptDirect` (aes.c:16900, 17015) after validating only pointers and sizes — so RFC 3394 key wrapping on an `Aes` that only saw `wc_AesInit()` still silently wraps with the zero key, the identical failure mode the PR cites for GCM. Note `WOLFSSL_AES_DIRECT` is also absent from the new test's outer `#if`, so a DIRECT-only build (e.g. AES-XTS-only) compiles none of the new test at all. For completeness, the AES-GCM *streaming* API is NOT a gap: `wc_AesGcmEncryptUpdate`/`Final` and the decrypt equivalents already check `aes->gcmKeySet` and return `MISSING_KEY` (aes.c:14172, 14259, 14334, 14409).

**Recommendation:** Guard wc_AesKeyWrap_ex/wc_AesKeyUnWrap_ex directly (note `#define wc_AesEncryptDirect(aes, out, in) wc_AesEncrypt(aes, in, out)` at aes.c:6106/6123 means a guard on the wrapper alone will not cover them), add WOLFSSL_AES_DIRECT to the test's outer #if, and add direct-API negative cases.

_Referenced code: `wolfcrypt/src/aes.c:6089-6120, 16862-16900, 16964-17037` (8 lines)_

---

#### keyInstalled semantics are backend-dependent: device-key ports set it while aes-key stays zero, so aes.c's software…
**File:** `wolfcrypt/src/port/kcapi/kcapi_aes.c:208-217; wolfcrypt/src/port/devcrypto/devcrypto_aes.c:123-130`
**Function:** `wc_AesGcmSetKey (kcapi) / wc_AesSetKey (devcrypto)`
**Severity:** Medium

aes.h:297-300 documents the flag as meaning the mode APIs will not run "with the all-zero key schedule left by wc_AesInit". But several of the new assignments do not imply `aes->key` holds anything. kcapi_aes.c:213 sets `keyInstalled` and then copies the key into `aes->devKey` only (line 216); `aes->key` stays zero. KCAPI does not get aes.c's CBC or GCM, but it DOES compile aes.c's generic ECB (`_AesEcbEncrypt`, guard at 15903), CFB/OFB (16166/16253/16711) and CCM (15048/15218) — all of which read `aes->key`. So after `wc_AesGcmSetKey` on a KCAPI build, those paths see "keyed" while the schedule is zero: the guard passes and the software cipher runs against an all-zero key. The same shape applies to devcrypto and to the `WOLFSSL_AFALG_XILINX_AES` config where the key lives in `msgBuf`. The flag conflates "some key material was configured for this backend" with "aes->key holds a usable schedule".

**Recommendation:** Either correct the aes.h comment to state the weaker guarantee, or restrict the flag to setters that populate aes->key and give device-key backends separate state, so the software guards mean what they claim.

_Referenced code: `wolfcrypt/src/port/kcapi/kcapi_aes.c:208-217; wolfcrypt/src/port/devcrypto/devcrypto_aes.c:123-131` (9 lines)_

---

#### wc_AesFree crypto-cb early return skips the ForceZero, leaving keyInstalled == 1 on a freed context
**File:** `wolfcrypt/src/aes.c:15617-15633`
**Function:** `wc_AesFree`
**Severity:** Medium

`wc_AesInit` correctly zeroes the whole struct (aes.c:15525) and the normal `wc_AesFree` tail does `ForceZero(aes, sizeof(Aes))` (aes.c:15703), so the flag is cleared on the ordinary path. But when `WOLF_CRYPTO_CB && WOLF_CRYPTO_CB_FREE` are defined and the callback handles the free, `wc_AesFree` returns at line 15630 without ever reaching the ForceZero, so `keyInstalled` stays 1. Reuse of that context (a common pattern with a static/stack `Aes`) now passes the new guard and proceeds into the software path with a stale or freed key schedule — the exact case the guard is meant to catch. The adjacent code already clears `aes->devCtx = NULL` on this path, so the fix is a one-liner in the same block.

**Recommendation:** Add `aes->keyInstalled = 0;` alongside the existing `aes->devCtx = NULL;` before the early return, so the flag is always cleared on free.

_Referenced code: `wolfcrypt/src/aes.c:15617-15623` (7 lines)_

---

#### Renesas FSPSM CRYPTONLY wc_AesSetKey is the one port key-install function the PR did not update
**File:** `wolfcrypt/src/port/Renesas/renesas_fspsm_aes.c:928-957`
**Function:** `wc_AesSetKey (WOLFSSL_RENESAS_FSPSM_CRYPTONLY)`
**Severity:** Medium

The PR added `aes->keyInstalled = 1` to six port-provided key setters (caam:92, ti:101, silabs:92, devcrypto:127, kcapi:213, af_alg:154 and :566) but missed this one. aes.c:4874-4876 delegates `wc_AesSetKey` to this file for `WOLFSSL_RENESAS_FSPSM_CRYPTONLY`, and aes.c:8458 makes `wc_AesGcmSetKey` return immediately after calling it — so a fully keyed FSPSM `Aes` always has `keyInstalled == 0` while all of its mode functions (CBC/CCM/ECB/CFB/OFB) come from the guarded aes.c. The main paths survive today only by accident: this port mandates `devId != INVALID_DEVID` (renesas_fspsm_aes.c:939-941) and the guards sit after the WOLF_CRYPTO_CB dispatch, so the callback intercepts first. But `wc_AesFeedbackCFB8` (aes.c:16485) and `wc_AesFeedbackCFB1` (aes.c:16549) have no crypto-cb dispatch and already hard-fail with BAD_FUNC_ARG on this backend.

**Recommendation:** Add the assignment next to `aes->ctx.keySize = keylen;` with the same comment used in the other six ports.

_Referenced code: `wolfcrypt/src/port/Renesas/renesas_fspsm_aes.c:928-933` (6 lines)_

---

### Skipped findings
- [High] ~~WOLF_CRYPTO_CB_SETKEY success path returns 0 without setting keyInstalled, breaking every subsequent software-fallback mode call~~
- [High] ~~New aes_no_key_set_test() will fail wolfcrypt_test on every hardware AES backend, including AF_ALG which the PR explicitly addressed~~
- [High] ~~keyInstalled is set before code that can still fail, so a FAILED wc_AesSetKey leaves the context marked as keyed~~
- [High] ~~TI port's wc_AesInit does not zero the Aes struct, so the new guards read an indeterminate keyInstalled bit~~
- [Medium] ~~F-6151 hardening is incomplete: ~10 aes.c hardware arms and 7 ports still run modes against the all-zero key schedule~~
- [Medium] ~~aes_no_key_set_test has no positive counterpart asserting the same calls succeed once a key is installed~~
- [Low] ~~Five added comment lines exceed the 80-column limit while the identical comment was wrapped correctly in two other files~~
- [Low] ~~Test comment and PR description describe a keylen == 0 check, but the implementation is a new keyInstalled bitfield~~
- [Low] ~~PSOC6 wc_AesSetKey wrapper: the added 'aes != NULL' test is dead code and reads as a backwards null check~~
- [Low] ~~AF_ALG wc_AesSetup guard rejects silently with no WOLFSSL_MSG, unlike every other guard in the PR~~
- [Low] ~~New aes_no_key_set_test call is placed outside the #ifndef HAVE_RENESAS_SYNC block that excludes the analogous aes_key_size_test~~
- [Low] ~~rsa_pss_test adds a second WC_MAX_DIGEST_SIZE stack buffer that is avoidable~~

dgarske (changes requested):
https://github.com/wolfSSL/wolfssl/pull/10762#pullrequestreview-4849120952

Also should we consider a new build option to turn off this check? I check imagine some edge cases where this new keyInstalled = 1 is not getting set.

dgarske (changes requested):
This PR could break many platforms, so might need a way to "opt out"
STM32 for example:
```
AES      test failed!
 error L=16590
 [fiducial line numbers: 13144 37941 63673 82657]
```

dgarske:
Jenkins retest this please

dgarske:
Jenkins retest this please

wolfcrypt/src/port/af_alg/afalg_aes.c
```diff
@@ -60,6 +60,10 @@ static int wc_AesSetup(Aes* aes, const char* type, const char* name, int ivSz, i
     byte* key = (byte*)aes->key;
 #endif
 
+    if (!WC_AES_KEY_IS_SET(aes)) {
```
dgarske, line 63: AF_ALG backend: MISSING_KEY guard is unreachable on CBC-decrypt, ECB-decrypt, DecryptDirect and GCM-encrypt, so aes_no_key_set_test fails on --enable-afalg

Recommendation: Either move the guard to the top of each AF_ALG entry point (preferred — the PR already committed to supporting this backend), or add `defined(WOLFSSL_AFALG) || defined(WOLFSSL_AFALG_XILINX_AES)` to `WC_AES_KEY_SET_CHECK_UNSUPPORTED` and drop the `wc_AesSetup` guard entirely. Confirm with an `--enable-afalg` build of `wolfcrypt/test`.

wolfcrypt/src/aes.c
dgarske, line 5707: Also needs `aes->keyInstalled = 1;` I think?

wolfcrypt/src/aes.c
```diff
@@ -5637,6 +5662,7 @@ static void AesSetKey_C(Aes* aes, const byte* key, word32 keySz, int dir)
 
         aes->keylen = (int)keylen;
         aes->rounds = (keylen/4) + 6;
+        aes->keyInstalled = 1;
```
dgarske: Move the assignment to the success return paths (or add `aes->keyInstalled = 0;` before each of the error returns listed above), and do the same in the nrf51 arm at `aes.c:4708`.

wolfcrypt/src/port/caam/caam_aes.c
```diff
@@ -88,6 +88,8 @@ int  wc_AesSetKey(Aes* aes, const byte* key, word32 len,
         default:
             return BAD_FUNC_ARG;
     }
+    /* Mark key installed so the shared aes.c mode guards accept this context. */
+    aes->keyInstalled = 1;
```
dgarske, line 92: MEDIUM wc_SECO_AesSetKeyID() installs a key without setting keyInstalled [SUGGEST] (api)
  File:       wolfcrypt/src/port/caam/wolfcaam_seco.c:541-545
  Function:   wc_SECO_AesSetKeyID

Recommendation: Set `keyInstalled` here so both SECO key-install routes agree. Add the NULL check while you are in the function.

wolfcrypt/src/aes.c
dgarske, line 16207: Recommendation: Set `keyInstalled` in `wc_AesInit_Id`/`wc_AesInit_Label`, or decide explicitly that key-by-id contexts must still be rejected by the software-only entry points and say so in the PR/migration notes.

dgarske (changes requested):

dgarske (dismissed):
Approving, however I do think this change is going to cause issues for some users and will produce support tickets. I am not convinced this AES key installed is needed.

julek-wolfssl:
@dgarske hopefully users who are using the API correctly are unaffected and it will only affect users with incorrect configurations.

philljj:
Retest this please.

no logs.

philljj:
Retest this please.

(no logs, I'm guessing from Monday's github disruption.)

philljj:
tests good with multi-test:
- linuxkm-aesni-insmod-kmemleak
- linuxkm-fips-dev-module-update-fips-hash-intelasm-LKCAPI-insmod
- quantum-safe-wolfssl-all-crypto-only-intelasm-sp-asm-linuxkm-insmod

philljj (approved):

PR #10765, opened 2026-06-23
Xilinx Versal Gen2 ASU port: TRNG, hashes, HMAC offload

night1rider:
Initial porting for Xilinx Versal Gen 2 ASU support, will need #10750


wolfcrypt/src/port/xilinx/versal_gen2_asu/asu_rng.c
```diff
@@ -0,0 +1,133 @@
...
+{
+    if (info == NULL) {
+        return BAD_FUNC_ARG;
+    }
+
+    switch (info->algo_type) {
+        case WC_ALGO_TYPE_SEED:
+            return wc_AsuRngSeed(info);
+        case WC_ALGO_TYPE_RNG:
+            return wc_AsuRngGenerate(info);
```
JacobBarthelmeh, line 127: The TRNG used as a seed is okay, but if the hardware is not a Hash-DRBG implementation I don't think it should directly replace the software Hash-DRBG.

wolfcrypt/src/port/xilinx/versal_gen2_asu/asu_util.c
```diff
@@ -0,0 +1,337 @@
...
+    byte*  pt;
+
+    if (keep == NULL || (in == NULL && inSz > 0)) {
+        return BAD_FUNC_ARG;
+    }
+    if (inSz == 0) {
+        return 0;
+    }
+    /* Reject an accumulated length that would overflow word32. */
+    if (inSz > (0xFFFFFFFFU - keep->used)) {
```
JacobBarthelmeh: Use the macro WOLFSSL_MAX_32BIT

wolfcrypt/src/port/xilinx/versal_gen2_asu/asu_util.c
```diff
@@ -0,0 +1,337 @@
...
+    params->AdditionalStatus = (u32)XST_FAILURE;
+}
+
+word32 wc_AsuWaitDone(AsuWait* wait)
+{
+    /* Unbounded spin: wait is stack-allocated in wc_AsuTransactOne and its address
+     * is the ASU completion ISR's callback reference, so returning before Done is
+     * set (e.g. on a timeout) would let a late ISR write into a reclaimed frame. A
+     * wedged ASU therefore hangs here by design rather than corrupting memory. */
+    while (wait->Done == 0) {
```
JacobBarthelmeh, line 73: We should follow up with ways to relax the loop and avoid taxing the CPU while waiting.

wolfcrypt/src/port/xilinx/versal_gen2_asu/asu_rng.c
night1rider: Updated to make it a it a opt in path, and that path is off by default.

There is a Hash DRGB on the ASU, but you cannot use it and the TRNG in the same firmware build. It also requires a different API.

wolfcrypt/src/port/xilinx/versal_gen2_asu/asu_util.c
night1rider: Updated

wolfcrypt/src/port/xilinx/versal_gen2_asu/asu_util.c
night1rider: Updated for the default path to call `Xil_WaitForEvent`, unless a `XYEILD` is defined for now.

JacobBarthelmeh (changes requested):
Great additions! Please resolve the remaining CI tests before this can be merged.

night1rider:
jenkins retest this please

JacobBarthelmeh (approved):

PR #10768, opened 2026-06-24
Add public alt-name list APIs

Frauschi:
Promotes wolfSSL's internal SAN-list helpers to a public `wc_` API so apps can build SAN entries, encode them, and attach them to a `Cert` before signing without using internal functions.

### New functions

| Function | Purpose |
|---|---|
| `wc_SetDNSEntry()` | Append a typed alt-name entry (DNS, IP, RFC822, URI, …) to a `DNS_entry` list. |
| `wc_FlattenAltNames()` | Encode a list into a DER `GeneralNames` SEQUENCE (general buffer primitive). |
| `wc_SetAltNamesFromList()` | Encode a list straight into `cert->altNames`/`altNamesSz` - mirrors `wc_SetAltNamesBuffer`. |

### Notes

- Prototypes live in `asn.h` (they use the `DNS_entry` type) and share the `WOLFSSL_ASN_API` export gate (`WOLFSSL_PUBLIC_ASN`/`OPENSSL_EXTRA`/…); doxygen documents this.
- Guarded by `WOLFSSL_CERT_GEN && WOLFSSL_ALT_NAMES`.

### Testing

Adds `flattenAltNames_test` covering success, NULL-list, NULL-output, `BUFFER_E`, and the `Cert` paths. Builds clean and passes under `--enable-all`.

Used in wolfCert.

### DIFF
--- a/wolfcrypt/src/asn.c
+++ b/wolfcrypt/src/asn.c
@@ -14520,6 +14520,46 @@ static int SetDNSEntry(void* heap, const char* str, int strLen,
 
     return ret;
 }
+
+/* Public wrapper for SetDNSEntry(): allocate an alt-name entry that copies the
+ * given name, set its type/length, and append it to the linked list. The list
+ * is freed with FreeAltNames() and can be flattened with wc_FlattenAltNames().
+ * Additionally requires WOLFSSL_ASN_TEMPLATE (its internal SetDNSEntry/
+ * AddDNSEntryToList helpers are template-only), on top of the
+ * WOLFSSL_CERT_GEN && WOLFSSL_ALT_NAMES gating of its companions
+ * wc_FlattenAltNames()/wc_SetAltNamesFromList(). Because this builder is the
+ * more restrictive of the set, a DNS_entry list built here can always be
+ * encoded by a public API in the same build. */
+#if defined(WOLFSSL_CERT_GEN) && defined(WOLFSSL_ALT_NAMES)
+int wc_SetDNSEntry(void* heap, const char* str, int strLen, int type,
+                   DNS_entry** entries)
+{
+    /* Validate caller-supplied arguments at the public boundary: a negative
+     * strLen would cast to a huge size_t in the internal allocate/copy, a NULL
+     * entries list head would be dereferenced, and a NULL str cannot be
+     * copied. */
+    if (str == NULL || entries == NULL || strLen < 0) {
+        return BAD_FUNC_ARG;
+    }
+
+    /* Reject unsupported GeneralName types so a public caller cannot OR an
+     * out-of-range value into the context tag emitted by FlattenAltNames(). */
+    switch (type) {
+        case ASN_OTHER_TYPE:
+        case ASN_RFC822_TYPE:
+        case ASN_DNS_TYPE:
+        case ASN_DIR_TYPE:
+        case ASN_URI_TYPE:
+        case ASN_IP_TYPE:
+        case ASN_RID_TYPE:
+            break;
+        default:
+            return BAD_FUNC_ARG;
+    }
+
+    return SetDNSEntry(heap, str, strLen, type, entries);
+}
+#endif /* WOLFSSL_CERT_GEN && WOLFSSL_ALT_NAMES */
 #endif
 
 /* Set the details of a subject name component into a certificate.
@@ -26960,6 +27000,15 @@ int FlattenAltNames(byte* output, word32 outputSz, const DNS_entry* names)
     return (int)idx;
 }
 
+/* Public wrapper for FlattenAltNames(): encode a linked list of alt-name
+ * entries into the DER GeneralNames SEQUENCE used as the subjectAltName
+ * extension value. Returns the encoded length, 0 for a NULL list, or a
+ * negative error code. */
+int wc_FlattenAltNames(byte* output, word32 outputSz, const DNS_entry* names)
+{
+    return FlattenAltNames(output, outputSz, names);
+}
+
 #endif /* WOLFSSL_ALT_NAMES */
 #endif /* WOLFSSL_CERT_GEN */
 
@@ -31505,6 +31554,26 @@ int wc_SetAltNamesBuffer(Cert* cert, const byte* der, int derSz)
     }
 
     return(ret);
+}
+
+/* Set cert alt names from a linked list of alt-name entries (e.g. built with
+ * wc_SetDNSEntry()). Encodes the list into cert->altNames and stores the
+ * length in cert->altNamesSz. Returns 0 on success or a negative error code. */
+int wc_SetAltNamesFromList(Cert* cert, const DNS_entry* names)
+{
+    int ret;
+
+    if (cert == NULL) {
+        return BAD_FUNC_ARG;
+    }
+
+    ret = FlattenAltNames(cert->altNames, sizeof(cert->altNames), names);
+    if (ret < 0) {
+        return ret;
+    }
+
+    cert->altNamesSz = ret;
+    return 0;
 }
 
 /* Set cert dates from DER buffer */
--- a/wolfcrypt/test/test.c
+++ b/wolfcrypt/test/test.c
@@ -1035,6 +1035,12 @@ WOLFSSL_TEST_SUBROUTINE wc_test_ret_t  certext_test(void);
     defined(WOLFSSL_CERT_EXT) && defined(WOLFSSL_CERT_GEN)
 WOLFSSL_TEST_SUBROUTINE wc_test_ret_t decodedCertCache_test(void);
 #endif
+#if defined(WOLFSSL_CERT_GEN) && defined(WOLFSSL_ALT_NAMES) && \
+    defined(WOLFSSL_ASN_TEMPLATE) && \
+    (defined(WOLFSSL_TEST_CERT) || defined(OPENSSL_EXTRA) || \
+     defined(OPENSSL_EXTRA_X509_SMALL) || defined(WOLFSSL_PUBLIC_ASN))
+WOLFSSL_TEST_SUBROUTINE wc_test_ret_t flattenAltNames_test(void);
+#endif
 WOLFSSL_TEST_SUBROUTINE wc_test_ret_t memory_test(void);
 #if defined(WOLFSSL_PUBLIC_MP) && \
     ((defined(WOLFSSL_SP_MATH_ALL) && !defined(WOLFSSL_RSA_VERIFY_ONLY)) || \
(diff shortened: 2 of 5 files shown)


dgarske (dismissed):

dgarske (commented):
## Skoll Code Review

### Posted findings
- [Medium] **New public wc_SetDNSEntry() does no argument validation, unlike its sibling wrappers** — `wolfcrypt/src/asn.c:14528-14532`
- [Medium] **Test does not cover documented NULL-names path of wc_SetAltNamesFromList or wc_SetDNSEntry error paths** — `wolfcrypt/test/test.c:26581-26606`
- [Low] **API availability mismatch: wc_SetDNSEntry can build a list that cannot be flattened in some build configs** — `wolfssl/wolfcrypt/asn.h:2393-2397`

wolfcrypt/src/asn.c
```diff
@@ -14520,6 +14520,16 @@ static int SetDNSEntry(void* heap, const char* str, int strLen,
 
     return ret;
 }
+
+/* Public wrapper for SetDNSEntry(): allocate an alt-name entry that copies the
+ * given name, set its type/length, and append it to the linked list. The list
+ * is freed with FreeAltNames() and can be flattened with
+ * wc_FlattenAltNames(). */
+int wc_SetDNSEntry(void* heap, const char* str, int strLen, int type,
```
dgarske, line 14534: **🟠 [Medium] New public wc_SetDNSEntry() does no argument validation, unlike its sibling wrappers**

This PR promotes the previously `static` internal `SetDNSEntry()` to a public API via a bare pass-through wrapper. The other two wrappers added in this same PR validate their pointer arguments (`wc_SetAltNamesFromList` checks `cert == NULL`, and `wc_FlattenAltNames`/`FlattenAltNames` checks `output == NULL`), but `wc_SetDNSEntry` performs no validation at all. Because it now accepts caller-supplied values from outside the library, malformed inputs reach `SetDNSEntry()` directly: a negative `strLen` flows into `XMALLOC((size_t)strLen + 1, ...)` and `XMEMCPY(dnsEntry_name, str, (size_t)strLen)` where the negative value casts to a huge `size_t` (out-of-bounds copy / crash); a NULL `entries` is dereferenced in `AddDNSEntryToList(*lst...)`; and a NULL `str` with non-zero `strLen` causes a NULL read in `XMEMCPY`. The doxygen for this function does not document any `BAD_FUNC_ARG` return for these cases. wolfSSL convention is for public `wc_` functions to return a non-void error and validate inputs at the boundary.

**Fix:** Add boundary validation (NULL `str`, NULL `entries`, negative `strLen`) in the public wrapper and document the `BAD_FUNC_ARG` return in the doxygen, to be consistent with the other two wrappers added in this PR and with wolfSSL public-API conventions.

wolfcrypt/test/test.c
```diff
@@ -26485,6 +26501,118 @@ WOLFSSL_TEST_SUBROUTINE wc_test_ret_t decodedCertCache_test(void)
...
+            ret = WC_TEST_RET_ENC_EC(len);
+    }
+    /* Output one byte too small is rejected with BUFFER_E. */
+    if (ret == 0) {
+        len = wc_FlattenAltNames(out, (word32)expSz - 1, list);
+        if (len != WC_NO_ERR_TRACE(BUFFER_E))
+            ret = WC_TEST_RET_ENC_EC(len);
+    }
+
+    /* wc_SetAltNamesFromList() encodes the same list straight into a Cert and
```
dgarske, line 26595: **🟠 [Medium] Test does not cover documented NULL-names path of wc_SetAltNamesFromList or wc_SetDNSEntry error paths**

The new test exercises the success paths, NULL-output, BAD_FUNC_ARG, BUFFER_E, and the NULL-cert path well. However, `wc_SetAltNamesFromList()` documents that `names` may be NULL, in which case `cert->altNamesSz` is set to 0 (this is a distinct branch: `FlattenAltNames` returns 0 and the function stores 0). That documented behavior is not asserted by any test. Likewise, none of the `wc_SetDNSEntry()` error returns are exercised. These are reasonable gaps to close while a dedicated test is being added.

**Fix:** Add an assertion that `wc_SetAltNamesFromList(cert, NULL)` returns 0 and sets `cert->altNamesSz == 0`, matching the documented contract. Optionally add a wc_SetDNSEntry error-path case.

wolfssl/wolfcrypt/asn.h
```diff
@@ -2390,6 +2390,11 @@ WOLFSSL_LOCAL int StreamOctetString(const byte* inBuf, word32 inBufSz,
 WOLFSSL_ASN_API void FreeAltNames(DNS_entry* altNames, void* heap);
 WOLFSSL_ASN_API DNS_entry* AltNameNew(void* heap);
 WOLFSSL_ASN_API DNS_entry* AltNameDup(DNS_entry* from, void* heap);
+#if defined(WOLFSSL_ASN_TEMPLATE) && \
```
dgarske: **🔵 [Low] API availability mismatch: wc_SetDNSEntry can build a list that cannot be flattened in some build configs**

`wc_SetDNSEntry` is gated by `WOLFSSL_ASN_TEMPLATE && (WOLFSSL_CERT_GEN || !defined(NO_CERTS))`, while its companions `wc_FlattenAltNames` and `wc_SetAltNamesFromList` are gated by `WOLFSSL_CERT_GEN && WOLFSSL_ALT_NAMES`. In a build with `!NO_CERTS` and `WOLFSSL_ASN_TEMPLATE` but without `WOLFSSL_CERT_GEN`/`WOLFSSL_ALT_NAMES`, a caller can construct a `DNS_entry` list via `wc_SetDNSEntry` (whose doxygen \sa points at `wc_FlattenAltNames`) but has no public function to encode it. This is a benign surface inconsistency, not a build break, but it makes the documented usage flow unavailable in that configuration.

**Fix:** Consider aligning the guards (or documenting the gap) so the wc_SetDNSEntry -> wc_FlattenAltNames flow advertised in the doxygen is available together. Mirroring the internal availability is acceptable, but a short note in the docs would avoid confusion.

wolfcrypt/src/asn.c
Frauschi: fixed

wolfcrypt/test/test.c
Frauschi: fixed

wolfssl/wolfcrypt/asn.h
Frauschi: Fixed by aligning the gating (`wc_SetDnsEntry` is now only available when the companioning `wc_FlattenAltNames` and `wc_SetAltNamesFromList` are also available. Otherwise this function is not relevant anyway).

Frauschi:
Jenkins retest this please

wolfcrypt/src/asn.c
```diff
@@ -14520,6 +14520,27 @@ static int SetDNSEntry(void* heap, const char* str, int strLen,
 
     return ret;
 }
+
+/* Public wrapper for SetDNSEntry(): allocate an alt-name entry that copies the
```
dgarske, line 14524: **🔵 [Low] wc_SetDNSEntry documentation/comment claims gating 'matches' its companions, but it additionally requires…** · _Logic_

Both the inline source comment (asn.c:14524-14528) and the doxygen `\note` (doc/dox_comments/header_files/asn_public.h:1162) claim wc_SetDNSEntry is gated identically to its companions wc_FlattenAltNames()/wc_SetAltNamesFromList(). Confirmed against the code: wc_SetDNSEntry sits inside the outer `#ifdef WOLFSSL_ASN_TEMPLATE` plus `#if defined(WOLFSSL_CERT_GEN) && defined(WOLFSSL_ALT_NAMES)` (asn.c:14529), whereas its companions wc_FlattenAltNames (asn.c:26988) and wc_SetAltNamesFromList (asn.c:31543) are gated by only `WOLFSSL_CERT_GEN && WOLFSSL_ALT_NAMES` and do NOT require WOLFSSL_ASN_TEMPLATE. So in a legacy non-template build (WOLFSSL_ASN_TEMPLATE undefined) with WOLFSSL_CERT_GEN && WOLFSSL_ALT_NAMES, the encode functions exist but the public list-builder wc_SetDNSEntry does not, contradicting the 'always available together' / 'matching its companions' wording. This is a comment/doc-accuracy issue, not a build break: declaration and definition guards match each other per-function, so there is no link error. The follow-on conclusion ('a DNS_entry list built here can always be encoded by a public API in the same build') happens to remain true (the builder's stricter gate implies the looser encoder gate), but the stated premise is wrong. Both contributing modes flagged the same root cause: the security mode rated it Info (Logic); the review mode rated it Low/NIT (style). Stricter severity (Low) retained.

**Fix:** Correct BOTH the inline comment (asn.c:14524-14528) and the asn_public.h doxygen note (asn_public.h:1162) to state that wc_SetDNSEntry additionally requires WOLFSSL_ASN_TEMPLATE (its internal SetDNSEntry/AddDNSEntryToList helpers are template-only). Clarify that because this builder is the more restrictive of the set, a DNS_entry list built here can always be encoded by a public API in the same build, but in non-template builds the public list can only be sourced from a parsed DecodedCert.altNames.

wolfcrypt/src/asn.c
```diff
@@ -14520,6 +14520,27 @@ static int SetDNSEntry(void* heap, const char* str, int strLen,
...
+ * given name, set its type/length, and append it to the linked list. The list
+ * is freed with FreeAltNames() and can be flattened with wc_FlattenAltNames().
+ * Gated the same as its companions wc_FlattenAltNames()/wc_SetAltNamesFromList()
+ * so the build-list -> encode flow is always available together. */
+#if defined(WOLFSSL_CERT_GEN) && defined(WOLFSSL_ALT_NAMES)
+int wc_SetDNSEntry(void* heap, const char* str, int strLen, int type,
```
dgarske, line 14534: **⚪ [Info] New public wc_SetDNSEntry() does not validate the GeneralName type argument** · _Security_

The new public wrapper `wc_SetDNSEntry()` validates `str`, `entries`, and `strLen`, but passes the caller-supplied `type` through to the internal `SetDNSEntry()` unchecked. `type` is stored verbatim in `DNS_entry.type` (an `int`) and later flows into the DER tag byte in `FlattenAltNames()`: `output[idx] = (byte)(ASN_CONTEXT_SPECIFIC | curName->type);`. Because this is now an external boundary (it was previously only reachable from internal call sites that always pass valid `ASN_*_TYPE` values), a caller can supply an out-of-range value (e.g. >0x1F or negative) which would OR into the tag class/constructed bits and produce a malformed GeneralNames encoding. Impact is limited to the caller corrupting a certificate they are building/signing themselves (garbage-in/garbage-out); there is no memory-safety consequence and pre-existing alt-name APIs are equally permissive, so this is a defense-in-depth note for the new public surface rather than a vulnerability.

**Fix:** Optionally reject unsupported GeneralName types at the public boundary (e.g. validate against the documented set: ASN_DNS_TYPE, ASN_IP_TYPE, ASN_RFC822_TYPE, ASN_URI_TYPE, ASN_DIR_TYPE, ASN_OTHER_TYPE, ASN_RID_TYPE) and return BAD_FUNC_ARG otherwise, so a public caller cannot emit an out-of-range context tag.

dgarske (commented):
## Skoll Multi-Scan Review

### Posted findings
- [Low] [review+review-security] **wc_SetDNSEntry documentation/comment claims gating 'matches' its companions, but it additionally requires WOLFSSL_ASN_TEMPLATE** — `wolfcrypt/src/asn.c:14524`
- [Info] [review-security] **New public wc_SetDNSEntry() does not validate the GeneralName type argument** — `wolfcrypt/src/asn.c:14530`

### Skipped findings
- [Medium] ~~New public cert setter wc_SetAltNamesFromList missing WOLFSSL_ABI annotation~~
- [Low] ~~256-byte stack buffer in test without WOLFSSL_SMALL_STACK handling~~

wolfcrypt/src/asn.c
Frauschi: Fixed by adding the check for only allowed values

wolfcrypt/src/asn.c
Frauschi: extended the comments to mark the distinction clearly.

dgarske (approved):

douzzer (approved):

PR #10773, opened 2026-06-24
24062026 Coverity fixes

rlm2002:
# Description

- Logically dead code
  - 561219: remove `XBADFILE` check
  - 561215/561216: remove dead check
- Untrusted loop bound
  - 561218: Add bounds check to test helper
- Overflowed return value
  - 561217: move casts and harden function
-  Unchecked return value
  - 561216/561205: Remove call to `remove(pubFile/privFile)`, add `(void)` cast
- Assign instead of compare
  - 561214: change `ExpectTrue()` to assignment
- Untrusted pointer read
  - 561207: Bound tainted lengths in ECH test helper

# Testing

### DIFF
--- a/src/ocsp.c
+++ b/src/ocsp.c
@@ -1188,10 +1188,6 @@ OcspResponse* wolfSSL_d2i_OCSP_RESPONSE(OcspResponse** response,
 
     if (data == NULL || *data == NULL || len <= 0)
         return NULL;
-    if (*data == NULL)
-        return NULL;
-    if (len <= 0)
-        return NULL;
 
     if (response != NULL)
         resp = *response;
--- a/tests/api.c
+++ b/tests/api.c
@@ -15284,6 +15284,9 @@ static int ech_find_extension(byte* buf, word16* idx_p, word16 extType)
     word16 innerExtLen;
 
     innerExtIdx = ech_seek_extensions(buf + *idx_p, &innerExtLen) + *idx_p;
+    if (innerExtLen > MAX_RECORD_SIZE) {
+        return BAD_FUNC_ARG;
+    }
     idx = innerExtIdx;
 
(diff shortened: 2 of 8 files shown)


src/ocsp.c
```diff
@@ -1190,8 +1190,6 @@ OcspResponse* wolfSSL_d2i_OCSP_RESPONSE(OcspResponse** response,
         return NULL;
     if (*data == NULL)
         return NULL;
```
dgarske: **⚪ [Info] Adjacent dead *data == NULL check left in place after dead-code cleanup**

This PR removes the redundant `if (len <= 0) return NULL;` because line 1189 (`if (data == NULL || *data == NULL || len <= 0) return NULL;`) already covers it. That same combined guard also already covers the immediately-following `if (*data == NULL) return NULL;` on lines 1191-1192, which is therefore unreachable/dead in exactly the same way. This is pre-existing (the diff did not add it), so it is out of strict scope, but since the PR's stated theme is removing logically-dead code, a reviewer would reasonably remove this adjacent duplicate at the same time. No correctness impact either way.

**Fix:** Optionally drop the redundant `if (*data == NULL) return NULL;` for consistency with the dead-code cleanup in this PR. Not blocking; pre-existing and harmless.

dgarske (commented):
## Skoll Code Review

### Posted findings
- [Info] **Adjacent dead *data == NULL check left in place after dead-code cleanup** — `src/ocsp.c:1189-1192`

src/ocsp.c
rlm2002: Fix applied, thanks.

dgarske (approved):

Frauschi (approved):

douzzer (approved):

PR #10775, opened 2026-06-25
Regression testing fixes: ARM/PP64 asm fixes, plus more

SparkiDev:
# Description

ARM64/ARM32/Thumb2:
  - Inline-asm use param names not registers.
  - Return value through first parameter and not first register
  - 32-bit values zero extended when loaded off stack
aes.c: Aligned-accedd GHASH for 32-bit CPUs.
test.c: #ifdef protection update.
sha3.h: make digest and block size constants defines for use in hash.h
sha512.h: internal Transform_Sha512* functions declared when only SHA-384.
sp_int.h: include WOLFSSL_SP_MATH_ALL in SP_INT_BITS selection
memory.c: fix printf format to be compatible with more platforms

# Testing

Regression testing.


SparkiDev:
Jenkins: retest this please

FIPS: failed
generic timedout

dgarske (approved):

SparkiDev:
Code generated with PR:
https://github.com/wolfSSL/scripts/pull/603

PR #10778, opened 2026-06-25
Time-Stamp Protocol (RFC 3161)

SparkiDev:
# Description

Implementation in wolfCrypt
OpenSSL compatibility layer in wolfSSL
Added tests, certificates, examples.

# Testing

Different configuration with --enable-tsp.


SparkiDev:
Jenkins: retest this please

wolfcrypt/src/tsp.c
```diff
@@ -0,0 +1,2370 @@
...
+    static const int monthDays[12] = {
+        0, 31, 59, 90, 120, 151, 181, 212, 243, 273, 304, 334
+    };
+    /* Years contributing a leap day - exclude this year before March. */
+    int y = year - ((mon <= 2) ? 1 : 0);
+    int leapDays = y / 4 - y / 100 + y / 400 -
+                   (1969 / 4 - 1969 / 100 + 1969 / 400);
+
+    return (time_t)(((((time_t)(year - 1970) * 365 + leapDays +
+        monthDays[mon - 1] + day - 1) * 24 + hour) * 60 + min) * 60 + sec);
```
SparkiDev, line 702: fixed

SparkiDev:
DecodeExtKeyUsage fixed

SparkiDev:
Jenkins: retest this please

Visual Studio Build: CRL script
generic aborted

src/ssl_asn1.c
```diff
@@ -1134,6 +1172,44 @@ static int wolfssl_asn1_integer_require_len(WOLFSSL_ASN1_INTEGER* a, int len,
...
+/* Create an ASN.1 INTEGER object holding a big-endian number in DER form.
+ *
+ * The data is the ASN.1 type and length followed by the number as supplied.
+ *
+ * @param [in] val  Big-endian encoding of number.
+ * @param [in] len  Length of number in bytes.
+ * @return  ASN.1 INTEGER object on success.
+ * @return  NULL on failure.
+ */
+static WOLFSSL_ASN1_INTEGER* wolfssl_asn1_integer_new_buf(
```
dgarske, line 1186: **🔵 [Low] wolfssl_asn1_integer_new_buf() builds a negative-looking DER INTEGER when the value's MSB is set** · _Logic_

This new helper backs the OpenSSL-compat getters wolfSSL_TS_REQ_get_nonce(), wolfSSL_TS_TST_INFO_get_serial(), and wolfSSL_TS_TST_INFO_get_nonce(). It builds the ASN1_INTEGER's `data` buffer as [ASN_INTEGER][len][val...] copied straight from the wc magnitude bytes, and leaves `a->negative = 0`. The wc layer strips leading 0x00 pad bytes when decoding INTEGERs (asn.c zeroPadded path), so a positive serial/nonce whose first byte is >= 0x80 (common for random serials and the test nonce 0xc3..., serial 0x9a...) is re-wrapped into a DER INTEGER whose first content octet has the high bit set. By DER that encodes a NEGATIVE integer, yet `a->negative` stays 0 - an internally inconsistent object. Any conformant consumer (i2d_ASN1_INTEGER round-trip, ASN1_INTEGER_cmp, ASN1_INTEGER_get/BN conversion) would interpret it as negative or mismatch. The canonical builder wolfSSL_ASN1_INTEGER_set() in the same file correctly prepends a 0x00 pad in exactly this case, confirming the omission is a defect. No memory-safety or trust impact: these getters are inspection-only views and are not used in the TSP verification/trust path (which operates on the wc structures directly), and wolfSSL_i2d_TS_REQ re-encodes from the wc request, not from this view. The existing tests pass only because they compare raw bytes (`num->data + 2`) rather than exercising the numeric ASN1_INTEGER API on a high-bit value.

**Fix:** Mirror wolfSSL_ASN1_INTEGER_set(): compute pad = ((len > 0) && (val[0] & 0x80)) ? 1 : 0, size/require the buffer for len+pad+hdr, encode SetLength(len+pad, ...), write the 0x00 pad byte (if any) then the value, and keep a->negative = 0. This yields a self-consistent positive DER INTEGER matching OpenSSL output.

wolfcrypt/src/asn_tsp.c
```diff
@@ -0,0 +1,1454 @@
...
+        GetASN_GetConstRef(&dataASN[TSPRESPASN_IDX_STAT_STR], &freeText,
+            &freeTextSz);
+        ret = TspGetPkiFreeTextStr(freeText, freeTextSz, &resp->statusString,
+            &resp->statusStringSz);
+    }
+    if (ret == 0) {
+        /* failInfo is optional. Length includes unused bits byte.
+         * Shift up by the number of bytes not encoded - bit 0 of the named
+         * bit string is the most significant bit of 32. */
+        if (dataASN[TSPRESPASN_IDX_STAT_FAIL].tag != 0) {
```
dgarske: **🔵 [Low] Possible shift-by-32 (undefined behavior) when decoding an empty PKIFailureInfo BIT STRING** · _Undefined Behavior_

The failInfo BIT STRING is read into a `word32` and then left-aligned with `resp->failInfo <<= (8 * (5 - dataASN[TSPRESPASN_IDX_STAT_FAIL].length))`. The design relies on `.length` being the full BIT STRING content length (unused-bits byte + data bytes), so length ranges 1..5. A validly-framed but empty BIT STRING (`03 01 00`, length == 1) is accepted (`tag != 0`), giving a shift count of `8 * (5 - 1) == 32`. Shifting a 32-bit value by 32 is undefined behavior in C (and would be flagged by UBSan). In practice the value is 0 so the result is benign on most targets, but it is still UB on a reachable, attacker-influenced decode path. This edge (present-but-empty failInfo) is not covered by the new decode tests in `test_wc_TspResponse_Decode`.

**Fix:** Guard the shift so the count stays in 0..24 (e.g. only shift when length is in [2,5], or reject an empty/over-long failInfo BIT STRING). Add a decode test for a present-but-empty `03 01 00` failInfo. Suggested fix:
if ((dataASN[TSPRESPASN_IDX_STAT_FAIL].tag != 0) &&
        (dataASN[TSPRESPASN_IDX_STAT_FAIL].length >= 2) &&
        (dataASN[TSPRESPASN_IDX_STAT_FAIL].length <= 5)) {
    resp->failInfo <<=
        (8 * (5 - dataASN[TSPRESPASN_IDX_STAT_FAIL].length));
}

src/ssl.c
```diff
@@ -455,6 +455,9 @@ WC_RNG* wolfssl_make_rng(WC_RNG* rng, int* local)
 #include "src/ssl_asn1.c"
 #endif /* OPENSSL_EXTRA_NO_ASN1 */
 
+#define WOLFSSL_SSL_TSP_INCLUDED
+#include "src/ssl_tsp.c"
```
dgarske, line 459: **🔵 [Low] ssl_tsp.c depends on ssl_asn1.c but the latter is conditionally excluded by OPENSSL_EXTRA_NO_ASN1** · _Build/Portability_

`src/ssl_tsp.c` is included unconditionally, while `src/ssl_asn1.c` is included only `#ifndef OPENSSL_EXTRA_NO_ASN1`. ssl_tsp.c calls the file-static helper `wolfssl_asn1_integer_new_buf()` defined in ssl_asn1.c (and many other ASN.1 compat helpers). A build with `OPENSSL_EXTRA + WOLFSSL_TSP + HAVE_PKCS7 + WOLFSSL_TSP_VERIFIER + OPENSSL_EXTRA_NO_ASN1` would activate the ssl_tsp.c body while ssl_asn1.c is omitted, producing an implicit-declaration/undefined-symbol build failure. This is a narrow, likely-unused configuration, but there is no guard preventing it.

**Fix:** Either add a compile-time guard that documents/forbids the unsupported combination, or gate the TSP OpenSSL-layer guards in ts.h/ssl_tsp.c on `!defined(OPENSSL_EXTRA_NO_ASN1)`. Suggested guard:
#if defined(WOLFSSL_TSP) && defined(OPENSSL_EXTRA) && \
    defined(OPENSSL_EXTRA_NO_ASN1)
    #error "TSP OpenSSL compat layer requires the OpenSSL ASN.1 API (OPENSSL_EXTRA_NO_ASN1 not supported with WOLFSSL_TSP)"
#endif

src/ssl_tsp.c
```diff
@@ -0,0 +1,2370 @@
...
+    WOLFSSL_TS_RESP* tsResp = NULL;
+    WOLFSSL_ASN1_INTEGER* serialInt = NULL;
+    const unsigned char* serial = NULL;
+    word32 serialSz = 0;
+#ifndef NO_ASN_TIME
+    byte genTime[ASN_GENERALIZED_TIME_SIZE];
+#endif
+    const byte* genTimePtr = NULL;
+    word32 genTimeSz = 0;
+    byte reqDer[2048];
```
dgarske: **🔵 [Low] Large 2048-byte on-stack request buffer in new responder path does not use WOLFSSL_SMALL_STACK pattern** · _Security_

`wolfSSL_TS_RESP_create_response()` declares `byte reqDer[2048]` on the stack (plus a `WC_RNG`, `TspRequest`, `TspTstInfo`, `TspResponse`). The wolfSSL coding convention is that stack buffers larger than ~100 bytes should use the `WOLFSSL_SMALL_STACK` heap-allocation pattern so the code remains usable on constrained/embedded targets and under -DWOLFSSL_SMALL_STACK. The companion examples (`examples/tsp/*.c`) already follow a heap-allocation pattern (TSP_DECL/TSP_ALLOC) for big buffers; the library compat layer should be at least as careful. The buffer size is fixed (not attacker-controlled), so there is no overflow - a request larger than 2048 bytes is read truncated and then fails to decode (returns NULL), which is acceptable responder-side behavior. This is a portability/footprint observation only. Both reviewers flagged this location; the review mode rated it severity Low / action NIT, the review-security mode rated it severity Info - the stricter Low view is kept here. `ts_check_data()` also uses a 256-byte `buf[256]`, which is above the threshold.

**Fix:** Wrap `reqDer` (and optionally `buf` in `ts_check_data`) with the WOLFSSL_SMALL_STACK / !WOLFSSL_NO_MALLOC heap-allocation pattern, freeing on all return paths, to match library conventions for >100B stack buffers and keep the responder usable on small-stack targets:
#ifdef WOLFSSL_SMALL_STACK
    byte* reqDer = (byte*)XMALLOC(2048, NULL, DYNAMIC_TYPE_TMP_BUFFER);
    if (reqDer == NULL)
        return NULL;
#else
    byte reqDer[2048];
#endif

dgarske (commented):
## Skoll Multi-Scan Review

### Posted findings
- [Low] [review-security] **wolfssl_asn1_integer_new_buf() builds a negative-looking DER INTEGER when the value's MSB is set** — `src/ssl_asn1.c:1186-1206`
- [Low] [review] **Possible shift-by-32 (undefined behavior) when decoding an empty PKIFailureInfo BIT STRING** — `wolfcrypt/src/asn_tsp.c:wc_TspResponse_Decode`
- [Low] [review] **ssl_tsp.c depends on ssl_asn1.c but the latter is conditionally excluded by OPENSSL_EXTRA_NO_ASN1** — `src/ssl.c:453-459`
- [Low] [review+review-security] **Large 2048-byte on-stack request buffer in new responder path does not use WOLFSSL_SMALL_STACK pattern** — `src/ssl_tsp.c:2243`

src/ssl_tsp.c
```diff
@@ -0,0 +1,2370 @@
+/* ssl_tsp.c
```
dgarske, line 1: Make sure this new file is added to all projects including the Visual Studio one, etc.

dgarske (changes requested):

src/ssl_asn1.c
SparkiDev: fixed

wolfcrypt/src/asn_tsp.c
SparkiDev: fixed

src/ssl.c
SparkiDev: fixed

src/ssl_tsp.c
SparkiDev: fixed

src/ssl_tsp.c
SparkiDev: fixed
ssl_tsp.c and asn_tsp.c are included into other files but tsp.c needed to be added.

SparkiDev:
Jenkins: retest this please

src/ssl_tsp.c
```diff
@@ -0,0 +1,2412 @@
...
+    if (ret == NULL)
+        ret = wolfSSL_TS_VERIFY_CTX_new();
+    if (ret == NULL)
+        return NULL;
+
+    /* Replace - not accumulate - the checks when reusing a context. */
+    ret->flags = WOLFSSL_TS_VFY_ALL_IMPRINT & ~WOLFSSL_TS_VFY_TSA_NAME;
+
+    /* Copy the message imprint's hash of the request. */
+    XFREE(ret->imprint, NULL, DYNAMIC_TYPE_OPENSSL);
```
dgarske, line 1648: **🔵 [Low] TS_REQ_to_TS_VERIFY_CTX leaves imprintSz set when imprint allocation fails on a reused ctx** · _bug_

When filling an existing (caller-owned, non-NULL) ctx, ret->imprintSz is assigned before the XMALLOC. If the allocation fails, the function returns NULL but does not free the caller's ctx (correct, since ctx != NULL) and leaves the ctx in an inconsistent state: imprint == NULL while imprintSz is non-zero. flags has also already been overwritten. It is benign in practice because ts_verify_token guards the imprint check with ctx->imprint == NULL and fails safe, but the dangling imprintSz is untidy if the caller ignores the NULL return and reuses the context. Confirmed against current source at src/ssl_tsp.c:1644-1652.

**Fix:** Set imprintSz only after the allocation succeeds (or reset it to 0 on the failure path) so a reused context is never left with a non-zero imprintSz and a NULL imprint.

src/ssl_tsp.c
```diff
@@ -0,0 +1,2412 @@
...
+/* Hash the data read from a BIO with the token's message imprint algorithm
+ * and compare to the imprint - the TS_VFY_DATA check. The data is hashed
+ * incrementally so it need not be held in memory all at once.
+ *
+ * @param [in] bio  BIO to read the time-stamped data from.
+ * @param [in] tst  Decoded TSTInfo with the expected message imprint.
+ * @return  1 when the hash of the data matches the imprint.
+ * @return  0 otherwise or on error.
+ */
+static int ts_check_data(WOLFSSL_BIO* bio, const TspTstInfo* tst)
```
dgarske, line 1756: **🔵 [Low] Stale 'not supported' comment for TS_VFY_DATA contradicts the implemented behavior** · _style_

The WOLFSSL_TS_VFY_DATA flag is annotated /* not supported */ in the header (wolfssl/openssl/ts.h), but ssl_tsp.c does implement it: ts_verify_token handles the WOLFSSL_TS_VFY_DATA flag by calling ts_check_data, which hashes the BIO data with the token's imprint algorithm and compares it to the imprint, and test_wolfSSL_TS_RESP_verify_data exercises this successfully. The stale comment (and the matching 'Hashing message data is not supported' comments in the OpenSSL tests) is misleading to API users. (Note: WOLFSSL_TS_VFY_TSA_NAME genuinely is unimplemented, so its annotation is accurate.)

**Fix:** Remove the '/* not supported */' annotation from WOLFSSL_TS_VFY_DATA (header is out of primary scope, but the inconsistency is directly tied to the new src implementation) so the documented capability matches the code.

dgarske (commented):
## Skoll Multi-Scan Review

### Posted findings
- [Low] [review] **TS_REQ_to_TS_VERIFY_CTX leaves imprintSz set when imprint allocation fails on a reused ctx** — `src/ssl_tsp.c:1645-1653`
- [Low] [review] **Stale 'not supported' comment for TS_VFY_DATA contradicts the implemented behavior** — `src/ssl_tsp.c:1751-1807`

### Skipped findings
- [Low] ~~TSA responder private key DER freed without zeroization~~

src/ssl_tsp.c
SparkiDev: fixed

src/ssl_tsp.c
SparkiDev: fixed

SparkiDev:
Jenkins: retest this please

FIPS fails
fips-ready fail x2

gencertbuf.pl
```diff
@@ -87,7 +89,11 @@
         [ "./certs/ca-cert.der", "ca_cert_der_2048" ],
         [ "./certs/ca-cert-chain.der", "ca_cert_chain_der" ],
         [ "./certs/server-key.der", "server_key_der_2048" ],
-        [ "./certs/server-cert.der", "server_cert_der_2048" ]
+        [ "./certs/server-cert.der", "server_cert_der_2048" ],
+        [ "./certs/tsa-key.der", "tsa_key_der_2048" ],
+        [ "./certs/tsa-cert.der", "tsa_cert_der_2048" ],
+        [ "./certs/tsa-bad-ku-cert.der", "tsa_bad_ku_cert_der_2048" ],
+        [ "./certs/tsa-extra-eku-cert.der", "tsa_extra_eku_cert_der_2048" ]
```
Frauschi: `tsa-chain-cert.der`, `tsa-chain-key.der`, and `intermediate/ca-int-cert.de`r are missing here (they are already in certs_test.h, however).

src/ssl_tsp.c
```diff
@@ -0,0 +1,2414 @@
...
+    /* Decode the request and start a TSTInfo. */
+    if ((wc_TspRequest_Decode(&req, reqDer, reqSz) != 0) ||
+            (wc_TspTstInfo_Init(&tst) != 0)) {
+        goto cleanup;
+    }
+
+    /* Get a serial number from the callback - a non-negative INTEGER whose
+     * magnitude bytes are the serial number. */
+    serialInt = ctx->serialCb(ctx, ctx->serialCbData);
+    if (ts_asn1_integer_get(serialInt, &serial, &serialSz) != 1)
```
Frauschi, line 2319: `ts_asn1_integer_get` returns the raw DER content, so a serial ≥ `0x80` keeps its leading `0x00` pad. `wc_TspTstInfo_SetFromRequest` doesn't strip it (unlike `wc_TspTstInfo_SetSerial`), and `TspCheckNum` in the encoder then rejects it — `create_response` returns `NULL` for any serial whose top byte has the high bit set. Strip the leading `0x00` here (or route through a stripping setter).

tests/api/test_ossl_tsp.c
```diff
@@ -0,0 +1,1857 @@
...
+ * response creation takes ownership of the returned ASN1_INTEGER. */
+static WOLFSSL_ASN1_INTEGER* test_tsp_serial_cb(WOLFSSL_TS_RESP_CTX* ctx,
+    void* data)
+{
+    WOLFSSL_ASN1_INTEGER* serial = wolfSSL_ASN1_INTEGER_new();
+
+    (void)ctx;
+    (void)data;
+    if (serial != NULL) {
+        wolfSSL_ASN1_INTEGER_set(serial, 0x1234);
```
Frauschi, line 240: This serial has the high bit clear, so its DER encoding has no leading `0x00` pad and never exercises the strip-needed path. Add a positive high-bit case (e.g. `0x80` or a multi-byte random-style serial) so the encode path for padded serials is covered.

Frauschi (changes requested):
Some smaller issues only

src/ssl_tsp.c
SparkiDev: Fixed.
wc_TspTstInfo_SetFromRequest modified to call set serial API and the zeros are stripped.

tests/api/test_ossl_tsp.c
SparkiDev: Tests added.

Frauschi (dismissed):
LGTM!

SparkiDev:
Jenkins: retest this please

Tiny TLS1.3 Tests

Frauschi:
the tiny-tls failures are unrelated and will be fixed by #10835

SparkiDev:
I rebased and it went crazy.

Frauschi:
Jenkins retest this please.

Frauschi (dismissed):

dgarske (commented):
## Skoll Multi-Scan Review

### Posted findings
- [Low] [review] **TS_REQ_set_policy_id accepts empty OID and reports success (inconsistent with set_def_policy)** — `src/ssl_tsp.c:~830-865`
- [Info] [review-security] **wc_TspResponse_Verify with NULL cert and no CM establishes no trust** — `wolfcrypt/src/tsp.c:2162`
- [Info] [review] **TSP_RESP_REQ_DER_SZ macro defined between the doxygen comment and the function signature** — `src/ssl_tsp.c:2238-2254`
- [Info] [review] **wolfssl_asn1_integer_new_buf relies solely on len0 guard, no defensive val NULL check** — `src/ssl_asn1.c:1187-1216`

### Skipped findings
- [Low] ~~TSA private key DER not zeroized before free in responder context~~

src/ssl_tsp.c
```diff
@@ -0,0 +1,2417 @@
...
+ */
+int wolfSSL_TS_REQ_set_policy_id(WOLFSSL_TS_REQ* a,
+    const WOLFSSL_ASN1_OBJECT* policy)
+{
+    const unsigned char* oid;
+    word32 oidSz;
+    word32 idx = 1;
+    int len = 0;
+
+    if ((a == NULL) || (policy == NULL) || (policy->obj == NULL))
```
dgarske, line 591: **🔵 [Low] TS_REQ_set_policy_id accepts empty OID and reports success (inconsistent with set_def_policy)** · _convention_

`wolfSSL_TS_REQ_set_policy_id` validates only `a`, `policy`, and `policy->obj` for NULL. If the supplied `WOLFSSL_ASN1_OBJECT` has `objSz == 0` (or a full-DER OID whose content length is 0, e.g. `{0x06,0x00}`), the strip path yields `oidSz == 0`, the `oidSz > MAX_OID_SZ` check passes, `XMEMCPY` copies zero bytes, `a->req.policySz` is set to 0, and the function returns success (1). Downstream, `policySz == 0` means "no policy present", so the caller believes a policy was set when none was. The analogous responder setter `wolfSSL_TS_RESP_CTX_set_def_policy` explicitly rejects this with `if ((oidSz == 0) || (oidSz > MAX_OID_SZ)) return 0;`. This is a minor API inconsistency, not a memory-safety issue.

**Fix:** Reject `oidSz == 0` in `wolfSSL_TS_REQ_set_policy_id` to match `wolfSSL_TS_RESP_CTX_set_def_policy` so a degenerate/empty OID does not silently report success while setting no policy.

wolfcrypt/src/tsp.c
```diff
@@ -0,0 +1,2373 @@
...
+    }
+    if (ret == 0) {
+        /* The token references the response - not modified by verify. */
+        ret = wc_TspTstInfo_VerifyWithPKCS7(pkcs7, (byte*)resp->token,
+            resp->tokenSz, tstInfo);
+    }
+    /* Establish trust in the signer. The token's signature was verified
+     * against the certificate in the token; trust pins that certificate to a
+     * known one or chains it to a trusted CA in the manager. */
+    if ((ret == 0) && (cm != NULL)) {
```
dgarske, line 2162: **⚪ [Info] wc_TspResponse_Verify with NULL cert and no CM establishes no trust** · _Security_

When `wc_TspResponse_Verify(resp, NULL, 0, tstInfo)` is called with `cert == NULL` (and therefore `cm == NULL`), neither the pin branch (`cert != NULL`) nor the chain branch (`cm != NULL`) executes, so the function returns success as long as the token's own embedded certificate cryptographically verifies the CMS signature and passes the EKU/KU/ESSCertID checks. No trust anchoring is performed - any self-signed certificate carrying the time-stamping EKU is accepted. This is explicitly documented in the function contract ("When the certificate is NULL the token must include the signer's certificate and no trust is established - the caller must trust the signer by other means") and the shipped examples (`examples/tsp/tsp_verify.c`) pass a real trusted certificate, and the OpenSSL compat layer (`ts_verify_token`) additionally fails closed when a store is absent. Reporting only for reviewer awareness so the NULL-cert convenience form is not mistaken for a trust-establishing call by downstream integrators; this is intended behavior, not a defect.

**Fix:** No code change required. Consider strengthening the doc note where the public API is exposed (wolfssl/wolfcrypt/tsp.h) to emphasize that the NULL-cert form performs signature verification only, not trust establishment, mirroring the fail-closed behavior already implemented in the OpenSSL compat wrapper.

src/ssl_tsp.c
```diff
@@ -0,0 +1,2417 @@
...
+ */
+int wolfSSL_TS_RESP_CTX_add_flags(WOLFSSL_TS_RESP_CTX* ctx, int flags)
+{
+    if (ctx == NULL)
+        return 0;
+    ctx->flags |= flags;
+    return 1;
+}
+
+/* Create a time-stamp response for a request.
```
dgarske, line 2244: **⚪ [Info] TSP_RESP_REQ_DER_SZ macro defined between the doxygen comment and the function signature** · _style_

The `#define TSP_RESP_REQ_DER_SZ 2048` is placed after the function's doxygen block (lines 2238-2250) and immediately before the function definition, visually splitting the doc comment from the code it documents. It is functionally correct (defined before first use in the `#else byte reqDer[TSP_RESP_REQ_DER_SZ]` and the `XMALLOC`), but the placement is unusual and slightly harms readability. The `WOLFSSL_SMALL_STACK` heap/stack pattern for `reqDer` itself is correctly applied per project convention.

**Fix:** Relocate the macro definition above the function's doc comment for readability. Non-blocking.

src/ssl_asn1.c
```diff
@@ -1134,6 +1172,50 @@ static int wolfssl_asn1_integer_require_len(WOLFSSL_ASN1_INTEGER* a, int len,
...
+    const unsigned char* val, word32 len)
```
dgarske, line 1187: **⚪ [Info] wolfssl_asn1_integer_new_buf relies solely on len0 guard, no defensive val NULL check** · _style_

`wolfssl_asn1_integer_new_buf` dereferences `val[0]` via `((len > 0) && (val[0] & 0x80))` and later `XMEMCPY(a->data + i, val, len)` without a NULL check on `val`. All current callers in `src/ssl_tsp.c` guarantee `val` is non-NULL (fixed-array `req.nonce`, or explicit `tst.serial != NULL` / `tst.nonce != NULL` checks before the call), and the `len > 0` short-circuit avoids the deref when `len == 0`, so there is no reachable defect today. Flagging only as a robustness note since this is a shared internal helper.

**Fix:** Consider a defensive NULL guard if the helper may be reused by callers that do not pre-validate `val`. Not required given current usage.

dgarske (changes requested):
https://github.com/wolfSSL/wolfssl/pull/10778#pullrequestreview-4621458799

src/ssl_tsp.c
SparkiDev: Fixed

wolfcrypt/src/tsp.c
SparkiDev: Documentation updated

src/ssl_tsp.c
SparkiDev: Macro moved

src/ssl_asn1.c
SparkiDev: Added NULL check

Frauschi (dismissed):

wolfcrypt/src/asn_tsp.c
```diff
@@ -0,0 +1,1454 @@
+/* asn_tsp.c
```
dgarske, line 1: Is there a good reason not to put these new functions directly in asn.c instead of introducing a new file?

wolfssl/wolfcrypt/settings.h
```diff
@@ -3970,6 +3970,30 @@
...
+#endif
+
+#ifdef WOLFSSL_TSP
+    /* Time-Stamp Protocol roles: requester (create requests, verify
+     * responses), verifier (verify tokens/responses only) and responder
+     * (read requests, create responses). When none is chosen, requester and
+     * responder are enabled. */
+    #if !defined(WOLFSSL_TSP_REQUESTER) && !defined(WOLFSSL_TSP_RESPONDER) && \
+        !defined(WOLFSSL_TSP_VERIFIER)
+        #define WOLFSSL_TSP_REQUESTER
```
dgarske, line 3984: Could/should defaults also gate on NO_WOLFSSL_CLIENT or NO_WOLFSSL_SERVER? Not sure if requester/responder is client/server specific.

doc/dox_comments/header_files/tsp.h
```diff
@@ -0,0 +1,1800 @@
+/*!
```
dgarske, line 1: Make sure you run the documentation repo doxy on this branch and make sure there are no issues

wolfcrypt/src/asn_tsp.c
```diff
@@ -0,0 +1,1454 @@
...
+ * @param [in] sz    Length of number in bytes.
+ * @return  0 when the number is encodable.
+ * @return  BAD_FUNC_ARG when the number is empty or has a leading zero byte.
+ */
+#define TspCheckNum(data, sz)                                       \
+    ((((sz) == 0) || (((sz) > 1) && ((data)[0] == 0x00))) ?         \
+        BAD_FUNC_ARG : 0)
+
+#ifdef WOLFSSL_TSP_REQUESTER
+/* Encode a TimeStampReq.
```
dgarske, line 108: Are we duplicating any of this inline documentation here and in the docs headers?

dgarske (changes requested):
This is a huge PR. I've tried to look over it all but very difficult.

wolfcrypt/src/asn_tsp.c
SparkiDev: Because asn.c is way too bog already and they are not used by anything else.
The file is included in asn.c so this is a code readability thing.
Looking to pull out OCSP into asn_ocsp.c

wolfcrypt/src/asn_tsp.c
SparkiDev: This is internal documentation.
The external documentation is for users of the function.
They should have different ways of writing.

doc/dox_comments/header_files/tsp.h
SparkiDev: Doxygen generates.

wolfssl/wolfcrypt/settings.h
SparkiDev: NO_WOLFSSL_CLIENT/SERVER is for SSL protocol - in my mind anyway.
TSP is outside TLS.

SparkiDev:
Jenkins: retest this please

src/ssl_asn1.c
```diff
@@ -1134,6 +1172,58 @@ static int wolfssl_asn1_integer_require_len(WOLFSSL_ASN1_INTEGER* a, int len,
...
+    word32 i = 0;
+
+    /* Defensive: all current callers pass a non-NULL magnitude, but this is a
+     * shared helper that dereferences val below. */
+    if (val == NULL)
+        return NULL;
+
+    /* Pad with a leading 0x00 when the top bit is set so the DER INTEGER stays
+     * positive - the wc layer supplies the magnitude without this pad. */
+    pad = ((len > 0) && (val[0] & 0x80)) ? 1 : 0;
```
dgarske: **🔵 [Low] wolfssl_asn1_integer_new_buf emits a non-canonical zero-length DER INTEGER (02 00) when len==0** · _bug_

With `len==0` (val non-NULL) the helper builds `02 00` (a->length==2), which is not valid DER -- the canonical encoding of value 0 is `02 01 00`. In practice this is unreachable: `wolfSSL_TS_REQ_get_nonce` returns early on `nonceSz==0`, and the TSTInfo serial/nonce getters only call the helper when the source pointer is non-NULL (which implies non-zero size for a validly decoded structure). The high-bit pad, `hdrSz`, `require_len` sizing, and `a->length = len + i` are all correct for `len > 0`.

**Fix:** Add an explicit `len == 0` guard so the shared helper's contract is robust independent of caller guards.

wolfcrypt/src/tsp.c
```diff
@@ -0,0 +1,2378 @@
...
+             (XMEMCMP(pkcs7->verifyCert, cert, certSz) != 0))) {
+        /* Pin: the signer must be the given trusted certificate. */
+        WOLFSSL_MSG("TSP signer is not the trusted TSA");
+        ret = TSP_VERIFY_E;
+    }
+    /* tstInfo references pkcs7->content, which wc_PKCS7_Free releases - the
+     * verify may copy the eContent into PKCS7-owned memory. Re-point the
+     * references into the caller's token, which holds the same TSTInfo DER,
+     * so tstInfo stays valid after this function returns. */
+    if ((ret == 0) && (tstInfo != NULL) && (pkcs7->contentSz > 0)) {
```
dgarske, line 2188: **⚪ [Info] TspResponse_Verify relies on an unenforced 'ret==0 implies contentSz0' invariant to avoid handing back dangling pointers** · _bug_

On success, the pointer fields of `*tstInfo` (policy, serial, genTime, nonce, tsa) reference `pkcs7->content`, which `wc_PKCS7_Free()` releases at the end of the function. The loop that rebases those pointers into the caller-owned `resp->token` is guarded by `if ((ret == 0) && (tstInfo != NULL) && (pkcs7->contentSz > 0))`, and the error-clearing `XMEMSET` at the bottom only runs when `ret != 0`. If `contentSz == 0` were ever true with `ret == 0`, rebasing is skipped AND the clear is skipped, so the caller receives pointers into freed memory. This is currently unreachable (a zero-length content fails `wc_TspTstInfo_VerifyWithPKCS7`, forcing `ret != 0`), so it is a latent structural dependency rather than a live bug.

**Fix:** Make the invariant explicit: treat `ret==0 && tstInfo && contentSz==0` as an error (which then triggers the existing XMEMSET), so a future change to the verify path cannot silently return dangling pointers.

wolfcrypt/src/asn_tsp.c
dgarske: **⚪ [Info] failInfo BIT STRING left-shift depends on a decode invariant in a separate function** · _Logic_

The new `wc_TspResponse_Decode` shifts the decoded failInfo word by `8 * (5 - length)`, where `length` is the BIT STRING content length (including the unused-bits byte). If `length` were ever 1 (empty BIT STRING) or >5 the shift amount would be 32 or negative -- undefined behavior for a `word32`. This is currently NOT reachable: `GetASN_StoreData` (asn.c:1523) rejects a post-unused-byte data length of 0 or >4 with `ASN_PARSE_E`, so `length` is always in [2,5] when the shift executes, giving a shift in [0,24]. Both boundary inputs are covered by tests (`emptyFailDer` and `longFailDer` in test_wc_TspResponse_Decode both expect ASN_PARSE_E). This is a safe-as-written observation, not a vulnerability -- noting the tight coupling between the shift and the `GetASN_StoreData` length check for future maintainers.

**Fix:** Optional hardening: bound the shift explicitly, e.g. only shift when `length` is in [2,5], so correctness does not silently depend on GetASN_StoreData's range check. No change required for current correctness.

wolfssl/wolfcrypt/asn.h
```diff
@@ -1855,6 +1908,8 @@ struct DecodedCert {
     byte    policyConstSkip;         /* Policy Constraints skip certs value */
     word16  extKeyUsage;             /* Key usage bitfield               */
     byte    extExtKeyUsage;          /* Extended Key usage bitfield      */
+    word32  extExtKeyUsageOidCnt;    /* Number of EKU KeyPurposeIds seen,
```
dgarske: **⚪ [Info] New DecodedCert member inserted mid-struct rather than at the end** · _Blast Radius_

The new `word32 extExtKeyUsageOidCnt;` field is inserted between `extExtKeyUsage` and the `#ifdef WOLFSSL_WOLFSSH extExtKeyUsageSsh` member, i.e. mid-struct, which shifts the offsets of all subsequent DecodedCert members and changes the internal struct ABI across versions. This is an internal (non-public) parse structure and it is zero-initialized by InitDecodedCert (asn.c:12498), so there is no uninitialized-read risk and no functional defect. Flagged only as an ABI/convention note (project convention favors appending new members at the end of the struct).

**Fix:** Consider appending the new member at the end of DecodedCert to preserve member offsets for out-of-tree code that may depend on the internal layout. No functional change needed.

src/ssl_tsp.c
```diff
@@ -0,0 +1,2417 @@
...
+#ifdef OPENSSL_EXTRA_NO_ASN1
+#error "TSP OpenSSL compat layer requires the OpenSSL ASN.1 API \
+(OPENSSL_EXTRA_NO_ASN1 not supported with WOLFSSL_TSP)"
+#endif
+
+#include <wolfssl/openssl/ts.h>
+#include <wolfssl/openssl/pkcs7.h>
+#include <wolfssl/wolfcrypt/tsp.h>
+
+/* Time-Stamp Protocol (RFC 3161) compatibility layer.
```
dgarske, line 47: **⚪ [Info] Stale 'Requester side only' file/module comment now that responder code is present** · _style_

The module comment states "Requester side only: create and encode requests, decode responses and verify time-stamp tokens." The file now also implements the full responder OpenSSL compat layer (`WOLFSSL_TS_RESP_CTX`, `wolfSSL_TS_RESP_CTX_*`, `wolfSSL_TS_RESP_create_response` under `WOLFSSL_TSP_RESPONDER`). The comment is misleading for a reader. Note also that the entire compat layer is gated on `WOLFSSL_TSP_VERIFIER` (line 32-33), so a responder-only OpenSSL build (`--enable-tsp --enable-opensslall -DWOLFSSL_TSP_RESPONDER`, per the new `tsp-responder` CI job) compiles this file out entirely and exposes no OpenSSL TS API -- consistent with `wolfssl/openssl/ts.h`, but worth a one-line note in the header so it is not surprising.

**Fix:** Update the module comment to reflect that both requester and responder are implemented, and that the layer requires WOLFSSL_TSP_VERIFIER.

dgarske (commented):
## Skoll Multi-Scan Review

### Posted findings
- [Low] [review] **wolfssl_asn1_integer_new_buf emits a non-canonical zero-length DER INTEGER (02 00) when len==0** — `src/ssl_asn1.c:1201`
- [Info] [review] **TspResponse_Verify relies on an unenforced 'ret==0 implies contentSz0' invariant to avoid handing back dangling pointers** — `wolfcrypt/src/tsp.c:2179`
- [Info] [review-security] **failInfo BIT STRING left-shift depends on a decode invariant in a separate function** — `wolfcrypt/src/asn_tsp.c:1018-1021`
- [Info] [review-security] **New DecodedCert member inserted mid-struct rather than at the end** — `wolfssl/wolfcrypt/asn.h:1911`
- [Info] [review] **Stale 'Requester side only' file/module comment now that responder code is present** — `src/ssl_tsp.c:47-50`

### Skipped findings
- [Low] ~~Unchecked cnt+1 arithmetic can bypass the WOLFSSL_NO_MALLOC bounds check and undersize the heap alloc~~
- [Low] ~~TspCheckNum function-like macro evaluates its size argument multiple times~~

dgarske (changes requested):
https://github.com/wolfSSL/wolfssl/pull/10778#pullrequestreview-4649626573

src/ssl_asn1.c
SparkiDev: Fixed.
Extended the existing guard to if ((val == NULL) || (len == 0)) return NULL;

wolfcrypt/src/tsp.c
SparkiDev: Fixed.
Made the ret==0 && tstInfo && contentSz==0 case an explicit error (TSP_VERIFY_E) before the rebasing block.

wolfcrypt/src/asn_tsp.c
SparkiDev: Fixed.
Bounded the shift to length in [2,5] (... .length >= 2 && ... <= 5), with a comment explaining the coupling.

wolfssl/wolfcrypt/asn.h
SparkiDev: Fixed.
Moved word32 extExtKeyUsageOidCnt from mid-struct (after extExtKeyUsage) to the end of the struct, preserving the offsets of all existing members (project convention).

src/ssl_tsp.c
SparkiDev: Fixed.
Rewrote it to state the layer implements both requester and responder (WOLFSSL_TSP_RESPONDER) and is gated on WOLFSSL_TSP_VERIFIER.

SparkiDev:
Jenkins: retest this please

dgarske (approved):

Frauschi (approved):

PR #10781, opened 2026-06-25
.github/scripts/parallel-make-check.py: emit literal status emoji, not :shortcodes:

julek-wolfssl:
The `parallel-make-check.py` summary table is printed both to the GitHub
step summary and to plain stdout (the Actions console log and local
terminals). GitHub emoji `:shortcodes:` like `:heavy_minus_sign:` only
expand on GitHub's Markdown surfaces, so in the console log they showed up
as literal `:shortcode:` text.

Emit the Unicode emoji directly instead, using `\N{...}` escapes so the
source stays ASCII. The glyphs match what the shortcodes mapped to, so the
rendered step summary is unchanged while the console log now shows real
status icons.


dgarske:
Jenkins retest this please. Flaky valgrind

dgarske (approved):

Frauschi (approved):

PR #10785, opened 2026-06-25
openssl compat errors and mldsa oid fix

stenslae:
# Description

- **Fix 1: verifyX509orX509REQ(): add ML-DSA case**

verifyX509orX509REQ() was missing a case for ML-DSA public keys, so they returned errors. Added an mldsaOID field that is from the cert's OID in wolfSSL_X509_get_pubkey(). With pkeys with no backing cert, the OID is cached and used for level recognition.

Also fixed the wc_MlDsaKey_PublicKeyDecode() so that auto-detect is checked before the draft mode.

- **Fix 2: return 0 when error queue not compiled in**

When WOLFSSL_HAVE_ERROR_QUEUE is not defined, all three functions returned (unsigned long)(0 - NOT_COMPILED_IN) (= 174), a non-zero value. This is non-OpenSSL-compatible, so changed all three #else branches to return 0.


dgarske (dismissed):

src/x509.c
```diff
@@ -8954,6 +8954,12 @@ static int verifyX509orX509REQ(WOLFSSL_X509* x509, WOLFSSL_EVP_PKEY* pkey,
             type = DSAk;
             break;
 
+    #if defined(WOLFSSL_HAVE_MLDSA)
+        case WC_EVP_PKEY_DILITHIUM:
+            type = x509->pubKeyOID;
```
Frauschi: **🟡 [Medium] ML-DSA verify uses subject key OID instead of the verifying key's OID for parameter-set selection**
*🚫 BLOCK*  `bug`

The new `WC_EVP_PKEY_DILITHIUM` case sets `type = x509->pubKeyOID` — the ML-DSA parameter-set OID (`ML_DSA_44k/65k/87k`) taken from the **certificate's own** SubjectPublicKeyInfo (the subject key). But `type` is then passed as the `pubKeyOID` for the **caller-supplied verifying key** `pkey` into `CheckCertSignaturePubKey`/`CheckCSRSignaturePubKey` → `ConfirmSignature` (`wolfcrypt/src/asn.c:23234`). There, `keyOID` selects the ML-DSA level (`asn.c:17470-17493`) used by `wc_MlDsaKey_SetParams()` / `wc_MlDsaKey_PublicKeyDecode()` to decode the **verifying key** (`asn.c:17507-17515`). So the level used to decode the issuer/CA key is taken from the subject cert instead of from the key itself.

That this OID must describe the *verifying* key is confirmed by the CA-lookup path in the same function, which sets `pubKeyOID = ca->keyOID` (`asn.c:23218`). Note the RSA/EC/DSA cases derive `type` from `pkey->type`; only the ML-DSA case reaches into `x509`. For RSA/EC/DSA the distinction is harmless (those OIDs do not encode a parameter set), and for self-signed / same-level PKIs the subject and issuer OIDs coincide, so it happens to work.

**Impact:** when the issuer's ML-DSA level differs from the subject's — e.g. an ML-DSA-65 leaf signed by an ML-DSA-87 CA whose key is passed as `pkey` — `x509->pubKeyOID` yields the wrong level and verification of a *validly*-signed certificate fails. The failure mode is **fail-safe**: a level/type mismatch makes key import or signature verification FAIL (false negative, returns `WOLFSSL_FAILURE`) and cannot produce a false accept, so there is no security regression, and this remains a net improvement over the pre-PR unconditional `WOLFSSL_FATAL_ERROR`.

**Recommendation:** Derive the ML-DSA level from the **verifying key** `pkey`, not from the subject cert. Cleanest is to read the algorithm OID from `pkey`'s own SPKI DER (`pkey->pkey.ptr`) so it always matches the key being decoded — the `WOLFSSL_EVP_PKEY` struct only stores `type = WC_EVP_PKEY_DILITHIUM` with no per-level field, which is presumably why `x509` was used here. Note `x509->sigOID` cannot be passed verbatim: it lives in the signature-OID space (`CTC_ML_DSA_44/65/87`) while the `keyOID` switch in `ConfirmSignature` requires the key-OID space (`ML_DSA_44k/65k/87k`) — see `asn.c:17019-17024` — so it would need mapping first. If only same-level / self-signed verification is intended, document that assumption with a brief comment. Either way, add a cross-level positive test (ML-DSA-65 leaf verified with an ML-DSA-87 CA key) plus a mismatched-level negative test to lock in the expected behavior — the path is currently unexercised (`tests/api/test_x509.c` has no ML-DSA cases).

Frauschi (changes requested):
## 🐺 Skoll Code Review

### Posted findings
- [Medium] **ML-DSA verify uses subject key OID instead of the verifying key's OID for parameter-set selection** — `src/x509.c:8959`

src/x509.c
stenslae: Good finding. Fixed it  by adding mldsaOID field to WOLFSSL_EVP_PKEY, and use wolfSSL_X509_get_pubkey() to assign the respective level's mldsaOID. Also included FIPS 204 draft OIDs in this change.

Added mldsaOidFromPKeyDer() to decode the SPKI DER to recover the OID when mldsaOID is unset in cases with d2i_PUBKEY. verifyX509orX509REQ() now assigns the type as using the mldsaOID, but falls back to mldsaOidFromPKeyDer() if mldsaOID is unset.

Also added tests in test_wolfSSL_X509_verify_mldsa_cross_level.

dgarske (changes requested):
Some of the CI is failing for PQC. Please check. Thanks @stenslae

dgarske (dismissed):
Still seeing CI failures "make check failed for: mldsa-no-asn1-fips204-draft-opensslextra (1 config(s) aborted by fail-fast)"

Frauschi:
**Consider folding the new ML-DSA fixtures into `certs/renewcerts.sh`**

`renewcerts.sh` already has an ML-DSA block, but it only regenerates the per-level `mldsa<N>-*` certs/keys and `ecc-leaf-mldsa44.pem`. The cross-level chain this PR adds (`mldsa87-ca-*`, `mldsa65-leaf87ca-*`) is documented only as manual commands in `README.txt`.

Worth wiring the chain into the script: it's the regression fixture for the OID-selection fix here, and since it isn't scripted, the next `renewcerts.sh` run regenerates the base certs but leaves the CA/leaf chain untouched — so they drift out of sync. It's pure OpenSSL 3.5+ and fits right after the `ecc-leaf-mldsa44` section.

I'd leave `_oqskeypair.der` (needs liboqs) and the raw key-shape vectors as documented-only. Not blocking.

Frauschi (approved):
LGTM

SparkiDev (approved):

Issue #10796, opened 2026-06-27
[Bug]: DefTicketEncCb: BAD_STATE_E on second handshake when ticket hint > WOLFSSL_TICKET_KEY_LIFETIME / 2


Spongman:
### Version

5.9.1, HEAD

### Description

Summary
When a WOLFSSL_CTX is reused across successive DTLS 1.3 handshakes, and wolfSSL_CTX_set_TicketHint() (or wolfSSL_CTX_set_timeout(), which forwards its argument directly as the hint) is called with a value greater than WOLFSSL_TICKET_KEY_LIFETIME / 2, DefTicketEncCb permanently fails on every handshake after the first.

TicketEncCbCtx_ChooseKey() returns BAD_STATE_E when both keys are simultaneously alive but neither covers the hint window. The invariant ticketHint < WOLFSSL_TICKET_KEY_LIFETIME / 2 is enforced at compile time for SESSION_TICKET_HINT_DEFAULT but not for runtime wolfSSL_CTX_set_TicketHint() or wolfSSL_CTX_set_timeout() calls.

Tests added
test_dtls13_server_ctx_reuse_ticket — repro, fails on unpatched wolfSSL
One server WOLFSSL_CTX reused across two DTLS 1.3 handshakes using a hybrid PQ X25519MLKEM768 key share (fragmented ClientHello). Hint set to WOLFSSL_TICKET_KEY_LIFETIME + 1 to arm the trigger. The second wolfSSL_accept() must succeed — it fails with BAD_STATE_E on current HEAD.

test_dtls13_server_ctx_reuse_no_ticket_ok — control, passes on all versions
Identical setup but with wolfSSL_CTX_no_ticket_TLSv13() suppressing server-issued tickets. Both handshakes succeed, isolating the fault to NewSessionTicket issuance with an oversized hint rather than CTX reuse generally.

Suggested fix
Either cap ctx->ticketHint at WOLFSSL_TICKET_KEY_LIFETIME / 2 inside wolfSSL_CTX_set_TicketHint() and wolfSSL_CTX_set_timeout(), or extend TicketEncCbCtx_ChooseKey() to generate a new key unconditionally when neither key covers the hint window and neither has expired. This PR adds only the regression tests.

### Reproduction steps

see PR #10797 

```
./configure \
  --prefix=/usr --libdir=/usr/lib64 --enable-shared --disable-static \
[...]

PR #10797, opened 2026-06-27
tests/api: add DTLS 1.3 server CTX reuse ticket regression test

Spongman:
test_dtls13_server_ctx_reuse_ticket reproduces a BAD_STATE_E fatal error that occurs when a WOLFSSL_CTX is reused across successive DTLS 1.3 handshakes after issuing a NewSessionTicket.

Root cause: TicketEncCbCtx_ChooseKey() refuses to use a key unless expirary[i] >= now + ticketHint, but only generates a replacement when expirary[i] < now.  When ticketHint > WOLFSSL_TICKET_KEY_LIFETIME / 2 (e.g. because wolfSSL_CTX_set_timeout() was called with a large value), both keys can be simultaneously alive yet unable to cover the hint window, hitting the BAD_STATE_E branch and causing wolfSSL_accept() to fail fatally on every handshake after the first.

test_dtls13_server_ctx_reuse_no_ticket_ok is the control: the same scenario with TLS 1.3 tickets suppressed passes, isolating the fault to NewSessionTicket issuance tainting the reused CTX.

Repros #10796

# Testing

(see bug report)

# Checklist

 - [x] added tests

### DIFF
--- a/tests/api/test_dtls.c
+++ b/tests/api/test_dtls.c
@@ -6352,6 +6352,11 @@ int test_dtls13_server_ctx_reuse_ticket(void)
     XMEMSET(&test_ctx, 0, sizeof(test_ctx));
     ExpectIntEQ(test_memio_setup(&test_ctx, &ctx_c, &ctx_s, &ssl_c, &ssl_s,
         wolfDTLSv1_3_client_method, wolfDTLSv1_3_server_method), 0);
+    /* Set hint > WOLFSSL_TICKET_KEY_LIFETIME / 2 to trigger the bug in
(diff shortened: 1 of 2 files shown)


dgarske:
Hi @Spongman , thank you for the issue report and reproducer. This has been assigned to @mattia-moffa to work on. For now I will close this PR, however I am very interested to hear about your use of wolfSSL and if you are planning any future PR's. We require a signed contributor agreement to be in place for any code we merge and typically only set it up for larger contributions.
Thanks, David Garske, wolfSSL

PR #10803, opened 2026-06-29
Fenrir fixes

Frauschi:
Fix various Fenrir issues:
- **F-6345**: Reject oversized lengths in the memory BIO write path so an inflated length can't grow a short buffer and read past the source.
- **F-6346**: Reject oversized lengths in `EVP_EncodeBlock` to prevent the base64 output-size calculation from overflowing.
- **F-6347**: Reject negative and oversized lengths in `EVP_EncodeUpdate` to stop integer overflow and out-of-bounds input reads.
- **F-6350**: Cap the `d2i_ASN1_OBJECT` parse window to the encoded OID size to avoid reading beyond the object.
- **F-6351**: Fix a use-after-free in `wolfSSL_ASN1_STRING_set` when the string is set from an alias of its own buffer.
- **F-6426**: Reject Camellia cipher operations when no key has been set, returning `MISSING_KEY` instead of running on uninitialized state.
- **F-6427**: Reject RC2 cipher operations when no key has been set, consistently across ECB and CBC entry points.
- **F-6547**: Reject TLS KeyUpdate messages on QUIC connections (both send and receive) per RFC 9001 §6.

### DIFF
--- a/src/ssl_asn1.c
+++ b/src/ssl_asn1.c
@@ -3174,42 +3174,52 @@ int wolfSSL_ASN1_STRING_set(WOLFSSL_ASN1_STRING* asn1, const void* data, int sz)
          * when sz == INT_MAX. By this point sz >= 0 (negative sz is
          * handled above as OpenSSL -1/strlen compat). */
         size_t allocSz = (size_t)sz + 1;
+        char* oldData = asn1->data;
+        int oldDynamic = asn1->isDynamic;
+        char* dst;
 
-        /* Dispose of any existing dynamic data. */
-        if (asn1->isDynamic) {
-            XFREE(asn1->data, NULL, DYNAMIC_TYPE_OPENSSL);
-            asn1->data = NULL;
-        }
-
-        /* Check string will fit - including NUL. */
+        /* Select the destination buffer WITHOUT disposing of the existing
+         * data yet. Deferring the free keeps the copy below safe even when
+         * the source aliases the object's own buffer (data == asn1->data),
+         * avoiding a use-after-free or clear-before-copy without needing an
+         * ephemeral allocation. */
         if (allocSz > CTC_NAME_SIZE) {
             /* Allocate new buffer. */
-            asn1->data = (char*)XMALLOC(allocSz, NULL,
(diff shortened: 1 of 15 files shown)


Frauschi:
Jenkins retest this please

src/ssl_asn1.c
```diff
@@ -3158,6 +3170,20 @@ int wolfSSL_ASN1_STRING_set(WOLFSSL_ASN1_STRING* asn1, const void* data, int sz)
...
+        copy = XMALLOC((size_t)sz + 1, NULL, DYNAMIC_TYPE_OPENSSL);
+        if (copy == NULL) {
+            ret = 0;
+        }
+        else {
+            XMEMCPY(copy, data, (size_t)sz);
+            data = copy;
+        }
+    }
+
```
douzzer: Is there a way we can make this safe without an ephemeral heap allocation here?

douzzer (changes requested):

tests/quic.c
```diff
@@ -1470,6 +1470,56 @@ static int test_quic_server_hello_fail(int verbose) {
...
+    ExpectIntEQ(wolfSSL_update_keys(tserver.ssl),
+                WC_NO_ERR_TRACE(BAD_FUNC_ARG));
+
+    /* RFC 9001 section 6: a TLS KeyUpdate handshake message must be rejected
+     * as a fatal unexpected_message connection error when running over QUIC.
+     * Feed a key_update (update_not_requested) as post-handshake CRYPTO data
+     * and confirm the server refuses to process it. */
+    len = fake_record(key_update, OPAQUE8_LEN, lbuffer);
+    lbuffer[HANDSHAKE_HEADER_SZ] = update_not_requested;
+    AssertIntEQ(wolfSSL_provide_quic_data(tserver.ssl,
```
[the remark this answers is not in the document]
Frauschi: fixed

src/ssl_asn1.c
Frauschi: fixed

Frauschi:
Jenkins retest this please.

douzzer (approved):

PR #10805, opened 2026-06-29
Fix Azure example to set NULL after free

### LINKED ISSUE #10800: [Bug]: Double free of `ssl` in Azure Sphere server example on the shutdown path
### Contact Details

_No response_

### Version

5.9.2

### Description

I found a possible double free of the `WOLFSSL *ssl` object in the Azure Sphere server example. On the normal shutdown path, `ssl` is freed at the end of the connection loop without being set to `NULL`, and then freed again by `util_Cleanup()` after the loop exits.

`IDE/VS-AZURE-SPHERE/server/server.c`

The shared cleanup helper frees `ssl` (`IDE/VS-AZURE-SPHERE/shared/util.h`):

```c
static void util_Cleanup(int sockfd, WOLFSSL_CTX* ctx, WOLFSSL* ssl)
{
    wolfSSL_free(ssl);      /* Free the wolfSSL object  */
    wolfSSL_CTX_free(ctx);
    wolfSSL_Cleanup();
    close(sockfd);
}
```

Inside the connection loop, each completed connection frees `ssl` but does not clear the pointer:

```c
    /* ... handle one connection ... */
    if (strncmp(buff, "shutdown", 8) == 0) {
        ...
        shutdown = 1;
    }
    ...
    wolfSSL_free(ssl);      /* Free the wolfSSL object */
    close(connd);           /* Close the connection    */
}   /* end of while (!shutdown) */

util_Cleanup(sockfd, ctx, ssl);   /* frees ssl again */
```

When a `"shutdown"` command is received, `shutdown` is set, the reply is written, then `wolfSSL_free(ssl)` runs at the bottom of the loop body. Because `ssl` is not set to `NULL`, the loop exits and `util_Cleanup(sockfd, ctx, ssl)` calls `wolfSSL_free(ssl)` a second time on the same pointer — a double free.

The same dangling-`ssl` problem also affects error handling on later iterations: after the loop body frees `ssl` without nulling it, if the next iteration's `accept()` or `wolfSSL_new()` fails, `util_Cleanup(sockfd, ctx, ssl)` is called with the already-freed `ssl`.

### Reproduction steps

_No response_

### Relevant log output

```shell

```

embhorn:
# Description

Fix Azure example to set NULL after free

Fixes #10800 

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/IDE/VS-AZURE-SPHERE/server/server.c
+++ b/IDE/VS-AZURE-SPHERE/server/server.c
@@ -197,6 +197,7 @@ int main(void)
 
         /* Cleanup after this connection */
         wolfSSL_free(ssl);      /* Free the wolfSSL object              */
+        ssl = NULL;             /* Avoid double-free in util_Cleanup    */
         close(connd);           /* Close the connection to the client   */
     }
 


embhorn:
Closing in favor of https://github.com/wolfSSL/wolfssl/pull/10854

PR #10806, opened 2026-06-29
STM32Cube: fix STM32L552 HASH build and ecc_test_buffers min digest length

aidangarske:


aidangarske:
already handled by david

PR #10807, opened 2026-06-29
AES-GCM-SIV: Add implementation in C and assembly

SparkiDev:
# Description

Added assembly for Intel x64, ARM64, ARM32, Thumb2.

# Testing

Tested each platform with AES-GCM-SIV.


dgarske (commented):
## Skoll Multi-Scan Review

**Modes:** `review-security`**Overall recommendation:** `COMMENT`

### Posted findings
- [Medium] [review-security] **SGX cpuid hard-set now forces CPUID_VAES | CPUID_AVX512, can crash AES-GCM on SGX CPUs lacking those extensions** — `wolfcrypt/src/cpuid.c:56-57`
- [Medium] [review-security] **New Thumb-2 AES-GCM-SIV asm and C wrappers are neither built nor run in the added CI** — `.github/workflows/aesgcm-siv.yml:60-148`

#### POLYVAL hash-key stack temporaries not zeroized (t, hrev)
**File:** `wolfcrypt/src/aes.c (AesGcmSivPolyvalInit asm branch; AesGcmSivPolyvalInitSw)`
**Function:** `AesGcmSivPolyvalInit / AesGcmSivPolyvalInitSw`
**Severity:** Low
**Category:** Zeroization

The per-message POLYVAL key is derived from the secret message-authentication-key. The persistent copies in `AesGcmSivPolyval` (`poly->h` / `poly->m` / `poly->hHw`) are correctly wiped by `AesGcmSivPolyvalFinal` via `ForceZero(poly, sizeof(*poly))`. However the intermediate stack buffers that hold transformed copies of that key are not wiped before going out of scope: `byte t[WC_AES_BLOCK_SIZE]` in the asm key-prep branch of `AesGcmSivPolyvalInit` (holds `ByteReverse(mulX(ByteReverse(h)))`), and `byte hrev[WC_AES_BLOCK_SIZE]` in both `AesGcmSivPolyvalInitSw` variants (holds `ByteReverse(h)`). These leave a permutation of the POLYVAL key on the stack after the function returns. This is defense-in-depth only (recovering it would let an attacker forge for that specific key+nonce), and the buffers are typically overwritten by later stack frames.

**Recommendation:** Add `ForceZero(t, sizeof(t));` before the asm key-prep block returns and `ForceZero(hrev, sizeof(hrev));` at the end of each `AesGcmSivPolyvalInitSw` variant, consistent with the zeroization used elsewhere in the SIV code.

_Referenced code: `wolfcrypt/src/aes.c (AesGcmSivPolyvalInit asm branch; AesGcmSivPolyvalInitSw)` (9 lines)_

---

#### ARM64 GHASH length conversion changed from ubfiz (#3,#32) to lsl #3, dropping upper-32-bit masking
**File:** `wolfcrypt/src/port/arm/armv8-aes-asm.S (multiple *_partial_done labels); wolfcrypt/src/port/arm/armv8-aes-asm_c.c (AES_GCM_encrypt_final/decrypt_final/_AARCH64 paths)`
**Function:** `AES_GCM_*_AARCH64 / aes_gcm_*_arm64_crypto_*_partial_done`
**Severity:** Info
**Category:** Logic

In the existing (non-SIV) AES-GCM AArch64 code regenerated by this PR, byte-length-to-bit-length conversions changed from `ubfiz xN, xN, #3, #32` to `lsl xN, xN, #3`, and standalone `ubfiz xN, xN, #0, #32` truncations (e.g. on tagSz in the *_part_tag paths) were removed. `ubfiz #3,#32` masks the operand to its low 32 bits before shifting; `lsl #3` shifts the full 64-bit register. The two are identical for any operand whose upper 32 bits are zero, which is always the case for the word32 length/tag-size arguments as materialized by wolfSSL callers (32-bit values are zero-extended). So no wrong tag results for valid wolfSSL usage, and these GCM paths are covered by the broader GCM test suite. The note is only that the new sequence is marginally less defensive against a caller leaving non-zero garbage in the upper 32 bits of the argument register.

**Recommendation:** No action required if these are deliberate asm-generator output; the behavior is equivalent for all in-tree callers. If the generator can emit it cheaply, retaining a 32-bit mask (or documenting that callers must pass zero-extended lengths) preserves the prior defensiveness.

_Referenced code: `wolfcrypt/src/port/arm/armv8-aes-asm.S (multiple *_partial_done labels); wolfcrypt/src/port/arm/armv8-aes-asm_c.c (AES_GCM_encrypt_final/decrypt_final/_AARCH64 paths)` (4 lines)_

---

wolfcrypt/src/cpuid.c
```diff
@@ -53,6 +53,8 @@
             new_cpuid_flags |= CPUID_ADX;
             new_cpuid_flags |= CPUID_MOVBE;
             new_cpuid_flags |= CPUID_BMI1;
+            new_cpuid_flags |= CPUID_VAES;
```
dgarske, line 61: **🟠 [Medium] SGX cpuid hard-set now forces CPUID_VAES | CPUID_AVX512, can crash AES-GCM on SGX CPUs lacking those extensions** · _Regression_

The SGX-specific `cpuid_set_flags()` does not call `cpuid` (it would SIGILL inside an enclave, per the file comment) and instead hard-codes the feature set. BEFORE this PR the hard-set list stopped at `CPUID_BMI1` (so Intel ASM in SGX used the AES-NI/PCLMUL paths). AFTER this PR it also sets `new_cpuid_flags |= CPUID_VAES;` and `new_cpuid_flags |= CPUID_AVX512;`. These flags gate the runtime dispatchers for ALL Intel AES-GCM ASM (not only the new GCM-SIV `AesGcmSivPolyvalAsm`/`AesGcmSivCtrAsm`, but the existing `AES_GCM_*` VAES/AVX512 routines). Many SGX-capable parts (Skylake/Coffee Lake/Comet Lake client, and 12th-gen+ client which dropped AVX512) do NOT implement AVX512 or VAES. On such a CPU, an enclave built after this change will dispatch to `vpclmulqdq`/`vaesenc`/zmm code and take a #UD (SIGILL) in AES-GCM/GCM-SIV. This narrows the set of CPUs an SGX build will run on, a behavioral regression introduced by a SIV feature PR. The path is explicitly a benchmarking hard-set (`@TODO`), which limits real-world exposure.

**Fix:** Do not unconditionally force CPUID_VAES/CPUID_AVX512 in the SGX hard-set path. Gate them behind a build macro (e.g. only when the integrator asserts the target CPU has them), or omit them so SGX continues to use the universally-available AES-NI/PCLMUL paths. If the intent was only to benchmark the new SIV asm, scope the addition behind WOLFSSL_AESGCM_SIV plus an explicit opt-in.

.github/workflows/aesgcm-siv.yml
```diff
@@ -0,0 +1,148 @@
...
+          ./configure ${{ matrix.config }}
+          make -j 4
+          make check
+
+      - name: Print errors
+        if: ${{ failure() }}
+        run: |
+          for file in scripts/*.log test-suite.log
+          do
+              if [ -f "$file" ]; then
```
dgarske, line 64: **🟠 [Medium] New Thumb-2 AES-GCM-SIV asm and C wrappers are neither built nor run in the added CI** · _Missing Tests_

This PR adds large new Thumb-2 implementations: `AES_GCMSIV_polyval_thumb2` / `AES_GCMSIV_ctr_thumb2` in `thumb2-aes-asm.S` and the matching inline-asm wrappers in `thumb2-aes-asm_c.c` (~3000 new lines). The added workflow's `cross_check` matrix builds only `arm64-pmull`, `arm64-no-hw-crypto`, and `armhf-crypto`; `make_check` is x86_64 only. The workflow comment acknowledges Thumb-2 (armv7-m) cannot run under qemu-user and claims coverage by 'bare-metal / membrowse builds' that are not part of this workflow. Consequently the Thumb-2 SIV POLYVAL/CTR code is not even compile-tested, let alone validated against the RFC 8452 KATs, in the CI introduced here. A defect in the Thumb-2 POLYVAL or CTR (e.g. a register/shift error) would produce wrong tags or wrong keystream and would not be caught.

**Fix:** Add at least a Thumb-2 build job (e.g. arm-none-eabi cross-compile of wolfcrypt with --enable-armasm --enable-aesgcm-siv) to compile the new thumb2 sources, and where feasible run aesgcm_siv_test under a Cortex-M emulator (the project's m33mu / QEMU armv7m) so the RFC 8452 KATs validate the Thumb-2 path.

dgarske (changes requested):
macOS/clang build break:

--enable-intelasm --enable-aesgcm-siv fails to compile on an Intel MacBookPro:

```
aes.c:885: error: unused function 'AesEcbEncryptBlocks' [-Werror,-Wunused-function]
aes.c:912: error: unused function 'AesEcbDecryptBlocks'
aes.c:997: error: unused function 'AesCtrEncryptBlocks'
```

These three new static helpers' definitions are compiled in the AES-NI config, but their call sites (lines 7796/15153/15252) are behind narrower #if guards, so without AVX512/AES-ECB they're defined-but-unused. gcc tolerates unused static inline; clang -Werror doesn't, which is why all the Linux x86 hosts passed and only the Mac caught it. 

Fix: align the guards (only define each helper when a caller is compiled) or mark them unused.

.github/workflows/aesgcm-siv.yml
SparkiDev: CI loop doesn't build or run Thumb2 assembly code.

SparkiDev:
Add #ifdefs for the three functions.

wolfcrypt/src/cpuid.c
SparkiDev: Put #ifdef protection around two new flags being ORed in.

SparkiDev:
Jenkins: retest this please

FIPS aborted
fips-ready failed

dgarske (approved):

wolfcrypt/src/aes_gcm_asm.asm
```diff
@@ -6469,6 +6469,567 @@ L_AES_GCM_decrypt_final_aesni_cmp_tag_done:
...
+AES_GCMSIV_polyval_aesni ENDP
+_TEXT ENDS
+_DATA SEGMENT
+ALIGN 16
+L_aes_gcmsiv_ctr_aesni_one QWORD \
+     0000000000000001h,  0000000000000000h
+ptr_L_aes_gcmsiv_ctr_aesni_one QWORD L_aes_gcmsiv_ctr_aesni_one
+_DATA ENDS
+_TEXT SEGMENT READONLY PARA
+AES_GCMSIV_ctr_aesni PROC
```
dgarske, line 6846: **🔴 [High] Windows MASM AES_GCMSIV_ctr_* clobber round-count and ctr-pointer registers (wrong ciphertext + wild 16-byte write)** · _Security_

All four new Windows MASM CTR routines mis-allocate registers versus the reference System V `.S` versions. On Windows x64 the args are in=rcx, out=rdx, length=r8, KS=r9, nr=[rsp+48], ctr=[rsp+56]. The prologue does `mov eax,[rsp+48]` (nr) and `mov r10,[rsp+56]` (ctr), then IMMEDIATELY executes `xor eax, eax` (destroying nr) and later `mov r10d, r8d` (destroying the ctr pointer with the length). Thereafter `eax` is used as the byte-offset index (`add eax,64`/`lea [rcx+rax]`) yet the AES round-count decisions are `cmp eax, 11` / `cmp eax, 13` -- comparing the byte offset instead of nr. In the correct SysV `.S` these are `cmpl $11, %r8d` (nr in r8) and the ctr write-back is `movdqu %xmm7, (%r9)` (ctr in r9, never clobbered). Consequences on Windows MASM builds (WOLFSSL_AESNI + WOLFSSL_AESGCM_SIV): (1) Incorrect keystream/ciphertext -- for AES-256 the first 64-byte block runs only 10 rounds while later blocks run 14; even AES-128 breaks past the first 64 bytes because eax>=13 forces 14 rounds. (2) Memory corruption -- at `L_..._done_enc` the `movdqu OWORD PTR [r10], xmm7` writes the 16-byte counter to address `r10 = length (masked)`, a caller-length-derived low address, not the ctr buffer (segfault or heap corruption). The AVX512 variant additionally uses `cmp eax, 11/13` to decide round-key caching, leaving zmm23-26 uninitialized for AES-256. The bug is systematic across all four functions (generator register-remap collision) and is not exercised by the added CI (Linux `.S` + ARM qemu only), so it was not caught.

**Fix:** Do not ship the MASM `.asm` as generated. Regenerate `aes_gcm_asm.asm` with a generator fix so the round count (`nr`) is kept in a register that survives (e.g. keep it out of `eax`, which is the byte index) and so the `ctr` pointer register (r10) is not overwritten by `length`; the final counter write-back must target the ctr pointer. Cross-check against the SysV `.S` (nr=r8d, ctr=r9). Add a Windows/MSVC MASM CI job (or at minimum a local Windows AES-GCM-SIV KAT run for AES-128 >64B and AES-256) to validate before merge.

.github/workflows/aesgcm-siv.yml
```diff
@@ -0,0 +1,148 @@
+name: AES-GCM-SIV (RFC 8452) tests
```
dgarske, line 1: **🟠 [Medium] New AES-GCM-SIV pure-C and assembly POLYVAL/CTR variants (Windows MASM, Thumb2, AArch32/AArch64 base, GCM_WORD32…** · _Missing Tests_

The added workflow provides good RFC 8452 KAT coverage but which POLYVAL multiply / CTR backend it exercises depends on build config. It covers native x86_64 `make check` (software word64 table + Intel `.S` asm) and cross-compiled AArch64 (PMULL/NEON/base) and AArch32 with `QEMU_CPU=max` (crypto/vmull.p64 path). It does NOT exercise these newly written, security-sensitive paths: the Windows MASM `.asm` variant (which in fact contains a correctness+memory-safety bug -- see the AES_GCMSIV_ctr finding -- that this gap allowed through), the `word32` table variant (`GCM_WORD32`, AesGcmSivGMult/AesGcmSivPolyvalInitSw word32 block in aes.c), the table-free `GCM_SMALL` variant, the AArch32 base (`WC_POLYVAL_ASM_AARCH32_BASE`) NO_HW_CRYPTO path (armhf job forces `QEMU_CPU=max`, selecting the crypto path), the AArch64 base (`AES_GCMSIV_polyval_base`, only reached with NEON+HW-crypto both disabled), and Thumb2 (`AES_GCMSIV_polyval_thumb2` / `gcm_siv_thumb2`, targeting armv7-m which qemu-user cannot run). A transposition/reduction/register error in any of these would not be caught by CI. Severity views: review flagged this Medium/SUGGEST (test coverage); review-security flagged Medium (missing-tests, and it is the reason the High MASM defect went undetected).

**Fix:** Add a Windows MSVC/MASM job that builds with AES-NI + AES-GCM-SIV and runs the KATs. Add at least one x86 `make check` config forcing the word32 and GCM_SMALL software multiplies (e.g. `CPPFLAGS=-DGCM_WORD32` and `CPPFLAGS=-DGCM_SMALL`) so the two untested pure-C POLYVAL variants run the RFC 8452 KATs at near-zero cost. Add an AArch32 NO_HW_CRYPTO base config (no `QEMU_CPU=max`) so the generated software-table CTR/POLYVAL variants are exercised. Note the Thumb-2 gap explicitly in release notes if it cannot be emulated.

dgarske (commented):
## Skoll Multi-Scan Review

### Posted findings
- [High] [review-security] **Windows MASM AES_GCMSIV_ctr_* clobber round-count and ctr-pointer registers (wrong ciphertext + wild 16-byte write)** — `wolfcrypt/src/aes_gcm_asm.asm`
- [Medium] [review+review-security] **New AES-GCM-SIV pure-C and assembly POLYVAL/CTR variants (Windows MASM, Thumb2, AArch32/AArch64 base, GCM_WORD32, GCM_SMALL) not exercised by added CI** — `.github/workflows/aesgcm-siv.yml:1-148`
- [Low] [review-security] **Feature-guarding of AES-NI bulk helpers removes WC_MAYBE_UNUSED; verify no unguarded callers** — `wolfcrypt/src/aes.c:908-1062`
- [Low] [review] **wc_AesGcmSivEncrypt lacks WARN_UNUSED_RESULT while decrypt has it** — `wolfssl/wolfcrypt/aes.h:838-847`

#### POLYVAL update leaves plaintext/AAD residue on the stack (block/rev not zeroized)
**File:** `wolfcrypt/src/aes.c`
**Function:** `AesGcmSivPolyvalUpdate`
**Severity:** Low
**Category:** Zeroization

`AesGcmSivPolyvalUpdate` copies the trailing partial block of the hashed data into local buffers `block[16]` and `rev[16]` (`XMEMCPY(block, data, partial)` / `AesGcmSivByteReverse(rev, ...)`). During tag computation this data is plaintext (and AAD). Neither `block` nor `rev` is zeroized before return. `AesGcmSivPolyvalFinal` does `ForceZero(poly, sizeof(*poly))` (covering the running sum and key/table material), but the Update-local buffers are separate stack storage and retain a 16-byte plaintext/AAD residue. Per the project rule to zero secrets with `ForceZero()`, this is a defense-in-depth gap. Impact is limited: the same plaintext already lives in caller-owned in/out buffers, and this matches wolfSSL's existing GHASH behavior, so it is Low.

**Recommendation:** Add `ForceZero(block, sizeof(block)); ForceZero(rev, sizeof(rev));` before returning from `AesGcmSivPolyvalUpdate` (both the ASM-partial and C-fallback exits) to avoid leaving plaintext/AAD on the stack.

_Referenced code: `wolfcrypt/src/aes.c` (13 lines)_

---

wolfcrypt/src/aes.c
```diff
@@ -907,7 +908,8 @@ block cipher mechanism that uses n-bit binary string parameter key with 128-bits
     /* Pick the widest available implementation at runtime.  Callers must
```
dgarske, line 908: **🔵 [Low] Feature-guarding of AES-NI bulk helpers removes WC_MAYBE_UNUSED; verify no unguarded callers** · _Logic_

These x86_64 AES-NI bulk helpers were previously always compiled with `WC_MAYBE_UNUSED WC_INLINE`. The diff now wraps them in `#ifdef HAVE_AES_ECB`, `#if defined(HAVE_AES_ECB) && defined(HAVE_AES_DECRYPT)`, and `#ifdef WOLFSSL_AES_COUNTER` respectively and drops `WC_MAYBE_UNUSED`. If any caller references one of these helpers in a configuration where the new guard is not satisfied (e.g. an AES-NI + AESGCM build without HAVE_AES_ECB / WOLFSSL_AES_COUNTER), the build would fail with an undefined-symbol error; conversely, removing WC_MAYBE_UNUSED could trigger unused-function warnings under -Werror in a sub-config where the guard is met but the function is not called. This is a portability/build concern rather than a runtime security issue, and the `--enable-all-crypto --enable-intelasm` CI config satisfies all guards, so it may be benign.

**Fix:** Confirm that every caller of AesEcbEncryptBlocks/AesEcbDecryptBlocks/AesCtrEncryptBlocks is itself guarded by the same macros (HAVE_AES_ECB / HAVE_AES_DECRYPT / WOLFSSL_AES_COUNTER), and build a narrow config such as AES-NI + AESGCM only (no AES_ECB, no AES_COUNTER) to ensure no regression. Keep WC_MAYBE_UNUSED if any guarded-but-uncalled path remains possible.

wolfssl/wolfcrypt/aes.h
```diff
@@ -824,6 +824,30 @@ int wc_AesSivDecrypt_ex(const byte* key, word32 keySz, const AesSivAssoc* assoc,
...
+ *   key   : key-generating-key, 16 (AES-128) or 32 (AES-256) bytes.
+ *   nonce : 12 bytes.
+ *   tag   : 16 bytes (the RFC 8452 authentication tag).
+ * The encrypted output is the same length as the plaintext; the tag is
+ * returned separately.
+ *
+ * The POLYVAL hash is constant-time wherever the CPU provides carry-less
+ * multiply (x86 PCLMUL, Arm PMULL/VMULL) - the runtime default on such CPUs.
+ * Software-only builds fall back to a key-dependent 4-bit table (a cache-timing
+ * trade-off matching GCM_TABLE GHASH); GCM_SMALL avoids the table entirely. */
```
dgarske, line 838: **🔵 [Low] wc_AesGcmSivEncrypt lacks WARN_UNUSED_RESULT while decrypt has it** · _Convention_

`wc_AesGcmSivDecrypt` is annotated `WOLFSSL_API WARN_UNUSED_RESULT` (correct - the auth result must not be ignored), but `wc_AesGcmSivEncrypt` is only `WOLFSSL_API`. An ignored encrypt return could let a caller transmit uninitialized/garbage output or a stale tag after a `BAD_FUNC_ARG` (e.g. wrong key size). This mirrors the existing `wc_AesGcmEncrypt` signature, so it is consistent with the codebase, hence a NIT rather than a defect.

**Fix:** Consider adding WARN_UNUSED_RESULT to wc_AesGcmSivEncrypt for symmetry and to force callers to check for parameter/setup failures; acceptable to leave as-is to match wc_AesGcmEncrypt.

dgarske (changes requested):
All tested correctly on my array of ARM and Intel machines, but Skoll review found some interesting things -> https://github.com/wolfSSL/wolfssl/pull/10807#pullrequestreview-4619669038

wolfcrypt/src/aes_gcm_asm.asm
SparkiDev: fixed

.github/workflows/aesgcm-siv.yml
SparkiDev: Added

wolfcrypt/src/aes.c
SparkiDev: Guards correct.

wolfssl/wolfcrypt/aes.h
SparkiDev: Added

wolfcrypt/src/aes.c
```diff
@@ -18183,6 +18188,1122 @@ int wc_AesSivDecrypt_ex(const byte* key, word32 keySz, const AesSivAssoc* assoc,
...
+        out[i] = in[WC_AES_BLOCK_SIZE - 1 - i];
+    }
+#endif
+}
+
+/* Multiply a GF(2^128) element (GHASH bit order: the most-significant bit of
+ * byte 0 is the x^0 coefficient) by x: shift the 128-bit value right by one
+ * and reduce with the GHASH polynomial. Branch free, so constant time. Used by
+ * the GCM_SMALL and byte-table multiplies. */
+static WC_INLINE void AesGcmSivMulX(byte* x)
```
dgarske: Clang indicates this is unused when not GCM_SMALL. Needs a macro guard.

dgarske (changes requested):
| Config | POLYVAL path | raptorlake (VAES) | tigerlake (AVX512+VAES) | IvyBridge |
|--------|--------------|---------------|-------------------------|---------------------|
| c-only (`--enable-aesgcm-siv`) | software word64 table | PASS | PASS | **BUILD FAIL (clang)** |
| word32 (`-DGCM_WORD32`) | software word32 table | PASS | PASS | **BUILD FAIL (clang)** |
| small (`-DGCM_SMALL`) | software table-free | PASS | PASS | PASS |
| intelasm | PCLMUL/VAES/AVX512 asm | PASS | PASS | PASS |
| all-crypto + intelasm | asm | PASS | PASS | PASS |
| smallstack | software word64 table | PASS | PASS | **BUILD FAIL (clang)** |
| api-asm (non-cryptonly) | `test_wc_AesGcmSiv*` API | PASS | PASS | PASS |

wolfcrypt/test/test.c
```diff
@@ -20298,6 +20307,236 @@ static wc_test_ret_t aesgcm_aes256_large_test(Aes* enc, Aes* dec)
 }
 #endif /* WOLFSSL_AES_256 */
 
+#ifdef WOLFSSL_AESGCM_SIV
+/* AES-GCM-SIV (RFC 8452) self test: known-answer vectors from RFC 8452
```
dgarske, line 20311: **🟠 [Medium] aesgcm_siv_test() uses AES-128 vectors unconditionally (breaks AES-256-only builds)**

The new self-test `aesgcm_siv_test()` is only guarded by `#ifdef WOLFSSL_AESGCM_SIV`. Its AES-256 vectors are correctly wrapped in `#ifdef WOLFSSL_AES_256`, but the AES-128 vectors (`key128`, `keyBig128` and their `wc_AesGcmSivEncrypt(key128, sizeof(key128), ...)` calls) run unconditionally. This diverges from the convention used everywhere else in `wolfcrypt/test/test.c` (e.g. `aesgcm_test` at lines 12916/13186, and every other AES helper) which guards AES-128 code with `#ifdef WOLFSSL_AES_128`. It also diverges from the sibling API test `test_wc_AesGcmSivEncryptDecrypt` in `tests/api/test_aes.c`, which explicitly requires `defined(WOLFSSL_AES_128)`. In a build with `WOLFSSL_AES_256` + `WOLFSSL_AESGCM_SIV` but AES-128 disabled, `wc_AesSetKey(aes, key128, 16, ...)` returns an error, so `aesgcm_siv_test()` fails and `make check` breaks even though the SIV feature itself is fine.

**Fix:** Guard the AES-128 vector blocks in aesgcm_siv_test() with `#ifdef WOLFSSL_AES_128` for consistency with the rest of test.c and with tests/api/test_aes.c.

dgarske (commented):
## Skoll Code Review

### Posted findings
- [Medium] **aesgcm_siv_test() uses AES-128 vectors unconditionally (breaks AES-256-only builds)** — `wolfcrypt/test/test.c:20311-20537`

### Skipped findings
- [Low] ~~Multi-block (128-byte) SIV KAT vectors are self-generated, not independently sourced~~

wolfcrypt/test/test.c
SparkiDev: Fixed

wolfcrypt/src/aes.c
SparkiDev: Fixed

SparkiDev:
Jenkins: retest this please

tests/api/test_aes.c
```diff
@@ -3705,6 +3705,242 @@ int test_wc_AesGcmMixedEncDecLongIV(void)
  * AES-GCM non-standard nonce lengths
  ******************************************************************************/
 
+#if defined(WOLFSSL_AESGCM_SIV) && !defined(NO_AES) && defined(HAVE_AESGCM)
```
dgarske: **🟠 [Medium] Test-helper guard broader than usage guard - unused-function -Werror break in AES-256-only SIV builds** · _convention_

The static helpers `gcmsiv_hex` and `gcmsiv_kat` are compiled under `#if defined(WOLFSSL_AESGCM_SIV) && !defined(NO_AES) && defined(HAVE_AESGCM)`, but their only call sites live in `test_wc_AesGcmSivEncryptDecrypt` under the stricter guard `... && defined(WOLFSSL_AES_128)`. `WOLFSSL_AES_128` is independently disableable (settings.h line 3664 gates on `!defined(WOLFSSL_AES_128) && !defined(WOLFSSL_AES_256)`), so a legitimate AES-256-only build with `--enable-aesgcm-siv` compiles `gcmsiv_kat` with no caller. Under `-Werror=unused-function` (which this same PR takes pains to satisfy in `aes.c` via the new `#ifdef HAVE_AES_ECB` refactor) this is a hard build break. Note the wolfcrypt `aesgcm_siv_test` in test.c was deliberately written to work in AES-256-only builds (key-size-agnostic invalid-param checks), so this asymmetry in test_aes.c is an oversight rather than intent.

**Fix:** Add `&& defined(WOLFSSL_AES_128)` to the helper `#if` guard so the helpers are compiled only when their call sites are, OR restructure the test to run key-size-agnostically like the wolfcrypt test.c version. Not blocking (test-only, narrow config not in the PR CI matrix) but should be fixed for consistency with the PR's own -Werror hygiene.

wolfcrypt/src/aes.c
```diff
@@ -18183,6 +18188,1127 @@ int wc_AesSivDecrypt_ex(const byte* key, word32 keySz, const AesSivAssoc* assoc,
...
+            }
+            AesGcmSivMulX(V);
+            y = (byte)(y << 1);
+        }
+    }
+    XMEMCPY(poly->s, Z, WC_AES_BLOCK_SIZE);
+}
+
+/* Store the hash key mulX_GHASH(ByteReverse(h)); no table to build. */
+static void AesGcmSivPolyvalInitSw(AesGcmSivPolyval* poly, const byte* h)
```
dgarske, line 18514: **🔵 [Low] POLYVAL hash-key-derived stack temporaries not zeroized** · _Zeroization_

The new AES-GCM-SIV code carefully zeroizes the top-level secrets (`authKey`, `encKey`, `tagTmp`/`expTag`) and wipes the whole POLYVAL state via `AesGcmSivPolyvalFinal` -> `ForceZero(poly, sizeof(*poly))` (aes.c:19000). However, three local stack buffers that hold the per-message POLYVAL hash key in transformed form are never cleared: in the word64 and word32 `AesGcmSivPolyvalInitSw` the buffer `byte hrev[WC_AES_BLOCK_SIZE]` holds `ByteReverse(H)` used to seed the multiply table (aes.c:18616 and 18741), and in `AesGcmSivPolyvalInit` the buffer `byte t[WC_AES_BLOCK_SIZE]` holds the prepared carry-less-multiply key `mulX_GHASH(ByteReverse(h))` for the asm path (aes.c:18933). H is the message-authentication-key derived per-nonce from the key-generating-key; recovering it enables tag forgery for that nonce, so it is secret material under wolfSSL's ForceZero policy (the same policy the GCM_SMALL variant honors by writing straight into `poly->h`, which Final wipes). Leaving these copies on the stack is a defense-in-depth gap (exploitation requires disclosure of freed stack memory), not a directly exploitable flaw. The BEFORE state had no such code; this is introduced by the PR.

**Fix:** Add `ForceZero(hrev, sizeof(hrev));` at the end of both the word64 and word32 `AesGcmSivPolyvalInitSw`, and `ForceZero(t, sizeof(t));` after `AesGcmSivByteReverse(poly->hHw, t)` in `AesGcmSivPolyvalInit`, to match wolfSSL's secret-zeroization policy.

wolfcrypt/src/cpuid.c
```diff
@@ -53,6 +53,10 @@
             new_cpuid_flags |= CPUID_ADX;
             new_cpuid_flags |= CPUID_MOVBE;
             new_cpuid_flags |= CPUID_BMI1;
+        #ifdef WOLFSSL_SGX_CPUID_AVX512_VAES
```
dgarske, line 56: **🔵 [Low] New SGX cpuid force macro affects all VAES/AVX512 paths, not just AES-GCM-SIV** · _convention_

The new `WOLFSSL_SGX_CPUID_AVX512_VAES` macro unconditionally OR-sets `CPUID_VAES | CPUID_AVX512` in the SGX cpuid path (where the CPUID instruction is unavailable in an enclave). This is a reasonable opt-in for SGX builds to reach the new SIV VAES/AVX512 asm, but it also enables those flags for regular AES-GCM and any other VAES/AVX512-gated code. The macro is documented in `.wolfssl_known_macro_extras`, and the integrator must only set it when the target CPU truly supports VAES+AVX512. Behavior is correct; just noting the blast radius is broader than SIV.

**Fix:** No change required. Consider a one-line comment clarifying that the integrator asserts CPU support (CPUID is unreliable under SGX) and that this affects all VAES/AVX512 code paths.

wolfcrypt/src/aes.c
```diff
@@ -18183,6 +18188,1127 @@ int wc_AesSivDecrypt_ex(const byte* key, word32 keySz, const AesSivAssoc* assoc,
...
+    AesGcmSivPolyvalInitSw(poly, h);
+}
+
+/* Add data to the POLYVAL sum. A trailing partial block is zero-padded to a
+ * full block, which is exactly the padding RFC 8452 applies to the AAD and
+ * the plaintext independently. */
+static void AesGcmSivPolyvalUpdate(AesGcmSivPolyval* poly, const byte* data,
+    word32 sz)
+{
+    byte block[WC_AES_BLOCK_SIZE];
```
dgarske, line 18962: **⚪ [Info] Plaintext/AAD-derived scratch blocks in POLYVAL update not zeroized** · _Zeroization_

`AesGcmSivPolyvalUpdate` uses local buffers `byte block[WC_AES_BLOCK_SIZE]` and `byte rev[WC_AES_BLOCK_SIZE]` (aes.c:18956-18957) that transiently hold byte-reversed plaintext/AAD blocks and a zero-padded trailing block. They are not zeroized on return. This is a minor hardening gap only: the underlying plaintext already lives in the caller-owned `in`/`out` buffers (which this API does not own), so it is far less sensitive than the hash-key temporaries in the primary finding. Noted for completeness.

**Fix:** Optionally `ForceZero` `block`/`rev` before returning for defense-in-depth consistency; low priority.

dgarske (commented):
## Skoll Multi-Scan Review

**Findings:** 7 total — 4 posted, 3 skipped

### Posted findings
- [Medium] [review] **Test-helper guard broader than usage guard - unused-function -Werror break in AES-256-only SIV builds** — `tests/api/test_aes.c:3708-3945`
- [Low] [review-security] **POLYVAL hash-key-derived stack temporaries not zeroized** — `wolfcrypt/src/aes.c:18616,18741,18933`
- [Low] [review] **New SGX cpuid force macro affects all VAES/AVX512 paths, not just AES-GCM-SIV** — `wolfcrypt/src/cpuid.c:56-59`
- [Info] [review-security] **Plaintext/AAD-derived scratch blocks in POLYVAL update not zeroized** — `wolfcrypt/src/aes.c:18956-18957`

### Skipped findings
- [Low] ~~Large stack buffers in aesgcm_siv_test do not use the WOLFSSL_SMALL_STACK pattern~~
- [Info] ~~Incidental curve25519/mlkem ARM asm regenerations bundled into an AES-GCM-SIV PR~~
- [Info] ~~aes.c inline-helper feature guards reviewed (intended -Werror fix, sound)~~

dgarske (changes requested):
https://github.com/wolfSSL/wolfssl/pull/10807#pullrequestreview-4647241633

tests/api/test_aes.c
SparkiDev: Fixed.
Added && defined(WOLFSSL_AES_128) to the gcmsiv_hex/gcmsiv_kat guard.

wolfcrypt/src/aes.c
SparkiDev: Fixed.
Added ForceZero for the transformed per-message hash key H.

wolfcrypt/src/cpuid.c
SparkiDev: Fixed.
Added a comment at WOLFSSL_SGX_CPUID_AVX512_VAES in cpuid.c.

wolfcrypt/src/aes.c
SparkiDev: Fixed.
Added ForceZero(block, …) on the asm-path return and ForceZero(block/rev, …) at the end of AesGcmSivPolyvalUpdate.

dgarske (approved):

SparkiDev:
Jenkins: retest this please

FIPS ready (aborted)

dgarske (approved):

PR #10810, opened 2026-06-29
Add support for id-on-bundleEID from RFC 9174.

### LINKED ISSUE #10694: [Bug]: X509 decoder fails to handle OtherName with value of an IA5String
### Contact Details

brian.sipos@gmail.com

### Version

5.8.4

### Description

The X509 component fails to decode any certificate with a SAN with value Other Name having a value that is an IA5String. It looks like Other Name values of UTF8String are handled correctly (although not visualized in a tool such as `wolfssl x509 ...`, which is fine).

This happens with online (D)TLS certificate exchange as well as with offline tool such as `wolfssl x509 ...` so it's easily isolated and tested.

It appears that the Other Name value is only considered value for a few specific types, when the definition in RFC 5280 allows the value to be any type:
```asn1
   OtherName ::= SEQUENCE {
        type-id    OBJECT IDENTIFIER,
        value      [0] EXPLICIT ANY DEFINED BY type-id }
```

### Reproduction steps

1. Get a certificate conforming to [RFC 9174](https://www.rfc-editor.org/info/rfc9174/#section-4.4.2), specifically having a Subject Alt Name extension with an item of type Other Name, and an Other Name value that is an IA5String. A dummy example is attached [node-transport.pem](https://github.com/user-attachments/files/28979674/node-transport.pem.txt).
2. Attempt to process the certificate using `wolfssl x509 -noout -text -in node-transport.pem`
3. Observe the tool fails to process.
4. Convert to DER form and process using `dumpasn1` tool and see it succeed, so the contents are well-formed ASN1.
5. Attempt to process the same file using `openssl x509 ...` and see it succeed, so the contents are valid PKIX cert.

The relevant portion from dumpasn1 is:
```
251  54:         SEQUENCE {
253   3:           OBJECT IDENTIFIER subjectAltName (2 5 29 17)
258  47:           OCTET STRING, encapsulates {
260  45:             SEQUENCE {
262  28:               [0] {
264   8:                 OBJECT IDENTIFIER '1 3 6 1 5 5 7 8 11'
274  16:                 [0] {
276  14:                   IA5String 'dtn://node000/'
       :                   }
       :                 }
292  13:               [2] 'node000.local'
       :               }
       :             }
       :           }
```

### Relevant log output

```shell
When wolfssl is configured with `--enable-debug` and `-DWOLFSSL_DEBUG_ASN_TEMPLATE` the following is output.

wolfSSL Entering DecodeAltNames
TEMPLATE: altNameASN
wolfSSL Entering GetASN_Items
 0:    2   28    [0]             
TEMPLATE: otherNameASN
wolfSSL Entering GetASN_Items
 0:    4    8    OBJECT ID       
 1:   14   16 +  [0]             
More data in constructed item at depth: 0
```

kareem-wolfssl:
# Description

Fixes #10694.

# Testing

Built in tests, added test, reproducer from original issue

# Checklist

 - [X] added tests
 - [ ] updated/added doxygen


wolfcrypt/src/asn.c
```diff
@@ -18724,12 +18724,20 @@ static int ConfirmNameConstraints(Signer* signer, DecodedCert* cert)
  *  (RFC3280 sec 4.2.1.7). Often used with FIPS 201 smartcard login.
  * FASC-N (Federal Agency Smart Credential Number), defined in the document
  *  fpki-x509-cert-policy-common.pdf. Used for a smart card ID.
+ *
+ * id-on-bundleEID (RFC 9174, sec 4.4.1), an Other Name whose value is an
+ *  IA5String holding a Bundle Protocol node/endpoint ID (e.g. "dtn://node/").
+ *  Only handled when WOLFSSL_DTN is defined as these OIDs are specific to
+ *  Delay-Tolerant Networking (DTN) / the Bundle Protocol.
```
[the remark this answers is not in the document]
kareem-wolfssl: This is intentional.

kareem-wolfssl:
Retest this please

dgarske (commented):
## Skoll Multi-Scan Review

#### New regression test never executes: WOLFSSL_DTN not enabled by any build config
**File:** `tests/api/test_asn.c:1508-1563; configure.ac:1307`
**Function:** `test_DecodeOtherName_bundleEID`
**Severity:** Medium
**Category:** test

The entire body of `test_DecodeOtherName_bundleEID` is guarded by `#if defined(WOLFSSL_DTN) && (defined(WOLFSSL_SEP) || defined(WOLFSSL_FPKI)) && ...`. A repo-wide search shows `WOLFSSL_DTN` is defined nowhere in the build system: it appears only in the 9 files this PR touches, and there is no `--enable-dtn` option or `-DWOLFSSL_DTN` in `configure.ac` (contrast `WOLFSSL_FPKI`, which is turned on at configure.ac:1308). As a result the new regression test always compiles to an empty `EXPECT_RESULT()` in CI and provides no protection against a future regression of the id-on-bundleEID decode path. The `DecodeOtherHelper`/`otherNameASN` changes in asn.c are therefore also completely unexercised by automated testing.

**Recommendation:** Wire WOLFSSL_DTN into the build/CI configuration so the added test exercises the new code path; otherwise the regression test is effectively dead.

_Referenced code: `tests/api/test_asn.c:1508-1563; configure.ac:1307-1310` (4 lines)_

---

#### WOLFSSL_DTN silently requires WOLFSSL_SEP or WOLFSSL_FPKI, and is undocumented
**File:** `wolfcrypt/src/asn.c:18716-18744; wolfcrypt/src/asn.c:18728-18731`
**Function:** `DecodeOtherName`
**Severity:** Medium
**Category:** convention

The `otherNameASN` template (with the new BEID entry) and `DecodeOtherName` live inside `#if defined(WOLFSSL_SEP) || defined(WOLFSSL_FPKI)`. If a user defines only `WOLFSSL_DTN` (the macro the feature is named for), none of the bundleEID handling is compiled and id-on-bundleEID OtherNames are still rejected with ASN_PARSE_E -- with no compile-time error or warning to indicate the macro had no effect. The header comment states the feature is 'Only handled when WOLFSSL_DTN is defined' but does not mention the SEP/FPKI prerequisite. Additionally, because WOLFSSL_DTN is off by default, the original report (#10694) remains unfixed for default builds; this appears intentional per the rationale comment, but it is worth confirming that gating the fix behind an off-by-default macro is the desired resolution.

**Recommendation:** Add documentation for the new WOLFSSL_DTN macro and its build prerequisites; confirm the off-by-default scoping is intended for issue #10694.

_Referenced code: `wolfcrypt/src/asn.c:18716-18744; wolfcrypt/src/asn.c:18728-18732` (5 lines)_

---

### Skipped findings
- [Low] ~~Generated oid_sum.h is hand-edited for BUNDLE_EID_OID~~

wolfcrypt/src/asn.c
```diff
@@ -18724,12 +18724,20 @@ static int ConfirmNameConstraints(Signer* signer, DecodedCert* cert)
...
  */
 static const ASNItem otherNameASN[] = {
 /* TYPEID   */ { 0, ASN_OBJECT_ID, 0, 0, 0 },
 /* VALUE    */ { 0, ASN_CONTEXT_SPECIFIC | ASN_OTHERNAME_VALUE, 1, 1, 0 },
 /* UPN      */     { 1, ASN_UTF8STRING, 0, 0, 2 },
 /* FASC-N   */     { 1, ASN_OCTET_STRING, 0, 0, 2 },
+#ifdef WOLFSSL_DTN
```
dgarske: Let's add WOLFSSL_DTN to the settings.h `WOLFSSL_ASN_ALL` so it gets tested.

dgarske: Document this new macro at top of asn.c. Thanks

dgarske (changes requested):
https://github.com/wolfSSL/wolfssl/pull/10810#pullrequestreview-4649436588

kareem-wolfssl:
Retest this please

kareem-wolfssl:
Retest this please

dgarske (dismissed):

SparkiDev (dismissed):

wolfcrypt/src/asn.c
```diff
@@ -19420,12 +19422,20 @@ static int ConfirmNameConstraints(Signer* signer, DecodedCert* cert)
  *  (RFC3280 sec 4.2.1.7). Often used with FIPS 201 smartcard login.
  * FASC-N (Federal Agency Smart Credential Number), defined in the document
  *  fpki-x509-cert-policy-common.pdf. Used for a smart card ID.
+ *
+ * id-on-bundleEID (RFC 9174, sec 4.4.1), an Other Name whose value is an
+ *  IA5String holding a Bundle Protocol node/endpoint ID (e.g. "dtn://node/").
+ *  Only handled when WOLFSSL_DTN is defined as these OIDs are specific to
+ *  Delay-Tolerant Networking (DTN) / the Bundle Protocol.
  */
 static const ASNItem otherNameASN[] = {
```
dgarske, line 20200: **🔴 [High] New IA5String CHOICE alternative makes mismatched OtherName value types parse successfully instead of failing**

The diff adds `{ 1, ASN_IA5_STRING, 0, 0, 2 }` as a fourth alternative in the group-2 CHOICE of `otherNameASN` (UPN=UTF8String, FASC-N=OCTET STRING, BEID=IA5String, HWN=SEQUENCE). `GetASN_Items()` accepts **any one** member of a choice group without regard to the `TYPEID` OID, and `DecodeOtherHelper()` then reads only the slot that corresponds to the OID. When the encoded value type does not match the OID, that slot is still zeroed from `CALLOC_ASNGETDATA` (`data.ref.data == NULL`, `length == 0`) and the malformed OtherName is accepted with an empty value instead of being rejected.

I built this branch with `./configure --enable-all` (which defines `WOLFSSL_SEP`, `WOLFSSL_FPKI` **and** `WOLFSSL_DTN` via `WOLFSSL_ASN_ALL` in settings.h) and mutated the new `certs/bundle-eid-cert.der` by one byte to confirm:

| SAN OtherName | before this diff | after this diff | asn_orig.c |
|---|---|---|---|
| bundleEID OID + IA5String | `-140` ASN_PARSE_E | `0`, `"dtn://node000/"` (the intended fix) | `0` |
| bundleEID OID + UTF8String | `0`, raw entry preserved (len 28) | **`0`, entry with `len == 0`, value silently dropped** | `-140` |
| HW_NAME_OID + IA5String | **`-140`** | **`0`, silently accepted** | `-140` |

Consequences introduced by the diff:
1. `SetDNSEntry()` (asn.c:14970) executes `XMEMCPY(dnsEntry_name, NULL, 0)`. Passing NULL to `memcpy` is undefined behavior (C99 7.1.4) and is diagnosed by UBSan's `nonnull-attribute` check.
2. A FASC-N OtherName carrying an IA5String now yields a zero-length entry with `oidSum == FASCN_OID`, so `wc_GetFASCNFromCert()` hands callers an…

**Fix:** Validate in `DecodeOtherHelper()` that the CHOICE alternative actually decoded matches the type OID, returning `ASN_PARSE_E` when the OID's slot is empty (`buf == NULL`). Additionally map `UPN_OID` onto the BEID slot so an IA5String-encoded UPN keeps working, matching `asn_orig.c`'s behaviour.

wolfssl/wolfcrypt/settings.h
```diff
@@ -4199,6 +4199,10 @@
     #if defined(WOLFSSL_ASN_TEMPLATE) && !defined(WOLFSSL_ACERT)
         #define WOLFSSL_ACERT
     #endif
+
+    /* Delay-Tolerant Networking (RFC 9174) support */
```
dgarske, line 4278: **🟠 [Medium] WOLFSSL_DTN has no configure/CMake option and no documentation**

`WOLFSSL_DTN` is only reachable by hand-defining `-DWOLFSSL_DTN` or implicitly via `WOLFSSL_ASN_ALL` (i.e. `--enable-all` / `--enable-asn=all`). The comparable `WOLFSSL_FPKI` gets its own configure option (`configure.ac:1377`). Adding the macro to `.wolfssl_known_macro_extras` is the correct workaround for a macro with no configure origin, but it also signals the gap. The PR checklist leaves "updated/added doxygen", "updated appropriate READMEs" and "Updated manual and documentation" unchecked; the only documentation added is the `ASN Options:` block comment in `asn.c:117-118`, which is good but not discoverable by users.

**Fix:** Add an `--enable-dtn` configure option (and the CMake equivalent), and mention RFC 9174 / id-on-bundleEID support in the README build-options list.

wolfcrypt/src/asn.c
```diff
@@ -19435,6 +19445,9 @@ enum {
     OTHERNAMEASN_IDX_VALUE,
     OTHERNAMEASN_IDX_UPN,
     OTHERNAMEASN_IDX_FASCN,
+#ifdef WOLFSSL_DTN
+    OTHERNAMEASN_IDX_BEID,
+#endif
     OTHERNAMEASN_IDX_HWN_SEQ,
     OTHERNAMEASN_IDX_HWN_TYPE,
     OTHERNAMEASN_IDX_HWN_NUM
```
dgarske, line 20218: **🔴 [High] DecodeSEP silently accepts a Hardware Module Name whose value is an IA5String, producing hwTypeSz/hwSerialNumSz of 0**

Same root cause as the finding above but a distinct failure site and a distinct fix location. `DecodeSEP()` unconditionally trusts that the `HWN_SEQ` CHOICE alternative was the one matched, and reads `OTHERNAMEASN_IDX_HWN_TYPE` / `OTHERNAMEASN_IDX_HWN_NUM` without checking they were populated. Before this diff that was safe: the choice group had no member an IA5String could satisfy, so `GetASN_Items()` returned `ASN_PARSE_E` for a `HW_NAME_OID` OtherName with an IA5String value. After this diff the IA5String matches the new BEID member, `GetASN_Items()` succeeds, and `DecodeSEP()` runs with `oidLen == 0` and `serialLen == 0`.

Verified empirically on this branch with `--enable-all` (defines both `WOLFSSL_SEP` and `WOLFSSL_DTN`), flipping the OID's last byte in `certs/bundle-eid-cert.der` from `0x0b` to `0x04`:
- before the template change: `wc_ParseCert() == -140` (ASN_PARSE_E)
- after: `wc_ParseCert() == 0`, `cert.hwTypeSz == 0`, `cert.hwSerialNumSz == 0`, and the OtherName vanishes from `cert.altNames` entirely.

A malformed Hardware Module Name identity is therefore silently swallowed rather than rejected. Two secondary effects: `XMALLOC(0)` is called for `cert->hwType` — on platforms where `malloc(0)` returns NULL this produces `MEMORY_E` from `wc_ParseCert()` instead of the correct `ASN_PARSE_E` — and both `XMEMCPY()` calls at asn.c:19476 and asn.c:19491 receive a NULL source with length 0 (UB per C99 7.1.4, flagged by UBSan's `nonnull-attribute` check).

`asn_orig.c` is unaffected: `DecodeSepHwAltName()` re-checks the tag itself and still returns `ASN_PARSE_E` for…

**Fix:** Add an explicit check in `DecodeSEP()` that the `HWN_SEQ` alternative was the one decoded, returning `ASN_PARSE_E` otherwise. This restores the pre-diff rejection of a `HW_NAME_OID` OtherName carrying a non-SEQUENCE value and keeps `WOLFSSL_SEP` behaviour identical whether or not `WOLFSSL_DTN` is enabled.

---
> **Note:** Referenced line (`wolfcrypt/src/asn.c:19459-19497`) is outside the diff hunk. Comment anchored to nearest changed region.

dgarske (commented):
## Skoll Code Review

**Findings:** 11 total — 3 posted, 8 skipped

### Posted findings
- [High] **New IA5String CHOICE alternative makes mismatched OtherName value types parse successfully instead of failing** — `wolfcrypt/src/asn.c:19436-19438, 19518-19524`
- [High] **DecodeSEP silently accepts a Hardware Module Name whose value is an IA5String, producing hwTypeSz/hwSerialNumSz of 0** — `wolfcrypt/src/asn.c:19459-19497`
- [Medium] **WOLFSSL_DTN has no configure/CMake option and no documentation** — `wolfssl/wolfcrypt/settings.h:4203-4205`

### Skipped findings
- [Medium] ~~No negative test coverage for OtherName OID/value-type mismatch~~
- [Medium] ~~certs/bundle-eid-cert.der does not chain to the repo's certs/ca-ecc-cert.pem~~
- [Medium] ~~DNS_entry.oidSum is not compiled in a WOLFSSL_DTN-only build, leaving the decoded bundleEID unidentifiable~~
- [Medium] ~~Gating the IA5String CHOICE alternative behind WOLFSSL_DTN leaves the reported parse failure in place for non-DTN builds~~
- [Low] ~~Test loads its certificate from disk while every other test in the file embeds DER inline~~
- [Low] ~~2 KB stack buffer for a 690-byte certificate~~
- [Low] ~~Magic length literals 14 and 13 duplicated in the SAN comparisons~~
- [Info] ~~Trailing blank line added at end of certs/renewcerts/wolfssl.cnf~~

dgarske (changes requested):
https://github.com/wolfSSL/wolfssl/pull/10810#pullrequestreview-4849148751

kareem-wolfssl:
Retest this please

philljj (changes requested):
Some of the Fenrir findings look valid, please review.

tests/api/test_asn.c
```diff
@@ -1538,6 +1538,204 @@ int test_DecodeAltNames_length_underflow(void)
...
+     * value of certs/fpki-cert.der to an IA5String must still parse and
+     * yield the same entry. */
+    {
+        /* UPN type-id (1.3.6.1.4.1.311.20.2.3). */
+        static const byte upnOid[] = {
+            0x06, 0x0a, 0x2b, 0x06, 0x01, 0x04, 0x01, 0x82, 0x37, 0x14,
+            0x02, 0x03
+        };
+        /* Offsets into fpki-cert.der, as above. */
+        #define UPN_OID_IDX 825
```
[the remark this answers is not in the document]
kareem-wolfssl: Intentional to avoid having to scan the entire cert.  Cert is generated with static parameters (serial etc) so that the offset is constant.

wolfssl/wolfcrypt/settings.h
```diff
@@ -4246,6 +4246,10 @@
...
+    #undef  WOLFSSL_DTN
+    #define WOLFSSL_DTN
```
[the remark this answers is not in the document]
kareem-wolfssl: Intentional as requested by earlier code review feedback

PR #10813, opened 2026-06-30
OpenSSL compatibility for libodbc

kojo1:
# Description

OpenSSL compatibility for libodbc.
- VERIFY_NONE by default when WOLFSSL_VERIFY_NONE_DEFAULT is enabled
- Added constant and macro functions

Fixes zd#21884

# Testing

Integration test with:

[ODBC application]
  ↓ unixODBC (libodbc.2)
  ├─→ psqlodbcw.so   ─→ libpq.5.dylib ─┐
  └─→ libtdsodbc.so  ─────────┴─→ libwolfssl.44.dylib 

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


kojo1:
Retest please.

wolfssl/openssl/objects.h
```diff
@@ -74,6 +74,20 @@
 #define NID_ad_OCSP WC_NID_ad_OCSP
 #define NID_ad_ca_issuers WC_NID_ad_ca_issuers
 
+/* OBJ_find_sigid_algs(): report SHA-256 / RSA for libpq's
+ * RSA-with-SHA-256 channel binding. Literal NIDs (672, 6) is for
+ * ASN is disabled. */
+#ifndef BUILDING_WOLFSSL
+static WC_INLINE int
```
dgarske: **🔴 [High] OBJ_find_sigid_algs stub hardcodes SHA-256/RSA and ignores sigid**

The new `wolfSSL_OBJ_find_sigid_algs()` casts away its `sigid` input (`(void)sigid;`) and unconditionally reports `*pdig = 672` (NID_sha256) and `*ppkey = 6` (NID_rsaEncryption) for ALL signature algorithms, then returns success (1). This is an always-success stub returning fixed values regardless of input, which the wolfSSL C coding conventions explicitly forbid ("No always-success stubs - return a 'not implemented' error instead"). Functionally it is wrong for any certificate not signed with `sha256WithRSAEncryption`: for an ECDSA cert or an RSA cert signed with SHA-384/SHA-512, this reports SHA-256+RSA. libpq's tls-server-end-point channel binding derives the binding hash from the server certificate's signature algorithm (via `X509_get_signature_nid()` -> `OBJ_find_sigid_algs()`); reporting the wrong digest/pkey will compute the wrong channel-binding hash and cause the bound connection to fail against a real peer, or silently produce a non-interoperable binding. The narrow libpq use case (RSA+SHA-256) works, but the function is a latent correctness/interop defect for the general case and gives no way for the caller to detect the substitution.

**Fix:** Do not ship an always-success stub that returns fixed crypto parameters. Either map the real signature algorithm, or return 0 for anything other than the algorithm(s) actually supported so callers can detect the unsupported case.

src/internal.c
```diff
@@ -2642,6 +2642,12 @@ int InitSSL_Ctx(WOLFSSL_CTX* ctx, WOLFSSL_METHOD* method, void* heap)
 
     XMEMSET(ctx, 0, sizeof(WOLFSSL_CTX));
 
+#ifdef WOLFSSL_VERIFY_NONE_DEFAULT
```
dgarske: **🟠 [Medium] No test coverage for WOLFSSL_VERIFY_NONE_DEFAULT security-relevant default change**

The new `WOLFSSL_VERIFY_NONE_DEFAULT` build option flips the default peer-verification behavior of a freshly created CTX (and every SSL it spawns, via `ssl->options.verifyNone = ctx->verifyNone`) from verify to no-verify. This is a security-relevant default and there is no test asserting (a) that with the macro set a new CTX has `verifyNone` and an SSL inherits it, and (b) that a subsequent `wolfSSL_CTX_set_verify(ctx, WOLFSSL_VERIFY_PEER, ...)` correctly clears the default. The PR checklist 'added tests' is unchecked.

**Fix:** Add an API test (guarded on WOLFSSL_VERIFY_NONE_DEFAULT) that creates a CTX, checks the default is VERIFY_NONE, verifies an SSL inherits it, and verifies wolfSSL_CTX_set_verify(WOLFSSL_VERIFY_PEER) overrides it. Also confirm interaction with wolfSSL_CTX_set_verify(WOLFSSL_VERIFY_DEFAULT), which resets verifyNone back to 0.

wolfssl/openssl/err.h
```diff
@@ -38,10 +38,21 @@
 #define WOLFSSL_SSL_F_SSL_CTX_USE_CERTIFICATE_FILE      2
 #define WOLFSSL_SSL_F_SSL_USE_PRIVATEKEY                3
 #define WOLFSSL_EC_F_EC_GFP_SIMPLE_POINT2OCT            4
+#define WOLFSSL_SSL_F_SSL_SET_FD                        5
 
 /* reasons */
 #define WOLFSSL_ERR_R_SYS_LIB                           1
 #define WOLFSSL_PKCS12_R_MAC_VERIFY_FAILURE             2
+#define WOLFSSL_ERR_R_BUF_LIB                           0
```
dgarske: **🔵 [Low] WOLFSSL_ERR_R_BUF_LIB defined as 0 (conflicts with 'no error' and OpenSSL's value)**

`WOLFSSL_ERR_R_BUF_LIB` (mapped to `ERR_R_BUF_LIB`) is defined as 0. An error reason value of 0 is conventionally 'no error/unset', and 0 is also used as the function-code sentinel in this file's `wolfSSL_SSLerr(0,...)` macros, so an app doing `if (ERR_GET_REASON(e) == ERR_R_BUF_LIB)` could match on a cleared/zero reason. It also does not match OpenSSL, where `ERR_R_BUF_LIB` is a nonzero library reason code. The other reasons in this block are nonzero (SYS_LIB=1, MAC_VERIFY_FAILURE=2).

**Fix:** Give ERR_R_BUF_LIB a distinct nonzero value (matching OpenSSL's convention if the consumer compares raw values), or confirm the consumer never treats 0 as a meaningful reason.

wolfssl/openssl/objects.h
```diff
@@ -74,6 +74,20 @@
 #define NID_ad_OCSP WC_NID_ad_OCSP
 #define NID_ad_ca_issuers WC_NID_ad_ca_issuers
 
+/* OBJ_find_sigid_algs(): report SHA-256 / RSA for libpq's
```
dgarske: **🔵 [Low] Garbled comment on OBJ_find_sigid_algs literal NIDs**

The comment 'Literal NIDs (672, 6) is for ASN is disabled.' is grammatically broken and unclear (double 'is'). It appears to intend 'Literal NIDs (672, 6) are used in case ASN is disabled.'

**Fix:** Fix the comment wording.

wolfssl/openssl/ssl.h
```diff
@@ -1568,6 +1569,12 @@ typedef WOLFSSL_SRTP_PROTECTION_PROFILE      SRTP_PROTECTION_PROFILE;
 #define SSL_get_state                   wolfSSL_get_state
 #define SSL_state_string_long           wolfSSL_state_string_long
 
+#define WOLFSSL_TLS_ST_OK               16
```
dgarske: **🔵 [Low] TLS_ST_OK hardcoded to 16, coupled to internal enum position**

`WOLFSSL_TLS_ST_OK` is defined as the literal 16. This is correct today because `wolfSSL_get_state()` returns `ssl->options.handShakeState`, and `HANDSHAKE_DONE` is the 17th member (index 16) of the internal `enum states`. However, 16 is a magic number duplicating an internal enum value: if any state is ever inserted before `HANDSHAKE_DONE`, this constant silently becomes wrong. The internal enum already carries a warning comment ('Adding state before HANDSHAKE_DONE will break session importing'), so risk is low, but the coupling is not obvious from the public header.

**Fix:** Add a comment noting that 16 must equal the internal HANDSHAKE_DONE enum value (wolfssl/internal.h enum states), so a future editor keeps them in sync.

dgarske (commented):
## Skoll Code Review

### Posted findings
- [High] **OBJ_find_sigid_algs stub hardcodes SHA-256/RSA and ignores sigid** — `wolfssl/openssl/objects.h:81-88`
- [Medium] **No test coverage for WOLFSSL_VERIFY_NONE_DEFAULT security-relevant default change** — `src/internal.c:2645-2649`
- [Low] **WOLFSSL_ERR_R_BUF_LIB defined as 0 (conflicts with 'no error' and OpenSSL's value)** — `wolfssl/openssl/err.h:46`
- [Low] **Garbled comment on OBJ_find_sigid_algs literal NIDs** — `wolfssl/openssl/objects.h:77-79`
- [Low] **TLS_ST_OK hardcoded to 16, coupled to internal enum position** — `wolfssl/openssl/ssl.h:1572`

#### SSL_R_UNKNOWN_PROTOCOL / WRONG_VERSION_NUMBER / UNSUPPORTED_PROTOCOL numeric value changed
**File:** `wolfssl/openssl/ssl.h:1748-1753, wolfssl/openssl/err.h:47-49`
**Function:** `N/A (macros)`
**Severity:** Medium

These three pre-existing public compat macros were changed from expanding to `VERSION_ERROR` (a negative wolfSSL internal error enum) to OpenSSL numeric reason codes (`WOLFSSL_SSL_R_UNKNOWN_PROTOCOL`=252, `WOLFSSL_SSL_R_WRONG_VERSION_NUMBER`=267, `WOLFSSL_SSL_R_UNSUPPORTED_PROTOCOL`=258). This is more OpenSSL-accurate, but it is a behavior change: any existing application that compared these macros against wolfSSL's internally reported error code (`VERSION_ERROR`, e.g. from `wolfSSL_get_error`/`SSL_get_error`) will no longer match, because wolfSSL still reports `VERSION_ERROR` internally, not 252/267/258. I confirmed there are no internal `src/*.c` or `tests/*.c` users of these specific macros (tests compare against `VERSION_ERROR` directly), so the build and current tests are unaffected, but downstream consumers relying on the old mapping would silently break.

**Recommendation:** Confirm this value change is intended and note it as a compatibility change in the PR/changelog. If wolfSSL's error surface (ERR_get_error/ERR_GET_REASON) does not translate internal VERSION_ERROR into these OpenSSL reason codes, apps checking `reason == SSL_R_WRONG_VERSION_NUMBER` will never match - verify the intended consumer (libodbc/libtds) actually reads these as raw constants and not via wolfSSL's error queue.

_Referenced code: `wolfssl/openssl/ssl.h:1748-1753, wolfssl/openssl/err.h:47-50` (4 lines)_

---

#### BIO compat stubs have latent multi-use hazards (fixed index, app_data aliasing)
**File:** `wolfssl/openssl/bio.h:196, wolfssl/openssl/ssl.h:1771-1774`
**Function:** `wolfSSL_BIO_get_new_index`
**Severity:** Medium

Two BIO compat shims work for a single custom BIO but are unsafe if used more than once: (1) `wolfSSL_BIO_get_new_index()` always returns the constant 1000 - in OpenSSL each call returns a distinct index, so an app that registers two custom BIO types would get identical type values and collide. (2) `BIO_get_app_data`/`BIO_set_app_data` are mapped onto `wolfSSL_BIO_get_data`/`wolfSSL_BIO_set_data`, which is the SAME storage slot that a custom BIO method already uses for its method-private data. In OpenSSL these are two independent slots; collapsing them means an app that uses both `BIO_set_data` (method state) and `BIO_set_app_data` (app state) on the same BIO will overwrite one with the other. The `wolfSSL_BIO_meth_get_*` getters returning NULL and `set_callback_ctrl` being a no-op are acceptable documented stubs.

**Recommendation:** Keep for the narrow libodbc/libtds use, but document the single-custom-BIO limitation in the header comment. If feasible, back get_new_index with a static incrementing counter, and keep app_data in a separate slot from method data to avoid silent corruption when both are used.

_Referenced code: `wolfssl/openssl/bio.h:196, wolfssl/openssl/ssl.h:1771-1774` (4 lines)_

---

dgarske (changes requested):
https://github.com/wolfSSL/wolfssl/pull/10813#pullrequestreview-4638899860

src/internal.c
kojo1: Added test_wolfSSL_CTX_verify_none_default(void)

wolfssl/openssl/objects.h
kojo1: Improved OBJ_find_sigid_algs

wolfssl/openssl/err.h
kojo1: Fixed the comment.

wolfssl/openssl/objects.h
kojo1: Fixed the comment.

wolfssl/openssl/ssl.h
kojo1: Fixed the comment.

dgarske:
@kojo1 can you rebase to latest master? That might resolve the CI issues

dgarske (commented):
## Skoll Code Review

### Posted findings
- [Medium] **OBJ_find_sigid_algs compat macro placed in ssl.h instead of objects.h with the other OBJ_* macros** — `wolfssl/openssl/ssl.h:1785`
- [Medium] **WOLFSSL_TLS_ST_OK hardcoded to 16, duplicating internal HANDSHAKE_DONE with no compile-time check** — `wolfssl/openssl/ssl.h:1572-1577`
- [Medium] **BIO_get_app_data/BIO_set_app_data aliased to BIO_get_data/BIO_set_data (shared storage)** — `wolfssl/openssl/ssl.h:1773-1775`
- [Medium] **No test coverage for new wolfSSL_OBJ_find_sigid_algs mapping logic** — `wolfssl/openssl/objects.h:79-127`
- [Low] **BIO compat stubs report success/fixed values with silent limitations** — `wolfssl/openssl/bio.h:194-220`
- [Low] **wolfSSL_OBJ_find_sigid_algs omits RSA-PSS, EdDSA, and MD5 signature NIDs** — `wolfssl/openssl/objects.h:83-121`

### Skipped findings
- [Medium] ~~SSL_R_* compat macros silently change value from VERSION_ERROR to positive OpenSSL codes~~

wolfssl/openssl/ssl.h
```diff
@@ -1738,16 +1747,43 @@ typedef WOLFSSL_SRTP_PROTECTION_PROFILE      SRTP_PROTECTION_PROFILE;
...
+#define BIO_get_new_index wolfSSL_BIO_get_new_index
+#define BIO_meth_get_gets wolfSSL_BIO_meth_get_gets
+#define BIO_meth_get_puts wolfSSL_BIO_meth_get_puts
+#define BIO_meth_get_ctrl wolfSSL_BIO_meth_get_ctrl
+#define BIO_meth_get_create wolfSSL_BIO_meth_get_create
+#define BIO_meth_get_destroy wolfSSL_BIO_meth_get_destroy
+#define BIO_meth_get_callback_ctrl wolfSSL_BIO_meth_get_callback_ctrl
+#define BIO_meth_set_callback_ctrl wolfSSL_BIO_meth_set_callback_ctrl
+#ifndef BUILDING_WOLFSSL
+#define OBJ_find_sigid_algs wolfSSL_OBJ_find_sigid_algs
```
dgarske: **🟠 [Medium] OBJ_find_sigid_algs compat macro placed in ssl.h instead of objects.h with the other OBJ_* macros**

Every other `OBJ_*` -> `wolfSSL_OBJ_*` compat macro is defined in `wolfssl/openssl/objects.h` (lines 53-69), co-located with the function it aliases. This PR defines the function `wolfSSL_OBJ_find_sigid_algs` in `objects.h` but puts the `#define OBJ_find_sigid_algs wolfSSL_OBJ_find_sigid_algs` alias in `ssl.h` (broad OPENSSL_EXTRA block), decoupling the macro from its definition. `ssl.h` only `#include`s `objects.h` under `WOLFSSL_QT || OPENSSL_ALL` (line 52-55); it works for plain OPENSSL_EXTRA builds only because `evp.h` happens to include `objects.h` at its tail. This transitive dependency is fragile: any build path where `evp.h` is skipped (e.g. WOLFCRYPT_ONLY) would define the macro while leaving `wolfSSL_OBJ_find_sigid_algs` undeclared, yielding an implicit-declaration error at the call site.

**Fix:** Move the `#define OBJ_find_sigid_algs wolfSSL_OBJ_find_sigid_algs` into objects.h next to the function definition and the other OBJ_* aliases, so the macro and its target are always defined together.

wolfssl/openssl/ssl.h
```diff
@@ -1568,6 +1569,14 @@ typedef WOLFSSL_SRTP_PROTECTION_PROFILE      SRTP_PROTECTION_PROFILE;
 #define SSL_get_state                   wolfSSL_get_state
 #define SSL_state_string_long           wolfSSL_state_string_long
 
+/* Must equal HANDSHAKE_DONE in the internal 'enum states'
```
dgarske: **🟠 [Medium] WOLFSSL_TLS_ST_OK hardcoded to 16, duplicating internal HANDSHAKE_DONE with no compile-time check**

`WOLFSSL_TLS_ST_OK` (and its aliases `TLS_ST_OK`/`SSL_ST_OK`) is hardcoded to 16 to match `HANDSHAKE_DONE` in the private `enum states` (wolfssl/internal.h). The value is correct today (HANDSHAKE_DONE is the 17th enumerator, index 16, and `wolfSSL_get_state()` returns `ssl->options.handShakeState`), but the coupling is a magic number with no static assertion. The comment even admits `16 today`. Because `internal.h` isn't a public header, a `static_assert` can't live in the public header, so a silent drift is possible if a state is ever inserted before HANDSHAKE_DONE (the enum comment warns this also breaks session import).

**Fix:** Add a compile-time check in a .c file that includes internal.h, e.g. `wc_static_assert(WOLFSSL_TLS_ST_OK == HANDSHAKE_DONE);`, so any future enum drift fails the build instead of silently returning a wrong state constant.

wolfssl/openssl/ssl.h
```diff
@@ -1738,16 +1747,43 @@ typedef WOLFSSL_SRTP_PROTECTION_PROFILE      SRTP_PROTECTION_PROFILE;
...
+#define SSL_R_TLSV1_ALERT_PROTOCOL_VERSION \
+    WOLFSSL_SSL_R_TLSV1_ALERT_PROTOCOL_VERSION
 #define SSL_R_CERTIFICATE_VERIFY_FAILED            VERIFY_CERT_ERROR
 #define SSL_R_CERT_CB_ERROR                        CLIENT_CERT_CB_ERROR
 #define SSL_R_NULL_SSL_METHOD_PASSED               BAD_FUNC_ARG
 #define SSL_R_CCS_RECEIVED_EARLY                   OUT_OF_ORDER_E
+#define ERR_R_BUF_LIB WOLFSSL_ERR_R_BUF_LIB
+#define BIO_TYPE_DESCRIPTOR WOLFSSL_BIO_TYPE_DESCRIPTOR
+#define BIO_TYPE_SOURCE_SINK WOLFSSL_BIO_TYPE_SOURCE_SINK
+#define BIO_get_app_data(bio) wolfSSL_BIO_get_data(bio)
```
dgarske: **🟠 [Medium] BIO_get_app_data/BIO_set_app_data aliased to BIO_get_data/BIO_set_data (shared storage)**

In OpenSSL, `BIO_set_data`/`BIO_get_data` access the method-implementation pointer, while `BIO_set_app_data`/`BIO_get_app_data` access separate application ex-data. This PR maps `BIO_get_app_data`/`BIO_set_app_data` onto the same `wolfSSL_BIO_get_data`/`wolfSSL_BIO_set_data` slot that `BIO_get_data`/`BIO_set_data` already use (bio.h:163-164). A custom source/sink BIO (the ODBC/FreeTDS transport this PR targets) that stores its connection state via BIO_set_data in its create/ctrl callbacks and also uses BIO_set_app_data for an application pointer will have the two collide - whichever is written last silently overwrites the other.

**Fix:** Back app_data with a distinct storage slot (e.g. BIO ex_data index) so it does not alias the method data pointer, or document the limitation prominently if the aliasing is acceptable for the intended consumers.

wolfssl/openssl/objects.h
```diff
@@ -74,6 +74,53 @@
 #define NID_ad_OCSP WC_NID_ad_OCSP
 #define NID_ad_ca_issuers WC_NID_ad_ca_issuers
 
+/* OBJ_find_sigid_algs(): map a certificate signature-algorithm NID to its
+ * digest and public-key NIDs; returns 0 for unsupported algorithms. */
+#ifndef BUILDING_WOLFSSL
```
dgarske: **🟠 [Medium] No test coverage for new wolfSSL_OBJ_find_sigid_algs mapping logic**

The PR adds a non-trivial mapping function (a switch over ~10 signature-algorithm NIDs producing digest+pkey NID pairs, returning 0 for unsupported), but adds no unit test for it. `grep` finds no reference to `OBJ_find_sigid_algs` in tests/, examples/, or testsuite/. Mapping tables are exactly the kind of code where a transposed NID pair would go unnoticed. The verifyNone default got a good targeted test (test_wolfSSL_CTX_verify_none_default); this helper deserves the same.

**Fix:** Add an api.c test that calls OBJ_find_sigid_algs for a couple of RSA and ECDSA sig NIDs, asserts the returned digest/pkey NIDs and return==1, verifies NULL out-params are tolerated, and asserts an unsupported NID returns 0 without touching the out-params.

wolfssl/openssl/bio.h
```diff
@@ -188,6 +188,37 @@
 #define BIO_meth_set_create        wolfSSL_BIO_meth_set_create
 #define BIO_meth_set_destroy       wolfSSL_BIO_meth_set_destroy
 
+#define WOLFSSL_BIO_TYPE_DESCRIPTOR  0x0100
+#define WOLFSSL_BIO_TYPE_SOURCE_SINK 0x0400
+
+/* OpenSSL allocates a fresh BIO type index per call; wolfSSL
```
dgarske: **🔵 [Low] BIO compat stubs report success/fixed values with silent limitations**

`wolfSSL_BIO_get_new_index()` returns a fixed 1000 for every call, so two distinct app-defined BIO types receive the same index and are indistinguishable if compared. `wolfSSL_BIO_meth_set_callback_ctrl()` is an always-success no-op that discards the callback and returns 1, which conflicts with the wolfSSL guideline of not shipping always-success stubs (a caller that relies on the callback later being invoked gets no indication it was dropped). The getters correctly return NULL. These mirror OpenSSL's return conventions and are documented in the comments, so this is a low-priority observation rather than a defect.

**Fix:** If uniqueness is ever needed, hand out a monotonically increasing index instead of a constant. Keep the comment explicit that callback_ctrl is unsupported so callers don't assume the callback fires.

wolfssl/openssl/objects.h
```diff
@@ -74,6 +74,53 @@
...
+static WC_INLINE int
+wolfSSL_OBJ_find_sigid_algs(int sigid, int *pdig, int *ppkey)
+{
+    int dig  = 0;
```
dgarske: **🔵 [Low] wolfSSL_OBJ_find_sigid_algs omits RSA-PSS, EdDSA, and MD5 signature NIDs**

The mapping covers SHA1/224/256/384/512 with RSA and ECDSA but returns 0 (unsupported) for other algorithms wolfSSL/OpenSSL know, e.g. NID_rsassaPss (WC_NID_rsassaPss=912), Ed25519/Ed448, and md5WithRSAEncryption (WC_NID_md5WithRSAEncryption=99), all of which are already defined in evp.h. For EdDSA in particular, OpenSSL's OBJ_find_sigid_algs returns 1 with pkey set and digest = undef; returning 0 here diverges and may cause a compat consumer's certificate-algorithm introspection to treat those certs as unknown. This is acceptable as a documented limitation for the current use case but is a divergence from OpenSSL semantics.

**Fix:** Consider extending the switch (guarded by the relevant feature macros) to cover RSA-PSS and Ed25519/Ed448 to match OpenSSL, or note the intentional scope limitation in the function comment.

wolfssl/openssl/ssl.h
kojo1: Moved it to ssl.h.

wolfssl/openssl/ssl.h
kojo1: Added wc_static_assert(WOLFSSL_TLS_ST_OK == HANDSHAKE_DONE)

wolfssl/openssl/ssl.h
kojo1: Back app_data with BIO ex_data slot 0.

wolfssl/openssl/objects.h
kojo1: Added TEST_DECL_GROUP("ossl_obj", test_wolfSSL_OBJ_find_sigid_algs).

wolfssl/openssl/bio.h
kojo1: Made it use BIO_get_new_index.

wolfssl/openssl/objects.h
kojo1: Added ED25519/448 cases.

kojo1:
jenkins, Retest this please.

src/bio.c
```diff
@@ -2119,6 +2119,27 @@ long wolfSSL_BIO_set_nbio(WOLFSSL_BIO* bio, long on)
...
 
+#ifndef WOLFSSL_BIO_TYPE_START
+/* Matches WOLFSSL_BIO_TYPE_START in wolfssl/openssl/bio.h (OpenSSL's
+ * BIO_TYPE_START), which is only included with OPENSSL_EXTRA. */
+#define WOLFSSL_BIO_TYPE_START 128
+#endif
+
+/* Return a new BIO type index, or WOLFSSL_FATAL_ERROR when exhausted.
+ * Not thread-safe. */
+int wolfSSL_BIO_get_new_index(void)
```
dgarske, line 2126: **🟠 [Medium] New public API wolfSSL_BIO_get_new_index has no test coverage**

This PR adds a new `WOLFSSL_API` function `wolfSSL_BIO_get_new_index()` (declared unconditionally in `wolfssl/ssl.h`), but no unit test exercises it. Grepping `tests/` finds no reference to `BIO_get_new_index`. The function has non-trivial behavior worth covering: the first call must return `WOLFSSL_BIO_TYPE_START + 1` (129), successive calls must monotonically increase, and the exhaustion boundary (`>= 0xff` returns `WOLFSSL_FATAL_ERROR`) is an off-by-one-prone edge that should be pinned by a test. The sibling `wolfSSL_OBJ_find_sigid_algs` and the `verifyNone` default both received dedicated tests in this same PR, so the omission here is inconsistent.

**Fix:** Add a focused unit test (e.g. test_wolfSSL_BIO_get_new_index) covering the initial value and monotonic increment; guard it under OPENSSL_EXTRA like the other BIO tests.

dgarske, line 2126: **🟠 [Medium] wolfSSL_BIO_get_new_index is not thread-safe, unlike OpenSSL's BIO_get_new_index**

OpenSSL's `BIO_get_new_index()` is explicitly thread-safe (it uses an atomic add under a lock). This compat shim uses a plain `static int` with a non-atomic `++`, and the header comment documents it as "Not thread-safe." The stated use case (PR description) is libodbc, where two ODBC drivers (psqlodbcw.so and libtdsodbc.so) both link the same shared `libwolfssl` and may register custom BIO types concurrently during driver load. Concurrent calls could hand out duplicate indices or observe a torn counter. In practice BIO-type registration is usually one-shot at init, so this is low-likelihood, but it is a behavioral divergence from the OpenSSL API contract being emulated.

**Fix:** Either add mutex protection to match OpenSSL semantics, or keep as-is but ensure the "not thread-safe" limitation is surfaced in the public documentation (not just an internal comment), since this is a drop-in compat function.

dgarske (commented):
## Skoll Code Review

### Posted findings
- [Medium] **New public API wolfSSL_BIO_get_new_index has no test coverage** — `src/bio.c:2130-2141`
- [Medium] **wolfSSL_BIO_get_new_index is not thread-safe, unlike OpenSSL's BIO_get_new_index** — `src/bio.c:2130-2141`

### Skipped findings
- [Low] ~~Missing WOLFSSL_LEAVE on error return path~~

src/bio.c
```diff
@@ -2119,6 +2119,27 @@ long wolfSSL_BIO_set_nbio(WOLFSSL_BIO* bio, long on)
...
+{
+    static int wolfssl_bio_type_index = WOLFSSL_BIO_TYPE_START;
+
+    WOLFSSL_ENTER("wolfSSL_BIO_get_new_index");
```
dgarske: This might be noisy, consider removing it... also missing a WOLFSSL_LEAVE if you decide to keep it.

dgarske (changes requested):

src/bio.c
kojo1: It was removed.

src/bio.c
kojo1: Added the test.

kojo1: Fixed by making the counter atomic.

dgarske (dismissed):
Over to @julek-wolfssl

src/bio.c
```diff
@@ -2119,6 +2119,40 @@ long wolfSSL_BIO_set_nbio(WOLFSSL_BIO* bio, long on)
...
+#else
+    static int wolfssl_bio_type_index = WOLFSSL_BIO_TYPE_START;
+
+    if (wolfssl_bio_type_index >= 0xff) {
+        return WOLFSSL_FATAL_ERROR;
+    }
+
+    return ++wolfssl_bio_type_index;
+#endif
+}
```
julek-wolfssl, line 2137: Don't bother with atomics. This should be routed to `wolfssl_local_get_ex_new_index` without the callbacks for now.

src/internal.c
```diff
@@ -2641,6 +2641,12 @@ int InitSSL_Ctx(WOLFSSL_CTX* ctx, WOLFSSL_METHOD* method, void* heap)
...
+    /* OpenSSL compat: default to SSL_VERIFY_NONE unless the app
+     * sets SSL_VERIFY_PEER. */
+    ctx->verifyNone = 1;
+#endif
```
julek-wolfssl: Isn't this already covered by `OPENSSL_COMPATIBLE_DEFAULTS`? Why a new macro?

src/internal.c
```diff
@@ -9817,13 +9823,15 @@ void WriteSEQ(WOLFSSL* ssl, int verifyOrder, byte* out)
...
+#if !defined(WOLFSSL_NO_TLS12) && defined(HAVE_AEAD)
 /* Same as WriteSEQ() but does not advance the per-direction TLS sequence
  * counter. Lets a caller share the 64-bit record sequence between fields
  * that are written more than once per record (e.g. AES-GCM nonce_explicit
  * and AAD seq_num): peek here, then a later WriteSEQ() inside
  * writeAeadAuthData() does the single mandated increment. For DTLS the
  * underlying GetSEQ is already read-only, so this is identical to
- * WriteSEQ() in that path. */
+ * WriteSEQ() in that path. Guarded to match its only caller in
+ * BuildMessage()'s TLS 1.2 AEAD path. */
```
julek-wolfssl: Why add this comment? Remove unnecessary comments.

src/ssl.c
```diff
@@ -8652,6 +8658,83 @@ void wolfSSL_THREADID_set_numeric(void* id, unsigned long val)
 
 #endif /* OPENSSL_ALL || OPENSSL_EXTRA */
 
+#if defined(OPENSSL_EXTRA) || defined(OPENSSL_EXTRA_X509_SMALL)
+/* Map a certificate signature-algorithm NID to its digest and public-key
+ * NIDs, like OpenSSL's OBJ_find_sigid_algs() (used by X509_get_signature_nid
+ * consumers such as libpq's tls-server-end-point channel binding).
+ * For RSA-PSS and EdDSA the digest is reported as WC_NID_undef, matching
+ * OpenSSL (the digest is carried in the algorithm parameters or fixed by
+ * the algorithm itself).
```
julek-wolfssl: This should be one sentence at most.

julek-wolfssl (changes requested):

src/bio.c
kojo1: Route BIO_get_new_index through wolfssl_local_get_ex_new_index.

src/internal.c
kojo1: Remove WOLFSSL_VERIFY_NONE_DEFAULT.

src/internal.c
kojo1: Clean up comment.

src/ssl.c
kojo1: Clean up comment.

wolfssl/openssl/bio.h
```diff
@@ -188,6 +188,38 @@
...
+{ (void)m; return NULL; }
+static WC_INLINE void *
+wolfSSL_BIO_meth_get_destroy(WOLFSSL_BIO_METHOD *m)
+{ (void)m; return NULL; }
+static WC_INLINE void *
+wolfSSL_BIO_meth_get_callback_ctrl(WOLFSSL_BIO_METHOD *m)
+{ (void)m; return NULL; }
+static WC_INLINE int
+wolfSSL_BIO_meth_set_callback_ctrl(WOLFSSL_BIO_METHOD *m, void *cb)
+{ (void)m; return (cb == NULL) ? 1 : 0; }
```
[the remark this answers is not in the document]
kojo1: Return stored callbacks from BIO_meth_get_* with proper callback types.

wolfssl/openssl/ssl.h
```diff
@@ -1739,16 +1749,43 @@ typedef WOLFSSL_SRTP_PROTECTION_PROFILE      SRTP_PROTECTION_PROFILE;
...
 #define SSL_R_CCS_RECEIVED_EARLY                   OUT_OF_ORDER_E
+#define ERR_R_BUF_LIB WOLFSSL_ERR_R_BUF_LIB
+#define BIO_TYPE_DESCRIPTOR WOLFSSL_BIO_TYPE_DESCRIPTOR
+#define BIO_TYPE_SOURCE_SINK WOLFSSL_BIO_TYPE_SOURCE_SINK
+#define BIO_TYPE_START WOLFSSL_BIO_TYPE_START
+/* Like OpenSSL, back app_data with BIO ex_data slot 0 so it does not share
+ * storage with BIO_get_data()/BIO_set_data() (requires HAVE_EX_DATA). */
+#define BIO_get_app_data(bio) wolfSSL_BIO_get_ex_data((bio), 0)
+#define BIO_set_app_data(bio, data) \
+    wolfSSL_BIO_set_ex_data((bio), 0, (data))
```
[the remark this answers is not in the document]
kojo1: Moved BIO compat macros to openssl/bio.h.

wolfssl/openssl/ssl.h
```diff
@@ -1739,16 +1749,43 @@ typedef WOLFSSL_SRTP_PROTECTION_PROFILE      SRTP_PROTECTION_PROFILE;
...
+#define SSL_R_TLSV1_ALERT_PROTOCOL_VERSION \
+    WOLFSSL_SSL_R_TLSV1_ALERT_PROTOCOL_VERSION
 #define SSL_R_CERTIFICATE_VERIFY_FAILED            VERIFY_CERT_ERROR
 #define SSL_R_CERT_CB_ERROR                        CLIENT_CERT_CB_ERROR
 #define SSL_R_NULL_SSL_METHOD_PASSED               BAD_FUNC_ARG
 #define SSL_R_CCS_RECEIVED_EARLY                   OUT_OF_ORDER_E
+#define ERR_R_BUF_LIB WOLFSSL_ERR_R_BUF_LIB
+#define BIO_TYPE_DESCRIPTOR WOLFSSL_BIO_TYPE_DESCRIPTOR
+#define BIO_TYPE_SOURCE_SINK WOLFSSL_BIO_TYPE_SOURCE_SINK
+#define BIO_TYPE_START WOLFSSL_BIO_TYPE_START
```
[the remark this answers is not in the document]
kojo1: Moved BIO compat macros to openssl/bio.h.

wolfssl/ssl.h
```diff
@@ -5188,6 +5189,7 @@ WOLFSSL_LOCAL int wc_OBJ_sn2nid(const char *sn);
 
 WOLFSSL_API const char* wolfSSL_OBJ_nid2sn(int n);
 WOLFSSL_API int wolfSSL_OBJ_obj2nid(const WOLFSSL_ASN1_OBJECT *o);
+WOLFSSL_API int wolfSSL_OBJ_find_sigid_algs(int sigid, int *pdig, int *ppkey);
```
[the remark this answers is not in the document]
kojo1: Moved BIO compat macros to openssl/bio.h.

src/bio.c
```diff
@@ -2119,6 +2119,30 @@ long wolfSSL_BIO_set_nbio(WOLFSSL_BIO* bio, long on)
...
+int wolfSSL_BIO_get_new_index(void)
+{
+    int idx;
+#ifdef HAVE_EX_DATA_CRYPTO
+    idx = wolfssl_local_get_ex_new_index(WOLF_CRYPTO_EX_INDEX_BIO, 0, NULL,
+            NULL, NULL, NULL);
+#else
+    static int wolfssl_bio_type_index = 0;
+    idx = wolfssl_bio_type_index++;
+#endif
```
julek-wolfssl: This doesn't make sense. The entire API is only used to get a new index for external data. This entire function should be gated by `#ifdef HAVE_EX_DATA_CRYPTO`.

julek-wolfssl (changes requested):

src/bio.c
kojo1: Gate wolfSSL_BIO_get_new_index by HAVE_EX_DATA_CRYPT0.

kojo1:
Removed WOLFSSL_VERIFY_NONE_DEFAULT

dgarske (dismissed):

dgarske:
Jenkins retest this please (FIPS ready config) history already gone.

src/bio.c
```diff
@@ -2119,6 +2119,26 @@ long wolfSSL_BIO_set_nbio(WOLFSSL_BIO* bio, long on)
...
+int wolfSSL_BIO_get_new_index(void)
+{
+    int idx = wolfssl_local_get_ex_new_index(WOLF_CRYPTO_EX_INDEX_BIO, 0, NULL,
+            NULL, NULL, NULL);
+    if (idx < 0 || idx > 0xff - (WOLFSSL_BIO_TYPE_START + 1)) {
+        return WOLFSSL_FATAL_ERROR;
+    }
+    return WOLFSSL_BIO_TYPE_START + 1 + idx;
+}
+#endif /* HAVE_EX_DATA_CRYPTO */
```
julek-wolfssl: So I was wrong here. This is not for external data. This is for custom BIO types. Don't use `wolfssl_local_get_ex_new_index`. Instead use a static variable initialized to `WOLFSSL_BIO_TYPE_START` and just do `return bio_idx++`.

src/ssl.c
```diff
@@ -8088,7 +8095,6 @@ int wolfssl_local_get_ex_new_index(int class_index, long ctx_l, void* ctx_ptr,
         case WOLF_CRYPTO_EX_INDEX_RSA:
         case WOLF_CRYPTO_EX_INDEX_ENGINE:
         case WOLF_CRYPTO_EX_INDEX_UI:
-        case WOLF_CRYPTO_EX_INDEX_BIO:
         case WOLF_CRYPTO_EX_INDEX_APP:
         case WOLF_CRYPTO_EX_INDEX_UI_METHOD:
         case WOLF_CRYPTO_EX_INDEX_DRBG:
```
julek-wolfssl, line 8093: Remove this change since I was wrong to request it.

julek-wolfssl (dismissed):

src/bio.c
kojo1: Use a static counter for BIO type indexes in BIO_get_new_index

src/ssl.c
kojo1: Restore WOLF_CRYPTO_EX_INDEX_BIO

wolfssl/openssl/ssl.h
```diff
@@ -1752,16 +1762,27 @@ typedef WOLFSSL_SRTP_PROTECTION_PROFILE      SRTP_PROTECTION_PROFILE;
 #define SSL_R_DATA_LENGTH_TOO_LONG                 BUFFER_ERROR
 #define SSL_R_ENCRYPTED_LENGTH_TOO_LONG            BUFFER_ERROR
 #define SSL_R_BAD_LENGTH                           BUFFER_ERROR
-#define SSL_R_UNKNOWN_PROTOCOL                     VERSION_ERROR
-#define SSL_R_WRONG_VERSION_NUMBER                 VERSION_ERROR
+#define SSL_R_UNKNOWN_PROTOCOL WOLFSSL_SSL_R_UNKNOWN_PROTOCOL
```
dgarske, line 1766: SSL_R_UNKNOWN_PROTOCOL / WRONG_VERSION_NUMBER / UNSUPPORTED_PROTOCOL remapped to OpenSSL numeric reason codes that wolfSSL never returns, reverting the nginx fix

wolfSSL's error queue stores its own internal error codes, not OpenSSL reason numbers. `wolfSSL_ERR_GET_REASON()` (src/ssl_err.c:436-470) returns the raw internal code when it falls inside the wolfCrypt/wolfSSL error spans:

Nothing in the library ever pushes 252 / 258 / 267 into the queue. So after this change, `ERR_GET_REASON(ERR_peek_error()) == SSL_R_UNKNOWN_PROTOCOL` can never be true, where before it matched every version error.

Recommendation: Do not change the three existing macros. Restore `VERSION_ERROR` for `SSL_R_UNKNOWN_PROTOCOL`, `SSL_R_WRONG_VERSION_NUMBER` and `SSL_R_UNSUPPORTED_PROTOCOL`, and add only the genuinely new reason codes with OpenSSL numeric values. If the numeric values really are required (e.g. to avoid duplicate `case` labels in a consumer's `switch`), then that has to be done for the whole `SSL_R_*` family at once AND `wolfSSL_ERR_GET_REASON()` must be taught to translate wolfSSL internal codes into those numeric reasons - otherwise the mapping is inert. Either way, add a regression test that drives a version mismatch and asserts `ERR_GET_REASON()` against `SSL_R_WRONG_VERSION_NUMBER`.

src/bio.c
```diff
@@ -2119,6 +2119,20 @@ long wolfSSL_BIO_set_nbio(WOLFSSL_BIO* bio, long on)
...
+/* Matches WOLFSSL_BIO_TYPE_START in wolfssl/openssl/bio.h (OpenSSL's
+ * BIO_TYPE_START), which is only included with OPENSSL_EXTRA. */
+#define WOLFSSL_BIO_TYPE_START 128
+#endif
+
+/* Return a new type index for a custom BIO, starting at
+ * WOLFSSL_BIO_TYPE_START like OpenSSL's BIO_get_new_index(). */
+int wolfSSL_BIO_get_new_index(void)
+{
+    static int bio_idx = WOLFSSL_BIO_TYPE_START;
```
dgarske, line 2134: wolfSSL_BIO_get_new_index() uses a non-atomic static counter - concurrent callers can receive the same type index

Recommendation: Restore the atomic increment (`wolfSSL_Atomic_Int_FetchAdd`), or guard the counter with a `wolfSSL_Mutex` initialised at library init. Add a multi-threaded test (guarded by `!SINGLE_THREADED`) that spawns N threads each calling `BIO_get_new_index()` and asserts all N returned values are distinct.

src/bio.c
```diff
@@ -2231,6 +2245,86 @@ int wolfSSL_BIO_meth_set_destroy(WOLFSSL_BIO_METHOD *biom,
...
+{
+    WOLFSSL_ENTER("wolfSSL_BIO_meth_get_callback_ctrl");
+    if (biom) {
+        return biom->ctrlInfoCb;
+    }
+    return NULL;
+}
+
+
+int wolfSSL_BIO_meth_set_callback_ctrl(WOLFSSL_BIO_METHOD *biom,
```
dgarske, line 2320: wolfSSL_BIO_meth_set_callback_ctrl() always fails for a non-NULL callback, making the paired getter dead code and breaking drop-in OpenSSL setup sequences

OpenSSL's `BIO_meth_set_callback_ctrl()` unconditionally stores the callback and returns 1. Since the whole point of this PR is drop-in compatibility for libodbc/freetds, a caller that follows the usual defensive porting pattern (`if (!BIO_meth_set_callback_ctrl(m, cb)) goto err;`) now aborts BIO method construction where it previously would not even have compiled - i.e. the newly-added symbol turns a link error into a silent runtime setup failure, which is harder to diagnose.

Second, the assignment `biom->ctrlInfoCb = NULL;` is a no-op: `wolfSSL_BIO_meth_new()` already `XMEMSET`s the whole method to 0, and the only path that reaches the assignment is the one where the argument is already NULL. As a consequence `wolfSSL_BIO_meth_get_callback_ctrl()` (src/bio.c:2303-2311) can only ever return NULL - it reads a field that no code path can ever set non-NULL. The added test at tests/api/test_ossl_bio.c:1714 asserts exactly that, so the getter is a permanently-dead accessor.

The wolfSSL house rule is "no always-success stubs - return a 'not implemented' error instead", which this satisfies in spirit, but the resulting API is confusing: it exposes a getter that is provably always NULL and a setter that only accepts the value the field already has.

Recommendation: Either (a) store the callback so the getter round-trips like OpenSSL and log a `WOLFSSL_MSG` that it is never invoked, or (b) drop the getter/setter pair entirely rather than shipping an accessor that is provably always NULL. If the current reject-non-NULL behaviour is kept deliberately, document it in the function's comment block and in the release notes as a known compatibility limitation.

wolfssl/openssl/bio.h
```diff
@@ -188,6 +188,36 @@
...
+#define BIO_TYPE_START             WOLFSSL_BIO_TYPE_START
+
+/* Like OpenSSL, back app_data with BIO ex_data slot 0 so it does not share
+ * storage with BIO_get_data()/BIO_set_data(); without HAVE_EX_DATA fall back
+ * to sharing the BIO data pointer. */
+#ifdef HAVE_EX_DATA
+#define BIO_get_app_data(bio)       wolfSSL_BIO_get_ex_data((bio), 0)
+#define BIO_set_app_data(bio, data) wolfSSL_BIO_set_ex_data((bio), 0, (data))
+#else
+#define BIO_get_app_data(bio)       wolfSSL_BIO_get_data(bio)
```
dgarske: Recommendation: Drop the `#else` fallback and leave `BIO_{get,set}_app_data` undefined when `HAVE_EX_DATA` is off - an undefined macro is a loud compile error, which is far better than silently returning the method's private pointer as the application's. If a fallback is required, at minimum wrap it so the return type matches (`(wolfSSL_BIO_set_data((bio),(data)), WOLFSSL_SUCCESS)`) and document the aliasing loudly. Add a unit test that sets both `BIO_set_data` and `BIO_set_app_data` on the same BIO and asserts they read back independently.

wolfssl/openssl/hmac.h
```diff
@@ -37,6 +37,12 @@
 
 #include <wolfssl/openssl/compat_types.h>
 #include <wolfssl/openssl/opensslv.h>
+/* OpenSSL's hmac.h pulls in evp.h; mirror it, but only on standalone
+ * include (WOLFSSL_SSL_H unset) to avoid an include cycle during
+ * wolfssl/ssl.h's own parse. */
+#ifndef WOLFSSL_SSL_H
```
dgarske: Recommendation: Use a dedicated in-parse sentinel that is `#undef`'d when ssl.h's include block finishes (see suggestion), or - simpler - have `wolfssl/ssl.h` include `<wolfssl/openssl/evp.h>` immediately before `<wolfssl/openssl/hmac.h>` so the ordering is satisfied for every entry point and hmac.h can include evp.h unconditionally. Either way, correct the misleading comment. Add a compile-only test TU that does `#include <wolfssl/ssl.h>` followed by `#include <wolfssl/openssl/hmac.h>` and references an `EVP_*` symbol.

wolfssl/openssl/bio.h
```diff
@@ -188,6 +188,36 @@
 #define BIO_meth_set_create        wolfSSL_BIO_meth_set_create
 #define BIO_meth_set_destroy       wolfSSL_BIO_meth_set_destroy
 
+#define WOLFSSL_BIO_TYPE_DESCRIPTOR  0x0100
```
dgarske: Recommendation: Move `WOLFSSL_BIO_TYPE_DESCRIPTOR/SOURCE_SINK/START` above the `#ifndef OPENSSL_COEXIST` in bio.h, and `WOLFSSL_TLS_ST_OK`/`WOLFSSL_SSL_ST_OK` above the `#if !defined(OPENSSL_COEXIST)` in openssl/ssl.h - matching the placement the PR already uses for the new `WOLFSSL_SSL_R_*` constants in err.h. That removes the need for the duplicated `128` in src/bio.c and lets the `wc_static_assert` run unconditionally. If the fallback must stay, add `wc_static_assert(WOLFSSL_BIO_TYPE_START == 128);` next to it.

src/bio.c
```diff
@@ -2231,6 +2245,86 @@ int wolfSSL_BIO_meth_set_destroy(WOLFSSL_BIO_METHOD *biom,
...
+    }
+    return NULL;
+}
+
+
+wolfssl_BIO_meth_ctrl_info_cb wolfSSL_BIO_meth_get_callback_ctrl(
+        const WOLFSSL_BIO_METHOD *biom)
+{
+    WOLFSSL_ENTER("wolfSSL_BIO_meth_get_callback_ctrl");
+    if (biom) {
```
dgarske: Recommendation: Change all six new getters to `if (biom != NULL)`. Do not reformat the pre-existing `wolfSSL_BIO_meth_set_*` functions in this PR.

wolfssl/openssl/ssl.h
```diff
@@ -1752,16 +1762,27 @@ typedef WOLFSSL_SRTP_PROTECTION_PROFILE      SRTP_PROTECTION_PROFILE;
...
+#define SSL_R_UNSUPPORTED_SSL_VERSION \
+    WOLFSSL_SSL_R_UNSUPPORTED_SSL_VERSION
+#define SSL_R_WRONG_SSL_VERSION WOLFSSL_SSL_R_WRONG_SSL_VERSION
+#define SSL_R_TLSV1_ALERT_PROTOCOL_VERSION \
+    WOLFSSL_SSL_R_TLSV1_ALERT_PROTOCOL_VERSION
 #define SSL_R_CERTIFICATE_VERIFY_FAILED            VERIFY_CERT_ERROR
 #define SSL_R_CERT_CB_ERROR                        CLIENT_CERT_CB_ERROR
 #define SSL_R_NULL_SSL_METHOD_PASSED               BAD_FUNC_ARG
 #define SSL_R_CCS_RECEIVED_EARLY                   OUT_OF_ORDER_E
+#define ERR_R_BUF_LIB WOLFSSL_ERR_R_BUF_LIB
```
dgarske: Recommendation: Move `#define ERR_R_BUF_LIB` from wolfssl/openssl/ssl.h into wolfssl/openssl/err.h's `#ifndef OPENSSL_COEXIST` block, next to the other `ERR_R_*` aliases.

dgarske (changes requested):

src/bio.c
kojo1: Make BIO_get_new_index thread-safe with atomics, add threaded test

src/bio.c
kojo1: Store callback_ctrl in BIO_meth_set_callback_ctrl like OpenSSL

wolfssl/openssl/bio.h
kojo1: Drop BIO_{get,set}_app_data fallback aliasing BIO_{get,set}_data

wolfssl/openssl/hmac.h
kojo1: Use a parse-in-progress sentinel for hmac.h's evp.h include

wolfssl/openssl/bio.h
kojo1: Move WOLFSSL_-prefixed BIO type and TLS state constants out of coexist guards

src/bio.c
kojo1: Use explicit NULL comparison in BIO_meth_get_* getters

wolfssl/openssl/ssl.h
kojo1: Move ERR_R_BUF_LIB alias to openssl/err.h

wolfssl/openssl/ssl.h
kojo1: Translate VERSION_ERROR to SSL_R_WRONG_VERSION_NUMBER in ERR_GET_REASON.
Added the requested regression test: a TLS 1.3-only client against a TLS 1.2-only server over memio, asserting ERR_GET_REASON() yields SSL_R_WRONG_VERSION_NUMBER.

Restoring VERSION_ERROR for the three pre-existing macros is not an option here: libpq's open_client_SSL() (fe-secure-openssl.c) lists all nine SSL_R_* version reasons as case labels in a single switch (ERR_GET_REASON(ecode)), so any two macros sharing a value produce a duplicate-case compile error. Unique OpenSSL numeric values are required for the drop-in goal of this PR.

dgarske (approved):

PR #10818, opened 2026-06-30
DH: honor explicit prime in wc_DhCheckPrivKey_ex (Fenrir F5317)

dgarske:
`wc_DhCheckPrivKey_ex` loaded `q` from the supplied `prime` (or `key->q`), then re-copied `key->q` over it before the `x <= q-1` check. With an explicit prime and a `p,g`-only key (empty `key->q`), the bound was discarded and a valid private key was rejected; when `key->q` differed from the prime, the prime was silently ignored. Removed the redundant copy; the `prime == NULL` path is unchanged. Added a regression test that fails on the prior code.

### DIFF
--- a/wolfcrypt/src/dh.c
+++ b/wolfcrypt/src/dh.c
@@ -1808,12 +1808,9 @@ int wc_DhCheckPrivKey_ex(DhKey* key, const byte* priv, word32 privSz,
     if (ret == 0) {
         if (mp_iszero(q) == MP_NO) {
             /* priv (x) shouldn't be greater than q - 1 */
-            if (mp_copy(&key->q, q) != MP_OKAY)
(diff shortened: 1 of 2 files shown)


dgarske:
Jenkins retest this please

kaleb-himes (approved):
Thank you @dgarske I see no issues with this change. 

Fix is good. It removes a redundant mp_copy that clobbered the caller-supplied prime with an empty key->q, which was wrongly rejecting valid keys.

No side-channel risk, the copy served no constant-time/cache purpose and only touched public data, with the secret-dependent branches unchanged.

No new null-deref either: it drops the key->q access rather than adding one, leaving only the already-initialized q and x.

Good change.

PR #10821, opened 2026-06-30
Add true zero-allocation X.509 certificate verification under WOLFSSL_NO_MALLOC

aidangarske:
Makes the X.509 verify path (wc_ParseCert) truly be no malloc, so embedded/no-heap builds can validate certificate chains without dynamic memory.

Previously a strict no-heap build failed to parse any real certificate. The public-key copy (StoreKey/StoreEccKey) and the SubjectAltName list (AltNameNew/SetDNSEntry) needed the heap and returned `MEMORY_E`. This references the public key and alt-name strings in place in the source DER (the same contract wolfSSL already uses for subjectCN), with SAN entries held in a fixed per-cert pool  (WC_ASN_MAX_ALTNAMES, default 8).

Gated by a new internal `WC_ASN_NO_HEAP` `(WOLFSSL_NO_MALLOC && NO_WOLFSSL_MEMORY && !XMALLOC_USER && !WOLFSSL_STATIC_MEMORY)`: only builds with genuinely no allocator take the in-place path; builds with a real allocator (callbacks, XMALLOC_USER, static memory) keep the existing heap paths. Default malloc builds are byte-for-byte unchanged
 
Under no-heap, IP/RID SANs (which need a heap-synthesized string) and CertManager CA storage are fail-closed (documented limitations). Adds a heap-free parse test (cert_no_malloc_test).

### DIFF
--- a/wolfcrypt/src/asn.c
+++ b/wolfcrypt/src/asn.c
@@ -156,6 +156,15 @@ ASN Options:
  * WOLFSSL_ALLOW_CRIT_AKID:  Allow critical Auth Key Identifier
  * WOLFSSL_ALLOW_CRIT_SKID:  Allow critical Subject Key Identifier
  * WC_ASN_UNKNOWN_EXT_CB:    Callback for unknown extensions
+ * WC_ASN_NO_HEAP:           Zero-allocation cert parse: reference key/alt-name
+    data in the source DER instead of heap copies, so the source buffer must
+    outlive the DecodedCert. Auto-defined when WOLFSSL_NO_MALLOC and
+    NO_WOLFSSL_MEMORY are set without XMALLOC_USER or WOLFSSL_STATIC_MEMORY.
+    Limitation: IP and registeredID SAN entries need a parsed string form that
+    has no in-place source, so such certs are rejected with ASN_PARSE_E. SAN
+    DNS_entry.name is NOT NUL-terminated in this mode; only .len is authoritative.
+ * WC_ASN_MAX_ALTNAMES:      No-heap SAN pool slot count (default 8); excess
+    subject alternative names are rejected.
  *
  * ASN.1 Parsing:
  * WOLFSSL_ASN_ALL:          Enable all ASN.1 features
@@ -12579,7 +12588,14 @@ void FreeAltNames(DNS_entry* altNames, void* heap)
             altNames->ridStringStored = 0;
         }
     #endif
-        XFREE(altNames,       heap, DYNAMIC_TYPE_ALTNAME);
+#ifdef WC_ASN_NO_HEAP
+        /* Only free heap nodes; no-heap pool nodes aren't owned. */
+        if (altNames->entryStored) {
+            XFREE(altNames, heap, DYNAMIC_TYPE_ALTNAME);
+        }
+#else
+        XFREE(altNames, heap, DYNAMIC_TYPE_ALTNAME);
+#endif
         altNames = tmp;
     }
 }
@@ -12591,6 +12607,9 @@ DNS_entry* AltNameNew(void* heap)
     ret = (DNS_entry*)XMALLOC(sizeof(DNS_entry), heap, DYNAMIC_TYPE_ALTNAME);
     if (ret != NULL) {
         XMEMSET(ret, 0, sizeof(DNS_entry));
+#ifdef WC_ASN_NO_HEAP
+        ret->entryStored = 1;   /* heap-allocated node; FreeAltNames frees it */
+#endif
     }
     (void)heap;
     return ret;
@@ -12739,7 +12758,9 @@ static int StoreKey(DecodedCert* cert, const byte* source, word32* srcIdx,
 {
     int ret;
     int length;
+#ifndef WC_ASN_NO_HEAP
     byte* publicKey;
+#endif
 
     ret = CheckBitString(source, srcIdx, &length, maxIdx, 1, NULL);
     if (ret == 0) {
@@ -12749,6 +12770,17 @@ static int StoreKey(DecodedCert* cert, const byte* source, word32* srcIdx,
     }
     if (ret == 0) {
     #endif
+#ifdef WC_ASN_NO_HEAP
+        /* No heap: reference the key in place; source must outlive the cert. */
+        cert->publicKey = (byte*)&source[*srcIdx];
+        cert->pubKeyStored = 0;
+        cert->pubKeySize   = (word32)length;
+    #ifdef HAVE_OCSP_RESPONDER
+        cert->publicKeyForHash = cert->publicKey;
+        cert->pubKeyForHashSize = cert->pubKeySize;
+    #endif
+        *srcIdx += (word32)length;
+#else
         publicKey = (byte*)XMALLOC((size_t)length, cert->heap,
                                    DYNAMIC_TYPE_PUBLIC_KEY);
         if (publicKey == NULL) {
@@ -12767,6 +12799,7 @@ static int StoreKey(DecodedCert* cert, const byte* source, word32* srcIdx,
 
             *srcIdx += (word32)length;
         }
+#endif
     }
 
     return ret;
@@ -13407,7 +13440,9 @@ static int StoreEccKey(DecodedCert* cert, const byte* source, word32* srcIdx,
 {
     int ret = 0;
     DECL_ASNGETDATA(dataASN, eccCertKeyASN_Length);
+#ifndef WC_ASN_NO_HEAP
     byte* publicKey;
+#endif
 
     /* Validate parameters. */
     if (pubKey == NULL) {
@@ -13477,6 +13512,12 @@ static int StoreEccKey(DecodedCert* cert, const byte* source, word32* srcIdx,
     #endif
         /* Store public key data length. */
         cert->pubKeySize = pubKeyLen;
+#ifdef WC_ASN_NO_HEAP
+        /* No heap: point at pubKey in the input DER, so that buffer must stay
+         * valid for the DecodedCert's lifetime. */
+        cert->publicKey = (byte*)pubKey;
+        cert->pubKeyStored = 0;
+#else
         /* Must allocated space for key.
          * Don't memcpy into constant pointer so use temp. */
         publicKey = (byte*)XMALLOC(cert->pubKeySize, cert->heap,
@@ -13491,6 +13532,7 @@ static int StoreEccKey(DecodedCert* cert, const byte* source, word32* srcIdx,
             /* Indicate publicKey needs to be freed. */
             cert->pubKeyStored = 1;
         }
+#endif
     }
     FREE_ASNGETDATA(dataASN, cert->heap);
 
@@ -14357,7 +14399,7 @@ static const byte rdnChoice[] = {
 };
 #endif
 
-#ifdef WOLFSSL_IP_ALT_NAME
+#if defined(WOLFSSL_IP_ALT_NAME) && !defined(WC_ASN_NO_HEAP)
 /* used to set the human readable string for the IP address with a ASN_IP_TYPE
  * DNS entry
  * return 0 on success
@@ -14426,9 +14468,9 @@ static int GenerateDNSEntryIPString(DNS_entry* entry, void* heap)
 
     return ret;
 }
-#endif /* WOLFSSL_IP_ALT_NAME */
+#endif /* WOLFSSL_IP_ALT_NAME && !WC_ASN_NO_HEAP */
 
-#ifdef WOLFSSL_RID_ALT_NAME
+#if defined(WOLFSSL_RID_ALT_NAME) && !defined(WC_ASN_NO_HEAP)
 /* used to set the human readable string for the registeredID with an
  * ASN_RID_TYPE DNS entry
  * return 0 on success
@@ -14525,7 +14567,7 @@ static int GenerateDNSEntryRIDString(DNS_entry* entry, void* heap)
 
     return ret;
 }
-#endif /* WOLFSSL_RID_ALT_NAME */
+#endif /* WOLFSSL_RID_ALT_NAME && !WC_ASN_NO_HEAP */
 
 #ifdef WOLFSSL_ASN_TEMPLATE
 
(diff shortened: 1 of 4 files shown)


wolfcrypt/src/asn.c
```diff
@@ -23949,7 +24018,14 @@ int FillSigner(Signer* signer, DecodedCert* cert, int type, DerBuffer *der)
...
         }
+#ifdef WC_ASN_NO_HEAP
+        else if ((cert->publicKey != NULL) && (cert->pubKeySize > 0)) {
+            /* Borrowed key (points into source); fail rather than retain it. */
+            ret = MEMORY_E;
+        }
+#endif
 
+        if (ret == 0) {
         if (cert->subjectCNStored) {
```
SparkiDev, line 24476: Doesn't look like indenting lines up.

wolfcrypt/src/asn.c
```diff
@@ -23949,7 +24018,14 @@ int FillSigner(Signer* signer, DecodedCert* cert, int type, DerBuffer *der)
             signer->publicKey      = cert->publicKey;
             signer->pubKeySize     = cert->pubKeySize;
         }
+#ifdef WC_ASN_NO_HEAP
+        else if ((cert->publicKey != NULL) && (cert->pubKeySize > 0)) {
```
SparkiDev: Not sure how likely this case is, but I only see it in OCSP stapling v2 and AddCA which seems pretty likely if you are doing TLS.
Is the expected use case non-TLS?

SparkiDev (changes requested):

wolfcrypt/src/asn.c
aidangarske: I removed this you are right FillSigner is unreachable under WC_ASN_NO_HEAP so I removed it

wolfcrypt/src/asn.c
```diff
@@ -12471,7 +12471,14 @@ void FreeAltNames(DNS_entry* altNames, void* heap)
             altNames->ridStringStored = 0;
         }
     #endif
+#ifdef WC_ASN_NO_HEAP
```
dgarske, line 12591: Please document `WC_ASN_NO_HEAP` at top of asn.c.

wolfssl/wolfcrypt/asn.h
```diff
@@ -1449,6 +1449,17 @@ enum KeyIdType {
     #define WOLFSSL_IP6_ADDR_LEN 16
 #endif /* OPENSSL_ALL || WOLFSSL_IP_ALT_NAME */
 
+/* No allocator: reference key/alt-name data in the source DER, not copies. */
+#if defined(WOLFSSL_NO_MALLOC) && defined(NO_WOLFSSL_MEMORY) && \
+    !defined(XMALLOC_USER) && !defined(WOLFSSL_STATIC_MEMORY)
+    #define WC_ASN_NO_HEAP
+#endif
+
+#ifndef WC_ASN_MAX_ALTNAMES
```
dgarske: Wrap this with `WC_ASN_NO_HEAP` so its clear this only applies to the NO malloc case.

wolfssl/wolfcrypt/asn.h
```diff
@@ -1470,6 +1481,9 @@ struct DNS_entry {
 #ifdef WOLFSSL_FPKI
     int        oidSum; /* provide oid sum for verification */
 #endif
+#ifdef WC_ASN_NO_HEAP
+    int        entryStored; /* 1 = heap node to free; 0 = no-heap pool node */
```
dgarske: Consider using WC_BITFIELD or :1 with unsigned to save strut space.

wolfcrypt/src/asn.c
```diff
@@ -14466,13 +14498,55 @@ static int AddDNSEntryToList(DNS_entry** lst, DNS_entry* entry)
...
     int ret = 0;
+#ifndef WC_ASN_NO_HEAP
     char *dnsEntry_name = NULL;
+#endif
 
+#ifdef WC_ASN_NO_HEAP
+    /* No heap: borrow a pool slot; name points into the source DER. */
+    (void)heap;
+#ifdef WOLFSSL_IP_ALT_NAME
+    /* No-heap path can't synthesise an ipString; reject rather than skip. */
```
dgarske: Consider `synthesise` -> `parse`. Please document this limitation (also RID) at top of asn.c along with the macro

wolfcrypt/src/asn.c
```diff
@@ -13369,6 +13395,11 @@ static int StoreEccKey(DecodedCert* cert, const byte* source, word32* srcIdx,
     #endif
         /* Store public key data length. */
         cert->pubKeySize = pubKeyLen;
+#ifdef WC_ASN_NO_HEAP
+        /* No heap: reference the key in place; source must outlive the cert. */
```
dgarske: What does "source must outlive the cert"? Is this comment referring to the source pubKey pointer provided?

dgarske (changes requested):

wolfcrypt/src/asn.c
```diff
@@ -14466,13 +14498,55 @@ static int AddDNSEntryToList(DNS_entry** lst, DNS_entry* entry)
...
     DNS_entry* dnsEntry;
     int ret = 0;
+#ifndef WC_ASN_NO_HEAP
     char *dnsEntry_name = NULL;
+#endif
 
+#ifdef WC_ASN_NO_HEAP
+    /* No heap: borrow a pool slot; name points into the source DER. */
+    (void)heap;
+#ifdef WOLFSSL_IP_ALT_NAME
```
dgarske, line 14639: **🔴 [High] Unused static function build break under WC_ASN_NO_HEAP + WOLFSSL_IP_ALT_NAME/WOLFSSL_RID_ALT_NAME** · _bug_

`GenerateDNSEntryIPString` (guarded only by `#ifdef WOLFSSL_IP_ALT_NAME`, line 14284) and `GenerateDNSEntryRIDString` (`#ifdef WOLFSSL_RID_ALT_NAME`, line 14355) are `static`, and their ONLY call sites are lines 14577 and 14583, both of which the PR moved into the new `#else` (non-`WC_ASN_NO_HEAP`) branch of `SetDNSEntry` (the `#ifdef WC_ASN_NO_HEAP ... #else ... #endif` closes at line 14594). When `WC_ASN_NO_HEAP` is defined together with `WOLFSSL_IP_ALT_NAME` or `WOLFSSL_RID_ALT_NAME`, those callers are compiled out but the function definitions remain, producing `defined but not used` static functions. Under wolfSSL's common `-Werror` build jobs this `-Wunused-function` warning is a hard compile failure. This is not a hypothetical combination: the PR's own `WC_ASN_NO_HEAP` branch contains explicit `#ifdef WOLFSSL_IP_ALT_NAME`/`#ifdef WOLFSSL_RID_ALT_NAME` reject code (asn.c:14521-14531), so the combo is intended to be buildable.

**Fix:** Gate the two helper definitions with `&& !defined(WC_ASN_NO_HEAP)` (e.g. `#if defined(WOLFSSL_IP_ALT_NAME) && !defined(WC_ASN_NO_HEAP)` and likewise for GenerateDNSEntryRIDString) so the no-heap + IP/RID-alt-name configuration compiles cleanly under -Werror. Add a CI build combining WC_ASN_NO_HEAP with WOLFSSL_IP_ALT_NAME to catch this.

wolfcrypt/test/test.c
```diff
@@ -26012,6 +26012,40 @@ static wc_test_ret_t cert_bad_asn1_test(void)
...
+/* Heap/file-free parse from a const DER buffer; checks in-place refs under
+ * WC_ASN_NO_HEAP. */
+static wc_test_ret_t cert_no_malloc_test(void)
+{
+    DecodedCert cert;
+    wc_test_ret_t ret;
+
+    WOLFSSL_ENTER("cert_no_malloc_test");
+
+    InitDecodedCert(&cert, server_cert_der_2048, sizeof_server_cert_der_2048,
```
dgarske: **🟠 [Medium] cert_no_malloc_test uses a cert whose IP SAN the no-heap path fail-closes** · _test_

The positive parse test uses `server_cert_der_2048`, which (per certs/server-cert.pem) carries `X509v3 Subject Alternative Name: DNS:example.com, IP Address:127.0.0.1`. Under `WC_ASN_NO_HEAP` combined with `WOLFSSL_IP_ALT_NAME`, the new `SetDNSEntry` rejects `ASN_IP_TYPE` with `ASN_PARSE_E` (asn.c:14521-14525), so `ParseCert` fails and `cert_no_malloc_test` returns that error, failing the whole wolfCrypt suite. The test therefore silently depends on `WOLFSSL_IP_ALT_NAME` being undefined, even though the code path it is meant to validate explicitly supports that macro. When -Werror is off this is the runtime symptom of the same combo that otherwise breaks the build.

**Fix:** Use a DER cert buffer that contains no iPAddress/registeredID SAN for the positive no-heap parse (or explicitly gate the WC_ASN_NO_HEAP assertions when WOLFSSL_IP_ALT_NAME/WOLFSSL_RID_ALT_NAME are defined), and add a separate negative test asserting ASN_PARSE_E for an IP-SAN cert under those macros.

wolfcrypt/test/test.c
```diff
@@ -26012,6 +26012,40 @@ static wc_test_ret_t cert_bad_asn1_test(void)
     return ret;
 }
 
+#if defined(USE_CERT_BUFFERS_2048) && !defined(NO_RSA)
+/* Heap/file-free parse from a const DER buffer; checks in-place refs under
+ * WC_ASN_NO_HEAP. */
+static wc_test_ret_t cert_no_malloc_test(void)
```
dgarske: **🟠 [Medium] New no-heap key/SAN in-place paths are under-tested** · _test_

`cert_no_malloc_test` exercises only the pre-existing RSA in-place public-key path plus (at most) a single DNS SAN. The genuinely new code is not covered: `StoreKey`/`StoreEccKey` in-place references for Ed25519/Ed448/ECC keys (`cert->pubKeyStored == 0`), SAN pool nodes whose `name` pointer must land inside `[source, source+maxIdx)`, the pool-exhaustion `MEMORY_E` path, and the IP/RID fail-closed path. The RSA-only positive test cannot detect regressions in the Store*Key ECC/Ed branches or in the SAN pool name references.

**Fix:** Broaden the no-heap test: walk the full cert.altNames chain and confirm every `name` pointer satisfies `source <= name < source+maxIdx` and `entryStored == 0`; add an ECC cert buffer to hit StoreEccKey no-heap; add a synthetic >WC_ASN_MAX_ALTNAMES case expecting MEMORY_E; cover the IP/RID fail-closed path.

wolfcrypt/src/asn.c
```diff
@@ -14466,13 +14498,55 @@ static int AddDNSEntryToList(DNS_entry** lst, DNS_entry* entry)
...
+#ifdef WOLFSSL_RID_ALT_NAME
+    if (type == ASN_RID_TYPE) {
+        return ASN_PARSE_E;
+    }
+#endif
+    if ((pool == NULL) || (*poolUsed >= WC_ASN_MAX_ALTNAMES)) {
+        ret = MEMORY_E;
+        dnsEntry = NULL;
+    }
+    else {
```
dgarske, line 14656: **🔵 [Low] No-heap SAN name is not NUL-terminated (contract divergence from heap path)** · _Security_

The heap path NUL-terminates `dnsEntry->name` (`dnsEntry_name[strLen] = '\0'`, line 14572), but the new `WC_ASN_NO_HEAP` path sets `dnsEntry->name = (char*)str` pointing directly into the source DER with NO terminator (the following byte is the next ASN.1 element), relying solely on `len`. Length-based consumers are safe: the core verification path (CheckForAltNames/MatchDomainName in src/internal.c) and the copy paths in src/x509.c use `altName->len`/`dns->len`. However, several OpenSSL-compat consumers treat `entry->name` as a NUL-terminated C string, e.g. `X509PrintName` (src/x509.c:6956 `XSNPRINTF(scratch, MAX_WIDTH, "DNS:%s", entry->name)`, and the RFC822/URI cases at 6974/6981). Under no-heap those would read past the SAN value into adjacent DER bytes until a stray `0x00`, potentially past the end of the cert buffer (out-of-bounds read). Reachability is low because those consumers require OPENSSL_EXTRA/OPENSSL_ALL, which in practice cannot be built under a genuine no-allocator configuration. Flagged because the heap path's NUL-termination is an implicit contract the new path silently breaks.

**Fix:** Document explicitly (in asn.h near WC_ASN_NO_HEAP and on struct DNS_entry) that under WC_ASN_NO_HEAP `name` is NOT NUL-terminated and only `len` is authoritative, mirroring the subjectCN contract. Audit `entry->name` string consumers (e.g. src/x509.c X509PrintName) to use `%.*s` with `entry->len` instead of `%s`, or compile-guard those print helpers out when WC_ASN_NO_HEAP is defined.

wolfssl/wolfcrypt/asn.h
```diff
@@ -1789,6 +1803,11 @@ struct DecodedCert {
 #endif
     int     version;                 /* cert version, 1 or 3             */
     DNS_entry* altNames;             /* alt names list of dns entries    */
+#ifdef WC_ASN_NO_HEAP
```
dgarske: **🔵 [Low] altNamePool embedded mid-struct enlarges stack-allocated DecodedCert/DecodedAcert under no-heap** · _Blast Radius_

Under WC_ASN_NO_HEAP an inline `DNS_entry altNamePool[WC_ASN_MAX_ALTNAMES]` (default 8) plus `altNamePoolUsed` is embedded directly in both DecodedCert (asn.h:1806-1811) and DecodedAcert (asn.h:3230-3234). These structs are frequently stack-allocated on exactly the embedded/no-heap targets this feature serves, so this adds roughly 8 * sizeof(DNS_entry) bytes (order of a few hundred bytes) to stack usage per parse, which can matter on stack-constrained MCUs. It is the intended zero-allocation trade-off and is behind #ifdef (heap builds unaffected, ABI-safe since the block is fully compiled out in non-no-heap builds). Note the new members are inserted mid-struct (after `altNames`) rather than appended at the end, contrary to the project's new-members-at-the-end convention; this is only ABI-safe because the whole block is conditional.

**Fix:** Prefer appending the new members at the end of the struct. Document the added stack cost and that WC_ASN_MAX_ALTNAMES directly scales it, so integrators can size stacks / reduce the pool for constrained targets.

wolfcrypt/src/asn.c
```diff
@@ -18988,7 +19063,7 @@ static int DecodeOtherHelper(ASNGetData* dataASN, DecodedCert* cert, int oid)
     }
 
     if (ret == 0) {
-        ret = SetDNSEntry(cert->heap, buf, (int)bufLen, ASN_OTHER_TYPE, &entry);
+        ret = SetDNSEntry(cert->heap, WC_DNS_POOL(cert), buf, (int)bufLen, ASN_OTHER_TYPE, &entry);
```
dgarske: **🔵 [Low] New SetDNSEntry call sites exceed the 80-column style limit** · _style_

Several updated `SetDNSEntry(cert->heap, WC_DNS_POOL(cert), ...)` call sites now exceed the wolfSSL 80-column convention (e.g. asn.c:19066 and 19128), unlike the neighboring wrapped calls. Minor formatting inconsistency introduced by threading the extra pool arguments through the macro.

**Fix:** Re-wrap the over-length lines to stay within 80 columns, matching the surrounding style.

dgarske (commented):
## Skoll Multi-Scan Review

### Posted findings
- [High] [review] **Unused static function build break under WC_ASN_NO_HEAP + WOLFSSL_IP_ALT_NAME/WOLFSSL_RID_ALT_NAME** — `wolfcrypt/src/asn.c:14289,14360,14577,14583`
- [Medium] [review] **cert_no_malloc_test uses a cert whose IP SAN the no-heap path fail-closes** — `wolfcrypt/test/test.c:26025-26027`
- [Medium] [review] **New no-heap key/SAN in-place paths are under-tested** — `wolfcrypt/test/test.c:26018-26046`
- [Low] [review+review-security] **No-heap SAN name is not NUL-terminated (contract divergence from heap path)** — `wolfcrypt/src/asn.c:14536-14542`
- [Low] [review+review-security] **altNamePool embedded mid-struct enlarges stack-allocated DecodedCert/DecodedAcert under no-heap** — `wolfssl/wolfcrypt/asn.h:1806-1811`
- [Low] [review] **New SetDNSEntry call sites exceed the 80-column style limit** — `wolfcrypt/src/asn.c:19066,19128,19274,19306,19322`

### Skipped findings
- [Medium] ~~Shared 8-slot SAN pool exhaustion and IP/RID fail-closed abort the entire certificate parse under no-heap~~

wolfcrypt/test/test.c
dgarske: Please add CI to exercise `cert_no_malloc_test`.

1. cert_no_malloc_test is unreachable in a real no-malloc build — the exact config it validates. It sits at the end of cert_test(), but cert_test's preamble does XMALLOC(FOURK_BUF,...) at test.c:26060, which returns NULL under WOLFSSL_NO_MALLOC, so cert_test returns early and never reaches it. It's also only called when WOLFSSL_TEST_CERT && !NO_FILESYSTEM, yet the test is deliberately heap/file-free. Recommend: hoist cert_no_malloc_test to run before the heap/filesystem preamble, or invoke it from the main runner independently. That's why I had to exercise the path with a standalone harness instead.
2. CI gap — .github/workflows/no-malloc.yml sets -DWOLFSSL_NO_MALLOC but never -DNO_WOLFSSL_MEMORY, so WC_ASN_NO_HEAP is never compiled in CI. The new path needs both. Recommend adding a config with both macros.

dgarske (changes requested):

dgarske (dismissed):

dgarske:
Jenkins retest this please

dgarske (approved):

SparkiDev (approved):

PR #10822, opened 2026-06-30
Fix session ticket key rotation when hint exceeds key lifetime

### LINKED ISSUE #10796: [Bug]: DefTicketEncCb: BAD_STATE_E on second handshake when ticket hint > WOLFSSL_TICKET_KEY_LIFETIME / 2

mattia-moffa:
# Description

Disallows a `ticketHint > WOLFSSL_TICKET_KEY_LIFETIME / 2`.

The default ticket encryption callback rotates two keys and can only honor a hint below half the ticket key lifetime; a larger hint previously left no key available for encryption, failing every handshake after the first on a reused server CTX. Now these values are rejected.

Fixes #10796

# Testing

```
./configure --enable-dtls13 --enable-session-ticket && make && make check
```

# Checklist

 - [x] added tests

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -40181,6 +40181,19 @@ static int AddPSKtoPreMasterSecret(WOLFSSL* ssl)
     }
 #endif
 
+    /* Returns 1 when the default ticket encryption callback is in use with a
+     * hint of at least half the key lifetime. */
+    int DefTicketHintTooLarge(WOLFSSL* ssl)
+    {
+    #if !defined(WOLFSSL_NO_DEF_TICKET_ENC_CB) && !defined(NO_WOLFSSL_SERVER)
+        return ssl->ctx->ticketEncCb == DefTicketEncCb &&
+               2 * ssl->ctx->ticketHint >= WOLFSSL_TICKET_KEY_LIFETIME;
+    #else
+        (void)ssl;
+        return 0;
+    #endif
+    }
+
     /* create a new session ticket, 0 on success
      * Do any kind of setup in SetupTicket */
     int CreateTicket(WOLFSSL* ssl)
@@ -41197,12 +41210,19 @@ cleanup:
             /* This will be set when SendHandshakeMsg returns WANT_WRITE. Create
              * a new ticket only once. */
             !ssl->options.buildingMsg) {
-            ret = SetupTicket(ssl);
-            if (ret != 0)
-                return ret;
-            ret = CreateTicket(ssl);
-            if (ret != 0)
-                return ret;
+            if (DefTicketHintTooLarge(ssl)) {
+                WOLFSSL_MSG("Ticket hint exceeds half the ticket key lifetime; "
+                            "skipping ticket");
+                ssl->session->ticketLen = 0;
+            }
+            else {
+                ret = SetupTicket(ssl);
+                if (ret != 0)
+                    return ret;
+                ret = CreateTicket(ssl);
+                if (ret != 0)
+                    return ret;
+            }
         }
 
         length += ssl->session->ticketLen;
@@ -41742,13 +41762,6 @@ static int DefTicketEncCb(WOLFSSL* ssl, byte key_name[WOLFSSL_TICKET_NAME_SZ],
     /* For decryption, check minimum internal ticket size */
     if ((!enc) && (inLen < (int)WOLFSSL_INTERNAL_TICKET_LEN)) {
         return BUFFER_E;
-    }
-
-    /* The two-key rotation can only honor a hint below half the key lifetime.
-     * Refuse to issue a ticket whose advertised lifetime it cannot cover. */
-    if (enc && 2 * ctx->ticketHint >= WOLFSSL_TICKET_KEY_LIFETIME) {
-        WOLFSSL_MSG("Ticket hint exceeds half the ticket key lifetime");
-        return WOLFSSL_TICKET_RET_FATAL;
     }
 
     /* Check we have setup the RNG, name and primary key. */
--- a/src/internal.c
+++ b/src/internal.c
@@ -41744,6 +41744,13 @@ static int DefTicketEncCb(WOLFSSL* ssl, byte key_name[WOLFSSL_TICKET_NAME_SZ],
         return BUFFER_E;
     }
 
+    /* The two-key rotation can only honor a hint below half the key lifetime.
+     * Refuse to issue a ticket whose advertised lifetime it cannot cover. */
+    if (enc && 2 * ctx->ticketHint >= WOLFSSL_TICKET_KEY_LIFETIME) {
+        WOLFSSL_MSG("Ticket hint exceeds half the ticket key lifetime");
+        return WOLFSSL_TICKET_RET_FATAL;
+    }
+
     /* Check we have setup the RNG, name and primary key. */
     if (keyCtx->expirary[0] == 0) {
 #ifndef SINGLE_THREADED
--- a/src/ssl_api_ext.c
+++ b/src/ssl_api_ext.c
@@ -1055,8 +1055,6 @@ int wolfSSL_CTX_set_TicketEncCb(WOLFSSL_CTX* ctx, SessionTicketEncCb cb)
  *
  * @param [in] ctx   SSL/TLS context object.
  * @param [in] hint  Lifetime hint in seconds. No more than 604800 (7 days).
- *                   With the default ticket encryption callback, must also be
- *                   less than WOLFSSL_TICKET_KEY_LIFETIME / 2.
  * @return  WOLFSSL_SUCCESS on success.
  * @return  BAD_FUNC_ARG when ctx is NULL or hint is out of range.
  */
@@ -1069,15 +1067,6 @@ int wolfSSL_CTX_set_TicketHint(WOLFSSL_CTX* ctx, int hint)
      * 604800 seconds (7 days). */
     if (hint < 0 || hint > 604800)
         return BAD_FUNC_ARG;
-
-#ifdef WOLFSSL_TICKET_KEY_LIFETIME
-    /* The default ticket encryption callback rotates two keys, each living
-     * WOLFSSL_TICKET_KEY_LIFETIME seconds. Its staggered rotation can only
-     * honor a hint below half that; a larger hint would leave no key available
-     * for encryption. */
-    if (hint >= WOLFSSL_TICKET_KEY_LIFETIME / 2)
-        return BAD_FUNC_ARG;
-#endif
 
     ctx->ticketHint = hint;
 
--- a/src/ssl_api_ext.c
+++ b/src/ssl_api_ext.c
@@ -1055,6 +1055,8 @@ int wolfSSL_CTX_set_TicketEncCb(WOLFSSL_CTX* ctx, SessionTicketEncCb cb)
  *
  * @param [in] ctx   SSL/TLS context object.
  * @param [in] hint  Lifetime hint in seconds. No more than 604800 (7 days).
+ *                   With the default ticket encryption callback, must also be
+ *                   less than WOLFSSL_TICKET_KEY_LIFETIME / 2.
  * @return  WOLFSSL_SUCCESS on success.
  * @return  BAD_FUNC_ARG when ctx is NULL or hint is out of range.
  */
@@ -1067,6 +1069,15 @@ int wolfSSL_CTX_set_TicketHint(WOLFSSL_CTX* ctx, int hint)
      * 604800 seconds (7 days). */
     if (hint < 0 || hint > 604800)
         return BAD_FUNC_ARG;
+
+#ifdef WOLFSSL_TICKET_KEY_LIFETIME
+    /* The default ticket encryption callback rotates two keys, each living
+     * WOLFSSL_TICKET_KEY_LIFETIME seconds. Its staggered rotation can only
+     * honor a hint below half that; a larger hint would leave no key available
+     * for encryption. */
+    if (hint >= WOLFSSL_TICKET_KEY_LIFETIME / 2)
+        return BAD_FUNC_ARG;
+#endif
 
     ctx->ticketHint = hint;
 
--- a/tests/api/test_ssl_ext.c
+++ b/tests/api/test_ssl_ext.c
@@ -188,7 +188,18 @@ int test_wolfSSL_CTX_set_TicketHint_ext(void)
     ExpectIntEQ(wolfSSL_CTX_set_TicketHint(ctx, 604801),
         WC_NO_ERR_TRACE(BAD_FUNC_ARG));
     ExpectIntEQ(wolfSSL_CTX_set_TicketHint(ctx, 0), WOLFSSL_SUCCESS);
+#ifdef WOLFSSL_TICKET_KEY_LIFETIME
+    /* The default ticket encryption callback can only honor a hint below half
+     * the ticket key lifetime; larger values are rejected. */
+    ExpectIntEQ(wolfSSL_CTX_set_TicketHint(ctx, WOLFSSL_TICKET_KEY_LIFETIME / 2),
+        WC_NO_ERR_TRACE(BAD_FUNC_ARG));
+    ExpectIntEQ(wolfSSL_CTX_set_TicketHint(ctx,
+        WOLFSSL_TICKET_KEY_LIFETIME / 2 - 1), WOLFSSL_SUCCESS);
+    ExpectIntEQ(wolfSSL_CTX_set_TicketHint(ctx, 604800),
+        WC_NO_ERR_TRACE(BAD_FUNC_ARG));
+#else
     ExpectIntEQ(wolfSSL_CTX_set_TicketHint(ctx, 604800), WOLFSSL_SUCCESS);
+#endif
 
     wolfSSL_CTX_free(ctx);
 #endif
--- a/tests/api/test_ssl_ext.c
+++ b/tests/api/test_ssl_ext.c
@@ -188,20 +188,80 @@ int test_wolfSSL_CTX_set_TicketHint_ext(void)
     ExpectIntEQ(wolfSSL_CTX_set_TicketHint(ctx, 604801),
         WC_NO_ERR_TRACE(BAD_FUNC_ARG));
     ExpectIntEQ(wolfSSL_CTX_set_TicketHint(ctx, 0), WOLFSSL_SUCCESS);
-#ifdef WOLFSSL_TICKET_KEY_LIFETIME
-    /* The default ticket encryption callback can only honor a hint below half
-     * the ticket key lifetime; larger values are rejected. */
-    ExpectIntEQ(wolfSSL_CTX_set_TicketHint(ctx, WOLFSSL_TICKET_KEY_LIFETIME / 2),
-        WC_NO_ERR_TRACE(BAD_FUNC_ARG));
-    ExpectIntEQ(wolfSSL_CTX_set_TicketHint(ctx,
-        WOLFSSL_TICKET_KEY_LIFETIME / 2 - 1), WOLFSSL_SUCCESS);
-    ExpectIntEQ(wolfSSL_CTX_set_TicketHint(ctx, 604800),
-        WC_NO_ERR_TRACE(BAD_FUNC_ARG));
-#else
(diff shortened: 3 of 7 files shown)


tests/api/test_ssl_ext.c
```diff
@@ -195,6 +195,81 @@ int test_wolfSSL_CTX_set_TicketHint_ext(void)
...
+        for (round = 0; round < 4; round++) {
+            /* Set up two new connections using the same contexts */
+            if (round > 0) {
+                XMEMSET(&test_ctx, 0, sizeof(test_ctx));
+                ExpectNotNull(ssl_c = wolfSSL_new(ctx_c));
+                ExpectNotNull(ssl_s = wolfSSL_new(ctx_s));
+                wolfSSL_SetIOReadCtx(ssl_c, &test_ctx);
+                wolfSSL_SetIOWriteCtx(ssl_c, &test_ctx);
+                wolfSSL_SetIOReadCtx(ssl_s, &test_ctx);
+                wolfSSL_SetIOWriteCtx(ssl_s, &test_ctx);
```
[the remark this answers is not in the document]
mattia-moffa: No longer relevant

mattia-moffa:
Jenkins retest this please

dgarske (approved):

src/ssl_api_ext.c
```diff
@@ -1068,6 +1070,15 @@ int wolfSSL_CTX_set_TicketHint(WOLFSSL_CTX* ctx, int hint)
     if (hint < 0 || hint > 604800)
         return BAD_FUNC_ARG;
 
+#ifdef WOLFSSL_TICKET_KEY_LIFETIME
```
dgarske: **🟠 [Medium] Hint rejection also fires with a custom TicketEncCb -- macro gate reflects macro presence, not the default callback** · _Logic_

The new check is guarded only by `#ifdef WOLFSSL_TICKET_KEY_LIFETIME`, which is defined whenever the built-in default ticket encryption callback is available -- regardless of whether the application installed its own callback via `wolfSSL_CTX_set_TicketEncCb` (or defined `WOLFSSL_NO_DEF_TICKET_ENC_CB`). The half-lifetime limitation only exists for the default two-key rotation; a user-supplied callback may have no such constraint and may legitimately want a 7-day hint. With this change, such users can no longer set a hint >= LIFETIME/2 even though their callback would honor it. Severity views differ (review: Medium/SUGGEST; review-security: Low) -- stricter Medium kept. Note: gating on `ctx->ticketEncCb == NULL` at call time is only partially reliable because set_TicketHint may be called before set_TicketEncCb, so this is a design trade-off worth deciding/documenting.

**Fix:** Guard the check on the default-callback condition rather than just the macro, e.g. `#if defined(WOLFSSL_TICKET_KEY_LIFETIME) && !defined(WOLFSSL_NO_DEF_TICKET_ENC_CB)`, and/or only enforce the cap when the default callback is actually in effect (`ctx->ticketEncCb == NULL`). At minimum, document the interaction in the Doxygen comment so custom-callback users know the cap still applies.

wolfssl/internal.h
```diff
@@ -2047,8 +2047,8 @@ WOLFSSL_LOCAL int NamedGroupIsPqcHybrid(int group);
         /* Default lifetime is 1 hour from issue of first ticket with key. */
         #define WOLFSSL_TICKET_KEY_LIFETIME       (60 * 60)
     #endif
-    #if WOLFSSL_TICKET_KEY_LIFETIME <= SESSION_TICKET_HINT_DEFAULT
-        #error "Ticket Key lifetime must be longer than ticket life hint."
+    #if WOLFSSL_TICKET_KEY_LIFETIME <= 2 * SESSION_TICKET_HINT_DEFAULT
```
dgarske: **🔵 [Low] Compile-time and runtime hint checks disagree by one for odd WOLFSSL_TICKET_KEY_LIFETIME** · _bug_

The compile-time guard requires `WOLFSSL_TICKET_KEY_LIFETIME > 2 * SESSION_TICKET_HINT_DEFAULT` (real arithmetic), while the runtime setter rejects `hint >= WOLFSSL_TICKET_KEY_LIFETIME / 2` (integer/floor division). For an odd LIFETIME these can disagree by one: e.g. LIFETIME=601, HINT_DEFAULT=300 passes the compile check (601 > 600) but the runtime setter rejects hint=300 (300 >= 601/2 == 300). Then the internal call `wolfSSL_CTX_set_TicketHint(ctx, SESSION_TICKET_HINT_DEFAULT)` would fail. This only bites unusual odd LIFETIME overrides (the default 3600 is even and unaffected), so it is minor, but the two checks are best kept in the same form.

**Fix:** Make the compile-time predicate mirror the runtime floor form, e.g. `#if SESSION_TICKET_HINT_DEFAULT >= WOLFSSL_TICKET_KEY_LIFETIME / 2`, so the default hint is guaranteed acceptable at runtime for any configured LIFETIME.

dgarske (commented):
## Skoll Multi-Scan Review

### Posted findings
- [Medium] [review+review-security] **Hint rejection also fires with a custom TicketEncCb -- macro gate reflects macro presence, not the default callback** — `src/ssl_api_ext.c:1073-1080`
- [Low] [review] **Compile-time and runtime hint checks disagree by one for odd WOLFSSL_TICKET_KEY_LIFETIME** — `wolfssl/internal.h:2050-2052`

#### New hint cap regresses OpenSSL-compat SSL_CTX_set_timeout for common timeouts (= 30 min)
**File:** `src/ssl_api_ext.c:1073-1080 (manifests via src/ssl_sess.c:958-968)`
**Function:** `wolfSSL_CTX_set_TicketHint / wolfSSL_CTX_set_timeout`
**Severity:** High
**Category:** Regression

`wolfSSL_CTX_set_TicketHint` is also called internally by the OpenSSL-compat `wolfSSL_CTX_set_timeout` (`SSL_CTX_set_timeout`). Under `OPENSSL_EXTRA && HAVE_SESSION_TICKET && !NO_WOLFSSL_SERVER`, `wolfSSL_CTX_set_timeout(ctx, to)` forwards the session timeout straight into `wolfSSL_CTX_set_TicketHint(ctx, (int)to)` (ssl_sess.c:968). The new check rejects any `hint >= WOLFSSL_TICKET_KEY_LIFETIME / 2`, i.e. any value >= 1800 seconds (30 min) with the default 3600s key lifetime. Session timeouts of 30+ minutes are the norm (OpenSSL's own default is 7200s; values like 3600/7200/86400 are routine), so `SSL_CTX_set_timeout(ctx, 7200)` -- which previously succeeded (only >604800 was rejected) -- now returns `BAD_FUNC_ARG`. This is a backward-incompatible break of a widely-used public compat API. Worse, `ctx->timeout` is already assigned at ssl_sess.c:958 BEFORE the ticket-hint call fails, so the session-cache timeout is silently applied while the function reports failure, leaving inconsistent state for callers that check the return code. Severity views differ (review: High/BLOCK; review-security: Medium) -- the stricter High is kept because this breaks a public API for common inputs; the session-cache timeout is conceptually independent of ticket-key rotation. Not covered by CI: `test_wolfSSL_CTX_set_timeout` only uses values below 1800.

**Recommendation:** Decouple the session-cache timeout from the ticket-hint constraint. In `wolfSSL_CTX_set_timeout`, clamp the value passed to `wolfSSL_CTX_set_TicketHint` (e.g. `min(to, WOLFSSL_TICKET_KEY_LIFETIME/2 - 1)`) or ignore a ticket-hint rejection when setting the session timeout, so a large session timeout still succeeds and ctx state stays consistent. Keep the hard rejection only for the direct public setter if desired. Add a `test_wolfSSL_CTX_set_timeout` case with a value >= WOLFSSL_TICKET_KEY_LIFETIME/2 (e.g. 3600) to lock in the intended behavior.

_Referenced code: `src/ssl_api_ext.c:1073-1080 (manifests via src/ssl_sess.c:958-968)` (10 lines)_

---

#### No test covers the wolfSSL_CTX_set_timeout - set_TicketHint regression path
**File:** `tests/api.c:27884-27913`
**Function:** `test_wolfSSL_CTX_set_timeout`
**Severity:** Medium
**Category:** test

The new tests in test_ssl_ext.c exercise the direct setter, but the highest-impact behavior change is via the internal caller `wolfSSL_CTX_set_timeout`. The existing `test_wolfSSL_CTX_set_timeout` only uses small values (20, 30, 100), all below the new 1800s cap, so it will neither catch the regression nor lock in whatever the intended behavior becomes. Since the change silently alters what `SSL_CTX_set_timeout` returns for common values, this path needs explicit coverage.

**Recommendation:** Add a case under OPENSSL_EXTRA + HAVE_SESSION_TICKET + server that calls wolfSSL_CTX_set_timeout with a value >= WOLFSSL_TICKET_KEY_LIFETIME/2 (e.g. 3600) and asserts the intended outcome (success with clamped hint per the fix above), and verify ctx state is consistent.

---

### Skipped findings
- [Info] ~~Stricter compile-time #error may break previously-valid custom lifetime/hint configurations~~

dgarske (changes requested):
https://github.com/wolfSSL/wolfssl/pull/10822#pullrequestreview-4649650699

SparkiDev (dismissed):

src/ssl_api_ext.c
mattia-moffa: The two proposed fixes wouldn't work:
- `&& !defined(WOLFSSL_NO_DEF_TICKET_ENC_CB)` is redundant: `WOLFSSL_TICKET_KEY_LIFETIME` is only defined inside `#if !defined(WOLFSSL_NO_DEF_TICKET_ENC_CB) && !defined(NO_WOLFSSL_SERVER)`. Also this is a runtime condition, so no `#ifdef` changes would solve it.
- `ctx->ticketEncCb == NULL` is wrong because by default it's not null, it's `DefTicketEncCb`. But `DefTicketEncCb` is a static symbol from a different file, so that wouldn't work either. Also `set_TicketHint` can be called before `set_TicketEncCb`, which would cause a false positive.

I picked the only remaining possibility, i.e. check it when the default callback actually runs (this means the user won't be able to notice the error at config time though).

dgarske:
Jenkins retest this please: "Error cloning remote repo 'origin'". Timeout

src/internal.c
```diff
@@ -41744,6 +41744,13 @@ static int DefTicketEncCb(WOLFSSL* ssl, byte key_name[WOLFSSL_TICKET_NAME_SZ],
         return BUFFER_E;
     }
 
+    /* The two-key rotation can only honor a hint below half the key lifetime.
+     * Refuse to issue a ticket whose advertised lifetime it cannot cover. */
+    if (enc && 2 * ctx->ticketHint >= WOLFSSL_TICKET_KEY_LIFETIME) {
```
dgarske: **🟠 [Medium] Oversized ticket hint returns FATAL and tears down the whole handshake instead of degrading to a ticket-less connection** · _Logic_

The new check `if (enc && 2 * ctx->ticketHint >= WOLFSSL_TICKET_KEY_LIFETIME) return WOLFSSL_TICKET_RET_FATAL;` runs on every ticket-issuing (enc) call of the default callback. FATAL (-1) propagates DefTicketEncCb -> CreateTicket (goto error) -> SendTls13NewSessionTicket / TLS1.2 SendTicket -> wolfSSL_accept returns WOLFSSL_FATAL_ERROR, so the entire TLS handshake fails and the client cannot connect at all. Session tickets are optional per RFC 8446, so a server that merely cannot issue a resumption ticket could instead complete the handshake and serve a ticket-less (no-resumption) connection; in TLS 1.3 the NewSessionTicket is sent after the handshake is already cryptographically complete, so a pure configuration problem is surfaced as a full connection failure. This is an easily-triggered functional regression, not attacker-controlled: in OpenSSL-compat builds (OPENSSL_EXTRA && HAVE_SESSION_TICKET && !NO_WOLFSSL_SERVER), wolfSSL_CTX_set_timeout(ctx, to) copies the session timeout straight into ctx->ticketHint (src/ssl_sess.c:968, clamped only to [0,604800]). With the default WOLFSSL_TICKET_KEY_LIFETIME of 3600, any timeout >= 1800 (e.g. 3600, or OpenSSL's common 7200) makes 2*ticketHint >= 3600 true, so EVERY handshake that issues a ticket now fails. BEFORE this PR the same misconfiguration let the FIRST connection succeed (it generated the secondary key and issued a ticket) and only failed reused-CTX connections thereafter (issue #10796); AFTER this PR even the first connection fails. The PR's own test fixture had to install a custom callback to avoid this new limit…

**Fix:** Confirm with maintainers that failing the handshake is the intended contract. Prefer a less destructive response: clamp the advertised ticket lifetime hint down to a value the two-key rotation can cover, or skip issuing the ticket (full handshake next time) via a distinct 'skip ticket' return path in CreateTicket, instead of returning FATAL. At minimum, guard the OpenSSL-compat propagation in wolfSSL_CTX_set_timeout so large session timeouts do not silently push ticketHint past WOLFSSL_TICKET_KEY_LIFETIME/2, and document in release notes that SSL_CTX_set_timeout / ticket-hint values >= half the ticket key lifetime disable default-callback tickets and will fail the handshake.

src/internal.c
```diff
@@ -41744,6 +41744,13 @@ static int DefTicketEncCb(WOLFSSL* ssl, byte key_name[WOLFSSL_TICKET_NAME_SZ],
...
+        WOLFSSL_MSG("Ticket hint exceeds half the ticket key lifetime");
+        return WOLFSSL_TICKET_RET_FATAL;
```
dgarske: **🔵 [Low] New failure uses WOLFSSL_TICKET_RET_FATAL while other encrypt-side returns in this function use REJECT** · _Convention_

Every other failure on the encryption (enc) path inside DefTicketEncCb returns WOLFSSL_TICKET_RET_REJECT (RNG failure at 41781, encrypt failure at 41843), while this new check returns WOLFSSL_TICKET_RET_FATAL. On the encrypt/create path both values are treated identically by the caller (both hit `goto error` at internal.c:40393 and fail the handshake), so this is not a functional defect. FATAL arguably communicates 'permanent misconfiguration' better than REJECT, but the inconsistency within the same function is worth a deliberate choice.

**Fix:** Either keep FATAL intentionally (permanent config error) with a brief comment, or use REJECT to match the other encrypt-side error returns in this function.

tests/api/test_ssl_ext.c
```diff
@@ -195,6 +195,77 @@ int test_wolfSSL_CTX_set_TicketHint_ext(void)
     return EXPECT_RESULT();
 }
 
+#if defined(HAVE_MANUAL_MEMIO_TESTS_DEPENDENCIES) && defined(WOLFSSL_TLS13) \
+    && defined(HAVE_SESSION_TICKET) && !defined(WOLFSSL_NO_DEF_TICKET_ENC_CB) \
+    && !defined(NO_WOLFSSL_SERVER)
+/* Trivial custom ticket encryption callback: it has no key-lifetime constraint,
+ * so it must be able to issue a ticket for any hint. */
+static int test_TicketHint_custom_encCb(WOLFSSL* ssl,
```
dgarske, line 203: **⚪ [Info] Trivial custom ticket-encryption callback duplicated across two test files** · _Style_

test_TicketHint_custom_encCb (test_ssl_ext.c) and test_tls13_max_lifetime_ticketEncCb (test_tls13.c:5737-5754) are byte-for-byte equivalent trivial XOR-0xA5 ticket callbacks. Cross-file duplication of small test helpers is common in this suite and not harmful, but a shared helper in tests/utils.* would avoid two copies drifting.

**Fix:** Optional: factor the trivial XOR ticket callback into a shared test utility if it is reused again; otherwise leave as-is.

dgarske (commented):
## Skoll Multi-Scan Review

### Posted findings
- [Medium] [review+review-security] **Oversized ticket hint returns FATAL and tears down the whole handshake instead of degrading to a ticket-less connection** — `src/internal.c:41749`
- [Low] [review] **New failure uses WOLFSSL_TICKET_RET_FATAL while other encrypt-side returns in this function use REJECT** — `src/internal.c:41751`
- [Info] [review] **Trivial custom ticket-encryption callback duplicated across two test files** — `tests/api/test_ssl_ext.c:203`

#### Compile-time ticket lifetime guard does not cover the new runtime 2*hint constraint
**File:** `wolfssl/internal.h:2050`
**Function:** `DefTicketEncCb`
**Severity:** Low
**Category:** Convention

internal.h statically enforces WOLFSSL_TICKET_KEY_LIFETIME > SESSION_TICKET_HINT_DEFAULT, but the new runtime rule requires 2 * ticketHint < WOLFSSL_TICKET_KEY_LIFETIME. With the defaults (lifetime 3600, default hint 300) there is no problem (600 < 3600). However, a user who overrides WOLFSSL_TICKET_KEY_LIFETIME to a value in the range (SESSION_TICKET_HINT_DEFAULT, 2*SESSION_TICKET_HINT_DEFAULT] (e.g. 400-600) passes the compile-time #error check yet now has every default-hint ticket handshake fail at runtime, with no build-time warning. This is a narrow, already-degenerate configuration, but the failure surfaces only at runtime.

**Recommendation:** Consider tightening the static assert to `#if WOLFSSL_TICKET_KEY_LIFETIME <= 2 * SESSION_TICKET_HINT_DEFAULT` (or documenting the 2*hint requirement near WOLFSSL_TICKET_KEY_LIFETIME) so users who shrink the key lifetime are warned at build time rather than seeing runtime handshake failures.

_Referenced code: `wolfssl/internal.h:2050-2052` (3 lines)_

---

tests/api/test_ssl_ext.c
mattia-moffa: De-duplicating the callbacks would require moving them to a shared header and C (e.g. `tests/utils.c`, `tests/utils.h`). I think they're simple enough that duplication is fine.

mattia-moffa:
Jenkins retest this please

dgarske (approved):

PR #10827, opened 2026-07-01
 Falcon: native implementation replacing liboqs, with crypto callbacks and ARM acceleration. Deprecate liboqs support.

danielinux:
# Description

Replaces wolfSSL's liboqs-based Falcon wrapper with a complete **native
wolfCrypt implementation** of Falcon (key generation, signing and
verification, levels 1 and 5), removes the liboqs dependency entirely, adds the
standard crypto-callback interface, and adds optional per-architecture
acceleration for x86-64 and ARM.

## What's in it

**Native implementation** (`falcon.c` + `wc_falcon*.c`, no liboqs):
- Public API `wc_falcon_*` / `falcon_key` (single surface; the redundant
  `wc_FnDsa_*` layer is gone). Verify uses an integer NTT mod q=12289 with
  cached twiddle tables; signing uses the FFT / ffLDL tree + discrete Gaussian
  sampler; keygen uses the RNS/big-integer NTRU solver.
- **Floating-point `fpr` seam** with selectable backends:
  - portable, constant-time **integer-emulated** FP — the default;
  - `--enable-falcon-double` — inline native `double` (fastest with a
    constant-time HW FPU);
  - `--enable-falcon-asm` — x86-64 SSE2 out-of-line `fpr` backend
    (generated; see companion wolfssl-scripts PR);
  - `--enable-falcon-avx2` — x86-64 AVX2 (4-wide) FFT.
- **Division-free (Barrett) verify NTT**, so no hardware divide is required on
  Cortex-M / embedded targets.

**Remove liboqs**: Falcon was the last liboqs consumer. Drops `--with-liboqs`
(and `-loqs`), the `wolfcrypt/src/port/liboqs` layer, the CMake `WOLFSSL_OQS`
option / `FindOQS.cmake`, and the VS/Zephyr/INTIME/Docker references. The
library links no liboqs. `HAVE_FALCON` is added to the experimental-settings
gate so the requirement holds across every build system.

**Crypto callbacks** (`WOLF_CRYPTO_CB`): keygen/sign/verify dispatch to
`wc_CryptoCb_MakePqcSignatureKey` / `PqcSign` / `PqcVerify`
(`WC_PQC_SIG_TYPE_FALCON`), with software fall-through. Adds
`WOLF_CRYPTO_CB_ONLY_FALCON` (first PQC signature to have it, mirroring
`_RSA`/`_ECC`): the callback becomes authoritative and the native core is
compiled out.

**ARM acceleration** (portable-C + intrinsics, no hand-asm):
- **Cortex-M DSP verify NTT** — auto-enabled on `__ARM_FEATURE_DSP`
  (Cortex-M4/M7/M33). Packed 2-lane butterflies (`SMLABB`/`SMLATB` +
  `SADD16`/`SSUB16` + `USUB16`/`SEL`) and an `SMUAD` squared-norm; bit-identical
  to the scalar path.
- **AArch64 NEON FFT** — `--enable-falcon-neon`, the 2-wide `float64x2_t` + FMA
  counterpart of the AVX2 FFT for the signing/keygen path.

# Testing

- `wolfcrypt/test` `falcon_test`: KAT verify + native keygen/sign/verify
  round-trip; the crypto-callback test drives every Falcon op through the
  callback and asserts the cb path was taken (incl. a CB_ONLY check).
- **Native ↔ liboqs interop** harness + CI workflow (both directions,
  keygen+sign vs verify).
- **Invariant fuzzer**: ~79k random seeds per level, checking the exact NTRU
  equation `f·G − g·F = q` on every keygen, `h·f ≡ g (mod q)`, the norm bound,
  sign→native+liboqs verify, and forgery rejection — zero failures observed with this set.
- **Emulator tests for ARM acceleration** (both under `IDE/`): `m33mu-falcon-verify` runs the DSP
  NTT on an STM32H563/Cortex-M33 (m33mu) — genuine sig accepted, tampered
  rejected; `qemu-falcon-neon` runs a full keygen→sign→verify at levels 1 and 5
  on qemu-system-aarch64 with the NEON FFT.
- Builds green (`testwolfcrypt` + full testsuite) across default / cryptocb /
  CB_ONLY / verify-only and each fpr backend.

## Build options

`--enable-falcon`            native Falcon (requires --enable-experimental)
`--enable-falcon-double`     inline native-double fpr backend
`--enable-falcon-asm`       x86-64 SSE2 fpr asm backend
`--enable-falcon-avx2`       x86-64 AVX2 FFT
`--enable-falcon-neon`       AArch64 NEON FFT

# Checklist

 - [x] added tests
 - [x] updated/added doxygen
 - [x] Updated manual and documentation


zephyr/CMakeLists.txt
```diff
@@ -125,7 +125,6 @@ if(CONFIG_WOLFSSL)
     zephyr_library_sources(${ZEPHYR_CURRENT_MODULE_DIR}/wolfcrypt/src/wolfevent.c)
     zephyr_library_sources(${ZEPHYR_CURRENT_MODULE_DIR}/wolfcrypt/src/wolfmath.c)
 
-    zephyr_library_sources(${ZEPHYR_CURRENT_MODULE_DIR}/wolfcrypt/src/port/liboqs/liboqs.c)
```
Frauschi, line 128: I think some of the new falcon-related source files must be added here as well? `falcon.c` is already in here.

configure.ac
```diff
@@ -1729,69 +1725,103 @@ then
...
+    [AS_HELP_STRING([--enable-falcon],[Enable Falcon post-quantum signatures (native, no liboqs; requires --enable-experimental) (default: disabled)])],
     [ ENABLED_FALCON=$enableval ],
     [ ENABLED_FALCON=no ])
 
-if test "$ENABLED_FALCON" = "yes" && test "$ENABLED_LIBOQS" = "no"; then
-    AC_MSG_ERROR([--enable-falcon requires --with-liboqs.])
+# --enable-falcon-asm selects the per-architecture assembly fpr backend for
+# Falcon (currently x86-64 SSE2 only); the portable constant-time integer
+# emulation remains the default. It implies --enable-falcon.
+AC_ARG_ENABLE([falcon-asm],
```
Frauschi: Instead of adding all these new standalone options, we should use the common approach of having these as additional options to the `--enable-falcon` option.

So something like `--enable-falcon=yes,asm`

configure.ac
```diff
@@ -12929,6 +12997,7 @@ echo "   * XMSS:                       $ENABLED_XMSS"
 echo "   * SLH-DSA                     $ENABLED_SLHDSA"
 echo "   * MLKEM:                      $ENABLED_MLKEM"
 echo "   * ML-DSA:                     $ENABLED_MLDSA"
+echo "   * Falcon:     $ENABLED_FALCON"
```
Frauschi: Spacing (also below)

wolfcrypt/src/falcon.c
```diff
@@ -27,17 +27,84 @@
...
     #define WOLFSSL_MISC_INCLUDED
     #include <wolfcrypt/src/misc.c>
 #endif
 
+/* Store a second copy of the public key in key->k immediately after the private
+ * key, reproducing the historical concat(private,public) layout that
+ * wc_falcon_check_key compares against. No-op unless both halves are set.
+ * Defined unconditionally: wc_falcon_import_public (a verify-only operation)
+ * calls it. */
+static void falcon_store_pub_behind_priv(falcon_key* key)
```
Frauschi: Is this concatenation still required with the native implementation? If not, I'd drop that to reduce the key memory footprint. I think this is also not exposed to the user in any way, so safe to do within the existing API.

Frauschi (changes requested):
Some smaller initial things regarding the infrastructure. I haven't reviewed the heavy math stuff, this is probably better handled by @SparkiDev.

wolfcrypt/src/wc_falcon_keygen.c
```diff
@@ -0,0 +1,1904 @@
...
+    for (u = 0; u < n; u++) {
+        if (ff[u] == 0) {
+            XFREE(zetas, heap, DYNAMIC_TYPE_TMP_BUFFER);
+            return 0;
+        }
+        h[u] = (word16)(((word64)h[u] * mq_modinv(ff[u])) % FALCON_Q);
+    }
+    mq_intt(h, n, izetas);
+
+    XFREE(zetas, heap, DYNAMIC_TYPE_TMP_BUFFER);
```
Frauschi: **🟡 [Medium] Private-key NTT(f) left in freed heap in falcon_compute_public (no ForceZero)**
*💡 SUGGEST*  `bug`

`falcon_compute_public` allocates a single `3*n` word16 scratch (`zetas`), whose last third `ff = izetas + n` holds the secret polynomial `f` and, after `mq_ntt(ff,...)`, the NTT image of `f` — enough to recover the Falcon private key. On both the rejection path (`if (ff[u] == 0)` at line 303) and the success path (line 310) the buffer is released with `XFREE` but never wiped first. `falcon_keygen` carefully `wc_ForceZero`s its main `tmpbuf`, but this secondary allocation escapes that hardening, violating the wolfSSL rule to zero sensitive data before free.

**Recommendation:** Call `wc_ForceZero` on the full `zetas` buffer immediately before each `XFREE` (both the reject and success returns).

wolfcrypt/src/wc_falcon_sampler.c
```diff
@@ -0,0 +1,362 @@
...
+    r = fpr_sub(mu, fpr_of((sword64)s));
+
+    /* dss = 1/(2*sigma^2) = 0.5 * isigma^2. */
+    dss = fpr_half(fpr_sqr(isigma));
+
+    /* ccs = sigma_min / sigma = sigma_min * isigma. */
+    ccs = fpr_mul(isigma, spc->sigma_min);
+
+    /* Sample on center r. */
+    for (;;) {
```
Frauschi: **🟡 [Medium] Sampler rejection loop can hang on a mid-signature PRNG failure**
*💡 SUGGEST*  `bug`

The PR adds a sticky PRNG error flag (`falcon_prng.err`) that `falcon_sign_core`/`falcon_do_sign_tree` check *between* full sign-tree attempts. But `falcon_sampler_z`'s inner `for(;;)` rejection loop never consults `spc->p.err`. If a SHAKE256 squeeze fails mid-sampling, `falcon_prng_get_u8`/`get_u64` return a constant byte (`buf[0]`), so `gaussian0`, the bimodal bit `b`, and `berexp` all become deterministic; if that fixed input makes `berexp` reject, the loop never returns to `do_sign_tree_once`, and the sticky-error check is never reached — an unbounded loop instead of a clean error. Only reachable under an already-failing RNG, but it defeats the very error-handling the PR engineered.

**Recommendation:** Add a `spc->p.err` check inside the `for(;;)` loop that breaks/returns once the sticky error is latched, so a failed squeeze terminates instead of potentially hanging.

wolfcrypt/src/wc_falcon_sign.c
```diff
@@ -0,0 +1,716 @@
...
+    /* tmp held the secret-derived Gram matrix and ffLDL intermediates. */
+    wc_ForceZero(tmp, (word32)((size_t)6 * n * sizeof(fpr)));
+    XFREE(tmp, heap, DYNAMIC_TYPE_TMP_BUFFER);
+    return 0;
+}
+
+/* ==================================================================== */
+/* Fast Fourier sampling.                                                */
+
+void falcon_ffSampling_fft(falcon_samplerZ samp, void* samp_ctx,
```
Frauschi: **🟡 [Medium] Recursion used in signing path violates wolfSSL no-recursion convention**
*💡 SUGGEST*  `convention`

`falcon_ffSampling_fft` and `ffLDL_fft_inner` are self-recursive (the general case at the bottom recurses on `logn-1`). wolfSSL C style bans recursion in favor of iteration. Recursion depth is bounded by `logn` (<=10) so stack use is safe, and these are faithful ports of the Falcon reference — but the convention deviation should be resolved rather than left implicit, since the no-recursion rule is a hard house rule.

**Recommendation:** Preferably convert to an explicit iterative/stack-based form. If the recursion is kept as a faithful reference port, add a header comment justifying the bounded-depth (<=10) recursion as an intentional, reviewed deviation from the no-recursion rule.

wolfcrypt/src/wc_falcon_sampler.c
```diff
@@ -0,0 +1,362 @@
...
+    /* Latch the first failure. get_u8/get_u64 have no error return, so a squeeze
+     * failure is made sticky here and checked by the signer (falcon_sign_core),
+     * which rejects any signature produced from an invalid PRNG state instead of
+     * consuming stale buffer bytes. */
+    if (ret != 0 && p->err == 0)
+        p->err = ret;
+    return ret;
+}
+
+int falcon_prng_init(falcon_prng* p, WC_RNG* rng)
```
Frauschi: **🔵 [Low] SHAKE256 context leaked when sampler init fails after wc_InitShake256 succeeds**
*💡 SUGGEST*  `bug`

`falcon_prng_init` calls `wc_InitShake256` and then, on the same path, may fail in `wc_Shake256_Absorb` or `falcon_prng_refill` (SqueezeBlocks). On that failure it returns non-zero WITHOUT calling `wc_Shake256_Free` on the sponge it just initialized. The only caller (`falcon_native_sign_msg`, wc_falcon.c:816) sets `haveSpc = 1` only when `falcon_sampler_init` returns 0, so the `out:` handler skips `wc_Shake256_Free(&spc.p.shake)`. In `WOLFSSL_ASYNC_CRYPT` builds `wc_InitShake256` allocates a device context that then leaks. Non-async builds are unaffected.

**Recommendation:** Free the SHAKE sponge on the error path after a successful `wc_InitShake256` (guard so it is only freed when it was initialized).

wolfcrypt/src/falcon.c
```diff
@@ -646,18 +639,23 @@ int wc_falcon_export_key(falcon_key* key, byte* priv, word32 *privSz,
...
+ *
+ * Note: this verifies both halves of the pair are loaded. It does not yet
+ * perform a full cryptographic cross-check (recomputing the public key h from
+ * the private (f, g) and comparing it against the stored public key); that is a
+ * TODO once a standalone public-key-from-private helper is exposed by the native
+ * core. The previous implementation compared the stored public key against a
+ * duplicate copy kept behind the private key, which was always a copy of the
+ * same bytes and so could never detect a mismatch.
  */
 int wc_falcon_check_key(falcon_key* key)
```
Frauschi, line 9276: **🔵 [Low] wc_falcon_check_key doxygen overstates behavior (no cryptographic cross-check)**
*💡 SUGGEST*  `question`

The published doxygen (doc/dox_comments/header_files/falcon.h:414-419) states the function "verif[ies] that the stored public key matches the private key" and returns `PUBLIC_KEY_E` "if the public and private keys are inconsistent." The new implementation only checks that both halves are *present* (`pubKeySet`/`prvKeySet`) and returns 0 otherwise — there is no cross-check (the in-code comment documents this as a TODO). A caller trusting the doc could import a mismatched public/private pair and get success from `check_key`. Not a regression (the old compare was itself ineffective), but the public documentation now misrepresents the contract.

**Recommendation:** Update the doxygen in doc/dox_comments/header_files/falcon.h to state it only checks that both key halves are loaded (not that they are cryptographically consistent), or implement the cross-check.

wolfcrypt/src/wc_falcon_codec.c
```diff
@@ -0,0 +1,359 @@
...
+        unsigned logn, unsigned bits)
+{
+    size_t n = (size_t)1 << logn;
+    size_t out_len = ((n * bits) + 7) >> 3;
+    size_t u, v;
+    int minv, maxv;
+    word32 acc = 0, mask;
+    unsigned acc_len = 0;
+
+    maxv = (1 << (bits - 1)) - 1;
```
Frauschi: **🔵 [Low] Undefined shift in trim_i8 encode/decode if bits is 0 or >= 32**
*🔧 NIT*  `bug`

`maxv = (1 << (bits - 1)) - 1` (line 189) and `mask2 = (word32)1 << (bits - 1)` (line 233) are UB if `bits == 0` (underflow to UINT_MAX shift) or `bits >= 32`. Both functions are `WOLFSSL_LOCAL` and only ever called with `falcon_max_fg_bits[]`/`falcon_max_FG_bits[]` values in 5..8, so not currently reachable — but neither function validates `bits`.

**Recommendation:** Add a `bits` range guard at the top of both `falcon_trim_i8_encode` and `falcon_trim_i8_decode` for defense in depth.

wolfcrypt/src/wc_falcon_keygen.c
```diff
@@ -0,0 +1,1904 @@
...
+
+        /* Success. */
+        break;
+    }
+
+    ret = 0;
+
+out:
+    falcon_rng_free(&rc);
+    if (h == NULL) {
```
Frauschi: **⚪ [Info] Internally allocated h scratch freed without ForceZero**
*🔧 NIT*  `convention`

When the caller passes `h == NULL`, `falcon_keygen` allocates `hwork` and frees it at exit without `wc_ForceZero`. By the time of free the buffer holds the *public* key `h` (secret `g` was overwritten in place by `mq_intt`), so residual secret exposure is minimal; flagged only for consistency with the surrounding `tmpbuf` hardening.

**Recommendation:** Zeroize `hwork` before free for defense in depth (low priority since it holds public data at that point).

wolfcrypt/src/wc_falcon_sampler.c
```diff
@@ -0,0 +1,362 @@
...
+    p->err = 0;
+    ForceZero(seed, (word32)sizeof(seed));
+
+    if (ret == 0)
+        ret = falcon_prng_refill(p);
+
+    return ret;
+}
+
+byte falcon_prng_get_u8(falcon_prng* p)
```
Frauschi: **⚪ [Info] PRNG re-squeezes on every call once error is latched**
*🔧 NIT*  `style`

After a refill failure `len` stays 0, so each subsequent `get_u8`/`get_u64` re-invokes `falcon_prng_refill` (which fails again) and reads stale-but-initialized `buf` bytes. No uninitialized read or security impact given the sticky `err` + `falcon_sign_core` post-check, but it wastes work re-issuing failing `SqueezeBlocks` calls and makes the failure contract implicit.

**Recommendation:** Early-return once `p->err` is latched to avoid repeated failing squeezes and make the error contract explicit.

Frauschi (changes requested):
## 🐺 Skoll Code Review

**Findings:** 8 total — 8 posted, 0 skipped

### Posted findings
- [Medium] **Private-key NTT(f) left in freed heap in falcon_compute_public (no ForceZero)** — `wolfcrypt/src/wc_falcon_keygen.c:310`
- [Medium] **Sampler rejection loop can hang on a mid-signature PRNG failure** — `wolfcrypt/src/wc_falcon_sampler.c:321`
- [Medium] **Recursion used in signing path violates wolfSSL no-recursion convention** — `wolfcrypt/src/wc_falcon_sign.c:367`
- [Low] **SHAKE256 context leaked when sampler init fails after wc_InitShake256 succeeds** — `wolfcrypt/src/wc_falcon_sampler.c:133`
- [Low] **wc_falcon_check_key doxygen overstates behavior (no cryptographic cross-check)** — `wolfcrypt/src/falcon.c:657`
- [Low] **Undefined shift in trim_i8 encode/decode if bits is 0 or >= 32** — `wolfcrypt/src/wc_falcon_codec.c:189`
- [Info] **Internally allocated h scratch freed without ForceZero** — `wolfcrypt/src/wc_falcon_keygen.c:1892`
- [Info] **PRNG re-squeezes on every call once error is latched** — `wolfcrypt/src/wc_falcon_sampler.c:158`

wolfcrypt/src/falcon.c
Frauschi: In the new comment, you also mention that currently no "public-key-from-private helper is exposed" for the `check_key` functionality. Is the core functionality already implemented and only a wrapper is needed, or would that require more effort to integrate? If easy, I'd suggest adding it so we have proper `check_key` functionality.

danielinux:
[405db5c](https://github.com/wolfSSL/wolfssl/pull/10827/commits/465db5cd7c0435989e6ad5b6367921400c893b80) ran in fuzzer with 60,000,144 sign/verify round-trips, zero failures.

wolfcrypt/src/wc_falcon.c
```diff
@@ -0,0 +1,1147 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1335, USA
+ */
+
+/* Native Falcon implementation for wolfCrypt.
```
Frauschi: **🔵 [Low] Stale "Phase 1: verification only" file header comment**
*🔧 NIT*  `style`

The file banner still says `Phase 1: verification only (integer arithmetic, no floating point). The signature/keygen paths and the floating-point primitive seam are added in later phases.` But this same file now contains `falcon_native_make_key`, `falcon_native_sign_msg`, and pulls in the FFT/sampler/keygen seams. The comment is misleading to a future maintainer.

**Suggestion:**
```suggestion
/* Native Falcon implementation for wolfCrypt: keygen, sign and verify cores
 * (falcon_native_*). Verify is integer-only; sign/keygen use the fpr FFT seam.
 * See wolfssl/wolfcrypt/falcon.h. */
```

Frauschi (commented):
## 🐺 Skoll Code Review

### Posted findings
- [Low] **Stale "Phase 1: verification only" file header comment** — `wolfcrypt/src/wc_falcon.c:22-26`
- [Low] **wc_falcon_verify_msg doc comment references non-existent contextLen** — `wolfcrypt/src/falcon.c:160-162`
- [Medium] **Sign compression-fit retry silently caps at 32 attempts** — `wolfcrypt/src/wc_falcon.c:824-855`
- [Medium] **No unit test exercises the concat(priv,pub) legacy import size path directly** — `wolfcrypt/src/falcon.c:448-484`

- [Low] Bare-scope { } blocks used to introduce mid-function declarations

</details>

wolfcrypt/src/falcon.c
```diff
@@ -149,9 +165,6 @@ int wc_falcon_verify_msg(const byte* sig, word32 sigLen, const byte* msg,
                         word32 msgLen, int* res, falcon_key* key)
```
Frauschi, line 8775: **🔵 [Low] wc_falcon_verify_msg doc comment references non-existent contextLen**
*🔧 NIT*  `style`

The Doxygen-ish comment says `returns BAD_FUNC_ARG when a parameter is NULL or contextLen is zero when and BUFFER_E when sigLen is less than FALCON_LEVEL1_SIG_SIZE`. There is no `contextLen` parameter, the sentence is garbled ("zero when and"), and BUFFER_E bounds are now level-dependent. This is adjacent to the changed wrapper body and worth correcting while here.

**Suggestion:**
```suggestion
*  returns BAD_FUNC_ARG when a parameter is NULL or the public key is not set,
 *          BUFFER_E when sigLen is out of range for the level,
 *          0 on a completed verification (check *res for validity).
```

wolfcrypt/src/wc_falcon.c
```diff
@@ -0,0 +1,1147 @@
...
+    }
+    ret = falcon_sampler_init(&spc, (int)logn, rng);
+    if (ret != 0) {
+        goto out;
+    }
+    haveSpc = 1;
+
+    /* Each attempt draws a fresh nonce and samples a signature; retry if the
+     * compressed form does not fit the level's maximum length. */
+    for (attempt = 0; attempt < 32; attempt++) {
```
Frauschi: **🟡 [Medium] Sign compression-fit retry silently caps at 32 attempts**
*💡 SUGGEST*  `question`

The outer loop retries at most 32 times drawing a fresh nonce until `falcon_comp_encode` fits within `sigMax`. If all 32 attempts overflow, the function returns `BUFFER_E`. Statistically 32 is more than sufficient (over-length compression is rare), so this is not a bug, but unlike the sign-tree restart loop in `falcon_do_sign_tree` (which documents/logs its bound and has an explicit `FALCON_SIGN_MAX_RESTARTS`), this 32 is an unnamed magic number with no comment on why 32 is safe. A wedged RNG returning constant bytes could also spin these 32 attempts before failing.

**Suggestion:**
```suggestion
#define FALCON_SIGN_MAX_ENCODE_RETRIES 32  /* over-length comp-encode is rare */
    for (attempt = 0; attempt < FALCON_SIGN_MAX_ENCODE_RETRIES; attempt++) {
```

wolfcrypt/src/falcon.c
```diff
@@ -646,18 +639,24 @@ int wc_falcon_export_key(falcon_key* key, byte* priv, word32 *privSz,
     return ret;
```
Frauschi, line 9257: **🟡 [Medium] No unit test exercises the concat(priv,pub) legacy import size path directly**
*💡 SUGGEST*  `test`

`wc_falcon_import_private_only` accepts two sizes (raw KEY_SIZE and legacy concat PRV_KEY_SIZE) and, in the concat case, recovers the public key from the trailing bytes and sets `pubKeySet`. `test_wc_falcon_import_export` covers `import_private_key(prvpub, prvpubLen, NULL, 0, ...)` which flows through this concat branch via check_key, but there is no direct assertion that a concat import leaves `pubKeySet=1` and that a subsequent verify succeeds without a separate public import. Given the legacy-compat intent, an explicit test would prevent a regression in the recover-pub-from-concat logic.

**Recommendation:** Add a test that imports a concat(priv,pub) buffer via wc_falcon_import_private_only, then verifies a signature using only that key (no separate import_public), asserting pubKeySet was recovered.

Frauschi:
**🟡 [Medium] Alternate fpr/FFT backends (asm, double, avx2, neon) not exercised by CI**
*💡 SUGGEST*  `test`

The added CI job builds only `./configure --enable-falcon --enable-experimental` (`.github/workflows/falcon-interop.yml:126`) — i.e. just the default portable-integer fpr backend on x86-64. The four selectable backends (`asm`, `double`, `avx2`, `neon`) are the highest-risk new code (SIMD intrinsics, generated asm), yet they're currently validated only out-of-band via the `IDE/` emulator harnesses.

Good news: **all four can run on GitHub-hosted runners** — none need self-hosted hardware or a mandatory emulator. Backend dispatch is compile-time with no runtime CPUID fallback, so the runner CPU is the deciding factor. What each needs to *execute*:

| Backend | ISA needed to run | Runner |
|---|---|---|
| `double` | none (hardware `double`) | any (`ubuntu-latest`) |
| `asm` | SSE2 scalar (`movsd`/`addsd`/`mulsd`) — baseline on all x86-64 | `ubuntu-latest` |
| `avx2` | AVX2 + FMA3 — present on all GitHub x86 runners (Xeon 8272CL/8370C, EPYC 7763) | `ubuntu-latest` |
| `neon` | ARM64 NEON + FMA | `ubuntu-24.04-arm` (free ARM64 runners for public repos) or `macos-14` (Apple Silicon) |

The `avx2` TU uses per-function `__attribute__((target("avx2,fma")))`, so it compiles cleanly without a global `-mavx2` — only the *runtime* CPU needs the features, and GitHub's standard runners have had AVX2+FMA3 for years.

**Suggested matrix** (respecting the configure constraints — `double` XOR `asm`, and `avx2`/`neon` imply `double`):

- **`ubuntu-latest` (x86-64):** matrix `--enable-falcon={asm,double,avx2}` → build + run `testwolfcrypt`. Covers 3 of 4 natively.
- **`ubuntu-24.04-arm` (ARM64, free for public repos):** `--enable-falcon=neon` → build + run. Covers NEON natively.

One caveat: since backend selection is compile-time with no CPUID guard, an `avx2` CI job is implicitly pinned to AVX2-capable runners. That's effectively always true on GitHub, but a one-line `grep -q avx2 /proc/cpuinfo` guard (or runtime dispatch) would make it robust if a runner class ever changes.

danielinux:
> **🟡 [Medium] Alternate fpr/FFT backends (asm, double, avx2, neon) not exercised by CI** _💡 SUGGEST_ `test`
> 
> The added CI job builds only `./configure --enable-falcon --enable-experimental` (`.github/workflows/falcon-interop.yml:126`) — i.e. just the default portable-integer fpr backend on x86-64. The four selectable backends (`asm`, `double`, `avx2`, `neon`) are the highest-risk new code (SIMD intrinsics, generated asm), yet they're currently validated only out-of-band via the `IDE/` emulator harnesses.
> 
> Good news: **all four can run on GitHub-hosted runners** — none need self-hosted hardware or a mandatory emulator. Backend dispatch is compile-time with no runtime CPUID fallback, so the runner CPU is the deciding factor. What each needs to _execute_:
> Backend 	ISA needed to run 	Runner
> `double` 	none (hardware `double`) 	any (`ubuntu-latest`)
> `asm` 	SSE2 scalar (`movsd`/`addsd`/`mulsd`) — baseline on all x86-64 	`ubuntu-latest`
> `avx2` 	AVX2 + FMA3 — present on all GitHub x86 runners (Xeon 8272CL/8370C, EPYC 7763) 	`ubuntu-latest`
> `neon` 	ARM64 NEON + FMA 	`ubuntu-24.04-arm` (free ARM64 runners for public repos) or `macos-14` (Apple Silicon)
> 
> The `avx2` TU uses per-function `__attribute__((target("avx2,fma")))`, so it compiles cleanly without a global `-mavx2` — only the _runtime_ CPU needs the features, and GitHub's standard runners have had AVX2+FMA3 for years.
> 
> **Suggested matrix** (respecting the configure constraints — `double` XOR `asm`, and `avx2`/`neon` imply `double`):
> 
>     * **`ubuntu-latest` (x86-64):** matrix `--enable-falcon={asm,double,avx2}` → build + run `testwolfcrypt`. Covers 3 of 4 natively.
> 
>     * **`ubuntu-24.04-arm` (ARM64, free for public repos):** `--enable-falcon=neon` → build + run. Covers NEON natively.
> 
> 
> One caveat: since backend selection is compile-time with no CPUID guard, an `avx2` CI job is implicitly pinned to AVX2-capable runners. That's effectively always true on GitHub, but a one-line `grep -q avx2 /proc/cpuinfo` guard (or runtime dispatch) would make it robust if a runner class ever changes.

I'll get there in the mc-dc coverage campaign
see #10845 and  #10876 -- I would keep those out for now, there will be new api tests added as soon as we get rid of liboqs.

Frauschi (dismissed):
LGTM, over to @SparkiDev.

Frauschi (commented):
## 🐺 Skoll Code Review

### Posted findings
- [Medium] **WOLFSSL_FALCON_SIGN_STATS references an undefined symbol (link error when enabled)** — `wolfcrypt/src/falcon.c:6947-6951`
- [Low] **configure.ac --enable-falcon help omits the accepted 'small-mem' sub-option** — `configure.ac:1840`
- [Info] **Native check_key uses hardware modulo despite division-free NTT elsewhere** — `wolfcrypt/src/falcon.c:8221-8227`

- [Low] Pervasive bare-scope `{ ... }` blocks in new Falcon C code

</details>

wolfcrypt/src/falcon.c
```diff
@@ -21,22 +21,8437 @@
...
+            sp = ffSampling_descend(stk, sp, samp, samp_ctx,
+                    f->tmp, f->tmp + hn,
+                    f->tree + n + ffLDL_treesize(f->logn - 1),
+                    f->z1, f->z1 + hn, f->logn - 1, f->tmp + n);
+            break;
+
+        case FF_SAMP_STAGE_T0:
+            f->stage = FF_SAMP_STAGE_DONE;
+            /* Merge the sampled halves back into z1. */
+
```
Frauschi: **🟡 [Medium] WOLFSSL_FALCON_SIGN_STATS references an undefined symbol (link error when enabled)**
*💡 SUGGEST*  `bug`

The optional instrumentation guarded by `WOLFSSL_FALCON_SIGN_STATS` declares `extern unsigned long falcon_sign_restart_count;` and increments it, but no translation unit in the PR defines this symbol (grep finds only this reference). Building with `WOLFSSL_FALCON_SIGN_STATS` defined will fail to link with an undefined-symbol error, so the debug knob is non-functional as shipped. Additionally, the `extern` is declared mid-function inside a bare block, which conflicts with the wolfSSL C89 declaration-at-top convention.

**Suggestion:** Provide a definition (e.g. `#ifdef WOLFSSL_FALCON_SIGN_STATS\nunsigned long falcon_sign_restart_count = 0;\n#endif` at file scope) and move the `extern` declaration to file scope / top of function, or drop the instrumentation if it isn't wired up.

configure.ac
```diff
@@ -1798,69 +1798,107 @@ then
...
+#            (x86-64); implies double. The AVX2 functions carry target("avx2,fma")
+#            attributes, so the TU builds under a baseline -march, but the host
+#            must support AVX2+FMA at run time. Signing is sampler-bound, so this
+#            is a modest (~1.1x) end-to-end speedup.
+#   neon   - AArch64 NEON (2-wide float64x2_t + FMA) vectorized FFT; implies
+#            double. Advanced SIMD is part of the ARMv8-A baseline, so no special
+#            -march is required.
 AC_ARG_ENABLE([falcon],
-    [AS_HELP_STRING([--enable-falcon],[Enable Falcon post-quantum signatures via liboqs (default: disabled)])],
+    [AS_HELP_STRING([--enable-falcon@<:@=OPTS@:>@],[Enable Falcon post-quantum signatures (native, no liboqs; requires --enable-experimental). OPTS is a comma-separated list of fpr/FFT backends: asm, double, avx2, neon (default: disabled)])],
```
Frauschi: **🔵 [Low] configure.ac --enable-falcon help omits the accepted 'small-mem' sub-option**
*🔧 NIT*  `style`

The `--enable-falcon=OPTS` help string lists only `asm, double, avx2, neon`, but the option parser accepts and acts on `small-mem` (`ENABLED_FALCON_SMALL_MEM=yes`), and the unknown-option error message even advertises it (`Valid options: asm, double, avx2, neon, small-mem`). The help text is inconsistent with the actual accepted set.

**Suggestion:** Add `small-mem` to the OPTS list in the AS_HELP_STRING (or document it as a memory-tradeoff sub-option).

wolfcrypt/src/falcon.c
Frauschi, line 8456: **⚪ [Info] Native check_key uses hardware modulo despite division-free NTT elsewhere**
*🔧 NIT*  `style`

The PR justifies a division-free Barrett verify NTT so embedded/Cortex-M targets need no hardware divide, but `falcon_native_check_key` uses `((word64)h[i] * ft[i]) % FALCON_Q` in its comparison loop. This is only compiled with native sign (not the pure verify-only embedded path), so it is not a correctness or portability problem, just a minor inconsistency with the division-free rationale stated for the verify path.

There is also a mild **constant-time** angle: `ft[i]` here is the NTT image of the secret polynomial `f`, so `% FALCON_Q` runs a hardware divide on secret-derived data. `falcon_barrett()` is documented as constant-time and bit-identical to `mod q`, whereas integer `%` timing is data-dependent on some cores (e.g. Cortex-M). `check_key` is not a hot path so the exposure is low, but reusing the helper matches wolfSSL's constant-time-by-default posture for secret data.

The verify path already does exactly this for its pointwise multiply (`t[i] = (word16)falcon_barrett((word32)t[i] * h[i]);`). Both `h[i]` and `ft[i]` are post-NTT values in `[0, q)`, so the product is in `[0, q^2)` — `falcon_barrett`'s documented input domain — and fits in `word32`, making the change safe:

```c
    for (i = 0; i < n; i++) {
        if (ft[i] == 0 ||
                (word16)falcon_barrett((word32)h[i] * ft[i]) != gt[i]) {
            ret = PUBLIC_KEY_E;
            break;
        }
    }
```

**Suggestion:** Reuse `falcon_barrett()` here as the verify path does — it drops the hardware divide and is constant-time on the secret-derived `ft` (the NTT of secret `f`). Consistency/robustness cleanup, no behavioral change.

Frauschi (dismissed):
LGTM

SparkiDev (dismissed):

tests/api/test_falcon.c
```diff
@@ -0,0 +1,791 @@
...
+    XMEMSET(&key, 0, sizeof(key));
+    ExpectIntEQ(wc_falcon_init(&key), 0);
+    ExpectIntEQ(wc_falcon_set_level(&key, FALCON_LEVEL1), 0);
+    ExpectIntEQ(wc_falcon_export_public(&key, NULL, &outLen),
+        WC_NO_ERR_TRACE(BAD_FUNC_ARG));
+    ExpectIntEQ(wc_falcon_export_public(&key, out, NULL),
+        WC_NO_ERR_TRACE(BAD_FUNC_ARG));
+    /* level set but no public key -> BAD_FUNC_ARG. */
+    outLen = (word32)sizeof(out);
+    ExpectIntEQ(wc_falcon_export_public(&key, out, &outLen),
```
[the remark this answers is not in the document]
danielinux: Fixed in d21ce34

danielinux:
jenkins retest this please

SparkiDev (approved):

PR #10836, opened 2026-07-02
Fix PKCS#7 padding for block-aligned input in wc_EncryptPKCS8Key_ex

yosuke-wolfssl:
## Description

`wc_EncryptPKCS8Key_ex` computed the PKCS#7 pad length with an extra
`(blockSz - 1)` mask:

```c
padSz = (word32)((blockSz - ((int)keySz & (blockSz - 1))) & (blockSz - 1));
```

When the plaintext length is an exact multiple of the cipher block size,
`keySz & (blockSz - 1) == 0`, so the inner term is `blockSz` and the trailing
mask collapses it to `0` — no padding is emitted. PKCS#7 (and CBC as OpenSSL
implements it) requires a pad length of `1..blockSz`, i.e. a **whole extra
block** when the input is already block-aligned.

The fix removes the mask and guards the computation on `blockSz > 1` so stream
ciphers (RC4, `blockSz == 1`) still receive no padding, matching the sibling
`wc_PkcsPad` helper and PKCS#7.

## Impact

Any private key whose plaintext PKCS#8 DER length is an exact multiple of the
cipher block size produced an `EncryptedPrivateKeyInfo` that non-wolfSSL parsers
reject (OpenSSL: `bad decrypt`). Examples with AES (16-byte block):

- **X25519 / Ed25519** — 48-byte PKCS#8 (a multiple of 16) → 0 padding → broken.
- **RSA / EC / DH** in common sizes are usually not block-aligned, so they padded
  correctly and worked, which is why this went unnoticed.

wolfCrypt's own `wc_DecryptPKCS8Key` strips padding by re-reading the inner DER
`SEQUENCE` length rather than validating PKCS#7, so wolfSSL↔wolfSSL round-trips
succeeded and the bug hid in self-tests — only wolfSSL↔OpenSSL interop exposed it.

## Changes

- `wolfcrypt/src/asn.c` — correct the `padSz` computation in
  `wc_EncryptPKCS8Key_ex`; gate padding on `blockSz > 1`.
- `tests/api.c` — add `enc_pkcs8_pad_check` helper and three cases covering
  block-aligned input at different block sizes plus the stream-cipher path:
  - PBES2 / AES-256-CBC (blockSz 16) — full 16-byte pad block
  - PBES1 / SHA1-DES (blockSz 8) — full 8-byte pad block, exercises the PBES1
    encoder branch
  - PKCS#12 / RC4 (blockSz 1) — no padding, exercises the `blockSz > 1` guard

  Each asserts the encrypted `OCTET STRING` length directly, since a decrypt
  round-trip alone false-passes on a missing pad block.

## Notes

- No public API or signature changes; decrypt side needs no change (correctly
  padded output already decrypts, and previously-written blobs still read).
- `blockSz` is always `1`, `8`, or `16` where `padSz` is computed
  (`CheckAlgo`/`GetAlgoV2` set it on every success path), so the mask arithmetic
  is well-defined.

## Testing

Built with `--enable-des3 --enable-arc4 --enable-aescbc` under ASan+UBSan
(macOS, no LeakSanitizer); all four PKCS#8 API tests pass, full API suite clean
(0 failures). Negative control confirmed: reverting the fix fails the
block-aligned length assertion at exactly the padding check.

### DIFF
--- a/tests/api.c
+++ b/tests/api.c
@@ -79,7 +79,9 @@
     #include <wolfssl/wolfcrypt/signature.h>
 #endif
 
-#ifdef WOLFSSL_SMALL_CERT_VERIFY
+#if defined(WOLFSSL_SMALL_CERT_VERIFY) || \
+    (defined(HAVE_PKCS8) && !defined(NO_ASN))
+    /* for ASN tag and PBE/PKCS enum values used by the PKCS#8 encrypt tests */
     #include <wolfssl/wolfcrypt/asn.h>
 #endif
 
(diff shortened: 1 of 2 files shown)


tests/api.c
```diff
@@ -21830,6 +21830,107 @@ static int test_wc_CreateEncryptedPKCS8Key(void)
...
+
+    /* The encrypted content is the trailing OCTET STRING, extending to the end
+     * of the buffer. Its length is plaintext + expected pad (short-form since
+     * it is < 128): a full block for a block cipher, none for a stream cipher. */
+    expEncLen = (word32)sizeof(plain) + expExtra;
+    ExpectIntGE(encKeySz, expEncLen + 2);
+    if (encKeySz >= expEncLen + 2) {
+        ExpectIntEQ(encKey[encKeySz - expEncLen - 2], ASN_OCTET_STRING);
+        ExpectIntEQ(encKey[encKeySz - expEncLen - 1], (byte)expEncLen);
+    }
```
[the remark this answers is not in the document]
yosuke-wolfssl: Fixed

tests/api.c
```diff
@@ -21830,6 +21835,108 @@ static int test_wc_CreateEncryptedPKCS8Key(void)
...
+        &encKeySz, password, (int)passwordSz, vPKCS, pbeOid, encAlgId, NULL, 0,
+        WC_PKCS12_ITT_DEFAULT, &rng, NULL), 0);
+
+    /* The encrypted content is the trailing OCTET STRING, extending to the end
+     * of the buffer. Its length is plaintext + expected pad (short-form since
+     * it is < 128): a full block for a block cipher, none for a stream cipher. */
+    expEncLen = (word32)sizeof(plain) + expExtra;
+    ExpectIntGE(encKeySz, expEncLen + 2);
+    if (EXPECT_SUCCESS() && (encKey != NULL) && (encKeySz >= expEncLen + 2)) {
+        ExpectIntEQ(encKey[encKeySz - expEncLen - 2], ASN_OCTET_STRING);
```
Frauschi, line 23208: **🟡 [Medium] Length check assumes short-form DER length**
*💡 SUGGEST*  `test`

This reads `encKey[encKeySz - expEncLen - 1]` as a single short-form length byte and compares it to `(byte)expEncLen`. That only holds while the encrypted content is < 128 bytes (short-form DER length), which is true for the current fixtures (48/56/64 bytes).

If this helper is later reused for larger keys — e.g. ML-DSA private keys, which run to several KB — the OCTET STRING length switches to long-form (`0x82 | len_hi | len_lo`). Both reads then land on the wrong bytes: the real tag sits two bytes earlier, and the compared value is a truncated 8-bit slice of a multi-KB length. The helper would silently false-fail (or false-pass).

Optional hardening: guard with `expEncLen < 128` so a future larger-plaintext caller fails loudly rather than mis-parsing the length field. Longer term, if PKCS#8-encrypt coverage is added for ML-DSA keys, the helper would need to decode long-form lengths properly.

tests/api.c
```diff
@@ -80,6 +80,11 @@
     #include <wolfssl/wolfcrypt/asn.h>
 #endif
 
+#if defined(HAVE_PKCS8) && !defined(NO_ASN)
+    /* for ASN tag and PBE/PKCS enum values used by the PKCS#8 encrypt tests */
+    #include <wolfssl/wolfcrypt/asn.h>
```
Frauschi, line 85: **🔵 [Low] Redundant asn.h include**
*🔧 NIT*  `convention`

This `#include <wolfssl/wolfcrypt/asn.h>` under `#if defined(HAVE_PKCS8) && !defined(NO_ASN)` duplicates the conditional include at lines 79–81 (under `WOLFSSL_SMALL_CERT_VERIFY`). asn.h has include guards so it's harmless, and the new guard does correctly ensure the ASN tag / PBE enum symbols are available when the small-cert path isn't compiled.

Optional: fold this into the existing include block to avoid the duplication, or leave as-is.

Frauschi (commented):
## 🐺 Skoll Code Review

### Posted findings
- [Medium] **Length check assumes short-form DER length** — `tests/api.c:21883`
- [Low] **Redundant asn.h include** — `tests/api.c:85`

tests/api.c
yosuke-wolfssl: Thanks for review.
I added the following guards with updated comments:
```
    expEncLen = (word32)sizeof(plain) + expExtra;
    ExpectIntLT(expEncLen, 128);
    ExpectIntGE(encKeySz, expEncLen + 2);

```

tests/api.c
yosuke-wolfssl: Thanks again.
I folded the two asn.h include sites into one:
```
#if defined(WOLFSSL_SMALL_CERT_VERIFY) || \
    (defined(HAVE_PKCS8) && !defined(NO_ASN))
    /* for ASN tag and PBE/PKCS enum values used by the PKCS#8 encrypt tests */
    #include <wolfssl/wolfcrypt/asn.h>
#endif
```

Frauschi:
Jenkins retest this please

Frauschi (approved):

yosuke-wolfssl:
Hi @Frauschi ,
Thanks for reviewing. All tests are passed now.
Can you merge this please ?

PR #10837, opened 2026-07-02
DNS name constraint fix

rlm2002:
# Description

A certificate with no dNSName SAN but another SAN type present (e.g. registeredID or iPAddress) bypassed the Subject CN dNSName name-constraint check. The CN-as-DNS fallback was gated on `cert->subjectCN != NULL && cert->altNames == NULL && !cert->isCA` instead of "no dNSName SAN", so an out-of-scope CN was accepted. The fallback now fires whenever no dNSName SAN is present, (RFC 6125 6.4.4 ). A non-dNSName SAN no longer suppresses the check.

Fixes zd#22078

# Testing

Adds additional cases to `test_wolfSSL_CertManagerNameConstraint_DNS_CN()`

# Checklist

 - [x] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/tests/api/test_certman.c
+++ b/tests/api/test_certman.c
@@ -1591,16 +1591,20 @@ int test_wolfSSL_CertManagerNameConstraint_DNS_CN(void)
     !defined(NO_WOLFSSL_CM_VERIFY) && !defined(NO_RSA) && \
     defined(OPENSSL_EXTRA) && defined(WOLFSSL_CERT_GEN) && \
     defined(WOLFSSL_CERT_EXT) && defined(WOLFSSL_ALT_NAMES) && \
-    !defined(NO_SHA256)
-    /* Test that DNS name constraints are enforced against the Subject CN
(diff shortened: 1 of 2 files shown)


Pig-Tail:
Thanks for putting this together, @rlm2002 — I verified the branch against my reproducer and the out-of-scope CN is now correctly rejected when only a non-dNSName SAN (registeredID/iPAddress) is present. The dnsSan == NULL gate is exactly right. 🙌

 Since this restores the protection from CVE-2026-6731 for the RID/iPAddress SAN path, would it be worth tracking it with its own CVE/advisory once the fix lands in a release? It'd help downstream integrators that rely on DNS name constraints for sub-CA containment pick it up. No urgency — happy to review advisory text or provide anything else that's useful. 

 Thanks again for the quick fix.

Frauschi (approved):

dgarske (approved):

PR #10841, opened 2026-07-03
internal.c: extract functions to make code cleaner

SparkiDev:
# Description

Extracted to functions pieces of code from very long functions.

# Testing

enable-all

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -3666,58 +3666,16 @@ int AllocateSuites(WOLFSSL* ssl)
     return 0;
 }
 
-void InitSuites(Suites* suites, ProtocolVersion pv, int keySz, word16 haveRSA,
-                word16 havePSK, word16 haveDH, word16 haveECDSAsig,
-                word16 haveECC, word16 haveStaticRSA, word16 haveStaticECC,
-                word16 haveAnon, word16 haveNull, word16 haveAES128,
-                word16 haveSHA1, word16 haveRC4, int side)
-{
-    word16 idx = 0;
-    int    tls    = pv.major == SSLv3_MAJOR && pv.minor >= TLSv1_MINOR;
-    int    tls1_2 = pv.major == SSLv3_MAJOR && pv.minor >= TLSv1_2_MINOR;
-    int    tls1_3 = IsAtLeastTLSv1_3(pv);
-    int    dtls   = 0;
-    int    haveRSAsig = 1;
-
-#ifdef WOLFSSL_DTLS
-    if (pv.major == DTLS_MAJOR) {
-        dtls   = 1;
-        tls    = 1;
-        /* May be dead assignments dependent upon configuration */
-        (void) dtls;
-        (void) tls;
-        tls1_2 = pv.minor <= DTLSv1_2_MINOR;
-    }
-#endif /* WOLFSSL_DTLS */
-
-    (void)tls;  /* shut up compiler */
-    (void)tls1_2;
-    (void)tls1_3;
-    (void)dtls;
-    (void)haveDH;
-    (void)havePSK;
-    (void)haveStaticRSA;
-    (void)haveStaticECC;
-    (void)haveECC;
-    (void)haveECDSAsig;
-    (void)side;
-    (void)haveRSA;    /* some builds won't read */
-    (void)haveRSAsig; /* non ecc builds won't read */
-    (void)haveAnon;   /* anon ciphers optional */
-    (void)haveNull;
-    (void)haveAES128;
-    (void)haveSHA1;
-    (void)haveRC4;
-
-    if (suites == NULL) {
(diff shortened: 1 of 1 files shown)


src/internal.c
```diff
@@ -14360,112 +14678,14 @@ int CopyDecodedToX509(WOLFSSL_X509* x509, DecodedCert* dCert)
...
-        if (x509->authInfoCaIssuer != NULL) {
-            XMEMCPY(x509->authInfoCaIssuer, dCert->extAuthInfoCaIssuer, dCert->extAuthInfoCaIssuerSz);
-            x509->authInfoCaIssuerSz = dCert->extAuthInfoCaIssuerSz;
-        }
-        else {
-            ret = MEMORY_E;
-        }
-    }
-    #endif
+    if (ret == 0)
```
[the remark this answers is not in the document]
SparkiDev: fixed

src/internal.c
```diff
@@ -7171,57 +7131,238 @@ int SetSSL_CTX(WOLFSSL* ssl, WOLFSSL_CTX* ctx, int writeDup)
...
-    ssl->CBIS         = ctx->CBIS;
-#endif
-    ssl->timeout = ctx->timeout;
-    ssl->verifyCallback    = ctx->verifyCallback;
-    /* If we are setting the ctx on an already initialized SSL object
-     * then we possibly already have a side defined. Don't overwrite unless
-     * the context has a well defined role. */
-    if (newSSL || ctx->method->side != WOLFSSL_NEITHER_END) {
-        ssl->options.side  = (word16)(ctx->method->side);
+#ifndef NO_CERTS
```
dgarske, line 7139: **⚪ [Info] New block scopes introduced solely to declare helper-return variables**

The refactor introduces bare `{ int certRet = ...; }` and `{ int copyRet = ...; }` scopes purely to hold a helper's return value. The project C89 style rule states "No new `{ ... }` scope just to declare." That said, this deviation is functionally justified: in `SetSSL_CTX` a separate `certRet` is required because assigning the helper's `0`-on-success result into the function-level `ret` would clobber the `WOLFSSL_SUCCESS` value that must be returned on the `writeDup != 0` path; in `CopyDecodedToX509` the `copyRet` pattern preserves error-code priority (only overwrite `ret` when the new op fails). So the choice is correct, just stylistically at odds with the stated convention.

**Fix:** Acceptable as-is given the semantics. Optionally hoist the temp declaration to the top of the enclosing function/block to match the C89 convention, but do NOT reuse the function-level `ret` in SetSSL_CTX (that would change the success return value).

src/internal.c
```diff
@@ -16672,52 +17492,9 @@ int ProcessPeerCerts(WOLFSSL* ssl, byte* input, word32* inOutIdx,
                 #endif
 #if defined(HAVE_CERTIFICATE_STATUS_REQUEST_V2)
                     if (ret == 0 && addToPendingCAs && !alreadySigner) {
```
dgarske, line 17488: **⚪ [Info] skipAddCA now set unconditionally before pending-CA helper call**

In the original inline code, `skipAddCA = 1;` was set only after `ParseCert`/`AllocDer`/`MakeSigner`/`FillSigner` all succeeded (just before `TLSX_CSR2_AddPendingSigner`). The refactor moves `skipAddCA = 1;` to run unconditionally before `ProcessPeerCertAddPendingCA()`. This is behaviorally equivalent because `skipAddCA` is only consulted later under an `if (ret == 0 && ... && !skipAddCA)` guard (line 17504-17505), and any failure inside the helper returns non-zero and immediately `goto exit_ppc`, so the differing `skipAddCA` value on the error path is never read. Flagging only so the intent is documented.

**Fix:** No change required; behavior is preserved. Optionally add a brief comment noting skipAddCA only matters on the ret==0 continuation path.

dgarske (commented):
## Skoll Code Review

### Posted findings
- [Info] **New block scopes introduced solely to declare helper-return variables** — `src/internal.c:7445-7449, 14615-14621, 14704-14708, 14713-14717, 14770-14774`
- [Info] **skipAddCA now set unconditionally before pending-CA helper call** — `src/internal.c:17494-17499`

dgarske (dismissed):
This refactor looks great. There is a few empty braces in the refactor. I did a function by function review and the refactor looks good.

src/internal.c
SparkiDev: Added a comment at the ProcessPeerCertAddPendingCA call site (no behavior change), capturing this.

src/internal.c
SparkiDev: Hoisted the block-scope temps (C89 "no new scope just to declare")

dgarske (approved):

SparkiDev:
Jenkins: retest this please

dgarske (approved):

PR #10845, opened 2026-07-06
Infrastructure for Mc/Dc test coverage + AES coverage campaign

danielinux:
# Description

This PR is test-only, additive infrastructure for structural (Modified Condition / Decision Coverage - MC/DC) test coverage - no library source or behavior changes.

Two parts:

1. Reusable MC/DC white-box harness — tests/unit-mcdc/ (new).
A standalone pattern for exercising decision conditions that are structurally unreachable through the public API: static / WOLFSSL_LOCAL helpers whose "impossible" operand combinations every wc_* entry point rejects first (e.g. a size != 0 argument paired with a NULL pointer). The program #includes the module .c directly so those helpers are in scope, and calls them with both halves of each MC/DC independence pair in one binary. It is not part of the wolfSSL build and not registered in tests/api — it's an opt-in supplement for coverage campaigns. The first instance, test_aes_whitebox.c, reaches the AES GHASH / GHASH_UPDATE pointer guards and _AesNew_common cross-argument checks; README.md documents the mechanism and how to add modules.

2. AES coverage campaign + seed coverage for adjacent modules.
A large batch of decision-, feature-, and argument-check tests, registered in the existing test groups so they run in normal CI:
- AES (groups aes, aes-eax, aes-siv; also exercises gmac/cmac): 14 new test_wc_Aes* functions covering key-wrap vectors, GCM/GMAC/CCM/XTS/CMAC/SIV/EAX and key-export/import argument checks, mode decision paths, and feature (streaming/multi-chunk) paths.
- Other modules touched by the same campaign: test_asn (*DecisionCoverage/*FeatureCoverage), test_rsa (same), test_ocsp, test_pkcs7 (InitWithCert guardrails), test_pkcs12 (create/parse guardrails), test_signature (falcon sign/verify), and OpenSSL-compat X.509 (test_ossl_x509, _ext name-constraints manual paths, _vp, test_certman).

All new tests/api tests are guarded by the same #if as the code under test, so they auto-report TEST_SKIPPED when the feature is compiled out.

Testing

- Functional: built with --enable-all and ran ./tests/unit.test (and the individual --group aes, aes-eax, aes-siv, gmac, cmac, plus asn, rsa, ocsp, pkcs7, pkcs12, signature, x509 groups); all new tests pass. Verified they compile out cleanly (report TEST_SKIPPED) in reduced configs where the feature is disabled.
- MC/DC measurement (clang source-based, -fcoverage-mcdc + llvm-cov): ran each module through per-user_settings.h variant builds and took the union of covered decisions across variants (necessary because a file selects mutually-exclusive implementations at compile time — e.g. the five GHASH backends). For AES this reaches 375/379 (98.94%) union MC/DC on wolfcrypt/src/aes.c; the remaining 4 are documented structural residuals (2 complementary-operand decisions where unique-cause MC/DC is unsatisfiable, 1 ret==0 internal-failure path, 1 dead loop-index guard).
- White-box harness: validated on clang 21 — compiles wolfcrypt/src/aes.c directly with the library's exact flags and links against libwolfssl.a with aes.o removed, so the instrumented TU is the single definition; flips exactly the 19 API-unreachable conditions.
- Measurement + a regression gate are automated by a companion Jenkins job (separate PR in wolfssl/testing); this PR only adds the tests.

```
./tests/unit.test -~ArgMcdc -~DecisionCoverage -~FeatureCoverage
```

Checklist

- [x] added tests
- [ ] updated/added doxygen — n/a (no public API added)
- [x] updated appropriate READMEs — added tests/unit-mcdc/README.md
- [ ] Updated manual and documentation — n/a (internal test infrastructure)


tests/api/test_aes.c
```diff
@@ -7777,179 +7783,2251 @@ int test_wc_AesEaxStream(void)
...
+        ExpectIntEQ(wc_AesCtrEncrypt(&aes, out, in, 5), 0);
+
+        /* Corrupted rounds + sz >= block size: on this platform the full
+         * blocks are consumed by the AES-NI batch path (which does not
+         * consult wc_AesEncrypt()'s rounds check), so this still lands on
+         * the "ret == 0 && sz != 0" (true) leftover-handling call, which
+         * itself then fails via the corrupted rounds. */
+        aes.rounds = 0;
+        ExpectIntEQ(wc_AesCtrEncrypt(&aes, out, in, 32),
+            WC_NO_ERR_TRACE(KEYUSAGE_E));
```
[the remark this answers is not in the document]
danielinux: fixed in 840f3ad

danielinux:
Jenkins retest this please

tests/api/test_signature.c
```diff
@@ -161,3 +168,46 @@ int test_wc_SignatureGetSize_rsa(void)
     return EXPECT_RESULT();
 } /* END test_wc_SignatureGetSize_rsa(void) */
 
+int test_wc_falcon_sign_verify(void)
+{
+    EXPECT_DECLS;
+#if defined(HAVE_PQC) && defined(HAVE_FALCON) && defined(HAVE_LIBOQS)
```
dgarske: **🟠 [Medium] Falcon coverage test is dead code — guarded on HAVE_PQC which is never defined**

The new Falcon sign/verify test (and its `falcon.h`/`oqs.h` includes) is wrapped in `#if defined(HAVE_PQC) && defined(HAVE_FALCON) && defined(HAVE_LIBOQS)`. `HAVE_PQC` is never defined anywhere in the build: `--enable-falcon` sets `-DHAVE_FALCON` (configure.ac:1864) and liboqs sets `HAVE_LIBOQS`, but nothing defines `HAVE_PQC` (grep across configure.ac, m4/, settings.h, options.h finds zero definitions — only doc comments and this test reference it). The library itself guards Falcon on `HAVE_FALCON` alone (wolfssl/wolfcrypt/falcon.h:38). Consequently this test always compiles out to TEST_SKIPPED and can NEVER execute, even with `--enable-all` or `--enable-falcon --with-liboqs`. This defeats the entire purpose of adding it in a coverage-campaign PR. The companion change adding `HAVE_PQC` to `.wolfssl_known_macro_extras` masks the underlying problem (it silences the macro-consistency checker for a macro that is referenced but never defined) rather than fixing it.

**Fix:** Change both guards to `defined(HAVE_FALCON) && defined(HAVE_LIBOQS)` (matching the library's own guard and how existing Falcon code is gated) so the test can actually run, and revert the `HAVE_PQC` addition to `.wolfssl_known_macro_extras` since it should no longer be referenced.

tests/api/test_pkcs12.c
```diff
@@ -199,6 +199,72 @@ int test_wc_PKCS12_create(void)
...
+#if !defined(NO_ASN) && !defined(NO_PWDBASED) && defined(HAVE_PKCS12)
+    WC_PKCS12* pkcs12 = NULL;
+    byte* outKey = NULL;
+    byte* outCert = NULL;
+    WC_DerCertList* outCa = (WC_DerCertList*)1;
+    word32 outKeySz = 0;
+    word32 outCertSz = 0;
+
+    ExpectIntEQ(wc_PKCS12_parse(NULL, "", &outKey, &outKeySz, &outCert,
+        &outCertSz, &outCa), BAD_FUNC_ARG);
```
dgarske: **🔵 [Low] New PKCS12 parse guardrail asserts use bare BAD_FUNC_ARG instead of WC_NO_ERR_TRACE()**

The new negative-path assertions compare the return value against bare `BAD_FUNC_ARG` (e.g. `ExpectIntEQ(wc_PKCS12_parse(...), BAD_FUNC_ARG)`), whereas most other new tests in this PR use `WC_NO_ERR_TRACE(BAD_FUNC_ARG)` to strip the error-trace wrapper from the expected value. Functionally the comparison still yields the correct integer, and this file already contains pre-existing bare usages (lines 719, 721), so it is consistent-with-file. Under `WOLFSSL_DEBUG_TRACE_ERROR_CODES` the bare form can emit spurious trace output when the RHS is evaluated. Minor consistency point only.

**Fix:** Optionally wrap the expected error codes in `WC_NO_ERR_TRACE(...)` for consistency with the rest of the PR; not required since it matches existing file style and is functionally correct.

dgarske (commented):
## Skoll Code Review

### Posted findings
- [Medium] **Falcon coverage test is dead code — guarded on HAVE_PQC which is never defined** — `tests/api/test_signature.c:36,174`
- [Low] **New PKCS12 parse guardrail asserts use bare BAD_FUNC_ARG instead of WC_NO_ERR_TRACE()** — `tests/api/test_pkcs12.c:239-258`

#### OPENSSL_EXTRA guard widening on TXT_DB / NCONF tests is ineffective (dead) due to outer OPENSSL_ALL wrapper
**File:** `tests/api.c:21464,21547`
**Function:** `test_wolfSSL_TXT_DB / test_wolfSSL_NCONF_negative_paths`
**Severity:** Low

The inner `#if` for `test_wolfSSL_TXT_DB` and the new `test_wolfSSL_NCONF_negative_paths` were widened from `OPENSSL_ALL` to `(OPENSSL_EXTRA || OPENSSL_ALL)`, apparently to broaden where they run. However both function definitions sit inside the enclosing `#if defined(OPENSSL_ALL)` block (api.c:21150 ... `#endif /* OPENSSL_ALL */` at 21583), and their TEST_DECL registrations are also under `#ifdef OPENSSL_ALL` (api.c:36469-36473). So under an `OPENSSL_EXTRA`-only build (no `OPENSSL_ALL`) these functions are never compiled and never registered — the widened inner guard cannot take effect. Additionally, `test_wolfSSL_NCONF_negative_paths` calls `NCONF_load`/`NCONF_get_*`, which are themselves compiled only under `OPENSSL_ALL` (src/conf.c line 36 region), so widening to `OPENSSL_EXTRA` would not even link if the outer wrapper were removed. The change is harmless (no build break) but misleading — the intended coverage broadening is not realized. Contrast with the CONF_CTX and ossl_x509_ext widenings in this same PR, which are effective because those APIs are genuinely available under `OPENSSL_EXTRA`.

**Recommendation:** Either revert these two inner-guard widenings to keep them plainly `OPENSSL_ALL`-scoped (matching the surrounding block and their OPENSSL_ALL-only APIs/registration), or, if the goal is to run TXT_DB/CONF under OPENSSL_EXTRA, move the function definitions and registrations out of the outer `#ifdef OPENSSL_ALL` and confirm the underlying APIs exist under OPENSSL_EXTRA. As written the change has no observable effect.

_Referenced code: `tests/api.c:21464,21547` (3 lines)_

---

dgarske (changes requested):
https://github.com/wolfSSL/wolfssl/pull/10845#pullrequestreview-4655989742

tests/api/test_signature.c
```diff
@@ -31,7 +31,14 @@
 #include <wolfssl/wolfcrypt/signature.h>
 #include <wolfssl/wolfcrypt/rsa.h>
 #include <wolfssl/wolfcrypt/ecc.h>
+#include <wolfssl/wolfcrypt/random.h>
 #include <wolfssl/wolfcrypt/types.h>
+#if defined(HAVE_PQC) && defined(HAVE_FALCON)
```
dgarske: Check `HAVE_PQC`. Doesn't seem to be a real macro. Consider cleaning up in this PR.

dgarske (changes requested):

dgarske (approved):

PR #10853, opened 2026-07-06
Add cross-library compile-testing for wolfSSL products

night1rider:
For each product (wolfSSH, wolfCLU, wolfTPM, wolfMQTT, wolfPKCS11, wolfProvider) this builds wolfSSL and then builds the product against it, at both the product's latest release tag and its master branch. It only checks that they compile, it does not run any tests. If a wolfSSL change breaks a product's latest release on purpose, you say so in a commit message with breaks-<product>=<tag>. A break on a product's master is not allowed and has to be fixed.

breaks-wolfssh=v1.5.0-stable

Note: wolfSSH v1.5.0-stable does not currently compile against wolfSSL. This commit did not break it, it adds the cross-library check that discovered the break. The token above declares it so the new check tracks it as a known break instead of failing red, until wolfSSH ships a fixed release.

### DIFF
--- a/.github/workflows/cross-library.yml
+++ b/.github/workflows/cross-library.yml
@@ -1,0 +1,279 @@
+name: cross-library (reusable)
+
+# Reusable engine for compile-testing a downstream wolfSSL product against the
+# wolfSSL in this checkout. It builds wolfSSL once from this checkout (the PR
+# merge commit) with the flags a product needs, then compiles the product
+# against that install at both the product's default-branch HEAD and its highest
+# release tag (compile-only, no `make check`).
+#
+# The wolfSSL build runs once and is shared: the build-wolfssl job installs
+# wolfSSL and uploads it as an artifact, and the compile matrix (head, latest)
+# downloads that artifact instead of rebuilding. This halves the wolfSSL builds
+# per product from two to one.
+
+on:
+  workflow_call:
+    inputs:
+      product:
+        description: 'Short product label (used for the job name)'
+        required: true
+        type: string
+      repo:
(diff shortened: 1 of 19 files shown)


night1rider:
Jenkins retest this please

night1rider:
Jenkins retest this please

dgarske (approved):

dgarske (changes requested):
This is great, but can you consider the following two items for optimizing the runner time/usage?

A. Shallow checkout for head legs (trivial, ~6 min saved, zero risk). Make depth conditional in cross-library.yml:
fetch-depth: ${{ inputs.ref_mode == 'latest' && 0 || 1 }}
fetch-tags:  ${{ inputs.ref_mode == 'latest' }}
Keep full history on latest — check-break.sh fails loud (never fails open) if history is missing, so we must not starve it. Safe because head legs demonstrably never call it.

B. Build wolfSSL once per distinct config and share it (bigger win, moderate effort, ~6-8 min saved). Restructure so wolfSSL is built in a prerequisite job keyed on the wolfssl_configure set + wolfSSL SHA, uploaded as an artifact, and both compile legs download it. Dedupes 12 builds -> ~5. This is the largest structural saving but a real rewrite of the reusable workflow (adds artifact upload/download, ~10-20s each — still net positive). A lighter-touch variant is actions/cache keyed on hash(configure)+SHA, though within a single PR run the head/latest legs start together and both miss the cache, so a shared build job dedupes intra-run more reliably than cache does.

dgarske (approved):

PR #10858, opened 2026-07-07
C89 compliance items

anhu:
Fixes ZD 22098

- Flexible array members ("type name[];") fixup
- avoid empty asn_orig.c when building standalone
- avoid trailing comma on the last enumerator
- drop some consts to match function prototypes


src/internal.c
```diff
@@ -29434,7 +29434,7 @@ int SetCipherBits(const char* enc) {
 }
 #endif /* WOLFSSL_QT || OPENSSL_ALL */
 
-const char* GetCipherNameIana(const byte cipherSuite0, const byte cipherSuite)
+const char* GetCipherNameIana(byte cipherSuite0, byte cipherSuite)
```
douzzer: The `const` attributes on the args need to stay.  These are all `WOLFSSL_LOCAL`, and adding `const` to API args would be harmless and binary-compatible anyway.  _Removing_ them is a different story.

douzzer (changes requested):

dgarske (dismissed):

wolfssl/wolfcrypt/types.h
```diff
@@ -221,6 +221,15 @@ typedef const char wcchar[];
...
 #endif
 
+#if defined(WOLF_C89)
+    /* C99 flexible array member, or the C89 "struct hack" fallback.
+     * See http://c-faq.com/struct/structhack.html */
+    #define WC_FLEXIBLE_ARRAY_MEMBER 1
+#else
+    /* empty: C99 flexible array member */
+    #define WC_FLEXIBLE_ARRAY_MEMBER
+#endif
```
douzzer: This needs to be named `WC_FLEXIBLE_ARRAY_SIZE`.  It also needs override capability.  The setup for it in WolfSentry is
```
#if defined(WOLFSENTRY_FLEXIBLE_ARRAY_SIZE)
    /* keep override value. */
#elif defined(__STRICT_ANSI__) || defined(WOLFSENTRY_PEDANTIC_C)
    #define WOLFSENTRY_FLEXIBLE_ARRAY_SIZE 1
#elif defined(__GNUC__) && !defined(__clang__)
    #define WOLFSENTRY_FLEXIBLE_ARRAY_SIZE
    /*!< \brief Value appropriate as a size for an array that will be allocated to a variable size.  Built-in value usually works. */
#else
    #define WOLFSENTRY_FLEXIBLE_ARRAY_SIZE 0
#endif
```

and that's my recommendation here too, with `|| defined(WOLF_C89)` added to the `__STRICT_ANSI__` condition.

douzzer (changes requested):

douzzer (dismissed):
Note a whitespace flub in the reindented comments in the `struct ExternalTicket` definition -- but def not worth holding up the PR for it.

anhu:
Fixed the whitespace.

douzzer (approved):

wolfssl/internal.h
```diff
@@ -5455,6 +5461,7 @@ typedef enum {
     STACK_TYPE_X509_REQ_ATTR      = 18,
     STACK_TYPE_GENERAL_SUBTREE    = 19,
     STACK_TYPE_X509_REVOKED       = 20,
+    WOLF_ENUM_DUMMY_LAST_ELEMENT(WOLF_STACK_TYPE)
```
douzzer: Yikes, nearly all of these `WOLF_ENUM_DUMMY_LAST_ELEMENT()` uses are frivolous and need to be removed.

The C89 syntax issues can all be resolved by simply removing the dangling comma on the last item in the `enum`s.

One of the `WOLF_ENUM_DUMMY_LAST_ELEMENT()` s, the one added here, is actually breaking:
```
./src/ssl_sk.c:129:13: error: enumeration value '_wolf_WOLF_STACK_TYPE_L5468_enum_dummy_last_element' not explicitly handled in switch [clang-diagnostic-switch-enum]
129 |     switch (node->type) {
|             ^~~~~~~~~~
```
(and many more like it)

`WOLF_ENUM_DUMMY_LAST_ELEMENT()` is only for use when the last element in an `enum` is conditionally present, so that a trick is needed to assure a proper trailing-comma-free final item.

Even then, it is inherently impossible to use `WOLF_ENUM_DUMMY_LAST_ELEMENT()` in an `enum` that will be the subject of a `switch()`, because the name is computed dynamically at build time.

wolfssl/internal.h
```diff
@@ -6786,6 +6793,7 @@ enum ContentType {
 #ifdef WOLFSSL_DTLS13
     ack                = 26,
 #endif /* WOLFSSL_DTLS13 */
+    WOLF_ENUM_DUMMY_LAST_ELEMENT(ContentType)
```
douzzer, line 6790: This is the only `WOLF_ENUM_DUMMY_LAST_ELEMENT()` in the PR that's actually needed.  It isn't currently used in any `switch()`es so you can go ahead and keep it this way.

douzzer (changes requested):

wolfssl/internal.h
```diff
@@ -2090,7 +2090,7 @@ enum states {
 #ifdef WOLFSSL_DTLS13
     SERVER_FINISHED_ACKED,
 #endif /* WOLFSSL_DTLS13 */
-
+    WOLF_ENUM_DUMMY_LAST_ELEMENT(states)
```
anhu, line 2093: I think this one needs to stay as well.

douzzer:
retest this please

anhu:
jenkins retest this please

wolfssl/internal.h
douzzer: Yup.  It's not the subject of a `switch()` `case` anywhere so it's OK.

wolfssl/internal.h
douzzer: Correction, `enum ContentType` _is_ used in a `switch()`, in `GetRecordHeader()`, but there is a `default:` clause.

douzzer (approved):

PR #10860, opened 2026-07-07
Use refcount on DER buffer to prevent UAF

padelsbach:
# Description

This change fixes a use-after-free when the certificate is updated in the SNI callback (triggered by the clienthello). The certificate and certChain buffer are pointed to by both the `WOLFSSL` object and the `WOLFSSL_CTX` object, and a call to  `wolfSSL_CTX_use_certificate_file()` triggers a free+null of the pointers in CTX but not the SSL object. 

Observable failure is an ASAN read failure in `wolfSSL_accept_TLSv13`. Requires `HAVE_SNI` enabled.

Fix is to add reference counter on the shared `DerBuffer`.

Fixes zd 22107.

# Testing

TDD using new unit test.

# Checklist

 - [x] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/tests/api/test_tls13.c
+++ b/tests/api/test_tls13.c
@@ -3639,9 +3639,10 @@ int test_tls13_pha(void)
 static int test_cert_alias_sni_cb(WOLFSSL* ssl, int* ad, void* arg)
 {
     WOLFSSL_CTX* ctx = wolfSSL_get_SSL_CTX(ssl);
+    int* cbRet = (int*)arg;
     (void)ad;
-    (void)arg;
-    (void)wolfSSL_CTX_use_certificate_file(ctx, svrCertFile,
+    /* Feed the reload result back to the parent function below. */
+    *cbRet = wolfSSL_CTX_use_certificate_file(ctx, svrCertFile,
         WOLFSSL_FILETYPE_PEM);
     /* 0 means ack the servername and continue the handshake. */
     return 0;
@@ -3659,6 +3660,7 @@ int test_tls13_cert_alias_uaf_sni(void)
     WOLFSSL *ssl_c = NULL, *ssl_s = NULL;
     struct test_memio_ctx test_ctx;
     const char* host = "example.com";
+    int cbRet = WOLFSSL_FATAL_ERROR;
 
     XMEMSET(&test_ctx, 0, sizeof(test_ctx));
     /* Server cert/key are loaded on ctx_s, so ssl_s->buffers.certificate
@@ -3668,6 +3670,7 @@ int test_tls13_cert_alias_uaf_sni(void)
 
     /* Server swaps its CTX certificate when the SNI arrives. */
     wolfSSL_CTX_set_servername_callback(ctx_s, test_cert_alias_sni_cb);
+    ExpectIntEQ(wolfSSL_CTX_set_servername_arg(ctx_s, &cbRet), WOLFSSL_SUCCESS);
 
     /* Client offers SNI so the server callback fires while parsing the
      * ClientHello. */
@@ -3677,6 +3680,8 @@ int test_tls13_cert_alias_uaf_sni(void)
     /* The callback loads a new certificate on the server CTX, which is
      * aliased by the SSL. The handshake must complete without a UAF. */
     ExpectIntEQ(test_memio_do_handshake(ssl_c, ssl_s, 10, NULL), 0);
+    /* Fail the test if the callback's cert reload failed or didn't run. */
+    ExpectIntEQ(cbRet, WOLFSSL_SUCCESS);
 
     wolfSSL_free(ssl_c);
     wolfSSL_free(ssl_s);
--- a/tests/api/test_tls13.c
+++ b/tests/api/test_tls13.c
@@ -3642,8 +3642,7 @@ static int test_cert_alias_sni_cb(WOLFSSL* ssl, int* ad, void* arg)
     int* cbRet = (int*)arg;
     (void)ad;
     /* Feed the reload result back to the parent function below. */
-    *cbRet = wolfSSL_CTX_use_certificate_file(ctx, svrCertFile,
-        WOLFSSL_FILETYPE_PEM);
+    *cbRet = wolfSSL_CTX_use_certificate_file(ctx, svrCertFile, CERT_FILETYPE);
     /* 0 means ack the servername and continue the handshake. */
     return 0;
 }
--- a/tests/api/test_tls13.c
+++ b/tests/api/test_tls13.c
@@ -3627,6 +3627,66 @@ int test_tls13_pha(void)
 }
 
 
+#if defined(HAVE_MANUAL_MEMIO_TESTS_DEPENDENCIES) && defined(WOLFSSL_TLS13) && \
+    defined(HAVE_SNI) && !defined(WOLFSSL_COPY_CERT) && \
+    !defined(NO_WOLFSSL_CLIENT) && !defined(NO_WOLFSSL_SERVER) && \
+    !defined(NO_RSA) && !defined(NO_CERTS)
+/* React to the client SNI by reloading the certificate on the existing CTX.
+ * Because the SSL was created without WOLFSSL_COPY_CERT, the cert pointer is
+ * identical between the CTX and SSL objects. This test validates the
+ * cert file replacement.
+ * Note the updated cert will not be used for this connection. */
+static int test_cert_alias_sni_cb(WOLFSSL* ssl, int* ad, void* arg)
+{
+    WOLFSSL_CTX* ctx = wolfSSL_get_SSL_CTX(ssl);
+    (void)ad;
+    (void)arg;
+    (void)wolfSSL_CTX_use_certificate_file(ctx, svrCertFile,
+        WOLFSSL_FILETYPE_PEM);
+    /* 0 means ack the servername and continue the handshake. */
+    return 0;
+}
+#endif
+
+int test_tls13_cert_alias_uaf_sni(void)
+{
+    EXPECT_DECLS;
+#if defined(HAVE_MANUAL_MEMIO_TESTS_DEPENDENCIES) && defined(WOLFSSL_TLS13) && \
+    defined(HAVE_SNI) && !defined(WOLFSSL_COPY_CERT) && \
+    !defined(NO_WOLFSSL_CLIENT) && !defined(NO_WOLFSSL_SERVER) && \
+    !defined(NO_RSA) && !defined(NO_CERTS)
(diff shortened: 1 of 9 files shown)


wolfcrypt/src/asn.c
```diff
@@ -24683,13 +24683,30 @@ int AllocDer(DerBuffer** pDer, word32 length, int type, void* heap)
...
     }
     return ret;
 }
 
+void RefDer(DerBuffer* der)
+{
+    if (der != NULL) {
+        int ref_err = 0;
+        wolfSSL_RefInc(&der->ref, &ref_err);
+        (void)ref_err;
```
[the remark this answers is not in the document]
padelsbach: fixed

tests/api/test_tls13.c
```diff
@@ -3627,6 +3627,66 @@ int test_tls13_pha(void)
...
+ * Because the SSL was created without WOLFSSL_COPY_CERT, the cert pointer is
+ * identical between the CTX and SSL objects. This test validates the
+ * cert file replacement.
+ * Note the updated cert will not be used for this connection. */
+static int test_cert_alias_sni_cb(WOLFSSL* ssl, int* ad, void* arg)
+{
+    WOLFSSL_CTX* ctx = wolfSSL_get_SSL_CTX(ssl);
+    (void)ad;
+    (void)arg;
+    (void)wolfSSL_CTX_use_certificate_file(ctx, svrCertFile,
```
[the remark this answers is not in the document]
padelsbach: fixed

padelsbach:
jenkins retest this please

tests/api/test_tls13.c
```diff
@@ -3627,6 +3627,70 @@ int test_tls13_pha(void)
 }
 
 
+#if defined(HAVE_MANUAL_MEMIO_TESTS_DEPENDENCIES) && defined(WOLFSSL_TLS13) && \
```
dgarske, line 3630: **🟠 [Medium] No test coverage for add0/add1_chain_cert on an SSL that aliases the CTX chain** · _Missing Tests_

The added test exercises only the leaf-certificate reload path (wolfSSL_CTX_use_certificate_file in the SNI callback), which correctly decrements the aliased reference via ProcessBufferCertHandleDer's now-unconditional FreeDer. The refcount change also affects the cert-CHAIN aliasing path, but there is no test that appends to an SSL's aliased chain via wolfSSL_add1_chain_cert / wolfSSL_add0_chain_cert: create an SSL from a CTX that loaded an intermediate chain (so ssl->buffers.certChain aliases ctx->certChain with refcount 2), then call wolfSSL_add1_chain_cert(ssl, x509) and free both under leak detection. That is exactly the path where the wolfssl_add_to_chain leak (see High finding) survives, which is why the leak was not caught. Severity: review Medium/SUGGEST, review-security Low; stricter view kept.

**Fix:** Add a test under OPENSSL_EXTRA + KEEP_OUR_CERT that (1) loads a certificate chain on the CTX, (2) creates a WOLFSSL so ssl->buffers.certChain aliases ctx->certChain, (3) calls wolfSSL_add1_chain_cert(ssl, x509) (or wolfSSL_add0_chain_cert), and (4) frees ssl then ctx, verifying no leak under ASAN/valgrind. This validates the chain-alias refcount balance and guards the fix for the High finding.

wolfssl/wolfcrypt/asn_public.h
```diff
@@ -241,8 +241,9 @@ typedef struct DerBuffer {
     byte*  buffer;
```
dgarske, line 241: **🟠 [Medium] Public DerBuffer struct grows a new member (ABI/size change; FreeDer now requires AllocDer-initialized ref)** · _API / ABI_

The public DerBuffer struct gains a `wolfSSL_Ref ref` member, correctly appended at the end (per wolfSSL struct convention, avoiding mid-struct ABI breakage). FreeDer now calls wolfSSL_RefDec(&der->ref, ...) on every buffer. Two consequences: (1) sizeof(DerBuffer) changes, so any application or FIPS-boundary build that allocates/embeds DerBuffer directly, or links against a prebuilt shared library, must be recompiled - a classic shared-library ABI break. (2) In mutex-backed refcount builds (!SINGLE_THREADED && !WOLFSSL_ATOMIC_OPS), a DerBuffer NOT created via AllocDer (which now runs wolfSSL_RefInit) has an uninitialized mutex and hits UB when freed with wc_FreeDer; every cert/key DerBuffer also now embeds and init/frees a mutex per AllocDer/FreeDer. All in-tree allocations go through AllocDer (verified: no sizeof(DerBuffer) allocation or raw XFREE of a DerBuffer struct exists outside wolfcrypt/src/asn.c), and the PR fixed the one internal offender (test_wolfSSL_add_to_chain_overflow now uses AllocDer/FreeDer); the stack `DerBuffer rawDer` in wolfcrypt/src/evp_pk.c is safe because it is never passed to FreeDer. Severity: review Medium/SUGGEST, review-security Info; stricter view kept.

**Fix:** Document in the ChangeLog/release notes the sizeof(DerBuffer) ABI change so downstream users rebuild rather than mix object files across the boundary, and add a contract note in the DerBuffer/FreeDer doxygen that a DerBuffer must be obtained from AllocDer/wc_AllocDer before being freed with FreeDer/wc_FreeDer. No functional code change required.

wolfcrypt/src/asn.c
```diff
@@ -24693,13 +24693,39 @@ int AllocDer(DerBuffer** pDer, word32 length, int type, void* heap)
...
+/* Increments the reference count on a shared DER buffer.
+ *
+ * A mutex-backed refcount can fail to lock, leaving the count un-incremented;
+ * return failure so callers don't proceed with an un-referenced alias.
+ *
+ * @param [in, out] der  DER buffer. May be NULL.
+ * @return  1 on success
+ * @return  0 on error
+ */
+int RefDer(DerBuffer* der)
```
dgarske, line 24720: **🔵 [Low] RefDer treats NULL as success; FreeDer discards mutex-lock failure (leak-on-lock-failure)** · _Style_

RefDer(NULL) returns 1 (success), which is intentional so aliasing a NULL ctx->certificate is a no-op; callers rely on this. Worth a brief note only: in the mutex build, a lock failure inside FreeDer's wolfSSL_RefDec sets isZero=0 and the buffer is intentionally not freed (leak-on-lock-failure), and ref_err is discarded via `(void)ref_err;`. This matches the library's existing refcount conventions (session/ctx refcounts), so no change is required, but the discarded error is easy to misread.

**Fix:** No functional change needed; consider a one-line comment at the FreeDer `(void)ref_err;` noting the deliberate leak-on-lock-failure behavior for parity with other refcounted frees.

dgarske (commented):
## Skoll Multi-Scan Review

### Posted findings
- [Medium] [review+review-security] **No test coverage for add0/add1_chain_cert on an SSL that aliases the CTX chain** — `tests/api/test_tls13.c:3630-3693`
- [Medium] [review+review-security] **Public DerBuffer struct grows a new member (ABI/size change; FreeDer now requires AllocDer-initialized ref)** — `wolfssl/wolfcrypt/asn_public.h:241-246`
- [Low] [review] **RefDer treats NULL as success; FreeDer discards mutex-lock failure (leak-on-lock-failure)** — `wolfcrypt/src/asn.c:24720-24727`

#### wolfssl_add_to_chain leaks aliased cert-chain reference (missed refcount release when weOwn==0)
**File:** `src/ssl_load.c:4924-4929`
**Function:** `wolfssl_add_to_chain`
**Severity:** High
**Category:** Resource Leak

This PR changes the ownership model so aliased cert/certChain buffers now hold a real reference. SetSSL_CTX_CertsAndKeys (src/internal.c:7180-7184) calls RefDer(ssl->buffers.certChain) when the SSL aliases the CTX chain, so an SSL holds a reference on ctx->certChain even though ssl->buffers.weOwnCertChain == 0 (refcount == 2). Every other pointer-drop site in the PR was converted to an UNCONDITIONAL FreeDer to release that reference (wolfSSL_certs_clear, ProcessUserChainRetain, ProcessBufferCertHandleDer, wolfSSL_ResourceFree). But wolfssl_add_to_chain was NOT converted: it still only frees the old chain when `if (weOwn) { FreeDer(chain); }` and then overwrites *chain. wolfSSL_add0_chain_cert / wolfSSL_add1_chain_cert (src/ssl_load.c:5225-5227) call it as wolfssl_add_to_chain(&ssl->buffers.certChain, ssl->buffers.weOwnCertChain, ...). When the chain is aliased (weOwn == 0), the aliased reference taken by RefDer is dropped WITHOUT a matching FreeDer: on wolfSSL_CTX_free the count decrements 2->1 and the CTX cert-chain buffer is never freed. Reachable via the public OpenSSL-compat APIs SSL_add0_chain_cert / SSL_add1_chain_cert (nginx/httpd) under OPENSSL_EXTRA + KEEP_OUR_CERT when the CTX loaded a certificate chain before the SSL was created. Leaks only on the success path of the add. The other two callers (lines 4970, 5162) pass weOwn=1 on a refcount-1 ctx->certChain, so unconditional free is also correct for them. Severity: review mode rated this High/BLOCK; review-security rated it Medium; stricter view kept.

**Recommendation:** Make the FreeDer(chain) unconditional (drop the `if (weOwn)` guard), matching the conversion applied to the other alias-drop sites in this PR. FreeDer now decrements a shared alias and only frees at refcount 0, so this is safe for both owned (weOwn=1 CTX callers) and aliased chains. The weOwn parameter then becomes redundant for the free decision and can be removed. Add a regression test that loads a chain on the CTX, creates an SSL (aliasing the chain), calls SSL_add1_chain_cert on the SSL, then frees both under ASAN/valgrind.

_Referenced code: `src/ssl_load.c:4924-4929` (6 lines)_

---

wolfssl/wolfcrypt/asn_public.h
```diff
@@ -241,8 +241,9 @@ typedef struct DerBuffer {
...
     void*  heap;
     word32 length;
-    int    type;    /* enum CertType */
-    int    dynType; /* DYNAMIC_TYPE_* */
+    int    type;     /* enum CertType */
+    int    dynType;  /* DYNAMIC_TYPE_* */
+    wolfSSL_Ref ref; /* refcount for aliased pointers to the der */
```
dgarske, line 246: Do we really need to always add "wolfSSL_Ref ref;" on all builds? I am concerned about portability and code size growth.

dgarske (changes requested):

padelsbach:
Will redesign a solution to prevent/block this scenario rather than try to fix it. Closing this PR.

PR #10864, opened 2026-07-08
Riscv port bug fixes

danielinux:
# Description

## Two bug fixes for RiscV

1. riscv: return KEYUSAGE_E for AES use without a key schedule (fc1bb6395d46d2ec6c0f3b67d4c48a5d17bcc57a)
    
    The generic wc_AesEncrypt/wc_AesDecrypt (aes.c) reject an AES object
    whose key schedule was never set (rounds outside 1..7 after halving)
    with KEYUSAGE_E, and every mode inherits that through their int return.
    The RISC-V port's block helpers are void, so nothing reported unkeyed
    use: wc_AesEncryptDirect and wc_AesCcmEncrypt/Decrypt silently
    processed with a garbage schedule, and wc_AesCtrEncrypt classified it
    as BAD_FUNC_ARG instead of KEYUSAGE_E.
    
    Align the port with the generic error contract:
    * wc_AesEncryptDirect / wc_AesDecryptDirect: add the generic rounds
      check (also fixes wc_CmacUpdate error reporting, which goes through
      wc_AesEncryptDirect on this port)
    * wc_AesCcmEncrypt / wc_AesCcmDecrypt: same check after the argument
      sanity block
    * wc_AesCtrEncrypt (both variants): the existing rounds switch now
      returns KEYUSAGE_E instead of BAD_FUNC_ARG
      
2. riscv: add missing argument checks in AES port (7b72cf6acff7a785f330b1783526306b203c97fc)
    
    Align the RISC-V AES port's argument validation with the generic aes.c
    implementations (and with the port's own vector/scalar-crypto siblings):
    
    * wc_AesSetKey (base assembly variant, i.e. neither
      WOLFSSL_RISCV_VECTOR_CRYPTO_ASM nor WOLFSSL_RISCV_SCALAR_CRYPTO_ASM):
      reject key == NULL. The vector and scalar-crypto variants of the same
      function already check it; the base variant passed NULL through to the
      key expansion and wc_AesGcmSetKey then dereferenced the uninitialized
      schedule, crashing on e.g. wc_AesGcmSetKey(aes, NULL, 16).
    
    * wc_AesCcmEncrypt / wc_AesCcmDecrypt: reject authIn == NULL when
      authInSz > 0, as the generic implementation does. The port otherwise
      walks the NULL authIn buffer while computing the CBC-MAC.
      
      
# Testing

Found by the ISO 26262 per-module MC/DC campaign's argument-matrix tests  (test_wc_AesGcmSetKey, test_wc_AesCcmArgMcdc) running the RISC-V port  under qemu-riscv64: upstream CI only exercises this port with the KAT  suite, which never passes invalid arguments.

### DIFF
--- a/wolfcrypt/src/port/riscv/riscv-64-aes.c
+++ b/wolfcrypt/src/port/riscv/riscv-64-aes.c
@@ -1375,8 +1375,10 @@ int wc_AesCtrEncrypt(Aes* aes, byte* out, const byte* in, word32 sz)
         #endif /* WOLFSSL_AES_256 */
                 break;
             default:
+                /* Not set or corrupted: match the generic wc_AesEncrypt
+                 * (aes.c) behavior for an unusable key schedule. */
                 WOLFSSL_MSG("Bad AES-CTR round value");
-                ret = BAD_FUNC_ARG;
+                ret = KEYUSAGE_E;
         }
     }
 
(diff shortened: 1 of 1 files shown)


douzzer (approved):
tested with
```
wolfssl-multi-test.sh ...
cross-riscv64-all-asm
cross-riscv32-all-asm
```

PR #10865, opened 2026-07-08
Add refcounts on key, keyMask, altKey, altKeyMask to prevent UAF

padelsbach:
# Description

This is a follow up to #10860. In this case, if the application replaces the private key file in the SNI callback (which frees and zeros the buffers) a subsequent sign operation would use stale or zeroed key. 

This is a more impactful bug than in #10860. Fix is again to use refcounters on the fields.

# Testing

Added unit test.

# Checklist

 - [x] added tests
 - [ ] updated/added doxygen


padelsbach:
jenkins retest this please

padelsbach:
Will redesign a solution to prevent/block this scenario rather than try to fix it. Closing this PR.

PR #10866, opened 2026-07-08
check RSA key size for PSS

kojo1:
# Description

Don't select RSA-PSS in MatchSigAlgo when the RSA key is too small for the signature.
When a TLS 1.2 server holds a small RSA certificate (e.g. 512bit) and the ClientHello's signature_algorithms extension offers rsa_pss_rsae_sha256 (0x0804), the handshake is aborted even though the client also offers rsa_pkcs1_sha256 (0x0401), which the key can use.

Fixes zd#22091

# Testing

Local test with 512bit cert and run time SetMinRsaKey_Sz.

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -31062,6 +31062,8 @@ int SetSuitesHashSigAlgo(Suites* suites, const char* list)
 #endif /* OPENSSL_EXTRA */
 
 #if !defined(NO_TLS) && (!defined(NO_WOLFSSL_SERVER) || !defined(NO_CERTS))
+/* Smallest RSA key able to make an RSA-PSS/SHA-256 signature: emLen >= 2*hLen+2 = 66 bytes (RFC 8017 9.1.1, salt len = hash len). */
+#define RSA_PSS_SHA256_MIN_SZ  66   /* bytes: 2*WC_SHA256_DIGEST_SIZE + 2 */
 static int MatchSigAlgo(WOLFSSL* ssl, int sigAlgo)
 {
 #ifdef HAVE_ED25519
@@ -31124,9 +31126,14 @@ static int MatchSigAlgo(WOLFSSL* ssl, int sigAlgo)
         if (IsAtLeastTLSv1_3(ssl->version))
             return sigAlgo == rsa_pss_sa_algo;
     #endif
-        /* TLS 1.2 and below - RSA-PSS allowed. */
-        if (sigAlgo == rsa_pss_sa_algo)
+        /* TLS 1.2 and below - RSA-PSS allowed, but only if the RSA key is big
+         * enough to actually produce an RSA-PSS/SHA-256 signature. */
+        if (sigAlgo == rsa_pss_sa_algo) {
+            if (ssl->buffers.keySz != 0 &&
+                    ssl->buffers.keySz < RSA_PSS_SHA256_MIN_SZ)
+                return 0;
             return 1;
+        }
     }
 #endif
 #ifdef HAVE_ECC_BRAINPOOL
--- a/src/internal.c
+++ b/src/internal.c
@@ -31062,10 +31062,42 @@ int SetSuitesHashSigAlgo(Suites* suites, const char* list)
 #endif /* OPENSSL_EXTRA */
 
 #if !defined(NO_TLS) && (!defined(NO_WOLFSSL_SERVER) || !defined(NO_CERTS))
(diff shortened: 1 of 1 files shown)


src/internal.c
```diff
@@ -31124,9 +31126,14 @@ static int MatchSigAlgo(WOLFSSL* ssl, int sigAlgo)
         if (IsAtLeastTLSv1_3(ssl->version))
             return sigAlgo == rsa_pss_sa_algo;
     #endif
-        /* TLS 1.2 and below - RSA-PSS allowed. */
-        if (sigAlgo == rsa_pss_sa_algo)
+        /* TLS 1.2 and below - RSA-PSS allowed, but only if the RSA key is big
```
Frauschi, line 31161: **🟡 [Medium] Key-size check only covers RSA-PSS/SHA-256, not SHA-384/SHA-512**
*💡 SUGGEST*  `bug`

The new guard hardcodes the SHA-256 minimum (66 bytes) but `MatchSigAlgo` receives only `sigAlgo` — it cannot see the hash. In `PickHashSigAlgo` (src/internal.c:31275-31284), the loop iterates over client-offered (hashAlgo, sigAlgo) pairs, and TLS 1.2 RSA-PSS may use SHA-384 (needs emLen >= 2*48+2 = 98 bytes) or SHA-512 (needs 2*64+2 = 130 bytes). A medium key — e.g. a 768-bit / 96-byte RSA cert — passes this check (96 >= 66), so `rsa_pss_sa_algo` with `rsa_pss_rsae_sha512` gets selected, and the actual sign later fails with emLen too small. The same handshake abort this PR aims to fix therefore still occurs for keys that are big enough for SHA-256 PSS but not for the SHA-384/SHA-512 PSS the peer offered. Consider passing `hashAlgo` into `MatchSigAlgo` (or checking the digest-derived minimum in the loop) so the bound scales with the actual hash: `emLen >= 2*hLen + 2`.

**Suggestion:**
```suggestion
/* Pass hashAlgo through and compute the min from the hash digest size */
/* minSz = 2 * <digest size for hashAlgo> + 2; reject if keySz < minSz */
```

src/internal.c
```diff
@@ -31062,6 +31062,8 @@ int SetSuitesHashSigAlgo(Suites* suites, const char* list)
 #endif /* OPENSSL_EXTRA */
 
 #if !defined(NO_TLS) && (!defined(NO_WOLFSSL_SERVER) || !defined(NO_CERTS))
+/* Smallest RSA key able to make an RSA-PSS/SHA-256 signature: emLen >= 2*hLen+2 = 66 bytes (RFC 8017 9.1.1, salt len = hash len). */
```
Frauschi: **🔵 [Low] Magic number 66 hardcoded instead of derived from WC_SHA256_DIGEST_SIZE**
*🔧 NIT*  `convention`

The macro value is written as the literal `66` with a comment explaining it equals `2*WC_SHA256_DIGEST_SIZE + 2`. Per the repo's C style (prefer named constants/expressions over magic numbers), defining it as the actual expression makes it self-checking if the digest size ever changes and removes the need for the explanatory comment. `WC_SHA256_DIGEST_SIZE` is already available (defined in wolfssl/internal.h and wolfcrypt headers).

**Suggestion:**
```suggestion
#define RSA_PSS_SHA256_MIN_SZ  (2 * WC_SHA256_DIGEST_SIZE + 2)
```

Frauschi (commented):
## 🐺 Skoll Code Review

### Posted findings
- [Medium] **Key-size check only covers RSA-PSS/SHA-256, not SHA-384/SHA-512** — `src/internal.c:31129-31136`
- [Low] **Magic number 66 hardcoded instead of derived from WC_SHA256_DIGEST_SIZE** — `src/internal.c:31065-31066`

- [Medium] No test added for the small-RSA-key PSS rejection path
- [Low] keySz is used as a proxy for emLen (loose by up to 1 byte)

</details>

src/internal.c
kojo1: added MinRsaPssKeySz

src/internal.c
kojo1: This is not a constant, as in MinRsaPssKeySz.

Frauschi:
Jenkins retest this please.

Frauschi (approved):
LGTM

PR #10879, opened 2026-07-10
dtls: fail explicitly on oversized HelloVerifyRequest cookie (#10608)

### LINKED ISSUE #10608: DTLS 1.2 client only retains HelloVerifyRequest cookies up to 32 bytes
# DTLS 1.2 client only retains HelloVerifyRequest cookies up to 32 bytes

## Summary
wolfSSL implements the DTLS 1.2 cookie exchange, but its client-side storage for the cookie received in `HelloVerifyRequest` is limited to `MAX_COOKIE_LEN`, which is 32 bytes. RFC 6347 represents the DTLS 1.2 cookie as an 8-bit-length opaque vector, so a conforming server can send a cookie larger than 32 bytes and up to 255 bytes.

## Standard Requirement
Official standard: https://www.rfc-editor.org/rfc/rfc6347

RFC 6347 Section 4.2.1, Denial-of-Service Countermeasures

```text
The server responds with a HelloVerifyRequest containing a stateless cookie. The client retransmits the ClientHello with the cookie added. The server verifies the cookie before continuing. Cookies SHOULD be generated as HMAC(Secret, Client-IP, Client-Parameters). When the server changes the Secret value, it SHOULD retain the previous value for a limited period and accept cookies generated with either secret.
```

RFC 6347 also defines the DTLS 1.2 `ClientHello.cookie` and `HelloVerifyRequest.cookie` fields as `opaque cookie<0..2^8-1>`. Taken together, the syntax allows a cookie length from 0 to 255 bytes, and the client is required to retransmit the `ClientHello` with the cookie received from `HelloVerifyRequest`.

## Relevant Source Code
```c
src/internal.c:31357
if (cookieSz <= MAX_COOKIE_LEN) {
    XMEMCPY(ssl->arrays->cookie, input + *inOutIdx, cookieSz);
    ssl->arrays->cookieSz = cookieSz;
}

wolfssl/internal.h:1572
MAX_COOKIE_LEN = 32
```

The retransmitted DTLS `ClientHello` uses the saved value:

```c
src/internal.c:31225
byte cookieSz = ssl->arrays->cookieSz;

src/internal.c:31227
output[idx++] = cookieSz;

src/internal.c:31229
XMEMCPY(&output[idx], ssl->arrays->cookie, cookieSz);
```

Server-side context:

```c
src/dtls.c:284
if (ch->cookie.size != DTLS_COOKIE_SZ)
    return 0;
```

## Implementation Behavior
When wolfSSL parses a DTLS `HelloVerifyRequest` as a client, it reads the one-byte cookie length but only copies the cookie into `ssl->arrays->cookie` when `cookieSz <= MAX_COOKIE_LEN`. Since `MAX_COOKIE_LEN` is 32, a syntactically valid 33..255 byte cookie is skipped rather than retained. The later `ClientHello` retransmission uses `ssl->arrays->cookieSz`, so a skipped larger cookie is not echoed back to the server.

The server-side `DTLS_COOKIE_SZ` check is a narrower point. wolfSSL generates and verifies its own fixed-size HMAC cookie, so requiring that size for cookies returned to its own server is mainly an implementation cookie-format policy. It is useful context, but the clearer compliance and interoperability gap is on the client side: wolfSSL cannot preserve and retransmit a larger cookie sent by another conforming DTLS 1.2 server.

## Inconsistency Reason
RFC 6347 permits DTLS 1.2 cookie values up to 255 bytes and requires the client to retransmit `ClientHello` with the received cookie added. wolfSSL's client-side receive path retains only cookies up to 32 bytes. Therefore, a valid `HelloVerifyRequest` carrying a 33..255 byte cookie can be processed without the cookie being saved, preventing the following `ClientHello` from carrying the required cookie.

This is a partial compliance issue rather than a complete absence of cookie support: wolfSSL handles its own 20-byte or 32-byte generated cookies, but it is not interoperable with conforming peers that choose larger DTLS 1.2 cookies within the RFC vector limit.

## Runtime Evidence
The verification script confirms `MAX_COOKIE_LEN = 32`, the guarded copy in `DoHelloVerifyRequest`, and the fixed-size server cookie comparison. The first two facts are the primary evidence for this issue.

## Impact
wolfSSL clients can fail to complete the DTLS 1.2 cookie exchange with servers that send legal `HelloVerifyRequest` cookies larger than 32 bytes.

## Fix Direction
Increase the client-side cookie storage path to preserve peer cookies up to the RFC 6347 vector limit of 255 bytes, then retransmit exactly the received cookie in the next `ClientHello`. The server can still keep its own fixed HMAC cookie format if desired, but the client parser should not silently drop valid peer cookies above 32 bytes.

94xhn:
Fixes #10608.

## Problem

RFC 6347 defines the DTLS 1.2 `HelloVerifyRequest`/`ClientHello` cookie as an 8-bit-length opaque vector (`cookie<0..2^8-1>`), so a conforming server can legally send a cookie up to 255 bytes. wolfSSL's client-side buffer for this cookie is `MAX_COOKIE_LEN` (`WOLFSSL_COOKIE_LEN`, default 32, user-overridable at compile time up to the RFC's 255-byte limit).

`DoHelloVerifyRequest()` previously handled a cookie larger than the configured buffer by simply not copying it:

```c
if (cookieSz <= MAX_COOKIE_LEN) {
    XMEMCPY(ssl->arrays->cookie, input + *inOutIdx, cookieSz);
    ssl->arrays->cookieSz = cookieSz;
}
/* no else - falls through as if nothing happened */

*inOutIdx += cookieSz; /* still correctly advances past the cookie bytes */
```

The function still returns success, and `ssl->arrays->cookieSz` is left at its previous value (0 on a fresh handshake). The retransmitted `ClientHello` then goes out without the cookie the server asked for, and the handshake fails later - typically as a generic verification failure or timeout on the server side - with nothing in the client-side return value indicating that an oversized cookie was silently dropped.

## Fix

When the received cookie doesn't fit in the client's configured buffer, return `BUFFER_ERROR` immediately instead of silently continuing. This surfaces the real cause of the handshake failure at the point where it actually occurs, and matches the sibling server-side check a few thousand lines away in the same file (the equivalent `peerCookieSz > MAX_COOKIE_LEN` check while parsing a `ClientHello`'s returned cookie already returns `BUFFER_ERROR` rather than silently proceeding).

I deliberately did **not** change the default value of `WOLFSSL_COOKIE_LEN` itself (e.g. bumping it to 255) - that's a memory/interoperability tradeoff (every DTLS SSL object's cookie buffer would grow accordingly) that's better left as a build-time choice via the existing `WOLFSSL_COOKIE_LEN` override, not something this fix should silently impose on every user by changing the shipped default.

## Verification

I don't have a live DTLS 1.2 peer configured to send an oversized cookie handy, so I verified the corrected branch logic with a standalone C reproduction of the exact `cookieSz`/`MAX_COOKIE_LEN` comparison and buffer-copy sequence:

- A 40-byte cookie (RFC-legal, exceeds the default 32-byte `MAX_COOKIE_LEN`): before the fix, the function returns success while silently leaving `cookieSz` at 0. After the fix, it returns `BUFFER_ERROR` immediately.
- A normal 20-byte cookie (within the default limit): unchanged before and after - copied correctly, `cookieSz` set, function returns success.

## Disclosure

Generative AI (Claude) was used to help investigate this issue, implement, and verify this fix. All changes were reviewed by me before submission.

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -32855,10 +32855,19 @@ static void MakePSKPreMasterSecret(Arrays* arrays, byte use_psk_key)
                 return BUFFER_ERROR;
 
 #ifdef WOLFSSL_DTLS
-            if (cookieSz <= MAX_COOKIE_LEN) {
-                XMEMCPY(ssl->arrays->cookie, input + *inOutIdx, cookieSz);
-                ssl->arrays->cookieSz = cookieSz;
-            }
+            /* RFC 6347 allows a cookie of up to 255 bytes
+             * (opaque cookie<0..2^8-1>). If the peer's cookie doesn't fit
+             * in our configured buffer (MAX_COOKIE_LEN, adjustable via
+             * WOLFSSL_COOKIE_LEN), fail explicitly instead of silently
+             * dropping it - a silently-dropped cookie is never echoed
+             * back in the retransmitted ClientHello, causing the
+             * handshake to fail later in a way that gives no indication
+             * of the real cause. */
+            if (cookieSz > MAX_COOKIE_LEN)
+                return BUFFER_ERROR;
+
+            XMEMCPY(ssl->arrays->cookie, input + *inOutIdx, cookieSz);
+            ssl->arrays->cookieSz = cookieSz;
 #endif
             *inOutIdx += cookieSz;
         }


dgarske:
Hi @94xhn , thank you for the report. Unfortunately we cannot accept pull requests from your region. You are welcome to submit an issue report by emailing support@wolfssl.com. Thanks, David Garske, wolfSSL

julek-wolfssl (commented):
We ignore oversized cookies because if they don't fit in our buffer then they were not generated by the current wolfSSL instance. There is no point to store it.

PR #10883, opened 2026-07-10
Add AES-GCM DEM, CryptoCb support, and devId threading to ECIES

night1rider:
Add AES-GCM (128/256) as an ECIES DEM next to the AES-CBC/CTR+HMAC modes.

Only the encryption key comes from the KDF; the mac salt is bound as GCM AAD and the 16-byte tag replaces the HMAC. 

The GCM DEM honors all three IV build modes, and default fixed-nonce GCM is gated behind the new WOLFSSL_ECIES_STATIC_GCM_NONCE opt-in. Adds ECIES CryptoCb encrypt/decrypt, the WOLF_CRYPTO_CB ctx getters, devId/heap threading into the DEM primitives, and test/benchmark/CI coverage.


night1rider:
Rebased to fix merge conflict

night1rider:
Jenkins retest this please

wolfcrypt/benchmark/benchmark.c
```diff
@@ -13997,62 +14023,143 @@ void bench_eccEncrypt(int curveId)
...
+            #ifdef WOLFSSL_AES_256
+            { ecAES_256_GCM, "AES256GCM" },
+            #endif
+        #endif
+            { 0, NULL } /* sentinel */
+        };
+        ecEncCtx* cliCtx = wc_ecc_ctx_new(REQ_RESP_CLIENT, &gRng);
+        ecEncCtx* srvCtx = wc_ecc_ctx_new(REQ_RESP_SERVER, &gRng);
+        char      encDesc[32];
+        char      decDesc[32];
```
JacobBarthelmeh: Avoid magic numbers here (32), maybe WC_BENCH_MAX_LINE_LEN if it should fit on one line. Or another macro.

wolfcrypt/src/ecc.c
```diff
@@ -15170,35 +15354,46 @@ int wc_ecc_encrypt_ex(ecc_key* privKey, ecc_key* pubKey, const byte* msg,
...
+    if (ret != 0)
+        return ret;
 
     if (keysLen > ECC_BUFSIZE) /* keys size */
         return BUFFER_E;
 
     if ((msgSz % blockSz) != 0)
         return BAD_PADDING_E;
 
+    if (ecc_is_gcm(ctx->encAlgo)) {
```
JacobBarthelmeh: This section of code is identical to the section at line 15700, please consider a helper function for the checks.

wolfcrypt/src/ecc.c
```diff
@@ -14978,6 +15034,12 @@ static void ecc_ctx_init(ecEncCtx* ctx, int flags, WC_RNG* rng)
         #else
             ctx->encAlgo  = ecAES_128_CTR;
         #endif
+    #elif !defined(NO_AES) && defined(HAVE_AESGCM)
```
JacobBarthelmeh: In a GCM-only build with default IV mode, ecc_ctx_init defaults to AES-GCM but encrypt/decrypt then reject GCM with NOT_COMPILED_IN, making ECIES unusable without a manual wc_ecc_ctx_set_algo override.

wolfssl/wolfcrypt/ecc.h
```diff
@@ -1059,6 +1061,8 @@ enum {
     IV_SIZE_64       =  8,
     IV_SIZE_128      = 16,
     ECC_MAX_IV_SIZE  = 16,
+    AES_GCM_NONCE_SZ = 12,   /* GCM IV/nonce length for ECIES DEM */
```
JacobBarthelmeh: I think these enums are duplicates of existing AES ones available. Use GCM_NONCE_MID_SZ for nonce size and in other places we use WC_AES_BLOCK_SIZE for a 16 bit tag.

JacobBarthelmeh (changes requested):

wolfcrypt/benchmark/benchmark.c
night1rider: Updated

wolfcrypt/src/ecc.c
night1rider: Created a static function helper

wolfcrypt/src/ecc.c
night1rider: Good observation, added a compile time check.

wolfssl/wolfcrypt/ecc.h
night1rider: removed the duplicates

night1rider:
Addressed comments, and rebased off current master.

JacobBarthelmeh (approved):

PR #10886, opened 2026-07-12
Cryptocb hooks rsapss ed448 cmac

night1rider:
Adds WOLF_CRYPTO_CB crypto callback hooks so a device can service three more operations:
  Ed448 sign and verify, mirroring the existing Ed25519 hooks.
  CMAC context free on wc_CmacFree (WOLF_CRYPTO_CB_FREE), letting a device release offload state.
  RSA-PSS verify with the digest (WOLF_CRYPTO_CB_RSA_PAD) so the device performs the full signature and padding check.
  Includes testwolfcrypt and API unit test coverage for each hook.


night1rider:
jenkins retest this please

wolfcrypt/src/ed448.c
```diff
@@ -447,6 +450,17 @@ int wc_ed448_sign_msg_ex(const byte* in, word32 inLen, byte* out,
                                      ((context == NULL) && (contextLen != 0))) {
         ret = BAD_FUNC_ARG;
     }
+
+#ifdef WOLF_CRYPTO_CB
+    if ((ret == 0) && (key->devId != INVALID_DEVID)) {
```
JacobBarthelmeh: This does not follow the same pattern for the crypto cb find compile option. i.e.

```
 #ifndef WOLF_CRYPTO_CB_FIND
    if (cmac->devId != INVALID_DEVID)
 #endif
```

wolfcrypt/src/ed448.c
```diff
@@ -902,6 +916,16 @@ int wc_ed448_verify_msg_ex(const byte* sig, word32 sigLen, const byte* msg,
         return BAD_LENGTH_E;
     }
 
+#ifdef WOLF_CRYPTO_CB
+    if (key->devId != INVALID_DEVID) {
```
JacobBarthelmeh: This does not follow the same pattern for the crypto cb find compile option. i.e.

```
 #ifndef WOLF_CRYPTO_CB_FIND
    if (cmac->devId != INVALID_DEVID)
 #endif
```

wolfcrypt/src/rsa.c
```diff
@@ -4670,6 +4673,27 @@ int wc_RsaPSS_VerifyCheckInline(byte* in, word32 inLen, byte** out,
             saltLen = RSA_PSS_SALT_MAX_SZ;
     #endif
 
+#if defined(WOLF_CRYPTO_CB) && defined(WOLF_CRYPTO_CB_RSA_PAD)
+    /* Let a device verify signature + padding in one shot (it gets the digest,
+     * which the RsaPad path does not). Fall through to software if unavailable. */
+    if (key != NULL && key->devId != INVALID_DEVID) {
```
JacobBarthelmeh: Check this and similar locations for the guard around devId check.

wolfcrypt/test/test.c
```diff
@@ -81151,6 +81260,196 @@ WOLFSSL_TEST_SUBROUTINE wc_test_ret_t cryptocb_test(void)
...
+#endif /* HAVE_ED448 */
+
+#if defined(WOLFSSL_CMAC) && defined(WOLF_CRYPTO_CB_FREE) && \
+    !defined(NO_AES) && defined(WOLFSSL_AES_DIRECT)
+    if (ret == 0) {
+        byte   cmacKey[16] = {
+            0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08,
+            0x09, 0x0a, 0x0b, 0x0c, 0x0d, 0x0e, 0x0f, 0x10
+        };
+        byte   cmacIn[16] = {
```
JacobBarthelmeh: using magic number 16 here and then WC_AES_BLOCK_SIZE for tag. Please make them all match with WC_AES_BLOCK_SIZE for a buffer of 16 bytes.

wolfssl/wolfcrypt/rsa.h
```diff
@@ -414,6 +414,8 @@ WOLFSSL_API int  wc_RsaPSS_CheckPadding_ex2(const byte* in, word32 inLen,
                                            const byte* sig, word32 sigSz,
                                            enum wc_HashType hashType,
                                            int saltLen, int bits, void* heap);
+/* Crypto callback (WOLF_CRYPTO_CB_RSA_PAD) path: a positive return still means
+ * success but *out is set to NULL, so callers must not dereference *out. */
```
JacobBarthelmeh: I don't think these extra 2 lines of comments add much value, would rather see this file not be in the "git diff" since this behavior was noted in other code comment sections.

JacobBarthelmeh (changes requested):

JacobBarthelmeh (dismissed):

dgarske (changes requested):
Please resolve merge conflicts. Thanks

wolfcrypt/src/cmac.c
dgarske, line 171: Consider also setting `cmac->type = type` here so the wc_CmacFree can always cleanup properly.

wolfcrypt/src/ed448.c
dgarske: Recommend we do this check for crypto callback too right? Maybe move it up higher? Also bring in the brace to same line if under 80 chars.

wolfcrypt/src/ed448.c
```diff
@@ -924,6 +943,19 @@ int wc_ed448_verify_msg_ex(const byte* sig, word32 sigLen, const byte* msg,
         return BAD_LENGTH_E;
     }
 
+#ifdef WOLF_CRYPTO_CB
+    #ifndef WOLF_CRYPTO_CB_FIND
+    if (key->devId != INVALID_DEVID)
+    #endif
+    {
+        ret = wc_CryptoCb_Ed448Verify(sig, sigLen, msg, msgLen, res, key, type,
```
dgarske, line 952: Set `*res = 0;` before, just in case.

dgarske (changes requested):

wolfcrypt/src/ed448.c
dgarske, line 450: Recommendation: Relocate the `WOLFSSL_CHECK_MEM_ZERO` registration block to after the crypto-callback hook (mirroring `wc_ed25519_sign_msg_ex`), or alternatively `goto` a common exit label instead of returning directly. Add a `--enable-cryptocb` + `WOLFSSL_CHECK_MEM_ZERO` build to the validation matrix for this PR.

wolfcrypt/src/rsa.c
```diff
@@ -4679,6 +4682,32 @@ int wc_RsaPSS_VerifyCheckInline(byte* in, word32 inLen, byte** out,
...
+        int res = 0;
+        ret = wc_CryptoCb_RsaPssVerify(in, inLen, digest, digestLen, hash, mgf,
+                                       saltLen, key, &res);
+        if (ret != WC_NO_ERR_TRACE(CRYPTOCB_UNAVAILABLE)) {
+            if (ret == 0) {
+                /* Device verified internally; no recovered PSS block to expose,
+                 * so report no inline output rather than a misleading pointer. */
+                if (out != NULL)
+                    *out = NULL;
+                ret = (res != 0) ? (int)inLen : SIG_VERIFY_E;
```
dgarske: wc_RsaPSS_VerifyCheck device path returns a different length and leaves `out` untouched, with no way for the caller to tell. 

The software path returns the value of `wc_RsaPSS_Verify_ex()`, which ultimately comes from `RsaUnPad_PSS` (rsa.c:2024): `return saltLen + hLen;` — 64 for SHA-256 with a 32-byte salt — and it fills the caller's `out` buffer with that recovered PSS block. The new crypto-callback path returns `(int)inLen` — 256 for a 2048-bit key — and **never writes `out` at all**.

Recommendation: Do not return `inLen` as if it were the recovered-PSS-block length. Either have the device report the actual output length and fill `out`, or define an unambiguous "device verified, no recovered data" signal. Whichever is chosen, document it in `doc/dox_comments/header_files/rsa.h` for `wc_RsaPSS_VerifyCheck` as well as `wc_RsaPSS_VerifyCheckInline`, and add a test that asserts the device return value equals the software return value (or the documented sentinel).

tests/api/test_rsa.c
```diff
@@ -1999,3 +2002,154 @@ int test_wc_RsaFeatureCoverage(void)
...
+
+    ExpectIntEQ(wc_CryptoCb_RegisterDevice(devId, rsa_pss_test_crypto_cb,
+                                           &pssVerifySeen), 0);
+    ExpectIntEQ(wc_InitRng(&rng), 0);
+    ExpectIntEQ(wc_InitRsaKey_ex(key, HEAP_HINT, devId), 0);
+    ExpectIntEQ(wc_MakeRsaKey(key, 2048, WC_RSA_EXPONENT, &rng), 0);
+    ExpectIntEQ(wc_RsaSetRNG(key, &rng), 0);
+
+    /* PSS sign runs in software (device declines). */
+    ExpectIntGT(sigLen = (word32)wc_RsaPSS_Sign(digest,
```
dgarske: Buffer overflow in new RSA-PSS test when wc_RsaPSS_Sign fails: word32 sigLen holds a cast negative error code

Recommendation: Capture the signature length in a signed `int` and only widen to `word32` after confirming it is positive, and additionally gate the plain-`if` inline sub-test on `EXPECT_SUCCESS()` so it does not run after a failure.

wolfcrypt/src/rsa.c
```diff
@@ -4679,6 +4682,32 @@ int wc_RsaPSS_VerifyCheckInline(byte* in, word32 inLen, byte** out,
...
+            }
+            return ret;
```
dgarske, line 4760: Recommendation: Explicitly map any positive device return to `SIG_VERIFY_E` (fail closed), and document in `cryptocb.h` that a `WC_PK_TYPE_RSA_PSS_VERIFY` handler must return exactly 0 (with `*res` set) or a negative error.

wolfcrypt/src/cmac.c
```diff
@@ -376,6 +380,17 @@ int wc_CmacFree(Cmac* cmac)
...
     if (cmac == NULL)
         return BAD_FUNC_ARG;
+#if defined(WOLF_CRYPTO_CB) && defined(WOLF_CRYPTO_CB_FREE)
+    /* Let the device release any per-context state it hung off cmac->devCtx
+     * before the struct is zeroed (e.g. an offload context never finalized). */
+    #ifndef WOLF_CRYPTO_CB_FIND
+    if (cmac->devId != INVALID_DEVID)
+    #endif
+    {
+        (void)wc_CryptoCb_Free(cmac->devId, WC_ALGO_TYPE_CMAC, (int)cmac->type,
```
dgarske, line 388: Reset `cmac->devId = INVALID_DEVID` after the `ForceZero` so a repeat `wc_CmacFree()` is inert

wolfcrypt/test/test.c
```diff
@@ -78834,6 +78844,51 @@ static int myCryptoDevCb(int devIdArg, wc_CryptoInfo* info, void* ctx)
         WOLFSSL_MSG_EX("CryptoDevCb: Pk Type %d\n", info->pk.type);
     #endif
 
+    #if defined(WC_RSA_PSS) && defined(WOLF_CRYPTO_CB_RSA_PAD)
```
dgarske: Add the same `WOLF_CRYPTO_CB_ONLY_RSA` escape as the sibling branches, or add `&& !defined(WOLF_CRYPTO_CB_ONLY_RSA)` to the block guard.

dgarske (changes requested):

wolfcrypt/src/cmac.c
night1rider: Writing `devId` back after the `ForceZero` breaks `WOLFSSL_CHECK_MEM_ZERO`, which requires the freed struct to be all zero. On the repeat-free concern: with nothing registered at `devId` 0 the dispatch finds no device and is a no-op, and if something is registered there it receives a fully zeroed context with a NULL `devCtx` and nothing to release. `wc_AesFree` and `wc_Sha256Free` behave the same way, so I left CMAC consistent with them.

night1rider:
Jenkins retest this please

wolfcrypt/src/ed448.c
night1rider: Addressed

wolfcrypt/src/ed448.c
night1rider: addressed

night1rider: Addressed. Moved the length check above the callback and put the brace on the same line.

wolfcrypt/src/rsa.c
night1rider: A device that recovers data fills out and we return its length. A device that only gives a verdict now returns saltLen + hLen, the same value software returns, and out is zeroed so no stale bytes are left behind. This is documented in doc/dox_comments/header_files/rsa.h, and the new test checks that the device return matches the software return and that the bytes match.

tests/api/test_rsa.c
night1rider: Addressed

wolfcrypt/src/rsa.c
night1rider: Addressed. A positive return from the device is now mapped to `SIG_VERIFY_E` in both `wc_RsaPSS_VerifyCheck` and `wc_RsaPSS_VerifyCheckInline`, and documented on the `rsa_pss_verify` struct in `cryptocb.h`.

wolfcrypt/test/test.c
night1rider: Addressed

wolfcrypt/src/rsa.c
```diff
@@ -4679,6 +4681,55 @@ int wc_RsaPSS_VerifyCheckInline(byte* in, word32 inLen, byte** out,
             saltLen = RSA_PSS_SALT_MAX_SZ;
     #endif
 
+#if defined(WOLF_CRYPTO_CB) && defined(WOLF_CRYPTO_CB_RSA_PAD)
+    /* Let a device verify signature + padding in one shot (it gets the digest,
+     * which the RsaPad path does not). Fall through to software if unavailable. */
+    #ifndef WOLF_CRYPTO_CB_FIND
```
dgarske: Could we create a helper so we don't have to duplicate this code and the one at 4793-4837? They look the same.

dgarske (changes requested):

dgarske (approved):

PR #10895, opened 2026-07-14
AES-GCM: constant-time output clear on decrypt auth failure

danielinux:
# Description

AES_GCM_decrypt_C cleared the output on a tag mismatch with 'if (ret != 0) ForceZero(out, sz)'. That is a conditional branch on the secret-dependent authentication result, which is not constant time and is flagged by the ct-valgrind constant-time test (Conditional jump depends on uninitialised value in AES_GCM_decrypt_C).

Mask the output with 'res' (already computed as all-ones on tag mismatch, zero on match) instead of branching, matching the constant-time idiom used for the tag comparison itself. C path only; the AES-NI/ASM paths are unaffected.

# Testing

Found while troubleshooting a recurring failing nightly test `nightly-ConstantTimeTest-v2`


wolfcrypt/src/aes.c
```diff
@@ -11439,8 +11441,11 @@ int WARN_UNUSED_RESULT AES_GCM_decrypt_C(
     ret = (ret & ~res);
     ret |= (res & WC_NO_ERR_TRACE(AES_GCM_AUTH_E));
 #endif
-    if (ret != 0) {
-        ForceZero(out, sz);
+    /* Mask the output on auth failure instead of branching, to keep the tag
+     * compare constant time. res is all-ones on mismatch, zero on match. */
+    mask = (byte)res;
```
Frauschi, line 11450: **🔵 [Low] Output masking now runs a full pass over out[] on every successful decrypt**
*💡 SUGGEST*  `question`

The previous code only touched `out` on the (rare) auth-failure path. The new masking loop iterates over all `sz` bytes and performs `out[i] &= 0xFF` on every call, including the common success path, adding an O(sz) read-modify-write pass over the plaintext for each decryption. This is the inherent cost of removing the secret-dependent branch (making it constant time), so it is expected — but it is a real per-record overhead on the decrypt hot path (e.g. TLS). Worth confirming this cost is acceptable versus, for example, folding the mask into the existing `xorbufout`/`xorbuf` decrypt loops so no extra pass is needed.

Note for the `WC_AES_GCM_DEC_AUTH_EARLY` build: in that config the tag is checked before decryption and a mismatch returns early (`if (ret != 0) return ret;`) before `out` is written, so `res` is provably 0 at this masking loop and the pass is a guaranteed no-op. It could be `#ifndef WC_AES_GCM_DEC_AUTH_EARLY`-guarded to skip the wasted pass entirely in that config.

**Suggestion:**
```suggestion
Consider whether the mask can be applied within the existing decryption loops (xorbufout/partial handling) to avoid a second full pass over the buffer, while keeping the operation unconditional/constant-time.
```

wolfcrypt/src/aes.c
```diff
@@ -11439,8 +11441,11 @@ int WARN_UNUSED_RESULT AES_GCM_decrypt_C(
     ret = (ret & ~res);
     ret |= (res & WC_NO_ERR_TRACE(AES_GCM_AUTH_E));
 #endif
-    if (ret != 0) {
-        ForceZero(out, sz);
+    /* Mask the output on auth failure instead of branching, to keep the tag
```
Frauschi, line 11445: **🔵 [Low] No test asserts output buffer is zeroed on tag mismatch for the C path**
*💡 SUGGEST*  `test`

Existing tests verify that `wc_AesGcmDecrypt` returns `AES_GCM_AUTH_E` on a tampered tag (wolfcrypt/test/test.c around lines 20540 and 20846), but none assert that the plaintext output buffer is cleared to zero on auth failure. Since this PR rewrites the exact code that performs that clearing, a regression in the masking logic (e.g. an inverted mask) would still pass the current tests. A small test that decrypts with a corrupted tag and then checks the output buffer is all-zero would lock in the intended behavior.

**Suggestion:**
```suggestion
Add a test: encrypt data, flip a tag byte, call wc_AesGcmDecrypt into a pre-filled (non-zero) output buffer, assert return == AES_GCM_AUTH_E AND the output buffer is all zero.
```

Frauschi (commented):
## 🐺 Skoll Code Review

### Posted findings
- [Low] **Output masking now runs a full pass over out[] on every successful decrypt** — `wolfcrypt/src/aes.c:11446-11449`
- [Low] **No test asserts output buffer is zeroed on tag mismatch for the C path** — `wolfcrypt/src/aes.c:11444-11449`

wolfcrypt/src/aes.c
danielinux: The folding-mask suggested approach is more expensive than single vectorizable path (measured 3.5% slower v. 0.4% impact of the original CT code).

Rejecting comment.

Frauschi (dismissed):
LGTM

wolfcrypt/test/test.c
```diff
@@ -19170,6 +19170,50 @@ static wc_test_ret_t aesgcm_setiv_test(Aes* enc, Aes* dec)
...
+    #if defined(WOLFSSL_ASYNC_CRYPT)
+        ret = wc_AsyncWait(ret, &dec->asyncDev, WC_ASYNC_FLAG_NONE);
+    #endif
+        resultT[0] ^= 0xB5;
+    #ifdef WOLFSSL_AESNI
+        dec->use_aesni = saved_use_aesni;
+    #endif
+        if (ret != WC_NO_ERR_TRACE(AES_GCM_AUTH_E))
+            ERROR_OUT(WC_TEST_RET_ENC_NC, out);
+    #if !defined(WOLFSSL_ARMASM) && !defined(WOLFSSL_ARMASM_NO_HW_CRYPTO) && \
```
Frauschi: **🟠 [High] Zero-check test fails on WC_AES_GCM_DEC_AUTH_EARLY builds**
*🚫 BLOCK*  `test`

The new negative test fills `resultP` with `0x5A` and then asserts every byte is `0` after a tampered-tag decrypt, on the assumption that the C path always clears `out` on auth failure. That assumption is false when `WC_AES_GCM_DEC_AUTH_EARLY` is defined. In that build the C path (`AES_GCM_decrypt_C`, aes.c:11352-11363) authenticates the tag *before* decryption and returns `AES_GCM_AUTH_E` early, so `out` is never written — this PR's own comment at aes.c:11448 states exactly this ("there the tag is checked before decryption, so out is never written on a mismatch"). Because the `#if !defined(...)` guard on the zero-check does NOT list `WC_AES_GCM_DEC_AUTH_EARLY`, the loop runs, finds `resultP[0] == 0x5A`, and calls `ERROR_OUT`. Result: the wolfCrypt self-test fails on any build configured with `WC_AES_GCM_DEC_AUTH_EARLY` (a supported, documented option per aes.c:66). The masking check in aes.c is correct; the defect is that the test does not match the code's own contract for the early-auth configuration.

**Suggestion:**
```suggestion
#if !defined(WOLFSSL_ARMASM) && !defined(WOLFSSL_ARMASM_NO_HW_CRYPTO) && \
    !defined(__aarch64__) && !defined(WOLFSSL_PPC64_ASM) && \
    !defined(WOLFSSL_PPC32_ASM) && !defined(STM32_CRYPTO_AES_GCM) && \
    !defined(WOLFSSL_AFALG_XILINX_AES) && !defined(WOLFSSL_XILINX_CRYPT) && \
    !defined(WOLFSSL_KCAPI_AES) && !defined(WOLFSSL_DEVCRYPTO_AES) && \
    !defined(WOLFSSL_ESP32_CRYPT) && !defined(WOLFSSL_IMXRT_DCP) && \
    !defined(WOLFSSL_SILABS_SE_ACCEL) && !defined(HAVE_FIPS) && \
    !defined(WC_AES_GCM_DEC_AUTH_EARLY)
```

wolfcrypt/test/test.c
```diff
@@ -19170,6 +19170,50 @@ static wc_test_ret_t aesgcm_setiv_test(Aes* enc, Aes* dec)
...
     if (XMEMCMP(p, resultP, sizeof(p)))
         ERROR_OUT(WC_TEST_RET_ENC_NC, out);
+
+    /* A tampered tag must fail decryption. On the software C path the output
+     * buffer is additionally cleared so unauthenticated plaintext is never
+     * released. The AES-NI/asm paths do not clear it, so the zero-check is
+     * limited to the C path, which is forced here by disabling AES-NI for
+     * this one call. */
+    {
+    #ifdef WOLFSSL_AESNI
```
Frauschi, line 19180: **🟡 [Medium] Zero-check denylist may fail under WOLFSSL_ASYNC_CRYPT with a real device**
*💡 SUGGEST*  `test`

The test forces the software path only by clearing `dec->use_aesni`, but the output-clearing guarantee it verifies is specific to `AES_GCM_decrypt_C`. When built with `WOLFSSL_ASYNC_CRYPT` against a real async device (non-`INVALID_DEVID`), `wc_AesGcmDecrypt` offloads to hardware regardless of `use_aesni`, and that path is not guaranteed to zero `out` on auth failure. The `#if !defined(...)` guard is a denylist of hardware macros and does not account for async offload, so the zero-check could fail on async-hardware builds. Toggling `use_aesni` also does not guarantee the C path on non-AES-NI accelerated builds not enumerated in the list.

**Suggestion:** Guard the zero-check additionally with `!defined(WOLFSSL_ASYNC_CRYPT)`, or gate it on `dec->devId == INVALID_DEVID`, so it only asserts clearing when the pure-C path is actually exercised.

Frauschi (changes requested):
## 🐺 Skoll Code Review

### Posted findings
- [High] **Zero-check test fails on WC_AES_GCM_DEC_AUTH_EARLY builds** — `wolfcrypt/test/test.c:19197-19214`
- [Medium] **Zero-check denylist may fail under WOLFSSL_ASYNC_CRYPT with a real device** — `wolfcrypt/test/test.c:19180-19214`

- [Low] Unrelated macro-list deletions bundled into AES-GCM fix
- [Low] Unconditional output pass runs even on auth success

</details>

danielinux:
Jenkins retest this please

Frauschi (approved):
LGTM

PR #10899, opened 2026-07-14
Zero-copy encrypt of AEAD application data

julek-wolfssl:
## Summary
`BuildMessage` copied the application data into the output buffer and encrypted it in place. For AEAD suites the cipher input is exactly the plaintext, so `Encrypt` can read it straight from the caller's buffer and write the record to the output buffer, skipping the copy.

- `EncryptDo` AES-GCM/CCM: `input == out` keeps the in-place record layout `[explicit IV | plaintext | tag space]`; `input != out` means input is the plaintext itself. ChaCha20-Poly1305 already reads only the plaintext from input. Other suites keep the copy: CBC/stream append MAC/pad to the cipher input, CID appends the real content type.
- Disabled for `WOLFSSL_ASYNC_CRYPT` and `WOLFSSL_THREADED_CRYPT`, where encryption can run after `BuildMessage` returns, and for debug hooks that expect the in-place layout. Define `WOLFSSL_BUILD_MSG_NO_ZERO_COPY` to force the old behavior.
- Measured with `examples/benchmark/dtls_bench -z` (DTLS 1.2, 1300 byte records, aesni + intelasm): +2.9% AES-256-GCM, +1.0% ChaCha20-Poly1305. The copy was ~3% of send-path CPU in perf.

Also fixes two bugs found while benchmarking with a larger `STATIC_BUFFER_LEN`:
- The `STATIC_BUFFER_LEN` bounds check had an unbalanced paren and compared enum constants in a preprocessor `#if`, where they evaluate to 0, so any user supplied value failed to compile. Keep the lower bound as a static assert. Drop the upper bound: values past the largest record are never used by the record layer but can be useful for memory layout reasons.
- `ShrinkInputBuffer` could run in the middle of record processing via `FreeHandshakeResources` on the first DTLS app data record, moving the record data and invalidating the idx based bookkeeping. The stale `curStartIdx` made the more-messages-per-record check dispatch a spurious second message over the AEAD tag bytes, handing garbage app data to the caller and corrupting the buffer state for the next read. Only reachable when the remaining record data fits in the static buffer, which it never does with the default `STATIC_BUFFER_LEN`. Only shrink between records; the deferred shrink happens at the next `wolfSSL_read`.

## Test plan
- [ ] CI passes
- [ ] Benchmark with `examples/benchmark/dtls_bench -z` shows no regression
- [ ] DTLS app-data read tests pass with a larger `STATIC_BUFFER_LEN`


julek-wolfssl:
retest this please

julek-wolfssl:
retest this please

julek-wolfssl:
retest this please history lost

SparkiDev (approved):

PR #10900, opened 2026-07-14
Bound remaining hangable operations in test scripts

julek-wolfssl:
## Summary
A test script that blocks forever burns the CI job's full timeout-minutes with no logs. e82ecdff93 and 5c5cbd3094 bounded the waited-on servers; this covers the remaining hang classes in `scripts/*.test`:

- Wrap foreground example client/server, `openssl s_client`, and `openssl ocsp` invocations in `timeout -s KILL 2m`. A client wedged before or without a live peer (e.g. blocked in first-seed entropy gathering, or DTLS with no reset from a dead peer) is not bounded by its peer's timeout.
- Add the macOS `timeout()` fallback shim to scripts that now use `timeout`.
- Bound the `get_first_free_port` scan loops (`nc -w 1`, 100-port cap).
- Add `-w 1` to the remaining `nc` probes and `dtls.test` UDP pcap markers.
- `ocsp-responder-openssl-interop.test`: bound the responder reap in cleanup: give each responder 5 s to exit after SIGTERM, then SIGKILL before waiting, so a wedged responder cannot hang the EXIT trap.
- `benchmark.test`: bound the clients but leave the `-i` servers unwrapped: the script ends them with `kill -6`, which `timeout(1)` does not forward, so wrapping would orphan the server.
- `trusted_peer`/`tls13`: kill the server with SIGTERM instead of SIGKILL in cleanup so the signal forwards through the timeout wrapper to the wrapped server.

## Test plan
- [ ] Affected `scripts/*.test` pass locally
- [ ] CI passes

### DIFF
--- a/scripts/openssl.test
+++ b/scripts/openssl.test
@@ -2,6 +2,12 @@
 
 # openssl.test
 
+# timeout(1) is GNU coreutils and absent on macOS; where it's missing, run the
+# command unbounded (the flaky hang this guards against is Linux-only CI).
+if ! command -v timeout >/dev/null 2>&1; then
+    timeout() { while [ "${1:-}" = "-s" ] || [ "${1:-}" = "-k" ]; do shift 2; done; shift; "$@"; }
+fi
+
 # Environment variables used:
 # OPENSSL            (openssl app to use)
 # OPENSSL_ENGINE_ID  (engine id if any i.e. "wolfengine")
@@ -363,13 +369,13 @@ do_wolfssl_client() {
         echo "#"
         echo "# $WOLFSSL_CLIENT -p $port -g $wolfssl_resume -l $wolfSuite -v $version $psk $adh \"$wolfssl_cert\" \"$wolfssl_key\" \"$wolfssl_caCert\" $crl"
         # shellcheck disable=SC2086
-        $WOLFSSL_CLIENT -p "$port" -g $wolfssl_resume -l "$wolfSuite" -v "$version" $psk $adh "$wolfssl_cert" "$wolfssl_key" "$wolfssl_caCert" $crl
+        timeout -s KILL 2m $WOLFSSL_CLIENT -p "$port" -g $wolfssl_resume -l "$wolfSuite" -v "$version" $psk $adh "$wolfssl_cert" "$wolfssl_key" "$wolfssl_caCert" $crl
     else
         echo "#"
         echo "# $WOLFSSL_CLIENT -p $port -g $wolfssl_resume -l $wolfSuite $psk $adh \"$wolfssl_cert\" \"$wolfssl_key\" \"$wolfssl_caCert\" $crl"
         # do all versions
         # shellcheck disable=SC2086
-        $WOLFSSL_CLIENT -p "$port" -g $wolfssl_resume -l "$wolfSuite" $psk $adh "$wolfssl_cert" "$wolfssl_key" "$wolfssl_caCert" $crl
+        timeout -s KILL 2m $WOLFSSL_CLIENT -p "$port" -g $wolfssl_resume -l "$wolfSuite" $psk $adh "$wolfssl_cert" "$wolfssl_key" "$wolfssl_caCert" $crl
     fi
 
     client_result=$?
@@ -425,11 +431,11 @@ do_openssl_client() {
     then
         echo "#"
         echo "# $OPENSSL s_client -connect localhost:$port -reconnect -legacy_renegotiation -cipher $cmpSuite $openssl_version $openssl_psk $openssl_cert1 \"$openssl_cert2\" $openssl_key1 \"$openssl_key2\" $openssl_caCert1 \"$openssl_caCert2\""
-        echo "Hello" | eval "$OPENSSL s_client -connect localhost:$port -reconnect -legacy_renegotiation -cipher $cmpSuite $openssl_version $openssl_psk $openssl_cert1 \"$openssl_cert2\" $openssl_key1 \"$openssl_key2\" $openssl_caCert1 \"$openssl_caCert2\""
+        echo "Hello" | eval "timeout -s KILL 2m $OPENSSL s_client -connect localhost:$port -reconnect -legacy_renegotiation -cipher $cmpSuite $openssl_version $openssl_psk $openssl_cert1 \"$openssl_cert2\" $openssl_key1 \"$openssl_key2\" $openssl_caCert1 \"$openssl_caCert2\""
     else
         echo "#"
         echo "# $OPENSSL s_client -connect localhost:$port -reconnect -legacy_renegotiation -ciphersuites=$cmpSuite $openssl_seclevel $openssl_version $openssl_psk $openssl_cert1 \"$openssl_cert2\" $openssl_key1 \"$openssl_key2\" $openssl_caCert1 \"$openssl_caCert2\""
-        echo "Hello" | eval "$OPENSSL s_client -connect localhost:$port -reconnect -legacy_renegotiation -ciphersuites=$cmpSuite $openssl_seclevel $openssl_version $openssl_psk $openssl_cert1 \"$openssl_cert2\" $openssl_key1 \"$openssl_key2\" $openssl_caCert1 \"$openssl_caCert2\""
+        echo "Hello" | eval "timeout -s KILL 2m $OPENSSL s_client -connect localhost:$port -reconnect -legacy_renegotiation -ciphersuites=$cmpSuite $openssl_seclevel $openssl_version $openssl_psk $openssl_cert1 \"$openssl_cert2\" $openssl_key1 \"$openssl_key2\" $openssl_caCert1 \"$openssl_caCert2\""
     fi
 
     client_result=$?
@@ -536,7 +542,7 @@ if [ "$wolf_certs" != "" ]
 then
     echo
     # Check if RSA certificates supported in wolfSSL
-    wolf_rsa=$($WOLFSSL_CLIENT -A "${CERT_DIR}/ca-cert.pem" 2>&1)
+    wolf_rsa=$(timeout -s KILL 2m $WOLFSSL_CLIENT -A "${CERT_DIR}/ca-cert.pem" 2>&1)
     case $wolf_rsa in
     *"ca file"*)
         echo "wolfSSL does not support RSA"
@@ -549,7 +555,7 @@ then
         echo "wolfSSL supports RSA"
     fi
     # Check if RSA-PSS certificates supported in wolfSSL
-    wolf_rsapss=$($WOLFSSL_CLIENT -A "${CERT_DIR}/rsapss/ca-rsapss.pem" 2>&1)
+    wolf_rsapss=$(timeout -s KILL 2m $WOLFSSL_CLIENT -A "${CERT_DIR}/rsapss/ca-rsapss.pem" 2>&1)
     case $wolf_rsapss in
     *"ca file"*)
         echo "wolfSSL does not support RSA-PSS"
@@ -562,7 +568,7 @@ then
         echo "wolfSSL supports RSA-PSS"
     fi
     # Check if ECC certificates supported in wolfSSL
-    wolf_ecc=$($WOLFSSL_CLIENT -A "${CERT_DIR}/ca-ecc-cert.pem" 2>&1)
+    wolf_ecc=$(timeout -s KILL 2m $WOLFSSL_CLIENT -A "${CERT_DIR}/ca-ecc-cert.pem" 2>&1)
     case $wolf_ecc in
     *"ca file"*)
         echo "wolfSSL does not support ECDSA"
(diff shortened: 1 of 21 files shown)


julek-wolfssl:
retest this please

julek-wolfssl:
retest this please

dgarske (commented):
## Skoll Code Review

**Findings:** 16 total — 3 posted, 13 skipped
**3 finding(s) not tied to a diff line** (full detail below)

#### New get_first_free_port 100-port cap uses exit 1 inside a command substitution, so it never terminates the script
**File:** `scripts/ocsp-stapling.test:288-291 (callers 298-300); scripts/ocsp-stapling2.test:271-274 (281-285); scripts/ocsp-stapling_tls13multi.test:289-292 (299-303); scripts/ocsp-stapling-with-ca-as-responder.test:216-219 (226-227); scripts/ocsp-stapling-with-wolfssl-responder.test:362-365 (372-376); scripts/ocsp-responder-openssl-interop.test:160-163 (237-241)`
**Function:** `get_first_free_port`
**Severity:** High

The new port-scan bound ends with `exit 1`, but in all six scripts `get_first_free_port` is invoked *only* inside a command substitution (`port1=$(get_first_free_port $base_port)`). `exit` therefore terminates only the command-substitution subshell; the parent script keeps running. I verified this empirically: a function that does `exit 1` inside `v=$(f)` leaves the parent alive with `v` empty and the script exiting 0. None of the six scripts sets `set -e` or `set -o pipefail` (verified by grep across all six), and none of the 25 call sites checks the assignment status. Consequence: the failing port variable becomes the empty string, and the next call computes `$((port1 + 1))`, which bash evaluates as 1 for an empty operand, so every subsequent port collapses toward privileged port 1. Servers are then launched with `-p` followed by the next flag (e.g. `-p -v`), and openssl URLs degrade to `http://127.0.0.1:/`. The run still eventually fails via `wait_for_readyFile`, but with a misleading `exited before creating ready file` message and after the automatic `RETRIES_REMAINING` re-execs — exactly the confusing failure mode this guard was added to prevent. The guard is currently dead code.

**Recommendation:** Make the failure visible to the caller. Minimal fix (the subshell already returns 1): append `|| exit 1` to every one of the 25 call sites across the six files. Cleaner: change the guard to `return 1` and have each caller do `|| { echo "ERROR: no free port" 1>&2; exit 1; }`.

_Referenced code: `scripts/ocsp-stapling.test:288-291 (callers 298-300); scripts/ocsp-stapling2.test:271-274 (281-285); scripts/ocsp-stapling_tls13multi.test:289-292 (299-303); scripts/ocsp-stapling-with-ca-as-responder.test:216-219 (226-227); scripts/ocsp-stapling-with-wolfssl-responder.test:362-365 (372-376); scripts/ocsp-responder-openssl-interop.test:160-163 (237-241)` (10 lines)_

---

#### macOS timeout() shim breaks $! for backgrounded servers, so cleanup no longer kills them (orphaned processes)
**File:** `scripts/trusted_peer.test:8-10,144,162,179,196,213,232,250,267,289 and cleanup at 98-104; scripts/tls13.test:128,144,166,192,209,240 and cleanup at 84-91`
**Function:** `timeout() shim / do_cleanup()`
**Severity:** High

The shim is a shell *function*. When a shell function is backgrounded (`timeout -s KILL 2m ./examples/server/server ... &`), bash forks a subshell to run it, so `$!` is the PID of that SUBSHELL, not of the server. `kill $server_pid` in `do_cleanup` then kills only the subshell and leaves the server running. I verified this empirically on bash 5.2: with the shim, `$!` resolved to a process with `comm=bash` while the real command was a separate PID; after `kill $!` the real child was still alive (orphan). With real GNU `timeout` the same test correctly killed the child, because timeout forwards SIGTERM (I confirmed that too — so the PR's `kill -9` -> `kill` change is itself correct on Linux). Bash does not apply its exec/no-fork optimisation to a function invoked asynchronously, so this is not version-specific, and `kill -9` on the subshell orphans the child just as `kill` does. Before this PR the servers were backgrounded directly, so `$!` was the real server PID and `kill -9 $server_pid` always worked. This only bites where the shim is active, i.e. macOS — exactly the platform the shim exists for. trusted_peer.test is the worst case: it had ZERO `timeout` uses before this PR (verified against HEAD~3) and now wraps 9 servers, including the `-i` infinite-accept server at line 289 that *only* ever stops via `do_cleanup`. It will leak on every macOS run.

**Recommendation:** Do not place a shell function between `&` and `$!`. Replace the function with a prefix variable that expands to nothing on macOS, so `$!` stays the real PID (I verified this works for both foreground and background use). Note that making the shim `exec "$@"` is NOT a valid fix — I tested it and in the foreground call sites it replaces the running script itself, terminating the test. Alternatively, leave the backgrounded servers unwrapped, as psk.test and benchmark.test already do.

_Referenced code: `scripts/trusted_peer.test:8-10,144,162,179,196,213,232,250,267,289 and cleanup at 98-104; scripts/tls13.test:128,144,166,192,209,240 and cleanup at 84-91` (6 lines)_

---

#### A timeout kill (137) is misread as "feature not enabled" and turned into a silent PASS (exit 0)
**File:** `scripts/psk.test:110-118, scripts/psk.test:138-146, scripts/trusted_peer.test:150-158`
**Function:** `n/a (top-level test flow)`
**Severity:** High

`timeout -s KILL 2m` returns **137**, not 124 — I verified this locally with coreutils (`timeout -s KILL 1 sleep 5` -> 137), because timeout signals its own process group and is itself SIGKILLed. These three sites feed that status straight into an `if [ $RESULT -ne 0 ]` branch whose action is to print a "not compiled in" style message and `exit 0`. So the exact hang this PR is meant to bound now produces a GREEN test result instead of a CI-level timeout failure. That is strictly *worse* observability than before the change: a real hang used to be visible as a stuck job, and now it is indistinguishable from "this feature isn't enabled in this build". The PR already added an explicit `>= 124` discriminator for the two version-probe pipes (commit 0b37af7a5f) but did not apply the same treatment to any of the plain `RESULT=$?` skip sites.

**Recommendation:** Insert an explicit timeout check before each skip branch, mirroring what the PR already did for the version probes. Applies to psk.test:114, psk.test:142 and trusted_peer.test:154.

_Referenced code: `scripts/psk.test:110-118, scripts/psk.test:138-146, scripts/trusted_peer.test:150-158` (9 lines)_

---

### Skipped findings
- [Medium] ~~dtlscid.test: client killed by the new 2m timeout is silently reported as PASS~~
- [Medium] ~~openssl.test: version probes treat a 2m timeout as "OpenSSL does not support this version" and silently skip the whole cipher matrix~~
- [Medium] ~~Negative tests accept a timeout kill as the expected failure, so a hang passes~~
- [Medium] ~~dtls.test: nc -u -w 1 adds exactly 1.00 s per PCAP marker (3 s per test), up to 4.2 hours in extended mode~~
- [Medium] ~~trusted_peer.test gives a section-lifetime -i server the same 2m budget as a single operation~~
- [Medium] ~~benchmark.test hard-caps a user-specified benchmark workload at 2 minutes and reports the abort as "Client failed!"~~
- [Low] ~~ocsp-stapling2.test: openssl s_server readiness loop left unbounded while its byte-identical twin was bounded~~
- [Low] ~~external.test: second external client connection left unbounded while the first was wrapped~~
- [Low] ~~google.test: a client killed by the new timeout becomes a hard FAILURE, bypassing the deliberate network-flakiness skip~~
- [Low] ~~Version-probe discriminator = 124 misattributes signal deaths and non-timeout statuses to a timeout~~
- [Low] ~~New exit 1 timeout branches bypass do_cleanup, unlike every other failure exit in these scripts~~
- [Info] ~~ocsp-responder-openssl-interop.test: responder reap escalation is sequential, up to 25 s per run~~
- [Info] ~~retry_with_backoff around a 2m timeout triples the worst-case wall time of the external test~~

dgarske (changes requested):
https://github.com/wolfSSL/wolfssl/pull/10900#pullrequestreview-4849256788

dgarske (dismissed):

julek-wolfssl:
retest this please

scripts/trusted_peer.test
```diff
@@ -227,10 +249,10 @@ if [[ $wrong_ca != *"ecc"* ]]; then
...
 port=0
-./examples/server/server -c "$server_cert" -k "$server_key" -R "$ready_file" -p $port &
+$TIMEOUT_KILL_2M ./examples/server/server -c "$server_cert" -k "$server_key" -R "$ready_file" -p $port &
 server_pid=$!
 create_port
-./examples/client/client -A "$wrong_ca" -E "$wrong_cert" -p $port
+$TIMEOUT_KILL_2M ./examples/client/client -A "$wrong_ca" -E "$wrong_cert" -p $port
 RESULT=$?
 remove_ready_file
 if [ $RESULT -eq 0 ]; then
```
philljj, line 263: Will a timeout status 137 trick this test into passing?

From Fenrir:

### HIGH-6: Trusted-peer negative tests accept timeout kills as certificate rejection
- **File:** `scripts/trusted_peer.test:255-262`

**Description:** The wrong-CA/wrong-peer test accepts every nonzero client result, including the newly introduced timeout status 137. The same pattern appears in the server-rejection case at lines 273-280 and the final multi-peer rejection at lines 323-329. Only the initial feature probe uses `timed_out`; these later hangs are therefore indistinguishable from the expected certificate failures and the script can report `ALL TESTS PASSED`.

scripts/openssl.test
```diff
@@ -550,7 +567,7 @@ then
         echo "wolfSSL supports RSA"
     fi
     # Check if RSA-PSS certificates supported in wolfSSL
-    wolf_rsapss=$($WOLFSSL_CLIENT -A "${CERT_DIR}/rsapss/ca-rsapss.pem" 2>&1)
+    wolf_rsapss=$("${TIMEOUT_KILL_2M[@]}" $WOLFSSL_CLIENT -A "${CERT_DIR}/rsapss/ca-rsapss.pem" 2>&1)
```
philljj, line 585: Do these need to have exit status checked, to avoid misinterpreting output? From Fenrir:

### HIGH-2: Certificate probe timeouts silently remove OpenSSL interop coverage
- **File:** `scripts/openssl.test:557-635`

**Description:** The PR wraps the wolfSSL certificate capability probes in `timeout`, but only examines their captured output and discards the exit status. A process killed before producing output leaves variables such as `wolf_rsapss`, `wolf_ecc`, `wolf_ed25519`, or `wolf_ed448` empty. Later conditions interpret an empty variable as lack of feature support and omit those interop cases, allowing a client deadlock to produce a successful run with reduced coverage.

scripts/openssl.test
```diff
@@ -871,7 +888,7 @@ do
 
         # double check that can actually do a sslv3 connection using
         # client-cert.pem to send but any file with EOF works
-        $OPENSSL s_client -ssl3 -no_ign_eof -host localhost -port "$openssl_port" < "${CERT_DIR}/client-cert.pem"
+        "${TIMEOUT_KILL_2M[@]}" $OPENSSL s_client -ssl3 -no_ign_eof -host localhost -port "$openssl_port" < "${CERT_DIR}/client-cert.pem"
         sslv3_sup=$?
         if [ "$sslv3_sup" != 0 ]
```
philljj, line 915: Will this misinterpret a timeout exit status?

### HIGH-3: OpenSSL protocol probe timeouts are misclassified as unsupported versions
- **File:** `scripts/openssl.test:891-929`

**Description:** The new timeout returns 137 when the SSLv3, TLS 1.0, or TLS 1.1 connection probe hangs. Each branch treats every nonzero result as `No OpenSSL support` and continues, while the script ultimately exits 0. Thus a hung `s_client` silently skips an entire protocol's interop matrix instead of failing the CI test.

scripts/psk.test
```diff
@@ -158,7 +190,7 @@ if [ $? -ne 0 ]; then
         ./examples/server/server -j -R "$ready_file" -p $port &
         server_pid=$!
         create_port
-        ./examples/client/client -x -p $port
+        $TIMEOUT_KILL_2M ./examples/client/client -x -p $port
         RESULT=$?
         remove_ready_file
         if [ $RESULT -eq 0 ]; then
```
philljj, line 201: Does this need to check for timeout exit status, so it doesn't confuse timeout with an expected auth failure?

scripts/tls13.test
```diff
@@ -142,10 +152,10 @@ echo ""
...
 port=0
-./examples/server/server -v 4 -R "$ready_file" -p $port -l TLS13-AES128-GCM-SHA256 &
+$TIMEOUT_KILL_2M ./examples/server/server -v 4 -R "$ready_file" -p $port -l TLS13-AES128-GCM-SHA256 &
 server_pid=$!
 create_port
-./examples/client/client -v 4 -p $port -l TLS13-AES256-GCM-SHA384
+$TIMEOUT_KILL_2M ./examples/client/client -v 4 -p $port -l TLS13-AES256-GCM-SHA384
 RESULT=$?
 remove_ready_file
 if [ $RESULT -eq 0 ]; then
```
philljj, line 166: Will this negative test get tricked by a timeout exit status?

philljj (changes requested):

julek-wolfssl:
@philljj 
Confirmed on all five, and fixed. timed_out() existed but was only used on the initial feature probe, so every later negative test read a timeout kill as the rejection it was testing for, and openssl.test never looked at probe exit status at all.

Each site now checks for 124/137 before interpreting the result: the three wrong-CA/wrong-peer cases in trusted_peer.test; psk.test's no-peer-cert case; tls13.test's cipher mismatch, mutual auth and both version-downgrade cases (the same pattern was in four more spots beyond the one you flagged); and in openssl.test a new abort_if_timed_out helper covering the seven certificate capability probes and the SSLv3/TLS 1.0/TLS 1.1 connection probes. Verified the guard fires on a real timeout -s KILL (137) while a normal non-zero exit still passes through, and that the probes still detect RSA/RSA-PSS/ECDSA/Ed25519/Ed448 correctly.

philljj (approved):

PR #10901, opened 2026-07-14
Add SLH-DSA support for the TLS 1.3 and DTLS 1.3 handshake

Frauschi:
### Summary

Adds **SLH-DSA (SPHINCS+, FIPS 205)** as an entity authentication algorithm for the TLS 1.3 and DTLS 1.3 handshake, following [`draft-reddy-tls-slhdsa`](https://datatracker.ietf.org/doc/draft-reddy-tls-slhdsa/). An SLH-DSA leaf key can now be used to sign (and a peer's SLH-DSA leaf to verify) the `CertificateVerify` message, for both server and mutual authentication.

All twelve parameter sets are wired into the handshake — the **SHAKE** and **SHA2** hash families across **128/192/256** in both the **f** (fast) and **s** (small) variants — with per-parameter-set gating so a build only offers, accepts, and maps the variants actually compiled in. Test certificates and configs cover the 128f and 128s sets for both hash families.

### Motivation

SLH-DSA is a stateless hash-based signature scheme standardized in FIPS 205. Its appeal is a conservative security foundation (hash functions only), but its signatures are **large** — from ~7.8 KB (128s) up to ~49 KB (256f) — which stresses several fixed-size buffers in the TLS/cert code that were sized for classical signatures. This PR adds the protocol support **and** the buffer-sizing work needed to make it viable, including on memory-constrained targets.

#### Handshake integration
- Map the SLH-DSA signature schemes to/from the wire in the `signature_algorithms` extension and `CertificateVerify`.
- Sign/verify `CertificateVerify` with an SLH-DSA entity key.
- Load SLH-DSA private keys and certificates and resolve their key/param types (`ssl_load.c`, `ssl.c`, `ssl_api_pk.c`, `asn.c`, `wc_slhdsa.h`).
- Advertisement, mapping, and OID handling are gated **per parameter set**, so partial builds (e.g. `--enable-slhdsa=sha2-128s`) never advertise or map a variant they cannot perform.

#### Streamed CertificateVerify send (large signatures)
When the `CertificateVerify` body exceeds a single record, the signature is generated once into a **connection-level buffer** and emitted **one record at a time**, so the output buffer never has to hold the whole signature.

- Resumes correctly across a non-blocking `WANT_WRITE` **without recomputing** the randomized signature (the already-sent records are bound into the transcript, so the resumed send must emit the identical bytes — this is a correctness requirement, not an optimization).
- Gated by `WOLFSSL_TLS13_STREAM_CERT_VERIFY` (TLS 1.3, non-async, PQC signatures). It is **algorithm-neutral** — it also applies to ML-DSA (or any scheme) whenever the body exceeds the negotiated `max_fragment_length`.
- **DTLS 1.3** and **`WOLFSSL_ASYNC_CRYPT`** builds keep the existing in-place fragmented path (see *Scope / carve-outs*).
- **Dual-algorithm** (`WOLFSSL_DUAL_ALG_CERTS`, BOTH) bodies stream as well: the combined two-signature body is sized from per-signature upper bounds and the exact length is recorded after signing (trailing slack is never sent).

Peak send-side memory is **≈ one signature + one record** (the record is at most 16 KB, or the negotiated `max_fragment_length` if smaller), versus the **~2×** a naive in-place send holds (the assembled body plus a copy). The reduction is largest exactly where it matters for constrained devices — a small `max_fragment_length`, where the second term is sub-KB and the peak is essentially the signature itself. It also stops the output buffer from staying inflated to the whole signature for the connection's lifetime.

#### Buffer sizing

Three concerns that previously shared one constant are now separated, because they have genuinely different requirements:

- **`MAX_X509_SIZE`** sizes the static per-certificate slot embedded by value in every cached session. It is a flat 9 KB for post-quantum builds and is deliberately **not** derived from a signature size: a certificate too large for a slot is simply not recorded in the chain. It is also not derived from the enabled ML-DSA level, since `MLDSA_MAX_SIG_SIZE` tracks the compiled level and a level-restricted build would otherwise silently drop certificates a full build keeps.
- **`MAX_CERT_WIRE_SZ`** (new) bounds the largest certificate that may appear in a handshake message, sized from the enabled post-quantum signatures, and feeds `MAX_CERTIFICATE_SZ`.
- **`MAX_CERT_MSG_DEPTH`** (new) is the chain depth assumed when sizing the certificate message. `MAX_CHAIN_DEPTH` bounds how deep a chain may be *verified*, whereas this sizes a buffer an *unauthenticated* peer can make us allocate, so it is trimmed to 5 once a single certificate can be large. Classic and ML-DSA builds keep the historical depth.

The `CertificateVerify` buffers are sized from the actual signature length instead of the worst-case `WC_MAX_CERT_VERIFY_SZ` (which balloons to ~50 KB with SLH-DSA). `WC_MAX_CERT_VERIFY_SZ` is **retained** for API compatibility; its growth is documented as a behavioral change in `README.md`.

The dual-algorithm PreTBS buffer keeps a tight estimate (certificate size minus both signatures) as a first attempt, since that allocation matters on constrained targets, and retries once at a size the canonical re-encode cannot exceed if the estimate is short. `wc_GeneratePreTBS` reports an encoder failure as `WOLFSSL_FAILURE`, which is `0`, so a non-positive result is now treated as an error rather than skipping `ConfirmSignature` and reading as a verified alternative signature.

#### Protocol version gating

SLH-DSA is defined for TLS 1.3 only. The schemes are no longer advertised to a TLS 1.2 peer, `MatchSigAlgo` and `PickHashSigAlgo` pin an SLH-DSA certificate to its exact parameter set and to TLS 1.3, and the TLS 1.2 `CertificateVerify` rejects any post-quantum key with `SIG_TYPE_E` rather than emitting a record sized for a classic signature.

#### Device-held private keys

An SLH-DSA private key that lives in a device and is referenced by id or label is now supported. The parameter set cannot be recovered from a device-side identifier, so it is carried from the key type down to `wc_SlhDsaKey_Init_id` / `wc_SlhDsaKey_Init_label`.

#### Tests and certificates
- SLH-DSA entity (client/server) certificates for the SHAKE and SHA2 128f/128s parameter sets, plus generation-script updates.
- TLS 1.3 and DTLS 1.3 entity-cert `CertificateVerify` test configs covering the **fragmented (128f)** and **single-record (128s)** send paths for both hash families, wired into `suites.c` (server-auth + mutual-auth scenarios).

### Memory impact

Measured on `--enable-all --enable-dilithium --enable-slhdsa --enable-experimental` (all parameter sets, so `WC_SLHDSA_MAX_SIG_LEN` = 49856):

| | Before separating the constants | After |
|---|---|---|
| `MAX_X509_SIZE` | 53952 | 9216 |
| `sizeof(WOLFSSL_X509_CHAIN)` | 485,608 | 83,016 |
| `sizeof(WOLFSSL_SESSION)` | 973,984 | 168,736 |
| Default session cache (3 x 11 = 33 sessions, `SESSION_CERTS`) | ~32 MB | ~5.6 MB |
| `MAX_CERT_WIRE_SZ` | n/a | 53952 |
| `MAX_HANDSHAKE_SZ` | 485,598 | 269,778 |

So roughly **26 MB less static BSS** on a default build, while the wire limit still accommodates an SLH-DSA certificate chain. `BIG_SESSION_CACHE` (7 x 2861 = 20,027 sessions) would not have linked at the earlier session size.

Both new constants scale with what is compiled in rather than with the family maximum: a `--enable-slhdsa=128s` build gets `MAX_CERT_WIRE_SZ` = 11952 and `MAX_HANDSHAKE_SZ` = 107,598. Classic builds are unchanged at 2048 / 18,462, and ML-DSA-only builds at 9216 / 82,974.

### Enabling

```sh
./configure --enable-slhdsa            # all parameter sets (SHAKE)
./configure --enable-slhdsa=sha2       # SHA2 family
./configure --enable-slhdsa=128s,128f  # specific parameter sets
```

TLS 1.3 is required (enabled by default). The SHA2 family additionally requires SHA-256, SHA-512, and HMAC.

### Scope / carve-outs

- **DTLS 1.3** is intentionally **not** on the streaming path. DTLS fragments at the handshake layer (offset/length) via `Dtls13HandshakeSend`, and its reliability model must retain the full message for retransmission until ACKed — so streaming yields no memory benefit there. A code comment in `Dtls13SendFragmented` records the follow-up (sign directly into the fragments buffer to avoid the transient second copy; the retransmit copy is a hard floor).
- **`WOLFSSL_ASYNC_CRYPT`** builds keep the existing in-place fragmented path. Async introduces a record-AEAD `WC_PENDING_E` resume point whose cursor-timing requirement conflicts with the non-blocking `WANT_WRITE` cursor; the in-place path already handles both correctly. Async targets (HW-offload servers) are not memory-constrained, so there is no beneficiary for the added complexity.

### Testing

- Full `./tests/unit.test` suite: **pass** (normal handshakes + all SLH-DSA configs + DTLS + crypto API tests).
- Streamed path (128f, server + mutual auth): pass.
- Single-record path (128s) and DTLS 1.3 in-place path: pass.
- Full library build with `-Werror`: clean; verified the streaming feature-macro
  **off** path (classical/async builds) also compiles clean.
- Negative coverage: a `CertificateVerify` that does not verify against the peer certificate's key aborts the handshake with `SIG_VERIFY_E`; a TLS 1.2 handshake presenting an SLH-DSA client certificate is rejected.
- CI configs added to `.github/configs/pq-all.json`: a SHA2-only SLH-DSA build, and an async-crypto build with dual-algorithm certificates. The latter is the only configuration that compiles the in-place fragmented send, and it caught a static-assertion build break in `Scv13Args` that no existing job exercised.

### Known limitations / follow-ups

- Test coverage exercises the 128f/128s fragmentation boundaries; the **192f/256f** parameter sets (deepest fragmentation, up to ~49 KB) are not yet covered at the TLS layer. Adding a 192f/256f entity config (guarded on the corresponding `WOLFSSL_SLHDSA_PARAM_*` macros) would strengthen the streaming/fragmentation coverage.
- The dual-algorithm (`WOLFSSL_DUAL_ALG_CERTS`, CKS `BOTH`) `CertificateVerify` sizing is compiled by CI but not executed by any test. `EncodeDualSigAlg` only accepts pairs where one side is post-quantum, while the existing dual-alg fixtures generate ECC-native + ECC-alt certificates, so covering it needs a new hybrid fixture (ECC native + ML-DSA alternative) for both root and server.
- The `WANT_WRITE` tests interrupt the `CertificateVerify` flight and assert the retried send emits identical bytes, but where the flight is flushed as a single write the retry happens below `SendTls13CertificateVerify`, so they do not by themselves reach the `fragOffset` resume branch.
- DTLS memory reduction (sign into the fragments buffer) is noted in-code as a possible future optimization.


Frauschi:
Jenkins retest this please

src/tls13.c
```diff
@@ -10714,29 +10934,144 @@ static int SendTls13CertificateVerify(WOLFSSL* ssl)
...
+                            ret = AddPacketInfo(ssl, "CertificateVerify",
+                                        handshake, output, recSz, WRITE_PROTO, 0,
+                                        ssl->heap);
+                            if (ret != 0)
+                                goto exit_scv;
+                        }
+                    }
+                #endif
+
+                    ssl->buffers.outputBuffer.length += (word32)recSz;
```
Frauschi, line 11268: this is intended, extended the comment explaining this

wolfcrypt/src/asn.c
```diff
@@ -24354,17 +24354,27 @@ int ParseCertRelative(DecodedCert* cert, int type, int verify, void* cm,
...
-                #else
-                    byte *der = (byte*)XMALLOC(WC_MAX_CERT_VERIFY_SZ, cert->heap,
-                                            DYNAMIC_TYPE_DCERT);
+                    word32 derSz;
+                    byte*  der;
+
+                    /* PreTBS upper bound = certificate size minus both
+                     * signatures. Validate before the unsigned subtraction to
+                     * avoid wrap-around on a malformed certificate. */
+                    if (cert->sigLength + (word32)cert->altSigValLen >=
```
Frauschi: fixed

src/tls13.c
```diff
@@ -10783,6 +11244,19 @@ static int SendTls13CertificateVerify(WOLFSSL* ssl)
     } /* switch(ssl->options.asyncState) */
 
 exit_scv:
+#ifdef WOLFSSL_TLS13_STREAM_CERT_VERIFY
+    /* A streamed CertificateVerify keeps its assembled body across a
+     * non-blocking WANT_WRITE so the send resumes without recomputing the
+     * signature; release it on completion or on any real error. */
+    if (ret != WC_NO_ERR_TRACE(WANT_WRITE) &&
```
Frauschi, line 11471: fixed

tests/suites.c
```diff
@@ -1370,6 +1370,103 @@ int SuiteTest(int argc, char** argv)
         goto exit;
     }
 #endif
+#if defined(WOLFSSL_HAVE_SLHDSA) && defined(WOLFSSL_SLHDSA_PARAM_128F) && \
```
Frauschi: Noting this as a deliberate scope decision rather than a gap we'll close in this PR.

The two structural axes that affect the TLS/DTLS 1.3 send path are already covered by the 128-bit configs:

- **Single-record vs. fragmented CertificateVerify** — 128s (~7.8 KB signature, single record) vs. 128f (~17 KB signature, multi-record) exercise both the non-fragmented path and the `fragOffset` streaming loop.
- **SHAKE vs. SHA2 families** — both are covered for the 128 parameter sets.

The 192/256 parameter sets differ only in signature size, which flows through the same size arithmetic (`BuildTls13Message`, `CheckAvailableSize`, the `fragOffset` bounds) that 128f already drives under fragmentation. They add larger constants, not a new code path, so the marginal coverage is low relative to the cost of generating and carrying 256f root+leaf certs, `.conf` files, and harness blocks.

Leaving the 192/256 handshake integration configs out for now; happy to revisit in a follow-up if we want max-size fragmentation coverage.

src/tls13.c
```diff
@@ -10996,6 +11486,41 @@ static int decodeMlDsaKey(WOLFSSL* ssl, int level)
...
+    word32 tmpIdx = 0;
+
+    if (ssl->peerSlhDsaKeyPresent)
+        return INVALID_PARAMETER;
+
+    keyRet = AllocKey(ssl, DYNAMIC_TYPE_SLHDSA, (void**)&ssl->peerSlhDsaKey);
+    if (keyRet != 0)
+        return PEER_KEY_ERROR;
+
+    ssl->peerSlhDsaKeyPresent = 1;
```
Frauschi: Fixed in the latest push. `peerSlhDsaKeyPresent` is now set only after both `wc_SlhDsaKey_Init()` and `wc_SlhDsaKey_PublicKeyDecode()` succeed, so a failure after the intermediate `wc_SlhDsaKey_Free()` no longer leaves the flag set on a half-initialised key. The allocated key block is still freed unconditionally by `FreeHandshakeResources()`/`ResourceFree()`, so the error paths remain leak-free.

src/internal.c
```diff
@@ -4848,22 +5146,37 @@ void DecodeSigAlg(const byte* input, byte* hashAlgo, byte* hsType)
...
             }
             else if (input[1] == MLDSA_87_SA_MINOR) {
                 *hsType = mldsa_87_sa_algo;
                 *hashAlgo = sha512_mac;
+                break;
+            }
+        #endif /* WOLFSSL_HAVE_MLDSA */
+        #ifdef WOLFSSL_HAVE_SLHDSA
+            slhType = SlhDsaSigMinorToType(input[1]);
+            if (slhType != (byte)invalid_sa_algo) {
```
Frauschi, line 5212: Fixed. `*hashAlgo` is now unconditionally initialised to `no_mac` at the top of `DecodeSigAlg`, alongside `*hsType = invalid_sa_algo`. This guarantees a defined hash value on every path — covering the new SLH-DSA case as well as the pre-existing FALCON/SM2 cases that shared the "unwritten `*hashAlgo` on unrecognised minor" pattern.

src/tls13.c
```diff
@@ -10199,14 +10301,155 @@ static int SendTls13CertificateVerify(WOLFSSL* ssl)
...
+
+            #ifdef WOLFSSL_DTLS13
+                if (ssl->options.dtls) {
+                    rem -= recordLayerHdrExtra + DTLS_HANDSHAKE_EXTRA;
+                    args->idx += recordLayerHdrExtra + DTLS_HANDSHAKE_EXTRA;
+                    args->verify += recordLayerHdrExtra + DTLS_HANDSHAKE_EXTRA;
+                }
+            #endif /* WOLFSSL_DTLS13 */
+
+                if (rem < 0 || (int)args->sigLen > rem) {
```
Frauschi, line 10584: Keeping this as an intentional defense-in-depth bound. `CheckAvailableSize()` does guarantee capacity for `args->sendSz`, so today the branch is redundant — but it's a cheap re-check immediately before writing `args->sigLen` signature bytes into the buffer, and it mirrors the guard on the streaming path. Added a comment documenting the intent.

src/tls13.c
```diff
@@ -11025,6 +11528,48 @@ static int decodeMlDsaKey(WOLFSSL* ssl, int level)
...
+ * SLH-DSA public key. Convert it into a usable public key. */
+static int decodeSlhDsaKey(WOLFSSL* ssl, int param)
+{
+    int keyRet;
+    word32 tmpIdx = 0;
+
+    if (ssl->peerSlhDsaKeyPresent)
+        return INVALID_PARAMETER;
+
+    keyRet = AllocKey(ssl, DYNAMIC_TYPE_SLHDSA, (void**)&ssl->peerSlhDsaKey);
```
Frauschi: After review we're keeping SLH-DSA consistent with the other signature algorithms on this path rather than adding an inline free.

On this same peer-key decode failure path, none of RSA/ECC/Ed25519/Falcon/ML-DSA free the peer key inline — the allocated `peerSlhDsaKey` (~1.1 KB, `sizeof(SlhDsaKey)`) is released by `FreeHandshakeResources()`/`ResourceFree()` at session teardown, which follows immediately once `PEER_KEY_ERROR` aborts the handshake. Peer-key decode runs once per handshake, so a failure can't accumulate allocations within a live session; it's a single ~1.1 KB block freed at the imminent teardown, not an unbounded leak. `peerSlhDsaKeyPresent` is already only set after full init, so the teardown `FreeKey()` cleans up correctly regardless.

Adding an inline free only for SLH-DSA would diverge from every other algorithm without a real memory benefit, so we're leaving cleanup to the common teardown path. A uniform inline-free across all algorithms could be a separate follow-up if desired.

src/dtls13.c
```diff
@@ -1152,6 +1152,16 @@ static int Dtls13SendFragmented(WOLFSSL* ssl, byte* message, word16 length,
```
Frauschi, line 1152: Acknowledged — this is the known transient ~2x peak-heap trade-off on the DTLS 1.3 CertificateVerify path, already documented in the code comment in `Dtls13SendFragmented()` (src/dtls13.c). An incremental DTLS fragmentation path analogous to `WOLFSSL_TLS13_STREAM_CERT_VERIFY` is a larger change that's out of scope for this PR; leaving it as a documented design trade-off for now, to revisit in a follow-up.

src/tls13.c
Frauschi: current implementation is aligned with the other algorithms, so won't change

src/dtls13.c
Frauschi: This is a known limitation for now, as adding this functionality would require a large refactoring of the DTLS path, which is out-of-scope for this PR. It is documented in the code as well

Frauschi:
Jenkins retest this please

src/tls13.c
```diff
@@ -10199,14 +10301,158 @@ static int SendTls13CertificateVerify(WOLFSSL* ssl)
...
+            }
+            else {
+                /* Reserve the output buffer, sized from the actual signature
+                 * length(s) plus record/handshake framing and encryption
+                 * slack. */
+                args->sendSz = RECORD_HEADER_SZ + HANDSHAKE_HEADER_SZ +
+                               HASH_SIG_SIZE + VERIFY_HEADER + (int)args->sigLen;
+            #ifdef WOLFSSL_DUAL_ALG_CERTS
+                args->sendSz += (int)args->altSigLen;
+            #endif
```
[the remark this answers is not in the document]
Frauschi: fixed

src/tls13.c
```diff
@@ -11025,6 +11528,48 @@ static int decodeMlDsaKey(WOLFSSL* ssl, int level)
 }
 #endif /* WOLFSSL_HAVE_MLDSA */
 
+#ifdef WOLFSSL_HAVE_SLHDSA
+/* ssl->peerCert->sapkiDer is the alternative public key. Hopefully it is an
+ * SLH-DSA public key. Convert it into a usable public key. */
```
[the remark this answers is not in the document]
Frauschi: fixed

src/tls13.c
```diff
@@ -11637,6 +12250,30 @@ static int DoTls13CertificateVerify(WOLFSSL* ssl, byte* input,
                 }
             }
         #endif /* WOLFSSL_HAVE_MLDSA */
+        #if defined(WOLFSSL_HAVE_SLHDSA)
+            if (IsSlhDsaSigAlgo(ssl->options.peerSigAlgo) &&
```
Frauschi, line 12372: fixed by adding a check

src/tls13.c
```diff
@@ -10996,6 +11536,48 @@ static int decodeMlDsaKey(WOLFSSL* ssl, int level)
...
+    if (ssl->peerSlhDsaKeyPresent)
+        return INVALID_PARAMETER;
+
+    keyRet = AllocKey(ssl, DYNAMIC_TYPE_SLHDSA, (void**)&ssl->peerSlhDsaKey);
+    if (keyRet != 0)
+        return PEER_KEY_ERROR;
+
+    /* AllocKey initialised the key with a placeholder parameter set; re-init
+     * with the parameter set negotiated in the CertificateVerify. */
+    wc_SlhDsaKey_Free(ssl->peerSlhDsaKey);
```
Frauschi: won't fix. This is a no-issue and aligned with all other algorithms that do the same

Frauschi:
The one CI failure ([Multiple compilers and versions / Compiler test](https://github.com/wolfSSL/wolfssl/actions/runs/29748076544/job/88371082411?pr=10901)) will be resolved by #10951 .

Frauschi:
Jenkins retest this please

src/ssl_load.c
```diff
@@ -1045,6 +1045,82 @@ static int ProcessBufferTryDecodeMlDsa(WOLFSSL_CTX* ctx, WOLFSSL* ssl,
...
+    /* Allocate an SLH-DSA key to parse into. */
+    key = (SlhDsaKey*)XMALLOC(sizeof(SlhDsaKey), heap, DYNAMIC_TYPE_SLHDSA);
+    if (key == NULL) {
+        return MEMORY_E;
+    }
+
+    /* Initialise with an always-present placeholder parameter set;
+     * wc_SlhDsaKey_PrivateKeyDecode selects the real one from the key's
+     * algorithm OID. */
+    ret = wc_SlhDsaKey_Init(key, WC_SLHDSA_DEFAULT_PARAM, heap, INVALID_DEVID);
```
Frauschi, line 1085: fixed

src/tls13.c
```diff
@@ -8842,14 +8844,79 @@ static WC_INLINE void EncodeSigAlg(const WOLFSSL * ssl, byte hashAlgo,
             output[1] = MLDSA_87_SA_MINOR;
             break;
 #endif
+#ifdef WOLFSSL_HAVE_SLHDSA
+    #ifdef WOLFSSL_SLHDSA_SHA2
```
Frauschi: fixed

src/tls13.c
```diff
@@ -10224,14 +10326,155 @@ static int SendTls13CertificateVerify(WOLFSSL* ssl)
...
+                               HASH_SIG_SIZE + VERIFY_HEADER + (int)args->sigLen;
+            #ifdef WOLFSSL_DUAL_ALG_CERTS
+                args->sendSz += (int)args->altSigLen;
+            #endif
+            #ifdef WOLFSSL_DTLS13
+                if (ssl->options.dtls)
+                    args->sendSz += recordLayerHdrExtra + DTLS_HANDSHAKE_EXTRA;
+            #endif /* WOLFSSL_DTLS13 */
+                /* Framing and always-on encryption expansion. */
+                args->sendSz += MAX_MSG_EXTRA + MAX_MSG_EXTRA;
```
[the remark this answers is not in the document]
Frauschi: fixed

src/tls13.c
```diff
@@ -10165,23 +10274,16 @@ static int SendTls13CertificateVerify(WOLFSSL* ssl)
...
+            word32 maxFrag;
+#endif
+            int doStream = 0;
 
+            /* Decode the private key first so the output buffer can be sized
+             * to the actual signature length rather than the worst-case
+             * WC_MAX_CERT_VERIFY_SZ (very large when SLH-DSA is enabled). */
             if (ssl->buffers.key == NULL) {
             #ifdef HAVE_PK_CALLBACKS
                 if (wolfSSL_CTX_IsPrivatePkSet(ssl->ctx))
```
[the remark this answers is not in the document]
Frauschi: fixed

Frauschi:
Jenkins retest this please

dgarske (commented):
## Skoll Code Review

**Findings:** 31 total — 17 posted, 14 skipped
**11 finding(s) posted as inline comments** (see file-level comments below)
**6 finding(s) not tied to a diff line** (full detail below)

### Posted findings
- [Critical] **Dual-alg PreTBS tight bound turns a non-fail-safe error path into an alternative-signature verification bypass** — `wolfcrypt/src/asn.c:24606-24652 and 24678-24723`
- [High] **TLS 1.2 SendCertificateVerify buffer under-sized for PQC keys - heap over-read hashed into the transcript and sent on the wire** — `src/internal.c:36690-36698`
- [High] **MAX_X509_SIZE tied to WC_SLHDSA_MAX_SIG_LEN inflates the static session cache from 2.3 MB to 15.3 MB of BSS** — `wolfssl/internal.h:2003-2028`
- [High] **New SLH-DSA entity .conf suite blocks are not excluded from verify-only builds - breaks the existing all-pq-verify-only-noasm CI job** — `tests/suites.c:1373,1389,1405,1424,1441,1456`
- [Medium] **SLH-DSA cert paths ignore SlhDsaParamToType() / wc_SlhDsaKey_PublicSizeFromParam() failure returns; two macro families can diverge** — `src/ssl_load.c:1988-2001 and 2229-2236`
- [Medium] **ssl_load.c accepts SLH-DSA alternative (dual-alg) public keys that tls13.c always rejects** — `src/ssl_load.c:2214-2239`
- [Medium] **Public WC_MAX_CERT_VERIFY_SZ silently jumps from 6000 to 50880 when SLH-DSA is enabled** — `wolfssl/wolfcrypt/asn.h:939-948`
- [Medium] **No CI configuration pairs SLH-DSA with --enable-asynccrypt, so the in-place fragmented CertificateVerify path is compiled nowhere** — `.github/workflows/pq-all.yml:150`
- [Medium] **ProcessBufferTryDecodeSlhDsa calls wc_SlhDsaKey_Free() on a key whose Init failed before zeroization** — `src/ssl_load.c:1081-1121`
- [Low] **New SLHDSA_ALL_NO_* macros are unprefixed in an installed public header** — `wolfssl/wolfcrypt/wc_slhdsa.h:471-536`
- [Low] **wc_SlhDsaParamToOid() return value stored into *keyFormat without a range check** — `src/ssl_load.c:1102-1106`

#### WOLFSSL_MAX_SIGALGO lean-build carve-out not extended for SLH-DSA; ADD_HASH_SIG_ALGO has no bounds check
**File:** `wolfssl/internal.h:1922-1933`
**Function:** `InitSuitesHashSigAlgo / ADD_HASH_SIG_ALGO`
**Severity:** Medium

The new `SIG_SLHDSA` block appends up to 12 two-byte entries (24 bytes) to `hashSigAlgo`. I verified that `ADD_HASH_SIG_ALGO` (`src/internal.c:3416-3423`) writes `out[*inOutIdx]` and `out[*inOutIdx + 1]` with **no bound check at all**, and that `InitSuitesHashSigAlgo` ends with a bare `*len = idx;` and no cap — so the caller's array size is the only protection. Consumers include `byte hashSigAlgo[WOLFSSL_MAX_SIGALGO]` (`internal.h:2431`, `:6402`) and the stack arrays at `internal.c:27322` and `:27348`.

`WOLFSSL_MAX_SIGALGO` is 128 in the default case but **44** for `(WOLFSSL_LEANPSK || WOLFSSL_LEANTLS) && !HAVE_FALCON && !WOLFSSL_HAVE_MLDSA`. The comment on that guard states its intent explicitly: lean builds keep the list small "unless they are post-quantum builds: those want to inter-op with OQS's OpenSSL that sends a lot more sigalgs, so they fall through to the larger default." SLH-DSA is exactly such a build and adds **four times** as many sigalgs as ML-DSA (12 vs 3), yet `!defined(WOLFSSL_HAVE_SLHDSA)` was not added to the condition.

I confirmed the default (non-lean) build is safe: worst case is ~36-39 entries = 72-78 bytes <= 128. I was **not** able to construct the overflow from any supported `./configure` invocation (`--enable-leantls` sets `-DNO_RSA -DNO_SHA`, keeping the count at ~42 B), so I am rating this Medium rather than High. But a `user_settings.h` lean build with ECC+brainpool+Ed25519+Ed448+SM2+SLH-DSA reaches 46 bytes and overruns, and user_settings.h builds are first-class in wolfSSL. The guard is objectively inconsistent with its own documented…

**Recommendation:** Add the one-line guard. Independently, consider giving ADD_HASH_SIG_ALGO a real bound (pass the array size, skip or error when `*inOutIdx + 2 > size`) so future sigalg families cannot silently overflow a fixed array.

_Referenced code: `wolfssl/internal.h:1922-1927` (6 lines)_

---

#### MAX_CERTIFICATE_SZ / MAX_HANDSHAKE_SZ grow to ~486 KB, raising the unauthenticated handshake-reassembly cap 6.6x
**File:** `wolfssl/internal.h:2052-2063`
**Function:** `MAX_CERTIFICATE_SZ`
**Severity:** Medium

`MAX_CERTIFICATE_SZ = CERT_HEADER_SZ + (MAX_X509_SIZE + CERT_HEADER_SZ) * MAX_CHAIN_DEPTH`, and `MAX_HANDSHAKE_SZ` aliases it. With the new MAX_X509_SIZE this goes from 73,758 to 485,598 (probe-measured). Those macros are the ceiling used to accept and buffer an inbound handshake message **before any authentication**: `src/dtls13.c:1916` gates DTLS 1.3 fragment reassembly, `src/tls13.c:14957` and `src/internal.c:20121/20158/20829` gate TLS message buffering, and `src/internal.c:17601` gates the certificate list size. A remote peer can therefore cause ~486 KB of buffering per connection/epoch pre-authentication instead of ~72 KB. Over DTLS (connectionless, spoofable source) that is a 6.6x memory-amplification improvement for an attacker.

**Recommendation:** Same root cause as the MAX_X509_SIZE finding — fixing that one (a separate constant for the static cache plus a tighter parse ceiling) resolves this too. At minimum note the new value in the PR description so DTLS deployments can pin MAX_HANDSHAKE_SZ.

_Referenced code: `wolfssl/internal.h:2052-2056` (5 lines)_

---

#### PickHashSigAlgo has no SLH-DSA arm; SLH-DSA silently unnegotiable without WOLFSSL_SHA512
**File:** `src/internal.c:31864-31890`
**Function:** `PickHashSigAlgo`
**Severity:** Medium

`DecodeSigAlg` (`internal.c:5170-5177`) reports `sha512_mac` for every SLH-DSA parameter set, including the SHAKE-128 sets that never touch SHA-512. `PickHashSigAlgo` has explicit early-accept arms keyed on `ssl->pkCurveOID` for Ed25519, Ed448, Falcon and ML-DSA that short-circuit before the hash switch, but the PR adds no SLH-DSA arm. SLH-DSA therefore falls through to the `switch (hashAlgo)` at `internal.c:31973-32035`, where `case sha512_mac:` is compiled only `#ifdef WOLFSSL_SHA512`; otherwise the entry hits `default: /* Support for hash algorithm not compiled in. */ break;` and SLH-DSA can never be selected. `ssl->pkCurveOID` is populated for SLH-DSA certs (the PR extends the pkCurveOID `#if` guards in internal.h), so the information for an explicit arm is available. Net effect: a SHAKE-only SLH-DSA build with SHA-512 disabled builds fine but never negotiates SLH-DSA, with no diagnostic.

**Recommendation:** Add the arm alongside the Falcon/ML-DSA ones, and add a build-matrix entry for `--enable-slhdsa` with SHA-512 disabled so the gap is caught in CI.

_Referenced code: `src/internal.c:31864-31872` (9 lines)_

---

#### Dual-alg streaming path over-claims the ECC signature output capacity by 2 bytes
**File:** `src/tls13.c:10604-10606 and 10704-10713`
**Function:** `SendTls13CertificateVerify`
**Severity:** Medium

For ECC, `args->sigLen` is recomputed as the remaining buffer: `args->sigLen = (word32)args->sendSz - args->idx - HASH_SIG_SIZE - VERIFY_HEADER;`.

On the **streaming** path `args->sendSz = bodySz` and `args->idx = 0` exactly (no `MAX_MSG_EXTRA` slack), where `bodySz = HASH_SIG_SIZE + VERIFY_HEADER + sigLen + altSigLen + 2*OPAQUE16_LEN` for dual-alg. So the recomputed capacity is `sigLen + altSigLen + 2*OPAQUE16_LEN`. But `sigOut` is advanced by an extra `OPAQUE16_LEN` for dual-alg (`tls13.c:10712`), leaving only `sigLen + altSigLen + OPAQUE16_LEN` bytes to the end of the allocation. The capacity handed to `wc_ecc_sign_hash` is therefore **2 bytes larger than the buffer actually has**.

The in-place path has the same 2-byte over-claim but absorbs it in the `MAX_MSG_EXTRA` slack included in its reservation; the streaming allocation is exactly `bodySz` with no slack. Not reachable today because ECC never writes anywhere near its maximum DER signature size, and the combination requires native-ECC + huge-alt + a small max_fragment_length. The PR also notes the dual-alg streamed path is untested.

**Recommendation:** Correct the arithmetic so the capacity passed to the sign call never exceeds the real remaining buffer, since the streaming allocation has no slack to absorb it.

_Referenced code: `src/tls13.c:10604-10606 and 10704-10713` (8 lines)_

---

#### (void)checkKeySz added only in the native cert path, not the alt cert path
**File:** `src/ssl_load.c:2001 and 2229-2237`
**Function:** `ProcessBufferCertPublicKey / ProcessBufferCertAltPublicKey`
**Severity:** Low

`ProcessBufferCertPublicKey`'s SLH-DSA case adds `(void)checkKeySz;` with a comment explaining it is needed so a build where SLH-DSA is the only certificate signature algorithm does not warn about an unused parameter. `ProcessBufferCertAltPublicKey` takes the same `int checkKeySz` parameter and its new SLH-DSA case omits the cast, yet it has the same exposure: its only unconditional arm is `case 0:` (no OID), which does not reference checkKeySz. In an SLH-DSA-only build with WOLFSSL_DUAL_ALG_CERTS and no RSA/ECC/Ed/Falcon/ML-DSA, checkKeySz becomes unused there and `-Wunused-parameter -Werror` would fail. Also, placing the cast inside a switch case is unusual; the file's convention is `(void)x;` near the top of the function.

**Recommendation:** Make the two cert paths consistent.

---

#### keySz semantics differ between the SLH-DSA private-key path (4n) and cert path (2n)
**File:** `src/ssl_load.c:1096 and 1999`
**Function:** `ProcessBufferTryDecodeSlhDsa / ProcessBufferCertPublicKey`
**Severity:** Low

`ProcessBufferTryDecodeSlhDsa` stores `wc_SlhDsaKey_PrivateSize(key)` (4*n, e.g. 64 for the 128-bit sets) into `ssl->buffers.keySz` / `ctx->privateKeySz`, while `ProcessBufferCertPublicKey` stores `wc_SlhDsaKey_PublicSizeFromParam()` (2*n, e.g. 32) into the same field. Whichever of certificate or key is loaded last wins. The ML-DSA arms of both functions store the same constant (`WC_MLDSA_xx_KEY_SIZE`) from both paths, so this is a new divergence. Impact is currently cosmetic — the field is only fed to `InitSuites()`/`InitSuitesHashSigAlgo()` where keySz gates RSA-PSS sig-alg advertisement — but a key/cert-dependent value is a latent surprise.

**Recommendation:** Make both SLH-DSA paths report the same size.

_Referenced code: `src/ssl_load.c:1096 and 1999` (2 lines)_

---

### Skipped findings
- [Medium] ~~SLH-DSA signature schemes are advertised in TLS 1.2 signature_algorithms / CertificateRequest~~
- [Medium] ~~want-write CertificateVerify test never asserts the interruption actually fired, and binds to the CV only by record position~~
- [Medium] ~~Dual-algorithm (WOLFSSL_DUAL_ALG_CERTS BOTH) streamed CertificateVerify is not covered by any test~~
- [Medium] ~~fragActive async-resume optimization silently depends on GrowOutputBuffer's copy semantics~~
- [Low] ~~wolfSSL_set_SSL_CTX does not propagate the new haveSlhDsaSig flag~~
- [Low] ~~Only 128-bit SLH-DSA parameter sets have entity-cert fixtures - 192f/256f deepest fragmentation is untested~~
- [Low] ~~SlhDsaSigMinorToType / SlhDsaParamToType round-trip is only exercised for the two 128-bit parameter sets~~
- [Low] ~~Dual-alg PreTBS path now requires a heap; WOLFSSL_NO_MALLOC builds lose alternative-signature verification~~
- [Low] ~~gen-slhdsa-mldsa-certs.sh still writes 20 .der files that are neither tracked nor in EXTRA_DIST~~
- [Low] ~~New brace-only scopes to declare slhParam depart from the file's C89 declaration style~~
- [Low] ~~ProcessPeerCertDecodeKey SLH-DSA arm omits WOLFSSL_ERROR_VERBOSE and any minimum-key-size policy~~
- [Low] ~~Cross-family re-init in wc_SlhDsaKey_PrivateKeyDecode discards WOLF_PRIVATE_KEY_ID id/label~~
- [Low] ~~Line-continuation backslash misaligned in the new test gating #if~~
- [Info] ~~haveSlhDsaSig is set but never read; SIG_ALL makes it moot~~

src/internal.c
```diff
@@ -36219,7 +36687,12 @@ int SendCertificateVerify(WOLFSSL* ssl)
                 return 0;  /* sent blank cert, can't verify */
             }
 
-            args->sendSz = WC_MAX_CERT_VERIFY_SZ + MAX_MSG_EXTRA;
+            /* TLS 1.2 and earlier only ever produce a classic signature
```
dgarske, line 37076: **🔴 [High] TLS 1.2 SendCertificateVerify buffer under-sized for PQC keys - heap over-read hashed into the transcript and sent on…**

The diff replaces `args->sendSz = WC_MAX_CERT_VERIFY_SZ + MAX_MSG_EXTRA` with `MAX_ENCODED_CLASSIC_SIG_SZ + MAX_MSG_EXTRA`, justified by the comment "TLS 1.2 and earlier only ever produce a classic signature (RSA/ECC/EdDSA); PQC signatures are TLS 1.3 only". **That invariant does not hold**, and I traced the full reachable path myself:

1. `InitSuitesHashSigAlgo` emits the Falcon/ML-DSA/SLH-DSA sigalgs with **no `tls1_3` gate**, and `GetServerCertReqHashSigAlgo` (`internal.c:27312`) copies `suites->hashSigAlgo` verbatim into the TLS 1.2 CertificateRequest.
2. `MatchSigAlgo` (`internal.c:31631-31670`) and `PickHashSigAlgo` (`internal.c:31864-31890`) have explicit Falcon and ML-DSA arms keyed on `ssl->pkCurveOID` that sit **above** the first `IsAtLeastTLSv1_3(ssl->version)` check (line 31676) — I verified there is no version gate on them, unlike the ECDSA/SM2 arms immediately below which do have one.
3. `DoCertificateRequest` sets `sendVerify = SEND_CERT` (`internal.c:34246`).
4. In `SendCertificateVerify`, `DecodePrivateKey` takes the Falcon/ML-DSA branch and returns `args->length` = the PQC signature size (Falcon-512 690, ML-DSA-44/65/87 = 2420/3309/4627).
5. The TLS 1.2 sign chain only has `DYNAMIC_TYPE_RSA/ECC/ED25519/ED448` arms, so `args->sigAlgo` stays 0 and **no signature is ever written** — but `args->length` is still the PQC size.
6. `TLS_ASYNC_END` calls `SendHandshakeMsg(ssl, args->output, args->length + args->extraSz + VERIFY_HEADER, ...)`, which `HashRaw()`s and `XMEMCPY()`s that many bytes out of the allocation (`internal.c:11723, 11737-11743`).

With RSA…

**Fix:** Either hard-fail before the buffer can be used with a PQC `ssl->hsType`, or size `args->input` from `args->length` after `DecodePrivateKey` (move the allocation into TLS_ASYNC_BUILD). Add a regression test doing TLS 1.2 mutual auth with an ML-DSA client cert — it should fail cleanly, not over-read.

wolfssl/internal.h
```diff
@@ -1981,8 +2001,26 @@ WOLFSSL_LOCAL int NamedGroupIsPqcHybrid(int group);
 #endif
 
 #ifndef MAX_X509_SIZE
```
dgarske, line 2009: **🔴 [High] MAX_X509_SIZE tied to WC_SLHDSA_MAX_SIG_LEN inflates the static session cache from 2.3 MB to 15.3 MB of BSS**

`MAX_X509_SIZE` becomes `(WC_SLHDSA_MAX_SIG_LEN + 4096)` = 53952 whenever `WOLFSSL_HAVE_SLHDSA` is defined — the SHAKE-256f worst case, regardless of which parameter sets are actually used at runtime. It is not only a parse limit: it sizes a **static** array. I verified the chain myself: `wolfssl/internal.h:4790` `typedef struct x509_buffer { int length; byte buffer[MAX_X509_SIZE]; }` -> `internal.h:4797` `struct WOLFSSL_X509_CHAIN { int count; x509_buffer certs[MAX_CHAIN_DEPTH]; }` (MAX_CHAIN_DEPTH=9, `internal.h:2033`) -> `internal.h:4941` `WOLFSSL_X509_CHAIN chain;` inside WOLFSSL_SESSION under SESSION_CERTS -> `src/ssl_sess.c:111` `static WC_THREADSHARED SessionRow SessionCache[SESSION_ROWS];` with sessions embedded by value.

Probe-measured against this tree:
| | baseline | with SLH-DSA |
|---|---|---|
| MAX_X509_SIZE | 8192 | 53952 |
| sizeof(WOLFSSL_SESSION) | 74,008 | 485,848 |
| default cache (11x3=33) | 2.33 MB | **15.29 MB** |
| MEDIUM (211x5=1055) | 74 MB | **489 MB** |
| BIG (2861x7=20027) | 1.4 GB | **9.3 GB** (will not link/run) |

`SESSION_CERTS` is not exotic: `configure.ac:1588` turns it on for `--enable-all-crypto` (so `--enable-all`), `configure.ac:5256-5265` for nginx/openvpn/lighty/netsnmp/strongswan/hitch/mosquitto **and unconditionally for TLS1.3+PSK**, and `configure.ac:1795` for the wolfguard preset. So `--enable-all --enable-all-quantum-crypto`, or plain `--enable-slhdsa --enable-tls13 --enable-psk`, yields a 15 MB BSS regression and makes the MEDIUM/BIG cache presets unusable. Raising the parse limit is necessary (an SLH-DSA-signed cert does not…

**Fix:** Pick one: (a) introduce a separate smaller constant for `x509_buffer`, leaving MAX_X509_SIZE for parse/reassembly bounds; (b) force SESSION_CACHE_DYNAMIC_MEM when WOLFSSL_HAVE_SLHDSA && SESSION_CERTS; or (c) at minimum `#error` when SLH-DSA is combined with SESSION_CERTS and a non-default cache preset. Quantify the BSS delta in the PR description either way.

tests/suites.c
dgarske: **🔴 [High] New SLH-DSA entity .conf suite blocks are not excluded from verify-only builds - breaks the existing…**

All six new .conf harness blocks gate only on `WOLFSSL_HAVE_SLHDSA` + the parameter macros + `WOLFSSL_TLS13`/`WOLFSSL_DTLS13`. They do **not** exclude `WOLFSSL_SLHDSA_VERIFY_ONLY`. Unlike the pre-existing SLH-DSA .conf tests (whose leaf is ML-DSA), these configs use an SLH-DSA **leaf** whose private key the example server must load and sign with. In a verify-only build `wc_slhdsa.c` compiles out the private-key import and sign paths, so the server aborts before the handshake.

`.github/workflows/pq-all.yml` already contains an entry that hits this exactly: `all-pq-verify-only-noasm` = `--enable-all` (which turns on TLS 1.3 **and** DTLS 1.3) plus `--enable-slhdsa=yes,small,verify-only` (which defines `WOLFSSL_SLHDSA_PARAM_128S` and `_128F`). That selects 4 of the 6 new blocks.

VERIFIED EMPIRICALLY by building a clean export of this branch with that configuration — `wolfssl/options.h` confirms `WOLFSSL_SLHDSA_VERIFY_ONLY`, `WOLFSSL_SLHDSA_PARAM_128S`, `_128F`, `WOLFSSL_TLS13` and `WOLFSSL_DTLS13` are all defined, and:
```
$ ./tests/unit.test tests/test-tls13-slhdsa-entity.conf
wolfSSL error: can't load server private key file, check file and run from wolfSSL home dir
EXIT=1
```
For contrast, the same four .conf files all pass in a signing build (`--enable-slhdsa=128f,128s`), and both SHA2 configs pass in the new `--enable-slhdsa=sha2` CI config — so the failure is specifically the verify-only axis. The pre-existing block at `tests/suites.c:1329` (ML-DSA leaf) has the identical structural gap and was reproduced the same way.

**Fix:** Add `!defined(WOLFSSL_SLHDSA_VERIFY_ONLY)` to all six new `#if` guards. Consider fixing the analogous pre-existing gap at line 1329/1359 (needs `!defined(WOLFSSL_MLDSA_VERIFY_ONLY)`) in the same change so the pq-all verify-only job goes green.

src/ssl_load.c
```diff
@@ -1826,6 +1971,37 @@ static int ProcessBufferCertPublicKey(WOLFSSL_CTX* ctx, WOLFSSL* ssl,
...
+        case SLH_DSA_SHAKE_192Fk:
+        case SLH_DSA_SHAKE_256Sk:
+        case SLH_DSA_SHAKE_256Fk:
+        {
+            int slhParam = wc_SlhDsaOidToParam((int)cert->keyOID);
+            if (slhParam < 0) {
+                ret = ALGO_ID_E;
+                break;
+            }
+            keyType = SlhDsaParamToType(slhParam);
```
dgarske, line 2008: **🟠 [Medium] SLH-DSA cert paths ignore SlhDsaParamToType() / wc_SlhDsaKey_PublicSizeFromParam() failure returns; two macro families…**

Both new cert cases do `keyType = SlhDsaParamToType(slhParam); keySz = wc_SlhDsaKey_PublicSizeFromParam(...)` with **no check on either result**, then fall through to the unconditional store into `ssl->buffers.keyType`/`keySz` with `ret == 0`.

The two lookups are gated on different macro families, and I confirmed the divergence is real and reachable: `wc_SlhDsaOidToParam` keys off `SLHDSA_BUILT_*`, which `wolfcrypt/src/asn.c:4956` defines as `!defined(WOLFSSL_SLHDSA_PARAM_NO_128S)` — it does **not** consider `NO_SMALL`. The new `SlhDsaParamToType` (`src/internal.c:5018`) keys off the positive `WOLFSSL_SLHDSA_PARAM_128S`, which `wolfssl/wolfcrypt/wc_slhdsa.h:122-126` defines as `!NO_128S && !NO_128 && !NO_SMALL`. Nothing derives `NO_128S` from `NO_SMALL`. So under `-DWOLFSSL_SLHDSA_PARAM_NO_SMALL` (which `configure.ac:2333` emits for a fast-only build), `wc_SlhDsaOidToParam(SLH_DSA_SHAKE_128Sk)` returns `SLHDSA_SHAKE128S` and passes the `slhParam < 0` check, while `SlhDsaParamToType()` falls to its default and returns `(byte)invalid_sa_algo == 255`.

The cert then loads successfully with `ssl->buffers.keyType == 255`; `DecodePrivateKey` (`internal.c:32857`) matches neither the ML-DSA `keyType == 0` arm nor `IsSlhDsaSigAlgo()`, so the failure surfaces much later as a generic algorithm error. `wc_SlhDsaKey_PublicSizeFromParam()` likewise returns `NOT_COMPILED_IN` (negative) on its default arm and that value would be stored straight into `keySz`. `ProcessBufferTryDecodeSlhDsa` in the same file **does** check this, so the three SLH-DSA entry points are inconsistent.

**Fix:** Validate both lookups in both functions, matching ProcessBufferTryDecodeSlhDsa. Separately, consider deriving `WOLFSSL_SLHDSA_PARAM_NO_\<variant>` from `NO_SMALL`/`NO_FAST`/`NO_128`/`NO_192`/`NO_256` in wc_slhdsa.h so the two macro families cannot diverge.

src/ssl_load.c
```diff
@@ -2036,6 +2212,31 @@ static int ProcessBufferCertAltPublicKey(WOLFSSL_CTX* ctx, WOLFSSL* ssl,
             }
             break;
     #endif /* WOLFSSL_HAVE_MLDSA */
```
dgarske, line 2238: **🟠 [Medium] ssl_load.c accepts SLH-DSA alternative (dual-alg) public keys that tls13.c always rejects**

The new SLH-DSA case in the alt-key switch makes a dual-algorithm certificate with an SLH-DSA SAPKI load successfully (previously it hit the `default:` arm and returned `NOT_COMPILED_IN` at load time). But the same PR hard-rejects SLH-DSA alt keys at handshake time: `DoTls13CertificateVerify` (`src/tls13.c:11787-11806`) has an explicit SLH-DSA arm that does `ERROR_OUT(ALGO_ID_E, exit_dcv)` with the comment "SLH-DSA is not supported as an alternative (dual-algorithm / CKS) key", and `SendTls13CertificateVerify`'s alt-sig-algo list still enumerates only ecc/falcon/mldsa/rsa. Net effect: the application successfully configures a certificate/key pair that can never complete a handshake, and the diagnostic moves from a clear load-time `NOT_COMPILED_IN` to a late, opaque `ALGO_ID_E`.

**Fix:** Make load-time and handshake-time policy agree on SLH-DSA as an alternative key — either reject at load, or complete the alt-key support in both Send/DoTls13CertificateVerify.

wolfssl/wolfcrypt/asn.h
```diff
@@ -936,7 +936,16 @@ extern const WOLFSSL_ObjectInfo wolfssl_object_info[];
     #endif
 #endif
 
-#if defined(HAVE_FALCON) || defined(WOLFSSL_HAVE_MLDSA)
+/* Maximum size of a CertificateVerify signature buffer. Retained as public API
```
dgarske, line 939: **🟠 [Medium] Public WC_MAX_CERT_VERIFY_SZ silently jumps from 6000 to 50880 when SLH-DSA is enabled**

The PR's claim that the macro is no longer used internally was verified: `grep -rn WC_MAX_CERT_VERIFY_SZ --include=*.c --include=*.h .` finds no remaining in-tree *uses* — the only hits outside asn.h are comments (`src/tls13.c:10267,10286`; `src/internal.c:36693`; `renesas_tsip_util.c:1714`). There is no in-tree stack-array blowup, and `asn.h:70` already includes `wc_slhdsa.h` under the same guard so there is no standalone-include build break.

The residual risk is downstream: `WC_MAX_CERT_VERIFY_SZ` is a **public** macro whose historical purpose was exactly `byte buf[WC_MAX_CERT_VERIFY_SZ]`, and it now evaluates to 50880 for any `--enable-slhdsa` build. Any application or port that kept such a stack buffer gets a ~50 KB frame with no compile-time signal.

**Fix:** Cheap to do and removes an invisible 50 KB stack hazard from every downstream consumer of a public header. If the value must change, add a `#pragma message` so downstream stack arrays are caught.

.github/workflows/pq-all.yml
```diff
@@ -147,6 +147,9 @@ jobs:
             "--enable-slhdsa=yes,small,verify-only",
             "--enable-dilithium=yes,small,verify-only", "--disable-qt",
             "CPPFLAGS=-pedantic -Wdeclaration-after-statement -Wnull-dereference -DWOLFCRYPT_TEST_LINT -DNO_WOLFSSL_CIPHER_SUITE_TEST -DTEST_LIBWOLFSSL_SOURCES_INCLUSION_SEQUENCE -DWOLFSSL_MLDSA_VERIFY_SMALL_MEM -DWOLFSSL_MLDSA_NO_LARGE_CODE"]},
+          {"name": "slhdsa-sha2-only", "minutes": 2,
```
dgarske: **🟠 [Medium] No CI configuration pairs SLH-DSA with --enable-asynccrypt, so the in-place fragmented CertificateVerify path is…**

`wolfssl/internal.h:5100` defines `WOLFSSL_TLS13_STREAM_CERT_VERIFY` only when `!defined(WOLFSSL_ASYNC_CRYPT)`; async builds therefore take the older in-place fragmented CertificateVerify path — the one the PR description itself calls out as a carve-out. Grepping `.github/workflows` for `slhdsa` matches only `pq-all.yml` and `wolfCrypt-Wconversion.yml`, and no entry in `async.yml` enables SLH-DSA (its `--enable-all` entries do not turn SLH-DSA on — `configure.ac` only defaults `enable_slhdsa` under `--enable-all-quantum-crypto`). So no job in the repository ever compiles the in-place fragmented path with a >16 KB signature, including the new `args->frag`/`fragOffset`/`fragActive` async-resume bookkeeping this PR adds. The new .conf files would provide that coverage for free if a matrix entry existed.

**Fix:** Add one matrix entry combining --enable-asynccrypt with --enable-slhdsa. The six new .conf tests then run against the in-place fragmented path with no additional test code.

src/ssl_load.c
```diff
@@ -1045,6 +1045,84 @@ static int ProcessBufferTryDecodeMlDsa(WOLFSSL_CTX* ctx, WOLFSSL* ssl,
...
+    /* Allocate an SLH-DSA key to parse into. */
+    key = (SlhDsaKey*)XMALLOC(sizeof(SlhDsaKey), heap, DYNAMIC_TYPE_SLHDSA);
+    if (key == NULL) {
+        return MEMORY_E;
+    }
+
+    /* Initialise with an always-present placeholder parameter set;
+     * wc_SlhDsaKey_PrivateKeyDecode selects the real one from the key's
+     * algorithm OID. */
+    ret = wc_SlhDsaKey_Init(key, WC_SLHDSA_DEFAULT_PARAM, heap, INVALID_DEVID);
```
dgarske, line 1085: **🟠 [Medium] ProcessBufferTryDecodeSlhDsa calls wc_SlhDsaKey_Free() on a key whose Init failed before zeroization**

The new function frees the key unconditionally, **outside** the `if (ret == 0)` Init-success block. The adjacent `ProcessBufferTryDecodeMlDsa` (`ssl_load.c:1039`) deliberately keeps `wc_MlDsaKey_Free()` inside that block. `wc_SlhDsaKey_Init()` (`wolfcrypt/src/wc_slhdsa.c:6655`) returns `BAD_FUNC_ARG` (key NULL) or `NOT_COMPILED_IN` (param set not in `SlhDsaParams[]`) **before** it reaches `XMEMSET(key, 0, sizeof(SlhDsaKey))`. On that path `key` is still raw XMALLOC memory, so the caller's `wc_SlhDsaKey_Free()` reads an uninitialized `key->params` and, if the garbage pointer is non-NULL, dereferences it in `ForceZero(key->sk, key->params->n * 2)` and `SLHDSA_IS_SHA2(key->params->param)`.

It is unreachable today only because `WC_SLHDSA_DEFAULT_PARAM` is derived from the `WOLFSSL_SLHDSA_PARAM_*` macros so it is always present in `SlhDsaParams[]` — a non-obvious invariant spanning two files. Note the in-code comment above the Free is factually wrong for that path: it claims the Free "is a no-op once the key's parameter set has been cleared", but on early-return Init failure the parameter set was never cleared nor set. The other failure modes (hash-init failure after the XMEMSET, and the family-switch re-Init inside `wc_SlhDsaKey_PrivateKeyDecode`) were traced and are safe.

**Fix:** Match the adjacent ML-DSA function: free the key only when Init succeeded. Alternatively, have wc_SlhDsaKey_Init() zeroize the key before its parameter lookup so a failed Init always leaves a Free-safe object.

wolfssl/wolfcrypt/wc_slhdsa.h
```diff
@@ -468,18 +468,85 @@
 
 #endif /* WOLFSSL_SLHDSA_SHA2 */
 
+/* ======== Combined family-absence guards ======== */
```
dgarske, line 539: **🔵 [Low] New SLHDSA_ALL_NO_* macros are unprefixed in an installed public header**

Eleven new object-like macros (`SLHDSA_ALL_NO_256/192/SMALL/FAST/128/256F/256S/192F/192S/128F/128S`) are defined in the installed public header without the `WOLFSSL_` / `WC_` prefix used by every neighbouring guard (`WOLFSSL_SLHDSA_PARAM_NO_*`, `WC_SLHDSA_*`). They leak into every translation unit that includes `wolfssl/wolfcrypt/asn.h`. The derivation logic itself was verified correct — `WOLFSSL_SLHDSA_NO_SHA2` is a real macro (`wc_slhdsa.h:326`), every `wc_slhdsa.c` maximum is monotone non-decreasing versus the old guards so nothing can be under-sized, and the SHAKE size constants are unconditional so there is no SHA2-only build break. This is purely namespace hygiene.

**Fix:** Rename before merge; the macros are new so there is no compatibility cost.

src/ssl_load.c
```diff
@@ -1045,6 +1045,84 @@ static int ProcessBufferTryDecodeMlDsa(WOLFSSL_CTX* ctx, WOLFSSL* ssl,
...
+                ret = ALGO_ID_E;
+            }
+            else {
+                keySizeTemp = wc_SlhDsaKey_PrivateSize(key);
+                if (keySizeTemp <= 0) {
+                    ret = ALGO_ID_E;
+                }
+            }
+
+            if (ret == 0) {
```
dgarske, line 1106: **🔵 [Low] wc_SlhDsaParamToOid() return value stored into *keyFormat without a range check**

`*keyFormat = wc_SlhDsaParamToOid((enum SlhDsaParam)param);` takes the return unchecked. `wc_SlhDsaParamToOid()` (`wolfcrypt/src/asn.c:5128`) returns `NOT_COMPILED_IN` for a known-but-disabled variant and `BAD_FUNC_ARG` for an unknown param — both negative. A negative `*keyFormat` would be propagated with `ret == 0`, and `ProcessBufferTryDecode`'s trailing sanity check (`*keyFormat == 0 || matchAnyKey == 0`) would not catch it, so the key would load with a nonsense format that later switch statements silently ignore. Unreachable today (param came from `key->params->param` after a successful decode), but the project style rule is to check every return code, and the sibling ML-DSA function does check its `mldsa_get_oid_sum()` result.

**Fix:** Check the OID lookup result before storing it.

wolfcrypt/src/asn.c
```diff
@@ -24606,17 +24606,30 @@ int ParseCertRelative(DecodedCert* cert, int type, int verify, void* cm,
             #ifdef WOLFSSL_DUAL_ALG_CERTS
```
dgarske, line 24687: **🟣 [Critical] Dual-alg PreTBS tight bound turns a non-fail-safe error path into an alternative-signature verification bypass**

Both new hunks replace the fixed `WC_MAX_CERT_VERIFY_SZ` PreTBS buffer with `derSz = cert->maxIdx - cert->sigLength - cert->altSigValLen`. That is not a provable upper bound on what `wc_GeneratePreTBS()` writes, because it does not copy the original TBSCertificate — it **re-encodes** it canonically (`CopyDecodedToX509` -> `wolfssl_x509_make_der` -> `MakeAnyCert`). The produced length is a function of wolfSSL's encoder, not of the input-length arithmetic. Instrumented measurement on `test_dual_alg_support`: `derSz=829` vs an actual PreTBS of `776` — only **53 bytes of headroom**, which is just the outer SEQUENCE header + signatureAlgorithm + BIT STRING header + the altSignatureValue extension wrapper. Anything that grows the re-encode (SPKI algorithm parameters NULL added where the original omitted them, extension normalisation, issuer/subject substitution, random-serial generation when `serialSz==0`) consumes it.

There is **no heap overflow** — `MakeAnyCert` bounds-checks (`if ((ret == 0) && (sz > derSz)) ret = BUFFER_E;` at `asn.c:30179`). The defect is what happens on that failure, which I verified line-by-line myself:
1. `wolfssl_x509_make_der` converts `BUFFER_E` to `WOLFSSL_FAILURE`, and `WOLFSSL_FAILURE == 0` (`wolfssl/ssl.h:3173`).
2. `wc_GeneratePreTBS` (`src/x509.c:8867-8870`) does `ret = wolfssl_x509_make_der(...); if (ret == WOLFSSL_SUCCESS) ret = derSz;` — a non-SUCCESS value propagates unchanged, so it **returns 0**.
3. `ParseCertRelative` does `if (ret > 0) ret = ConfirmSignature(...)` — **skipped** — then `if (ret != 0) { return ret; } else {…

**Fix:** Apply BOTH changes before merge: (1) make the path fail-closed on `ret <= 0`, and (2) add explicit headroom to `derSz` (e.g. `+ MAX_ALGO_SZ + MAX_SEQ_SZ + MAX_LENGTH_SZ`), or retry once with a larger buffer on BUFFER_E. Fixing only the sizing leaves the bypass latent for any third-party dual-alg certificate whose canonical re-encode exceeds the ~50-byte margin; fixing only the `ret == 0` handling turns the tight bound into a spurious verification failure. Add a unit test that forces `wc_GeneratePreTBS` to fail and asserts `ParseCertRelative` returns non-zero.

---
> **Note:** Referenced line (`wolfcrypt/src/asn.c:24606-24652 and 24678-24723`) is outside the diff hunk. Comment anchored to nearest changed region.

dgarske (changes requested):
Please fix merge conflict too.

https://github.com/wolfSSL/wolfssl/pull/10901#pullrequestreview-4849192930

src/internal.c
Frauschi: Confirmed and fixed, and I reproduced it under ASAN before touching anything: an ML-DSA-65 client cert over TLS 1.2 gave `heap-buffer-overflow, READ of size 64, 0 bytes after 614-byte region`, allocated at `internal.c:36992` and read at `SendHandshakeMsg -> HashRaw -> Sha256Update`. Reverting only that one line removed it, so the regression was exactly as you described.

Took your first option: no signature scheme below TLS 1.3 covers a post-quantum key, so `SendCertificateVerify` now rejects Falcon/ML-DSA/SLH-DSA with `SIG_TYPE_E` rather than sizing a record for a case that must never happen. Added the version gating you implied as well - `MatchSigAlgo` and `PickHashSigAlgo` now pin an SLH-DSA certificate to its exact parameter set and to TLS 1.3, and the schemes are no longer advertised to a TLS 1.2 peer.

Regression test added (`test_slhdsa_tls12_client_cert_rejected`): TLS 1.2 mutual auth with a PQC client certificate must fail cleanly. It asserts `INVALID_PARAMETER`, from `DoCertificateRequest` when `PickHashSigAlgo` finds no usable scheme. Proven non-vacuous - disabling the `MatchSigAlgo` arm makes the test fail, and the error then becomes `SIG_TYPE_E` from the `SendCertificateVerify` guard, so both layers are exercised independently.

wolfssl/internal.h
Frauschi: Confirmed, and went with your option (a). `MAX_X509_SIZE` is now a fixed 9 KB and a new `MAX_CERT_WIRE_SZ` carries the post-quantum-derived bound for `MAX_CERTIFICATE_SZ`, so the wire limit still accommodates an SLH-DSA chain while the static per-certificate slot does not scale with a signature.

Measured on `--enable-all` + SLH-DSA: `sizeof(WOLFSSL_SESSION)` drops from 973,984 to 168,736, about 26 MB less BSS on the default cache. Deliberately not derived from the enabled ML-DSA level either - `MLDSA_MAX_SIG_SIZE` tracks the compiled level, so a level-restricted build would otherwise have silently dropped certificates a full build keeps.

I have not yet added the BSS figures to the PR description; will do that separately.

tests/suites.c
Frauschi: Guards added to all six new entity blocks, plus `!defined(WOLFSSL_MLDSA_VERIFY_ONLY)` on the two pre-existing ML-DSA-leaf blocks you pointed at.

One correction on the severity, though: this does not break `all-pq-verify-only-noasm` today. That config passes `-DNO_WOLFSSL_CIPHER_SUITE_TEST` in its CPPFLAGS, which compiles out the whole `SuiteTest()` call in `tests/unit.c`, so no `.conf` test runs there at all. I ran the config to check - it passes, and no `.conf` banner appears in its `scripts/unit.log`. Your reproduction invoked the `.conf` by hand, which bypasses that guard. I also scanned every `.github/configs/*.json` and no config pairs verify-only with `.conf` suites. Worth fixing as a latent trap, which is why I did, but it is not a live CI break.

src/ssl_load.c
Frauschi: Both validated now, in `ProcessBufferCertPublicKey` and `ProcessBufferCertAltPublicKey`, matching `ProcessBufferTryDecodeSlhDsa`. Also took your second suggestion: the per-parameter-set `WOLFSSL_SLHDSA_PARAM_NO_*` macros are now derived from the group-level `NO_SMALL` / `NO_FAST` / `NO_128` exclusions, for both the SHAKE and SHA2 families, so the parameter table and the positive macros the TLS mappings use cannot diverge.

While in there I found a related bug in the same function: the SLH-DSA arm read `cert->keyOID` even though the switch is over `cert->sapkiOID`, so the native key's OID was picking the alternative key's parameter set. Fixed.

src/ssl_load.c
Frauschi: Left as is for now - dual-algorithm support for SLH-DSA is out of scope for this PR, and we would rather not half-implement it. Agreed the diagnostic is worse than it should be; if you would prefer the load-time rejection as an interim step, say so and I will add it.

wolfssl/wolfcrypt/asn.h
Frauschi: Added a "Behavioral change" entry to README.md under Enhancements, alongside the existing SLH-DSA and RSA-PSS entries, describing the growth and pointing downstream users at heap allocation or sizing from the parameter set actually in use.

I did not add a `#pragma message`. It fires on every translation unit that includes the header, including builds that never touch the macro, which felt noisier than the problem. Happy to add it if you disagree.

.github/workflows/pq-all.yml
Frauschi: Added `pq-asynccrypt-dual-alg` to `.github/configs/pq-all.json`, combining `--enable-asynccrypt-sw`, `--enable-dual-alg-certs` and SLH-DSA.

That entry immediately earned its keep: it fails to build on the current code, because `Scv13Args` gained four unguarded members and no longer fit `ssl->async->args` - `static assertion failed: sizeof(ssl->async->args) >= sizeof(*args)` at `tls13.c:10268`. The holder is 96 bytes and the struct went 88 to 112 with dual-alg; without dual-alg it lands at exactly 88 and still fits, which is why nothing caught it. Fixed by ordering the struct widest-member-first so it carries no interior padding: 88 bytes, with margin. The config passes now.

src/ssl_load.c
Frauschi: Fixed. Took your second suggestion in spirit and zeroize before `wc_SlhDsaKey_Init` so a failed Init always leaves a Free-safe object, which keeps the unconditional Free correct. Your note about the comment being wrong on that path was right; it has been rewritten.

Applied the same fix to `AllocKey`, which had the identical shape.

wolfssl/wolfcrypt/wc_slhdsa.h
Frauschi: Renamed to `WC_SLHDSA_ALL_NO_*` across the header and `wc_slhdsa.c`.

src/ssl_load.c
Frauschi: Fixed - the OID is validated before it is stored in `*keyFormat`.

wolfcrypt/src/asn.c
Frauschi: Confirmed, and this was the most valuable finding in the set. I traced the same chain: `wolfssl_x509_make_der` sets `ret = WOLFSSL_FAILURE`, which is 0; `wc_GeneratePreTBS` returns it unchanged; `ParseCertRelative` then does `if (ret > 0) ret = ConfirmSignature(...)`, skips it, and the following `if (ret != 0)` reads 0 as success. The alternative signature is never verified and the certificate is accepted.

Applied both halves, as you asked. The sizing is now a `GeneratePreTBSBuffer()` helper shared by both call sites: it tries the tight `maxIdx - sigLength - altSigValLen` first, because keeping that allocation small on constrained targets is deliberate, and retries once at a size the canonical re-encode cannot exceed if the estimate is short. A non-positive result is then an error rather than a skipped check.

All three paths are proven by forcing the sizes: normal case `derSz=1928` against a 1837-byte PreTBS with no retry; tight estimate forced 200 bytes short gives `result=0` - your fail-open value, observed directly - the retry fires and verification still succeeds; both forced short returns `ASN_PARSE_E` and the certificate is rejected.

Separately worth flagging: getting the path to execute at all required adding a `wolfSSL_CertManagerVerifyBuffer` call to `test_dual_alg_ecdsa_mldsa`, because every `ParseCertRelative` call in the dual-alg tests ran with `verify=0` and the alternative-signature block was never reached. Doing so exposed a bug in the test fixture itself: it set the altSignatureAlgorithm extension to `CTC_SHA256wECDSA` while signing with ML-DSA-44. Corrected, and that path now has real coverage.

I could not add the "force wc_GeneratePreTBS to fail and assert ParseCertRelative returns non-zero" test you asked for, because there is no injection point from the public API. The assertion that a 1-byte buffer returns `<= 0` covers the fail-open source directly instead.

dgarske (dismissed):

dgarske (approved):

dgarske:
Jenkins retest this please

PR #10913, opened 2026-07-15
ECH with multiple tenants

sebastian-carpenter:
# Description

Add support for ECH with multiple tenants (i.e., multiple private SNI's).

Ran into some issues with the SNI_Callback function when running the new callback tests: `*_multi_sni` and `*_multi_ctx`. The server never ack's an SNI that was accepted. This was fixed by adding `TLSX_SetResponse(ssl, TLSX_SERVER_NAME)` to the default case. Not sure if a new case would be preferred (ack_return?) or if this should be placed in the default since it could affect many users.

ECH testing:
* Reworked testing (where appropriate) to make use of a 'public' and 'private' cert, switched based on the SNI the client provides.
* Added multi-sni / multi-ctx tests
* Added more sni_parse vectors based on the ordering of ECH and SNI in the extensions list
* Optimized / refactored some tests

ECH code:
* Reworked some code to make it easier to look at
* Added / moved some guards
* echConfig is now referenced through the ech extension only instead of through the ctx as well
* echConfigs are copied when the ssl's ctx is swapped

client/server examples:
* Updated to accept an ECH cert/key.
* Improved logging

openssl interop:
* Reject config is now built at runtime instead of from a stored config
* Improved the --reject arg
* Updated to use the multiple certs

Added certs/keys:
* ech-public-cert.(pem|der)
* ech-public-key.(pem|der)
* tenant-a-cert.(pem|der), uses server-key
* tenant-b-cert.(pem|der), uses server-key

Updated certs scripts:
* renewcerts.sh now contains the added certs/keys
* wolfssl.cnf updated with the cert info
* gencertbuf.pl updated for builds with no filesystem
* Updated certs/include.am with the new certs/keys

# Checklist

 - [X] added tests
 - [ ] updated/added doxygen


src/ssl_ech.c
```diff
@@ -906,6 +906,113 @@ int GetEchConfigsEx(WOLFSSL_EchConfig* configs, byte* output, word32* outputLen)
...
+
+    echX = TLSX_Find(ssl->extensions, TLSX_ECH);
+    if (echX == NULL || echX->data == NULL)
+        return 0;
+
+    ech = (WOLFSSL_ECH*)echX->data;
+
+    /* the client's configs will be owned by the ssl, and only copied once */
+    if (ssl->options.side != WOLFSSL_SERVER_END || ech->copiedConfig)
+        return 0;
```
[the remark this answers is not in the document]
sebastian-carpenter: fixed

tests/api.c
```diff
@@ -16014,9 +16290,25 @@ static int test_wolfSSL_Tls13_ECH_rejected_empty_client_cert(void)
...
+    const char* sSNI;           /* server-side SNI                            */
+    const char* innerName;      /* client private SNI                         */
+    const char* outerName;      /* client public SNI                          */
+    const char* authoritative;  /* value expected from wolfSSL_SNI_GetRequest */
+    int         outerRet;       /* expected return from parsing the outer CH  */
+    int         innerRet;       /* expected return from parsing the inner CH  */
+    int         reject;         /* setup client's ECH to be rejected          */
+    byte        sniOpt;         /* options for the server SNI                 */
+    byte        reorder;        /* swap outer SNI/ECH so ECH parses first     */
+    byte        reparseSni;     /* re-parse outer SNI after accept (post-HRR) */
```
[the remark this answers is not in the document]
sebastian-carpenter: fixed

sebastian-carpenter:
Jenkins retest this please.

src/ssl_ech.c
```diff
@@ -906,6 +906,113 @@ int GetEchConfigsEx(WOLFSSL_EchConfig* configs, byte* output, word32* outputLen)
...
+    XMEMCPY(copy->cipherSuites, src->cipherSuites, suitesSz);
+    copy->configId = src->configId;
+    copy->kemId = src->kemId;
+    copy->numCipherSuites = src->numCipherSuites;
+    copy->maxNameLen = src->maxNameLen;
+    XMEMCPY(copy->receiverPubkey, src->receiverPubkey, HPKE_Npk_MAX);
+
+    /* publicName shouldn't be NULL here, but make sure since the TLS code
+     * relies on it existing */
+    if (src->publicName == NULL) {
```
SparkiDev: Check this before doing anything.

SparkiDev (dismissed):

src/ssl_ech.c
sebastian-carpenter: Fixed. Moved publicName check up to top and also added one for ciphersuites.

examples/server/server.c
```diff
@@ -1610,6 +1616,87 @@ static int server_srtp_test(WOLFSSL *ssl, func_args *args)
 }
 #endif
 
+#if defined(WOLFSSL_TLS13) && defined(HAVE_ECH) && !defined(NO_CERTS)
```
dgarske, line 1619: **🔴 [High] ECH cert-swap code in examples/server.c is not guarded by HAVE_SNI, breaking --enable-ech --disable-sni builds**

The new ECH multi-tenant code introduces three undefined-symbol errors when `HAVE_ECH` is enabled but `HAVE_SNI` is not. (1) `sniHostName` is declared only under `#ifdef HAVE_SNI` (server.c:1846-1848) but is read at server.c:3112 (`echPrivateCert.name = sniHostName;`) inside a block guarded only by `#if defined(WOLFSSL_TLS13) && defined(HAVE_ECH) && !defined(NO_CERTS)`. (2) `wolfSSL_CTX_set_servername_callback()`, `wolfSSL_CTX_set_servername_arg()` and `CallbackSniRecv` are declared only inside `#ifdef HAVE_SNI` in `wolfssl/ssl.h:5769-5777`, but are called at server.c:3368-3369 under an ECH-only guard. (3) `WOLFSSL_SNI_HOST_NAME` and `wolfSSL_SNI_GetRequest()` are inside `#ifdef HAVE_SNI` in `wolfssl/ssl.h:4637-4687` and are used by `EchSwapToPrivateCert()` at server.c:1656. `HAVE_ECH` does NOT imply `HAVE_SNI`: `configure.ac:2387` only sets a default (`test "$enable_sni" = "" && enable_sni=yes`), and `configure.ac:8695-8708` honours an explicit `--disable-sni` and then never emits `-DHAVE_SNI`. There is no `HAVE_ECH -> HAVE_SNI` rule in `wolfssl/wolfcrypt/settings.h`. The library itself still builds (the corresponding library code at `src/tls13.c:7832` and `src/internal.c:43885` is HAVE_SNI-guarded); only the example breaks. This is a regression introduced by this PR: before it, the `#ifdef HAVE_ECH` block in server.c only called `GenerateEchConfig`/`GetEchConfigs`, so `--enable-ech --disable-sni` compiled cleanly.

**Fix:** Add `defined(HAVE_SNI)` to the guards at server.c:1619, 1854, 3093 and 3367. Alternatively, force `ENABLED_SNI=yes` when ECH is enabled in configure.ac and add a `#error` in settings.h so the dependency is enforced once rather than at each use site. Please add a `--enable-ech --disable-sni` build to CI to keep this configuration covered.

src/internal.c
```diff
@@ -43913,6 +43913,7 @@ static int DefTicketEncCb(WOLFSSL* ssl, byte key_name[WOLFSSL_TICKET_NAME_SZ],
                     WOLFSSL_MSG("Server quietly not acking servername.");
                     break;
                 default:
```
dgarske, line 43915: **🔴 [High] TLSX_SetResponse in the SNI callback default case can emit an unsolicited empty server_name extension**

This is the change the PR description explicitly flagged as uncertain. Two independent analyses converged on the same defect. `SNI_Callback()` is invoked unconditionally by the server after `TLSX_Parse` (`src/internal.c:40010` for TLS 1.2, `src/tls13.c:7833` for TLS 1.3), whether or not the ClientHello actually contained a `server_name` extension. `TLSX_SetResponse()` is a no-op only when `TLSX_Find(ssl->extensions, TLSX_SERVER_NAME)` returns NULL. However `TLSX_SERVER_NAME` can be present on `ssl->extensions` with NO client SNI received, via the supported server-side API `wolfSSL_UseSNI(ssl, ...)` -> `TLSX_UseSNI(&ssl->extensions, ...)` (`src/ssl_api_ext.c:46-52`); that this is a supported server pattern is confirmed by `wolfSSL_SNI_SetOptions(WOLFSSL*, ...)` living under `#ifndef NO_WOLFSSL_SERVER` and by `tests/api.c:16760`. The chain: `wolfSSL_UseSNI(ssl,...)` creates the extension with `resp=0`; the client sends no `server_name` so `TLSX_SNI_Parse` never runs and never sets the response; `SNI_Callback` runs anyway, returns 0, hits `default:`, and sets `resp=1`; the write path only skips an extension when `!isRequest && !extension->resp` (`src/tls.c:15380-15381`) and `TLSX_SERVER_NAME` is masked neither by the TLS 1.2 server_hello semaphore nor by encrypted_extensions, so a 4-byte empty `server_name` is emitted. RFC 8446 Sect 4.2 forbids sending an extension the client did not offer, and wolfSSL itself enforces this: `TLSX_SNI_Parse` with `isRequest==0` and no local extension calls `TLSX_HandleUnsupportedExtension` (`src/tls.c:2411-2414`), sending a fatal…

**Fix:** Gate the ack on the client having actually sent a server_name. `sni->status` is initialised to `WOLFSSL_SNI_NO_MATCH` by `TLSX_SNI_New` and is raised only inside `TLSX_SNI_Parse`, making it an exact 'client sent SNI and we recorded it' signal; `TLSX_SNI_Status()` is already `WOLFSSL_LOCAL` (`wolfssl/internal.h:3335`). Add a regression test: server with `wolfSSL_UseSNI(ssl,...)` plus a servername callback, client with no SNI, assert the handshake still succeeds.

dgarske, line 43915: **🟠 [Medium] SSL_TLSEXT_ERR_OK has no explicit case, so default: also acks unrecognized callback return codes (fail-open)**

This directly answers the author's question in the PR description ('Not sure if a new case would be preferred (ack_return?) or if this should be placed in the default'). The answer is yes, a dedicated case is preferred. The switch is on the raw `int` returned by `ssl->ctx->sniRecvCb`. The mapping is: `SSL_TLSEXT_ERR_OK`=0 (`wolfssl/openssl/ssl.h:1638`) has no case label and falls to `default:`; `SSL_TLSEXT_ERR_ALERT_WARNING`=1==`warning_return`, `SSL_TLSEXT_ERR_ALERT_FATAL`=2==`fatal_return`, `SSL_TLSEXT_ERR_NOACK`=3==`noack_return` (`wolfssl/ssl.h:1034-1038`). So `default:` is reached by exactly one legitimate value (0 = OK) plus every illegitimate one: negative wolfSSL error codes, stray values, 4 and above, and uninitialised returns. Before this diff `default:` was inert, so a garbage return produced no protocol action; after it, a garbage return produces an affirmative 'servername accepted' assertion on the wire. Failing open on an unrecognized callback result is the wrong default for a multi-tenant vhost selector, which is the exact use case this PR targets. Git history confirms the silence was not a deliberate anti-ack decision being regressed: the switch came from `1b6b16c2c39` ('HaProxy 2.4-dev18 support'), which only refactored an earlier if/else chain and added the NOACK arm; `default:` was simply 'nothing left to do'. So the fix direction is right and addresses a genuine long-standing omission -- only the placement is wrong.

**Fix:** Give OK its own case and leave `default:` inert. Optionally add `ack_return = 0` to `enum SNICbReturn` in wolfssl/ssl.h (ABI-safe, since the other values are pinned to alert_warning/alert_fatal) and use `case ack_return:`. Merge this with the 'client actually sent SNI' guard from the preceding finding -- one combined fix resolves both.

tests/api.c
```diff
@@ -15393,41 +15552,45 @@ static int test_wolfSSL_Tls13_ECH_retry_configs(void)
 }
 
 /* Test retry configs are cleared when authentication fails */
```
dgarske, line 15554: **🟠 [Medium] HRR variant of the ECH retry-config auth-failure test was deleted, losing coverage**

The test was `test_wolfSSL_Tls13_ECH_retry_configs_auth_fail_ex(int hrr)` with a wrapper running both `hrr=0` and `hrr=1`; the `hrr=1` pass added `if (hrr) ExpectIntEQ(wolfSSL_NoKeyShares(test_ctx.c_ssl), WOLFSSL_SUCCESS);` so the 'retry configs must be cleared when outer-hello auth fails' assertion was exercised across a HelloRetryRequest. The PR collapses it to a `(void)` function, deletes the hrr branch and deletes the wrapper. No other ECH test combines `wolfSSL_NoKeyShares` with a `DOMAIN_NAME_MISMATCH` outer-hello failure plus a negative `wolfSSL_GetEchRetryConfigs` assertion, so this scenario is now entirely uncovered. Note that the parallel removal of `TEST_DECL(test_wolfSSL_Tls13_ECH_HRR)` is NOT a loss -- `test_wolfSSL_Tls13_ECH` (tests/api.c:15028-15038) now calls `test_wolfSSL_ECH_conn_ex(..., 0)` and `(..., 1)`, so HRR is still run there.

**Fix:** Restore the `_ex(int hrr)` parameterization (keeping the new scoped-block cleanup and the `retryConfigs` buffer fix, which are both genuine improvements) and re-add a `(void)` wrapper calling it with 0 and 1.

tests/api.c
```diff
@@ -15062,6 +15136,87 @@ static int ech_find_extension(byte* buf, word16* idx_p, word16 extType)
     return BAD_FUNC_ARG;
 }
 
+/* swap two extensions in a CH record so they appear in the opposite order
+ * returns 0 on success, error otherwise. */
+static int ech_swap_extensions(byte* buf, word16 typeFirst, word16 typeSecond)
```
dgarske, line 15141: **🔵 [Low] ech_swap_extensions performs unchecked word16 length arithmetic that can underflow**

The copy algebra itself is correct and was verified: `tmp` is sized `middle = lenA + midLen` and holds `[A][middle]`; the reads `tmp+lenA` for `midLen` bytes and `tmp` for `lenA` bytes are in bounds; `XMEMMOVE(buf+idxA, buf+idxB, lenB)` reads a not-yet-overwritten region and correctly uses memmove for the overlap when `lenB > lenA+midLen`; the total written back (`lenB+midLen+lenA`) exactly equals the original span, so no record or handshake length prefix is invalidated. The gap is input validation. `ech_find_extension` (15104-15137) returns as soon as the type matches and never validates the matched extension's own length -- only the lengths of extensions it skips. So `lenA`/`lenB` are unvalidated 16-bit wire values: `lenA += 2 * OPAQUE16_LEN` can wrap, and `midLen = idxB - (idxA + lenA)` is an unsigned subtraction with no guarantee that `idxA + lenA <= idxB`. A single malformed length yields `midLen` near 0xFFFF, a ~64KB `XMALLOC` and a 64KB out-of-bounds read and write. This cannot be triggered today because the ClientHellos are self-generated by the test, but the helper is unguarded for future reuse. The adjacent-extension case (`midLen == 0`) is benign since `middle == lenA >= 4`.

**Fix:** Add the bounds guard before the `XMALLOC` (using `word32` for the sum to catch the `+=` wrap). The allocation lifetime is already correct -- both early returns precede the XMALLOC, so no XFREE is missed.

tests/api.c
```diff
@@ -16383,18 +16690,33 @@ static int ech_server_parse_inner_ch(WOLFSSL* ssl)
...
-    const char* innerName;      /* client private SNI                         */
-    const char* outerName;      /* client public SNI                          */
-    const char* authoritative;  /* value expected from wolfSSL_SNI_GetRequest */
-    int         outerRet;       /* expected return from parsing the outer CH  */
-    int         innerRet;       /* expected return from parsing the inner CH  */
-    int         reject;         /* setup client's ECH to be rejected          */
-    byte        sniOpt;         /* options for the server SNI                 */
-} echSniVec;
+/* Re-parse the outer server_name through TLSX_Parse */
+static int ech_server_reparse_outer_sni(WOLFSSL* ssl, byte* chRecord,
```
dgarske, line 16694: **🔵 [Low] ech_server_reparse_outer_sni performs a no-op extension swap and its comment is wrong**

Only the SNI extension's own bytes are handed to `TLSX_Parse`: `idx` points at the extension TYPE and the length passed is `sniLen + OPAQUE16_LEN + OPAQUE16_LEN` (type + len + body), which is the correct framing for TLSX_Parse's walk. The ECH extension is therefore never reached by the parse, so the preceding `ech_swap_extensions` call and its comment ('extensions must be swapped, otherwise ECH would cause an error') are both wrong. Worse, both `reparse` vectors in the table have `reorder=0` (tests/api.c:16899-16904), so the dead swap always executes, rewriting `test_ctx.s_buff` in place after `ech_server_parse_outer_ch` already stored `ech->aad = chBody` pointing into that same buffer. The test still passes, but it is mutating state under a live alias for no reason.

**Fix:** Delete the swap block and its comment. If the intent was to re-parse the whole post-HRR extension list, pass `extLen` from `ech_seek_extensions` instead of `sniLen + 4` -- then the swap becomes necessary and the comment becomes true.

dgarske (commented):
## Skoll Code Review

**Findings:** 24 total — 24 posted, 0 skipped
**18 finding(s) posted as inline comments** (see file-level comments below)
**6 finding(s) not tied to a diff line** (full detail below)

### Posted findings
- [High] **ECH cert-swap code in examples/server.c is not guarded by HAVE_SNI, breaking --enable-ech --disable-sni builds** — `examples/server/server.c:1619,1656,3093-3112,3367-3373`
- [High] **TLSX_SetResponse in the SNI callback default case can emit an unsolicited empty server_name extension** — `src/internal.c:43915-43917`
- [Medium] **SSL_TLSEXT_ERR_OK has no explicit case, so default: also acks unrecognized callback return codes (fail-open)** — `src/internal.c:43915-43917`
- [Medium] **HRR variant of the ECH retry-config auth-failure test was deleted, losing coverage** — `tests/api.c:15554-15614`
- [Low] **ech_swap_extensions performs unchecked word16 length arithmetic that can underflow** — `tests/api.c:15141-15191`
- [Low] **ech_server_reparse_outer_sni performs a no-op extension swap and its comment is wrong** — `tests/api.c:16694-16719`
- [Low] **New BAD_STATE_E ClientHello2 guard uses = while the adjacent CH2 early-out uses ==, and emits no trace** — `src/tls.c:14834-14838`
- [Low] **New failure path in wolfSSL_set_SSL_CTX is silent** — `src/ssl.c:8482-8487`
- [Low] **ECH wire-SNI test matrix reduced from 4 cases to 2; echAccepted assertion dropped** — `tests/api.c:15929-16047`
- [Low] **Copied ECH configs leave raw/rawLen NULL while the comment claims they are 'derived later'** — `src/ssl_ech.c:947-956`
- [Low] **Omission of receiverPrivkey is safe only via an undocumented, unenforced cached-HPKE-context invariant** — `src/ssl_ech.c:911-957`
- [Low] **hello_retry_request branch still dereferences echX-data without the NULL check the client_hello branch just gained** — `src/tls.c:14790-14801`
- [Low] **ECH response flag now depends on TLSX_ECH_Parse running, silently coupling the HRR confirmation offset to it** — `src/tls.c:14803-14808`
- [Low] **--ech with -P (PK callbacks) now hard-aborts the server example** — `examples/server/server.c:2767-2771`
- [Info] **CopyEchConfigsPub reports success for three undocumented no-op cases** — `src/ssl_ech.c:960-1007`
- [Info] **openssl-ech.sh: operand-less wait always returns 0, making the added || clause dead** — `.github/scripts/openssl-ech.sh:217-219, 327-329`
- [Info] **corrupt_ech_config depends on bash-4 mapfile and GNU base64/od** — `.github/scripts/openssl-ech.sh:111-126`
- [Info] **Multi-tenant CTX-swap test drops the SNI dispatch callback for the rest of the connection** — `tests/api.c:16261-16296`

#### New ech-public-key.pem is PKCS#8 while every other RSA test key is PKCS#1, breaking NO_PKCS8 builds
**File:** `certs/renewcerts.sh:1074, certs/ech-public-key.pem:1, wolfssl/certs_test.h:3043-3045`
**Function:** `run_renewcerts`
**Severity:** Medium

`openssl genrsa` under OpenSSL 3 emits PKCS#8, and the subsequent `openssl rsa -outform DER` preserved it. `certs/ech-public-key.pem` line 1 is `-----BEGIN PRIVATE KEY-----` (PKCS#8) rather than `-----BEGIN RSA PRIVATE KEY-----`, and `ech_public_key_der_2048` starts `30 82 04 BC 02 01 00 30 0D 06 09 2A 86 48 86 F7 0D 01 01 01 05 00 04 82` (PrivateKeyInfo), whereas `server_key_der_2048` (certs_test.h:1883) starts `30 82 04 A5 02 01 00 02 82 01 01 00` (bare PKCS#1). Every other RSA key used by the api tests is PKCS#1. Because this PR makes ech-public-key the server key for essentially every ECH memio test and for examples/server ECH mode, `NO_PKCS8` builds (`--disable-pkcs8`, or `user_settings_baremetal.h:202` / `user_settings_wolftpm.h:185`) will now fail at key load where they previously worked with server-key.pem. PKCS#8 is on by default in `configure.ac:10391`, so mainline CI will not catch this.

**Recommendation:** Use `openssl genrsa -traditional` (or normalize afterwards with `openssl rsa -traditional`) and regenerate `ech-public-key.der` plus the `certs_test.h` buffer so the fixture matches the format of every other RSA key in certs/.

_Referenced code: `certs/renewcerts.sh:1074, certs/ech-public-key.pem:1, wolfssl/certs_test.h:3043-3046` (4 lines)_

---

#### No test asserts the presence or absence of the server_name ack on the callback path
**File:** `tests/api.c:8670-8712, tests/api/test_tls.c:1734-1738`
**Function:** `verify_SNI_real_matching / accept_any_sni_cb`
**Severity:** Medium

The `TLSX_SetResponse(TLSX_SERVER_NAME)` change alters on-the-wire behavior for every non-ECH user of `wolfSSL_CTX_set_servername_callback` / `SSL_CTX_set_tlsext_servername_callback`, but nothing in the suite pins that behavior. The only tests that assert the ack state are in `test_wolfSSL_UseSNI_connection` (tests/api.c:8720), whose helpers `verify_SNI_no_matching` / `verify_SNI_real_matching` / `verify_SNI_fake_matching` (tests/api.c:8681-8712) read `wolfSSL_SNI_Status()` on the client -- i.e. exactly whether the server acked. Those cases all drive the NON-callback path (`use_SNI_at_ctx` / `use_SNI_at_ssl`), which already set the response before this diff, so they are unaffected and do not cover the new code. The existing non-ECH tests that DO register a servername callback returning 0 -- `accept_any_sni_cb` used by `test_tls12_session_id_resumption_sni_mismatch` (tests/api/test_tls.c:1764) and `test_tls13_session_resumption_sni_mismatch` (tests/api/test_tls.c:1836), plus `inspect_SNI` in tests/quic.c:1539 -- now exercise the new ack but assert only on `wolfSSL_session_reused()` and callback ordering, never on the SNI response. Nothing would catch a regression in either direction, and nothing covers the no-client-SNI case that the BLOCK finding describes.

**Recommendation:** Add a non-ECH memio matrix over TLS 1.2 and TLS 1.3 with the server registering a servername callback: (1) callback returns 0 and client sent SNI -> handshake succeeds and client `wolfSSL_SNI_Status()==WOLFSSL_SNI_REAL_MATCH`; (2) callback returns `noack_return` -> succeeds with `NO_MATCH`; (3) server also calls `wolfSSL_UseSNI(ssl,...)` and client sends NO SNI -> handshake still succeeds with no `unsupported_extension` alert; (4) callback returns an unrecognized value (e.g. 99) -> no ack.

_Referenced code: `tests/api.c:8670-8712, tests/api/test_tls.c:1734-1739` (6 lines)_

---

#### ECH extension absence changed from fail-closed to a silent ECH downgrade
**File:** `src/tls13.c:8374-8377, src/tls13.c:7781-7792`
**Function:** `SendTls13ServerHello / DoTls13ClientHello`
**Severity:** Low

The PR replaces the guard `ssl->ctx->echConfigs != NULL && !ssl->options.disableECH` plus a hard `if (echX == NULL) return WOLFSSL_FATAL_ERROR;` with a condition that folds `TLSX_Find` into the `if` and silently skips the whole ECH block when the extension is absent. The same change was made in `DoTls13ClientHello`, where `ERROR_OUT(WOLFSSL_FATAL_ERROR, exit_dch)` was deleted. `git log -S` confirms both hard errors originate from the original ECH commit `6b6ad38e4f` -- they are defensive assertions, not targeted bug fixes, so this is not re-introducing a previously fixed bug, and they were effectively unreachable. The loosening is also intentional and necessary for multi-tenant CTX switching, since `wolfSSL_set_SSL_CTX` now calls `CopyEchConfigsPub()` so `ech->echConfig` survives a swap while `ssl->ctx->echConfigs` becomes NULL. The residual risk is that the invariant is now gone entirely: if the extension is ever missing while ECH was expected, the server emits a ServerHello without the acceptance confirmation instead of failing, which the client reads as ECH rejection -- a silent privacy downgrade rather than a loud failure.

**Recommendation:** Keep the loosened guard, but preserve the invariant with a targeted fail-closed check keyed on `ssl->options.echAccepted` in both functions. At minimum add a `WOLFSSL_MSG()` on the skip path so the downgrade is diagnosable in a debug build.

_Referenced code: `src/tls13.c:8374-8377, src/tls13.c:7781-7790` (10 lines)_

---

#### PrintEchStatus prints on every handshake in any ECH-enabled build
**File:** `wolfssl/test.h:4929-4954, examples/server/server.c:3936-3939, examples/client/client.c:4304-4307`
**Function:** `PrintEchStatus`
**Severity:** Low

The helper has no early-out, so every connection in an `--enable-all` build (configure.ac:1364 turns ECH on there) gains an `ECH status: not offered` line, including all non-ECH test runs that use the examples. Contrast `EarlyDataStatus()` (test.h:4909), which returns early when the feature is not in play. The CI script now greps for these strings, so they do serve a purpose -- but only the accepted/rejected values are ever asserted.

**Recommendation:** Return early on `WOLFSSL_ECH_STATUS_NOT_OFFERED` so the line only appears when ECH was actually in play, matching the `EarlyDataStatus` precedent.

---

#### TLSX_SetResponse prototype guard does not match the definition guard, and the definition lacks WOLFSSL_LOCAL
**File:** `wolfssl/internal.h:3301, src/tls.c:1754`
**Function:** `TLSX_SetResponse`
**Severity:** Info

The relocated prototype is added under `#ifdef HAVE_TLS_EXTENSIONS`, but the definition (src/tls.c:1754) is inside `#if !defined(NO_WOLFSSL_SERVER) || defined(WOLFSSL_TLS13)`. In a `NO_WOLFSSL_SERVER` + `!WOLFSSL_TLS13` build the header declares a function that is never compiled. This is harmless in practice -- there are no call sites in that configuration, and `HAVE_SNI` without `HAVE_TLS_EXTENSIONS` is already a hard `#error` at internal.h:3315-3317, so the new caller in src/internal.c is always covered. It is noted only for consistency: by contrast, `CopyEchConfigsPub` at internal.h:3245 IS correctly placed inside the `WOLFSSL_TLS13 && HAVE_ECH` block, matching both its definition and its call site. The definition also still lacks the `WOLFSSL_LOCAL` that the new declaration carries.

**Recommendation:** Guard the prototype with the same condition as the definition and add `WOLFSSL_LOCAL` to the definition. Note the related `WOLFSSL_TEST_VIS` change on `TLSX_SNI_GetRequest` is the reverse and is correct -- the header at internal.h:3388 already carried `WOLFSSL_TEST_VIS` and the definition was the one out of sync, so the PR fixes it.

_Referenced code: `wolfssl/internal.h:3301, src/tls.c:1754-1760` (7 lines)_

---

#### tenant-a and tenant-b certificates are issued over the same RSA key pair
**File:** `certs/renewcerts.sh:1101-1113, certs/tenant-a-cert.pem:14-32`
**Function:** `run_renewcerts`
**Severity:** Info

Both CSRs use `-key server-key.pem`, and both certs carry modulus `00:c0:95:08:e1:57:41:f2:71:6d:b7:d2:45:41:27:...`, which also matches certs/server-cert.pem. The tests correspondingly use `svrKeyFile` for both tenants (tests/api.c:16242-16243, 16266-16267). The consequence is that a bug which swapped the certificate but not the matching key -- or picked the wrong tenant's key -- would still complete the handshake, so only cert selection (via SAN plus `wolfSSL_check_domain_name`) is genuinely under test. Verified as fine: serials 0x21/0x22/0x23 were checked against every `-set_serial` in renewcerts.sh and all committed PEMs and are unique under ca-cert.pem; validity Jul 10 2026 - Apr 5 2029 (`-days 1000`) matches the repo norm (server-cert.pem: Jun 11 2026 - Mar 7 2029); the new `wolfssl.cnf` sections omit `keyUsage` just like the existing `[wolfssl_opts]`; and renewcerts.sh runs the new sections from `certs/` before the `gencertbuf.pl` step, so the `.der` files exist when the buffers are generated.

**Recommendation:** Give at least one tenant a distinct key if key-swap coverage matters; otherwise add a comment stating the shared key is intentional so a future reader does not assume the test proves more than it does.

_Referenced code: `certs/renewcerts.sh:1101-1113, certs/tenant-a-cert.pem:14-15` (2 lines)_

---

src/tls.c
```diff
@@ -14831,6 +14830,13 @@ static int TLSX_ECH_Parse(WOLFSSL* ssl, const byte* readBuf, word16 size,
             /* MUST process INNER in inner hello and OUTER in outer hello */
             return INVALID_PARAMETER;
         }
+
+        /* expect hpke and its context to be available on ClientHello2 */
```
dgarske, line 14834: **🔵 [Low] New BAD_STATE_E ClientHello2 guard uses = while the adjacent CH2 early-out uses ==, and emits no trace**

The new guard fires when `serverState >= SERVER_HELLO_RETRY_REQUEST_COMPLETE` and `ech->hpke` or `ech->hpkeContext` is NULL. Reading the enclosing function, this combination cannot occur today: 24 lines earlier the pre-existing early-out at `src/tls.c:14812-14816` returns as soon as `(!ssl->options.echAccepted && serverState == SERVER_HELLO_RETRY_REQUEST_COMPLETE)`, so the guard is only reached with `echAccepted == 1`; `echAccepted` is set only after a successful `TLSX_ExtractEch` (`src/tls.c:14988`), and `TLSX_ExtractEch` leaves both `ech->hpke` and `ech->hpkeContext` non-NULL on success and frees both together on failure (`src/tls.c:14722-14730`). The concern is the asymmetry: the two checks disagree on the comparison operator (`==` vs `>=`). Per the enum at `wolfssl/internal.h:2068-2094`, the states above `SERVER_HELLO_RETRY_REQUEST_COMPLETE` are `SERVER_HELLO_COMPLETE` and later; a ClientHello parsed in those states would skip the early-out but trip the new `BAD_STATE_E`. No such flow could be constructed (TLS 1.3 has no renegotiation and `got_client_hello` is capped at 2), so this is not reported as reachable -- but the inconsistency looks unintentional and is fragile against future state additions. Additionally the return emits no `WOLFSSL_MSG`/`WOLFSSL_ERROR_VERBOSE`, so if it ever does fire it surfaces with no trace and no specific alert mapping.

**Fix:** Make both checks use the same predicate (`==`, matching the early-out above) so a future state addition cannot make them disagree, and add a WOLFSSL_MSG/WOLFSSL_ERROR_VERBOSE before the return.

src/ssl.c
```diff
@@ -8479,6 +8479,13 @@ WOLFSSL_CTX* wolfSSL_set_SSL_CTX(WOLFSSL* ssl, WOLFSSL_CTX* ctx)
         InitSSL_CTX_Suites(ctx);
     }
 
+#if defined(WOLFSSL_TLS13) && defined(HAVE_ECH)
```
dgarske, line 8482: **🔵 [Low] New failure path in wolfSSL_set_SSL_CTX is silent**

The placement of the new `CopyEchConfigsPub` call is correct and was verified in detail: the `return NULL` at src/ssl.c:8486 sits before `wolfSSL_RefWithMutexInc` (8489), before `wolfSSL_CTX_free(ssl->ctx)` (8500) and before `ssl->ctx = ctx` (8501), so nothing on the WOLFSSL object has been mutated yet -- no dropped refcount, no torn `ssl->buffers.*`, no leak. The copy also correctly captures the OLD ctx's configs (the ones that decrypted CH1), which is the intent. The issue is diagnosability: the failure emits no `WOLFSSL_MSG`/`WOLFSSL_ERROR_VERBOSE` and does not set `ssl->error`, and it adds a new allocation-sized failure source to a function that some integrations call without checking the return. In-tree callers do check (tests/api.c:13882, 29658), but an unchecked caller would silently keep the front-end cert.

**Fix:** Add the message and verbose error. `ret` is already declared at the top of the function, so this stays C89-clean.

tests/api.c
```diff
@@ -15763,63 +15916,35 @@ static int test_wolfSSL_Tls13_ECH_GREASE(void)
...
-
-    return EXPECT_RESULT();
-}
-
-/* The public name must be visible in plaintext
- * useCtx installs the inner SNI on the client ctx */
-static int test_wolfSSL_Tls13_ECH_wire_sni_ex(int accept, int useCtx)
+/* Test the client properly inherits an SNI from the ctx,
+ * the public name must be visible in plaintext */
+static int test_wolfSSL_Tls13_ECH_ctx_sni_ex(int reject)
```
dgarske, line 15929: **🔵 [Low] ECH wire-SNI test matrix reduced from 4 cases to 2; echAccepted assertion dropped**

`test_wolfSSL_Tls13_ECH_wire_sni_ex(accept, useCtx)` previously ran four combinations: (0,0), (1,0), (0,1), (1,1). The replacement `test_wolfSSL_Tls13_ECH_ctx_sni_ex(reject)` hard-wires the CTX form (`wolfSSL_CTX_UseSNI` on `c_ctx`) and deletes the `else { wolfSSL_UseSNI(c_ssl, ...) }` arm, so two scenarios are lost: accept + per-SSL SNI, and reject + per-SSL SNI. The assertion `ExpectIntEQ(test_ctx.c_ssl->options.echAccepted, accept);` was deleted with no replacement anywhere in the suite. Mitigating factor: the new `test_wolfSSL_Tls13_ECH_multi_sni_ex` (tests/api.c:16119) does use per-SSL `wolfSSL_UseSNI` and does call `test_ech_assert_wire_sni` on both CH1 and CH2, so the per-SSL path is not entirely uncovered.

**Fix:** Either keep a second `useCtx` parameter and restore the 4-way matrix, or add a comment recording that the per-SSL path is now covered by `multi_sni`. Re-add `ExpectIntEQ(test_ctx.c_ssl->options.echAccepted, !reject);` in either case.

src/ssl_ech.c
```diff
@@ -906,6 +906,107 @@ int GetEchConfigsEx(WOLFSSL_EchConfig* configs, byte* output, word32* outputLen)
...
+    /* publicName */
+    nameLen = (word32)XSTRLEN(src->publicName) + 1;
+    copy->publicName = (char*)XMALLOC(nameLen, heap, DYNAMIC_TYPE_TMP_BUFFER);
+    if (copy->publicName == NULL) {
+        XFREE(copy->cipherSuites, heap, DYNAMIC_TYPE_TMP_BUFFER);
+        XFREE(copy, heap, DYNAMIC_TYPE_TMP_BUFFER);
+        return MEMORY_E;
+    }
+    XMEMCPY(copy->publicName, src->publicName, nameLen);
+
```
dgarske, line 947: **🔵 [Low] Copied ECH configs leave raw/rawLen NULL while the comment claims they are 'derived later'**

Copies leave `raw == NULL` and `rawLen == 0`. All server-side consumers are fine: `GetEchConfig` (src/ssl_ech.c:354-460) re-serializes the wire form from the struct fields and never reads `config->raw`, so retry_configs produced from a copy are byte-identical; `wolfSSL_CTX_GenerateEchConfigEx` also produces `raw == NULL`, so this state is normal. However `TLSX_FinalizeEch` dereferences `ech->echConfig->raw`/`rawLen` unguarded when building the HPKE info. That is a client-only path and therefore unreachable for these server-side copies today, but the copy is structurally invalid for it, and the comment implies a fill-in step that does not actually exist anywhere.

**Fix:** Correct the comment to state that raw/rawLen are only used by the client seal path and that the server re-serializes via `GetEchConfig()`. Optionally add a NULL guard on `ech->echConfig->raw` in `TLSX_FinalizeEch` for defence in depth.

src/ssl_ech.c
```diff
@@ -906,6 +906,107 @@ int GetEchConfigsEx(WOLFSSL_EchConfig* configs, byte* output, word32* outputLen)
     return WOLFSSL_SUCCESS;
 }
 
+/* allocate *out and copy the public fields from src into it
+ * returns 0 on success, error otherwise */
+static int EchConfigCopyPub(WOLFSSL_EchConfig** out,
```
dgarske, line 911: **🔵 [Low] Omission of receiverPrivkey is safe only via an undocumented, unenforced cached-HPKE-context invariant**

Verified that HRR + CTX-swap is NOT broken by the missing private key: `TLSX_ExtractEch` touches `echConfig->receiverPrivkey` only inside `if (ech->hpke == NULL)` (src/tls.c:14649), and on CH2 the cached `ech->hpke`/`ech->hpkeContext` from CH1 are reused, so the privkey is never needed after the swap. The ordering also holds -- the only first-time decrypt is in `TLSX_Parse` of CH1, and `SNI_Callback` runs after it (src/tls13.c:7833). Worst case, `wc_HpkeInitOpenContext` rejects a NULL receiverKey with `BAD_FUNC_ARG` (hpke.c:1174), so a clean error rather than a crash. The point is that this fairly subtle dependency is captured only by a one-line comment in ssl.c and is neither documented on the copy function nor asserted anywhere.

**Fix:** Document on `EchConfigCopyPub` that the copies are public-only: usable for retry_configs, public-name matching, and continuing an already-established HPKE open context -- never for starting a new decryption.

src/tls.c
```diff
@@ -14801,12 +14800,12 @@ static int TLSX_ECH_Parse(WOLFSSL* ssl, const byte* readBuf, word16 size,
 
         ech->confBuf = (byte*)readBuf;
```
dgarske, line 14801: **🔵 [Low] hello_retry_request branch still dereferences echX-data without the NULL check the client_hello branch just gained**

The PR hardened the `client_hello` branch by adding `echX->data != NULL` to the condition, and `src/tls13.c:8374-8377` in this same PR adopts the same style. The sibling `hello_retry_request` branch immediately above was left as-is: it assigns `ech = (WOLFSSL_ECH*)echX->data;` then writes `ech->confBuf = (byte*)readBuf;` with only an `echX != NULL` check. This is pre-existing and currently safe (a TLSX_ECH node is always pushed with a non-NULL WOLFSSL_ECH by `TLSX_ECH_Use`/`TLSX_GreaseECH_Use`/`TLSX_ServerECH_Use`), so it is not a new defect -- but the diff now makes the two adjacent branches visibly inconsistent about the same invariant, which is the kind of asymmetry that invites a future bug.

**Fix:** While touching this function, make both branches enforce the same invariant.

src/tls.c
```diff
@@ -14801,12 +14800,12 @@ static int TLSX_ECH_Parse(WOLFSSL* ssl, const byte* readBuf, word16 size,
...
     }
-    else if (msgType == client_hello && ssl->ctx->echConfigs != NULL) {
-        /* get extension */
-        echX = TLSX_Find(ssl->extensions, TLSX_ECH);
-        if (echX == NULL)
-            return BAD_FUNC_ARG;
+    else if (msgType == client_hello &&
```
dgarske, line 14803: **🔵 [Low] ECH response flag now depends on TLSX_ECH_Parse running, silently coupling the HRR confirmation offset to it**

The PR removes `if (ret == 0) TLSX_SetResponse(ssl, TLSX_ECH);` from `TLSX_PopulateExtensions` (src/tls.c:16765-16771) and sets it here instead, only when a ClientHello actually carries an ECH extension. Every message type that can emit an ECH response was traced and the behavior is preserved: accepted-ECH keeps `resp=1` through `TLSX_EchReplaceExtensions` (early return at src/tls.c:18864), the rejected path keeps `resp=1` because src/tls.c:18882-18884 only clears resp when `ech->state == ECH_WRITE_NONE`, CH2 re-sets it on each parse, and the accepted ServerHello path removes the whole TLSX anyway (src/tls13.c:8410). So retry_configs in EncryptedExtensions and the 8-byte HRR confirmation both still work. The observation is the new coupling: `SendTls13ServerHello` computes `acceptOffset = (word32)(ech->confBuf - output)` for hello_retry_request (src/tls13.c:8383), and `ech->confBuf` is written ONLY by `TLSX_ECH_Write` when the extension is actually serialized (src/tls.c:14017), which now requires `resp==1`, which now requires `TLSX_ECH_Parse` to have run. No reachable path exists today, so this is hardening rather than a live bug.

**Fix:** Guard the HRR branch in `SendTls13ServerHello` with `if (ech->confBuf != NULL)` before computing `acceptOffset`, so the offset can never be derived from an unwritten confirmation buffer. Optionally add a comment at this `TLSX_SetResponse` noting that the HRR accept-confirmation write depends on it.

examples/server/server.c
```diff
@@ -2652,6 +2764,12 @@ THREAD_RETURN WOLFSSL_THREAD server_test(void* args)
         err_sys_ex(runWithErrors, "Cannot use DTLS with both UDP and SCTP.");
     }
 
+#if defined(WOLFSSL_TLS13) && defined(HAVE_ECH) && defined(HAVE_PK_CALLBACKS)
```
dgarske, line 2767: **🔵 [Low] --ech with -P (PK callbacks) now hard-aborts the server example**

The new `err_sys_ex(catastrophic, "PK callbacks not tested with ECH.")` turns a previously runnable command line into an immediate exit. The underlying reason is real -- `EchSwapToPrivateCert` swaps the cert/key on the SSL object (server.c:1676-1691) without updating `pkCbInfo.ourKey`, which is set once at server.c:3161 -- but the message states neither the constraint nor the workaround, and `catastrophic` is a fairly blunt response to a combination that merely lacks test coverage.

**Fix:** Reword to state the actual constraint (e.g. 'ECH cert swap does not update PK callback key material; -P with --ech is unsupported'), or update `pkCbInfo.ourKey` in the callback so the combination works.

src/ssl_ech.c
```diff
@@ -906,6 +906,107 @@ int GetEchConfigsEx(WOLFSSL_EchConfig* configs, byte* output, word32* outputLen)
...
+    copy->maxNameLen = src->maxNameLen;
+    XMEMCPY(copy->receiverPubkey, src->receiverPubkey, HPKE_Npk_MAX);
+
+    /* don't copy raw and rawLen, these are derived later */
+
+    *out = copy;
+    return 0;
+}
+
+/* copy the public parts of the ech configs into ssl owned memory
```
dgarske, line 960: **⚪ [Info] CopyEchConfigsPub reports success for three undocumented no-op cases**

The function returns 0 for three no-op cases that the leading comment does not describe: no ECH extension present, client side, and already-copied. All three were verified correct. `ssl->options.side` IS reliable at this point (`wolfSSL_accept` sets `SERVER_END` via `InitSSL_Side` before any ClientHello is parsed, ssl_api_hs.c:554). Excluding the client is correct because the client's `ech->echConfig` points at `ssl->echConfigs` (tls.c:16758), which a CTX swap does not affect. The empty-list case sets `echConfig=NULL` with `copiedConfig=1`, which is safe due to the guard at tls.c:15024. Minor efficiency note: the side check runs after `TLSX_Find`, so clients pay an extension-list walk for nothing.

**Fix:** Spell out the no-op cases in the function comment; optionally hoist the `side` check above the `TLSX_Find` call.

.github/scripts/openssl-ech.sh
```diff
@@ -184,11 +214,16 @@ openssl_server(){
         $wolfssl_extra \
         &>> "$TMP_LOG" || [ "$REJECT" -ne 0 ]
 
+    # let s_server finish writing its ech_success= line before grepping;
+    # on reject it sees a fatal alert, so tolerate a nonzero exit
+    wait || [ "$REJECT" -ne 0 ]
```
dgarske, line 219: **⚪ [Info] openssl-ech.sh: operand-less wait always returns 0, making the added || clause dead**

There IS a background job in both functions (line 190 backgrounds s_server, line 269 backgrounds the wolfSSL server), so adding a `wait` is a genuine fix for the grep race the comment describes. However bash's operand-less `wait` returns 0 unconditionally, so the `|| [ "$REJECT" -ne 0 ]` clause is unreachable and the comment describes behavior that never occurs; in particular a `timeout 30` kill (exit 124) is silently ignored. Separately verified as NOT a problem: `MAX_WAIT=50` is an iteration count over `sleep 0.1` (5 s total), so it does not outlast the new 30 s timeout. The new greps are also stronger rather than vacuous -- `ECH status:` comes only from `PrintEchStatus()` (wolfssl/test.h:4930-4946, strings match exactly), and in each mode only one wolfSSL process writes to `$TMP_LOG` while the paired marker (`ech_success=`, `ECH: success: 1`, `ECH: Got 1 retry-configs`) comes from OpenSSL, so the two greps check opposite ends of the handshake.

**Fix:** Capture the PID after the `&` and wait on it specifically, or drop the `||` clause and fix the comment.

.github/scripts/openssl-ech.sh
```diff
@@ -89,18 +96,35 @@ WOLFSSL_CLIENT=${WOLFSSL_CLIENT:-"$WORKSPACE/examples/client/client"}
...
+# ECH public name; matches certs/ech-public-cert.pem, the public-facing cert the
+# outer (or a rejected) handshake authenticates against
+PUB_NAME="public.com"
 MAX_WAIT=50
 
+# --------------------------------------------------------------------------
+# Flip a bit in the HPKE public key of the config the server published. The
+# client will offer ECH, but the server can't decrypt so ECH is rejected.
+# --------------------------------------------------------------------------
+corrupt_ech_config() {
```
dgarske, line 111: **⚪ [Info] corrupt_ech_config depends on bash-4 mapfile and GNU base64/od**

The shebang is `#!/bin/bash` and the workflow runs the script on ubuntu-24.04 (openssl-ech.yml:96,149-177), so `mapfile`, arrays, `local`, `base64 -w 0` and GNU `od` are all fine in CI. It does break local developer runs on macOS (bash 3.2 has no `mapfile`; BSD base64 has no `-w`). Verified correct: the byte offset is right -- 0-1 list len, 2-3 version, 4-5 config len, 6 config_id, 7-8 kem_id, 9-10 pubkey len, so index 11 is the first key byte and bash arrays are 0-indexed; this matches tests/api.c:14306-14312. The `--reject`/`--suite` mutual-exclusion guard is also correct: `SUITE` is initialized to "" at line 22 and set only at line 43, and `set -u` is not in effect. `printf "\\x$b"` is fine since bash's builtin printf supports `\xHH` and `$b` is always two hex digits.

**Fix:** Optional: switch to an xxd-based transform and `base64 | tr -d '\n'`, or add a comment noting that bash 4 plus GNU coreutils are required so macOS developers are not surprised.

tests/api.c
```diff
@@ -15997,6 +16116,185 @@ static int test_wolfSSL_Tls13_ECH_disable_conn(void)
...
+        TEST_SUCCESS);
+
+    return EXPECT_RESULT();
+}
+
+#if defined(OPENSSL_ALL) || defined(OPENSSL_EXTRA)
+/* a test for the wolfSSL_CTX_set_servername_callback with ECH
+ * tests ctx swapping with wolfSSL_set_SSL_CTX
+ *  - This is more tricky because echConfig inheritance must be considered */
+static int test_wolfSSL_Tls13_ECH_multi_ctx(void)
```
dgarske, line 16261: **⚪ [Info] Multi-tenant CTX-swap test drops the SNI dispatch callback for the rest of the connection**

The callback and arg are installed only on `test_ctx.s_ctx` (tests/api.c:16158-16162); the per-tenant CTXs never get `wolfSSL_CTX_set_servername_callback`. Dispatch goes through `ssl->ctx->sniRecvCb` (src/internal.c:43900-43902), so once `wolfSSL_set_SSL_CTX` repoints `ssl->ctx` the multi-tenant dispatch is off for the remainder of the handshake -- which is precisely the HRR call `multi_sni_ex(tenants, &tenants[1], 1, 0)`. The test still passes because `TLSX_SNI_Parse` early-returns on CH2 (src/tls.c:2509-2510), so nothing is asserted post-swap. CTX lifetime was checked and is correct: `wolfSSL_set_SSL_CTX` increments the new refcount and frees the old reference, so calling `test_ssl_memio_cleanup` inside `multi_sni_ex` before the tenant free loop is safe, and the free loop is a plain statement so it still runs after an `Expect` failure -- no leak.

**Fix:** Install the callback and arg on each tenant CTX as well, or add a comment recording that the CTX swap intentionally drops dispatch and that CH2 is handled without re-dispatch.

dgarske (changes requested):
https://github.com/wolfSSL/wolfssl/pull/10913#pullrequestreview-4832810975

JacobBarthelmeh:
@sebastian-carpenter please go through and address comments / merge conflicts. When ready for re-review and merge assign back to wolfSSL-bot and unassigned yourself in the assignees.

PR #10915, opened 2026-07-15
Retry the ISO TP flow control wait up to N_Bs

aidangarske:
`isotp_receive_flow_control()` waits for flow control with a single 100ms `recv_fn` call and treats a miss as fatal, while every other ISO-TP receive in the same file retries (`ISOTP_Receive` loops `do { ... } while (ret == 0)`). A peer that is a few milliseconds late aborts an otherwise healthy transfer.

ISO 15765-2 calls this timeout N_Bs and puts it around 1000ms, so this retries to that deadline instead of giving up after one poll. Still bounded, no behaviour change when flow control arrives promptly.

Honest scope: I found this while investigating a `can-bus` stall in wolfSSL/wolfssl-examples#598 and initially believed it was the cause. It is not — building that example against this branch still fails the same way, so the stall has a different root cause I have not yet identified. This change stands on its own as a correctness fix to the flow control wait, but it does not close that issue.

### DIFF
--- a/src/wolfio.c
+++ b/src/wolfio.c
@@ -3851,9 +3851,21 @@ static int isotp_send_flow_control(struct isotp_wolfssl_ctx *ctx,
 static int isotp_receive_flow_control(struct isotp_wolfssl_ctx *ctx)
 {
     int ret;
+    int waited;
     enum isotp_frame_type type;
     enum isotp_flow_control flow_control;
-    ret = ctx->recv_fn(&ctx->frame, ctx->arg, ISOTP_DEFAULT_TIMEOUT);
+    /* Wait up to N_Bs for flow control rather than a single 100ms poll: a peer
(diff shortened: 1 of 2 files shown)


aidangarske:
Closing this. I opened it believing it was the cause of a can-bus stall found by CI in wolfSSL/wolfssl-examples#598. Testing that example against this branch showed it is not: the client there never calls wolfSSL_read, so flow control was never sent at all and no timeout value would have helped. The real cause was in the example.

The underlying observation may still be worth something to someone with a real ISO-TP peer: isotp_receive_flow_control() waits for flow control with a single 100ms poll and treats a miss as fatal, while every other ISO-TP receive in wolfio.c retries, and ISO 15765-2 puts N_Bs nearer 1000ms. But I have no reproducer for it, so it should not sit here as an open PR on a justification I already disproved.

PR #10912, opened 2026-07-15
MC/DC coverage for wolfCrypt modules - Part 3

danielinux:
# Description

Adds/extends tests/api MC/DC DecisionCoverage + FeatureCoverage and tests/unit-mcdc white-box supplements for the remaining wolfCrypt primitives (first-pass per-module union; gap-closing to follow):

- classic bigint: tfm.c, integer.c (extend test_wolfmath.c; tfm/integer whiteboxes)
- coding.c, error.c, wc_encrypt.c, pwdbased.c
- ed448.c + curve448.c (shared fe_448/ge_448 backends; ed448 whitebox)
- random.c deepened + new wolfentropy.c (MemUse SP800-90B), whiteboxes
- hash.c, kdf.c, signature.c, chacha20_poly1305.c
- LMS/XMSS (wc_lms*.c, wc_xmss*.c) with impl whiteboxes
- ML-KEM (wc_mlkem*.c), ML-DSA (wc_mldsa.c), SLH-DSA (wc_slhdsa.c) with whiteboxes

New tests register in tests/api.c + tests/api/include.am. Frozen-module guards (!HAVE_SELFTEST && !HAVE_FIPS) applied to new random.c/pwdbased.c tests that call post-v4.1.0 symbols.

Follow up to #10845 , #10876

PR #10927, opened 2026-07-16
Hash: consistent error code for uninitialized wc_Hash ops under DEBUG_WOLFSSL

danielinux:
# Description

Under DEBUG_WOLFSSL the hash->type != type check in wc_HashUpdate, wc_HashFinal and wc_HashFree fired for an uninitialized hash (hash->type == WC_HASH_TYPE_NONE), returning BAD_FUNC_ARG where a non-debug build returns HASH_TYPE_E from the type switch, so the returned error code depended on whether DEBUG_WOLFSSL was defined. Only apply the mismatch check to initialized hashes; the genuine init-then-wrong-type misuse check is preserved.

# Testing

Discovered via MC/DC campaign part 3 #10912


danielinux:
Jenkins retest this please

JacobBarthelmeh (approved):

PR #10929, opened 2026-07-16
CMake: fix XMSS wiring and complete LMS/XMSS sub-option parity

Frauschi:
### Problem
The CMake build accepted `-DWOLFSSL_XMSS=1` but never emitted `WOLFSSL_HAVE_XMSS` — neither onto the library compile line nor into the installed `options.h`. Since every XMSS body is wrapped in `#ifdef WOLFSSL_HAVE_XMSS`, the option was a no-op: `wc_xmss.c` compiled empty and consumers saw no XMSS support. LMS had the correct wiring. Autotools (`--enable-xmss`) was unaffected.

While fixing this, CMake was also missing several LMS sub-options that autotools `--enable-lms` already supports.

### Changes

**XMSS (build-break fix)** — `32ab46a`
- `CMakeLists.txt` — add the missing `if (WOLFSSL_XMSS)` block (mirroring LMS), plus `WOLFSSL_XMSSVERIFYONLY` / `WOLFSSL_XMSSSMALL` sub-options for parity with autotools `--enable-xmss=verify-only,small`.
- `cmake/options.h.in` — expose `WOLFSSL_XMSS_VERIFY_ONLY` and `WOLFSSL_WC_XMSS_SMALL`.
- `configure.ac` — drop the unused legacy `XMSS_VERIFY_ONLY` macro (the un-prefixed twin is dead in-tree), so both build systems emit only `WOLFSSL_XMSS_VERIFY_ONLY`.
- `.github/workflows/cmake.yml` — enable `WOLFSSL_XMSS` in the main CMake CI build (it tested LMS but not XMSS — the blind spot that let this through).

**LMS (parity follow-on)** — `f377a54`
- Add the three LMS sub-options CMake lacked: `WOLFSSL_LMSVERIFYONLY`, `WOLFSSL_LMSSMALL`, `WOLFSSL_LMSSHAKE256` (emitting `WOLFSSL_LMS_VERIFY_ONLY`, `WOLFSSL_WC_LMS_SMALL`, `WOLFSSL_LMS_SHAKE256`), matching autotools `--enable-lms=verify-only,small,shake256`.

### Verification
- `-DWOLFSSL_XMSS=1` → `options.h` now emits `#define WOLFSSL_HAVE_XMSS`; `wc_XmssKey_*` symbols exported (were absent before).
- Both `verify-only` builds drop `wc_{Xmss,Lms}Key_Sign`/`MakeKey`, keep `Verify` — matches autotools.
- Sub-options don't leak into `options.h` when their parent (`WOLFSSL_XMSS` / `WOLFSSL_LMS`) is off.
- Generated `options.h` no longer contains the dead bare `XMSS_VERIFY_ONLY`.

### Downstream
Unblocks [wolfPKCS11 #206](https://github.com/wolfSSL/wolfPKCS11/pull/206), whose CMake CI clones wolfSSL master and requires `WOLFSSL_HAVE_XMSS`.


Frauschi:
Jenkins retest this please

JacobBarthelmeh (approved):

PR #10931, opened 2026-07-16
socat CI: prevent rare shard hang in inverted OPENSSL tests

julek-wolfssl:
Fixes a rare socat shard hang that stalls the job until the 15-minute timeout kills it with no diagnosable output. Seen once on a run whose re-run passed; reproduced deterministically by delaying the listener past the retry budget.

### Root cause
The inverted `testserversec` OPENSSL tests (`OPENSSLCERTCLIENT`, the commonname test and the fips test) run a foreground one-shot SSL listener with no port-wait, while their background connector exits after ~2s (`retry=2`, 1s apart). On a loaded runner the listener can bind *after* the connector is already gone, then block in `accept()` forever. Nothing bounds the shard, so it stalls until the job-level timeout fires and the log says nothing about which test was running.

### Fix
- **Widen the connector budget to `retry=30`.** Retries only happen while the listener is not yet up, and `testserversec` kills the connector as soon as the client returns, so passing runs are unaffected — this only buys the listener a wider window to win the race.
- **Bound each shard with `timeout(1)` at 10 minutes** (normal shard time is ~2 min). Any future hang then fails fast, and the shard output names the test that was running instead of dying to the opaque job-level kill.

### DIFF
--- a/.github/workflows/socat.yml
+++ b/.github/workflows/socat.yml
@@ -127,17 +127,12 @@ jobs:
           # (|| true), errors left visible so a runner without IPv6 still runs the
           # IPv4 tests and any failure stays diagnosable in the log.
           #
-          # The inverted OPENSSL testserversec tests (OPENSSLCERTCLIENT etc.)
-          # run a foreground one-shot SSL listener with no port-wait; their
-          # background connector gives up after 2s (retry=2, 1s apart). On a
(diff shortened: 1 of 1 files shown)


julek-wolfssl:
retest this please

julek-wolfssl:
retest this please

SparkiDev (approved):

PR #10937, opened 2026-07-16
Fix PKCS7 SignerIdentifier SKID to implicit [0] tagging

embhorn:
# Description

PKCS7_EncodeSigned emitted the SignedData SignerIdentifier subjectKeyIdentifier as an explicit [0] wrapper around an OCTET STRING (0xA0 len 0x04 len value). RFC 5652 defines the CMS module with IMPLICIT TAGS, so the [0] choice must use an implicit context tag that replaces the OCTET STRING tag (0x80 len value). Strict DER parsers such as OpenSSL reject the explicit form.

Encode with SetImplicit(ASN_OCTET_STRING, ...) and remove the now unused inner OCTET STRING header and its ESD struct members. The decoder already accepts both the explicit and implicit forms, so SignedData produced by earlier releases still verifies.

Fixes zd22160

# Testing

Add a regression test asserting the implicit wire form and round-trip verify, plus a decode test that verifies a captured explicit-form fixture so the backward-compatible parse path stays covered.

# Checklist

 - [x] added tests
 - [ ] updated/added doxygen


Frauschi:
Jenkins retest this please

Frauschi (approved):
LGTM

dgarske (approved):

PR #10940, opened 2026-07-17
ChaCha20/Poly1305 ASM: AVX512 and improvements to other Intel x64

SparkiDev:
# Description

Add AVX512 implementations of ChaCha20 and Poly1305. Add fused asm implementations of ChaCha20-Poly1305. Wire the fused APIs into new API.
Use new API in TLS 1.2 and TLS 1.3.

# Testing

Tested ChaCha20/Poly1305


SparkiDev:
Code generated with PR:
https://github.com/wolfSSL/scripts/pull/630

SparkiDev:
Jenkins: retest this please

dgarske (commented):
Host: tigerlake-come (i7-1185G7E, Tiger Lake, AVX-512), pinned to core 2, --enable-intelasm, 3-pass average, base-10 MB/s.

```
┌─────────────────────────────┬──────────────────────────┬─────────────────────┬─────────┬────────────────────┐
│          Algorithm          │     Before (parent       │     After (PR       │ Speedup │   cpb before ->    │
│                             │        c1ee61c8d)        │     98a1f6313)      │         │       after        │
├─────────────────────────────┼──────────────────────────┼─────────────────────┼─────────┼────────────────────┤
│ ChaCha20                    │ 3,672 MB/s               │ 6,733 MB/s          │ 1.83x   │ 0.76 -> 0.42       │
├─────────────────────────────┼──────────────────────────┼─────────────────────┼─────────┼────────────────────┤
│ Poly1305                    │ 7,767 MB/s               │ 12,216 MB/s         │ 1.57x   │ 0.36 -> 0.23       │
├─────────────────────────────┼──────────────────────────┼─────────────────────┼─────────┼────────────────────┤
│ ChaCha20-Poly1305 AEAD      │ 2,503 MB/s               │ 4,898 MB/s          │ 1.96x   │ 1.12 -> 0.57       │
│ (enc)                       │                          │                     │         │                    │
└─────────────────────────────┴──────────────────────────┴─────────────────────┴─────────┴────────────────────┘
```

wolfcrypt/src/chacha_asm.asm
```diff
@@ -1423,4 +1766,9225 @@ L_chacha20_avx2_end256:
...
+        vpmadd52huq	zmm27, zmm16, zmm19
+        vpmadd52huq	zmm27, zmm17, zmm23
+        vpmadd52huq	zmm27, zmm18, zmm22
+        vpmadd52luq	zmm25, zmm16, zmm20
+        vpmadd52luq	zmm25, zmm17, zmm19
+        vpmadd52luq	zmm25, zmm18, zmm23
+        vpmadd52huq	zmm28, zmm16, zmm20
+        vpmadd52huq	zmm28, zmm17, zmm19
+        vpmadd52huq	zmm28, zmm18, zmm23
+
```
dgarske: ### CRITICAL-1: MASM chacha20_poly1305_small_enc reads aadLen and tag from wrong stack offsets (inside its own xmm save area)
- **File:** `wolfcrypt/src/chacha_asm.asm:10017`
- **Function:** `chacha20_poly1305_small_enc`
- **Action:** BLOCK
- **Tag:** bug
- **Confidence:** High

**Description:** The Windows/MASM prologue pushes 8 registers (r12,r13,r14,r15,rdi,rsi,rbx,rbp = 64 bytes) then `sub rsp, 352`. Args 5 and 6 are correctly fetched BEFORE the sub (`mov r10,[rsp+104]` = entry+40 = mLen, `mov r11,[rsp+112]` = entry+48 = aad). But args 7 (aadLen) and 8 (tag) are fetched AFTER the sub at `[rsp+280]` and `[rsp+288]`. On Windows x64 the correct offsets are `[rsp+472]` and `[rsp+480]` (56/64 arg slot + 64 push area + 352 frame). The emitted offsets land squarely in this function's own nonvolatile xmm save area: xmm11 is stored at `[rsp+272..287]` and xmm12 at `[rsp+288..303]`. So `mov eax,[rsp+280]` loads bytes 8-11 of the caller's saved xmm11 as `aadLen`, and `mov rax,[rsp+288]` loads the caller's saved xmm12 low qword as the `tag` pointer. Both are spilled to locals 128/152 and then used as the AAD hashing loop length and, in the epilogue, `mov QWORD PTR [rbx],r15` / `[rbx+8]` writes the 16-byte MAC through that bogus pointer. Result: out-of-bounds AAD read with an arbitrary length plus a 16-byte wild write. The GAS twin is correct (6 pushes + `subq $0xc0`=192 gives `movl 248(%rsp)` / `movq 256(%rsp)` = entry+8/entry+16 for SysV). The generator appears to have used a 160-byte frame instead of 352 when computing arg7/arg8 (56+64+160=280, 64+64+160=288). This is reachable on any AVX2 Windows x86-64 build: `chacha20_poly1305.c` gates it only on `WOLFSSL_CHACHA20_POLY1305_SHORT` + `IS_INTEL_AVX2()` with `sz <= 64`, and `chacha_asm.asm` wraps it in `IFDEF HAVE_INTEL_AVX2` which the file auto-defines. Independently confirmed by compiling `offsetof` probes: all struct offsets used by the new asm are valid, so this is purely a stack-argument computation defect.

**Code:**
```
chacha20_poly1305_small_enc PROC
        push r12 / r13 / r14 / r15 / rdi / rsi / rbx / rbp   ; 8 pushes = 64 bytes
        mov  r10, QWORD PTR [rsp+104]        ; arg5 mLen  (correct)
        mov  r11, QWORD PTR [rsp+112]        ; arg6 aad   (correct)
        sub  rsp, 352
        vmovdqu OWORD PTR [rsp+272], xmm11   ; occupies 272..287
        vmovdqu OWORD PTR [rsp+288], xmm12   ; occupies 288..303
        ...
        mov  eax, DWORD PTR [rsp+280]        ; BUG: aadLen; should be [rsp+472]
        mov  QWORD PTR [rsp+128], rax
        mov  rax, QWORD PTR [rsp+288]        ; BUG: tag;    should be [rsp+480]
        mov  QWORD PTR [rsp+152], rax
```

**Recommendation:** Fix the generator's Windows stack-argument offset computation for args 7+ so it adds the local frame size in addition to the push area and the Windows arg-slot base. Preferably, match how args 5/6 are already handled and fetch args 7/8 into scratch registers before `sub rsp, 352`. Add an ml64 assemble + run step to CI for the .asm twins.

dgarske: ### CRITICAL-2: MASM chacha20_poly1305_small_dec reads aadLen and received tag from wrong stack offsets
- **File:** `wolfcrypt/src/chacha_asm.asm:10510`
- **Function:** `chacha20_poly1305_small_dec`
- **Action:** BLOCK
- **Tag:** bug
- **Confidence:** High

**Description:** Same defect as `chacha20_poly1305_small_enc` with the decrypt frame size. 8 pushes (64 bytes) then `sub rsp, 328`. Args 5/6 fetched correctly before the sub at `[rsp+104]`/`[rsp+112]`; args 7 (aadLen) and 8 (tag) fetched after the sub at `[rsp+256]` and `[rsp+264]`, where the correct offsets are `[rsp+448]` and `[rsp+456]` (56/64 + 64 + 328). The emitted offsets fall inside this function's own xmm save area: xmm11 at `[rsp+248..263]`, xmm12 at `[rsp+264..279]`. The epilogue then does `mov rbx, QWORD PTR [rsp+144]` followed by `xor r15, QWORD PTR [rbx]` / `xor rdi, QWORD PTR [rbx+8]` -- dereferencing a garbage tag pointer. Because the function's return value is the constant-time verify result consumed by `chacha20_poly1305_decrypt_short()`, a wrong compare can cause a bad tag to be accepted or a good tag rejected, in addition to the likely access violation. The GAS twin is correct (6 pushes + `subq $0xa8`=168 gives 224/232 = entry+8/entry+16).

**Code:**
```
chacha20_poly1305_small_dec PROC
        push r12/r13/r14/r15/rdi/rsi/rbx/rbp   ; 64 bytes
        mov  r10, QWORD PTR [rsp+104]          ; arg5 ctLen (correct)
        mov  r11, QWORD PTR [rsp+112]          ; arg6 aad   (correct)
        sub  rsp, 328
        vmovdqu OWORD PTR [rsp+248], xmm11
        vmovdqu OWORD PTR [rsp+264], xmm12
        ...
        mov  eax, DWORD PTR [rsp+256]          ; BUG: aadLen; should be [rsp+448]
        mov  rax, QWORD PTR [rsp+264]          ; BUG: tag;    should be [rsp+456]
        mov  QWORD PTR [rsp+144], rax
        ...
        mov  rbx, QWORD PTR [rsp+144]
        xor  r15, QWORD PTR [rbx]              ; deref of garbage pointer
```

**Recommendation:** Emit `[rsp+448]` / `[rsp+456]`, or fetch args 7 and 8 before the `sub rsp, 328` as is already done for args 5 and 6.

wolfcrypt/src/chacha_asm.asm
```diff
@@ -1423,4 +1766,9225 @@ L_chacha20_avx2_end256:
...
+        vprold	xmm20, xmm20, 12
+        vprold	xmm21, xmm21, 12
+        vprold	xmm22, xmm22, 12
+        vprold	xmm23, xmm23, 12
+        vpaddd	xmm16, xmm16, xmm20
+        vpaddd	xmm17, xmm17, xmm21
+        vpaddd	xmm18, xmm18, xmm22
+        vpaddd	xmm19, xmm19, xmm23
+        vpxord	xmm28, xmm28, xmm16
+        vpxord	xmm29, xmm29, xmm17
```
dgarske, line 3148: ### HIGH-3: GAS-suffixed `bsrq` / `btq` emitted into the MASM file - poly1305_asm.asm will not assemble
- **File:** `wolfcrypt/src/poly1305_asm.asm:3144,3217`
- **Function:** `poly1305_fold_avx512ifma`
- **Action:** BLOCK
- **Tag:** bug
- **Confidence:** High

**Description:** Two instructions in the new `poly1305_fold_avx512ifma` PROC keep the AT&T `q` size suffix that the generator strips everywhere else: `bsrq rax, rdx` and `btq rax, rdx`. MASM/ml64 has no `BSRQ` or `BTQ` mnemonic -- the mnemonics are `BSR` and `BT`, with operand size taken from the register operands. These lines fail to assemble, breaking the whole `poly1305_asm.asm` translation unit and therefore any Windows x86-64 intelasm build. A full mnemonic inventory of `chacha_asm.asm` + `poly1305_asm.asm` confirms every other GAS-suffixed form is correctly translated (movq->mov, shrq->shr, shrdq->shrd, addq->add, negq->neg, decq->dec, leaq->lea), and `bsrq`/`btq` are the only two suffixed GPR mnemonics left in any .asm under wolfcrypt/src -- so this is a missing generator table entry, not an accepted alias. The GAS twin is correct: `poly1305_asm.S:3326 bsrq %rsi,%rax` and `poly1305_asm.S:3399 btq %rdx,%rax`.

**Code:**
```
poly1305_asm.asm:3144:        bsrq	rax, rdx      ; ml64: invalid mnemonic
poly1305_asm.asm:3217:        btq	rax, rdx      ; ml64: invalid mnemonic
```

**Recommendation:** Add `bsr` and `bt` to the generator's GAS->MASM mnemonic table so the `q` suffix is stripped, and re-generate `poly1305_asm.asm`. Add an ml64 assemble step to CI for the .asm twins so this class of defect is caught.

wolfcrypt/src/chacha_asm.asm
```diff
@@ -459,6 +435,373 @@ L_chacha_x64_done:
...
+ptr_L_chacha20_sse3_one QWORD L_chacha20_sse3_one
+_DATA ENDS
+_TEXT SEGMENT READONLY PARA
+chacha_encrypt_sse3 PROC
+        push	r12
+        push	r13
+        push	r14
+        push	r15
+        sub	rsp, 160
+        vmovdqu	OWORD PTR [rsp], xmm6
```
dgarske: ### HIGH-4: MASM chacha_encrypt_sse3 uses VEX-encoded vmovdqu but is dispatched on SSSE3-only (pre-AVX) CPUs
- **File:** `wolfcrypt/src/chacha_asm.asm:464-473,786-795`
- **Function:** `chacha_encrypt_sse3`
- **Action:** BLOCK
- **Tag:** bug
- **Confidence:** High

**Description:** `chacha_encrypt_sse3` is guarded by `IFDEF HAVE_INTEL_SSSE3` and its body is pure legacy SSE/SSSE3 (`movdqu`, `movdqa`, `pshufb`, `paddd`, `pxor`, `pslld`/`psrld`, `pshufd`) -- the GAS twin at `wolfcrypt/src/chacha_asm.S:511-836` contains zero VEX instructions. But the MASM version saves and restores the Windows nonvolatile xmm6-xmm15 with the AVX (VEX-encoded) form `vmovdqu OWORD PTR [rsp+N], xmmK` in the prologue and `vmovdqu xmmK, OWORD PTR [rsp+N]` in the epilogue. VEX encodings require AVX (CPUID.1:ECX.AVX plus OSXSAVE/XCR0), which an SSSE3-only part does not have. `wolfcrypt/src/chacha.c` dispatches here precisely when AVX is absent: the trailing `else if (IS_INTEL_SSSE3(cpuidFlags))` at chacha.c:524 is only reached after the AVX2 and AVX1 branches fail, and the new short-message branch at chacha.c:472 is also reachable with SSSE3 only. So on Core 2 / Nehalem / Westmere / Bonnell-Silvermont-class Windows hosts the very first instruction of the prologue raises #UD. The established repo convention for non-AVX MASM is the legacy encoding -- see `wolfcrypt/src/aes_gcm_asm.asm:57` using `movdqu OWORD PTR [rsp], xmm6` in the AES-NI (non-AVX) path.

**Code:**
```
chacha_encrypt_sse3 PROC
        push    r12 / r13 / r14 / r15
        sub     rsp, 160
        vmovdqu OWORD PTR [rsp], xmm6        ; VEX -> #UD on an SSSE3-only CPU
        vmovdqu OWORD PTR [rsp+16], xmm7
        ...
        vmovdqu OWORD PTR [rsp+144], xmm15
```

**Recommendation:** Emit legacy `movdqu` (not `vmovdqu`) for the xmm6-15 save/restore in any PROC gated on `HAVE_INTEL_SSSE3`, matching `aes_gcm_asm.asm`. If instead the intent is that this path only runs on AVX-capable CPUs, the chacha.c dispatch must also require `IS_INTEL_AVX1()` -- but that defeats the stated purpose of an SSSE3 baseline path.

wolfssl/wolfcrypt/poly1305.h
```diff
@@ -52,6 +52,29 @@
 #if defined(USE_INTEL_SPEEDUP) && !defined(NO_POLY1305_ASM)
     #define USE_INTEL_POLY1305_SPEEDUP
     #define HAVE_INTEL_AVX1
+    /* 8-way AVX-512 path.  Enabling it appends r^5..r^8 to the state (see the
+     * struct below); define WOLFSSL_POLY1305_NO_AVX512 to keep the smaller
+     * state and drop the path. */
+    #if !defined(WOLFSSL_POLY1305_NO_AVX512)
+        #define WOLFSSL_POLY1305_AVX512
```
dgarske, line 60: ### HIGH-5: Feature macros in poly1305.h ignore NO_AVX512_SUPPORT / NO_AVX2_SUPPORT - those builds fail to link
- **File:** `wolfssl/wolfcrypt/poly1305.h:55-78`
- **Function:** `n/a (preprocessor configuration)`
- **Action:** BLOCK
- **Tag:** bug
- **Confidence:** High

**Description:** `chacha_asm.S` and `poly1305_asm.S` gate their new kernels with the established tree-wide macros: `#ifndef NO_AVX512_SUPPORT / #define HAVE_INTEL_AVX512` (chacha_asm.S:54-58, poly1305_asm.S:54-58) and `#ifndef NO_AVX2_SUPPORT / #define HAVE_INTEL_AVX2` (chacha_asm.S:44-48). `wolfcrypt/src/poly1305.c:92` correctly mirrors this with `#if defined(WOLFSSL_POLY1305_AVX512) && !defined(NO_AVX512_SUPPORT)`. But the new feature macros in the public header do NOT: `WOLFSSL_POLY1305_AVX512` is gated only on `WOLFSSL_POLY1305_NO_AVX512` (a different macro), and `WOLFSSL_CHACHA20_POLY1305_FUSED` / `_FUSED_IFMA` reference neither `NO_AVX512_SUPPORT` nor `NO_AVX2_SUPPORT`. Likewise `WOLFSSL_CP_SMALL_ASM` in `chacha20_poly1305.c:257` is gated only on `USE_INTEL_SPEEDUP && WOLFSSL_X86_64_BUILD && !WOLFSSL_NO_CHACHA20_POLY1305_SMALL_ASM`. Consequences on x86-64 with `--enable-intelasm`: (a) `-DNO_AVX512_SUPPORT` still defines `WOLFSSL_CHACHA20_POLY1305_FUSED` and `_FUSED_IFMA`, so chacha20_poly1305.c declares and calls `chacha20_poly1305_avx512`, `chacha20_poly1305_ifma`, `chacha20_poly1305_ifma_decrypt`, `poly1305_calc_powers_avx512ifma` and `poly1305_fold_avx512ifma`, none of which the .S files emit -> undefined references; (b) `-DNO_AVX2_SUPPORT` still defines `WOLFSSL_CP_SMALL_ASM` and `WOLFSSL_CHACHA20_POLY1305_FUSED`, so `chacha20_poly1305_small_enc`, `chacha20_poly1305_small_dec` and `poly1305_calc_powers_avx2` are referenced but not emitted -> undefined references. `NO_AVX512_SUPPORT` and `NO_AVX2_SUPPORT` are long-standing supported knobs used by sp_x86_64_asm.S, sha512_asm.S, aes_gcm_asm.S, wc_mldsa_asm.S and others, so these are real supported configurations.

**Code:**
```
/* wolfssl/wolfcrypt/poly1305.h - missing the NO_*_SUPPORT guards */
    #if !defined(WOLFSSL_POLY1305_NO_AVX512)
        #define WOLFSSL_POLY1305_AVX512
    #endif
    #if !defined(WOLFSSL_NO_CHACHA20_POLY1305_FUSED)
        #define WOLFSSL_CHACHA20_POLY1305_FUSED
    #endif
    #if defined(WOLFSSL_POLY1305_AVX512) && \
        !defined(WOLFSSL_NO_CHACHA20_POLY1305_FUSED_IFMA)
        #define WOLFSSL_CHACHA20_POLY1305_FUSED_IFMA
    #endif
```

**Recommendation:** Mirror the `NO_AVX512_SUPPORT` / `NO_AVX2_SUPPORT` guards used by the .S files in poly1305.h and in the `WOLFSSL_CP_SMALL_ASM` guard, and add `--enable-intelasm CFLAGS=-DNO_AVX512_SUPPORT` and `-DNO_AVX2_SUPPORT` to the CI build matrix.

wolfcrypt/src/chacha20_poly1305.c
```diff
@@ -41,6 +42,558 @@ or Authenticated Encryption with Additional Data (AEAD) algorithm.
...
+WOLFSSL_LOCAL void poly1305_fold_avx512ifma(Poly1305* ctx, word32 nBlocks);
+#ifdef __cplusplus
+}
+#endif
+
+/* Minimum length to stitch.  The kernel processes 1024-byte units and a
+ * once-per-op power precompute, and any sub-1024 remainder is authenticated by
+ * the slower scalar Poly1305; below this the two-pass wins (measured crossover
+ * on Zen5).  Above it the stitch wins 1.1-1.4x, growing with size. */
+#ifndef CHACHA20_POLY1305_STITCH_MIN
```
dgarske, line 152: ### HIGH-6: CHACHA20_POLY1305_STITCH_MIN is user-overridable but values < 1024 cause the IFMA kernel to run off the end of the buffer
- **File:** `wolfcrypt/src/chacha20_poly1305.c:140-142`
- **Function:** `chacha20_poly1305_stitch_chunk / chacha20_poly1305_ifma`
- **Action:** BLOCK
- **Tag:** bug
- **Confidence:** High

**Description:** `CHACHA20_POLY1305_STITCH_MIN` is declared with `#ifndef` (documented as a tunable) and every stitch entry point computes `bulk = sz & ~(word32)0x3ff` then passes `bulk` straight to the kernel. The assembly kernels have NO entry guard for a short or zero length: in `chacha20_poly1305_ifma` the first length test is `subl $0x400, %r8d; jz L_cp512_epi` which occurs only AFTER a full 1024-byte chunk has already been encrypted and hashed; `chacha20_poly1305_ifma_decrypt` is a do-while (`jnz L_cp512_dstart` at the bottom) with the same property; `chacha20_poly1305_avx512` has `subl $0x100, %r8d; jz` after a full 256-byte chunk. So if a user sets `-DCHACHA20_POLY1305_STITCH_MIN=512` (or anything < 1024), a 512..1023-byte message yields `bulk == 0`, the kernel still processes 1024 bytes of out-of-bounds memory in both directions, `subl $0x400` underflows `%r8d` to 0xFFFFFC00, `jz` is not taken, and the loop runs ~4 million more iterations. That is arbitrary OOB read plus OOB write. The default (4096) is safe, and nothing in-tree lowers it, but the macro is exposed with no documented lower bound and no compile-time or runtime check. The same latent hazard applies to the hard-coded `inPlaintextLen >= 256` gate on `chacha20_poly1305_encrypt_fused`.

**Code:**
```
#ifndef CHACHA20_POLY1305_STITCH_MIN
#define CHACHA20_POLY1305_STITCH_MIN 4096
#endif
...
    if (sz >= CHACHA20_POLY1305_STITCH_MIN &&
            chacha20_poly1305_use_fused_ifma()) {
        word32 bulk = sz & ~(word32)0x3ff;   /* == 0 if STITCH_MIN < 1024 */
        ...
        ret = chacha20_poly1305_stitch_chunk(chacha, poly, in, out, bulk, 0);

/* kernel: first length check is AFTER one full 1024-byte chunk */
chacha20_poly1305_ifma:
        ... 1024 bytes encrypted + hashed unconditionally ...
        subl    $0x400, %r8d
        jz      L_cp512_epi
```

**Recommendation:** Add the `#error` guard for `CHACHA20_POLY1305_STITCH_MIN < 1024` and an early `if (bulk == 0) return 0;` in `chacha20_poly1305_stitch_chunk()`. Document the 1024-byte-unit constraint next to the macro.

wolfcrypt/src/chacha20_poly1305.c
```diff
@@ -41,6 +42,558 @@ or Authenticated Encryption with Additional Data (AEAD) algorithm.
...
+            CHACHA20_POLY1305_AEAD_INITIAL_COUNTER + 1);
+    if (ret == 0)
+        ret = wc_Poly1305SetKey(poly, polyKey, sizeof(polyKey));
+    ForceZero(polyKey, sizeof(polyKey));
+    if (ret != 0)
+        return ret;
+
+#ifdef WOLFSSL_CHACHA20_POLY1305_FUSED_IFMA
+    if (sz >= CHACHA20_POLY1305_STITCH_MIN &&
+            chacha20_poly1305_use_fused_ifma()) {
```
dgarske, line 468: ### MEDIUM-7: AVX-512VL CPUs take the scalar path for msglen <= 64, making the AVX1 fast path a performance inversion
- **File:** `wolfcrypt/src/chacha.c:443-462`
- **Function:** `wc_Chacha_Process`
- **Action:** SUGGEST
- **Tag:** bug
- **Confidence:** High

**Description:** The AVX1 one-block fast path is guarded with `IS_INTEL_AVX512VL(cpuidFlags) == 0`, and its comment says "The AVX-512VL path already uses SIMD for one block; match that here" -- i.e. the intent is that AVX-512VL hosts fall through to `chacha_encrypt_avx512vl` at line 502. But the very next statement is an unconditional `if (msglen <= CHACHA_CHUNK_BYTES) { chacha_encrypt_x64(...); return 0; }`, which catches every `msglen <= 64` case before line 502 can be reached. Net effect: on an AVX-512VL CPU a 32- or 64-byte ChaCha (exactly the per-record Poly1305 key derivation in the two-pass ChaCha20-Poly1305 path, and the TLS-record hot path this PR is optimising) runs the SCALAR kernel, while an AVX1-only CPU gets the faster AVX1 kernel. The `IS_INTEL_AVX512VL == 0` guard is therefore not just useless but actively harmful, and the `msglen < 8 * CHACHA_CHUNK_BYTES` AVX-512VL branch is dead for msglen <= 64.

**Code:**
```
#if defined(HAVE_INTEL_AVX1) && !defined(WOLFSSL_LINUXKM)
    if (msglen <= CHACHA_CHUNK_BYTES && IS_INTEL_AVX512VL(cpuidFlags) == 0 &&
            IS_INTEL_AVX1(cpuidFlags)) {
        SAVE_VECTOR_REGISTERS(return _svr_ret;);
        chacha_encrypt_avx1(ctx, input, output, msglen);
        RESTORE_VECTOR_REGISTERS();
        return 0;
    }
#endif
    /* unconditional - swallows the AVX-512VL <=64 case */
    if (msglen <= CHACHA_CHUNK_BYTES) {
        chacha_encrypt_x64(ctx, input, output, msglen);
        return 0;
    }
```

**Recommendation:** Add an explicit AVX-512VL one-block branch before the unconditional scalar fallback (or drop the now-pointless `IS_INTEL_AVX512VL == 0` condition and reorder), and re-benchmark the <=64-byte case on an AVX-512VL part to confirm the intended dispatch.

wolfcrypt/src/chacha_asm.asm
```diff
@@ -1423,4 +1766,9225 @@ L_chacha20_avx2_end256:
...
+ALIGN 16
+L_chacha20_avx512_add QWORD 0000000100000000h, 0000000300000002h
+        QWORD 0000000500000004h, 0000000700000006h
+        QWORD 0000000900000008h, 0000000b0000000ah
+        QWORD 0000000d0000000ch, 0000000f0000000eh
+ptr_L_chacha20_avx512_add QWORD L_chacha20_avx512_add
+_DATA ENDS
+_TEXT SEGMENT READONLY PARA
+chacha_encrypt_avx512 PROC
+        sub	rsp, 168
```
dgarske, line 2389: ### MEDIUM-8: MASM chacha_encrypt_avx512 calls chacha_encrypt_avx2 without reserving 32-byte home space; saved xmm6/xmm7 live inside it
- **File:** `wolfcrypt/src/chacha_asm.asm:2387,2641`
- **Function:** `chacha_encrypt_avx512`
- **Action:** SUGGEST
- **Tag:** bug
- **Confidence:** High

**Description:** The MASM prologue does `sub rsp, 168` (no pushes) and stores nonvolatile xmm6..xmm15 at `[rsp+8]`..`[rsp+152]`. At the end it does `call chacha_encrypt_avx2` and only afterwards reloads xmm6-15. Stack alignment is fine (entry rsp == 8 mod 16, minus 168 gives rsp == 0 mod 16 at the call). But the Windows x64 ABI requires the CALLER to reserve 32 bytes of home/shadow space that the callee may freely spill its four register parameters into; relative to rsp at the `call` that is `[rsp+0, rsp+32)`. Here `[rsp+8..24)` holds the saved xmm6 and `[rsp+24..40)` holds the saved xmm7, so the whole xmm6 slot and the low half of the xmm7 slot are inside the callee's home space. It is benign only because the current `chacha_encrypt_avx2` never writes its home area -- an instrumented, recompiled or future version would silently corrupt the caller's nonvolatile xmm6/xmm7. The GAS twin needs no equivalent (SysV has no home space) and its `subq $8,%rsp` is purely for alignment.

**Code:**
```
chacha_encrypt_avx512 PROC
        sub     rsp, 168
        vmovdqu OWORD PTR [rsp+8], xmm6     ; 8..23  -> inside callee home space
        vmovdqu OWORD PTR [rsp+24], xmm7    ; 24..39 -> half inside home space
        ...
L_chacha20_avx512_end512:
        call    chacha_encrypt_avx2         ; callee owns [rsp+0, rsp+32)
        vzeroupper
        vmovdqu xmm6, OWORD PTR [rsp+8]     ; may reload corrupted data
```

**Recommendation:** Reserve 32 bytes of home space below the saved registers for any MASM PROC that emits a `call`, keeping rsp 16-byte aligned at the call. Apply the rule generally in the generator.

wolfcrypt/src/poly1305_asm.S
```diff
@@ -1142,6 +1142,2447 @@ L_poly1305_avx2_final_cmp_copy:
...
+.align	16
+poly1305_setkey_avx512ifma:
+#else
+.section	__TEXT,__text
+.globl	_poly1305_setkey_avx512ifma
+.p2align	4
+_poly1305_setkey_avx512ifma:
+#endif /* __APPLE__ */
+#ifndef __APPLE__
+        callq	poly1305_setkey_avx@plt
```
dgarske, line 2881: ### LOW-9: poly1305_setkey_avx512ifma returns with dirty zmm upper state (no vzeroupper)
- **File:** `wolfcrypt/src/poly1305_asm.S:2861`
- **Function:** `poly1305_setkey_avx512ifma`
- **Action:** SUGGEST
- **Tag:** bug
- **Confidence:** High

**Description:** The new `poly1305_setkey_avx512ifma` zeroes 192 bytes of `ifma_h` with three 512-bit stores (`vpxorq %zmm0,%zmm0,%zmm0` then `vmovdqu64 %zmm0, 752/816/880(%rdi)`) and returns without `vzeroupper`. Same in the MASM twin (`wolfcrypt/src/poly1305_asm.asm:2729-2735`). Every other new AVX-512 function in this PR (`poly1305_blocks_avx512`, `poly1305_final_avx512`, `poly1305_blocks_avx512ifma`, `poly1305_final_avx512ifma`, `chacha_encrypt_avx512`, `chacha20_poly1305_avx512`, `chacha20_poly1305_ifma`, `chacha20_poly1305_ifma_decrypt`) does emit `vzeroupper` on its return path. Leaving the upper 384 bits of zmm0 non-zero puts the core in the dirty-upper state, so the next legacy-SSE code pays the AVX-SSE transition penalty (tens of cycles each way on Intel) -- and this PR newly introduces legacy-SSE code (`chacha_encrypt_sse3`) into the same call graph, alongside AES-NI. This mirrors the pre-existing `poly1305_setkey_avx2` (ymm0, also no vzeroupper), so it is a generator-wide gap, but the code is new here.

**Code:**
```
poly1305_setkey_avx512ifma:
        callq   poly1305_setkey_avx
        vpxorq  %zmm0, %zmm0, %zmm0
        vmovdqu64 %zmm0, 752(%rdi)
        vmovdqu64 %zmm0, 816(%rdi)
        vmovdqu64 %zmm0, 880(%rdi)
        movq    $0x00, 608(%rdi)
        movw    $0x00, 616(%rdi)
        repz retq              /* no vzeroupper */
```

**Recommendation:** Emit `vzeroupper` before the `ret` in `poly1305_setkey_avx512ifma` (both .S and .asm), and ideally make the generator emit it for any function that touches ymm/zmm. Alternatively use four 128-bit `vmovdqu` stores per 64-byte range, which do not dirty the upper state.

wolfssl/wolfcrypt/chacha20_poly1305.h
```diff
@@ -104,6 +104,20 @@ int wc_ChaCha20Poly1305_Decrypt(
                 const byte inAuthTag[CHACHA20_POLY1305_AEAD_AUTHTAG_SIZE],
                 byte* outPlaintext);
 
+/* Encrypt+authenticate one message with pre-keyed ChaCha20 and Poly1305
```
dgarske, line 107: No API test coverage for new API's

wolfcrypt/src/chacha20_poly1305.c
```diff
@@ -41,6 +42,558 @@ or Authenticated Encryption with Additional Data (AEAD) algorithm.
...
+    if (ret == 0)   /* message data starts at counter 1 */
+        ret = wc_Chacha_SetIV(chacha, nonce,
+            CHACHA20_POLY1305_AEAD_INITIAL_COUNTER + 1);
+    if (ret == 0)
+        ret = wc_Poly1305SetKey(poly, polyKey, sizeof(polyKey));
+    ForceZero(polyKey, sizeof(polyKey));
+    if (ret != 0)
+        return ret;
+
+#ifdef WOLFSSL_CHACHA20_POLY1305_FUSED_IFMA
```
dgarske, line 466: ### MEDIUM-11: No ChaCha20-Poly1305 test vector reaches 4096 bytes - the default-on IFMA stitch path is never executed by any test
- **File:** `wolfcrypt/src/chacha20_poly1305.c:441,531`
- **Function:** `chacha20_poly1305_stitch_chunk / chacha20_poly1305_encrypt_fused_ifma`
- **Action:** SUGGEST
- **Tag:** test
- **Confidence:** High

**Description:** Every new IFMA stitch entry point is gated on `sz >= CHACHA20_POLY1305_STITCH_MIN` (4096). The largest ChaCha20-Poly1305 message in the entire test tree is 265 bytes (`wolfcrypt/test/test.c:11938`, exercised at `test.c:12142`); the largest randomised one is 1024 (`tests/api/test_chacha20_poly1305.c:350` `MC_CHACHA20P1305_MAX_SZ 1024`, used at :386). Nothing anywhere reaches 4096. That leaves these blocks with ZERO execution in CI: the FUSED_IFMA branch in `wc_ChaCha20Poly1305_Encrypt`, the same branch in `Encrypt_ex`, the decrypt stitch in `Decrypt_ex`, `chacha20_poly1305_encrypt_fused_ifma()`, `chacha20_poly1305_stitch_chunk()` in both directions, the `poly1305_fold_avx512ifma()` AAD fold, and the sub-1024 scalar tail. This is not dead code in shipped builds: `wolfssl/wolfcrypt/poly1305.h:74-76` turns `WOLFSSL_CHACHA20_POLY1305_FUSED_IFMA` ON BY DEFAULT for any `USE_INTEL_SPEEDUP` + `WOLFSSL_X86_64_BUILD` (i.e. `--enable-intelasm`) build, runtime-gated only on AVX-512 + IFMA. An entire assembly kernel plus its scalar fold/tail glue ships enabled and untested. All sub-bands are missing: no vector with `sz>=4096 && sz%1024!=0` (stitch + tail), none with `sz>=4096 && aadSz>0` (stitch + AAD fold), none at an exact multiple of 1024.

**Code:**
```
#ifndef CHACHA20_POLY1305_STITCH_MIN
#define CHACHA20_POLY1305_STITCH_MIN 4096
#endif
...
    if (sz >= CHACHA20_POLY1305_STITCH_MIN &&
            chacha20_poly1305_use_fused_ifma()) {

/* largest existing vectors */
wolfcrypt/test/test.c:11938              265 bytes
tests/api/test_chacha20_poly1305.c:350   MC_CHACHA20P1305_MAX_SZ 1024
```

**Recommendation:** Add the large-message KAT/self-consistency test above, round-trip Decrypt, and also assert the bad-tag path at >=4096 since the decrypt stitch produces plaintext before the tag is checked.

wolfcrypt/src/chacha20_poly1305.c
```diff
@@ -41,6 +42,558 @@ or Authenticated Encryption with Additional Data (AEAD) algorithm.
...
+    !defined(WOLFSSL_NO_CHACHA20_POLY1305_SMALL_ASM)
+#define WOLFSSL_CP_SMALL_ASM
+
+/* Fused single-call ChaCha20-Poly1305 encrypt for a one-block (<=64 byte)
+ * record (in chacha_asm.S): SSSE3 crypt2 produces the Poly1305 key block and
+ * the single data block together; the scalar poly1305_*_avx do the MAC. */
+#ifdef __cplusplus
+extern "C" {
+#endif
+WOLFSSL_LOCAL void chacha20_poly1305_small_enc(ChaCha* chacha, Poly1305* poly,
```
dgarske, line 292: ### MEDIUM-12: New WOLFSSL_CP_SMALL_ASM small-message asm kernel is never called with a non-empty AAD
- **File:** `wolfcrypt/src/chacha20_poly1305.c:267-276`
- **Function:** `chacha20_poly1305_small_enc / chacha20_poly1305_small_dec`
- **Action:** SUGGEST
- **Tag:** test
- **Confidence:** High

**Description:** The new fused single-call kernels take an `(aad, aadLen)` pair and fold the AAD into the Poly1305 MAC inside assembly. They are reached only when `sz <= 64` AND AVX2 is present. Enumerating every one-shot / `_ex` call site in the test tree: (a) `wolfcrypt/test/test.c` KATs use 114-byte (`test.c:11878`) and 265-byte (`test.c:11938`) plaintexts -- both > 64; (b) `tests/api/test_chacha20_poly1305.c:39` uses the same 114-byte vector; :831 (InPlace), :888 (UnalignedBuffers), :963 (CrossCipher) all use 67 bytes -- also > 64, by three bytes; (c) `tests/api/test_chacha20_poly1305.c:354` MonteCarlo is the ONLY producer of `sz <= 64` one-shot calls, and it hard-codes NULL/0 AAD at :389 and :391. So the `aadLen>0` code inside the two new asm kernels has never run under any test -- which is exactly the code the two Critical MASM offset defects above corrupt. The 8-/16-byte plaintexts at :652 and :758 do not help (streaming API, which does not use the short path); the XChaCha tests at :161/:287 use 16-byte plaintext WITH AAD but `wc_XChaCha20Poly1305_crypt_oneshot` drives `wc_Chacha_Process` + `wc_Poly1305Update` directly, bypassing the short path. Secondary problem: even the no-AAD coverage of the `sz<=64` band is probabilistic (random `plainLen` in 1..1024 over 100 iterations), so a regression there is flaky rather than deterministic.

**Code:**
```
tests/api/test_chacha20_poly1305.c:386-392
        plainLen = (plainLen % MC_CHACHA20P1305_MAX_SZ) + 1;
        ExpectIntEQ(wc_RNG_GenerateBlock(&rng, plain, plainLen), 0);
        ExpectIntEQ(wc_ChaCha20Poly1305_Encrypt(key, nonce, NULL, 0,
            plain, plainLen, cipher, tag), 0);       /* <-- AAD always NULL/0 */
```

**Recommendation:** Add the deterministic small-with-AAD KATs; extending MonteCarlo to a random AAD closes the AAD gap across every size band at once. On a Windows/ml64 build these tests would immediately expose the two Critical stack-offset defects.

wolfcrypt/src/chacha20_poly1305.c
```diff
@@ -247,8 +866,33 @@ int wc_ChaCha20Poly1305_UpdateData(ChaChaPoly_Aead* aead,
     if (dataLen > CHACHA20_POLY1305_MAX - aead->dataLen)
         return CHACHA_POLY_OVERFLOW;
 
+#ifdef WOLFSSL_CHACHA20_POLY1305_FUSED_IFMA
```
dgarske, line 917: ### MEDIUM-13: Streaming UpdateData is never called with >= 4096 bytes, so the in-UpdateData stitch switch and buffered-AAD re-hash are unreachable in tests
- **File:** `wolfcrypt/src/chacha20_poly1305.c:869-891`
- **Function:** `wc_ChaCha20Poly1305_UpdateData`
- **Action:** SUGGEST
- **Tag:** test
- **Confidence:** High

**Description:** The streaming API is otherwise well covered -- `wolfcrypt/test/test.c:12249` runs Init + UpdateAad + chunked UpdateData + Final + CheckTag for both directions over the 114- and 265-byte vectors, and `tests/api/test_chacha20_poly1305.c:409` does a 2-chunk AAD + 3-chunk data pass. But every UpdateData call in the tree is tiny: 32 or 64 bytes (`test.c:12242-12244`, used at `test.c:12274`), 38 bytes (`tests/api/test_chacha20_poly1305.c:478/480/482`), 8 bytes (:665), 16 bytes (:774). The PR adds two blocks to UpdateData that both require `dataLen >= CHACHA20_POLY1305_STITCH_MIN` (4096): (1) the stitch-mode entry that flips `poly.forceScalar`/`finished` and, critically, COPIES AND RE-HASHES the vector-buffered AAD through a fresh scalar `wc_Poly1305Update` using a stack `byte aadBuf[8 * POLY1305_BLOCK_SIZE]`; and (2) the per-chunk stitch that additionally requires `(aead->dataLen & 63) == 0`. Neither can fire under any existing test. The AAD-then-large-data case the PR comment explicitly calls out ("re-hash the buffered AAD cleanly through the scalar path") is exactly the untested one, and it is the case most likely to produce a wrong tag if the leftover/pad accounting is off. The mid-stream 64-byte-alignment guard is unexercised in both its true and false directions.

**Code:**
```
    if (!aead->poly.forceScalar && aead->poly.started == 0 &&
            aead->dataLen == 0 && dataLen >= CHACHA20_POLY1305_STITCH_MIN &&
            chacha20_poly1305_use_fused_ifma()) {
        word32 aadN = (word32)aead->poly.leftover;
        byte   aadBuf[8 * POLY1305_BLOCK_SIZE];
        if (aadN > 0)
            XMEMCPY(aadBuf, aead->poly.buffer, aadN);

/* largest UpdateData in any test: */
wolfcrypt/test/test.c:12244    #define TEST_SMALL_CHACHA_CHUNKS 64
```

**Recommendation:** Add the three streaming cases above. Case (b) is the highest value -- it is the only exercise of the buffered-AAD re-hash and its 128-byte stack copy.

dgarske (changes requested):

wolfcrypt/src/chacha_asm.asm
SparkiDev: fixed

SparkiDev: fixed

wolfcrypt/src/chacha_asm.asm
SparkiDev: fixed

wolfcrypt/src/chacha_asm.asm
SparkiDev: fixed

wolfssl/wolfcrypt/poly1305.h
SparkiDev: fixed

wolfcrypt/src/chacha20_poly1305.c
SparkiDev: fixed

wolfcrypt/src/chacha20_poly1305.c
SparkiDev: fixed

wolfcrypt/src/chacha_asm.asm
SparkiDev: fixed

wolfcrypt/src/poly1305_asm.S
SparkiDev: fixed

wolfcrypt/src/chacha20_poly1305.c
SparkiDev: fixed

wolfcrypt/src/chacha20_poly1305.c
SparkiDev: fixed

wolfcrypt/src/chacha20_poly1305.c
SparkiDev: fixed

wolfssl/wolfcrypt/chacha20_poly1305.h
SparkiDev: Added tests.

dgarske (approved):
## PR #10940 test results -- ChaCha20/Poly1305 AVX512

Tested `3717560a01` (current PR head) against its parent `067c10d71e` on three x86 machines spanning every SIMD tier. Both sides built from detached worktrees with `--enable-intelasm --enable-mlkem --enable-chacha --enable-poly1305`.

### Correctness (build + full `make check`)

| Machine | CPU / OS | SIMD tier | Before | After |
|---|---|---|---|---|
| tigerlake-come | i7-1185G7E (Tiger Lake), Linux | AVX-512 (f/bw/dq/vl/vbmi/ifma/vnni/gfni/vaes) | PASS | PASS |
| i9w680 | i9-14900K (Raptor Lake), Linux | AVX2 only (AVX-512 fused off) | PASS | PASS |
| MBP-i7 | i7-3720QM (Ivy Bridge, 2012), macOS 10.15, Apple clang 12 | AVX1 only | PASS | PASS |

Every run: `TOTAL 10 / PASS 5 / SKIP 5 / FAIL 0 / ERROR 0`.

The new assembly is genuinely built in, not silently skipped: the `after` libraries export `chacha20_poly1305_avx512`, `chacha_encrypt_avx512`, `chacha_encrypt_avx512vl` and the `poly1305_avx512_blocks_*` set (60 ChaCha/Poly AVX-512 symbols on Linux, 11 on macOS). Apple clang 12 assembles the AVX512/VBMI mnemonics on a CPU that predates AVX2, and the runtime CPUID dispatch falls back cleanly on both non-AVX-512 machines.

### Performance -- tigerlake-come (native AVX-512)

Pinned to one core, 3-pass average, base-10 MB/s. Pass-to-pass spread under 1.5%.

| Algorithm | Before | After | Change | Cycles/byte |
|---|---|---|---|---|
| ChaCha20 | 3670 | 6787 | **+84.9%** (1.85x) | 0.76 -> 0.41 |
| Poly1305 | 7736 | 12246 | **+58.3%** (1.58x) | 0.36 -> 0.23 |
| ChaCha20-Poly1305 AEAD (enc) | 2499 | 4919 | **+96.8%** (1.97x) | 1.12 -> 0.57 |

The PR also adds finer-grained AEAD benchmarks the base build did not have. After-side figures (MB/s): enc 4919, dec 4401, ex-enc 4903, ex-dec 4892, stream 4905. The AEAD row above compares against the base build's single combined `CHA-POLY` number, which is the one-shot encrypt path.

### Summary

No regressions anywhere, and the AVX-512 gains are large and reproducible: ChaCha20 1.85x, Poly1305 1.58x, AEAD 1.97x. These are within about 1% of what the same PR measured on 2026-07-31 (1.83x / 1.57x / 1.96x), so the fixes pushed since then cost nothing on the fast path.

The AVX2 and AVX1 machines were also benchmarked before/after as a regression check. Every delta there landed inside run-to-run noise (under 0.5%, both directions), so those tiers are unchanged in performance and are reported above for correctness only.

PR #10941, opened 2026-07-17
Compliance to new RFCs

Frauschi:
RFC compliance updates for TLS 1.3, the sniffer key-log reader, and the external-PSK certificate extension. Six focused, independent changes, grouped by the RFC they address. No public API is changed and every change stays behind its existing compile-time guards.

### RFC 9846 (RFC 8446 successor, TLS 1.3 clarifications & errata)

- **`general_error(117)` alert** — define the new generic alert from §6.2 in the `AlertDescription` enum and `AlertTypeToString`. In TLS 1.3 it is processed as a fatal alert by the existing alert handling, matching the spec.
- **Cap sender key updates at 2^48-1** — §4.7.3 requires a *sending* implementation to never let its key-update count exceed 2^48-1, while a *receiver* MUST NOT enforce this on the peer. Stream TLS 1.3 previously tracked no count (only the DTLS 1.3 epoch was gated). A `keyUpdateCount` field is added at the **end** of the `Keys` TLS 1.3 block; `SendTls13KeyUpdate` refuses once the ceiling is reached (`BAD_STATE_E`) and increments after deriving each new sending key. The receive path is untouched.
- **Accept empty `CertificateRequest` extensions** — §4.4.2 sets the lower bound on `CertificateRequest.extensions` to 0, so a zero-length block is valid. The client no longer rejects it early with `INVALID_PARAMETER`; the empty block is instead caught by the existing mandatory `signature_algorithms` check, which returns the correct `missing_extension` alert.

### RFC 9850 (SSLKEYLOGFILE format)

- **Line-by-line sniffer key-log parsing** — the reader used `fscanf` with three whitespace-delimited tokens, which is not line-aware, so a comment line was parsed as fields spanning the line boundary and desynchronized the rest of the file. It now reads one line at a time with `fgets`, skips empty and `#` comment lines (§1), and silently skips malformed lines that do not yield the three expected fields (§2) so secrets are still recoverable from a corrupted file. Field buffers are cleared each iteration, the fixed-length client-random field is length-validated, and the line buffer is zeroed on every return path.
- **Honor the `SSLKEYLOGFILE` environment variable** — RFC 9850 notes the informal `SSLKEYLOGFILE` env var used by curl, NSS, Wireshark and various other applications to share one key-log path. Both `tls13ShowSecrets` and `tlsShowSecrets` can now prefer `XGETENV("SSLKEYLOGFILE")` (when set and non-empty), falling back to the compile-time `WOLFSSL_SSLKEYLOGFILE_OUTPUT`. This is **opt-in** behind a new `WOLFSSL_SSLKEYLOGFILE_USE_ENV` macro (default off), so a build that exports the variable for other applications is unaffected and no `XGETENV` reference is emitted unless requested. It remains under the existing `SHOW_SECRETS` / `WOLFSSL_SSLKEYLOGFILE` guards, so production builds cannot log secrets regardless of the environment, per the RFC security considerations.

### RFC 9973 (certificates with an external PSK)

- **Documentation-only reference update** — RFC 9973 is now published and obsoletes RFC 8773; the `cert_with_extern_psk` implementation was written against the 8773bis draft that became RFC 9973 and is already compliant. Only textual references (comments, configure help, extension-codepoint comment, a test comment, Doxygen) are updated. The `WOLFSSL_CERT_WITH_EXTERN_PSK` macro, the `--enable-cert-with-extern-psk` option, the public API, and extension codepoint 33 (0x0021) are all unchanged.

## Testing

- **New unit test** `test_tls13_KeyUpdate_sender_limit` (group `tls13`) exercises both sides of the 2^48-1 boundary: seeded at 2^48-2, `wolfSSL_update_keys` succeeds and advances the count to exactly 2^48-1; at the ceiling the next call is refused with `BAD_STATE_E` and the count is unchanged. This pins the increment side-effect and the `>=` vs `>` boundary.
- Built with `./configure --enable-tls13 --enable-debug && make` and ran `./tests/unit.test --group tls13` — the new test runs (not skipped) and passes; group result `0 failed / 50 skipped / 30 passed`, exit 0.
- The sniffer key-log change adds a comment line, a blank line, and a malformed (`INCOMPLETE_KEYLOG_LINE`) line to the TLS 1.3 key-log test data so the sniffer keylog test exercises the new skip paths.
- Remaining changes are guarded and behavior-preserving for existing configs; full `make check` recommended in CI across the standard build matrix.


Frauschi:
Jenkins retest this please.

SparkiDev (approved):

Issue #16, opened 2015-02-25
OpenSSL header still pulls in cyassl SSL header


aburks:
The OpenSSL header, `wolfssl/openssl/ssl.h` still pulls in the cyassl SSL header. Here's the exact line: https://github.com/wolfSSL/wolfssl/blob/master/wolfssl/openssl/ssl.h#L32

toddouska:
Thanks!

Issue #12, opened 2015-02-24
Remove use of CyaSSL_Mutex in wolfcrypt


aburks:
There are a few vestigial uses of CyaSSL_Mutex in wolfcrypt:

AES: https://github.com/wolfSSL/wolfssl/blob/master/wolfcrypt/src/aes.c#L1438
ECC: https://github.com/wolfSSL/wolfssl/blob/master/wolfcrypt/src/ecc.c#L2653

toddouska:
Thanks!  Good finds.

PR #10949, opened 2026-07-20
AES asm: Add GCM 8-bit tabl, fixes

SparkiDev:
# Description

Added assembly to do 8-bit-table GCM_gmult_len.
Wired it into aes.c and wired small to use C code.
Fixed guards around assembly.

# Testing

Regression tested AES assembly for ARM64, ARM32 and Thumb2.


SparkiDev:
Code generated by:
https://github.com/wolfSSL/scripts/pull/631

wolfcrypt/src/aes.c
```diff
@@ -8138,8 +8138,12 @@ static WC_INLINE void IncrementGcmCounter(byte* inOutCtr)
 #endif
 #endif /* !FREESCALE_LTC_AES_GCM */
 
+/* 32-bit ARMv8 crypto assembly has no standalone GHASH, so it uses this
+ * portable table setup too for AES-GCM streaming (as Aarch64 and the
+ * no-crypto ARM builds already do). */
 #if !defined(WOLFSSL_ARMASM) || defined(__aarch64__) || \
```
SparkiDev: fixed

SparkiDev:
Jenkins: retest this please

FIPS FAILED

wolfcrypt/src/aes.c
```diff
@@ -11067,9 +11101,12 @@ int wc_AesGcmEncrypt(Aes* aes, byte* out, const byte* in, word32 sz,
 #if defined(WOLFSSL_ARMASM)
 #ifndef WOLFSSL_ARMASM_NO_HW_CRYPTO
 #if !defined(__aarch64__)
+    /* Reflect H back to the form the PMULL assembly wants for this call. */
```
dgarske: **🟠 [Medium] wc_AesGcmEncrypt/Decrypt mutate key-derived aes-gcm.H in place; comment claims a copy is used**

The happy path is correct — `AES_GCM_set_key_AARCH32` really does bit-reflect each byte (verified at `armv8-32-aes-asm.S:3958-3967`: the 0x55 / 0x33 / nibble `vshl`+`vsri`+`vbif` sequence), and per-byte reflection is its own inverse, so the paired `GcmReflectH` calls round-trip. There is no early return or error path between the two calls in either function, and `AES_GCM_{set_key,encrypt,decrypt}_AARCH32` have no other callers in the tree.

The concern is that `aes->gcm.H` — key-derived state that was previously read-only during encrypt/decrypt — is now write-shared. If one `Aes` object is used concurrently from two threads, or a call is abandoned (longjmp, cancellation, fatal signal handler), `H` is left permanently in the wrong domain and every subsequent operation on that key silently produces wrong tags with no error return. Scratch fields `aes->tmp`/`aes->reg` were already mutated so concurrency was already unsupported; the difference is that scratch corruption is self-healing per call whereas `H` corruption is persistent and silent.

The fix is cheap because the assembly only reads `gcm_h` — a single `vld1.32 {q8}, [r8]` at `armv8-32-aes-asm.S:3992`, no store. Additionally, the comment at `aes.c:8424-8425` says "the bulk assembly reflects **a copy** back per call", which does not match the code — `GcmReflectH(aes->gcm.H)` reflects in place.

**Fix:** Reflect into a 16-byte stack local, pass that to the assembly, and `ForceZero` it on the way out. This removes the shared-state hazard, halves the reflection work, and makes the code match its own comment. Also fix the `aes.c:8424-8425` comment, which currently describes behaviour the code does not implement.

wolfcrypt/src/aes.c
```diff
@@ -8328,6 +8325,28 @@ void GCM_generate_m0_avx2(const unsigned char *h, unsigned char *m)
 #endif
 #endif /* WOLFSSL_AESNI && GCM_TABLE_4BIT && WC_C_DYNAMIC_FALLBACK */
 
+#if defined(WOLFSSL_ARMASM) && !defined(__aarch64__) && \
+    !defined(WOLFSSL_ARMASM_NO_HW_CRYPTO) && defined(HAVE_AESGCM)
+/* AES_GCM_set_key_AARCH32 produces H with the bits of each byte reflected -
+ * the form the PMULL bulk assembly wants.  The portable GHASH used for AES-GCM
+ * streaming needs plain H, so H is kept un-reflected and reflected back only
+ * around the bulk assembly calls.  Reflecting per byte is its own inverse. */
+static WC_INLINE void GcmReflectH(byte* h)
```
dgarske, line 8365: **🟠 [Medium] Per-call GcmReflectH is avoidable in the default table configurations**

`GcmReflectH` adds 2 x 128 bit-loop iterations (~500-1000 cycles) to every `wc_AesGcmEncrypt`/`wc_AesGcmDecrypt` call. That is negligible for bulk records but a multiple-x regression for small AEAD messages — DTLS, CoAP, handshake records — in exactly the configuration the PMULL assembly is meant to accelerate.

It is also unnecessary in the shipping default. `wolfssl/wolfcrypt/aes.h:39-41` forces `GCM_TABLE_4BIT` for any `WOLFSSL_ARMASM` build that has not chosen a layout, and for both table layouts the software paths read only `gcm.M0` (`aes.c:9588-9646`, `9679`) — `gcm.H` is never read after `GenerateM0`. Only `GCM_SMALL` (`aes.c:8751`, `8815`) and `GCM_WORD32` need plain `H`.

**Fix:** For `GCM_TABLE` / `GCM_TABLE_4BIT`, keep `H` in the reflected form the assembly produces and build `M0` from a temporary un-reflected copy at SetKey time (e.g. add a `GenerateM0_ex(Gcm*, const byte* h)`), which eliminates the hot-path reflection entirely. Keep reflect-on-use only for `GCM_SMALL`/`GCM_WORD32`. If a per-call reflect must stay, use the 3-step SWAR form above or a 256-byte reverse-bits LUT instead of the 128-iteration bit loop.

wolfcrypt/src/port/arm/armv8-aes-asm_c.c
```diff
@@ -56490,6 +56491,367 @@ void GCM_gmult_len(unsigned char* x, const unsigned char** m,
...
+        "ldp	x8, x9, [x11]\n\t"
+        "rev	x8, x8\n\t"
+        "rev	x9, x9\n\t"
+        "eor	x4, x4, x8\n\t"
+        "eor	x5, x5, x9\n\t"
+        "rev	x4, x4\n\t"
+        "rev	x5, x5\n\t"
+        "stp	x4, x5, [%x[x]]\n\t"
+        "subs	%x[len], %x[len], #16\n\t"
+        "add	%x[data], %x[data], #16\n\t"
```
dgarske, line 56849: **🔵 [Low] New AArch64 inline asm modifies the input-only operand 'data'**

The new inline asm advances the block pointer with `add %x[data], %x[data], #16` while `data` is declared input-only (`[data] "r" (data)`) and is not in the clobber list. Under GCC/Clang extended-asm rules an operand written by the asm must be an output (`"+r"`) or explicitly clobbered; writing an input-only operand is undefined behaviour, and the compiler may assume the register still holds the original value or tie `data` to another equal-valued operand.

This is **not introduced** by the PR: the pre-existing 4-bit `GCM_gmult_len` (line 56486) and `AES_GCMSIV_polyval_base` (line 59667) use the identical body line and identical constraint list; grepping for `"+r" (data)` in the file returns zero hits. The new code faithfully reproduces the generator's existing output, and real exposure is limited because `GCM_gmult_len` is a non-static out-of-line function, so absent LTO the compiler has no post-asm use of `data` to mis-optimise. Flagged only so the generator gets fixed once for all three sites. (The ARM32/Thumb-2 `_c.c` forms declare all operands `"+r"` under `!WOLFSSL_NO_VAR_ASSIGN_REG` and match their pre-existing siblings exactly, so they are not affected.)

**Fix:** Fix in the assembly generator so every emitted AArch64 inline-asm block that advances a pointer declares it read-write, and apply the same change to the pre-existing `GCM_gmult_len` (line 56486) and `AES_GCMSIV_polyval_base` (line 59667) so the file is internally consistent.

wolfcrypt/src/port/arm/armv8-aes-asm_c.c
```diff
@@ -56490,6 +56491,367 @@ void GCM_gmult_len(unsigned char* x, const unsigned char** m,
     );
 }
 
+#endif /* !defined(GCM_SMALL) && !defined(GCM_TABLE) */
+#ifdef GCM_TABLE
+XALIGNED(4) static const word8 L_GCM_gmult_len_r[] = {
```
dgarske, line 56500: **⚪ [Info] L_GCM_gmult_len_r alignment diverges between the three AArch64 output forms**

The same 512-byte reduction table is emitted with three different alignments across otherwise byte-for-byte equivalent forms: `armv8-aes-asm.S:56569` uses `.align 3` (8 bytes), `armv8-aes-asm.asm:55701` uses `ALIGN 8`, but `armv8-aes-asm_c.c:56496` uses `XALIGNED(4)`. The pre-existing 4-bit table is 8-byte aligned in all three. There is no functional impact — the new asm accesses the table only with `ldrb` at `%[r] + x13*2` — but the `.S` comment directly above the directive still reads `# 8-byte aligned, 64-bit aligned`, which now contradicts the C form. (`word8` itself is safe: unconditionally typedef'd at `wolfssl/wolfcrypt/types.h:99` in the same block as `word32`.)

**Fix:** Emit `XALIGNED(8)` in the `_c.c` form so the three generated variants agree with each other and with the pre-existing 4-bit table.

dgarske (commented):
## Skoll Code Review

**Findings:** 11 total — 9 posted, 2 skipped
**5 finding(s) not tied to a diff line** (full detail below)

### Posted findings
- [Medium] **wc_AesGcmEncrypt/Decrypt mutate key-derived aes-gcm.H in place; comment claims a copy is used** — `wolfcrypt/src/aes.c:11105,11109,11914,11918 (comment at 8330-8333 and 8424-8425)`
- [Medium] **Per-call GcmReflectH is avoidable in the default table configurations** — `wolfcrypt/src/aes.c:8334-8348`
- [Low] **New AArch64 inline asm modifies the input-only operand 'data'** — `wolfcrypt/src/port/arm/armv8-aes-asm_c.c:56845`
- [Info] **L_GCM_gmult_len_r alignment diverges between the three AArch64 output forms** — `wolfcrypt/src/port/arm/armv8-aes-asm_c.c:56496`

#### New 8-bit-table GCM_gmult_len uses LDM/STM on caller-supplied pointers, adding a 4-byte alignment requirement the 4-bit…
**File:** `wolfcrypt/src/port/arm/thumb2-aes-asm.S:6509-6515 (also thumb2-aes-asm_c.c:7165-7166, armv8-32-aes-asm.S:24340-24346, armv8-32-aes-asm_c.c:25280-25286)`
**Function:** `GCM_gmult_len (GCM_TABLE / 8-bit variant)`
**Severity:** High

The newly added `GCM_TABLE` variant of `GCM_gmult_len` loads the caller-supplied `data` block and the running tag `x` with multi-register transfers: `LDM r0, {r4,r5,r6,r7}` / `LDM r2, {r8,r9,r10,r11}` / `STM r0, {...}`. On ARM32 and Thumb-2, `LDM`/`STM` **always** fault on a non-word-aligned address — unlike `LDR`/`STR`, they are never covered by the unaligned-access support of ARMv6+/ARMv7-M. The result is an alignment Data Abort on ARMv7-A/R and a `UsageFault (UNALIGNED)` on ARMv7-M, regardless of `SCTLR.A` / `CCR.UNALIGN_TRP`.

This is a **divergence from the pre-existing 4-bit function in the very same files**, which reads `data` only with single-word loads at fixed offsets (`armv8-32-aes-asm.S:23572` `ldr r3, [r2, #12]`; `thumb2-aes-asm.S:5895` `LDR r3, [r2, #12]`) and reserves `LDM` for `gcm->M0`, which is `ALIGN16`. The same convention holds for `AES_GCM_encrypt` in these files, which reads/writes the user `in`/`out` buffers with four discrete `LDR`/`STR`. The AArch64 sibling added in this same commit is unaffected because `ldp x8, x9, [x2]` does tolerate unaligned Normal memory.

`data` is a raw caller pointer with no alignment guarantee. Confirmed call sites in `wolfcrypt/src/aes.c`: line 9092 (`a` = user AAD in `GHASH`), 9121 (`c` = ciphertext), 10862 (`authIn`), 10894 (`out`), 11710 (`in`). `x` is likewise a plain `byte x[WC_AES_BLOCK_SIZE]` local at `aes.c:10843` with no alignment attribute.

Reachability confirmed end to end for Thumb-2: `configure.ac:4117-4118` sets `-DWOLFSSL_ARMASM_THUMB2 -DWOLFSSL_ARMASM_NO_HW_CRYPTO`; `thumb2-aes-asm.S` has no…

**Recommendation:** Change the generator so the `data` block is read with four discrete word loads instead of `LDM r2, {r8-r11}`, matching both the pre-existing 4-bit `GCM_gmult_len` and `AES_GCM_encrypt` in the same files. Keep `LDM`/`STM` for `M0` (ALIGN16). Note that a KAT-only regression test will NOT catch this — wolfcrypt test vectors are aligned static arrays, while a real TLS 1.2 record puts the ciphertext at record header (5) + explicit IV (8) = offset 13, reliably unaligned. If `LDM` is kept for speed, the alignment precondition must be documented and `aes.c` must bounce unaligned `data` through an aligned scratch block.

_Referenced code: `wolfcrypt/src/port/arm/thumb2-aes-asm.S:6509-6515 (also thumb2-aes-asm_c.c:7165-7166, armv8-32-aes-asm.S:24340-24346, armv8-32-aes-asm_c.c:25280-25286)` (13 lines)_

---

#### ARM32 (armv8-32) half of the new 8-bit-table assembly is unreachable in every configuration
**File:** `wolfcrypt/src/port/arm/armv8-32-aes-asm.S:24252-25157 (and armv8-32-aes-asm_c.c:25184-26110)`
**Function:** `GCM_gmult_len (GCM_TABLE / 8-bit variant, armv8-32)`
**Severity:** High

The ~1,836 lines of new ARM32 code (907 in the `.S`, 929 in the `_c.c`) plus two copies of a 512-byte table can never be called, because the condition under which they are **compiled** and the condition under which `aes.c` **calls** them are mutually exclusive.

Compiled when: `armv8-32-aes-asm.S` wraps its entire AES body in `#ifndef WOLFSSL_ARMASM_NO_HW_CRYPTO` (opened line 34, closed line 30450; `_c.c` line 53 to 31485). So the new `GCM_gmult_len` exists only for `WOLFSSL_ARMASM && !__aarch64__ && !WOLFSSL_ARMASM_THUMB2 && !WOLFSSL_ARMASM_NO_HW_CRYPTO && HAVE_AESGCM && GCM_TABLE`.

Called when: `aes.c:8892-8899` routes `GCM_GMULT_LEN` to `GCM_gmult_len` only inside `#if defined(WOLFSSL_ARMASM) && (defined(__aarch64__) || defined(WOLFSSL_ARMASM_NO_HW_CRYPTO))`. For 32-bit non-aarch64 that requires `WOLFSSL_ARMASM_NO_HW_CRYPTO` to be **defined**.

So: (a) ARM32 + HW crypto + `GCM_TABLE` emits the asm and the table — and, being `.globl`, they are not link-GC'd — while `aes.c` falls through to the portable `#else` branch at `aes.c:8904` (`static R[256][2]` + portable C `GMULT`) and never references them; (b) ARM32 + `WOLFSSL_ARMASM_NO_HW_CRYPTO` preprocesses the whole file away, so the symbol does not exist at all.

Contrast with the AArch64 sibling added in the same commit: in `armv8-aes-asm.S` the `#ifdef GCM_TABLE` block sits **after** `#endif /* !WOLFSSL_ARMASM_NO_HW_CRYPTO */` and after `#endif /* !WOLFSSL_ARMASM_NO_NEON */`, so it is always compiled and reachable for `aarch64 + WOLFSSL_ARMASM_NO_NEON`. Thumb-2 has no such guard and is also reachable. The commit…

**Recommendation:** Decide the intent and make the guards match. Either (a) move the `#ifdef GCM_TABLE` `GCM_gmult_len` block (and the pre-existing `#if !defined(GCM_SMALL) && !defined(GCM_TABLE)` block, which has the same problem) outside the `#ifndef WOLFSSL_ARMASM_NO_HW_CRYPTO` region in both armv8-32 files, mirroring `armv8-aes-asm.S`; or (b) drop the ARM32 half so ~4 KB of code and a 512-byte table are not emitted into every ARM32 armasm + GCM_TABLE build. Either way, add a build-matrix job for ARM32 + `--enable-aesgcm=table` so the wiring is actually exercised.

_Referenced code: `wolfcrypt/src/port/arm/armv8-32-aes-asm.S:24252-25157 (and armv8-32-aes-asm_c.c:25184-26110)` (12 lines)_

---

#### New WOLFSSL_ARMASM_NO_NEON guard removes Transform_Sha512_Len_crypto while sha512.c still references it
**File:** `wolfcrypt/src/port/arm/armv8-sha512-asm.S:33 (closes at :1649); armv8-sha512-asm_c.c:36 (closes :1568); armv8-sha512-asm.asm:28 (closes :1571)`
**Function:** `Transform_Sha512_Len_crypto`
**Severity:** Medium

The new `#ifndef WOLFSSL_ARMASM_NO_NEON` wraps the entire `WOLFSSL_SHA512 || WOLFSSL_SHA384` body, so it also excludes `Transform_Sha512_Len_crypto` (`.S:1050-1648`, `_c.c:998-1567`, `.asm:1001-1570`). That routine is self-contained — it has its own K table `L_SHA512_trans_crypto_len_k` and shares nothing with the NEON path.

The consumer was not updated to match. `wolfcrypt/src/sha512.c:1478` defines `Transform_Sha512_crypto_aarch64` / `Transform_Sha512_Len_crypto_aarch64` under a bare `#ifdef WOLFSSL_ARMASM_CRYPTO_SHA512`, and `Sha512_SetTransform()` installs them at `sha512.c:1522-1528` under the same bare guard — while the adjacent NEON adapters (`sha512.c:1491-1503`) and NEON dispatch arm (`sha512.c:1529-1535`) **are** correctly qualified with `#ifndef WOLFSSL_ARMASM_NO_NEON`. So a `WOLFSSL_ARMASM && __aarch64__ && WOLFSSL_ARMASM_NO_NEON && WOLFSSL_ARMASM_CRYPTO_SHA512` build still emits calls with no definition.

Two facts keep this at Medium rather than High: (1) the config was already degraded pre-PR — `wolfssl/wolfcrypt/sha512.h:232-239` nests the `Transform_Sha512_Len_crypto` prototype inside `#if !defined(WOLFSSL_ARMASM_NO_NEON)`, so it hit an implicit-declaration diagnostic first; and (2) it is unreachable via autoconf, since `configure.ac:4078` unconditionally sets `ENABLED_ARMASM_NEON=yes` on aarch64, so `-DWOLFSSL_ARMASM_NO_NEON` is only emitted for 32-bit ARM. It IS reachable via a hand-written `user_settings.h`, which is the normal path for wolfSSL's embedded users. The PR turns a compile-time warning into a hard link failure for that config.

**Recommendation:** Since this commit is explicitly a guard-consistency fix, finish it. Either (A) treat SHA-512 crypto as NEON-dependent — defensible, since FEAT_SHA512 implies FEAT_AdvSIMD and `sha512.h` already assumes it — and qualify `sha512.c:1478` and `:1522` with `&& !defined(WOLFSSL_ARMASM_NO_NEON)`; or (B) close the new asm guard just before the `WOLFSSL_ARMASM_CRYPTO_SHA512` section in all three asm files and lift the `sha512.h:235-238` prototype out of the `!NO_NEON` block. The parallel ChaCha guard move in this PR was verified correct and needs no change.

_Referenced code: `wolfcrypt/src/port/arm/armv8-sha512-asm.S:33 (closes at :1649); armv8-sha512-asm_c.c:36 (closes :1568); armv8-sha512-asm.asm:28 (closes :1571)` (8 lines)_

---

#### No test or CI coverage added; the workflow workaround this PR fixes is left in place with a now-stale comment
**File:** `.github/workflows/aesgcm-siv.yml:128-146`
**Function:** `n/a (build matrix)`
**Severity:** Medium

The commit touches 15 files and adds 4,089 lines with no test, no CI job, and no changelog entry. Two specific gaps:

1. The primary aes.c change — removing the `#if !defined(WOLFSSL_ARMASM) || defined(__aarch64__) || defined(WOLFSSL_ARMASM_NO_HW_CRYPTO)` guards — exists precisely to unblock `WOLFSSL_AESGCM_STREAM` on 32-bit ARM. `.github/workflows/aesgcm-siv.yml:130` documents that exact failure: "--disable-aesgcm-stream: WOLFSSL_AESGCM_STREAM's software GHASH only defines its macros for __aarch64__ armasm, not 32-bit __arm__ armasm, so all-crypto + armasm otherwise fails to build aes.c (pre-existing, unrelated to SIV)." Both the `armhf-crypto` (line 133) and `armhf-base` (line 142) jobs still pass `--disable-aesgcm-stream`, so the newly-enabled code path is not compiled, let alone executed, anywhere in CI. This is also the single best regression test available for the `GcmReflectH` round-trip.

2. No job anywhere in `.github/workflows/` builds with `--enable-aesgcm=table` (`-DGCM_TABLE`), which is the only configuration that exercises any of the ~3,100 lines of new 8-bit-table assembly across the four ARM files.

**Recommendation:** Drop `--disable-aesgcm-stream` and the stale comment from the `armhf-crypto` and `armhf-base` jobs — if this PR is correct they should now build and pass, which is the cleanest possible proof of the fix. Add at least one `--enable-aesgcm=table` job for aarch64 (`+ -DWOLFSSL_ARMASM_NO_NEON`) and one for Thumb-2 to exercise the new 8-bit assembly. Ensure at least one case runs GHASH over an intentionally misaligned AAD/ciphertext buffer, since aligned KAT vectors will not surface the `LDM` alignment defect reported separately.

_Referenced code: `.github/workflows/aesgcm-siv.yml:128-132` (5 lines)_

---

#### Redundant push/pop of r3 on every loop iteration in the ARM32 8-bit variant
**File:** `wolfcrypt/src/port/arm/armv8-32-aes-asm.S:24347 and :25151 (also armv8-32-aes-asm_c.c:25287, :26091)`
**Function:** `GCM_gmult_len (GCM_TABLE / 8-bit variant, armv8-32)`
**Severity:** Low

The new ARM32 8-bit `GCM_gmult_len` saves and restores r3 (the `len` loop counter) around every loop iteration. This is required in the pre-existing 4-bit function, where `%[len]`/r3 is used as general scratch throughout the body in every configuration. In the new 8-bit function r3 / `%[len]` is used as scratch **only** inside the `#if defined(WOLFSSL_ARM_ARCH) && (WOLFSSL_ARM_ARCH < 6)` REV-emulation blocks. On ARMv6 and later (the `rev r4, r4` path) r3 is untouched between the push and the pop, so the pair costs two extra memory accesses per 16-byte block for nothing. The Thumb-2 sibling omits it entirely, confirming the generator can emit the block without it. Stack balance itself is correct on all paths including the `bne` back edge.

**Recommendation:** Have the generator emit the `push {r3}` / `pop {r3}` pair only inside the `WOLFSSL_ARM_ARCH < 6` guard, or use one of the already-clobbered registers (r4-r11) as the REV scratch so no save/restore is needed.

_Referenced code: `wolfcrypt/src/port/arm/armv8-32-aes-asm.S:24347 and :25151 (also armv8-32-aes-asm_c.c:25287, :26091)` (10 lines)_

---

### Skipped findings
- [Info] ~~New aes.h comment is spliced onto the pre-existing one and reads as a single garbled sentence~~
- [Low] ~~Generated asm files edited without the generators, which live outside this repository~~

dgarske (changes requested):
https://github.com/wolfSSL/wolfssl/pull/10949#pullrequestreview-4832643665

wolfcrypt/src/aes.c

wolfcrypt/src/aes.c
SparkiDev: fixed

wolfcrypt/src/port/arm/armv8-aes-asm_c.c
SparkiDev: fixed

wolfcrypt/src/port/arm/armv8-aes-asm_c.c
SparkiDev: fixed

SparkiDev:
rebased

SparkiDev:
Jenkins: retest this please

dgarske (approved):

PR #10951, opened 2026-07-20
build: distribute IDE project trees per-directory to fix "make dist"

Frauschi:
### Summary

`make dist` can fail on Linux with:

```
/bin/bash: Argument list too long
make[2]: *** [distdir-am] Error 127
```

This changes the `IDE/` integration trees to be distributed one directory at a time (`EXTRA_DIST += IDE/<dir>`) instead of enumerating every file, shrinking the top-level `DISTFILES` from ~123 KB to ~97 KB and getting `make dist` well clear of the limit. The set of files placed in the release tarball is unchanged.

### Root cause

wolfSSL uses a single non-recursive `Makefile.am`, so every distributed file lands in one top-level `$(DISTFILES)`. Automake's `distdir-am` recipe inlines that entire list into a single shell command (`list='$(DISTFILES)'; ...`), which `make` passes to `/bin/bash -c` as one argument. Linux caps a single argument at `MAX_ARG_STRLEN` (128 KB), independent of the 2 MB total `ARG_MAX`. With `DISTFILES` sitting at ~123 KB, the recipe was right at that limit, so a modest addition of `EXTRA_DIST` entries pushed it over and `execve` returned `E2BIG`. macOS has no per-argument cap, which is why this only reproduces on Linux.

### Change

The `IDE/` directories were enumerated file-by-file via 53 `include IDE/*/include.am` fragments (~745 `EXTRA_DIST` entries). Automake already copies a directory-valued `EXTRA_DIST` entry recursively, and `IDE/include.am` already used that form for a few trees (`IDE/Espressif`, `IDE/HEXIWEAR`, ...). This change extends it to the rest:

- Replace the per-file enumeration with one `EXTRA_DIST += IDE/<dir>` entry per directory (~52 entries).
- Remove the 53 now-unused `include.am` fragments.
- Preserve the `DISTCLEANFILES` entry for `IDE/iotsafe/build`.

Net effect on the top-level `DISTFILES`:

| | Before | After |
| --- | --- | --- |
| Bytes | 122,948 | 97,366 |
| Entries | 3,363 | 2,824 |
| `IDE/` contribution | ~36 KB / ~910 entries | ~12.5 KB / ~52 entries |

That leaves roughly 34 KB of headroom under the 128 KB limit.

#### Directories kept enumerated

Three directories intentionally omit some tracked files (private IDE configs, `.gitkeep` placeholders, helper scripts), so they stay listed file-by-file to avoid shipping those: `IDE/apple-universal`, `IDE/MPLABX16`, and `IDE/Renesas`.

### Why not a recursive `SUBDIRS`

Splitting `IDE/` into a recursively-distributed subdirectory is the other textbook fix, but wolfSSL deliberately avoids recursive `distdir`: `linuxkm/Makefile` is hand-written (not in `AC_CONFIG_FILES`), and `Makefile.am` documents that the `SUBDIRS_OPT` aux-variable exists specifically to keep `make distdir` from recursing. Extending the wholesale-`EXTRA_DIST` idiom already present in `IDE/include.am` stays on-convention.

### Testing

- Before/after `make dist` tarball comparison: **no `IDE/` project file added or removed**. The only difference is that the 53 now-unused `include.am` build fragments are no longer packaged.
- `make dist` verified on Ubuntu 24.04 / automake 1.16.5 (matching the `Multiple compilers and versions` CI toolchain).
- A dirty in-tree build is out of scope: as with the pre-existing wholesale entries, `make dist` is expected to run from a clean checkout.

### Checklist

- [x] No functional/library code changed (build metadata only).
- [x] Distributed file set unchanged (verified by tarball diff).
- [x] `make dist` passes on the CI toolchain.


douzzer:
retest this please
(aborted "group 6")

douzzer (approved):

PR #10832, opened 2026-07-01
Add crypto callback only mode for curve25519

padelsbach:
# Description

Companion PR to #10830. Applied for `make_key` and `shared_secret` operations. No streaming or async support.

Note: some fe_operations math routines linger due to some curve25519 operations without `CRYPTO_CB` support (regular or "only" mode). About 6kB of code can be trimmed in CB-only mode. I'm considering that as a follow-up change. GE math is fully trimmed however.

# Testing

Added unit tests

# Checklist

 - [x] added tests
 - [ ] updated/added doxygen
 - [ ] updated appropriate READMEs
 - [ ] Updated manual and documentation

PR #10955, opened 2026-07-20
Extend curve25519 crypto cb and cb-only for footprint savings

padelsbach:
# Description

Follow up PR to #10832

Add `make_pub` and `generic` methods for curve25519 to the crypto cb and cb-only modes. Saves about 6kB of code footprint on my aarch64 machine.

# Testing

New/updated unit tests

# Checklist

 - [x] added tests
 - [ ] updated/added doxygen


dgarske (changes requested):
Please resolve merge conflicts. Thanks @padelsbach

padelsbach:
jenkins retest this please

wolfcrypt/src/curve25519.c
```diff
@@ -165,6 +168,16 @@ int wc_curve25519_make_pub(int public_size, byte* pub, int private_size,
     if (ret != 0)
         return ret;
 
+#ifdef WOLF_CRYPTO_CB
+    ret = wc_CryptoCb_Curve25519MakePub(public_size, pub, private_size, priv);
```
dgarske, line 172: Could this behavior only happens with `WOLF_CRYPTO_CB_ONLY_CURVE25519` or when devId == INVALID_DEVID ? This function uses the Find Callback which seems fragile. I'd prefer to use the devId if known and I don't see this code path doing that.

wolfcrypt/src/cryptocb.c
```diff
@@ -1142,6 +1142,67 @@ int wc_CryptoCb_Curve25519(curve25519_key* private_key,
 
     return wc_CryptoCb_TranslateErrorCode(ret);
 }
+
+int wc_CryptoCb_Curve25519MakePub(int public_size, byte* pub,
```
dgarske, line 1148: Make sure and add new PK's to the GetPkTypeStr() function.

```
case WC_PK_TYPE_ED25519_MAKE_PUB: return "ED25519 MakePub";
case WC_PK_TYPE_ED25519_CHECK_KEY: return "ED25519 CheckKey";
```

wolfcrypt/src/curve25519.c
```diff
@@ -339,6 +355,16 @@ int wc_curve25519_make_pub_blind(int public_size, byte* pub, int private_size,
...
```
dgarske, line 359: Same here. If we have a devId why not use that specifically. Also isn't this only needed for `WOLF_CRYPTO_CB_ONLY_CURVE25519`?

wolfcrypt/src/fe_operations.c
```diff
@@ -23,9 +23,7 @@
 
  /* Based On Daniel J Bernstein's curve25519 Public Domain ref10 work. */
 
-/* under WOLF_CRYPTO_CB_ONLY_ED25519 the callback device does all Ed25519
- * field math, so Ed25519 alone no longer pulls this file in */
-#if defined(HAVE_CURVE25519) || \
+#if (defined(HAVE_CURVE25519) && !defined(WOLF_CRYPTO_CB_ONLY_CURVE25519)) || \
```
dgarske, line 26: Don't forget the `CURVE_SMALL` case in `wolfcrypt/src/fe_low_mem.c`

dgarske (changes requested):

wolfcrypt/src/curve25519.c
padelsbach: Background on this approach:

1. The public API (eg `wc_curve25519_make_pub` in this case) receives plain raw bytes, so there is no clean path to add the `devID` field into an existing struct without disrupting the API
2. The `FindCb` mechanism is the unambiguous "right" way for the crypto device to be specified imho -- but if they don't have that set, we just use the first registered device.
3. Ideally don't want to limit to `CRYPTO_CB_ONLY` in hopes of broadening the main cryptocb feature

If you still feel strongly about it, I can come up with another approach.

wolfcrypt/src/curve25519.c
padelsbach: No existing `devId` at this layer with the existing API

wolfcrypt/src/fe_operations.c
padelsbach: done

wolfcrypt/src/cryptocb.c
padelsbach: good catch, done

dgarske (approved):

dgarske:
Jenkins retest this please

PR #10957, opened 2026-07-21
ARM32 assembly: Get CPU Id flags to choose assembly.

SparkiDev:
# Description

Added flag determination for Linux/Android/BSD/Windows/privilege-mode.
Made AES and SHA-256 runtime dispatch.
Updated the ARM32 assembly to support having multiple implementations compiled in.

# Testing

Tested ARM32 builds.

Code generated by PR:
https://github.com/wolfSSL/scripts/pull/633


wolfcrypt/src/aes.c
```diff
@@ -8416,7 +8557,8 @@ int wc_AesGcmSetKey(Aes* aes, const byte* key, word32 len)
             AES_ECB_encrypt_NEON(iv, aes->gcm.H, WC_AES_BLOCK_SIZE,
                 (const unsigned char*)aes->key, aes->rounds);
         }
-#elif defined(__aarch64__) || defined(WOLFSSL_ARMASM_NO_HW_CRYPTO)
+#elif defined(__aarch64__) || defined(WOLFSSL_ARMASM_NO_HW_CRYPTO) || \
+      defined(WOLFSSL_ARM32_AES_DISPATCH)
         {
             AES_ECB_encrypt(iv, aes->gcm.H, WC_AES_BLOCK_SIZE,
```
SparkiDev, line 8630: fixed

wolfcrypt/src/aes.c
```diff
@@ -19450,10 +19672,22 @@ static AesGcmSivPolyvalFn AesGcmSivPolyvalAsm(void)
...
 {
-#ifndef WOLFSSL_ARMASM_NO_HW_CRYPTO
+#ifdef WOLFSSL_ARM32_AES_DISPATCH
+#ifndef WOLFSSL_AESGCM_STREAM
+    cpuid_get_flags_ex(&cpuid_flags);
+    if (IS_ARM32_PMULL(cpuid_flags)) {
+        return &AES_GCMSIV_polyval_crypto;
+    }
+#endif
+    return &AES_GCMSIV_polyval_base;
```
SparkiDev, line 19900: fxied

SparkiDev:
Jenkins: retest this please

SparkiDev:
Jenkins: retest this please

wolfcrypt/src/cpuid.c
```diff
@@ -567,6 +568,256 @@
...
+                new_cpuid_flags |= CPUID_PMULL;
+            if (features & ANDROID_CPU_ARM_FEATURE_SHA2)
+                new_cpuid_flags |= CPUID_SHA256;
+        #endif
+
+            (void)wolfSSL_Atomic_Uint_CompareExchange
+                (&cpuid_flags, &old_cpuid_flags, new_cpuid_flags);
+        }
+    }
+#elif defined(__FreeBSD__) || defined(__OpenBSD__)
```
Frauschi: **🟡 [Medium] OpenBSD grouped with FreeBSD but calls FreeBSD-only elf_aux_info()**
*💡 SUGGEST*

The new AArch32 feature-detection block adds a branch `#elif defined(__FreeBSD__) || defined(__OpenBSD__)` that `#include <sys/auxv.h>` and calls `elf_aux_info(AT_HWCAP, ...)` / `elf_aux_info(AT_HWCAP2, ...)`. `elf_aux_info()` is a FreeBSD/NetBSD interface; OpenBSD does not provide it (nor the same `<sys/auxv.h>`/`elf_aux_info` API for reading HWCAP). On a 32-bit ARM (armv7) OpenBSD build with `WOLFSSL_ARMASM`, this branch is selected and the translation unit fails to compile/link (implicit declaration of `elf_aux_info`, or missing header), breaking `cpuid.c` for that supported platform. FreeBSD is unaffected. Because `__OpenBSD__` is grouped here rather than falling through to the safe `#else` compiled-features fallback, the run-time-dispatch feature turns a previously buildable configuration into a build break.

**Recommendation:** Restrict this branch to platforms that actually provide `elf_aux_info()` (e.g. `#elif defined(__FreeBSD__)`), and give OpenBSD its own branch using its supported mechanism (e.g. `sysctl` CTL_MACHDEP) or let OpenBSD fall through to the `#else` compiled-features fallback.

wolfcrypt/src/cpuid.c
```diff
@@ -567,6 +568,256 @@
...
+        #ifdef __ARM_NEON
+            new_cpuid_flags |= CPUID_ASIMD;
+        #endif
+        #endif
+
+            (void)wolfSSL_Atomic_Uint_CompareExchange
+                (&cpuid_flags, &old_cpuid_flags, new_cpuid_flags);
+        }
+    }
+#elif defined(__linux__)
```
Frauschi, line 650: **🔵 [Low] Android CPUID branch is unreachable (falls into the __linux__ branch first)**
*💡 SUGGEST*

In the new `HAVE_CPUID_ARM32` block the platform selection places `#elif defined(__linux__)` (getauxval, line 639) before `#elif defined(__ANDROID__) || defined(ANDROID)` (cpu-features.h / android_getCpuFeatures, line 689). Android NDK/bionic toolchains define `__linux__` in addition to `__ANDROID__`, so the `__linux__` branch always matches first and the dedicated Android branch is dead code that is never compiled. On modern Android the `getauxval`/`AT_HWCAP2` path works (API 18+/21+), so run-time detection still functions, but the intended NDK `cpu-features` fallback (written for pre-API-18 bionic that lacked `getauxval`) can never be reached. Note this exactly mirrors the pre-existing AArch64 code (cpuid.c:267 vs 329), so it may be an accepted convention rather than an oversight. Severity assessed as Low by both modes; confidence bumped to High on cross-mode agreement.

**Recommendation:** Order the `#elif defined(__ANDROID__) || defined(ANDROID)` branch before the generic `#elif defined(__linux__)` branch (in both the ARM32 and, for consistency, the AArch64 blocks) so the Android-specific detection is actually selectable. If getauxval is preferred everywhere, consider dropping the dead Android branch instead.

wolfcrypt/src/aes.c
```diff
@@ -1085,6 +1085,17 @@ block cipher mechanism that uses n-bit binary string parameter key with 128-bits
     #endif /* HAVE_AES_DECRYPT */
 
 #elif defined(WOLFSSL_ARMASM)
+/* On 32-bit Arm both the base (table) AES and the Armv8 crypto-extension AES
+ * are compiled into one object by default, and the implementation is chosen at
+ * run time through aes->use_aes_hw_crypto - mirroring the AArch64 path.  The
+ * base fallback can be dropped with WOLFSSL_ARMASM_NO_BASE_IMPL (crypto
+ * always present), and WOLFSSL_ARMASM_NO_HW_CRYPTO keeps only the base, both
+ * of which revert to direct calls with no run-time check. */
+#if !defined(__aarch64__) && !defined(WOLFSSL_ARMASM_NO_HW_CRYPTO) && \
```
Frauschi: **🔵 [Low] WOLFSSL_ARM32_AES_DISPATCH does not exclude Thumb-2 (unlike SHA256_ARM32_DISPATCH)**
*🔧 NIT*

The SHA-256 dispatch guard explicitly excludes Thumb-2 (`sha256.c:1350` uses `!defined(WOLFSSL_ARMASM_THUMB2)`), but the analogous AES guard `WOLFSSL_ARM32_AES_DISPATCH` (`#if !defined(__aarch64__) && !defined(WOLFSSL_ARMASM_NO_HW_CRYPTO) && !defined(WOLFSSL_ARMASM_NO_BASE_IMPL)`) does not. The Thumb-2 AES assembly (`thumb2-aes-asm.S`) provides only the base implementation and none of the crypto-extension `AES_*_AARCH32` symbols, so a Thumb-2 build that reaches this dispatch path would fail to link against `AES_encrypt_AARCH32`. This is safe in autotools builds because configure always pairs `WOLFSSL_ARMASM_THUMB2` with `WOLFSSL_ARMASM_NO_HW_CRYPTO` (configure.ac:4139-4140), which suppresses the dispatch macro. It could bite a hand-rolled user_settings.h that sets `WOLFSSL_ARMASM_THUMB2` + crypto without `NO_HW_CRYPTO` (already broken pre-diff, so not a regression — just missing defense-in-depth/consistency with the SHA-256 guard).

**Recommendation:** Add `!defined(WOLFSSL_ARMASM_THUMB2)` to the `WOLFSSL_ARM32_AES_DISPATCH` definition to match the SHA-256 dispatch guard and make the Thumb-2 restriction explicit at the code level rather than relying solely on configure.

Frauschi (commented):
## 🐺 Skoll Code Review

### Posted findings
- [Medium] **OpenBSD grouped with FreeBSD but calls FreeBSD-only elf_aux_info()** — `wolfcrypt/src/cpuid.c:719-767`
- [Low] **Android CPUID branch is unreachable (falls into the __linux__ branch first)** — `wolfcrypt/src/cpuid.c:639-718`
- [Low] **WOLFSSL_ARM32_AES_DISPATCH does not exclude Thumb-2 (unlike SHA256_ARM32_DISPATCH)** — `wolfcrypt/src/aes.c:1094-1097`

wolfcrypt/src/aes.c

wolfcrypt/src/cpuid.c
SparkiDev: fixed

wolfcrypt/src/cpuid.c
SparkiDev: fixed

wolfcrypt/src/aes.c
```diff
@@ -7300,8 +7418,16 @@ int wc_AesCbcEncrypt(Aes* aes, byte* out, const byte* in, word32 sz)
 #if defined(WOLFSSL_ARMASM)
 #ifndef WOLFSSL_ARMASM_NO_HW_CRYPTO
```
Frauschi, line 7404: **🟠 [High] Dangling `else` swallows the CBC-decrypt fallback in 32-bit dispatch builds with WOLFSSL_ARMASM_NEON_NO_TABLE_LOOKUP**
*🚫 BLOCK*  `bug`

The diff adds `WOLFSSL_ARM32_AES_DISPATCH` to the guard on the table-based fallback (line 7451), but that fallback is *also* wrapped in `#ifndef WOLFSSL_ARMASM_NEON_NO_TABLE_LOOKUP` (line 7453). Before the diff, a 32-bit crypto build never reached that guard - the ARM32 arm of the `#if` unconditionally called `AES_CBC_decrypt_AARCH32()` with no `else`. Now, if a 32-bit ARM dispatch build also defines `WOLFSSL_ARMASM_NEON_NO_TABLE_LOOKUP`, the preprocessor emits:

```c
if (aes->use_aes_hw_crypto) { AES_CBC_decrypt_AARCH32(...); }
else
    return 0;
```

The NEON branch at 7438 is `__aarch64__`-gated so it contributes nothing on 32-bit. The result compiles cleanly and, on a CPU without the crypto extension, `wc_AesCbcDecrypt()` returns success with the output buffer completely untouched - silent plaintext corruption rather than an error. The same class of hole is opened in `wc_AesCtrEncrypt` (aes.c:7945-7951, full blocks left unencrypted) and in `AES_GCM_encrypt_ASM`/`AES_GCM_decrypt_ASM` (aes.c:11007-11012, 11844), both of which are now compiled for 32-bit via the new `WOLFSSL_ARM32_AES_DISPATCH` term. Note the single-block paths (`wc_AesEncrypt`/`wc_AesDecrypt` at 1180-1192, 3611-3623, 4450-4462) use `#if defined(__aarch64__) && ... #elif ...` and are correctly unaffected - only these four sites use the standalone `#ifndef` form.

**Also raised by the `portability` scan** (wolfcrypt/src/aes.c:7418-7461):

In a 32-bit Arm dispatch build the new `if (aes->use_aes_hw_crypto) { AES_CBC_decrypt_AARCH32(...); } else` is followed by two preprocessor blocks. The AArch64 NEON block (7439-7449) is compiled out on 32-bit, and the newly-widened base block (7451-7458) is additionally wrapped in `#ifndef WOLFSSL_ARMASM_NEON_NO_TABLE_LOOKUP`. If a user defines `WOLFSSL_ARMASM_NEON_NO_TABLE_LOOKUP` in `user_settings.h` for a 32-bit dispatch build, both blocks vanish and the `else` at line 7426 binds to the `return 0;` at line 7461. The hardware branch then falls off the end of the `#if defined(WOLFSSL_ARMASM)` section without returning a value (undefined behaviour / garbage return code), and the non-hardware branch performs no decryption at all. Before this diff the 32-bit crypto build called `AES_CBC_decrypt_AARCH32` unconditionally and reached `return 0;` normally, so the diff is what introduces the dangling `else`.

**Suggestion:**
```suggestion
/* WOLFSSL_ARMASM_NEON_NO_TABLE_LOOKUP only makes sense where a NEON AES
 * implementation exists, i.e. AArch64.  On 32-bit Arm the table variant is
 * the only fallback, so it must always be compiled in. */
#if defined(WOLFSSL_ARM32_AES_DISPATCH) && \
    defined(WOLFSSL_ARMASM_NEON_NO_TABLE_LOOKUP)
    #error "WOLFSSL_ARMASM_NEON_NO_TABLE_LOOKUP is not supported on 32-bit Arm"
#endif
```

**Recommendation:** Either add an explicit `#error` (or configure-time rejection) for `WOLFSSL_ARMASM_NEON_NO_TABLE_LOOKUP` on 32-bit Arm, or change the four `#ifndef WOLFSSL_ARMASM_NEON_NO_TABLE_LOOKUP` guards at aes.c:7453, 7947, 11009 and 11844 to `#if !defined(WOLFSSL_ARMASM_NEON_NO_TABLE_LOOKUP) || defined(WOLFSSL_ARM32_AES_DISPATCH)` so the table fallback is never dropped when it is the only fallback.

wolfcrypt/src/aes.c
```diff
@@ -1098,6 +1111,46 @@ static void Check_CPU_support_HwCrypto(Aes* aes)
...
+#elif defined(WOLFSSL_ARM32_AES_DISPATCH)
+static cpuid_flags_t cpuid_flags = WC_CPUID_INITIALIZER;
+
+/* Record on the Aes object whether this CPU implements the Armv8 AES and PMULL
+ * crypto-extension instructions, so the per-operation code can select the
+ * crypto or the base assembly at run time.  Called from key setup.
+ *
+ * @param [in, out] aes  AES object whose use_aes_hw_crypto /
+ *                       use_pmull_hw_crypto flags are set. */
+static void Check_CPU_support_HwCrypto(Aes* aes)
```
Frauschi, line 1114: **🟡 [Medium] WOLFSSL_AESGCM_STREAM disables the Armv8 crypto extension for every AES mode on 32-bit Arm, not just GCM**
*💡 SUGGEST*  `bug`

**Sites:** `wolfcrypt/src/aes.c:1123-1155`; `wolfcrypt/src/aes.c:1125-1134`
The streaming carve-out clears `aes->use_aes_hw_crypto` for the whole `Aes` object at key-setup time. Because that one flag is what every 32-bit dispatch site consults (`wc_AesEncrypt`, `wc_AesDecrypt`, `wc_AesCbcEncrypt`/`Decrypt`, `wc_AesCtrEncrypt`, `_AesEcbEncrypt`/`Decrypt`, `wc_AesXtsEncrypt`/`Decrypt`), enabling `--enable-aesgcm-stream` silently drops AES-CBC, AES-CTR, AES-ECB, AES-XTS, AES-CCM, AES-CFB and AES-OFB back to the table-based software AES on 32-bit Arm - none of which have anything to do with GCM streaming. `WOLFSSL_AESGCM_STREAM` is not on by default, but it is force-enabled for `--enable-linuxkm-lkcapi-register` (configure.ac:11546), `--enable-bsdkm-register` (configure.ac:932), `--enable-wolfzfs` (configure.ac:3715) and FIPS dev builds (configure.ac:6946, 7187, 7411). Those configurations get zero AES acceleration on 32-bit Arm, and the table-based AES is also the more cache-timing-exposed implementation. The technical reason (streaming GHASH needs `gcm.M0`, which the crypto key setup does not build) is correct and well documented, but the fix is applied far too broadly.

**Also raised by the `portability` scan** (wolfcrypt/src/aes.c:1123-1153):

The new AArch32 `Check_CPU_support_HwCrypto()` forces `aes->use_aes_hw_crypto = 0` (and `use_pmull_hw_crypto = 0`) for *every* `Aes` object whenever `WOLFSSL_AESGCM_STREAM` is defined. The stated rationale is GCM-specific (streaming GHASH needs `gcm.H` in standard form, while `AES_GCM_set_key_AARCH32` stores it PMULL-reflected), but the flag is consulted by `wc_AesEncrypt`, `wc_AesDecrypt`, `wc_AesCbcEncrypt`, `wc_AesCbcDecrypt`, `wc_AesCtrEncrypt`, `_AesEcbEncrypt`, `_AesEcbDecrypt`, `wc_AesXtsEncrypt`, `wc_AesXtsDecrypt` and AES-CCM as well. `WOLFSSL_AESGCM_STREAM` is auto-enabled by configure.ac for FIPS v6 (`configure.ac:6946`, `7187`, `7411`), wolfZFS (`configure.ac:3715`), and Linux/FreeBSD kernel-module crypto registration (`configure.ac:932`, `11546`). In every such build on a 32-bit Armv8 CPU the library silently drops from the constant-time `aese`/`aesd` crypto-extension AES (which the pre-diff code called unconditionally via `AES_encrypt_AARCH32`) to the table-driven `AES_ECB_encrypt`/`AES_CBC_*`/`AES_CTR_encrypt` base assembly, which performs secret-index Te/Td table lookups. That is both a several-fold throughput regression versus the pre-diff behaviour and a reintroduction of cache-timing-observable AES for CBC/CTR/ECB/XTS/CCM traffic that has nothing to do with GCM streaming. The constraint should gate only the GCM/GHASH path (or the key schedule used for GCM), not the whole `Aes` object.

**Also, at `wolfcrypt/src/aes.c:1125-1134` (Info, review scan) — Pointless CPUID read when streaming forces the base implementation:**

In a `WOLFSSL_AESGCM_STREAM` build the function reads the CPUID flags (potentially a `getauxval()` syscall on the first call) and then unconditionally assigns 0 to both flags without ever consulting the result. Moving the read inside the `#else` arm makes the intent obvious and skips the work.

**Suggestion:**
```suggestion
/* Keep the CPU-derived flags here so CBC/CTR/ECB/XTS stay accelerated, and
 * demote only the GCM object in wc_AesGcmSetKey():
 *
 *   #if defined(WOLFSSL_AESGCM_STREAM) && defined(WOLFSSL_ARM32_AES_DISPATCH)
 *       aes->use_aes_hw_crypto = 0;
 *       aes->use_pmull_hw_crypto = 0;
 *       AES_set_encrypt_key(key, len * 8, (byte*)aes->key);
 *   #endif
 * (re-expanding the schedule with the base variant before H/M0 are derived).
 */
```

**Recommendation:** Move the streaming override out of `Check_CPU_support_HwCrypto()` and into `wc_AesGcmSetKey()` so only GCM objects fall back, or generate the `M0` table alongside the crypto-form `H` so streaming can use the crypto key schedule. At minimum, document the build-wide performance impact in the configure help text.

[wolfcrypt/src/aes.c:1125-1134] Move the `cpuid_get_flags_ex()` call into the non-streaming branch.

wolfcrypt/src/sha256.c
```diff
@@ -1342,6 +1342,48 @@ int wc_InitSha256_ex(wc_Sha256* sha256, void* heap, int devId)
...
+ * and call it directly. */
+#if !defined(WOLFSSL_ARMASM_THUMB2) && !defined(WOLFSSL_ARMASM_NO_NEON) && \
+    !defined(WOLFSSL_ARMASM_NO_HW_CRYPTO) && \
+    !defined(WOLFSSL_ARMASM_NO_NEON_IMPL)
+    #define SHA256_ARM32_DISPATCH
+#endif
+
+#ifdef SHA256_ARM32_DISPATCH
+
+static int sha256_transform_check = 0;
```
Frauschi, line 1368: **🟡 [Medium] IS_ARM32_ASIMD is detected on every AArch32 path but no dispatch site consults it (SHA-256 NEON transform is ungated)**
*💡 SUGGEST*  `bug`

**Sites:** `wolfcrypt/src/sha256.c:1358-1382`; `wolfssl/wolfcrypt/cpuid.h:145-150, wolfcrypt/src/sha256.c:1368-1381`
`cpuid.c` now populates `CPUID_ASIMD` for ARM32 and `cpuid.h` exports `IS_ARM32_ASIMD()`, but the SHA-256 selector never consults it: the pointer is statically initialised to `Transform_Sha256_Len_neon` and the only run-time decision is crypto-vs-NEON. On a build compiled with NEON assembly running on a CPU that does not implement Advanced SIMD, this executes NEON instructions and takes an undefined-instruction fault. The AArch64 sibling (sha256.c:1290-1302) does gate its NEON transform on `IS_AARCH64_ASIMD(cpuid_flags)`, so the omission is an inconsistency as much as a hazard. The inline comment ('valid on any Armv8-32/NEON CPU') asserts the safety property rather than checking it, and the ARM32 Linux detection explicitly reads `HWCAP_NEON` - a value that is then thrown away.

**Also, at `wolfssl/wolfcrypt/cpuid.h:145-150, wolfcrypt/src/sha256.c:1368-1381` (Info, portability scan) — IS_ARM32_ASIMD is detected but never consulted by any dispatch site:**

`CPUID_ASIMD` / `IS_ARM32_ASIMD()` are added and populated by all four AArch32 detection branches, but no caller in the tree uses them. In particular `Sha256_SetTransform()` falls back to `Transform_Sha256_Len_neon` without checking `IS_ARM32_ASIMD(sha256_cpuid_flags)`, and the comment claims that transform "is valid on any Armv8-32/NEON CPU". That is only true because `SHA256_ARM32_DISPATCH` requires `!WOLFSSL_ARMASM_NO_NEON`, i.e. the whole build already assumes NEON via `-mfpu=crypto-neon-fp-armv8`. The AArch64 sibling (`sha256.c:1298`) does gate on `IS_AARCH64_ASIMD`, so the asymmetry is likely to confuse future maintainers, and the detected ASIMD flag currently costs a `getauxval` read for no benefit.

**Suggestion:**
```suggestion
if (IS_ARM32_SHA256(sha256_cpuid_flags))
        Transform_Sha256_Len_p = Transform_Sha256_Len_crypto;
    else if (IS_ARM32_ASIMD(sha256_cpuid_flags))
        Transform_Sha256_Len_p = Transform_Sha256_Len_neon;
    else
        Transform_Sha256_Len_p = Transform_Sha256_Len_base;
```

**Recommendation:** Add an `IS_ARM32_ASIMD()` tier falling back to `Transform_Sha256_Len_base`, matching the AArch64 selector; or, if a NEON-capable CPU is a hard build precondition, say so explicitly in the comment instead of leaving `IS_ARM32_ASIMD()` defined but unused.

[wolfssl/wolfcrypt/cpuid.h:145-150, wolfcrypt/src/sha256.c:1368-1381] Either mirror the AArch64 structure and gate the NEON transform on `IS_ARM32_ASIMD()` (falling back to `Transform_Sha256_Len_base`, which would then also need to be compiled in), or document in cpuid.h that `CPUID_ASIMD` on AArch32 is informational only because NEON is a build-time requirement.

wolfcrypt/src/cpuid.c
```diff
@@ -567,6 +568,231 @@
...
+    /* getauxval() reports NEON in AT_HWCAP and the crypto-extension features in
+     * AT_HWCAP2 on 32-bit Arm.  Android is covered here too: bionic defines
+     * __linux__ and has provided getauxval() since API level 18. */
+
+    #include <sys/auxv.h>
```
Frauschi, line 655: **🟡 [Medium] AT_HWCAP2 used without a fallback definition in the new AArch32 cpuid detection**
*💡 SUGGEST*

**Sites:** `wolfcrypt/src/cpuid.c:644-671, 695-737`
The new 32-bit Arm Linux and FreeBSD detection branches defensively `#ifndef`-guard `HWCAP_NEON`, `HWCAP2_AES`, `HWCAP2_PMULL` and `HWCAP2_SHA2` ("in case the kernel headers predate them"), but call `getauxval(AT_HWCAP2)` / `elf_aux_info(AT_HWCAP2, ...)` without guarding `AT_HWCAP2` itself. `AT_HWCAP2` comes from `<elf.h>`/`<linux/auxvec.h>` and is absent on older glibc (< 2.18), older uClibc-ng, and cross toolchains whose kernel headers predate it - exactly the class of toolchain that would also be missing the `HWCAP2_*` bits the code does guard. In that case `wolfcrypt/src/cpuid.c` fails to compile for any 32-bit Arm `--enable-armasm` build. The PPC64 branch in the same file (`cpuid.c:800-802`) already demonstrates the correct pattern with `#ifndef AT_HWCAP2 / #define AT_HWCAP2 26 / #endif`, so the omission is inconsistent within the file.

**Recommendation:** Add `#ifndef AT_HWCAP2 / #define AT_HWCAP2 26 / #endif` (and, for symmetry, `AT_HWCAP` 16) to both the `__linux__` and `__FreeBSD__` AArch32 branches, matching the existing PPC64 branch at cpuid.c:800.

wolfssl/wolfcrypt/cpuid.h
```diff
@@ -43,6 +43,15 @@
     #define HAVE_CPUID
     #define HAVE_CPUID_AARCH64
 #endif
+/* 32-bit Arm (AArch32 / Thumb-2).  __arm__ is defined for both the A32 and
```
Frauschi: **🟡 [Medium] HAVE_CPUID_ARM32 has no architecture floor - enabled for Thumb-2/Cortex-M, ARMv6 and ARMv4, where the CP15 ID_ISAR5 read does not exist**
*💡 SUGGEST*

**Sites:** `wolfssl/wolfcrypt/cpuid.h:46-54`; `wolfcrypt/src/cpuid.c:591-638`
The new gate is `(defined(WOLFSSL_ARM32_BUILD) || (defined(__arm__) && defined(WOLFSSL_ARMASM))) && !defined(__aarch64__) && !defined(WOLFSSL_NO_ASM)`. `__arm__` is defined for every 32-bit Arm target, and configure.ac enables `WOLFSSL_ARMASM` on `armv7m*` (Cortex-M/Thumb-2, configure.ac:4137-4152), `armv6*` (configure.ac:4155) and `armv4*` (configure.ac:4163) hosts. None of those implement the Armv8 crypto extension, and ARMv7-M/ARMv6-M have no CP15 coprocessor at all. `HAVE_CPUID`/`HAVE_CPUID_ARM32` are nonetheless now defined for them, pulling in the ARM32 detection block, the `cpuid_flags` atomic global and the `cpuid_get_flags`/`cpuid_select_flags`/`cpuid_set_flag`/`cpuid_clear_flag` API. Worse, if `WOLFSSL_ARM32_PRIVILEGE_MODE` is set on such a target, `cpuid.c:612-617` emits `mrc p15, 0, %[reg], c0, c2, 5` unconditionally: ID_ISAR5 was introduced with ARMv7-A/R, and Cortex-M has no CP15, so the build either fails to assemble or executes an undefined instruction (HardFault) at run time. The comment in cpuid.h claims `__arm__` "cleanly distinguishes the 32-bit build", but it does not distinguish Armv8-A/32 from pre-Armv8 or M-profile.

**Also, at `wolfcrypt/src/cpuid.c:591-638` (Low, review scan) — WOLFSSL_ARM32_PRIVILEGE_MODE emits a CP15 MRC that faults on Cortex-M:**

`HAVE_CPUID_ARM32` is enabled for any `__arm__ && WOLFSSL_ARMASM` build, which includes Thumb-2/ARMv7-M (`WOLFSSL_ARMASM_THUMB2`) targets. If such a target also defines `WOLFSSL_ARM32_PRIVILEGE_MODE` - which reads privileged and is the natural choice for bare-metal - the `mrc p15, 0, %[reg], c0, c2, 5` either fails to assemble or takes an undefined-instruction fault, because ARMv7-M has no CP15 coprocessor interface (ID_ISAR5 is a memory-mapped SCB register there). Thumb-2 also has no crypto-extension AES and no run-time dispatch, so there is no consumer of the flags on that target in the first place.

**Recommendation:** Add an architecture floor to the gate, e.g. exclude `WOLFSSL_ARMASM_THUMB2` and require `!defined(WOLFSSL_ARM_ARCH) || WOLFSSL_ARM_ARCH >= 8` (or at minimum `>= 7`), and additionally guard the `mrc p15 ... c0, c2, 5` read in the `WOLFSSL_ARM32_PRIVILEGE_MODE` branch on an A/R-profile ARMv7+ target.

[wolfcrypt/src/cpuid.c:591-638] Exclude `WOLFSSL_ARMASM_THUMB2` from `HAVE_CPUID_ARM32` in cpuid.h, or add an `#error` in the privileged branch for ARMv7-M (`__ARM_ARCH_PROFILE == 'M'`).

Frauschi: **🔵 [Low] WOLFSSL_ARM32_BUILD is referenced by cpuid.h but never defined by the build system**
*💡 SUGGEST*  `question`

The new `HAVE_CPUID_ARM32` trigger mirrors the AArch64 one, which keys off `WOLFSSL_AARCH64_BUILD` - a macro `configure.ac:1225` emits for `host_cpu = aarch64`. There is no analogous `-DWOLFSSL_ARM32_BUILD` for `*arm*` hosts anywhere in `configure.ac`, so that arm of the condition is dead in autotools builds; only `__arm__ && WOLFSSL_ARMASM` ever fires. The macro was added to `.wolfssl_known_macro_extras`, which silences the macro linter but leaves the intent unclear: is it meant to be user-defined (like `WOLFSSL_ARM32_PRIVILEGE_MODE`), or was the configure.ac hunk forgotten?

**Suggestion:**
```suggestion
# configure.ac, next to the existing aarch64 line 1225:
AS_IF([test "x$host_cpu" = "xarm" || test "x$host_cpu" = "xarmv7l"],
      [AM_CFLAGS="$AM_CFLAGS -DWOLFSSL_ARM32_BUILD"])
```

**Recommendation:** Either wire `WOLFSSL_ARM32_BUILD` into configure.ac mirroring line 1225, or drop the term from the condition and note in the comment that it is a user-supplied macro for non-autotools builds.

wolfcrypt/src/cpuid.c
Frauschi, line 655: **🔵 [Low] cpuid.c ARM32 Linux path pulls in the kernel <asm/hwcap.h>**
*💡 SUGGEST*  `convention`

Every other Linux branch in this file (AArch64 at cpuid.c:~250, PPC64 at cpuid.c:823) includes only `<sys/auxv.h>`. The new ARM32 branch additionally includes `<asm/hwcap.h>`, a kernel UAPI header that is not guaranteed present on every 32-bit Arm toolchain/sysroot (musl and uClibc setups, and cross toolchains built without kernel headers). Since the code already provides `#ifndef HWCAP_NEON`/`HWCAP2_*` fallbacks with the correct bit values immediately below, the include buys nothing and only adds a build-break surface. It is also a possible source of macro-redefinition diagnostics under `-Werror` if a toolchain defines the same names non-identically.

**Suggestion:**
```suggestion
#include <sys/auxv.h>

    /* Feature bits defined locally: glibc's <bits/hwcap.h> spells the 32-bit
     * Arm NEON bit HWCAP_ARM_NEON, and the HWCAP2_* names come only from the
     * kernel UAPI header, which is not present in every sysroot. */
    #ifndef HWCAP_NEON
        #define HWCAP_NEON      (1 << 12)
    #endif
```

**Recommendation:** Drop `#include <asm/hwcap.h>` and rely on the local `#ifndef` definitions, matching the AArch64 and PPC64 branches.

Frauschi (commented):
## 🐺 Skoll Code Review

**Findings:** 14 total — 10 posted, 4 skipped

### Posted findings
- [High] **Dangling `else` swallows the CBC-decrypt fallback in 32-bit dispatch builds with WOLFSSL_ARMASM_NEON_NO_TABLE_LOOKUP** — `wolfcrypt/src/aes.c:7419-7461`
- [Medium] **WOLFSSL_AESGCM_STREAM disables the Armv8 crypto extension for every AES mode on 32-bit Arm, not just GCM** — `wolfcrypt/src/aes.c:1123-1155`
- [Medium] **IS_ARM32_ASIMD is detected on every AArch32 path but no dispatch site consults it (SHA-256 NEON transform is ungated)** — `wolfcrypt/src/sha256.c:1358-1382`
- [Medium] **AT_HWCAP2 used without a fallback definition in the new AArch32 cpuid detection** — `wolfcrypt/src/cpuid.c:644-671, 695-737`
- [Medium] **HAVE_CPUID_ARM32 has no architecture floor - enabled for Thumb-2/Cortex-M, ARMv6 and ARMv4, where the CP15 ID_ISAR5 read does not exist** — `wolfssl/wolfcrypt/cpuid.h:46-54`
- [Low] **WOLFSSL_ARM32_BUILD is referenced by cpuid.h but never defined by the build system** — `wolfssl/wolfcrypt/cpuid.h:46-54`
- [Low] **cpuid.c ARM32 Linux path pulls in the kernel <asm/hwcap.h>** — `wolfcrypt/src/cpuid.c:644-659`
- [Low] **Aes struct grows for 32-bit builds that never dispatch** — `wolfssl/wolfcrypt/aes.h:334-349`
- [Low] **Sha256_SetTransform writes the transform pointer and its guard flag without synchronization** — `wolfcrypt/src/sha256.c:1360-1382`
- [Info] **Stale #endif comments after the guard conditions changed** — `wolfcrypt/src/aes.c:1153, wolfssl/wolfcrypt/aes.h:981`

- [Medium] No CI job exercises the new 32-bit run-time dispatch on a non-crypto CPU
- [Low] configure.ac accepts no-base / no-neon-impl combinations that are contradictory or meaningless, and silently ignores them
- [Low] AES-GCM-SIV key schedule and CTR variant can disagree when HAVE_AESGCM is not defined
- [Info] GCM-SIV POLYVAL selector uses PMULL alone while every other 32-bit selector requires AES && PMULL

</details>

wolfssl/wolfcrypt/aes.h
```diff
@@ -331,14 +331,21 @@ struct Aes {
         #define WC_FLAG_DONT_USE_VECTOR_OPS 2
     #endif
 #endif /* WOLFSSL_AESNI */
-#if defined(__aarch64__) && defined(WOLFSSL_ARMASM) && \
-    !defined(WOLFSSL_ARMASM_NO_HW_CRYPTO)
+/* Run-time implementation-selection flags.  Present when a hardware-crypto and
```
Frauschi, line 360: **🔵 [Low] Aes struct grows for 32-bit builds that never dispatch**
*🔧 NIT*  `api`

The header's new condition (`WOLFSSL_ARMASM && !NO_HW_CRYPTO && (__aarch64__ || !NO_BASE_IMPL)`) is broader than aes.c's `WOLFSSL_ARM32_AES_DISPATCH` (which additionally requires `!WOLFSSL_ARMASM_THUMB2` and that the `#elif defined(WOLFSSL_ARMASM)` arm of the implementation chain at aes.c:1085 be taken). So Thumb-2/Cortex-M builds - and 32-bit builds that select another implementation branch such as `STM32_CRYPTO` - now carry `use_aes_hw_crypto` and `use_pmull_hw_crypto` in every `Aes` object, never written and never read. Two dead bytes per object on the most memory-constrained targets wolfSSL supports. No correctness risk today (nothing reads them in those builds), but the divergence between the two conditions is the kind of thing that becomes a real uninitialised-read bug on the next refactor.

**Suggestion:**
```suggestion
#if defined(WOLFSSL_ARMASM) && !defined(WOLFSSL_ARMASM_NO_HW_CRYPTO) && \
    (defined(__aarch64__) || (!defined(WOLFSSL_ARMASM_THUMB2) && \
     !defined(WOLFSSL_ARMASM_NO_BASE_IMPL)))
    byte use_aes_hw_crypto;
```

**Recommendation:** Add `!defined(WOLFSSL_ARMASM_THUMB2)` to the 32-bit arm of the header condition so it matches `WOLFSSL_ARM32_AES_DISPATCH`, or better, promote the dispatch macro to a header so struct layout and dispatch logic derive from a single expression.

wolfcrypt/src/sha256.c
```diff
@@ -1342,6 +1342,48 @@ int wc_InitSha256_ex(wc_Sha256* sha256, void* heap, int devId)
...
+static cpuid_flags_atomic_t sha256_cpuid_flags = WC_CPUID_ATOMIC_INITIALIZER;
+
```
Frauschi, line 1370: **🔵 [Low] Sha256_SetTransform writes the transform pointer and its guard flag without synchronization**
*🔧 NIT*  `convention`

**Sites:** `wolfcrypt/src/sha256.c:1360-1382`; `wolfcrypt/src/sha256.c:1356-1381`
`sha256_cpuid_flags` is a `cpuid_flags_atomic_t` read through `cpuid_get_flags_atomic()`, but the two pieces of state it feeds - `sha256_transform_check` (plain `int`) and `Transform_Sha256_Len_p` (plain function pointer) - are read and written without any synchronisation from `wc_InitSha256_ex()`, which applications call concurrently. The writes are idempotent so the race is benign in practice, but the file already contains a stricter template a few hundred lines up: the PPC64 selector (sha256.c:1155-1160) uses `wolfSSL_Atomic_Uint` with an explicit comment justifying the idempotent double-write. Following either that pattern or the AArch64 one consistently, rather than half of each, would make the intent legible.

**Also, at `wolfcrypt/src/sha256.c:1356-1381` (Low, portability scan) — Sha256_SetTransform writes the transform pointer and guard flag without synchronization:**

`sha256_cpuid_flags` is correctly declared `cpuid_flags_atomic_t` and read through `cpuid_get_flags_atomic()`, but the two objects that actually drive dispatch - `static int sha256_transform_check` and `static void (*Transform_Sha256_Len_p)(...)` - are plain statics written non-atomically from `wc_InitSha256_ex()`. Two threads calling `wc_InitSha256_ex()` concurrently constitute a data race, which is undefined behaviour under C11/C++11 and is flagged by ThreadSanitizer and by MISRA/CERT CON43-C. In practice on Arm the observable outcome is a benign stale-pointer read (the initializer is the always-valid NEON transform), but the construct is not portable to compilers/platforms where an aligned pointer store is not single-copy atomic, and the same pattern would be unsafe if a third variant were ever added. This is newly introduced code.

**Suggestion:**
```suggestion
/* Both writes are idempotent - every caller computes the same value from the
 * atomic master flags - so a concurrent double-write is harmless.  (Same
 * argument as the PPC64 selector above.) */
static int sha256_transform_check = 0;
```

**Recommendation:** Either store the selection in a `wolfSSL_Atomic_Uint` index like the PPC64 path, or add the one-line benign-race justification so the mixed atomicity is clearly deliberate.

[wolfcrypt/src/sha256.c:1356-1381] Drop the separate `sha256_transform_check` and derive the decision from the already-atomic `sha256_cpuid_flags` (i.e. select the transform inline at each call, or store the selection in an atomic word), so no non-atomic shared object is written after initialization.

wolfcrypt/src/aes.c
```diff
@@ -1098,6 +1111,46 @@ static void Check_CPU_support_HwCrypto(Aes* aes)
...
+     * base key schedule.  Only take the crypto AES path when PMULL is present
+     * too, so the whole cipher stays consistent.  (A CPU implementing AES but
+     * not PMULL is rare - the crypto extension provides them together.) */
+    aes->use_aes_hw_crypto = IS_ARM32_AES(cpuid_flags) &&
+                             aes->use_pmull_hw_crypto;
+#else
+    aes->use_aes_hw_crypto = IS_ARM32_AES(cpuid_flags);
+#endif
+#endif /* WOLFSSL_AESGCM_STREAM */
+}
```
Frauschi, line 1131: **⚪ [Info] Stale #endif comments after the guard conditions changed**
*🔧 NIT*  `style`

Two closing comments no longer describe the conditions they close. aes.c:1153 reads `#endif /* __aarch64__ && !WOLFSSL_ARMASM_NO_HW_CRYPTO */` but now also terminates the `#elif defined(WOLFSSL_ARM32_AES_DISPATCH)` branch added by the diff. aes.h:981 reads `#endif /* __aarch64__ || WOLFSSL_ARMASM_NO_HW_CRYPTO */` while the condition is now `__aarch64__ || WOLFSSL_ARMASM_NO_HW_CRYPTO || !WOLFSSL_ARMASM_NO_BASE_IMPL`. Elsewhere the diff was careful to update these (aes.c:7459-7460, 11265-11266, 12082-12083), so the two misses stand out. Similarly aes.c:7952 still says `/* __aarch64__ || WOLFSSL_ARMASM_NO_HW_CRYPTO */` for a block whose guard gained `WOLFSSL_ARM32_AES_DISPATCH` at line 7920.

**Suggestion:**
```suggestion
#endif /* (__aarch64__ && !WOLFSSL_ARMASM_NO_HW_CRYPTO) ||
        * WOLFSSL_ARM32_AES_DISPATCH */
```

**Recommendation:** Update the three stale `#endif` comments to match their conditions.

wolfcrypt/src/aes.c
SparkiDev: fixed

wolfcrypt/src/aes.c
SparkiDev: Fixed.

wolfcrypt/src/sha256.c
SparkiDev: Fixed

wolfcrypt/src/aes.c
SparkiDev: Fixed

wolfcrypt/src/sha256.c
SparkiDev: None needed.

wolfssl/wolfcrypt/aes.h
SparkiDev: Fixed

wolfcrypt/src/cpuid.c
SparkiDev: Fixed

wolfssl/wolfcrypt/cpuid.h
SparkiDev: Not fixing - same as WOLFSSL_AARCH64_BUILD.

wolfcrypt/src/cpuid.c
SparkiDev: Fixed

wolfssl/wolfcrypt/cpuid.h
SparkiDev: fixed

tests/api.c
```diff
@@ -10160,7 +10160,11 @@ static int test_wolfSSL_SCR_check_enabled(void)
     ExpectIntEQ(1, wolfSSL_CTX_get_scr_check_enabled(ctx_c));
     ExpectIntEQ(1, wolfSSL_get_scr_check_enabled(ssl_c));
     ExpectIntEQ(0, test_memio_do_handshake(ssl_c, ssl_s, 10, NULL));
+#ifdef HAVE_SERVER_RENEGOTIATION
```
Frauschi: **🟠 [High] test_wolfSSL_SCR_check_enabled guards on a macro that does not exist, inverting the assertion in every build**
*🚫 BLOCK*  `bug`

The new guard uses `HAVE_SERVER_RENEGOTIATION`, which is defined nowhere in the tree - every other reference (configure.ac, internal.h, the enclosing `#if` of this very test at tests/api.c:10144, tests/api/test_ssl_ext.c, tests/api/test_dtls.c) uses `HAVE_SERVER_RENEGOTIATION_INFO`, and the object-level feature macro is `HAVE_SECURE_RENEGOTIATION`. Because the macro is never defined, the `#ifdef` branch is dead in every configuration and the `#else` is always taken. `wolfSSL_SSL_get_secure_renegotiation_support()` (src/ssl_api_ext.c:968) returns exactly 0 or 1, so the original `ExpectIntNE(0, ...)` ("SCR was negotiated") has been replaced by `ExpectIntNE(1, ...)` ("SCR was NOT negotiated") - the exact opposite - for all builds. In a `--enable-secure-renegotiation` build (configure.ac:8925 defines both `HAVE_SECURE_RENEGOTIATION` and `HAVE_SERVER_RENEGOTIATION_INFO`, so this test runs) the phase-1 handshake does negotiate SCR, the getter returns 1, and `ExpectIntNE(1, 1)` fails. In builds that define only `HAVE_SERVER_RENEGOTIATION_INFO` (e.g. configure.ac:12729) the getter returns 0 and the assertion passes vacuously, which is presumably the failure the change was meant to fix - but the typo makes it fix one config by breaking the other, and it silently removes the assertion the test exists for. Two further notes: (a) the assertion should be `ExpectIntEQ(0, ...)` rather than `ExpectIntNE(1, ...)`, which passes for any non-1 value; (b) the wolfSSL unknown-macro check consumes `.wolfssl_known_macro_extras`, and this PR edits that file for the two new ARM32 macros but does not add `HAVE_SERVER_RENEGOTIATION` - further evidence this is a typo rather than an intentional out-of-tree knob. Finally, this hunk is unrelated to the PR's subject (32-bit ARM CPU-ID dispatch) and has no explanation in the commit.

**Suggestion:**
```suggestion
#ifdef HAVE_SECURE_RENEGOTIATION
    ExpectIntNE(0, (int)wolfSSL_SSL_get_secure_renegotiation_support(ssl_c));
#else
    /* Without HAVE_SECURE_RENEGOTIATION the SSL object has no
     * secure_renegotiation state, so the getter always reports 0. */
    ExpectIntEQ(0, (int)wolfSSL_SSL_get_secure_renegotiation_support(ssl_c));
#endif
```

**Recommendation:** Use `HAVE_SECURE_RENEGOTIATION` (the macro that controls whether `ssl->secure_renegotiation` is allocated) and make the negative branch an exact `ExpectIntEQ(0, ...)`. Verify against both `--enable-secure-renegotiation` and a build with only `HAVE_SERVER_RENEGOTIATION_INFO`. Because this is unrelated to the ARM32 dispatch work, split it into its own commit with a one-line rationale.

Frauschi (commented):
## 🐺 Skoll Code Review

**Findings:** 10 total — 5 posted, 5 skipped

### Posted findings
- [High] **test_wolfSSL_SCR_check_enabled guards on a macro that does not exist, inverting the assertion in every build** — `tests/api.c:10163-10167`
- [Medium] **SHA-224 now calls Sha256_SetTransform() in every WOLFSSL_ARMASM build, including ones where it is never defined** — `wolfcrypt/src/sha256.c:2587-2591`
- [Medium] **WOLFSSL_AESGCM_STREAM disables crypto-extension AES for every AES-GCM object on 32-bit ARM, including one-shot GCM** — `wolfcrypt/src/aes.c:8517-8530`
- [Medium] **--enable-armasm=no-neon-impl removes the SHA-256 run-time dispatch entirely, stranding the base transform that is still assembled** — `wolfcrypt/src/sha256.c:1351-1355`
- [Low] **ARM32 SHA-256 selector inverts the ASIMD test inside a #ifndef, diverging from the AArch64 selector** — `wolfcrypt/src/sha256.c:1378-1395`

- [Medium] No CI job exercises the new 32-bit run-time fallback (dispatch build on a CPU without the crypto extension)
- [Low] New configure options no-base / no-neon-impl have no build coverage
- [Low] Public aes.h now pulls in cpuid.h, exporting generic CPUID_* / HAVE_CPUID* macros to every consumer
- [Low] no-base / no-neon-impl are silently ignored when combined with no-crypto
- [Info] ARM32 privilege-mode branch reads AES/SHA2 from ID_ISAR5 but takes NEON from the compiler

</details>

wolfcrypt/src/sha256.c
```diff
@@ -2517,8 +2584,9 @@ static WC_INLINE int Transform_Sha256_Len(wc_Sha256* sha256, const byte* data,
     #else
         Sha256_SetTransform();
     #endif
-    #elif defined(WOLFSSL_ARMASM) && defined(__aarch64__) && \
-          !defined(WOLF_CRYPTO_CB_ONLY_SHA256)
+    #elif defined(WOLFSSL_ARMASM) && !defined(WOLF_CRYPTO_CB_ONLY_SHA256)
```
Frauschi: **🟡 [Medium] SHA-224 now calls Sha256_SetTransform() in every WOLFSSL_ARMASM build, including ones where it is never defined**
*💡 SUGGEST*  `bug`

The call-site guard was widened from `WOLFSSL_ARMASM && __aarch64__ && !WOLF_CRYPTO_CB_ONLY_SHA256` to `WOLFSSL_ARMASM && !WOLF_CRYPTO_CB_ONLY_SHA256`. `Sha256_SetTransform` (or its `WC_DO_NOTHING` macro form) is only defined inside the two ARMASM arms of the SHA-256 implementation `#elif` chain - sha256.c:1217 (AArch64) and sha256.c:1343 (AArch32). That chain is ordered, and many hardware-hash ports come first: `FREESCALE_MMCAU_SHA` (line 782), `WOLFSSL_PIC32MZ_HASH` (878), `WOLFSSL_SCE` (988), `WOLFSSL_USE_ESP32_CRYPT_HASH_HW` (1045), `WOLFSSL_IMXRT_DCP` (1119), `WOLFSSL_NXP_HASHCRYPT_SHA` (1123), `WOLFSSL_SILABS_SE_ACCEL` (1126) and others. Several of those are 32-bit ARM MCU ports where `--enable-armasm` (Thumb-2) is a supported combination, and several of them are *not* excluded from the software SHA-224 `#else` branch at line 2543 (`WOLFSSL_IMXRT_DCP`, `WOLFSSL_NXP_HASHCRYPT_SHA`, `FREESCALE_MMCAU_SHA`, `WOLFSSL_TI_HASH`, `WOLFSSL_CRYPTOCELL`, ...). In such a build `Sha256_SetTransform` is never defined but is now called - an implicit-declaration error under the project's `-Werror` builds. On AArch64 the same shape existed but was unreachable in practice; on 32-bit ARM it is not. Contrast the PPC call site immediately below (lines 2592-2596), whose guard `WOLFSSL_PPC64_ASM && WOLFSSL_PPC64_ASM_CRYPTO` mirrors its definition guard at line 1144 exactly.

**Suggestion:**
```suggestion
/* In the ARMASM SHA-256 branches (sha256.c:1217 / :1343): */
#define WOLFSSL_ARMASM_SHA256_TRANSFORM

/* At the SHA-224 call site: */
    #elif defined(WOLFSSL_ARMASM_SHA256_TRANSFORM)
        /* SHA-224 shares the SHA-256 transform, on AArch32 as well as AArch64;
         * a no-op in builds that compile a single variant. */
        Sha256_SetTransform();
    #endif
```

**Recommendation:** Have each ARMASM SHA-256 branch define a marker macro (e.g. `WOLFSSL_ARMASM_SHA256_TRANSFORM`) alongside `Sha256_SetTransform`, and key the SHA-224 call site off that marker so the call can never outlive its definition - the same discipline the PPC call site already follows. Add a build of `--enable-armasm` plus one of the ARM hardware-hash ports (e.g. `-DWOLFSSL_IMXRT_DCP`) with `WOLFSSL_SHA224` to confirm.

wolfcrypt/src/aes.c
```diff
@@ -8384,6 +8514,20 @@ int wc_AesGcmSetKey(Aes* aes, const byte* key, word32 len)
 #endif
     XMEMSET(iv, 0, WC_AES_BLOCK_SIZE);
     ret = wc_AesSetKey(aes, key, len, iv, AES_ENCRYPTION);
+#if defined(WOLFSSL_ARM32_AES_DISPATCH) && defined(WOLFSSL_AESGCM_STREAM)
```
Frauschi: **🟡 [Medium] WOLFSSL_AESGCM_STREAM disables crypto-extension AES for every AES-GCM object on 32-bit ARM, including one-shot GCM**
*💡 SUGGEST*  `bug`

The carve-out clears `use_aes_hw_crypto` / `use_pmull_hw_crypto` and re-expands the base key schedule for *every* AES-GCM key whenever `WOLFSSL_AESGCM_STREAM` is merely compiled in - not only when the streaming API is actually used. As a result, in any build that defines `WOLFSSL_AESGCM_STREAM` (FIPS v6 via configure.ac:7002/7243/7467, wolfZFS at 3770, bsdkm at 933, linuxkm-LKCAPI at 11665, or an explicit `--enable-aesgcm-stream`), one-shot `wc_AesGcmEncrypt`/`wc_AesGcmDecrypt` on a crypto-capable ARMv8-A 32-bit CPU stops using `AES_GCM_encrypt_AARCH32` / `AES_GCM_decrypt_AARCH32` and falls all the way back to table AES plus the software 4-bit-table GHASH - a large throughput cliff on exactly the CPUs the PR is meant to accelerate, and one that is invisible from the configure summary. It also silently downgrades any other mode driven from the same `Aes` object. The AArch64 sibling `Check_CPU_support_HwCrypto` does no such thing. The root cause is real (there is no `AES_GCM_init_AARCH32`, so streaming has no crypto assembly, and the two key schedules are incompatible), but paying for it on every GCM object is heavier than necessary.

**Suggestion:**
```suggestion
/* Retain the raw key for this build so the base schedule can be built lazily
 * when (and only when) the streaming API is entered, keeping the one-shot GCM
 * path on the crypto extension. */
#if defined(WOLFSSL_ARM32_AES_DISPATCH) && defined(WOLFSSL_AESGCM_STREAM)
    /* in wc_AesGcmInit()/wc_AesGcmEncryptInit()/wc_AesGcmDecryptInit(): */
    if (aes->use_aes_hw_crypto) {
        aes->use_aes_hw_crypto = 0;
        aes->use_pmull_hw_crypto = 0;
        AES_set_encrypt_key(aes->gcmRawKey, aes->keylen * 8, (byte*)aes->key);
        AES_ECB_encrypt(zeroIv, aes->gcm.H, WC_AES_BLOCK_SIZE,
            (const unsigned char*)aes->key, aes->rounds);
        GenerateM0(&aes->gcm);
    }
#endif
```

**Recommendation:** Move the downgrade to the streaming entry points (`wc_AesGcmInit` / `wc_AesGcmEncryptInit` / `wc_AesGcmDecryptInit`), retaining the raw key on the `Aes` object for this configuration so the schedule and `gcm.H` can be rebuilt there; that keeps one-shot GCM on the crypto extension. If that is judged too costly, keep the current behaviour but call it out explicitly in the configure summary / ChangeLog so users of FIPS v6, wolfZFS and linuxkm builds know AES-GCM is not hardware-accelerated on AArch32.

wolfcrypt/src/sha256.c
```diff
@@ -1342,6 +1342,62 @@ int wc_InitSha256_ex(wc_Sha256* sha256, void* heap, int devId)
 
 #elif defined(WOLFSSL_ARMASM) && !defined(WOLF_CRYPTO_CB_ONLY_SHA256)
 
+/* On 32-bit Arm a NEON build compiles in the base and NEON and (unless
+ * disabled) the Armv8 crypto-extension block transforms, so the best supported
+ * one is chosen at run time - mirroring the AArch64 path.  Thumb-2, no-NEON,
+ * no-crypto (NO_HW_CRYPTO) and crypto-only (NO_NEON_IMPL) builds compile a
+ * single variant and call it directly, as does a build with no run-time
+ * detection to dispatch on (HAVE_CPUID_ARM32). */
+#if !defined(WOLFSSL_ARMASM_THUMB2) && !defined(WOLFSSL_ARMASM_NO_NEON) && \
```
Frauschi, line 1359: **🟡 [Medium] --enable-armasm=no-neon-impl removes the SHA-256 run-time dispatch entirely, stranding the base transform that is still assembled**
*💡 SUGGEST*  `bug`

`SHA256_ARM32_DISPATCH` requires `!defined(WOLFSSL_ARMASM_NO_NEON_IMPL)`, so `--enable-armasm=no-neon-impl` (without `no-base`) turns off dispatch completely and `Transform_Sha256`/`Transform_Sha256_Len` fall through to the unconditional `Transform_Sha256_Len_crypto` call. Meanwhile the base transform is still assembled: armv8-32-sha256-asm.S guards it with `#if !defined(WOLFSSL_ARMASM_NO_BASE_IMPL) || defined(WOLFSSL_ARMASM_NO_NEON)`, which is true in that configuration. The result is (a) ~2.5k lines of `Transform_Sha256_Len_base` linked in but unreachable, and (b) a SIGILL on any CPU without FEAT_SHA256 even though a perfectly usable fallback is present in the binary. The user asked to drop the *NEON* fallback, not the base one; the natural behaviour would be a two-way crypto-vs-base dispatch. (`no-base` alone is handled correctly: it disables `WOLFSSL_ARM32_AES_HW_FLAGS`/dispatch precisely because no fallback remains.)

**Suggestion:**
```suggestion
#if !defined(WOLFSSL_ARMASM_THUMB2) && !defined(WOLFSSL_ARMASM_NO_NEON) && \
    !defined(WOLFSSL_ARMASM_NO_HW_CRYPTO) && \
    !(defined(WOLFSSL_ARMASM_NO_NEON_IMPL) && \
      defined(WOLFSSL_ARMASM_NO_BASE_IMPL)) && defined(HAVE_CPUID_ARM32)
    #define SHA256_ARM32_DISPATCH
#endif

/* ...and in Sha256_SetTransform(), skip the NEON arm when NO_NEON_IMPL. */
```

**Recommendation:** Only disable dispatch when *both* fallbacks are gone (`no-neon-impl` and `no-base`); otherwise keep dispatching and just drop the NEON arm from `Sha256_SetTransform`. Alternatively, if `no-neon-impl` is meant to imply "crypto is always present", make the asm drop the base transform too so at least no dead code is linked - and say so in the `--enable-armasm` help text.

wolfcrypt/src/sha256.c
```diff
@@ -1342,6 +1342,62 @@ int wc_InitSha256_ex(wc_Sha256* sha256, void* heap, int devId)
...
+#endif
+
+static void (*Transform_Sha256_Len_p)(wc_Sha256* sha256, const byte* data,
+    word32 len) = SHA256_ARM32_TRANSFORM_INIT;
+
+/* Select the crypto-extension transform when the CPU implements FEAT_SHA256,
+ * the NEON transform when it implements Advanced SIMD, and otherwise the base
+ * transform - which is compiled in alongside them unless dropped with
+ * WOLFSSL_ARMASM_NO_BASE_IMPL. */
+static void Sha256_SetTransform(void)
```
Frauschi, line 1388: **🔵 [Low] ARM32 SHA-256 selector inverts the ASIMD test inside a #ifndef, diverging from the AArch64 selector**
*🔧 NIT*  `style`

The three-way selection is written as `if (SHA256) crypto; #ifndef NO_BASE_IMPL else if (!ASIMD) base; #endif else neon;` - a conditionally-compiled `else if` spliced between an `if` and its `else`, with the feature test inverted. It is correct but hard to read, and it differs in shape from the AArch64 selector at sha256.c:1284-1308, which reads `if (SHA256) crypto; else if (ASIMD) neon; else base;` with the `#if`s wrapping whole arms. Matching the existing shape would also make the `WOLFSSL_ARMASM_NO_BASE_IMPL` case read naturally (no base arm at all) instead of relying on the reader to work out that the inverted test falls through to NEON.

**Suggestion:**
```suggestion
if (IS_ARM32_SHA256(sha256_cpuid_flags)) {
        Transform_Sha256_Len_p = Transform_Sha256_Len_crypto;
    }
#ifndef WOLFSSL_ARMASM_NO_BASE_IMPL
    else if (IS_ARM32_ASIMD(sha256_cpuid_flags)) {
        Transform_Sha256_Len_p = Transform_Sha256_Len_neon;
    }
    else {
        Transform_Sha256_Len_p = Transform_Sha256_Len_base;
    }
#else
    else {
        Transform_Sha256_Len_p = Transform_Sha256_Len_neon;
    }
#endif
```

**Recommendation:** Restructure to mirror the AArch64 selector: positive feature tests, one arm per implementation, `#if`s around whole arms.

dgarske (changes requested):
Please resolve merge conflicts. Thanks @SparkiDev

tests/api.c
SparkiDev: fixed

wolfcrypt/src/sha256.c
SparkiDev: fixed

wolfcrypt/src/aes.c

wolfcrypt/src/sha256.c
SparkiDev: fixed

wolfcrypt/src/sha256.c
SparkiDev: fixed

dgarske (changes requested):
Please fix merge conflicts. Thanks

dgarske (changes requested):
FAIL: scripts/unit.test in several jobs. Please check

SparkiDev:
Jenkins: retest this please

SparkiDev:
Jenkins: retest this please

dgarske (approved):
Correctness — one crypto-capable binary, three CPU classes:

```
┌───────────────┬────────────┬───────────────────────┬─────────────────────┬───────────────────────────────┐
│     Node      │ crypto ext │    detected flags     │ after testwolfcrypt │ before (parent) testwolfcrypt │
├───────────────┼────────────┼───────────────────────┼─────────────────────┼───────────────────────────────┤
│ Pi5 (A76)     │ yes        │ aes pmull sha256 neon │        PASS         │             PASS              │
├───────────────┼────────────┼───────────────────────┼─────────────────────┼───────────────────────────────┤
│ Pi4 (A72)     │ no         │ neon                  │  PASS (fell back)   │        SIGILL (rc=132)        │
├───────────────┼────────────┼───────────────────────┼─────────────────────┼───────────────────────────────┤
│ RobotPi (A53) │ no         │ neon                  │  PASS (fell back)   │        SIGILL (rc=132)        │
└───────────────┴────────────┴───────────────────────┴─────────────────────┴───────────────────────────────┘
```

That middle result is the whole point: the parent crashes with an illegal instruction on non-crypto ARM hardware (runs crypto asm unconditionally), while the PR detects the absence and safely falls back — same binary passes everywhere.

Benefit — Pi5 crypto vs NEON (MiB/s), the PR unlocking the A76's crypto ext for a portable build:

```
┌─────────────────┬──────┬────────┬─────────┐
│      Algo       │ NEON │ crypto │ Speedup │
├─────────────────┼──────┼────────┼─────────┤
│ SHA-256         │  256 │   1463 │  5.7x   │
├─────────────────┼──────┼────────┼─────────┤
│ AES-128-CBC-enc │  217 │   1380 │  6.4x   │
├─────────────────┼──────┼────────┼─────────┤
│ AES-128-CBC-dec │  216 │   3192 │  14.7x  │
├─────────────────┼──────┼────────┼─────────┤
│ AES-128-GCM-enc │   98 │   1641 │  16.7x  │
├─────────────────┼──────┼────────┼─────────┤
│ AES-256-GCM-enc │   84 │   1396 │  16.6x  │
└─────────────────┴──────┴────────┴─────────┘
```

No regression on the no-crypto nodes: the crypto build's runtime fallback matches the pure-NEON build exactly (Pi4 ~124 MiB/s SHA-256; RobotPi ~69).

PR #10958, opened 2026-07-21
wolfCrypt security hardening, portability fixes, and negative test coverage

Frauschi:
## Summary

This branch is a batch of eleven independent fixes across wolfCrypt and several hardware ports, from a security review. They fall into four groups: fail-open/validation hardening, undefined-behavior/portability fixes, hardware-port correctness, and negative regression test coverage. Each change is a self-contained commit; there is no shared refactoring.

## Validation and fail-open hardening

- **ChaCha20 rejects an unset key.** `wc_Chacha_Process` had no key-state check, so a zero-initialized context that received only a nonce would encrypt with an all-zero, attacker-predictable key and return success. A `keySet` flag is now set in `wc_Chacha_SetKey` and checked in `wc_Chacha_Process`, which returns `MISSING_KEY`, mirroring `wc_Arc4Process`.
- **DH validates untrusted moduli with random-witness Miller-Rabin.** `wc_DhSetKey_ex` loads parameters as untrusted but ran a fixed-base primality test (bases 2..19) when no RNG was supplied, which a crafted strong pseudoprime can pass. It now creates a temporary RNG so `mp_prime_is_prime_ex` runs with random witnesses, and falls back to the deterministic test only when no RNG can be obtained, so there is no new failure mode. Named FFDHE primes still short-circuit.
- **ECDH rejects a point-at-infinity shared secret.** The non-SP path of `wc_ecc_shared_secret_gen_sync` copied the affine x-coordinate to the output without checking for the identity point, so a shared secret that computed to infinity was returned as an all-zero secret with a success code. It now returns `ECC_INF_E`, as required by SP 800-56Ar3 section 5.7.1.2.

## Undefined behavior and portability

- **RSA (SE050): alignment-safe key-id read.** `wc_InitRsaKey_Id` cast the `byte id[]` array to `word32*` and dereferenced it, an unaligned 32-bit read that faults or misreads on strict-alignment targets and violates strict aliasing. It now uses `readUnalignedWord32`, matching the existing `wc_ecc_init_id`.
- **ML-KEM: alignment-safe reads and writes.** The little-endian fast paths of `mlkem_cbd_eta3` (input) and `mlkem_vec_compress_10_c` (output ciphertext) reinterpreted caller byte buffers through `word32*` pointers. Both now use `readUnalignedWord32` / `writeUnalignedWord32`, matching the neighboring `mlkem_cbd_eta2` and `mldsa_encode_w1_88_c`.

## Hardware-port correctness

- **PSoC6: bound ECDSA r/s before serialization.** `psoc6_ecc_verify_hash_ex` wrote attacker-supplied `r`/`s` of unbounded size into a fixed 132-byte stack buffer, a pre-authentication stack overflow reachable during TLS signature verification. It now rejects any component larger than the key size, matching the generic path's `wc_ecc_check_r_s_range`.
- **Intel QAT: surface hardware failures.** The synchronous `IntelQaSymCipher` never translated the `cpaCySymPerformOp` completion status into its return code, so a hardware failure returned success with the plaintext copy written to the ciphertext output. A non-success status now yields `ASYNC_OP_E`, matching the asynchronous port.
- **ESP32-C3: fix transposed register and mask.** The RSA accelerator deactivation passed the register address and bit mask in the wrong argument order (with a spurious base offset), so the accelerator was never powered down and stray writes hit a bogus address. The arguments now match the activation path and the other targets.

## Negative test coverage

- **`wc_ecc_check_key`.** Added a test that imports secp224r1 public keys that are off the curve and out of coordinate range, asserting `IS_POINT_E` and `ECC_OUT_OF_RANGE_E`. This closed a mutation-survival gap: the software on-curve and range checks previously had no negative coverage in a software build.
- **Ed448 signature S-range.** Added a test that signs a message and then verifies crafted signatures whose scalar `S` equals or exceeds the group order, including the malleability case `S + L`, asserting rejection. Without the range check, `(R, S + L)` verifies as a second valid signature for the same message.

### DIFF
--- a/doc/dox_comments/header_files/ecc.h
+++ b/doc/dox_comments/header_files/ecc.h
@@ -316,6 +316,8 @@ void wc_ecc_key_free(ecc_key* key);
     shared key
     \return MP_MEM may be returned if there is an error while computing the
     shared key
+    \return ECC_INF_E returned when the computed shared secret is the point at
+    infinity
 
     \param private_key pointer to the ecc_key structure containing the local
     private key
--- a/tests/api/test_dh.c
+++ b/tests/api/test_dh.c
@@ -317,6 +317,41 @@ int test_wc_DhSetKey(void)
 }
 
 /*
+ * wc_DhSetKey_ex validates untrusted parameters, so it must reject a composite
+ * modulus even when that composite is a strong pseudoprime to the fixed
+ * small-prime Miller-Rabin bases. n = 341550071728321 = 10670053 * 32010157 is
+ * a strong pseudoprime to bases 2, 3, 5, 7, 11, 13, 17 and 19, which the
+ * deterministic fixed-base test wrongly accepts as prime. Random-witness
+ * testing rejects it. As with the library's own primality check, rejection is
+ * probabilistic: eight random Miller-Rabin rounds leave a negligible
+ * (well under 1e-4) chance of accepting the composite, so a one-off failure
+ * here is statistical, not a regression.
+ */
+int test_wc_DhSetKey_ex_pseudoprime(void)
+{
+    EXPECT_DECLS;
+    /* wc_DhSetKey_ex takes no RNG, so it has to create one to get random
+     * witnesses. Builds that cannot allocate it fall back to the fixed-base
+     * test by design, and that test accepts this composite. */
+#if !defined(NO_DH) && !defined(HAVE_SELFTEST) && !defined(HAVE_FIPS) && \
+    !defined(WC_NO_RNG) && !defined(WOLFSSL_NO_MALLOC) && \
+    !defined(WOLFSSL_STATIC_MEMORY)
+    DhKey key;
+    /* n = 341550071728321, big-endian. */
+    byte p[] = { 0x01, 0x36, 0xA3, 0x52, 0xB2, 0xC8, 0xC1 };
+    byte g[] = { 0x02 };
+
+    XMEMSET(&key, 0, sizeof(key));
+
+    ExpectIntEQ(wc_InitDhKey(&key), 0);
+    ExpectIntEQ(wc_DhSetKey_ex(&key, p, sizeof(p), g, sizeof(g), NULL, 0),
+        WC_NO_ERR_TRACE(DH_CHECK_PUB_E));
+    wc_FreeDhKey(&key);
+#endif
+    return EXPECT_RESULT();
+}
+
+/*
  * Testing wc_DhSetNamedKey(), wc_DhGetNamedKeyParamSize(),
  * wc_DhCopyNamedKey() and wc_DhCmpNamedKey().
  */
--- a/wolfcrypt/src/dh.c
+++ b/wolfcrypt/src/dh.c
@@ -2670,10 +2670,37 @@ static int _DhSetKey(DhKey* key, const byte* p, word32 pSz, const byte* g,
         else
         #endif
         {
-            if (rng != NULL)
-                ret = mp_prime_is_prime_ex(keyP, 8, &isPrime, rng);
-            else
+#ifndef WC_NO_RNG
+            WC_RNG* checkRng = rng;
+            WC_RNG* tmpRng = NULL;
+
+            /* A fixed-base Miller-Rabin test can be fooled by a crafted
+             * composite, so use random witnesses when an RNG is available.
+             * Create a temporary RNG when the caller did not supply one. */
+            if (checkRng == NULL) {
+                if (wc_rng_new_ex(&tmpRng, NULL, 0, key->heap,
+                        INVALID_DEVID) == 0) {
+                    checkRng = tmpRng;
+                }
+            }
+
+            if (checkRng != NULL) {
+                ret = mp_prime_is_prime_ex(keyP, 8, &isPrime, checkRng);
+            }
+            else {
+                /* Fall back to the deterministic test rather than failing the
+                 * parameter load. This is the weaker check: a composite
+                 * crafted against the fixed bases is accepted as prime. */
+                WOLFSSL_MSG("DH: no RNG, primality test uses fixed bases");
                 ret = mp_prime_is_prime(keyP, 8, &isPrime);
+            }
+
+            /* Safe on NULL and zeroizes the RNG state before freeing. */
+            wc_rng_free(tmpRng);
+#else
+            (void)rng;
+            ret = mp_prime_is_prime(keyP, 8, &isPrime);
+#endif
         }
 
         if (ret == 0 && isPrime == 0)
--- a/wolfcrypt/src/ecc.c
+++ b/wolfcrypt/src/ecc.c
@@ -4869,6 +4869,14 @@ int wc_ecc_shared_secret_gen_sync(ecc_key* private_key, ecc_point* point,
 {
     int err = MP_OKAY;
     mp_int* k = ecc_get_k(private_key);
+#ifdef WOLFSSL_SE050
+    /* A key-id handle holds no software private scalar - ECDH for such a key
+     * runs on the SE050 - so the identity this function computes from the
+     * all-zero scalar is not a shared secret and must not be rejected here. */
+    int checkInf = !private_key->keyIdSet;
+#else
+    int checkInf = 1;
+#endif
 #ifdef HAVE_ECC_CDH
     WC_DECLARE_VAR(k_lcl, mp_int, 1, 0);
 #endif
@@ -4997,6 +5005,7 @@ int wc_ecc_shared_secret_gen_sync(ecc_key* private_key, ecc_point* point,
 #endif
 #if defined(WOLFSSL_SP_MATH)
     {
+        (void)checkInf;
         err = WC_KEY_SIZE_E;
         goto errout;
     }
@@ -5065,6 +5074,14 @@ int wc_ecc_shared_secret_gen_sync(ecc_key* private_key, ecc_point* point,
             err = ecc_map_ex(result, curve->prime, mp, 1);
         }
         if (err == MP_OKAY) {
+            /* SP 800-56Ar3 5.7.1.2: the shared secret must not be the point at
+             * infinity. ecc_map_ex reports that case as x, y of zero with a
+             * success code, so check the point explicitly. */
+            if (checkInf && wc_ecc_point_is_at_infinity(result)) {
+                err = ECC_INF_E;
+            }
+        }
+        if (err == MP_OKAY) {
             x = mp_unsigned_bin_size(curve->prime);
             if (*outlen < (word32)x || x < mp_unsigned_bin_size(result->x)) {
                 err = BUFFER_E;
@@ -5084,6 +5101,25 @@ int wc_ecc_shared_secret_gen_sync(ecc_key* private_key, ecc_point* point,
 
         wc_ecc_curve_free(curve);
         FREE_CURVE_SPECS();
+    }
+#endif
+
+#ifdef WOLFSSL_HAVE_SP_ECC
+    if ((err == MP_OKAY) && checkInf) {
+        /* The single precision implementations above serialize the point at
+         * infinity as an all-zero x-coordinate and report success, so the
+         * identity has to be recognized from the output. SP 800-56Ar3 5.7.1.2
+         * requires an error and stop in that case. Accumulate over the whole
+         * buffer so the scan does not branch on the secret. */
+        word32 i;
+        byte   acc = 0;
+
+        for (i = 0; i < *outlen; i++) {
+            acc |= out[i];
+        }
+        if (acc == 0) {
+            err = ECC_INF_E;
+        }
     }
 #endif
 
--- a/wolfcrypt/src/port/Espressif/esp32_mp.c
+++ b/wolfcrypt/src/port/Espressif/esp32_mp.c
@@ -567,12 +567,10 @@ static int esp_mp_hw_unlock(void)
          * This releases the RSA Accelerator from reset.*/
         portENTER_CRITICAL_SAFE(&wc_rsa_reg_lock);
         {
-            DPORT_REG_CLR_BIT(
-                 (volatile void *)(DR_REG_RSA_BASE + SYSTEM_CRYPTO_RSA_CLK_EN),
-                                   SYSTEM_PERIP_CLK_EN1_REG);
-            DPORT_REG_SET_BIT(
-                (volatile void *)(DR_REG_RSA_BASE + SYSTEM_RSA_MEM_PD),
-                                  SYSTEM_RSA_PD_CTRL_REG);
+            DPORT_REG_CLR_BIT((volatile void *)(SYSTEM_PERIP_CLK_EN1_REG),
+                              SYSTEM_CRYPTO_RSA_CLK_EN);
+            DPORT_REG_SET_BIT((volatile void *)(SYSTEM_RSA_PD_CTRL_REG),
+                              SYSTEM_RSA_MEM_PD);
         }
         portEXIT_CRITICAL_SAFE(&wc_rsa_reg_lock);
 #elif defined(CONFIG_IDF_TARGET_ESP32C6)
--- a/wolfcrypt/src/port/cypress/psoc6_crypto.c
+++ b/wolfcrypt/src/port/cypress/psoc6_crypto.c
@@ -2093,7 +2093,11 @@ int psoc6_ecc_verify_hash_ex(MATH_INT_T* r, MATH_INT_T* s, const byte* hash,
     uint8_t k[MAX_ECC_KEYSIZE] = { 0 };
 
     if (!key || !verif_res || !r || !s || !hash)
-        return -BAD_FUNC_ARG;
+        return BAD_FUNC_ARG;
+
+    /* Fail closed on both channels: every early return below leaves the
+     * caller's result flag in the failed state. */
+    *verif_res = 0;
 
     /* Enable CRYPTO block if not enabled */
     if (!Cy_Crypto_Core_IsEnabled(crypto_base)) {
@@ -2105,7 +2109,12 @@ int psoc6_ecc_verify_hash_ex(MATH_INT_T* r, MATH_INT_T* s, const byte* hash,
     sSz   = mp_unsigned_bin_size(s);
 
     if (keySz > MAX_ECC_KEYSIZE)
-        return -BAD_FUNC_ARG;
+        return BAD_FUNC_ARG;
+
+    /* Reject r or s values that would overflow their keySz slot in
+     * signature_buf when serialized. */
+    if (rSz > keySz || sSz > keySz)
+        return BAD_FUNC_ARG;
 
     /* Prepare ECC key */
     ecc_key.type     = PK_PUBLIC;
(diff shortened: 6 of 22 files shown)


wolfcrypt/src/dh.c
```diff
@@ -2670,10 +2670,38 @@ static int _DhSetKey(DhKey* key, const byte* p, word32 pSz, const byte* g,
...
+            }
+
+            /* Fall back to the deterministic test when no RNG could be
+             * obtained rather than failing the parameter load. */
+            if (checkRng != NULL)
+                ret = mp_prime_is_prime_ex(keyP, 8, &isPrime, checkRng);
             else
                 ret = mp_prime_is_prime(keyP, 8, &isPrime);
+
+            if (tmpRng != NULL) {
```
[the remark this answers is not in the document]
Frauschi: fixed

wolfcrypt/src/ecc.c
```diff
@@ -5043,6 +5043,15 @@ int wc_ecc_shared_secret_gen_sync(ecc_key* private_key, ecc_point* point,
             /* Use constant time map if compiled in */
             err = ecc_map_ex(result, curve->prime, mp, 1);
         }
+        if (err == MP_OKAY) {
```
dgarske, line 5076: **🟠 [Medium] ECDH infinity check is skipped for a zero scalar, which is the only way infinity occurs on prime-order curves**

The new guard is `if (!mp_iszero(k) && wc_ecc_point_is_at_infinity(result))`. On a prime-order curve (all NIST curves, cofactor 1) with a valid peer point, `k*P == infinity` if and only if `k == 0 mod n`. The carve-out therefore disables the check for the single most likely way to reach the identity in a software build: `key->k == 0`. A key imported with `d = 0` is rejected by `_ecc_import_raw_private` (`mp_iszero(key->k)` -> BAD_FUNC_ARG), but a caller that builds an `ecc_key` by other means (or an SE050 key-id key reaching `wc_ecc_shared_secret_ex` / `wc_ecc_shared_secret_ssh` directly, which is exactly the case the comment cites) still gets an all-zero secret with a success return. The check does still catch `k == n`, `k == 2n`, and small-subgroup results on cofactor>1 curves, so it is not useless, but the comment's rationale ("leave that case to the offload path") does not hold for `wc_ecc_shared_secret_ex`, which never dispatches to the offload path.

**Fix:** Either drop the `!mp_iszero(k)` exception (a zero scalar cannot produce a usable secret on this path anyway), or, if the SE050 key-id case genuinely reaches here with `k == 0`, gate the exception on that condition explicitly (e.g. `private_key->keyIdSet`) rather than on the scalar value.

dgarske, line 5076: **🟠 [Medium] Infinity rejection is absent from the SP shared-secret paths, so the hardening does not apply to P-256/384/521 in SP…**

The check was added only to the generic (non-SP) branch of `wc_ecc_shared_secret_gen_sync`. The SP branches above it call `sp_ecc_secret_gen_256/384/521`, which do not report the identity: `sp_ecc_secret_gen_256` runs `sp_256_ecc_mulmod_5(point, point, k, 1, 1, heap)` and then unconditionally does `sp_256_to_bin_5(point->x, out); *outLen = 32;` with `err == MP_OKAY`, so an identity result is serialized as a 32-byte all-zero secret with a success return - exactly the behavior this PR set out to remove. The new test deliberately selects secp224r1 "which is not single precision accelerated", which confirms the gap is known but it is not recorded anywhere in the code. In `--enable-sp` builds (the common TLS configuration) P-256/384/521 keep the old behavior.

**Fix:** Either extend the rejection to the SP paths (add the identity check inside `sp_ecc_secret_gen_*`, or add a post-condition check in `wc_ecc_shared_secret_gen_sync` after the SP calls), or add a comment on the SP branch documenting that the identity case is not detected there, so the limitation is discoverable.

wolfcrypt/src/dh.c
```diff
@@ -2670,10 +2670,39 @@ static int _DhSetKey(DhKey* key, const byte* p, word32 pSz, const byte* g,
         else
         #endif
         {
-            if (rng != NULL)
-                ret = mp_prime_is_prime_ex(keyP, 8, &isPrime, rng);
+#ifndef WC_NO_RNG
```
dgarske, line 2673: **🟠 [Medium] DH temporary RNG hand-rolls the allocate/init/free sequence instead of using wc_rng_new_ex()/wc_rng_free()**

wolfCrypt already provides `wc_rng_new_ex(&rng, nonce, nonceSz, heap, devId)` / `wc_rng_free(rng)` (wolfcrypt/src/random.c:2177, :2202) for precisely this pattern: it XMALLOCs, calls `_InitRng`, and on failure frees the block and NULLs the pointer; `wc_rng_free` also `ForceZero`s the struct before freeing. The new code duplicates that logic and, on a failed `wc_InitRng_ex`, calls `wc_FreeRng(tmpRng)` on the failed object (safe today only because `_InitRng` memsets the struct before it can fail) and then `XFREE`s it without zeroizing. Using the existing helper is shorter, removes the reliance on `_InitRng`'s internal memset ordering, and lets you pass a devId if that is ever wanted.

**Fix:** Replace the XMALLOC/wc_InitRng_ex/wc_FreeRng/XFREE sequence with `wc_rng_new_ex()` + `wc_rng_free()`.

dgarske, line 2673: **🟠 [Medium] wc_DhSetKey_ex now instantiates a DRBG per call and silently downgrades to the fixed-base test when it cannot**

Two behavioral consequences of the new path are worth calling out explicitly. (1) `wc_DhSetKey_ex()` changes from a pure-math parameter load into a function that allocates memory and consumes entropy on every non-FFDHE untrusted call; a full DRBG instantiation (seed gather plus, in FIPS builds, the instantiate health test) is not cheap on embedded targets and can block on entropy. (2) When the allocation or `wc_InitRng_ex` fails, the code falls back to the same fixed-base test the commit describes as defeatable, and returns success for the crafted composite. That means the hardening is unconditionally absent in `WOLFSSL_NO_MALLOC` builds (there `XMALLOC` is `((void)(s), (void)(h), (void)(t), NULL)`), and an attacker who can induce allocation failure gets the old behavior back with no signal. The fallback is documented as intentional to avoid a new failure mode, so this is a design question rather than a defect - but it should be a conscious decision.

**Fix:** Keep the fallback if the no-new-failure-mode requirement stands, but log it (`WOLFSSL_MSG`) so the downgrade is observable, and note the new entropy/allocation dependency of `wc_DhSetKey_ex()` in the API documentation (doc/dox_comments/header_files/dh.h).

tests/api/test_dh.c
```diff
@@ -316,6 +316,37 @@ int test_wc_DhSetKey(void)
...
+ * modulus even when that composite is a strong pseudoprime to the fixed
+ * small-prime Miller-Rabin bases. n = 341550071728321 = 10670053 * 32010157 is
+ * a strong pseudoprime to bases 2, 3, 5, 7, 11, 13, 17 and 19, which the
+ * deterministic fixed-base test wrongly accepts as prime. Random-witness
+ * testing rejects it. As with the library's own primality check, rejection is
+ * probabilistic: eight random Miller-Rabin rounds leave a negligible
+ * (well under 1e-4) chance of accepting the composite, so a one-off failure
+ * here is statistical, not a regression.
+ */
+int test_wc_DhSetKey_ex_pseudoprime(void)
```
dgarske, line 330: **🟠 [Medium] New DH pseudoprime test asserts an outcome the implementation only provides on a best-effort basis**

The test asserts `wc_DhSetKey_ex(...) == DH_CHECK_PUB_E` unconditionally, but the code under test deliberately falls back to the fixed-base Miller-Rabin test whenever the temporary RNG cannot be created - in which case 341550071728321 is accepted as prime and the call returns 0. The guard covers `WC_NO_RNG`, FIPS and selftest builds, but not `WOLFSSL_NO_MALLOC` (where `XMALLOC` is defined to return NULL, so the fallback always triggers) nor a static-memory build whose pool is exhausted. In those configurations this test fails even though the library is behaving as designed. The test comment already acknowledges the separate probabilistic-failure axis (~4^-8), which is fine.

**Fix:** Add `!defined(WOLFSSL_NO_MALLOC)` (and consider `!defined(WOLFSSL_STATIC_MEMORY)`) to the test guard, since the implementation cannot obtain an RNG in those builds by construction.

dgarske (commented):
## Skoll Code Review

**10 finding(s) posted as inline comments** (see file-level comments below)

### Posted findings
- [Medium] **ECDH infinity check is skipped for a zero scalar, which is the only way infinity occurs on prime-order curves** — `wolfcrypt/src/ecc.c:5046-5055`
- [Medium] **Infinity rejection is absent from the SP shared-secret paths, so the hardening does not apply to P-256/384/521 in SP builds** — `wolfcrypt/src/ecc.c:5046-5055`
- [Medium] **DH temporary RNG hand-rolls the allocate/init/free sequence instead of using wc_rng_new_ex()/wc_rng_free()** — `wolfcrypt/src/dh.c:2673-2706`
- [Medium] **wc_DhSetKey_ex now instantiates a DRBG per call and silently downgrades to the fixed-base test when it cannot** — `wolfcrypt/src/dh.c:2673-2695`
- [Medium] **New DH pseudoprime test asserts an outcome the implementation only provides on a best-effort basis** — `tests/api/test_dh.c:330-350`
- [Low] **PSoC6: new attacker-reachable early return leaves *verif_res unmodified** — `wolfcrypt/src/port/cypress/psoc6_crypto.c:2110-2114`
- [Low] **New PSoC6 rejection returns a positive error value (-BAD_FUNC_ARG)** — `wolfcrypt/src/port/cypress/psoc6_crypto.c:2113`
- [Low] **Two new lines exceed the 80-column source-text limit** — `wolfcrypt/src/dh.c:2684`
- [Low] **Japanese doxygen for ecc.h not updated alongside the English copy** — `doc/dox_comments/header_files/ecc.h:319-320`
- [Low] **ESP32-C3 register fix and other hardware-port changes have no automated coverage** — `wolfcrypt/src/port/Espressif/esp32_mp.c:567-576`

wolfcrypt/src/port/cypress/psoc6_crypto.c
```diff
@@ -2107,6 +2107,11 @@ int psoc6_ecc_verify_hash_ex(MATH_INT_T* r, MATH_INT_T* s, const byte* hash,
     if (keySz > MAX_ECC_KEYSIZE)
         return -BAD_FUNC_ARG;
 
+    /* Reject r or s values that would overflow their keySz slot in
```
dgarske, line 2114: **🔵 [Low] PSoC6: new attacker-reachable early return leaves *verif_res unmodified**

The new bound check returns before `*verif_res` is ever written (it is only assigned on the hardware-success path via `*verif_res = stat;`). Since `wc_ecc_verify_hash_ex` returns this function's value directly for `WOLFSSL_PSOC6_CRYPTO`, a caller that inspects only the result flag would observe a stale value for an attacker-supplied oversized r/s. In-tree callers are safe today (`EccVerify` in src/internal.c checks `ret != 0 || ssl->eccVerifyRes == 0`, and `wolfcrypt/src/asn.c` / `signature.c` check the return code first), and the two pre-existing early returns in this function have the same shape - but since this new path is the one an attacker can trigger, it is worth failing closed on both channels.

**Fix:** Set `*verif_res = 0` immediately after the NULL-argument check at the top of the function so every early return leaves the flag in the failed state.

wolfcrypt/src/port/cypress/psoc6_crypto.c
```diff
@@ -2107,6 +2107,11 @@ int psoc6_ecc_verify_hash_ex(MATH_INT_T* r, MATH_INT_T* s, const byte* hash,
...
+     * signature_buf when serialized. */
+    if (rSz > keySz || sSz > keySz)
+        return -BAD_FUNC_ARG;
```
dgarske: **🔵 [Low] New PSoC6 rejection returns a positive error value (-BAD_FUNC_ARG)**

`BAD_FUNC_ARG` is already negative (-173), so `return -BAD_FUNC_ARG;` propagates +173 out of `wc_ecc_verify_hash_ex`. Callers that test `ret != 0` (all in-tree ones) still fail closed, but callers using the common wolfSSL `if (ret < 0)` idiom would treat it as success. The new line follows the existing convention in this file, so changing it is a wider cleanup - flagging only so the choice is deliberate.

**Fix:** Consider returning `BAD_FUNC_ARG` (already negative) here, and separately normalizing the other `-BAD_FUNC_ARG` returns in this port.

wolfcrypt/src/dh.c
```diff
@@ -2670,10 +2670,39 @@ static int _DhSetKey(DhKey* key, const byte* p, word32 pSz, const byte* g,
...
+            WC_RNG* tmpRng = NULL;
+
+            /* A fixed-base Miller-Rabin test can be fooled by a crafted
+             * composite, so use random witnesses when an RNG is available.
+             * Create a temporary RNG when the caller did not supply one. */
+            if (checkRng == NULL) {
+                tmpRng = (WC_RNG*)XMALLOC(sizeof(WC_RNG), key->heap,
+                    DYNAMIC_TYPE_RNG);
+                if ((tmpRng != NULL) &&
+                        (wc_InitRng_ex(tmpRng, key->heap, INVALID_DEVID) == 0)) {
```
dgarske: **🔵 [Low] Two new lines exceed the 80-column source-text limit**

wolfSSL's source-text check flags lines longer than 80 columns on changed lines. Two lines added by this PR are 81 columns: `wolfcrypt/src/dh.c:2684` (`(wc_InitRng_ex(tmpRng, key->heap, INVALID_DEVID) == 0)) {`) and `tests/api/test_ecc.c:314` (`!defined(NO_ECC_CHECK_PUBKEY_ORDER) && !defined(WOLF_CRYPTO_CB_ONLY_ECC) && \`). All other added lines in the diff are within the limit.

**Fix:** Re-wrap both lines to 80 columns and run the check-source-text lint before pushing.

doc/dox_comments/header_files/ecc.h
```diff
@@ -316,6 +316,8 @@ void wc_ecc_key_free(ecc_key* key);
     shared key
     \return MP_MEM may be returned if there is an error while computing the
     shared key
+    \return ECC_INF_E returned when the computed shared secret is the point at
```
dgarske, line 319: **🔵 [Low] Japanese doxygen for ecc.h not updated alongside the English copy**

The ChaCha doc change was mirrored into `doc/dox_comments/header_files-ja/chacha.h`, but the new `ECC_INF_E` return value for `wc_ecc_shared_secret` was only added to the English `doc/dox_comments/header_files/ecc.h`; `doc/dox_comments/header_files-ja/ecc.h` was left unchanged. That file already documents `ECC_INF_E` for `wc_ecc_check_key`, so the two copies now disagree on `wc_ecc_shared_secret`.

**Fix:** Add the matching `\return ECC_INF_E` line to the Japanese ecc.h doc block for consistency with the ChaCha change in this same PR.

wolfcrypt/src/port/Espressif/esp32_mp.c
```diff
@@ -567,12 +567,10 @@ static int esp_mp_hw_unlock(void)
          * This releases the RSA Accelerator from reset.*/
```
dgarske, line 567: **🔵 [Low] ESP32-C3 register fix and other hardware-port changes have no automated coverage**

The register/mask swap is clearly correct by inspection - it now mirrors the C3 activation path at esp_mp_hw_lock() exactly (SET/CLR of SYSTEM_CRYPTO_RSA_CLK_EN in SYSTEM_PERIP_CLK_EN1_REG and SYSTEM_RSA_MEM_PD in SYSTEM_RSA_PD_CTRL_REG) - and the ordering is safe: every caller reads results with esp_memblock_to_mpint() before calling esp_mp_hw_unlock(), so powering down RSA memory after the read cannot corrupt results. Note the C3 unlock still does not call periph_module_disable() the way the ESP32/S2/S3 paths do (it now matches C6); that is benign because the lock path re-sets the clock bit explicitly rather than relying on the ESP-IDF refcount, but it does leave the IDF refcount permanently at 1. This change, plus the QAT and SE050 RSA changes, cannot be exercised by CI - they need a bench run.

**Fix:** Run wolfcrypt_test (RSA/bignum) on real ESP32-C3 hardware across repeated operations to confirm the accelerator re-activates correctly after the now-effective power-down, and note the result in the PR.

dgarske (changes requested):
https://github.com/wolfSSL/wolfssl/pull/10958#pullrequestreview-4831733000

wolfcrypt/src/ecc.c
Frauschi: Agreed - and worse than described: on a prime-order curve `k == 0` is the only way to reach infinity, so the carve-out disabled the check for the exact case it was added for. Now gated on `!private_key->keyIdSet` under `WOLFSSL_SE050`; that case reaches here via `wc_ecc_shared_secret_ssh`, which bypasses the hardware dispatch.

Frauschi: Fixed, not documented. On `--enable-all --enable-sp` a P-256 agreement with `d = n` returned `0` with an all-zero secret; it now returns `ECC_INF_E`. The test covers both backends, and the check costs ~28 of ~164,500 cycles/op.

wolfcrypt/src/dh.c
Frauschi: Done - also removes the `_InitRng` memset-ordering reliance and the missing zeroize.

Frauschi: Keeping the fallback, but added a `WOLFSSL_MSG` so it is observable and documented the entropy/allocation dependency in `dh.h`.

tests/api/test_dh.c
Frauschi: Agreed - the guard now excludes `WOLFSSL_NO_MALLOC` and `WOLFSSL_STATIC_MEMORY`.

wolfcrypt/src/port/cypress/psoc6_crypto.c
Frauschi: Fixed - `*verif_res = 0` after the NULL check, so every early return fails closed on both channels.

wolfcrypt/src/port/cypress/psoc6_crypto.c
Frauschi: Fixed. Narrower than a convention change: `-BAD_FUNC_ARG` appears 3 times in the file, all in this function, while the other ~40 returns already use plain `BAD_FUNC_ARG`.

wolfcrypt/src/dh.c
Frauschi: Both re-wrapped; the `dh.c` one vanished with the `wc_rng_new_ex()` rewrite. `check-source-text.sh` is clean.

doc/dox_comments/header_files/ecc.h
Frauschi: Added the matching `\return ECC_INF_E` line.

wolfcrypt/src/port/Espressif/esp32_mp.c
Frauschi: Agreed, including the IDF refcount staying at 1 - benign since the lock path sets the clock bit explicitly. I have no C3 board, so this is inspection-only; happy for someone with one to bench it, but I would rather merge than leave the transposed register in place.

dgarske (approved):

PR #10961, opened 2026-07-21
New API for CRL unknown extension callback

anhu:
Adds public entry points mirroring the existing X.509 unknown extension callback so callers can register a handler for unrecognized CRL extensions instead of failing with ASN_CRIT_EXT_E.

Fixes ZD 21887

### DIFF
--- a/src/crl.c
+++ b/src/crl.c
@@ -899,6 +899,13 @@ int BufferLoadCRL(WOLFSSL_CRL* crl, const byte* buff, long sz, int type,
     }
 
     InitDecodedCRL(dcrl, crl->heap);
+#ifdef WC_ASN_UNKNOWN_EXT_CB
+    if (crl->cm != NULL) {
+        dcrl->unknownExtCallback      = crl->cm->crlUnknownExtCallback;
+        dcrl->unknownExtCallbackEx    = crl->cm->crlUnknownExtCallbackEx;
+        dcrl->unknownExtCallbackExCtx = crl->cm->crlUnknownExtCallbackExCtx;
+    }
+#endif
     ret = ParseCRL(currentEntry->certs, dcrl, myBuffer, (word32)sz,
                    verify, crl->cm);
 
@@ -1224,6 +1231,13 @@ int GetCRLInfo(WOLFSSL_CRL* crl, CrlInfo* info, const byte* buff,
     }
 
     InitDecodedCRL(dcrl, crl->heap);
+#ifdef WC_ASN_UNKNOWN_EXT_CB
+    if (crl->cm != NULL) {
+        dcrl->unknownExtCallback      = crl->cm->crlUnknownExtCallback;
+        dcrl->unknownExtCallbackEx    = crl->cm->crlUnknownExtCallbackEx;
+        dcrl->unknownExtCallbackExCtx = crl->cm->crlUnknownExtCallbackExCtx;
+    }
+#endif
     ret = ParseCRL(crle->certs, dcrl, myBuffer, (word32)sz,
                    0, crl->cm);
     if (ret != 0 && !(ret == WC_NO_ERR_TRACE(ASN_CRL_NO_SIGNER_E))) {
--- a/tests/api/test_certman.c
+++ b/tests/api/test_certman.c
@@ -2846,12 +2846,10 @@ int test_wolfSSL_CRL_critical_idp(void)
     return EXPECT_RESULT();
 }
 
-int test_wolfSSL_CRL_unknown_critical_ext(void)
-{
-    EXPECT_DECLS;
 #if !defined(NO_CERTS) && defined(HAVE_CRL) && !defined(NO_RSA)
-
-    static const unsigned char ca_cert[] = {
+/* claim-root CA, DER.  Shared by the unknown-critical-CRL-extension tests
+ * below. */
+static const unsigned char crl_unk_ca_der[] = {
         0x30, 0x82, 0x03, 0x1b, 0x30, 0x82, 0x02, 0x03, 0xa0, 0x03, 0x02,
         0x01, 0x02, 0x02, 0x14, 0x1e, 0x25, 0xc1, 0x5d, 0x6f, 0x02, 0x21,
         0xa0, 0xf0, 0x14, 0x15, 0x9c, 0x3b, 0x4d, 0x1d, 0x73, 0x16, 0x00,
@@ -2927,10 +2925,10 @@ int test_wolfSSL_CRL_unknown_critical_ext(void)
         0xa2, 0xb0, 0x3e, 0x61, 0x16, 0x69, 0x8f
     };
 
-    /* CRL with critical obsolete extension OID 2.5.29.1, 422 bytes DER.
-     * OID 2.5.29.1 is the old X.509v2 Authority Key Identifier, permanently
-     * superseded by 2.5.29.35. No implementation will ever support it. */
-    static const unsigned char crl_obsolete_critical[] = {
+/* CRL with a critical obsolete CRL-level extension OID 2.5.29.1, 422 bytes
+ * DER.  OID 2.5.29.1 is the old X.509v2 Authority Key Identifier, permanently
+ * superseded by 2.5.29.35. No implementation will ever support it. */
+static const unsigned char crl_obsolete_critical[] = {
         0x30, 0x82, 0x01, 0xa6, 0x30, 0x81, 0x8f, 0x02, 0x01, 0x01, 0x30,
         0x0d, 0x06, 0x09, 0x2a, 0x86, 0x48, 0x86, 0xf7, 0x0d, 0x01, 0x01,
         0x0b, 0x05, 0x00, 0x30, 0x15, 0x31, 0x13, 0x30, 0x11, 0x06, 0x03,
@@ -2970,13 +2968,18 @@ int test_wolfSSL_CRL_unknown_critical_ext(void)
         0xff, 0xad, 0x4d, 0x6e, 0x10, 0x73, 0x68, 0xfa, 0x54, 0x9e, 0xdc,
         0x34, 0x70, 0xe4, 0x5d, 0x9e, 0x7c, 0xfa, 0x59, 0x97, 0xde, 0x35,
         0x17, 0xbb, 0xaf, 0xa0, 0x28, 0x78, 0x13, 0xbf
-    };
+};
+#endif /* !NO_CERTS && HAVE_CRL && !NO_RSA */
 
+int test_wolfSSL_CRL_unknown_critical_ext(void)
+{
+    EXPECT_DECLS;
+#if !defined(NO_CERTS) && defined(HAVE_CRL) && !defined(NO_RSA)
     WOLFSSL_CERT_MANAGER* cm = NULL;
 
     ExpectNotNull(cm = wolfSSL_CertManagerNew());
-    ExpectIntEQ(wolfSSL_CertManagerLoadCABuffer(cm, ca_cert,
-        sizeof(ca_cert), WOLFSSL_FILETYPE_ASN1), WOLFSSL_SUCCESS);
+    ExpectIntEQ(wolfSSL_CertManagerLoadCABuffer(cm, crl_unk_ca_der,
+        sizeof(crl_unk_ca_der), WOLFSSL_FILETYPE_ASN1), WOLFSSL_SUCCESS);
     ExpectIntEQ(wolfSSL_CertManagerEnableCRL(cm, WOLFSSL_CRL_CHECKALL),
         WOLFSSL_SUCCESS);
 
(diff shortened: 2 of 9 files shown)


anhu:
jenkins retest this please.

anhu:
jenkins retest this please

tests/api/test_certman.c
```diff
@@ -3014,6 +3017,175 @@ int test_wolfSSL_CRL_unknown_critical_entry_ext(void)
     return EXPECT_RESULT();
 }
 
+#if !defined(NO_CERTS) && defined(HAVE_CRL) && !defined(NO_RSA) && \
```
Frauschi, line 3020: **🔵 [Low] No test coverage for the non-Ex setter wolfSSL_CertManagerSetCRLUnknownExtCallback**
*🔧 NIT*  `test`

All three new tests exercise only the `...Ex` variant (`wolfSSL_CertManagerSetCRLUnknownExtCallbackEx`). The non-Ex public API `wolfSSL_CertManagerSetCRLUnknownExtCallback` (which sets `crlUnknownExtCallback` and drives the `dcrl->unknownExtCallback(...)` branch in both parse paths) is never called by any test, so that branch is uncovered.

**Suggestion:** Add a case that registers the non-Ex callback via wolfSSL_CertManagerSetCRLUnknownExtCallback and asserts it fires (the non-Ex callback signature omits the ctx argument).

Frauschi (commented):
## 🐺 Skoll Code Review

### Posted findings
- [Medium] **CRL unknown-ext callback is a silent no-op in non-template (WOLFSSL_NO_ASN_TEMPLATE) builds** — `wolfcrypt/src/asn_orig.c:9194, wolfcrypt/src/asn_orig.c:9394-9541`
- [Low] **No test coverage for the non-Ex setter wolfSSL_CertManagerSetCRLUnknownExtCallback** — `tests/api/test_certman.c:3020-3190`
- [Low] **Positive-return sanitization untested on the CRL-level extension path** — `tests/api/test_certman.c:3160-3190, wolfcrypt/src/asn.c:37351-37357`

wolfcrypt/src/asn.c
```diff
@@ -37279,9 +37320,48 @@ static int ParseCRL_Extensions(DecodedCRL* dcrl, const byte* buf, word32 idx,
                     mp_free(m);
                     FREE_MP_INT_SIZE(m, NULL, DYNAMIC_TYPE_TMP_BUFFER);
                 }
-                else if (critical) {
-                    WOLFSSL_MSG("Unknown critical CRL extension");
-                    ret = ASN_CRIT_EXT_E;
+                else {
```
Frauschi, line 37336: **🟡 [Medium] Accept-all CRL callback also silently bypasses recognized-but-unimplemented critical extensions (IDP, deltaCRLIndicator)**
*💡 SUGGEST*  `api`

The new dispatch sits in the `else` arm that catches *everything* except `AUTH_KEY_OID` and `CRL_NUMBER_OID`. That arm is not only genuinely-unknown OIDs: it is also where `issuingDistributionPoint` (2.5.29.28) and `deltaCRLIndicator` (2.5.29.27) land, because neither is in the `oidCertExtType` table and neither is decoded by wolfSSL. Today a critical IDP is rejected with `ASN_CRIT_EXT_E` — that rejection is exactly what `test_wolfSSL_CRL_critical_idp` (tests/api/test_certman.c) asserts. Once any accept-all callback is registered, `handled = 1` and that rejection disappears, so a scoped CRL (`onlyContainsUserCerts` / `onlyContainsCACerts` / `onlySomeReasons`) is loaded and then applied as if it were full-scope. Contrast this with the X.509 analogue at asn.c:22188, where `DecodeExtensionType` sets `isUnknownExt` only for OIDs it truly did not handle, so the cert-side callback has a narrower reach than this new CRL-side one. The ssl.h doc comment says the callback fires for extensions 'whose OID the parser does not recognize', which is technically true but will not lead a caller to expect CRL-scoping enforcement to switch off.

**Suggestion:**
```suggestion
/* Either exclude the security-relevant OIDs the parser knows about but
 * does not implement... */
else if (oid == ISSUE_DIST_POINT_OID || oid == DELTA_CRL_INDICATOR_OID) {
    if (critical) {
        WOLFSSL_MSG("Critical CRL scoping extension not supported");
        ret = ASN_CRIT_EXT_E;
    }
}
else {
    /* ...unknown-OID callback dispatch as written... */
}
```

Frauschi (commented):
## 🐺 Skoll Code Review

**Findings:** 25 total — 6 posted, 19 skipped

### Posted findings
- [Medium] **Accept-all CRL callback also silently bypasses recognized-but-unimplemented critical extensions (IDP, deltaCRLIndicator)** — `wolfcrypt/src/asn.c:37323-37364`
- [Medium] **New CRL unknown-extension callback is a silent no-op on every OpenSSL-compat CRL load path** — `src/crl.c:900-907`
- [Medium] **Zero-length entry-extension OID hands the callback an uninitialized 64-byte stack array** — `wolfcrypt/src/asn.c:36814-36879`
- [Medium] **Registering a permissive callback makes CRL parsing STRICTER when DecodeObjectId cannot decode the OID** — `wolfcrypt/src/asn.c:36872-36888`
- [Medium] **Observe-only callback silently disables RFC 5280 §5.3 enforcement for all critical CRL extensions, including Issuing Distribution Point** — `wolfssl/ssl.h:4477-4494`
- [Low] **Registering an accepting CRL callback turns previously-tolerated unknown non-critical entry extensions into hard load failures** — `wolfcrypt/src/asn.c:36858-36890`

- [Medium] Callback error code reaches the application differently depending on where the extension sits
- [Medium] Bare BAD_FUNC_ARG in test assertions instead of WC_NO_ERR_TRACE(BAD_FUNC_ARG)
- [Medium] GetCRLInfo callback wiring is untested
- [Medium] No test that the callback fires for a non-critical unknown CRL extension
- [Medium] New CRL unknown-extension callback is a silent no-op in non-template ASN builds, and the new unit tests fail there
- [Low] Registering a callback turns previously-skipped malformed entry extensions into hard load failures
- [Low] CRL-level dispatch reads the ASNGetData union through buffer.data instead of the buf + localIdx idiom used by the rest of the function
- [Low] Buffer-only callback tests are gated on !NO_FILESYSTEM
- [Low] Callback error codes are discarded by the caller, contradicting the preservation logic and the new ssl.h documentation
- [Low] Callback's negative error code is flattened to ASN_PARSE_E, contradicting the code's own comment and the X.509 callback it claims to mirror
- [Low] New unconditional ASN_PARSE_E when an unknown CRL entry extension's extnValue is malformed
- [Low] Ex callback and its context pointer are published and consumed as two unsynchronized accesses
- [Low] New setters return int while the adjacent unknown-extension setter returns void, and the return contract is undocumented
- [Info] Moved static test arrays keep function-body indentation at file scope
- [Info] New public API functions lack doxygen parameter/return blocks
- [Info] Cert-manager cert-side callback still has no Ex variant, leaving an asymmetric API surface
- [Info] New tests compare against bare BAD_FUNC_ARG instead of WC_NO_ERR_TRACE(BAD_FUNC_ARG)
- [Info] Callback/context pair is read unsynchronized and now fires from the CRL monitor thread
- [Info] Undocumented dispatch semantics: both callbacks chain, and neither can tell a CRL-level from a revoked-entry extension

</details>

src/crl.c
```diff
@@ -899,6 +899,13 @@ int BufferLoadCRL(WOLFSSL_CRL* crl, const byte* buff, long sz, int type,
     }
```
Frauschi, line 900: **🟡 [Medium] New CRL unknown-extension callback is a silent no-op on every OpenSSL-compat CRL load path**
*💡 SUGGEST*

Propagation of the new callback into the `DecodedCRL` is gated on `crl->cm != NULL`. Several public CRL-loading paths parse through a `WOLFSSL_CRL` whose `cm` is NULL, so the setter has no effect there:

- `wolfSSL_d2i_X509_CRL()` (src/x509.c:9480) does `InitCRL(newcrl, NULL)` then `BufferLoadCRL(newcrl, ...)`. This is the terminus for `d2i_X509_CRL`, `d2i_X509_CRL_bio`, `d2i_X509_CRL_fp`, `PEM_read_bio_X509_CRL` and `wolfSSL_X509_load_crl_file` (src/x509.c:9352, 9373).
- `wolfSSL_X509_crl_new()` / `wolfSSL_X509_CRL_new()` (src/crl.c:2417) likewise pass NULL.
- `wolfSSL_X509_STORE_add_crl()` (src/crl.c:1543) only splices an *already-parsed* entry list into `store->cm->crl`; attaching the CRL to a cert manager afterwards cannot retroactively apply the callback.

Concrete caller path: an `OPENSSL_EXTRA`/`OPENSSL_ALL` application calls `wolfSSL_CertManagerSetCRLUnknownExtCallbackEx(cm, cb, ctx)` (returns `WOLFSSL_SUCCESS`), then loads its CRL the OpenSSL way (`PEM_read_bio_X509_CRL` + `X509_STORE_add_crl`, or `X509_STORE_load_locations`). The CRL with a critical unknown extension is still rejected with `ASN_CRIT_EXT_E` inside `d2i`, which returns NULL — the callback never fires and `X509_STORE_add_crl` is never reached. Nothing in the return value or the header comment signals that the registration had no effect. The wolfSSL-native paths (`wolfSSL_CertManagerLoadCRL*`, `ProcessFile`, the CRL monitor via `InitCRL(tmp, crl->cm)`, HAVE_CRL_IO) do work, so behaviour depends on which of two equivalent-looking load APIs the caller picked.

**Recommendation:** Either add an internal `d2i` variant that takes the store/cert manager (with `wolfSSL_X509_load_crl_file` and the `X509_LOOKUP` helpers passing `ctx->store->cm`), or add a `WOLFSSL_X509_CRL`-level setter so a standalone CRL object can carry the callback. At minimum, state the limitation explicitly in the ssl.h doc comment: only CRLs parsed through this cert manager are covered; CRLs parsed with `d2i_X509_CRL`/`PEM_read_bio_X509_CRL` are not.

wolfcrypt/src/asn.c
```diff
@@ -36781,10 +36782,15 @@ static const byte crlReasonOid[] = { 0x55, 0x1d, 0x15 };
...
 static int ParseCRL_EntryExtensions(const byte* buff, word32 idx, word32 maxIdx,
-                                    int* reasonCode)
+                                    int* reasonCode, DecodedCRL* dcrl)
 {
+#ifndef WC_ASN_UNKNOWN_EXT_CB
+    (void)dcrl;
+#endif
     while (idx < maxIdx) {
         int len;
         int oidLen;
```
Frauschi, line 36796: **🟡 [Medium] Zero-length entry-extension OID hands the callback an uninitialized 64-byte stack array**
*💡 SUGGEST*

The hand-rolled entry-extension parser validates the OID only with `GetASNTag()` + `GetLength()` + a bounds check; unlike the template path it never runs `GetASN_ObjectId()` (asn.c:1457), which enforces `length >= 3` and that the last content octet has bit 8 clear. `GetLength_ex()` returns 0 (not a negative error) for a zero-length item (asn.c:2664-2669) and the check at asn.c:36814 is `< 0`, so `oidLen == 0` is accepted.

`DecodeObjectId(buff + oidContent, 0, decOid, &decOidSz)` then never enters its `while (inSz--)` loop, returns 0, and sets `*outSz = 0` without writing a single element of `out`. The newly added dispatch immediately passes that fully uninitialized `word16 decOid[MAX_OID_SZ]` (64 bytes of stack) to the caller's callback with `oidSz == 0`.

Reachable with an attacker-supplied CRL containing the entry extension `30 05 06 00 04 01 00`, parsed *before* signature verification. A third-party callback that reads `oid[0]`/`oid[1]` without first checking `oidSz` (a very natural way to write an OID-arc whitelist) reads uninitialized stack. This path did not exist before the PR; site (B) in `ParseCRL_Extensions` is not affected because `GetASN_ObjectId()` rejects both cases first.

**Recommendation:** Apply the same OID sanity checks the template path uses before calling `DecodeObjectId()` — require `oidLen >= 3` and reject an OID whose last content octet has bit 8 set (or call `GetASN_ObjectId(buff, oidContent, oidLen)` directly). Additionally `XMEMSET(decOid, 0, sizeof(decOid));` so a callback can never be handed uninitialized stack data.

wolfcrypt/src/asn.c
```diff
@@ -36849,11 +36855,46 @@ static int ParseCRL_EntryExtensions(const byte* buff, word32 idx, word32 maxIdx,
...
+                word16 decOid[MAX_OID_SZ];
+                word32 decOidSz = MAX_OID_SZ;
+                word32 valIdx = idx;
+                int    valLen = 0;
+                int    cbRet;
+
+                if (GetOctetString(buff, &valIdx, &valLen, end) < 0) {
+                    return ASN_PARSE_E;
+                }
+                cbRet = DecodeObjectId(buff + oidContent, (word32)oidLen,
```
Frauschi, line 36885: **🟡 [Medium] Registering a permissive callback makes CRL parsing STRICTER when DecodeObjectId cannot decode the OID**
*💡 SUGGEST*

Both new dispatch sites call `DecodeObjectId()` on the extension OID and treat any failure as fatal for the whole CRL, regardless of criticality and without the callback ever being consulted.

Site (A), asn.c:36872-36888: on `cbRet != 0` the callbacks are skipped and `return (cbRet < 0) ? cbRet : ASN_PARSE_E;` aborts the parse. Site (B), asn.c:37334-37358: `ret` stays negative, the callbacks are skipped, `handled = 1` is set *anyway* (so the historical criticality check is bypassed), and the negative `ret` ends the loop.

`DecodeObjectId` (asn.c:7469-7509) fails in two CA-reachable ways: `ASN_OBJECT_ID_E` when any sub-identifier occupies 5+ bytes (`if (cnt == 4) return ASN_OBJECT_ID_E;`) and `BUFFER_E` when the OID has more arcs than `MAX_OID_SZ` (= 32, asn.h:1352). The 5-byte case is not exotic: the standardised UUID arc `2.25.<128-bit uuid>` always encodes a sub-identifier far above 2^28, so no `2.25.x` OID can ever decode.

Regression: before this PR an unknown NON-critical CRL extension with such an OID was silently skipped and the CRL loaded. After it, the same legitimately-signed CRL fails to load for any caller that registered the callback — the exact opposite of the API's stated purpose ("register a handler instead of failing with ASN_CRIT_EXT_E"). Revocation checking then breaks for that issuer.

**Recommendation:** Do not let a `DecodeObjectId()` failure change the outcome for extensions that would previously have been accepted: log it and fall through to the historical criticality check (leave `handled = 0`, reset `ret` to 0) instead of aborting the parse. In site (B), set `handled = 1` only after a callback has actually been invoked and returned 0.

wolfssl/ssl.h
```diff
@@ -4474,6 +4474,24 @@ WOLFSSL_API void wolfSSL_CTX_SetPerformTlsRecordProcessingCb(WOLFSSL_CTX* ctx,
     WOLFSSL_API void wolfSSL_CertManagerSetUnknownExtCallback(
         WOLFSSL_CERT_MANAGER* cm,
         wc_UnknownExtCallback cb);
+#if defined(HAVE_CRL)
```
Frauschi, line 4477: **🟡 [Medium] Observe-only callback silently disables RFC 5280 §5.3 enforcement for all critical CRL extensions, including Issuing Distribution Point**
*💡 SUGGEST*

The callback contract has only two outcomes: return 0 = accept the extension, return non-zero = reject the whole CRL. There is no neutral "I observed this, keep the default behaviour" return. The most natural first use of a discovery hook — register a logger to find out which extensions wolfSSL does not handle — is written as `{ log(oid); return 0; }`, and that single line turns off the `ASN_CRIT_EXT_E` check for *every* unknown critical CRL extension at once (a configuration cliff).

This matters more for CRLs than for certificates because `ParseCRL_Extensions()` only implements `AUTH_KEY_OID` and `CRL_NUMBER_OID`; everything else falls into the new `else` branch — including standard, security-relevant extensions wolfSSL deliberately rejects today. Issuing Distribution Point (2.5.29.28) is the concrete case: `test_wolfSSL_CRL_critical_idp` in tests/api/test_certman.c asserts that a CRL carrying a critical IDP must be rejected. With an accept-everything callback registered, a CRL scoped by a critical IDP (`onlyContainsUserCerts`, `onlyContainsCACerts`, `onlySomeReasons`) is accepted and then applied to certificates outside its declared scope, since wolfSSL never parses the scoping constraint.

**Recommendation:** Define a distinct return value meaning "observed, apply default handling" (e.g. a positive sentinel or a dedicated `WC_UNKNOWN_EXT_DEFAULT` code) so a logging callback cannot accidentally relax criticality enforcement, and spell out in the ssl.h comment that returning 0 accepts *critical* extensions the library does not implement — naming Issuing Distribution Point explicitly as an example of what enforcement is being waived.

wolfcrypt/src/asn.c
```diff
@@ -36849,11 +36855,46 @@ static int ParseCRL_EntryExtensions(const byte* buff, word32 idx, word32 maxIdx,
                 }
             }
         }
-        else if (critical) {
-            /* RFC 5280 Section 5.3: reject CRL with unknown critical
-             * entry extension. */
-            WOLFSSL_MSG("Unknown critical CRL entry extension");
-            return ASN_CRIT_EXT_E;
+        else {
```
Frauschi, line 36858: **🔵 [Low] Registering an accepting CRL callback turns previously-tolerated unknown non-critical entry extensions into hard load failures**
*💡 SUGGEST*

In the new unknown-extension branch, two failures unrelated to the callback's decision now abort the whole CRL parse, but only when a callback happens to be registered:

- `if (GetOctetString(buff, &valIdx, &valLen, end) < 0) return ASN_PARSE_E;` (line 36869)
- `cbRet = DecodeObjectId(...)`, whose non-zero result falls through to `return (cbRet < 0) ? cbRet : ASN_PARSE_E;` (line 36887) without ever calling the user callback.

Without a callback registered, the same extension takes the `!handled && critical` path; for a **non-critical** unknown entry extension that condition is false, control falls to `idx = end;` and the extension is silently skipped — the CRL loads fine.

So a CRL carrying a revoked entry with a non-critical unknown extension whose `extnValue` is not a well-formed OCTET STRING, or whose OID has more components than `MAX_OID_SZ` (making `DecodeObjectId` return `BUFFER_E`), loads successfully today and fails with `ASN_PARSE_E` after the caller registers a callback that would have accepted it. Since the parse happens before signature verification, the input is attacker-influenced, so this is a way to make a previously-usable CRL stop loading purely as a side effect of callback registration.

The `DecodeObjectId` half mirrors existing `DecodeCertExtensions` behaviour for certificates (which at least logs `"DecodeObjectId() failed. Corruption?"`); the `GetOctetString` half is new to this PR and has no certificate analogue.

**Recommendation:** Keep the pre-callback behaviour for extensions that would not have been rejected anyway: on a malformed value or an undecodable OID, fall through to the historical critical/non-critical decision instead of failing outright.

            if (dcrl != NULL && (dcrl->unknownExtCallback != NULL ||
                                 dcrl->unknownExtCallbackEx != NULL)) {
                word32 valIdx = idx;
                int    valLen = 0;
                int    cbRet  = 0;

                if ((GetOctetString(buff, &valIdx, &valLen, end) >= 0) &&
                    (DecodeObjectId(buff + oidContent, (word32)oidLen,
                                    decOid, &decOidSz) == 0)) {
                    if (dcrl->unknownExtCallback != NULL) {
                        cbRet = dcrl->unknownExtCallback(decOid, decOidSz,
                            critical, buff + valIdx, (word32)valLen);
                    }
                    if ((cbRet == 0) && (dcrl->unknownExtCallbackEx != NULL)) {
                        cbRet = dcrl->unknownExtCallbackEx(decOid, decOidSz,
                            critical, buff + valIdx, (word32)valLen,
                            dcrl->unknownExtCallbackExCtx);
                    }
                    if (cbRet != 0) {
                        return (cbRet < 0) ? cbRet : ASN_PARSE_E;
                    }
                    handled = 1;
                }
                else {
                    WOLFSSL_MSG("Malformed unknown CRL entry extension");
                }
            }

If failing closed is intentional, apply it unconditionally (with and without a callback) so behaviour does not depend on registration, and say so in the ssl.h doc comment.

wolfssl/ssl.h
anhu: "With an accept-everything callback registered...."   This reasoning assumes a really silly user. 

Won't Fix.

wolfcrypt/src/asn.c
anhu: "Once any accept-all callback is registered,...." this assumes a silly user.  

Won't fix.

src/crl.c
anhu: This is a silly problem because people using the compat layer didn't have a way to register a callback in the first place. 

Wont fix.

wolfcrypt/src/asn.c
anhu: This is silly.  If the OID can't be decoded, its bad and should fail.  If there is a problem in the OID parser, it should be fixed there.  

Won't fix.

wolfcrypt/src/asn.c
anhu: Taking this suggestion.  Stay tuned for a commit.

wolfcrypt/src/asn.c
anhu: Low.  Wont fix.

wolfcrypt/src/asn.c
```diff
@@ -36813,6 +36819,16 @@ static int ParseCRL_EntryExtensions(const byte* buff, word32 idx, word32 maxIdx,
...
             break;
         }
+        /* Validate the OID encoding itself. GetLength() returns 0 (not an
+         * error) for a zero-length item, and an OID whose last content octet
+         * has bit 8 set encodes no complete sub-identifier. Either case makes
+         * DecodeObjectId() return 0 with *outSz == 0 having written nothing to
+         * the output array, which would hand the unknown-extension callback an
+         * uninitialized OID buffer. The template path gets this check via
+         * GetASN_ObjectId(); this hand-rolled parser must do it explicitly. */
+        if (GetASN_ObjectId(buff, oidContent, oidLen) != 0) {
```
Frauschi: GetASN_ObjectId() is only compiled under WOLFSSL_ASN_TEMPLATE, so this causes a compile-error with the old non-template asn code (also the cause for the PRB-master failure).

Frauschi (approved):
LGTM

dgarske (approved):

PR #10963, opened 2026-07-21
Fix wolfSSL_CTX/SSL_set_cipher_list() silently ignoring TLS 1.3-only cipher lists on a non-TLS1.3 ctx/ssl

miyazakh:
## Summary(f3228)

- `wolfSSL_CTX_set_cipher_list()` / `wolfSSL_set_cipher_list()` returned `WOLFSSL_SUCCESS` without touching the configured suites when given a cipher list that names only TLS 1.3 suites, but the `ctx`/`ssl` cannot negotiate TLS 1.3 (`OPENSSL_EXTRA` builds only, via `wolfSSL_parse_cipher_list()` in `src/ssl.c`). Found via an AI-assisted scan for potential issues.
- A caller that treats the non-failing return as confirmation that the restriction took effect (e.g. "harden this connection to TLS 1.3 AEAD suites only") will silently keep negotiating whatever `<= TLS 1.2` suites were already configured, including weaker RSA-key-exchange or CBC-mode suites.

## Changes

- `src/ssl.c`: `wolfSSL_parse_cipher_list()` now returns `WOLFSSL_FAILURE` unconditionally in this case, instead of silently ignoring the list.
- `src/internal.c`: no doc changes needed (no new macro).
- `tests/api/test_tls13.c`: `test_tls13_cipher_list_no_tls13_ctx` updated to assert `WOLFSSL_FAILURE` and that the previously configured suites are left unchanged, for both the `ctx` and `ssl` code paths.

## Testing

- [x] `./configure --enable-opensslextra --enable-tls13 --enable-debug && make` — builds clean, no new warnings.
- [x] `./tests/unit.test -1523` (`test_tls13_cipher_list_no_tls13_ctx`) — passed.
- [x] Full cipher-list/ciphersuites test set (`test_wolfSSL_CTX_set_cipher_list_bytes`, `test_wolfSSL_set_cipher_list_tls12_keeps_tls13`, `test_wolfSSL_set_cipher_list_tls13_keeps_tls12`, `test_wolfSSL_set_cipher_list_tls12_with_version`, `test_wolfSSL_set_cipher_list_tls13_with_version`, `test_wolfSSL_CTX_set_ciphersuites`) — all passed, confirming existing TLS 1.3-capable (`wolfSSLv23_*_method()`) call sites are unaffected.
- [x] Manual check: `wolfSSL_CTX_set_cipher_list(ctx, "TLS13-AES128-GCM-SHA256")` on a `wolfTLSv1_2_client_method()` ctx now returns `WOLFSSL_FAILURE`; a `<= TLS 1.2` suite on the same ctx still returns `WOLFSSL_SUCCESS` as before.

# Checklist

 - [x] added tests
 - [ ] updated/added doxygen


src/ssl.c
```diff
@@ -3895,8 +3895,17 @@ static int wolfSSL_parse_cipher_list(WOLFSSL_CTX* ctx, WOLFSSL* ssl,
...
         if ((ctx != NULL && !IsAtLeastTLSv1_3(ctx->method->version)) ||
                 (ssl != NULL && !IsAtLeastTLSv1_3(ssl->version))) {
+#ifdef WOLFSSL_STRICT_CIPHER_LIST
+            /* TLS 1.3 is not negotiable on this ctx/ssl: fail instead of
+             * leaving the previously configured <= TLS 1.2 suites
+             * untouched while reporting success. */
+            WOLFSSL_MSG("Cipher list has only TLS 1.3 suites but TLS 1.3 "
+                        "is not negotiable");
+            return WOLFSSL_FAILURE;
+#else
```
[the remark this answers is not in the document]
miyazakh: **Decision:** Not addressed.

**Reason:** The other failure paths in `wolfSSL_parse_cipher_list()` (NULL argument, `XMALLOC` failure, `SetCipherList_ex` failure) also skip `WOLFSSL_ERROR()` and rely on `WOLFSSL_MSG` only, so adding it just here would be inconsistent. Also, `wolfSSL_get_error()` reads `ssl->error` and needs a non-NULL `ssl`, so it can't observe this failure via `wolfSSL_CTX_set_cipher_list()` (`ssl == NULL`) regardless — the relevant API would be `wolfSSL_ERR_get_error()`, which this function doesn't use anywhere. A uniform fix across all failure paths would be a separate change.

miyazakh:
retest this please

tests/api/test_tls13.c
```diff
@@ -2463,6 +2463,31 @@ int test_tls13_cipher_suites(void)
...
+#ifdef WOLFSSL_STRICT_CIPHER_LIST
+    ExpectIntEQ(wolfSSL_CTX_set_cipher_list(ctx, "TLS13-AES128-GCM-SHA256"),
+        WOLFSSL_FAILURE);
+#else
+    /* Default OpenSSL-compat behavior: succeed but leave ctx->suites
+     * untouched since the TLS 1.3-only list is unusable on this ctx. */
+    ExpectIntEQ(wolfSSL_CTX_set_cipher_list(ctx, "TLS13-AES128-GCM-SHA256"),
+        WOLFSSL_SUCCESS);
+#endif
+    wolfSSL_CTX_free(ctx);
```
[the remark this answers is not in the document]
miyazakh: Finding is incorrect.

wolfSSL_CTX_set_cipher_list(ctx, ...) is wrapped in ExpectIntEQ, whose body only runs "if (EXPECT_SUCCESS())" (tests/unit.h:208-214). If the earlier ExpectNotNull(ctx = wolfSSL_CTX_new(...)) fails, that guard becomes false, so the call is never evaluated -- no NULL deref possible.

wolfSSL_CTX_free(ctx) does run unconditionally, but it NULL-checks internally (src/ssl.c:677-685), so that's safe too.

No fix needed -- this is the standard wolfSSL test-harness short-circuit
pattern.

miyazakh:
retest this please

miyazakh:
retest this please

miyazakh:
retest this please

tests/api/test_tls13.c
```diff
@@ -2463,6 +2463,31 @@ int test_tls13_cipher_suites(void)
     return EXPECT_RESULT();
 }
 
+int test_tls13_cipher_list_no_tls13_ctx(void)
```
Frauschi, line 2497: **🟡 [Medium] Test only exercises the CTX path, not the SSL path**
*💡 SUGGEST*  `test`

The new macro gates the return value inside `wolfSSL_parse_cipher_list()`, which backs BOTH `wolfSSL_CTX_set_cipher_list()` (via ctx) and `wolfSSL_set_cipher_list()` (via ssl, `src/ssl.c:4041`). The PR body states the change and test cover both entry points, but the test only calls `wolfSSL_CTX_set_cipher_list()`. The `ssl->version` branch of the condition (`ssl != NULL && !IsAtLeastTLSv1_3(ssl->version)`) is never exercised, so a regression that broke only the SSL path would pass CI.

**Suggestion:** Add a parallel case that creates a WOLFSSL* from a TLS 1.2 ctx and calls wolfSSL_set_cipher_list(ssl, "TLS13-AES128-GCM-SHA256"), asserting WOLFSSL_FAILURE (strict) / WOLFSSL_SUCCESS (default), to cover the ssl->version branch.

tests/api/test_tls13.c
```diff
@@ -2463,6 +2463,31 @@ int test_tls13_cipher_suites(void)
...
+    defined(BUILD_TLS_AES_128_GCM_SHA256)
+    WOLFSSL_CTX* ctx = NULL;
+
+    /* ctx->method caps the connection at TLS 1.2, so a cipher list that
+     * names only TLS 1.3 suites can never take effect on it. */
+    ExpectNotNull(ctx = wolfSSL_CTX_new(wolfTLSv1_2_client_method()));
+#ifdef WOLFSSL_STRICT_CIPHER_LIST
+    ExpectIntEQ(wolfSSL_CTX_set_cipher_list(ctx, "TLS13-AES128-GCM-SHA256"),
+        WOLFSSL_FAILURE);
+#else
```
Frauschi: **🔵 [Low] Test asserts return value but not that suites are actually left untouched**
*💡 SUGGEST*  `test`

The default-path comment claims the call succeeds "but leave ctx->suites untouched," yet the test only checks the WOLFSSL_SUCCESS return code. The distinguishing behavior (that the prior <=TLS1.2 suites remain and the TLS1.3-only list did NOT overwrite them) is not asserted. This is the exact silent-no-op behavior the PR is calling out as a footgun, so it would be valuable to pin it down.

**Suggestion:** Capture ctx->suites->suiteSz (or the suite bytes) before the call and assert they are unchanged after, confirming the TLS1.3-only list was actually ignored rather than merely returning success.

src/ssl.c
```diff
@@ -3895,8 +3895,17 @@ static int wolfSSL_parse_cipher_list(WOLFSSL_CTX* ctx, WOLFSSL* ssl,
         tls13Only = 1;
```
Frauschi, line 3134: **🔵 [Low] WOLFSSL_STRICT_CIPHER_LIST can now fail ctx creation under WOLFSSL_SYS_CRYPTO_POLICY**
*💡 SUGGEST*  `question`

Beyond the two public setters named in the PR, `wolfSSL_parse_cipher_list()` is also called during ctx construction when `WOLFSSL_SYS_CRYPTO_POLICY` is enabled. With `WOLFSSL_STRICT_CIPHER_LIST` defined, a system crypto policy that lists only TLS 1.3 suites applied to a <=TLS1.2 method would now make `wolfSSL_parse_cipher_list()` return WOLFSSL_FAILURE, causing `wolfSSL_CTX_new_ex` to free the ctx and return NULL where it previously succeeded. This is a plausibly-intended consequence of opting into strict behavior (fail loudly), but it is a broader blast radius than the two setters described in the PR body and is not covered by tests. Worth confirming this interaction is intended and documenting it.

**Recommendation:** Confirm the WOLFSSL_SYS_CRYPTO_POLICY + WOLFSSL_STRICT_CIPHER_LIST interaction (ctx creation can now fail) is intended, and note it in the macro documentation.

Frauschi (commented):
## 🐺 Skoll Code Review

### Posted findings
- [Medium] **Test only exercises the CTX path, not the SSL path** — `tests/api/test_tls13.c:2466-2489`
- [Low] **Test asserts return value but not that suites are actually left untouched** — `tests/api/test_tls13.c:2480-2484`
- [Low] **WOLFSSL_STRICT_CIPHER_LIST can now fail ctx creation under WOLFSSL_SYS_CRYPTO_POLICY** — `src/ssl.c:636-641`

tests/api/test_tls13.c
miyazakh: Added a test case

tests/api/test_tls13.c
miyazakh: Added a test case

src/ssl.c
miyazakh: Add the macro documentation

miyazakh:
Jenkins retest this please.

src/ssl.c
```diff
@@ -3895,8 +3895,17 @@ static int wolfSSL_parse_cipher_list(WOLFSSL_CTX* ctx, WOLFSSL* ssl,
...
         if ((ctx != NULL && !IsAtLeastTLSv1_3(ctx->method->version)) ||
                 (ssl != NULL && !IsAtLeastTLSv1_3(ssl->version))) {
+#ifdef WOLFSSL_STRICT_CIPHER_LIST
```
Frauschi: **🟡 [Medium] WOLFSSL_STRICT_CIPHER_LIST also breaks SSL_CTX_set_ciphersuites() OpenSSL semantics**
*💡 SUGGEST*  `api`

`wolfssl/openssl/ssl.h:395` defines `SSL_CTX_set_ciphersuites` as `wolfSSL_CTX_set_cipher_list`, so the new strict return path applies to *both* compat APIs, not just `set_cipher_list()`. The comment three lines above the new code (`src/ssl.c:3893`) even says this branch exists to "simulate set_ciphersuites() compatibility layer API". In OpenSSL, `SSL_CTX_set_ciphersuites()` is deliberately version-independent: it configures the separate TLS 1.3 suite list and returns 1 for any syntactically valid suite name regardless of the ctx's max protocol version. Under a strict build the very common OpenSSL idiom now fails:

```c
ctx = SSL_CTX_new(TLS_client_method());
SSL_CTX_set_max_proto_version(ctx, TLS1_2_VERSION);   /* lowers ctx->method->version.minor, see ssl.c:7463 */
SSL_CTX_set_ciphersuites(ctx, "TLS13-AES128-GCM-SHA256"); /* OpenSSL: 1, strict wolfSSL: 0 */
```

`Set_CTX_max_proto_version()` writes `ctx->method->version.minor = TLSv1_2_MINOR` (`src/ssl.c:7463-7465`), so `IsAtLeastTLSv1_3(ctx->method->version)` goes false and the new `WOLFSSL_FAILURE` triggers — even though the caller used the API whose whole purpose is setting 1.3 suites. Because both APIs funnel through the same function, the strict flag cannot distinguish "user called set_cipher_list" from "user called set_ciphersuites", so opting in unavoidably breaks the OpenSSL contract for the second one. It is opt-in and off by default, so nothing regresses for existing builds — but the doc block in `src/internal.c` only mentions `set_cipher_list()`, which understates the blast radius.

**Suggestion:**
```suggestion
* WOLFSSL_STRICT_CIPHER_LIST: Fail wolfSSL_CTX/SSL_set_cipher_list()
 *                  when given a TLS 1.3-only suite list but TLS 1.3
 *                  is not negotiable, instead of leaving the prior
 *                  suites untouched and returning success. NOTE: since
 *                  SSL_CTX_set_ciphersuites() maps to
 *                  wolfSSL_CTX_set_cipher_list(), this also makes that
 *                  API fail, which diverges from OpenSSL (there
 *                  set_ciphersuites() succeeds regardless of the
 *                  configured max protocol version). ...
```

src/ssl.c
Frauschi, line 3134: **🟡 [Medium] WOLFSSL_STRICT_CIPHER_LIST guarantee is narrower than documented: version-cap via SSL_OP_NO_TLSv1_3 / set_max_proto_version is not detected**
*💡 SUGGEST*

The new opt-in macro is documented (both in `src/internal.c` and in the added code comment) as failing `wolfSSL_CTX_set_cipher_list()` / `wolfSSL_set_cipher_list()` when "TLS 1.3 is not negotiable". The guard it hangs off, however, only inspects `ctx->method->version` and `ssl->version`. Those fields are NOT updated by two common OpenSSL-compat ways of capping a connection at TLS 1.2:

- `wolfSSL_CTX_set_options(ctx, WOLFSSL_OP_NO_TLSv1_3)` only sets `ctx->mask` (`src/ssl.c:6823`); `ctx->method->version` stays at TLSv1_3.
- `wolfSSL_set_max_proto_version(ssl, TLS1_2_VERSION)` -> `Set_SSL_max_proto_version()` only sets `ssl->options.mask` (`src/ssl.c:7659`) and leaves `ssl->version` at TLSv1_3. (`wolfSSL_CTX_set_max_proto_version()` *does* lower `ctx->method->version.minor`, so that ctx-level path is covered by the guard; these two are not.)

So with the macro enabled, `SSL_CTX_set_options(ctx, WOLFSSL_OP_NO_TLSv1_3); SSL_CTX_set_ciphersuites(ctx, "TLS13-AES128-GCM-SHA256")` passes the guard and `SetCipherList_ex()` installs the TLS 1.3-only suite list with `suites->setSuites = 1`. Measured against this PR head in a `-DWOLFSSL_STRICT_CIPHER_LIST` build (`--enable-opensslextra --enable-tls13 --enable-debug`): the call returns `WOLFSSL_SUCCESS` and leaves `ctx->suites->suiteSz == 2`, `setSuites == 1`, holding the single suite `0x1301` (TLS_AES_128_GCM_SHA256). Because TLS 1.3 is masked off at handshake time (`src/internal.c:7089`), the ctx now advertises only a suite it can never negotiate - a worse outcome than the silent-ignore the macro was added to eliminate, and exactly the misuse the caller enabled the macro to be protected from. A user who opts in reasonably concludes "a non-negotiable TLS 1.3-only list now returns WOLFSSL_FAILURE"; that conclusion holds only for the method-version path the added unit test exercises.

**Recommendation:** Extend the strict-mode predicate to consider the option mask as well as the method/ssl version, e.g. also treat `(ctx != NULL && (ctx->mask & WOLFSSL_OP_NO_TLSv1_3))` and `(ssl != NULL && (ssl->options.mask & WOLFSSL_OP_NO_TLSv1_3))` as "TLS 1.3 not negotiable". If that is deliberately out of scope, narrow the wording in `src/internal.c` and in the code comment to say the check is on the ctx/ssl method version only, so opt-in users do not over-trust the guarantee, and add a test for the `SSL_OP_NO_TLSv1_3` case.

tests/api/test_tls13.c
```diff
@@ -2463,6 +2463,68 @@ int test_tls13_cipher_suites(void)
...
+    WOLFSSL_CTX* ctx = NULL;
+    WOLFSSL* ssl = NULL;
+#ifndef WOLFSSL_STRICT_CIPHER_LIST
+    word16 suiteSzBefore;
+#endif
+
+    /* ctx->method caps the connection at TLS 1.2, so a cipher list that
+     * names only TLS 1.3 suites can never take effect on it. */
+    ExpectNotNull(ctx = wolfSSL_CTX_new(wolfTLSv1_2_client_method()));
+#ifdef WOLFSSL_STRICT_CIPHER_LIST
```
Frauschi: **🔵 [Low] Strict branch of the new test is never compiled by any CI configuration**
*🔧 NIT*  `test`

The `#ifdef WOLFSSL_STRICT_CIPHER_LIST` half of the test only builds when a developer passes `CPPFLAGS=-DWOLFSSL_STRICT_CIPHER_LIST` by hand (as the PR description shows). No in-tree build config defines the macro, so the branch that the PR actually adds to `src/ssl.c` is dead code in every CI job and will silently rot. Several CI workflows already build with ad-hoc macro lists (e.g. `.github/workflows/os-check.yml` uses `CPPFLAGS="-DWOLFSSL_..."`), so there is an established place to add coverage.

**Suggestion:** Add `-DWOLFSSL_STRICT_CIPHER_LIST` to one existing opensslextra+tls13 CI matrix entry (e.g. in .github/workflows/os-check.yml) so the strict path is built and the new assertions actually run.

tests/api/test_tls13.c
```diff
@@ -2463,6 +2463,68 @@ int test_tls13_cipher_suites(void)
     return EXPECT_RESULT();
 }
 
+int test_tls13_cipher_list_no_tls13_ctx(void)
```
Frauschi, line 2497: **🔵 [Low] New regression test's "suites unchanged" assertions are vacuous (0 == 0) and never verify preservation of configured suites**
*💡 SUGGEST*

The non-strict branch claims to "Confirm the suite count is actually unchanged, not just that the call reported success", but no suites are ever configured before the measurement. `ctx->suites` is NULL right after `wolfSSL_CTX_new()`, so `suiteSzBefore` is 0; `wolfSSL_CTX_set_cipher_list()` then calls `AllocateCtxSuites()`, which `XMEMSET`s the new `Suites` to zero (`src/internal.c:3651-3663`), so the post-call `suiteSz` is also 0. The same holds for the ssl case: `AllocateSuites()` copies the (empty) ctx suites (`src/internal.c:3666-3681`). Both `ExpectIntEQ(..., suiteSzBefore)` checks therefore compare 0 with 0 and would still pass if the permissive path wiped a previously configured suite list. Symmetrically, the strict branch asserts only the `WOLFSSL_FAILURE` return and never checks that the prior suites survived the rejected call, which is the actual justification given for the new macro.

**Recommendation:** Configure a real <= TLS 1.2 list first (e.g. `wolfSSL_CTX_set_cipher_list(ctx, "ECDHE-RSA-AES128-GCM-SHA256")`), assert `suiteSz > 0`, snapshot the suite bytes, then issue the TLS 1.3-only call and compare both `suiteSz` and the suite bytes. Apply the same snapshot/compare in the `WOLFSSL_STRICT_CIPHER_LIST` branch so the "fail without mutating state" contract is actually tested.

Frauschi (commented):
## 🐺 Skoll Code Review

**Findings:** 10 total — 4 posted, 6 skipped

### Posted findings
- [Medium] **WOLFSSL_STRICT_CIPHER_LIST also breaks SSL_CTX_set_ciphersuites() OpenSSL semantics** — `src/ssl.c:3898-3905`
- [Medium] **WOLFSSL_STRICT_CIPHER_LIST guarantee is narrower than documented: version-cap via SSL_OP_NO_TLSv1_3 / set_max_proto_version is not detected** — `src/ssl.c:3891-3910`
- [Low] **Strict branch of the new test is never compiled by any CI configuration** — `tests/api/test_tls13.c:2481-2483`
- [Low] **New regression test's "suites unchanged" assertions are vacuous (0 == 0) and never verify preservation of configured suites** — `tests/api/test_tls13.c:2466-2523`

- [Medium] Default-path test asserts 0 == 0; it does not prove prior suites are preserved
- [Medium] New macro documented in src/internal.c although it is only used in src/ssl.c
- [Low] Public API return contract changes but doxygen for set_cipher_list is not updated
- [Low] Strict mode is not applied to the *_set_cipher_list_bytes() variants
- [Low] Enabling WOLFSSL_STRICT_CIPHER_LIST turns wolfSSL_CTX_new() into a failing call under a TLS 1.3-only system crypto policy
- [Info] Strict-mode rejection is indistinguishable from other set_cipher_list failures at the API boundary

</details>

src/ssl.c
miyazakh: `SSL_CTX/SSL_set_ciphersuites()` alias to `set_cipher_list()`, so strict-mode failures apply to them too, but the doc only mentioned `set_cipher_list()`. Added a note in `src/internal.c` (~line 44) that `set_ciphersuites()` is also affected, including via `SSL_CTX_set_max_proto_version()`.

src/ssl.c
miyazakh: The strict-mode guard only checks `ctx->method->version` / `ssl->version`, not the `WOLFSSL_OP_NO_TLSv1_3` option mask (`SSL_CTX_set_options()` / `wolfSSL_set_max_proto_version()`). Narrowed the doc comments to say so.

- `src/internal.c` (~line 44): macro doc block now notes the check is version-only and doesn't catch `WOLFSSL_OP_NO_TLSv1_3`.
- `src/ssl.c` (~line 3898): inline comment reworded to match.

Both are comment-only changes, no logic change.

Follow-up not done: extending the guard to also check the option mask, and adding a test for the `SSL_OP_NO_TLSv1_3` case.

tests/api/test_tls13.c
miyazakh: No CI workflow defined `WOLFSSL_STRICT_CIPHER_LIST`, so the `#ifdef WOLFSSL_STRICT_CIPHER_LIST` branch of the new test in `tests/api/test_tls13.c` was dead code in every CI job — only the `#else` (silent-ignore) path ever ran. Piggybacked on the existing `opensslextra-no-ca-names` matrix entry instead of adding a new job, to avoid extra CI time.

- `.github/workflows/os-check.yml` (~line 215-217): appended `-DWOLFSSL_STRICT_CIPHER_LIST` to that entry's existing `CPPFLAGS`.

tests/api/test_tls13.c
miyazakh: `test_tls13_cipher_list_no_tls13_ctx` measured `suiteSzBefore` right after `wolfSSL_CTX_new()`/`wolfSSL_new()`, before any suites were ever configured, so both the before and after suite counts were 0 — the assertion could never catch a regression that wiped a real, previously configured suite list. The strict branch didn't check state preservation at all.

Fixed in `tests/api/test_tls13.c` (`test_tls13_cipher_list_no_tls13_ctx`):

- Configure a real `<= TLS 1.2` suite (`ECDHE-RSA-AES128-GCM-SHA256`) first, assert `suiteSz > 0`, and snapshot the full `Suites` struct via `XMEMCPY`.
- After the TLS 1.3-only call (both the strict-failure and non-strict-success paths, for both `ctx` and `ssl`), compare the current `Suites` struct against the snapshot with `XMEMCMP` instead of just comparing `suiteSz`.
- Added build guards (`!NO_RSA`, `HAVE_ECC`, `!NO_AES`, `HAVE_AESGCM`, `!NO_SHA256`, `BUILD_TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256`) needed for the new TLS 1.2 suite name, matching the convention used elsewhere in `tests/api/test_tls.c`.

Verified the modified file compiles cleanly (no errors, no unused-variable warnings under `-Wall -Wextra -Wunused-variable`) using this repo's actual `AM_CFLAGS`, confirming the new branch is genuinely active (not compiled out) in this build configuration.

Frauschi (dismissed):

src/internal.c
```diff
@@ -41,6 +41,27 @@
...
+ *                  when given a TLS 1.3-only suite list but TLS 1.3
+ *                  is not negotiable on ctx->method->version /
+ *                  ssl->version, instead of leaving the prior suites
+ *                  untouched and returning success. Also applies to
+ *                  wolfSSL_CTX_new()/_ex() when WOLFSSL_SYS_CRYPTO_POLICY
+ *                  is enabled and the system policy is TLS 1.3-only: ctx
+ *                  creation fails instead of succeeding with the policy
+ *                  silently unapplied. Since SSL_CTX/SSL_set_ciphersuites()
+ *                  are aliased to the same underlying set_cipher_list()
+ *                  implementation, this also makes them fail when the
```
dgarske: **🟠 [Medium] Macro documentation contradicts itself and misstates what SSL_CTX_set_max_proto_version() does**

The new `WOLFSSL_STRICT_CIPHER_LIST` doc entry makes two mutually exclusive claims about the same API. Lines 54-56 say the strict check fires "when the ctx/ssl's method version excludes TLS 1.3 (e.g. after SSL_CTX_set_max_proto_version() lowers it to TLS 1.2)", while lines 58-62 say "TLS 1.3 disabled solely via the WOLFSSL_OP_NO_TLSv1_3 option mask (e.g. SSL_CTX_set_options() or wolfSSL_set_max_proto_version()) is not detected". The first claim is also factually wrong: `wolfSSL_CTX_set_max_proto_version()` -> `Set_CTX_max_proto_version()` (src/ssl.c:7386-7415) only calls `wolfSSL_CTX_set_options(ctx, WOLFSSL_OP_NO_TLSv1_3)`, and `wolfSSL_CTX_set_options()` (src/ssl.c:6820-6844) only ORs into `ctx->mask` -- it never touches `ctx->method->version`. `CheckSslMethodVersion()` receives the major byte by value and cannot mutate it either. So `SSL_CTX_set_max_proto_version(ctx, TLS1_2_VERSION)` leaves `ctx->method->version.minor` at TLSv1_3_MINOR and the new check does *not* fire. A user reading this table to decide whether to enable the macro gets the opposite of the real behavior.

**Fix:** Delete the incorrect `SSL_CTX_set_max_proto_version()` example on lines 55-56 and replace it with a method that really does cap the version (e.g. `wolfTLSv1_2_client_method()`), so the entry no longer contradicts the option-mask sentence that follows.

src/ssl.c
```diff
@@ -3895,8 +3895,21 @@ static int wolfSSL_parse_cipher_list(WOLFSSL_CTX* ctx, WOLFSSL* ssl,
...
+            /* ctx->method->version/ssl->version caps below TLS 1.3: fail
```
dgarske: **🟠 [Medium] Strict check is asymmetric between SSL_set_options() and SSL_CTX_set_options(); the ssl.c comment claims neither is…**

The new comment states the check "only checks the method/ssl version, not WOLFSSL_OP_NO_TLSv1_3 set via wolfSSL_CTX_set_options()/wolfSSL_set_max_proto_version()", and src/internal.c:58-62 generalizes this to "the WOLFSSL_OP_NO_TLSv1_3 option mask ... is not detected". That generalization is wrong at the SSL level: `wolfSSL_set_options()` (src/ssl.c:10482-10486) does `if (ssl->version.minor == TLSv1_3_MINOR) ssl->version.minor = TLSv1_2_MINOR;` when the NO_TLSv1_3 bit is set. So the feature behaves inconsistently for the two halves of the same OpenSSL idiom: `SSL_set_options(ssl, SSL_OP_NO_TLSv1_3)` followed by `SSL_set_ciphersuites(ssl, "TLS13-AES128-GCM-SHA256")` now returns WOLFSSL_FAILURE (ssl->version was lowered), while `SSL_CTX_set_options(ctx, SSL_OP_NO_TLSv1_3)` followed by `SSL_CTX_set_ciphersuites(ctx, "TLS13-AES128-GCM-SHA256")` still returns WOLFSSL_SUCCESS with the suites installed but never negotiable -- which is exactly the silent no-op class the macro is meant to surface. Note `wolfSSL_set_max_proto_version()` -> `Set_SSL_max_proto_version()` (src/ssl.c:7638+) sets `ssl->options.mask` directly and bypasses the version lowering, so only that one API matches the comment.

**Fix:** Either extend the strict check to also consult `ctx->mask` / `ssl->options.mask` for WOLFSSL_OP_NO_TLSv1_3 so CTX and SSL behave the same, or, if the narrow scope is intentional, correct both comments to say that `wolfSSL_set_options()` IS detected (because it lowers ssl->version.minor) while `wolfSSL_CTX_set_options()` and `wolfSSL_set_max_proto_version()` are not.

tests/api/test_tls13.c
```diff
@@ -2463,6 +2463,110 @@ int test_tls13_cipher_suites(void)
     return EXPECT_RESULT();
 }
 
+#if defined(WOLFSSL_TLS13) && defined(OPENSSL_EXTRA) && \
```
dgarske, line 2479: **🔵 [Low] Long feature-guard condition duplicated verbatim and partly redundant**

The same 5-line `#if defined(...)` condition is repeated verbatim for the helper and for the test body, and any future edit has to be applied to both. Several terms are also implied by others already in the list: `defined(BUILD_TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256)` cannot be true unless `!NO_RSA`, `HAVE_ECC`, `!NO_AES`, `HAVE_AESGCM` and `!NO_SHA256` all hold (see the BUILD_* derivation in wolfssl/internal.h), so those five terms are dead weight that make the guard harder to read and to keep in sync.

**Fix:** Define a single local feature macro once and use `#ifdef` on it for both the helper and the test body; drop the terms already implied by the two `BUILD_*` suite macros.

src/internal.c
```diff
@@ -41,6 +41,27 @@
  *                  Enable SHA-2 cipher suites for pre-TLS 1.2         default: off
  * WOLFSSL_NO_STRICT_CIPHER_SUITE:
  *                  Relax strict cipher suite validation               default: off
+ * WOLFSSL_STRICT_CIPHER_LIST: Fail wolfSSL_CTX/SSL_set_cipher_list()
```
dgarske: **🔵 [Low] 21-line entry breaks the terse one-to-three-line format of the build-options table**

Every other entry in this table is one to three lines with the `default: off` column aligned at the right (e.g. `NO_CHAPOL_AEAD:  Disable ChaCha20-Poly1305 AEAD suites   default: off`). The new entry runs 21 lines and buries its `default: off` at the end of a long paragraph, which makes the table considerably harder to scan and increases the chance the alignment drifts on future edits. The detail is valuable but belongs next to the implementation, where the PR already places a shorter version of the same explanation (src/ssl.c:3899-3905).

**Fix:** Trim the table entry to the two- or three-line house format with a pointer to `wolfSSL_parse_cipher_list()`, and keep the full caveats (crypto-policy interaction, set_ciphersuites aliasing, option-mask limits) in the ssl.c implementation comment.

dgarske (commented):
## Skoll Code Review

**Findings:** 12 total — 5 posted, 7 skipped

### Posted findings
- [Medium] **Macro documentation contradicts itself and misstates what SSL_CTX_set_max_proto_version() does** — `src/internal.c:54-62`
- [Medium] **Strict check is asymmetric between SSL_set_options() and SSL_CTX_set_options(); the ssl.c comment claims neither is detected** — `src/ssl.c:3899-3905`
- [Low] **Long feature-guard condition duplicated verbatim and partly redundant** — `tests/api/test_tls13.c:2466-2470,2488-2492`
- [Low] **21-line entry breaks the terse one-to-three-line format of the build-options table** — `src/internal.c:44-64`
- [Info] **New macro name is easily confused with the adjacent, unrelated WOLFSSL_NO_STRICT_CIPHER_SUITE** — `src/internal.c:42-44`

### Skipped findings
- [Medium] ~~WOLFSSL_STRICT_CIPHER_LIST is a silent no-op without OPENSSL_EXTRA, and the doc entry does not say so~~
- [Medium] ~~Documented WOLFSSL_SYS_CRYPTO_POLICY interaction (CTX_new returning NULL) has no test and is not built in CI~~
- [Medium] ~~New test covers only the rejection path; no guard against strict mode over-rejecting~~
- [Medium] ~~New macro piggybacked onto the existing opensslextra-no-ca-names CI config instead of getting its own entry~~
- [Medium] ~~Public API doxygen not updated with the new WOLFSSL_STRICT_CIPHER_LIST failure mode~~
- [Low] ~~suitesBefore is conditionally initialized but unconditionally read; risks -Wmaybe-uninitialized under -Werror~~
- [Low] ~~Three different NULL-guard idioms for the same check inside one test function~~

src/internal.c
```diff
@@ -41,6 +41,27 @@
  *                  Enable SHA-2 cipher suites for pre-TLS 1.2         default: off
  * WOLFSSL_NO_STRICT_CIPHER_SUITE:
```
dgarske, line 42: **⚪ [Info] New macro name is easily confused with the adjacent, unrelated WOLFSSL_NO_STRICT_CIPHER_SUITE**

`WOLFSSL_STRICT_CIPHER_LIST` is placed immediately below `WOLFSSL_NO_STRICT_CIPHER_SUITE` in the "Cipher Suite Selection" table. The two names differ only by the `NO_` prefix and `LIST` vs `SUITE`, yet they govern unrelated things: the existing one relaxes the server-hello suite-echo validation at src/internal.c:33225, while the new one changes a configuration-time API return value. Reading the table, it is natural to assume `WOLFSSL_STRICT_CIPHER_LIST` is simply the inverse of its neighbor.

**Fix:** Consider a name that describes the API behavior rather than reusing "STRICT" (e.g. `WOLFSSL_CIPHER_LIST_NO_SILENT_IGNORE`), or at minimum separate the two entries in the table so they do not read as a pair. Renaming is cheap now and impossible after release, since the macro is public build surface.

dgarske (changes requested):
Please resolve merge conflicts too.
https://github.com/wolfSSL/wolfssl/pull/10963#pullrequestreview-4857422379

src/internal.c
miyazakh: `Set_CTX_max_proto_version()` (ssl.c:7386-7488) does mutate `ctx->method->version`.minor directly at lines 7467-7468 — the finding only quoted lines 7386-7415 and missed this part. Since the strict check (ssl.c:3896) reads ctx->method->version via IsAtLeastTLSv1_3(), `SSL_CTX_set_max_proto_version()` really does trigger it, so the doc's example on lines 55-56 is correct.

The contrast on lines 58-62 refers to a different function: SSL-level `wolfSSL_set_max_proto_version()` → `Set_SSL_max_proto_version()` (ssl.c:7638-7680), which only ORs into ssl->options.mask and never touches `ssl->version`. So the two doc statements describe two different APIs correctly, not a contradiction. Recommend rejecting this finding; no doc change needed.

src/ssl.c
miyazakh: Comments updated in both `src/ssl.c` and `src/internal.c` to clarify the asymmetry: `wolfSSL_set_options(ssl, WOLFSSL_OP_NO_TLSv1_3)` lowers `ssl->version` and is caught by the strict check, whereas `wolfSSL_CTX_set_options()` and `wolfSSL_set_max_proto_version()` (both CTX and SSL variants) only set the option mask and are not detected. No behavior change — documentation only, per your decision to fix the comments rather than extend the check.

tests/api/test_tls13.c
miyazakh: Trimmed both guards to 3 lines, dropping the 4 terms already implied by `BUILD_TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256 `(!NO_RSA, !NO_AES, HAVE_AESGCM, !NO_SHA256). Kept HAVE_ECC since it's not actually implied — the BUILD_ macro's outer condition is `HAVE_ECC || HAVE_CURVE25519 || HAVE_CURVE448`, so it can be satisfied without ECC. Left the two guards as separate inline conditions rather than consolidating into a shared macro, matching this file's existing convention of writing `#if defined(...)` directly at each use site rather than through indirection.

src/internal.c
miyazakh: Trimmed the table entry from 24 lines down to 6, matching the house style (compare `WOLFSSL_HARDEN_TLS` at 6 lines), with default: off right-aligned to match neighboring entries. The full caveats (crypto-policy interaction, `set_ciphersuites` aliasing, option-mask detection limits) stay in the ssl.c:3898-3909 implementation comment; the table entry now just points to `wolfSSL_parse_cipher_list()` for details.

src/ssl.c
```diff
@@ -3119,8 +3119,26 @@ static int wolfSSL_parse_cipher_list(WOLFSSL_CTX* ctx, WOLFSSL* ssl,
         tls13Only = 1;
         if ((ctx != NULL && !IsAtLeastTLSv1_3(ctx->method->version)) ||
                 (ssl != NULL && !IsAtLeastTLSv1_3(ssl->version))) {
+#ifdef WOLFSSL_CIPHER_LIST_NO_SILENT_IGNORE
```
dgarske: Do we need a new option for this? Or should this be on by default and optionally disable? If this is for openssl compatibility then perhaps OPENSSL_EXTRA should set it or drive it? I just don't like another obscure macro for a tiny change like this. I suspect many customers are not checking the return code on set_cipher_list. If its for compatibility with an OSP then perhaps the macro should indicate that? In fact I am wondering if this change is really needed?

dgarske (changes requested):

src/ssl.c
dgarske: Also PR description is not updated to match the code. Please update PR description.

miyazakh: Thinking back to the original issue, I agree with your point.

I dropped the macro entirely.

Background: the original issue was that `set_cipher_list`() returns `WOLFSSL_SUCCESS `while silently discarding a TLS-1.3-only suite list on a ctx/ssl that can't negotiate TLS 1.3, leaving the prior (possibly weaker, <=TLS1.2) suites active. A caller treating that success as confirmation that hardening was applied would be wrong, and as you noted, most callers don't check.

Given that, gating it behind an opt-in macro didn't really solve the problem it was found for — it only protects the (approximately zero) callers who know to enable it. So instead of a new build flag, the fix is now unconditional: `set_cipher_list`() returns `WOLFSSL_FAILURE `for this case.

No macro, no `OPENSSL_EXTRA `tie-in, no doc/CI wiring. It's a small, narrowly-scoped behavior change (a TLS-1.3-only list on a non-TLS-1.3-capable ctx/ssl), and I'll flag it as such in the PR description.

miyazakh:
retest this please

dgarske:
Jenkins retest this please "Make dist check" remote timeout

dgarske (approved):

philljj (approved):

philljj:
Note: the `Build wolfSSL Rust Wrapper (ubuntu-24.04-arm, --enable-all)` succeeded on the 2nd attempt (timeout on 1st attempt), but it doesn't show in the PR checklist.

PR #10964, opened 2026-07-21
TFM bounds checks for safety

SparkiDev:
# Description

fp_to_unsigned_bin_len did not deal with negative input length - fails early now. No code in wolfSSL calls mp_to_unsigned_bin_len with a negative len.

fp_is_bit_set and fp_set_bit check for greater than or equal to FP_MAX_BITS instead of just greater. Check doesn't cause issues as FP_MAX_SIZE has padding but it was the wrong check.

# Testing

Tested with --enable-fastmath.

### DIFF
--- a/wolfcrypt/src/tfm.c
+++ b/wolfcrypt/src/tfm.c
@@ -3969,6 +3969,10 @@ int fp_to_unsigned_bin_len(fp_int *a, unsigned char *b, int c)
   int j = 0;
   int x;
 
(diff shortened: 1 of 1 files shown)


SparkiDev:
Jenkins: retest this please

Visual Studio: remote connection dropped

Frauschi (approved):
LGTM

dgarske (approved):

PR #10965, opened 2026-07-22
Fix multiple reported issues

kareem-wolfssl:
# Description

Fixes zd#22133

# Testing

Built in tests, provided reproducers

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -8271,13 +8271,32 @@ int InitSSL(WOLFSSL* ssl, WOLFSSL_CTX* ctx, int writeDup)
         }
 #endif
 
-        if (ctx->suites == NULL) {
-            /* suites */
-            ret = AllocateCtxSuites(ctx);
-            if (ret != 0)
-                return ret;
-            InitSSL_CTX_Suites(ctx);
+        /* Lazily allocating and deriving ctx->suites mutates shared CTX state.
+         * A CTX is commonly configured once and then shared by many threads
+         * that each call wolfSSL_new(), so guard the check and initialization
+         * with the CTX mutex. Without it, two threads can both observe
+         * ctx->suites == NULL and both allocate, leaking the Suites that loses
+         * the race. */
+        ret = wolfSSL_RefWithMutexLock(&ctx->ref);
+        if (ret != 0) {
+            WOLFSSL_MSG("Failed to lock CTX mutex for suites init");
+            return ret;
         }
+        if (ctx->suites == NULL) {
+            ret = AllocateCtxSuites(ctx);
+            if (ret == 0)
+                InitSSL_CTX_Suites(ctx);
+        }
+        if (wolfSSL_RefWithMutexUnlock(&ctx->ref) != 0) {
+            WOLFSSL_MSG("Failed to unlock CTX mutex after suites init");
+            /* A stuck lock would deadlock the up_ref in SetSSL_CTX() below, so
+             * surface it as an error. Keep any earlier suites-init error, as
+             * that is the more meaningful failure. */
+            if (ret == 0)
+                ret = BAD_MUTEX_E;
+        }
+        if (ret != 0)
+            return ret;
 #ifdef OPENSSL_ALL
         ssl->suitesStack = NULL;
 #endif
--- a/src/internal.c
+++ b/src/internal.c
@@ -8651,6 +8651,12 @@ int AllocKey(WOLFSSL* ssl, int type, void** pKey)
         return MEMORY_E;
     }
 
+    /* Zero the allocation before initializing. XMALLOC does not clear memory,
+     * and the key-specific init below may fail before it has zeroed/initialized
+     * the structure's members.  Starting from all-zero makes any partial-init
+     * failure safe to free. */
+    XMEMSET(*pKey, 0, sz);
+
     /* Initialize key */
     switch (type) {
     #ifndef NO_RSA
--- a/src/internal.c
+++ b/src/internal.c
@@ -3854,6 +3854,44 @@ int AllocateCtxSuites(WOLFSSL_CTX* ctx)
     return 0;
 }
 
+/* Lazily allocate and derive ctx->suites, holding the CTX mutex so the paths
+ * that may share a CTX between threads (wolfSSL_new, wolfSSL_set_SSL_CTX)
+ * cannot both allocate and leak one of the Suites. The CTX setup APIs
+ * (wolfSSL_CTX_set_cipher_list, wolfSSL_CTX_use_certificate, ...) still
+ * allocate without the lock and must not be called concurrently with these
+ * paths on a shared CTX.
+ *
+ * returns 0 on success, otherwise an error code.
+ */
+int InitCtxSuitesWithMutex(WOLFSSL_CTX* ctx)
+{
+    int ret;
+
+    if (ctx == NULL)
+        return BAD_FUNC_ARG;
+
+    ret = wolfSSL_RefWithMutexLock(&ctx->ref);
+    if (ret != 0) {
+        WOLFSSL_MSG("Failed to lock CTX mutex for suites init");
+        return ret;
+    }
+    if (ctx->suites == NULL) {
+        ret = AllocateCtxSuites(ctx);
+        if (ret == 0)
+            InitSSL_CTX_Suites(ctx);
+    }
+    if (wolfSSL_RefWithMutexUnlock(&ctx->ref) != 0) {
+        WOLFSSL_MSG("Failed to unlock CTX mutex after suites init");
+        /* A stuck lock would deadlock every later use of ctx->ref, so surface
+         * it as an error. Keep any earlier suites-init error, as that is the
+         * more meaningful failure. */
+        if (ret == 0)
+            ret = BAD_MUTEX_E;
+    }
+
+    return ret;
+}
+
 /* Call this when the ssl object needs to have its own ssl->suites object */
 int AllocateSuites(WOLFSSL* ssl)
 {
@@ -5325,6 +5363,19 @@ void FreeX509Name(WOLFSSL_X509_NAME* name)
 }
 
 
+/* Zero the X509 and set up its fields. The ref is zeroed too; the caller
+ * must initialize or restore it afterward. */
+static void InitX509Fields(WOLFSSL_X509* x509, int dynamicFlag, void* heap)
+{
+    XMEMSET(x509, 0, sizeof(WOLFSSL_X509));
+
+    x509->heap = heap;
+    InitX509Name(&x509->issuer, 0, heap);
+    InitX509Name(&x509->subject, 0, heap);
+    x509->dynamicMemory  = (byte)dynamicFlag;
+}
+
+
 /* Initialize wolfSSL X509 type */
 void InitX509(WOLFSSL_X509* x509, int dynamicFlag, void* heap)
 {
@@ -5333,12 +5384,7 @@ void InitX509(WOLFSSL_X509* x509, int dynamicFlag, void* heap)
         return;
     }
 
-    XMEMSET(x509, 0, sizeof(WOLFSSL_X509));
-
-    x509->heap = heap;
-    InitX509Name(&x509->issuer, 0, heap);
-    InitX509Name(&x509->subject, 0, heap);
-    x509->dynamicMemory  = (byte)dynamicFlag;
+    InitX509Fields(x509, dynamicFlag, heap);
 #if defined(OPENSSL_EXTRA_X509_SMALL) || defined(OPENSSL_EXTRA)
     {
         int ret;
@@ -5349,8 +5395,8 @@ void InitX509(WOLFSSL_X509* x509, int dynamicFlag, void* heap)
 }
 
 
-/* Free wolfSSL X509 type */
-void FreeX509(WOLFSSL_X509* x509)
+/* Free the contents of an X509, leaving the ref untouched */
+static void FreeX509Contents(WOLFSSL_X509* x509)
 {
     #if defined(WOLFSSL_CERT_REQ) && defined(OPENSSL_ALL) \
     &&  defined( WOLFSSL_CUSTOM_OID)
@@ -5464,10 +5510,46 @@ void FreeX509(WOLFSSL_X509* x509)
         x509->altSigValDer= NULL;
     }
     #endif /* WOLFSSL_DUAL_ALG_CERTS */
+}
 
+
+/* Free wolfSSL X509 type */
+void FreeX509(WOLFSSL_X509* x509)
+{
+    if (x509 == NULL)
+        return;
+
+    FreeX509Contents(x509);
     #if defined(OPENSSL_EXTRA_X509_SMALL) || defined(OPENSSL_EXTRA)
         wolfSSL_RefFree(&x509->ref);
     #endif
+}
+
+
+/* Free an X509's contents and re-initialize it for reuse. Unlike FreeX509()
+ * followed by InitX509(), the object's heap and reference state carry across,
+ * so references held elsewhere stay counted. */
+void ReinitX509(WOLFSSL_X509* x509)
+{
+#if defined(OPENSSL_EXTRA_X509_SMALL) || defined(OPENSSL_EXTRA)
+    wolfSSL_Ref ref;
+#endif
+    void* heap;
+    byte  dynamic;
+
+    if (x509 == NULL)
+        return;
+
+    heap = x509->heap;
+    dynamic = x509->dynamicMemory;
+#if defined(OPENSSL_EXTRA_X509_SMALL) || defined(OPENSSL_EXTRA)
+    ref = x509->ref;
+#endif
+    FreeX509Contents(x509);
+    InitX509Fields(x509, dynamic, heap);
+#if defined(OPENSSL_EXTRA_X509_SMALL) || defined(OPENSSL_EXTRA)
+    x509->ref = ref;
+#endif
 }
 
 
@@ -8700,13 +8782,11 @@ int InitSSL(WOLFSSL* ssl, WOLFSSL_CTX* ctx, int writeDup)
         }
 #endif
 
-        if (ctx->suites == NULL) {
-            /* suites */
-            ret = AllocateCtxSuites(ctx);
-            if (ret != 0)
-                return ret;
-            InitSSL_CTX_Suites(ctx);
-        }
+        /* suites, allocated and derived under the CTX mutex as the CTX may be
+         * shared with other threads */
+        ret = InitCtxSuitesWithMutex(ctx);
+        if (ret != 0)
+            return ret;
 #ifdef OPENSSL_ALL
         ssl->suitesStack = NULL;
 #endif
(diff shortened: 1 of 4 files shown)


src/internal.c
```diff
@@ -8271,13 +8271,32 @@ int InitSSL(WOLFSSL* ssl, WOLFSSL_CTX* ctx, int writeDup)
...
+            if (ret == 0)
+                InitSSL_CTX_Suites(ctx);
         }
+        if (wolfSSL_RefWithMutexUnlock(&ctx->ref) != 0) {
+            WOLFSSL_MSG("Failed to unlock CTX mutex after suites init");
+            /* A stuck lock would deadlock the up_ref in SetSSL_CTX() below, so
+             * surface it as an error. Keep any earlier suites-init error, as
+             * that is the more meaningful failure. */
+            if (ret == 0)
+                ret = BAD_MUTEX_E;
```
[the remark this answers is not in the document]
kareem-wolfssl: Keeping the current logic intentionally, as the mutex failure is logged but suites failure is not.

PhilipTaronQ (approved):
Works for me.

kareem-wolfssl:
Retest this please

src/internal.c
```diff
@@ -8271,13 +8271,32 @@ int InitSSL(WOLFSSL* ssl, WOLFSSL_CTX* ctx, int writeDup)
         }
 #endif
 
+        /* Lazily allocating and deriving ctx->suites mutates shared CTX state.
```
Frauschi: **🟡 [Medium] suites race fix not applied to the parallel lazy-alloc path in wolfSSL_set_SSL_CTX**
*💡 SUGGEST*  `question`

The PR correctly guards the `ctx->suites == NULL` -> `AllocateCtxSuites` -> `InitSSL_CTX_Suites` sequence with the CTX ref mutex to prevent two threads from both allocating (leaking one Suites). The identical unguarded lazy-alloc sequence still exists in `wolfSSL_set_SSL_CTX` (ssl.c:8475-8480), which can run on an SSL after handshake while the CTX is shared. That path is pre-existing and outside the changed files, so it is out of scope for this PR, but it exposes the same double-allocate race the PR is fixing. Flagging so the fix's completeness can be confirmed as intentional.

**Suggestion:** Wrap the ctx->suites lazy-alloc in wolfSSL_set_SSL_CTX with the same wolfSSL_RefWithMutexLock/Unlock(&ctx->ref) guard used in InitSSL, or extract a small helper (e.g. EnsureCtxSuites(ctx)) shared by both call sites.

Frauschi (commented):
## 🐺 Skoll Code Review

### Posted findings
- [Medium] **suites race fix not applied to the parallel lazy-alloc path in wolfSSL_set_SSL_CTX** — `src/internal.c:8274-8299`

- [Medium] No regression test added for the concurrent suites-init and peerCert re-decode fixes
- [Low] Redundant FreeX509/InitX509 in wolfSSL_get_peer_certificate now that DecodeToX509 frees internally

</details>

PhilipTaronQ:
@qcwallace

Frauschi:
Jenkins retest this please

Frauschi (commented):
## 🐺 Skoll Code Review

**Findings:** 6 total — 3 posted, 3 skipped

### Posted findings
- [Medium] **Lazy ctx->suites init in wolfSSL_set_SSL_CTX left unguarded, defeating the mutex added to InitSSL** — `src/ssl.c:8486-8491`
- [Medium] **CTX refcount leaked when suites allocation fails in wolfSSL_set_SSL_CTX** — `src/ssl.c:8475-8491`
- [Low] **Redundant FreeX509/InitX509 now duplicated inside DecodeToX509** — `src/ssl_api_cert.c:2765-2766`

- [Medium] CTX reference leaked when AllocateCtxSuites fails after reorder
- [Medium] Lazy suites init in wolfSSL_set_SSL_CTX left unguarded while InitSSL now locks
- [Low] Suites-init race the PR fixes in InitSSL remains open in wolfSSL_set_SSL_CTX

</details>

src/ssl.c
```diff
@@ -8489,6 +8482,14 @@ WOLFSSL_CTX* wolfSSL_set_SSL_CTX(WOLFSSL* ssl, WOLFSSL_CTX* ctx)
 #else
     (void)ret;
 #endif
+
+    if (ctx->suites == NULL) {
```
Frauschi: **🟡 [Medium] Lazy ctx->suites init in wolfSSL_set_SSL_CTX left unguarded, defeating the mutex added to InitSSL**
*💡 SUGGEST*

This PR fixes a lazy-init race by wrapping the `ctx->suites == NULL` check plus `AllocateCtxSuites()`/`InitSSL_CTX_Suites()` in the CTX ref mutex inside `InitSSL()` (src/internal.c:8280-8299). The stated goal (see the new comment) is to stop two threads from both observing `ctx->suites == NULL` and both allocating, leaking the loser's `Suites`. However the sibling lazy-init in `wolfSSL_set_SSL_CTX()` — which this same PR relocated to just after `wolfSSL_RefWithMutexInc()` — performs the identical check-then-allocate WITHOUT holding `ctx->ref`. `AllocateCtxSuites()` only re-checks `ctx->suites == NULL` with no lock (src/internal.c:3644), so the double-check is not safe. A `WOLFSSL_CTX` is commonly shared: one thread in `wolfSSL_new()`→`InitSSL()` can run concurrently with another thread performing an SNI-driven `wolfSSL_set_SSL_CTX()` onto the same (new) CTX, or two SSL objects can switch to the same CTX concurrently. Because only one of the two paths takes the mutex, both can still see `ctx->suites == NULL`, both `XMALLOC` a `Suites`, and one allocation is overwritten and leaked — plus the concurrent write to `ctx->suites` and the `InitSSL_CTX_Suites()` field population race on shared CTX state. The lock added to `InitSSL` does not serialize against the `set_SSL_CTX` path, so the race the PR intends to close remains open through this second door.

**Recommendation:** Guard the lazy `ctx->suites` init in `wolfSSL_set_SSL_CTX()` with the same CTX mutex used in `InitSSL()`: wrap the `ctx->suites == NULL` check plus `AllocateCtxSuites()`/`InitSSL_CTX_Suites()` in `wolfSSL_RefWithMutexLock(&ctx->ref)` / `wolfSSL_RefWithMutexUnlock(&ctx->ref)`, mirroring the internal.c change (and preserving the existing NULL-return-with-refcount-restore error handling). Better, factor the locked lazy-init into a single helper both call sites use so the two paths cannot drift again.

src/ssl.c
```diff
@@ -8472,13 +8472,6 @@ WOLFSSL_CTX* wolfSSL_set_SSL_CTX(WOLFSSL* ssl, WOLFSSL_CTX* ctx)
...
         return ssl->ctx;
 
-    if (ctx->suites == NULL) {
-        /* suites */
-        if (AllocateCtxSuites(ctx) != 0)
-            return NULL;
-        InitSSL_CTX_Suites(ctx);
-    }
-
     wolfSSL_RefWithMutexInc(&ctx->ref, &ret);
```
Frauschi, line 8721: **🟡 [Medium] CTX refcount leaked when suites allocation fails in wolfSSL_set_SSL_CTX**
*💡 SUGGEST*

The PR moved the lazy suites initialization from BEFORE `wolfSSL_RefWithMutexInc(&ctx->ref, ...)` to AFTER it. `wolfSSL_RefWithMutexInc` increments `ctx->ref`. If `AllocateCtxSuites(ctx)` then fails (memory allocation failure), the function does `return NULL` without ever assigning `ssl->ctx = ctx` and without decrementing `ctx->ref`. The reference count is now permanently inflated by one, so the CTX (and everything it owns) is never freed at final `wolfSSL_CTX_free`. Before this change the suites init ran before the increment, so an allocation failure returned cleanly with no leaked reference. `wolfSSL_set_SSL_CTX` is a public API that returns NULL on error, so the caller has no way to know a reference was taken and cannot compensate. Trigger: shared-CTX SNI callback path where `ctx->suites == NULL` and the `Suites` XMALLOC fails.

**Recommendation:** On the suites-alloc failure path, release the reference taken above before returning, e.g.:
    if (ctx->suites == NULL) {
        if (AllocateCtxSuites(ctx) != 0) {
            wolfSSL_CTX_free(ctx); /* undo the RefWithMutexInc */
            return NULL;
        }
        InitSSL_CTX_Suites(ctx);
    }
Alternatively, keep the suites initialization before `wolfSSL_RefWithMutexInc` as it was prior to this PR.

src/ssl_api_cert.c
```diff
@@ -2755,6 +2760,10 @@ int wolfSSL_get0_chain_certs(WOLFSSL *ssl, WOLF_STACK_OF(WOLFSSL_X509) **sk)
                 ret = wolfSSL_X509_dup(&ssl->peerCert);
 #ifdef SESSION_CERTS
             else if (ssl->session->chain.count > 0) {
+                /* Re-decoding into ssl->peerCert would orphan any contents it
+                 * already holds (e.g. the handshake's copy). Free them first. */
+                FreeX509(&ssl->peerCert);
```
Frauschi: **🔵 [Low] Redundant FreeX509/InitX509 now duplicated inside DecodeToX509**
*🔧 NIT*  `style`

The PR adds `FreeX509(&ssl->peerCert); InitX509(&ssl->peerCert, 0, ssl->heap);` immediately before `DecodeToX509(&ssl->peerCert, ...)`. But the same PR also teaches `DecodeToX509` to do `FreeX509(x509); InitX509(x509, 0, x509->heap);` internally for any non-dynamic x509 (which `ssl->peerCert` is) on successful parse. The result is a double Free+Init of peerCert on the success path. It is harmless (after the first Init the second Free operates on all-NULL members) and defensive, but redundant. This branch is only reached when `ssl->peerCert.issuer.sz == 0` (peerCert unpopulated), so the outer free is essentially a no-op in normal flow.

**Suggestion:**
```suggestion
else if (ssl->session->chain.count > 0) {
    if (DecodeToX509(&ssl->peerCert,
            ssl->session->chain.certs[0].buffer,
            ssl->session->chain.certs[0].length) == 0) {
        ret = wolfSSL_X509_dup(&ssl->peerCert);
    }
}
```

Frauschi:
Jenkins retest this please

src/ssl_api_cert.c
```diff
@@ -2721,8 +2721,13 @@ int wolfSSL_get0_chain_certs(WOLFSSL *ssl, WOLF_STACK_OF(WOLFSSL_X509) **sk)
         InitDecodedCert(cert, (byte*)in, (word32)len, NULL);
         if ((ret = ParseCertRelative(cert, CERT_TYPE, 0, NULL, NULL)) == 0) {
```
Frauschi: **🟡 [Medium] New FreeX509() in DecodeToX509 turns a peerCert leak into real frees from an unsynchronized getter**
*💡 SUGGEST*

The PR adds `FreeX509(x509)` before `InitX509()` on the non-dynamic path of `DecodeToX509()`. The only non-dynamic object that reaches that path is `&ssl->peerCert`: the other two callers pre-set `dynamicMemory=1` (`ssl_api_cert.c:2169` `InitX509(x509, 1, NULL)` and `:2846` `wolfSSL_X509_new_ex()`), so only `wolfSSL_get_peer_certificate()` at `:2758` gets here.

Before this change the path ran `InitX509()` alone, which just memsets the struct, so a racer saw a leaked-but-still-valid heap block. Now the same path performs real `XFREE()`s on `derCert`, `sig.buffer`, `pubKey.buffer`, `altNames` and the `X509_NAME` contents, from an application-callable getter that holds no lock.

The concrete race: a handshake/renegotiation thread runs `DoCert` (`src/internal.c:15402`: `FreeX509(&ssl->peerCert); InitX509(...)`) while an application thread calls `wolfSSL_get_peer_certificate()`. Pre-PR only the handshake side freed; post-PR both sides free the same buffers, and `CopyDecodedToX509()` on one thread writes into a struct the other is freeing.

Scope note, so this is not overstated: the free only runs while `ssl->peerCert.issuer.sz == 0`. After a successful decode `issuer.sz` is non-zero and later calls take the `wolfSSL_X509_dup()` path, so this is a one-shot window rather than something every call hits.

Secondary consequence worth a look: `FreeX509()` ends in `wolfSSL_RefFree(&x509->ref)` (`src/internal.c:5141`) and `InitX509()` re-runs `wolfSSL_RefInit()`, resetting the count to 1. `wolfSSL_get0_verified_chain()` (`src/ssl_api_cert.c:2935`) hands `&ssl->peerCert` to `wolfSSL_X509_STORE_CTX_init()`, which `up_ref`s it (`src/x509_str.c:673`), so a counted reference can be silently reset. In builds without `WOLFSSL_ATOMIC_OPS` (and not `SINGLE_THREADED`) `wolfSSL_RefFree` is `wc_FreeMutex()`, so the refcount mutex is torn down and re-created as well. This teardown pattern is not new in itself, `DoCert` and four other `src/internal.c` sites already do `FreeX509()`+`InitX509()` on `peerCert`, but this is the first time it happens from an app-callable getter that can run concurrently with them.

**Recommendation:** The leak being fixed is real, but freeing from the getter is a heavy way to fix it. Options, roughly in order of preference:

1. Add a `FreeX509Contents()`-style helper that releases the decoded allocations but leaves `x509->ref` (and its mutex) and `x509->heap` untouched, and call that instead of `FreeX509()`+`InitX509()` here.
2. Keep the free in the handshake-owned paths only (`DoCert`) and have `DecodeToX509()` leave an already-populated `ssl->peerCert` alone.
3. If the current shape stands, document that `wolfSSL_get_peer_certificate()` is now a mutating call that must not be issued concurrently with an in-flight handshake or with `wolfSSL_get0_verified_chain()` on the same `WOLFSSL`.

src/internal.c
```diff
@@ -3653,6 +3653,42 @@ int AllocateCtxSuites(WOLFSSL_CTX* ctx)
     return 0;
 }
 
+/* Lazily allocate and derive ctx->suites, holding the CTX mutex so concurrent
```
Frauschi: **🔵 [Low] CTX suites lock covers only two of six lazy-init sites; the documented invariant does not hold**
*💡 SUGGEST*

The new helper's comment states: "All lazy-init call sites must use this helper; one that skips the lock re-opens the race for the others." That invariant does not hold in the tree as it stands. Only two sites were converted (`InitSSL` at `src/internal.c:8312` and `wolfSSL_set_SSL_CTX` at `src/ssl.c:8478`). Four still call a bare `AllocateCtxSuites(ctx)` with no lock held:

- `src/ssl.c:3217` `wolfSSL_CTX_set_cipher_list()`
- `src/ssl.c:3237` `wolfSSL_CTX_set_cipher_list_bytes()`
- `src/ssl_api_ext.c:1463` `wolfSSL_CTX_set1_sigalgs_list()`
- `src/ssl_load.c:2355` `ProcessBufferResetSuites()`, the server-side `CERT_TYPE` reset reached from `wolfSSL_CTX_use_certificate*`

(`src/ssl.c:610` is inside `wolfSSL_CTX_new_ex()` before the CTX is published, so it is fine.)

The last one is the one that matters: it is exactly the cert-rotation-on-a-live-CTX case that the shared-CTX threat model this PR adopts is about. Thread A in `wolfSSL_new()` holds `ctx->ref` and is inside `AllocateCtxSuites`; thread B in `wolfSSL_CTX_use_certificate_chain_file()` runs `AllocateCtxSuites` unlocked. Both observe `ctx->suites == NULL` and one `Suites` allocation is orphaned, which is the original defect. Because the unlocked side never takes the mutex, the lock on the converted side does not exclude it.

Worth noting alongside this: `AllocateCtxSuites()` assigns `ctx->suites` from `XMALLOC` *before* the `XMEMSET`, so the pointer is briefly visible pointing at uninitialized memory, and the callers above then fill the block after it is already published. That is pre-existing rather than introduced here, but it is the reason the unlocked sites are worth converting rather than just documenting.

So the fix closes the `InitSSL` vs `InitSSL`/`set_SSL_CTX` race but not the general one, while the comment tells a future reader the invariant is enforced.

**Recommendation:** Either route the four remaining sites through `InitCtxSuitesWithMutex()` (better: hold the lock across allocation *and* population, since those sites also write into `ctx->suites` afterwards), or, if converting the setup-time APIs is out of scope for this fix, soften the comment so it does not assert something the tree does not do. For example:

    /* ... Used by the run-time paths (wolfSSL_new, wolfSSL_set_SSL_CTX). The
     * CTX setup APIs (wolfSSL_CTX_set_cipher_list, wolfSSL_CTX_use_certificate*,
     * ...) still allocate unlocked and must not be called concurrently with
     * them on a shared CTX. */

Frauschi (commented):
## 🐺 Skoll Code Review

**Findings:** 11 total — 2 posted, 9 skipped

### Posted findings
- [Medium] **New FreeX509() in DecodeToX509 turns a peerCert leak into real frees from an unsynchronized getter** — `src/ssl_api_cert.c:2722-2732`
- [Low] **CTX suites lock covers only two of six lazy-init sites; the documented invariant does not hold** — `src/internal.c:3656-3690`

- [Medium] Suites lazy-init race only partially closed; helper comment overstates coverage
- [Medium] No tests added for any of the three fixes
- [Medium] FreeX509() on ssl->peerCert destroys and re-creates a live reference-count mutex
- [Medium] InitCtxSuitesWithMutex leaves four unsynchronized AllocateCtxSuites call sites, so the ctx->suites race is only half closed
- [Low] New helper name diverges from the existing suites-function naming
- [Low] Refcount mutex now held across the full suite derivation
- [Low] AllocKey memset is redundant with every wc_*_init in-tree
- [Info] Reference-count mutex reused as a general-purpose CTX lock; non-recursive, so any future nested use self-deadlocks
- [Info] AllocKey zeroing duplicates work already done by every key init function

</details>

src/ssl_api_cert.c
```diff
@@ -2865,11 +2865,14 @@ static int DecodeToX509(WOLFSSL_X509* x509, const byte* in, int len)
         /* Create a DecodedCert object and copy fields into WOLFSSL_X509
          * object. */
         InitDecodedCert(cert, (byte*)in, (word32)len, NULL);
-        ret = ParseCertRelative(cert, CERT_TYPE, 0, NULL, NULL);
-        if (ret == 0) {
-            /* Initialize when not done by wolfSSL_X509_new() already. */
+        if ((ret = ParseCertRelative(cert, CERT_TYPE, 0, NULL, NULL)) == 0) {
```
JacobBarthelmeh: Why squash the ret assignment and check of ret into one line? I find it harder to read and it is also a concern sometimes with things like MISRA.

JacobBarthelmeh (changes requested):
Please squash the commits. Code changes look okay, one minor nit comment.

kareem-wolfssl:
Retest this please

kareem-wolfssl:
Retest this please

PhilipTaronQ:
I'm out until Monday; I've pinged my co-workers @qcwallace and @moonrei to see if they have cycles to test this prior to Monday or Tuesday.

JacobBarthelmeh (approved):

JacobBarthelmeh:
A GitHub comment with the phrase "Retest this please" is one we use to trigger an automated Jenkins tests re-run on any failed CI tests. Thanks @PhilipTaronQ for initially reviewing the changes.

PR #10970, opened 2026-07-22
STM32 and RealTek: plaintext-key AES device + hardware AES-GCM/ECDSA via crypto callbacks

dgarske:
## Summary

Hardware AES and ECDSA through the crypto-callback framework for STM32 (CubeMX/HAL and bare-metal) and RealTek RTL8735B, so the code-size `WOLF_CRYPTO_CB_ONLY_AES` / `WOLF_CRYPTO_CB_ONLY_ECC` configs run fully on hardware. Adds a plaintext-key AES device that coexists with the existing HUK/DHUK seed-key device (selectable per `Aes` by devId), full hardware AES-GCM on the bare STM32 and RealTek paths, an explicit-word-order key-wrap API for offline provisioning, and a portable `asn.c` NULL-guard fix.

## Features

- **STM32 CubeMX crypto-callback device**: hardware AES (ECB, and AES-GCM serviced in-callback on the HAL GCM engine) plus HW ECDSA sign + verify via the PKA and CCB-protected sign via `HAL_CCB`. One devId serves AES + ECDSA + CCB, making `WOLF_CRYPTO_CB_ONLY_AES` / `WOLF_CRYPTO_CB_ONLY_ECC` usable on the HAL build (that macro compiles out the direct `ecc.c` PKA path, so the callback has to provide HW ECDSA instead).
- **Plaintext-key AES device, selectable by devId**: a new device (STM32 CubeMX + bare, and RealTek) runs a caller-supplied AES key directly on the HW engine, registered alongside the DHUK/HUK seed-key device. An application picks per `Aes` whether its key bytes are used verbatim (`WOLFSSL_STM32_AES_DEVID`, default 806) or as a hardware-unique-key derivation seed (`WC_DHUK_DEVID`, 808). Same key bytes therefore give a standard AES-GCM result on one device and a device-bound, non-portable result on the other. 128/192/256-bit keys on the plaintext device; the DHUK device remains 256-bit-seed only.
- **Full hardware AES-GCM** on the bare STM32 (TinyAES peripheral: encrypt + decrypt-verify with AAD and partial blocks; CRYP IP: whole-block enc + dec) and RealTek RTL8735B. Bare TinyAES GCM previously fell back to software. AES peripherals with no GCM mode fall back to a software GHASH whose AES blocks still run on hardware. Only the standard 12-byte GCM IV takes the HW path; other IV lengths fall back to software.
- **RealTek RTL8735B plaintext AES device** (`WOLFSSL_RTL8735B_AES`, `WC_RTL8735B_AES_DEVID` default 811) with 128/192/256-bit keys, coexisting with the HUK device. Shapes the HW engine cannot service (non-block-multiple size, zero-length, oversized AAD) are declined with `CRYPTOCB_UNAVAILABLE` so software GCM completes them under the same literal key.
- **`wc_Stm32_Aes_Wrap_ex()` with an explicit blob word order** (`WC_STM32_WRAP_ORDER_RAW` / `WC_STM32_WRAP_ORDER_LEGACY`). Both build paths now share one wrap implementation, but each keeps its historical default so already-provisioned key material stays valid: CubeMX defaults to `LEGACY` (matching the `HAL_CRYPEx_WrapKey` behavior shipped in 5.9.0 - 5.9.2), bare-metal to `RAW`. `RAW` is recommended for new provisioning - it is the one format both builds agree on and the format the DHUK derive path already uses. `WC_STM32_WRAP_DEFAULT_RAW_ORDER` changes the plain `wc_Stm32_Aes_Wrap()` default build-wide. Both orders are documented with an off-target formula and measured vectors, so an offline provisioning tool can reproduce them with any AES library.
- **`wc_ecc_import_wrapped_private()` widened to the CubeMX build**: its guard was `WOLFSSL_STM32_BARE`-only, so the DHUK ECC sign path was unreachable from a HAL build. `ecc.h` now includes `stm32.h` under `WOLFSSL_DHUK` so the prototype and the `ecc.c` definition are gated identically - previously the guard silently evaluated false in any translation unit that had not already seen `stm32.h`.

## Fixes

- **`asn.c`: guard NULL `OidFromId` / `GetAlgoV2` results in `wc_EncryptPKCS8Key_ex`** - fixes a GCC `-Wnonnull` break (crypto-callback-only PKCS8 build at `-O2 -Werror`) and closes a latent NULL-deref on the `pbeOidBuf` path, matching the existing pattern in the sibling `EncryptContentPBES2`. No API change. Adds two `tests/api.c` regression tests (`test_wc_EncryptPKCS8Key_ex_badHmac` / `_goodHmac`) covering the bad-`hmacOid` rejection and a positive encrypt/decrypt round-trip control.
- **STM32 PKA sign/verify gating**: `WC_STM32_PKA_VERIFY_ONLY` now also compiles out the ECC scalar-multiply helpers, not just sign. The STM32H563 "light" PKA has no scalar-mul engine, so those helpers were dead weight that would not link on a verify-only part.
- **STM32 build-config fixes**: add the missing `WOLFSSL_STM32U3` PKA include arm; `#ifndef`-guard the bare HASH-legacy macros so `WOLFSSL_STM32_BARE` coexists with the ST Cube HAL headers in a single translation unit (e.g. under Zephyr).

## Documentation

- `wolfcrypt/src/port/st/README.md`: new sections on the crypto-callback HW path under `WOLF_CRYPTO_CB_ONLY_AES` / `_ECC`, the plaintext-key vs DHUK-seed device split with a worked devId example, and the key-wrap word orders (comparison table, off-target formula, measured NUCLEO-U385RG-Q vectors).
- `wolfcrypt/src/port/realtek/README.md`: the plaintext-key device, its decline-vs-hard-fail contract, and HUK coexistence.
- `doc/dox_comments/header_files/stm32.h`: API docs for `wc_Stm32_Aes_Wrap_ex` and the blob word-order constants.

## Testing

Hardware, all `Result: 0 (PASS)`:

- **NUCLEO-U385RG-Q** (CubeMX + bare): plaintext AES-GCM matches published vectors (encrypt + decrypt + tamper-reject), coexistence with the DHUK device (same key bytes -> distinct ciphertext per devId), CubeMX HW ECDSA sign/verify + CCB, and the `RAW` / `LEGACY` wrap-blob vectors quoted in the README.
- **NUCLEO-U585** (bare): plaintext AES-GCM and DHUK-device coexistence.
- **NUCLEO-F439ZI** (CRYP IP): direct-call HW AES-GCM encrypt + decrypt KAT.
- **RealTek RTL8735B** (AmebaPro2 EVB): 67/67 including plaintext AES-GCM on the HW engine and HUK coexistence.

Host / CI:

- `./configure --enable-rtl8735b` compile-test through the HAL shim; `testwolfcrypt` passes (RTL8735B self-test + crypto-callback test).
- `./configure --enable-all` + `make check`: 17 pass / 6 skip / 0 fail.

Examples and on-target tests: https://github.com/wolfSSL/wolfssl-examples-stm32/pull/16


dgarske:
Jenkins retest this please

dgarske:
Jenkins retest this please

wolfcrypt/src/port/st/stm32.c
```diff
@@ -2334,2954 +2496,3456 @@ int wc_Stm32_Aes_Cbc(struct Aes* aes, byte* out, const byte* in,
...
+    }
+    pka_ecc.hash       = Hashbin;
+    pka_ecc.integer    = Intbin;
+    pka_ecc.privateKey = Keybin;
+    pka_ecc_out.RSign  = Rbin;
+    pka_ecc_out.SSign  = Sbin;
 
-    return 0;
-}
+    status = HAL_PKA_ECDSASign(&hpka, &pka_ecc, HAL_MAX_DELAY);
```
philljj, line 5707: Does this need a
```
Stm32Pka_EnsureInit();
```
preceding the HAL_PKA_ECDSASign() call? The similar code block at line 6458 has it:

```
    Stm32Pka_EnsureInit(); 
    status = HAL_PKA_ECDSASign(&hpka, &pka_ecc, HAL_MAX_DELAY); 
    if (status != HAL_OK) { 
        ForceZero(Keybin, sizeof(Keybin)); 
        ForceZero(Intbin, sizeof(Intbin));
        HAL_PKA_RAMReset(&hpka); 
        return WC_HW_E;
    }       
    HAL_PKA_ECDSASign_GetResult(&hpka, &pka_ecc_out, NULL);
```

wolfcrypt/src/port/st/stm32.c
```diff
@@ -2334,2954 +2496,3456 @@ int wc_Stm32_Aes_Cbc(struct Aes* aes, byte* out, const byte* in,
...
+/* ---------------------------------------------------------------------------
+ * DHUK crypto-callback backend -- shared by the bare-metal and CubeMX/HAL
+ * builds. Placed after the BARE / CUBEMX / StdPeriph split (and after the
+ * CubeMX cipher/PKA helpers) so the dispatcher can call into either build's
+ * helpers without forward declarations. The SAES primitives it runs on live
+ * in the shared block near the top of the STM32_CRYPTO section.
+ * ------------------------------------------------------------------------- */
+#if defined(WOLFSSL_DHUK) && defined(WC_STM32_HAS_DHUK) && \
+    defined(WOLF_CRYPTO_CB) && \
+    (defined(WOLFSSL_STM32_BARE) || defined(WOLFSSL_STM32_CUBEMX))
```
philljj, line 5003: dumb question: with `WOLFSSL_STM32_CUBEMX`  it's possible to be FREERTOS / multithreaded (according to `examples/configs/user_settings_stm32.h`).

Are `g_stm32DhukSeed*` and Stm32Dhuk_StageSeed guarded by locks everywhere?

I do see MutexLock calls around seed access in Stm32Dhuk_Gmac.

I don't see MutexLock calls around Stm32Dhuk_PkSign -> Stm32Dhuk_StageSeed call chain.

wolfcrypt/src/port/st/stm32.c
```diff
@@ -2334,2954 +2496,3456 @@ int wc_Stm32_Aes_Cbc(struct Aes* aes, byte* out, const byte* in,
...
+    if (ivSz == 0u) {
+        return BAD_FUNC_ARG;
+    }
+    if (tagSz < 4u || tagSz > WC_AES_BLOCK_SIZE) {
+        return BAD_FUNC_ARG;
+    }
+    if (aad == NULL && aadSz > 0u) {
+        return BAD_FUNC_ARG;
+    }
+    if (g_stm32DhukSeedSz != 32u) {
```
philljj: nit: should `g_stm32DhukSeedSz` be checked after `wolfSSL_CryptHwMutexLock`?

wolfssl/wolfcrypt/port/st/stm32.h
```diff
@@ -944,29 +1003,55 @@ int stm32_ecc_sign_hash_ex(const byte* hash, word32 hashlen, struct WC_RNG* rng,
...
     void wc_Stm32_DhukUnRegister(int devId);
 #endif
 
-#endif /* WOLFSSL_STM32_BARE && WC_STM32_HAS_DHUK */
+/* Crypto-callback device id for the plaintext-key AES device (distinct from the
+ * DHUK/SAES device ids 807/808/809). Registering both lets an application
+ * choose, per Aes, whether its key is used verbatim (this devId) or as a DHUK
+ * derivation seed (WC_DHUK_DEVID) -- selected by the devId passed to wc_AesInit.
+ * Override before include if it collides. */
+/* Guard is the exact union of the two device definitions in stm32.c: the CubeMX
```
philljj: Was this meant to be one comment and something got cutoff? comment reads strange

philljj (changes requested):
First pass. Looks good so far, just a few questions.

tests/api.c
```diff
@@ -24018,6 +24018,88 @@ static int test_wc_EncryptPKCS8Key_rc4NoPad(void)
     return EXPECT_RESULT();
 }
 
+/* Negative: PKCS5v2 with a valid encryption algorithm but an hmacOid that maps
+ * to no OID must return a clean ALGO_ID_E, not dereference a NULL OidFromId()
+ * result. Exercises the NULL guards added in wc_EncryptPKCS8Key_ex. */
+static int test_wc_EncryptPKCS8Key_ex_badHmac(void)
```
philljj, line 24197: These tests may need a FIPS define guard. Seeing several of these in PRB fips harness tests:
```
 824: test_wc_EncryptPKCS8Key_ex_goodHmac                :
ERROR - tests/api.c line 24087 failed with:
    expected: encSz = (word32)wc_EncryptPKCS8Key_ex(plain, (word32)sizeof(plain), enc, &encSz, password, (int)strlen((password)), PKCS5, PBES2, AES256CBCb, ((void *)0), 0, WC_PKCS12_ITT_DEFAULT, HMAC_SHA256_OID, &rng, ((void *)0)) > 0
    result:   -200 <= 0
```

philljj:
Retest this please.

(PRB passed, but results didn't report back).

philljj (approved):

PR #10971, opened 2026-07-22
SE050: create ECDH derive target object for applet 7.2 middleware

LinuxJedi:
Middleware built for SE05x applet >= 07_02 (required for SE052) derives the ECDH shared secret with Se05x_API_ECDHGenerateSharedSecret_InObject, which requires TLV[TAG_7] to reference an existing HMACKey object sized exactly to the shared secret; otherwise the applet returns SW 0x6985 (conditions not satisfied). The port never created this object, so ECDH offload failed with WC_HW_E on such builds.

Create the derive target as an HMACKey object of the exact secret size before the derive, for both ECC and Curve25519 shared secrets. Read the result back as AES type since sss_se05x_key_store_get_key has no HMAC read case.

Fixes zd#22069

# Testing

wolfSSL/simulators#9


Frauschi (dismissed):
LGTM

LinuxJedi:
Also fixes ZD#22212

LinuxJedi:
Updated PR with some missing policy pieces.

Frauschi (dismissed):
@LinuxJedi PRB-master-job fails, as `SSS_HAVE_SE05X_VER_GTE_07_02` is missing from `.wolfssl_known_macro_extras`. Otherwise LGTM.

LinuxJedi:
I need to do further testing before I'm ready to call this good.

LinuxJedi:
Should all be good now.

wolfcrypt/src/port/nxp/se050_port.c
```diff
@@ -2878,6 +2882,59 @@ int se050_ecc_shared_secret(ecc_key* private_key, ecc_key* public_key,
...
+    if (status == kStatus_SSS_Success) {
+        /* Middleware built for applet >= 7.2 derives into an SE05x
+         * resident object, but the applet refuses to export a symmetric
+         * key object regardless of the policy attached at its creation
+         * (verified on SE051 applet 7.2.0 hardware), so a derived secret
+         * stored in an object can never be read back. Use the direct
+         * APDU that returns the shared secret in the response instead,
+         * as the middleware itself does whenever the derived key lives
+         * in a host keystore. */
+        byte peerPoint[SE050_ECC_DER_MAX];
```
Frauschi, line 2909: **🟡 [Medium] Peer public key is still uploaded to the SE050 on the applet 7.2 path where it is never used**
*💡 SUGGEST*  `question`

On the new `SSS_HAVE_SE05X_VER_GTE_07_02` path the shared secret is computed with `Se05x_API_ECDHGenerateSharedSecret`, which carries the peer point in TLV[TAG_2] of the command - nothing about the derive needs the peer key to be resident on the SE050. Yet the block at lines 2804-2844 still runs unconditionally: when `public_key->keyIdSet == 0` it does `wc_EccPublicKeyToDer` + `se050_allocate_key` + `sss_key_object_allocate_handle` + `sss_key_store_erase_key` + `sss_key_store_set_key`, and the resulting object is then ignored because the 7.02 branch re-exports the point from the software key with `wc_ecc_export_x963`. Consequences on the new path: (1) several wasted APDU round trips per ECDH; (2) in the default (non-`WOLFSSL_SE050_ONLY_KEY_ID`) build the object is created with `kKeyObject_Mode_Persistent` and never erased, so every distinct peer key permanently consumes an SE050 object slot and NVM write endurance for no benefit; (3) a new failure mode is inherited - if the SE050 is out of objects or the write fails, `status` becomes `kStatus_SSS_Fail` and the ECDH aborts even though the direct APDU needed nothing on-chip. The same applies to `se050_curve25519_shared_secret` (lines 3459-3495, transient there so only the wasted round trips apply). This is not a regression versus master (master needs the object for `sss_derive_key_dh`), but the diff makes the upload unnecessary on the 7.02 path and it would be cheap to skip it.

**Suggestion:** Guard the peer-key upload block (se050_port.c:2806-2840 and 3461-3490) with the same `#if !(defined(SSS_HAVE_SE05X_VER_GTE_07_02) && SSS_HAVE_SE05X_VER_GTE_07_02)` used for the derive, so on 7.02 builds the peer key is only read back when it was already SE050-resident (`keyIdSet != 0`).

**Recommendation:** Skip the peer public key upload on the >= 7.02 path, or document why the object is still wanted (e.g. so `public_key->keyIdSet` becomes set for later use) - as written it costs SE050 flash slots and adds a failure path the direct APDU does not need.

.github/workflows/se050-sim.yml
```diff
@@ -100,4 +102,8 @@ jobs:
           cache-to: ${{ ((github.event_name == 'schedule' || github.event_name == 'workflow_dispatch') && matrix.name == 'default') && 'type=registry,ref=ghcr.io/wolfssl/wolfssl-sim-cache:se050,mode=max' || '' }}
 
       - name: Run wolfCrypt tests against simulator
```
Frauschi, line 104: **🟡 [Medium] CI only builds the applet >= 7.2 side of the new #if; the restored pre-7.2 Binary path is never compiled or run**
*💡 SUGGEST*  `test`

The CI image builds NXP plug-and-trust at tag `v04.07.01`, whose stock `fsl_sss_ftr.h` sets `#define SSS_HAVE_SE05X_VER_07_02 1` (line 90), which in turn yields `#define SSS_HAVE_SE05X_VER_GTE_07_02 1` (line 537). The simulator's `wolfcrypt-test/patch_ftr.py` only rewrites `SSS_HAVE_EC_*` and `SSS_HAVE_HOSTCRYPTO_NONE`, so it does not change the applet version. Therefore both matrix legs compile the new `#else` (direct-APDU) branch, and the `#if !(...)` branch - the Binary derive target that the PR deliberately restored for applet 3.x parts after the SE050C hardware regression (commit 'SE050: keep Binary ECDH derive target on pre-7.2 middleware') - has no build or run coverage at all. A regression in that branch would only surface on real applet 3.x silicon. Note this also means the `deriveKeyCreated`/erase-before-derive reordering in `se050_curve25519_shared_secret` (lines 3512-3532), which this PR changes, is untested.

**Suggestion:**
```suggestion
- name: applet-3x
            extra_cflags: ''
            # plus a build-arg/patch that flips fsl_sss_ftr.h to
            # SSS_HAVE_SE05X_VER_03_XX 1 / SSS_HAVE_SE05X_VER_07_02 0
```

**Recommendation:** Add a matrix leg that builds the SDK with the 03_XX applet version (PTMW_SE05X_Ver=03_XX, or a patch_ftr.py tweak) so both sides of the new preprocessor guard are compiled and exercised; at minimum state in a comment that the pre-7.2 branch is hardware-verified only.

wolfcrypt/src/port/nxp/se050_port.c
```diff
@@ -2878,6 +2882,59 @@ int se050_ecc_shared_secret(ecc_key* private_key, ecc_key* public_key,
...
+            if (status == kStatus_SSS_Success && derSz >= pointSz &&
+                    peerPoint[derSz - pointSz] == 0x04) {
+                XMEMMOVE(peerPoint, peerPoint + derSz - pointSz, pointSz);
+                peerPointSz = pointSz;
+            }
+            else {
+                status = kStatus_SSS_Fail;
+            }
+        }
+        if (status == kStatus_SSS_Success) {
```
Frauschi, line 2936: **🔵 [Low] ECC direct-APDU path accepts any returned shared-secret length without validating it**
*💡 SUGGEST*  `bug`

The ECC branch propagates whatever length the applet reports straight into `*outlen` with no sanity check, while the Curve25519 branch added in the same diff does check (`if (sm == SM_OK && outSz == CURVE25519_KEYSIZE)`). For a NIST curve the ECDH secret is always exactly `private_key->dp->size` bytes, so a short response indicates a malformed/truncated R-APDU; accepting it silently hands the caller a truncated premaster secret (`wc_ecc_shared_secret` returns 0 and the caller trusts `*outlen`). Adding the check costs nothing and makes the two new branches consistent.

**Suggestion:**
```suggestion
if (sm == SM_OK && outSz == (size_t)keySize) {
                *outlen = (word32)outSz;
            }
```

**Recommendation:** Validate `outSz == (size_t)keySize` before assigning `*outlen`, mirroring the Curve25519 branch.

.github/workflows/se050-sim.yml
```diff
@@ -100,4 +102,8 @@ jobs:
...
-        run: docker run --rm wolfssl-se050-sim:ci-${{ matrix.name }}
+        # SE050_SIM_STRICT_ECDH=1 makes the simulator enforce the applet 7.2
```
Frauschi: **🔵 [Low] Workflow comment describes a strict-ECDH contract no code path in this PR follows**
*🔧 NIT*  `style`

The comment states that `SE050_SIM_STRICT_ECDH=1` makes the simulator require the Tag7 target to be pre-created as an HMACKey object. That was true of an intermediate commit in this PR, but the final port never sends TLV[TAG_7] at all: the 7.02 branch uses the direct (Tag7-less) variant and the pre-7.2 branch goes through `sss_derive_key_dh`, which on a < 7.2 middleware also omits Tag7. At the pinned simulator commit (`895ff41b`, `handlers/ec.rs:466-471`) `strict` only gates the InObject form, so with the CI's 7.02 build the switch cannot influence any request the port makes. Keeping the flag is defensible as a guard against reverting to the InObject flow, but the comment now misdescribes what CI verifies.

**Suggestion:**
```suggestion
# SE050_SIM_STRICT_ECDH=1 enforces the applet 7.2 InObject contract
        # (a Tag7 target must already exist as an exactly-sized HMACKey).
        # The port now always uses the Tag7-less direct variant, so this is a
        # regression guard: it fails the run if the InObject flow comes back.
```

**Recommendation:** Reword the comment to say the flag is a regression guard rather than the mechanism that reproduces the original failure.

wolfcrypt/src/port/nxp/se050_port.c
```diff
@@ -2878,6 +2882,59 @@ int se050_ecc_shared_secret(ecc_key* private_key, ecc_key* public_key,
...
+                private_key->keyId, peerPoint, peerPointSz, out, &outSz);
+            if (sm == SM_OK) {
+                *outlen = (word32)outSz;
+            }
+            else {
+                status = kStatus_SSS_Fail;
+            }
+        }
+    }
+    (void)ctx_derive_key;
```
Frauschi: **🔵 [Low] Derive-key locals and cleanup are dead on the 7.02 path and silenced with (void) casts**
*🔧 NIT*  `style`

On the 7.02 path `deriveKeyCreated` can never become 1, so the `if (deriveKeyCreated)` cleanup at 2938-2941 / 3606-3609 is dead, and `sss_key_object_init(&deriveKey, &host_keystore)` at 2846 / 3497 still runs for an object nothing on that path touches. Two sizeable SSS structs (`sss_object_t`, `sss_derive_key_t`) also sit on the stack for nothing, which matters on the constrained targets this port serves. Note that `(void)deriveKey;` is redundant even as written: `deriveKey` is used by the `sss_key_object_init` call, so it would never draw an unused-variable warning. Only `ctx_derive_key` needs suppressing, which is itself the sign that it belongs inside the guard. Moving the declarations, the `sss_key_object_init` call and the cleanup block inside the `#if !(...)` arm removes both the dead code and the `(void)` casts.

**Suggestion:** Scope `sss_object_t deriveKey;`, `sss_derive_key_t ctx_derive_key;`, `int deriveKeyCreated`, the `sss_key_object_init(&deriveKey, ...)` call and the `if (deriveKeyCreated) {...}` cleanup inside the pre-7.2 `#if` arm instead of casting them to void.

**Recommendation:** Scope the derive-key state to the branch that uses it rather than declaring it unconditionally and voiding it.

Frauschi (commented):
## 🐺 Skoll Code Review

### Posted findings
- [Medium] **Peer public key is still uploaded to the SE050 on the applet 7.2 path where it is never used** — `wolfcrypt/src/port/nxp/se050_port.c:2895-2921`
- [Medium] **CI only builds the applet >= 7.2 side of the new #if; the restored pre-7.2 Binary path is never compiled or run** — `.github/workflows/se050-sim.yml:104-109`
- [Low] **ECC direct-APDU path accepts any returned shared-secret length without validating it** — `wolfcrypt/src/port/nxp/se050_port.c:2922-2933`
- [Low] **Workflow comment describes a strict-ECDH contract no code path in this PR follows** — `.github/workflows/se050-sim.yml:105-109`
- [Low] **Derive-key locals and cleanup are dead on the 7.02 path and silenced with (void) casts** — `wolfcrypt/src/port/nxp/se050_port.c:2935-2941`

- [Info] Explanatory comment inserted in the middle of a function argument list
- [Info] Unrelated .wolfssl_known_macro_extras change bundled into the SE050 PR

</details>

dgarske (approved):

PR #10967, opened 2026-07-22
MC/DC decision coverage part 4:  remaining wolfCrypt primitives

danielinux:
Add DecisionCoverage/FeatureCoverage tests and tests/unit-mcdc white-box supplements for the remaining reachable wolfCrypt primitive sources in the ISO 26262 per-module MC/DC campaign (excluding asn* and the EVP/OpenSSL compat layer, which are  mot yet covered).

tests/api:
- legacy ciphers / digests: des3, camellia, ascon, blake2, siphash
- niche PK: srp, eccsi, sakke, hpke
- native PQC KEM: frodokem
- math: wolfmath

tests/unit-mcdc white-box drivers (#include the .c to reach file-static helpers and impl-selected paths, standalone main()/WB_NOTE harness):
- blake2b, blake2s
- eccsi, sakke, hpke
- SP host backends: sp_x86_64, sp_c64, sp_c32
- infra: cryptocb (dev && dev->cb dispatch three-vector, 127/127), logging (per-thread + global error-queue impls, 19/19), memory (static-pool allocator, 46/48)

Registrations in tests/api.c, tests/api/include.am and CMakeLists.txt.

PR #10973, opened 2026-07-22
Bug Fixes: issues discovered during MC/DC coverage step 4

danielinux:
# Description

3x wolfcrypt bugfixes.

# Testing
https://github.com/wolfSSL/wolfssl/pull/10967

# Checklist

 - [x] added tests

### DIFF
--- a/wolfcrypt/src/rsa.c
+++ b/wolfcrypt/src/rsa.c
@@ -5062,8 +5062,18 @@ static int wc_CompareDiffPQ(mp_int* p, mp_int* q, int size, int* valid)
         return BAD_FUNC_ARG;
 
 #ifdef WOLFSSL_SMALL_STACK
-    if (((c = (mp_int *)XMALLOC(sizeof(*c), NULL, DYNAMIC_TYPE_WOLF_BIGINT)) == NULL) ||
-        ((d = (mp_int *)XMALLOC(sizeof(*d), NULL, DYNAMIC_TYPE_WOLF_BIGINT)) == NULL))
+    c = (mp_int *)XMALLOC(sizeof(*c), NULL, DYNAMIC_TYPE_WOLF_BIGINT);
+    d = (mp_int *)XMALLOC(sizeof(*d), NULL, DYNAMIC_TYPE_WOLF_BIGINT);
+    /* Zero any struct that was allocated so the cleanup path's mp_clear()/
+     * mp_forcezero() are safe even when the sibling allocation fails and the
+     * mp_init_multi() below is skipped on MEMORY_E. Without this, clearing an
+     * allocated-but-uninitialized mp_int reads a garbage ->used and corrupts
+     * the heap. */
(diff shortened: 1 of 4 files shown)


philljj:
Retest this please.

(not clear what failed).

philljj:
Retest this please.

windows test:
```
[0] Incorrect return -308
```

Frauschi (dismissed):
LGTM

wolfcrypt/src/rsa.c
```diff
@@ -5062,8 +5062,18 @@ static int wc_CompareDiffPQ(mp_int* p, mp_int* q, int size, int* valid)
...
-        ((d = (mp_int *)XMALLOC(sizeof(*d), NULL, DYNAMIC_TYPE_WOLF_BIGINT)) == NULL))
+    c = (mp_int *)XMALLOC(sizeof(*c), NULL, DYNAMIC_TYPE_WOLF_BIGINT);
+    d = (mp_int *)XMALLOC(sizeof(*d), NULL, DYNAMIC_TYPE_WOLF_BIGINT);
+    /* Zero any struct that was allocated so the cleanup path's mp_clear()/
+     * mp_forcezero() are safe even when the sibling allocation fails and the
+     * mp_init_multi() below is skipped on MEMORY_E. Without this, clearing an
+     * allocated-but-uninitialized mp_int reads a garbage ->used and corrupts
+     * the heap. */
+    if (c != NULL)
+        XMEMSET(c, 0, sizeof(*c));
```
SparkiDev: Don't like the XMEMSET.
The mp_init_multi should be doing that if necessary.
Don't want to be doing multiple mp_init calls either.
Is the XMEMSET really needed?

SparkiDev (changes requested):

wolfcrypt/src/rsa.c
danielinux: Yes something is needed here, but you're right XMEMSET is not the right tool. 

I've changed it to work without it now.

This issue isn't that mp_init_multi() does not zero, it's that the path is never reached. The temporary buffers are XMALLOC'd in a short circuit condition chain `||` so when the next allocation returns NULL we set MEMORY_E and skip the call to mp_init_multi, while there's still a cleanup running (mp_clear + mp_forcezero) on all non-null pointers, including the struct that was allocated byt never initialized.

With sp_math that implies a ForceZero with an uninitialized size, on tfm/integer there's a free on a dangling `dp`.

The fail can be reproduced by forcing the second XMALLOC to fail in wc_CheckProbablePrime_ex() with p allocated and q=NULL. 

This new version restores the original short-circuit condition and removes the XMEMSET.  I've changed the failure path so the two temp buffers are freed and their pointers set to NULL to 'skip' the common cleanup path.

danielinux:
jenkins retest this please

dgarske (approved):

SparkiDev (approved):

PR #10975, opened 2026-07-22
Move WOLFSSL_X509_TINY test to the unit test suite and run

aidangarske:
- Add CI configurations for the minimal `WOLFSSL_X509_TINY` profile and its extension add-backs.
  - Keep `cert_x509_tiny_test()` in `wolfcrypt/test/test.c` so it remains available to crypt-only and embedded test targets.
  - Expand the test to cover retained extensions, stripped non-critical extensions, restored SKI/AKI/AIA/CRL-DP/certificatePolicies extensions, and critical
  name constraints.
  - Make the RSA-PSS certificate test compatible with TINY builds without AKI by verifying same-subject roots independently.
  - Run the full `make check` suite for both new CI configurations.


tests/api.c
```diff
@@ -3730,6 +3730,162 @@ static int test_wolfSSL_CertRsaPss(void)
...
+{
+    EXPECT_DECLS;
+    DecodedCert cert;
+
+    wc_InitDecodedCert(&cert, tinyCert_plain, (word32)sizeof(tinyCert_plain),
+        NULL);
+    ExpectIntEQ(wc_ParseCert(&cert, CERT_TYPE, NO_VERIFY, NULL), 0);
+    wc_FreeDecodedCert(&cert);
+
+#ifndef WOLFSSL_NO_ASN_STRICT
```
Frauschi: **🔵 [Low] critUnknown assertion gated on WOLFSSL_NO_ASN_STRICT, but the code does not relax it**
*💡 SUGGEST*  `test`

The new critical-unknown-extension assertion is wrapped in `#ifndef WOLFSSL_NO_ASN_STRICT`, implying the rejection only happens in strict builds. But the source is explicit that this is not relaxed by `WOLFSSL_NO_ASN_STRICT`: `wolfcrypt/src/asn.c:22053-22057` comments "RFC 5280 4.2 makes this a MUST that WOLFSSL_NO_ASN_STRICT does not relax", and the `default` case at `asn.c:22058-22061` sets `ASN_CRIT_EXT_E` unconditionally on a critical unknown extension. I confirmed the code returns `ASN_CRIT_EXT_E` regardless of the strict flag, and that it propagates out of `wc_ParseCert` even under `NO_VERIFY` (`asn.c:24766-24767`). So the guard needlessly drops this assertion in `WOLFSSL_NO_ASN_STRICT` builds and slightly contradicts the source comment. Note this is not exercised by the x509-tiny CI job (which does not define `WOLFSSL_NO_ASN_STRICT`), so it is a coverage/clarity issue, not a CI failure.

**Suggestion:**
```suggestion
/* Unknown critical extension is rejected regardless of WOLFSSL_NO_ASN_STRICT
 * (RFC 5280 4.2 MUST, see asn.c DecodeExtensionType default case). */
wc_InitDecodedCert(&cert, tinyCert_critUnknown,
    (word32)sizeof(tinyCert_critUnknown), NULL);
ExpectIntEQ(wc_ParseCert(&cert, CERT_TYPE, NO_VERIFY, NULL),
    WC_NO_ERR_TRACE(ASN_CRIT_EXT_E));
wc_FreeDecodedCert(&cert);
```

tests/api.c
```diff
@@ -3730,6 +3730,162 @@ static int test_wolfSSL_CertRsaPss(void)
...
+    ExpectIntEQ(wc_ParseCert(&cert, CERT_TYPE, NO_VERIFY, NULL), 0);
+#else
+    ExpectIntEQ(wc_ParseCert(&cert, CERT_TYPE, NO_VERIFY, NULL),
+        WC_NO_ERR_TRACE(ASN_NAME_INVALID_E));
+#endif
+    wc_FreeDecodedCert(&cert);
+#endif
+    return EXPECT_RESULT();
+}
+#endif /* WOLFSSL_X509_TINY && HAVE_ECC && !NO_ECC256 && !NO_ECC_SECP */
```
Frauschi: **⚪ [Info] Double blank line after the moved test function**
*🔧 NIT*  `style`

The moved block ends with `#endif` followed by two blank lines before `test_wolfSSL_CTX_load_verify_locations_ex`. Single blank-line separation is the norm elsewhere in this file. Purely cosmetic.

**Suggestion:**
```suggestion
#endif /* WOLFSSL_X509_TINY && HAVE_ECC && !NO_ECC256 && !NO_ECC_SECP */

static int test_wolfSSL_CTX_load_verify_locations_ex(void)
```

Frauschi (commented):
## 🐺 Skoll Code Review

### Posted findings
- [Low] **critUnknown assertion gated on WOLFSSL_NO_ASN_STRICT, but the code does not relax it** — `tests/api.c:3866-3872`
- [Info] **Double blank line after the moved test function** — `tests/api.c:3887-3889`

tests/api.c
```diff
@@ -3730,6 +3730,157 @@ static int test_wolfSSL_CertRsaPss(void)
...
+    0x18, 0x05, 0xAC, 0xAD, 0x49, 0x88, 0x28, 0x39, 0xBD, 0xCD, 0x73, 0xFE,
+    0x0A, 0x74, 0x98, 0x11, 0x3F, 0xFA, 0xF1, 0x18, 0xCF, 0x4E, 0x4E, 0x93,
+};
+#endif
+
+/* WOLFSSL_X509_TINY: a normal cert parses; a critical-unknown extension and
+ * nameConstraints fail closed unless the add-back is set. */
+static int test_wolfSSL_X509_tiny(void)
+{
+    EXPECT_DECLS;
```
Frauschi: **🟡 [Medium] tinyCert_plain assertion proves nothing TINY-specific**
*💡 SUGGEST*  `test`

The first case only asserts `wc_ParseCert(...) == 0`. `tinyCert_plain` carries SKI, AKI, basicConstraints, SAN, keyUsage and extKeyUsage, all of which parse to `0` in a **full** (non-TINY) build too, so this assertion cannot distinguish a working TINY profile from a build where `WOLFSSL_X509_TINY` was silently not applied. The observable TINY behaviour is that the stripped decoders never run: `wolfcrypt/src/asn.c:21861` and `:21893` handle `AUTH_KEY_OID` / `SUBJ_KEY_OID` with the explicit comment "No SET_OID, bit means decoded", so `extAuthKeyIdSet` and `extSubjKeyIdSet` stay `0` under TINY and are `1` otherwise. Asserting those makes the test actually verify the profile.

**Suggestion:**
```suggestion
wc_InitDecodedCert(&cert, tinyCert_plain, (word32)sizeof(tinyCert_plain),
        NULL);
    ExpectIntEQ(wc_ParseCert(&cert, CERT_TYPE, NO_VERIFY, NULL), 0);
    /* Stripped decoders never run, so their "decoded" bits stay clear. */
#ifndef WOLFSSL_X509_TINY_SKI
    ExpectIntEQ(cert.extSubjKeyIdSet, 0);
#endif
#ifndef WOLFSSL_X509_TINY_AKI
    ExpectIntEQ(cert.extAuthKeyIdSet, 0);
#endif
    wc_FreeDecodedCert(&cert);
```

.github/workflows/os-check.yml
```diff
@@ -448,7 +448,11 @@ jobs:
           {"comment": "wolfSSL <= 5.9.0 DTLS 1.3 legacy_session_id echo compatibility shim",
            "name": "dtls13-echo-legacy-session-id",
            "configure": ["--enable-dtls", "--enable-dtls13", "--enable-session-ticket",
-            "--enable-ecc", "CPPFLAGS=-DWOLFSSL_DTLS13_ECHO_LEGACY_SESSION_ID"]}
+            "--enable-ecc", "CPPFLAGS=-DWOLFSSL_DTLS13_ECHO_LEGACY_SESSION_ID"]},
+          {"name": "x509-tiny", "minutes": 1.0,
```
Frauschi: **🟡 [Medium] No CI coverage for any WOLFSSL_X509_TINY add-back macro**
*💡 SUGGEST*  `test`

The new `x509-tiny` entry defines only the bare `WOLFSSL_X509_TINY`. That means the `#ifdef WOLFSSL_X509_TINY_NAME_CONSTRAINTS` branch of the new test (`tests/api.c:3873-3874`) is never compiled by CI, and none of the add-back macros — `_NAME_CONSTRAINTS`, `_AKI`, `_SKI`, `_AIA`, `_CRL_DP`, `_POLICIES`, `_SAN_EMAIL/_IP/_DIR/_RID` — get any build coverage. The settings.h forcing logic at `wolfssl/wolfcrypt/settings.h:4791-4817` (NAME_CONSTRAINTS implying the four SAN add-backs, HAVE_OCSP implying AIA, HAVE_CRL_IO implying CRL_DP) is likewise unbuilt. Half the new test is dead code in CI.

**Suggestion:**
```suggestion
{"name": "x509-tiny", "minutes": 1.0,
 "comment": "Minimal-extension X.509 profile; runs just its unit test.",
 "configure": ["--enable-ecc", "CPPFLAGS=-DWOLFSSL_X509_TINY"],
 "check": false, "run": [["tests/unit.test", "-test_wolfSSL_X509_tiny"]]},
{"name": "x509-tiny-addbacks", "minutes": 1.0,
 "comment": "TINY profile with the add-back macros on; covers the settings.h implication chain and the NAME_CONSTRAINTS branch of the unit test.",
 "configure": ["--enable-ecc",
  "CPPFLAGS=-DWOLFSSL_X509_TINY -DWOLFSSL_X509_TINY_NAME_CONSTRAINTS -DWOLFSSL_X509_TINY_AKI -DWOLFSSL_X509_TINY_SKI -DWOLFSSL_X509_TINY_POLICIES"],
 "check": false, "run": [["tests/unit.test", "-test_wolfSSL_X509_tiny"]]}
```

tests/api.c
```diff
@@ -3730,6 +3730,157 @@ static int test_wolfSSL_CertRsaPss(void)
...
+    0x4E, 0xBD, 0xDB, 0x1A, 0x65, 0x3C, 0xEB, 0xCC, 0xF3, 0x01, 0xB8, 0xBD,
+    0xF6, 0x02, 0x21, 0x00, 0x83, 0xAE, 0x5E, 0x31, 0xE7, 0xF3, 0x7A, 0x70,
+    0x18, 0x05, 0xAC, 0xAD, 0x49, 0x88, 0x28, 0x39, 0xBD, 0xCD, 0x73, 0xFE,
+    0x0A, 0x74, 0x98, 0x11, 0x3F, 0xFA, 0xF1, 0x18, 0xCF, 0x4E, 0x4E, 0x93,
+};
+#endif
+
+/* WOLFSSL_X509_TINY: a normal cert parses; a critical-unknown extension and
+ * nameConstraints fail closed unless the add-back is set. */
+static int test_wolfSSL_X509_tiny(void)
```
Frauschi: **🟡 [Medium] TINY coverage lost for cryptonly / embedded builds**
*💡 SUGGEST*  `test`

The test moved out of `wolfcrypt/test/test.c` into `tests/api.c`. `tests/api.c` is only compiled under the `BUILD_TESTS` automake conditional (`tests/include.am:6`), i.e. a TLS-enabled autotools build. `wolfcrypt/test/test.c` is what `wolfcrypt/test/testwolfcrypt` is built from (`wolfcrypt/test/include.am:12`) and is the harness that `--enable-cryptonly`, `--enable-usersettings` and embedded users actually run. `WOLFSSL_X509_TINY` is explicitly a minimal/embedded profile, so this removes the self-test from exactly the audience the feature targets. Note the eight existing `testwolfcrypt-*` os-check entries all run `wolfcrypt/test/testwolfcrypt` and would have picked it up; the new entry runs `tests/unit.test` instead.

**Suggestion:** Either keep a thin wolfCrypt-level `cert_x509_tiny_test()` in wolfcrypt/test/test.c alongside the new api.c test, or state in the PR/commit that TINY is only validated in TLS-enabled builds from now on.

tests/api.c
```diff
@@ -3730,6 +3730,157 @@ static int test_wolfSSL_CertRsaPss(void)
 #endif /* WOLFSSL_TEST_APPLE_NATIVE_CERT_VALIDATION */
 #endif /* HAVE_CERT_CHAIN_VALIDATION */
 
+#if defined(WOLFSSL_X509_TINY) && defined(HAVE_ECC) && \
+    !defined(NO_ECC256) && !defined(NO_ECC_SECP)
+static const byte tinyCert_plain[] = {
```
Frauschi: **🔵 [Low] tinyCert_plain has a one-year validity while its siblings have ten**
*🔧 NIT*  `test`

`tinyCert_plain` is notBefore 2026-06-29 / notAfter **2027-06-29**, whereas `tinyCert_critUnknown` and `tinyCert_nameConstr` both run to 2036. This is benign today: `wolfcrypt/src/asn.c:22592` and `:22604` gate `CheckDate()` on `(verify != NO_VERIFY)`, and the test passes `NO_VERIFY`, so the expiry is never consulted. The PR is nonetheless what first puts this cert on CI, and a future change to `VERIFY`/`VERIFY_SKIP_DATE` would turn it into a dated failure. Regenerating it with the same ten-year validity as the other two costs nothing.

**Suggestion:** Regenerate tinyCert_plain with -days 3650 so all three test certs share a 2036 notAfter.

Frauschi (commented):
## 🐺 Skoll Code Review

**Findings:** 7 total — 4 posted, 3 skipped

### Posted findings
- [Medium] **tinyCert_plain assertion proves nothing TINY-specific** — `tests/api.c:3856-3862`
- [Medium] **No CI coverage for any WOLFSSL_X509_TINY add-back macro** — `.github/workflows/os-check.yml:452-455`
- [Medium] **TINY coverage lost for cryptonly / embedded builds** — `tests/api.c:3854`
- [Low] **tinyCert_plain has a one-year validity while its siblings have ten** — `tests/api.c:3735-3745`

- [Medium] Critical-extension case duplicates existing coverage and misses the TINY-specific reject paths
- [Low] "check": false silently opts the config out of --private-dir=certs
- [Low] Test name does not follow the api.c wolfCrypt-test convention

</details>

aidangarske:
Jenkins retest this please

tests/api.c
```diff
@@ -3785,6 +3785,168 @@ static int test_wolfSSL_CertRsaPss(void)
...
+    0x30, 0x0A, 0x06, 0x08, 0x2A, 0x86, 0x48, 0xCE, 0x3D, 0x04, 0x03, 0x02,
+    0x03, 0x49, 0x00, 0x30, 0x46, 0x02, 0x21, 0x00, 0xF8, 0x72, 0x10, 0x93,
+    0x20, 0x76, 0xC8, 0x6C, 0x2A, 0x29, 0x30, 0x00, 0xC6, 0xB5, 0x90, 0x26,
+    0x95, 0x0A, 0xD0, 0x41, 0xB1, 0x45, 0x3A, 0xBC, 0x1C, 0xDE, 0x9D, 0x61,
+    0x5E, 0x07, 0xC9, 0x8F, 0x02, 0x21, 0x00, 0xD6, 0x1E, 0x4A, 0xDF, 0x78,
+    0x33, 0x1B, 0xC3, 0x7E, 0x9D, 0x5F, 0x4D, 0xCA, 0x99, 0x0B, 0xCE, 0xBC,
+    0xC2, 0xC0, 0x1E, 0xA5, 0x2E, 0xD4, 0x69, 0x74, 0x84, 0x3A, 0xAF, 0x15,
+    0xFF, 0xB0, 0xEB,
+};
+static const byte tinyCert_critUnknown[] = {
```
Frauschi: **🟡 [Medium] Critical-unknown-extension case asserts behavior that is identical in every build, so it adds no TINY coverage**
*💡 SUGGEST*  `test`

`tinyCert_critUnknown` carries a critical private extension `1.3.6.1.4.1.99999.1` (confirmed by decoding the array with `openssl x509 -text`). That OID is unrecognized in *every* wolfSSL configuration, so it lands in the generic `default:` arm of `DecodeCertExtension` (`wolfcrypt/src/asn.c:22064-22077`) and yields `ASN_CRIT_EXT_E` with or without `WOLFSSL_X509_TINY` and with or without any add-back. The assertion therefore behaves identically in the `x509-tiny` job, the `x509-tiny-addbacks` job, and a normal build - it exercises nothing the TINY profile changed. Meanwhile the arms that *are* TINY-specific, the `#elif defined(WOLFSSL_X509_TINY)` stubs that fail closed on a critical AIA/AKI/SKI even under `WOLFSSL_ALLOW_CRIT_AIA` / `_AKID` / `_SKID` (`wolfcrypt/src/asn.c:21838-21846`, `21876-21884`, `21908-21916`), have no coverage at all in either new CI job.

**Suggestion:**
```suggestion
/* A cert with a *critical* SKI (or AKI) exercises the TINY-only stub:
 * a full build with WOLFSSL_ALLOW_CRIT_SKID accepts it, TINY must not. */
```

**Recommendation:** Replace or supplement the private-OID fixture with a cert whose critical extension is one TINY strips (critical SKI/AKI, or critical certificatePolicies with `WOLFSSL_X509_TINY_POLICIES` off), so the assertion actually distinguishes the TINY profile from a full build.

Frauschi: **🟡 [Medium] tinyCert_critUnknown duplicates critUnknownExtCert byte for byte**
*💡 SUGGEST*  `convention`

The 415-byte `tinyCert_critUnknown` array added to `tests/api.c` is byte-for-byte identical to `critUnknownExtCert` in `wolfcrypt/test/test.c:30806-30842` (verified by extracting and comparing both arrays: 415 == 415, identical). The assertion around it also duplicates `cert_crit_unknown_ext_test()` (`wolfcrypt/test/test.c:30903`), which already asserts `ParseCert() == ASN_CRIT_EXT_E` on the same bytes. The PR therefore ships a second copy of a fixture and a second copy of a check, with no behavioral difference between them (see the previous finding). Two copies drift independently the next time the encoding needs a tweak.

**Recommendation:** Drop the duplicate fixture (the generic critical-unknown case is already covered in test.c), or, if TINY-specific critical-extension coverage is the goal, generate a distinct cert whose critical extension is one TINY strips.

tests/api.c
```diff
@@ -3785,6 +3785,168 @@ static int test_wolfSSL_CertRsaPss(void)
...
+#endif
+
+/* WOLFSSL_X509_TINY: a normal cert parses; a critical-unknown extension and
+ * nameConstraints fail closed unless the add-back is set. */
+static int test_wolfSSL_X509_tiny(void)
+{
+    EXPECT_DECLS;
+    DecodedCert cert;
+
+    wc_InitDecodedCert(&cert, tinyCert_plain, (word32)sizeof(tinyCert_plain),
```
Frauschi: **🟡 [Medium] Plain-cert case only asserts SKI/AKI bits; retained extensions and the ignore-non-critical path are unchecked**
*💡 SUGGEST*  `test`

`tinyCert_plain` carries basicConstraints (critical, CA:TRUE), SAN (DNS a.test, www.a.test), keyUsage and extendedKeyUsage - all extensions the TINY profile is supposed to keep decoding - but the test only checks the return code plus `extSubjKeyIdSet` / `extAuthKeyIdSet`. A regression that made TINY silently stop populating `altNames`, `extKeyUsage`, `extExtKeyUsage` or `isCA` would still return 0 and still pass. The other half of the profile's contract is also untested: a *non-critical* stripped extension (e.g. a non-critical CRL distribution point with `WOLFSSL_X509_TINY_CRL_DP` off) must be ignored and the parse must succeed - no fixture covers that, so the profile's "ignore quietly when non-critical, fail closed when critical" split is only half-tested.

**Suggestion:**
```suggestion
ExpectIntEQ(wc_ParseCert(&cert, CERT_TYPE, NO_VERIFY, NULL), 0);
    /* Extensions TINY keeps must still decode. */
    ExpectIntEQ(cert.isCA, 1);
    ExpectIntEQ(cert.extBasicConstSet, 1);
    ExpectIntEQ(cert.extSubjAltNameSet, 1);
    ExpectIntEQ(cert.extKeyUsageSet, 1);
    ExpectIntEQ(cert.extExtKeyUsageSet, 1);
    ExpectNotNull(cert.altNames);
```

**Recommendation:** Assert the retained-extension bits on `tinyCert_plain`, and add a fixture with a non-critical stripped extension to prove TINY ignores it rather than erroring.

Frauschi (commented):
## 🐺 Skoll Code Review

**Findings:** 6 total — 5 posted, 1 skipped

### Posted findings
- [Medium] **Critical-unknown-extension case asserts behavior that is identical in every build, so it adds no TINY coverage** — `tests/api.c:3830-3866,3930-3934`
- [Medium] **tinyCert_critUnknown duplicates critUnknownExtCert byte for byte** — `tests/api.c:3830-3866`
- [Medium] **Plain-cert case only asserts SKI/AKI bits; retained extensions and the ignore-non-critical path are unchecked** — `tests/api.c:3914-3928`
- [Low] **x509-tiny-addbacks comment overstates what the config builds** — `.github/configs/os-check-linux.json:333-337`
- [Low] **tinyCert_plain fixture is a CA cert whose keyUsage lacks keyCertSign** — `tests/api.c:3790-3828`

- [Medium] Moving the test out of wolfcrypt/test/test.c drops it from cryptonly and no-example builds

</details>

.github/configs/os-check-linux.json
```diff
@@ -325,5 +325,14 @@
 {"comment": "wolfSSL <= 5.9.0 DTLS 1.3 legacy_session_id echo compatibility shim",
  "name": "dtls13-echo-legacy-session-id",
  "configure": ["--enable-dtls", "--enable-dtls13", "--enable-session-ticket",
-  "--enable-ecc", "CPPFLAGS=-DWOLFSSL_DTLS13_ECHO_LEGACY_SESSION_ID"]}
+  "--enable-ecc", "CPPFLAGS=-DWOLFSSL_DTLS13_ECHO_LEGACY_SESSION_ID"]},
+{"name": "x509-tiny",
+ "comment": "Minimal-extension X.509 profile; runs just its unit test.",
+ "configure": ["--enable-ecc", "CPPFLAGS=-DWOLFSSL_X509_TINY"],
+ "check": false, "run": [["tests/unit.test", "-test_wolfSSL_X509_tiny"]]},
+{"name": "x509-tiny-addbacks",
```
Frauschi, line 333: **🔵 [Low] x509-tiny-addbacks comment overstates what the config builds**
*🔧 NIT*  `style`

The comment says the entry "builds the settings.h implication chain", but only one of the three implications actually fires. `WOLFSSL_X509_TINY_NAME_CONSTRAINTS` -> `SAN_EMAIL`/`SAN_IP`/`SAN_DIR`/`SAN_RID` (`wolfssl/wolfcrypt/settings.h:4869-4881`) does fire; the `HAVE_OCSP` -> `WOLFSSL_X509_TINY_AIA` and `HAVE_CRL_IO` -> `WOLFSSL_X509_TINY_CRL_DP` implications (`settings.h:4886-4893`) are `#ifndef`-guarded no-ops here, because the config already defines `-DWOLFSSL_X509_TINY_AIA` and `-DWOLFSSL_X509_TINY_CRL_DP` on the command line and enables neither OCSP nor CRL_IO. Also worth noting: none of AIA, CRL_DP or POLICIES is exercised by any of the three fixtures, so for those add-backs this job is compile-only coverage. (Verification note: the sandbox in this session denied `configure`/`make`, so the two new configs' compilability was checked by reading the guards in `wolfcrypt/src/asn.c` and `wolfssl/wolfcrypt/asn.h`, not by building. The AIA/CRL-DP decoders are gated purely on the TINY macros and their `DecodedCert` fields are unconditional, so the build should be clean.)

**Suggestion:**
```suggestion
{"name": "x509-tiny-addbacks",
 "comment": "TINY with every add-back defined directly: compiles all add-back decoders and takes the NAME_CONSTRAINTS/SKI/AKI branches of the unit test. Only the NAME_CONSTRAINTS -> SAN_* settings.h implication fires here.",
```

**Recommendation:** Reword the comment, or add `--enable-ocsp --enable-crl` (dropping the explicit `-DWOLFSSL_X509_TINY_AIA -DWOLFSSL_X509_TINY_CRL_DP`) so the settings.h implications are what the job actually builds.

tests/api.c
```diff
@@ -3785,6 +3785,168 @@ static int test_wolfSSL_CertRsaPss(void)
 #endif /* WOLFSSL_TEST_APPLE_NATIVE_CERT_VALIDATION */
 #endif /* HAVE_CERT_CHAIN_VALIDATION */
 
+#if defined(WOLFSSL_X509_TINY) && defined(HAVE_ECC) && \
+    !defined(NO_ECC256) && !defined(NO_ECC_SECP)
+static const byte tinyCert_plain[] = {
```
Frauschi: **🔵 [Low] tinyCert_plain fixture is a CA cert whose keyUsage lacks keyCertSign**
*🔧 NIT*  `test`

Decoding the regenerated `tinyCert_plain` shows basicConstraints critical CA:TRUE combined with keyUsage = digitalSignature only and EKU = serverAuth - an internally inconsistent cert (RFC 5280 4.2.1.3: a cert asserting cA should assert keyCertSign). wolfSSL only enforces the inverse today (`!isCA && keyCertSign` -> `KEYUSAGE_E`, `wolfcrypt/src/asn.c:24346-24356`), and the test parses with `NO_VERIFY`, so nothing fails now. It is still a fragile fixture: any future CA/keyUsage consistency check, or a switch of this test to `VERIFY`, invalidates it. The self-signature itself is valid on all three fixtures (verified with `openssl dgst -sha256 -verify` against each cert's own public key).

**Recommendation:** When the fixture is next regenerated, either add keyCertSign to the CA cert's keyUsage or make it a plain end-entity cert (CA:FALSE) so the profile it encodes is self-consistent.

aidangarske:
Jenkins retest this please

dgarske (approved):

Frauschi (approved):

PR #10976, opened 2026-07-22
GPLv2 license exception for Cisco Systems, Inc. use of wolfCrypt in U-Boot only

dgarske:
Pairs with https://github.com/wolfSSL/wolfTPM/pull/557


toddouska (approved):
Looks good.

PR #10978, opened 2026-07-22
Add Enable/DisableNormalMasterSecret APIs to enforce EMS requirement at runtime.  Disable server EMS when DisableExtendedMasterSecret is called, not just client EMS.

kareem-wolfssl:
# Description

Fixes zd#21980

# Testing

Built-in tests plus added tests

# Checklist

 - [X] added tests
 - [X] updated/added doxygen


kareem-wolfssl:
Retest this please

wolfssl/ssl.h
```diff
@@ -5076,6 +5076,10 @@ WOLFSSL_API int wolfSSL_CTX_set_num_tickets(WOLFSSL_CTX* ctx, size_t mxTickets);
 /* TLS Extended Master Secret Extension */
 WOLFSSL_API int wolfSSL_DisableExtendedMasterSecret(WOLFSSL* ssl);
 WOLFSSL_API int wolfSSL_CTX_DisableExtendedMasterSecret(WOLFSSL_CTX* ctx);
+WOLFSSL_API int wolfSSL_DisableNormalMasterSecret(WOLFSSL* ssl);
```
dgarske: Consider inline documentation for what "Normal" is or perhaps a better API name? I think its the "standard (non-extended)"

src/ssl_api_ext.c
```diff
@@ -1307,11 +1317,116 @@ int wolfSSL_DisableExtendedMasterSecret(WOLFSSL* ssl)
...
     ssl->options.haveEMS = 0;
+    ssl->options.disableEMS = 1;
+    /* Disabling EMS is mutually exclusive with requiring it. */
+    ssl->options.requireEMS = 0;
+
+    return WOLFSSL_SUCCESS;
+}
+
+
+/* Disable the standard (non-extended) master secret on the context.
```
dgarske: Consider putting the doxy in`doc/dox_comments/header_files/ssl.h` so they are actually in the user manual

dgarske (changes requested):

src/tls.c
```diff
@@ -705,6 +705,19 @@ int MakeTlsMasterSecret(WOLFSSL* ssl)
 {
     int ret;
 
+#ifdef HAVE_EXTENDED_MASTER
```
dgarske: **🔴 [High] requireEMS is never enforced on a TLS 1.2 abbreviated (resumption) handshake**

The new `requireEMS` check is placed at the top of `MakeTlsMasterSecret()`, which is the *only* place in the library that reads `ssl->options.requireEMS` (verified by grep: every other occurrence is a write). `MakeMasterSecret()` (`src/keys.c:4232`) is reached from exactly two handshake sites: `SendClientKeyExchange` (`src/internal.c:35970`) and `DoClientKeyExchange` (`src/internal.c:43810`). Both are full-handshake-only. On an abbreviated handshake the master secret is copied from the cached session and keys are derived directly, so the check is never executed.

**Server bypass** — with `wolfSSL_CTX_DisableNormalMasterSecret(ctx)` set, a client that sends *no* `extended_master_secret` extension but presents a valid session ID / ticket for a cached non-EMS session resumes successfully. `ssl->options.haveEMS` stays 0 (the only setters are `src/tls.c:18306` and `src/internal.c:40103`, both inside `extId == HELLO_EXT_EXTMS`). `HandleTlsResumption` at `src/internal.c:39419` evaluates `session->haveEMS (0) != ssl->options.haveEMS (0)` as false, skips both RFC 7627 5.3 branches, falls into the success `else` at `src/internal.c:39439`, and derives keys at `src/internal.c:39481-39495`. The server that was configured to *require* EMS silently completes a connection on a legacy, triple-handshake-vulnerable master secret.

**Client bypass** — `wolfSSL_DisableNormalMasterSecret(ssl)` sets `requireEMS = 1, haveEMS = 1`, but `wolfSSL_SetSession` then unconditionally clobbers it at `src/ssl_sess.c:1610` (`ssl->options.haveEMS = (ssl->session->haveEMS) ? 1 : 0;`). With a cached non-EMS…

**Fix:** Add the `requireEMS` gate to the resumption paths as well: in `HandleTlsResumption` (src/internal.c ~39419, server) and in `CheckResumptionConsistency` (src/internal.c ~23763, client), and stop `wolfSSL_SetSession` from clearing `haveEMS` when `requireEMS` is set. Add a handshake test that requires EMS and then attempts to resume a cached non-EMS session on both sides.

src/tls.c
```diff
@@ -18285,11 +18298,15 @@ WOLFSSL_TEST_VIS int TLSX_Parse(WOLFSSL* ssl, const byte* input, word16 length,
                 if (size != 0)
                     return BUFFER_ERROR;
 
+                /* Honor a user request to disable EMS by ignoring the peer's
```
dgarske, line 18886: **🔴 [High] disableEMS honored only on parse, not on advertise - client advertises EMS then discards the server's echo, breaking…**

The PR consults the new `disableEMS` flag only in the *receive* path (`src/tls.c:18303` and `src/internal.c:40102`). The *send* path is still gated purely on `ssl->options.haveEMS` (size: `src/tls.c:17186`, write: `src/tls.c:17410`). Two code paths unconditionally re-arm `haveEMS = 1` on a client after the user has called `wolfSSL_DisableExtendedMasterSecret()`:

1. `InitSSL_Side()` at `src/internal.c:2359-2369` sets `ssl->options.haveEMS = 1` for any client. It is called from `wolfSSL_connect()` when side is `WOLFSSL_NEITHER_END` (`src/ssl_api_hs.c:98`) and *unconditionally* from `wolfSSL_set_connect_state()` (`src/ssl_api_hs.c:1346`). Both are inside `OPENSSL_EXTRA || WOLFSSL_EITHER_SIDE` / `OPENSSL_EXTRA || WOLFSSL_WPAS_SMALL` - the common OpenSSL-compat build.
2. `wolfSSL_SetSession()` at `src/ssl_sess.c:1610`, when the cached session used EMS.

Resulting sequence (no cached session needed):
```
ctx = wolfSSL_CTX_new(wolfSSLv23_method());   /* NEITHER_END */
ssl = wolfSSL_new(ctx);
wolfSSL_DisableExtendedMasterSecret(ssl);      /* haveEMS=0, disableEMS=1 */
wolfSSL_connect(ssl);                          /* InitSSL_Side -> haveEMS=1  */
```
The ClientHello now carries `extended_master_secret`. The server echoes it in ServerHello. The new guard makes `TLSX_Parse` skip `pendingEMS = 1`, so `src/tls.c:18767` (`else if (!isRequest && ssl->options.haveEMS && !pendingEMS) ssl->options.haveEMS = 0;`) clears `haveEMS` back to 0. The client then takes the non-EMS branch at `src/tls.c:736-737` and derives a **normal** master secret while the server derives an **extended** one ->…

**Fix:** Make `disableEMS` authoritative in the advertise path too, and stop `InitSSL_Side()` (src/internal.c:2359) and `wolfSSL_SetSession()` (src/ssl_sess.c:1610) from re-arming `haveEMS` when `disableEMS` is set. Add a regression test that calls `wolfSSL_DisableExtendedMasterSecret()` followed by `wolfSSL_set_connect_state()` / `wolfSSL_connect()` on a `wolfSSLv23_method()` object and asserts the handshake still completes (without EMS).

tests/api/test_tls_ext.c
```diff
@@ -803,6 +928,57 @@ int test_wolfSSL_DisableExtendedMasterSecret(void)
...
+    ExpectNotNull(ssl = wolfSSL_new(ctx));
+
+    /* error cases */
+    ExpectIntNE(WOLFSSL_SUCCESS, wolfSSL_CTX_DisableNormalMasterSecret(NULL));
+    ExpectIntNE(WOLFSSL_SUCCESS, wolfSSL_DisableNormalMasterSecret(NULL));
+    ExpectIntNE(WOLFSSL_SUCCESS, wolfSSL_CTX_EnableNormalMasterSecret(NULL));
+    ExpectIntNE(WOLFSSL_SUCCESS, wolfSSL_EnableNormalMasterSecret(NULL));
+
+    /* success cases */
+    ExpectIntEQ(WOLFSSL_SUCCESS, wolfSSL_CTX_DisableNormalMasterSecret(ctx));
```
dgarske: **🟠 [Medium] test_wolfSSL_DisableNormalMasterSecret assertions are vacuous - the (re)enable behavior is never actually exercised**

The test comment says "Requiring EMS must (re)enable advertising it on a client", but both asserted values are already 1 *before* the call: `InitSSL_Ctx` sets `ctx->haveEMS = 1` for a client method at `src/internal.c:2861-2866`, and `InitSSL` copies it into `ssl->options.haveEMS` at `src/internal.c:8168`. So `ExpectIntEQ(ctx->haveEMS, 1)` and `ExpectIntEQ(ssl->options.haveEMS, 1)` would pass even if the `if (side == WOLFSSL_CLIENT_END) haveEMS = 1;` lines were deleted from `wolfSSL_CTX_DisableNormalMasterSecret` / `wolfSSL_DisableNormalMasterSecret`. The most subtle part of the new code - the "undo a previous disable" logic - has zero real coverage.

The same applies to the later block at lines 961-966: `ctx->haveEMS` / `ssl->options.haveEMS` are asserted to still be 1, but they were never driven to 0.

**Fix:** Disable EMS first so `haveEMS` is genuinely 0, then assert `DisableNormalMasterSecret` restores it to 1 and clears `disableEMS`.

tests/api/test_tls_ext.c
```diff
@@ -313,6 +313,131 @@ int test_tls_ems_resumption_server_downgrade(void)
...
+    wolfSSL_CTX_free(ctx_s);
+#endif
+    return EXPECT_RESULT();
+}
+
+
+#if !defined(WOLFSSL_NO_TLS12) && defined(HAVE_EXTENDED_MASTER) && \
+        !defined(NO_WOLFSSL_CLIENT) && !defined(NO_WOLFSSL_SERVER) && \
+        defined(HAVE_MANUAL_MEMIO_TESTS_DEPENDENCIES)
+/* Exercise wolfSSL_DisableNormalMasterSecret. When serverSide is set the
```
dgarske: **🟠 [Medium] No test coverage for the CTX-level APIs in a handshake, resumption, or the non-client method branch**

The new handshake tests are a good start, but three significant gaps remain in code the PR introduces:

1. **Resumption** - not covered at all. This is where the enforcement is missing entirely (see the resumption-bypass finding). A test that requires EMS and then resumes a cached non-EMS session would fail today.
2. **CTX-level APIs in a handshake** - `wolfSSL_CTX_DisableNormalMasterSecret` / `wolfSSL_CTX_EnableNormalMasterSecret` / `wolfSSL_CTX_DisableExtendedMasterSecret` are only exercised for flag state on a fresh CTX (`test_wolfSSL_DisableNormalMasterSecret`, no handshake). The new `InitSSL` propagation at `src/internal.c:8169-8170` is never verified end-to-end.
3. **Non-client method branch** - `if (ctx->method != NULL && ctx->method->side == WOLFSSL_CLIENT_END)` at `src/ssl_api_ext.c:1351` is only ever entered with a client method. The server-method and `wolfSSLv23_method()` (`WOLFSSL_NEITHER_END`) branches are untested.
4. **`wolfSSL_set_connect_state()` / `wolfSSL_connect()` on a NEITHER_END object after a disable** - the regression path described in the `disableEMS`-advertise finding. `test_memio_setup` always uses explicit `wolfTLSv1_2_client_method` / `wolfTLSv1_2_server_method`, so `InitSSL_Side` is never re-run and the break is invisible to the suite.

**Fix:** Add (a) a resumption test that proves `requireEMS` is enforced on abbreviated handshakes, (b) a CTX-level handshake test proving `InitSSL` propagation works, and (c) a `wolfSSLv23_method()` + `wolfSSL_set_connect_state()` case covering the `WOLFSSL_NEITHER_END` branch.

src/ssl_api_ext.c
```diff
@@ -1307,11 +1317,116 @@ int wolfSSL_DisableExtendedMasterSecret(WOLFSSL* ssl)
...
+ */
+int wolfSSL_EnableNormalMasterSecret(WOLFSSL* ssl)
+{
+    if (ssl == NULL)
+        return BAD_FUNC_ARG;
+
+    ssl->options.requireEMS = 0;
 
     return WOLFSSL_SUCCESS;
 }
```
dgarske, line 1787: **🟠 [Medium] disableEMS is a one-way latch with no matching Enable API**

The PR introduces an asymmetric four-function API. `wolfSSL_EnableNormalMasterSecret` / `wolfSSL_CTX_EnableNormalMasterSecret` clear only `requireEMS`; they deliberately leave `disableEMS` alone. There is no `wolfSSL_EnableExtendedMasterSecret` anywhere in the tree (grep confirms zero matches), and nothing else in the library clears `disableEMS`. The net result is that once `wolfSSL_DisableExtendedMasterSecret()` is called, the *only* way to undo it is to call `wolfSSL_DisableNormalMasterSecret()` - a function whose name says the opposite of "re-enable EMS" and which additionally forces `requireEMS = 1`. A caller who just wants to return to the default "EMS if the peer supports it" state has no way to express it.

This is worth resolving now while the API is still unreleased. Two state bits with three legal combinations, mutually-exclusive cross-clearing, and a name that reads as a double negative ("disable normal" == "require extended") is easy to misuse. A single tri-state setter would be materially harder to get wrong.

**Fix:** Add `wolfSSL_EnableExtendedMasterSecret()` / `wolfSSL_CTX_EnableExtendedMasterSecret()` so `disableEMS` can be cleared without side effects, or replace the four-function pair with a single tri-state mode setter before this API ships.

src/ssl_api_ext.c
```diff
@@ -1307,11 +1317,116 @@ int wolfSSL_DisableExtendedMasterSecret(WOLFSSL* ssl)
...
+ */
+int wolfSSL_CTX_DisableNormalMasterSecret(WOLFSSL_CTX* ctx)
+{
+    if (ctx == NULL)
+        return BAD_FUNC_ARG;
+
+    ctx->requireEMS = 1;
+    /* Requiring EMS is mutually exclusive with disabling it. */
+    ctx->disableEMS = 0;
+    /* A client must advertise the extension for it to be negotiated. Undo any
```
dgarske: **🔵 [Low] wolfSSL_CTX_DisableNormalMasterSecret's re-enable branch silently skipped for WOLFSSL_NEITHER_END methods**

The comment states "Undo any previous disable so the extension is offered", but the branch is gated on `ctx->method->side == WOLFSSL_CLIENT_END`. For a `wolfSSLv23_method()` CTX the side is `WOLFSSL_NEITHER_END`, so `ctx->haveEMS` is left at 0 and the stated undo does not happen.

I traced this through and it is masked in practice: `wolfSSLv23_method()` only exists under `OPENSSL_EXTRA || WOLFSSL_EITHER_SIDE` (`src/ssl.c:3414`), which is the same guard as the `InitSSL_Side(ssl, WOLFSSL_CLIENT_END)` call in `wolfSSL_connect()` (`src/ssl_api_hs.c:96-104`), and `InitSSL_Side` sets `haveEMS = 1` for clients at `src/internal.c:2359-2369`. So the state is recovered at connect time. Flagging only because the comment promises something the code does not do on its own, and the correctness depends on a non-obvious coupling to a different file.

**Fix:** Either widen the condition to `!= WOLFSSL_SERVER_END` so it is self-contained, or extend the comment to record that the `WOLFSSL_NEITHER_END` case is handled later by `InitSSL_Side()`.

dgarske (commented):
## Skoll Code Review

**Findings:** 11 total — 7 posted, 4 skipped

### Posted findings
- [High] **requireEMS is never enforced on a TLS 1.2 abbreviated (resumption) handshake** — `src/tls.c:708-719`
- [High] **disableEMS honored only on parse, not on advertise - client advertises EMS then discards the server's echo, breaking the handshake** — `src/tls.c:18301-18309`
- [Medium] **test_wolfSSL_DisableNormalMasterSecret assertions are vacuous - the (re)enable behavior is never actually exercised** — `tests/api/test_tls_ext.c:950-957`
- [Medium] **No test coverage for the CTX-level APIs in a handshake, resumption, or the non-client method branch** — `tests/api/test_tls_ext.c:360-437`
- [Medium] **disableEMS is a one-way latch with no matching Enable API** — `src/ssl_api_ext.c:1290-1428`
- [Low] **wolfSSL_CTX_DisableNormalMasterSecret's re-enable branch silently skipped for WOLFSSL_NEITHER_END methods** — `src/ssl_api_ext.c:1348-1352`

#### disableEMS semantics differ between HAVE_TLS_EXTENSIONS and non-TLSX builds (client ServerHello parse not guarded)
**File:** `src/internal.c:33330-33331`
**Function:** `DoServerHello`
**Severity:** Medium

The PR added the `!ssl->options.disableEMS` guard to the server's non-TLSX ClientHello parse (`src/internal.c:40102`) but left the matching client-side non-TLSX ServerHello parse untouched:
```c
                        if (extId == HELLO_EXT_EXTMS)
                            pendingEMS = 1;
```
In a `!HAVE_TLS_EXTENSIONS` build, a client with `disableEMS = 1` whose `haveEMS` has been re-armed (see the `InitSSL_Side` / `wolfSSL_SetSession` finding) will set `pendingEMS = 1`, so `src/internal.c:33346-33347` leaves `haveEMS = 1` and the connection quietly uses EMS. The identical configuration in a `HAVE_TLS_EXTENSIONS` build hard-fails the handshake. Two build configurations give opposite outcomes for the same API call, which makes the disable semantics unpredictable and untestable.

**Recommendation:** Apply the same `!ssl->options.disableEMS` guard to the non-TLSX client ServerHello parse so the flag behaves identically in both build configurations.

_Referenced code: `src/internal.c:33330-33338` (9 lines)_

---

### Skipped findings
- [Medium] ~~requireEMS is bypassed for SSLv3 and for the session-secret (EAP-FAST) callback path~~
- [Low] ~~SNI CTX switch does not propagate disableEMS/requireEMS, and cannot express a per-SNI EMS policy~~
- [Info] ~~Stray double blank line in new test function~~
- [Info] ~~Preprocessor directive indentation inconsistent with surrounding block in TLSX_Parse~~

src/internal.c
```diff
@@ -40576,6 +40599,18 @@ static int AddPSKtoPreMasterSecret(WOLFSSL* ssl)
         }
 #endif /* !WOLFSSL_NO_TICKET_EXPIRE && !NO_ASN_TIME */
 
+#ifdef HAVE_EXTENDED_MASTER
+        else if (ssl->options.requireEMS && !ssl->options.haveEMS) {
```
dgarske: This `else if` has fragile dependencies on earlier options.

Recommendation: Do not chain a security check off a conditionally-compiled `if`. Make the requireEMS test a self-contained `if` with its own early return, placed before the `#if !defined(WOLFSSL_NO_TICKET_EXPIRE) && !defined(NO_ASN_TIME)` block. While there, consider fixing the pre-existing dangling `else` so the whole chain no longer depends on WOLFSSL_NO_TICKET_EXPIRE / NO_ASN_TIME / HAVE_SNI / HAVE_ALPN. Add `--enable-all CFLAGS=-DWOLFSSL_NO_TICKET_EXPIRE` to the config test matrix so this class of failure is caught.

src/ssl_sess.c
```diff
@@ -1607,7 +1607,17 @@ int wolfSSL_SetSession(WOLFSSL* ssl, WOLFSSL_SESSION* session)
 #endif
     }
     ssl->options.resuming = 1;
-    ssl->options.haveEMS = (ssl->session->haveEMS) ? 1 : 0;
+#ifdef HAVE_EXTENDED_MASTER
+    /* A user EMS override takes precedence over the session's EMS state. */
+    if (ssl->options.requireEMS)
+        ssl->options.haveEMS = 1;
```
dgarske, line 1613: I think this needs to have `side == WOLFSSL_CLIENT_END` check added like done elsewhere.

src/tls.c
```diff
@@ -704,6 +704,19 @@ int MakeTlsMasterSecret(WOLFSSL* ssl)
 {
     int ret;
 
+#ifdef HAVE_EXTENDED_MASTER
+    /* The user disabled the standard master secret and requires the Extended
+     * Master Secret extension (RFC 7627). If it was not negotiated with the
+     * peer, abort rather than derive a standard master secret. Only reachable
+     * for TLS 1.2 and earlier; TLS 1.3 uses a separate key schedule. */
+    if (ssl->options.requireEMS && !ssl->options.haveEMS) {
```
dgarske: Recommendation: Move the enforcement from `MakeTlsMasterSecret` up into `MakeMasterSecret` so the SSLv3 path is covered too, or add an equivalent guard in `MakeSslMasterSecret`.

src/ssl_api_ext.c
```diff
@@ -1362,11 +1372,121 @@ int wolfSSL_DisableExtendedMasterSecret(WOLFSSL* ssl)
...
+ *
+ * If EMS (RFC 7627) is not negotiated the connection is aborted with
+ * EXT_MASTER_SECRET_NEEDED_E instead of deriving a standard master secret.
+ * TLS 1.3 has its own key schedule and is unaffected.
+ *
+ * @param [in] ctx  SSL/TLS context object.
+ * @return  WOLFSSL_SUCCESS on success.
+ * @return  BAD_FUNC_ARG when ctx is NULL.
+ */
+int wolfSSL_CTX_RequireExtendedMasterSecret(WOLFSSL_CTX* ctx)
```
dgarske, line 1727: Should this be wrapped with `#ifndef WOLFSSL_NO_TLS12`?

src/internal.c
```diff
@@ -41257,7 +41292,9 @@ static int AddPSKtoPreMasterSecret(WOLFSSL* ssl)
                         i += hashSigAlgoSz;
                     }
 #ifdef HAVE_EXTENDED_MASTER
-                    else if (extId == HELLO_EXT_EXTMS)
+                    /* Honor a user request to disable EMS on the server by
+                     * ignoring the peer's extension. */
+                    else if (extId == HELLO_EXT_EXTMS && !ssl->options.disableEMS)
```
dgarske: Check 80 line length.

src/internal.c
```diff
@@ -24487,24 +24490,36 @@ static void DropAndRestartProcessReply(WOLFSSL* ssl)
...
+
     if (ssl->session == NULL) /* nothing to compare against */
         return 0;
-    /* EMS must match (RFC 7627 5.3); skip EAP-FAST (session-secret callback). */
-    if (
 #ifdef HAVE_SECRET_CALLBACK
-        !(ssl->sessionSecretCb != NULL
+    /* Skip the EMS checks for EAP-FAST (session-secret callback): the master
+     * secret comes from the callback rather than the cached session. */
+    skipEmsCheck = ssl->sessionSecretCb != NULL
```
dgarske: Add an explicit `? 1 : 0` so the narrowing is intentional and portable to stricter warning configurations.

dgarske (changes requested):

src/internal.c
```diff
@@ -40626,7 +40662,12 @@ static int AddPSKtoPreMasterSecret(WOLFSSL* ssl)
         }
 #endif /* !WOLFSSL_NO_TICKET_EXPIRE && !NO_ASN_TIME */
 
-        else if (session->haveEMS != ssl->options.haveEMS) {
+        if (!ssl->options.resuming) {
```
dgarske, line 41140: **🔵 [Low] "Expired above" comment is wrong when the ticket-expiry check is compiled out**

The restructure that replaced the `else if` chain is a genuine improvement — I confirmed `--enable-all CPPFLAGS=-DWOLFSSL_NO_TICKET_EXPIRE` now builds and passes the full unit suite, whereas before the `else if` at the old line 40662 bound to whatever `if` preceded the removed `#if` block. The added CI entry is well justified.

The comment on the replacement guard is only accurate in the default configuration:

```c
        if (!ssl->options.resuming) {
            /* Expired above: DoClientHello falls back to a full handshake. */
            return ret;
        }
```

When `WOLFSSL_NO_TICKET_EXPIRE` or `NO_ASN_TIME` is defined there is nothing "above" — the preceding `#if` block is gone, `ssl->options.resuming` is provably 1 at this point, and the guard is dead. Stating the invariant instead of the cause keeps the comment true in every build.

Also note the new CI entry only covers `WOLFSSL_NO_TICKET_EXPIRE`; `NO_ASN_TIME` reaches the identical control flow through the same `#if` and remains uncovered by that job.

**Fix:** Reword the comment to describe the invariant rather than the (config-dependent) cause.

dgarske (commented):
## Skoll Code Review

**Findings:** 8 total — 5 posted, 3 skipped
**4 finding(s) not tied to a diff line** (full detail below)

### Posted findings
- [Low] **"Expired above" comment is wrong when the ticket-expiry check is compiled out** — `src/internal.c:40665-40668`

#### Client enforces requireEMS only after the ClientKeyExchange is already on the wire
**File:** `src/keys.c:4232-4245 (enforcement) / src/internal.c:37182 (call site)`
**Function:** `MakeMasterSecret / SendClientKeyExchange`
**Severity:** Medium

The new `requireEMS` gate lives at the top of `MakeMasterSecret()`, but on the client `MakeMasterSecret()` is only reached *after* the ClientKeyExchange has been fully built, appended to `ssl->buffers.outputBuffer`, and flushed by `SendBuffered()`:

```c
ssl->buffers.outputBuffer.length += (word32)args->sendSz;
if (!ssl->options.groupMessages) {
    ret = SendBuffered(ssl);
}
if (ret == 0 || ret == WC_NO_ERR_TRACE(WANT_WRITE)) {
    int tmpRet = MakeMasterSecret(ssl);
```

So a client that called `wolfSSL_RequireExtendedMasterSecret()` still performs the full key-exchange operation (RSA encrypt of the premaster, or ECDHE public-key generation) and **transmits it to a server it is about to reject**, before aborting. `ssl->options.haveEMS` is already final much earlier — it is fixed the moment the ServerHello extensions are parsed (`TLSX_Parse()` clears it at src/tls.c:19225, and `DoServerHello` clears it at src/internal.c:34480 when no extensions are present). Enforcing there would abort before any key material is computed or sent.

Secondary effect of the same placement: the new early `return EXT_MASTER_SECRET_NEEDED_E;` bypasses `CleanPreMaster(ssl)` (src/keys.c:4065), which is the function that `ForceZero`s and frees `ssl->arrays->preMasterSecret`. The buffer is eventually zeroed by `FreeArrays()` (src/internal.c), so this is not a leak — but the premaster stays live in the heap for longer than on every other exit path from `MakeMasterSecret`.

**Recommendation:** Move (or duplicate) the client-side check to ServerHello processing so the abort happens before the key exchange is computed and transmitted, and call `CleanPreMaster(ssl)` on the new error path in `MakeMasterSecret()`.

_Referenced code: `src/keys.c:4232-4245 (enforcement) / src/internal.c:37182 (call site)` (12 lines)_

---

#### Server-side EMS disable turns pre-existing EMS sessions/tickets into fatal handshake failures instead of a…
**File:** `src/internal.c:40670-40688 (reached via new src/tls.c:18747 behavior)`
**Function:** `HandleTlsResumption / TLSX_Parse`
**Severity:** Medium

This PR makes `wolfSSL_(CTX_)DisableExtendedMasterSecret()` take effect on the server: `TLSX_Parse()` now ignores the client's `extended_master_secret` extension, so `ssl->options.haveEMS` stays 0 (src/tls.c:18747).

That interacts badly with the pre-existing RFC 7627 5.3 resumption rule further down `HandleTlsResumption()`:

```c
if (session->haveEMS != ssl->options.haveEMS) {
    ...
    /* if old sess used EMS, but new doesn't, MUST abort */
    else if (session->haveEMS && !ssl->options.haveEMS) {
        ...
        ret = EXT_MASTER_SECRET_NEEDED_E;
```

The abort rule exists to catch a *client* that drops EMS on resumption. But with server-side disable, `ssl->options.haveEMS == 0` is the server's own self-inflicted state, not what the client sent. A perfectly well-behaved, EMS-offering client that presents a session-cache entry or session ticket created before the disable (`it->haveEMS` is restored at src/internal.c:42853) now gets a **fatal handshake_failure**, when the correct/graceful response is to decline resumption and do a full handshake — the client did offer EMS.

Realistic trigger: a mixed fleet or a rolling config change. Server A (EMS on) issues tickets with `haveEMS=1`; server B shares ticket keys / session cache but has EMS disabled. Every client that lands on B with an outstanding ticket hard-fails until its ticket expires. This is a new availability regression, since before this PR `disableEMS` had no effect on the server and `haveEMS` always reflected the client's offer.

**Recommendation:** Fall back to a full handshake rather than a fatal alert when the EMS mismatch is caused by the local `disableEMS` setting, and document the ticket/session-cache implications of enabling server-side EMS disable on a live deployment.

_Referenced code: `src/internal.c:40670-40688 (reached via new src/tls.c:18747 behavior)` (13 lines)_

---

#### requireEMS silently unenforced on the session-secret-callback (EAP-FAST) path, contradicting the new doxygen
**File:** `src/internal.c:24541-24563 and src/internal.c:40564-40601`
**Function:** `CheckResumptionConsistency / HandleTlsResumption`
**Severity:** Medium

Both new resumption enforcement points are skipped whenever `HAVE_SECRET_CALLBACK` is active with a session ticket:

1. `CheckResumptionConsistency()` (client) computes `skipEmsCheck` from `ssl->sessionSecretCb != NULL && ssl->session->ticketLen > 0`, and the new require gate is written as `if (!skipEmsCheck && ssl->options.requireEMS && !ssl->options.haveEMS)` — so the requirement is dropped.
2. `HandleTlsResumption()` (server) returns from the `#ifdef HAVE_SECRET_CALLBACK` block at the top of the function (`WOLFSSL_LEAVE("HandleTlsResumption", ret); return ret;`) *before* ever reaching the new require check at line 40647.

Resumption also never calls `MakeMasterSecret()`, so there is no backstop. Net effect: with EAP-FAST, `wolfSSL_RequireExtendedMasterSecret()` is a no-op and the handshake completes.

That directly contradicts the doxygen added by this PR: *"Makes the TLS Extended Master Secret extension (RFC 7627) mandatory on the SSL object: if it is not negotiated, **including on resumption**, the connection is aborted with EXT_MASTER_SECRET_NEEDED_E."* A caller relying on that promise gets a silent downgrade. Skipping the *consistency* check for EAP-FAST is defensible (the master secret comes from the callback, not the cached session); silently skipping an explicit *user requirement* is a different decision and should be either enforced or documented.

**Recommendation:** Either enforce requireEMS on the EAP-FAST path or document the exemption in doc/dox_comments/header_files/ssl.h so the API contract matches the implementation.

_Referenced code: `src/internal.c:24541-24563 and src/internal.c:40564-40573` (10 lines)_

---

#### Enable/Require re-arm ctx-haveEMS without the version gate applied by InitSSL_Ctx
**File:** `src/ssl_api_ext.c:1643-1646 and src/ssl_api_ext.c:1706-1710`
**Function:** `wolfSSL_CTX_EnableExtendedMasterSecret / wolfSSL_CTX_RequireExtendedMasterSecret`
**Severity:** Low

The new re-arm is a bare side test:

```c
if (ctx->method != NULL && ctx->method->side == WOLFSSL_CLIENT_END)
    ctx->haveEMS = 1;
```

but `InitSSL_Ctx` (src/internal.c:2892-2902) applies a version gate as well:

```c
if (method->side == WOLFSSL_CLIENT_END) {
    if ((method->version.major == SSLv3_MAJOR) &&
         (method->version.minor >= TLSv1_MINOR)) {
        ctx->haveEMS = 1;
    }
#ifdef WOLFSSL_DTLS
    if (method->version.major == DTLS_MAJOR)
        ctx->haveEMS = 1;
#endif
}
```

For an SSLv3-only client method (`WOLFSSL_ALLOW_SSLV3`, `minor == SSLv3_MINOR == 0`) the default is `ctx->haveEMS == 0`, but `wolfSSL_CTX_EnableExtendedMasterSecret()` sets it to 1 — so "Enable" does not restore the documented default ("Re-enables ... (the default)"), it sets a state the library never produces on its own. The object-level `wolfSSL_EnableExtendedMasterSecret` / `wolfSSL_RequireExtendedMasterSecret` have the same divergence from `InitSSL_Side` (src/internal.c:2366-2375).

Impact is limited: `DoServerHello` clears `haveEMS` when the ServerHello carries no extensions (src/internal.c:34480), so an SSLv3 handshake still ends with `haveEMS == 0` and a `Require` still aborts correctly. It is a state-model inconsistency rather than a live bug.

**Recommendation:** Apply the same version/DTLS gate used by `InitSSL_Ctx` / `InitSSL_Side`, or add a comment explaining why the unconditional re-arm is safe.

_Referenced code: `src/ssl_api_ext.c:1643-1646 and src/ssl_api_ext.c:1706-1710` (5 lines)_

---

### Skipped findings
- [Medium] ~~Triplicated requireEMS abort block with inconsistent alert behavior - server resumption aborts with no alert in default builds~~
- [Medium] ~~No handshake-level coverage for CTX-level server disable, for the Enable APIs, or for DTLS~~
- [Low] ~~if/else chain split across a #endif in wolfSSL_SetSession~~

dgarske (dismissed):

dgarske (dismissed):

src/ssl_sess.c
```diff
@@ -1607,7 +1607,17 @@ int wolfSSL_SetSession(WOLFSSL* ssl, WOLFSSL_SESSION* session)
 #endif
     }
     ssl->options.resuming = 1;
-    ssl->options.haveEMS = (ssl->session->haveEMS) ? 1 : 0;
+#ifdef HAVE_EXTENDED_MASTER
+    /* A user EMS override takes precedence over the session's EMS state. */
+    if (ssl->options.requireEMS && ssl->options.side == WOLFSSL_CLIENT_END)
+        ssl->options.haveEMS = 1;
+    else if (ssl->options.disableEMS)
+        ssl->options.haveEMS = 0;
```
JacobBarthelmeh, line 1615: This leaves resuming = 1 while forcing haveEMS = 0, which sets up a resumption offer the peer is required to reject. A client that established a session with EMS, then calls wolfSSL_DisableExtendedMasterSecret() and wolfSSL_set_session(), will send a ClientHello carrying the EMS-bound session ID/ticket but without the  extended_master_secret extension (the extension is emitted purely on ssl->options.haveEMS, see src/tls.c:17606/17852). Per RFC 7627 5.3 item 1 the server MUST abort that, and our own server does: the new graceful-decline arm in HandleTlsResumption checks the server's disableEMS, so a client-side disable falls through to the EXT_MASTER_SECRET_NEEDED_E abort. Even against a lenient server, the client kills it itself in CheckResumptionConsistency() (session->haveEMS != options.haveEMS).

src/internal.c
```diff
@@ -40650,13 +40707,25 @@ static int AddPSKtoPreMasterSecret(WOLFSSL* ssl)
...
-            #endif
-                ret = EXT_MASTER_SECRET_NEEDED_E;
-                WOLFSSL_ERROR_VERBOSE(ret);
+#ifdef HAVE_EXTENDED_MASTER
+                if (ssl->options.disableEMS) {
+                    /* Local disable, not a client downgrade: decline the
+                     * resumption and do a full handshake. */
+                    WOLFSSL_MSG("EMS disabled locally, declining resumption "
+                                "of an EMS session. Do full handshake.");
+                    ssl->options.resuming = 0;
```
JacobBarthelmeh, line 41164: When this declines the resumption it should also clear peerAuthGood. If the client presented a ticket, DoClientTicket has already set ssl->options.peerAuthGood = 1 ("SERVER: ticket is peer auth", src/tls.c:6564/6576); falling back to a full handshake with that flag still set means the client-auth fail-safe (case TICKET_SENT: if (!ssl->options.peerAuthGood) ...) is satisfied by the discarded ticket rather than by whatever certificate the client presents in the full handshake.

JacobBarthelmeh (changes requested):

kareem-wolfssl:
Jenkins retest this please

kareem-wolfssl:
Retest this please

kareem-wolfssl:
Jenkins retest this please

src/internal.c
```diff
@@ -40841,13 +40898,27 @@ static int AddPSKtoPreMasterSecret(WOLFSSL* ssl)
             }
```
philljj, line 41155: This looks valid, similar to the comment lower at line 40907.

philljj (changes requested):

src/ssl_sess.c
```diff
@@ -1607,7 +1607,25 @@ int wolfSSL_SetSession(WOLFSSL* ssl, WOLFSSL_SESSION* session)
...
+        ssl->options.haveEMS = 1;
+    else if (ssl->options.disableEMS) {
+        ssl->options.haveEMS = 0;
+        /* An EMS session cannot be offered without the extension
+         * (RFC 7627 5.3): decline it and do a full handshake. TLS 1.3
+         * sessions resume independently of EMS; their haveEMS is only the
+         * RFC 8446 Appendix D indicator. */
+        if (ssl->session->haveEMS &&
+                !IsAtLeastTLSv1_3(ssl->session->version))
+            ssl->options.resuming = 0;
```
philljj, line 1623: My local skoll run also hit this:

### HIGH-1: Declined EMS session still pins the full handshake to the old protocol
- **File:** `src/ssl_sess.c:1614-1655`
- **Function:** `wolfSSL_SetSession`
- **Action:** BLOCK
- **Tag:** bug
- **Confidence:** High

**Description:** The PR's new disable branch correctly clears `ssl->options.resuming` for a pre-TLS-1.3 EMS session, but then falls through and imports that declined session's protocol version and cipher suite into the live connection. `SendClientHello` chooses its TLS 1.3 versus legacy path from `ssl->version`, so a flexible TLS-1.3-capable client given a TLS 1.2 EMS session will perform its supposed full fallback as TLS 1.2 instead of offering a fresh TLS 1.3 handshake. The added client-resumption test uses fixed TLS 1.2 methods on both connections, which masks this state leak.

**Code:**
```
if (ssl->session->haveEMS &&
        !IsAtLeastTLSv1_3(ssl->session->version))
    ssl->options.resuming = 0;
...
ssl->version = ssl->session->version;
...
ssl->options.cipherSuite0 = ssl->session->cipherSuite0;
ssl->options.cipherSuite  = ssl->session->cipherSuite;
```

**Recommendation:** Short-circuit the declined-session path and add a test that supplies a TLS 1.2 EMS session to an EMS-disabled flexible client, then verifies that the full handshake negotiates TLS 1.3.

src/internal.c
```diff
@@ -2362,7 +2362,8 @@ int InitSSL_Side(WOLFSSL* ssl, word16 side)
 #endif /* WOLFSSL_HAVE_SLHDSA */
 
 #if defined(HAVE_EXTENDED_MASTER) && !defined(NO_WOLFSSL_CLIENT)
-    if (ssl->options.side == WOLFSSL_CLIENT_END) {
+    /* Don't re-arm EMS advertising that the user disabled. */
```
philljj, line 2366: ### HIGH-2: EMS policy state is not reset when a WOLFSSL object is reused
- **File:** `src/internal.c:2364-2375`

**Description:** The PR introduces persistent `requireEMS` and `disableEMS` policy bits while continuing to use `haveEMS` as mutable per-handshake negotiation state. The new re-arming logic only runs when assigning a side; `wolfSSL_clear()` on an already-sided object does not call it or otherwise reset `haveEMS`. Consequently, a requiring client whose first handshake sees no EMS retains `requireEMS == 1` but `haveEMS == 0` after clear and no longer advertises EMS on the next connection, even to a supporting server. Conversely, a reused requiring server can retain `haveEMS == 1` from the previous peer; an absent EMS extension in the next ClientHello does not clear request-side state, so the required-EMS check is bypassed and the server can emit an unsolicited EMS response and derive incompatible keys. Every new EMS test allocates fresh objects, so object reuse is uncovered.

**Code:**
```
if (ssl->options.side == WOLFSSL_CLIENT_END &&
        !ssl->options.disableEMS) {
    ...
    ssl->options.haveEMS = 1;
}
```

**Recommendation:** Reset negotiated EMS state during object reuse without clearing the selected policy, and add client- and server-side `wolfSSL_clear()` tests that change peer EMS support between consecutive handshakes.

src/ssl_api_ext.c
```diff
@@ -1600,13 +1627,147 @@ int wolfSSL_DisableExtendedMasterSecret(WOLFSSL* ssl)
...
+/* Re-enable the Extended Master Secret extension on the object (default).
+ *
+ * Undoes a previous disable or require: EMS is used when the peer supports it
+ * but is not mandatory.
+ *
+ * @param [in] ssl  SSL/TLS object.
+ * @return  WOLFSSL_SUCCESS on success.
+ * @return  BAD_FUNC_ARG when ssl is NULL.
+ */
+int wolfSSL_EnableExtendedMasterSecret(WOLFSSL* ssl)
```
philljj, line 1688: ### MEDIUM-3: New object setters can overwrite negotiated EMS state mid-handshake
- **File:** `src/ssl_api_ext.c:1679-1765`

**Description:** Both new object-level setters unconditionally set `ssl->options.haveEMS = 1` for a client, even though that bit also records whether the peer actually negotiated EMS. A nonblocking application can regain control after parsing a non-EMS ServerHello but before master-secret derivation; calling either setter then changes the recorded negotiation result. `MakeMasterSecret` can consequently derive an EMS secret that the server is not using instead of reporting `EXT_MASTER_SECRET_NEEDED_E`. The public documentation lists no requirement that these APIs be called before the handshake starts.

**Code:**
```
ssl->options.requireEMS = 1;
ssl->options.disableEMS = 0;
if (ssl->options.side == WOLFSSL_CLIENT_END &&
        EmsAllowedForVersion(ssl->ctx->method->version))
    ssl->options.haveEMS = 1;
```

**Recommendation:** Define and enforce the valid call window in code and Doxygen, then add a nonblocking-handshake test covering a setter call after ServerHello processing.

philljj (dismissed):
Fenrir and Skoll found some overlapping issues, please review.

kareem-wolfssl:
Retest this please

kareem-wolfssl:
Retest this please

kareem-wolfssl:
Retest this please

JacobBarthelmeh (dismissed):

JacobBarthelmeh:
Retest this please Jenkins.

Windows test failure:

```
  connecting to 127.0.0.1:30602
  [0] Incorrect return -308
  Failure Reason Given is -1
  "DLL Release x64 - FAILED"
  "BUILDING SOLUTION"
  ```

JacobBarthelmeh:
Retest this please Jenkins . FIPS-ready error

JacobBarthelmeh (approved):

PR #10979, opened 2026-07-22
Route wc_MlDsaKey_SignCtx through the crypto callback before the prvKeySet check for device keys

aidangarske:
# Description 
- wc_MlDsaKey_SignCtx checked !prvKeySet and returned BAD_FUNC_ARG before the WOLF_CRYPTO_CB dispatch. A device/id‑backed key (private key in a TPM/HSM, referenced by devId) has no local private material (prvKeySet == 0), so it was rejected before the callback could run.
- Moves the prvKeySet check to after the crypto‑callback dispatch, matching the sibling wc_MlDsaKey_Sign / wc_MlDsaKey_SignCtxHash which already dispatch first.
- Enables device‑key ML‑DSA signing via crypto callback. e.g. a TLS 1.3 server signing CertificateVerify on a TPM. Reference: wolfTPM PR
  (aidangarske:mldsa-tls-cryptocb).


Frauschi:
Jenkins retest this please

Frauschi (approved):

dgarske (approved):

PR #10980, opened 2026-07-23
20260720-linuxkm-CMAC-native

douzzer:
Implement glue logic for Linux kernel registration of native `cmac(aes)` implementation, and fix gaps in several other `struct shash_alg` glue implementations:

`linuxkm/lkcapi_aes_glue.c`, `linuxkm/lkcapi_glue.c`, `configure.ac`: implement linuxkm `cmac(aes)` glue.

`linuxkm/lkcapi_sha_glue.c`: fix `sha.W` lifecycle management to prevent leaks:

* separate `WC_LINUXKM_SHA_IMPLEMENT()` into `WC_LINUXKM_SHA1_IMPLEMENT()` (no fixes
  needed) and `WC_LINUXKM_SHA2_IMPLEMENT()` (with associated new helpers
  `WC_LINUXKM_SHA2_FREE_W()`, `WC_LINUXKM_SHA2_DECL_W()`, `WC_LINUXKM_SHA2_PUSH_W()`,
  and `WC_LINUXKM_SHA2_POP_W()`, that move `.W` to a stack buffer).

* Reimplement SHA-2 one-shot digest callback to use only direct wolfCrypt calls
  rather than proxy to other callbacks.

`linuxkm/linuxkm_wc_port.h`, `linuxkm/module_hooks.c`: implement `magic` checks on kernel `struct wolfSSL_Mutex` in `WOLFSSL_LINUXKM_VERBOSE_DEBUG` builds.

`linuxkm/lkcapi_sha_glue.c`: refactor SHA-3 state around `struct km_Sha3TfmCtx`,
  adding `km_sha3_init_tfm()`, `km_sha3_exit_tfm()`, `km_sha3_alloc_tstate()`, and
  `km_sha3_free_tstate()`, implementing garbage collection of abandoned `shash_desc`
  objects at `.exit_tfm`.

`linuxkm/lkcapi_sha_glue.c`: add `struct km_sha3_export_state`, and implement `km_sha3_export()` and per-alg `_import()` methods, and `km_sha3_test_export_import()`.

`linuxkm/lkcapi_sha_glue.c`: fix export/import for HMAC:
* add to `struct km_sha_hmac_pstate` members `desc_list_lock`, `desc_list`, `export_list`, and `export_list_len`;
* add `struct km_sha_hmac_export_state`;
* add `km_hmac_alloc_tstate()`, `km_hmac_export()`, `km_hmac_import()`, `km_hmac_test_export_import()`;
* and update `km_hmac_*()` and `WC_LINUXKM_HMAC_IMPLEMENT`() to implement.

test/:
* fixes for `NO_DH`;
* fixes in `test_wc_ed448_import_public()` and `test_wc_Ed448DecisionCoverage()` for FIPS v6;
* fixes in `tests/api/test_sha3.c` for KMAC keysize in FIPS builds.

tested with
```
wolfssl-multi-test.sh ...
check-source-text-fips-dev
'.*insmod.*'
all-crypto-openssl-extra-coexist-fips-dev
```


philljj:
Retest this please.

```
FATAL: [ssh-agent] Unable to start agent
```

linuxkm/lkcapi_sha_glue.c
```diff
@@ -877,12 +1517,386 @@ WC_MAYBE_UNUSED static int km_hmac_finup(struct shash_desc *desc, const u8 *data
...
+    struct Hmac *h;
+    int ret;
+
+    h = (struct Hmac *)malloc(sizeof *h);
+    if (! h)
+        return -ENOMEM;
+
+    ret = wc_HmacCopy(&p_ctx->wc_hmac, h);
+    if (ret != 0) {
+        free(h);
```
douzzer, line 1566: False positive and wrong advice.  After `wc_HmacCopy` fails on an uninitialized `Hmac` object, the `Hmac` object is in an indeterminate state that is unsafe to `wc_HmacFree()` -- the only guarantee is that there are no dangling allocations under the `Hmac` object.

A `ForceZero()` of `h` before `free()` is appropriate, however (fixed).

wolfcrypt/src/sha256.c
```diff
@@ -747,7 +747,7 @@ static int InitSha256(wc_Sha256* sha256)
         sha256->devCtx = NULL;
     #endif
     #ifdef WOLFSSL_SMALL_STACK_CACHE
```
douzzer, line 749: Fixed (reverted)

douzzer, line 749: Fixed (reverted)

linuxkm/lkcapi_sha_glue.c
```diff
@@ -877,12 +1517,386 @@ WC_MAYBE_UNUSED static int km_hmac_finup(struct shash_desc *desc, const u8 *data
...
+#if defined(HAVE_FIPS) && FIPS_VERSION3_LT(6,0,0)
+        ForceZero(evicted, sizeof(*evicted));
+#endif
+        free(evicted);
+    }
+
+    /* Zero first so no uninitialized padding leaks into the caller's buffer. */
+    XMEMSET(blob, 0, sizeof(*blob));
+    blob->magic = WC_LINUXKM_HMAC_EXPORT_MAGIC;
+    blob->snapshot = snapshot;
```
douzzer: Fixed (refactored to use random cookies and sequential IDs).

linuxkm/lkcapi_sha_glue.c
```diff
@@ -451,28 +474,101 @@ struct km_sha_state {
...
+    /* Don't need to lock the mutex to clean up, because the API contract
+     * forbids any use of descs at/after exit of the associated TFM -- i.e. the
+     * list holds only abandoned descs -- and we're deallocating the lock
+     * besides.  Moreover, we definitely don't want to lock, so that the
+     * iteration and heap operations aren't in a locked context that might make
+     * desc deallocation awkward or impossible (leak).
+     */
+    list_for_each_entry_safe(s_ctx_i, next_ent, &t_ctx->desc_list, desc_ent) {
+        list_del(&s_ctx_i->desc_ent);
+        ForceZero(s_ctx_i, sizeof(*s_ctx_i));
```
douzzer, line 527: Fixed.

linuxkm/lkcapi_sha_glue.c
```diff
@@ -788,43 +1380,117 @@ WC_MAYBE_UNUSED static int linuxkm_hmac_setkey_common(struct crypto_shash *tfm,
...
+     */
+    list_for_each_entry_safe(node_i, next_ent, &p_ctx->desc_list, desc_ent) {
+        list_del(&node_i->desc_ent);
+        wc_HmacFree(&node_i->wc_hmac);
+#if defined(HAVE_FIPS) && FIPS_VERSION3_LT(6,0,0)
+        ForceZero(node_i, sizeof(*node_i));
+#endif
+        free(node_i);
+    }
+    list_for_each_entry_safe(node_i, next_ent, &p_ctx->export_list, desc_ent) {
```
douzzer, line 1474: False positive.  `wc_HmacFree(()` in mainline and FIPS v6+ does the `ForceZero()` itself.

linuxkm/lkcapi_glue.c
```diff
@@ -224,7 +224,7 @@ static wolfSSL_Atomic_Int linuxkm_lkcapi_registering_now = WOLFSSL_ATOMIC_INITIA
 static int linuxkm_lkcapi_register(void);
 static int linuxkm_lkcapi_unregister(void);
 
-#if defined(HAVE_FIPS) && defined(WC_LINUXKM_HAVE_SELFTEST)
+#if defined(HAVE_FIPS) && defined(WC_LINUX_CONFIG_SELFTESTS)
```
Frauschi: **🟡 [Medium] enabled_fips declaration/use guard mismatch can break -Werror kernel builds**
*🚫 BLOCK*  `bug`

This PR renamed `WC_LINUXKM_HAVE_SELFTEST` to `WC_LINUX_CONFIG_SELFTESTS` and, at the two use sites of `enabled_fips` (lines 270 and 348/357), additionally added `defined(CONFIG_CRYPTO_FIPS)` to the preprocessor guard. However, the *declaration* of `enabled_fips` at line 227 was only renamed and did NOT get the `CONFIG_CRYPTO_FIPS` guard. As a result, when a target is built with `HAVE_FIPS && WC_LINUX_CONFIG_SELFTESTS` but WITHOUT `CONFIG_CRYPTO_FIPS` (a valid combination: wolfCrypt FIPS build on a kernel that has CONFIG_CRYPTO_MANAGER selftests but not CONFIG_CRYPTO_FIPS), `static int enabled_fips` is declared but never referenced. That yields a `-Wunused-variable` warning, which is fatal in the typical `-Werror` linuxkm build. Before this PR the declaration and use guards matched exactly, so this is a regression introduced by the diff.

**Suggestion:**
```suggestion
#if defined(HAVE_FIPS) && defined(CONFIG_CRYPTO_FIPS) && \
    defined(WC_LINUX_CONFIG_SELFTESTS)
static int enabled_fips = 0;
#endif
```

configure.ac
```diff
@@ -7374,9 +7374,9 @@ AS_CASE([$FIPS_VERSION],
                (test "$FIPS_VERSION" != "lean-aesgcm-dev" || test "$enable_sha384" != "yes")],
             [enable_sha384="no"; ENABLED_SHA384="no"; AM_CFLAGS="$AM_CFLAGS -UWOLFSSL_SHA384 -UWOLFSSL_SHA384"])
 
-        AS_IF([test "$ENABLED_SHA512" != "no" &&
-               (test "$FIPS_VERSION" != "lean-aesgcm-dev" || test "$enable_sha512" != "yes")],
-            [enable_sha512="no"; ENABLED_SHA512="no"; ENABLED_SHA512_DRBG="no"; AM_CFLAGS="$AM_CFLAGS -UWOLFSSL_SHA512 -UWOLFSSL_SHA384"])
+        AS_IF([test "$ENABLED_SHA512" = "no" &&
```
Frauschi, line 7377: **🟡 [Medium] configure.ac lean-aesgcm-dev SHA512 logic inverted from disable to force-enable**
*💡 SUGGEST*  `question`

The SHA512 block was inverted: the condition changed from `test "$ENABLED_SHA512" != "no"` (with the action DISABLING SHA512: `-UWOLFSSL_SHA512`) to `test "$ENABLED_SHA512" = "no"` (with the action force-ENABLING it: `enable_sha512="yes"; ... -DWOLFSSL_SHA512 -DWOLFSSL_SHA384`). This now mirrors the SHA256 block just above (line 7353), so SHA512 becomes mandatory in the lean-aesgcm-dev FIPS variant (presumably required by the SHA512-based DRBG). The surrounding SHA3/SHAKE/SHA224/SHA384 blocks still use the old disable pattern, so SHA512 is now the odd one out. The PR description does not mention this configure change. It looks intentional and consistent with SHA256, but it is a meaningful behavior flip in FIPS config selection worth an explicit confirmation.

**Recommendation:** Confirm SHA512 is meant to be force-enabled (not disabled) in the lean-aesgcm-dev FIPS variant, and consider noting it in the PR description since it is not part of the stated linuxkm CMAC scope.

Frauschi (changes requested):
## 🐺 Skoll Code Review

### Posted findings
- [Medium] **enabled_fips declaration/use guard mismatch can break -Werror kernel builds** — `linuxkm/lkcapi_glue.c:227`
- [Medium] **configure.ac lean-aesgcm-dev SHA512 logic inverted from disable to force-enable** — `configure.ac:7377-7379`

- [Medium] km_hmac_import() calls allocating wc_HmacCopy() while holding spinlock (sleep-in-atomic)
- [Low] cpuid_flag: reg[num] = 0 is dead code (overwritten by cpuid())

</details>

configure.ac
douzzer:
```
commit 0a411897b4a9b261b8bb8ed8bf6cdfec4a20cf91
Author: Daniel Pouzzner <douzzer@wolfssl.com>
Date:   Thu Jul 23 17:46:30 2026 -0500

    configure.ac: enable SHA512 for fips=lean-aesgcm, to support SHA512 integrity check and DRBG.
```

linuxkm/lkcapi_glue.c
douzzer: Fixed.

douzzer:
retest this please
```
PRB-aarch64-tests-v2

ERROR: Error cloning remote repo 'origin'
  hudson.plugins.git.GitException: Command "git fetch --tags --force --progress -- git@github.com:wolfssl/wolfssl +refs/heads/*:refs/remotes/origin/*" returned status code 128:
  stdout: 
  stderr: git@github.com: Permission denied (publickey).
  
PRB-makedist-check

ERROR: Error cloning remote repo 'origin'
  hudson.plugins.git.GitException: Command "git fetch --tags --force --progress -- git@github.com:wolfssl/wolfssl +refs/heads/*:refs/remotes/origin/*" returned status code 128:
  stdout: 
  stderr: git@github.com: Permission denied (publickey).
```

linuxkm/lkcapi_aes_glue.c
```diff
@@ -4981,6 +5006,655 @@ static int linuxkm_test_aesecb(void) {
...
+    Cmac *cmac;
+
+    if (t_ctx->pristine == NULL)
+        return -ENOKEY;
+
+    cmac = (Cmac *)malloc(sizeof(*cmac));
+    if (! cmac)
+        return -ENOMEM;
+
+    XMEMCPY(cmac, t_ctx->pristine, sizeof(*cmac));
```
Frauschi, line 5187: **🟠 [High] CMAC transient-copy design breaks under WC_DEBUG_CIPHER_LIFECYCLE: dangling/double-free of Aes.CipherLifecycleTag**
*🚫 BLOCK*  `bug`

The new CMAC glue documents an invariant — "the embedded Aes owns no heap allocations in module configurations: wc_AesFree()'s only mainstream XFREE() target is the GCM-streaming streamData buffer" — and enforces it with a single `#error` for `WOLF_CRYPTO_CB_FREE`. That invariant is false under `WC_DEBUG_CIPHER_LIFECYCLE`: `wolfssl/wolfcrypt/aes.h:444` adds `void *CipherLifecycleTag`, `wolfcrypt/src/aes.c:15475` XMALLOCs it in `wc_AesInit()`, and `wolfcrypt/src/aes.c:15553` XFREEs it (and NULLs it) in `wc_AesFree()`. `km_AesCmacMaterialize()` struct-copies the tfm-owned pristine `Cmac` (including that pointer), and every `km_AesCmacUpdate()`/`Final()`/`Finup()` then destroys the copy via `wc_CmacFree()`/`wc_CmacFinal()` -> `wc_AesFree()`, which frees the tag while `t_ctx->pristine->aes.CipherLifecycleTag` still points at it. The next materialize copies the dangling pointer and `wc_debug_CipherLifecycleCheck()` reads freed kernel memory. `WOLF_CRYPTO_CB` without `_FREE` is a second unguarded case: `Cmac.devCtx` is a per-instance pointer that would likewise ride every copy.

**Suggestion:**
```suggestion
#if defined(WOLF_CRYPTO_CB) && defined(WOLF_CRYPTO_CB_FREE)
    #error LINUXKM_LKCAPI_REGISTER_AESCMAC is incompatible with WOLF_CRYPTO_CB_FREE.
#endif
#ifdef WC_DEBUG_CIPHER_LIFECYCLE
    #error LINUXKM_LKCAPI_REGISTER_AESCMAC is incompatible with WC_DEBUG_CIPHER_LIFECYCLE.
#endif
```

configure.ac
Frauschi, line 7377: **🟠 [High] configure.ac lean-aesgcm SHA-512 clause is logically inverted, force-ENABLING SHA-512/SHA-384 in the lean FIPS profile**
*🚫 BLOCK*

Line 7377-7379 sits inside the `[lean-aesgcm|lean-aesgcm-ready|lean-aesgcm-dev]` case arm (opened at configure.ac:7236). Every one of the ~40 neighbouring `AS_IF` clauses in that arm follows the same shape: `test "$ENABLED_FOO" != "no" && (FIPS_VERSION != lean-aesgcm-dev || enable_foo != "yes")` -> force-DISABLE. The PR inverts all four operands of the SHA-512 clause simultaneously (`!= "no"` -> `= "no"`, `!= "yes"` -> `!= "no"`, `-UWOLFSSL_SHA512 -UWOLFSSL_SHA384` -> `-DWOLFSSL_SHA512 -DWOLFSSL_SHA384`, and `ENABLED_SHA512_DRBG="no"` -> `"yes"`), turning it into a force-ENABLE. Two concrete regressions: (1) SHA-512 is on by default in wolfSSL, so `ENABLED_SHA512` is `yes` on entry, the new guard `test "$ENABLED_SHA512" = "no"` is false, and the clause no longer removes SHA-512/SHA-384 -- lean-aesgcm FIPS modules now ship SHA-512 and SHA-384 outside the lean algorithm boundary that the original code existed to enforce. (2) With `--enable-fips=lean-aesgcm --disable-sha512` (so `ENABLED_SHA512=no`, `enable_sha512=no`), `FIPS_VERSION != "lean-aesgcm-dev"` is TRUE, so the clause fires and force-re-enables SHA-512 against the user's explicit `--disable-sha512`. Additionally, the SHA-384 clause three lines above appends `-UWOLFSSL_SHA384` while setting `ENABLED_SHA384="no"`; when this clause fires it appends `-DWOLFSSL_SHA384` later on the same command line, so `WOLFSSL_SHA384` ends up DEFINED while `ENABLED_SHA384` remains `no` -- sha512.c compiles SHA-384 code that the automake conditionals and the FIPS algorithm manifest believe is absent. The PR description documents only the `cmac(aes)` addition to configure.ac and says nothing about SHA-512, which suggests this is a leftover debug/bisect edit rather than an intended change.

**Recommendation:** Revert the clause to the force-disable form matching every sibling clause in the arm:

        AS_IF([test "$ENABLED_SHA512" != "no" &&
               (test "$FIPS_VERSION" != "lean-aesgcm-dev" || test "$enable_sha512" != "yes")],
            [enable_sha512="no"; ENABLED_SHA512="no"; ENABLED_SHA512_DRBG="no"; AM_CFLAGS="$AM_CFLAGS -UWOLFSSL_SHA512 -UWOLFSSL_SHA384"])

If the intent really was to make SHA-512 mandatory for lean-aesgcm, it needs its own commit with a rationale, and the duplicated `-UWOLFSSL_SHA384 -UWOLFSSL_SHA384` typo in the SHA-384 clause above should be fixed to `-UWOLFSSL_SHA512 -UWOLFSSL_SHA384` or deduplicated.

linuxkm/lkcapi_sha_glue.c
```diff
@@ -659,75 +1119,135 @@ static int km_ ## name ## _finup(struct shash_desc *desc, const u8 *data,  \
...
+    return 0;                                                              \
 }                                                                          \
                                                                            \
+wc_static_assert((block_size) <=                                           \
+                 sizeof(((struct km_sha3_export_state *)0)->t));           \
+                                                                           \
+                                                                           \
 static struct shash_alg name ## _alg =                                     \
 {                                                                          \
+    .init_tfm       =       km_sha3_init_tfm,                              \
```
Frauschi, line 1181: **🟠 [High] SHA-3 and CMAC shash_algs adopt .init_tfm/.exit_tfm without extending the >= 5.6 kernel-version guard**
*💡 SUGGEST*

`struct shash_alg`'s `init_tfm`/`exit_tfm` members were added in Linux 5.6. The file already encodes that fact in an explicit guard at lines 409-420 -- `#error LINUXKM_LKCAPI_REGISTER for HMACs is supported only on Linux kernel versions >= 5.6.0.` -- whose condition enumerates only the eight `*_HMAC` register macros, because HMAC was the only user of those callbacks. This PR adds `.init_tfm = km_sha3_init_tfm` and `.exit_tfm = km_sha3_exit_tfm` to the SHA-3 `shash_alg` (lines 1177/1188) and `.exit_tfm = km_AesCmacExitTfm` to `cmacAesAlg` (lkcapi_aes_glue.c:5350), but does not extend the guard's condition to cover `LINUXKM_LKCAPI_REGISTER_SHA3_224/256/384/512` or `LINUXKM_LKCAPI_REGISTER_AESCMAC`. The module advertises support back to 5.4 (`lkcapi_glue.c:30-32`: `#error LINUXKM_LKCAPI_REGISTER is supported only on Linux kernel versions >= 5.4.0.`). Trigger: build against a 5.4 or 5.5 kernel with `--enable-linuxkm-lkcapi-register=sha3` (or `all`, or `cmac(aes)`). Instead of the intended clear `#error`, the build fails with an opaque 'struct shash_alg has no member named init_tfm' from the macro expansion, or -- worse if a distro backport added only one of the two members -- compiles with the callback silently dropped, which for SHA-3 means `desc_list_lock` is never initialized and `km_sha3_alloc_tstate()` locks an uninitialized spinlock on the first digest.

**Recommendation:** Extend the guard's condition to every alg that now installs an init_tfm/exit_tfm callback, and retitle the message:

#if (LINUX_VERSION_CODE < KERNEL_VERSION(5, 6, 0)) &&  \
    (defined(LINUXKM_LKCAPI_REGISTER_SHA1_HMAC) ||     \
     ... existing HMAC entries ...               \
     defined(LINUXKM_LKCAPI_REGISTER_SHA3_224) ||      \
     defined(LINUXKM_LKCAPI_REGISTER_SHA3_256) ||      \
     defined(LINUXKM_LKCAPI_REGISTER_SHA3_384) ||      \
     defined(LINUXKM_LKCAPI_REGISTER_SHA3_512))
    #error LINUXKM_LKCAPI_REGISTER for HMACs and SHA-3 is supported only on Linux kernel versions >= 5.6.0.
#endif

and add the matching guard for LINUXKM_LKCAPI_REGISTER_AESCMAC in lkcapi_aes_glue.c.

linuxkm/lkcapi_sha_glue.c
```diff
@@ -877,12 +1543,387 @@ WC_MAYBE_UNUSED static int km_hmac_finup(struct shash_desc *desc, const u8 *data
...
+    if (p_ctx->export_list_len >= WC_LINUXKM_HMAC_EXPORT_LIST_MAX) {
+        evicted = list_last_entry(&p_ctx->export_list,
+                                  struct km_sha_hmac_node, desc_ent);
+        list_del(&evicted->desc_ent);
+        p_ctx->export_list_len--;
+    }
+    snapshot->desc_id = p_ctx->cur_desc_id++;
+    /* list_add() prepends, so the tail from list_last_entry() is the oldest. */
+    list_add(&snapshot->desc_ent, &p_ctx->export_list);
+    p_ctx->export_list_len++;
```
Frauschi, line 1635: **🟡 [Medium] km_hmac_export reads snapshot->desc_id after releasing the list lock (use-after-free window)**
*🚫 BLOCK*  `bug`

The function's own design note says unlinking under the lock is what makes the ring safe: "the evicted node is now unlinked and unreachable ... so free it outside the lock." But after `wc_UnLockMutex()` the code dereferences `snapshot` again to fill in the blob (`blob->desc_id = snapshot->desc_id;`). Once the lock is dropped, a concurrent `km_hmac_export()` on the same tfm can evict this very node (it is the list head, becoming the tail after `WC_LINUXKM_HMAC_EXPORT_LIST_MAX` further exports) and `free()` it, so the read races a free. Every other post-unlock dereference in the function is on the deliberately-unlinked `evicted` node; this one is on a still-published node.

**Suggestion:**
```suggestion
word64 new_desc_id;
    ...
    new_desc_id = snapshot->desc_id = p_ctx->cur_desc_id++;
    list_add(&snapshot->desc_ent, &p_ctx->export_list);
    p_ctx->export_list_len++;
    (void)wc_UnLockMutex(&p_ctx->desc_list_lock);
    ...
    blob->desc_id = new_desc_id;
```

linuxkm/linuxkm_wc_port.h
```diff
@@ -2005,6 +2036,10 @@
```
Frauschi, line 2036: **🟡 [Medium] wc_UnLockMutex returns -1 without unlocking on magic mismatch, leaving the spinlock held with IRQs disabled**
*💡 SUGGEST*  `bug`

The new magic check returns early before `spin_unlock_irqrestore()`. The scenario it is designed to catch (a mutex that was `wc_FreeMutex()`d — which now sets `magic = 0` — while still held) therefore ends with the lock held and interrupts disabled on that CPU, and most call sites in this PR discard the result (`(void)wc_UnLockMutex(...)`), so the failure is silent until the CPU hangs. It also returns a bare `-1`, while the sibling check added in `module_hooks.c:183` returns `BAD_FUNC_ARG`.

**Suggestion:**
```suggestion
#ifdef WOLFSSL_LINUXKM_VERBOSE_DEBUG
    if ((m == NULL) || (m->magic != WC_LINUXKM_SPINLOCK_MAGIC)) {
        WARN_ONCE(1, "wolfSSL: wc_UnLockMutex on invalid mutex %p\n", m);
        return BAD_MUTEX_E;
    }
#endif
```

wolfcrypt/src/aes.c
```diff
@@ -13992,9 +13992,19 @@ int wc_AesGcmInit(Aes* aes, const byte* key, word32 len, const byte* iv,
 
         #ifdef WOLFSSL_AESNI
```
Frauschi, line 13993: **🟡 [Medium] wc_AesGcmInit() WC_C_DYNAMIC_FALLBACK permanently clears aes->use_aesni instead of scoping the fallback to the call**
*💡 SUGGEST*

The established `WC_C_DYNAMIC_FALLBACK` idiom in aes.c is `VECTOR_REGISTERS_PUSH` / `VECTOR_REGISTERS_POP` (defined at aes.c:5984-6005). PUSH saves `int orig_use_aesni = aes->use_aesni;` and clears `use_aesni` only if `SAVE_VECTOR_REGISTERS2()` fails; POP then does `if (aes->use_aesni) RESTORE_VECTOR_REGISTERS(); else aes->use_aesni = orig_use_aesni;` -- i.e. the C fallback is scoped to a single operation and the AESNI capability is restored afterwards. The new code in `wc_AesGcmInit` hand-rolls the pattern but omits the restore: `aes->use_aesni = 0;` is permanent for the lifetime of that `Aes` object. `SAVE_VECTOR_REGISTERS2()` failure is transient and context-dependent (it fails in nested/softirq contexts, and under `DEBUG_VECTOR_REGISTER_ACCESS_FUZZING` it fails pseudo-randomly), so a single unlucky `wc_AesGcmInit` call silently demotes a long-lived `Aes` -- e.g. an LKCAPI AEAD tfm reused for every request -- to pure C for the rest of its life. Output stays correct (`key_C_fallback` is populated at setkey time under WC_C_DYNAMIC_FALLBACK, aes.c:2573/3676/5212/5374), so this is a silent permanent throughput regression and a state-consistency divergence from every other AESNI site in the file, not a wrong-answer bug.

**Recommendation:** Use the existing scoped macros so the capability is restored, matching every other fallback site in aes.c:

            if (aes->use_aesni) {
                VECTOR_REGISTERS_PUSH;
                if (aes->use_aesni)
                    ret = AesGcmInit_aesni(aes, iv, ivSz);
                else
                    ret = AesGcmInit_C(aes, iv, ivSz);
                VECTOR_REGISTERS_POP;
            }

If the hand-rolled form must stay, save and restore `orig_use_aesni` explicitly around the fallback.

Frauschi (changes requested):
## 🐺 Skoll Code Review

**Findings:** 47 total — 6 posted, 41 skipped

### Posted findings
- [High] **CMAC transient-copy design breaks under WC_DEBUG_CIPHER_LIFECYCLE: dangling/double-free of Aes.CipherLifecycleTag** — `linuxkm/lkcapi_aes_glue.c:5168`
- [High] **configure.ac lean-aesgcm SHA-512 clause is logically inverted, force-ENABLING SHA-512/SHA-384 in the lean FIPS profile** — `configure.ac:7377-7379`
- [High] **SHA-3 and CMAC shash_algs adopt .init_tfm/.exit_tfm without extending the >= 5.6 kernel-version guard** — `linuxkm/lkcapi_sha_glue.c:1177`
- [Medium] **km_hmac_export reads snapshot->desc_id after releasing the list lock (use-after-free window)** — `linuxkm/lkcapi_sha_glue.c:1630-1648`
- [Medium] **wc_UnLockMutex returns -1 without unlocking on magic mismatch, leaving the spinlock held with IRQs disabled** — `linuxkm/linuxkm_wc_port.h:2036-2045`
- [Medium] **wc_AesGcmInit() WC_C_DYNAMIC_FALLBACK permanently clears aes->use_aesni instead of scoping the fallback to the call** — `wolfcrypt/src/aes.c:13993-14008`

- [Medium] km_hmac_import and SHA-3 .init/.import orphan the desc's previous working node onto an uncapped list
- [Medium] km_hmac_import performs wc_HmacCopy (which may allocate) while holding the IRQ-disabling spinlock
- [Medium] WC_SHA256_W_SIZE / WC_SHA512_W_SIZE duplicate private allocation sizes from sha256.c / sha512.c
- [Medium] WC_LINUXKM_SHA2_FREE_W frees an XMALLOC'd buffer with raw free()
- [Medium] The entire SHA-2 stack-W path is compiled out in default kernel builds and therefore ships untested
- [Medium] HMAC .export handle is not self-contained state and is valid only within one tfm and one eviction window
- [Medium] km_hmac_export() reads the snapshot node after publishing it to export_list and dropping the lock (use-after-free read)
- [Medium] New export/import self-tests allocate their tfm by cra_name and assert wolfSSL-private statesize, breaking registration under LINUXKM_LKCAPI_PRIORITY_ALLOW_MASKING
- [Medium] km_hmac_import() orphans a working node per abandoned desc, giving unbounded kernel memory growth on the AF_ALG accept()/close() path
- [Medium] AES-CMAC transient-copy design use-after-frees the Aes CipherLifecycleTag under WC_DEBUG_CIPHER_LIFECYCLE
- [Medium] km_AesCmacSetKey() frees the tfm-owned pristine Cmac that concurrent descs dereference
- [Medium] km_hmac_export() reads snapshot->desc_id after dropping desc_list_lock, racing eviction/free
- [Medium] HMAC export-snapshot ring is sized by CPU count, not in-flight accepts, causing spurious -EINVAL under concurrent AF_ALG accept()
- [Low] SHA-2 one-shot .digest still heap-allocates W, contradicting the stated no-heap-after-init design
- [Low] types.h: undocumented !defined(__GNUC__) on the C23 static_assert branch changes which implementation GCC/clang select
- [Low] configure.ac 'cmac(aes)' defines HAVE_AES_ECB (unused) instead of WOLFSSL_AES_DIRECT (required by the glue gate)
- [Low] configure.ac lean-aesgcm: SHA-512 force-enable adds -DWOLFSSL_SHA384 while ENABLED_SHA384 stays "no"
- [Low] km_AesCmacFinal/Finup bypass km_AesCmacDispose and bare-free the Cmac
- [Low] SHA-3 one-shot digest places a ~480-byte struct on the kernel stack
- [Low] km_AesCmacSetKey does not fail closed on the key-length rejection path
- [Low] CMAC transient copy does not defensively NULL aes.streamData, unlike km_AesGet() in the same file
- [Low] cpuid.c: replacing XMEMSET of the register array with a single-element zero is unrelated to this PR
- [Low] Bare scope blocks introduced in km_AesCmacFinal / km_AesCmacFinup
- [Low] KMAC-256 FIPS path drops the zero-length-key check instead of asserting KMAC_MIN_KEYLEN_E
- [Low] SHA-3 .init and .import orphan any prior state node onto an unbounded per-TFM desc_list
- [Low] Bounded HMAC export ring makes a previously valid state blob permanently unimportable, so concurrent AF_ALG accept() can fail spuriously
- [Low] km_AesCmacSetKey() does not fail closed on the invalid-key-length path, contradicting its own rekey invariant
- [Low] SHA-3 export blob carries no algorithm/variant tag, permitting cross-variant state import
- [Low] cpuid.c XMEMSET removal not applied to cpuid_is_intel()/cpuid_is_amd(), which run first on the same path
- [Low] km_hmac_import() performs the deep wc_HmacCopy() while holding the IRQ-disabling spinlock
- [Low] wc_FreeMutex() now invalidates the lock, and wc_UnLockMutex() can bail out while still holding the spinlock
- [Low] km_hmac_free_tstate()/km_sha3_free_tstate() silently skip cleanup when the list lock cannot be taken
- [Info] linuxkm_test_aescmac ForceZeros an uninitialized stack struct on early-exit paths
- [Info] wc_FreeMutex uses memset (not XMEMSET/ForceZero) and keeps a now-vestigial (void)m
- [Info] WC_LINUXKM_SPINLOCK_MAGIC is an unexplained decimal literal
- [Info] list_last_entry() without a list_empty() guard, protected only by an assert that can be silently skipped
- [Info] Empty #if/#endif block left behind in the selftest-macro rename
- [Info] settings.h #undefs WOLFSSL_SMALL_STACK before headers are parsed, changing struct layouts inside the sha256.c/sha512.c translation units
- [Info] WC_LINUXKM_SHA2_FREE_W() releases an XMALLOC'd buffer with raw free()
- [Info] WC_LINUXKM_SHA2_PUSH_W/POP_W generate bare brace-scope blocks in every SHA-2 shash entry point
- [Info] SHA-3 .digest places a ~448-byte struct on the stack, contradicting the HMAC digest's stated rationale

</details>

configure.ac
douzzer: False positive.  It _has_ its own commit with the rationale, as already noted above on the previous Skoll review:
```
commit 168893520a6dcf87ae0fe401a98f10deb7658a9a
Author: Daniel Pouzzner <douzzer@wolfssl.com>
Date:   Thu Jul 23 17:46:30 2026 -0500

    configure.ac: enable SHA512 for fips=lean-aesgcm, to support SHA512 integrity check and DRBG.
```

linuxkm/lkcapi_sha_glue.c
douzzer: Good catch.  Fixed.

linuxkm/linuxkm_wc_port.h
douzzer: False positive.  If `.magic` doesn't match, then `m` is invalid/corrupt (e.g. was previously ` wc_FreeMutex()`ed) and can't be used.  Note also this is all gated on `WOLFSSL_LINUXKM_VERBOSE_DEBUG`.

wolfcrypt/src/aes.c
douzzer: False positive, wrong fix.  Incumbent `AesSetKey_C()` persistently disables `.aesni` on the `Aes` object if `SAVE_VECTOR_REGISTERS2()` fails, and must do so due to missing AES-NI setup.  The fix for `wc_AesGcmInit()` functions similarly.

linuxkm/lkcapi_sha_glue.c
douzzer: Fixed.

linuxkm/lkcapi_aes_glue.c
douzzer: Fixed.

Frauschi (dismissed):

SparkiDev (dismissed):

Frauschi (approved):

PR #10981, opened 2026-07-23
ML-KEM assembly: AVX512F and AVX512VBMI

SparkiDev:
# Description

New assembly for ML-KEM for Intel x64 machines that support AVX512 extensions.

# Testing

Regression testing of ML-KEM.

# Generator

https://github.com/wolfSSL/scripts/pull/638


wolfcrypt/src/wc_mlkem_poly.c
```diff
@@ -4280,6 +4741,73 @@ static int mlkem_get_noise_k2_avx2(MLKEM_PRF_T* prf, sword16* vec1,
...
+#ifdef WOLFSSL_MLKEM_HAVE_INTEL_AVX512
+/* Get eight lanes of ETA3-length random bytes using eight-way AVX-512 SHA3.
+ * Lane j uses seed count j; its output (two blocks) is written to
+ * rand + j * PRF_RAND_SZ. ETA2 samplers may read the same lanes (they consume
+ * fewer bytes of the identical SHAKE stream).
+ *
+ * @param  [out]  rand  Random number byte array (8 * PRF_RAND_SZ bytes).
+ * @param  [in]   seed  Seed to generate random from.
+ */
+static void mlkem_get_noise_x8_eta3_avx512(byte* rand, byte* seed)
```
Frauschi: **🟡 [Medium] New AVX512 k2 noise path uses large unconditional stack buffers, unlike its AVX2 sibling**
*💡 SUGGEST*  `convention`

`mlkem_get_noise_k2_avx512` declares `byte rand[8 * PRF_RAND_SZ]` (PRF_RAND_SZ = 2*SHA3_256_BYTES = 272, so 2176 bytes) as a plain fixed stack array, and its helper `mlkem_get_noise_x8_eta3_avx512` (and the eta2 helper) each declare `word64 state[25 * 8]` (1600 bytes) on the stack. The corresponding AVX2 sibling `mlkem_get_noise_k2_avx2` instead respects `WOLFSSL_SMALL_STACK` via `WC_DECLARE_VAR(rand, byte, 4 * PRF_RAND_SZ, 0)` + `WC_ALLOC_VAR_EX(...)`, and the new AVX512 matrix generators (`mlkem_gen_matrix_k2/k3/k4_avx512`) also use the XMALLOC small-stack pattern. So the new k2 noise path both doubles the buffer and drops the small-stack handling its own AVX2 twin provides, violating the project rule that large (>~100 byte) stack buffers use the `WOLFSSL_SMALL_STACK` pattern. Practical impact is low because these paths only execute on AVX-512-capable x86-64 (rarely a small-stack target), but it is an internal inconsistency worth aligning.

**Suggestion:** Use the WC_DECLARE_VAR / WC_ALLOC_VAR_EX pattern (as mlkem_get_noise_k2_avx2 does) for the rand and state buffers in mlkem_get_noise_k2_avx512 and the x8 helpers, returning MEMORY_E on allocation failure. Note k3/k4 avx512 mirror their avx2 siblings' plain-stack arrays, so this concern is specific to the k2 path.

Frauschi (commented):
## 🐺 Skoll Code Review

### Posted findings
- [Medium] **New AVX512 k2 noise path uses large unconditional stack buffers, unlike its AVX2 sibling** — `wolfcrypt/src/wc_mlkem_poly.c:4754-4809`

wolfcrypt/src/wc_mlkem_poly.c
SparkiDev: Changed as suggested.

wolfcrypt/src/wc_mlkem_poly.c
```diff
@@ -4298,21 +4843,57 @@ static int mlkem_get_noise_k3_avx2(sword16* vec1, sword16* vec2, sword16* poly,
...
+ * @param  [out]  vec2  Second Vector of polynomials.
+ * @param  [out]  poly  Polynomial.
+ * @param  [in]   seed  Seed to use when calculating random.
+ * @return  0 on success.
+ */
+static int mlkem_get_noise_k3_avx512(sword16* vec1, sword16* vec2,
+    sword16* poly, byte* seed)
+{
+    int ret;
+    byte rand[8 * ETA2_RAND_SIZE];
```
Frauschi: **🟡 [Medium] mlkem_get_noise_k3/k4_avx512 double the stack buffer and drop the WOLFSSL_SMALL_STACK handling used by the k2 sibling**
*💡 SUGGEST*  `convention`

Both new functions declare `byte rand[8 * ETA2_RAND_SIZE]` as a plain automatic array. `ETA2_RAND_SIZE` is `(2 * MLKEM_N) / 4` = 128, so this is 1024 bytes of stack, unconditionally, in every build - including `WOLFSSL_SMALL_STACK`. That is exactly double the 512 bytes their AVX2 counterparts (`mlkem_get_noise_k3_avx2:4843`, `mlkem_get_noise_k4_avx2:4914`) use. The inconsistency is internal to this commit: `mlkem_get_noise_k2_avx512:4800` uses `WC_DECLARE_VAR(rand, byte, 8 * PRF_RAND_SZ, 0)` + `WC_ALLOC_VAR_EX(...)`, and the shared helper `mlkem_get_noise_x8_eta2_avx512:4590` heap-allocates its 1600-byte state via `WC_DECLARE_VAR` - so a k3 call under WOLFSSL_SMALL_STACK heap-allocates the 1600-byte state while its caller has already put 1024 bytes on the stack. The AVX2 versions predate the small-stack macros so the 512-byte arrays are grandfathered in, but new code doubling them is a regression for embedded and deeply-threaded builds. (The `mlkem_gen_matrix_k*_avx512` functions do handle this correctly: `rand[8 * GEN_MATRIX_SIZE + 4]` = 4036 B plus `state[25 * 8]` = 1600 B is ~5.6 KB versus the AVX2 ~2.8 KB, but it is properly wrapped in `#ifdef WOLFSSL_SMALL_STACK`.)

**Suggestion:**
```suggestion
static int mlkem_get_noise_k3_avx512(sword16* vec1, sword16* vec2,
    sword16* poly, byte* seed)
{
    int ret;
    WC_DECLARE_VAR(rand, byte, 8 * ETA2_RAND_SIZE, 0);

    WC_ALLOC_VAR_EX(rand, byte, 8 * ETA2_RAND_SIZE, NULL,
        DYNAMIC_TYPE_TMP_BUFFER, return MEMORY_E);

    ret = mlkem_get_noise_x8_eta2_avx512(rand, seed);
    ...
    ForceZero(rand, 8 * ETA2_RAND_SIZE);
    WC_FREE_VAR_EX(rand, NULL, DYNAMIC_TYPE_TMP_BUFFER);
    return ret;
}
```

wolfssl/wolfcrypt/wc_mlkem.h
```diff
@@ -583,6 +583,19 @@ WOLFSSL_LOCAL
 int mlkem_check_reduced(const sword16* p, int k);
 
 #ifdef USE_INTEL_SPEEDUP
+/* AVX512 assembly for ML-KEM is built (and dispatched at runtime on capable
```
Frauschi, line 588: **🟡 [Medium] NO_AVX512_SUPPORT no longer disables all AVX-512 code emission: the VBMI assembly block is not nested under the AVX-512 guard**
*💡 SUGGEST*

The new comment block added here states that `WOLFSSL_MLKEM_HAVE_INTEL_AVX512_VBMI` "matches the auto-enabled HAVE_INTEL_AVX512_VBMI guard around the generated assembly". That claim does not hold. On the C side the VBMI define is nested three deep: `#ifdef USE_INTEL_SPEEDUP` -> `#ifndef NO_AVX512_SUPPORT` -> `#ifndef NO_AVX512_VBMI_SUPPORT`. On the assembly side the base AVX-512 block is gated by `#ifndef NO_AVX512_SUPPORT` (wolfcrypt/src/wc_mlkem_asm.S:54-58, wc_mlkem_asm.asm:15440), but the ~1400-line VBMI block that follows is opened at top level with only `#ifndef NO_AVX512_VBMI_SUPPORT` (wc_mlkem_asm.S:34874-34880; wc_mlkem_asm.asm:33931-33937), independent of `NO_AVX512_SUPPORT`. Building with `-DNO_AVX512_SUPPORT` alone therefore produces an inconsistent result: the C side declares and calls nothing (correct), while the assembler still emits nine `mlkem_*_avx512_vbmi` routines full of EVEX/ZMM encodings and 58 `vpermb` instructions. `NO_AVX512_SUPPORT` is the established, tree-wide opt-out that every other wolfSSL assembly file (aes_gcm_asm.S, aes_xts_asm.S, sha3_asm.S, sp_x86_64_asm.S, chacha_asm.S, ...) uses to gate 100% of its AVX-512 emission, precisely so a toolchain whose assembler cannot encode AVX-512 can still build. After this commit wc_mlkem_asm.S/.asm break that contract: a user on an older binutils/clang who sets NO_AVX512_SUPPORT for exactly that reason now fails to assemble, and every build that sets it links ~1400 lines of unreferenced dead code. Verified that no cross-references exist between the two blocks, so the fix is a pure guard change.

**Recommendation:** Make VBMI a strict subset of AVX512 in the generated assembly so it matches the header. Wrap wc_mlkem_asm.S:34875-34879 in `#ifndef NO_AVX512_SUPPORT` / `#endif` and wc_mlkem_asm.asm:33932-33936 in `IFNDEF NO_AVX512_SUPPORT` / `ENDIF`, or move the whole VBMI block inside the existing `#ifdef HAVE_INTEL_AVX512` region. Since these files are generated, fix the generator template so it does not regress.

wolfssl/wolfcrypt/wc_mlkem.h
```diff
@@ -649,6 +662,111 @@ void mlkem_decompress_5_avx2(sword16* p, const byte* r);
...
+void mlkem_from_bytes_avx512_vbmi(sword16* p, const byte* b);
+WOLFSSL_LOCAL
+void mlkem_to_bytes_avx512_vbmi(byte* b, sword16* p);
+#endif
+WOLFSSL_LOCAL
+void mlkem_from_msg_avx512(sword16* p, const byte* msg);
+WOLFSSL_LOCAL
+void mlkem_to_msg_avx512(byte* msg, sword16* p);
+WOLFSSL_LOCAL
+void mlkem_compress_4_avx512(byte* b, sword16* p);
```
Frauschi: **🔵 [Low] New AVX512 declarations are inconsistently const-qualified, named, and ordered**
*🔧 NIT*  `api`

The AVX2 block directly above (lines 644-660) is a clean, grouped, uniformly const-qualified list. The new AVX512 block diverges on three axes. (1) const-ness: `mlkem_compress_4_avx2(byte* r, const sword16* p)` but `mlkem_compress_4_avx512(byte* b, sword16* p)` (non-const) and `mlkem_compress_4_avx512_vbmi(byte* b, const sword16* p)` (const again) - three declarations of the same operation, two different signatures. `mlkem_compress_5_avx512` and `mlkem_compress_5_avx512_vbmi` are both non-const while `mlkem_compress_5_avx2` is const. (2) parameter naming: the AVX2 declarations use `r` for the byte output, the AVX512 ones switch to `b`. (3) ordering: `mlkem_compress_5_avx512` (line 712) sits between `mlkem_decompress_11_avx512` and `mlkem_compress_10_avx512`, and `mlkem_compress_4_avx512_vbmi` (line 721) is grouped with `mlkem_compress_10_avx512_vbmi` rather than next to `mlkem_compress_4_avx512` at line 696. None of this breaks the build - all call sites pass non-const `sword16*` - but it makes the header harder to audit for missing dispatches.

**Suggestion:**
```suggestion
WOLFSSL_LOCAL
void mlkem_compress_4_avx512(byte* r, const sword16* p);
WOLFSSL_LOCAL
void mlkem_decompress_4_avx512(sword16* p, const byte* r);
WOLFSSL_LOCAL
void mlkem_compress_5_avx512(byte* r, const sword16* p);
WOLFSSL_LOCAL
void mlkem_decompress_5_avx512(sword16* p, const byte* r);
#ifdef WOLFSSL_MLKEM_HAVE_INTEL_AVX512_VBMI
WOLFSSL_LOCAL
void mlkem_compress_4_avx512_vbmi(byte* r, const sword16* p);
WOLFSSL_LOCAL
void mlkem_compress_5_avx512_vbmi(byte* r, const sword16* p);
#endif
```

wolfcrypt/src/wc_mlkem_poly.c
```diff
@@ -4132,13 +4569,42 @@ static void mlkem_get_noise_x4_eta2_avx2(byte* rand, byte* seed, byte o)
...
+    mlkem_redistribute_16_rand_ins(state, rand + 0 * ETA2_RAND_SIZE,
         rand + 1 * ETA2_RAND_SIZE, rand + 2 * ETA2_RAND_SIZE,
         rand + 3 * ETA2_RAND_SIZE);
 
     /* state is secret-seeded; caller zeroizes rand. */
     ForceZero(state, sizeof(state));
 }
+
+#ifdef WOLFSSL_MLKEM_HAVE_INTEL_AVX512
+/* Get eight lanes of ETA2 random bytes using eight-way AVX-512 SHA3. Lane j
```
Frauschi, line 4604: **🔵 [Low] Doc comments on the new x8 helpers omit @return and misstate lane counts**
*🔧 NIT*  `style`

Four small doc-comment defects in the new code. (1) `mlkem_get_noise_x8_eta2_avx512` and `mlkem_get_noise_x8_eta3_avx512` both return `int` and can return `MEMORY_E` from `WC_ALLOC_VAR_EX`, but neither has an `@return` line - every other documented function in this file, including the AVX2 siblings and the new `mlkem_gen_matrix_k*_avx512`, documents both the 0 and MEMORY_E cases. (2) `mlkem_get_noise_k3_avx512`'s comment says "All seven ETA2 lanes are produced in one batch" - eight lanes are produced, seven are consumed when `poly != NULL` and only six when it is NULL. (3) `mlkem_gen_matrix_k3_avx512`'s comment says "First eight polynomials (rows 0..2, up to column 1)", which is confusing shorthand for "indices 0..7 of the row-major 3x3, i.e. through (2,1)". (4) `mlkem_get_noise_k2_avx512` documents `seed` as `@param [in, out]` but, unlike `mlkem_get_noise_k2_avx2` which writes `seed[WC_ML_KEM_SYM_SZ] = 4`, it never writes to the buffer - so it should be `[in]`. (That behavioural difference is itself benign: I checked both call sites in `wolfcrypt/src/wc_mlkem.c:1223` and `:934`, and `r`/`sigma` are not read after `mlkem_get_noise` returns on the non-small-mem paths that pass a non-NULL `poly`.)

**Suggestion:**
```suggestion
* @param  [out]  rand  Random number byte array (8 * ETA2_RAND_SIZE bytes).
 * @param  [in]   seed  Seed to generate random from.
 * @return  0 on success.
 * @return  MEMORY_E when dynamic memory allocation fails. Only possible when
 *          WOLFSSL_SMALL_STACK is defined.
 */
static int mlkem_get_noise_x8_eta2_avx512(byte* rand, byte* seed)
```

wolfssl/wolfcrypt/wc_mlkem.h
```diff
@@ -583,6 +583,19 @@ WOLFSSL_LOCAL
 int mlkem_check_reduced(const sword16* p, int k);
 
 #ifdef USE_INTEL_SPEEDUP
+/* AVX512 assembly for ML-KEM is built (and dispatched at runtime on capable
+ * CPUs) whenever the Intel speedups are enabled and AVX512 is not opted out.
+ * Matches the HAVE_INTEL_AVX512 guard around the generated assembly. */
+#ifndef NO_AVX512_SUPPORT
+    #define WOLFSSL_MLKEM_HAVE_INTEL_AVX512
+    /* AVX512VBMI (vpermb) functions are built (matching the auto-enabled
```
Frauschi: **🔵 [Low] NO_AVX512_VBMI_SUPPORT does not remove all VBMI-family instructions: vpcompressw is emitted from the plain AVX-512 block**
*💡 SUGGEST*

The new header comment instructs users to "Opt out with NO_AVX512_VBMI_SUPPORT if the toolchain's assembler cannot emit VBMI." That instruction is incomplete. `mlkem_rej_uniform_avx512` (wc_mlkem_asm.S:31340) and `mlkem_rej_uniform_n_avx512` (wc_mlkem_asm.S:31463) both execute `vpcompressw %zmm4, (%rdi){%k1}` (lines 31366 and 31489), which is an AVX-512VBMI2 instruction, yet both sit inside the plain `#ifdef HAVE_INTEL_AVX512` block rather than a VBMI-gated one. wc_mlkem_asm.asm has the identical distribution (vpcompressw at 30855, 30950 inside the non-VBMI PROCs). Defining NO_AVX512_VBMI_SUPPORT therefore leaves VBMI2-encoded instructions in the object file. Combined with the guard-nesting defect above, an assembler that handles AVX-512F but not the VBMI/VBMI2 encodings cannot be satisfied by either macro alone - only by defining BOTH NO_AVX512_SUPPORT and NO_AVX512_VBMI_SUPPORT, which nothing documents. The runtime dispatch itself is correct: wc_mlkem_poly.c:2330 and :2344 check `IS_INTEL_AVX512_VBMI2(cpuid_flags)` before calling these, so Skylake-X/Cascade Lake correctly fall back to AVX2 and there is no SIGILL risk.

**Recommendation:** Introduce a distinct HAVE_INTEL_AVX512_VBMI2 assembly guard (CPUID_AVX512_VBMI2 already exists on the C side after this commit) and move the two vpcompressw-based rej_uniform variants under it, or at minimum correct the wc_mlkem.h comment to state that NO_AVX512_VBMI_SUPPORT only removes the vpermb-based `*_vbmi` routines and that assemblers lacking VBMI2 also need NO_AVX512_SUPPORT.

Frauschi (commented):
## 🐺 Skoll Code Review

**Findings:** 18 total — 5 posted, 13 skipped

### Posted findings
- [Medium] **mlkem_get_noise_k3/k4_avx512 double the stack buffer and drop the WOLFSSL_SMALL_STACK handling used by the k2 sibling** — `wolfcrypt/src/wc_mlkem_poly.c:4877`
- [Medium] **NO_AVX512_SUPPORT no longer disables all AVX-512 code emission: the VBMI assembly block is not nested under the AVX-512 guard** — `wolfssl/wolfcrypt/wc_mlkem.h:586-598`
- [Low] **New AVX512 declarations are inconsistently const-qualified, named, and ordered** — `wolfssl/wolfcrypt/wc_mlkem.h:696-726`
- [Low] **Doc comments on the new x8 helpers omit @return and misstate lane counts** — `wolfcrypt/src/wc_mlkem_poly.c:4581-4587`
- [Low] **NO_AVX512_VBMI_SUPPORT does not remove all VBMI-family instructions: vpcompressw is emitted from the plain AVX-512 block** — `wolfssl/wolfcrypt/wc_mlkem.h:591-594`

- [Medium] AVX512BW-dependent ML-KEM assembly is dispatched on the AVX512F CPUID bit alone
- [Medium] Four dispatch tiers added with no test that deterministically exercises each one
- [Medium] ML-KEM AVX-512 dispatch gated on the AVX-512F CPUID bit while the assembly requires AVX-512BW (and AVX-512VL)
- [Low] mlkem_csubq_avx512 is declared but never called and has no AVX2 counterpart declaration
- [Low] New `_ins` suffix is undocumented and four helpers are generated by a function-defining macro
- [Low] VBMI guard nesting in wc_mlkem.h does not actually match the assembly guard it claims to mirror
- [Low] mlkem_get_noise_k3_avx512 / k4_avx512 use 1 KB plain stack arrays, ignoring WOLFSSL_SMALL_STACK
- [Low] New AVX-512 CPUID flags are set from feature bits with no OSXSAVE/XCR0 state check
- [Info] AVX-512 matrix generators roughly double the non-small-stack frame; k2 pays the full eight-lane cost to use four lanes
- [Info] The four-lane mlkem_redistribute_*_rand_avx512 dispatch is unreachable in normal builds
- [Info] mlkem_csubq_avx512 is newly declared in the header but has no caller
- [Info] New x8 noise helpers can return MEMORY_E but their doc comments omit @return; k3 comment miscounts lanes
- [Info] No test exercises the new CPUID dispatch variants, so cross-implementation output equivalence is unverified

</details>

wolfcrypt/src/wc_mlkem_poly.c
SparkiDev: fixed

wolfssl/wolfcrypt/wc_mlkem.h
SparkiDev: fixed

wolfssl/wolfcrypt/wc_mlkem.h
SparkiDev: fixed

wolfcrypt/src/wc_mlkem_poly.c
SparkiDev: fixed

wolfssl/wolfcrypt/wc_mlkem.h
SparkiDev: fixed

SparkiDev:
Jenkins: retest this please

SparkiDev:
Jenkins: retest this please

Visual Studio: CRL monitor test failed.

wolfcrypt/src/wc_mlkem_poly.c
```diff
@@ -5102,6 +5796,20 @@ static void mlkem_vec_compress_10_c(byte* r, sword16* v, unsigned int k)
 void mlkem_vec_compress_10(byte* r, sword16* v, unsigned int k)
 {
 #ifdef USE_INTEL_SPEEDUP
+#ifdef WOLFSSL_MLKEM_HAVE_INTEL_AVX512_VBMI
```
Frauschi, line 5799: **🟠 [High] AVX512-VBMI/VBMI2 dispatch branches never test CPUID_AVX512, so cpuid_clear_flag(CPUID_AVX512) no longer disables AVX512 code**
*🚫 BLOCK*  `bug`

Every new VBMI/VBMI2 dispatch branch added by this diff gates only on the sub-feature flag and never on `IS_INTEL_AVX512(cpuid_flags)`. The affected sites are `mlkem_vec_compress_10` (~5799), `mlkem_vec_decompress_10` (~6024), `mlkem_vec_decompress_11` (~6126), `mlkem_compress_4` (~6309), `mlkem_compress_5` (~6401), `mlkem_from_bytes` (~6858), `mlkem_to_bytes` (~6951), plus `mlkem_rej_uniform_n_ins`/`mlkem_rej_uniform_ins` (2326-2369). Because the VBMI branch is tested *before* the AVX512 branch, it wins whenever the VBMI CPUID bit is set.

wolfSSL exposes `cpuid_clear_flag()` as a `WOLFSSL_API` (declared `wolfssl/wolfcrypt/cpuid.h:197`, implemented `wolfcrypt/src/cpuid.c:710`) precisely so integrators can switch an instruction set off at run time. After this diff, `cpuid_clear_flag(CPUID_AVX512)` leaves `CPUID_AVX512_VBMI`/`CPUID_AVX512_VBMI2` set, and ML-KEM keeps executing AVX512 instructions through the `*_vbmi`/`*_vbmi2` routines. The AVX512-only branch immediately below is correctly gated, so the behaviour is inconsistent within the same function.

This also diverges from the established repo convention. `wolfcrypt/src/wc_frodokem_mat.c:1723-1735` nests the sub-feature test (`IS_INTEL_VAES`) *inside* a block already gated on `IS_INTEL_AVX512(cpuid_flags) && (SAVE_VECTOR_REGISTERS2() == 0)`. The ML-KEM code inverts that structure.

(The narrow architectural point that VBMI implies AVX512F on real silicon does not save this: the flags are software state that callers are allowed to modify, and the code depends on an unstated invariant.)

**Suggestion:**
```suggestion
#ifdef WOLFSSL_MLKEM_HAVE_INTEL_AVX512_VBMI
    if (IS_INTEL_AVX512(cpuid_flags) && IS_INTEL_AVX512_VBMI(cpuid_flags) &&
            (SAVE_VECTOR_REGISTERS2() == 0)) {
        mlkem_compress_10_avx512_vbmi(r, v, (int)k);
        RESTORE_VECTOR_REGISTERS();
    }
    else
#endif
```

Frauschi (commented):
## 🐺 Skoll Code Review

**Findings:** 15 total — 1 posted, 14 skipped

### Posted findings
- [High] **AVX512-VBMI/VBMI2 dispatch branches never test CPUID_AVX512, so cpuid_clear_flag(CPUID_AVX512) no longer disables AVX512 code** — `wolfcrypt/src/wc_mlkem_poly.c:5799-5812`

- [Medium] New WOLFSSL_MLKEM_ASM_TEST macro gates assembly with no prototypes and no in-tree consumer
- [Medium] No test exercises the four new rejection-sampler variants or the AVX512 paths on non-AVX512 CI
- [Medium] mlkem_get_noise_k2_avx512 does not advance seed[WC_ML_KEM_SYM_SZ] while its AVX2 sibling does
- [Medium] AVX512 matrix generators double the stack frame in default (non-small-stack) builds
- [Medium] No configure-time probe for the new NO_AVX512_VBMI_SUPPORT / NO_AVX512_VBMI2_SUPPORT opt-outs
- [Medium] AVX-512 VBMI/VBMI2 dispatch bypasses the CPUID_AVX512 master flag, defeating cpuid_clear_flag(CPUID_AVX512)
- [Medium] New AVX-512 matrix generators roughly double single-frame stack usage in non-WOLFSSL_SMALL_STACK builds
- [Low] mlkem_csubq_avx512 is declared and assembled but never called
- [Low] New CPUID flag defines break column alignment and are missing from the benchmark feature banner
- [Low] Unused lanes 4..7 in mlkem_gen_matrix_k2_avx512 duplicate lane 0's XOF input and omit the (word32) cast
- [Low] The _ins suffix on the new dispatch helpers is undocumented and novel to the codebase
- [Low] New static dispatch wrappers are unreferenced in single-parameter-set builds
- [Info] SGX CPUID path cannot express the new VBMI/VBMI2 flags
- [Info] AVX-512 and AVX-2 ML-KEM-512 noise paths now differ in whether they mutate seed[WC_ML_KEM_SYM_SZ]

</details>

wolfcrypt/src/wc_mlkem_poly.c
SparkiDev: fixed

Frauschi (approved):
LGTM

dgarske (commented):
Host: tigerlake-come (i7-1185G7E, Tiger Lake — has AVX512F and AVX512VBMI, both required by this PR). Pinned to core 2, --enable-intelasm --enable-mlkem, ops/sec.

```
┌───────────────┬────────┬───────────────────┬─────────────────────┬─────────┐
│ Parameter set │   Op   │ Before (f6990377) │ After (PR f79d42ed) │ Speedup │
├───────────────┼────────┼───────────────────┼─────────────────────┼─────────┤
│ ML-KEM-512    │ keygen │ 288,015           │ 320,820             │ 1.11x   │
├───────────────┼────────┼───────────────────┼─────────────────────┼─────────┤
│ ML-KEM-512    │ encap  │ 258,098           │ 311,477             │ 1.21x   │
├───────────────┼────────┼───────────────────┼─────────────────────┼─────────┤
│ ML-KEM-512    │ decap  │ 169,250           │ 184,107             │ 1.09x   │
├───────────────┼────────┼───────────────────┼─────────────────────┼─────────┤
│ ML-KEM-768    │ keygen │ 167,535           │ 255,189             │ 1.52x   │
├───────────────┼────────┼───────────────────┼─────────────────────┼─────────┤
│ ML-KEM-768    │ encap  │ 165,123           │ 249,301             │ 1.51x   │
├───────────────┼────────┼───────────────────┼─────────────────────┼─────────┤
│ ML-KEM-768    │ decap  │ 108,039           │ 137,692             │ 1.27x   │
├───────────────┼────────┼───────────────────┼─────────────────────┼─────────┤
│ ML-KEM-1024   │ keygen │ 114,003           │ 193,557             │ 1.70x   │
├───────────────┼────────┼───────────────────┼─────────────────────┼─────────┤
│ ML-KEM-1024   │ encap  │ 107,443           │ 172,734             │ 1.61x   │
├───────────────┼────────┼───────────────────┼─────────────────────┼─────────┤
│ ML-KEM-1024   │ decap  │ 74,098            │ 99,984              │ 1.35x   │
└───────────────┴────────┴───────────────────┴─────────────────────┴─────────┘
```

dgarske (approved):

PR #10983, opened 2026-07-23
Zephyr: wolfSSL module support for the wolfPSA provider and native RTOS use

Frauschi:
## Summary

Extends the wolfSSL Zephyr module to support two coupled efforts: **wolfPSA as Zephyr's PSA Crypto provider** and the **wolfSSL secure-sockets TLS backend**.

## What's included

**Native RTOS threading**
- wolfCrypt's Zephyr port now uses `k_mutex` / `k_thread` / `k_condvar` directly, with no dependency on `CONFIG_POSIX_THREADS`; `k_condvar` use is gated on the kernel version.
- `WOLFSSL_SINGLE_THREADED` defaults from `!MULTITHREADING`, so single-threaded kernels need no manual opt-in and consumers (wolfPSA) don't have to select it.
- New `zephyr/tests/wolfssl_condvar` ztest exercises the native condvar path, wired into CI.

**Configuration interface**
- A user-supplied `CONFIG_WOLFSSL_SETTINGS_FILE` is authoritative; the module never layers Kconfig `#define`s over it.
- Many new Kconfig knobs (`WOLFSSL_CRYPTO_ONLY`, `WOLFSSL_SINGLE_THREADED`) plus new classic-crypto / TLS / PQC feature knobs — RSA, ECC, ChaCha20-Poly1305, Curve25519, SNI, session cache/tickets, and ML-KEM / ML-DSA / LMS / XMSS / Falcon — each with its memory-reduction "small" option.
- Replaces the hardcoded `HAVE_FIPS_VERSION 5` / `_MINOR 3` with a Kconfig `choice` under `WOLFCRYPT_FIPS` (default v6). Version triples mirror `configure.ac`'s `--enable-fips=` mapping exactly. Also defines the `HAVE_FIPS` master switch, which was previously missing and left the version macros inert.

**DRBG seeding**
- `wc_GenerateSeed()` on Zephyr draws seed material from the hardware entropy driver when present (chunked to the entropy API's `uint16_t` length, `DT_HAS_CHOSEN`-guarded), falling back to `sys_rand_get()`. `HAVE_HASHDRBG` stays guarded by `WC_NO_HASHDRBG`.

**Misc**
- `z_time()` reads the native_sim simulator RTC for a real wall clock on native targets, regardless of libc.
- Benchmark stack tracking: `mem_track.h` gains a Zephyr `StackSizeCheck` that runs the benchmark in a measured `k_thread` and reports high-water-mark stack usage.


wolfssl/wolfcrypt/mem_track.h
```diff
@@ -733,7 +733,112 @@ int StackSizeHWMReset(void)
...
+      printf("stack used = %lu\n", (unsigned long)used);
+    }
+#endif
+
+out:
+    XFREE(tid, wolfsslThreadHeapHint, DYNAMIC_TYPE_TMP_BUFFER);
+    XFREE(threadStack, wolfsslThreadHeapHint,
+          DYNAMIC_TYPE_TMP_BUFFER);
+
+    return ret;
```
Frauschi, line 832: fixed

Frauschi:
Jenkins retest this please

wolfssl/wolfcrypt/types.h
```diff
@@ -1958,6 +1958,19 @@ WOLFSSL_API word32 CheckRunTimeSettings(void);
     } THREAD_TYPE;
     #define WOLFSSL_THREAD
     extern void* wolfsslThreadHeapHint;
+    /* Native Zephyr condition variable (k_condvar) built on a k_mutex; no
+     * POSIX pthread layer required. Only reached when !SINGLE_THREADED, so
+     * wolfSSL_Mutex is k_mutex and <zephyr/kernel.h> is already included.
+     * k_condvar was introduced in Zephyr 2.4, so gate the capability on the
+     * kernel version: an older target builds without condition-variable
+     * support (WOLFSSL_COND undefined), exactly as it did before. */
+    #if KERNEL_VERSION_NUMBER >= 0x20400
```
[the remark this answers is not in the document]
Frauschi: This is a false positive. The Zephyr `COND_TYPE`/`WOLFSSL_COND` block is an `#elif defined(WOLFSSL_ZEPHYR)` arm of the `#ifdef SINGLE_THREADED` chain that begins at `types.h:1861`. In a `SINGLE_THREADED` build the first `#ifdef SINGLE_THREADED` arm is selected and this block is never compiled, so `struct k_condvar` can't be incomplete and `WOLFSSL_COND` is never defined here. This mirrors `wc_port.h:328`, which includes `<zephyr/kernel.h>` under the same `#ifndef SINGLE_THREADED`. Adding `!defined(SINGLE_THREADED)` would be redundant with the enclosing chain.

wolfcrypt/src/random.c
```diff
@@ -5328,10 +5328,53 @@ int wc_GenerateSeed(OS_Seed* os, byte* output, word32 sz)
         #include <posix/time.h>
     #endif
 
+    #if defined(CONFIG_ENTROPY_HAS_DRIVER) && DT_HAS_CHOSEN(zephyr_entropy)
```
danielinux, line 5451: This condition invokes DT_HAS_CHOSEN before anything in this block includes <zephyr/devicetree.h> 
 (the includes above it are only version.h, sys/types.h, random/random.h, posix/time.h). The preprocessor expands the whole expression regardless of the &&, so if DT_HAS_CHOSEN isn't a macro it becomes 0 (0) hard error ("missing binary operator before token '('"), not a graceful fallback. It works today because wc_port.h:336-340 pulls in <zephyr/kernel.h>  but only #ifndef SINGLE_THREADED, which is exactly the config this PR newly auto-enables (Kconfig:121, default y if !MULTITHREADING). CI is green because native_sim has MULTITHREADING=y.

Proposed fix: Verify on a CONFIG_MULTITHREADING=n build. Either include <zephyr/devicetree.h> explicitly at  line 5330, or nest the guards: `#ifdef CONFIG_ENTROPY_HAS_DRIVER` / `#if DT_HAS_CHOSEN(zephyr_entropy)`.

.wolfssl_known_macro_extras
```diff
@@ -840,15 +861,8 @@ WOLFSSL_EVP_PRINT
 WOLFSSL_EXPORT_INT
 WOLFSSL_EXPORT_SPC_SZ
 WOLFSSL_EXTRA
-WOLFSSL_FALCON_FFT_AVX2
```
danielinux: These removals don't seem related. Rebase on latest master and ensure known macro tests succeed

danielinux (changes requested):

Frauschi:
Jenkins retest this please

wolfcrypt/src/random.c
Frauschi: fixed by adding the include of `devicetree.h`.

.wolfssl_known_macro_extras
Frauschi: Re-added them.

danielinux (dismissed):

zephyr/user_settings.h
```diff
@@ -82,9 +89,38 @@ extern "C" {
 /* FIPS */
 /* ------------------------------------------------------------------------- */
 #ifdef CONFIG_WOLFCRYPT_FIPS
-    /* FIPS Ready */
-    #define HAVE_FIPS_VERSION 5
-    #define HAVE_FIPS_VERSION_MINOR 3
+    /* HAVE_FIPS is the master switch that routes the wolfCrypt algorithms
```
dgarske, line 92: Let's make sure `HAVE_FIPS_VERSION`, `HAVE_FIPS_VERSION_MINOR` and `HAVE_FIPS_VERSION_PATCH` can be overridden manually.

zephyr/user_settings.h
```diff
@@ -358,13 +409,52 @@ extern "C" {
...
+    #define HAVE_FALCON
+    /* Small-memory dynamic signer instead of the fast tree-signer (keeps full
+     * sign+verify); the default portable integer FPR backend needs no ASM. */
+    #define WOLFSSL_FALCON_SIGN_SMALL_MEM
+#endif
+
+/* SHA-3 / SHAKE are required by ML-KEM, ML-DSA, and Falcon; enable them (small
+ * variant) when any is on, and explicitly disable SHAKE otherwise to keep a
+ * non-PQC build lean. */
+#if defined(CONFIG_WOLFSSL_MLKEM) || defined(CONFIG_WOLFSSL_MLDSA) || \
```
dgarske, line 506: Do we need XMSS in this list too? `CONFIG_WOLFSSL_XMSS`

wolfssl/wolfcrypt/mem_track.h
```diff
@@ -733,7 +733,112 @@ int StackSizeHWMReset(void)
...
+
+#ifdef HAVE_STACK_SIZE_VERBOSE
+    struct stack_size_debug_context shim_args;
+#endif
+
+    tid = (struct k_thread*)XMALLOC(
+            Z_KERNEL_STACK_SIZE_ADJUST(sizeof(struct k_thread)),
+            wolfsslThreadHeapHint, DYNAMIC_TYPE_TMP_BUFFER);
+    if (tid == NULL) {
+        printf("error: XMALLOC tid failed");
```
dgarske: Does this need `\n`?

wolfssl/wolfcrypt/mem_track.h
```diff
@@ -733,7 +733,112 @@ int StackSizeHWMReset(void)
...
+
+#ifndef WOLFSSL_ZEPHYR_STACK_SZ
+    #define WOLFSSL_ZEPHYR_STACK_SZ (48*1024)
+#endif
+
+    threadStack = (void*)XMALLOC(
+                Z_KERNEL_STACK_SIZE_ADJUST(WOLFSSL_ZEPHYR_STACK_SZ),
+                wolfsslThreadHeapHint, DYNAMIC_TYPE_TMP_BUFFER);
+    if (threadStack == NULL) {
+        printf("error: k_thread_stack_alloc threadStack failed");
```
dgarske: Does this need `\n`?

wolfssl/wolfcrypt/mem_track.h
```diff
@@ -733,7 +733,112 @@ int StackSizeHWMReset(void)
...
+    }
+
+#ifdef HAVE_STACK_SIZE_VERBOSE
+    printf("stack used = %lu\n", StackSizeCheck_stackSizeHWM > (WOLFSSL_ZEPHYR_STACK_SZ - i)
+        ? (unsigned long)StackSizeCheck_stackSizeHWM
+        : (unsigned long)(WOLFSSL_ZEPHYR_STACK_SZ - i));
+    StackSizeCheck_myStack = NULL;
+    StackSizeCheck_stackOffsetPointer = NULL;
+#else
+    {
```
dgarske: Please avoid the empty braces.

wolfssl/wolfcrypt/mem_track.h
```diff
@@ -733,7 +733,112 @@ int StackSizeHWMReset(void)
 
 #endif /* HAVE_STACK_SIZE_VERBOSE */
 
-#ifdef HAVE_PTHREAD
+#if defined(__ZEPHYR__)
```
dgarske: Maybe use `WOLFSSL_ZEPHYR` (combined with the `!SINGLE_THREADED`)

dgarske (changes requested):

wolfssl/wolfcrypt/mem_track.h
Frauschi: Changed to `#if defined(WOLFSSL_ZEPHYR) && !defined(SINGLE_THREADED)`. The `!SINGLE_THREADED` half turned out to be load bearing: this PR auto-enables `WOLFSSL_SINGLE_THREADED` when `CONFIG_MULTITHREADING=n`, and `types.h` only declares `THREAD_TYPE` and `wolfsslThreadHeapHint` in the non-single-threaded chain, so that configuration failed to compile before this change. Verified both ways: single-threaded now builds, and the multithreaded build reports the same stack figure as it did with `__ZEPHYR__`.

wolfssl/wolfcrypt/mem_track.h
Frauschi: Yes, added.

wolfssl/wolfcrypt/mem_track.h
Frauschi: Added, and fixed the wording: the call above it is an `XMALLOC`, not `k_thread_stack_alloc`. The missing newline was actively hiding a real failure, see the general note on the PR.

wolfssl/wolfcrypt/mem_track.h
Frauschi: Removed, the subtraction is inlined into the `printf` now. There is an identical block in the `HAVE_PTHREAD` branch that predates this PR; happy to clean that up here too if you want it in scope.

zephyr/user_settings.h
Frauschi: Each version macro is now `#ifndef` guarded so a `-D` on the compile line or a project header wins, and `HAVE_FIPS_VERSION_MAJOR` is derived the way `configure.ac` sets it. While checking this I found we were not defining `WOLFSSL_FIPS_READY`, which `configure.ac` sets for both `--enable-fips=v7` and `=ready`. That made a Zephyr v7 build select `WC_FIPS_186_5` where the autotools v7 build selects `WC_FIPS_186_4`. Both arms define it now, and all five arms match `configure.ac`. One caveat worth noting: `settings.h` `#undef`s the triple unconditionally for FIPS-Ready/dev, so a manual override is only honoured for v2/v5/v6.

zephyr/user_settings.h
Frauschi: Good catch, added. `wc_xmss.h` derives `WC_XMSS_SHAKE128/256` straight from `WOLFSSL_SHAKE128/256`, so an XMSS-only build was hitting the `#else` arm and silently losing every SHAKE based XMSS/XMSSMT parameter set. LMS I left out deliberately: its SHAKE-256 support is opt-in through `WOLFSSL_LMS_SHAKE256` and is never derived from `WOLFSSL_SHAKE256`. Flash cost for an XMSS-only build is +6.7 KB.

Frauschi:
Addressed all six review comments, thanks. Both commits were amended and force-pushed.

One thing worth flagging that came out of the newline fix: the stack tracking was never actually running. No Kconfig option enabled `HAVE_STACK_SIZE`, and the benchmark sample's 8 KB malloc arena could not hold the 48 KB thread stack, so the allocation failed every time and the error was invisible precisely because of the missing newline. The sample now sets `HAVE_STACK_SIZE` and a 64 KB arena (48 KB stack plus the `k_thread` object, leaving about 16 KB for the benchmark's own allocations), and prints the stack figure.

dgarske (dismissed):

dgarske (approved):

dgarske:
Jenkins retest this please

dgarske:
Jenkins retest this please

PR #7018, opened 2023-11-30
Fixes for TI AES and SHA

dgarske:
# Description

Fixes for TI AES and SHA on TM4C with HW acceleration.
Adds full AES GCM and CCM support with TLS.

Fixes ZD 17079

# Testing

CCS and TM4C

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen
 - [ ] updated appropriate READMEs
 - [ ] Updated manual and documentation

PR #10986, opened 2026-07-23
Fix TI port AES-CTR streaming and hashCopy correctness bugs

dgarske:
## Description

Fixes two bugs in the TI (TivaWare TM4C129 AES/SHA-MD5 hardware) port. Both silently produce wrong output rather than failing, and both trace to PR #7018.

### F-3078 - `wc_AesCtrEncrypt` drops partial streaming input (`wolfcrypt/src/port/ti/ti-aes.c`)

When AES-CTR is called in streaming fashion and a follow-up chunk does not complete the buffered block (`aes->left + sz < WC_AES_BLOCK_SIZE`), the leading carry-over branch buffered the new plaintext but never encrypted it, never wrote the caller's output, and never advanced `aes->left`. The caller saw uninitialized/stale output, and the next call overwrote the still-buffered bytes, silently losing plaintext. Fixed by adding the `else` path that mirrors the existing tail-handling block: encrypt the partial block with `AES_CFG_MODE_CTR_NOCTR` (no counter advance), emit the new bytes, and do `aes->left += odd`.

### F-2646 - `hashCopy` discards accumulated message (`wolfcrypt/src/port/ti/ti-hash.c`)

`hashCopy` zeroed `msg`/`used`/`len` and copied only the never-populated `hash` field, so a copied context finalized to the hash of an empty message instead of the source data. Affects `wc_Md5Copy`, `wc_ShaCopy`, `wc_Sha224Copy`, and `wc_Sha256Copy`. Fixed by allocating a fresh `dst->msg` and copying `src->len` bytes (independent allocation preserves the original anti-double-free intent), matching the reference `wc_Sha256Copy` in `devcrypto_hash.c`.

### DIFF
--- a/wolfcrypt/src/port/ti/ti-aes.c
+++ b/wolfcrypt/src/port/ti/ti-aes.c
@@ -251,6 +251,17 @@ int wc_AesCtrEncrypt(Aes* aes, byte* out, const byte* in, word32 sz)
             aes->left = 0;
             XMEMSET(tmp, 0x0, WC_AES_BLOCK_SIZE);
         }
+        else if (odd > 0) {
+            /* buffered bytes do not fill a block: encrypt the partial block
+             * without advancing the counter, emit the new bytes, and keep
+             * the remaining buffered bytes for the next call */
+            ret = AesProcess(aes, (byte*)out_block, (byte const *)tmp, WC_AES_BLOCK_SIZE,
+                        AES_CFG_DIR_ENCRYPT, AES_CFG_MODE_CTR_NOCTR);
+            if (ret != 0)
+                return ret;
+            XMEMCPY(out, out_block+aes->left, odd);
+            aes->left += odd;
+        }
         in += odd;
         out+= odd;
         sz -= odd;
--- a/wolfcrypt/src/port/ti/ti-hash.c
+++ b/wolfcrypt/src/port/ti/ti-hash.c
@@ -127,10 +127,24 @@ static int hashCopy(wolfssl_TI_Hash *src, wolfssl_TI_Hash *dst)
 {
     if (src == NULL || dst == NULL)
         return BAD_FUNC_ARG;
-    /* only copy hash, zero the rest of the struct to avoid double-free */
+    if (src == dst)
+        return 0;
+    /* release any buffer dst already owns, then copy the accumulated message
+     * into a fresh buffer so each descriptor owns its own allocation and can
+     * be freed independently */
+    XFREE(dst->msg, NULL, DYNAMIC_TYPE_TMP_BUFFER);
     dst->msg = NULL;
-    dst->used = 0;
-    dst->len = 0;
+    if ((src->msg != NULL) && (src->len > 0)) {
+        dst->msg = (byte*)XMALLOC(src->len, NULL, DYNAMIC_TYPE_TMP_BUFFER);
+        if (dst->msg == NULL) {
+            dst->used = 0;
+            dst->len  = 0;
+            return MEMORY_E;
+        }
+        XMEMCPY(dst->msg, src->msg, src->len);
+    }
+    dst->used = src->used;
+    dst->len  = src->len;
     XMEMCPY(dst->hash, src->hash, sizeof(dst->hash));
     return 0;
 }


dgarske:
Jenkins retest this please. valgrind with openssl-all

Frauschi (commented):
## 🐺 Skoll Code Review

**Findings:** 9 total — 3 posted, 6 skipped

### Posted findings
- [Medium] **hashCopy leaves dst in an inconsistent state when XMALLOC fails** — `wolfcrypt/src/port/ti/ti-hash.c:132-138`
- [Medium] **hashCopy overwrites a pre-existing dst->msg without freeing it** — `wolfcrypt/src/port/ti/ti-hash.c:130`
- [Low] **wc_AesCtrEncrypt(sz == 0) with buffered bytes now performs a pointless HW AES block** — `wolfcrypt/src/port/ti/ti-aes.c:254`

- [Medium] TI port fixes are not exercised by CI; confirm on-target validation
- [Low] Copying src->len instead of src->used couples the copy to the allocator's capacity
- [Low] New CTR carry-over branch duplicates the existing tail-fragment block
- [Low] hashCopy leaves dst in an inconsistent state when the new allocation fails
- [Info] Added lines exceed the 80-column limit
- [Info] New AES-CTR carry-over branch runs a hardware AES pass for a zero-length call

</details>

wolfcrypt/src/port/ti/ti-hash.c
```diff
@@ -127,10 +127,19 @@ static int hashCopy(wolfssl_TI_Hash *src, wolfssl_TI_Hash *dst)
 {
     if (src == NULL || dst == NULL)
         return BAD_FUNC_ARG;
-    /* only copy hash, zero the rest of the struct to avoid double-free */
-    dst->msg = NULL;
-    dst->used = 0;
-    dst->len = 0;
+    /* copy the accumulated message into a fresh buffer so each descriptor
+     * owns its own allocation and can be freed independently */
+    dst->used = src->used;
```
Frauschi: **🟡 [Medium] hashCopy leaves dst in an inconsistent state when XMALLOC fails**
*💡 SUGGEST*  `bug`

`dst->used` and `dst->len` are assigned *before* the allocation is attempted. If `XMALLOC` fails, the function returns `MEMORY_E` with `dst->msg == NULL` but `dst->used`/`dst->len` set to non-zero values from `src`. That state is unsafe: `hashGetHash()`/`hashFinal()` on such a context calls `ROM_SHAMD5DataProcess(SHAMD5_BASE, (uint32_t*)hash->msg /* NULL */, hash->used /* non-zero */, h)`, i.e. a DMA read from address 0, and `hashUpdate()` would append at `hash->msg + hash->used` over a freshly-allocated buffer whose first `used` bytes are garbage, silently producing a wrong digest. Before this PR `hashCopy` could not fail after the NULL check, so no caller has ever had to defend against a half-copied context. Every in-tree caller does check the return code, so this only bites under OOM plus an ignoring caller — but the fix is two lines and the invariant is worth keeping.

**Suggested shape** (replaces lines 132-142, ordering fix only):

```c
    if ((src->msg != NULL) && (src->len > 0)) {
        dst->msg = (byte*)XMALLOC(src->len, NULL, DYNAMIC_TYPE_TMP_BUFFER);
        if (dst->msg == NULL) {
            dst->used = 0;
            dst->len  = 0;
            return MEMORY_E;
        }
        XMEMCPY(dst->msg, src->msg, src->len);
    }
    else {
        dst->msg = NULL;
    }
    dst->used = src->used;
    dst->len  = src->len;
```


**Suggestion:** Publish `dst->used`/`dst->len` only after the allocation succeeds, and zero them on the `MEMORY_E` path, so a failed copy leaves `dst` fully reset.

wolfcrypt/src/port/ti/ti-hash.c
```diff
@@ -127,10 +127,19 @@ static int hashCopy(wolfssl_TI_Hash *src, wolfssl_TI_Hash *dst)
 {
     if (src == NULL || dst == NULL)
         return BAD_FUNC_ARG;
-    /* only copy hash, zero the rest of the struct to avoid double-free */
-    dst->msg = NULL;
-    dst->used = 0;
-    dst->len = 0;
+    /* copy the accumulated message into a fresh buffer so each descriptor
```
Frauschi: **🟡 [Medium] hashCopy overwrites a pre-existing dst->msg without freeing it**
*💡 SUGGEST*  `bug`

The new comment states the intent is that "each descriptor owns its own allocation and can be freed independently", but `dst->msg` is assigned unconditionally without releasing whatever `dst` already owned. A `dst` that has been updated before the copy leaks its buffer. This is not a new leak — the previous code did `dst->msg = NULL` and leaked the same buffer — but the PR is precisely a rework of this ownership contract, and the contract is not complete without releasing `dst`. The path is live in-tree: `wolfcrypt/test/test.c` `sha256_copy_test()` does `wc_Sha256Update(shaCopy, "abc", 3)` and then `wc_Sha256Copy(sha, shaCopy)`, i.e. it copies into a context that already holds an allocation. The reference implementation the PR description cites, `wc_Sha256Copy()` in `wolfcrypt/src/port/devcrypto/devcrypto_hash.c:220`, calls `wc_InitSha256_ex(dst, ...)` first for exactly this reason (note that a plain re-init would still leak here, since TI's `hashInit` only zeroes `msg`; TI needs `hashFree(dst)`).

**Suggested shape** (insert after the NULL check at line 129):

```c
    if (src == dst)
        return 0;
    /* release any buffer dst already owns before taking a new one */
    XFREE(dst->msg, NULL, DYNAMIC_TYPE_TMP_BUFFER);
    dst->msg = NULL;
```


**Suggestion:** Free whatever `dst` already owns before taking a new allocation, and short-circuit the `src == dst` case.

wolfcrypt/src/port/ti/ti-aes.c
```diff
@@ -251,6 +251,17 @@ int wc_AesCtrEncrypt(Aes* aes, byte* out, const byte* in, word32 sz)
             aes->left = 0;
             XMEMSET(tmp, 0x0, WC_AES_BLOCK_SIZE);
         }
+        else {
```
Frauschi: **🔵 [Low] wc_AesCtrEncrypt(sz == 0) with buffered bytes now performs a pointless HW AES block**
*🔧 NIT*  `bug`

When the function is called with `sz == 0` while `aes->left != 0`, the first branch is entered, `odd` is computed as `sz` (0), and control now falls into the new `else`. That runs a full `AesProcess()` — CCM mutex lock, `ROM_AESReset`, key/IV load, a 16-byte block through the HW engine — and then does `XMEMCPY(out, out_block + aes->left, 0)` and `aes->left += 0`. The result is functionally correct (nothing is written, no counter movement, state unchanged), but it is a wasted hardware round-trip and a wasted mutex acquisition on a call that previously did nothing. Reachability is narrow: no in-tree caller hits it. The zero-length calls in `tests/api/test_aes.c:2447-2459` all pass at least one NULL pointer and return `BAD_FUNC_ARG` at the argument check, and `wolfcrypt/src/evp.c:631` is only reached from `evpCipherBlock()` with `ctx->block_size` or `blocks * ctx->block_size`, never 0. It takes a direct `wc_AesCtrEncrypt(aes, out, in, 0)` from application code, which the public API permits.

**Suggestion:**
```suggestion
else if (odd > 0) {
```

Frauschi (approved):

PR #10990, opened 2026-07-24
Add compile-time crypto callback async poll for record ciphers

julek-wolfssl:
## Summary
- Adds `WOLF_CRYPTO_CB_ASYNC_POLL`, giving crypto callback devices a QAT/Nitrox-style "poll to fill output" completion model for TLS record ciphers (AES and 3DES) instead of the re-invoke model, fixing silent record corruption when a callback returns `WC_PENDING_E`.
- Without the feature (and without a software/QAT/Cavium backend), a pending bulk cipher op now errors out with `ASYNC_OP_E` instead of corrupting the record.

## Test plan
- [ ] `tests/api/test_async.c` covers direct AES-GCM/CBC/CCM and 3DES poll completion at multiple pend depths
- [ ] Negative cases for cipher types defined but not dispatched (ChaCha, single DES)
- [ ] Full TLS 1.3 handshake+echo including the no-poll failure path


src/internal.c
```diff
@@ -21626,7 +21583,16 @@ static WC_INLINE int Encrypt(WOLFSSL* ssl, byte* out, const byte* input,
         #ifdef WOLFSSL_ASYNC_CRYPT
             /* If pending, then leave and return will resume below */
             if (ret == WC_NO_ERR_TRACE(WC_PENDING_E)) {
+            #if defined(WOLF_CRYPTO_CB) && \
+                !defined(WOLF_CRYPTO_CB_ASYNC_POLL) && \
+                !defined(WOLFSSL_ASYNC_CRYPT_SW) && \
+                !defined(HAVE_INTEL_QA) && !defined(HAVE_CAVIUM)
+                /* No completion path for a pending bulk cipher op. */
+                WOLFSSL_ERROR_VERBOSE(ASYNC_OP_E);
+                return ASYNC_OP_E;
```
dgarske, line 21592: Should we also reset `ssl->encrypt.state = CIPHER_STATE_BEGIN;`? I think it might be handled elsewhere in the async code path, but worth checking.

wolfcrypt/src/cryptocb.c
```diff
@@ -453,6 +460,37 @@ int wc_CryptoCb_GetDevIdAtIndex(int startIdx)
     return devId;
 }
 
+#if defined(WOLFSSL_ASYNC_CRYPT) && defined(WOLF_CRYPTO_CB_ASYNC_POLL)
+/* Returns WC_PENDING_E while in-flight, 0 or an error when done. */
+int wc_CryptoCb_Poll(int devId)
```
dgarske, line 465: Should this API also support passing `WC_ASYNC_DEV* dev`?

src/internal.c
```diff
@@ -21626,7 +21583,16 @@ static WC_INLINE int Encrypt(WOLFSSL* ssl, byte* out, const byte* input,
         #ifdef WOLFSSL_ASYNC_CRYPT
             /* If pending, then leave and return will resume below */
             if (ret == WC_NO_ERR_TRACE(WC_PENDING_E)) {
+            #if defined(WOLF_CRYPTO_CB) && \
```
dgarske: Consider a new helper macro to avoid this logic and put in settings.h?

wolfssl/wolfcrypt/async.h
```diff
@@ -391,6 +391,12 @@ typedef struct WC_ASYNC_DEV {
 #elif defined(WOLFSSL_ASYNC_CRYPT_SW)
     WC_ASYNC_SW         sw;
 #endif
+#if defined(WOLF_CRYPTO_CB) && defined(WOLF_CRYPTO_CB_ASYNC_POLL)
+    /* Crypto callback to re-enter at poll time. INVALID_DEVID: none. */
+    struct {
```
dgarske: Do we really need a struct? Maybe just use devId directly? I guess this might make it easier to extend.

wolfcrypt/src/async.c
```diff
@@ -720,11 +775,27 @@ int wolfAsync_EventQueuePoll(WOLF_EVENT_QUEUE* queue, void* context_filter,
...
+                         * so a completed result is not overwritten. */
+                        if (event->ret == WC_NO_ERR_TRACE(WC_PENDING_E)) {
+                            event->ret =
+                                wc_CryptoCb_Poll(asyncDev->cryptocb.devId);
+                        }
+                    }
+                    else
+                    #endif
+                    {
+                        /* Re-invoke model: crypto/PK callbacks manage their own
```
dgarske: This block is empty. Consider removing it.

dgarske (changes requested):

wolfcrypt/src/cryptocb.c
julek-wolfssl: cryptocb API only operate on the `devId`.

julek-wolfssl:
@dgarske addressed review

dgarske (dismissed):

julek-wolfssl:
retest this please

JacobBarthelmeh (approved):

PR #10991, opened 2026-07-24
CI: set ccache path so settings are saved on initial seed

padelsbach:
# Description

Set the path prior to running `ccache --set-config` so that the settings end up in the right path on the first run (or after the weekly re-seed).

Prior to this change, the settings are saved to a different directory. Cache keys contain the settings, so the first run has incorrect keys. On subsequent loads, the cache is already restored and the keys are correct, but there are cache misses on this run unnecessarily.

Note this does not impact existing caches.

Found while adding `ccache` to wolfHSM.

# Testing

How did you test?

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/.github/actions/ccache-setup/action.yml
+++ b/.github/actions/ccache-setup/action.yml
@@ -83,6 +83,13 @@ runs:
     - name: Configure ccache and PATH
       shell: bash
       run: |
+        # Set CCACHE_DIR for the commands below too, not only via
+        # GITHUB_ENV (which applies to later steps): without it the
+        # settings land in a different default dir on runs where
+        # ~/.ccache does not exist yet, and seed runs then compile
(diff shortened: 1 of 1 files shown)


padelsbach:
jenkins retest this please

.github/actions/ccache-setup/action.yml
dgarske: Change to `echo "CCACHE_DIR=$CCACHE_DIR" >> "$GITHUB_ENV"` to ensure a change above doesn't break things here in the future.

dgarske (changes requested):

.github/actions/ccache-setup/action.yml
padelsbach: fixed

dgarske (approved):

PR #10994, opened 2026-07-27
Xilinx Versal Gen2 ASU port: AES ciphers, CMAC, GMAC offload

night1rider:
Add AES, CMAC, and GMAC offload for Xilinx ASU port


wolfcrypt/src/port/xilinx/versal_gen2_asu/asu_cmac.c
```diff
@@ -0,0 +1,318 @@
...
+
+    /* init(): capture the key and start a per-context buffer. AES-192 and other
+     * unsupported key sizes are declined so wolfSSL handles the whole CMAC. */
+    if (info->cmac.key != NULL) {
+        u32 keySize = 0;
+        if (info->cmac.keySz > sizeof(((AsuCmacKeep*)0)->key) ||
+            (wc_AsuCmacKeySize(info->cmac.keySz, &keySize) != 0)) {
+            return CRYPTOCB_UNAVAILABLE;
+        }
+        keep = (AsuCmacKeep*)XMALLOC(sizeof(AsuCmacKeep), NULL,
```
night1rider, line 278: https://github.com/wolfSSL/wolfssl/pull/10886 Free callback will be added here.

wolfcrypt/src/port/xilinx/versal_gen2_asu/asu_cmac.c
```diff
@@ -0,0 +1,321 @@
...
+        return BAD_FUNC_ARG;
+    }
+    if (info->cmac.type != WC_CMAC_AES) {
+        return CRYPTOCB_UNAVAILABLE;
+    }
+    cmac = info->cmac.cmac;
+
+    /* Single-call generate (key, message and output all supplied). A whole-block
+     * message goes to the ASU; anything else runs in wolfSSL software. */
+    if (info->cmac.key != NULL && info->cmac.out != NULL) {
```
JacobBarthelmeh, line 253: There is a case where "key" could be NULL but both "in" and "out" are set. Looking through the conditions here I think that edge case would cause an incorrect CMAC tag to be created with the last input block not accounted for. Try with:

```
wc_InitCmac_ex()
wc_AesCmacGenerate_ex() // where 'key == NULL'
```

JacobBarthelmeh (changes requested):

night1rider:
rebased off master to fix some CI test issues

wolfcrypt/src/port/xilinx/versal_gen2_asu/asu_cmac.c
night1rider: Fixed. When the key was set at init and the one shot generate is called with key as `NULL`, the message arrives as `info->cmac.in` on the final call. That branch now adds it to the buffered message before running the CMAC, so the last input is no longer dropped.

JacobBarthelmeh (approved):

PR #10997, opened 2026-07-28
X509_verify_cert: avoid mutating shared store cert stack

julek-wolfssl:
`wolfSSL_X509_verify_cert` built certificate chains directly on
`ctx->store->certs`, which is shared by every connection using the
store, with no locking. Caller intermediates were pushed onto it,
retries reordered it, and injected entries were removed again on
exit, so concurrent verifications could race and corrupt the shared
stack.

Now the verification works on a per-verification shallow copy of the
cert stack instead, leaving `store->certs` and `setTrustedSk`
read-only.

Fixes F-7229

### DIFF
--- a/src/x509_str.c
+++ b/src/x509_str.c
@@ -654,13 +654,17 @@ static int X509StoreChainPush(WOLF_STACK_OF(WOLFSSL_X509)* chain,
 static int X509StoreCertInStack(WOLF_STACK_OF(WOLFSSL_X509)* stack,
         WOLFSSL_X509* cert)
 {
-    WOLFSSL_STACK* node;
+    int i;
+    int num;
 
     if (stack == NULL || cert == NULL)
         return 0;
 
-    for (node = stack; node != NULL; node = node->next) {
-        if (node->data.x509 == cert)
+    /* Index by logical position like the other helpers in this file rather
+     * than walking raw nodes. */
+    num = wolfSSL_sk_X509_num(stack);
+    for (i = 0; i < num; i++) {
+        if (wolfSSL_sk_X509_value(stack, i) == cert)
             return 1;
     }
 
@@ -708,7 +712,11 @@ static int X509DerEquals(WOLFSSL_X509* cur, WOLFSSL_X509* x509)
  * set0_trusted_stack override - captured before any intermediates were
  * injected for this verification call) or in store->trusted.  Returns 0
  * otherwise.  Used by the X509_V_FLAG_PARTIAL_CHAIN fallback to confirm that
- * a chain actually terminates at a caller-trusted certificate. */
+ * a chain actually terminates at a caller-trusted certificate.
+ * NOTE: origTrustedSk is a private snapshot, but store->trusted is read live
+ * and unlocked here (as it is at the terminal issuer lookup); this mitigation
+ * is deliberately asymmetric, so a single X509_STORE must not be shared across
+ * threads verifying concurrently. */
 static int X509StoreCertIsTrusted(WOLFSSL_X509_STORE* store,
         WOLFSSL_X509* x509, WOLF_STACK_OF(WOLFSSL_X509)* origTrustedSk)
 {
@@ -866,16 +874,19 @@ int wolfSSL_X509_verify_cert(WOLFSSL_X509_STORE_CTX* ctx)
      * are appended and X509VerifyCertSetupRetry moves failed certs out of it.
      * store->certs is shared by every connection using this store and
      * setTrustedSk is owned by the caller, so build a per-verification shallow
-     * copy (certsToUse) and leave both untouched.
+     * copy (certsToUse) and leave both untouched. This removes the write-side
+     * corruption a concurrent verification used to inflict on those stacks, but
+     * it does NOT make concurrent use of one X509_STORE safe: the dup is itself
+     * an unlocked read, and store->trusted is still walked live (see the NOTE
+     * below). A single store must not be shared across threads that verify
+     * concurrently.
      *
      * The X509_V_FLAG_PARTIAL_CHAIN fallback needs the set of certs that were
-     * caller-trusted before any intermediates were injected. Take a second,
-     * independent snapshot (origTrustedSk) for that check rather than borrowing
-     * a pointer into store->certs/setTrustedSk: chain building reorders
-     * certsToUse, and another thread may add to or detach store->certs while
-     * this verification is in flight, so a borrowed pointer could be reordered
-     * or freed underneath us. Both dups hold borrowed references and are
-     * shallow-freed at exit. */
+     * caller-trusted before any intermediates were injected. Snapshot it from
+     * certsToUse - the private copy, taken before addAllButSelfSigned() injects
+     * intermediates - rather than walking the shared stack a second time, so
+     * the two snapshots are provably identical. Both dups hold borrowed
+     * references and are shallow-freed at exit. */
     callerTrusted = ctx->store->certs;
     if (ctx->setTrustedSk != NULL) {
         callerTrusted = ctx->setTrustedSk;
@@ -883,7 +894,8 @@ int wolfSSL_X509_verify_cert(WOLFSSL_X509_STORE_CTX* ctx)
 
     if (callerTrusted != NULL) {
         certsToUse = wolfSSL_shallow_sk_dup(callerTrusted);
-        origTrustedSk = wolfSSL_shallow_sk_dup(callerTrusted);
+        if (certsToUse != NULL)
+            origTrustedSk = wolfSSL_shallow_sk_dup(certsToUse);
         if (origTrustedSk == NULL) {
             ret = WOLFSSL_FAILURE;
             goto exit;
@@ -1073,7 +1085,10 @@ int wolfSSL_X509_verify_cert(WOLFSSL_X509_STORE_CTX* ctx)
              * chain.  setTrustedSk / store->trusted are searched by name+AKID,
              * not by signature, and setTrustedSk is no longer pruned during
              * chain building, so X509StoreGetIssuerEx can return a same-subject
-             * cert that was tried and rejected. */
+             * cert that was tried and rejected.
+             * Under WOLFSSL_SIGNER_DER_CERT the issuer above is a freshly
+             * allocated CM copy, never pointer-equal to a failedCerts entry, so
+             * this guard is a no-op on that path. */
             if (issuer != NULL && !X509StoreCertInStack(failedCerts, issuer)) {
                 X509StoreChainPush(ctx->chain, issuer);
             }
--- a/src/x509_str.c
+++ b/src/x509_str.c
@@ -632,37 +632,6 @@ static int X509StoreMoveCert(WOLFSSL_STACK *certs_stack,
     return WOLFSSL_FAILURE;
 }
 
-/* Remove the first node referencing `cert` (by pointer identity) from `stack`.
- * The certificate object itself is not freed - the stack only holds a borrowed
- * reference. Returns WOLFSSL_SUCCESS if a node was removed, WOLFSSL_FAILURE if
- * `cert` was not present, or WOLFSSL_FATAL_ERROR if `stack`/`cert` is NULL.
- * The only caller performs best-effort cleanup and intentionally ignores the
- * return value.
- *
- * Walks the linked list once (O(n)) rather than indexing with
- * wolfSSL_sk_X509_value() per position (which would re-walk from the head each
- * time, O(n^2)). */
-static int X509StoreRemoveCert(WOLFSSL_STACK *stack, WOLFSSL_X509 *cert) {
-    WOLFSSL_STACK* node;
-    int idx;
-    int num;
-
-    if (stack == NULL || cert == NULL)
-        return WOLFSSL_FATAL_ERROR;
-
-    num = wolfSSL_sk_X509_num(stack);
-    for (node = stack, idx = 0; idx < num && node != NULL;
-            node = node->next, idx++) {
-        if (node->data.x509 == cert) {
-            (void)wolfSSL_sk_pop_node(stack, idx);
-            return WOLFSSL_SUCCESS;
-        }
-    }
-
-    return WOLFSSL_FAILURE;
-}
-
-
 /* Push x509 onto the ctx chain with its own reference, like OpenSSL.
  * The chain owns a reference to each of its certs. */
 static int X509StoreChainPush(WOLF_STACK_OF(WOLFSSL_X509)* chain,
@@ -715,12 +684,11 @@ static int X509DerEquals(WOLFSSL_X509* cur, WOLFSSL_X509* x509)
                    x509->derCert->length) == 0;
 }
 
-/* Returns 1 if x509's DER matches an entry in either origTrustedSk (an
- * immutable snapshot of the caller's trusted set captured before any
- * intermediates were injected for this verification call) or in
- * store->trusted.  Returns 0 otherwise.  Used by the
- * X509_V_FLAG_PARTIAL_CHAIN fallback to confirm that a chain actually
- * terminates at a caller-trusted certificate. */
+/* Returns 1 if x509's DER matches an entry in either origTrustedSk (the
+ * caller's trusted stack - store->certs or the set0_trusted_stack override -
+ * which chain building leaves unmodified) or in store->trusted.  Returns 0
+ * otherwise.  Used by the X509_V_FLAG_PARTIAL_CHAIN fallback to confirm that
+ * a chain actually terminates at a caller-trusted certificate. */
 static int X509StoreCertIsTrusted(WOLFSSL_X509_STORE* store,
         WOLFSSL_X509* x509, WOLF_STACK_OF(WOLFSSL_X509)* origTrustedSk)
 {
@@ -859,8 +827,6 @@ int wolfSSL_X509_verify_cert(WOLFSSL_X509_STORE_CTX* ctx)
     int ret = WC_NO_ERR_TRACE(WOLFSSL_FAILURE);
     int done = 0;
     int added = 0;
-    int i = 0;
-    int numFailedCerts = 0;
     int depth = 0;
     int origDepth = 0;
     WOLFSSL_X509 *issuer = NULL;
@@ -876,54 +842,33 @@ int wolfSSL_X509_verify_cert(WOLFSSL_X509_STORE_CTX* ctx)
         return WOLFSSL_FATAL_ERROR;
     }
 
-    certs = ctx->store->certs;
-
+    /* Chain building mutates the working stack: caller-supplied intermediates
+     * are appended and X509VerifyCertSetupRetry moves failed certs out of it.
+     * store->certs is shared by every connection using this store and
+     * setTrustedSk is owned by the caller, so build a per-verification
+     * shallow copy and leave both untouched. */
+    origTrustedSk = ctx->store->certs;
     if (ctx->setTrustedSk != NULL) {
-        certs = ctx->setTrustedSk;
+        origTrustedSk = ctx->setTrustedSk;
     }
 
-    if (certs == NULL &&
-        wolfSSL_sk_X509_num(ctx->ctxIntermediates) > 0) {
-        certsToUse = wolfSSL_sk_X509_new_null();
-        if (certsToUse == NULL) {
-            ret = WOLFSSL_FAILURE;
-            goto exit;
-        }
-        ret = addAllButSelfSigned(certsToUse, ctx->ctxIntermediates, NULL);
-        /* certsToUse holds only injected intermediates, none are trusted, so
-         * leave origTrustedSk NULL (empty snapshot). */
-        certs = certsToUse;
+    if (origTrustedSk != NULL) {
+        certsToUse = wolfSSL_shallow_sk_dup(origTrustedSk);
     }
     else {
-        /* Snapshot the caller-trusted entries before injecting the
-         * caller-supplied untrusted intermediates.  Only the entries already
-         * present count as trusted for the partial-chain check below, and
-         * we need a stable reference because X509VerifyCertSetupRetry may
-         * remove nodes from `certs` during chain building. */
-        if (certs != NULL && wolfSSL_sk_X509_num(certs) > 0) {
-            int j;
-            int n = wolfSSL_sk_X509_num(certs);
-            origTrustedSk = wolfSSL_sk_X509_new_null();
-            if (origTrustedSk == NULL) {
-                ret = WOLFSSL_FAILURE;
-                goto exit;
-            }
-            for (j = 0; j < n; j++) {
-                if (wolfSSL_sk_X509_push(origTrustedSk,
-                        wolfSSL_sk_X509_value(certs, j)) <= 0) {
-                    ret = WOLFSSL_FAILURE;
-                    goto exit;
-                }
-            }
-        }
-        /* Add the intermediates provided on init to the list of untrusted
-         * intermediates to be used.  They are removed again from `certs` in the
-         * exit cleanup (by identity, recomputed from ctxIntermediates). */
-        ret = addAllButSelfSigned(certs, ctx->ctxIntermediates, NULL);
+        certsToUse = wolfSSL_sk_X509_new_null();
     }
+    if (certsToUse == NULL) {
+        ret = WOLFSSL_FAILURE;
+        goto exit;
+    }
+    /* Add the intermediates provided on init to the list of untrusted
+     * intermediates to be used. */
+    ret = addAllButSelfSigned(certsToUse, ctx->ctxIntermediates, NULL);
     if (ret != WOLFSSL_SUCCESS) {
         goto exit;
     }
+    certs = certsToUse;
 
     if (ctx->chain != NULL) {
         wolfSSL_sk_X509_pop_free(ctx->chain, NULL);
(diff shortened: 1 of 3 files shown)


tests/api/test_ossl_x509_str.c
```diff
@@ -1535,18 +1535,16 @@ static int test_untrusted_inter_depth_exhaustion(X509* leafDeep, X509* inter,
     return EXPECT_RESULT();
 }
 
-/* Intermediate-stack cleanup: the caller-supplied intermediates that the
- * verifier temporarily appends to its working cert list must be removed from
- * the exact stack they were added to once verification finishes.  When a
- * trusted_stack is in use (X509_STORE_CTX_set0_trusted_stack), they are
- * appended to that caller-owned stack; if they are not removed again, a later
- * verification reusing the stack/ctx would snapshot them as trust anchors.
+/* Caller-owned trusted stack (X509_STORE_CTX_set0_trusted_stack): chain
```
Frauschi, line 1771: **🟡 [Medium] Trusted-stack test does not cover the retry/reorder case the removed cleanup guarded against**
*💡 SUGGEST*  `test`

The deleted `X509StoreRemoveCert()` cleanup carried an explicit rationale: identity removal (rather than a positional pop) was required because `X509VerifyCertSetupRetry` **reorders** the working stack. The new test `test_untrusted_inter_store_stack_unchanged()` builds exactly that scenario (tampered candidate ahead of the genuine one, forcing the retry path) and asserts contents *and order* — but only for `store->certs`.

`test_untrusted_inter_trusted_stack_cleanup()`, the `X509_STORE_CTX_set0_trusted_stack` counterpart, was only reworded. It still pushes a single `root` and asserts `sk_X509_num(trusted) == 1`, so it never drives the retry path over the caller's trusted stack. A regression that re-introduced mutation of `setTrustedSk` specifically on the retry/move path would pass this test.

**Suggestion:** Mirror the new store-stack test on the trusted stack: push `tamperedInter` (or another retry-forcing candidate) ahead of `root` in `trusted`, supply the genuine `inter` via the untrusted stack, then assert both the count and the exact order of `trusted` after verification:

    ExpectIntGT(sk_X509_push(trusted, tamperedInter), 0);
    ExpectIntGT(sk_X509_push(trusted, root), 0);
    /* ... verify ... */
    ExpectIntEQ(sk_X509_num(trusted), 2);
    ExpectPtrEq(sk_X509_value(trusted, 0), tamperedInter);
    ExpectPtrEq(sk_X509_value(trusted, 1), root);

Separately, the reworded header comment on this function no longer matches the inline comment further down, which still says "the injected intermediate appended during verification must have been removed, leaving only root" - nothing is appended to the caller's stack any more.

src/x509_str.c
```diff
@@ -876,54 +842,33 @@ int wolfSSL_X509_verify_cert(WOLFSSL_X509_STORE_CTX* ctx)
...
 
-    certs = ctx->store->certs;
-
+    /* Chain building mutates the working stack: caller-supplied intermediates
+     * are appended and X509VerifyCertSetupRetry moves failed certs out of it.
+     * store->certs is shared by every connection using this store and
+     * setTrustedSk is owned by the caller, so build a per-verification
+     * shallow copy and leave both untouched. */
+    origTrustedSk = ctx->store->certs;
     if (ctx->setTrustedSk != NULL) {
```
Frauschi, line 972: **🟡 [Medium] setTrustedSk is no longer pruned by retries, so a failed candidate can now be appended to the reported chain**
*💡 SUGGEST*  `question`

Behavioral delta worth confirming is intended. Previously, when `ctx->setTrustedSk != NULL`, `certs` **was** `ctx->setTrustedSk` (same pointer), so `X509VerifyCertSetupRetry` -> `X509StoreMoveCert` physically removed failed candidates from it. The final "finish building the chain" lookup at line 1026 therefore searched a stack that had already had the failed candidates pruned, and could not return one.

Now `certs` is a private copy, so `ctx->setTrustedSk` keeps every entry. If the store holds two CA certs with the same subject (one broken/expired) and the broken one was moved to `failedCerts` during retry, `X509StoreGetIssuerEx(&issuer, ctx->setTrustedSk, ctx->current_cert)` — which matches on `wolfSSL_X509_check_issued()` (names/AKID, not signature) — can now return that already-failed candidate and `X509StoreChainPush()` puts it on top of `ctx->chain`.

The verdict is unaffected (`ctx->current_cert` was verified against the CM before this point), but `X509_STORE_CTX_get_chain()` / `get1_chain()` would report a terminal certificate that failed verification. Note the `#else` `store->trusted` branch and the `WOLFSSL_SIGNER_DER_CERT` branch already behave this way, so the change makes the `setTrustedSk` path *consistent* — it may well be the desired outcome, it is just not called out anywhere and no test asserts chain contents after a retry.

**Suggestion:** If intended, note it in the commit message / a comment; if not, filter this lookup against `failedCerts` (still in scope at this point) before pushing:

    /* Candidates that already failed verification must not terminate the
     * reported chain: setTrustedSk is no longer pruned by the retry path. */
    if (issuer != NULL && !X509StoreCertInStack(failedCerts, issuer)) {
        X509StoreChainPush(ctx->chain, issuer);
    }

src/x509_str.c
```diff
@@ -876,54 +842,33 @@ int wolfSSL_X509_verify_cert(WOLFSSL_X509_STORE_CTX* ctx)
...
     }
 
-    certs = ctx->store->certs;
-
+    /* Chain building mutates the working stack: caller-supplied intermediates
+     * are appended and X509VerifyCertSetupRetry moves failed certs out of it.
+     * store->certs is shared by every connection using this store and
+     * setTrustedSk is owned by the caller, so build a per-verification
+     * shallow copy and leave both untouched. */
+    origTrustedSk = ctx->store->certs;
```
Frauschi: **🟡 [Medium] origTrustedSk now aliases the live shared store->certs instead of a private snapshot**
*💡 SUGGEST*

The removed code built `origTrustedSk` as a thread-private stack of pointers copied out of `certs` at function entry, and the `WOLFSSL_PARTIAL_CHAIN` fallback walked that private copy. The new code sets `origTrustedSk = ctx->store->certs` (or `ctx->setTrustedSk`) and keeps that raw pointer live for the whole verification, so `X509StoreCertIsTrusted()` traverses the *shared* store stack near the end of chain building, with no lock held and after an arbitrarily long chain walk (CertManager verifies, verify callbacks, CRL checks). Concurrently, `wolfSSL_X509_STORE_add_cert()` mutates the same `store->certs` through `wolfSSL_sk_insert()` with no lock either: it does `stack->num++` *before* walking to the tail and linking the new node, so a reader can observe `num` larger than the number of linked nodes. Worse, `X509StorePushCertsToCM()` (reached from `SSL_CTX_set_cert_store()`) does `wolfSSL_sk_X509_pop_free(store->certs, NULL); store->certs = NULL;` -- if that runs on another thread while a verification is in flight, the `origTrustedSk` pointer captured at entry now refers to freed stack nodes and freed X509 objects, and the partial-chain trust check dereferences them. The PR's own rationale is that `store->certs` is "shared by every connection using this store"; holding a borrowed pointer into it across the entire verification and re-walking it late contradicts that rationale, and it is strictly weaker than the snapshot the PR deleted. Impact is limited to unusual timing (an application that adds certs to, or re-attaches, a store while another thread is verifying against it), hence Medium.

**Recommendation:** Keep the partial-chain trust check on thread-private state: take a second `wolfSSL_shallow_sk_dup()` of the caller-trusted set before injecting the intermediates and use that as `origTrustedSk`, so no pointer into `store->certs` outlives the copy step.

Note: the obvious cheaper variant - recording the pre-injection node count and having `X509StoreCertIsTrusted()` scan only the first N entries of `certsToUse` - does not work. `X509VerifyCertSetupRetry()` -> `X509StoreMoveCert()` pops nodes out of the working copy during chain building, so a positional window drifts off the original trusted entries. That is the same reordering hazard the removed `X509StoreRemoveCert()` documented as its reason for removing by identity rather than by position.

src/x509_str.c
```diff
@@ -876,54 +842,33 @@ int wolfSSL_X509_verify_cert(WOLFSSL_X509_STORE_CTX* ctx)
...
+    /* Add the intermediates provided on init to the list of untrusted
+     * intermediates to be used. */
+    ret = addAllButSelfSigned(certsToUse, ctx->ctxIntermediates, NULL);
     if (ret != WOLFSSL_SUCCESS) {
         goto exit;
     }
+    certs = certsToUse;
 
     if (ctx->chain != NULL) {
         wolfSSL_sk_X509_pop_free(ctx->chain, NULL);
```
Frauschi, line 1000: **🔵 [Low] `certs` is now a redundant alias for `certsToUse`**
*🔧 NIT*  `style`

Before this change `certs` and `certsToUse` were genuinely distinct: `certs` aliased `store->certs`/`setTrustedSk` in the common case and only pointed at the throwaway `certsToUse` in the `certs == NULL` fallback. That distinction drove the exit cleanup (`certsToUse == NULL` meant "we mutated a caller stack, remove the injected certs"). After the rewrite there is exactly one working stack and `certs = certsToUse;` is an unconditional alias, so two locals now track the same object.

**Suggestion:** Drop `certs` and use `certsToUse` at the three call sites (`X509StoreGetIssuerEx`, and the two `X509VerifyCertSetupRetry` calls), or drop `certsToUse` and free `certs` at exit — either removes the alias.

src/x509_str.c
```diff
@@ -876,54 +842,33 @@ int wolfSSL_X509_verify_cert(WOLFSSL_X509_STORE_CTX* ctx)
         return WOLFSSL_FATAL_ERROR;
     }
 
-    certs = ctx->store->certs;
-
+    /* Chain building mutates the working stack: caller-supplied intermediates
+     * are appended and X509VerifyCertSetupRetry moves failed certs out of it.
+     * store->certs is shared by every connection using this store and
+     * setTrustedSk is owned by the caller, so build a per-verification
+     * shallow copy and leave both untouched. */
```
Frauschi: **🔵 [Low] Shared-state race only partly closed: verification still adds/removes temp CAs on the shared store->cm**
*💡 SUGGEST*

The PR moves chain building onto a per-verification shallow copy so that `ctx->store->certs` / `ctx->setTrustedSk` are no longer mutated, which is correct and does remove the stack-corruption race. However, the same function still mutates a second piece of store-shared, unlocked state on every verification: it loads caller-supplied intermediates into `ctx->store->cm` as `WOLFSSL_TEMP_CA` via `X509StoreAddCa()`, and later calls `wolfSSL_CertManagerUnloadTempIntermediateCerts(ctx->store->cm)`, which unloads *all* `WOLFSSL_TEMP_CA` signers in that CertManager, not just the ones this verification added. Two threads verifying concurrently against the same `X509_STORE` therefore still interfere: (a) thread A's unload can drop the intermediate thread B just loaded, making B's chain fail spuriously; and (b) more seriously, in the `WOLFSSL_FAILURE` branch the code unloads temp intermediates *and then* calls `X509StoreVerifyCert(ctx)` expecting only genuinely trusted CAs to be present - if thread B adds its (attacker-supplied, unauthenticated) intermediate as a `TEMP_CA` in that window, thread A can anchor its path on it. The condition is exactly the multi-threaded shared-store usage that motivates this change, so the fix as written does not make concurrent `X509_verify_cert()` on a shared store safe. This mechanism is pre-existing code, not introduced by the diff - it is reported because the PR is specifically framed as making shared-store verification race-free.

**Recommendation:** Either scope the temp-CA add/unload per verification (track the signers this call added and remove only those, e.g. by returning a handle from X509StoreAddCa and unloading by subject key id), or serialize the whole verification against the store/CertManager with the CM lock. At minimum, document in the commit/PR body that concurrent verification on one X509_STORE remains unsupported so the fix is not read as a complete answer to the reported race.

Frauschi (commented):
## 🐺 Skoll Code Review

**Findings:** 11 total — 5 posted, 6 skipped

### Posted findings
- [Medium] **Trusted-stack test does not cover the retry/reorder case the removed cleanup guarded against** — `tests/api/test_ossl_x509_str.c:1538-1577`
- [Medium] **setTrustedSk is no longer pruned by retries, so a failed candidate can now be appended to the reported chain** — `src/x509_str.c:851`
- [Medium] **origTrustedSk now aliases the live shared store->certs instead of a private snapshot** — `src/x509_str.c:850`
- [Low] **`certs` is now a redundant alias for `certsToUse`** — `src/x509_str.c:874`
- [Low] **Shared-state race only partly closed: verification still adds/removes temp CAs on the shared store->cm** — `src/x509_str.c:849`

- [Medium] Copy-on-verify fixes the stack race but X509_verify_cert is still not safe for concurrent use of one store
- [Low] Every verification now allocates one stack node per store certificate
- [Low] Shallow copy of store->certs is taken without synchronization, freezing a possibly inconsistent (num, node list) pair
- [Info] Verification still mutates the store's shared CertManager, so the shared-store race is only partly closed
- [Info] New regression test asserts the single-threaded invariant only; no concurrent verification coverage
- [Info] wolfSSL_shallow_sk_dup() walks the shared store->certs list without holding a lock

</details>

Frauschi:
Jenkins retest this please

src/x509_str.c
```diff
@@ -1082,7 +1068,13 @@ int wolfSSL_X509_verify_cert(WOLFSSL_X509_STORE_CTX* ctx)
                     ctx->setTrustedSk, ctx->current_cert);
             }
     #endif
-            if (issuer != NULL) {
+            /* A candidate that already failed verification (moved to
```
Frauschi: **🟡 [Medium] The new failedCerts guard is narrowly reachable and untested, and when it does fire it drops the anchor off the reported chain**
*💡 SUGGEST*  `test`

This is the guard added for the round-1 review comment, and the fix is right in principle - thanks. Two things to settle before it lands.

**It currently has no coverage.** I verified this by mutation on the PR head (7051bc9): neutralising the guard so it always pushes (`X509StoreCertInStack(...) || 1`) leaves the *entire* `tests/unit.test` suite green, and an instrumented build that prints whenever the guard suppresses a push records zero firings across the `ossl_x509_store` group. So nothing in tree exercises it.

That matches the reachability analysis: `failedCerts` is only ever fed from `certsToUse`, a shallow dup of `ctx->setTrustedSk` or `ctx->store->certs`. In the default path the terminal lookup searches `ctx->store->trusted`, and `wolfSSL_X509_STORE_add_cert()` sorts self-signed certs into `store->trusted` and non-self-signed into `store->certs` - disjoint sets, so pointer identity can never match. The guard can only fire with `setTrustedSk` in use *and* the chain terminating through the `done = 1` CertManager path. `test_untrusted_inter_trusted_stack_unchanged` terminates through the self-issued `ctx->current_cert == issuer` break instead, and asserts only the caller stack, never `X509_STORE_CTX_get0_chain()`.

Also worth a word in the comment: under `WOLFSSL_SIGNER_DER_CERT` the guard is a silent no-op, since `x509GetIssuerFromCM()` returns a freshly allocated X509 that can never be pointer-equal to a `failedCerts` entry.

**When it does fire, `done = 1` still runs and the function reports success with `ctx->chain` missing its top element.** `X509StoreCheckPathLen()` (same file, 764-840) treats `sk_X509_value(ctx->chain, num - 1)` as the trust anchor - it seeds the budget from that cert and skips it in the `num-2 .. 1` loop. With the anchor suppressed, the last intermediate is mistaken for the anchor and is never charged against the path-length budget; if the chain drops below three entries the check returns early altogether. Separately, callers reading the last element of `X509_STORE_CTX_get1_chain()` to identify the anchor now get an intermediate. The verdict is unaffected either way (`ctx->current_cert` was verified against the CM before this point), so this is not a bypass - but the reported chain ends up inconsistent with both its internal consumer and the OpenSSL contract.

**Suggestion:** Add a case that puts the store's real anchor in the CertManager (e.g. an SSL_CTX-owned store, or CertManagerLoadCA) while a same-subject rejected candidate sits in the set0_trusted_stack, then assert X509_STORE_CTX_get0_chain() does not contain the rejected cert. Also note in the comment that under WOLFSSL_SIGNER_DER_CERT the issuer is a CM copy and the identity check does not apply.

src/x509_str.c
```diff
@@ -876,51 +862,43 @@ int wolfSSL_X509_verify_cert(WOLFSSL_X509_STORE_CTX* ctx)
         return WOLFSSL_FATAL_ERROR;
     }
 
-    certs = ctx->store->certs;
-
+    /* Chain building mutates the working stack: caller-supplied intermediates
```
Frauschi, line 954: **🟡 [Medium] Both shallow dups re-walk the shared store stack, so the two snapshots can disagree - take the second one from certsToUse**
*💡 SUGGEST*  `question`

Moving the PARTIAL_CHAIN trust check onto its own snapshot addresses the round-1 comment, but taking that snapshot from `callerTrusted` again re-reads the shared stack a second time:

```c
certsToUse    = wolfSSL_shallow_sk_dup(callerTrusted);
origTrustedSk = wolfSSL_shallow_sk_dup(callerTrusted);
```

The two walks are independent and unsynchronised, so they are not guaranteed to observe the same set. If thread B calls `X509_STORE_add_cert(store, C)` between them, thread A verifies with a *trust* snapshot containing C while its *working* stack does not - and the PARTIAL_CHAIN fallback (1028-1032) can then accept a chain terminating at C even though C was never a chain-building candidate for this verification. The reverse interleaving rejects a terminus the verification did consider. That makes the trust decision depend on timing rather than on store contents at any single instant, and it is the one part of this that is free to fix: `origTrustedSk` needs the caller's set *before* `addAllButSelfSigned()` injects intermediates, which is exactly what `certsToUse` already holds at that point. Duplicating from the private copy removes the second shared-list walk and makes the two sets provably identical.

The residual points are worth a note rather than a change, since the PR already documents (972-979) that concurrent verification on one store stays unsupported:

- The dup is itself an unlocked read. `WOLFSSL_X509_STORE` has no mutex over `certs`/`trusted`, and `wolfSSL_sk_insert()` bumps `stack->num` *before* linking the new node, so a concurrent `add_cert` can be observed mid-update (harmless today - `wolfSSL_X509_check_issued()` is NULL-safe on the trailing indices). `X509StorePushCertsToCM()` pop-frees the whole stack and NULLs it, which the walk cannot survive at all. So the write-side corruption is genuinely gone, but the comment at 865-878 and the PR/commit wording read as if the race is closed; they should say what the NOTE at 972-979 says.
- `origTrustedSk`'s only consumer is `X509StoreCertIsTrusted()` inside the PARTIAL_CHAIN branch, and both flag sources are readable at entry - so the snapshot could be taken only when the flag is set. (Not a new allocation class: the deleted code also built one node per store cert.)
- Minor: `origTrustedSk` is NULL-checked inside the `if (callerTrusted != NULL)` block while `certsToUse` is checked by the shared test below. The result is correct - both reach `goto exit` and the exit path shallow-frees whichever succeeded - but it reads as if only one allocation is checked, and the second dup is still attempted after the first has failed.

**Suggestion:** Reword the 865-878 comment (and the PR description) to say the copy removes the write-side corruption but does not make concurrent use of one X509_STORE safe, since the copy is itself an unlocked read - consistent with the NOTE already added at lines 972-979.

src/x509_str.c
```diff
@@ -679,6 +648,25 @@ static int X509StoreChainPush(WOLF_STACK_OF(WOLFSSL_X509)* chain,
     return ret;
 }
 
+/* Returns 1 if `cert` (by pointer identity) is present in `stack`, else 0.
```
Frauschi, line 725: **🔵 [Low] X509StoreCertInStack walks nodes without bounding by wolfSSL_sk_X509_num()**
*🔧 NIT*  `convention`

The new helper iterates the raw node list (`for (node = stack; node != NULL; node = node->next)`) while every sibling helper in this file bounds the walk by the stack count - `X509StoreMoveCert()` uses `wolfSSL_sk_X509_num()`, and the `X509StoreRemoveCert()` deleted by this PR used `idx < num && node != NULL` for exactly that reason. It is safe today (a wolfSSL empty stack is a single head node with `data.x509 == NULL`, and `cert` is NULL-checked, so no false positive), but it relies on the node-list length always matching `num`, an invariant the rest of the file does not assume.

**Also raised by the `bugs` scan** (src/x509_str.c:651-668):

The new helper iterates raw nodes (`for (node = stack; node != NULL; node = node->next)`) and never consults `wolfSSL_sk_X509_num()`. The helper it replaces in the same diff, `X509StoreRemoveCert()`, deliberately bounded its walk with `num = wolfSSL_sk_X509_num(stack)` and carried a comment explaining that a wolfSSL stack's logical length is `head->num`, not the physical node count.

Today this is still correct for the only argument passed (`failedCerts`): that stack is only ever appended to via `wolfSSL_sk_push()`, an empty head node carries `data.x509 == NULL`, and `cert` is checked non-NULL, so no false positive is possible. But the two representations are not interchangeable in general - `wolfSSL_sk_insert()` increments `stack->num` *before* linking the new node (src/ssl_sk.c:767-778), so `num` and the node chain are transiently inconsistent - and any future caller that passes a stack which has had entries removed positionally, or a stack shared with another writer, gets a silently different answer than every other iteration site in this file.

Since the guard at line 1077 decides whether a certificate is reported in `ctx->chain`, a wrong answer here is a wrong verified chain, not just a cosmetic difference.

**Suggestion:**
```suggestion
int idx;
    int num = wolfSSL_sk_X509_num(stack);

    for (node = stack, idx = 0; idx < num && node != NULL;
            node = node->next, idx++) {
        if (node->data.x509 == cert)
            return 1;
    }
```

Frauschi (commented):
## 🐺 Skoll Code Review

**Findings:** 6 total — 5 posted, 1 skipped

### Posted findings
- [Medium] **The new failedCerts guard is narrowly reachable and untested, and when it does fire it drops the anchor off the reported chain** — `src/x509_str.c:1071-1079`
- [Medium] **Both shallow dups re-walk the shared store stack, so the two snapshots can disagree - take the second one from certsToUse** — `src/x509_str.c:865-898`
- [Low] **X509StoreCertInStack walks nodes without bounding by wolfSSL_sk_X509_num()** — `src/x509_str.c:651-668`
- [Low] **Both new negative sub-cases assert only the return value, not ctx->error** — `tests/api/test_ossl_x509_str.c:1618-1627`
- [Info] **Snapshot mitigation is asymmetric: store->trusted and setTrustedSk are still walked live during verification** — `src/x509_str.c:729-735,1062-1069`

- [Info] X509_STORE_CTX_get1_issuer() from a verify callback no longer sees caller-supplied intermediates

</details>

tests/api/test_ossl_x509_str.c
```diff
@@ -1578,6 +1577,61 @@ static int test_untrusted_inter_trusted_stack_cleanup(X509* leaf, X509* inter,
...
+    ExpectIntEQ(X509_verify_cert(ctx), 1);
+    X509_STORE_CTX_free(ctx);
+    ctx = NULL;
+
+    /* Trusted stack unchanged in contents and order. */
+    ExpectIntEQ(sk_X509_num(trusted), 2);
+    ExpectPtrEq(sk_X509_value(trusted, 0), tamperedInter);
+    ExpectPtrEq(sk_X509_value(trusted, 1), root);
+
+    /* Failing verification on the same trusted stack: no genuine issuer. */
```
Frauschi, line 1851: **🔵 [Low] Both new negative sub-cases assert only the return value, not ctx->error**
*🔧 NIT*  `test`

The two failing verifications added by this PR assert `ExpectIntEQ(X509_verify_cert(ctx), 0)` and then re-check the caller stack, but never pin the error. Every sibling negative test in this file also does `ExpectIntNE(X509_STORE_CTX_get_error(ctx), X509_V_OK)` - see `test_untrusted_inter_tampered`, `test_untrusted_inter_reused_store`, `test_untrusted_inter_no_stale_anchor`, `test_untrusted_inter_retry`.

It matters more than usual here because of the path these two drive: `X509VerifyCertSetupRetry()` pops an already-empty `ctx->chain`, leaves `ctx->current_cert == NULL`, and the next `X509StoreGetIssuerEx()` bails to `exit` with `WOLFSSL_FATAL_ERROR`. That is exactly the path most likely to return failure with a stale or unset `ctx->error`. `test_untrusted_inter_store_stack_unchanged` (1753-1757) has the same gap.

**Suggestion:**
```suggestion
ExpectIntEQ(X509_verify_cert(ctx), 0);
    ExpectIntNE(X509_STORE_CTX_get_error(ctx), X509_V_OK);
    X509_STORE_CTX_free(ctx);
```

src/x509_str.c
```diff
@@ -716,11 +704,11 @@ static int X509DerEquals(WOLFSSL_X509* cur, WOLFSSL_X509* x509)
...
- * X509_V_FLAG_PARTIAL_CHAIN fallback to confirm that a chain actually
- * terminates at a caller-trusted certificate. */
+ * immutable snapshot of the caller's trusted set - store->certs or the
+ * set0_trusted_stack override - captured before any intermediates were
+ * injected for this verification call) or in store->trusted.  Returns 0
+ * otherwise.  Used by the X509_V_FLAG_PARTIAL_CHAIN fallback to confirm that
+ * a chain actually terminates at a caller-trusted certificate. */
 static int X509StoreCertIsTrusted(WOLFSSL_X509_STORE* store,
         WOLFSSL_X509* x509, WOLF_STACK_OF(WOLFSSL_X509)* origTrustedSk)
 {
```
Frauschi, line 796: **⚪ [Info] Snapshot mitigation is asymmetric: store->trusted and setTrustedSk are still walked live during verification**
*🔧 NIT*

**Sites:** `src/x509_str.c:729-735,1062-1069`
The PR's rationale for snapshotting is that "another thread may add to or detach store->certs while this verification is in flight, so a borrowed pointer could be reordered or freed underneath us" (src/x509_str.c:871-878). That rationale applies verbatim to `ctx->store->trusted`, which `wolfSSL_X509_STORE_add_cert()` also appends to under no lock (src/x509_str.c:2059) - yet `store->trusted` is still walked live in two places the PR touched or relies on:

  * `X509StoreCertIsTrusted()` (src/x509_str.c:729-735) iterates `store->trusted` directly on every PARTIAL_CHAIN fallback. Only the `origTrustedSk` half of that function got a snapshot; the `store->trusted` half did not.
  * the terminal issuer lookup (src/x509_str.c:1062-1069) calls `X509StoreGetIssuerEx()` on `ctx->store->trusted` / `ctx->setTrustedSk` live, immediately before the new `X509StoreCertInStack(failedCerts, issuer)` guard.

So after this PR one shared stack (`store->certs`) is snapshotted while its sibling (`store->trusted`) is not, for no stated reason. The PR is honest that full concurrency is not achieved - the new NOTE at src/x509_str.c:972-979 records that caller-supplied intermediates are still loaded into the shared `ctx->store->cm` as `WOLFSSL_TEMP_CA` and that the unload drops *all* temp CAs in that CertManager, so two threads verifying against one `X509_STORE` still corrupt each other's temporary trust set. This is a design observation, not a defect introduced by the diff: the change is a net improvement (no verification path mutates `store->certs` or the caller's `setTrustedSk` any more, which removes the previous concurrent-verify reorder/use-after-free on those stacks), but the shared-store verification path as a whole is still not thread safe.

**Recommendation:** For consistency with the `store->certs` treatment, either snapshot `store->trusted` the same way at the top of `wolfSSL_X509_verify_cert()` and pass the snapshot into `X509StoreCertIsTrusted()` and the terminal issuer lookup, or - preferably - introduce a single `WOLFSSL_X509_STORE`-level mutex guarding `certs`, `trusted`, and `owned`, taken by `wolfSSL_X509_STORE_add_cert()` and by the snapshot/lookup paths here. Until then, promote the `store->cm` NOTE at src/x509_str.c:972-979 into the public API documentation for `X509_STORE`/`X509_verify_cert` so callers know a store must not be shared across concurrently verifying threads.

julek-wolfssl:
fenrir addressed

philljj:
Retest this please.

FAIL: TLS1.2 default negotiated: connection failed

philljj (approved):

PR #10999, opened 2026-07-28
Parse x500UniqueIdentifier (OID 2.5.4.45) in certificate DN

embhorn:
# Description

Adds support for the x500UniqueIdentifier attribute (X.520 OID 2.5.4.45) in certificate subject/issuer distinguished names. Unlike the usual DN attributes, its value is encoded as a BIT STRING rather than a DirectoryString, which the RDN parser previously rejected with ASN_PARSE_E.

Changes:
- Accept ASN_BIT_STRING in the RDN tag choice list (rdnChoice[]).
- Handle ASN_X500_UNIQUE_ID (0x2d) in GetRDN/GetCertName: set the type string and WC_NID_x500UniqueIdentifier, and strip the leading unused-bits count octet so the stored value is the raw bytes.
- Map the BIT STRING encoding to WOLFSSL_V_ASN1_BIT_STRING for the WOLFSSL_X509_NAME entry (OpenSSL-extra path), so the full value is reachable by NID even when it contains an embedded NUL that would truncate the display string.
- Add the x500UniqueIdentifier NID (503), short/long names, and object-info table entry.

Fixes zd22154

# Testing

New regression test test_wc_ParseCert_uniqueIdentifier in tests/api.c using self-signed ECDSA P-256 certs whose DNs carry an x500UniqueIdentifier. The DER was generated with asn1crypto/pyca and verified with OpenSSL, and covers both a normal value and one with an embedded 0x00.

# Checklist

 - [x] added tests
 - [ ] updated/added doxygen


julek-wolfssl (changes requested):
The comments are too verbose. Please shorten.

tests/api.c
```diff
@@ -671,6 +671,211 @@ defined(USE_CERT_BUFFERS_2048)
...
+/* Test parsing an x500UniqueIdentifier (OID 2.5.4.45) in a subject/issuer DN.
+ * Its value is a BIT STRING, which the RDN parser used to reject. */
+static int test_wc_ParseCert_uniqueIdentifier(void)
+{
+    EXPECT_DECLS;
+#if !defined(NO_CERTS) && defined(WOLFSSL_ASN_TEMPLATE) && defined(HAVE_ECC)
+    /* Self-signed ECDSA P-256 certificate with an x500UniqueIdentifier BIT
+     * STRING in both subject and issuer DN. Built with asn1crypto and
+     * pyca/cryptography, verified with OpenSSL. */
+    static const byte leafUniqueIdDer[] = {
```
dgarske: What happens when this certificate expires? How was it created or how can it be updated?

tests/api.c
```diff
@@ -671,6 +671,211 @@ defined(USE_CERT_BUFFERS_2048)
...
+    for (i = 0; (i + 6) <= (word32)sizeof(badDer); i++) {
+        /* OID 2.5.4.45 (55 04 2d), BIT STRING (03), len 9 (09), 0 unused (00) */
+        if ((badDer[i] == 0x55) && (badDer[i + 1] == 0x04) &&
+                (badDer[i + 2] == 0x2D) && (badDer[i + 3] == 0x03) &&
+                (badDer[i + 4] == 0x09) && (badDer[i + 5] == 0x00)) {
+            badDer[i + 5] = 0x03; /* 3 unused bits - not byte aligned */
+            flipped++;
+        }
+    }
+    ExpectIntGT(flipped, 0); /* sanity: mutation was applied */
```
dgarske: Use `ExpectIntEQ(flipped, 2)`

dgarske (changes requested):

tests/api.c
embhorn: Agreed. I added a script and moved the cert into a header file.

wolfssl/wolfcrypt/asn.h
```diff
@@ -1101,6 +1103,10 @@ extern const WOLFSSL_ObjectInfo wolfssl_object_info[];
 #define WC_LN_userId "userId"
 #define WC_NID_userId 458
 
+#define WC_SN_x500UniqueIdentifier "x500UniqueIdentifier"
+#define WC_LN_x500UniqueIdentifier "X500 unique identifier"
+#define WC_NID_x500UniqueIdentifier 503     /* 2.5.4.45 */
```
dgarske, line 1108: Add `case WC_NID_x500UniqueIdentifier: return ASN_X500_UNIQUE_ID;` to `ConvertNIDToWolfSSL()`. 
Add the `WC_NID_x500UniqueIdentifier` case to `get_dn_attr_by_nid()`.

wolfcrypt/src/asn.c
```diff
@@ -15427,6 +15460,25 @@ static int GetCertName(DecodedCert* cert, char* full, byte* hash, int nameType,
...
+                        WOLFSSL_MSG("BIT STRING RDN value only valid for "
+                                    "x500UniqueIdentifier");
+                        ret = ASN_PARSE_E;
+                    }
+                    else if ((strLen == 0) || (str[0] != 0)) {
+                        WOLFSSL_MSG("BIT STRING RDN value not byte-aligned");
+                        ret = ASN_PARSE_E;
+                    }
+                    else {
+                        str++;
```
dgarske: The `WOLFSSL_NO_ASN_STRICT` guard rejecting `strLen < 1` runs at `wolfcrypt/src/asn.c:15454` **before** the new BIT STRING unused-bits octet is stripped at `asn.c:15476-15479`. For a BIT STRING encoded `03 01 00` (an empty bit string), the raw `strLen` is 1, so the check passes; after stripping, `strLen` becomes 0 and a zero-length X509_NAME entry is created. Every other RDN value type is rejected at length 0. The same asymmetry exists in `GetRDN` — `asn.c:15299` only rejects `strLen == 0` on the *raw* length, so an empty BIT STRING is accepted there too and `/x500UniqueIdentifier=` is appended to the display string with no value. I confirmed this by generating a cert with `make_cert(b'')` from the PR's own generator script:

```
empty-bitstring parse=0 subject=[/C=US/O=Test Org/CN=LEAF Test Card/x500UniqueIdentifier=]
idx=3 len=0 type=3
```

No crash or OOB (`wolfSSL_ASN1_STRING_set` handles `sz == 0` safely at `src/ssl_asn1.c:3261`), but the strictness intent is inconsistent, and `EscapeString_RFC2253` rejects `inSz == 0`.

  Code:
```
                #ifndef WOLFSSL_NO_ASN_STRICT
                    /* RFC 5280 section 4.1.2.4 lists a DirectoryString as being
                     * 1..MAX in length */
                    if (ret == 0 && strLen < 1) {   /* <-- runs on the RAW length */
                        ...
                        ret = ASN_PARSE_E;
                    }
                #endif
    
                    if (tag == ASN_BIT_STRING) {
                        ...
                        else {
                            str++;
                            strLen--;   /* <-- strip happens AFTER the check */
                        }
                    }
```

Either move the strip above the strict length check so an empty x500UniqueIdentifier is rejected consistently, or explicitly decide that an empty BIT STRING is legal (RFC 4517 allows `''B`) and add a comment plus a regression test for the `03 01 00` case.

wolfcrypt/src/asn.c
```diff
@@ -15427,6 +15460,25 @@ static int GetCertName(DecodedCert* cert, char* full, byte* hash, int nameType,
...
+                 * supported as the count isn't stored. */
+                if (tag == ASN_BIT_STRING) {
+                    if (nid != WC_NID_x500UniqueIdentifier) {
+                        WOLFSSL_MSG("BIT STRING RDN value only valid for "
+                                    "x500UniqueIdentifier");
+                        ret = ASN_PARSE_E;
+                    }
+                    else if ((strLen == 0) || (str[0] != 0)) {
+                        WOLFSSL_MSG("BIT STRING RDN value not byte-aligned");
+                        ret = ASN_PARSE_E;
```
dgarske: Consider adding this in the new error code paths: `WOLFSSL_ERROR_VERBOSE(ASN_PARSE_E);` which matches existing convention in this file.

src/ssl_asn1.c
```diff
@@ -5068,6 +5068,8 @@ const WOLFSSL_ObjectInfo wolfssl_object_info[] = {
 
     /* Rarely used and legacy attributes. */
     { WC_NID_userId, WC_NAME_USER_ID_OID, oidCertNameType, "UID", "userId"},
+    { WC_NID_x500UniqueIdentifier, WC_NAME_X500_UNIQUE_ID_OID, oidCertNameType,
+      "x500UniqueIdentifier", "X500 unique identifier"},
```
dgarske: Use `WC_SN_x500UniqueIdentifier` / `WC_LN_x500UniqueIdentifier` in the table so the strings have a single definition.

wolfssl/wolfcrypt/oid_sum.h
```diff
@@ -1942,6 +1942,9 @@ enum CertName_Sum {
     /* 0x55,0x04,0x2b  */
     /* 2.5.4.43 */
     WC_NAME_INITIALIS_OID               = 132,
+    /* 0x55,0x04,0x2d  */
+    /* 2.5.4.45 */
+    WC_NAME_X500_UNIQUE_ID_OID          = 134,
```
dgarske, line 1947: Note this collides with `CRL_NUMBER_OID              = 134`, but I think that is acceptable since this is the old method.

tests/api/create_x500_unique_id_certs.py
```diff
@@ -0,0 +1,215 @@
+#!/usr/bin/env python3
```
dgarske, line 1: Can / should these x500 cert creation/header be in the certs/ dir?

dgarske (changes requested):

tests/api/create_x500_unique_id_certs.py
embhorn: These test certs use attributes that cannot be created with the shell scripts

dgarske (approved):

dgarske (approved):

julek-wolfssl (approved):

PR #11002, opened 2026-07-29
autoconf: build wolfevent.c under --enable-usersettings

danielinux:
# Description

wolfevent.c (the wolfEvent completion queue) only compiled under --enable-asynccrypt: AM_CONDITIONAL([BUILD_WOLFEVENT],...) tested only ENABLED_ASYNCCRYPT. But HAVE_WOLF_EVENT is a user-settable macro that internal.c/ssl.c gate their wolfEventQueue_* references on, so a --enable-usersettings build that defines HAVE_WOLF_EVENT referenced the queue API while the object was never built -> link error (undefined wolfEventQueue_Init/Free). BUILD_WOLFEVENT was also the only BUILD_* conditional missing the standard "|| ENABLED_USERSETTINGS = yes" escape hatch that lets a header-driven build compile a source and defer the decision to the file's own #ifdef.

Add that clause (matching BUILD_FALCON/BUILD_MLKEM/etc.) so usersettings builds compile wolfevent.c; its internal #ifdef HAVE_WOLF_EVENT still gates whether the code is active, so non-event builds get an empty object.

Also silence wolfEvent_Poll's unused-parameter -Werror in non-async builds (event/flags are consumed only by the async hardware poll), which the above change surfaces now that the file compiles without WOLFSSL_ASYNC_CRYPT.

# Testing

MC/DC  step 4 (in progress, #10967 ) wants to cover wolfevent.c using `--enable-usersettings`


danielinux:
jenkins retest this please

dgarske (approved):

PR #11003, opened 2026-07-29
SSL_get_cipher_list: enumerate the configured cipher list

julek-wolfssl:
`SSL_get_cipher_list()` was mapped to `wolfSSL_get_cipher_list_ex()`, which
only returned the negotiated cipher at index 0 and NULL for every other
index, so callers walking the list saw a single entry instead of the full
configured list as OpenSSL does.

`wolfSSL_get_cipher_list_ex()` had a fallback meant to enumerate the list,
but it was unreachable (it only triggers when the internal cipher name
lookup returns NULL, which hasn't happened since the lookup started
returning the literal "None" instead). Even if reached, it indexed the
global `cipher_names[]` table rather than the suites actually configured on
the SSL, so it wouldn't have matched OpenSSL's behavior anyway.

Add `wolfSSL_get_cipher_list_compat()` and point the `SSL_get_cipher_list`
macro at it instead. It walks `WOLFSSL_SUITES(ssl)` with the same SCSV and
min/max version filtering used by `wolfSSL_get_ciphers_compat()`, so
`SSL_get_cipher_list()` and `SSL_get_ciphers()` now report the same suites
in the same order, named the way `SSL_CIPHER_get_name()` does.
`wolfSSL_get_cipher_list_ex()` itself is left unchanged for direct callers.

### DIFF
--- a/src/ssl.c
+++ b/src/ssl.c
@@ -9305,12 +9305,9 @@ WOLF_STACK_OF(WOLFSSL_CIPHER) *wolfSSL_get_ciphers_compat(const WOLFSSL *ssl)
         if (ssl->suitesStack == NULL)
             return NULL;
 
-        /* higher priority of cipher suite will be on top of stack */
-#if defined(OPENSSL_ALL)
-        for (i = suites->suiteSz - 2; i >=0; i-=2)
-#else
-        for (i = 0; i < suites->suiteSz; i+=2)
-#endif
+        /* Walk lowest priority first: each suite is inserted at index 0, so
+         * the highest priority suite ends up on top of the stack. */
+        for (i = suites->suiteSz - 2; i >= 0; i -= 2)
         {
             struct WOLFSSL_CIPHER cipher;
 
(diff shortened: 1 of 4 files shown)


julek-wolfssl:
Failures are stale results left when converting from draft.

philljj:
Retest this please

(no logs).

philljj (changes requested):
This looks good, but I think it needs a rebase + force-push to get PRB CI to actually run.

philljj (changes requested):
Please review if these flagged low / info issues are worth addressing.

julek-wolfssl:
retest this please

julek-wolfssl:
retest this please

julek-wolfssl:
retest this please

julek-wolfssl:
retest this please

julek-wolfssl:
retest this please

julek-wolfssl:
retest this please

philljj:
Retest this please.

(no logs)

julek-wolfssl:
retest this please

philljj (approved):

PR #11006, opened 2026-07-29
20260729 Coverity fixes

rlm2002:
# Description

CID 561836: Untrusted value as argument - Bound untrusted DER length before allocation in `wolfssl_read_der_bio`
CID 561978: Uninitialized scalar variable - Initialize `enc` in test_coding.c to avoid use before set.
CID 562433/562424/562425: Resource leak - Free port, path, host for potential failure path in `test_ssl_crl_ocsp.c`
CID 562432: Resource leak - Free the i2d output buffer in `test_wc_PKCS12DecisionCoverage`
CID 562428: Resource leak - set `names = wolfSSL_load_client_CA_file("does/not/exist.pem")` to `NULL` before continuing

# Testing

`./configure --enable-jni && make check`

### DIFF
--- a/src/pk_rsa.c
+++ b/src/pk_rsa.c
@@ -651,6 +651,12 @@ static int wolfssl_read_der_bio(WOLFSSL_BIO* bio, unsigned char** out)
         WOLFSSL_ERROR_MSG("DER SEQUENCE decode failed");
         err = 1;
     }
+    /* Cap at 8x the maximum modulus size, leaves headroom for the full
+     * private key encoding. */
+    if ((!err) && (derLen > RSA_MAX_SIZE)) {
+        WOLFSSL_ERROR_MSG("DER length too large");
+        err = 1;
+    }
     /* Allocate a buffer to read DER data into. */
     if ((!err) && ((der = (unsigned char*)XMALLOC((size_t)derLen, bio->heap,
             DYNAMIC_TYPE_TMP_BUFFER)) == NULL)) {
--- a/src/ssl_sess.c
+++ b/src/ssl_sess.c
@@ -2165,13 +2165,37 @@ void AddSession(WOLFSSL* ssl)
         if (ssl->rng != NULL)
             rng = ssl->rng;
 #if defined(HAVE_GLOBAL_RNG) && defined(OPENSSL_EXTRA)
-        else if (initGlobalRNG == 1 || wolfSSL_RAND_Init() == WOLFSSL_SUCCESS) {
+        else if (initGlobalRNG == 1 ||
+                 wolfSSL_RAND_Init() == WOLFSSL_SUCCESS) {
             rng = &globalRNG;
+        }
+        if (rng == &globalRNG) {
+            if (wc_LockMutex(&globalRNGMutex) != 0) {
+                WOLFSSL_MSG("Bad Lock Mutex rng");
+                return;
+            }
+            /* The above access requires initGlobalRNG recheck now
+             * that we have the lock. */
+            if (initGlobalRNG == 0) {
+                wc_UnLockMutex(&globalRNGMutex);
+                return;
+            }
         }
 #endif
         if (wc_RNG_GenerateBlock(rng, ssl->session->altSessionID,
-                ID_LEN) != 0)
+                ID_LEN) != 0) {
+#if defined(HAVE_GLOBAL_RNG) && defined(OPENSSL_EXTRA)
+            if (rng == &globalRNG) {
+                wc_UnLockMutex(&globalRNGMutex);
+            }
+#endif
             return;
+        }
+#if defined(HAVE_GLOBAL_RNG) && defined(OPENSSL_EXTRA)
+        if (rng == &globalRNG) {
+            wc_UnLockMutex(&globalRNGMutex);
+        }
+#endif
         ssl->session->haveAltSessionID = 1;
         id = ssl->session->altSessionID;
         idSz = ID_LEN;
--- a/tests/api.c
+++ b/tests/api.c
@@ -20491,6 +20491,83 @@ static int test_wolfSSL_d2i_PrivateKeys_bio(void)
 }
 #endif /* OPENSSL_ALL || (WOLFSSL_ASIO && !NO_RSA) */
 
+/* Build gate for the hook + test so they can't drift: OPENSSL_EXTRA (compat
+ * BIO/RSA) + a d2i_RSAPrivateKey_bio profile + RSA/key-gen + swappable alloc.
+ * DEBUG_MEMORY excluded on purpose: the hook is single-arg only. */
+#if defined(OPENSSL_EXTRA) &&                                                 \
+    (defined(OPENSSL_ALL) || defined(WOLFSSL_ASIO) ||                         \
+     defined(WOLFSSL_HAPROXY) || defined(WOLFSSL_NGINX)) &&                   \
+    !defined(NO_RSA) && defined(WOLFSSL_KEY_GEN) &&                           \
+    defined(USE_WOLFSSL_MEMORY) && !defined(WOLFSSL_NO_MALLOC) &&             \
+    !defined(WOLFSSL_STATIC_MEMORY) && !defined(WOLFSSL_DEBUG_MEMORY)
+    #define TEST_DER_CAP_MALLOC_HOOK
+#endif
+
+#ifdef TEST_DER_CAP_MALLOC_HOOK
+/* Refuses allocations >= der_cap_threshold; forwards the rest to the installed
+ * allocator (native if none) so the alloc triple stays consistent. */
+static wolfSSL_Malloc_cb der_cap_prev_malloc = NULL;
+static size_t der_cap_threshold = 0; /* 0 = disabled */
+static int    der_cap_attempts  = 0;
+
+static void* der_cap_malloc_cb(size_t size)
+{
+    if (der_cap_threshold != 0 && size >= der_cap_threshold) {
+        der_cap_attempts++;
+        return NULL; /* refuse; records the attempt */
+    }
+    if (der_cap_prev_malloc != NULL)
+        return der_cap_prev_malloc(size);
+    return malloc(size);
+}
+#endif /* TEST_DER_CAP_MALLOC_HOOK */
+
+/* Oversized-DER cap must reject before allocating. NULL alone doesn't prove it
+ * (uncapped returns NULL too), so assert no large alloc was attempted. */
+static int test_wolfSSL_d2i_RSAPrivateKey_bio_oversized(void)
+{
+    EXPECT_DECLS;
+#ifdef TEST_DER_CAP_MALLOC_HOOK
+    /* SEQUENCE, canonical 4-byte length 0x01000000 (16 MB), over the cap.
+     * Must be canonical -- 0x00FFFFFF is rejected by the parser first. */
+    static const unsigned char hugeSeq[] =
+        { 0x30, 0x84, 0x01, 0x00, 0x00, 0x00 };
+    BIO* bio = NULL;
+    RSA* rsa = NULL;
+    wolfSSL_Malloc_cb  prev_mc = NULL;
+    wolfSSL_Free_cb    prev_fc = NULL;
+    wolfSSL_Realloc_cb prev_rc = NULL;
+    int allocators_set = 0;
+
+    ExpectIntEQ(wolfSSL_GetAllocators(&prev_mc, &prev_fc, &prev_rc), 0);
+    der_cap_prev_malloc = prev_mc;
+    /* Override only malloc; keep the installed free/realloc. */
+    ExpectIntEQ(wolfSSL_SetAllocators(der_cap_malloc_cb, prev_fc, prev_rc), 0);
+    if (EXPECT_SUCCESS())
+        allocators_set = 1;
+
+    ExpectNotNull(bio = BIO_new(BIO_s_mem()));
+    ExpectIntGT(BIO_write(bio, hugeSeq, (int)sizeof(hugeSeq)), 0);
+
+    der_cap_attempts = 0;
+    der_cap_threshold = 0x10000;
+    ExpectNull(d2i_RSAPrivateKey_bio(bio, &rsa));
+    der_cap_threshold = 0;
+
+    ExpectIntEQ(der_cap_attempts, 0);
+
+    BIO_free(bio);
+    RSA_free(rsa);
+
+    if (allocators_set)
+        (void)wolfSSL_SetAllocators(prev_mc, prev_fc, prev_rc);
+    der_cap_prev_malloc = NULL;
+#endif
+    return EXPECT_RESULT();
+}
+
+#undef TEST_DER_CAP_MALLOC_HOOK
+
 #endif /* !NO_BIO */
 
 
@@ -38233,6 +38310,9 @@ TEST_CASE testCases[] = {
     TEST_DECL(test_wolfSSL_d2i_PrivateKeys_bio),
 #endif /* !NO_BIO */
 #endif
+#ifndef NO_BIO
+    TEST_DECL(test_wolfSSL_d2i_RSAPrivateKey_bio_oversized),
+#endif /* !NO_BIO */
 
 
 #if !defined(NO_WOLFSSL_CLIENT) && !defined(NO_WOLFSSL_SERVER)
--- a/tests/api.c
+++ b/tests/api.c
@@ -21696,6 +21696,83 @@ static int test_wolfSSL_d2i_PrivateKeys_bio(void)
 }
 #endif /* OPENSSL_ALL || (WOLFSSL_ASIO && !NO_RSA) */
 
+/* Build gate for the hook + test so they can't drift: OPENSSL_EXTRA (compat
+ * BIO/RSA) + a d2i_RSAPrivateKey_bio profile + RSA/key-gen + swappable alloc.
+ * DEBUG_MEMORY excluded on purpose: the hook is single-arg only. */
+#if defined(OPENSSL_EXTRA) &&                                                 \
+    (defined(OPENSSL_ALL) || defined(WOLFSSL_ASIO) ||                         \
+     defined(WOLFSSL_HAPROXY) || defined(WOLFSSL_NGINX)) &&                   \
+    !defined(NO_RSA) && defined(WOLFSSL_KEY_GEN) &&                           \
+    defined(USE_WOLFSSL_MEMORY) && !defined(WOLFSSL_NO_MALLOC) &&             \
+    !defined(WOLFSSL_STATIC_MEMORY) && !defined(WOLFSSL_DEBUG_MEMORY)
+    #define TEST_DER_CAP_MALLOC_HOOK
+#endif
+
+#ifdef TEST_DER_CAP_MALLOC_HOOK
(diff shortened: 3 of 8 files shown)


src/ssl_sess.c
```diff
@@ -2165,13 +2165,24 @@ void AddSession(WOLFSSL* ssl)
         if (ssl->rng != NULL)
             rng = ssl->rng;
 #if defined(HAVE_GLOBAL_RNG) && defined(OPENSSL_EXTRA)
-        else if (initGlobalRNG == 1 || wolfSSL_RAND_Init() == WOLFSSL_SUCCESS) {
+        else if (initGlobalRNG == 1 ||
+                 wolfSSL_RAND_Init() == WOLFSSL_SUCCESS) {
             rng = &globalRNG;
         }
+        if (rng == &globalRNG && wc_LockMutex(&globalRNGMutex) != 0) {
```
Frauschi: **🟠 [High] globalRNGMutex leaked when wc_RNG_GenerateBlock() fails in AddSession()**
*🚫 BLOCK*  `bug`

The new locking added by this PR takes `globalRNGMutex` before `wc_RNG_GenerateBlock()` but only releases it on the success path. If `wc_RNG_GenerateBlock()` returns non-zero (e.g. DRBG reseed failure, entropy source failure, `WC_NO_ERR_TRACE(RNG_FAILURE_E)`), the function does a bare `return;` while still holding the mutex. `globalRNGMutex` is a plain (non-recursive, non-owner-tracked) `wolfSSL_Mutex`, so every subsequent user of the global RNG in the process — `wolfSSL_RAND_bytes()`, `wolfSSL_RAND_seed()`, `wolfSSL_RAND_Cleanup()`, `wolfSSL_d2i_PKCS12_bio` paths in `src/ssl_p7p12.c:1563`, and `AddSession()` itself — blocks forever on `wc_LockMutex()`. `wolfSSL_Cleanup()` (`src/ssl.c:3726`) will also fail to `wc_FreeMutex()` a still-held mutex. Before this PR the same error path simply returned with no lock held, so this deadlock is introduced by the change. Note the pre-existing early `return` on the lock-failure branch is fine (nothing held), but the RNG-failure return is not.

**Suggestion:**
```suggestion
if (wc_RNG_GenerateBlock(rng, ssl->session->altSessionID,
                ID_LEN) != 0) {
        #if defined(HAVE_GLOBAL_RNG) && defined(OPENSSL_EXTRA)
            if (rng == &globalRNG) {
                wc_UnLockMutex(&globalRNGMutex);
            }
        #endif
            return;
        }
#if defined(HAVE_GLOBAL_RNG) && defined(OPENSSL_EXTRA)
        if (rng == &globalRNG) {
            wc_UnLockMutex(&globalRNGMutex);
        }
#endif
```

src/ssl_sess.c
```diff
@@ -2165,13 +2165,24 @@ void AddSession(WOLFSSL* ssl)
         if (ssl->rng != NULL)
             rng = ssl->rng;
 #if defined(HAVE_GLOBAL_RNG) && defined(OPENSSL_EXTRA)
```
Frauschi, line 2171: **🟡 [Medium] initGlobalRNG is not re-checked after taking the lock (documented racy pattern elsewhere)**
*💡 SUGGEST*  `bug`

The decision `rng = &globalRNG` is made from an unlocked read of `initGlobalRNG`, and the mutex is only taken afterwards. `wolfSSL_RAND_Cleanup()` (`src/ssl_crypto.c:3980-3985`) calls `wc_FreeRng(&globalRNG)` and clears `initGlobalRNG` under that same mutex, so a concurrent cleanup that lands between the unlocked read and `wc_LockMutex()` leaves this code generating from a freed RNG. The existing in-tree pattern explicitly handles this — see `src/ssl_crypto.c:4071-4106`, which comments "the above access to initGlobalRNG is racey -- recheck it now that we have the lock" and re-tests `initGlobalRNG` while holding the mutex. Since this PR is specifically a data-race fix for this call site, it is worth closing the same window rather than leaving a narrower version of it.

**Suggestion:**
```suggestion
if (rng == &globalRNG) {
            if (wc_LockMutex(&globalRNGMutex) != 0) {
                WOLFSSL_MSG("Bad Lock Mutex rng");
                return;
            }
            /* the above access to initGlobalRNG is racey -- recheck it now
             * that we have the lock. */
            if (initGlobalRNG == 0) {
                wc_UnLockMutex(&globalRNGMutex);
                return;
            }
        }
```

src/pk_rsa.c
```diff
@@ -651,6 +651,12 @@ static int wolfssl_read_der_bio(WOLFSSL_BIO* bio, unsigned char** out)
         WOLFSSL_ERROR_MSG("DER SEQUENCE decode failed");
         err = 1;
     }
+    /* Cap at 8x the maximum modulus size, leaves headroom for the full
```
Frauschi, line 654: **🟡 [Medium] No test exercises the new oversized-DER rejection path**
*💡 SUGGEST*  `test`

The only existing coverage for this code path is `tests/api.c:20400-20416` (`d2i_RSAPrivateKey_bio` with NULL args, an empty BIO, and a valid key). Nothing feeds a BIO whose outer SEQUENCE header declares a length above the new cap, so the added branch is never executed and a future refactor of the bound would go unnoticed. This is cheap to cover because only the SEQUENCE header needs to be well-formed — the body never gets read once the length check trips.

**Suggestion:**
```suggestion
/* SEQUENCE, 4-byte length = 0x00FFFFFF, followed by nothing: the header
 * parses, the declared length exceeds the cap, so no allocation happens. */
static const byte hugeSeq[] = { 0x30, 0x84, 0x00, 0xFF, 0xFF, 0xFF,
                                0x00, 0x00, 0x00, 0x00 };
ExpectNotNull(bio = BIO_new_mem_buf(hugeSeq, (int)sizeof(hugeSeq)));
ExpectNull(d2i_RSAPrivateKey_bio(bio, &rsa));
```

tests/api/test_coding.c
```diff
@@ -335,13 +335,15 @@ int test_wc_Base64_EncodeDecisionCoverage(void)
         outLen = (word32)sizeof(enc);
```
Frauschi, line 337: **🔵 [Low] test_coding.c change guards the read but does not initialize `enc` as the PR describes**
*🔧 NIT*  `question`

The PR description states "Initialize `enc` in test_coding.c to avoid use before set" (CID 561978), but the change actually wraps the reads in `if (EXPECT_SUCCESS())`. The guard is functionally correct — `EXPECT_SUCCESS()` (tests/unit.h:157) is false once `ExpectIntEQ` on `Base64_Encode` fails, so `enc` and `outLen` are only read after a successful encode — but `enc` remains uninitialized, and Coverity's uninitialized-read checker generally does not model the `_ret` state machine, so the CID may well survive. Adding an explicit initializer is one line and makes the intent unambiguous for both the reader and the checker.

**Suggestion:**
```suggestion
byte enc[128];

        XMEMSET(enc, 0, sizeof(enc));
```

Frauschi (changes requested):
## 🐺 Skoll Code Review

**Findings:** 12 total — 4 posted, 8 skipped

### Posted findings
- [High] **globalRNGMutex leaked when wc_RNG_GenerateBlock() fails in AddSession()** — `src/ssl_sess.c:2172-2185`
- [Medium] **initGlobalRNG is not re-checked after taking the lock (documented racy pattern elsewhere)** — `src/ssl_sess.c:2167-2177`
- [Medium] **No test exercises the new oversized-DER rejection path** — `src/pk_rsa.c:654-659`
- [Low] **test_coding.c change guards the read but does not initialize `enc` as the PR describes** — `tests/api/test_coding.c:328-347`

- [High] globalRNGMutex leaked on wc_RNG_GenerateBlock failure path in AddSession
- [High] globalRNGMutex left locked when wc_RNG_GenerateBlock() fails in AddSession()
- [Medium] DER size cap expression `(RSA_MAX_SIZE / 8) * 8` reduces to `RSA_MAX_SIZE` and contradicts its comment
- [Medium] AddSession() global-RNG locking path has no regression test
- [Medium] initGlobalRNG not re-checked after acquiring globalRNGMutex in AddSession (races wolfSSL_RAND_Cleanup)
- [Low] Misindented return block does not match wolfSSL brace/indent style
- [Info] DER cap expression `(RSA_MAX_SIZE / 8) * 8` is a no-op and its comment contradicts the code
- [Info] New error-return block in AddSession() uses non-conforming indentation

</details>

src/pk_rsa.c
rlm2002: New test has been added see `test_wolfSSL_d2i_RSAPrivateKey_bio_oversized()`

src/ssl_sess.c
rlm2002: Fixed thanks.

src/ssl_sess.c
rlm2002: Fixed thanks.

tests/api/test_coding.c
rlm2002: Fixed, thanks.

tests/api.c
```diff
@@ -20491,6 +20491,68 @@ static int test_wolfSSL_d2i_PrivateKeys_bio(void)
 }
 #endif /* OPENSSL_ALL || (WOLFSSL_ASIO && !NO_RSA) */
 
+#if defined(OPENSSL_EXTRA) && !defined(NO_RSA) && defined(WOLFSSL_KEY_GEN) && \
```
Frauschi: **🟠 [High] New test's preprocessor guard does not match the API it calls - link failure in OPENSSL_EXTRA-without-OPENSSL_ALL builds**
*🚫 BLOCK*  `bug`

The new test body is guarded on `defined(OPENSSL_EXTRA) && !defined(NO_RSA) && defined(WOLFSSL_KEY_GEN) && ...`, but the function it calls, `d2i_RSAPrivateKey_bio` -> `wolfSSL_d2i_RSAPrivateKey_bio`, is compiled only under a *different* guard. In `src/pk_rsa.c:622-625` the definition is wrapped in `#if defined(OPENSSL_ALL) || defined(WOLFSSL_ASIO) || defined(WOLFSSL_HAPROXY) || defined(WOLFSSL_NGINX) || defined(WOLFSSL_QT)` plus `#if defined(WOLFSSL_KEY_GEN) && !defined(NO_BIO)`, and the prototype in `wolfssl/ssl.h:1567-1587` carries the same guard. `OPENSSL_EXTRA` does *not* imply `OPENSSL_ALL` anywhere in `configure.ac` or `settings.h` (only the reverse holds). The `d2i_RSAPrivateKey_bio` macro in `wolfssl/openssl/ssl.h:1398` is defined unconditionally, so the call site compiles (with an implicit-declaration warning, which is an error under the `-Werror` CI configs) and then fails at link with an undefined reference to `wolfSSL_d2i_RSAPrivateKey_bio`. Realistic configurations that hit this: `--enable-wpas` (configure.ac:3118-3144 sets `-DWOLFSSL_KEY_GEN` and `-DOPENSSL_EXTRA` but never `-DOPENSSL_ALL`; used in `.github/workflows/hostap-vm.yml:59`), `--enable-fortress` combined with `--enable-opensslextra`, and plain `--enable-opensslextra --enable-keygen`. The author's stated test config `--enable-jni` masks this because configure.ac:9435-9439 force-enables `OPENSSL_ALL` for JNI. Compare the sibling test in the same file, `test_wolfSSL_d2i_PrivateKeys_bio`, which is correctly guarded on `OPENSSL_ALL || (WOLFSSL_ASIO && !NO_RSA)`.

Verified by building the PR head: `./configure --enable-opensslextra --enable-keygen` fails with `tests/api.c:20542:16: error: use of undeclared identifier 'd2i_RSAPrivateKey_bio'`.

**Suggestion:**
```suggestion
#if (defined(OPENSSL_ALL) || defined(WOLFSSL_ASIO) ||                        \
     defined(WOLFSSL_HAPROXY) || defined(WOLFSSL_NGINX) ||                   \
     defined(WOLFSSL_QT)) &&                                                 \
    !defined(NO_RSA) && defined(WOLFSSL_KEY_GEN) &&                          \
    defined(USE_WOLFSSL_MEMORY) && !defined(WOLFSSL_NO_MALLOC) &&            \
    !defined(WOLFSSL_STATIC_MEMORY) && !defined(WOLFSSL_DEBUG_MEMORY) &&     \
    !defined(WOLFSSL_TRACK_MEMORY)
```

tests/api.c
```diff
@@ -20491,6 +20491,68 @@ static int test_wolfSSL_d2i_PrivateKeys_bio(void)
...
+    defined(USE_WOLFSSL_MEMORY) && !defined(WOLFSSL_NO_MALLOC) &&             \
+    !defined(WOLFSSL_STATIC_MEMORY) && !defined(WOLFSSL_DEBUG_MEMORY)
+/* Counts and refuses allocations >= big_alloc_threshold, to observe the
+ * oversized DER allocation in wolfssl_read_der_bio(). */
+static size_t big_alloc_threshold = 0; /* 0 = disabled */
+static int    big_alloc_attempts  = 0;
+
+static void* big_malloc_cb(size_t size)
```
Frauschi: **🟠 [High] Test installs a raw malloc callback while leaving the previous free/realloc callbacks in place - invalid free in --enable-trackmemory builds**
*🚫 BLOCK*  `bug`

`wolfSSL_SetAllocators(big_malloc_cb, prev_fc, prev_rc)` swaps in a malloc hook that returns memory from the C library `malloc()`, but keeps the *previous* free and realloc callbacks installed. That breaks the allocator triple whenever the previous allocators are not plain `malloc`/`free`/`realloc`. This is exactly the case in a `--enable-trackmemory` build: `configure.ac:5955-5965` defines only `WOLFSSL_TRACK_MEMORY` (no `WOLFSSL_DEBUG_MEMORY`, which is the only tracking-related macro the guard excludes), and `wolfCrypt_Init()` at `wolfcrypt/src/wc_port.c:525-531` calls `InitMemoryTracker()`, which installs `TrackMalloc/TrackFree/TrackRealloc` globally (`wolfssl/wolfcrypt/mem_track.h:381`). `TrackMalloc` over-allocates by `sizeof(memoryTrack)` and returns an interior pointer; `TrackFree` (`mem_track.h:243-290`) does `mt = (memoryTrack*)((byte*)ptr - sizeof(memoryTrack))`, reads `header->thisSize`, and on Linux/macOS (`DO_MEM_LIST`) also unlinks `header` from `ourMemList` before `free(mt)`. Concrete failure: with `./configure --enable-all --enable-trackmemory && make check`, the test's own `ExpectNotNull(bio = BIO_new(BIO_s_mem()))` allocates through `big_malloc_cb` (raw `malloc`), then `BIO_free(bio)` at line 20547 routes to `prev_fc == TrackFree`, which frees a pointer 32+ bytes before a valid heap block and writes through garbage `next`/`prev` pointers - heap corruption / crash, not just bad statistics. The existing precedent this test was modeled on, `test_write_dup_oom` (tests/api.c:35660-35662), correctly replaces all three callbacks with `oom_malloc_cb`/`oom_free_cb`/`oom_realloc_cb`, which are consistent plain `malloc`/`free`/`realloc` wrappers.

Verified by building and running the PR head: `./configure --enable-opensslall --enable-keygen --enable-trackmemory` builds cleanly, then `./tests/unit.test test_wolfSSL_d2i_RSAPrivateKey_bio_oversized` segfaults (exit 139). Crash backtrace: `TrackFree` <- `wolfSSL_BIO_free` <- `test_wolfSSL_d2i_RSAPrivateKey_bio_oversized` -- the BIO is allocated by the raw-malloc hook and freed by `TrackFree`, which dereferences a `memoryTrack` header that was never written.

**Suggestion:**
```suggestion
/* Supply a matching free/realloc so the allocator triple stays consistent. */
static void big_free_cb(void* ptr)
{
    free(ptr);
}

static void* big_realloc_cb(void* ptr, size_t size)
{
    if (big_alloc_threshold != 0 && size >= big_alloc_threshold) {
        big_alloc_attempts++;
        return NULL;
    }
    return realloc(ptr, size);
}
...
    ExpectIntEQ(wolfSSL_SetAllocators(big_malloc_cb, big_free_cb,
            big_realloc_cb), 0);
```

Frauschi (changes requested):
## 🐺 Skoll Code Review

### Posted findings
- [High] **New test's preprocessor guard does not match the API it calls - link failure in OPENSSL_EXTRA-without-OPENSSL_ALL builds** — `tests/api.c:20494-20520`
- [High] **Test installs a raw malloc callback while leaving the previous free/realloc callbacks in place - invalid free in --enable-trackmemory builds** — `tests/api.c:20502-20551`

- [Medium] Bit-count constant RSA_MAX_SIZE compared against a byte length
- [Medium] Triple-duplicated unlock blocks and a garbled comment in AddSession
- [Medium] No test coverage for the new globalRNG locking path in AddSession
- [Low] enc[outLen - 1] still underflows if Base64_Encode returns outLen == 0
- [Low] Redundant #ifndef NO_BIO block in testCases[] and a free of an always-NULL pointer

</details>

tests/api.c
rlm2002: Fixed gating on test. Thanks.

tests/api.c
rlm2002: Reworked test to avoid `trackmemory` issue

Frauschi (dismissed):
LGTM

dgarske (dismissed):
Over to @philljj as he's reviewing another PR with similar global mutex fixes

src/ssl_sess.c
```diff
@@ -2165,13 +2165,37 @@ void AddSession(WOLFSSL* ssl)
...
-        else if (initGlobalRNG == 1 || wolfSSL_RAND_Init() == WOLFSSL_SUCCESS) {
+        else if (initGlobalRNG == 1 ||
```
philljj: I can't find a code path where ssl->rng would ever be null. Are we sure this isn't dead code?

We could also try to use a local rng first, and only fallback to the global RNG on error. We follow this pattern in src/pk.c and other places:

```
        int initTmpRng = 0; 
        WC_RNG *rng = NULL;
        WC_DECLARE_VAR(tmpRng, WC_RNG, 1, 0);
                                                                                
        WC_ALLOC_VAR_EX(tmpRng, WC_RNG, 1, NULL, DYNAMIC_TYPE_RNG,
            return WOLFSSL_FATAL_ERROR);
        if (wc_InitRng(tmpRng) == 0) {
            rng = tmpRng;
            initTmpRng = 1;
        }
        else {
            WOLFSSL_MSG("Bad RNG Init, trying global");
            rng = wolfssl_get_global_rng();
        }
```

Then we would only hit the global lock when the local temp rng init fails (which should be rare).

philljj (changes requested):

src/ssl_sess.c
philljj: I was wrong, this branch is reachable per this other PR comment:
- https://github.com/wolfSSL/wolfssl/pull/11048#pullrequestreview-4860583267

Both PR #11048 and this PR are making this same change to `ssl_sess.c`.

philljj:
Merge conflict. Another PR (https://github.com/wolfSSL/wolfssl/pull/11048) just fixed the same issue in src/ssl_sess.c

src/ssl_sess.c
rlm2002: Got it, I'll take the `globalRNG` changes out.

philljj (approved):

PR #11009, opened 2026-07-30
Fix multiple reported issues.

kareem-wolfssl:
# Description

Fixes zd#22229

# Testing

Built in tests, provided reproducers.

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -12108,11 +12108,6 @@ retry:
 void ShrinkOutputBuffer(WOLFSSL* ssl)
 {
     WOLFSSL_MSG("Shrinking output buffer");
-    /* Only the dynamic buffer is heap allocated. When dynamicFlag is clear the
-     * buffer already is the inline staticBuffer array, and in a
-     * LARGE_STATIC_BUFFERS build a pending flight can live there, so freeing
-     * unconditionally would pass an interior pointer of the WOLFSSL struct to
-     * XFREE. */
     if (ssl->buffers.outputBuffer.dynamicFlag) {
         XFREE(ssl->buffers.outputBuffer.buffer -
                   ssl->buffers.outputBuffer.offset,
@@ -12122,15 +12117,8 @@ void ShrinkOutputBuffer(WOLFSSL* ssl)
     ssl->buffers.outputBuffer.bufferSize  = STATIC_BUFFER_LEN;
     ssl->buffers.outputBuffer.dynamicFlag = 0;
     ssl->buffers.outputBuffer.offset      = 0;
-    /* Drop idx/length with the buffer they described. bufferSize shrinks to
-     * STATIC_BUFFER_LEN here, so leaving them set would break the
-     * idx + length <= bufferSize invariant that the rest of the output
-     * accounting relies on: CheckAvailableSize()'s word32 space computation
-     * would underflow instead of growing, and SendBuffered()/GetOutputBuffer()
-     * would address past the end of staticBuffer. Callers reaching this
-     * function with output still queued are abandoning that output. */
-    ssl->buffers.outputBuffer.idx    = 0;
-    ssl->buffers.outputBuffer.length = 0;
+    ssl->buffers.outputBuffer.idx         = 0;
+    ssl->buffers.outputBuffer.length      = 0;
 }
 
 
--- a/src/internal.c
+++ b/src/internal.c
@@ -12110,11 +12110,6 @@ retry:
 void ShrinkOutputBuffer(WOLFSSL* ssl)
 {
     WOLFSSL_MSG("Shrinking output buffer");
-    /* Only the dynamic buffer is heap allocated. When dynamicFlag is clear the
-     * buffer already is the inline staticBuffer array, and in a
-     * LARGE_STATIC_BUFFERS build a pending flight can live there, so freeing
-     * unconditionally would pass an interior pointer of the WOLFSSL struct to
-     * XFREE. */
     if (ssl->buffers.outputBuffer.dynamicFlag) {
         XFREE(ssl->buffers.outputBuffer.buffer -
                   ssl->buffers.outputBuffer.offset,
@@ -12124,15 +12119,8 @@ void ShrinkOutputBuffer(WOLFSSL* ssl)
     ssl->buffers.outputBuffer.bufferSize  = STATIC_BUFFER_LEN;
     ssl->buffers.outputBuffer.dynamicFlag = 0;
     ssl->buffers.outputBuffer.offset      = 0;
-    /* Drop idx/length with the buffer they described. bufferSize shrinks to
-     * STATIC_BUFFER_LEN here, so leaving them set would break the
-     * idx + length <= bufferSize invariant that the rest of the output
-     * accounting relies on: CheckAvailableSize()'s word32 space computation
-     * would underflow instead of growing, and SendBuffered()/GetOutputBuffer()
-     * would address past the end of staticBuffer. Callers reaching this
-     * function with output still queued are abandoning that output. */
-    ssl->buffers.outputBuffer.idx    = 0;
-    ssl->buffers.outputBuffer.length = 0;
+    ssl->buffers.outputBuffer.idx         = 0;
+    ssl->buffers.outputBuffer.length      = 0;
 }
 
 
--- a/src/internal.c
+++ b/src/internal.c
@@ -12104,17 +12104,16 @@ retry:
 }
 
 
-/* Switch dynamic output buffer back to static, discarding any pending output.
- * Safe to call whether or not the buffer is dynamic; callers need not check
- * dynamicFlag first. */
+/* Switch dynamic output buffer back to static, discarding pending output.
+ * Safe on a static buffer. */
 void ShrinkOutputBuffer(WOLFSSL* ssl)
 {
     WOLFSSL_MSG("Shrinking output buffer");
     if (ssl->buffers.outputBuffer.dynamicFlag) {
-        /* Discarded records may hold plaintext, so wipe before releasing,
-         * as ShrinkInputBuffer does. */
+        /* Only the region still holding record data; 0 after a full flush. */
         ForceZero(ssl->buffers.outputBuffer.buffer,
-                  ssl->buffers.outputBuffer.bufferSize);
+                  ssl->buffers.outputBuffer.idx +
+                      ssl->buffers.outputBuffer.length);
         XFREE(ssl->buffers.outputBuffer.buffer -
                   ssl->buffers.outputBuffer.offset,
               ssl->heap, DYNAMIC_TYPE_OUT_BUFFER);
--- a/src/internal.c
+++ b/src/internal.c
@@ -12104,11 +12104,17 @@ retry:
 }
 
 
-/* Switch dynamic output buffer back to static, discarding any pending output */
+/* Switch dynamic output buffer back to static, discarding any pending output.
+ * Safe to call whether or not the buffer is dynamic; callers need not check
+ * dynamicFlag first. */
 void ShrinkOutputBuffer(WOLFSSL* ssl)
 {
     WOLFSSL_MSG("Shrinking output buffer");
     if (ssl->buffers.outputBuffer.dynamicFlag) {
+        /* Discarded records may hold plaintext, so wipe before releasing,
+         * as ShrinkInputBuffer does. */
+        ForceZero(ssl->buffers.outputBuffer.buffer,
+                  ssl->buffers.outputBuffer.bufferSize);
         XFREE(ssl->buffers.outputBuffer.buffer -
                   ssl->buffers.outputBuffer.offset,
               ssl->heap, DYNAMIC_TYPE_OUT_BUFFER);
--- a/src/internal.c
+++ b/src/internal.c
@@ -12104,17 +12104,33 @@ retry:
 }
 
 
-/* Switch dynamic output buffer back to static, buffer is assumed clear */
+/* Switch dynamic output buffer back to static, discarding any pending output */
 void ShrinkOutputBuffer(WOLFSSL* ssl)
 {
     WOLFSSL_MSG("Shrinking output buffer");
-    XFREE(ssl->buffers.outputBuffer.buffer - ssl->buffers.outputBuffer.offset,
-          ssl->heap, DYNAMIC_TYPE_OUT_BUFFER);
+    /* Only the dynamic buffer is heap allocated. When dynamicFlag is clear the
+     * buffer already is the inline staticBuffer array, and in a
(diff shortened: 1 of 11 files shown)


src/x509_str.c
```diff
@@ -997,11 +996,19 @@ int wolfSSL_X509_verify_cert(WOLFSSL_X509_STORE_CTX* ctx)
...
+                 * caller-supplied issuer left behind would sit in the shared
+                 * CertManager as a fully usable WOLFSSL_TEMP_CA anchor -
+                 * GetCA()/GetCAByName() do not inspect signer->type, so every
+                 * other consumer of this CertManager (native TLS peer
+                 * verification, OCSP, CRL, CM_VerifyBuffer_ex) would treat it
+                 * like a configured trust root. X509VerifyCertSetupRetry()
+                 * below only removes ctx->current_cert, the child. */
+                X509StoreRemoveCa(ctx->store, issuer, WOLFSSL_TEMP_CA);
                 X509VerifyCertSetupRetry(ctx, certs, failedCerts,
                     &depth, origDepth);
```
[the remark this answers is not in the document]
julek-wolfssl: This doesn't solve the problem. It just acts as a retry. Erroring out directly from `X509StoreRemoveCa` should be fine.

julek-wolfssl (changes requested):
comments are way too verbose

kareem-wolfssl:
I removed the X509StoreRemoveCa call.  Adding error checking would have caused issues in the case that the cert was already added previously.  AddCA would return success, but the cert would not be added as TEMP_CA type, so X509StoreRemoveCa would error out.

The TEMP_CAs will be removed either way by the unconditional wolfSSL_CertManagerUnloadTempIntermediateCerts call.

src/ssl.c
```diff
@@ -5659,6 +5659,17 @@ size_t wolfSSL_get_client_random(const WOLFSSL* ssl, unsigned char* out,
         ssl->earlyData = no_early_data;
         ssl->earlyDataSz = 0;
     #endif
+    #ifdef HAVE_RPK
+        {
```
dgarske: Let's not use empty brace. Thanks

src/x509_str.c
```diff
@@ -1149,12 +1142,11 @@ int wolfSSL_X509_verify_cert(WOLFSSL_X509_STORE_CTX* ctx)
...
     }
-    /* Remove intermediates that were added to CM */
+    /* Remove intermediates that were added to CM. Unconditional: anything left
+     * resident anchors verification for every other user of this CM. */
     if (ctx != NULL) {
         if (ctx->store != NULL) {
-            if (added == 1) {
-                wolfSSL_CertManagerUnloadTempIntermediateCerts(ctx->store->cm);
-            }
+            wolfSSL_CertManagerUnloadTempIntermediateCerts(ctx->store->cm);
```
dgarske: Consider comment indicating return code is not checked or have `(void)` or front or even consider a WOLFSSL_MSG on failure.

src/internal.c
```diff
@@ -11691,17 +11691,21 @@ static int wolfSSLReceive(WOLFSSL* ssl, byte* buf, word32 sz)
...
 
-/* Switch dynamic output buffer back to static, buffer is assumed clear */
+/* Switch dynamic output buffer back to static, discarding any pending output */
 void ShrinkOutputBuffer(WOLFSSL* ssl)
 {
     WOLFSSL_MSG("Shrinking output buffer");
-    XFREE(ssl->buffers.outputBuffer.buffer - ssl->buffers.outputBuffer.offset,
-          ssl->heap, DYNAMIC_TYPE_OUT_BUFFER);
+    if (ssl->buffers.outputBuffer.dynamicFlag) {
+        XFREE(ssl->buffers.outputBuffer.buffer -
```
dgarske, line 12119: Does this need ForceZero()?

src/x509_str.c
```diff
@@ -1023,13 +1019,10 @@ int wolfSSL_X509_verify_cert(WOLFSSL_X509_STORE_CTX* ctx)
                     != WOLFSSL_SUCCESS) {
                 /* Could not guarantee the temporary intermediates were
                  * dropped; fail closed rather than risk verifying the current
-                 * certificate against one.  Leave `added` set: they are still
-                 * loaded, so the exit cleanup makes a final attempt to drop
-                 * them. */
+                 * certificate against one. */
```
dgarske, line 1105: At line 1030 drop the `added == 1` reference now that the variable no longer exists.

tests/api/test_tls13.c
```diff
@@ -4078,6 +4078,60 @@ int test_tls13_rpk_trust(void)
...
 
+int test_tls13_rpk_unoffered_cert_type(void)
+{
+    EXPECT_DECLS;
+#if defined(HAVE_RPK) && defined(WOLFSSL_TLS13) && \
+    !defined(NO_WOLFSSL_CLIENT) && !defined(NO_WOLFSSL_SERVER)
+    int isServerType;
+
+    /* round 0: server_cert_type. round 1: the client_cert_type twin. */
+    for (isServerType = 1; isServerType >= 0; isServerType--) {
```
dgarske: Should you add `&& EXPECT_SUCCESS()` to the loop condition to match the sibling test added in this PR?

src/internal.c
```diff
@@ -11691,17 +11691,21 @@ static int wolfSSLReceive(WOLFSSL* ssl, byte* buf, word32 sz)
...
 
 
-/* Switch dynamic output buffer back to static, buffer is assumed clear */
+/* Switch dynamic output buffer back to static, discarding any pending output */
 void ShrinkOutputBuffer(WOLFSSL* ssl)
 {
     WOLFSSL_MSG("Shrinking output buffer");
-    XFREE(ssl->buffers.outputBuffer.buffer - ssl->buffers.outputBuffer.offset,
-          ssl->heap, DYNAMIC_TYPE_OUT_BUFFER);
+    if (ssl->buffers.outputBuffer.dynamicFlag) {
```
dgarske, line 12114: Now that the `XFREE` is guarded internally, the callers at src/internal.c:9326 (FreeSSL), src/internal.c:11876 (SendBuffered) and src/sniffer.c:6638 all wrap the call in an outer `if (ssl->buffers.outputBuffer.dynamicFlag)` that no longer serves a purpose - only the DTLS `SendAlert()` site (src/internal.c:28569) actually needed the new guard. Leaving both makes it ambiguous which layer owns the precondition. This is cosmetic and does not affect behavior; flagged only so the PR can decide where the check should live.
Recommendation: Either remove the now-redundant caller-side guards or document that the function is safe to call unconditionally

dgarske (changes requested):

dgarske:
Jenkins retest this please "python port test"

src/x509_str.c
```diff
@@ -1777,13 +1772,15 @@ static void X509StoreFreeObjList(WOLFSSL_X509_STORE* store,
...
+         * CRL object appended after them must not shift this window. */
+        if (obj != NULL && obj->type == WOLFSSL_X509_LU_X509) {
+            /* The inner X509 is owned by somebody else, NULL out the ref */
             obj->type = (WOLFSSL_X509_LOOKUP_TYPE)0;
             obj->data.ptr = NULL;
+            cnt--;
         }
-        cnt--;
         i--;
     }
```
julek-wolfssl: `X509StoreFreeObjList` is a mess. Let's instead increase the reference counter of the objects placed inside of a `WOLFSSL_X509_OBJECT` (both x509 and x509_crl have ref counters) and just always call the correct `_free` function.

src/x509_str.c
```diff
@@ -2625,8 +2626,28 @@ WOLF_STACK_OF(WOLFSSL_X509_OBJECT)* wolfSSL_X509_STORE_get0_objects(
...
+                cur = (WOLFSSL_X509_OBJECT*)wolfSSL_sk_X509_OBJECT_value(ret,
+                                                                         j);
+                if (cur != NULL) {
+                    cur->type = (WOLFSSL_X509_LOOKUP_TYPE)0;
+                    cur->data.ptr = NULL;
+                }
+            }
+            wolfSSL_sk_X509_OBJECT_pop_free(ret, NULL);
+        }
+    }
```
julek-wolfssl: Again, just increase the reference counter when placing objects inside the stack and simplify the cleanup logic to always call the free.

julek-wolfssl (dismissed):

src/tls.c
```diff
@@ -13543,6 +13543,22 @@ static int TLSX_ClientCertificateType_Parse(WOLFSSL* ssl, const byte* input,
...
+                WOLFSSL_MSG("client_cert_type received but never offered");
+                WOLFSSL_ERROR_VERBOSE(UNSUPPORTED_EXTENSION);
+                return UNSUPPORTED_EXTENSION;
+            }
+            if (!IsCertTypeListed(*input,
+                    ssl->options.rpkState.sending_ClientCertTypeCnt,
+                    ssl->options.rpkState.sending_ClientCertTypes)) {
+                WOLFSSL_MSG("client_cert_type value was not offered");
+                WOLFSSL_ERROR_VERBOSE(UNSUPPORTED_EXTENSION);
+                return UNSUPPORTED_EXTENSION;
```
dgarske, line 13561: Do we need to add `SendAlert(ssl, alert_fatal, unsupported_extension);` before each of the four new `return UNSUPPORTED_EXTENSION` statements, matching the existing unsolicited-extension convention in this file?

src/internal.c
```diff
@@ -12104,17 +12104,27 @@ static int wolfSSLReceive(WOLFSSL* ssl, byte* buf, word32 sz)
...
+ * dynamicFlag first. */
 void ShrinkOutputBuffer(WOLFSSL* ssl)
 {
     WOLFSSL_MSG("Shrinking output buffer");
-    XFREE(ssl->buffers.outputBuffer.buffer - ssl->buffers.outputBuffer.offset,
-          ssl->heap, DYNAMIC_TYPE_OUT_BUFFER);
+    if (ssl->buffers.outputBuffer.dynamicFlag) {
+        /* Discarded records may hold plaintext, so wipe before releasing,
+         * as ShrinkInputBuffer does. */
+        ForceZero(ssl->buffers.outputBuffer.buffer,
```
dgarske, line 12116: ShrinkOutputBuffer now ForceZeros the whole buffer on every flush of every record

`SendBuffered()` calls `ShrinkOutputBuffer()` unconditionally at the end of every successful flush when the buffer is dynamic (src/internal.c:12294-12296), and `GrowOutputBuffer()` re-allocates for the next record. The new `ForceZero(buffer, bufferSize)` therefore adds a full-buffer wipe -- up to `MAX_RECORD_SIZE` + overhead (~16.6 KB, larger with `WOLFSSL_MAX_FRAGMENT`) -- to the hot per-record send path. At that call site `idx` and `length` are already 0 and the residue is ciphertext (records are encrypted in place before being buffered), so the wipe buys little there; the case the comment actually describes is the `SendAlert()` discard path at src/internal.c:29234. Note also that the comment says the discarded data 'may hold plaintext', which is only accurate for pre-encryption handshake records.

Recommendation: Limit the wipe to `idx + length` (the region that actually holds record data) so the common post-flush shrink costs nothing, while the SendAlert discard path still gets wiped. If the full-buffer wipe is deliberate, note the per-record cost in the comment.

src/x509_str.c
```diff
@@ -1239,11 +1228,17 @@ int wolfSSL_X509_verify_cert(WOLFSSL_X509_STORE_CTX* ctx)
...
         }
     }
-    /* Remove intermediates that were added to CM */
+    /* Remove intermediates that were added to CM. Unconditional: anything left
+     * resident anchors verification for every other user of this CM. */
     if (ctx != NULL) {
         if (ctx->store != NULL) {
-            if (added == 1) {
-                wolfSSL_CertManagerUnloadTempIntermediateCerts(ctx->store->cm);
+            if (wolfSSL_CertManagerUnloadTempIntermediateCerts(ctx->store->cm)
```
dgarske, line 1236: Recommendation: Either keep a cheap 'we touched the CM' flag to skip the unload when this call provably added nothing, or document that concurrent `X509_verify_cert()` / `wolfSSL_get0_verified_chain()` on one store is unsupported. At minimum, add a comment at the call site noting the shared-CM interaction so the next reader does not have to re-derive it.

src/x509_str.c
```diff
@@ -2703,33 +2671,44 @@ WOLF_STACK_OF(WOLFSSL_X509_OBJECT)* wolfSSL_X509_STORE_get0_objects(
...
+        if (wolfSSL_X509_up_ref(x509) != WOLFSSL_SUCCESS) {
+            WOLFSSL_MSG("wolfSSL_X509_up_ref error");
+            goto err_cleanup;
+        }
+        if (wolfSSL_sk_X509_push(cert_stack, x509) <= 0) {
+            WOLFSSL_MSG("wolfSSL_sk_X509_push error");
+            wolfSSL_X509_free(x509);
+            goto err_cleanup;
         }
+        store->numAdded++;
```
dgarske: store->numAdded is now write-only dead state

`X509StoreFreeObjList()` was the only reader of `store->numAdded`. With it removed, the field is written in three places (`wolfSSL_X509_STORE_new` at line 1829, and the reset/increment at 2666 and 2687) and read nowhere in the tree. Its declaration comment in `wolfssl/ssl.h:892` -- 'Number of objs in objs that are in certs sk' -- no longer describes anything the code relies on, since every object now owns its own reference. Leaving live-looking bookkeeping behind will mislead the next reader into thinking the ownership split still matters.


Recommendation: Drop the `store->numAdded` writes (keeping the struct field for ABI), or add a one-line comment on the ssl.h declaration noting it is retained for ABI only and no longer used.

src/internal.c
```diff
@@ -12104,17 +12104,27 @@ static int wolfSSLReceive(WOLFSSL* ssl, byte* buf, word32 sz)
 }
 
 
-/* Switch dynamic output buffer back to static, buffer is assumed clear */
+/* Switch dynamic output buffer back to static, discarding any pending output.
```
dgarske: ShrinkOutputBuffer comment claims callers need not check dynamicFlag, but all four still do

The new doc comment states 'Safe to call whether or not the buffer is dynamic; callers need not check dynamicFlag first.' All four existing call sites still guard on it: src/internal.c:9712-9713 (`SSL_ResourceFree`), src/internal.c:12295-12296 (`SendBuffered`), and src/sniffer.c:6656-6657. Only the SendAlert path at src/internal.c:29234 calls it unguarded. The comment and the code now disagree about the intended contract.

Recommendation: Either remove the now-redundant `if (ssl->buffers.outputBuffer.dynamicFlag)` guards at the three call sites, or soften the comment to 'safe to call on a static buffer' without implying the callers were updated.

dgarske (changes requested):

dgarske (approved):

PR #11016, opened 2026-07-30
dtls13: epoch mangament improvement and bug fixes

rizlik:
1. NULL-ify Dtls13DecryptEcpoch if the pointed epoch is evicted
2. Use only epoch that can decrypt as candidate for decrypting records
3. Never evict Peer epoch - 1 as it can be active


julek-wolfssl (dismissed):

dgarske (dismissed):

dgarske (commented):
Please resolve merge conflicts. Thanks @rizlik

dgarske (approved):

PR #11017, opened 2026-07-30
Fix truncation and noDegenerate bypasses in wc_PKCS7_VerifySignedData

miyazakh:
## Summary

`wc_PKCS7_VerifySignedData()` had two bugs that let signature verification be silently bypassed (F-7302). Both returned `0` (success) with `pkcs7->content` populated from unauthenticated data.

1. Default streaming build (`NO_PKCS7_STREAM` undefined): a bundle truncated 0-5 bytes short of a complete `signerInfos` SET was accepted as a valid degenerate (certs-only) end.(f7302)
2. Both build modes: a well-formed bundle with non-empty `digestAlgorithms` and empty `signerInfos` bypassed `wc_PKCS7_AllowDegenerate(pkcs7, 0)` entirely.

## Bug 1: truncation accepted as a valid degenerate end

`WC_PKCS7_VERIFY_STAGE3` treated any input with fewer than 6 bytes remaining after the content as a finished, successful parse, unable to distinguish a genuine minimal degenerate end (`31 00`) from a truncated bundle. Dates back to `3cad0bfe5`, which left the gap with an acknowledging comment.

**Fix:** stage 3 no longer returns early. It caps `pkcs7->stream->expected` to the bytes actually remaining, then lets stage 4/5/6's existing bounds checks run normally, so truncated input fails there instead of being silently accepted.

## Bug 2: noDegenerate bypass via non-empty digestAlgorithms

`wc_PKCS7_ParseSignerInfo()` only enforced `noDegenerate` when the whole remaining buffer was empty, never when the accurately-computed `degenerate` flag (from the real `signerInfos` SET) was true. No truncation needed to trigger this.

**Fix:** one line, broadened the guard to also check `degenerate == 1`:

```c
if (pkcs7->noDegenerate == 1 && (inSz == 0 || degenerate == 1)) {
```

## Tests added

- `test_wc_PKCS7_VerifySignedData_DegenerateMinimal` -- minimal legit degenerate bundle still succeeds.
- `test_wc_PKCS7_VerifySignedData_TruncSignerInfosTag` / `NoSignerInfosTag` -- truncated/missing `signerInfos` rejected.
- `test_wc_PKCS7_VerifySignedData_DegenerateNonEmptyDigestAlgos` -- bug 2 case returns `PKCS7_NO_SIGNER_E`.
- Dropped the `NO_PKCS7_STREAM`-only gate on `test_wc_PKCS7_VerifySignedData_TruncCertSetTag`; it now passes in streaming mode too.

## Testing

Verified in isolated `git worktree` builds (`--enable-pkcs7 --enable-opensslextra`, plus `CFLAGS=-DNO_PKCS7_STREAM` for the non-streaming build), since the primary tree doesn't enable PKCS7. Each bug: negative test failed pre-fix, passed post-fix. No regressions in `pkcs7_sd`, `wolfcrypt/test/testwolfcrypt`, or the full `tests/unit.test` suite, in both build modes.

### DIFF
--- a/tests/api/test_pkcs7.c
+++ b/tests/api/test_pkcs7.c
@@ -6483,6 +6483,10 @@ int test_wc_PKCS7_VerifySignedData_NoSignerInfosTag(void)
     EXPECT_DECLS;
 #if defined(HAVE_PKCS7)
     PKCS7* pkcs7 = NULL;
+#ifndef NO_PKCS7_STREAM
+    int ret;
+    word32 idx;
+#endif
 
     WOLFSSL_SMALL_STACK_STATIC byte der[] = {
         0x30, 0x61,
@@ -6507,11 +6511,35 @@ int test_wc_PKCS7_VerifySignedData_NoSignerInfosTag(void)
     };
     word32 derSz = (word32)sizeof(der);
 
+    /* single-shot: must fail with a real parse error, not WANT_READ_E
+     * (no more bytes will ever arrive per the outer length) */
     ExpectNotNull(pkcs7 = wc_PKCS7_New(HEAP_HINT, testDevId));
     ExpectIntEQ(wc_PKCS7_Init(pkcs7, HEAP_HINT, INVALID_DEVID), 0);
     ExpectIntEQ(wc_PKCS7_InitWithCert(pkcs7, NULL, 0), 0);
     ExpectIntNE(wc_PKCS7_VerifySignedData(pkcs7, der, derSz), 0);
+    ExpectIntNE(wc_PKCS7_VerifySignedData(pkcs7, der, derSz),
+            WC_NO_ERR_TRACE(WC_PKCS7_WANT_READ_E));
     wc_PKCS7_Free(pkcs7);
+    pkcs7 = NULL;
+
+#ifndef NO_PKCS7_STREAM
+    /* same bundle fed one byte at a time: must not end stuck on
+     * WANT_READ_E once all available bytes are consumed */
+    ExpectNotNull(pkcs7 = wc_PKCS7_New(HEAP_HINT, testDevId));
+    ExpectIntEQ(wc_PKCS7_Init(pkcs7, HEAP_HINT, INVALID_DEVID), 0);
+    ExpectIntEQ(wc_PKCS7_InitWithCert(pkcs7, NULL, 0), 0);
+
+    ret = WC_NO_ERR_TRACE(WC_PKCS7_WANT_READ_E);
+    for (idx = 0; idx < derSz && ret != 0; idx++) {
+        ret = wc_PKCS7_VerifySignedData(pkcs7, der + idx, 1);
+        if (ret < 0 && ret != WC_NO_ERR_TRACE(WC_PKCS7_WANT_READ_E)) {
+            break;
+        }
+    }
+    ExpectIntNE(ret, 0);
+    ExpectIntNE(ret, WC_NO_ERR_TRACE(WC_PKCS7_WANT_READ_E));
+    wc_PKCS7_Free(pkcs7);
+#endif /* !NO_PKCS7_STREAM */
 
 #endif /* HAVE_PKCS7 */
     return EXPECT_RESULT();
--- a/tests/api/test_pkcs7.c
+++ b/tests/api/test_pkcs7.c
@@ -6466,6 +6466,8 @@ int test_wc_PKCS7_VerifySignedData_TruncSignerInfosTag(void)
     ExpectIntEQ(wc_PKCS7_Init(pkcs7, HEAP_HINT, INVALID_DEVID), 0);
     ExpectIntEQ(wc_PKCS7_InitWithCert(pkcs7, NULL, 0), 0);
     ExpectIntNE(wc_PKCS7_VerifySignedData(pkcs7, der, derSz), 0);
+    ExpectIntNE(wc_PKCS7_VerifySignedData(pkcs7, der, derSz),
+            WC_NO_ERR_TRACE(WC_PKCS7_WANT_READ_E));
     wc_PKCS7_Free(pkcs7);
 
 #endif /* HAVE_PKCS7 */
--- a/tests/api/test_pkcs7.c
+++ b/tests/api/test_pkcs7.c
@@ -6285,19 +6285,15 @@ int test_wc_PKCS7_VerifySignedData_TruncEContentTag(void)
 /*
  * SignedData bundle truncated at the certificates [0] IMPLICIT tag.
  * Verifies that the parser rejects the malformed input rather than
- * dereferencing past the end of the buffer.
- *
- * TODO: limited to NO_PKCS7_STREAM because the streaming parser's stage 3
- * early-exit check (pkcs7.c near line 6594) accepts any bundle
- * whose remaining footer is < 6 bytes as a successful degenerate end,
- * so the bounds check at line 6765 is unreachable in streaming mode.
- * Drop the NO_PKCS7_STREAM gate if/when the early-exit check becomes
- * more accurate.
+ * dereferencing past the end of the buffer. Runs in both streaming and
+ * NO_PKCS7_STREAM builds: the streaming parser's stage 3 early-exit check
+ * used to accept any bundle whose remaining footer was < 6 bytes as a
+ * successful degenerate end, silently accepting this truncated input.
  */
 int test_wc_PKCS7_VerifySignedData_TruncCertSetTag(void)
 {
     EXPECT_DECLS;
-#if defined(HAVE_PKCS7) && defined(NO_PKCS7_STREAM)
+#if defined(HAVE_PKCS7)
     PKCS7* pkcs7 = NULL;
 
     WOLFSSL_SMALL_STACK_STATIC byte der[] = {
@@ -6337,7 +6333,188 @@ int test_wc_PKCS7_VerifySignedData_TruncCertSetTag(void)
     ExpectIntNE(wc_PKCS7_VerifySignedData(pkcs7, der, derSz), 0);
     wc_PKCS7_Free(pkcs7);
 
-#endif /* HAVE_PKCS7 && NO_PKCS7_STREAM */
+#endif /* HAVE_PKCS7 */
+    return EXPECT_RESULT();
+}
+
+/*
+ * SignedData bundle that is a genuine, non-truncated degenerate
+ * (certs-only) bundle: no certificates, no CRLs, and an empty signerInfos
+ * SET ("31 00") closing the bundle right after the content. This is the
+ * shortest legitimate ending the stage 3 tail check in the streaming
+ * parser can see, and must still succeed after fixing that check to reject
+ * truncated bundles.
+ */
+int test_wc_PKCS7_VerifySignedData_DegenerateMinimal(void)
+{
+    EXPECT_DECLS;
+#if defined(HAVE_PKCS7)
+    PKCS7* pkcs7 = NULL;
+
+    WOLFSSL_SMALL_STACK_STATIC byte der[] = {
+        /* outer ContentInfo SEQUENCE (99 bytes content) */
+        0x30, 0x63,
+        /* contentType OID signedData */
+        0x06, 0x09, 0x2A, 0x86, 0x48, 0x86, 0xF7, 0x0D, 0x01, 0x07, 0x02,
+        /* [0] EXPLICIT (86 bytes content) */
+        0xA0, 0x56,
+        /* SignedData SEQUENCE (84 bytes content) */
+        0x30, 0x54,
+        /* version INTEGER 1 */
+        0x02, 0x01, 0x01,
+        /* digestAlgorithms SET (empty - degenerate) */
+        0x31, 0x00,
+        /* encapContentInfo SEQUENCE (75 bytes content) */
+        0x30, 0x4B,
+        /* eContentType OID 1.2.840.113549.1.7.1 (data) */
+        0x06, 0x09, 0x2A, 0x86, 0x48, 0x86, 0xF7, 0x0D, 0x01, 0x07, 0x01,
+        /* eContent [0] EXPLICIT (62 bytes content) */
+        0xA0, 0x3E,
+        /* OCTET STRING (60 bytes content) */
+        0x04, 0x3C,
+        0x00, 0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07,
+        0x08, 0x09, 0x0A, 0x0B, 0x0C, 0x0D, 0x0E, 0x0F,
+        0x10, 0x11, 0x12, 0x13, 0x14, 0x15, 0x16, 0x17,
+        0x18, 0x19, 0x1A, 0x1B, 0x1C, 0x1D, 0x1E, 0x1F,
+        0x20, 0x21, 0x22, 0x23, 0x24, 0x25, 0x26, 0x27,
+        0x28, 0x29, 0x2A, 0x2B, 0x2C, 0x2D, 0x2E, 0x2F,
+        0x30, 0x31, 0x32, 0x33, 0x34, 0x35, 0x36, 0x37,
+        0x38, 0x39, 0x3A, 0x3B,
+        /* signerInfos SET (empty - degenerate end) */
+        0x31, 0x00
+    };
+    word32 derSz = (word32)sizeof(der);
+
+    /* single-shot call */
+    ExpectNotNull(pkcs7 = wc_PKCS7_New(HEAP_HINT, testDevId));
+    ExpectIntEQ(wc_PKCS7_Init(pkcs7, HEAP_HINT, INVALID_DEVID), 0);
+    ExpectIntEQ(wc_PKCS7_InitWithCert(pkcs7, NULL, 0), 0);
+    ExpectIntEQ(wc_PKCS7_VerifySignedData(pkcs7, der, derSz), 0);
+    if (pkcs7 != NULL) {
+        ExpectIntEQ(pkcs7->contentSz, 60);
+        ExpectNotNull(pkcs7->content);
+    }
+    wc_PKCS7_Free(pkcs7);
+    pkcs7 = NULL;
+
+#ifndef NO_PKCS7_STREAM
+    /* same bundle fed one byte at a time */
+    {
+    int ret;
+    word32 idx;
+
+    ExpectNotNull(pkcs7 = wc_PKCS7_New(HEAP_HINT, testDevId));
+    ExpectIntEQ(wc_PKCS7_Init(pkcs7, HEAP_HINT, INVALID_DEVID), 0);
+    ExpectIntEQ(wc_PKCS7_InitWithCert(pkcs7, NULL, 0), 0);
+
+    ret = WC_NO_ERR_TRACE(WC_PKCS7_WANT_READ_E);
+    for (idx = 0; idx < derSz && ret != 0; idx++) {
+        ret = wc_PKCS7_VerifySignedData(pkcs7, der + idx, 1);
+        if (ret < 0 && ret != WC_NO_ERR_TRACE(WC_PKCS7_WANT_READ_E)) {
+            break;
+        }
+    }
+    ExpectIntEQ(ret, 0);
+    if (pkcs7 != NULL) {
+        ExpectIntEQ(pkcs7->contentSz, 60);
+        ExpectNotNull(pkcs7->content);
+    }
+    wc_PKCS7_Free(pkcs7);
+    }
+#endif /* !NO_PKCS7_STREAM */
+
+#endif /* HAVE_PKCS7 */
+    return EXPECT_RESULT();
+}
+
+/*
+ * Same shape as test_wc_PKCS7_VerifySignedData_DegenerateMinimal, but the
+ * final 2 bytes are "31 01" instead of "31 00": a signerInfos SET claiming
+ * one byte of content that the buffer never supplies. Must be rejected,
+ * not treated as a successful degenerate end.
+ */
+int test_wc_PKCS7_VerifySignedData_TruncSignerInfosTag(void)
+{
+    EXPECT_DECLS;
+#if defined(HAVE_PKCS7)
+    PKCS7* pkcs7 = NULL;
+
+    WOLFSSL_SMALL_STACK_STATIC byte der[] = {
+        0x30, 0x63,
+        0x06, 0x09, 0x2A, 0x86, 0x48, 0x86, 0xF7, 0x0D, 0x01, 0x07, 0x02,
+        0xA0, 0x56,
+        0x30, 0x54,
+        0x02, 0x01, 0x01,
+        0x31, 0x00,
+        0x30, 0x4B,
+        0x06, 0x09, 0x2A, 0x86, 0x48, 0x86, 0xF7, 0x0D, 0x01, 0x07, 0x01,
+        0xA0, 0x3E,
+        0x04, 0x3C,
+        0x00, 0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07,
+        0x08, 0x09, 0x0A, 0x0B, 0x0C, 0x0D, 0x0E, 0x0F,
+        0x10, 0x11, 0x12, 0x13, 0x14, 0x15, 0x16, 0x17,
+        0x18, 0x19, 0x1A, 0x1B, 0x1C, 0x1D, 0x1E, 0x1F,
+        0x20, 0x21, 0x22, 0x23, 0x24, 0x25, 0x26, 0x27,
+        0x28, 0x29, 0x2A, 0x2B, 0x2C, 0x2D, 0x2E, 0x2F,
+        0x30, 0x31, 0x32, 0x33, 0x34, 0x35, 0x36, 0x37,
+        0x38, 0x39, 0x3A, 0x3B,
+        /* signerInfos SET claims 1 byte of content, buffer ends here */
+        0x31, 0x01
+    };
+    word32 derSz = (word32)sizeof(der);
+
+    ExpectNotNull(pkcs7 = wc_PKCS7_New(HEAP_HINT, testDevId));
+    ExpectIntEQ(wc_PKCS7_Init(pkcs7, HEAP_HINT, INVALID_DEVID), 0);
+    ExpectIntEQ(wc_PKCS7_InitWithCert(pkcs7, NULL, 0), 0);
+    ExpectIntNE(wc_PKCS7_VerifySignedData(pkcs7, der, derSz), 0);
+    wc_PKCS7_Free(pkcs7);
+
+#endif /* HAVE_PKCS7 */
+    return EXPECT_RESULT();
+}
+
+/*
+ * Same shape again, but the signerInfos SET is missing entirely: the
+ * buffer ends right after the content, with nothing following. signerInfos
+ * is a mandatory field, so this must be rejected rather than accepted as
+ * a bundle with no more elements.
+ */
+int test_wc_PKCS7_VerifySignedData_NoSignerInfosTag(void)
+{
+    EXPECT_DECLS;
+#if defined(HAVE_PKCS7)
+    PKCS7* pkcs7 = NULL;
+
+    WOLFSSL_SMALL_STACK_STATIC byte der[] = {
+        0x30, 0x61,
+        0x06, 0x09, 0x2A, 0x86, 0x48, 0x86, 0xF7, 0x0D, 0x01, 0x07, 0x02,
+        0xA0, 0x54,
+        0x30, 0x52,
+        0x02, 0x01, 0x01,
+        0x31, 0x00,
+        0x30, 0x4B,
+        0x06, 0x09, 0x2A, 0x86, 0x48, 0x86, 0xF7, 0x0D, 0x01, 0x07, 0x01,
+        0xA0, 0x3E,
+        0x04, 0x3C,
+        0x00, 0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07,
+        0x08, 0x09, 0x0A, 0x0B, 0x0C, 0x0D, 0x0E, 0x0F,
+        0x10, 0x11, 0x12, 0x13, 0x14, 0x15, 0x16, 0x17,
+        0x18, 0x19, 0x1A, 0x1B, 0x1C, 0x1D, 0x1E, 0x1F,
+        0x20, 0x21, 0x22, 0x23, 0x24, 0x25, 0x26, 0x27,
+        0x28, 0x29, 0x2A, 0x2B, 0x2C, 0x2D, 0x2E, 0x2F,
+        0x30, 0x31, 0x32, 0x33, 0x34, 0x35, 0x36, 0x37,
+        0x38, 0x39, 0x3A, 0x3B
+        /* buffer ends here -- no signerInfos SET at all */
+    };
+    word32 derSz = (word32)sizeof(der);
+
+    ExpectNotNull(pkcs7 = wc_PKCS7_New(HEAP_HINT, testDevId));
+    ExpectIntEQ(wc_PKCS7_Init(pkcs7, HEAP_HINT, INVALID_DEVID), 0);
+    ExpectIntEQ(wc_PKCS7_InitWithCert(pkcs7, NULL, 0), 0);
+    ExpectIntNE(wc_PKCS7_VerifySignedData(pkcs7, der, derSz), 0);
+    wc_PKCS7_Free(pkcs7);
+
+#endif /* HAVE_PKCS7 */
     return EXPECT_RESULT();
 }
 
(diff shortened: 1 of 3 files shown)


wolfcrypt/src/pkcs7.c
```diff
@@ -7506,26 +7506,23 @@ static int PKCS7_VerifySignedData(wc_PKCS7* pkcs7, const byte* hashBuf,
...
-                break;
-            }
-            /* expect data length to be enough to check set and seq of certs */
             pkcs7->stream->expected = (ASN_TAG_SZ + MAX_LENGTH_SZ) * 2;
+            if (pkcs7->stream->totalRd < pkcs7->stream->maxLen &&
+                    pkcs7->stream->expected >
+                    pkcs7->stream->maxLen - pkcs7->stream->totalRd) {
+                pkcs7->stream->expected =
+                    pkcs7->stream->maxLen - pkcs7->stream->totalRd;
+            }
```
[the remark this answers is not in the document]
miyazakh: Confirmed. Reproduced with `NoSignerInfosTag` (bundle truncated exactly at the outer SignedData length, no signerInfos at all): both single-shot and streaming calls returned `WC_PKCS7_WANT_READ_E` (-270) instead of a deterministic error, and streaming consumed all available bytes without resolving.

Fixed by capping `pkcs7->stream->expected` to 0 when `totalRd >= maxLen`, instead of skipping the cap in that case, so stage 4's existing `idx >= pkiMsg2Sz` bounds check now fires immediately and returns `BUFFER_E`.

Also strengthened `test_wc_PKCS7_VerifySignedData_NoSignerInfosTag` to assert the return code is not `WC_PKCS7_WANT_READ_E` (the prior `ExpectIntNE(ret, 0)` check passed even on the stall), and added a streaming byte-at-a-time variant of the same case.

Fixed in 8777c2b47.

tests/api/test_pkcs7.c
```diff
@@ -6337,7 +6333,280 @@ int test_wc_PKCS7_VerifySignedData_TruncCertSetTag(void)
...
+        0x38, 0x39, 0x3A, 0x3B,
+        /* signerInfos SET claims 1 byte of content, buffer ends here */
+        0x31, 0x01
+    };
+    word32 derSz = (word32)sizeof(der);
+
+    ExpectNotNull(pkcs7 = wc_PKCS7_New(HEAP_HINT, testDevId));
+    ExpectIntEQ(wc_PKCS7_Init(pkcs7, HEAP_HINT, INVALID_DEVID), 0);
+    ExpectIntEQ(wc_PKCS7_InitWithCert(pkcs7, NULL, 0), 0);
+    ExpectIntNE(wc_PKCS7_VerifySignedData(pkcs7, der, derSz), 0);
```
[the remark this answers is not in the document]
miyazakh: Confirmed — good catch. `ExpectIntNE(ret, 0)` was passing trivially since single-shot on this truncated buffer returns `WC_PKCS7_WANT_READ_E` (-270), not a real rejection. Added the missing `ExpectIntNE(ret, WC_NO_ERR_TRACE(WC_PKCS7_WANT_READ_E))`, matching NoSignerInfosTag.

That assertion then exposed an actual bug: the STAGE6→STAGE7 transition in `PKCS7_VerifySignedData` (pkcs7.c) requested the signerInfos content length without capping it against the bundle's remaining declared length, so a truncated SET header stalled forever instead of failing. Fixed by adding the same totalRd >= maxLen zero-cap guard STAGE3 already uses. Verified with full `tests/unit.test` (0 failures) and testwolfcrypt to rule out regressions in the encode/decode streaming paths.

miyazakh:
retest this please

wolfcrypt/src/pkcs7.c
```diff
@@ -7506,26 +7506,33 @@ static int PKCS7_VerifySignedData(wc_PKCS7* pkcs7, const byte* hashBuf,
...
-             * to pkcs7->content and hash to pkcs7->hash.
-             *
-             * NOTE: this check returns success whenever fewer than 6 bytes
-             * follow the content within the outer ContentInfo, which also
-             * accepts truncated bundles whose footer was cut short (e.g. a
-             * lone certificates [0] tag with no length). Distinguishing a
-             * legitimate degenerate end (such as an empty signerInfos SET
-             * "31 00") from truncated junk would require peeking at the
-             * remaining bytes or making stage 4's `expected` window smaller.
+            /* expect data length to be enough to check set and seq of certs,
```
dgarske, line 7516: **🟣 [Critical] Stage 3 fall-through breaks valid degenerate bundles at nearly every chunk size (permanent WC_PKCS7_WANT_READ_E stall)**

The removed stage-3 early return caused short-footer bundles to never enter stages 4-7. The new code always falls through into stage 4 with `pkcs7->stream->expected` capped to exactly the remaining byte count and zero slack. Stages 4/5/6 cannot run correctly with that zero-slack window: stage 4's `wc_PKCS7_StreamEndCase(pkcs7, &stateIdx, &idx)` (line 7746) over-counts `totalRd` by the caller's chunk offset (`stateIdx` is still 0 because `wc_PKCS7_HandleOctetStrings` exits through the buffered branch of `StreamEndCase` at line 502, which never writes `*tmpIdx`). The stage 4 cap (line 7755) and stage 6 cap (line 8121) then either under-size `expected` by that amount (`maxIdx` one byte short, `GetSet_ex` at line 8085 returns `ASN_PARSE_E`) or, once `totalRd` has passed `maxLen`, the unguarded `maxLen - totalRd` subtraction wraps to `0xFFFFFFFF`, no cap applies, and `expected` stays larger than the bytes that will ever arrive so `wc_PKCS7_AddDataToStream` returns `WC_PKCS7_WANT_READ_E` forever.

I measured this on a clean checkout of this PR (`472ec6258`) against a build of the merge-base, both `./configure --enable-pkcs7 --enable-opensslextra`, feeding **the PR's own `test_wc_PKCS7_VerifySignedData_DegenerateMinimal` DER vector** through `wc_PKCS7_VerifySignedData` at every chunk size 1..101:

```
base (master): FAILED 0 of 101 chunk sizes
this PR:       FAILED 86 of 101 chunk sizes
  chunk=6  -> -270 (WC_PKCS7_WANT_READ_E)
  chunk=7  -> -140 (ASN_PARSE_E)
  chunk=8  -> -270
  chunk=12..96 -> -270 (mostly)
  chunk=97 -> -132 (BUFFER_E)
  chunk=98 -> -140
```

Only chunk=1 and…

**Fix:** Do not merge as-is. The PR's own DegenerateMinimal vector regresses from 'succeeds at every chunk size' to 'fails at 86 of 101'. Fix the stage 3/4 interaction, then extend the new tests to sweep chunk sizes 1..derSz instead of only chunk=1 and single-shot.

wolfcrypt/src/pkcs7.c
```diff
@@ -7506,26 +7506,33 @@ static int PKCS7_VerifySignedData(wc_PKCS7* pkcs7, const byte* hashBuf,
...
              */
-            if (ret == 0 && pkcs7->stream->maxLen > 0 &&
-                    (pkcs7->stream->maxLen - pkcs7->stream->totalRd)
-                                                < ASN_TAG_SZ + MAX_LENGTH_SZ) {
-
-                ret = 0;
-                break;
-            }
-            /* expect data length to be enough to check set and seq of certs */
             pkcs7->stream->expected = (ASN_TAG_SZ + MAX_LENGTH_SZ) * 2;
```
dgarske, line 7526: **🔴 [High] Stage 3 'bytes remaining' cap omits + stream-length, diverging chunked from single-shot results on valid input**

The new cap computes remaining bytes as `(stream->maxLen - stream->totalRd)`. Everywhere else in this file the same quantity is written `(stream->maxLen - stream->totalRd) + stream->length` -- see lines 7755-7758, 7991-7994, 8121-8122, 12909, 16311. The `+ stream->length` term is required: `wc_PKCS7_AddDataToStream()` charges bytes to `totalRd` the moment they are copied into the internal buffer (line 392), while `wc_PKCS7_StreamEndCase()` only shrinks `stream->length` when they are actually parsed (lines 496-506). So `stream->length` is exactly the count of bytes already charged to `totalRd` but not yet consumed.

`stream->length` is routinely non-zero at the end of stage 3 whenever the caller feeds the bundle in chunks: `wc_PKCS7_HandleOctetStrings` buffers a 6-byte lookahead at line 6760 before returning. When the unparsed remainder lives entirely in that buffer, `totalRd >= maxLen` is trivially true and `expected` is forced to 0 even though six perfectly good footer bytes are sitting in `stream->buffer`. Stage 4 then computes `maxIdx = idx + 0` (line 7675) and `GetLength_ex(..., maxIdx, ...)` at line 7730 fails immediately with `BUFFER_E` -> `ASN_PARSE_E`.

Measured on a clean checkout of this PR, using the DegenerateMinimal bundle with the footer `A0 00 A1 00 31 00` (empty certificates SET, empty crls SET, empty signerInfos SET -- structurally valid CMS):

```
  R  | single-shot | 1-byte chunks | 4-byte chunks
-----+-------------+---------------+--------------
   6 |           0 |          -140 |          -140   <-- MISMATCH
   7 |           0 |          -140 |…

**Fix:** Use the same remaining-bytes expression as the rest of the file, including the `+ stream->length` term, and underflow-guard the subtraction. Add a regression test that feeds a valid degenerate bundle carrying `[0]`/`[1]` SETs at chunk sizes 2..N.

File: tests/api/test_pkcs7.c [hunk 5]
```diff
@@ -6337,7 +6333,282 @@ int test_wc_PKCS7_VerifySignedData_TruncCertSetTag(void)
...
```
dgarske: **🔴 [High] TruncSignerInfosTag asserts on a double-call artifact, masking that the PR's stated goal is not met for this vector**

The test calls `wc_PKCS7_VerifySignedData(pkcs7, der, derSz)` twice on the SAME `pkcs7` object and asserts a different property of each return value as though they were one value. They are not. Measured on a clean checkout of this PR:

```
TruncSignerInfosTag same-object: call1=-270 call2=-141 call3=-270
  test asserts call1 != 0        -> PASS
  test asserts call2 != -270     -> PASS
  TRUE single-call result is -270 (WC_PKCS7_WANT_READ_E -- stalls, NOT a parse error)
```

The oscillation `-270 / -141 / -270` proves call2 is a stateful artifact. `PKCS7_VerifySignedData` only resets the parser at wolfcrypt/src/pkcs7.c:8273 under `if (ret != 0 && ret != WC_NO_ERR_TRACE(WC_PKCS7_WANT_READ_E))`. Because call1 returns `WANT_READ_E` the reset is deliberately skipped, so call2 does not re-parse -- `wc_PKCS7_AddDataToStream` appends the 101 bytes of `der` to the 101 already buffered and the parser chokes on the second copy's version INTEGER. `-141` (`ASN_VERSION_E`) says nothing about whether the truncated `signerInfos` SET is rejected.

Consequences: (1) written idiomatically -- save `ret` once and assert twice, the pattern used at lines 5148-5155 and 5180-5190 of this same file -- **the test would FAIL**, because `ret == -270 == WC_PKCS7_WANT_READ_E`. (2) It contradicts the PR's own sibling test at line 6516-6517, whose comment states the contract as 'must fail with a real parse error, not WANT_READ_E (no more bytes will ever arrive per the outer length)'. (3) Commit `8777c2b47` explicitly states the goal is 'a deterministic parse error instead of WC_PKCS7_WANT_READ_E' -- for…

**Fix:** Collapse to one call and assert twice on the saved value. That will expose that this bundle currently returns `WC_PKCS7_WANT_READ_E`. Then either extend the pkcs7.c fix so 'signerInfos SET length exceeds the remaining outer length' yields a real parse error (consistent with NoSignerInfosTag), or, if WANT_READ_E is the intended contract, drop the second assertion and correct the comment at lines 6429-6434. Never leave a second call on the same object -- it is not a repeat of the first parse.

tests/api/test_pkcs7.c
```diff
@@ -6337,7 +6333,282 @@ int test_wc_PKCS7_VerifySignedData_TruncCertSetTag(void)
...
+    ExpectIntEQ(wc_PKCS7_InitWithCert(pkcs7, NULL, 0), 0);
+    ExpectIntEQ(wc_PKCS7_VerifySignedData(pkcs7, der, derSz), 0);
+    if (pkcs7 != NULL) {
+        ExpectIntEQ(pkcs7->contentSz, 60);
+        ExpectNotNull(pkcs7->content);
+    }
+    wc_PKCS7_Free(pkcs7);
+    pkcs7 = NULL;
+
+#ifndef NO_PKCS7_STREAM
```
dgarske, line 6468: **🟠 [Medium] New tests only cover chunk=1 and single-shot, which is why the stage 3 regression passed CI**

The streaming sub-tests feed the bundle one byte at a time only. That is the single most forgiving chunking for this code path, because `wc_PKCS7_AddDataToStream` always takes the buffered path and never mixes buffered and direct-from-input parsing. The PR rewrites `pkcs7->stream->expected` arithmetic that is chunk-size dependent by construction, so a one-byte loop plus a single-shot call cannot validate it. Concretely, the DegenerateMinimal vector passes at chunk=1 and single-shot but fails at 86 of the other 99 chunk sizes on this PR (see the BLOCK finding above); the two covered cases are exactly the two that work.

**Fix:** Add a chunk-size sweep helper for the streaming sub-tests. Any fix to the stage 3 arithmetic must be validated across chunk sizes, not just chunk=1.

tests/api/test_pkcs7.c
```diff
@@ -6337,7 +6333,282 @@ int test_wc_PKCS7_VerifySignedData_TruncCertSetTag(void)
...
+        0x38, 0x39, 0x3A, 0x3B,
+        /* signerInfos SET (empty -- genuinely degenerate) */
+        0x31, 0x00
+    };
+    word32 derSz = (word32)sizeof(der);
+
+    ExpectNotNull(pkcs7 = wc_PKCS7_New(HEAP_HINT, testDevId));
+    ExpectIntEQ(wc_PKCS7_Init(pkcs7, HEAP_HINT, INVALID_DEVID), 0);
+    ExpectIntEQ(wc_PKCS7_InitWithCert(pkcs7, NULL, 0), 0);
+    wc_PKCS7_AllowDegenerate(pkcs7, 0);
```
dgarske, line 6730: **🟠 [Medium] No regression test that noDegenerate=1 still ACCEPTS a bundle with a real signer**

This PR widens the noDegenerate rejection in `wc_PKCS7_ParseSignerInfo` (wolfcrypt/src/pkcs7.c:6394) from `inSz == 0` to `inSz == 0 || degenerate == 1`. That adds a new way for a caller who set `wc_PKCS7_AllowDegenerate(pkcs7, 0)` to get `PKCS7_NO_SIGNER_E`, so the obvious failure mode is a false positive: a good signed bundle rejected because `degenerate` was set for some other reason. The new test only covers the positive-rejection direction.

Grepping the repo, `wc_PKCS7_AllowDegenerate(pkcs7, 0)` appears in exactly three places -- tests/api/test_pkcs7.c:5168, :5178 and the new :6606 -- and all three feed a degenerate bundle and expect `PKCS7_NO_SIGNER_E`. `wolfcrypt/test/test.c:70677` only uses `AllowDegenerate(pkcs7, 1)`. Nothing anywhere asserts that noDegenerate=1 plus a real signerInfos still returns 0.

The change itself analyses as safe: `wc_PKCS7_ParseSignerInfo` has a single call site (pkcs7.c:8177), `degenerate` there is set at pkcs7.c:8093-8098 from the *signerInfos* SET length immediately before the call (`degenerate = (length == 0) ? 1 : 0`), the early digestAlgorithms heuristic at 7111-7113 is always overwritten first, and the value is strictly 0/1. Verified empirically: a real RSA signed bundle streamed at every chunk size with `AllowDegenerate(0)` still returns 0. But that behaviour has no CI signal.

**Fix:** `certs/test-ber-exp02-05-2022.p7b` (already used by `test_wc_PKCS7_BER`) returns 0 under noDegenerate=1 and works for this, or reuse the RSA bundle built by `test_wc_PKCS7_VerifySignedData_RSA`.

tests/api/test_pkcs7.c
```diff
@@ -6337,7 +6333,282 @@ int test_wc_PKCS7_VerifySignedData_TruncCertSetTag(void)
...
+    wc_PKCS7_Free(pkcs7);
+    pkcs7 = NULL;
+
+#ifndef NO_PKCS7_STREAM
+    /* same bundle fed one byte at a time */
+    ExpectNotNull(pkcs7 = wc_PKCS7_New(HEAP_HINT, testDevId));
+    ExpectIntEQ(wc_PKCS7_Init(pkcs7, HEAP_HINT, INVALID_DEVID), 0);
+    ExpectIntEQ(wc_PKCS7_InitWithCert(pkcs7, NULL, 0), 0);
+
+    ret = WC_NO_ERR_TRACE(WC_PKCS7_WANT_READ_E);
```
dgarske: **🟠 [Medium] Streaming loops never assert that all bytes were consumed, so an early success is indistinguishable from a full parse**

The loop terminates on `ret != 0`, so the moment the parser reports success the loop stops -- leaving `idx` possibly well short of `derSz` -- and the following `ExpectIntEQ(ret, 0)` passes regardless. Nothing checks `idx == derSz`.

This is not hypothetical: running the identical loop over the DegenerateMinimal blob against the *pre-fix* pkcs7.c returns 0 after consuming only 99 of 101 bytes (the buggy stage 3 early-exit accepted the bundle before the trailing `31 00` signerInfos SET was ever fed) and the assertion still passes. The same blind spot applies to the `NoSignerInfosTag` loop at 6535.

On the other aspects of this loop shape: the inner `if (ret < 0 && ret != WANT_READ_E) break;` is NOT redundant (a hard error is non-zero, so the loop condition alone would spin re-feeding bytes), but having two exit conditions both expressing 'stop' is confusing; `ret` is `int` and the `WC_NO_ERR_TRACE(...)` comparisons are type-correct; and if the loop exhausts the buffer still on WANT_READ_E, `ExpectIntEQ(ret, 0)` does fail with a readable `-270 != 0`. The initialisation `ret = WC_NO_ERR_TRACE(WC_PKCS7_WANT_READ_E)` purely to enter the loop also differs from this file's existing idiom of `ret = -1` (lines 5148, 5181).

**Fix:** Add a consumption assertion after each streaming loop so a premature success is caught. Optionally simplify to a single exit condition: `for (idx = 0; idx < derSz; idx++) { ret = ...; if (ret != WC_NO_ERR_TRACE(WC_PKCS7_WANT_READ_E)) break; }`.

tests/api/test_pkcs7.c
```diff
@@ -6337,7 +6333,282 @@ int test_wc_PKCS7_VerifySignedData_TruncCertSetTag(void)
...
+        /* buffer ends here -- no signerInfos SET at all */
+    };
+    word32 derSz = (word32)sizeof(der);
+
+    /* single-shot: must fail with a real parse error, not WANT_READ_E
+     * (no more bytes will ever arrive per the outer length) */
+    ExpectNotNull(pkcs7 = wc_PKCS7_New(HEAP_HINT, testDevId));
+    ExpectIntEQ(wc_PKCS7_Init(pkcs7, HEAP_HINT, INVALID_DEVID), 0);
+    ExpectIntEQ(wc_PKCS7_InitWithCert(pkcs7, NULL, 0), 0);
+    ExpectIntNE(wc_PKCS7_VerifySignedData(pkcs7, der, derSz), 0);
```
dgarske: **🔵 [Low] NoSignerInfosTag repeats the double-call idiom; currently self-consistent but fragile**

Same anti-pattern as `TruncSignerInfosTag`: two `wc_PKCS7_VerifySignedData` calls on one `pkcs7` object, each asserted for a different property of 'the' return value. Here it happens to work today -- measured call1 = -132 (`BUFFER_E`) and call2 = -132, because a hard error does trigger `wc_PKCS7_ResetStream` + `wc_PKCS7_ChangeState(WC_PKCS7_START)` at wolfcrypt/src/pkcs7.c:8273-8278, so the second call genuinely re-parses. But the equivalence is incidental: the moment the first call returns `WC_PKCS7_WANT_READ_E` -- exactly the regression this test guards against -- the reset is skipped and the second call becomes a doubled-input parse that silently satisfies the assertion for the wrong reason. The test also does double the work for no benefit, and reuses the object post-failure without re-init, which the API does not document as supported.

**Fix:** Save the return once and assert twice, as the streaming half of this very same test already does at lines 6541-6542. The `int ret;` declaration would need to move out of the `#ifndef NO_PKCS7_STREAM` guard since it is then used in both modes.

wolfcrypt/src/pkcs7.c
```diff
@@ -8122,6 +8129,18 @@ static int PKCS7_VerifySignedData(wc_PKCS7* pkcs7, const byte* hashBuf,
             }
             else  {
                 pkcs7->stream->expected = (word32)length;
+                if (pkcs7->stream->totalRd >= pkcs7->stream->maxLen) {
```
dgarske: **🔵 [Low] Stage 6 zero-cap comment describes a stage 7 bounds check that does not exist**

The new zero-cap states it will 'force stage 7's bounds check to fail now'. `WC_PKCS7_VERIFY_STAGE7` (lines 8150-8178) has no bounds check on `pkcs7->stream->expected`: unlike stage 6, which does `maxIdx = idx + pkcs7->stream->expected; if (maxIdx > pkiMsg2Sz) ret = BUFFER_E;` (lines 8029-8033), stage 7 goes straight from `wc_PKCS7_AddDataToStream()` to `wc_PKCS7_ParseSignerInfo()`. With `expected == 0`, `wc_PKCS7_AddDataToStream()` always SUCCEEDS -- if `stream->length == 0` the first branch matches (`inSz - rdSz >= 0` is trivially true) and returns the caller's whole `srcSz`; if `stream->length > 0` the second branch matches (`stream->length >= 0`) and returns the internal buffer. Nothing fails; `wc_PKCS7_ParseSignerInfo()` is simply handed whatever window is available and errors out later.

The guard itself is functionally sound. Instrumented runs over the full `tests/unit.test` corpus (99 stage-6 traversals) plus ~4000 chunked-stream permutations show the cap fires only when `length == 0`, where `expected` was already 0 -- a no-op on every well-formed input. The detached `_ex` path cannot fire because lines 7126-7128 inflate `maxLen` by `pkiMsg2Sz + contentSz`; indefinite BER takes the other branch at 8127. Unlike the stage 3 hunk, this one is safe -- but the rationale in the comment is wrong, and it is the only remaining-bytes test in the file that ignores `stream->length`.

**Fix:** Fix the comment. If a hard failure is really wanted here, set the error explicitly (`ret = BUFFER_E; break;`) rather than relying on a downstream parse error. For consistency with lines 7755, 7991, 8121 and 12909, express the remaining-bytes test as `((maxLen - totalRd) + stream->length)`.

tests/api/test_pkcs7.c
```diff
@@ -6337,7 +6333,282 @@ int test_wc_PKCS7_VerifySignedData_TruncCertSetTag(void)
...
+        /* signerInfos SET (empty -- genuinely degenerate) */
+        0x31, 0x00
+    };
+    word32 derSz = (word32)sizeof(der);
+
+    ExpectNotNull(pkcs7 = wc_PKCS7_New(HEAP_HINT, testDevId));
+    ExpectIntEQ(wc_PKCS7_Init(pkcs7, HEAP_HINT, INVALID_DEVID), 0);
+    ExpectIntEQ(wc_PKCS7_InitWithCert(pkcs7, NULL, 0), 0);
+    wc_PKCS7_AllowDegenerate(pkcs7, 0);
+    ExpectIntEQ(wc_PKCS7_VerifySignedData(pkcs7, der, derSz),
```
dgarske, line 6731: **🔵 [Low] Bare PKCS7_NO_SIGNER_E instead of WC_NO_ERR_TRACE(PKCS7_NO_SIGNER_E)**

tests/api/test_pkcs7.c uses `WC_NO_ERR_TRACE(...)` 147 times when comparing against wolfCrypt error codes, including the two other `PKCS7_NO_SIGNER_E` comparisons in the same file at lines 5170 and 5190. Line 6608 is the only bare error constant in an `Expect*` comparison introduced by this diff.

This is not a build break: wolfssl/wolfcrypt/error-crypt.h:377-380 gates the tracing redefinition on `defined(WOLFSSL_DEBUG_TRACE_ERROR_CODES) && (defined(BUILDING_WOLFSSL) || defined(WOLFSSL_DEBUG_TRACE_ERROR_CODES_ALWAYS))`, and `-DBUILDING_WOLFSSL` is only added to `src_libwolfssl_la_CFLAGS`, not to the test objects, so under a plain `--enable-debug-trace-errcodes` the test TU still sees the plain enumerator. Under `WOLFSSL_DEBUG_TRACE_ERROR_CODES_ALWAYS` the bare label expands to `WC_ERR_TRACE(...)` and would emit a spurious `ERR TRACE: tests/api/test_pkcs7.c L 6608 PKCS7_NO_SIGNER_E (-269)` line from inside the harness -- exactly the noise `WC_NO_ERR_TRACE` exists to prevent.

**Fix:** Wrap it to match lines 5170 and 5190.

dgarske (commented):
## Skoll Code Review

**9 finding(s) posted as inline comments** (see file-level comments below)

### Posted findings
- [Critical] **Stage 3 fall-through breaks valid degenerate bundles at nearly every chunk size (permanent WC_PKCS7_WANT_READ_E stall)** — `wolfcrypt/src/pkcs7.c:7509-7535`
- [High] **Stage 3 'bytes remaining' cap omits + stream-length, diverging chunked from single-shot results on valid input** — `wolfcrypt/src/pkcs7.c:7519-7535`
- [High] **TruncSignerInfosTag asserts on a double-call artifact, masking that the PR's stated goal is not met for this vector** — `tests/api/test_pkcs7.c:6468-6470`
- [Medium] **New tests only cover chunk=1 and single-shot, which is why the stage 3 regression passed CI** — `tests/api/test_pkcs7.c:6404-6423`
- [Medium] **No regression test that noDegenerate=1 still ACCEPTS a bundle with a real signer** — `tests/api/test_pkcs7.c:6606-6608`
- [Medium] **Streaming loops never assert that all bytes were consumed, so an early success is indistinguishable from a full parse** — `tests/api/test_pkcs7.c:6410-6417`
- [Low] **NoSignerInfosTag repeats the double-call idiom; currently self-consistent but fragile** — `tests/api/test_pkcs7.c:6521-6523`
- [Low] **Stage 6 zero-cap comment describes a stage 7 bounds check that does not exist** — `wolfcrypt/src/pkcs7.c:8132-8143`
- [Low] **Bare PKCS7_NO_SIGNER_E instead of WC_NO_ERR_TRACE(PKCS7_NO_SIGNER_E)** — `tests/api/test_pkcs7.c:6607-6608`

#### TruncSignerInfosTag and DegenerateNonEmptyDigestAlgos have no streaming sub-test, unlike their siblings
**File:** `tests/api/test_pkcs7.c:6435,6559`
**Function:** `test_wc_PKCS7_VerifySignedData_TruncSignerInfosTag`
**Severity:** Medium

The PR's justification for these tests is the streaming stage 3/4/7 rework, yet only two of the four new tests exercise the streaming path. `DegenerateMinimal` (6404-6423) and `NoSignerInfosTag` (6527-6543) add a `#ifndef NO_PKCS7_STREAM` one-byte-at-a-time loop; `TruncSignerInfosTag` (6435) and `DegenerateNonEmptyDigestAlgos` (6559) do not.

The single-shot path is not equivalent: for `TruncSignerInfosTag` the single-shot call returns `-270` (WANT_READ_E) while the byte-at-a-time loop returns `-141` after consuming 100/101 bytes -- two entirely different code paths with different outcomes, and only the weaker one is covered. Both missing loops work when added (measured: TruncSignerInfosTag bytewise -> -141; DegenerateNonEmptyDigestAlgos bytewise -> -269 `PKCS7_NO_SIGNER_E`).

Related gaps: none of the four new blobs carries a certificates `[0]` SET or a CRLs `[1]` SET, so the stage 3 -> 4 transition with real optional footers is untested by any new test -- which is exactly the shape that breaks (see BLOCK finding 2). No new test covers an indefinite-length (BER) bundle even though this PR added an explicit `indefLen` carve-out at stage 3 in commit `59c4c7087`; that path is only covered incidentally by the pre-existing `test_wc_PKCS7_BER`.

**Recommendation:** Also add (a) a degenerate bundle carrying a real certificates `[0]` SET and a CRLs `[1]` SET, and (b) an indefinite-length/BER variant, to pin the `indefLen` carve-out this PR introduces.

_Referenced code: `tests/api/test_pkcs7.c:6435,6559` (8 lines)_

---

dgarske (changes requested):
https://github.com/wolfSSL/wolfssl/pull/11017#pullrequestreview-4870852451

wolfcrypt/src/pkcs7.c
miyazakh: Commit `8777c2b47 `caps `stream->expected` to 0 when totalRd >= maxLen so stage 4's bounds check fires deterministically instead of looping on `WC_PKCS7_WANT_READ_E`, and 59c4c7087 restricts that cap to definite-length bundles only (it was also firing on BER indefinite-length streams, where maxLen is a running estimate, not a hard bound — that had broken `EncodeSignedData/VerifySignedData_RSA/BER`).

The `DegenerateMinimal `test now sweeps chunk sizes 2..derSz (not just chunk=1/single-shot) per your suggestion, and I re-ran it locally: 0/101 failures across all chunk sizes, full` --enable-all` unit suite passes (1746/1746, 0 failures). Verified against the current branch head (59c4c7087).

wolfcrypt/src/pkcs7.c
miyazakh: Stage 3 now computes the cap as `(maxLen - totalRd) + stream->length`, matching the expression used at stages 4/6 and lines 12919/13026/13324/13499, instead of omitting the `+ length` term. Two related fixes went in alongside it:

`wc_PKCS7_HandleOctetStrings `now re-syncs `*tmpIdx = *idx` after `wc_PKCS7_StreamEndCase `returns through the buffered branch, so bytes charged to `totalRd `during buffering aren't double-charged once the loop reads directly from in.
Stage 4 and stage 6 now re-sync `stateIdx = idx` for the same reason, so `totalRd `doesn't carry a stale offset from stage 3's internal buffer.

tests/api/test_pkcs7.c
miyazakh: Fixed. `TruncSignerInfosTag `now saves the return value once and asserts both `!= 0` and `!= WC_PKCS7_WANT_READ_E`. On the pkcs7.c side, stage 6 now checks that the claimed signerInfos SET content length actually fits within what can still arrive (using `totalRd/maxLen`, adjusted for bytes already consumed within the current stage before the `stateIdx `resync), and fails immediately with `ASN_PARSE_E `if not — instead of falling through to stage 7 and stalling on `WC_PKCS7_WANT_READ_E `forever. Skipped for indefinite-length (BER) bundles since `maxLen `there is only a running estimate, consistent with the existing stage 3/4/6 caps.

tests/api/test_pkcs7.c
miyazakh: Added a `test_wc_PKCS7_VerifySignedData_ChunkSweep_once `helper (follows the existing _once + EXPECT_TEST() composition pattern already used in [`test_asn.c`]) that feeds a DER buffer at every chunk size from a given minimum up to `derSz`, asserting either a consistent success with a given content size, or a consistent real parse error that's never `WC_PKCS7_WANT_READ_E`.

tests/api/test_pkcs7.c
miyazakh: Added `test_wc_PKCS7_VerifySignedData_NoDegenerateAcceptsRealSigner`: builds a real RSA-signed bundle via the existing `CreatePKCS7SignedData `helper, calls `wc_PKCS7_AllowDegenerate(pkcs7, 0)`, and asserts `wc_PKCS7_VerifySignedData `still returns 0.

tests/api/test_pkcs7.c
miyazakh: Added a fed tracker to `ChunkSweep_once`: it now records the byte offset actually fed into `wc_PKCS7_VerifySignedData `at the moment ret stops being WC_PKCS7_WANT_READ_E, and for the success path asserts `fed == derSz` before checking `ret == 0`. A premature accept that consumed less than the full bundle now fails the sweep instead of silently passing.

tests/api/test_pkcs7.c
miyazakh: fixed. `NoSignerInfosTag's `single-shot check now saves the return value once (ret = `wc_PKCS7_VerifySignedData(...)`) and `asserts != 0` and `!= WC_PKCS7_WANT_READ_E` against that same value, matching the streaming half's existing pattern and the fix already applied to `TruncSignerInfosTag`. int ret; moved out of the `#ifndef NO_PKCS7_STREAM` guard since it's now used unconditionally.

wolfcrypt/src/pkcs7.c
miyazakh: Addressed

tests/api/test_pkcs7.c
miyazakh: Wrapped it

JacobBarthelmeh (approved):

dgarske (approved):

PR #11018, opened 2026-07-30
Security and correctness fixes, plus a DTLS 1.3 scheduled work API

Frauschi:
## Summary

Ten of the twelve commits fix distinct Fenrir-found memory-safety, key-hygiene, and timing issues in wolfSSL and wolfCrypt, each one independent and reviewable on its own. The last two add a small DTLS 1.3 API and test coverage for a protocol-level gap that only appears under `WOLFSSL_RW_THREADED`.

## Fixes

| Finding | Area | Issue |
|---|---|---|
| F-7240 | `src/internal.c` | `TimingPadVerify` passed `ssl->hmac` a length that underflows when the padding byte exceeds the record, so an out-of-range byte skipped the MAC entirely while a valid one paid for a full HMAC. That is a Lucky13-style timing oracle. Now clamped in constant time; the rejection decision is unchanged. |
| F-7223 | `src/x509.c` | `X509PrintDirType` measured a raw-DER `directoryName` with `XSTRLEN`. A measured length below five wrapped an unsigned loop bound to near `UINT32_MAX` and the tag scan ran far past the entry. Reached from ordinary certificate printing. Now uses `entry->len`. |
| F-7248 | `src/x509.c` | `wolfSSL_X509_check_host` accepts a length-delimited buffer, then handed it to `CheckIPAddr`, which discards the length and re-measures with `XSTRLEN`, reading past the end on every call. Now calls `CheckHostName` with the caller's length, which also fixes a missed `iPAddress` match. |
| F-7084 | `wolfcrypt/src/srp.c` | On a small-stack allocation failure, cleanup gated zeroization on the return code rather than on initialization, so it called `mp_forcezero` on uninitialized memory and took the length from an unset size field. That is a write of arbitrary length past the allocation. Now gated on an init flag. |
| F-7083 | `wolfcrypt/src/hpke.c` | Encap and decap install a temporary RNG into a caller-owned key and free it without removing it, leaving the key pointing at freed memory for any later use that consults it. ECH holds exactly such a long-lived key. The previous RNG is now saved and restored. |
| F-7258 | `src/ssl.c` | `wolfSSL_clear` left the previous connection's key material in a recycled object: the `Keys` struct, the TLS 1.3 traffic secrets, the DTLS 1.3 epoch keys, the handshake arrays and the tls-unique binding. All now force-zeroed, with the arrays wiped in place so the post-connection accessors keep working. |
| F-7256, F-7257 | `src/ssl_crypto.c`, `src/ssl_api_ext.c` | `WOLFSSL_HMAC_CTX` keeps the HMAC pads outside the wolfCrypt object, and for any key up to the block size the key falls out of a single exclusive or. Cleanup never wiped them, so `HMAC_CTX_free` handed the key back to the allocator. The session ticket key callback leaked the ticket HMAC key onto the stack the same way, on every encrypt and decrypt. |
| F-7082 | `wolfcrypt/src/ecc.c` | `wc_ecc_ctx_reset` cleared the heap hint `wc_ecc_ctx_new_ex` had just recorded, so the context was freed to the default allocator rather than the heap it came from. Under static memory that is a free of memory never allocated there. |
| F-7085 | `wolfcrypt/src/rng_bank.c` | `wc_rng_bank_init` personalized each DRBG from `&ctx->rngs[i]`, hashing the just-cleared struct instead of the instance address, so every instance in the bank got the same all-zero string. Now matches `wc_rng_bank_inst_reinit`. |
| F-7222 | `src/ssl_api_dtls.c`, `src/dtls.c`, `src/internal.c` | `wolfSSL_dtls_set_pending_peer` mutated shared state under a read lock, which admits several holders, so two threads could both free `pendingPeer.sa` while the send path read it. Now takes the write lock. `DtlsResetState` took no lock at all and now holds it too. |

## DTLS 1.3 scheduled work

Under `WOLFSSL_RW_THREADED` the read path performs no scheduled work, because transmitting from the reader would race the write thread over the output buffer and the sending key schedule, neither of which is covered by a lock. Post-handshake the only remaining consumers sit on the write path, and `wolfSSL_dtls_retransmit()` only helps while the handshake is unfinished. An application that reads without writing therefore never acknowledges a NewSessionTicket, KeyUpdate or connection ID message, and the peer keeps retransmitting. RFC 9147 depends on those ACKs, so this is a protocol break rather than a missed optimisation.

This adds two entry points so such an application can drive that work from its write thread:

- `wolfSSL_dtls13_pending_work()` reports whether anything is waiting.
- `wolfSSL_dtls13_do_scheduled_work()` sends it.

Both ask the same helpers rather than each testing conditions of their own, so the predicate cannot drift into promising work the pump then declines, which would leave a drain loop spinning. The predicate errs towards reporting work when it cannot tell, so a loop surfaces an error rather than stopping silently. Refusing an object is treated as a usage error and deliberately leaves `ssl->error` alone: that field is sticky, and `wolfSSL_write()` skips its write-dup drain while it is set, so recording an error there would disable the very drain a write-dup application depends on.

Two things are deliberately out of scope. Write-dup pairs already park the read side's work in the shared `WriteDup` struct, which `wolfSSL_write()` reconciles, so they have a drain. Completing a key update we started ourselves needs the peer's acknowledgement processed, which rotates the sending keys and creates an epoch, and the epoch table has no locking while the read thread mutates it as well. That is a larger change and is not attempted here.

These last two commits have no finding ID behind them. They came out of investigating why `test_wolfSSL_dtls_AEAD_limit` fails under `WOLFSSL_RW_THREADED`, which it does on master as well.

### DIFF
--- a/doc/dox_comments/header_files/ssl.h
+++ b/doc/dox_comments/header_files/ssl.h
@@ -3952,6 +3952,87 @@ int  wolfSSL_dtls_got_timeout(WOLFSSL* ssl);
 int wolfSSL_dtls_retransmit(WOLFSSL* ssl);
 
 /*!
+    \ingroup Setup
+
+    \brief Sends the DTLS 1.3 work that was scheduled while reading. With
+    WOLFSSL_RW_THREADED the read path never transmits, because that would race
+    the write thread over the output buffer and the sending key schedule, so
+    ACKs, retransmissions and a KeyUpdate the peer asked for are only sent from
+    the write side. An application that reads without writing must call this,
+    from the same thread it uses for writing, or those messages are never sent
+    and the peer keeps retransmitting what it is waiting to have acknowledged.
+    Not for write-dup applications, which drain through wolfSSL_write().
+
+    A KeyUpdate is only sent while none of ours is still unacknowledged, since
+    DTLS must not have two in flight. Completing one we started needs the
+    peer's acknowledgement processed, which rotates the sending keys and so
+    does not happen here, so in a WOLFSSL_RW_THREADED build a peer request
+    arriving after that point is held rather than answered.
+
+    A send that could only write part of a record returns WOLFSSL_FATAL_ERROR
+    with wolfSSL_get_error() reporting SSL_ERROR_WANT_WRITE. That is not a
+    failure of the connection: the record is held and the next call sends the
+    rest, so wolfSSL_dtls13_pending_work() keeps reporting work until it is
+    out. Treat it as a retry rather than as a reason to stop draining.
+
+    \return WOLFSSL_SUCCESS on success, including when there was nothing to do.
+    \return WOLFSSL_FATAL_ERROR if ssl is NULL, is not a DTLS 1.3 object, is
+    part of a write-dup pair, or the send failed. Call wolfSSL_get_error() to
+    tell a retryable SSL_ERROR_WANT_WRITE from a real failure.
+
+    \param ssl a pointer to a WOLFSSL structure, created using wolfSSL_new().
+
+    _Example_
+    \code
+    WOLFSSL* ssl;
+    ...
+    while (wolfSSL_dtls13_pending_work(ssl)) {
+        if (wolfSSL_dtls13_do_scheduled_work(ssl) != WOLFSSL_SUCCESS) {
+            if (wolfSSL_get_error(ssl, 0) == SSL_ERROR_WANT_WRITE) {
+                // the socket is full, wait for it and call again
+                break;
+            }
+            // a real error
+            break;
+        }
+    }
+    \endcode
+
+    \sa wolfSSL_dtls13_pending_work
+    \sa wolfSSL_dtls_retransmit
+*/
+int wolfSSL_dtls13_do_scheduled_work(WOLFSSL* ssl);
+
+/*!
+    \ingroup Setup
+
+    \brief Reports whether the object has DTLS 1.3 work waiting to be sent by
+    wolfSSL_dtls13_do_scheduled_work(). Only meaningful with
+    WOLFSSL_RW_THREADED. The answer is advisory and can change as soon as it is
+    returned. Only work the pump can actually carry out is reported, so a drain
+    loop over the pair terminates; waiting for the peer to acknowledge a key
+    update we sent is not reported, as there is nothing to send for it. A
+    record the pump could only write in part is reported, so that the retry it
+    owes is not lost.
+
+    \return 1 if there is work to send, or if it could not be determined.
+    \return 0 if there is nothing to do, or ssl is not supported here.
+
+    \param ssl a pointer to a WOLFSSL structure, created using wolfSSL_new().
+
+    _Example_
+    \code
+    WOLFSSL* ssl;
+    ...
+    if (wolfSSL_dtls13_pending_work(ssl))
+        wolfSSL_dtls13_do_scheduled_work(ssl);
+    \endcode
+
+    \sa wolfSSL_dtls13_do_scheduled_work
+*/
+int wolfSSL_dtls13_pending_work(WOLFSSL* ssl);
+
+/*!
     \brief This function is used to determine if the SSL session has been
     configured to use DTLS.
 
--- a/src/dtls.c
+++ b/src/dtls.c
@@ -66,6 +66,11 @@
 
 void DtlsResetState(WOLFSSL* ssl)
 {
+#if defined(WOLFSSL_DTLS) && defined(WOLFSSL_DTLS_CID) && \
+    defined(WOLFSSL_RW_THREADED)
+    int locked;
+#endif
+
     /* Reset the state so that we can statelessly await the
      * ClientHello that contains the cookie. Don't gate on IsAtLeastTLSv1_3
      * to handle the edge case when the peer wants a lower version. */
@@ -98,6 +103,17 @@ void DtlsResetState(WOLFSSL* ssl)
     ssl->options.tls1_1 = 0;
     ssl->options.tls1_3 = 0;
 #if defined(WOLFSSL_DTLS) && defined(WOLFSSL_DTLS_CID)
+#ifdef WOLFSSL_RW_THREADED
+    /* wolfSSL_dtls_set_pending_peer() may run on another thread, so take the
+     * lock the record layer uses for these two fields before dropping what
+     * that call left behind. Callers must not already hold peerLock:
+     * re-acquiring a write lock is undefined and deadlocks rather than fails,
+     * so the only failure this can report is a broken or uninitialised lock.
+     * Clear the fields anyway in that case, since resetting the state is the
+     * whole point of this call and leaving a pending peer behind is worse than
+     * the race. dtlsProcessPendingPeer() and ProcessReplyEx() do the same. */
+    locked = (wc_LockRwLock_Wr(&ssl->buffers.dtlsCtx.peerLock) == 0);
+#endif
     ssl->buffers.dtlsCtx.processingPendingRecord = 0;
     /* Clear the pending peer in case user set */
     XFREE(ssl->buffers.dtlsCtx.pendingPeer.sa, ssl->heap,
@@ -105,6 +121,10 @@ void DtlsResetState(WOLFSSL* ssl)
     ssl->buffers.dtlsCtx.pendingPeer.sa = NULL;
     ssl->buffers.dtlsCtx.pendingPeer.sz = 0;
     ssl->buffers.dtlsCtx.pendingPeer.bufSz = 0;
+#ifdef WOLFSSL_RW_THREADED
+    if (locked)
+        (void)wc_UnLockRwLock(&ssl->buffers.dtlsCtx.peerLock);
+#endif
 #endif
 }
 
--- a/src/internal.c
+++ b/src/internal.c
@@ -22730,11 +22730,13 @@ int TimingPadVerify(WOLFSSL* ssl, const byte* input, int padLen, int macSz,
 
     XMEMSET(verify, 0, WC_MAX_DIGEST_SIZE);
     good = MaskPadding(input, pLen, macSz);
-    /* 4th argument has potential to underflow, ssl->hmac function should
-     * either increment the size by (macSz + padLen + 1) before use or check on
-     * the size to make sure is valid. */
-    ret = ssl->hmac(ssl, verify, input, (word32)(pLen - macSz - padLen - 1), padLen,
-                                                        content, 1, PEER_ORDER);
+    /* An out of range padding length byte is already recorded in good, but the
+     * length handed to ssl->hmac must not underflow. Clamp it in constant time
+     * so that the same amount of hashing is done for every value of the
+     * padding length byte. */
+    padLen &= ctMaskIntGTE(pLen - macSz - 1, padLen);
+    ret = ssl->hmac(ssl, verify, input, (word32)(pLen - macSz - padLen - 1),
+                                        padLen, content, 1, PEER_ORDER);
     good |= MaskMac(input, pLen, ssl->specs.hash_size, verify);
 
     /* Non-zero on failure. */
--- a/src/internal.c
+++ b/src/internal.c
@@ -23737,6 +23737,16 @@ static int dtlsRecordIsNewest(WOLFSSL* ssl)
  */
 static void dtlsProcessPendingPeer(WOLFSSL* ssl, int deprotected, int isNewest)
 {
+#ifdef WOLFSSL_RW_THREADED
+    int locked;
+
+    /* A failure here means the lock itself is broken, not that another holder
+     * is in the way, so carry on regardless: this bookkeeping is the caller's
+     * own and leaving it stale is worse than the race. Only the promotion into
+     * dtlsCtx.peer below is skipped, since EmbedSendTo reads that buffer
+     * without the lock. DtlsResetState() and ProcessReplyEx() do the same. */
+    locked = (wc_LockRwLock_Wr(&ssl->buffers.dtlsCtx.peerLock) == 0);
+#endif
     if (ssl->buffers.dtlsCtx.pendingPeer.sa != NULL) {
         if (!deprotected) {
             /* Here we have just read an entire record from the network. It is
@@ -23752,11 +23762,30 @@ static void dtlsProcessPendingPeer(WOLFSSL* ssl, int deprotected, int isNewest)
                     !ssl->buffers.dtlsCtx.processingPendingRecord;
         }
         else {
-            /* Pending peer present and record deprotected. Update the peer. */
-            if (isNewest) {
-                (void)wolfSSL_dtls_set_peer(ssl,
-                        ssl->buffers.dtlsCtx.pendingPeer.sa,
-                        ssl->buffers.dtlsCtx.pendingPeer.sz);
+            /* Pending peer present and record deprotected. Promote it here
+             * rather than through wolfSSL_dtls_set_peer, which would take this
+             * same lock again. */
+            if (isNewest
+        #ifdef WOLFSSL_RW_THREADED
+                    /* Only this touches the buffer the send path reads
+                     * without the lock, so it is the one thing a failed
+                     * acquisition has to skip. */
+                    && locked
+        #endif
+                ) {
+                WOLFSSL_SOCKADDR* from = &ssl->buffers.dtlsCtx.pendingPeer;
+
+                if (wolfssl_local_SockAddrSet(&ssl->buffers.dtlsCtx.peer,
+                        from->sa, from->sz, ssl->heap) == WOLFSSL_SUCCESS) {
+                    ssl->buffers.dtlsCtx.userSet = 1;
+                }
+                else {
+                    /* wolfSSL_dtls_set_peer clears this when it fails to store
+                     * a peer, and the receive path only checks the sender
+                     * while peer.sz is non zero, so leaving it set would drop
+                     * the check rather than fail safe. */
+                    ssl->buffers.dtlsCtx.userSet = 0;
+                }
             }
             ssl->buffers.dtlsCtx.processingPendingRecord = 0;
             dtlsClearPeer(&ssl->buffers.dtlsCtx.pendingPeer);
@@ -23765,6 +23794,10 @@ static void dtlsProcessPendingPeer(WOLFSSL* ssl, int deprotected, int isNewest)
     else {
         ssl->buffers.dtlsCtx.processingPendingRecord = 0;
     }
+#ifdef WOLFSSL_RW_THREADED
+    if (locked)
+        (void)wc_UnLockRwLock(&ssl->buffers.dtlsCtx.peerLock);
+#endif
 }
 #endif
 static int DoDecrypt(WOLFSSL *ssl)
@@ -25104,6 +25137,10 @@ int ProcessReply(WOLFSSL* ssl)
 int ProcessReplyEx(WOLFSSL* ssl, int allowSocketErr)
 {
     int ret;
+#if defined(WOLFSSL_DTLS) && defined(WOLFSSL_DTLS_CID) && \
+    defined(WOLFSSL_RW_THREADED)
+    int locked;
+#endif
 
     ret = DoProcessReplyEx(ssl, allowSocketErr);
 
@@ -25116,8 +25153,20 @@ int ProcessReplyEx(WOLFSSL* ssl, int allowSocketErr)
                 && ret != WC_NO_ERR_TRACE(WC_PENDING_E)
 #endif
             ) {
+        #ifdef WOLFSSL_RW_THREADED
+            /* Drop the pending peer even when the lock cannot be taken, as
+             * DtlsResetState() and dtlsProcessPendingPeer() do. A failure here
+             * means the lock is broken rather than held, and nothing below
+             * touches dtlsCtx.peer, which is the buffer the send path reads
+             * without this lock. */
+            locked = (wc_LockRwLock_Wr(&ssl->buffers.dtlsCtx.peerLock) == 0);
+        #endif
             dtlsClearPeer(&ssl->buffers.dtlsCtx.pendingPeer);
             ssl->buffers.dtlsCtx.processingPendingRecord = 0;
+        #ifdef WOLFSSL_RW_THREADED
+            if (locked)
+                (void)wc_UnLockRwLock(&ssl->buffers.dtlsCtx.peerLock);
+        #endif
         }
     }
 #endif
--- a/src/ssl.c
+++ b/src/ssl.c
@@ -5679,7 +5679,94 @@ size_t wolfSSL_get_client_random(const WOLFSSL* ssl, unsigned char* out,
                 ssl->buffers.inputBuffer.bufferSize);
         #endif
         }
-        ssl->keys.encryptionOn = 0;
+        /* Recycling the object for a new connection must not carry the
+         * previous connection's key material along with it. A freshly
+         * created object has all of this zeroed. */
+        ForceZero(&ssl->keys, sizeof(Keys));
+    #ifdef WOLFSSL_MULTICAST
+        if (ssl->options.haveMcast) {
+            int i;
+
+            for (i = 0; i < WOLFSSL_DTLS_PEERSEQ_SZ; i++)
+                ssl->keys.peerSeq[i].peerId = INVALID_PEER_ID;
+        }
+    #endif
+    #ifdef WOLFSSL_TLS13
+        ForceZero(ssl->clientSecret, sizeof(ssl->clientSecret));
+        ForceZero(ssl->serverSecret, sizeof(ssl->serverSecret));
+        /* A key update the previous connection asked for has nothing to say
+         * about the next one, and the keys it would rotate are gone. */
+        ssl->options.sendKeyUpdate = 0;
+    #endif
+    #ifdef WOLFSSL_HAVE_TLS_UNIQUE
+        /* The channel binding of the connection that just ended. Leaving it
+         * in place would let the next caller bind to the wrong session. */
+        ForceZero(ssl->clientFinished, TLS_FINISHED_SZ_MAX);
+        ForceZero(ssl->serverFinished, TLS_FINISHED_SZ_MAX);
+        ssl->clientFinished_len = 0;
+        ssl->serverFinished_len = 0;
+    #endif
+    #ifdef WOLFSSL_DTLS13
+        /* Per-epoch traffic keys, IVs and sequence number keys. */
+        ForceZero(ssl->dtls13Epochs, sizeof(ssl->dtls13Epochs));
+        /* Only InitSSL() sets up the unprotected epoch 0 and aims the epoch
+         * pointers at it, and this object is being reused rather than freed,
+         * so put it back or the next handshake has no valid epoch. */
+        ssl->dtls13Epochs[0].isValid = 1;
+        ssl->dtls13Epochs[0].side = ENCRYPT_AND_DECRYPT_SIDE;
+        ssl->dtls13EncryptEpoch = &ssl->dtls13Epochs[0];
+        ssl->dtls13DecryptEpoch = &ssl->dtls13Epochs[0];
+        /* The numbers that say which epoch to use sit outside the table and
+         * only ever move up, so wiping the table has to be matched here by
+         * hand. InitSSL() gets this for free from clearing the whole object.
+         * Left behind, they ask the next handshake to send under an epoch the
+         * table no longer holds, and Dtls13SetEpochKeys() fails the connection
+         * with BAD_STATE_E. */
+        w64Zero(&ssl->dtls13Epoch);
+        w64Zero(&ssl->dtls13PeerEpoch);
+        w64Zero(&ssl->dtls13InvalidateBefore);
+        ssl->dtls13WaitKeyUpdateAck = 0;
+        ssl->dtls13DoKeyUpdate = 0;
+        ssl->dtls13SendingAckOrRtx = 0;
+        /* Anything still queued for retransmission or acknowledgement is
+         * tagged with an epoch that no longer exists, so it can never be sent
+         * and would only be released when the object is freed. */
+        Dtls13FreeFsmResources(ssl);
+        ssl->dtls13Rtx.sendAcks = 0;
+        ssl->dtls13Rtx.retransmit = 0;
+    #endif
+        /* The handshake arrays hold the master and pre-master secrets. Wipe
+         * those in place rather than releasing the arrays, so that the
+         * allocation stays valid for wolfSSL_set_secret(), the exporter and
+         * the other accessors that read from it once a connection has ended.
+         * An application that asked to keep the arrays still gets back
+         * everything the API can hand it, so only the rest goes. */
+        if (ssl->arrays != NULL) {
+            if (ssl->arrays->preMasterSecret != NULL) {
+                ForceZero(ssl->arrays->preMasterSecret, ENCRYPT_LEN);
+                /* The key agreement routines take this as the size of the
+                 * buffer they may write, so put back what a freshly
+                 * allocated Arrays would carry. */
+                ssl->arrays->preMasterSz = ENCRYPT_LEN;
+            }
+        #if defined(HAVE_SESSION_TICKET) || !defined(NO_PSK)
+            ForceZero(ssl->arrays->psk_key, MAX_PSK_KEY_LEN);
+            ssl->arrays->psk_keySz = 0;
+        #endif
+        #ifdef WOLFSSL_TLS13
+            /* The key schedule secret. No API hands this one back. */
+            ForceZero(ssl->arrays->secret, SECRET_LEN);
+        #endif
+            if (!ssl->options.saveArrays) {
+                ForceZero(ssl->arrays->masterSecret, SECRET_LEN);
+            #ifdef HAVE_KEYING_MATERIAL
+                /* Tls13_Exporter() reads this one, and exporting keying
+                 * material requires the arrays to be kept, so it only goes
+                 * when the application did not ask for that. */
+                ForceZero(ssl->arrays->exporterSecret, WC_MAX_DIGEST_SIZE);
+            #endif
+            }
+        }
         XMEMSET(&ssl->msgsReceived, 0, sizeof(ssl->msgsReceived));
 
         /* Discard any partial handshake-message reassembly on reuse. */
--- a/src/ssl_api_dtls.c
+++ b/src/ssl_api_dtls.c
@@ -778,6 +778,13 @@ int wolfSSL_set_secret(WOLFSSL* ssl, word16 epoch,
         ret = BAD_FUNC_ARG;
     }
 
+    /* The handshake arrays are released once the handshake resources are
+     * freed, so a reused object may not have them any more. */
+    if (ret == 0 && ssl->arrays == NULL) {
+        WOLFSSL_MSG("Handshake arrays not available");
+        ret = BAD_FUNC_ARG;
+    }
+
     if (ret == 0 && ssl->arrays->preMasterSecret == NULL) {
         ssl->arrays->preMasterSz = ENCRYPT_LEN;
         ssl->arrays->preMasterSecret = (byte*)XMALLOC(ENCRYPT_LEN, ssl->heap,
--- a/src/ssl_api_dtls.c
+++ b/src/ssl_api_dtls.c
@@ -1369,6 +1369,227 @@ int wolfSSL_dtls_retransmit(WOLFSSL* ssl)
     return WOLFSSL_SUCCESS;
 }
 
+#ifdef WOLFSSL_DTLS13
+/* Is this object the kind the scheduled-work API operates on?
+ *
+ * A write-dup pair parks the read side's scheduled work in the shared WriteDup
+ * struct, and only wolfSSL_write() reconciles it back onto the write side.
+ * Such applications already have a working drain and are deliberately out of
+ * scope here, on either side of the pair.
+ */
+static int Dtls13ScheduledWorkObject(WOLFSSL* ssl)
+{
+    if (!ssl->options.dtls || !IsAtLeastTLSv1_3(ssl->version))
+        return 0;
+#ifdef HAVE_WRITE_DUP
+    if (ssl->dupWrite != NULL)
+        return 0;
+#endif
+
+    return 1;
+}
+
+/* Is there any point running the scheduled work now?
+ *
+ * While the handshake runs, wolfSSL_connect()/wolfSSL_accept() and
+ * wolfSSL_dtls_retransmit() already drive it. Kept separate from the object
+ * check above so the pump can tell a caller error, which it reports, from
+ * simply having nothing to do yet, which it does not.
+ */
+static int Dtls13ScheduledWorkReady(WOLFSSL* ssl)
+{
+    return Dtls13ScheduledWorkObject(ssl) && ssl->options.handShakeDone;
+}
+
+/* Can a key update be sent right now?
+ *
+ * DTLS must not have two in flight, so not while one of ours is still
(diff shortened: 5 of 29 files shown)


src/dtls.c
```diff
@@ -98,13 +98,23 @@ void DtlsResetState(WOLFSSL* ssl)
     ssl->options.tls1_1 = 0;
     ssl->options.tls1_3 = 0;
 #if defined(WOLFSSL_DTLS) && defined(WOLFSSL_DTLS_CID)
+    /* wolfSSL_dtls_set_pending_peer() may run on another thread, so take the
+     * lock the record layer uses for these two fields before dropping what
+     * that call left behind. */
+#ifdef WOLFSSL_RW_THREADED
+    if (wc_LockRwLock_Wr(&ssl->buffers.dtlsCtx.peerLock) == 0) {
```
Frauschi: fixed

src/ssl_crypto.c
```diff
@@ -1845,6 +1845,12 @@ int wolfSSL_HMAC_Init(WOLFSSL_HMAC_CTX* ctx, const void* key, int keylen,
             WOLFSSL_MSG("hmac set key error");
             WOLFSSL_ERROR(rc);
             wc_HmacFree(&ctx->hmac);
```
Frauschi, line 1847: fixed

src/internal.c
```diff
@@ -23734,11 +23740,33 @@ static void dtlsProcessPendingPeer(WOLFSSL* ssl, int deprotected, int isNewest)
...
+                        to->bufSz = 0;
+                    }
+                    else {
+                        to->bufSz = from->sz;
+                    }
+                }
+                if (to->sa != NULL) {
+                    XMEMCPY(to->sa, from->sa, from->sz);
+                    to->sz = from->sz;
+                    ssl->buffers.dtlsCtx.userSet = 1;
```
Frauschi, line 23780: fixed

src/ssl.c
```diff
@@ -5679,7 +5679,73 @@ size_t wolfSSL_get_client_random(const WOLFSSL* ssl, unsigned char* out,
...
+    #endif
+    #ifdef WOLFSSL_HAVE_TLS_UNIQUE
+        /* The channel binding of the connection that just ended. Leaving it
+         * in place would let the next caller bind to the wrong session. */
+        ForceZero(ssl->clientFinished, TLS_FINISHED_SZ_MAX);
+        ForceZero(ssl->serverFinished, TLS_FINISHED_SZ_MAX);
+        ssl->clientFinished_len = 0;
+        ssl->serverFinished_len = 0;
+    #endif
+    #ifdef WOLFSSL_DTLS13
```
dgarske, line 5709: **🔴 [High] wolfSSL_clear() wipes the DTLS 1.3 epoch table but leaves ssl-dtls13Epoch stale, breaking the next handshake with…**

The new block does `ForceZero(ssl->dtls13Epochs, sizeof(ssl->dtls13Epochs))` and then restores only epoch 0 and the two epoch pointers. It does not reset the epoch *counters* that live outside the array: `ssl->dtls13Epoch`, `ssl->dtls13PeerEpoch` and `ssl->dtls13InvalidateBefore` (wolfssl/internal.h:6471-6477). `InitSSL()` gets those for free from `XMEMSET(ssl, 0, sizeof(WOLFSSL))`; `wolfSSL_clear()` does not, so after the wipe the table and the counters disagree. A grep confirms `ssl->dtls13Epoch` is only ever assigned upward (src/tls13.c:8540, 12320, 12357, 14180; src/dtls13.c:2776) and never reset. `ssl->dtls13Rtx` (seenRecords/rtxRecords), `ssl->dtls13WaitKeyUpdateAck`, `ssl->dtls13DoKeyUpdate` and `ssl->options.sendKeyUpdate` are likewise untouched. Before this PR the table was left intact, so the table and counters stayed consistent (with stale keys — which is exactly what the PR set out to fix); the wipe without the counter reset turns that into a hard failure, so this is a regression introduced here. Note `wolfSSL_shutdown()` calls `wolfSSL_clear()` on success under `OPENSSL_EXTRA`/`WOLFSSL_WPAS_SMALL` (src/ssl_api_rw.c:726), so an application does not have to call `SSL_clear()` explicitly to hit it.

**Fix:** Reset the DTLS 1.3 epoch counters (and ideally the retransmit/ACK lists in `ssl->dtls13Rtx`, which still hold records tagged with epochs that no longer exist) alongside the table wipe. Concrete failure: after a completed DTLS 1.3 connection `ssl->dtls13Epoch == 3` (DTLS13_EPOCH_TRAFFIC0). The next `wolfSSL_connect()`/`wolfSSL_accept()` sends ClientHello/ServerHello through `Dtls13HandshakeSend()` (src/tls13.c:5184, 8451), which at src/dtls13.c:2124 sees `dtls13EncryptEpoch->epochNumber == 0 != dtls13Epoch == 3` and calls `Dtls13SetEpochKeys(ssl, 3, ENCRYPT_SIDE_ONLY)`; `Dtls13GetEpoch()` finds no epoch 3 in the wiped table and `Dtls13SetEpochKeys()` returns `BAD_STATE_E` (src/dtls13.c:2511-2513), aborting the handshake. Add a DTLS 1.3 reuse test (handshake, `wolfSSL_clear()`, handshake…

src/ssl_api_dtls.c
```diff
@@ -1355,6 +1362,188 @@ int wolfSSL_dtls_retransmit(WOLFSSL* ssl)
...
+        WOLFSSL_ERROR(ret);
+        return WOLFSSL_FATAL_ERROR;
+    }
+
+    /* Answer a KeyUpdate the peer requested. The read path defers this the
+     * same way, and SendData() is otherwise the only thing that clears it.
+     * Skip it while one of ours is still unacknowledged: DTLS must not have
+     * two KeyUpdates in flight, and Dtls13DoScheduledWork() above may have
+     * just started one. */
+    if (ssl->options.sendKeyUpdate && Dtls13CanSendKeyUpdate(ssl)) {
```
dgarske, line 1563: **🔴 [High] wolfSSL_dtls13_do_scheduled_work() leaves options.sendKeyUpdate set after WANT_WRITE, sending a duplicate KeyUpdate**

The new code clears `ssl->options.sendKeyUpdate` only when `SendTls13KeyUpdate()` returns exactly 0 ("Only drop the request once it is actually on the wire"). For DTLS that premise is wrong: `SendTls13KeyUpdate()` reaches `Dtls13HandshakeSend()`, which sets `ssl->dtls13WaitKeyUpdateAck = 1` unconditionally (src/dtls13.c:2134) and then calls `Dtls13SendOneFragmentRtx()`. That path commits the record to the retransmit list on `ret == 0 || ret == WANT_WRITE` (src/dtls13.c:1032-1035) and bumps `ssl->keys.dtls_handshake_number` on the same condition (src/dtls13.c:2152-2153). So on `WANT_WRITE` the KeyUpdate *is* committed — queued for retransmit, handshake number consumed, awaiting the peer ACK — but the request flag stays set. `Dtls13CanSendKeyUpdate()` then suppresses it until the ACK arrives, after which the next pump (or `wolfSSL_write()` -> `SendData()`) sends a second, spurious KeyUpdate and rotates the epoch again. `SendData()` gets this right by clearing the flag *before* the call (src/internal.c:27909-27911), so the two paths now have opposite semantics for the same flag. The convention is stated explicitly at src/ssl_api_rw.c:198-200: "WANT_WRITE is safe to clear. Data is buffered in output buffer or in DTLS RTX queue".

**Fix:** Clear `ssl->options.sendKeyUpdate` before calling `SendTls13KeyUpdate()`, matching `SendData()`, or at minimum treat `WC_NO_ERR_TRACE(WANT_WRITE)` as success for the purpose of dropping the request.

src/ssl_api_dtls.c
```diff
@@ -1355,6 +1362,188 @@ int wolfSSL_dtls_retransmit(WOLFSSL* ssl)
...
+    if (!Dtls13ScheduledWorkObject(ssl)) {
+        WOLFSSL_MSG("Not a DTLS 1.3 object this API handles");
+        WOLFSSL_LEAVE("wolfSSL_dtls13_do_scheduled_work", WOLFSSL_FATAL_ERROR);
+        return WOLFSSL_FATAL_ERROR;
+    }
+
+    if (!Dtls13ScheduledWorkReady(ssl))
+        return WOLFSSL_SUCCESS;
+
+    ret = Dtls13DoScheduledWork(ssl);
```
dgarske, line 1551: **🟠 [Medium] The scheduled-work pump has no way to flush a partially sent record, and the predicate cannot report one**

`Dtls13DoScheduledWork()` clears `ssl->dtls13Rtx.sendAcks` under the mutex and flushes the seen-record list before `SendDtls13Ack()` ends in `SendBuffered()` (src/dtls13.c:2834-2836, 3210-3212). If `SendBuffered()` returns `WANT_WRITE`, the ACK record sits in `ssl->buffers.outputBuffer` with `sendAcks` already 0, the seen list already flushed, and `ssl->dtls13SendingAckOrRtx` stuck at 1. The new API offers no retry: the pump never inspects `ssl->buffers.outputBuffer.length`, and `wolfSSL_dtls13_pending_work()` only reports `dtls13DoKeyUpdate` / `options.sendKeyUpdate` / `sendAcks` / `retransmit`, all now zero — so the documented drain loop reports "nothing to do" and exits. Every other caller of `Dtls13DoScheduledWork()` pairs it with exactly this retry: src/ssl_api_hs.c:177-210 and 713-725, src/tls13.c:14540-14552, 15785-15795, 16401-16412 all do `if (outputBuffer.length > 0) SendBuffered(); ... ssl->dtls13SendingAckOrRtx = 0;`. Since the entire purpose of this API is an application whose read thread never calls `wolfSSL_write()`, there is no other code path left to flush the buffer, so the peer keeps retransmitting waiting for an ACK that is already built but never sent — the exact protocol break the PR set out to fix.

**Fix:** Flush a pending output buffer at the top of the pump and clear `ssl->dtls13SendingAckOrRtx` on success, and include `ssl->buffers.outputBuffer.length > 0` in `wolfSSL_dtls13_pending_work()` so a drain loop knows a retry is owed. Add a test that makes the send callback return `WANT_WRITE` once and asserts the ACK still reaches the peer on the next pump.

src/ssl_crypto.c
```diff
@@ -1845,6 +1845,16 @@ int wolfSSL_HMAC_Init(WOLFSSL_HMAC_CTX* ctx, const void* key, int keylen,
...
+            /* Context is no longer keyed, so drop any previous key's pads
```
dgarske, line 1848: **🟠 [Medium] HMAC_Init failure path also clears ctx-type, breaking a retry with the same digest**

Wiping the saved pads on the `wc_HmacSetKey()` failure path is correct, but the change also sets `ctx->type = WC_HASH_TYPE_NONE`. `ctx->type` was already populated from the caller's `EVP_MD` a few lines earlier by `wolfssl_evp_md_to_hash_type()`. Clearing it means the documented FIPS retry — the adjacent comment says "in FIPS mode a key < 14 characters will fail here" — no longer works: `HMAC_Init_ex(ctx, longerKey, len, NULL /* same md */, NULL)` is now rejected by the `(type == NULL) && (ctx->type == WC_HASH_TYPE_NONE)` guard at src/ssl_crypto.c:1800-1803. The security goal (never recover a MAC from wiped pads) is fully met by `ctx->hmac.macType = WC_HASH_TYPE_NONE` alone, since the unkeyed-recover path is gated on `(key == NULL) && (ctx->hmac.macType == WC_HASH_TYPE_NONE)`.

**Fix:** Drop `ctx->type = WC_HASH_TYPE_NONE;` from the set-key failure path (keep it in `wolfSSL_HMAC_cleanup`, where a full reset is the intended semantic).

src/internal.c
```diff
@@ -23734,11 +23740,38 @@ static void dtlsProcessPendingPeer(WOLFSSL* ssl, int deprotected, int isNewest)
                     !ssl->buffers.dtlsCtx.processingPendingRecord;
         }
         else {
-            /* Pending peer present and record deprotected. Update the peer. */
+            /* Pending peer present and record deprotected. Update the peer
+             * here rather than through wolfSSL_dtls_set_peer, which would take
```
dgarske: **🟠 [Medium] dtlsProcessPendingPeer duplicates SockAddrSet instead of sharing it**

Inlining the peer copy to avoid recursively taking `peerLock` through `wolfSSL_dtls_set_peer()` is the right call — that would have been a self-deadlock on a POSIX write lock. But the result is a byte-for-byte reimplementation of `SockAddrSet()` (src/ssl_api_dtls.c:127-153): same `sz > bufSz` grow test, same free/alloc, same `sz = 0; bufSz = 0` on failure, same `XMEMCPY` + `sz` assignment, plus the `userSet` handling from `wolfSSL_dtls_set_peer()`. I verified the two copies are currently equivalent, but they now have to be kept in sync by hand across two files, and the comment ("exactly as SockAddrSet does") is the only thing tying them together.

**Fix:** Promote `SockAddrSet()` to `WOLFSSL_LOCAL` (or add a `wolfssl_local_dtls_set_peer_nolock()` helper) and call it from both places, so the buffer-reuse policy lives in one function.

src/internal.c
```diff
@@ -23719,6 +23721,10 @@ static int dtlsRecordIsNewest(WOLFSSL* ssl)
  */
 static void dtlsProcessPendingPeer(WOLFSSL* ssl, int deprotected, int isNewest)
 {
+#ifdef WOLFSSL_RW_THREADED
```
dgarske, line 23740: **🟠 [Medium] Three new peerLock call sites use three different lock-failure policies**

The PR adds `peerLock` acquisition in three places with three different behaviours on failure. `DtlsResetState()` (src/dtls.c:112-120) clears the pending peer *anyway*, with a comment arguing that leaving one behind is worse than the race. `dtlsProcessPendingPeer()` returns immediately, which also skips the `ssl->buffers.dtlsCtx.processingPendingRecord = 0` reset that the function performed unconditionally before this change — so a failed lock now leaves that flag latched, and the next record takes the `processingPendingRecord` branch it should not. `ProcessReplyEx()` (src/internal.c:25137-25145) silently skips the clear. Only the first has a documented rationale.

**Fix:** Pick one policy for a failed `peerLock` acquisition across all three sites and state it once. If clearing anyway is right for `DtlsResetState()`, it is also right for the `processingPendingRecord` reset in `dtlsProcessPendingPeer()` and `ProcessReplyEx()`, since none of those touch the shared address buffer.

wolfcrypt/src/hpke.c
```diff
@@ -1069,13 +1075,17 @@ static int wc_HpkeDecap(Hpke* hpke, void* receiverKey, const byte* pubKey,
                     break;
                 }
 
+                prevRng = ((ecc_key*)receiverKey)->rng;
```
dgarske: **🔵 [Low] HPKE RNG save/restore mutates a key the ECH server shares across connections**

The restore writes back into the caller's key object. For the main in-tree consumer that is a *shared* object: src/tls.c:14730 passes `echConfig->receiverPrivkey` — owned by the `WOLFSSL_CTX` and reused by every `WOLFSSL` created from it — straight into the decap path. Two concurrent ClientHellos therefore interleave as: T1 saves prevRng=NULL and installs rngT1; T2 saves prevRng=rngT1 and installs rngT2; T1 restores NULL and frees rngT1; T2 restores rngT1, which is now freed. The underlying data race is pre-existing (the `wc_ecc_set_rng(receiverKey, rng)` write on the shared key was already there), and the pre-patch steady state was arguably worse — the key permanently held a dangling RNG. But the patch's stated goal, handing the caller's key back unchanged, is not actually achieved for this consumer, and it adds a new outcome: because `wc_ecc_make_key_ex` never stores the RNG in `key->rng`, the correctly restored value is `NULL`, so a racing thread can read `private_key->rng == NULL` at wolfcrypt/src/ecc.c:5029 and get `MISSING_RNG_E` on an otherwise valid ECH decrypt.

**Fix:** Prefer not mutating the caller's key at all: add an internal shared-secret entry point that takes the RNG explicitly, or have `wc_HpkeDecap()` work on a local copy of the private key. If the save/restore stays, document in wolfssl/wolfcrypt/hpke.h that `wc_HpkeDecap`/`wc_HpkeOpenBase` are not re-entrant on the same key object.

doc/dox_comments/header_files/ssl.h
```diff
@@ -3951,6 +3951,72 @@ int  wolfSSL_dtls_got_timeout(WOLFSSL* ssl);
...
+    Not for write-dup applications, which drain through wolfSSL_write().
+
+    A KeyUpdate is only sent while none of ours is still unacknowledged, since
+    DTLS must not have two in flight. Completing one we started needs the
+    peer's acknowledgement processed, which rotates the sending keys and so
+    does not happen here, so in a WOLFSSL_RW_THREADED build a peer request
+    arriving after that point is held rather than answered.
+
+    \return WOLFSSL_SUCCESS on success, including when there was nothing to do.
+    \return WOLFSSL_FATAL_ERROR if ssl is NULL, is not a DTLS 1.3 object, is
```
dgarske, line 3979: **🔵 [Low] New API docs omit that WOLFSSL_FATAL_ERROR can be a retryable WANT_WRITE**

The function sets `ssl->error = ret` and returns `WOLFSSL_FATAL_ERROR` for any negative return, and `WANT_WRITE` is among them: `SendDtls13Ack()` ends in `SendBuffered()` (src/dtls13.c:3212), and `Dtls13RtxSendBuffered()`/`Dtls13HandshakeSend()`/`Dtls13SendOneFragmentRtx()` all propagate it. The docs say only "...or the send failed", and the documented example simply `break`s out of the drain loop, so a caller has no signal that it should consult `wolfSSL_get_error()` and retry. This matches how the neighbouring `wolfSSL_dtls_retransmit()` behaves, so it is not an internal inconsistency — but this is the call applications are explicitly told to poll in a loop.

**Fix:** Document the `wolfSSL_get_error()` follow-up and the `WANT_WRITE` retry contract, and show the retry in the example loop.

tests/api.c
```diff
@@ -9849,6 +9897,142 @@ static int test_wolfSSL_clear_secure_renegotiation(void)
     return EXPECT_RESULT();
 }
 
+#if defined(HAVE_MANUAL_MEMIO_TESTS_DEPENDENCIES) && defined(WOLFSSL_TLS13) && \
```
dgarske, line 9906: **🔵 [Low] No test covers DTLS 1.3 object reuse after wolfSSL_clear**

The new test exercises `wolfSSL_clear()` on a TLS 1.3 object only. The riskiest part of the change — the `#ifdef WOLFSSL_DTLS13` epoch-table wipe and the epoch-0 restore at src/ssl.c:5706-5716 — has no coverage at all, which is why the epoch-counter inconsistency reported above would pass CI. `test_dtls_communication()` (tests/api/test_dtls.c:1131) does not call shutdown, so no existing DTLS test reaches `wolfSSL_clear()` either.

**Fix:** Add a DTLS 1.3 variant: complete a handshake, assert the epoch keys are populated, call `wolfSSL_clear()` on both ends, assert `ssl->dtls13Epochs[]` is zeroed apart from epoch 0, then run a second full handshake on the same objects and `test_dtls_communication()` over it.

tests/api.c
```diff
@@ -35677,6 +35928,131 @@ static int test_write_dup_oom(void)
...
+    if (EXPECT_SUCCESS()) {
+        srpInited = 1;
+    }
+
+    ExpectIntEQ(wolfSSL_GetAllocators(&prev_mc, &prev_fc, &prev_rc), 0);
+    ExpectIntEQ(wolfSSL_SetAllocators(srp_oom_malloc_cb, srp_oom_free_cb,
+            srp_oom_realloc_cb), 0);
+    if (EXPECT_SUCCESS()) {
+        allocators_set = 1;
+    }
```
dgarske, line 36033: **🔵 [Low] test_wc_SrpComputeKey_oom hard-codes the allocation ordinal, making it build-fragile**

The test arms a global failing allocator at exactly the 6th allocation, relying on the comment's claim that `wc_SrpComputeKey` allocates "hash, digest and the four mp_ints in that order". That ordinal is not enforced by anything and shifts with the math backend (SP vs. fastmath vs. integer.c), with `WOLFSSL_SMALL_STACK` hash contexts, and with any future allocation added ahead of the mp_ints. If it drifts, the call returns a different error and `ExpectIntEQ(..., MEMORY_E)` fails for a reason unrelated to the fix — or, worse, passes while failing a different allocation and no longer exercising the uninitialized-mp_int cleanup path at all.

**Fix:** Either sweep `srp_oom_fail_at` over a range (1..8) and assert the call always returns an error and never crashes — which exercises the cleanup path regardless of ordinal drift — or key the failure on the allocation size (`sizeof(mp_int)`) plus an occurrence count rather than a raw global ordinal.

wolfcrypt/test/test.c
```diff
@@ -37976,6 +38033,13 @@ static wc_test_ret_t hpke_test_single(Hpke* hpke, int kem, int kdf, int aead)
             ret = WC_TEST_RET_ENC_NC;
     }
 
+    /* Seal and open install a temporary RNG on the caller's key and then free
```
dgarske, line 38036: **🔵 [Low] hpke_test_multi exercises the fixed encap/decap paths but got no RNG-restore assertion**

The new `hpke_test_key_rng_kept()` assertions were added only to `hpke_test_single()`. `hpke_test_multi()` drives `wc_HpkeInitSealContext`/`wc_HpkeInitOpenContext`, which funnel through `wc_HpkeSetupBaseSender` -> `wc_HpkeEncap` and `wc_HpkeSetupBaseReceiver` -> `wc_HpkeDecap` — exactly the code being fixed — and holds its own `ephemeralKey`/`receiverKey`. A regression reintroduced only on the context-API path would go uncaught. (The helpers themselves are correctly guarded and the RNG is still alive at the assertion point, which runs before `wc_FreeRng`.)

**Fix:** Apply the same `hpke_test_set_key_rng()`/`hpke_test_key_rng_kept()` pair in `hpke_test_multi()`, keyed off `hpke->kem`.

src/dtls.c
```diff
@@ -98,13 +103,27 @@ void DtlsResetState(WOLFSSL* ssl)
     ssl->options.tls1_1 = 0;
     ssl->options.tls1_3 = 0;
 #if defined(WOLFSSL_DTLS) && defined(WOLFSSL_DTLS_CID)
+#ifdef WOLFSSL_RW_THREADED
```
dgarske, line 106: **🔵 [Low] DtlsResetState comment misstates what a recursive write-lock does**

The comment justifies ignoring the lock failure with "The realistic failure here is this thread already holding the lock, which means the access is exclusive in any case." That is not what happens: `wc_LockRwLock_Wr()` maps to `pthread_rwlock_wrlock()`, and re-acquiring a write lock the calling thread already holds is undefined behaviour under POSIX — on glibc it deadlocks rather than returning an error. The tolerated-failure path can never be reached for that reason, and the stated rationale invites a future call site that already holds `peerLock`, which would hang. The code as written is safe: both call sites (src/internal.c:39809, src/tls13.c:14306) are reached from `DoProcessReplyEx()`, and every `peerLock` acquisition on that path releases before returning.

**Fix:** Reword so the tolerated failure is a broken/uninitialised lock, and state the caller requirement that `peerLock` must not already be held.

tests/api/test_dtls.c
```diff
@@ -4012,6 +4016,20 @@ static void test_AEAD_get_limits(WOLFSSL* ssl, w64wrapper* hardLimit,
     }
 }
 
+/* A DTLS 1.3 key update is scheduled while reading but is only transmitted
+ * from the write path. With WOLFSSL_RW_THREADED the read path does no
+ * scheduled work at all, because the reader must not transmit while a writer
+ * thread may be doing so, so an application that reads without writing has to
+ * pump the deferred work itself. Do what such an application would do. */
+static void test_AEAD_drain_scheduled_work(WOLFSSL* ssl)
```
dgarske, line 4024: **🔵 [Low] test_AEAD_drain_scheduled_work omits the !WOLFSSL_LEANPSK guard used everywhere else**

`wolfSSL_dtls13_do_scheduled_work()` is declared in wolfssl/ssl.h under `#if defined(WOLFSSL_DTLS13) && !defined(WOLFSSL_LEANPSK)` and defined inside the same `#ifndef WOLFSSL_LEANPSK` block of src/ssl_api_dtls.c. This helper guards only on `WOLFSSL_RW_THREADED`, and its enclosing block (tests/api/test_dtls.c:3950-3952) has no `WOLFSSL_LEANPSK` exclusion, so a `WOLFSSL_LEANPSK` + `WOLFSSL_DTLS13` + `WOLFSSL_RW_THREADED` test build would hit an implicit declaration. The same PR's `TEST_DTLS13_PUMP` in tests/utils.h:57-58 does include `!defined(WOLFSSL_LEANPSK)`, so this reads as an oversight rather than a deliberate choice.

**Fix:** Match the availability condition used in tests/utils.h and wolfssl/ssl.h.

wolfcrypt/src/hpke.c
```diff
@@ -1019,6 +1024,7 @@ static int wc_HpkeDecap(Hpke* hpke, void* receiverKey, const byte* pubKey,
     int ret;
 #if defined(ECC_TIMING_RESISTANT) || defined(WOLFSSL_CURVE25519_BLINDING)
```
dgarske: **🔵 [Low] prevRng in wc_HpkeDecap is declared under a broader guard than any of its uses**

The declaration is gated on `#if defined(ECC_TIMING_RESISTANT) || defined(WOLFSSL_CURVE25519_BLINDING)`, but the two uses are gated on `HAVE_ECC && ECC_TIMING_RESISTANT` (lines 1078/1088) and `HAVE_CURVE25519 && !NO_SHA256 && WOLFSSL_CURVE25519_BLINDING` (lines 1103/1112). `ECC_TIMING_RESISTANT` is defined by `--enable-harden` regardless of whether ECC is enabled, and `WOLFSSL_CURVE25519_BLINDING` is suppressed for `USE_INTEL_SPEEDUP`/`WOLFSSL_ARMASM`/`CURVE25519_SMALL`, so `--enable-ech --enable-curve25519 --disable-ecc --enable-intelasm` declares both `rng` and `prevRng` with zero uses. This is not new breakage — the pre-existing `rng` has the identical over-broad guard and already warns there — but the change doubles the exposure on a line it touches. `wc_HpkeEncap` got this right: its guard `#if defined(ECC_TIMING_RESISTANT) && defined(HAVE_ECC)` exactly matches its single use site. All eight guard combinations were enumerated; none gives a compile error or an uninitialized read (the `break` on `rng == NULL` precedes the `prevRng` assignment, and `prevRng` is never read afterwards).

**Fix:** Tighten both declarations to match the use sites while the line is being touched.

wolfcrypt/src/hpke.c
```diff
@@ -829,13 +830,17 @@ static int wc_HpkeEncap(Hpke* hpke, void* ephemeralKey, void* receiverKey,
...
             wc_ecc_set_rng((ecc_key*)ephemeralKey, rng);
 #endif
 
             ret = wc_ecc_shared_secret((ecc_key*)ephemeralKey,
                 (ecc_key*)receiverKey, dh, &dh_len);
 
 #ifdef ECC_TIMING_RESISTANT
+            /* The key belongs to the caller, so put back whatever RNG it had
+             * before this RNG is freed. */
+            wc_ecc_set_rng((ecc_key*)ephemeralKey, prevRng);
```
dgarske: **⚪ [Info] New set_rng restore calls discard their return code without a (void) cast**

The three restore calls added in hpke.c (lines 843, 1088, 1112) drop the `int` return of `wc_ecc_set_rng`/`wc_curve25519_set_rng` with no `(void)` cast, while the same PR does use an explicit `(void)` at src/ssl_ech.c:130 and wolfcrypt/test/test.c:37605 and 37612. The project rule is to check every return code; an explicit `(void)` documents a deliberate discard. Both functions can only fail with `BAD_FUNC_ARG` on a NULL key, which cannot happen here.

**Fix:** Prefix the three restore calls with `(void)` to match ssl_ech.c and test.c in this same change set.

src/internal.c
```diff
@@ -25098,8 +25134,15 @@ int ProcessReplyEx(WOLFSSL* ssl, int allowSocketErr)
                 && ret != WC_NO_ERR_TRACE(WC_PENDING_E)
 #endif
             ) {
+        #ifdef WOLFSSL_RW_THREADED
```
dgarske, line 25156: **⚪ [Info] ProcessReplyEx body not re-indented under the new conditional**

The two statements now guarded by `if (wc_LockRwLock_Wr(...) == 0) {` keep their old indentation, and the closing brace sits at a deeper level than the opening one. It reads as if the statements are unconditional, which is exactly the thing a reviewer needs to see at a glance here.

**Fix:** Indent the guarded statements one level so the control flow matches the layout.

dgarske (commented):
## Skoll Code Review

**Findings:** 21 total — 19 posted, 2 skipped
**19 finding(s) posted as inline comments** (see file-level comments below)

### Posted findings
- [High] **wolfSSL_clear() wipes the DTLS 1.3 epoch table but leaves ssl-dtls13Epoch stale, breaking the next handshake with BAD_STATE_E** — `src/ssl.c:5706-5716`
- [High] **wolfSSL_dtls13_do_scheduled_work() leaves options.sendKeyUpdate set after WANT_WRITE, sending a duplicate KeyUpdate** — `src/ssl_api_dtls.c:1530-1539`
- [Medium] **The scheduled-work pump has no way to flush a partially sent record, and the predicate cannot report one** — `src/ssl_api_dtls.c:1518-1524`
- [Medium] **HMAC_Init failure path also clears ctx-type, breaking a retry with the same digest** — `src/ssl_crypto.c:1848-1857`
- [Medium] **dtlsProcessPendingPeer duplicates SockAddrSet instead of sharing it** — `src/internal.c:23744-23775`
- [Medium] **Three new peerLock call sites use three different lock-failure policies** — `src/internal.c:23724-23727`
- [Low] **HPKE RNG save/restore mutates a key the ECH server shares across connections** — `wolfcrypt/src/hpke.c:1078-1090`
- [Low] **New API docs omit that WOLFSSL_FATAL_ERROR can be a retryable WANT_WRITE** — `doc/dox_comments/header_files/ssl.h:3973-3976`
- [Low] **No test covers DTLS 1.3 object reuse after wolfSSL_clear** — `tests/api.c:9900-10035`
- [Low] **test_wc_SrpComputeKey_oom hard-codes the allocation ordinal, making it build-fragile** — `tests/api.c:36030-36042`
- [Low] **hpke_test_multi exercises the fixed encap/decap paths but got no RNG-restore assertion** — `wolfcrypt/test/test.c:38036-38041`
- [Low] **DtlsResetState comment misstates what a recursive write-lock does** — `src/dtls.c:106-113`
- [Low] **test_AEAD_drain_scheduled_work omits the !WOLFSSL_LEANPSK guard used everywhere else** — `tests/api/test_dtls.c:4024-4031`
- [Low] **prevRng in wc_HpkeDecap is declared under a broader guard than any of its uses** — `wolfcrypt/src/hpke.c:1025-1028`
- [Info] **New set_rng restore calls discard their return code without a (void) cast** — `wolfcrypt/src/hpke.c:843`
- [Info] **ProcessReplyEx body not re-indented under the new conditional** — `src/internal.c:25137-25145`
- [Info] **Unnecessary address-of on array arguments to ForceZero** — `src/ssl.c:5695-5696`
- [Info] **New test guards use RC_NO_RNG, which looks like a typo for WC_NO_RNG** — `tests/api.c:35932,36006`
- [Info] **TEST_DTLS13_PUMP is not statement-safe** — `tests/utils.h:57-62`

### Skipped findings
- [Low] ~~AEAD hard-limit and send-limit coverage is silently dropped under WOLFSSL_RW_THREADED~~
- [Low] ~~dtls13Rtx.mutex is taken for fields nothing else guards with it~~

src/ssl.c
```diff
@@ -5679,7 +5679,73 @@ size_t wolfSSL_get_client_random(const WOLFSSL* ssl, unsigned char* out,
...
+    #ifdef WOLFSSL_MULTICAST
+        if (ssl->options.haveMcast) {
+            int i;
+
+            for (i = 0; i < WOLFSSL_DTLS_PEERSEQ_SZ; i++)
+                ssl->keys.peerSeq[i].peerId = INVALID_PEER_ID;
+        }
+    #endif
+    #ifdef WOLFSSL_TLS13
+        ForceZero(&ssl->clientSecret, sizeof(ssl->clientSecret));
```
dgarske: **⚪ [Info] Unnecessary address-of on array arguments to ForceZero**

`ssl->clientSecret` and `ssl->serverSecret` are `byte[SECRET_LEN]` (wolfssl/internal.h:6246-6247), so `&ssl->clientSecret` yields `byte (*)[SECRET_LEN]` where the plain array name already decays to `byte*`. It is correct but inconsistent with the adjacent lines in the same hunk, which pass `ssl->clientFinished` and `ssl->dtls13Epochs` without the `&`.

**Fix:** Drop the `&` on both lines to match the surrounding code.

tests/api.c
```diff
@@ -35677,6 +35928,131 @@ static int test_write_dup_oom(void)
     return EXPECT_RESULT();
 }
 
+#if defined(OPENSSL_EXTRA) && defined(WOLFCRYPT_HAVE_SRP) &&                   \
+    defined(WOLFSSL_SMALL_STACK) && !defined(RC_NO_RNG) &&                     \
```
dgarske: **⚪ [Info] New test guards use RC_NO_RNG, which looks like a typo for WC_NO_RNG**

Both new guard blocks test `!defined(RC_NO_RNG)`. wolfSSL's macro is `WC_NO_RNG` (used correctly elsewhere in the same file at lines 611, 1643, 11431). `RC_NO_RNG` is never defined anywhere in the tree — it only appears in `.wolfssl_known_macro_extras` and in one pre-existing guard at tests/api.c:155, so the source-text lint will not flag it, but the condition is inert. Note the test does not actually need an RNG (`wc_SrpComputeKey` does not use one), so the guard may simply be copy-paste that can be dropped.

**Fix:** Drop the clause, or correct it to `!defined(WC_NO_RNG)` if an RNG dependency was actually intended.

tests/utils.h
```diff
@@ -51,6 +51,18 @@ extern const char* currentTestName;
 #define HAVE_MANUAL_MEMIO_TESTS_DEPENDENCIES_BUILD
 #endif
 
+/* With WOLFSSL_RW_THREADED the read path never transmits, so anything a read
+ * schedules to be sent, an ACK in particular, is only sent from the write
+ * side. Stand in for the application, which is required to pump that work
+ * from its write thread. A no-op in builds where reads send for themselves. */
```
dgarske, line 57: **⚪ [Info] TEST_DTLS13_PUMP is not statement-safe**

Neither arm of the macro is wrapped, so `if (x) TEST_DTLS13_PUMP(ssl); else ...` would misparse or dangle depending on which arm is selected. All four current call sites (tests/api/test_dtls.c:855, 906, 975, 2157 and tests/api/test_dtls13.c:1231-1232) are plain statements in `EXPECT_DECLS` functions, so nothing is broken today — this is about the next caller.

**Fix:** Wrap both arms in `do { ... } while (0)`.

dgarske (changes requested):
https://github.com/wolfSSL/wolfssl/pull/11018#pullrequestreview-4831542394

src/ssl.c
Frauschi: Confirmed, and a regression from this PR. Now resets `dtls13Epoch`, `dtls13PeerEpoch`, `dtls13InvalidateBefore`, the key-update flags and `options.sendKeyUpdate`, and calls `Dtls13FreeFsmResources()`. New `test_dtls13_reuse_after_clear` fails without the reset, passes with it.

src/ssl_api_dtls.c
Frauschi: Confirmed. The flag is now cleared before the call, as `SendData()` does.

src/ssl_api_dtls.c
Frauschi: Fixed, with one change: the flush and the predicate are gated on `dtls13SendingAckOrRtx`, not `outputBuffer.length` alone, so the pump cannot flush a partial *application* record out from under the caller's `wolfSSL_write()` retry. Same test `src/tls13.c:16496` uses.

src/ssl_crypto.c
Frauschi: Agreed, dropped the `ctx->type` clear. `ctx->hmac.macType` alone gates the recover path.

src/internal.c
Frauschi: Agreed. Promoted to `wolfssl_local_SockAddrSet()` and called from both places.

src/internal.c
Frauschi: Agreed, and the latched `processingPendingRecord` was a good catch. One policy now: a failed acquisition means a broken lock, so the bookkeeping proceeds and only the `dtlsCtx.peer` promotion is skipped.

wolfcrypt/src/hpke.c
Frauschi: Done via a private-only copy (new `wc_HpkeCopyPrivateKey()`), so the caller's key is never written. Only `devId == INVALID_DEVID` keys are copied - a device-backed key keeps the save/restore, since it computes on the device and never reads `key->rng`, and copying would silently move it into software. A probe confirmed the copy engages on all four KEMs and on the ECH path.

doc/dox_comments/header_files/ssl.h
Frauschi: Agreed, both entry points now document the `WANT_WRITE` retry and the example shows it.

tests/api.c
Frauschi: Agreed. Added `test_dtls13_reuse_after_clear`, in `tests/api/test_dtls13.c` with the other DTLS 1.3 memio tests.

tests/api.c
Frauschi: Agreed. Now sweeps 1..8, asserts only the ordinals that land, and checks at least one did. Still aborts with SIGTRAP if the `mpInited` fix is reverted.

wolfcrypt/test/test.c
Frauschi: Agreed, same set/assert pair added to `hpke_test_multi()`, keyed off `hpke->kem`.

src/dtls.c
Frauschi: Correct. Reworded: the tolerated failure is a broken or uninitialised lock, and callers must not already hold `peerLock`.

tests/api/test_dtls.c
Frauschi: Oversight, fixed.

wolfcrypt/src/hpke.c
Frauschi: Agreed, each declaration now matches its use sites.

wolfcrypt/src/hpke.c
Frauschi: Fixed.

src/internal.c
Frauschi: Fixed - the block is no longer conditional, so layout and control flow agree.

src/ssl.c
Frauschi: Fixed.

tests/api.c
Frauschi: Right, and `wc_SrpComputeKey` uses no RNG anyway. Dropped from both; the pre-existing one at line 155 left alone.

tests/utils.h
Frauschi: Fixed.

dgarske (dismissed):

douzzer (dismissed):
```
wolfcrypt/src/hpke.c:1087:    int ret = NOT_COMPILED_IN;
```

Frauschi:
> ```
> unescaped error code operands (missing WC_NO_ERR_TRACE()):
> wolfcrypt/src/hpke.c:1087:    int ret = NOT_COMPILED_IN;
> ```

Fixed

dgarske (approved):

PR #11019, opened 2026-07-30
Fixed encode policy oid

stenslae:
# Description

Fixed a bug in EncodePolicyOID where combined first identifiers (the first two OID arcs) were improperly cast to a single byte when exceeding 127, rather than using the required ITU-T X.690 base-128 continuation encoding. Also added improved bounds checking.

# Testing

Added new test coverage to exercise these changes.

# Checklist

 - [X] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/src/ssl_asn1.c
+++ b/src/ssl_asn1.c
@@ -5845,7 +5845,8 @@ int wc_OBJ_sn2nid(const char *sn)
      * buffer.
      *
      * String is of the form "1.2.840.113549.1.9.1" and is always NUL
-     * terminated. Truncated when the buffer is too small.
+     * terminated. Fails when the buffer is too small; unlike OpenSSL's
+     * OBJ_obj2txt(), it does not truncate and report the needed length.
      *
      * @param [out] buf     Buffer to hold string.
      * @param [in]  bufLen  Length of buffer in bytes.
@@ -5853,6 +5854,7 @@ int wc_OBJ_sn2nid(const char *sn)
      * @return  Length of string that would be written, excluding the NUL
      *          terminator, on success.
      * @return  0 when decoding the object fails.
+     * @return  ASN_PARSE_E when the object's length cannot be parsed.
      */
     static int wolfssl_obj2txt_numeric(char *buf, int bufLen,
                                        const WOLFSSL_ASN1_OBJECT *a)
@@ -5861,6 +5863,11 @@ int wc_OBJ_sn2nid(const char *sn)
         int    length;
         word32 idx = 0;
         byte   tag;
+
+        /* Fail closed: keep buf a valid (empty) NUL-terminated string on
+         * every error path, since the doc contract promises it is always
+         * NUL terminated even when decoding fails. */
+        buf[0] = '\0';
 
         if (GetASNTag(a->obj, &idx, &tag, a->objSz) != 0) {
             return WOLFSSL_FAILURE;
@@ -5876,17 +5883,17 @@ int wc_OBJ_sn2nid(const char *sn)
             return ASN_PARSE_E;
         }
 
-        /* save an extra byte for null term. */
-        if (bufLen < MAX_OID_STRING_SZ) {
-            bufSz = bufLen - 1;
-        }
-        else {
-            bufSz = MAX_OID_STRING_SZ - 1;
-        }
+        /* wc_DecodePolicyOID() accounts for the NUL terminator itself, so
+         * pass the caller's full buffer length rather than reserving a
+         * byte here. Do not clamp to MAX_OID_STRING_SZ: that would turn a
+         * caller-supplied buffer larger than 64 bytes into a spurious
+         * failure for OID strings >= 64 chars. */
+        bufSz = bufLen;
 
-        if ((bufSz = DecodePolicyOID(buf, (word32)bufSz, a->obj + idx,
+        if ((bufSz = wc_DecodePolicyOID(buf, (word32)bufSz, a->obj + idx,
                     (word32)length)) <= 0) {
             WOLFSSL_MSG("Error decoding OID");
+            buf[0] = '\0';
             return WOLFSSL_FAILURE;
         }
 
@@ -6259,7 +6266,7 @@ int wc_OBJ_sn2nid(const char *sn)
         }
 
     #ifdef WOLFSSL_CERT_EXT
-        ret = EncodePolicyOID(out, &outSz, s, NULL);
+        ret = wc_EncodePolicyOID(out, &outSz, s, NULL);
         if (ret == 0) {
             /* sum OID */
             sum = wc_oid_sum(out, outSz);
@@ -6324,7 +6331,7 @@ int wc_OBJ_sn2nid(const char *sn)
             return NULL;
 
         /* If s is numerical value, try to sum oid */
-        ret = EncodePolicyOID(out, &outSz, s, NULL);
+        ret = wc_EncodePolicyOID(out, &outSz, s, NULL);
         if (ret == 0 && outSz > 0) {
             /* If numerical encode succeeded then just
              * create object from that because sums are
--- a/src/x509.c
+++ b/src/x509.c
@@ -1358,7 +1358,7 @@ static int wolfssl_x509_add_custom_ext(WOLFSSL_X509 *x509,
 
     if (err == 0) {
         XMEMCPY(val, ext->value.data, ext->value.length);
-        if (wolfSSL_OBJ_obj2txt(oid, MAX_OID_STRING_SZ, ext->obj, 1) < 0) {
+        if (wolfSSL_OBJ_obj2txt(oid, MAX_OID_STRING_SZ, ext->obj, 1) <= 0) {
             err = 1;
         }
     }
--- a/tests/api/test_asn.c
+++ b/tests/api/test_asn.c
@@ -2081,6 +2081,559 @@ int test_wc_DecodeObjectId(void)
     return EXPECT_RESULT();
 }
 
+int test_wc_EncodePolicyOID(void)
+{
+    EXPECT_DECLS;
+
+#if (defined(WOLFSSL_CERT_GEN) && defined(WOLFSSL_CERT_EXT)) \
+        || defined(OPENSSL_EXTRA)
+    {
+        byte out[MAX_OID_SZ];
+        word32 outSz;
+
+        /* Test 1: combined first identifier fits in one byte.
+         * "1.2.3.4.5" -> 40*1+2=42 (0x2a), then 3, 4, 5 */
+        outSz = sizeof(out);
+        ExpectIntEQ(wc_EncodePolicyOID(out, &outSz, "1.2.3.4.5", NULL), 0);
+        ExpectIntEQ((int)outSz, 4);
+        ExpectIntEQ(out[0], 0x2a);
+        ExpectIntEQ(out[1], 3);
+        ExpectIntEQ(out[2], 4);
+        ExpectIntEQ(out[3], 5);
+
+        /* Test 2: combined first identifier needs multi-byte base-128
+         * encoding. "2.100" -> 40*2+100=180 -> 0x81 0x34 */
+        outSz = sizeof(out);
+        ExpectIntEQ(wc_EncodePolicyOID(out, &outSz, "2.100", NULL), 0);
+        ExpectIntEQ((int)outSz, 2);
+        ExpectIntEQ(out[0], 0x81);
+        ExpectIntEQ(out[1], 0x34);
+
+        /* Test 3: X=2 combined identifier well past 127, still valid
+         * per X.690 since X=2 has no upper bound on Y. "2.999" ->
+         * 40*2+999=1079 -> 0x88 0x37 */
+        outSz = sizeof(out);
+        ExpectIntEQ(wc_EncodePolicyOID(out, &outSz, "2.999", NULL), 0);
+        ExpectIntEQ((int)outSz, 2);
+        ExpectIntEQ(out[0], 0x88);
+        ExpectIntEQ(out[1], 0x37);
+
+        /* Test 4: X=0/X=1 with Y>=40 cannot round-trip (combined value
+         * would collide with the X=2 range) and must be rejected. */
+        outSz = sizeof(out);
+        ExpectIntEQ(wc_EncodePolicyOID(out, &outSz, "1.200", NULL),
+                    WC_NO_ERR_TRACE(ASN_OBJECT_ID_E));
+        outSz = sizeof(out);
+        ExpectIntEQ(wc_EncodePolicyOID(out, &outSz, "0.40", NULL),
+                    WC_NO_ERR_TRACE(ASN_OBJECT_ID_E));
+
+        /* Test 5: X=0/X=1 with Y just under 40 is the valid boundary */
+        outSz = sizeof(out);
+        ExpectIntEQ(wc_EncodePolicyOID(out, &outSz, "1.39", NULL), 0);
+        ExpectIntEQ((int)outSz, 1);
+        ExpectIntEQ(out[0], 40 + 39);
+
+        /* Test 5b: single-byte combined identifiers in [120,127] (arcs
+         * "2.40" through "2.47") are exactly the range where the old
+         * decoder's naive "val/40, val%40" formula (applied to a value
+         * it read as a single byte) produced a wrong-but-plausible
+         * result instead of failing loudly, so pin the encode side of
+         * that boundary here. "2.40.1" -> 40*2+40=120 (0x78), then 1. */
+        outSz = sizeof(out);
+        ExpectIntEQ(wc_EncodePolicyOID(out, &outSz, "2.40.1", NULL), 0);
+        ExpectIntEQ((int)outSz, 2);
+        ExpectIntEQ(out[0], 0x78);
+        ExpectIntEQ(out[1], 1);
+
+        /* Test 5c: "2.47.1" -> 40*2+47=127 (0x7F), the largest single-byte
+         * combined identifier, then 1. */
+        outSz = sizeof(out);
+        ExpectIntEQ(wc_EncodePolicyOID(out, &outSz, "2.47.1", NULL), 0);
+        ExpectIntEQ((int)outSz, 2);
+        ExpectIntEQ(out[0], 0x7F);
+        ExpectIntEQ(out[1], 1);
+
+        /* Test 5d: "2.48" -> 40*2+48=128, the smallest combined value that
+         * needs base-128 multi-byte encoding (0x81 0x00). The old encoder
+         * wrote this as a bare 0x80 - a continuation byte with nothing
+         * after it, i.e. undecodable DER - so this pins the direct
+         * regression the encoder fix addresses. */
+        outSz = sizeof(out);
+        ExpectIntEQ(wc_EncodePolicyOID(out, &outSz, "2.48", NULL), 0);
+        ExpectIntEQ((int)outSz, 2);
+        ExpectIntEQ(out[0], 0x81);
+        ExpectIntEQ(out[1], 0x00);
+
+        /* Test 6: first arc out of range */
+        outSz = sizeof(out);
+        ExpectIntEQ(wc_EncodePolicyOID(out, &outSz, "3.1", NULL),
+                    WC_NO_ERR_TRACE(ASN_OBJECT_ID_E));
+
+        /* Test 7: negative arc components must not wrap to a huge word32
+         * and slip past the range checks (e.g. "2.-1" wrapping firstArc +
+         * val back down to a small, valid-looking value). */
+        outSz = sizeof(out);
+        ExpectIntEQ(wc_EncodePolicyOID(out, &outSz, "2.-1", NULL),
+                    WC_NO_ERR_TRACE(ASN_OBJECT_ID_E));
+        outSz = sizeof(out);
+        ExpectIntEQ(wc_EncodePolicyOID(out, &outSz, "1.-5", NULL),
+                    WC_NO_ERR_TRACE(ASN_OBJECT_ID_E));
+        outSz = sizeof(out);
+        ExpectIntEQ(wc_EncodePolicyOID(out, &outSz, "-1.2", NULL),
+                    WC_NO_ERR_TRACE(ASN_OBJECT_ID_E));
+
+        /* Test 8: a later (non-combined) arc >= 128 encodes correctly
+         * with a large-enough buffer. "1.2.16384" -> combined id 42
+         * (0x2a) + arc 16384 -> 0x81 0x80 0x00 (same multi-byte value
+         * as Test 9 below, but here with room to check the actual
+         * bytes at the later-arc call site, not just BUFFER_E). */
+        outSz = sizeof(out);
+        ExpectIntEQ(wc_EncodePolicyOID(out, &outSz, "1.2.16384", NULL), 0);
+        ExpectIntEQ((int)outSz, 4);
+        ExpectIntEQ(out[0], 0x2a);
+        ExpectIntEQ(out[1], 0x81);
+        ExpectIntEQ(out[2], 0x80);
+        ExpectIntEQ(out[3], 0x00);
+
+        /* Test 9: BUFFER_E when the output buffer is too small for the
+         * combined first identifier's multi-byte encoding.
+         * "2.16304" -> 40*2+16304=16384 -> needs 3 bytes (0x81 0x80 0x00),
+         * outSz=2 is one byte short. */
+        outSz = 2;
+        ExpectIntEQ(wc_EncodePolicyOID(out, &outSz, "2.16304", NULL),
+                    WC_NO_ERR_TRACE(BUFFER_E));
+
+        /* Test 10: BUFFER_E when the output buffer is too small for a
+         * later (non-combined) arc's multi-byte encoding.
+         * "1.2.16384" -> combined id 42 (1 byte) + arc 16384 (3 bytes,
+         * same as above) -> needs 4 bytes total, outSz=3 is one short. */
+        outSz = 3;
+        ExpectIntEQ(wc_EncodePolicyOID(out, &outSz, "1.2.16384", NULL),
+                    WC_NO_ERR_TRACE(BUFFER_E));
+
+        /* Test 11: NULL/bad args */
+        outSz = sizeof(out);
+        ExpectIntEQ(wc_EncodePolicyOID(NULL, &outSz, "1.2.3", NULL),
+                    WC_NO_ERR_TRACE(BAD_FUNC_ARG));
+        ExpectIntEQ(wc_EncodePolicyOID(out, NULL, "1.2.3", NULL),
+                    WC_NO_ERR_TRACE(BAD_FUNC_ARG));
+        ExpectIntEQ(wc_EncodePolicyOID(out, &outSz, NULL, NULL),
+                    WC_NO_ERR_TRACE(BAD_FUNC_ARG));
+        outSz = 1; /* *outSz < 2 is rejected up front */
+        ExpectIntEQ(wc_EncodePolicyOID(out, &outSz, "1.2.3", NULL),
+                    WC_NO_ERR_TRACE(BAD_FUNC_ARG));
+
+        /* Test 12: fewer than the two arcs needed to form the combined
+         * first identifier must be rejected, not silently succeed with
+         * an empty (zero-length) output. */
+        outSz = sizeof(out);
+        ExpectIntEQ(wc_EncodePolicyOID(out, &outSz, "2", NULL),
+                    WC_NO_ERR_TRACE(ASN_OBJECT_ID_E));
+        outSz = sizeof(out);
+        ExpectIntEQ(wc_EncodePolicyOID(out, &outSz, "", NULL),
+                    WC_NO_ERR_TRACE(ASN_OBJECT_ID_E));
+
+        /* Test 13: non-digit characters anywhere in an arc, including
+         * trailing junk after otherwise-valid digits, must be rejected
+         * rather than silently truncated the way atoi() would. */
+        outSz = sizeof(out);
+        ExpectIntEQ(wc_EncodePolicyOID(out, &outSz, "1.2.abc", NULL),
+                    WC_NO_ERR_TRACE(ASN_OBJECT_ID_E));
+        outSz = sizeof(out);
+        ExpectIntEQ(wc_EncodePolicyOID(out, &outSz, "1.2.3junk", NULL),
+                    WC_NO_ERR_TRACE(ASN_OBJECT_ID_E));
+
+        /* Test 13b: empty arcs (leading/trailing/doubled '.') must be
+         * rejected. XSTRTOK collapses runs of the delimiter and skips
+         * leading/trailing ones, so without an explicit check these would
+         * silently encode identically to the non-empty-arc string. */
+        outSz = sizeof(out);
+        ExpectIntEQ(wc_EncodePolicyOID(out, &outSz, "1..2.3", NULL),
+                    WC_NO_ERR_TRACE(ASN_OBJECT_ID_E));
+        outSz = sizeof(out);
+        ExpectIntEQ(wc_EncodePolicyOID(out, &outSz, ".1.2.3", NULL),
+                    WC_NO_ERR_TRACE(ASN_OBJECT_ID_E));
+        outSz = sizeof(out);
+        ExpectIntEQ(wc_EncodePolicyOID(out, &outSz, "1.2.3.", NULL),
+                    WC_NO_ERR_TRACE(ASN_OBJECT_ID_E));
+
+        /* Test 14: an arc value that overflows a word32 must be rejected
+         * rather than silently wrapping. It's a well-formed decimal
+         * number, just too big, so it gets the distinct
+         * ASN_OID_ARC_TOO_BIG_E rather than ASN_OBJECT_ID_E - mirrors
+         * DecodeOidArc()'s same split on the decode side. */
+        outSz = sizeof(out);
+        ExpectIntEQ(wc_EncodePolicyOID(out, &outSz, "1.2.4294967296", NULL),
+                    WC_NO_ERR_TRACE(ASN_OID_ARC_TOO_BIG_E));
+
+        /* Test 14b: a Y value that fits in a word32 on its own but wraps
+         * the 40*X+Y sum for the combined first identifier (X=2) must
+         * also be rejected, not silently wrap to a small, wrong OID.
+         * Same "well-formed but too big" story as Test 14. */
+        outSz = sizeof(out);
+        ExpectIntEQ(wc_EncodePolicyOID(out, &outSz, "2.4294967295", NULL),
+                    WC_NO_ERR_TRACE(ASN_OID_ARC_TOO_BIG_E));
+
+        /* Test 15: round-trip through wc_DecodePolicyOID for a combined
+         * first identifier that needs multi-byte base-128 encoding -
+         * this is the same "2.100" value as Test 2, exercising the
+         * decode side's matching fix. */
+        {
+            char decoded[32];
+
+            outSz = sizeof(out);
+            ExpectIntEQ(wc_EncodePolicyOID(out, &outSz, "2.100", NULL), 0);
+            ExpectIntGT(wc_DecodePolicyOID(decoded, sizeof(decoded), out,
+                        outSz), 0);
+            ExpectStrEQ(decoded, "2.100");
+
+            /* and a case in the [80,120) gap where the old naive
+             * "val/40, val%40" decode formula happened to still be
+             * correct, alongside the multi-byte "2.100"/"2.999" cases
+             * above, at/above 120, where it was not. The single-byte
+             * [120,127] sub-range ("2.40" through "2.47") is covered
+             * separately by Tests 5b/5c above and the decode-side check
+             * right below.
+             * (wc_DecodePolicyOID requires inSz >= 2, so a trailing arc
+             * is added to get past the single combined-identifier
+             * byte that "2.5" alone would produce.) */
+            outSz = sizeof(out);
+            ExpectIntEQ(wc_EncodePolicyOID(out, &outSz, "2.5.4", NULL), 0);
+            ExpectIntGT(wc_DecodePolicyOID(decoded, sizeof(decoded), out,
+                        outSz), 0);
+            ExpectStrEQ(decoded, "2.5.4");
+
+            /* decode side of the single-byte [120,127] boundary: "2.40.1"
+             * -> 0x78 0x01. The old decoder read the combined identifier
+             * as a single byte and applied "val/40, val%40" unconditionally,
+             * giving the wrong "3.0.1"; the fix must give "2.40.1". */
+            {
+                const byte oid2_40[] = { 0x78, 0x01 };
+                ExpectIntGT(wc_DecodePolicyOID(decoded, sizeof(decoded),
+                            oid2_40, sizeof(oid2_40)), 0);
+                ExpectStrEQ(decoded, "2.40.1");
+            }
+
+            outSz = sizeof(out);
+            ExpectIntEQ(wc_EncodePolicyOID(out, &outSz, "2.999", NULL), 0);
+            ExpectIntGT(wc_DecodePolicyOID(decoded, sizeof(decoded), out,
+                        outSz), 0);
+            ExpectStrEQ(decoded, "2.999");
+
+            outSz = sizeof(out);
+            ExpectIntEQ(wc_EncodePolicyOID(out, &outSz, "1.2.3.4.5", NULL),
+                        0);
+            ExpectIntGT(wc_DecodePolicyOID(decoded, sizeof(decoded), out,
+                        outSz), 0);
+            ExpectStrEQ(decoded, "1.2.3.4.5");
+
+            /* Test 16: a combined first identifier truncated mid
+             * continuation sequence (every byte has the high bit set,
+             * no terminating byte) must be rejected, not decoded as
+             * whatever partial value had accumulated so far. */
+            {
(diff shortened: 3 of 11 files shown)


stenslae:
jenkins retest this please

wolfcrypt/src/asn.c
```diff
@@ -32515,12 +32514,41 @@ int wc_SetDatesBuffer(Cert* cert, const byte* der, int derSz)
...
+    /* push value in the right order */
+    while (i >= 0)
+        out[(*idx)++] = oid[i--];
+
+    return 0;
+}
+
 /* Encode OID string representation to ITU-T X.690 format */
-int EncodePolicyOID(byte *out, word32 *outSz, const char *in, void* heap)
+int wc_EncodePolicyOID(byte *out, word32 *outSz, const char *in, void* heap)
```
dgarske, line 33657: Do we need to or should we make similar fixes in `DecodePolicyOID()`? Perhaps your new test can use the wc_EncodePolicyOID and a new wc_PolicyDecodeOID?

wolfcrypt/src/asn.c
```diff
@@ -32540,61 +32568,60 @@ int EncodePolicyOID(byte *out, word32 *outSz, const char *in, void* heap)
     token = XSTRTOK(str, ".", &ptr);
     while (token != NULL)
     {
-        word32 val = (word32)XATOI(token);
+        int ival = XATOI(token);
```
dgarske: I'd prefer to replace the `XATOI` + `ival < 0` pair with an explicit bounded decimal parse that rejects any non-digit character, an empty token, and arithmetic overflow. Add tests for `"1.2.abc"`, `"1.2.3junk"`, and an overflowing arc.

wolfcrypt/src/asn.c
```diff
@@ -32540,61 +32568,60 @@ int EncodePolicyOID(byte *out, word32 *outSz, const char *in, void* heap)
...
+            if (ret != 0)
+                break;
         }
 
         token = XSTRTOK(NULL, ".", &ptr);
         nb_val++;
     }
 
-    *outSz = idx;
+    if (ret == 0)
```
dgarske, line 33740: The refactor now computes `firstArc = 40 * val` at `nb_val == 0` and defers emission to the `nb_val == 1` branch. If the input has only one arc (e.g. `"2"`), that branch never runs, so nothing is written, `*outSz` is set to 0 at line 32621, and the function returns success. 
Recommendation: Reject `nb_val < 2` after the loop with `ASN_OBJECT_ID_E` and add a test for a single-arc input such as `"2"` and for the empty string.

wolfcrypt/src/asn.c
```diff
@@ -32515,12 +32514,41 @@ int wc_SetDatesBuffer(Cert* cert, const byte* der, int derSz)
...
 #if (defined(WOLFSSL_CERT_GEN) && defined(WOLFSSL_CERT_EXT)) \
         || defined(OPENSSL_EXTRA)
+/* Base-128 (DER) encode a single arc value at out[*idx], advancing *idx.
+ * Shared by the combined first identifier and every later arc, since both
+ * need the same continuation-bit encoding once the value reaches 128. */
+static int EncodeOidArc(byte *out, word32 *idx, word32 outSz, word32 val)
+{
+    word32 tb = 0;
+    int    i = 0;
+    byte   oid[MAX_OID_SZ];
```
dgarske: MAX_OID_SZ can beuser overridden (default is 32). Consider just using 5 (the actual max)

wolfcrypt/src/asn.c
```diff
@@ -32540,61 +32568,60 @@ int EncodePolicyOID(byte *out, word32 *outSz, const char *in, void* heap)
...
-                XFREE(str, heap, DYNAMIC_TYPE_TMP_BUFFER);
-                return ASN_OBJECT_ID_E;
+                ret = ASN_OBJECT_ID_E;
+                break;
             }
 
-            out[idx] = (byte)(40 * val);
+            /* first two arcs combine into one X.690 identifier (40*X+Y);
+             * defer encoding until Y is known below. */
+            firstArc = 40 * val;
```
dgarske: Two small readability points in the new code:

1. `firstArc` holds `40 * X`, and the validity test is written as `if (firstArc < 80 && val >= 40)`. The reader has to divide by 40 in their head to see that `firstArc < 80` means "X is 0 or 1". Keeping the raw first arc in a variable and multiplying at the point of use makes the intent self-evident and removes the magic 80.

2. In `EncodeOidArc`, `tb` is a `word32` counter used purely as a boolean (`((tb++) ? 0x80 : 0)`), and it is always exactly equal to `i` at each use. The final `tb++` at line 32535 increments a value that is never read again. Testing `i` directly removes a variable and the dead increment.

Both are carried over in spirit from the original code, but the PR is rewriting these exact lines.

  Code:
```
    firstArc = 40 * val;
    ...
    if (firstArc < 80 && val >= 40) {
        ret = ASN_OBJECT_ID_E;
        break;
    }
    ret = EncodeOidArc(out, &idx, *outSz, firstArc + val);
```

Recommendation: Store the raw first arc (X) and multiply by 40 at the point of encoding, so the range test reads as `firstArc < 2`. Drop `tb` from `EncodeOidArc` and test `i` directly.

dgarske (changes requested):

wolfcrypt/src/asn.c
stenslae: Both applied. FirstArc now holds raw X (0/1/2), with the range check reading as firstArc < 2 and 40 * firstArc computed only at the encode call site. tb removed from EncodeOidArc and loop index i is tested directly instead.

wolfcrypt/src/asn.c
stenslae: Changed to byte oid[5]

wolfcrypt/src/asn.c
stenslae: Added a check after the loop rejecting nb_val < 2 with ASN_OBJECT_ID_E, plus tests for "2" and ""

wolfcrypt/src/asn.c
stenslae: I replaced XATOI+ival<0 with ParseOidArc(), which rejects empty tokens, any non-digit character, and word32 overflow. And added the requested tests.

wolfcrypt/src/asn.c
stenslae: Yes same issue on decode route, I added the DecodePolicyOID() fixes.

wolfcrypt/src/asn.c
```diff
@@ -21229,55 +21229,112 @@ static int DecodeNameConstraints(const byte* input, word32 sz,
...
+        /* write val as text into out */
+        w = XSNPRINTF(out + outIdx, outSz - outIdx, ".%u", val);
+        if (w < 0 || (word32)w >= outSz - outIdx) {
+            w = BUFFER_E;
+            goto exit;
         }
-        inIdx++;
+        outIdx += (word32)w;
     }
     if (outIdx == outSz)
```
stenslae: Fixed as suggested

tests/api/test_asn.c
```diff
@@ -2081,6 +2081,368 @@ int test_wc_DecodeObjectId(void)
...
+             * whatever partial value had accumulated so far. */
+            {
+                const byte truncated[] = { 0x81, 0x81 };
+                ExpectIntLT(DecodePolicyOID(decoded, sizeof(decoded),
+                            truncated, sizeof(truncated)), 0);
+            }
+
+            /* Test 17: reject over-long first identifier (cnt > 4) */
+            {
+                const byte overlong[] =
```
stenslae, line 3161: Fixed as suggested

wolfcrypt/src/asn.c
```diff
@@ -21229,58 +21229,114 @@ static int DecodeNameConstraints(const byte* input, word32 sz,
...
+ * tolerance for malformed data after the first identifier). */
+static int DecodeOidArc(const byte *in, word32 *idx, word32 inSz,
+        word32 *val, int strict)
+{
+    word32 v = 0;
+    int cnt = 0;
+
+    while (*idx < inSz) {
+        if (in[*idx] & 0x80) {
+            if (cnt == 0 && in[*idx] == 0x80)
```
dgarske, line 21997: The new `DecodeOidArc()` applies its non-minimal-encoding check (`cnt == 0 && in[*idx] == 0x80`) and its word32-overflow check (`v > (0xFFFFFFFFU >> 7)`) unconditionally, i.e. to *every* arc, not just the first identifier. The `strict` flag only changes what happens when the input ends mid-continuation. The old loop tolerated both of these for arcs after the first identifier: a leading `0x80` simply contributed 0, and an overflowing shift silently wrapped. Because `DecodeCertPolicy()` (asn.c:21443) treats any `<= 0` return as fatal and returns `ASN_PARSE_E`, a certificate carrying such a policy OID now fails to parse *entirely* where it previously parsed. I confirmed this against the built library: `{0x2a, 0x80, 0x03}` used to decode to "1.2.3" and now returns -144 (ASN_OBJECT_ID_E); `{0x2a, 0x90, 0x80, 0x80, 0x80, 0x00}` (arc == 2^32) used to render as ".0" and now returns -144. Rejecting is the more correct DER posture and is probably intended, but the PR body only describes an *encoder* fix plus "improved bounds checking", and the new doxygen note on `wc_DecodePolicyOID` only mentions the first-identifier strictness and the 2.40-2.47 rendering change.

Recommendation: Confirm the parse-side strictness is intended, then (a) extend the `wc_DecodePolicyOID` note to cover later arcs

tests/api/test_asn.c
```diff
@@ -2081,6 +2081,368 @@ int test_wc_DecodeObjectId(void)
...
+                        outSz), 0);
+            ExpectStrEQ(decoded, "1.2.3.4.5");
+
+            /* Test 16: a combined first identifier truncated mid
+             * continuation sequence (every byte has the high bit set,
+             * no terminating byte) must be rejected, not decoded as
+             * whatever partial value had accumulated so far. */
+            {
+                const byte truncated[] = { 0x81, 0x81 };
+                ExpectIntLT(DecodePolicyOID(decoded, sizeof(decoded),
```
dgarske: `DecodeOidArc()`'s `strict` parameter changes behavior in exactly one place: end-of-input mid-continuation returns `ASN_OBJECT_ID_E` when strict and `ASN_OID_ARC_TRUNCATED` (silently dropping the partial arc) when not. Test 16 covers only the strict side (`{0x81, 0x81}` as a first identifier). Nothing in the new suite exercises a truncated *trailing* arc, so the `if (ret == ASN_OID_ARC_TRUNCATED) break;` path in `wc_DecodePolicyOID` and the `return strict ? ... : ASN_OID_ARC_TRUNCATED` line are both untested. I verified the intended behavior against the built library: `{0x2a, 0x81}` returns 3 with "1.2". Similarly, the newly-strict later-arc paths described in the previous finding (non-minimal `{0x2a, 0x80, 0x03}` and overflow `{0x2a, 0x90, 0x80, 0x80, 0x80, 0x00}`, both -144) have no assertions, so a future relaxation of that behavior would pass CI silently. The `heap != NULL` path through `wc_EncodePolicyOID` (which is what every in-tree caller except the OpenSSL-compat layer uses) is also never exercised.

Recommendation: Add the truncated-trailing-arc, non-minimal-later-arc, overflow-later-arc, and max-word32-arc cases above, plus at least one `wc_EncodePolicyOID` call with a non-NULL `heap` so the XMALLOC/XFREE pairing on the heap-hint path is exercised.

wolfcrypt/src/asn.c
```diff
@@ -32564,20 +32570,79 @@ int wc_SetDatesBuffer(Cert* cert, const byte* der, int derSz)
...
 
     if (out == NULL || outSz == NULL || *outSz < 2 || in == NULL)
         return BAD_FUNC_ARG;
 
-    /* duplicate string (including terminator) */
+    /* reject empty arcs (leading/trailing/doubled '.') as XSTRTOK skips them */
     len = (word32)XSTRLEN(in);
+    if (len == 0 || in[0] == '.' || in[len - 1] == '.')
+        return ASN_OBJECT_ID_E;
+    if (XSTRSTR(in, "..") != NULL)
```
dgarske: The new empty-arc rejection calls `XSTRSTR(in, "..")`. Grepping the tree, this is the *only* `XSTRSTR` use in `wolfcrypt/src/*.c`, and `wolfssl/wolfcrypt/types.h:1021-1023` carries an explicit note directly above the definition: "strstr, strncmp, strcmp, and strncat only used by wolfSSL proper, not required for wolfCrypt only". Ports that build with `STRING_USER` and supply only the wolfCrypt-required X-macros will now fail to compile `wolfcrypt/src/asn.c` under `--enable-certgen --enable-certext`, a wolfCrypt-only configuration. The check is also redundant work: the function already walks the whole string in `ParseOidArc` per token, and the three pre-scans (`XSTRLEN`, `in[0]`, `in[len-1]`, `XSTRSTR`) can collapse into one pass.

Recommendation: Replace `XSTRSTR` with an inline scan so `wolfcrypt/src/asn.c` does not gain a dependency on a string routine that types.h documents as wolfSSL-proper-only.

wolfssl/wolfcrypt/asn.h
```diff
@@ -2516,10 +2507,15 @@ WOLFSSL_API int wc_SetUnknownExtCallbackEx(DecodedCert* cert,
...
-WOLFSSL_LOCAL int DecodePolicyOID(char *out, word32 outSz, const byte *in,
-                                  word32 inSz);
-WOLFSSL_LOCAL int EncodePolicyOID(byte *out, word32 *outSz,
-                                  const char *in, void* heap);
+WOLFSSL_TEST_VIS int wc_DecodePolicyOID(char *out, word32 outSz,
+                                         const byte *in, word32 inSz);
+WOLFSSL_TEST_VIS int wc_EncodePolicyOID(byte *out, word32 *outSz,
+                                         const char *in, void* heap);
+/* Deprecated public API names kept for backwards build compatibility */
+#define DecodePolicyOID(out, outSz, in, inSz) \
```
dgarske: We actually do not need the old macros because they were WOLFSSL_LOCAL. If anywhere else in our code base is using them we should rename those. If customers are using this in a static build (for example) they will need to update their code.

wolfcrypt/src/asn_orig.c
```diff
@@ -5547,7 +5547,7 @@ static int SetCertificatePolicies(byte *output,
         oidSz = sizeof(oid);
         XMEMSET(oid, 0, oidSz);
 
-        ret = EncodePolicyOID(oid, &oidSz, input[i], heap);
+        ret = wc_EncodePolicyOID(oid, &oidSz, input[i], heap);
```
dgarske, line 5588: The PR renamed `EncodePolicyOID` -> `wc_EncodePolicyOID` at `asn_orig.c:5550` but left `DecodePolicyOID` at `asn_orig.c:4227` on the old spelling, relying on the new header macro. Same file, same PR, two conventions. `asn_orig.c` is `#include`d into `asn.c` (asn.c:39546) so it compiles either way, but the inconsistency is the reason the compat macros cannot be deleted.

Recommendation: Rename this call site to `wc_DecodePolicyOID` to match the other in-tree call sites updated by this PR.

wolfcrypt/src/asn.c
```diff
@@ -21229,58 +21229,114 @@ static int DecodeNameConstraints(const byte* input, word32 sz,
 #if defined(WOLFSSL_CERT_EXT) || \
     defined(OPENSSL_EXTRA) || defined(OPENSSL_EXTRA_X509_SMALL)
 
+/* returned by DecodeOidArc() (in place of 0/ASN_OBJECT_ID_E) when the
+ * input ends mid-continuation with no terminating byte and strict==0 */
+#define ASN_OID_ARC_TRUNCATED 1
```
dgarske: Recommendation: Either rename out of the `ASN_*` namespace and `#undef` at the end of the guarded block, or replace the sentinel with an `int* truncated` out-parameter so the function keeps asn.c's usual 0/negative return contract.

src/ssl_asn1.c
```diff
@@ -5874,15 +5874,14 @@ int wc_OBJ_sn2nid(const char *sn)
...
         if (bufLen < MAX_OID_STRING_SZ) {
             bufSz = bufLen - 1;
         }
         else {
-            bufSz = MAX_OID_STRING_SZ - 1;
+            bufSz = MAX_OID_STRING_SZ;
         }
 
-        if ((bufSz = DecodePolicyOID(buf, (word32)bufSz, a->obj + idx,
+        if ((bufSz = wc_DecodePolicyOID(buf, (word32)bufSz, a->obj + idx,
```
dgarske, line 5887: The doc comment states "String is of the form ... and is always NUL terminated. Truncated when the buffer is too small." The PR tightens `wc_DecodePolicyOID`'s length checks from `w > outSz - outIdx` to `w >= outSz - outIdx` and adds a bound check on the first `XSNPRINTF` that had none, so the exact-fit case that previously produced a silently truncated (and therefore wrong) OID string now returns `BUFFER_E`, which this function maps to `WOLFSSL_FAILURE`. The stricter behavior is the right call - a silently shortened OID string is worse than an error, and the missing first-`XSNPRINTF` check was an out-of-bounds `out[outIdx] = 0` write when the first identifier alone overflowed the caller's buffer - but the comment now describes behavior the code no longer has, and it diverges from OpenSSL's `OBJ_obj2txt`, which truncates and returns the length that would have been needed.

  Code:
```
         * String is of the form "1.2.840.113549.1.9.1" and is always NUL
         * terminated. Truncated when the buffer is too small.
         ...
         * @return  0 when decoding the object fails.
```

Recommendation: Update the comment to say the call fails rather than truncates on a short buffer, and note the OpenSSL-compat divergence if it matters for callers of `wolfSSL_OBJ_obj2txt(buf, small, obj, 1)`.

dgarske (changes requested):

tests/api/test_asn.c
stenslae: Added suggested test coverage

wolfcrypt/src/asn_orig.c
stenslae: Changed to wc_

wolfcrypt/src/asn.c
stenslae: Used inline scan instead

wolfcrypt/src/asn.c
stenslae: Replaced the sentinel with an int* truncated out-parameter to maintain the standard 0/negative return contract.

wolfcrypt/src/asn.c
stenslae: Confirmed the parse-side strictness is intended and added fixes on the decode route.

wolfssl/wolfcrypt/asn.h
stenslae: The old macros have been removed and any remaining internal usages were renamed to use the wc_ prefix.

src/ssl_asn1.c
stenslae: I've updated the doc comment to explicitly state that the function will fail and return BUFFER_E when the buffer is too small. I also added a note regarding the divergence from OpenSSL's OBJ_obj2txt behavior.

wolfcrypt/src/asn.c
```diff
@@ -21229,64 +21229,117 @@ static int DecodeNameConstraints(const byte* input, word32 sz,
...
-        /* first bit is used to set if value is coded on 1 or multiple bytes */
-        if (in[inIdx] & 0x80) {
-            if (cnt == 4) {
-                w = ASN_OBJECT_ID_E;
-                goto exit;
-            }
-            val += in[inIdx] & 0x7F;
-            val <<= 7;
-            cnt++;
+        ret = DecodeOidArc(in, &inIdx, inSz, &val, 1, &truncated);
```
stenslae: All arcs now reject truncation with ASN_OBJECT_ID_E.

tests/api/test_ossl_obj.c
```diff
@@ -130,6 +130,15 @@ int test_wolfSSL_OBJ(void)
     ExpectIntEQ(OBJ_txt2nid(buf), NID_sha256);
 #endif
     ExpectIntGT(OBJ_obj2txt(buf, (int)sizeof(buf), obj, 0), 0);
+    /* Numeric form ("2.16.840.1.101.3.4.2.1", 22 chars) must fail cleanly
+     * rather than truncate when the caller's buffer is too small - unlike
+     * OpenSSL's OBJ_obj2txt(), which would truncate and report the needed
+     * length. */
+    {
+        char smallBuf[4];
+        ExpectIntEQ(OBJ_obj2txt(smallBuf, (int)sizeof(smallBuf), obj, 1),
```
stenslae, line 140: Added [5] and [8] cases to hit both bounds findings wanted

tests/api/test_asn.c
```diff
@@ -2081,6 +2081,449 @@ int test_wc_DecodeObjectId(void)
...
+            ExpectIntGT(wc_DecodePolicyOID(decoded, sizeof(decoded), out,
+                        outSz), 0);
+            ExpectStrEQ(decoded, "1.2.4294967295");
+
+            /* Test 26: exercise the non-NULL heap hint path, which is
+             * what every in-tree caller but the OpenSSL-compat layer
+             * uses, so the XMALLOC/XFREE pairing is covered. */
+            outSz = sizeof(out);
+            ExpectIntEQ(wc_EncodePolicyOID(out, &outSz, "1.2.840.113549.1",
+                        HEAP_HINT), 0);
```
stenslae, line 3265: HEAP_HINT is NULL here; dropped the false "non-NULL heap" claim from the comment

tests/api/test_ossl_obj.c
stenslae, line 140: Added [5] and [8] cases to hit both bounds findings wanted

tests/api/test_asn.c
```diff
@@ -2081,6 +2081,449 @@ int test_wc_DecodeObjectId(void)
...
+                ExpectIntEQ(wc_DecodePolicyOID(small, sizeof(small),
+                            firstIdBig, sizeof(firstIdBig)),
+                            WC_NO_ERR_TRACE(BUFFER_E));
+            }
+
+            /* Test 20: return BUFFER_E when string exact-fits outSz but lacks NUL space */
+            {
+                char buf4[4];
+                const byte oidExactFit[] = { 0x5A, 0x01 };
+                ExpectIntEQ(wc_DecodePolicyOID(buf4, sizeof(buf4),
```
stenslae, line 3191: Added suggested test

dgarske (dismissed):

src/ssl_asn1.c
```diff
@@ -5881,7 +5882,7 @@ int wc_OBJ_sn2nid(const char *sn)
             bufSz = MAX_OID_STRING_SZ;
         }
 
-        if ((bufSz = DecodePolicyOID(buf, (word32)bufSz, a->obj + idx,
+        if ((bufSz = wc_DecodePolicyOID(buf, (word32)bufSz, a->obj + idx,
```
stenslae, line 5887: Added suggested

tests/api/test_ossl_asn1.c
```diff
@@ -1085,7 +1085,10 @@ int test_wolfSSL_i2a_ASN1_OBJECT(void)
     /* DER encoding */
     p = notObjDer;
     ExpectNotNull(a = c2i_ASN1_OBJECT(NULL, &p, 3));
-    ExpectIntEQ(wolfSSL_i2a_ASN1_OBJECT(bio, a), 5);
+    /* notObjDer's content has a truncated trailing arc (0xff with no
+     * terminator), now rejected instead of silently dropped, so this
+     * falls through to the "<INVALID>" + hex dump path. */
+    ExpectIntEQ(wolfSSL_i2a_ASN1_OBJECT(bio, a), 70);
```
stenslae: Added suggested

wolfcrypt/src/asn.c
```diff
@@ -29087,9 +29131,9 @@ static int EncodeExtensions(Cert* cert, byte* output, word32 maxSz,
              int idx = CERTEXTSASN_IDX_START_CUSTOM + (i * 4);
              word32 encodedOidSz = MAX_OID_SZ;
              idx++; /* Skip one for for SEQ. */
-             /* EncodePolicyOID() will never return error since we parsed this
+             /* wc_EncodePolicyOID() will never return error since we parsed this
               * OID when it was set. */
-             EncodePolicyOID(&encodedOids[i * MAX_OID_SZ], &encodedOidSz,
+             wc_EncodePolicyOID(&encodedOids[i * MAX_OID_SZ], &encodedOidSz,
```
stenslae: Added suggested

wolfcrypt/src/asn.c
```diff
@@ -32514,86 +32558,155 @@ int wc_SetDatesBuffer(Cert* cert, const byte* der, int derSz)
...
+static int EncodeOidArc(byte *out, word32 *idx, word32 outSz, word32 val)
+{
+    int    i = 0;
+    byte   oid[WC_OID_ARC_MAX_BYTES];
+
+    while (val >= 128) {
+        word32 x = val % 128;
+        val /= 128;
+        /* set continuation bit */
+        oid[i] = (byte) ((i > 0 ? 0x80 : 0) | x);
```
stenslae, line 33619: Added suggestion

wolfcrypt/src/asn.c
```diff
@@ -21229,64 +21229,105 @@ static int DecodeNameConstraints(const byte* input, word32 sz,
...
+    int cnt = 0;
+
+    while (*idx < inSz) {
+        if (in[*idx] & 0x80) {
+            if (cnt == 0 && in[*idx] == 0x80)
+                return ASN_OBJECT_ID_E;
+            if (cnt == WC_OID_ARC_MAX_BYTES - 1)
+                return ASN_OBJECT_ID_E;
+            /* reject shift that overflows word32 (v is a multiple of 128 here) */
+            if (v > (0xFFFFFFFFU >> 7))
```
stenslae, line 22003: Allowed for wrapping

src/ssl_asn1.c
stenslae, line 5887: Applied suggested fix

tests/api/test_asn.c
```diff
@@ -2081,6 +2081,486 @@ int test_wc_DecodeObjectId(void)
...
+                /* inSz < 2 is rejected up front, even for a single-byte
+                 * input that would otherwise decode to "1.2". */
+                ExpectIntEQ(wc_DecodePolicyOID(decoded, sizeof(decoded),
+                            validIn, 1),
+                            WC_NO_ERR_TRACE(BAD_FUNC_ARG));
+                /* inSz >= ASN_LONG_LENGTH (0x80) is rejected up front;
+                 * content is irrelevant since the check runs before any
+                 * byte of "in" is read. */
+                ExpectIntEQ(wc_DecodePolicyOID(decoded, sizeof(decoded),
+                            validIn, 0x80),
```
stenslae: Applied suggested fix

tests/api/test_ossl_asn1.c
stenslae: Applied suggested fix

tests/api/test_ossl_obj.c
```diff
@@ -130,6 +130,34 @@ int test_wolfSSL_OBJ(void)
...
+    /* Numeric form ("2.16.840.1.101.3.4.2.1", 22 chars) must fail cleanly
+     * rather than truncate when the caller's buffer is too small - unlike
+     * OpenSSL's OBJ_obj2txt(), which would truncate and report the needed
+     * length. */
+    {
+        char smallBuf[4];
+        /* bufLen 4 fails the first-identifier length check inside
+         * wc_DecodePolicyOID: sha256's combined first identifier "2.16"
+         * needs 5 bytes (4 chars + NUL) and doesn't fit in bufSz = 4. */
+        ExpectIntEQ(OBJ_obj2txt(smallBuf, (int)sizeof(smallBuf), obj, 1),
```
stenslae, line 140: Added suggested

tests/api/test_asn.c
```diff
@@ -2081,6 +2081,559 @@ int test_wc_DecodeObjectId(void)
...
+         * on wc_DecodePolicyOID() directly. "2.100.3" encodes to
+         * { 0x81, 0x34, 0x03 }; flip the trailing arc's terminating byte
+         * to a continuation byte with nothing after it (same length, so
+         * no other offset in the cert shifts). */
+        if (EXPECT_SUCCESS()) {
+            int i;
+            int found = 0;
+            const byte oidBytes[] = { 0x81, 0x34, 0x03 };
+
+            for (i = 0; i + (int)sizeof(oidBytes) <= certSz; i++) {
```
stenslae, line 3381: Added suggested

src/x509.c
```diff
@@ -1358,7 +1358,7 @@ static int wolfssl_x509_add_custom_ext(WOLFSSL_X509 *x509,
 
     if (err == 0) {
         XMEMCPY(val, ext->value.data, ext->value.length);
-        if (wolfSSL_OBJ_obj2txt(oid, MAX_OID_STRING_SZ, ext->obj, 1) < 0) {
+        if (wolfSSL_OBJ_obj2txt(oid, MAX_OID_STRING_SZ, ext->obj, 1) <= 0) {
```
stenslae: Fixed as suggested

stenslae: Added suggested

tests/api/test_ossl_obj.c
stenslae, line 140: Added  suggested

wolfcrypt/src/asn.c
```diff
@@ -29191,10 +29152,20 @@ static int EncodeExtensions(Cert* cert, byte* output, word32 maxSz,
...
-             EncodePolicyOID(&encodedOids[i * MAX_OID_SZ], &encodedOidSz,
-                             cert->customCertExt[i].oid, NULL);
+             /* wc_SetCustomExtension() already parsed this OID with
+              * wc_EncodePolicyOID() when it was set, so this call is not
+              * expected to fail. Check anyway: on error, encodedOidSz is
+              * left untouched (still MAX_OID_SZ) and encodedOids is
+              * uninitialized, so encoding without checking would write
+              * uninitialized bytes into the certificate as the extension
+              * OID. */
+             ret = wc_EncodePolicyOID(&encodedOids[i * MAX_OID_SZ],
```
stenslae, line 30100: Added suggested

wolfcrypt/src/asn.c
```diff
@@ -21437,10 +21435,22 @@ static int DecodeCertPolicy(const byte* input, word32 sz, DecodedCert* cert)
...
+             * how other malformed extensions are handled in this file.
+             * An otherwise well-formed arc whose magnitude just doesn't
+             * fit a word32 (ASN_OID_ARC_TOO_BIG_E, e.g. a 2.25 UUID-based
+             * OID) is more lenient: skip only this policy entry rather
+             * than failing the whole certificate. */
+            int decodeRet = wc_DecodePolicyOID(
                     cert->extCertPolicies[cert->extCertPoliciesNb],
-                    MAX_CERTPOL_SZ, data, length) <= 0) {
+                    MAX_CERTPOL_SZ, data, length);
+            if (decodeRet == WC_NO_ERR_TRACE(ASN_OID_ARC_TOO_BIG_E)) {
```
stenslae: Intentionally lenient, added the decoded cert flag

wolfssl/wolfcrypt/asn_public.h
```diff
@@ -1012,6 +1017,10 @@ WOLFSSL_API int wc_GetDecodedCertIssuer(const struct DecodedCert* cert,
                                         char* buf, word32* bufSz);
 WOLFSSL_API int wc_GetDecodedCertSerial(const struct DecodedCert* cert,
                                         byte* buf, word32* bufSz);
+#ifdef WOLFSSL_CERT_EXT
+WOLFSSL_API int wc_CertGetPoliciesTruncated(struct DecodedCert* cert,
```
stenslae: Fixed as suggested

tests/unit-mcdc/test_asn_ext_whitebox.c
```diff
@@ -1063,38 +1063,38 @@ static void wb_decode_policy_oid(void)
...
-    WB_CHECK(DecodePolicyOID(out, sizeof(out), oidBytes, 1)
+    WB_CHECK(wc_DecodePolicyOID(out, sizeof(out), oidBytes, 1)
                 == WC_NO_ERR_TRACE(BAD_FUNC_ARG), "inSz<2");
-    ret = DecodePolicyOID(out, sizeof(out), oidBytes, sizeof(oidBytes));
+    ret = wc_DecodePolicyOID(out, sizeof(out), oidBytes, sizeof(oidBytes));
     WB_CHECK(ret > 0 && strcmp(out, "2.5.29.32.0") == 0,
             "all args valid (all 4 operands false)");
 
-    WB_NOTE("DecodePolicyOID(): output buffer overflow guard [:21272]");
+    WB_NOTE("wc_DecodePolicyOID(): output buffer overflow guard [:21369]");
```
stenslae: Fixed as suggested

tests/api/test_asn.c
```diff
@@ -2081,6 +2081,582 @@ int test_wc_DecodeObjectId(void)
...
+                const byte oidTailExact[] = { 0x2a, 0x22 };
+                ExpectIntEQ(wc_DecodePolicyOID(buf6, sizeof(buf6),
+                            oidTailExact, sizeof(oidTailExact)),
+                            WC_NO_ERR_TRACE(BUFFER_E));
+            }
+
+            /* Test 21: reject non-minimal base-128 first identifier */
+            {
+                const byte nonMinimal[] = { 0x80, 0x2a };
+                ExpectIntLT(wc_DecodePolicyOID(decoded, sizeof(decoded),
```
stenslae: Fixed as suggested

wolfcrypt/src/asn.c
```diff
@@ -21437,10 +21490,23 @@ static int DecodeCertPolicy(const byte* input, word32 sz, DecodedCert* cert)
...
+                WOLFSSL_MSG("\tSkipping policy OID with an oversized arc");
+                skipPolicy = 1;
+                cert->extCertPoliciesTruncated = 1;
+            }
+            else if (decodeRet <= 0) {
```
stenslae: Fixed as suggested

wolfcrypt/src/asn.c
```diff
@@ -21452,7 +21518,8 @@ static int DecodeCertPolicy(const byte* input, word32 sz, DecodedCert* cert)
          * extension". This is a sanity check for duplicates.
          * extCertPolicies should only have OID values, additional
          * qualifiers need to be stored in a separate array. */
-        for (i = 0; (ret == 0) && (i < cert->extCertPoliciesNb); i++) {
+        for (i = 0; (ret == 0) && !skipPolicy && (i < cert->extCertPoliciesNb);
```
stenslae, line 22200: Fixed as suggested

tests/api/test_asn.c
```diff
@@ -2081,6 +2081,540 @@ int test_wc_DecodeObjectId(void)
...
+                    break;
+                }
+            }
+            ExpectIntEQ(found, 1);
+
+            XMEMSET(&decoded, 0, sizeof(decoded));
+            wc_InitDecodedCert(&decoded, der, (word32)certSz, HEAP_HINT);
+            ExpectIntEQ(wc_ParseCert(&decoded, CERT_TYPE, NO_VERIFY, NULL),
+                    0);
+            ExpectIntEQ(decoded.extCertPoliciesNb, 1);
```
stenslae, line 3445: Added suggested fix

tests/api/test_asn.c
```diff
@@ -2081,6 +2081,540 @@ int test_wc_DecodeObjectId(void)
     return EXPECT_RESULT();
 }
 
+int test_wc_EncodePolicyOID(void)
+{
+    EXPECT_DECLS;
+
+#if (defined(WOLFSSL_CERT_GEN) && defined(WOLFSSL_CERT_EXT)) \
```
stenslae: Fix already applied

wolfcrypt/src/asn.c
```diff
@@ -21437,10 +21486,28 @@ static int DecodeCertPolicy(const byte* input, word32 sz, DecodedCert* cert)
...
+             * for the buffer (BUFFER_E) only skip this policy. */
+            decodeRet = wc_DecodePolicyOID(
                     cert->extCertPolicies[cert->extCertPoliciesNb],
-                    MAX_CERTPOL_SZ, data, length) <= 0) {
+                    MAX_CERTPOL_SZ, data, length);
+            if ((decodeRet == WC_NO_ERR_TRACE(ASN_OID_ARC_TOO_BIG_E)) ||
+                    (decodeRet == WC_NO_ERR_TRACE(BUFFER_E))) {
+                WOLFSSL_MSG("\tSkipping policy OID that doesn't fit");
+                skipPolicy = 1;
+                cert->extCertPoliciesTruncated = 1;
```
stenslae: Fix applied

tests/unit-mcdc/test_asn_ext_whitebox.c
```diff
@@ -1321,26 +1330,73 @@ static void wb_decode_policy_oid(void)
          * never a negative value for these arguments. */
         char tiny6[6];
         char tiny5[5];
+        char tiny4[4];
+        static const byte firstIdBig[] = { 0x7F, 0x01 };
 
-        ret = DecodePolicyOID(tiny6, sizeof(tiny6), oidBytes,
+        ret = wc_DecodePolicyOID(tiny6, sizeof(tiny6), oidBytes,
                 sizeof(oidBytes));
         WB_CHECK(ret > 0, ":21326 2nd operand false (segment fits exactly)");
```
stenslae: Applied suggested fix

src/internal.c
```diff
@@ -15593,6 +15593,7 @@ int CopyDecodedToX509(WOLFSSL_X509* x509, DecodedCert* dCert)
                 XMEMCPY(x509->certPolicies[i], dCert->extCertPolicies[i],
                                                                 MAX_CERTPOL_SZ);
             x509->certPoliciesNb = dCert->extCertPoliciesNb;
+            x509->certPoliciesTruncated = dCert->extCertPoliciesTruncated;
```
stenslae: Applied suggested fix of guard

wolfcrypt/src/asn.c
```diff
@@ -28700,7 +28765,7 @@ static int SetCertificatePolicies(byte *output,
         XMEMSET(oid, 0, oidSz);
         dataASN[POLICYINFOASN_IDX_QUALI].noOut = 1;
 
-        ret = EncodePolicyOID(oid, &oidSz, input[i], heap);
+        ret = wc_EncodePolicyOID(oid, &oidSz, input[i], heap);
         if (ret == 0) {
             XMEMSET(dataASN, 0, sizeof(dataASN));
```
stenslae, line 28819: Applied suggested fix

tests/api/test_ossl_asn1.c
```diff
@@ -1085,7 +1085,23 @@ int test_wolfSSL_i2a_ASN1_OBJECT(void)
...
+     * on the marker rather than a byte count, which also covers the hex
+     * dump and so varies with NO_FILESYSTEM. */
+    {
+        const char* invalid = "<INVALID>";
+        int invalidLen = (int)XSTRLEN(invalid);
+        BIO* invBio = NULL;
+        char marker[16];
+
+        ExpectTrue((invBio = BIO_new(BIO_s_mem())) != NULL);
+        ExpectIntGE(wolfSSL_i2a_ASN1_OBJECT(invBio, a), invalidLen);
```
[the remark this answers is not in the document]
stenslae: Applied suggested fix

src/ssl_asn1.c
```diff
@@ -5876,17 +5903,11 @@ int wc_OBJ_sn2nid(const char *sn)
...
-        if (bufLen < MAX_OID_STRING_SZ) {
-            bufSz = bufLen - 1;
-        }
-        else {
-            bufSz = MAX_OID_STRING_SZ - 1;
-        }
-
-        if ((bufSz = DecodePolicyOID(buf, (word32)bufSz, a->obj + idx,
+        /* DecodePolicyOID() accounts for the NUL terminator itself. */
+        if ((bufSz = DecodePolicyOID(buf, (word32)bufLen, a->obj + idx,
```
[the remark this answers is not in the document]
stenslae: Applied suggested fix

wolfcrypt/src/asn.c
```diff
@@ -22168,9 +22203,26 @@ static int DecodeCertPolicy(const byte* input, word32 sz, DecodedCert* cert)
...
+    #ifdef WOLFSSL_CERT_EXT
+        if (ret == 0) {
+            /* clear the slot for duplicate check */
+            XMEMSET(cert->extCertPolicies[cert->extCertPoliciesNb], 0,
+                    MAX_CERTPOL_SZ);
+
+            /* drop policy on failure */
+            if (DecodePolicyOID(
+                    cert->extCertPolicies[cert->extCertPoliciesNb],
+                    MAX_CERTPOL_SZ, data, length) <= 0) {
```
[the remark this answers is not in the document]
stenslae: Applied suggested fix

src/x509.c
```diff
@@ -7942,8 +7987,13 @@ static int X509PrintReqAttributes(WOLFSSL_BIO* bio, WOLFSSL_X509* x509,
     do {
         attr = wolfSSL_X509_REQ_get_attr(x509, i);
         if (attr != NULL) {
-            char lName[NAME_SZ/4]; /* NAME_SZ default is 80 */
-            int lNameSz = NAME_SZ/4;
+            /* Sized for a numeric-form OID, not the NAME_SZ/4 column width:
+             * wolfSSL_OBJ_obj2txt() fails rather than truncates when the
+             * name does not fit, and an attribute OID outside the object
+             * table prints in dotted-decimal form. */
+            char lName[MAX_OID_STRING_SZ];
```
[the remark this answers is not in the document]
stenslae: Applied suggested fix in PR #11418

PR #11021, opened 2026-07-30
Remove carry over of old buffers/sizes in X509 ext function.

kareem-wolfssl:
# Description

Fixes zd#22223

# Testing

Built in tests, added tests, provided reproducer

# Checklist

 - [X] added tests
 - [ ] updated/added doxygen


wolfcrypt/src/evp.c
```diff
@@ -9840,25 +9847,17 @@ static int ECC_populate_EVP_PKEY(WOLFSSL_EVP_PKEY* pkey, WOLFSSL_EC_KEY *key)
     }
     else if (ecc->type == ECC_PUBLICKEY) {
```
dgarske, line 9985: **🟠 [Medium] ECC_populate_EVP_PKEY now keeps the previous key's cached DER on the serialization error path**

The rewritten `ECC_PUBLICKEY` branch changes the error-path semantics. Previously, when `wc_EccPublicKeyToDer()` failed the code did `pkey->pkey.ptr = NULL; pkey->pkey_sz = 0;`, so the EVP_PKEY was left with no cached DER. After this change, on that same failure only the newly allocated `derBuf` is freed and the *old* buffer/size are left in place. `wolfSSL_EVP_PKEY_set1_EC_KEY()` calls `clearEVPPkeyKeys()` (which does not touch `pkey->pkey.ptr`) and installs `pkey->ecc = key` *before* calling this function, so on failure the EVP_PKEY ends up with the NEW EC key but the OLD key's DER still cached in `pkey->pkey.ptr`/`pkey->pkey_sz`. Any caller that ignores the `WOLFSSL_FAILURE` return (or that later reads the cached DER via `wolfSSL_EVP_PKEY_get_der()`, `i2d_*`, `EVP_PKEY_cmp()`, `wolfSSL_CTX_use_PrivateKey()`) then silently operates on a different key rather than getting an error. It is not a memory-safety issue (size still matches the retained buffer), but it is exactly the stale-carry-over class of state this PR sets out to eliminate. Note also that this block does not reset `pkey->pkcs8HeaderSz` (pre-existing), so a PKCS#8-wrapped key followed by a public-only EC key leaves a stale non-zero header size next to a DER that has no PKCS#8 wrapper - worth handling while the block is being rewritten.

**Fix:** Free and NULL `pkey->pkey.ptr` (and zero `pkey->pkey_sz`) before/when the serialization fails so a failed `EVP_PKEY_set1_EC_KEY()` cannot leave a DER that decodes to the previously installed key. Mirrors the `pkey->pkey_sz = 0;` that was added to `PopulateRSAEvpPkeyDer()` in this same PR.

certs/test/catalog.txt
```diff
@@ -12,3 +12,9 @@ digsigku.pem:
   set.
 cert-ext-joi.pem:
   Simple certificate that includes OIDs for JurisdictionOfIncorporation
+cert-ext-oid-collide.der:
```
dgarske, line 15: **🔵 [Low] New test certificate has no generation recipe alongside the other cert-ext-* certs**

Every other `certs/test/cert-ext-*` artifact is produced by `certs/test/gen-ext-certs.sh` from a committed `.cfg`. `cert-ext-oid-collide.der` is committed as an opaque blob with no recipe; `catalog.txt` documents what it is (a 4-byte OID that hashes to the prime256v1 NID, invalid signature, parse-only) but not how to rebuild it. If `wc_oid_sum()` ever changes, no one can regenerate a colliding OID without redoing the analysis. (The certificate itself is fine for the test: `wolfSSL_X509_load_certificate_file()` parses with `NO_VERIFY`, so neither the bogus signature nor the 2029 notAfter matters.)

**Fix:** Add a generation recipe or document how the colliding OID was derived so the artifact is reproducible.

dgarske (commented):
## Skoll Code Review

### Posted findings
- [Medium] **ECC_populate_EVP_PKEY now keeps the previous key's cached DER on the serialization error path** — `wolfcrypt/src/evp.c:9848-9868`
- [Low] **New test certificate has no generation recipe alongside the other cert-ext-* certs** — `certs/test/catalog.txt:15-20`

### Skipped findings
- [Medium] ~~New EC test arm is gated on the wrong curve macros (P-256 arm guarded by ECC_MIN_KEY_SZ = 521 / HAVE_ECC256)~~
- [Medium] ~~X509 OID-collision test asserts a value that also holds when the collision does not occur~~
- [Low] ~~Unused pem.h include added to tests/api/test_evp.c~~
- [Low] ~~Inconsistent NULL-check-before-XFREE across the three sites touched by this PR~~
- [Info] ~~Macro continuation formatting in TEST_OSSL_X509_EXT_DECLS~~

dgarske (changes requested):
https://github.com/wolfSSL/wolfssl/pull/11021#pullrequestreview-4860106678

kareem-wolfssl:
https://github.com/wolfSSL/wolfssl/pull/11061 fixed the EVP issue originally addressed by this PR, so the EVP changes were dropped.  I left test_wolfSSL_CTX_use_PrivateKey_pkcs8_repopulate in as this covers a unique test case not covered by #11061.

kareem-wolfssl:
Retest this please

dgarske (approved):

JacobBarthelmeh (approved):

PR #11025, opened 2026-07-31
Fix: Test memory leak and Failing test when md5 and sha are enabled

aidankeefe2022:
# Description

Fixed a memory leak where if _ret is already failing, the rngInit function will not run, causing rngFree to free garbage data. The fix was to have rngInit always run to match with rngFree by swapping ExpectInt to DoExpectInt. Fixed failing test where WC_SIG_MIN_HASH_TYPE was not properly set for --enable-sha and --enable-md5, so we skip under those conditions.

Found while testing wolfCLU

# Testing

Code no longer leaks, and code no longer fails with the wrong code for the test when SHA and MD5 are enabled

### DIFF
--- a/tests/api/test_signature.c
+++ b/tests/api/test_signature.c
@@ -41,13 +41,6 @@
 #include <tests/api/api.h>
 #include <tests/api/test_signature.h>
 
-/* Must be kept in sync with wolfcrypt/src/signature.c */
-#ifndef WC_SIG_MIN_HASH_TYPE
-    #define TEST_SIG_MIN_HASH_TYPE WC_HASH_TYPE_SHA256
-#else
-    #define TEST_SIG_MIN_HASH_TYPE WC_SIG_MIN_HASH_TYPE
-#endif
-
 /* Testing wc_SignatureGetSize() for signature type ECC */
 int test_wc_SignatureGetSize_ecc(void)
 {
--- a/tests/api/test_signature.c
+++ b/tests/api/test_signature.c
@@ -41,6 +41,13 @@
 #include <tests/api/api.h>
 #include <tests/api/test_signature.h>
 
+/* Must be kept in sync with wolfcrypt/src/signature.c */
+#ifndef WC_SIG_MIN_HASH_TYPE
+    #define TEST_SIG_MIN_HASH_TYPE WC_HASH_TYPE_SHA256
+#else
+    #define TEST_SIG_MIN_HASH_TYPE WC_SIG_MIN_HASH_TYPE
+#endif
+
 /* Testing wc_SignatureGetSize() for signature type ECC */
 int test_wc_SignatureGetSize_ecc(void)
 {
--- a/tests/api/test_signature.c
+++ b/tests/api/test_signature.c
@@ -182,7 +182,7 @@ int test_wc_falcon_sign_verify(void)
     XMEMSET(&key, 0, sizeof(key));
     ExpectIntEQ(wc_falcon_init(&key), 0);
     ExpectIntEQ(wc_falcon_set_level(&key, 1), 0);
-    ExpectIntEQ(wc_InitRng(&rng), 0);
+    DoExpectIntEQ(wc_InitRng(&rng), 0);
 
     /* Use the embedded benchmark key rather than generating one: this is a
      * wrapper-level smoke test (the native suite lives in test_falcon.c), so
@@ -304,7 +304,7 @@ int test_wc_SignatureDecisionCoverage(void)
             WC_RNG rng;
 
             XMEMSET(data, 0x5A, sizeof(data));
-            ExpectIntEQ(wc_InitRng(&rng), 0);
+            DoExpectIntEQ(wc_InitRng(&rng), 0);
 
             ExpectIntEQ(wc_SignatureVerify(WC_HASH_TYPE_SHA256, sig_type,
                 NULL, data_len, sig, sig_len, &ecc, key_len),
@@ -401,7 +401,7 @@ int test_wc_SignatureDecisionCoverage(void)
             word32 genSigLen = (word32)eccSigMax;
             WC_RNG rng;
 
-            ExpectIntEQ(wc_InitRng(&rng), 0);
+            DoExpectIntEQ(wc_InitRng(&rng), 0);
 
             ExpectIntEQ(wc_SignatureGenerateHash_ex(WC_HASH_TYPE_SHA256,
                 sig_type, NULL, hash_len, genSig, &genSigLen, &ecc, key_len,
@@ -478,7 +478,7 @@ int test_wc_SignatureDecisionCoverage(void)
             WC_RNG rng;
 
             XMEMSET(data, 0xA5, sizeof(data));
-            ExpectIntEQ(wc_InitRng(&rng), 0);
+            DoExpectIntEQ(wc_InitRng(&rng), 0);
 
             ExpectIntEQ(wc_SignatureGenerate(WC_HASH_TYPE_SHA256, sig_type,
                 NULL, data_len, genSig, &genSigLen, &ecc, key_len, &rng),
@@ -639,7 +639,7 @@ int test_wc_SignatureDecisionCoverage(void)
 
             XMEMSET(data, 0x24, sizeof(data));
             XMEMSET(otherData, 0x99, sizeof(otherData));
-            ExpectIntEQ(wc_InitRng(&rng), 0);
+            DoExpectIntEQ(wc_InitRng(&rng), 0);
             ExpectIntEQ(wc_SignatureGenerate(WC_HASH_TYPE_SHA256,
                 WC_SIGNATURE_TYPE_RSA, data, data_len, realSig, &realSigLen,
                 &rsa_key, key_len, &rng), 0);
@@ -688,7 +688,7 @@ int test_wc_SignatureFeatureCoverage(void)
 
         ExpectIntEQ(wc_ecc_init(&ecc), 0);
         ExpectIntEQ(wc_ecc_import_raw(&ecc, qx, qy, d, "SECP256R1"), 0);
-        ExpectIntEQ(wc_InitRng(&rng), 0);
+        DoExpectIntEQ(wc_InitRng(&rng), 0);
 
         /* wc_SignatureGetSize(): positive path for ECC */
         ExpectIntGT(wc_SignatureGetSize(sig_type, &ecc, key_len), 0);
@@ -774,7 +774,7 @@ int test_wc_SignatureFeatureCoverage(void)
         ExpectIntEQ(wc_InitRsaKey_ex(&rsa_key, HEAP_HINT, testDevId), 0);
         ExpectIntEQ(wc_RsaPrivateKeyDecode(tmp, &idx, &rsa_key,
             (word32)bytes), 0);
-        ExpectIntEQ(wc_InitRng(&rng), 0);
+        DoExpectIntEQ(wc_InitRng(&rng), 0);
 
         /* wc_SignatureGetSize(): positive path for RSA */
         ExpectIntGT(wc_SignatureGetSize(WC_SIGNATURE_TYPE_RSA, &rsa_key,
--- a/tests/api/test_signature.c
+++ b/tests/api/test_signature.c
@@ -180,9 +180,10 @@ int test_wc_falcon_sign_verify(void)
     static const byte msg[] = "wolfssl falcon coverage";
 
     XMEMSET(&key, 0, sizeof(key));
+    XMEMSET(&rng, 0, sizeof(rng));
     ExpectIntEQ(wc_falcon_init(&key), 0);
     ExpectIntEQ(wc_falcon_set_level(&key, 1), 0);
-    DoExpectIntEQ(wc_InitRng(&rng), 0);
+    ExpectIntEQ(wc_InitRng(&rng), 0);
 
     /* Use the embedded benchmark key rather than generating one: this is a
      * wrapper-level smoke test (the native suite lives in test_falcon.c), so
@@ -304,7 +305,8 @@ int test_wc_SignatureDecisionCoverage(void)
             WC_RNG rng;
 
             XMEMSET(data, 0x5A, sizeof(data));
-            DoExpectIntEQ(wc_InitRng(&rng), 0);
+            XMEMSET(&rng, 0, sizeof(rng));
+            ExpectIntEQ(wc_InitRng(&rng), 0);
 
             ExpectIntEQ(wc_SignatureVerify(WC_HASH_TYPE_SHA256, sig_type,
                 NULL, data_len, sig, sig_len, &ecc, key_len),
@@ -401,7 +403,8 @@ int test_wc_SignatureDecisionCoverage(void)
             word32 genSigLen = (word32)eccSigMax;
             WC_RNG rng;
 
-            DoExpectIntEQ(wc_InitRng(&rng), 0);
+            XMEMSET(&rng, 0, sizeof(rng));
+            ExpectIntEQ(wc_InitRng(&rng), 0);
 
             ExpectIntEQ(wc_SignatureGenerateHash_ex(WC_HASH_TYPE_SHA256,
                 sig_type, NULL, hash_len, genSig, &genSigLen, &ecc, key_len,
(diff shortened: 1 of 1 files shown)


tests/api/test_signature.c
```diff
@@ -335,9 +335,11 @@ int test_wc_SignatureDecisionCoverage(void)
                 data, data_len, sig, sig_len, &ecc, key_len),
                 WC_NO_ERR_TRACE(BAD_FUNC_ARG));
 
-#ifndef NO_SHA
-            /* Hash weaker than WC_SIG_MIN_HASH_TYPE (default SHA-256)
-             * rejected by wc_SignatureCheckHashStrength() */
+#if !defined(NO_SHA) && !defined(WC_SIG_MIN_HASH_TYPE)
```
Frauschi: **🟡 [Medium] Weak-hash guard is skipped even when WC_SIG_MIN_HASH_TYPE raises the floor, and silently self-destructs if the default ever moves into a header**
*💡 SUGGEST*  `test`

The new guard `#if !defined(NO_SHA) && !defined(WC_SIG_MIN_HASH_TYPE)` drops the SHA-1-rejection assertion whenever the macro is defined *at all*, not just when it is defined to something weaker than SHA-256. If a downstream config sets `WC_SIG_MIN_HASH_TYPE=WC_HASH_TYPE_SHA384` (a stricter floor), `wc_SignatureCheckHashStrength()` still rejects SHA-1 with `BAD_FUNC_ARG`, but the test is skipped anyway — coverage is lost for no reason.

There is a second, more fragile coupling. The default `#ifndef WC_SIG_MIN_HASH_TYPE / #define WC_SIG_MIN_HASH_TYPE WC_HASH_TYPE_SHA256 / #endif` currently lives in `wolfcrypt/src/signature.c:62-64`, *not* in `wolfssl/wolfcrypt/signature.h`. The test only compiles the assertion because that default is invisible to this translation unit. The moment anyone moves that default into `signature.h` (a natural refactor — `signature.h` is included at line 31 of this file, before the guard), `WC_SIG_MIN_HASH_TYPE` becomes defined in every build and both assertion blocks become permanently dead code, with no compile error and no test failure to signal it.

A size comparison against the configured floor keeps coverage in every config and removes the dependence on where the default happens to be defined.

**Suggestion:**
```suggestion
/* Near the top of the file, mirroring wolfcrypt/src/signature.c: */
#ifndef WC_SIG_MIN_HASH_TYPE
    #define TEST_SIG_MIN_HASH_TYPE WC_HASH_TYPE_SHA256
#else
    #define TEST_SIG_MIN_HASH_TYPE WC_SIG_MIN_HASH_TYPE
#endif

/* At the call site: */
#ifndef NO_SHA
            /* Hash weaker than WC_SIG_MIN_HASH_TYPE is rejected by
             * wc_SignatureCheckHashStrength(). wolfCLU with MD5 compiled in
             * lowers the floor to MD5, which lets SHA-1 through. */
            if (wc_HashGetDigestSize(TEST_SIG_MIN_HASH_TYPE) >
                    WC_SHA_DIGEST_SIZE) {
                ExpectIntEQ(wc_SignatureVerify(WC_HASH_TYPE_SHA, sig_type,
                    data, data_len, sig, sig_len, &ecc, key_len),
                    WC_NO_ERR_TRACE(BAD_FUNC_ARG));
            }
#endif
```

Frauschi: **🟡 [Medium] Lowered-floor branch of wc_SignatureCheckHashStrength() has no coverage at all**
*💡 SUGGEST*  `test`

Both new guards are bare `#if ... #endif` with no `#else`. In exactly the configuration that motivated this PR (wolfCLU with MD5, which sets `-DWC_SIG_MIN_HASH_TYPE=WC_HASH_TYPE_MD5` at `configure.ac:12052`), the test now asserts *nothing* about hash-strength enforcement — it does not verify that SHA-1 is accepted under the lowered floor, only that it is no longer checked.

That matters because this file is explicitly a decision-coverage suite (see the header comment at lines 209-227 about independence pairs). Silently skipping the only assertion that touches `wc_SignatureCheckHashStrength()` means a regression that made the floor unconditionally SHA-256 — breaking wolfCLU's MD5 support, which is precisely the behaviour this override exists to preserve — would not be caught by any test.

**Suggestion:**
```suggestion
#if !defined(NO_SHA) && !defined(WC_SIG_MIN_HASH_TYPE)
            ExpectIntEQ(wc_SignatureVerify(WC_HASH_TYPE_SHA, sig_type,
                data, data_len, sig, sig_len, &ecc, key_len),
                WC_NO_ERR_TRACE(BAD_FUNC_ARG));
#elif !defined(NO_SHA)
            /* Floor lowered (e.g. wolfCLU with MD5): SHA-1 must get past the
             * strength check and fail on the signature itself instead. */
            ExpectIntNE(wc_SignatureVerify(WC_HASH_TYPE_SHA, sig_type,
                data, data_len, sig, sig_len, &ecc, key_len),
                WC_NO_ERR_TRACE(BAD_FUNC_ARG));
#endif
```

Frauschi (commented):
## 🐺 Skoll Code Review

**Findings:** 6 total — 2 posted, 4 skipped

### Posted findings
- [Medium] **Weak-hash guard is skipped even when WC_SIG_MIN_HASH_TYPE raises the floor, and silently self-destructs if the default ever moves into a header** — `tests/api/test_signature.c:338`
- [Medium] **Lowered-floor branch of wc_SignatureCheckHashStrength() has no coverage at all** — `tests/api/test_signature.c:338`

- [Low] PR description attributes the failure to --enable-sha/--enable-md5, but the trigger is --enable-wolfclu
- [Low] Two different init/free-pairing conventions now coexist in this file
- [Low] Weak-hash assertions skipped for any WC_SIG_MIN_HASH_TYPE override, including stronger floors
- [Info] Guard depends on WC_SIG_MIN_HASH_TYPE's default living in signature.c, not a public header

</details>

tests/api/test_signature.c
aidankeefe2022: merged the two issues into one fix

aidankeefe2022: merged the two issues into one fix

aidankeefe2022:
Retest this please

Frauschi (commented):
## 🐺 Skoll Code Review

### Posted findings
- [Medium] **Lowered-floor else branch at the generate site calls wc_SignatureVerify() instead of wc_SignatureGenerate()** — `tests/api/test_signature.c:538-544`
- [Low] **Guard compares wc_HashType enum values; the library compares digest sizes, and the two disagree in real configs** — `tests/api/test_signature.c:345,532`

tests/api/test_signature.c
```diff
@@ -516,15 +528,20 @@ int test_wc_SignatureDecisionCoverage(void)
...
+            if (TEST_SIG_MIN_HASH_TYPE > WC_HASH_TYPE_SHA) {
                 word32 lenCopy = (word32)eccSigMax;
                 ExpectIntEQ(wc_SignatureGenerate(WC_HASH_TYPE_SHA, sig_type,
                     data, data_len, genSig, &lenCopy, &ecc, key_len, &rng),
                     WC_NO_ERR_TRACE(BAD_FUNC_ARG));
             }
+            else {
+                /* Floor lowered : SHA-1 must get past the
+                 * strength check and fail on the signature itself instead. */
+                ExpectIntEQ(wc_SignatureVerify(WC_HASH_TYPE_SHA, sig_type,
```
Frauschi: **🟡 [Medium] Lowered-floor else branch at the generate site calls wc_SignatureVerify() instead of wc_SignatureGenerate()**
*💡 SUGGEST*  `test`

The `else` added at the generate site calls `wc_SignatureVerify()`, not `wc_SignatureGenerate()` — it is a copy of the block at line 352 with the arguments unchanged (`sig`/`sig_len` instead of `genSig`/`lenCopy`). Three consequences:

1. **The gap this branch was added to close is still open.** Under a lowered floor, `wc_SignatureGenerate()` with `WC_HASH_TYPE_SHA` has no assertion anywhere in the suite. The `if` side is compiled but never taken, and the `else` side tests a different function.
2. **The duplicated assertion is vacuous.** `sig[128]` is still the all-zero buffer declared at line 251 — nothing writes a real signature into it; the genuine sign/verify round-trips at lines 363-394 use local `realSig`/`realSig2` buffers. So this re-runs the same all-zeros verify already asserted 190 lines earlier, on `data` filled with 0xA5 instead of 0x5A. It passes for the same reason (`wolfcrypt/src/signature.c:263-265` collapses every ECC verify failure to `SIG_VERIFY_E`), which is why the mistake is silent.
3. **The expected error is wrong for the generate path.** I built this PR head with `CFLAGS=-DWC_SIG_MIN_HASH_TYPE=WC_HASH_TYPE_MD5` and replaced the branch body with the intended call: `wc_SignatureGenerate(WC_HASH_TYPE_SHA, ...)` returns **0**, not `SIG_VERIFY_E`. There is no signature to fail — generate gets past `wc_SignatureCheckHashStrength()` and simply signs. The comment ("fail on the signature itself instead") describes the verify site, not this one.

`genSig` is also left unused in this branch.

**Suggestion:**
```suggestion
else {
                word32 lenCopy = (word32)eccSigMax;
                /* Floor lowered (e.g. wolfCLU with MD5): SHA-1 gets past the
                 * strength check and the signature is generated normally. */
                ExpectIntEQ(wc_SignatureGenerate(WC_HASH_TYPE_SHA, sig_type,
                    data, data_len, genSig, &lenCopy, &ecc, key_len, &rng),
                    0);
            }
```

tests/api/test_signature.c
```diff
@@ -334,13 +341,18 @@ int test_wc_SignatureDecisionCoverage(void)
...
                 WC_NO_ERR_TRACE(BAD_FUNC_ARG));
-
-#ifndef NO_SHA
-            /* Hash weaker than WC_SIG_MIN_HASH_TYPE (default SHA-256)
-             * rejected by wc_SignatureCheckHashStrength() */
-            ExpectIntEQ(wc_SignatureVerify(WC_HASH_TYPE_SHA, sig_type,
-                data, data_len, sig, sig_len, &ecc, key_len),
-                WC_NO_ERR_TRACE(BAD_FUNC_ARG));
+#if !defined(NO_SHA)
+            if (TEST_SIG_MIN_HASH_TYPE > WC_HASH_TYPE_SHA) {
```
Frauschi: **🔵 [Low] Guard compares wc_HashType enum values; the library compares digest sizes, and the two disagree in real configs**
*🔧 NIT*  `convention`

This is a real improvement over the `!defined(WC_SIG_MIN_HASH_TYPE)` guard and it fixes both of the previously-raised problems. The remaining nit is that it uses enum ordering as a proxy for hash strength, while `wc_SignatureCheckHashStrength()` (`wolfcrypt/src/signature.c:65-84`) compares `wc_HashGetDigestSize()` results. The proxy holds for MD5/SHA-1/SHA-2/SHA-3 floors, but diverges in two reachable cases:

- **Floor not compiled in.** If `wc_HashGetDigestSize(WC_SIG_MIN_HASH_TYPE)` returns negative, the library takes the `min_sz < 0` path and *skips enforcement entirely* — SHA-1 is accepted. The test would still take the `if` side and assert `BAD_FUNC_ARG`, so it fails. Reachable with e.g. a `WC_HASH_TYPE_SHAKE128` floor (not handled by `wc_HashGetDigestSize()` at all) or a SHA-512 floor in a `--disable-sha512` build.
- **The FIPS/selftest enum is a different numbering.** Under `HAVE_SELFTEST` or `HAVE_FIPS_VERSION <= 2` (`wolfssl/wolfcrypt/types.h:1478-1506`) the values are `MD5 = 0, SHA = 1, SHA256 = 2` but `MD2 = 16, MD4 = 17` — so `MD2 > WC_HASH_TYPE_SHA` is true there while MD2's 16-byte digest is weaker than SHA-1's 20. The realistic floors still order correctly, so nothing breaks today.

A digest-size comparison is the same shape and matches the library exactly, including the not-compiled-in case (`wc_HashGetDigestSize()` returns negative, which is not `> WC_SHA_DIGEST_SIZE`, so the assertion is skipped just as enforcement is).

**Suggestion:**
```suggestion
if (wc_HashGetDigestSize(TEST_SIG_MIN_HASH_TYPE) >
                    WC_SHA_DIGEST_SIZE) {
```

tests/api/test_signature.c
Frauschi: Only half of this one is addressed. The verify site (line 351) now has the `else` branch and it is correct.

The generate site does not: the `else` at line 539 calls `wc_SignatureVerify()` rather than `wc_SignatureGenerate()`, so `wc_SignatureGenerate()` with `WC_HASH_TYPE_SHA` under a lowered floor still has no assertion anywhere. Details in the new comment on line 541.

tests/api/test_signature.c
aidankeefe2022: fixed

tests/api/test_signature.c
aidankeefe2022: fixed

aidankeefe2022:
Fix for failing tests was resolved by https://github.com/wolfSSL/wolfssl/pull/11026 but fix for memory leak on failure is real and DoExpect for rng init should still go in

Frauschi:
Jenkins retest this please

tests/api/test_signature.c
```diff
@@ -304,7 +304,7 @@ int test_wc_SignatureDecisionCoverage(void)
             WC_RNG rng;
 
             XMEMSET(data, 0x5A, sizeof(data));
-            ExpectIntEQ(wc_InitRng(&rng), 0);
+            DoExpectIntEQ(wc_InitRng(&rng), 0);
```
Frauschi: **🟡 [Medium] RNG fix deviates from the file's own (and the repo's dominant) zero-then-init idiom**
*💡 SUGGEST*  `convention`

The fix is correct: `wc_FreeRng()` dereferences `rng->status`, `rng->drbg`, `rng->drbg512` and hands `rng->heap` to `XFREE()` (`wolfcrypt/src/random.c:2615-2680`), so pairing an unconditional `DoExpectIntEQ(wc_FreeRng(...))` with a conditional `ExpectIntEQ(wc_InitRng(...))` really did free garbage off an uninitialized stack `WC_RNG`. And a *failing* `wc_InitRng` is safe to free, because `_InitRng()` does `XMEMSET(rng, 0, sizeof(*rng))` at `random.c:1838` before anything can fail. So there is no bug here.

The concern is that the PR fixes the asymmetry from the wrong end, and in a way this file uses nowhere else. Every *other* handle in `test_signature.c` follows the standard idiom -- zero the struct, init conditionally, free unconditionally: `ecc_key` (`XMEMSET(&ecc, 0, ...)` at lines 59/253/686 + `ExpectIntEQ(wc_ecc_init(...))` + `DoExpectIntEQ(wc_ecc_free(...))`), `RsaKey` (lines 104/555/742), and `falcon_key` (line 182). The zeroing is precisely what makes those unconditional frees safe. `WC_RNG` is now the sole handle in the file that is left uninitialized and instead relies on the init macro never being downgraded.

That also matches the wider repo: across `tests/api/`, `XMEMSET(&rng, 0, sizeof(rng))` appears 226 times in 29 files and `ExpectIntEQ(wc_InitRng` 324 times, while `DoExpectIntEQ(wc_InitRng` appears only in these 7 new lines. A zeroed `WC_RNG` is safe to free -- `WC_DRBG_NOT_INIT` is 0 and `WC_DRBG_BANKREF` is 4 (`wolfssl/wolfcrypt/random.h:346-352`), so the `WC_RNG_BANK_SUPPORT` early-return in `wc_FreeRng` is not tripped.

Secondary effect of the chosen approach: `wc_InitRng()` now instantiates a DRBG and draws entropy even when the test has already failed and every following assertion will be skipped. Harmless on CI, but it is wasted work on entropy-constrained targets and is the opposite of what the skip machinery is for.

**Suggestion:**
```suggestion
XMEMSET(&rng, 0, sizeof(rng));
            DoExpectIntEQ(wc_InitRng(&rng), 0);
```

**Recommendation:** Prefer adding `XMEMSET(&rng, 0, sizeof(rng));` at each of the 7 sites (lines 185, 307, 404, 481, 642, 691, 777). Either keep `ExpectIntEQ` for the init -- matching the 324 other call sites and the ecc/rsa/falcon handling in this very file -- or keep `DoExpectIntEQ` and add the memset anyway as defense in depth, so the free stays safe regardless of which macro a future refactor or merge-conflict resolution leaves behind. If the `DoExpectIntEQ` form is deliberately preferred going forward, say so in the PR so the repo can converge rather than carry two idioms. Note separately that the same conditional-init/unconditional-free asymmetry exists in other `tests/api/` files that lack the memset -- worth a follow-up sweep, but out of scope here.

Frauschi (commented):
## 🐺 Skoll Code Review

### Posted findings
- [Medium] **RNG fix deviates from the file's own (and the repo's dominant) zero-then-init idiom** — `tests/api/test_signature.c:307`

- [Low] PR description no longer matches the diff: the WC_SIG_MIN_HASH_TYPE half already landed on master
- [Info] Fix is unverified by execution in this review environment

</details>

tests/api/test_signature.c
aidankeefe2022: reverted DoExpect* and now only memset rng object on initialization so it is always safe to free.

aidankeefe2022:
retest this please

Frauschi (approved):
LGTM

dgarske (approved):

PR #11026, opened 2026-07-31
Honor WC_SIG_MIN_HASH_TYPE override in signature decision test

kojo1:
# Description

test_wc_SignatureDecisionCoverage asserts that SHA-1 input is rejected
by wc_SignatureVerify/Generate, assuming the default
SHA-256 hash strength floor. A build may lower the floor - configuring
with --enable-wolfclu.

Fixes zd#22151

# Testing

fixed test_wc_SignatureGetSize_ecc

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/tests/api/test_signature.c
+++ b/tests/api/test_signature.c
@@ -41,6 +41,13 @@
 #include <tests/api/api.h>
 #include <tests/api/test_signature.h>
 
+/* Effective hash floor used by wc_SignatureVerify/Generate; mirrors the
+ * default in wolfcrypt/src/signature.c. A build may lower it (e.g.
+ * --enable-wolfclu defines WC_SIG_MIN_HASH_TYPE=WC_HASH_TYPE_MD5). */
+#ifndef WC_SIG_MIN_HASH_TYPE
+    #define WC_SIG_MIN_HASH_TYPE WC_HASH_TYPE_SHA256
+#endif
+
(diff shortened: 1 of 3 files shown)


tests/api/test_signature.c
```diff
@@ -41,6 +41,13 @@
 #include <tests/api/api.h>
 #include <tests/api/test_signature.h>
 
+/* Effective hash floor used by wc_SignatureVerify/Generate; mirrors the
+ * default in wolfcrypt/src/signature.c. A build may lower it (e.g.
+ * --enable-wolfclu defines WC_SIG_MIN_HASH_TYPE=WC_HASH_TYPE_MD5). */
+#ifndef WC_SIG_MIN_HASH_TYPE
```
dgarske: Let's move this into a header (like signature.h) and use it from there. This same pattern is duplicated at top of signature.c. That seems better overall, but I am okay with this PR as-is if needed.

dgarske (changes requested):
Also double check the PR description which indicates "test_wc_SignatureGetSize_ecc", but no code changes in it.

tests/api/test_signature.c
kojo1: Move WC_SIG_MIN_HASH_TYPE default into signature.h.

dgarske (approved):

PR #11027, opened 2026-07-31
Fixes for OCSP stapling, cert manager, and certificate_status_request_v2 handling

Frauschi:
Three independent fixes:
1. **certificate_status_request_v2.** Enforce RFC 8446 4.4.2.1 on the client (ClientHello only under TLS 1.3, `EXT_NOT_ALLOWED` elsewhere), and align certificate checks with `AddCA()`. Pooled signers now use `WOLFSSL_CHAIN_CA`, so `wolfSSL_CertManagerUnloadIntermediateCerts()` can remove them. Fixes F-7227.
2. **OCSP stapling request ownership.** Drop the shared `OcspRequest.ssl` back-pointer that concurrent handshakes raced on, pass the connection explicitly, move publication of the CTX-cached request into `CreateOcspResponse()` behind an explicit `ctxOwnsRequest` flag, and key the cache lock off `ssl->ctx->cm`. Plus a leak fix on the `WOLFSSL_CSR2_OCSP_MULTI` allocation-failure paths. Fixes F-7230 and F-7231.
3. **Cert manager CRL/OCSP init.** Build each object in a local and publish only once initialized, under `caLock`; free a half-built object after the unlock, since `FreeCRL()` can join the monitor thread, which takes `crlLock` before `caLock`. Fixes F-7235.

**ABI:** `struct OcspRequest` loses its trailing `void* ssl`. It is fully defined in an installed header and re-exported as `OCSP_REQUEST`, so this is a size change for unrebuilt consumers. No other offsets move; no in-tree consumer affected. `wolfSSL_CertManagerEnable{CRL,OCSP,OCSPStapling}()` gain a `BAD_MUTEX_E` return.

**Tests:** new `test_TLSX_CSR2_tls13_msg_type_validation` and `test_ocsp_ctx_request_cache`, the latter run by a new `ocsp.yml` job in plain-stapling, `--enable-all -DWOLFSSL_NO_COPY_CERT`, and ASan builds. Zero failures across four local configurations; removing the `!ctxOwnsRequest` guard makes the cache test die under ASan, so it can actually fail.

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -16505,6 +16505,32 @@ static int ProcessPeerCertAddPendingCA(WOLFSSL* ssl, buffer* cert)
     if (ret != 0) {
         goto exit_req_v2;
     }
+    /* Only a certificate that is actually usable as a CA may enter the pending
+     * signer pool. ParseCertRelative() consults that pool ahead of the
+     * certificate manager and uses the signer as a verification key without
+     * further checks, so admission has to enforce the same capabilities AddCA()
+     * requires of a chain CA.
+     *
+     * A non-CA is skipped rather than reported as an error, because
+     * ProcessPeerCerts() only offers a chain cert to AddCA() when isCA is set:
+     * such a certificate is already left out of the certificate manager without
+     * failing the handshake. */
+    if (!dCertAdd->isCA) {
+        WOLFSSL_MSG("Chain cert is not a CA, not adding as pending CA");
+        goto exit_req_v2;
+    }
+#ifndef ALLOW_INVALID_CERTSIGN
+    /* Per RFC 5280 an absent Key Usage extension implies all usages, so only
+     * enforce certificate signing when the extension is actually present.
+     * AddCA() rejects such a certificate outright, so report the same error
+     * here rather than quietly leaving it out of the pool. */
+    if (!dCertAdd->selfSigned && dCertAdd->extKeyUsageSet &&
+            (dCertAdd->extKeyUsage & KEYUSE_KEY_CERT_SIGN) == 0) {
+        WOLFSSL_MSG("Chain cert doesn't have key usage certificate signing");
+        ret = NOT_CA_ERROR;
+        goto exit_req_v2;
+    }
+#endif
     ret = AllocDer(&derBuffer, cert->length, CA_TYPE, ssl->heap);
     if (ret != 0 || derBuffer == NULL) {
         goto exit_req_v2;
@@ -16515,7 +16541,11 @@ static int ProcessPeerCertAddPendingCA(WOLFSSL* ssl, buffer* cert)
         ret = MEMORY_E;
         goto exit_req_v2;
     }
-    ret = FillSigner(s, dCertAdd, CA_TYPE, derBuffer);
+    /* WOLFSSL_CHAIN_CA, not CA_TYPE: TLSX_CSR2_MergePendingCA() promotes this
+     * signer into the certificate manager, and the unload path
+     * (wolfSSL_CertManagerUnloadIntermediateCerts()) selects entries by that
+     * type. AddCA() records chain CAs the same way. */
+    ret = FillSigner(s, dCertAdd, WOLFSSL_CHAIN_CA, derBuffer);
     if (ret != 0) {
         goto exit_req_v2;
     }
@@ -17746,14 +17776,25 @@ int ProcessPeerCerts(WOLFSSL* ssl, byte* input, word32* inOutIdx,
                     }
                 #endif
 #if defined(HAVE_CERTIFICATE_STATUS_REQUEST_V2)
-                    if (ret == 0 && addToPendingCAs && !alreadySigner) {
-                        /* skipAddCA is only consulted later on the ret == 0
-                         * continuation path; on helper failure we goto
-                         * exit_ppc, so setting it up-front is safe. */
+                    if (ret == 0 && addToPendingCAs && !alreadySigner &&
+                            !ssl->options.verifyNone && !skipAddCA) {
+                        /* The verifyNone and skipAddCA conditions mirror the
+                         * guards on the AddCA() call below. A certificate the
+                         * surrounding code has already declined to admit as a
+                         * CA must not enter the pending pool either, and must
+                         * not be failed against CA rules. skipAddCA is set here
+                         * so AddCA() is not also run for a certificate that did
+                         * enter the pool; it is only consulted later on the
+                         * ret == 0 continuation path, so setting it up-front is
+                         * safe. */
                         skipAddCA = 1;
                         ret = ProcessPeerCertAddPendingCA(ssl,
                                 &args->certs[args->certIdx]);
-                        if (ret != 0)
+                        /* A certificate turned away for not being a usable CA
+                         * falls through to the shared error handling below so
+                         * that it fails the handshake the same way a rejection
+                         * from AddCA() does. */
+                        if (ret != 0 && ret != WC_NO_ERR_TRACE(NOT_CA_ERROR))
                             goto exit_ppc;
                     }
 #endif /* HAVE_CERTIFICATE_STATUS_REQUEST_V2 */
--- a/src/internal.c
+++ b/src/internal.c
@@ -26412,22 +26412,36 @@ int SendFinished(WOLFSSL* ssl)
         (defined(HAVE_CERTIFICATE_STATUS_REQUEST) || \
          defined(HAVE_CERTIFICATE_STATUS_REQUEST_V2))) || \
     (defined(WOLFSSL_TLS13) && defined(HAVE_CERTIFICATE_STATUS_REQUEST))
-/* Parses and decodes the certificate then initializes "request". In the case
- * of !ssl->buffers.weOwnCert, ssl->ctx->certOcspRequest gets set to "request".
+/* Returns the lock that guards the OCSP request cache on the WOLFSSL_CTX.
+ *
+ * The cache is a field of the CTX, so it has to be serialized with a lock whose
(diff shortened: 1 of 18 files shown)


wolfssl/wolfcrypt/asn.h
```diff
@@ -3099,7 +3099,6 @@ struct OcspRequest {
     byte   nonce[MAX_OCSP_NONCE_SZ];
     int    nonceSz;
     void*  heap;
-    void*  ssl;
```
philljj, line 3102: This is an installed header so this will be an ABI change. (it may be a necessary change, just making a note)

Frauschi: If we want to prevent an ABI change (if a SO version bump is not required otherwise in the next release anyway), we can re-add that. But technically the pointer is not used in the code anymore.

src/internal.c
```diff
@@ -16505,6 +16505,32 @@ static int ProcessPeerCertAddPendingCA(WOLFSSL* ssl, buffer* cert)
     if (ret != 0) {
         goto exit_req_v2;
     }
+    /* Only a certificate that is actually usable as a CA may enter the pending
+     * signer pool. ParseCertRelative() consults that pool ahead of the
+     * certificate manager and uses the signer as a verification key without
+     * further checks, so admission has to enforce the same capabilities AddCA()
+     * requires of a chain CA.
+     *
+     * A non-CA is skipped silently rather than reported as an error, because
```
philljj: It's not skipped silently, because it's logged right here in `WOLFSSL_MSG`?

src/ssl_certman.c
```diff
@@ -1777,30 +1778,65 @@ int wolfSSL_CertManagerEnableCRL(WOLFSSL_CERT_MANAGER* cm, int options)
...
+                    if (crl == NULL) {
+                        ret = MEMORY_E;
+                    }
+                    else {
+                        /* Reset fields of CRL object. */
+                        XMEMSET(crl, 0, sizeof(WOLFSSL_CRL));
+                        /* Initialize CRL object. */
+                        if (InitCRL(crl, cm) != 0) {
+                            WOLFSSL_MSG("Init CRL failed");
+                            /* Dispose of it outside the critical section:
```
philljj: Calling FreeCRL() inside the critical section can deadlock?

If that's what the comment is referring to, then would prefer to state that more plainly.

src/ssl_certman.c
```diff
@@ -2192,23 +2229,48 @@ int wolfSSL_CertManagerEnableOCSP(WOLFSSL_CERT_MANAGER* cm, int options)
...
+                    if (ocsp == NULL) {
+                        ret = MEMORY_E;
+                    }
+                    else {
+                        /* Reset the fields of the OCSP object. */
+                        XMEMSET(ocsp, 0, sizeof(WOLFSSL_OCSP));
+                        /* Initialize the OCSP object. */
+                        if (InitOCSP(ocsp, cm) != 0) {
+                            WOLFSSL_MSG("Init OCSP failed");
+                            /* Dispose of it outside the critical section, for
```
philljj: Calling FreeOCSP inside critical section can deadlock? Same comment as above.

src/internal.c
```diff
@@ -26420,50 +26461,90 @@ int CreateOcspRequest(WOLFSSL* ssl, OcspRequest* request,
...
+                      der->length);
+        }
+
+        /* Only a request built from a CTX-held certificate may be cached, since
+         * the cache outlives this connection. */
+        if (ret == 0 && !ssl->buffers.weOwnCert) {
+            wolfSSL_Mutex* ocspLock = GetCtxOcspLock(ssl);
+
+            if (ocspLock == NULL) {
+                WOLFSSL_MSG("No CTX OCSP lock, not caching the request");
```
philljj, line 26544: Does this `if` or `else if` warrant returning an error?

src/ocsp.c
```diff
@@ -480,7 +482,7 @@ int CheckOcspResponse(WOLFSSL_OCSP *ocsp, byte *response, int responseSz,
 #define OCSP_MAX_REQUEST_SZ 2048
 #endif
 int CheckOcspRequest(WOLFSSL_OCSP* ocsp, OcspRequest* ocspRequest,
-                     buffer* responseBuffer, void* heap)
+                     buffer* responseBuffer, void* heap, WOLFSSL* ssl)
```
philljj: this heap argument seems redundant now.

src/ssl_certman.c
```diff
@@ -2301,23 +2364,48 @@ int wolfSSL_CertManagerEnableOCSPStapling(WOLFSSL_CERT_MANAGER* cm)
...
+                    if (ocsp == NULL) {
+                        ret = MEMORY_E;
+                    }
+                    else {
+                        /* Reset the fields of the OCSP object. */
+                        XMEMSET(ocsp, 0, sizeof(WOLFSSL_OCSP));
+                        /* Initialize the OCSP stapling object. */
+                        if (InitOCSP(ocsp, cm) != 0) {
+                            WOLFSSL_MSG("Init OCSP failed");
+                            /* Dispose of it outside the critical section, for
```
philljj: Same.

src/internal.c
```diff
@@ -26475,9 +26556,8 @@ int CreateOcspResponse(WOLFSSL* ssl, OcspRequest** ocspRequest,
     }
 
     if (ret == 0) {
-        request->ssl = ssl;
         ret = CheckOcspRequest(SSL_CM(ssl)->ocsp_stapling, request, response,
-                               ssl->heap);
+                               ssl->heap, ssl);
```
philljj: Looks like we're passing `ssl->heap, ssl)` everywhere in `CheckOcspRequest()` now. The separate heap arg is redundant now.

tests/api/test_ocsp.c
```diff
@@ -1845,6 +1846,204 @@ int test_tls13_nonblock_ocsp_low_mfl(void)
...
+    *respBuf = (unsigned char*)resp_server1_cert;
+    return (int)sizeof(resp_server1_cert);
+}
+
+/* The leaf OCSP request that stapling builds is cached on the WOLFSSL_CTX, and
+ * every later connection on that CTX reuses it. The CTX owns it from that point
+ * on and frees it in wolfSSL_CTX_free(), so no connection may free it.
+ *
+ * Runs three handshakes over one CTX pair:
+ *   pass 0 - builds the request and publishes it on the CTX,
```
philljj, line 1907: Is there a way to interleave these tests to make them less serial?

wolfssl/wolfcrypt/asn.h
philljj: It looks like a justified change

philljj (changes requested):
Forgot to select  `Request changes`

wolfssl/wolfcrypt/asn.h
Frauschi: Agreed it is one. Nothing reads the field any more, so I have left it removed and will get it into the release notes.

src/internal.c
Frauschi: Dropped the word - the `WOLFSSL_MSG` does log it. Only meant that it is not an error.

src/ssl_certman.c
Frauschi: Checked it: no, it cannot deadlock. `crlLock` -> `caLock` exists (`CheckCertCRLList()`, then `GetCA()`) but nothing goes the other way, and `InitCRL()` sets `tid` to `INVALID_THREAD_VAL` before every failure return, so `FreeCRL()` never reaches the monitor join. Comment and commit message corrected.

It did expose a real bug, now a separate commit: the three `Enable*` functions and `wolfSSL_d2i_X509_CRL()` called `FreeCRL()` / `FreeOCSP()` after a *failed* init, double-destroying the lock and condition variable. Fixing it dropped the deferred free with it.

src/ssl_certman.c
Frauschi: Not the same reason - `FreeOCSP()` takes no lock. Both comments say that now, and the free is a plain `XFREE` at this point anyway.

src/ssl_certman.c
Frauschi: Not the same reason - `FreeOCSP()` takes no lock. Both comments say that now, and the free is a plain `XFREE` at this point anyway.

src/internal.c
Frauschi: No. `ctxOwns` stays 0, so this connection frees the request as usual and only cross-connection reuse is lost. The `NULL` branch is reachable, not padding: `SSL_CM(ssl)` can resolve through `ssl->x509_store_pt`. Both noted in the comment.

src/ocsp.c
Frauschi: Agreed, dropped. It was only the hint for `responseBuffer->buffer`, and every caller with a real one passed `ssl->heap`, so deriving it inside is identical at all six sites. `CheckOcspResponse()` keeps its heap arg for `wc_CheckCertOcspResponse()`, which has no `ssl`.

src/internal.c
Frauschi: Agreed, dropped. It was only the hint for `responseBuffer->buffer`, and every caller with a real one passed `ssl->heap`, so deriving it inside is identical at all six sites. `CheckOcspResponse()` keeps its heap arg for `wc_CheckCertOcspResponse()`, which has no `ssl`.

tests/api/test_ocsp.c
Frauschi: The serial order is what is under test: publish, then clean reuse, then marked reuse. Run concurrently, a wrong ownership decision stops being observable. It is three memio handshakes, no sockets or sleeps.

philljj (approved):

dgarske (approved):

PR #11028, opened 2026-07-31
Remove dead/unreachable code

danielinux:
# Description

Eliminates all conditions that are impossible to reach from the existing wolfcrypt code. 

Each dead condition is self-containerd and documented in a separate commit, with proof of death through object code comparison vs. gcc -O2:

# Testing

These are leftover (uncoverable) conditions discovered during the the first four phases of MC/DC campaign

## Proven by gcc -O2 object equivalence (15)

The optimiser had already folded the condition — compiling the file before and after the commit yields byte-identical code, so the removal is provably behaviour-preserving.

- 00f098e6e coding.c: Base64_SkipNewline, the len operand at :137. Entry check rejects len==0; the loop decrements only under len>1.
- 03e40fdc2 coding.c : the out operand at :281 and :395 in both Base64 decoders; entry checks already reject out==NULL.
- 413fcc550 coding.c : the in operand at :663 in Base16_Decode; :660 rejects in==NULL unconditionally.
- 4acfa9984 sp_int.c : nested if (err == MP_OKAY) in sp_todecimal, dominated by :18887; only char writes and a discarded sp_div_d between.
- d851828e0 sp_int.c : same pattern in _sp_lcm; only two void _sp_init_size() calls between.
- 738454dc4 memory.c : res == NULL in wolfSSL_Realloc; both writes to res are on branches mutually exclusive with this one.
- 04774352f wc_mldsa.c : (ret == 1) loop guard in mldsa_vec_check_low_c; callee returns only 0 or 1 and the body breaks on 0.
- 5aecf75c7 wc_mldsa.c : (ret == 0) in wc_MlDsaKey_CheckKey; every call in the dominated span returns void.
- cd1324ff1 wc_mldsa.c : (ret == 0) at the small-mem sign entry; ret unwritten between declaration and its only evaluation.
- 5cd3ba0b8 wc_mldsa.c : (ret == 0) in two inner loops whose every ret write is followed by an unconditional break.
- 23b3a7afc integer.c : the six mp_init() != MP_OKAY guards and their mp_clear cascades in mp_init_multi; mp_init fails only on NULL, already excluded. (Confirmed under --enable-heapmath, the config that compiles the file.)
- 5cdce9689 aes.c : bit >= 0 in wc_AesFeedbackCFB1; bit is normalised back to 7 on every decrement. bit < 7 kept — it's live.
- c62fad325 ed448.c : !key->privKeySet in wc_ed448_check_key, plus the nested always-true if (ret == 0); both follow from the sibling if's short-circuit.
- 7a07129d5 kdf.c : ret == 0 in the KDA-KDF CMAC while; the body's only exit is an unconditional if (ret != 0) break.
- 25ca23099 pwdbased.c : DoPKCS12Hash's NULL check; static, two call sites, all four paths (× WOLFSSL_SMALL_STACK) NULL-check and return first.

### Proved by exhaustive enumeration [no optimizer can reach these]  (6) 

- fc72ef83f md5.c : len == 0 in wc_Md5Update. With data == NULL held true, all 2³² word32 values of len enumerated: every one passing the :345 guard has len == 0. Zero counterexamples. (gcc won't correlate the entry guard's two operands across the branch.)
- f04324101 tfm.c : fp_gcd's two redundant operands plus the unreachable fp_iszero(a) block. fp_iszero is side-effect-free over ->used and neither operand is written in the span, so the chain depends on exactly two booleans: all 4 (Za,Zb) states enumerated, 0 counterexamples.
- e11b5e1e4 integer.c : tmpc == NULL in s_mp_add. mp_grow has exactly three exits, enumerated in full; the only MP_OKAY one is reached after a->dp = tmp (NULL-checked) or through the else if (a->dp == NULL) return MP_VAL filter. Hence MP_OKAY ⇒ dp != NULL.
- aeb153c46 wc_mlkem.c : the BAD_STATE_E re-check. All 9 writes to key->flags in the TU enumerated (:453, 621, 852, 1001, 1003, 2283, 2327, 2430, 2794); check_h calls only 4 functions, of which just EncodePublicKey touches flags — and it sets H_SET on ret == 0. No reachable path clears it.
- 7283dc884 wc_lms_impl.c : !priv_key->inited. HssPrivKey declares byte* priv as a pointer member (wc_lms.h:726) alongside word8 inited:1 (:741); the callee receives the pointer value, so reaching inited through it would be an out-of-bounds write. (gcc must conservatively assume aliasing.)
- 0e18af6fb ed448.c : the Y == p case. The byte walk pins 56 of 57 key bytes, leaving p[0] as the only free variable: all 256 values enumerated against the real ed448_is_small_order(). Two (0xff = y=p, 0xfe = y=p−1) are rejected early by the table; the other 254 satisfy the operand. Zero counterexamples over the complete state space.

### Compile-gated, no host evidence (1)

- 0090560d3 sha3.c  the same len == 0 idiom at :1189/:1868/:2166, all inside PSOC6_HASH_SHA3, which needs the Cypress SDK. Proof is textual (nothing executes between guard and site); the commit says plainly it is not compile-verified. Deliberately does not touch the four look-alikes at :1123/:1285/:1967/:2263, where len == 0 is load-bearing.

### DIFF
--- a/wolfcrypt/src/ed448.c
+++ b/wolfcrypt/src/ed448.c
@@ -1521,40 +1521,38 @@ int wc_ed448_check_key(ed448_key* key)
         }
     }
     /* No private key, check Y is valid. */
-    else if ((ret == 0) && (!key->privKeySet)) {
+    else if (ret == 0) {
         /* Verify that xQ and yQ are integers in the interval [0, p - 1].
          * Only have yQ so check that ordinate.
          * p = 2^448-2^224-1 = 0xff..fe..ff
          */
-        if (ret == 0) {
-            int i;
-            ret = PUBLIC_KEY_E;
+        int i;
+        ret = PUBLIC_KEY_E;
 
-            /* Check top part before 0xFE. */
-            for (i = ED448_PUB_KEY_SIZE - 1; i > ED448_PUB_KEY_SIZE/2; i--) {
-                if (key->p[i] < 0xff) {
-                    ret = 0;
-                    break;
-                }
+        /* Check top part before 0xFE. */
+        for (i = ED448_PUB_KEY_SIZE - 1; i > ED448_PUB_KEY_SIZE/2; i--) {
+            if (key->p[i] < 0xff) {
+                ret = 0;
+                break;
             }
-            if (ret == WC_NO_ERR_TRACE(PUBLIC_KEY_E)) {
-                /* Check against 0xFE. */
-                if (key->p[ED448_PUB_KEY_SIZE/2] < 0xfe) {
-                    ret = 0;
-                }
-                else if (key->p[ED448_PUB_KEY_SIZE/2] == 0xfe) {
-                    /* Check bottom part before last byte. */
-                    for (i = ED448_PUB_KEY_SIZE/2 - 1; i > 0; i--) {
-                        if (key->p[i] != 0xff) {
-                            ret = 0;
-                            break;
-                        }
-                    }
-                    /* Check last byte. */
-                    if ((ret == WC_NO_ERR_TRACE(PUBLIC_KEY_E)) &&
-                        (key->p[0] < 0xff)) {
+        }
+        if (ret == WC_NO_ERR_TRACE(PUBLIC_KEY_E)) {
+            /* Check against 0xFE. */
+            if (key->p[ED448_PUB_KEY_SIZE/2] < 0xfe) {
+                ret = 0;
+            }
+            else if (key->p[ED448_PUB_KEY_SIZE/2] == 0xfe) {
+                /* Check bottom part before last byte. */
+                for (i = ED448_PUB_KEY_SIZE/2 - 1; i > 0; i--) {
+                    if (key->p[i] != 0xff) {
                         ret = 0;
+                        break;
                     }
+                }
+                /* Check last byte. */
+                if ((ret == WC_NO_ERR_TRACE(PUBLIC_KEY_E)) &&
+                    (key->p[0] < 0xff)) {
+                    ret = 0;
                 }
             }
         }
--- a/wolfcrypt/src/ed448.c
+++ b/wolfcrypt/src/ed448.c
@@ -251,7 +251,10 @@ static int ed448_is_small_order(const byte p[ED448_PUB_KEY_SIZE])
      * (fe448_from_bytes) reads bytes 0-55 modulo p with no canonical-form
      * check, so y = p decodes to 0 and y = p+1 decodes to 1; both must
      * be rejected here. Only {y, y + p} fits in 56 bytes (2p overflows),
-     * so listing y and y + p exhausts the reachable encodings. */
+     * so listing y and y + p exhausts the reachable encodings.
+     * wc_ed448_check_key() depends on the y = p row: its Y-range test
+     * accepts that encoding, so dropping the row would let a y outside
+     * [0, p - 1] through. */
     static const byte small_order_y[][ED448_PUB_KEY_SIZE] = {
         /* order 1: identity y = 1, x = 0 */
         {0x01,0x00,0x00,0x00,0x00,0x00,0x00,0x00,
@@ -1537,23 +1540,13 @@ int wc_ed448_check_key(ed448_key* key)
             }
         }
         if (ret == WC_NO_ERR_TRACE(PUBLIC_KEY_E)) {
-            /* Check against 0xFE. */
-            if (key->p[ED448_PUB_KEY_SIZE/2] < 0xfe) {
+            /* Every byte above this one is 0xff here, so y > p whenever this
+             * byte is 0xff, and y == p is then the only remaining encoding
+             * outside [0, p - 1]. It is already rejected by
+             * ed448_is_small_order() above, whose table carries y == p as a
+             * non-canonical encoding, so the low bytes need no check. */
+            if (key->p[ED448_PUB_KEY_SIZE/2] <= 0xfe) {
                 ret = 0;
-            }
-            else if (key->p[ED448_PUB_KEY_SIZE/2] == 0xfe) {
-                /* Check bottom part before last byte. */
-                for (i = ED448_PUB_KEY_SIZE/2 - 1; i > 0; i--) {
-                    if (key->p[i] != 0xff) {
-                        ret = 0;
-                        break;
-                    }
-                }
-                /* Check last byte. */
-                if ((ret == WC_NO_ERR_TRACE(PUBLIC_KEY_E)) &&
-                    (key->p[0] < 0xff)) {
-                    ret = 0;
-                }
             }
         }
 
--- a/wolfcrypt/src/integer.c
+++ b/wolfcrypt/src/integer.c
@@ -98,12 +98,10 @@ word32 CheckRunTimeSettings(void)
 }
 
 
-/* handle up to 6 inits */
+/* handle up to 6 inits; returns MP_OKAY */
 int mp_init_multi(mp_int* a, mp_int* b, mp_int* c, mp_int* d, mp_int* e,
                   mp_int* f)
 {
-    int res = MP_OKAY;
-
     if (a) XMEMSET(a, 0, sizeof(mp_int));
     if (b) XMEMSET(b, 0, sizeof(mp_int));
     if (c) XMEMSET(c, 0, sizeof(mp_int));
@@ -111,35 +109,18 @@ int mp_init_multi(mp_int* a, mp_int* b, mp_int* c, mp_int* d, mp_int* e,
     if (e) XMEMSET(e, 0, sizeof(mp_int));
     if (f) XMEMSET(f, 0, sizeof(mp_int));
 
-    if (a && ((res = mp_init(a)) != MP_OKAY))
-        return res;
+    /* mp_init() has exactly one failure mode, a NULL argument, and the guard
+     * on each call excludes it, so every one of these returns MP_OKAY and the
+     * result is discarded. Same shape as sp_init_multi() and tfm.c's
+     * mp_init_multi(), which also return MP_OKAY unconditionally. */
+    if (a) (void)mp_init(a);
+    if (b) (void)mp_init(b);
+    if (c) (void)mp_init(c);
+    if (d) (void)mp_init(d);
+    if (e) (void)mp_init(e);
+    if (f) (void)mp_init(f);
 
-    if (b && ((res = mp_init(b)) != MP_OKAY)) {
-        mp_clear(a);
-        return res;
-    }
-
-    if (c && ((res = mp_init(c)) != MP_OKAY)) {
-        mp_clear(a); mp_clear(b);
-        return res;
-    }
-
-    if (d && ((res = mp_init(d)) != MP_OKAY)) {
-        mp_clear(a); mp_clear(b); mp_clear(c);
-        return res;
-    }
-
-    if (e && ((res = mp_init(e)) != MP_OKAY)) {
-        mp_clear(a); mp_clear(b); mp_clear(c); mp_clear(d);
-        return res;
-    }
-
-    if (f && ((res = mp_init(f)) != MP_OKAY)) {
-        mp_clear(a); mp_clear(b); mp_clear(c); mp_clear(d); mp_clear(e);
-        return res;
-    }
-
-    return res;
+    return MP_OKAY;
 }
 
 
--- a/wolfcrypt/src/integer.c
+++ b/wolfcrypt/src/integer.c
@@ -1792,8 +1792,7 @@ int s_mp_add (mp_int * a, mp_int * b, mp_int * c)
     tmpc = c->dp;
 
     /* sanity-check dp pointers. */
-    if ((min_ab > 0) &&
-        ((tmpa == NULL) || (tmpb == NULL) || (tmpc == NULL)))
+    if ((min_ab > 0) && ((tmpa == NULL) || (tmpb == NULL)))
     {
         return MP_VAL;
     }
--- a/wolfcrypt/src/pwdbased.c
+++ b/wolfcrypt/src/pwdbased.c
@@ -365,17 +365,15 @@ int wc_PBKDF2(byte* output, const byte* passwd, int pLen, const byte* salt,
 
 #ifdef HAVE_PKCS12
 
-/* helper for PKCS12_PBKDF(), does hash operation */
+/* helper for PKCS12_PBKDF(), does hash operation.
+ * buffer and Ai are guaranteed non-NULL by the caller: each is either a stack
+ * array or an XMALLOC result whose failure returns MEMORY_E before the call. */
 static int DoPKCS12Hash(enum wc_HashType hashT, byte* buffer, word32 totalLen,
     byte* Ai, word32 u, int iterations)
 {
     int i;
     int ret = 0;
     WC_DECLARE_VAR(hash, wc_HashAlg, 1, 0);
-
-    if ((buffer == NULL) || (Ai == NULL)) {
-        return BAD_FUNC_ARG;
-    }
 
     /* initialize hash */
     WC_ALLOC_VAR_EX(hash, wc_HashAlg, 1, NULL, DYNAMIC_TYPE_HASHCTX,
--- a/tests/api/test_ed448.c
+++ b/tests/api/test_ed448.c
@@ -1286,7 +1286,8 @@ int test_wc_ed448_import_private_only(void)
  * (the is_small_order VALUE operand's independence is already shown by
  * test_wc_ed448_reject_small_order_keys()), the (ret == 0 &&
  * XMEMCMP(...) != 0) recomputed-vs-stored public key mismatch check in the
- * have-private-key branch, and the deep Y-range check's final byte compare.
+ * have-private-key branch, and both decisions of the Y-range walk: the
+ * top-byte loop and the 0xFE compare that follows it.
  */
 int test_wc_ed448_check_key_decisions(void)
 {
@@ -1336,15 +1337,15 @@ int test_wc_ed448_check_key_decisions(void)
     key.p[0] = (byte)(key.p[0] ^ 0xff);
 
     /* Deep Y-range check: a Y value that is not one of
-     * ed448_is_small_order()'s tabulated points but still forces both
-     * range-check loops all the way down to the final byte compare (every
-     * byte except p[0] matches the encoded field prime p). Only reachable
-     * via a trusted import, which skips wc_ed448_check_key() at import
-     * time so the crafted (curve-invalid) point can be handed to a
-     * *direct* wc_ed448_check_key() call below -- same technique as
-     * test_wc_ed448_reject_small_order_keys(). Whatever the later
-     * curve-decode step decides is fine; the range-check decision itself
-     * is what's targeted here. */
+     * ed448_is_small_order()'s tabulated points but still runs the top-byte
+     * loop to exhaustion (every byte above the 0xFE position is 0xff), so
+     * the loop's index operand goes false and the 0xFE compare below it is
+     * reached and taken. Only reachable via a trusted import, which skips
+     * wc_ed448_check_key() at import time so the crafted (curve-invalid)
+     * point can be handed to a *direct* wc_ed448_check_key() call below --
+     * same technique as test_wc_ed448_reject_small_order_keys(). Whatever
+     * the later curve-decode step decides is fine; the range-check decision
+     * itself is what's targeted here. */
     XMEMSET(near_p, 0xff, sizeof(near_p));
     near_p[28] = 0xfe;
     near_p[0]  = 0x00;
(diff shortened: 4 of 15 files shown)


danielinux:
jenkins retest this please (build timeout)

wolfcrypt/src/ed448.c
```diff
@@ -1521,41 +1521,29 @@ int wc_ed448_check_key(ed448_key* key)
...
-                        }
-                    }
-                    /* Check last byte. */
-                    if ((ret == WC_NO_ERR_TRACE(PUBLIC_KEY_E)) &&
-                        (key->p[0] < 0xff)) {
-                        ret = 0;
-                    }
-                }
+        }
+        if (ret == WC_NO_ERR_TRACE(PUBLIC_KEY_E)) {
```
Frauschi, line 1542: **🟡 [Medium] ed448 range check is now correct only because ed448_is_small_order() tabulates y == p**
*💡 SUGGEST*  `bug`

The rewritten Y-range check drops the low-byte scan and the final `key->p[0] < 0xff` compare in favour of a single `if (key->p[ED448_PUB_KEY_SIZE/2] <= 0xfe) ret = 0;`. That is behaviour-preserving today, but only via a non-local invariant.

Reaching the compare at all requires `p[29..56]` to all be `0xff`. In that region old and new code diverge on exactly one encoding: `p[0..27] == 0xff, p[28] == 0xfe, p[29..55] == 0xff` -- i.e. `y == p`. The old code returned `PUBLIC_KEY_E`; the new code returns 0. It is still rejected only because `ed448_is_small_order()` (called ~30 lines earlier at :1511) carries `y = p` as a tabulated non-canonical encoding and masks byte 56 on both sides before comparing, so it also catches the `p[56] == 0xff` variant that is the only way to reach this block. I verified that table row matches bit-for-bit and that `ed448_is_small_order()` is compiled unconditionally, so there is no configuration where the case escapes, and the failure code is `PUBLIC_KEY_E` on both paths.

The fragility: the function's own doc comment still promises it verifies `y` is in `[0, p - 1]`, but the implementation no longer covers the `y == p` boundary itself. If the two non-canonical rows are ever trimmed from `small_order_y[]` -- a plausible refactor if canonicality enforcement moves into `fe448_from_bytes()` -- or if the table check is gated behind a macro or moved after this block, `wc_ed448_check_key()` silently starts returning 0 for a key it currently rejects, with no test failing. The new comment records the dependency at the consumer; `ed448_is_small_order()` carries no reciprocal note.

Not a BLOCK: there is no reachable behaviour change on the current tree.

**Suggestion:**
```suggestion
/* NOTE: this shortcut is only correct while ed448_is_small_order()
 * tabulates the non-canonical encodings y == p and y == p - 1. If those
 * entries are ever removed from small_order_y[], the low-byte scan below
 * must be restored or y == p will be accepted here. */
if (ret == WC_NO_ERR_TRACE(PUBLIC_KEY_E)) {
    if (key->p[ED448_PUB_KEY_SIZE/2] <= 0xfe) {
        ret = 0;
    }
}
```

wolfcrypt/src/integer.c
```diff
@@ -102,44 +102,22 @@ word32 CheckRunTimeSettings(void)
...
-    if (b && ((res = mp_init(b)) != MP_OKAY)) {
-        mp_clear(a);
-        return res;
-    }
-
-    if (c && ((res = mp_init(c)) != MP_OKAY)) {
-        mp_clear(a); mp_clear(b);
-        return res;
-    }
+    /* mp_init() only fails on a NULL argument, already excluded here. */
```
Frauschi: **🟡 [Medium] mp_init_multi now discards every mp_init() result and is structurally incapable of reporting failure**
*💡 SUGGEST*  `convention`

The six guarded `mp_init()` calls and their `mp_clear()` cascades become `(void)mp_init(x)` plus an unconditional `return MP_OKAY;`. The analysis is correct for the current tree -- `mp_init()` (integer.c:125-144) returns `MP_VAL` only for a NULL argument, which the `if (a)` guards already exclude, and it performs no allocation ("defer allocation until mp_grow") -- so no caller behaviour changes today.

The concern is the contract. `mp_init_multi()` is declared in wolfssl/wolfcrypt/integer.h with a return type that still advertises failure, and call sites across the tree still write `if ((res = mp_init_multi(...)) != MP_OKAY) return res;` (e.g. integer.c:1089, :1233). Those guards are now provably dead at *every* call site rather than at one. If `mp_init()` here ever regains a failure mode -- an allocation, a HW-backed init, a FIPS state check -- it is silently swallowed at six sites, and the `mp_clear()` cascade this patch deleted was the only cleanup for that partial-construction state. The repo convention is to check return codes, not cast them away.

Worth noting the calls are also close to vestigial now: the `XMEMSET(x, 0, sizeof(mp_int))` immediately above already produces exactly the state `mp_init()` writes (`dp = NULL`, `used = 0`, `alloc = 0`, `sign = MP_ZPOS`; `wc_bigint_init()` under `HAVE_WOLF_BIGINT` only zeroes `raw`).

**Suggestion:** Keep the checked form but collapse the cascade, e.g.:

int res = MP_OKAY;
...
if ((res == MP_OKAY) && a) res = mp_init(a);
if ((res == MP_OKAY) && b) res = mp_init(b);
...
if (res != MP_OKAY) {
    mp_clear(a); mp_clear(b); mp_clear(c); mp_clear(d); mp_clear(e);
}
return res;

(mp_clear() is NULL-safe, so the cascade collapses to one block.)

wolfcrypt/src/pwdbased.c
```diff
@@ -373,10 +373,6 @@ static int DoPKCS12Hash(enum wc_HashType hashT, byte* buffer, word32 totalLen,
     int ret = 0;
     WC_DECLARE_VAR(hash, wc_HashAlg, 1, 0);
 
-    if ((buffer == NULL) || (Ai == NULL)) {
-        return BAD_FUNC_ARG;
-    }
-
     /* initialize hash */
```
Frauschi, line 378: **🟡 [Medium] DoPKCS12Hash: this partially reverts hardening commit f15199906 from four months ago**
*💡 SUGGEST*  `convention`

I verified the unreachability claim at both call sites. `DoPKCS12Hash` is `static` with exactly two callers, pwdbased.c:551 (the `WC_PKCS12_PBKDF_USING_MP_API` variant) and :767 (the non-mp variant). At both, `buffer` starts as `&staticBuffer[0]` and is only replaced by an `XMALLOC` whose failure returns `MEMORY_E` before the call (:509-515, :743-749); `Ai`/`B` is either a stack array (:436, :681) or an `XMALLOC` with an immediate NULL check (:476-477, :738-739). No combination of `WOLFSSL_SMALL_STACK`, `WC_PKCS12_PBKDF_USING_MP_API` or `WOLFSSL_CHECK_MEM_ZERO` reaches either call with NULL. So the removal is safe today.

Two process points rather than a live bug. First, `git log -S "(buffer == NULL) || (Ai == NULL)"` shows the guard was added deliberately by f15199906 (2026-03-23, "SP int: fixes from AI review"); this PR removes it without saying so. Second, the guard's most likely original motivation -- a static-analysis NULL-deref report on `wc_HashUpdate(hash, hashT, buffer, totalLen)` / `wc_HashFinal(hash, hashT, Ai)` -- will resurface, and `Ai`/`B` come through `WC_DECLARE_VAR`, whose small-stack expansion is what analyzers most often flag. The function hashes `totalLen` bytes out of `buffer` and writes the digest into `Ai`, so a NULL slipping through from a future `#ifdef` arm becomes a crash inside the hash loop instead of a clean `BAD_FUNC_ARG`.

**Recommendation:** Either keep the guard (optionally under `DEBUG_WOLFSSL`), or call out in the PR description that f15199906 is being partially reverted, document "caller guarantees buffer and Ai are non-NULL" in the function header comment, and confirm the static-analysis job that motivated it stays clean.

wolfcrypt/src/integer.c
```diff
@@ -1792,8 +1770,7 @@ int s_mp_add (mp_int * a, mp_int * b, mp_int * c)
     tmpc = c->dp;
 
     /* sanity-check dp pointers. */
```
Frauschi, line 1775: **🟡 [Medium] No tests or assertions added for removals that rest on invariants owned by another function**
*💡 SUGGEST*  `test`

`git diff --stat master...HEAD -- tests/ wolfcrypt/test/` is empty: the PR removes 114 lines of code across 14 files and adds no test and no assertion. Several of the removals are correct only because of an invariant owned by a *different* function, and none of those invariants is documented at the site that provides it: (1) `wc_ed448_check_key()`'s range check now relies on `ed448_is_small_order()`'s table containing `y == p`; (2) `wc_mlkemkey_check_h()` relies on `wc_MlKemKey_EncodePublicKey()` setting `MLKEM_FLAG_H_SET` on every success path; (3) `s_mp_add()` relies on `mp_grow()` returning `MP_OKAY` only with `dp != NULL`; (4) `DoPKCS12Hash()` relies on both call sites always passing non-NULL `buffer`/`Ai`; (5) `wolfSSL_Realloc()` relies on the three `res`-assigning arms being mutually exclusive. Each is currently true — I verified all five by reading the providers — but each is now load-bearing without being stated where it is produced, which is the failure mode that makes 'provably dead today' removals rot.

**Suggestion:** Add a one-line comment at each *provider* of the invariant, e.g. in mp_grow():

  /* Callers (s_mp_add, s_mp_sub) rely on MP_OKAY implying a->dp != NULL. */

and in wc_MlKemKey_EncodePublicKey():

  /* wc_mlkemkey_check_h() relies on H_SET being set on every ret == 0 path. */

Frauschi (commented):
## 🐺 Skoll Code Review

**Findings:** 10 total — 5 posted, 5 skipped

### Posted findings
- [Medium] **ed448 range check is now correct only because ed448_is_small_order() tabulates y == p** — `wolfcrypt/src/ed448.c:1539-1548`
- [Medium] **tests/api/test_ed448.c: the MC/DC stanzas now document and target deleted code** — `tests/api/test_ed448.c:1339-1370`
- [Medium] **mp_init_multi now discards every mp_init() result and is structurally incapable of reporting failure** — `wolfcrypt/src/integer.c:112-120`
- [Medium] **DoPKCS12Hash: this partially reverts hardening commit f15199906 from four months ago** — `wolfcrypt/src/pwdbased.c:376`
- [Medium] **No tests or assertions added for removals that rest on invariants owned by another function** — `wolfcrypt/src/integer.c:1772`

- [Medium] s_mp_add: the tmpc guard exists to silence clang-analyzer, which cannot carry mp_grow()'s postcondition across the call
- [Medium] wc_mlkemkey_check_h(): this one is a cross-function invariant assertion, not dead error handling
- [Low] wc_hss_reload_key: this hunk deletes a store as well as an operand, so the object-code-equivalence framing does not apply
- [Low] sha3.c: the three PSOC6 hunks are not compile-verified by any build
- [Low] kdf.c: the KDA-KDF CMAC loop guard no longer carries an error exit

</details>

wolfcrypt/src/integer.c
danielinux: addressed in MC/DC coverage campaign

wolfcrypt/src/integer.c
danielinux: See commit comments. mp_init() has only one failure possible which is x=NULL, which is already guarded earlier in the code. This comment does not take that into account.

wolfcrypt/src/ed448.c
danielinux: This is the same analysis the commit makes, so we agree on the facts:

> Proof: reaching :1554 requires the preceding range walk to have found p[56..29] all 0xff (loop at :1533), p[28] == 0xfe (:1543) and p[27..1] all 0xff (loop at :1546). If p[0] were also 0xff the input would be exactly the y = p encoding modulo byte 56 -- which ed448_is_small_order masks -- so :1511 would have set ret = PUBLIC_KEY_E and the else-if arm containing :1554 would never have been entered.

and the machine check:

> the byte walk pins 56 of the 57 public-key bytes before the final test, leaving p[0] as the only free variable, so all 256 values were enumerated against the real ed448_is_small_order(). Two values (p[0] = 0xff, y = p; and p[0] = 0xfe, y = p-1) are rejected early by the table; the other 254 reach the final test and all satisfy p[0] < 0xff. Zero counterexamples over the complete state space.

The substantive ask -- a note where the invariant is produced -- is taken. `ed448_is_small_order()` now carries:

```c
     * wc_ed448_check_key() depends on the y = p row: its Y-range test
     * accepts that encoding, so dropping the row would let a y outside
     * [0, p - 1] through. */
```

The consumer-side comment is tightened at the same time to name `y == p` as the single encoding involved. `y == p - 1` is a canonical, in-range value that the old byte walk accepted as well (its `p[0] < 0xff` test is satisfied by `0xfe`); it is rejected as an order-2 point, not as out of range, so it is not part of what moved to the table.

wolfcrypt/src/integer.c
danielinux: To add to the above, on the convention point specifically: this commit does not introduce a new convention, it removes an outlier. Both other math backends already implement the same API exactly this way.

`wolfcrypt/src/sp_int.c`, `sp_init_multi()` (the default backend, `#define mp_init_multi sp_init_multi`, sp_int.h:1361):

```c
if (n1 != NULL) { _sp_init_size(n1, SP_INT_DIGITS); }
...
return MP_OKAY;
```

`wolfcrypt/src/tfm.c:4390`, `mp_init_multi()` (fastmath):

```c
if (a) fp_init(a);
...
return MP_OKAY;
```

Both NULL-guard each operand, call a `void` initialiser and return `MP_OKAY` unconditionally. So the call-site guards you list (`integer.c:1089`, `:1233`) were already vacuous in every default and fastmath build before this PR; heapmath was the only configuration where they were not, and only because of the six branches this commit shows are unreachable.

The callee-body dependency is real, and the commit flags it rather than hiding it:

> Caveat, since this is the one proof here that rests on the callee's body rather than a guard in the same function: mp_init is public API whose contract permits failure, so mp_init_multi now depends on integer.c's mp_init staying failure-free. Kept as the last commit of the non-FIPS group so it can be dropped on its own.

One clarification now in that message: `sp_init()` also fails only on a NULL argument (`sp_int.c`, `if (a == NULL) err = MP_VAL;`), so there is no backend in the tree today whose init can fail on a non-NULL operand.

The reasoning is recorded at the site as well:

```c
    /* mp_init() has exactly one failure mode, a NULL argument, and the guard
     * on each call excludes it, so every one of these returns MP_OKAY and the
     * result is discarded. Same shape as sp_init_multi() and tfm.c's
     * mp_init_multi(), which also return MP_OKAY unconditionally. */
```

The `XMEMSET` observation is correct and points the other way: with the memset above producing `dp = NULL, used = 0, alloc = 0, sign = MP_ZPOS`, and `wc_bigint_init()` only zeroing `raw`, the `mp_init()` calls are already redundant. They are kept so the function still reads as an init.

On the suggested rewrite: the checked-but-collapsed form leaves the same six `!= MP_OKAY` operands with unreachable independence pairs, which is what the MC/DC campaign flagged in the first place. If wolfSSL prefers defence in depth here, the cleaner answer is to drop this commit -- it is deliberately last in the non-FIPS group, so that costs nothing else.

wolfcrypt/src/pwdbased.c
danielinux: The attribution is wrong: `f15199906` did not add this guard, it re-parenthesised one that has been there since at least 2018. Its own diff shows the reformat, not an addition:

```
-    if (buffer == NULL || Ai == NULL) {
+    if ((buffer == NULL) || (Ai == NULL)) {
```

The unparenthesised form is present in `v3.15.0-stable` (`wolfcrypt/src/pwdbased.c`, in `DoPKCS12Hash`) and in every tag I checked since -- v4.0.0, v4.8.0, v5.0.0, v5.6.0, v5.8.0. `git log -S "(buffer == NULL) || (Ai == NULL)"` matches only the reformatted spelling, which is why it points at `f15199906`; that commit ("SP int: fixes from AI review") re-implemented `wc_PKCS12_PBKDF()` without MP and changed this signature to take `enum wc_HashType` directly. So there is no four-month-old hardening change being reverted here, and the inferred static-analysis motivation does not follow from it either.

The unreachability itself we agree on; the commit enumerates the same four paths:

> four paths reach the function (2 call sites x WOLFSSL_SMALL_STACK on and off) and none can pass a NULL: Ai / B -- SMALL_STACK: XMALLOC at :479 / :741, each immediately followed by `if (... == NULL) return MEMORY_E;`. Otherwise a stack array [...] buffer -- initialised to staticBuffer (:431 / :687) and only replaced by the XMALLOC at :513 / :748, each immediately followed by `if (buffer == NULL) { ...free...; return MEMORY_E; }`.

The caller obligation is now stated in the function header rather than restated as a runtime test:

```c
/* helper for PKCS12_PBKDF(), does hash operation.
 * buffer and Ai are guaranteed non-NULL by the caller: each is either a stack
 * array or an XMALLOC result whose failure returns MEMORY_E before the call. */
```

On the analyzer concern: CI on this PR is green across all legs, including `make and analyze` and the `analyze` cross-compile matrix.

wolfcrypt/src/integer.c
danielinux: To be concrete about where the five invariants are stated: each is written out in the commit that consumes it, with the enumeration behind it --

- ed448 / `small_order_y[]`: all 256 values of `p[0]` enumerated against the real `ed448_is_small_order()`, "Zero counterexamples over the complete state space."
- mlkem `H_SET`: "Verified instead by enumerating every write to key->flags in the translation unit -- all nine, at wc_mlkem.c:453, 621, 852, 1001, 1003, 2283, 2327, 2430 and 2794 [...] the sole flag write reachable from it is :2794, which SETS MLKEM_FLAG_H_SET on ret == 0."
- `mp_grow` post-condition: "mp_grow has exactly three exits, enumerated in full [...] Hence MP_OKAY implies a->dp != NULL." The commit also notes this "needs no data-structure invariant -- mp_grow's own post-condition suffices."
- `DoPKCS12Hash` call sites: the four-path enumeration quoted in the other thread.
- `wolfSSL_Realloc`: the two `res` writes are shown to lie in arms mutually exclusive with the site, checked "against every combination of WOLFSSL_STATIC_MEMORY, WOLFSSL_STATIC_MEMORY_LEAN, WOLFSSL_DEBUG_MEMORY, WOLFSSL_HEAP_TEST, WOLFSSL_NO_MALLOC, SINGLE_THREADED and WOLFSSL_MALLOC_CHECK."

The point I take is that a commit message is not where a future refactorer will look. The two that cross function boundaries in a non-obvious way now carry a note at the provider: `small_order_y[]` in ed448.c, and the `key->flags |= MLKEM_FLAG_H_SET` site in `wc_MlKemKey_EncodePublicKey()`. `mp_init_multi` and `DoPKCS12Hash` state their assumption at their own site; `s_mp_add` relies only on `mp_grow`'s documented return values.

A test cannot cover any of these by construction: the removed branches have no reachable input, which is the property being asserted. What the campaign provides is the negative evidence -- `llvm-cov` MC/DC records each of these conditions with no covered independence pair across the variant matrix, cited per commit.

danielinux:
Thanks -- replies inline on the four line comments. Two things need answering here rather than on a diff line.

**The `tests/api/test_ed448.c` finding didn't post.** It is listed in the review summary but GitHub dropped it, since that file was not in the diff. It was correct, and the branch now carries a commit for it. `test_wc_ed448_check_key_decisions()` had a stanza whose comment read "*Same construction but with an extra byte (p[1]) perturbed so the second range-check loop exits early with ret == 0 before the final byte compare runs*", and both that loop and that compare are removed here. The assertions were loose enough (`rc == 0 || rc == PUBLIC_KEY_E`) that nothing failed, which is exactly the rot being pointed at. The three stanzas now target the two surviving decisions:

- every byte above the 0xFE position 0xff, that byte 0xfe: the walk runs to exhaustion and the 0xFE compare is reached and taken;
- the same value with a byte inside the walked range cleared: the walk breaks with `ret == 0`, covering the byte compare's true side and the following `(ret == PUBLIC_KEY_E)` guard's false side;
- every byte 0xff including the 0xFE position: no table row matches, the walk finds nothing below 0xff, the compare is false and the key is rejected as out of range -- an exact `PUBLIC_KEY_E` assertion rather than a curve-decode-dependent one.

**On the five skipped findings**, for the record, since four are already answered in the commit bodies:

- *`wc_hss_reload_key` deletes a store as well as an operand, so the object-code-equivalence framing does not apply* -- that framing is not claimed for this commit. It says "gcc does not fold this -- it must assume the call may alias the bitfield" and argues from the pointer member instead. The deleted store is covered too: "The `priv_key->inited = 0;` at :3593 goes with it: it was only ever observed by this operand, and :3620 assigns inited unconditionally on every path out." That final `priv_key->inited = (ret == 0);` runs on the expand-failure path as well, so a failed reload still leaves the key marked uninitialised.
- *sha3 PSOC6 hunks are not compile-verified* -- stated in the commit ("Not compile-verified: PSOC6_HASH_SHA3 needs the Cypress SDK, which is not available on the host build matrix") and given its own section in the PR description.
- *kdf loop guard no longer carries an error exit* -- "The `break` is what terminates the loop on error and is kept; ret is still tested at :1672 and returned."
- *mlkem `check_h` is a cross-function invariant assertion, not dead error handling* -- agreed, and the commit says so: "The comment called it an 'implementation issue' check, i.e. an assertion against a future encoder that forgets the flag; it is being removed rather than retained because no call path can reach it." The flag site now carries the reciprocal note.
- *`s_mp_add`: the tmpc guard exists to silence clang-analyzer* -- checked, and it does not. clang 21's analyzer over `integer.c` in an `--enable-heapmath` build, with the default checkers (which include `core.NullDereference`), reports nothing -- with the guard and without it. Repeated with `-analyzer-config ipa=none`, which forces `s_mp_add` to be analyzed as a top-level function so `mp_grow()`'s post-condition is not carried across the call, i.e. the exact condition the finding describes: still clean both ways.

SparkiDev (approved):

danielinux:
Jenkins retest this please

some tests timed out

Frauschi (approved):

PR #11029, opened 2026-07-31
Fix various issues

kareem-wolfssl:
# Description

Fixes zd#22229

# Testing

Built in tests, provided reproducers

# Checklist

 - [X] added tests
 - [ ] updated/added doxygen


kareem-wolfssl:
Retest this please

dgarske:
Jenkins retest this please

wolfcrypt/src/pkcs7.c
```diff
@@ -7510,26 +7510,31 @@ static int PKCS7_VerifySignedData(wc_PKCS7* pkcs7, const byte* hashBuf,
                 pkcs7->content   = pkcs7->contentDynamic;
             }
 
-            /* check if bundle has more elements or footer, if not, set content
-             * to pkcs7->content and hash to pkcs7->hash.
+            /* expect data length to be enough to check set and seq of certs,
```
dgarske, line 7520: Recommendation: Do not merge as-is. The residual-byte expression `(maxLen - totalRd) + stream->length` is not a valid measure of "bytes left to parse" at the point stage 3 hands off: stage 3 does not call `wc_PKCS7_StreamEndCase()` before computing it, so `totalRd` (advanced when bytes are copied into the stream buffer) and `stream->length` (bytes still buffered, some already consumed by `wc_PKCS7_HandleOctetStrings`) are not in sync, and the resulting `expected` can be smaller or larger than what stages 4/5 actually need. Either sync the stream bookkeeping before computing the residual, or keep the residual purely as an emptiness test (`remaining == 0` -> PKCS7_NO_SIGNER_E) and reinstate master's behavior for a short-but-nonzero footer instead of clamping the stage-4 window to it. Whichever route, add the chunk-size sweep above -- the current one-shot-only tests cannot see this class of bug.

src/x509.c
dgarske: Recommendation: Route this through `CheckIPAddr()` now that it accepts an explicit length, and delete the stale comment. If the direct `CheckHostName()` call is deliberate for some other reason, update the comment to state the real reason.

dgarske (changes requested):

wolfcrypt/src/pkcs7.c
```diff
@@ -7513,20 +7513,41 @@ static int PKCS7_VerifySignedData(wc_PKCS7* pkcs7, const byte* hashBuf,
             /* check if bundle has more elements or footer, if not, set content
```
dgarske: **🔴 [High] Streaming SignedData verifier now reports truncated 6-11 byte footers as successfully verified**

The PR closes the `remaining == 0` hole but simultaneously **widens** the 'treat as degenerate end, return success' window. Master short-circuited only when the residual was `< ASN_TAG_SZ + MAX_LENGTH_SZ` (6 bytes); the new code short-circuits when it is `< (ASN_TAG_SZ + MAX_LENGTH_SZ) * 2` (12 bytes). Residuals of 6..11 bytes previously fell through to stage 4, whose `expected` read window of 12 bytes could never be satisfied, so `wc_PKCS7_VerifySignedData()` returned `WC_PKCS7_WANT_READ_E` (-270) - an error. They now return `0`, i.e. *verified*, for a bundle that contains no `signerInfos` field at all and a truncated/garbage `certificates [0]` section.

I built both this branch and the merge-base (`65836b4069`) with `--enable-pkcs7` and measured a SignedData carrying a non-empty `digestAlgorithms` SET, an `eContent`, and a footer of `A0 \<n> 30 ... 30` (no signerInfos, no valid cert):

| footer bytes | master | this PR |
|---|---|---|
| 0 | `0` (accepted) | `-269` PKCS7_NO_SIGNER_E (fixed) |
| 2..5 | `0` | `0` |
| **6..11** | **`-270` WC_PKCS7_WANT_READ_E** | **`0` ACCEPTED** |
| >= 12 | `-132` ASN_PARSE_E | `-132` ASN_PARSE_E |

So for footers of 6-11 bytes the change converts a hard error into a positive verification result. The mitigating factor is that `wc_PKCS7_AllowDegenerate(pkcs7, 0)` returns `-269` across the whole 0..11 range - but that is not the default, and this is a security-hardening PR whose net effect on this code path should be strictly a tightening. It also widens the streaming-vs-non-streaming divergence already flagged by the TODO in…

**Fix:** Do not let the 6..11 byte residual reach `ret = 0`. Either keep the original `ASN_TAG_SZ + MAX_LENGTH_SZ` threshold for the success branch (so the change is purely additive tightening at `remaining == 0`), or - better - parse the residual footer inside stage 3 as the test TODO contemplates, so only a genuinely parseable degenerate end (`31 00` / absent optional fields) returns 0 and everything else returns a parse error. Re-run the empirical footer-size sweep against master before and after to confirm no size class moves from error to success.

wolfssl/wolfcrypt/settings.h
```diff
@@ -407,6 +407,17 @@
     #define WOLF_CRYPT_FIPS_H
 #endif
 
+/* A configure-generated options.h is sometimes copied in as user_settings.h to
```
dgarske, line 410: **🔵 [Low] settings.h silently discards a detected header result with asymmetric Windows guards**

Placement is correct - the block sits at line 410, after the `#include "user_settings.h"` at line 375 - so a POSIX-generated `options.h` copied in as `user_settings.h` really does get scrubbed, and both library and application TUs see the same result (`wolfio.h`, `callbacks.h` and `wc_port.h` all reach `settings.h` first, so there is no layout skew).

Two things worth a second look:
1. The guards are asymmetric - `_MSC_VER` for `HAVE_SYS_TIME_H` but `_WIN32` for `HAVE_SYS_UN_H`. The commit message explains it (MinGW has `sys/time.h` but not `sys/un.h`), but the code comment does not spell out *why the guards differ*, which is the part a future reader will trip on.
2. It is an unconditional override with no diagnostic. If a Windows-hosted toolchain does ship `\<sys/un.h>`, `configure`/CMake will detect it, write it into `options.h`, and `settings.h` will then silently drop AF_UNIX BIO support with no warning. This also permanently encodes a workaround for a user mistake (copying `options.h` as `user_settings.h`) into the platform switchboard.

**Fix:** Extend the comment to state why the two guards differ (MSVC-only for time, all of Windows for un), so the asymmetry does not look like an oversight. Optionally allow an explicit opt-out (e.g. skip the undef if the user defined the macro deliberately) for Windows toolchains that do provide the headers.

wolfcrypt/src/rsa.c
```diff
@@ -5086,6 +5086,12 @@ static int wc_CompareDiffPQ(mp_int* p, mp_int* q, int size, int* valid)
     else
         ret = 0;
 
+    /* zeroed so the cleanup below no-ops if the init is skipped */
```
dgarske: **🔵 [Low] Added mp_int zeroing in rsa.c is dead code - the failure path already frees and NULLs**

Unlike the equivalent changes in `srp.c`, `sakke.c` and `ecc.c` - which are genuinely needed, because there a partial allocation failure leaves an allocated-but-uninitialized `mp_int` that the cleanup then `mp_clear()`s - the two `rsa.c` sites already handle this. Both allocation blocks `XFREE()` everything and set every pointer to `NULL` on failure (with a comment explaining exactly why). So on the `MEMORY_E` path all pointers are `NULL` and the new `if (x != NULL) XMEMSET(...)` guards evaluate false; on the success path they run immediately before `mp_init_multi()`, which fully initializes the structs anyway. The memsets can never change behavior.

Minor secondary cost: with SP math `sizeof(mp_int)` is the full `sp_int` including the digit array (several KB when large RSA sizes are configured), so this zeroes a few KB per call on the keygen path for no effect.

**Fix:** Drop the two added memset blocks in `rsa.c`, or - if the intent is one uniform idiom across all five files - drop the now-redundant free-and-NULL block instead so there is exactly one mechanism per site. Leave the `srp.c` / `sakke.c` / `ecc.c` additions as they are; those close real uninitialized-`mp_clear()` paths (e.g. `wc_SrpGetPublic`, where `i` succeeds and `j` fails, and `accel_fp_mul2add`, where `tka` succeeds and `tkb` fails before `goto done`).

---
> **Note:** Referenced line (`wolfcrypt/src/rsa.c:5089-5094, 5362-5369`) is outside the diff hunk. Comment anchored to nearest changed region.

dgarske (commented):
## Skoll Code Review

**Findings:** 9 total — 3 posted, 6 skipped

### Posted findings
- [High] **Streaming SignedData verifier now reports truncated 6-11 byte footers as successfully verified** — `wolfcrypt/src/pkcs7.c:7513-7550`
- [Low] **Added mp_int zeroing in rsa.c is dead code - the failure path already frees and NULLs** — `wolfcrypt/src/rsa.c:5089-5094, 5362-5369`
- [Low] **settings.h silently discards a detected header result with asymmetric Windows guards** — `wolfssl/wolfcrypt/settings.h:410-420`

### Skipped findings
- [Medium] ~~New PKCS#7 tests do not cover the 1..11 byte residual window the change introduces~~
- [Medium] ~~New residual arithmetic is unsound for indefinite-length bundles and is skipped entirely when maxLen  totalRd~~
- [Medium] ~~HAVE_SYS_UN_H / HAVE_SYS_TIME_H propagation is an intentional ABI change that needs a release note~~
- [Medium] ~~callbacks.h typedefs struct timeval without including sys/time.h, now exposed to every application TU~~
- [Low] ~~New PKCS#7 code inserted between a doc comment and the function it documents~~
- [Low] ~~Bare brace block introduced solely to declare a variable~~

configure.ac
```diff
@@ -221,9 +221,14 @@ then
     fi
 fi
 
-AC_CHECK_HEADERS([arpa/inet.h fcntl.h netdb.h netinet/in.h stddef.h time.h sys/ioctl.h sys/socket.h sys/time.h errno.h sys/un.h ctype.h sys/random.h])
+AC_CHECK_HEADERS([arpa/inet.h fcntl.h netdb.h netinet/in.h stddef.h time.h sys/ioctl.h sys/socket.h errno.h ctype.h sys/random.h])
```
dgarske: **🔴 [High] Moving sys/time.h out of AC_CHECK_HEADERS silently breaks the AC_CHECK_DECLS prologue and drops HAVE_GETTIMEOFDAY from…**

The diff removes `sys/time.h` (and `sys/un.h`) from the plural `AC_CHECK_HEADERS` list at line 224 and re-checks them with the singular `AC_CHECK_HEADER` at lines 230-231. `AC_CHECK_HEADERS` (plural) is what emits `#define HAVE_SYS_TIME_H 1` into autoconf's `confdefs.h`; `AC_CHECK_HEADER` (singular) with a custom action-if-found does **not**. `confdefs.h` is prepended to every subsequent conftest, and the `AC_CHECK_DECLS` invocation at configure.ac:249 uses a prologue that is guarded on exactly that macro:

```
#ifdef HAVE_SYS_TIME_H
    #include <sys/time.h>
#endif
```
(configure.ac:270-272)

With `HAVE_SYS_TIME_H` no longer in `confdefs.h`, `\<sys/time.h>` is never included, so `AC_CHECK_DECLS([... gettimeofday ...])` fails. The action-if-not-declared block at configure.ac:250-258 then deliberately *supersedes* the earlier positive `AC_CHECK_FUNCS` result: it sets `ac_cv_func_gettimeofday=no` and `sed`s `#define HAVE_GETTIMEOFDAY 1` out of `confdefs.h`.

I verified this empirically on this branch (Linux x86_64, glibc, gcc 13). `./configure` now produces in `config.h`:
```
#define HAVE_DECL_GETTIMEOFDAY 0
/* #undef HAVE_GETTIMEOFDAY */
```
and `config.log` shows the conftest failing with `error: 'gettimeofday' undeclared`, with the dumped `confdefs.h` listing `HAVE_TIME_H`, `HAVE_SYS_IOCTL_H`, `HAVE_SYS_SOCKET_H` but no `HAVE_SYS_TIME_H`. Before this PR `sys/time.h` was in the same `AC_CHECK_HEADERS` list as those, so the prologue compiled and both macros were set. Configure also now prints a spurious `note: earlier check for gettimeofday superseded.`

This is an…

**Fix:** Keep `sys/time.h` (and `sys/un.h`) in the `AC_CHECK_HEADERS` list and only *additionally* append `-DHAVE_SYS_TIME_H=1` / `-DHAVE_SYS_UN_H=1` to `AM_CPPFLAGS` via the singular checks (the cached result means no extra compile). That gets the macros into `wolfssl/options.h` as intended without dropping them from `confdefs.h`/`config.h`. Re-run `./configure` afterwards and confirm `config.h` again contains `#define HAVE_GETTIMEOFDAY 1` and `#define HAVE_DECL_GETTIMEOFDAY 1`.

wolfcrypt/src/pkcs7.c
```diff
@@ -7636,7 +7656,76 @@ static int PKCS7_VerifySignedData(wc_PKCS7* pkcs7, const byte* hashBuf,
                 contentSz = pkcs7->contentSz;
             }
         #endif /* !NO_PKCS7_STREAM */
-            wc_PKCS7_ChangeState(pkcs7, WC_PKCS7_VERIFY_STAGE4);
+            wc_PKCS7_ChangeState(pkcs7, shortFooter ?
+                    WC_PKCS7_VERIFY_STAGE3_FOOTER : WC_PKCS7_VERIFY_STAGE4);
+
+            FALL_THROUGH;
+
+        case WC_PKCS7_VERIFY_STAGE3_FOOTER:
```
dgarske: **🟠 [Medium] New WC_PKCS7_VERIFY_STAGE3_FOOTER reads from in and ignores the separate footer buffer in2**

The new short-footer stage calls `wc_PKCS7_AddDataToStream(pkcs7, in, inSz, ...)` unconditionally. Every other stage that reads footer bytes first selects the caller-supplied footer buffer when `wc_PKCS7_VerifySignedData_ex()` was used with a separate `pkiMsgFoot`:

```c
/* WC_PKCS7_VERIFY_STAGE4, pkcs7.c:7734-7743 (also STAGE6 at :8088) */
if (in2 && in2Sz > 0) { src = in2; srcSz = in2Sz; pkiMsg2 = in2; }
else                  { src = in;  srcSz = inSz;  pkiMsg2 = in;  }
ret = wc_PKCS7_AddDataToStream(pkcs7, src, srcSz, pkcs7->stream->expected, &pkiMsg2, &idx);
```

Stage 3 computes `remaining` from `pkcs7->stream->maxLen` (the *whole* outer ContentInfo length, read in stage 1) minus `totalRd` (bytes consumed from `in` only), so on the `_ex` path `remaining` covers the bytes that live in `in2`. If `remaining` ever lands under 12, `shortFooter` is set and the new stage then parses the wrong buffer.

In practice a real `_ex` footer carries a SignerInfo with a signature and is far larger than 12 bytes, so I could not construct a reachable case - hence SUGGEST rather than BLOCK. But the asymmetry is a latent trap: any future change that widens the `shortFooter` threshold, or a detached bundle with an unusually small footer, silently reads content bytes as the footer and reports `ASN_PARSE_E` (or spins on `WC_PKCS7_WANT_READ_E` if `in` is exhausted) instead of parsing the real footer.

**Fix:** Mirror the `in2`/`in2Sz` selection used by `WC_PKCS7_VERIFY_STAGE4` and `WC_PKCS7_VERIFY_STAGE6` before calling `wc_PKCS7_AddDataToStream()`, and add a `wc_PKCS7_VerifySignedData_ex()` case to the chunk sweep so the `_ex` footer path is exercised.

tests/api/test_pkcs7.c
```diff
@@ -6961,6 +7018,134 @@ static int pkcs7_build_digestparam_mismatch(byte* cert, word32 certSz,
...
+/*
+ * The streaming verifier decides at the stage 3 handoff whether the bundle has
+ * a footer left, and the amount of input handed over per call changes which
+ * read windows are satisfied from the caller's buffer and which from the
+ * internal one. Sweep every chunk size over a well formed bundle, which must
+ * verify however it is split, and over a bundle whose outer ContentInfo ends
+ * with the eContent, which carries no signerInfos field at all and must never
+ * be reported as verified.
+ */
+int test_wc_PKCS7_VerifySignedData_ChunkSweep(void)
```
dgarske: **🟠 [Medium] New footer-parser tests are gated behind RSA/USE_CERT_BUFFERS_2048 even though most of them need neither**

The whole of `test_wc_PKCS7_VerifySignedData_ChunkSweep` is gated on `HAVE_PKCS7 && !NO_PKCS7_STREAM && !NO_RSA && !NO_SHA256 && USE_CERT_BUFFERS_2048`. Only the first sweep (`msg`, built by `pkcs7_sign_digest_params()`) actually needs RSA and the 2048-bit cert buffers. The three sweeps that exercise the code this PR adds - `noSigner` (the `remaining == 0` -> `PKCS7_NO_SIGNER_E` path), `footTrunc` (malformed short footer -> `ASN_PARSE_E`), and `footEmptySi` (well-formed empty signerInfos -> degenerate accept) - are static byte arrays that need no crypto at all.

Consequence: in an ECC-only or `NO_RSA` streaming build, none of the new `WC_PKCS7_VERIFY_STAGE3_FOOTER` logic is covered. `test_wc_PKCS7_VerifySignedData_NoSignerInfos` still covers the `remaining == 0` case (it is gated only on `HAVE_PKCS7 && !NO_SHA256`), but the footer *parser* itself - the `A0 00` / `A1 00` / `31 00` shape checks and the `footIdx != footEnd` trailing-byte check at pkcs7.c:7682-7713 - goes entirely untested there. I confirmed the tests pass on this branch under `--enable-pkcs7 --enable-opensslall` (pkcs7_sd group: 23/23 pass), so this is about coverage breadth, not a failure.

**Fix:** Split the test body into two `#if` blocks so the three static-blob sweeps run under `HAVE_PKCS7 && !NO_PKCS7_STREAM && !NO_SHA256` and only the well-formed-bundle sweep keeps the `!NO_RSA && USE_CERT_BUFFERS_2048` gate.

tests/api/test_pkcs7.c
```diff
@@ -6961,6 +7018,134 @@ static int pkcs7_build_digestparam_mismatch(byte* cert, word32 certSz,
  * mismatch). The fix is symmetric, so both directions are exercised, over both
  * the attribute-free and signed-attribute signing paths.
  */
+#if defined(HAVE_PKCS7) && !defined(NO_PKCS7_STREAM) && !defined(NO_RSA) && \
+    !defined(NO_SHA256) && defined(USE_CERT_BUFFERS_2048)
+/* Feed der[0..derSz) to the verifier in fixed size chunks, mimicking a caller
+ * that streams the bundle in. Returns the final return code, which is
+ * WC_PKCS7_WANT_READ_E if the parser never reached a decision. */
+static int pkcs7_verify_chunked(const byte* der, word32 derSz, word32 chunkSz)
```
dgarske: **🟠 [Medium] pkcs7_verify_chunked setup failures are indistinguishable from a successful rejection**

The helper returns `MEMORY_E` when `wc_PKCS7_New()` fails and `BAD_FUNC_ARG` when `wc_PKCS7_Init()`/`wc_PKCS7_InitWithCert()` fails. Both are non-zero, so the two negative sweeps - `ExpectIntNE(pkcs7_verify_chunked(noSigner, ...), 0)` at :7129 and `ExpectIntNE(pkcs7_verify_chunked(footTrunc, ...), 0)` at :7135 - will report *pass* if the PKCS7 object could not even be created. If a future refactor breaks `wc_PKCS7_InitWithCert(pkcs7, NULL, 0)`, these two sweeps silently stop testing anything while still going green.

The two `ExpectIntEQ(..., 0)` sweeps do not have this problem.

**Fix:** Either assert the setup separately (e.g. return the setup status through an out-parameter and `ExpectIntEQ(setupOk, 1)` in the caller), or reserve a sentinel return value for setup failure and assert against it explicitly in the negative sweeps.

wolfssl/wolfcrypt/pkcs7.h
```diff
@@ -102,6 +102,7 @@ enum PKCS7_STATE {
 
     WC_PKCS7_VERIFY_STAGE2,
     WC_PKCS7_VERIFY_STAGE3,
+    WC_PKCS7_VERIFY_STAGE3_FOOTER,
```
dgarske: **🔵 [Low] New enumerator inserted mid-enum in the public PKCS7_STATE, renumbering all later states**

`WC_PKCS7_VERIFY_STAGE3_FOOTER` is inserted between `WC_PKCS7_VERIFY_STAGE3` and `WC_PKCS7_VERIFY_STAGE4` in a public header, which shifts the numeric value of `WC_PKCS7_VERIFY_STAGE4` and every enumerator after it (STAGE5/6/7, the whole INFOSET, ENV, AUTHENV and DECRYPT_* blocks). The project convention for public types is to append rather than insert mid-structure so existing values stay stable.

Practical impact is small here: the value is only stored in `word32 state;` inside `wc_PKCS7` (pkcs7.h:361), which is a per-process runtime field, and the enum is not serialized. So this is a convention/hygiene note, not a live break - but out-of-tree code that switches on or logs these values (or an application built against an older header and linked against a newer library) would see a mismatch.

**Fix:** Append `WC_PKCS7_VERIFY_STAGE3_FOOTER` at the end of `enum PKCS7_STATE` (the `case` label position in `PKCS7_VerifySignedData()` does not need to match the enum order) so existing enumerator values are preserved.

src/internal.c
```diff
@@ -16070,7 +16070,8 @@ int DoVerifyCallback(WOLFSSL_CERT_MANAGER* cm, WOLFSSL* ssl, int cert_err,
         /* perform IP address check on the peer certificate */
         if ((args->dCertInit != 0) && (args->dCert != NULL) && (ssl != NULL) &&
```
dgarske, line 16102: **🔵 [Low] XSTRLEN(ssl-param-ipasc) computed twice in DoVerifyCallback**

The new explicit-length call re-walks the string that the guard just measured. `ssl->param->ipasc` is a `WOLFSSL_MAX_IPSTR` char array, so the cost is bounded, but the double scan is avoidable and the two calls could in principle disagree if anything between them touched the buffer.

**Fix:** Compute the length once into a local (declared at the top of the function per the project C89 style) and reuse it for both the guard and the `CheckIPAddr()` argument.

dgarske (commented):
## Skoll Code Review

**Findings:** 9 total — 6 posted, 3 skipped

### Posted findings
- [High] **Moving sys/time.h out of AC_CHECK_HEADERS silently breaks the AC_CHECK_DECLS prologue and drops HAVE_GETTIMEOFDAY from config.h** — `configure.ac:224-231`
- [Medium] **New WC_PKCS7_VERIFY_STAGE3_FOOTER reads from in and ignores the separate footer buffer in2** — `wolfcrypt/src/pkcs7.c:7664-7680`
- [Medium] **New footer-parser tests are gated behind RSA/USE_CERT_BUFFERS_2048 even though most of them need neither** — `tests/api/test_pkcs7.c:7063-7147`
- [Medium] **pkcs7_verify_chunked setup failures are indistinguishable from a successful rejection** — `tests/api/test_pkcs7.c:7026-7047`
- [Low] **New enumerator inserted mid-enum in the public PKCS7_STATE, renumbering all later states** — `wolfssl/wolfcrypt/pkcs7.h:105`
- [Low] **XSTRLEN(ssl-param-ipasc) computed twice in DoVerifyCallback** — `src/internal.c:16071-16074`

### Skipped findings
- [Low] ~~Doc comment for test_wc_PKCS7_VerifySignedData_NoDigestParams is now detached from its function~~
- [Info] ~~Stray unrelated blank line in wc_CheckProbablePrime_ex~~
- [Info] ~~New brace scope opened purely to declare a test-local variable~~

dgarske (dismissed):

kareem-wolfssl:
The reported PKCS7 issue was fixed in #11017, so I dropped the PKCS7 changes from this branch.

dgarske:
Jenkins retest this please

dgarske (approved):

JacobBarthelmeh (approved):

PR #11033, opened 2026-08-02
Add support for wolfHAL crypto backend

AlexLanzano:
- Adds a wolfHAL port so wolfCrypt can offload symmetric crypto and entropy to a
    board's hardware accelerators.
- Uses the existing crypto callback framework rather than touching the algorithm
    implementations. wolfHAL registers as a crypto device at init and operations
    route to it by devId.
- AES ECB, CBC, GCM, and CCM are offloaded only when the board declares a device
    for that mode; everything else falls back to software. The RNG has no fallback,
    since it replaces the entropy source.
- The application supplies a `board.h` on the include path naming its devices,
    the same contract as `user_settings.h`. Stock wolfHAL board headers work as-is.
- Enabled with `--with-wolfhal=PATH --with-wolfhal-board=PATH`, both validated at
    configure time.
    
I have added examples using this new wolfHAL backend in wolfssl-examples. Here is the PR for that. https://github.com/wolfSSL/wolfssl-examples/pull/615

### DIFF
--- a/wolfcrypt/src/port/wolfHAL/wolfhal.c
+++ b/wolfcrypt/src/port/wolfHAL/wolfhal.c
@@ -1,0 +1,328 @@
+/* wolfhal.c
+ *
+ * Copyright (C) 2006-2026 wolfSSL Inc.
+ *
+ * This file is part of wolfSSL.
+ *
+ * wolfSSL is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * wolfSSL is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1335, USA
+ */
+
+#ifdef HAVE_CONFIG_H
+    #include <config.h>
+#endif
+
+#include <wolfssl/wolfcrypt/settings.h>
+
+#ifdef WOLFSSL_WOLFHAL
+
+/* Application-supplied, on the include path, same contract settings.h has
+ * with user_settings.h. Must come first: it brings in the wolfHAL platform
+ * driver headers (and with them the direct API mapping defines) and names the
+ * device for each algorithm. See wolfhal.h for what it has to provide. */
+#include "board.h"
+
+#include <wolfssl/wolfcrypt/error-crypt.h>
+#include <wolfssl/wolfcrypt/logging.h>
+#include <wolfssl/wolfcrypt/port/wolfHAL/wolfhal.h>
+#include <wolfssl/wolfcrypt/cryptocb.h>
+#include <wolfssl/wolfcrypt/aes.h>
+
+#include <wolfHAL/error.h>
+
+/* A stock wolfHAL board.h already names its devices BOARD_*_DEV. Accept those
+ * so such a board needs no wolfSSL-specific additions. */
+#if !defined(WC_WOLFHAL_AES_ECB_DEV) && defined(BOARD_AES_ECB_DEV)
+    #define WC_WOLFHAL_AES_ECB_DEV BOARD_AES_ECB_DEV
+#endif
+#if !defined(WC_WOLFHAL_AES_CBC_DEV) && defined(BOARD_AES_CBC_DEV)
+    #define WC_WOLFHAL_AES_CBC_DEV BOARD_AES_CBC_DEV
+#endif
+#if !defined(WC_WOLFHAL_AES_GCM_DEV) && defined(BOARD_AES_GCM_DEV)
+    #define WC_WOLFHAL_AES_GCM_DEV BOARD_AES_GCM_DEV
+#endif
+#if !defined(WC_WOLFHAL_AES_CCM_DEV) && defined(BOARD_AES_CCM_DEV)
+    #define WC_WOLFHAL_AES_CCM_DEV BOARD_AES_CCM_DEV
+#endif
+#if !defined(WC_WOLFHAL_RNG_DEV) && defined(BOARD_RNG_DEV)
+    #define WC_WOLFHAL_RNG_DEV BOARD_RNG_DEV
+#endif
+
+/* Offload gates. A mode is dispatched to hardware only when wolfCrypt enables it
+ * and board.h names a device for it. Anything else stays out of the dispatch,
+ * returns CRYPTOCB_UNAVAILABLE, and falls back to software, so a board offloads
+ * the subset its hardware covers without constraining the wolfSSL build. */
+#if defined(WOLF_CRYPTO_CB) && !defined(NO_AES)
+    #if defined(HAVE_AES_CBC) && defined(WC_WOLFHAL_AES_CBC_DEV)
+        #define WC_WOLFHAL_OFFLOAD_AES_CBC
+    #endif
+    #if defined(HAVE_AESGCM) && defined(WC_WOLFHAL_AES_GCM_DEV)
+        #define WC_WOLFHAL_OFFLOAD_AES_GCM
+    #endif
+    #if defined(HAVE_AESCCM) && defined(WC_WOLFHAL_AES_CCM_DEV)
+        #define WC_WOLFHAL_OFFLOAD_AES_CCM
+    #endif
+    #if defined(HAVE_AES_ECB) && defined(WC_WOLFHAL_AES_ECB_DEV)
+        #define WC_WOLFHAL_OFFLOAD_AES_ECB
+    #endif
+#endif
+
+/* No fallback for the RNG: CUSTOM_RAND_GENERATE_BLOCK replaces the entropy
+ * source outright rather than sitting behind a dispatch that can decline, so
+ * there is nothing to fall back to. */
+#if defined(WOLFSSL_WOLFHAL_RNG) && !defined(WC_WOLFHAL_RNG_DEV)
+    #error "board.h must define WC_WOLFHAL_RNG_DEV (or BOARD_RNG_DEV)"
+#endif
+
+static int wc_wolfHAL_TranslateError(whal_Error err)
+{
+    switch (err) {
+        case WHAL_SUCCESS:   return 0;
+        case WHAL_EINVAL:    return BAD_FUNC_ARG;
+        case WHAL_ENOTSUP:   return CRYPTOCB_UNAVAILABLE;
+        case WHAL_EHARDWARE: return WC_HW_E;
+        case WHAL_ETIMEOUT:  return WC_TIMEOUT_E;
+        default:             return WC_HW_E;
+    }
+}
+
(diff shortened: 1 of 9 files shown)


AlexLanzano:
retest this please

AlexLanzano:
Retest this please

JacobBarthelmeh:
Retest this please Jenkins

JacobBarthelmeh:
@dgarske and @danielinux I think it would be best to get your expert reviews on HAL implementation/integration. Please take a look at this PR.

wolfcrypt/src/port/wolfHAL/wolfhal.c
```diff
@@ -0,0 +1,328 @@
...
+ * there is nothing to fall back to. */
+#if defined(WOLFSSL_WOLFHAL_RNG) && !defined(WC_WOLFHAL_RNG_DEV)
+    #error "board.h must define WC_WOLFHAL_RNG_DEV (or BOARD_RNG_DEV)"
+#endif
+
+static int wc_wolfHAL_TranslateError(whal_Error err)
+{
+    switch (err) {
+        case WHAL_SUCCESS:   return 0;
+        case WHAL_EINVAL:    return BAD_FUNC_ARG;
```
danielinux, line 92: Only `WHAL_ENOTSUP` routes to `CRYPTOCB_UNAVAILABLE`, so it is the only code that produces a software fallback. `WHAL_EINVAL` becomes `BAD_FUNC_ARG`, which `aes.c` treats as fatal.

That makes the whole fallback story depend on an unwritten convention in an out-of-tree library. Concretely: a device that does AES-128 but not AES-256 (like DCP, see `dcp_port.c:368`) will hard-fail the handshake rather than fall back, if its driver returns `EINVAL` for the unsupported key size, which is the natural choice for that condition.

Could we document the contract in the port README ("decline with `ENOTSUP`, never `EINVAL`"), and ideally have the handlers check what they can before calling? I would avoid the shortcut of also mapping `EINVAL` to fallback, since that silently swallows genuine programming errors.

wolfcrypt/src/port/wolfHAL/wolfhal.c
```diff
@@ -0,0 +1,328 @@
...
+    #define WC_WOLFHAL_AES_GCM_DEV BOARD_AES_GCM_DEV
+#endif
+#if !defined(WC_WOLFHAL_AES_CCM_DEV) && defined(BOARD_AES_CCM_DEV)
+    #define WC_WOLFHAL_AES_CCM_DEV BOARD_AES_CCM_DEV
+#endif
+#if !defined(WC_WOLFHAL_RNG_DEV) && defined(BOARD_RNG_DEV)
+    #define WC_WOLFHAL_RNG_DEV BOARD_RNG_DEV
+#endif
+
+/* Offload gates. A mode is dispatched to hardware only when wolfCrypt enables it
```
danielinux: The offload gate is compile-time and per-mode only. Real accelerators decline at finer granularity: key length (DCP is AES-128 only, see the `keylen == 16` guards at `aes.c:6083` and `aes.c:7286`, which are what make its partial support work), GCM IV size (most hardware GCM is 12-byte-IV only), CCM nonce and tag sizes, buffer alignment and DMA-ability.

If this port is going to be the pattern for migrating the existing `wolfcrypt/src/port/*` drivers, a runtime capability check, `whal_Aes_Supported(dev, mode, keylen)` or equivalent, is what would let those ports move across without regressing.

wolfcrypt/src/port/wolfHAL/wolfhal.c
```diff
@@ -0,0 +1,328 @@
...
+
+#include <wolfssl/wolfcrypt/settings.h>
+
+#ifdef WOLFSSL_WOLFHAL
+
+/* Application-supplied, on the include path, same contract settings.h has
+ * with user_settings.h. Must come first: it brings in the wolfHAL platform
+ * driver headers (and with them the direct API mapping defines) and names the
+ * device for each algorithm. See wolfhal.h for what it has to provide. */
+#include "board.h"
```
danielinux: `#include "board.h"` from a library translation unit is risky: `board.h` is one of the most collision-prone names in embedded C (NXP MCUXpresso, Zephyr and mbed all ship one), and the `""` form searches the including file's directory first.

Consider `wolfssl_board.h`, or a `WOLFSSL_WOLFHAL_BOARD_H` macro the application sets to name its own header.

wolfcrypt/src/port/wolfHAL/wolfhal.c
```diff
@@ -0,0 +1,328 @@
...
+}
+
+void wc_wolfHAL_UnRegisterDevice(int devId)
+{
+    wc_CryptoCb_UnRegisterDevice(devId);
+}
+#endif /* WOLF_CRYPTO_CB */
+
+#ifdef WOLFSSL_WOLFHAL_RNG
+int wc_wolfHAL_GenerateBlock(byte* output, word32 sz)
```
danielinux: Wiring the TRNG in as `CUSTOM_RAND_GENERATE_BLOCK` replaces the Hash_DRBG outright: no reseed, no health tests, and output quality is whatever the peripheral gives. The in-tree convention that keeps the DRBG is `wc_GenerateSeed()` / `WC_RNG_SEED_CB`.

For something intended as the general HAL template, seeding the DRBG seems the safer default and keeps FIPS and entropy-assessment options open. If `CUSTOM_RAND_GENERATE_BLOCK` is deliberate, could the README say why?

wolfssl/wolfcrypt/port/wolfHAL/wolfhal_settings.h
```diff
@@ -0,0 +1,72 @@
...
+ * Setting both to different values is the deliberate multi-device case. */
+#ifndef WOLFSSL_WOLFHAL_DEVID
+    #ifdef WC_USE_DEVID
+        #define WOLFSSL_WOLFHAL_DEVID WC_USE_DEVID
+    #else
+        #define WOLFSSL_WOLFHAL_DEVID 0x5748 /* 'WH' for wolfHAL */
+    #endif
+#endif
+
+/* Let the unmodified wolfcrypt test and benchmark route every operation
```
danielinux: This says the point of mapping `WC_USE_DEVID` is to let "the unmodified wolfcrypt test and benchmark route every operation through this device", but `configure.ac` forces `ENABLED_CRYPT_TESTS="no"` whenever `--with-wolfhal` is set, so in an autotools build the test is never built. The mapping only pays off in the external example project.

Worth either scoping the comment to `user_settings.h`/IDE builds, or leaving crypttests buildable.

wolfcrypt/src/wc_port.c
```diff
@@ -582,6 +586,17 @@ int wolfCrypt_Init(void)
             WOLFCRYPT_INIT_RAISE_BAD_STATE();
         }
     #endif
+
+    /* Register the wolfHAL device so wolfCrypt operations route to the board's
+     * crypto accelerator. The board's peripherals must already be brought up
+     * by the application with whal_Board_Init(). */
+    #if defined(WOLFSSL_WOLFHAL) && defined(WOLF_CRYPTO_CB)
+        ret = wc_wolfHAL_RegisterDevice(WOLFSSL_WOLFHAL_DEVID);
```
danielinux, line 594: Registration is automatic here, but `wc_wolfHAL_UnRegisterDevice()` is never called from `wolfCrypt_Cleanup()`. Asymmetric: an init/cleanup/init cycle re-registers over a stale entry.

wolfssl/wolfcrypt/port/wolfHAL/wolfhal.h
```diff
@@ -0,0 +1,92 @@
...
+
+/* Crypto device callback for wolfHAL hardware acceleration */
+WOLFSSL_LOCAL int wc_wolfHAL_CryptoDevCb(int devId, struct wc_CryptoInfo* info,
+                                          void* ctx);
+#endif
+
+#ifdef WOLFSSL_WOLFHAL_RNG
+/* Generate a block of random data using the wolfHAL RNG device.
+ * Suitable for use as CUSTOM_RAND_GENERATE_BLOCK. */
+WOLFSSL_API int wc_wolfHAL_GenerateBlock(unsigned char* output,
```
danielinux: Declared `(unsigned char*, unsigned int)` but defined as `(byte*, word32)` in `wolfhal.c`. The same types on the current targets, but it would break where `word32` is `unsigned long`, and it is inconsistent with the rest of the API surface.

danielinux (changes requested):

wolfcrypt/src/port/wolfHAL/wolfhal.c
AlexLanzano: This will need to be resolved in the wolfHAL library itself since wolfHAL expects this to be named "board.h". I will put up a separate PR to rename this to "wolfHAL_board.h"

wolfcrypt/src/port/wolfHAL/wolfhal.c
AlexLanzano: Will replace the generate block implementation with generate seed instead

wolfcrypt/src/wc_port.c
AlexLanzano: `wolfCrypt_Cleanup` calls `wc_CryptoCb_Cleanup` which walks the entire registry and unregisters any registered devices. So this will get cleaned up automatically as well

wolfssl/wolfcrypt/port/wolfHAL/wolfhal.h
AlexLanzano: I removed this function in place of `wc_wolfHAL_GenerateBlock` which has consistent types for the definition and declaration

wolfssl/wolfcrypt/port/wolfHAL/wolfhal_settings.h
AlexLanzano: Removing this comment

wolfcrypt/src/port/wolfHAL/wolfhal.c
AlexLanzano: Documented this in the README.md.

wolfcrypt/src/port/wolfHAL/wolfhal.c
AlexLanzano: Hmmm, Im not sure if im following this. Wouldn't `whal_Aes_Supported` be essentially a redundant check we do in the other `whal_Aes*` functions? What's the issue with not calling `whal_Aes_Supported`, and just calling the `whal_Aes*` function and seeing if it returns WHAL_ENOTSUP or not?

wolfcrypt/src/port/wolfHAL/wolfhal.c
AlexLanzano: done

wolfssl/wolfcrypt/port/wolfHAL/wolfhal_settings.h
AlexLanzano: Done

wolfssl/wolfcrypt/port/wolfHAL/wolfhal.h
AlexLanzano: Done

wolfcrypt/src/port/wolfHAL/wolfhal.c
danielinux: I guess the point is automatic software rollback in this case, but as long as the errors are consistent in the contract ,this will be fine as well (if possible for the caller to repeat the call in SW if the algo mode is not supported by the HAL)

PR #11038, opened 2026-08-03
Deep copy revoked cert entry extensions in DupCRL_Entry()

gasbytes:
# Description

Deep copy revoked cert entry extensions in DupCRL_Entry() when CRL_STATIC_REVOKED_LIST and OPENSSL_EXTRA are enabled, and add a regression test for the X509_CRL_dup path.

# Testing

Added regression test test_wolfSSL_CRL_static_revoked_list_dup in the testsuite;

# Checklist

 - [x] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/src/crl.c
+++ b/src/crl.c
@@ -1411,6 +1411,37 @@ static CRL_Entry* DupCRL_Entry(const CRL_Entry* ent, void* heap)
         CRL_Entry_free(dupl, heap);
         return NULL;
     }
+#elif defined(OPENSSL_EXTRA)
+    {
+        int i;
+
+        /* certs is an in-struct array living after verifyMutex, so the bulk
+         * copy above aliased every extensions pointer with the original's.
+         * Cleaning the whole array first. */
+        for (i = 0; i < CRL_MAX_REVOKED_CERTS; i++) {
+            dupl->certs[i].extensions = NULL;
+            dupl->certs[i].extensionsSz = 0;
+        }
+
+        /* Deep copy the entry extensions, as DupRevokedCertList() does for the
+         * linked-list build. */
+        for (i = 0; i < ent->totalCerts; i++) {
+            if (ent->certs[i].extensions == NULL ||
+                    ent->certs[i].extensionsSz == 0) {
+                continue;
+            }
+            dupl->certs[i].extensions = (byte*)XMALLOC(
+                    ent->certs[i].extensionsSz, heap, DYNAMIC_TYPE_REVOKED);
+            if (dupl->certs[i].extensions == NULL) {
+                WOLFSSL_MSG("Failed to allocate revoked cert extensions");
+                CRL_Entry_free(dupl, heap);
+                return NULL;
+            }
+            XMEMCPY(dupl->certs[i].extensions, ent->certs[i].extensions,
+                    ent->certs[i].extensionsSz);
+            dupl->certs[i].extensionsSz = ent->certs[i].extensionsSz;
+        }
+    }
 #endif
 #ifdef OPENSSL_EXTRA
     dupl->issuer = wolfSSL_X509_NAME_dup(ent->issuer);
--- a/src/crl.c
+++ b/src/crl.c
@@ -1326,65 +1326,77 @@ static WOLFSSL_X509_CRL* wolfSSL_X509_crl_new(WOLFSSL_CERT_MANAGER* cm)
     return ret;
 }
 
-#ifndef CRL_STATIC_REVOKED_LIST
-/* returns head of copied list that was alloc'd */
-static RevokedCert *DupRevokedCertList(RevokedCert* in, void* heap)
+/* Copy a single revoked cert, deep copying any entry extensions.
+ * returns 0 on success and MEMORY_E on fail */
+static int DupRevokedCert(RevokedCert* out, const RevokedCert* in, void* heap)
 {
-    RevokedCert* head = NULL;
-    RevokedCert* current = in;
-    RevokedCert* prev = NULL;
-    while (current) {
-        RevokedCert* tmp = (RevokedCert*)XMALLOC(sizeof(RevokedCert), heap,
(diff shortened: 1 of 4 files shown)


gasbytes:
Jenkins retest this please

gasbytes:
Jenkins retest this please

src/crl.c
```diff
@@ -1416,6 +1411,37 @@ static CRL_Entry* DupCRL_Entry(const CRL_Entry* ent, void* heap)
...
+    {
+        int i;
+
+        /* certs is an in-struct array living after verifyMutex, so the bulk
+         * copy above aliased every extensions pointer with the original's.
+         * Cleaning the whole array first. */
+        for (i = 0; i < CRL_MAX_REVOKED_CERTS; i++) {
+            dupl->certs[i].extensions = NULL;
+            dupl->certs[i].extensionsSz = 0;
+        }
```
julek-wolfssl: Why not just use `XMEMSET`?

julek-wolfssl (changes requested):

src/crl.c
julek-wolfssl: I was wrong about the struct here. I thought it only contained these fields. Looking at the struct `RevokedCert* next` needs to be checked/cleared too.

julek-wolfssl (changes requested):

src/crl.c
gasbytes: adding the line to clear that field too, thanks for the feedback.

src/crl.c
```diff
@@ -1416,6 +1411,38 @@ static CRL_Entry* DupCRL_Entry(const CRL_Entry* ent, void* heap)
...
+        int i;
+
+        /* certs is an in-struct array living after verifyMutex, so the bulk
+         * copy above aliased every extensions pointer with the original's.
+         * Cleaning the whole array first. */
+        for (i = 0; i < CRL_MAX_REVOKED_CERTS; i++) {
+            dupl->certs[i].next = NULL;
+            dupl->certs[i].extensions = NULL;
+            dupl->certs[i].extensionsSz = 0;
+        }
```
julek-wolfssl: Let's move `certs` before verifyMutex instead - the bulk XMEMCPY then stays true to its plain-value-data contract and DupCRL_Entry copies the entries explicitly, rather than copy-then-repair.

src/crl.c
```diff
@@ -1416,6 +1411,38 @@ static CRL_Entry* DupCRL_Entry(const CRL_Entry* ent, void* heap)
...
+            if (dupl->certs[i].extensions == NULL) {
+                WOLFSSL_MSG("Failed to allocate revoked cert extensions");
+                CRL_Entry_free(dupl, heap);
+                return NULL;
+            }
+            XMEMCPY(dupl->certs[i].extensions, ent->certs[i].extensions,
+                    ent->certs[i].extensionsSz);
+            dupl->certs[i].extensionsSz = ent->certs[i].extensionsSz;
+        }
+    }
```
julek-wolfssl: Refactor this into `DupRevokedCertList` (where implementation of copying the field depends on `CRL_STATIC_REVOKED_LIST`.

julek-wolfssl (changes requested):

julek-wolfssl:
Also update the comment in `CRL_Entry` to assert that only static fields are allowed after `verifyMutex`.

src/crl.c
```diff
@@ -1331,65 +1326,77 @@ static WOLFSSL_X509_CRL* wolfSSL_X509_crl_new(WOLFSSL_CERT_MANAGER* cm)
...
+/* Copy a single revoked cert, deep copying any entry extensions.
+ * returns 0 on success and MEMORY_E on fail */
+static int DupRevokedCert(RevokedCert* out, const RevokedCert* in, void* heap)
+{
+    XMEMCPY(out->serialNumber, in->serialNumber, EXTERNAL_SERIAL_SIZE);
+    out->serialSz = in->serialSz;
+    XMEMCPY(out->revDate, in->revDate, MAX_DATE_SIZE);
+    out->revDateFormat = in->revDateFormat;
+    out->reasonCode = in->reasonCode;
+    out->next = NULL;
```
julek-wolfssl, line 1338: Why don't we want to copy `next`? Is it not used? If its not used it should be removed.

julek-wolfssl (changes requested):

src/crl.c
gasbytes: it's used, DupRecokvedCertList() links each copy in itself, and the NULL has to stay because tmp comes from a bare XMALLOC

so I think without it the tail node keeps garbage next that CRL_Entry_free() would walk into.

julek-wolfssl (dismissed):

tests/api/test_certman.c
```diff
@@ -2602,6 +2603,54 @@ int test_wolfSSL_CRL_static_revoked_list(void)
...
+    ExpectNotNull(dupl = wolfSSL_X509_CRL_dup(crl));
+
+    /* Every duplicated revoked cert must own its own extensions buffer. This
+     * fails if the entries are shallow-copied. */
+    if (crl != NULL && dupl != NULL && crl->crlList != NULL &&
+            dupl->crlList != NULL) {
+        int i;
+        for (i = 0; i < crl->crlList->totalCerts; i++) {
+            if (crl->crlList->certs[i].extensions != NULL) {
+                ExpectPtrNE(dupl->crlList->certs[i].extensions,
```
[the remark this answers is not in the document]
philljj: This seems like a valid nit.

If the crl extensions != NULL, then check dupl copy extensions are also != NULL and crl != dupl extensions.

And same logic for extensionsSz member.

philljj (changes requested):

tests/api/test_certman.c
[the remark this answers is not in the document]
gasbytes: duplicate

gasbytes:

philljj (approved):

PR #11042, opened 2026-08-04
Add Altera Agilex 5 SDM crypto offload port

aidangarske:
# Description  
  
  - Adds libfcs based Agilex 5 SDM crypto callback integration
  - Offloads RNG SHA256 AES ECC and HMAC operations
  - Keeps unsupported and short operations on safe software fallback paths
  - Preserves device resident ECC key isolation for signing and key agreement
  - Adds lifecycle rekey copy cleanup and hardware usage regression coverage
  - Supports Autotools CMake and user settings configurations
  - Validates the BitBake packaged wolfCrypt test on Agilex 5 hardware
  - Requires the matching meta-wolfSSL integration for the documented image flow

partner pr here: https://github.com/wolfSSL/meta-wolfssl/pull/176


aidangarske:
Jenkins retest this please

wolfcrypt/src/wc_port.c
```diff
@@ -817,6 +869,26 @@ int wolfCrypt_Cleanup(void)
...
+        }
+        ret = wc_AlteraFcsCryptoCb_UnRegisterDeviceEx(
+            WOLFSSL_ALTERA_FCS_DEVID);
+        if (ret != 0) {
+            int ret2 = wc_local_InitDownAbort(&wolfcrypt_init_state);
+            return (ret2 == 0) ? ret : ret2;
+        }
+#endif
+#ifdef WOLF_CRYPTO_CB
+        ret = wc_CryptoCb_Cleanup();
```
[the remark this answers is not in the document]
aidangarske: Fixed. wolfCrypt_Cleanup no longer tears the FCS device down ahead of the loop. It unregisters through wc_CryptoCb_Cleanup's reverse-order pass, which stops at the first BUSY_E and leaves every device before it registered.

wolfcrypt/src/wc_port.c
```diff
@@ -817,6 +869,26 @@ int wolfCrypt_Cleanup(void)
     else if (ret == WC_INIT_STATE_INITED)
         return 0;
     else {
+#if defined(WOLFSSL_ALTERA_FCS) && defined(WOLF_CRYPTO_CB)
+        if (wc_AlteraFcs_ResourceActive()) {
+            WOLFSSL_MSG("wolfCrypt_Cleanup() called with active Altera FCS objects");
+            ret = wc_local_InitDownAbort(&wolfcrypt_init_state);
+            return (ret == 0) ? BUSY_E : ret;
+        }
+        ret = wc_AlteraFcsCryptoCb_UnRegisterDeviceEx(
```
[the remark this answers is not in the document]
aidangarske: Fixed. Removed the FCS pre-teardown that ran before wc_CryptoCb_Cleanup. The device now unregisters inside that loop and its callback returns BUSY_E without closing the SDM session, so an aborted cleanup leaves it registered and no re-registration is needed.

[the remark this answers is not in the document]
aidangarske: Fixed. Same change as the other wc_port.c abort findings. FCS is no longer torn down before wc_CryptoCb_Cleanup, so a BUSY_E abort leaves g_devId intact and the session open.

wolfcrypt/src/cryptocb.c
```diff
@@ -539,11 +552,20 @@ void wc_CryptoCb_UnRegisterDevice(int devId)
...
+            return ret;
     }
 #endif
     wc_CryptoCb_ClearDev(dev);
+    return 0;
+}
+
+void wc_CryptoCb_UnRegisterDevice(int devId)
+{
+    (void)wc_CryptoCb_UnRegisterDeviceEx(devId);
```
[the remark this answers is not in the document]
aidangarske: Fixed. wc_CryptoCb_UnRegisterDeviceEx is now a public int-returning API so callers can observe BUSY_E. The void wc_CryptoCb_UnRegisterDevice wrapper stays for source compatibility.

wolfssl/wolfcrypt/settings.h
```diff
@@ -5800,6 +5800,12 @@ blinding by defining WC_BLINDING_NO_RNG_ACKNOWLEDGE_WEAKNESS."
 #endif
 
 /* Crypto Callback Rules */
+#if defined(WOLFSSL_ALTERA_FCS) && !defined(WOLF_CRYPTO_CB_CMD)
+    #define WOLF_CRYPTO_CB_CMD
+#endif
+#if defined(WOLFSSL_ALTERA_FCS) && !defined(WOLF_CRYPTO_CB_SETKEY)
```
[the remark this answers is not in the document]
aidangarske: Fixed. settings.h now also auto-defines WOLF_CRYPTO_CB_COPY and WOLF_CRYPTO_CB_FREE for WOLFSSL_ALTERA_FCS, matching what configure.ac already emits for build-system builds.

wolfcrypt/src/port/altera/altera_fcs_glue.c
```diff
@@ -0,0 +1,594 @@
...
+
+int wc_AlteraFcs_KeyIdNew(word32* keyId)
+{
+    int ret;
+
+    if (keyId == NULL) {
+        return BAD_FUNC_ARG;
+    }
+    ret = wc_AlteraFcs_OrphanKey(0);
+    if (ret != 0) {
```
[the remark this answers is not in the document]
aidangarske: Fixed. wc_AlteraFcs_KeyIdNew now ignores the best-effort sweep result and allocates the id regardless, since session close reclaims any orphan.

wolfcrypt/test/test.c
```diff
@@ -77855,6 +77876,1197 @@ WOLFSSL_TEST_SUBROUTINE wc_test_ret_t blob_test(void)
...
+            ERROR_OUT(WC_TEST_RET_ENC_EC(ret), exit_fcs_hash);
+        hwInit = 1;
+        ret = wc_InitSha256_ex(&swSha, HEAP_HINT, INVALID_DEVID);
+        if (ret != 0)
+            ERROR_OUT(WC_TEST_RET_ENC_EC(ret), exit_fcs_hash);
+        swInit = 1;
+
+        ret = wc_Sha256Update(&hwSha, msg, 2048);
+        if (ret == 0)
+            ret = wc_InitSha256_ex(&copySha, HEAP_HINT,
```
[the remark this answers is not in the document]
aidangarske: Fixed. copySha and copyInit are hoisted to function scope and freed at exit_fcs_hash, so the ERROR_OUTs in the copy blocks no longer leak the context.

wolfcrypt/test/test.c
```diff
@@ -77855,6 +77876,1197 @@ WOLFSSL_TEST_SUBROUTINE wc_test_ret_t blob_test(void)
...
+        copyInit = 1;
+
+        /* Copying over a live destination must release its existing state. */
+        ret = wc_Sha256Update(&copySha, msg, 64);
+        if (ret != 0)
+            ERROR_OUT(WC_TEST_RET_ENC_EC(ret), exit_fcs_hash);
+
+        ret = wc_Sha256Copy(&hwSha, &hwSha);
+        if (ret != WC_NO_ERR_TRACE(BAD_FUNC_ARG))
+            ERROR_OUT(WC_TEST_RET_ENC_NC, exit_fcs_hash);
```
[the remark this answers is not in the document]
aidangarske: Fixed. copyInit is now tracked in function scope and copySha is freed at exit_fcs_hash, so the assertion-failure ERROR_OUTs release it.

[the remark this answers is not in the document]
aidangarske: Fixed. copySha and copyInit moved to function scope and are freed at exit_fcs_hash alongside hwSha and swSha.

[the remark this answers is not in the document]
aidangarske: Fixed. copySha is freed at exit_fcs_hash after hoisting it to function scope, so the early ERROR_OUTs no longer leave the port resource count elevated.

wolfcrypt/test/test.c
```diff
@@ -81200,6 +82412,77 @@ static int myCryptoCbFind(int currentId, int algoType)
...
+                                     cryptoCbBusyTestCb, &testCtx);
+    if (ret != 0) {
+        return WC_TEST_RET_ENC_EC(ret);
+    }
+    wc_CryptoCb_UnRegisterDevice(CRYPTOCB_BUSY_TEST_DEVID);
+    if (!wc_CryptoCb_IsDeviceRegistered(CRYPTOCB_BUSY_TEST_DEVID)) {
+        return WC_TEST_RET_ENC_NC;
+    }
+    wc_CryptoCb_UnRegisterDevice(CRYPTOCB_BUSY_TEST_DEVID);
+    if (wc_CryptoCb_IsDeviceRegistered(CRYPTOCB_BUSY_TEST_DEVID)) {
```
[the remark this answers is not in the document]
aidangarske: Fixed. testCtx is now static, so the device never holds a pointer into a dead stack frame on any early exit.

wolfcrypt/test/test.c
```diff
@@ -81200,6 +82412,77 @@ static int myCryptoCbFind(int currentId, int algoType)
...
+
+    testCtx.busy = 1;
+    ret = wc_CryptoCb_RegisterDevice(CRYPTOCB_BUSY_TEST_DEVID,
+                                     cryptoCbBusyTestCb, &testCtx);
+    if (ret != 0) {
+        return WC_TEST_RET_ENC_EC(ret);
+    }
+    ret = wolfCrypt_Cleanup();
+    if (ret != WC_NO_ERR_TRACE(BUSY_E)) {
+        return WC_TEST_RET_ENC_NC;
```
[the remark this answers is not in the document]
aidangarske: Fixed. testCtx is static now, and the tail is restructured so wolfCrypt_Init runs on every exit after the cleanup cycle.

wolfcrypt/test/test.c
```diff
@@ -81200,6 +82412,77 @@ static int myCryptoCbFind(int currentId, int algoType)
...
+
+    testCtx.busy = 1;
+    ret = wc_CryptoCb_RegisterDevice(CRYPTOCB_BUSY_TEST_DEVID,
+                                     cryptoCbBusyTestCb, &testCtx);
+    if (ret != 0) {
+        return WC_TEST_RET_ENC_EC(ret);
+    }
+    wc_CryptoCb_UnRegisterDevice(CRYPTOCB_BUSY_TEST_DEVID);
+    if (!wc_CryptoCb_IsDeviceRegistered(CRYPTOCB_BUSY_TEST_DEVID)) {
+        return WC_TEST_RET_ENC_NC;
```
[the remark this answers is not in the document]
aidangarske: Fixed. testCtx has static storage, so a later wolfCrypt_Cleanup dispatching to cryptoCbBusyTestCb no longer reads a dead frame.

wolfcrypt/test/test.c
```diff
@@ -81200,6 +82412,77 @@ static int myCryptoCbFind(int currentId, int algoType)
...
+    }
+    wc_CryptoCb_UnRegisterDevice(CRYPTOCB_BUSY_TEST_DEVID);
+    if (wc_CryptoCb_IsDeviceRegistered(CRYPTOCB_BUSY_TEST_DEVID)) {
+        return WC_TEST_RET_ENC_NC;
```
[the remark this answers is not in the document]
aidangarske: Fixed. testCtx is file-scope static, so the UNREGISTER dispatch at process exit no longer touches a freed stack frame.

wolfcrypt/test/test.c
```diff
@@ -81200,6 +82412,77 @@ static int myCryptoCbFind(int currentId, int algoType)
...
+    }
+    if (!wc_CryptoCb_IsDeviceRegistered(CRYPTOCB_BUSY_TEST_DEVID)) {
+        return WC_TEST_RET_ENC_NC;
+    }
+    ret = wolfCrypt_Cleanup();
+    if (ret != 0) {
+        return WC_TEST_RET_ENC_EC(ret);
+    }
+    if (wc_CryptoCb_IsDeviceRegistered(CRYPTOCB_BUSY_TEST_DEVID)) {
```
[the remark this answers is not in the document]
aidangarske: Fixed. The post-cleanup paths now fall through to a single wolfCrypt_Init before returning, so the library is never left torn down for the rest of the suite.

wolfcrypt/test/test.c
```diff
@@ -81223,7 +82510,8 @@ WOLFSSL_TEST_SUBROUTINE wc_test_ret_t cryptocb_test(void)
 
     /* set devId to something other than INVALID_DEVID */
     devId = 1;
-    ret = wc_CryptoCb_RegisterDevice(devId, myCryptoDevCb, &myCtx);
+    if (ret == 0)
+        ret = wc_CryptoCb_RegisterDevice(devId, myCryptoDevCb, &myCtx);
     if (ret != 0)
         ret = WC_TEST_RET_ENC_EC(ret);
```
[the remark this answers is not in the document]
aidangarske: Fixed. The register call and its encoding are now inside the if (ret == 0) branch, so an already-encoded failure from cryptoCbBusyUnregisterTest is preserved instead of being re-encoded.

wolfcrypt/test/test.c
```diff
@@ -81200,6 +82412,77 @@ static int myCryptoCbFind(int currentId, int algoType)
...
+        return WC_TEST_RET_ENC_NC;
+    }
+
+    testCtx.busy = 1;
+    ret = wc_CryptoCb_RegisterDevice(CRYPTOCB_BUSY_TEST_DEVID,
+                                     cryptoCbBusyTestCb, &testCtx);
+    if (ret != 0) {
+        return WC_TEST_RET_ENC_EC(ret);
+    }
+    ret = wolfCrypt_Cleanup();
```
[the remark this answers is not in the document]
aidangarske: The full wolfCrypt_Cleanup/Init cycle is the behavior under test, so I kept it rather than dropping to wc_CryptoCb_Cleanup, which would not exercise the BUSY_E abort and INITED restore. The test now guarantees wolfCrypt_Init runs on every exit so the library is never left torn down.

wolfcrypt/test/test.c
```diff
@@ -81200,6 +82412,77 @@ static int myCryptoCbFind(int currentId, int algoType)
...
+    if (ret != WC_NO_ERR_TRACE(BUSY_E)) {
+        return WC_TEST_RET_ENC_NC;
+    }
+    if (!wc_CryptoCb_IsDeviceRegistered(CRYPTOCB_BUSY_TEST_DEVID)) {
+        return WC_TEST_RET_ENC_NC;
+    }
+    ret = wolfCrypt_Cleanup();
```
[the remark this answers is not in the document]
aidangarske: Kept the full-library cycle since that is what this test validates. The re-init is now guaranteed on every path. WC_SWDEV_ID is only registered under the WOLFSSL_SWDEV harness, which this test path does not build with.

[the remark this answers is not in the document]
aidangarske: Intentional. Exercising wolfCrypt_Cleanup's BUSY_E abort and restore requires the full cycle. The fix is that wolfCrypt_Init now always runs before return, not switching to wc_CryptoCb_Cleanup.

aidangarske:
Jenkins retest this please "logs lost"

philljj:
Retest this please.

(no logs)

wolfcrypt/src/cryptocb.c
```diff
@@ -464,14 +464,22 @@ void wc_CryptoCb_Init(void)
     }
 }
 
-void wc_CryptoCb_Cleanup(void)
+int wc_CryptoCb_Cleanup(void)
```
night1rider: The cleanup loop now quits as soon as one device says it is busy, so devices it had not reached yet stay registered. This is shared code, so the new behavior applies to every hardware port in wolfSSL and not just this one.

aidangarske:
Jenkins retest this please

wolfcrypt/src/port/altera/altera_fcs_aes.c
```diff
@@ -0,0 +1,774 @@
...
+#endif
+#ifndef WOLF_CRYPTO_CB_SETKEY
+    #error "WOLFSSL_ALTERA_FCS_AES requires WOLF_CRYPTO_CB_SETKEY for re-keying"
+#endif
+
+/* The driver refuses any non-GCM length that is not a multiple of this. */
+#define WC_ALTERA_FCS_AES_ALIGN 32
+
+#define FCS_KEY_OBJ_MAGIC     0x43736B4FU
+#define FCS_KEY_DATA_MAGIC    0x43736B64U
```
night1rider: Comment on what this "magic" represents would be helpful.

wolfcrypt/src/port/altera/altera_fcs_aes.c
```diff
@@ -0,0 +1,774 @@
...
+    /* The declared size covers the object but not the trailing MAC field. */
+    objSz = FCS_KEY_DATA_OFFSET + padded;
+    wc_AlteraFcs_Put32(out + 4, ((word32)FCS_KEY_OBJ_VER << 16) |
+                                (objSz & 0xFFFF));
+
+    *outSz = objSz + FCS_KEY_MAC_SZ;
+    return 0;
+}
+
+static int wc_AlteraFcs_KeyRemove(word32 keyId)
```
night1rider: Kinda of a nit: `wc_AlteraFcs_KeyRemove()` (`aes.c:185`) and `wc_AlteraFcs_EccKeyRemove()` (`ecc.c:212`) have identical bodies that only hand their argument to `wc_AlteraFcs_RemoveServiceKey()` and return its result. The HMAC path does the same job by calling `wc_AlteraFcs_RemoveServiceKey()` directly (`hmac.c:325`), so two of the three algorithms wrap it and one does not. Could you either have them call the function directly or add a wrapper for the hmac. I think I would prefer if they just call the function directly, unless you have a specific reason behind this.

wolfcrypt/src/port/altera/altera_fcs_aes.c
```diff
@@ -0,0 +1,774 @@
...
+    }
+
+    /* The declared size covers the object but not the trailing MAC field. */
+    objSz = FCS_KEY_DATA_OFFSET + padded;
+    wc_AlteraFcs_Put32(out + 4, ((word32)FCS_KEY_OBJ_VER << 16) |
+                                (objSz & 0xFFFF));
+
+    *outSz = objSz + FCS_KEY_MAC_SZ;
+    return 0;
+}
```
night1rider: `wc_AlteraFcs_Put32()` is byte for byte identical in all three files, and the three key-object builders write the same six fields at the same six offsets with the same size and padding maths, differing only in which size codes are legal, the type value, the usage value, and whether key bytes are copied. Six constants that describe the header are each defined three times, so the SDM key object layout is recorded in three independent places.

### Where the duplicates are

| What | `altera_fcs_aes.c` | `altera_fcs_ecc.c` | `altera_fcs_hmac.c` |
|---|---|---|---|
| `FCS_KEY_OBJ_MAGIC` | 71 | 64 | 109 |
| `FCS_KEY_DATA_MAGIC` | 72 | 65 | 110 |
| `FCS_KEY_OBJ_VER` | 73 | 66 | 111 |
| `FCS_KEY_MAC_SZ` | 78 | 69 | 113 |
| `FCS_KEY_DATA_OFFSET` | 79 | 70 | 114 |
| `FCS_KEY_STATUS_SZ` | 80 | 72 | 116 |
| `FCS_KEY_ALIGN` | -- | 71 | 115 |
| `FCS_KEY_USAGE_SIGN_VERIFY` | -- | 76 | 120 |
| `wc_AlteraFcs_Put32()` | 130-136 | 127-133 | 129-135 |
| key-object builder | 141-183 | 174-209 | 176-220 |
| context lookup | 105 (`AesCtx`) | 109 (`EccCtx`) | -- |

`wc_AlteraFcs_EccKeyObject()` is already the most general of the three, taking `keyType` and `usage` as parameters, so it is close to serving all three if it also copied key bytes when given them.

Maybe have a generic _utils file or use the glue file to pull out common functionality?

wolfcrypt/src/port/altera/altera_fcs_aes.c
```diff
@@ -0,0 +1,774 @@
...
+    }
+    if (sz < WC_ALTERA_FCS_AES_ALIGN || sz > WC_ALTERA_FCS_MAX_XFER) {
+        return 0;
+    }
+    if ((sz % WC_ALTERA_FCS_AES_ALIGN) != 0) {
+        return 0;
+    }
+    if (sz < WOLFSSL_ALTERA_FCS_AES_MIN) {
+        return 0;
+    }
```
night1rider: I do not think there should be a minimum at all, if the hardware can handle the buffer size, it should do it. The README says the value of the SDM is key isolation rather than throughput, so someone who turns this on should not end up thinking they are using hardware when the port has quietly sent the work to software. Does this have a max buffer limit?

wolfcrypt/src/port/altera/altera_fcs_aes.c
```diff
@@ -0,0 +1,774 @@
...
+/* The imported key is cached against the material it was built from, because
+ * wc_AesSetKey may re-key a context without clearing devCtx and a stale id
+ * would silently encrypt under the previous key. */
+typedef struct {
+    void*  heap;
+    word32 tag;                   /* WC_ALTERA_FCS_AES_TAG, else stale memory */
+    word32 keyId;                 /* SDM service key slot id */
+    int    keyLen;
+    int    origin;                /* WC_ALTERA_FCS_AES_IMPORTED | _RESIDENT */
+    byte   key[AES_256_KEY_SIZE]; /* imported origin only; zero for resident */
```
night1rider: `AlteraAesKey` holds the plaintext key in `keyCtx->key` (line 101), copied in at line 282 and kept for the life of the context, and the only thing that reads it is the `ConstantCompare()` at line 217 which checks the tracked `keyId` still matches the current key material. Since the port already requires `WOLF_CRYPTO_CB_SETKEY` (line 65) and wolfSSL calls `wc_CryptoCb_SetKey()` on every `wc_AesSetKey()` (`aes.c:5100` and `:5711`), a real re-key has already freed the context through `wc_AlteraFcs_AesSetKey()` before that compare can run. Is the copy still needed, and if some path can change `aes->devKey` without the callback firing, would a digest of the key material serve the same purpose without keeping key bytes?

AES is the only one of the three that does this. HMAC imports caller-supplied plaintext keys the same way but the bytes only reach a local key object that is wiped with `ForceZero()` at `altera_fcs_hmac.c:298`, leaving the caller holding just a `keyId`, and `AlteraEccKey` has no key field at all. Resident AES keys are also unaffected, since the field is left zero for them, and both AES free paths `ForceZero` the context at lines 236 and 515.

wolfcrypt/src/port/altera/altera_fcs_aes.c
```diff
@@ -0,0 +1,774 @@
...
+    wc_AlteraFcs_AesKeyFree(aes);
+    return CRYPTOCB_UNAVAILABLE;
+}
+
+/* Create a device resident AES key. The key is generated inside the SDM from a
+ * zero data key object, so it never exists in HPS memory, and CBC/CTR are then
+ * offloaded by handle with no software fallback. Requires
+ * wc_AesInit(aes, heap, WOLFSSL_ALTERA_FCS_DEVID) first and an empty devCtx.
+ * Do not call wc_AesSetKey on the resulting context. */
+int wc_AlteraFcsAes_MakeKey(Aes* aes, int keyBits)
```
night1rider: I am not personally a fan of adding device specific functions to the public api.

Would a second devId, say `WOLFSSL_ALTERA_FCS_AES_KEY_DEVID`, be worth adding so a device key can be created through the standard `wc_AesSetKey()` rather than the Altera-only `wc_AlteraFcsAes_MakeKey()`? The devId is what asks for a key to be generated, and the port's existing SETKEY handler would then require `userKey` to be NULL as confirmation, erroring out if it is not, so that a caller who passed real key material can never have it silently discarded in favour of a generated one. A caller would init with that devId and call `wc_AesSetKey(&aes, NULL, 32, iv, dir)`, where the length says 128 or 256, and the handler would create the key in the device, set `aes->devId` to the normal `WOLFSSL_ALTERA_FCS_DEVID` and return 0. Any context left on the key-generation devId outside that one path would fail rather than quietly do something else.

wolfcrypt/src/port/altera/altera_fcs_aes.c
night1rider: I do not think there should be a minimum at all, if the hardware can handle the buffer size, it should do it. The README says the value of the SDM is key isolation rather than throughput, so someone who turns this on should not end up thinking they are using hardware when the port has quietly sent the work to software. Does this have a max buffer limit?

wolfcrypt/src/port/altera/altera_fcs_ecc.c
```diff
@@ -0,0 +1,729 @@
...
+    }
+
+    if (curveId == ECC_CURVE_DEF) {
+        if (sizeHint == 32) {
+            curveId = ECC_SECP256R1;
+        }
+        else if (sizeHint == 48) {
+            curveId = ECC_SECP384R1;
+        }
+    }
```
night1rider: Is `sizeHint` still needed here, or is it left over from something? `wc_AlteraFcs_EccCreate()` is `static` with two callers, `wc_AlteraFcsEcc_MakeSigningKey()` at line 393 and `wc_AlteraFcsEcc_MakeExchangeKey()` at line 404, and both pass 0.

wolfcrypt/src/port/altera/altera_fcs_ecc.c
```diff
@@ -0,0 +1,729 @@
...
+         * possibly short response is exposed to the caller. ECDSA verify is
+         * deliberately declined by this callback and completes in software. */
+        ret = wc_ecc_verify_hash(info->pk.eccsign.out,
+                                 *info->pk.eccsign.outlen,
+                                 info->pk.eccsign.in,
+                                 info->pk.eccsign.inlen, &verified, key);
+        if (ret == 0 && verified != 1) {
+            ret = WC_HW_E;
+        }
+    }
```
night1rider: Every device ECDSA sign also runs a full software `wc_ecc_verify_hash()` before the signature is returned, justified by the comment that the Linux driver "does not return a trustworthy response length". I tested the sign path with libfcs only and the length came back as 64, which is correct for P-256 `r||s`, so on this board and firmware it looked reliable. Is the length genuinely wrong somewhere, or was it inferred from the separate `dst_len` overrun, since if it can be trusted then every sign is paying for a verify it may not need?

wolfcrypt/src/port/altera/altera_fcs_glue.c
```diff
@@ -0,0 +1,591 @@
...
+        ret = libfcs_init((FCS_OSAL_CHAR*)"error");
+        if (ret != 0) {
+            WOLFSSL_MSG("libfcs_init failed");
+            ret = wc_AlteraFcs_MapError(ret);
+        }
+        else {
+            g_libReady = 1;
+            g_processPid = getpid();
+        }
+    }
```
night1rider, line 387: When `libfcs_init()` fails the port leaves `g_libReady` at 0, so the next call to `wc_AlteraFcs_Init()` calls it again, but libfcs sets its own `ctx.state` to `in_progress` before doing any work and returns early on failure without resetting it, so every later call takes the `ctx.state != un_initialized` path and returns 0. The port reads that as success, sets `g_libReady = 1` and reports the hardware as ready when the library is unusable. Remembering the first failure and declining instead of retrying would avoid it, since the retry can never succeed.

The symptom on the board is a warning and two errors that contradict each other, "FCS library already initialized or initialization in progress" followed by "FCS library not initialized", visible in `repro/logs/03b-teardown-damage.log`. The root cause is in libfcs rather than wolfSSL, but the port is the thing that retries and then trusts the answer.

wolfcrypt/src/port/altera/altera_fcs_glue.c
```diff
@@ -0,0 +1,591 @@
...
+    *keyId = g_nextKeyId;
+    /* Key IDs are carried by a few legacy signed atomic trackers. Keep the
+     * allocator below INT_MAX so those trackers never observe a negative ID;
+     * zero remains reserved for the orphan-key sweep. */
+    if (g_nextKeyId >= 0x7FFFFFFFU) {
+        g_nextKeyId = WOLFSSL_ALTERA_FCS_KEY_ID_BASE;
+    }
+    else {
+        g_nextKeyId++;
+    }
```
night1rider: `wc_AlteraFcs_KeyIdNew()` hands out `g_nextKeyId` and increments it, and on reaching `0x7FFFFFFF` resets to `WOLFSSL_ALTERA_FCS_KEY_ID_BASE` without checking whether that id is still held by a live key. It would take roughly 683 million keys in a single session to get there, so this is a note rather than a real risk, but a reused id would collide with an existing device key rather than fail cleanly.

wolfcrypt/src/port/altera/altera_fcs_glue.c
```diff
@@ -0,0 +1,591 @@
...
+
+    if (sessionId == NULL) {
+        return BAD_FUNC_ARG;
+    }
+
+    /* pthread_atfork protects threaded builds from inheriting a locked mutex.
+     * The PID check also covers SINGLE_THREADED builds. */
+    if (g_ownerPid != 0 && g_ownerPid != getpid()) {
+        wc_AlteraFcs_ResetForkChild();
+    }
```
night1rider: `wc_AlteraFcs_SessionAcquire()` compares `g_ownerPid` against `getpid()` and may call `wc_AlteraFcs_ResetForkChild()`, which clears the session, refcount, orphan list and lock descriptor, all before `wc_LockMutex(&g_lock)` is reached. The comment says `pthread_atfork` covers the threaded case and that is fair, but the read and the reset are still unprotected. The same pattern appears again at line 564 in `wc_AlteraFcs_Cleanup()`, where it does run under the mutex.

wolfcrypt/src/port/altera/altera_fcs_hash.c
```diff
@@ -0,0 +1,831 @@
...
+            }
+            wc_AlteraFcs_KeepFree(keep);
+            *devCtxPtr = NULL;
+            return ret;
+        }
+#endif
+
+        /* Oversized messages exceed a single SDM transaction, so they are
+         * finished in software rather than rejected by the driver. */
+        if (hwOk && msgLen >= WOLFSSL_ALTERA_FCS_HASH_MIN &&
```
night1rider: `WOLFSSL_ALTERA_FCS_HASH_MIN` also defaults to 4096, so any message shorter than that is never offered to the device and is finished by `wc_AlteraFcs_SoftDigest()` instead. So unless the chip has a minimum input, I dont think we should be silently defaulting to software here.

wolfcrypt/src/port/altera/altera_fcs_hash.c
```diff
@@ -0,0 +1,831 @@
...
+            msgLen <= WC_ALTERA_FCS_MAX_XFER) {
```
night1rider: The port only ever calls the one shot entry points and never `fcs_get_digest_streaming()`, `fcs_aes_crypt_streaming()` or `fcs_mac_verify_streaming()`, which is what forces both the 4 MiB single transaction ceiling and the buffering that ends in a software digest. Those streaming calls take file paths rather than buffers, but `memfd_create()` creates an anonymous file that lives entirely in memory and is referenced by path as `/proc/self/fd/N`, so nothing has to be written to storage to use them. Between dropping the minimum and using the streaming path, the device could serve every size rather than only messages that happen to land between 4096 bytes and 4 MiB.

`memfd_create()` is Linux only rather than standard C, but so is the rest of the port, which already uses `flock()` from `sys/file.h`, `getpid()` and `pthread_atfork()`, and libfcs itself is a Linux userspace library talking to a Linux kernel driver.

wolfcrypt/src/port/altera/altera_fcs_cryptocb.c
```diff
@@ -0,0 +1,420 @@
...
+    if (info == NULL) {
+        return BAD_FUNC_ARG;
+    }
+
+#ifdef WOLF_CRYPTO_CB_CMD
+    if (info->algo_type == WC_ALGO_TYPE_NONE) {
+        if (info->cmd.type == WC_CRYPTOCB_CMD_TYPE_UNREGISTER) {
+            ret = wc_AlteraFcsCryptoCb_UnRegisterDeviceEx(devId);
+            return (ret == 0) ? 0 : BUSY_E;
+        }
```
night1rider, line 208: `return (ret == 0) ? 0 : BUSY_E;` turns every failure from `wc_AlteraFcsCryptoCb_UnRegisterDeviceEx()` into `BUSY_E`, whatever actually went wrong. That matters more than a usual error squash because `wc_CryptoCb_Cleanup()` treats `BUSY_E` as a reason to stop the teardown loop, so an unrelated failure would halt cleanup for every other registered device. Passing the real error through, and reserving `BUSY_E` for a device that is genuinely busy, would keep that signal meaningful.

wolfcrypt/src/port/altera/altera_fcs_cryptocb.c
```diff
@@ -0,0 +1,420 @@
...
+        return CRYPTOCB_UNAVAILABLE;
+    }
+#endif
```
night1rider, line 211: `wc_CryptoCb_RegisterDevice()` already calls the callback with `WC_CRYPTOCB_CMD_TYPE_REGISTER` as an init hook, and already unwinds on failure by calling `wc_CryptoCb_ClearDev(dev)` and returning the error, so a port can do its device setup there and have a failed setup undo the registration for free. The port handles `WC_CRYPTOCB_CMD_TYPE_UNREGISTER` but returns `CRYPTOCB_UNAVAILABLE` for REGISTER at line 139, which wolfSSL reads as "not implemented", and instead does the registration from `wolfCrypt_Init()` with the hand written `WOLFCRYPT_INIT_FCS_ROLLBACK` macro. Can the device setup happen in the REGISTER handler, where the framework already provides the hook and the unwinding, which would leave no Altera specific code in `wc_port.c` at all.

night1rider (changes requested):
Apologies for this taking longer than I wanted, got caught up with other work and had issues getting my Agilex 5 in a state that I could do some testing. Please reach out if you have any questions about these review comments.

wolfcrypt/src/port/altera/altera_fcs_aes.c
aidangarske: They are ASCII "CskO" and "Cskd" from the fcs_prepare key object layout. I added a note about it

wolfcrypt/src/port/altera/altera_fcs_aes.c
aidangarske: I dropped both wrappers, callers use wc_AlteraFcs_RemoveServiceKey and wc_AlteraFcs_DiscardServiceKey directly.

wolfcrypt/src/port/altera/altera_fcs_aes.c
aidangarske: Moved Put32/Get32, the key object builder and the header constants into altera_fcs_glue.c and altera_fcs.h. fcs_prepare uses keyBits/128 as the size code for every key type so one builder serves AES, ECC and HMAC.

wolfcrypt/src/port/altera/altera_fcs_aes.c
aidangarske: Thanks, I removed the copy and the compare, the struct holds no key bytes now. Every re-key on this target runs the SETKEY hook first (AesSetKeyLocal_body in aes.c), so the context is freed before that compare could be reached.

wolfcrypt/src/port/altera/altera_fcs_aes.c
aidangarske: Makes sense added a WOLFSSL_ALTERA_FCS_AES_KEY_DEVID (0x4144) takes a NULL key in wc_AesSetKey, creates the resident key and moves the context to the main devId, and wc_AlteraFcsAes_MakeKey is gone.

wolfcrypt/src/port/altera/altera_fcs_aes.c
aidangarske: Removed the minimum, the only rule left is the driver 32 byte multiple. Past CRYPTO_MAX_SZ the request now goes through fcs_aes_crypt_streaming, so that is no longer a ceiling either.

wolfcrypt/src/port/altera/altera_fcs_ecc.c
aidangarske: I think its leftover, removed along with the ECC_CURVE_DEF mapping.

wolfcrypt/src/port/altera/altera_fcs_ecc.c
aidangarske: I Measured it on the board with canaries the kernel writes 2*keySz bytes at dst_len (64 for P-256, 96 for P-384) but the first word is the correct length. So the software verify is gone, sign checks that word, and the sink stays with the comment corrected.

wolfcrypt/src/port/altera/altera_fcs_glue.c
aidangarske: Fixed, the first libfcs_init failure is latched and later wc_AlteraFcs_Init calls return it without touching libfcs again.

wolfcrypt/src/port/altera/altera_fcs_glue.c
aidangarske: Changed to refuse with BAD_STATE_E at the top instead of wrapping, since nothing tracks which ids are still live.

wolfcrypt/src/port/altera/altera_fcs_glue.c
aidangarske: Moved the pid check and reset under g_lock inside wc_AlteraFcs_Init and dropped the duplicate in SessionAcquire, matching the one in Cleanup.

wolfcrypt/src/port/altera/altera_fcs_hash.c
aidangarske: Okay I removed The only hardware rule that showed up on the board is that the SDM rejects lengths that are not a multiple of 8 bytes (one shot errors, streaming returns a wrong digest), so those finish on the software shadow and everything else goes to the device.

wolfcrypt/src/port/altera/altera_fcs_hash.c
aidangarske: Okay done, The message accumulates in a memfd and final calls fcs_get_digest_streaming on /proc/self/fd/N, AES and HMAC verify do the same past 4 MiB, and all three were checked across the chunk boundary on the board.

wolfcrypt/src/port/altera/altera_fcs_cryptocb.c
aidangarske: Thanks, Fixed

wolfcrypt/src/port/altera/altera_fcs_cryptocb.c
aidangarske: Moved wc_AlteraFcs_Init and the state setup into the REGISTER handler so a failed setup unwinds inside wc_CryptoCb_RegisterDevice, and dropped the rollback macro. wolfCrypt_Init keeps one plain register call like the Versal ASU one above it.

aidangarske: Thanks for all the good feedback!

PR #11046, opened 2026-08-04
Zero EncryptedInfo in ProcessChainBufferCRL and reset info->set on parse

yosuke-wolfssl:
### Problem
`ProcessChainBufferCRL()` declared a stack `EncryptedInfo` and passed it to `PemToDer()` without zeroing it — unlike every other caller in the tree. A PEM CRL carries no `Proc-Type:` header, so `wc_EncryptedInfoParse()` returns without ever writing `info->set`, leaving `PemToDer()` to read stack residue. When that residue is non-zero:

- **`encrypted_key`** is set spuriously.
- **The CRL header** does not match the non-encrypted early return, so control falls into the encrypted-key block.
- **`info->passwd_cb`** is then invoked through the uninitialized pointer, with an uninitialized argument, on a path that should never touch password callbacks.

Reachable from the public `wolfSSL_CTX_load_verify_buffer()` with `WOLFSSL_FILETYPE_PEM`. `--enable-wpas` alone enables all three required macros (`WOLFSSL_WPAS`, `HAVE_CRL`, `WOLFSSL_ENCRYPTED_KEYS`) — the configuration used to link wpa_supplicant/hostapd.

### Fix (`src/ssl_load.c`)

```diff
     WOLFSSL_MSG("Trying a CRL");
+    XMEMSET(&info, 0, sizeof(info));
     ret = PemToDer(buff, sz, CRL_TYPE, &der, NULL, &info, NULL);
```

Defense in depth in `wolfcrypt/src/asn.c`: `wc_EncryptedInfoParse()` now clears `info->set` on entry, so no future caller can be bitten by a stale flag. `set` has exactly two references tree-wide, so nothing relied on it persisting across a parse.

Closes f-7352.

### Tests (`tests/api.c`)
- **`test_wolfSSL_CTX_load_verify_buffer_pem_crl`** — first coverage of the CRL-in-chain fall-through. Loads `ca-cert.pem` + `crl/crl.pem` as one PEM buffer and asserts the CRL actually reached the cert manager.
- **A case in `test_wc_PemToDer`** passing a deliberately dirtied `info.set = 1` with `CRL_TYPE`. This is the deterministic regression guard: it needs no sanitizer and runs in ordinary builds. Appended last and re-zeroed so it cannot contaminate the existing assertions.

### Verification

| Build | Result |
|---|---|
| `--enable-all` | 1770 passed, 0 failed |
| `--enable-wpas --enable-opensslextra` | 1106 passed, 0 failed |
| `make check`, both configs | 0 failures |

Negative controls: reverting the `asn.c` reset makes the new `wc_PemToDer` case fail with `NO_PASSWORD` (`-176`) and nothing else. Forcing worst-case residue with both fixes reverted aborts the suite with SIGBUS on the indirect call, confirming the reported impact. ASan + UBSan clean; no warnings under `-Werror`.

### DIFF
--- a/src/ssl_load.c
+++ b/src/ssl_load.c
@@ -2564,19 +2564,29 @@ int ProcessBuffer(WOLFSSL_CTX* ctx, const unsigned char* buff, long sz,
 static int ProcessChainBufferCRL(WOLFSSL_CTX* ctx, const unsigned char* buff,
     long sz, long* consumed)
 {
-    int           ret;
-    DerBuffer*    der = NULL;
-    EncryptedInfo info;
+    int        ret = 0;
+    DerBuffer* der = NULL;
+    WC_DECLARE_VAR(info, EncryptedInfo, 1, 0);
 
     WOLFSSL_MSG("Trying a CRL");
-    ret = PemToDer(buff, sz, CRL_TYPE, &der, NULL, &info, NULL);
+
+    /* A PEM CRL has no Proc-Type header, so the encryption information must
+     * start zeroed or PemToDer() acts on a stale encrypted-key flag. */
+    WC_CALLOC_VAR_EX(info, EncryptedInfo, 1, ctx->heap,
+        DYNAMIC_TYPE_ENCRYPTEDINFO, ret = MEMORY_E);
+
+    if (ret == 0) {
+        ret = PemToDer(buff, sz, CRL_TYPE, &der, NULL, info, NULL);
+    }
     if (ret == 0) {
         WOLFSSL_MSG("   Processed a CRL");
         wolfSSL_CertManagerLoadCRLBuffer(ctx->cm, der->buffer, der->length,
             WOLFSSL_FILETYPE_ASN1);
         FreeDer(&der);
-        *consumed = info.consumed;
+        *consumed = info->consumed;
     }
+
+    WC_FREE_VAR_EX(info, ctx->heap, DYNAMIC_TYPE_ENCRYPTEDINFO);
 
     return ret;
 }
--- a/tests/api.c
+++ b/tests/api.c
@@ -3929,6 +3929,59 @@ static int test_wolfSSL_CTX_load_verify_chain_buffer_format(void)
     return EXPECT_RESULT();
 }
 
+/* PEM buffer holding a CA certificate followed by a CRL. The CRL block fails
+ * ProcessBuffer and falls through to ProcessChainBufferCRL. */
+static int test_wolfSSL_CTX_load_verify_buffer_pem_crl(void)
+{
+    EXPECT_DECLS;
+#if defined(WOLFSSL_WPAS) && defined(HAVE_CRL) && defined(WOLFSSL_PEM_TO_DER) \
+    && !defined(NO_FILESYSTEM) && !defined(NO_CERTS) && !defined(NO_RSA) && \
+    !defined(NO_TLS) && !defined(NO_WOLFSSL_CLIENT)
+    WOLFSSL_CTX*          ctx = NULL;
+    WOLFSSL_CERT_MANAGER* cm = NULL;
+    byte*  caBuf = NULL;
+    byte*  crlBuf = NULL;
+    byte*  pemBuf = NULL;
+    byte*  certBuf = NULL;
+    size_t caSz = 0;
+    size_t crlSz = 0;
+    size_t certSz = 0;
+
+    ExpectIntEQ(load_file("./certs/ca-cert.pem", &caBuf, &caSz), 0);
+    ExpectIntEQ(load_file("./certs/crl/crl.pem", &crlBuf, &crlSz), 0);
+    ExpectIntEQ(load_file("./certs/server-cert.der", &certBuf, &certSz), 0);
+    if ((caBuf != NULL) && (crlBuf != NULL)) {
+        ExpectNotNull(pemBuf = (byte*)XMALLOC(caSz + crlSz, NULL,
+            DYNAMIC_TYPE_TMP_BUFFER));
+    }
+    if (pemBuf != NULL) {
+        XMEMCPY(pemBuf, caBuf, caSz);
+        XMEMCPY(pemBuf + caSz, crlBuf, crlSz);
+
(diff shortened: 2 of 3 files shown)


tests/api.c
```diff
@@ -3918,6 +3918,58 @@ static int test_wolfSSL_CTX_load_verify_chain_buffer_format(void)
...
+    byte*  crlBuf = NULL;
+    byte*  pemBuf = NULL;
+    byte*  certBuf = NULL;
+    size_t caSz = 0;
+    size_t crlSz = 0;
+    size_t certSz = 0;
+
+    ExpectIntEQ(load_file("./certs/ca-cert.pem", &caBuf, &caSz), 0);
+    ExpectIntEQ(load_file("./certs/crl/crl.pem", &crlBuf, &crlSz), 0);
+    ExpectIntEQ(load_file("./certs/client-cert.der", &certBuf, &certSz), 0);
```
dgarske: Should this be `./certs/server-cert.der`?
New CRL integration test asserts nothing - wrong test certificate makes it pass with zero CRLs loaded

tests/api.c
```diff
@@ -3918,6 +3918,58 @@ static int test_wolfSSL_CTX_load_verify_chain_buffer_format(void)
...
+        ExpectIntEQ(wolfSSL_CTX_load_verify_buffer(ctx, pemBuf,
+            (long)(caSz + crlSz), WOLFSSL_FILETYPE_PEM), WOLFSSL_SUCCESS);
+
+        /* CRL must reach the manager, not be mistaken for an encrypted PEM. */
+        ExpectNotNull(cm = wolfSSL_CTX_GetCertManager(ctx));
+        ExpectIntEQ(wolfSSL_CertManagerEnableCRL(cm, WOLFSSL_CRL_CHECK),
+            WOLFSSL_SUCCESS);
+        /* Cast is safe - the test certificate is a fixed, small file. */
+        ExpectIntNE(wolfSSL_CertManagerCheckCRL(cm, certBuf, (int)certSz),
+            WC_NO_ERR_TRACE(CRL_MISSING));
```
dgarske: Tighten `ExpectIntNE(..., CRL_MISSING)` to `ExpectIntEQ(..., WOLFSSL_SUCCESS)`.

src/ssl_load.c
```diff
@@ -2569,6 +2569,7 @@ static int ProcessChainBufferCRL(WOLFSSL_CTX* ctx, const unsigned char* buff,
     EncryptedInfo info;
 
     WOLFSSL_MSG("Trying a CRL");
+    XMEMSET(&info, 0, sizeof(info));
```
dgarske: Since we are here... the "info" is 152 bytes. Let's consider switching it to the `WC_DECLARE_VAR`/`WC_ALLOC_VAR_EX`/`WC_FREE_VAR_EX` pattern for small stack.

tests/api.c
```diff
@@ -3918,6 +3918,58 @@ static int test_wolfSSL_CTX_load_verify_chain_buffer_format(void)
     return EXPECT_RESULT();
 }
 
+/* PEM buffer holding a CA certificate followed by a CRL. The CRL block fails
+ * ProcessBuffer and falls through to ProcessChainBufferCRL. */
+static int test_wolfSSL_CTX_load_verify_buffer_pem_crl(void)
+{
+    EXPECT_DECLS;
+#if defined(WOLFSSL_WPAS) && defined(HAVE_CRL) && !defined(NO_FILESYSTEM) && \
```
dgarske: Add `defined(WOLFSSL_PEM_TO_DER)` to the guard to match `test_wc_PemToDer` at tests/api.c:11914.

dgarske (changes requested):

tests/api.c
yosuke-wolfssl: You are correct, I fixed it

tests/api.c
yosuke-wolfssl: Fixed

src/ssl_load.c
yosuke-wolfssl: Done. Used `WC_CALLOC_VAR_EX` rather than `WC_ALLOC_VAR_EX` — it folds the allocation and the zeroing into one call, so the fix this PR is about survives the refactor on both branches:

- **`WOLFSSL_SMALL_STACK`** — `XMALLOC` + `XMEMSET`
- **otherwise** — `EncryptedInfo info[1]` on the stack + `XMEMSET`, i.e. byte-for-byte the original fix

With plain `WC_ALLOC_VAR_EX` the non-small-stack expansion is `WC_DO_NOTHING`, so the `XMEMSET` would have to be written by hand alongside it — reintroducing exactly the omission that caused this bug.

```c
    int        ret = 0;
    DerBuffer* der = NULL;
    WC_DECLARE_VAR(info, EncryptedInfo, 1, 0);

    WC_CALLOC_VAR_EX(info, EncryptedInfo, 1, ctx->heap,
        DYNAMIC_TYPE_ENCRYPTEDINFO, ret = MEMORY_E);
```

`ONFAIL` is discarded on the stack path, so `ret` correctly stays `0` where no allocation can fail. Cleanup is a single unconditional `WC_FREE_VAR_EX` at the bottom — a no-op on the stack path, and `XFREE(NULL, ...)` is safe on the allocation-failure path, so no `goto` was needed.

Verified in both configurations:

| Config | Result |
|---|---|
| `--enable-wpas --enable-opensslextra --enable-crl` | 1109 passed, 0 failed |
| same + `--enable-smallstack` | 1109 passed, 0 failed |
| same + ASan/UBSan, small stack | 1109 passed, 0 failed, 0 reports |
| `make check` | 13 pass, 5 skip, 0 fail |

Negative control on the heap branch: forcing the allocation to fail propagates `MEMORY_E`, leaves the CRL unloaded, and draws no sanitizer report.

tests/api.c
yosuke-wolfssl: Added as suggested

yosuke-wolfssl:
Hello @dgarske ,
Thank you for reviewing. I fixed issues.
Could you please check this again ?

dgarske (approved):

PR #11048, opened 2026-08-04
Fix unsynchronized globalRNG use in BN_rand, AddSession, ECDH and X25519

yosuke-wolfssl:
### Problem

`globalRNG` (`src/ssl.c`) is a single process-wide `WC_RNG` with no internal
per-instance locking; its designated lock is `globalRNGMutex`. The `RAND_*`
family in `src/ssl_crypto.c` takes that mutex at all eleven of its call sites.
Four OpenSSL-compat entry points using `globalRNG` as their primary RNG took no
lock at all, so two threads could interleave DRBG generation on the same V/C
state and produce torn or duplicate output used directly as key material.

Closes f-7543.

### Fix

Two remedies, depending on whether the site can cheaply own its RNG:

| Site | File | Fix |
|---|---|---|
| `wolfSSL_BN_rand` | `src/ssl_bn.c` | hold `globalRNGMutex` across `wc_RNG_GenerateBlock` |
| `AddSession` | `src/ssl_sess.c` | hold `globalRNGMutex` across `wc_RNG_GenerateBlock`, global path only |
| `wolfSSL_ECDH_compute_key` | `src/pk_ec.c` | `wolfssl_make_rng()` in place of `wolfssl_make_global_rng()` |
| `wolfSSL_EC25519_shared_key` | `src/pk.c` | same, for the `WOLFSSL_CURVE25519_BLINDING` RNG |

The two lock sites are open-coded in the style already used in those files, so
no new locking convention is introduced.

The two shared-secret paths instead take a **per-call RNG** via
`wolfssl_make_rng()` — the helper the rest of `pk.c`/`pk_ec.c` already uses —
removing the sharing rather than serializing on it. Holding `globalRNGMutex`
across the whole shared-secret operation (the first version of this PR) would
have serialized all ECDH process-wide, including keys that carry their own RNG.
Both sites now NULL-check the RNG; `wolfSSL_ECDH_compute_key` additionally
guards the shared-secret call on `!err` and clears `key->rng` before the local
RNG is freed, so no dangling reference is left behind.

### Known trade-off

ECDH and X25519 now init and seed a fresh DRBG per call instead of reusing the
global one, which costs performance on those paths. Accepted as an interim
state — still better than the unsynchronized global.

### Intentionally not in this PR

- Roughly eighteen further sites reach the global only through
  `wolfssl_make_rng()`'s fallback, after a local `wc_InitRng()` has already
  failed. Same defect, far lower reachability.
- A concurrently shared `EC_KEY` is unsafe independent of the RNG
  (`wc_ecc_shared_secret_ex` writes `private_key->state` and dispatches on it).
  That needs per-key locking, which neither wolfSSL nor OpenSSL provides.

### Verification

- Builds clean, no new warnings, across `opensslall+keygen`, `opensslextra+all`,
  `dh+ecc` (no `OPENSSL_EXTRA`), `singlethreaded` and `wpas`, plus a
  `-DWOLFSSL_CURVE25519_BLINDING` build for the Curve25519 path.
- `make check`: 17/17 on `--enable-all`, 6/6 on `opensslall`, 0 failures.
- ThreadSanitizer clean: 32 threads x 150 iterations hammering `BN_rand`,
  `ECDH_compute_key`, `EC25519_shared_key` and `RAND_bytes`, 0 races.
  Negative control: reverting the `BN_rand` lock reports races in
  `Hash512_DRBG_Generate`.

No regression test is added: a threaded test is nondeterministic and there is no
ThreadSanitizer job in CI, so it would add flake without signal.

### DIFF
--- a/src/pk.c
+++ b/src/pk.c
@@ -5187,6 +5187,11 @@ int wolfSSL_EC25519_shared_key(unsigned char *shared, unsigned int *sharedSz,
     int res = 1;
     curve25519_key privkey;
     curve25519_key pubkey;
+#ifdef WOLFSSL_CURVE25519_BLINDING
+    WC_RNG* rng = NULL;
+    WC_DECLARE_VAR(tmpRng, WC_RNG, 1, 0);
+    int initTmpRng = 0;
+#endif
 
     WOLFSSL_ENTER("wolfSSL_EC25519_shared_key");
 
@@ -5206,8 +5211,13 @@ int wolfSSL_EC25519_shared_key(unsigned char *shared, unsigned int *sharedSz,
     }
     if (res) {
     #ifdef WOLFSSL_CURVE25519_BLINDING
-        /* An RNG is needed. */
-        if (wc_curve25519_set_rng(&privkey, wolfssl_make_global_rng()) != 0) {
+        /* An RNG is needed for blinding - create local or get global. */
+        rng = wolfssl_make_rng(tmpRng, &initTmpRng);
+        if (rng == NULL) {
+            WOLFSSL_MSG("wolfSSL_EC25519_shared_key failed to make RNG");
+            res = 0;
+        }
+        else if (wc_curve25519_set_rng(&privkey, rng) != 0) {
             res = 0;
         }
         else
@@ -5249,6 +5259,14 @@ int wolfSSL_EC25519_shared_key(unsigned char *shared, unsigned int *sharedSz,
         }
         wc_curve25519_free(&privkey);
     }
+
+#ifdef WOLFSSL_CURVE25519_BLINDING
+    /* Disposed of after privkey, which references it for blinding. */
+    if (initTmpRng) {
+        wc_FreeRng(rng);
+        WC_FREE_VAR_EX(rng, NULL, DYNAMIC_TYPE_RNG);
+    }
+#endif
 
     return res;
 #else
--- a/src/pk_ec.c
+++ b/src/pk_ec.c
@@ -5489,7 +5489,10 @@ int wolfSSL_ECDH_compute_key(void *out, size_t outLen,
     ecc_key* key = NULL;
 #if defined(ECC_TIMING_RESISTANT) && !defined(HAVE_SELFTEST) && \
     (!defined(HAVE_FIPS) || FIPS_VERSION_GE(5,0))
-    int setGlobalRNG = 0;
+    WC_RNG* rng = NULL;
+    WC_DECLARE_VAR(tmpRng, WC_RNG, 1, 0);
+    int initTmpRng = 0;
+    int setKeyRng = 0;
 #endif
 
     /* TODO: support using the KDF. */
@@ -5524,30 +5527,44 @@ int wolfSSL_ECDH_compute_key(void *out, size_t outLen,
 
     #if defined(ECC_TIMING_RESISTANT) && !defined(HAVE_SELFTEST) && \
         (!defined(HAVE_FIPS) || FIPS_VERSION_GE(5,0))
-        /* An RNG is needed. */
+        /* An RNG is needed - create local or get global. */
         if (key->rng == NULL) {
-            key->rng = wolfssl_make_global_rng();
-            /* RNG set and needs to be unset. */
-            setGlobalRNG = 1;
+            rng = wolfssl_make_rng(tmpRng, &initTmpRng);
+            if (rng == NULL) {
+                WOLFSSL_MSG("wolfSSL_ECDH_compute_key failed to make RNG");
+                err = 1;
+            }
+            else {
+                key->rng = rng;
+                /* RNG set and needs to be unset. */
+                setKeyRng = 1;
+            }
         }
     #endif
 
-        PRIVATE_KEY_UNLOCK();
-        /* Create secret using wolfSSL. */
-        ret = wc_ecc_shared_secret_ex(key, (ecc_point*)pubKey->internal,
-            (byte *)out, &len);
-        PRIVATE_KEY_LOCK();
-        if (ret != MP_OKAY) {
-            WOLFSSL_MSG("wc_ecc_shared_secret failed");
-            err = 1;
+        if (!err) {
+            PRIVATE_KEY_UNLOCK();
+            /* Create secret using wolfSSL. */
+            ret = wc_ecc_shared_secret_ex(key, (ecc_point*)pubKey->internal,
+                (byte *)out, &len);
+            PRIVATE_KEY_LOCK();
+            if (ret != MP_OKAY) {
+                WOLFSSL_MSG("wc_ecc_shared_secret failed");
+                err = 1;
+            }
         }
     }
 
 #if defined(ECC_TIMING_RESISTANT) && !defined(HAVE_SELFTEST) && \
     (!defined(HAVE_FIPS) || FIPS_VERSION_GE(5,0))
-    /* Remove global from key. */
-    if (setGlobalRNG) {
+    /* Clear before the RNG is disposed of - key must not keep a dangling
+     * reference to a local RNG. */
+    if (setKeyRng) {
         key->rng = NULL;
+    }
+    if (initTmpRng) {
+        wc_FreeRng(rng);
+        WC_FREE_VAR_EX(rng, NULL, DYNAMIC_TYPE_RNG);
     }
 #endif
 
--- a/src/ssl_bn.c
+++ b/src/ssl_bn.c
@@ -2146,10 +2146,19 @@ int wolfSSL_BN_rand(WOLFSSL_BIGNUM* bn, int bits, int top, int bottom)
             WOLFSSL_MSG("Failed to allocate buffer.");
             ret = 0;
         }
-        /* Generate bytes to cover bits. */
-        if ((ret == 1) && wc_RNG_GenerateBlock(rng, buff, len) != 0) {
-            WOLFSSL_MSG("wc_RNG_GenerateBlock failed");
+        /* Global RNG is shared, lock it while generating. */
+        if ((ret == 1) && (wc_LockMutex(&globalRNGMutex) != 0)) {
+            WOLFSSL_MSG("Bad Lock Mutex rng");
             ret = 0;
+        }
+
+        /* Generate bytes to cover bits. */
+        if (ret == 1) {
+            if (wc_RNG_GenerateBlock(rng, buff, len) != 0) {
+                WOLFSSL_MSG("wc_RNG_GenerateBlock failed");
+                ret = 0;
+            }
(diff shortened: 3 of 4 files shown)


src/pk.c
```diff
@@ -5235,6 +5235,13 @@ int wolfSSL_EC25519_shared_key(unsigned char *shared, unsigned int *sharedSz,
                     res = 0;
                 }
             }
+        #ifdef WOLFSSL_CURVE25519_BLINDING
+            /* Global RNG is shared, lock it while blinding uses it. */
+            if (res && (wc_LockMutex(&globalRNGMutex) != 0)) {
+                WOLFSSL_MSG("Bad Lock Mutex rng");
+                res = 0;
+            }
+        #endif
```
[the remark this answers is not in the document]
yosuke-wolfssl: Fixed

src/pk_ec.c
```diff
@@ -5524,30 +5525,44 @@ int wolfSSL_ECDH_compute_key(void *out, size_t outLen,
 
     #if defined(ECC_TIMING_RESISTANT) && !defined(HAVE_SELFTEST) && \
         (!defined(HAVE_FIPS) || FIPS_VERSION_GE(5,0))
-        /* An RNG is needed. */
+        /* An RNG is needed. Hold the global lock across use of key->rng. */
         if (key->rng == NULL) {
```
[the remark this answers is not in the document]
yosuke-wolfssl: Accurate observation, but not fixed here, and I don't think the suggested fix
is the right one. Three reasons:

**Locking before the check would serialize all ECDH.** `globalRNGMutex` would
be taken on every call, including keys carrying their own RNG that never touch
the global. That is a process-wide throughput cost paid to protect a field
those callers do not share.

**It would not close the window.** In the scenario described, the second
thread takes the lock, sees `key->rng != NULL`, and must then either release
it (back to unprotected) or hold it for the whole operation regardless of RNG
provenance. Even holding it, the first thread's cleanup sets
`key->rng = NULL` while the second is mid-operation, which returns
`MISSING_RNG_E` (`ecc.c:5047`). A different failure, still broken.

**A shared `EC_KEY` is already unsound here, independent of the RNG.**
`wc_ecc_shared_secret_ex` runs a state machine on the shared key: it writes
`private_key->state` at `ecc.c:5266`, `5285` and `5312` and dispatches on it
with `default: err = BAD_STATE_E`. Two concurrent callers on one key corrupt
that and can spuriously fail. No amount of `globalRNGMutex` fixes it; it needs
a per-key lock, which neither wolfSSL nor OpenSSL provides for `EC_KEY`.

The unlocked `key->rng == NULL` read is pre-existing and unchanged by this PR,
so there is no regression. The case that matters in practice, each thread
holding its own key and falling back to the global RNG, is fully fixed and is
what the ThreadSanitizer run exercises.

Happy to file the shared-`EC_KEY` concern separately if that is useful, since
it is a per-key locking problem rather than an RNG one.

src/ssl_sess.c
```diff
@@ -2162,15 +2162,29 @@ void AddSession(WOLFSSL* ssl)
      * this point, it won't on resumption. */
     if (idSz == 0 && ssl->options.side == WOLFSSL_CLIENT_END) {
         WC_RNG* rng = NULL;
+        int genRet;
+#if defined(HAVE_GLOBAL_RNG) && defined(OPENSSL_EXTRA)
+        int rngLocked = 0;
+#endif
         if (ssl->rng != NULL)
```
philljj, line 2169: I can't find a code path where `ssl->rng` would ever be null.

I think we should confirm if this is dead code before adding more complexity to it.

src/pk.c
philljj, line 1203: General comment: Why not follow this pattern and avoid touching the global RNG?

Or use the `wolfssl_make_rng()` pattern, which tries to generate a new local rng, and then falls back to global rng.

src/pk_ec.c
```diff
@@ -5524,30 +5525,44 @@ int wolfSSL_ECDH_compute_key(void *out, size_t outLen,
...
-            key->rng = wolfssl_make_global_rng();
-            /* RNG set and needs to be unset. */
-            setGlobalRNG = 1;
+            rng = wolfssl_make_global_rng();
```
philljj: If you use `wolfssl_make_rng()` you can attempt to use a local rng first, and only fallback to the global.

src/pk_ec.c
```diff
@@ -5524,30 +5525,44 @@ int wolfSSL_ECDH_compute_key(void *out, size_t outLen,
...
+            if (rng == NULL) {
+                WOLFSSL_MSG("Failed to use global RNG.");
```
philljj: I'm not sure this extra rng checking logic before `wc_ecc_shared_secret_ex()` and `wc_curve25519_shared_secret_ex()` makes sense, and it could have surprising side effects.

1. wc_ecc_shared_secret_ex already checks if `rng == NULL` and will error handle that case.

2. If CryptoCB is being used, the rng could be ignored, which means we're holding a global RNG lock for no reason.

3. Alternatively, if the user CryptoCB code also tries to lock the global RNG, then this will deadlock.

4. These asymmetric ops can be non-deterministic, and holding a global lock across them could have an ugly performance penalty.

Is there a way to defer setting the rng, or allow the logic to be handled in ecc.c? E.g. `ecc_blind_k_rng()` falls back to a `WC_RNG local_rng;` when `rng == NULL`.

src/ssl_bn.c
philljj, line 2138: Could use `wolfssl_make_rng()` instead, and only fall back to global if local rng fails.

src/pk.c
```diff
@@ -5206,8 +5209,13 @@ int wolfSSL_EC25519_shared_key(unsigned char *shared, unsigned int *sharedSz,
     }
     if (res) {
     #ifdef WOLFSSL_CURVE25519_BLINDING
-        /* An RNG is needed. */
-        if (wc_curve25519_set_rng(&privkey, wolfssl_make_global_rng()) != 0) {
+        /* An RNG is needed for blinding. */
+        rng = wolfssl_make_global_rng();
```
philljj: I think you could use `wolfssl_make_rng()` here and only fallback to global rng if needed.

philljj (changes requested):

src/pk_ec.c
yosuke-wolfssl: Done. Now uses `wolfssl_make_rng()`, following the `wolfSSL_ECDSA_sign`
template.

Renamed `setGlobalRNG` to `setKeyRng`, since the flag may now refer to a local
RNG.

src/pk.c
yosuke-wolfssl: Done. Now uses `wolfssl_make_rng()`, following the `wolfSSL_ECDSA_sign`
template.

The RNG is freed after `wc_curve25519_free(&privkey)`, since privkey holds the
blinding reference to it.

src/pk_ec.c
yosuke-wolfssl: All four land, and (3) settled it: a user CryptoCB that also takes
`globalRNGMutex` would deadlock, and that would be a hazard this PR introduced
rather than inherited. The lock across both shared-secret calls is gone, and
the redundant pre-check with it.

Your `ecc_blind_k_rng()` pointer is a good end state. I left it out here since
it changes wolfCrypt behavior for non compat callers too, but happy to do it as
a follow up.

src/ssl_bn.c
yosuke-wolfssl: I'd keep this one on global plus lock. `BN_rand` produces a handful of bytes,
so `wolfssl_make_rng()` would mean a full `wc_InitRng()` per call, paid 100% of
the time, to avoid an essentially uncontended mutex. Every cheap byte producer
in the tree goes this way, `RAND_bytes()` included.

The rarity of `wc_InitRng()` failing cuts the same direction: it means the
fallback rarely fires, which is what makes the lock cheap rather than what
makes it unnecessary.

Happy to flip it if you disagree.

src/ssl_sess.c
yosuke-wolfssl: Reachable, though only in some configs. `FreeHandshakeResources()` NULLs
`ssl->rng` at `internal.c:9569-9573`, and for TLS 1.3 the clause at
`internal.c:9562-9563` covers the client when `HAVE_SESSION_TICKET` is defined
and `WOLFSSL_POST_HANDSHAKE_AUTH` is not. `wolfSSL_connect_TLSv13()` calls it at
`tls13.c:14943` when the handshake ends; the NewSessionTicket arrives after
that, and `DoSessionTicket()` calls `AddSession()` at `tls13.c:12985` with
`ssl->rng` already NULL. The `tls1_1 == 0` disjunct at `internal.c:9552` reaches
it for a TLS 1.0 client too.

Under `--enable-all`, `WOLFSSL_POST_HANDSHAKE_AUTH` is on and the client keeps
its RNG, which is probably why it reads as dead.

This is a static read of the `#if` chain; I have not built that config and
confirmed empirically. Can do if you'd like, otherwise I'll add a short comment
recording the reachable config.

src/pk.c
yosuke-wolfssl: Thanks for reviewing.
Your suggestion made me survey what the tree actually does. All 12 existing
`wolfssl_make_rng()` callers are keygen, prime, or sign operations, where
`wc_InitRng()` (entropy draw, 2-3 `XMALLOC`s, DRBG instantiate, health test) is
rounding error. The cheap byte producers all go the other way: `RAND_bytes()`,
`RAND_pseudo_bytes()` and the SMIME boundary use `&globalRNG` plus the mutex.

So the convention looks cost proportional, and I've split the PR on that line:

| Site | Change |
|---|---|
| `ECDH_compute_key` (`pk_ec.c`) | now `wolfssl_make_rng()`, no lock |
| `EC25519_shared_key` (`pk.c`) | now `wolfssl_make_rng()`, no lock |
| `BN_rand`, `AddSession` | unchanged, global plus lock |

I'm happy to flip them if you have different design.

yosuke-wolfssl:
Hello @philljj ,
I reworked on this. Could you please review it again ?

src/ssl_bn.c
philljj: That's fair and a good point.

philljj (approved):

dgarske (approved):
Previously (before this PR) ECDH and X25519 shared-secret paths used the global RNG, but now always instantiate and seed a fresh DRBG on every call. This could impact performance for some customers. I think originally this PR grabbed the global and locked it during the entire ECDH/X25519 which caused a different issue. I am not sure what the best solution is and it requires a larger refactor, which @philljj has agreed to take on. For now we'll merge this PR, but it does need some improvement soon (before next release). Its still better than what was in master. Also the PR title and description are wrong, so please correct those. Thanks

PR #11053, opened 2026-08-04
Xilinx asu ecc standalone

night1rider:


wolfcrypt/src/port/xilinx/versal_gen2_asu/asu_ecc.c
```diff
@@ -0,0 +1,842 @@
...
+#ifdef HAVE_ED25519
+        case WC_PK_TYPE_ED25519_SIGN:
+            return wc_AsuEd25519Sign(info);
+        case WC_PK_TYPE_ED25519_VERIFY:
+            return wc_AsuEd25519Verify(info);
+#endif
+#ifdef HAVE_ED448
+        case WC_PK_TYPE_ED448:
+            return wc_AsuEd448Sign(info);
+        case WC_PK_TYPE_ED448_VERIFY:
```
night1rider, line 951: https://github.com/wolfSSL/wolfssl/pull/10886

night1rider, line 951: https://github.com/wolfSSL/wolfssl/pull/10886

wolfcrypt/src/port/xilinx/versal_gen2_asu/asu_ecies.c
```diff
@@ -0,0 +1,603 @@
...
+    sc.req = mem.req;
+    sc.isEncrypt = 0;
+    status = wc_AsuTransact(wc_AsuEciesSubmit, &sc, &addl);
+
+    wc_AsuCacheInvalidate(out, ctLen);
+
+    WC_ASU_PRINTF("[ASU] ecies dec st=%u addl=0x%x\r\n",
+        (unsigned int)status, (unsigned int)addl);
+
+    if (status != XST_SUCCESS) {
```
night1rider, line 605: Hardware should be returning with an error, checking addl would only be for a more specific error code return.

wolfcrypt/src/port/xilinx/versal_gen2_asu/asu_ecc.c
```diff
@@ -0,0 +1,942 @@
...
+    if (msgLen != 0) {
+        XFREE(msg, NULL, DYNAMIC_TYPE_TMP_BUFFER);
+    }
+
+    WC_ASU_PRINTF("[ASU] ed25519 verify st=%u addl=0x%x\r\n",
+        (unsigned int)status, (unsigned int)addl);
+
+    /* Only VERIFIED passes. An S that is too big is a bad argument, and
+     * anything else is just a bad signature. */
+    if (status == XST_SUCCESS && addl == (word32)XASU_ECC_SIGNATURE_VERIFIED) {
```
night1rider, line 682: This should be handled by the hardware/asu firmware.

wolfcrypt/src/port/xilinx/versal_gen2_asu/asu_ecc.c
```diff
@@ -0,0 +1,942 @@
...
+    /* The ASU only read msg, so free it now. */
+    if (msgLen != 0) {
+        XFREE(msg, NULL, DYNAMIC_TYPE_TMP_BUFFER);
+    }
+
+    WC_ASU_PRINTF("[ASU] ed448 verify st=%u addl=0x%x\r\n",
+        (unsigned int)status, (unsigned int)addl);
+
+    /* Same rules as Ed25519 above. */
+    if (status == XST_SUCCESS && addl == (word32)XASU_ECC_SIGNATURE_VERIFIED) {
```
night1rider, line 902: This should be handled by the hardware/asu firmware.

wolfcrypt/src/port/xilinx/versal_gen2_asu/asu_ecies.c
```diff
@@ -0,0 +1,603 @@
...
+    }
+
+    /* uncompressed point: 0x04 then Qx and Qy */
+    pubKeySz = 1U + (2U * (word32)keyLen);
+    /* Check the length first so the size math below cannot overflow. */
+    if (msgSz > 0xFFFFFFFFU - pubKeySz - (word32)WC_ASU_ECIES_NONCE_SZ -
+            (word32)WC_ASU_ECIES_TAG_SZ) {
+        return BAD_FUNC_ARG;
+    }
+    need = pubKeySz + (word32)WC_ASU_ECIES_NONCE_SZ + msgSz +
```
night1rider, line 360: No, ECIES accepts an arbitrary range here, and the hardware will error out when it hits the dma limit for input size.

wolfcrypt/benchmark/benchmark.c
```diff
@@ -7018,6 +7075,26 @@ void bench_aesccm(int useDeviceID)
...
 }
+
+void bench_aesccm(int useDeviceID)
+{
+#ifdef WC_BENCH_AES_IV_SWEEP
+    /* One enc/dec row per nonce length from 7 to 13 bytes. */
+    word32 nsz;
+    char encLabel[28], decLabel[28];
+
+    for (nsz = 7; nsz <= 13; nsz++) {
```
night1rider, line 7146: addressed

wolfcrypt/src/port/xilinx/versal_gen2_asu/asu_ecies.c
```diff
@@ -0,0 +1,603 @@
...
+        if (macSaltSz > 0U) {
+            return CRYPTOCB_UNAVAILABLE;
+        }
+        if (infoSz == 0U) {
+            WC_ASU_PRINTF("[ASU] ecies: GCM needs a non-empty KDF context\r\n");
+            return CRYPTOCB_UNAVAILABLE;
+        }
+        /* The ASU uses the first half of the derived key, so only the client
+         * side is allowed here. */
+        if (proto != REQ_RESP_CLIENT) {
```
night1rider, line 330: addressed

wolfcrypt/benchmark/benchmark.c
```diff
@@ -7018,6 +7075,26 @@ void bench_aesccm(int useDeviceID)
...
+        (void)XSNPRINTF(encLabel, sizeof(encLabel),
+            "AES-CCM-n%u-enc", (unsigned)nsz);
+        (void)XSNPRINTF(decLabel, sizeof(decLabel),
+            "AES-CCM-n%u-dec", (unsigned)nsz);
+        bench_aesccm_internal(useDeviceID, nsz, encLabel, decLabel);
```
night1rider: addressed

wolfcrypt/benchmark/benchmark.c
```diff
@@ -6377,6 +6404,35 @@ void bench_gmac(int useDeviceID)
...
+#endif
+
+    bench_gmac_internal(useDeviceID, 12, gmacStr);
+
+#ifdef WC_BENCH_AES_IV_SWEEP
+    /* Extra row using a 16-byte IV alongside the 12-byte default above. */
+    {
+        char gmacIvStr[40];
+        (void)XSNPRINTF(gmacIvStr, sizeof(gmacIvStr), "%s-iv16", gmacStr);
+        bench_gmac_internal(useDeviceID, 16, gmacIvStr);
```
night1rider, line 6490: addressed

wolfcrypt/src/port/xilinx/versal_gen2_asu/asu_ecies.c
```diff
@@ -0,0 +1,603 @@
...
+
+    if (status != XST_SUCCESS) {
+        /* A bad tag and a hardware fault look the same here, so wipe the
+         * output and never hand back unchecked plaintext. */
+        ForceZero(out, ctLen);
+        /* Push the zeros out to memory, or only the cache copy is cleared. */
+        wc_AsuCacheFlush(out, ctLen);
+        *info->pk.eciesdecrypt.outSz = 0;
+        wc_AsuEciesReqFree(&mem);
+        return WC_HW_E;
```
night1rider, line 613: No, this will not work, we only get back a generic failure, not that the tag was good or bad.

wolfcrypt/src/port/xilinx/versal_gen2_asu/asu_ecc.c
```diff
@@ -0,0 +1,942 @@
...
+    }
+
+    ret = wc_AsuEccReqNew(&mem);
+    if (ret != 0) {
+        return ret;
+    }
+
+    XMEMSET(mem.req, 0, sizeof(*mem.req));
+    /* If the key does not fit in keyLen bytes we fall back to software. */
+    if (mp_to_unsigned_bin_len(wc_ecc_key_get_priv(key), mem.req->key,
```
night1rider, line 315: addressed

wolfcrypt/src/port/xilinx/versal_gen2_asu/asu_ecdh.c
```diff
@@ -0,0 +1,284 @@
...
+
+    ret = wc_AsuEcdhReqNew(&mem);
+    if (ret != 0) {
+        return ret;
+    }
+
+    XMEMSET(mem.req, 0, sizeof(*mem.req));
+    /* A key too big for keyLen bytes is not something the ASU can take, so
+     * fall back to software rather than report a hardware error. */
+    if (mp_to_unsigned_bin_len(wc_ecc_key_get_priv(priv), mem.req->privKey,
```
night1rider, line 242: addressed

wolfcrypt/benchmark/benchmark.c
```diff
@@ -15398,7 +15724,7 @@ void bench_curve448KeyAgree(void)
 #endif /* HAVE_CURVE448 */
 
 #ifdef HAVE_ED448
-void bench_ed448KeyGen(void)
+void bench_ed448KeyGen(int useDeviceID)
```
philljj: I don't think the wc_ed448_make_key implementation supports crypto callbacks (it's only sign / verify). So the useDeviceID is unused.

wolfcrypt/benchmark/benchmark.c
```diff
@@ -6377,6 +6456,36 @@ void bench_gmac(int useDeviceID)
...
+    const char* gmacStr = "GMAC Default";
+#endif
+
+    bench_gmac_internal(useDeviceID, 12, gmacStr);
+
+#ifdef WC_BENCH_AES_IV_SWEEP
+    /* Extra row using a 16-byte IV alongside the 12-byte default above. The
+     * stats list keeps the label by pointer, so it must outlive this call. */
+    {
+        static char gmacIvStr[40];
```
philljj, line 6488: I'm not sure this is memory or thread safe.

gmacIvStr gets passed down

1. bench_gmac_internal
2. bench_stats_sym_finish
3. bench_stats_add()

In bench_stats_add() it becomes `const char * desc`, which does
```
bstat->desc == desc 
```

so it shallow copies the gmacIvStr pointer.

wolfcrypt/benchmark/benchmark.c
```diff
@@ -5389,7 +5468,8 @@ void bench_rng_sha512(void)
     bench_stats_prepare();
 
 #ifndef HAVE_FIPS
-    ret = wc_InitRng_ex(&myrng, HEAP_HINT, devId);
+    ret = wc_InitRng_ex(&myrng, HEAP_HINT,
+        useDeviceID ? devId : INVALID_DEVID);
 #else
     ret = wc_InitRng(&myrng);
```
philljj, line 5480: The useDeviceID is unused when HAVE_FIPS

philljj (dismissed):

wolfcrypt/benchmark/benchmark.c
night1rider: yes, you are right. I  overreached on accident when updating the benchmarks. Should be reverted now.

wolfcrypt/benchmark/benchmark.c
night1rider: I think it is only an issue if `WC_ENABLE_BENCH_THREADING` is defined. I changed it so `WC_BENCH_AES_IV_SWEEP` will not build with that macro. The AES-CCM labels had the same problem and are covered too.

wolfcrypt/benchmark/benchmark.c
night1rider: Under `HAVE_FIPS` both `RNG` benchmarks call `wc_InitRng`, which takes no device id, so the second row ran the same code as the first while being labeled hardware. I stopped that second row from running in a FIPS build, for both DRBGs, and left the FIPS branch inside the functions alone.

IDE/XilinxSDK/vitis_sdt/lib/sw_services/wolfssl/src/user_settings_template.h
```diff
@@ -0,0 +1,88 @@
...
+#define NO_MAIN_DRIVER
+#define WOLFSSL_IGNORE_FILE_WARN
+
+/* No wall clock: certificate date checks are skipped. For production,
+ * remove this and provide XTIME and XGMTIME instead. */
+#define NO_ASN_TIME
+
+/* WARNING: test-only seed so builds work out of the box. For production
+ * use real entropy: the ASU offload below or CUSTOM_RAND_GENERATE_SEED. */
+#define WOLFSSL_GENSEED_FORTEST
```
JacobBarthelmeh: Let's avoid having this in the user_settings template if possible. It should be the TRNG from the hardware or the user explicitly setting this define.

wolfcrypt/src/port/xilinx/versal_gen2_asu/asu_ecdh.c
```diff
@@ -0,0 +1,289 @@
...
+    if (ret != 0) {
+        return ret;
+    }
+    /* Both keys must be on the same supported curve. */
+    if (pub->dp == NULL || pub->dp->id != priv->dp->id) {
+        return CRYPTOCB_UNAVAILABLE;
+    }
+    /* If the peer point is all zeros, let software return the proper error
+     * instead of the ASU failing. */
+    if (mp_iszero(pub->pubkey.x) && mp_iszero(pub->pubkey.y)) {
```
JacobBarthelmeh, line 225: Not sure this is the proper response? Am thinking of builds where it is crypto callback only. Is this something that the ASU documentation says they do not handle?

wolfcrypt/src/port/xilinx/versal_gen2_asu/asu_ecies.c
```diff
@@ -0,0 +1,615 @@
...
+    /* The caller passes a key here, but the ASU makes its own. */
+    if (privKey == NULL) {
+        return CRYPTOCB_UNAVAILABLE;
+    }
+    /* Both keys have to be on the same curve. */
+    if (privKey->dp == NULL || privKey->dp->id != pubKey->dp->id) {
+        return CRYPTOCB_UNAVAILABLE;
+    }
+    /* If the peer point is all zeros, let software return the proper error. */
+    if (mp_iszero(pubKey->pubkey.x) && mp_iszero(pubKey->pubkey.y)) {
```
JacobBarthelmeh, line 344: Same comment as the ECDH case.

wolfcrypt/src/port/xilinx/versal_gen2_asu/asu_ecies.c
```diff
@@ -0,0 +1,615 @@
...
+        }
+        /* The ASU always uses the first half of the derived key, which is the
+         * server half here. A context is never built with protocol zero. */
+        if (proto != REQ_RESP_SERVER) {
+            return CRYPTOCB_UNAVAILABLE;
+        }
+    }
+
+    /* The sender key must be uncompressed, so turn down compressed. */
+    if (msgSz < 1U || msg[0] != (byte)ECC_POINT_UNCOMP) {
```
JacobBarthelmeh, line 527: If logging is available a debug log message would be helpful here. It might be hard debugging the failure reason otherwise.

JacobBarthelmeh: Additionally, does this fall back to the software implementation? I could see users accidentally thinking they are using the ASU when in fact it was falling back to software.

IDE/XilinxSDK/vitis_sdt/lib/sw_services/wolfssl/src/user_settings_template.h
night1rider: `WOLFSSL_GENSEED_FORTEST` is gone from the template. Its now replaced by an `#error` in an `#if 1` so a new BSP stops until someone picks a seed source on purpose.

The comment beside it lists the choices, including `WOLF_CRYPTO_CB` and the two Xilinx ports that seed wolfSSL themselves.

wolfcrypt/src/port/xilinx/versal_gen2_asu/asu_ecdh.c
night1rider: Not from the ASU documentation.

I added a debug message and kept sending this to software, so the exact error comes back, software says ECC_INF_E for a zero point, while the ASU only reports a bad point.

It is also a cheap check to do here rather than build a request, send it to hardware and wait for the answer. In a crypto callback only build the caller would get an invalid result, but the debug message still shows why.

wolfcrypt/src/port/xilinx/versal_gen2_asu/asu_ecies.c
night1rider: Same as the ECDH case. I added a debug message and kept sending this to software, so the exact error comes back.

wolfcrypt/src/port/xilinx/versal_gen2_asu/asu_ecies.c
night1rider: Added a debug for this, also added this warning:

```
/* The ASU takes raw x and y only, so a compressed key always runs in software.
 * Define WOLFSSL_VERSAL_GEN2_ASU_NO_COMP_KEY_WARN to silence this. */
#if defined(HAVE_COMP_KEY) && \
    !defined(WOLFSSL_VERSAL_GEN2_ASU_NO_COMP_KEY_WARN)
    #warning "ASU ECIES cannot offload compressed keys, software handles them"
#endif
```

IDE/XilinxSDK/vitis_sdt/lib/sw_services/wolfssl/src/user_settings_template.h
JacobBarthelmeh: For the template I think it should default to the Versal TRNG use and then only error out if that define was removed. I think this should be a template for builds to use for Versal hardware acceleration, where most users can copy/paste it and get started with a build that has Versal gen 2 hardware acceleration on (minus the pitfall of a fake seed source). For other options on generic macros to set users can look at ./examples/configs/

wolfcrypt/src/port/xilinx/versal_gen2_asu/asu_ecdh.c
JacobBarthelmeh: Can we just return ECC_INF_E here since checking? Am leery of having a fallback to software path when it was set to attempt hardware offloading, even for having the software path doing checks like this. The fail case is a suggestion though and not a blocker.

wolfcrypt/src/port/xilinx/versal_gen2_asu/asu_ecies.c
JacobBarthelmeh: After thinking more on this specific case, here we should error out rather than fall back if it's not supported. Think of cases where users want additional hardening or side channel protection by offloading to the hardware ports. They may not be checking debug logs and it falling back to software removes the extra protections they thought they were getting. Better to fail the operation then continue with software only.

JacobBarthelmeh (approved):

PR #11059, opened 2026-08-05
LMS/XMSS: don't skip the software reload when a read callback is set

Frauschi:
## Description

`wc_LmsKey_Reload` and `wc_XmssKey_Reload` return success without doing any work whenever `key->devId != INVALID_DEVID`, assuming a device-bound key keeps its private state inside the device. That does not hold for a caller that sets a devId only to route primitives to a hardware accelerator while keeping the key state in its own storage - a wolfHSM server configures a server-wide devId for AES/ECC/RSA, then installs LMS/XMSS read/write callbacks and calls `Reload` before signing.

With such a devId configured, `Reload` returned 0 immediately and left the key unusable for the first sign that follows:

- **LMS**: `priv_data` stays NULL, `wc_hss_sign` finds `priv.inited` clear and calls `wc_hss_init_auth_path`, which derives its first read from a NULL `priv` pointer - a bus fault on a target with nothing mapped at low addresses.
- **XMSS**: `key->sk` is never allocated (`sk_len` stays 0), so the sign fails with `IO_FAILED_E` for a read callback that honours the length it is given, and faults for one that writes a fixed-size record.

Key generation hid the problem: `wc_LmsKey_MakeKey` already treats devId as "offer the operation to the callback, fall back to software on `CRYPTOCB_UNAVAILABLE`". `Reload` had no equivalent fall-through - and no cryptocb reload op exists to offer it to - so the same key and the same devId were interpreted two different ways by the same API.

The fix qualifies the short-circuit with `key->read_private_key == NULL`: a caller that installed a read callback is asking for the software reload to fetch state through it, while a genuinely device-backed key installs no such callback. Key generation still offers the operation to the crypto callback, so a port with real stateful-hash-signature hardware is unaffected.

## Testing

`test_wc_LmsKey_reload_devid` and `test_wc_XmssKey_reload_devid` cover both arms of the new condition, driven by a crypto callback that declines every operation with `CRYPTOCB_UNAVAILABLE` (the accelerator this fix is about):

- read callback set: `Reload` must come back with the private key expanded (`priv_data` / `sk`), and sign/verify must succeed;
- no read callback: `Reload` must still short-circuit and leave the key state in the device.

Asserting on the expanded key means the old behaviour fails the assertion rather than the NULL dereference it leads to. Verified against `./configure --enable-lms --enable-xmss --enable-cryptocb --enable-debug`: both tests fail before the fix and pass after it, with the rest of the `lms` and `xmss` groups unaffected (every other key there uses `INVALID_DEVID`). Also builds clean in an `--enable-all` tree, where the whole group compiles out.

The public Doxygen for both Reload functions now documents that the read callback, not the devId, decides whether the software reload runs, and that a `0` return can mean nothing was loaded.

### DIFF
--- a/doc/dox_comments/header_files/wc_lms.h
+++ b/doc/dox_comments/header_files/wc_lms.h
@@ -359,9 +359,14 @@ int wc_LmsKey_MakeKey(LmsKey* key, WC_RNG* rng);
     the private key bytes; the parameter set is metadata the
     application owns).
 
-    \return 0 on success.
+    With WOLF_CRYPTO_CB, a key on a devId with no read callback is
+    taken to be device-backed and Reload is a no-op. With a read
+    callback the software reload runs as usual.
+
+    \return 0 on success, including the device-backed no-op.
     \return BAD_FUNC_ARG if any required pointer is NULL.
     \return WC_LMS_RC_* mapped error if the read callback fails.
+    \return IO_FAILED_E if the private key could not be read.
 
     \param [in,out] key Pointer to an LmsKey with parameters and read
     callback set.
--- a/doc/dox_comments/header_files/wc_xmss.h
+++ b/doc/dox_comments/header_files/wc_xmss.h
@@ -270,9 +270,15 @@ int wc_XmssKey_MakeKey(XmssKey* key, WC_RNG* rng);
     The same parameter set selected at key-generation time must be
     reapplied with wc_XmssKey_SetParamStr() before calling Reload.
 
-    \return 0 on success.
+    With WOLF_CRYPTO_CB, a key on a devId with no read callback is
+    taken to be device-backed and Reload is a no-op. With a read
+    callback the software reload runs as usual, which needs the write
+    callback set too.
+
+    \return 0 on success, including the device-backed no-op.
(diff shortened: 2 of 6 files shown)


Frauschi:
Jenkins retest this please

dgarske:
Jenkins retest this please "PRB-fips-ready-config'"

dgarske:
Jenkins retest this please

doc/dox_comments/header_files/wc_lms.h
```diff
@@ -359,9 +359,18 @@ int wc_LmsKey_MakeKey(LmsKey* key, WC_RNG* rng);
     the private key bytes; the parameter set is metadata the
     application owns).
 
-    \return 0 on success.
+    In a WOLF_CRYPTO_CB build, a key initialized with a devId decides
```
dgarske: Again the inline code comments are way too verbose. Please compress. Thanks

dgarske (changes requested):

doc/dox_comments/header_files/wc_lms.h
Frauschi: Fixed

dgarske (approved):

PR #11060, opened 2026-08-05
Make `--enable-all --enable-asynccrypt-sw` work end to end

Frauschi:
## Description

Nothing in CI builds an asynchronous configuration that actually returns `WC_PENDING_E` from a full feature set. `--enable-all` turns on cryptocb, and `configure.ac` only auto-enables the software simulator when no backend is set, so the `asynccrypt-all` entries define `WOLFSSL_ASYNC_CRYPT` without ever suspending an operation. The `asynccrypt-sw` entries do suspend, but they build only OCSP stapling, so they compile neither TLS 1.3 post-handshake authentication nor DTLS.

Building that combination surfaced five defects: three in the wolfCrypt tests and ECCSI, two in the TLS record layer. This series fixes all five and adds the CI entry that covers them.

### wolfCrypt

- **ECIES crypto callback test** - `ecc_encrypt_cryptocb_test()` checked `wc_ecc_make_key()` directly while the sibling `ecc_encrypt_test()` already waited. Mirrors the sibling.
- **`wc_MakeEccsiKey()`** (library) - returned `WC_PENDING_E` to a caller that cannot use it: ECCSI exposes no async API, no event to poll and no resume path. Waits before returning. Only this one of the three `wc_ecc_make_key_ex()` calls in `eccsi.c` needs it; the other two are preceded by `wc_ecc_free()`, which clears the marker their pending path is gated on.
- **`cryptocb_test()`** - the mock device clears `devId` but not `asyncDev.marker`, so its inner keygen still goes pending and hands `WC_PENDING_E` back to its own caller. Waits at the call site, as every other keygen in that file does.

### TLS record layer (`src/internal.c`)

- **Post-handshake auth leaves the record unfinished.** `DoTls13CertificateRequest()` sends the client's flight from inside `DoTls13HandShakeMsg()`. When that flight goes pending, `DoProcessReplyEx()` returns before the end-of-record accounting, leaving the record short by `ssl->keys.padSz`. The client resumes through `wolfSSL_negotiate()` rather than by reprocessing the record, so the next `ProcessReply()` parses the record's own MAC as a record header and fails with `-326 VERSION_ERROR`. Fix mirrors both halves of the end-of-record block: set `runProcessingOneMessage` when content remains, advance past the padding only at the boundary.
- **The record size probe clobbers a suspended build.** `SendData()` sizes its buffer with `BuildMessage(..., sizeOnly=1)`, which keeps its arguments on the stack but drives the shared `ssl->options.buildMsgState`. A re-probe on retry rewinds the state, so the resumed build re-runs its begin and size stages against arguments that already carry the overhead and overruns the buffer it correctly sized, failing with `-132 BUFFER_E`. Fix saves and restores `buildMsgState` and `buildArgsSet` around the probe, and skips `SetKeysSide()` when `sizeOnly` is set.

### CI

Adds `asynccrypt-sw-all-dtls13` (`--enable-asynccrypt-sw --enable-all --enable-dtls13`) to `.github/workflows/async.yml`, which covers both reproducers.

## Reproducers

Both need `./configure --enable-all --enable-asynccrypt-sw`.

```bash
# Post-handshake auth, tests/test-tls13.conf: client dies with -326
./examples/server/server -v 4 -l TLS13-AES128-GCM-SHA256 -Q -2 -p 11119 &
./examples/client/client -v 4 -l TLS13-AES128-GCM-SHA256 -Q -2 -p 11119

# DTLS MTU, tests/test-dtls-mtu.conf: client dies with -132
./examples/server/server -e -u -f -v 3 -l ECDHE-RSA-AES128-SHA -2 -p 11151 &
./examples/client/client -B 4000,1310 -u -f -v 3 -l ECDHE-RSA-AES128-SHA -2 -p 11151
```

## Testing

- `--enable-all --enable-asynccrypt-sw`: `testwolfcrypt` passes in full, `make check` 17 passed / 6 skipped / 0 failed. Both reproducers complete.
- Plain `--enable-all`: `make check` 17 passed / 6 skipped / 0 failed. Non-async paths are unaffected; the changes are guarded by `WOLFSSL_ASYNC_CRYPT`.
- Both warning-free under the CFLAGS the os-check workflow applies.
- New `test_record_size_preserves_build_msg_state` (TLS 1.2 and TLS 1.3) parks a connection at `BUILD_MSG_ENCRYPT`, runs the probe and requires the state to survive. Proven non-vacuous by stubbing the restore back out.
- The new CI entry measured `pass [1.6 min]` locally with `parallel-make-check.py`.

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -25480,6 +25480,22 @@ static int DoProcessReplyEx(WOLFSSL* ssl, int allowSocketErr)
                                             ssl->buffers.inputBuffer.buffer,
                                             &ssl->buffers.inputBuffer.idx,
                                             ssl->curStartIdx + ssl->curSize);
+    #ifdef WOLFSSL_ASYNC_CRYPT
+                        /* Post-handshake authentication runs the connect state
+                         * machine from inside this handler and resumes through
+                         * wolfSSL_negotiate() rather than by reprocessing this
+                         * record, which is why it leaves processReply at
+                         * doProcessInit. Finish the record's accounting here,
+                         * or the trailing MAC is parsed as the next record
+                         * header and the read fails with VERSION_ERROR. An
+                         * ordinary pending message leaves processReply at
+                         * runProcessingOneMessage and must not be advanced. */
+                        if (ret == WC_NO_ERR_TRACE(WC_PENDING_E) &&
+                                ssl->options.processReply == doProcessInit &&
+                                IsEncryptionOn(ssl, 0)) {
+                            ssl->buffers.inputBuffer.idx += ssl->keys.padSz;
+                        }
+    #endif
     #ifdef WOLFSSL_EARLY_DATA
                         if (ret != 0)
                             return ret;
--- a/src/internal.c
+++ b/src/internal.c
@@ -25480,7 +25480,7 @@ static int DoProcessReplyEx(WOLFSSL* ssl, int allowSocketErr)
                                             ssl->buffers.inputBuffer.buffer,
                                             &ssl->buffers.inputBuffer.idx,
                                             ssl->curStartIdx + ssl->curSize);
-    #ifdef WOLFSSL_ASYNC_CRYPT
+    #if defined(WOLFSSL_ASYNC_CRYPT) && defined(WOLFSSL_POST_HANDSHAKE_AUTH)
                         /* Post-handshake authentication runs the connect state
                          * machine from inside this handler and resumes through
                          * wolfSSL_negotiate() rather than by reprocessing this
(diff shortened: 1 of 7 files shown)


src/internal.c
```diff
@@ -25480,6 +25480,36 @@ static int DoProcessReplyEx(WOLFSSL* ssl, int allowSocketErr)
...
+                         * record has to be resumed from inside rather than at a
+                         * record header. */
+                        if (ret == WC_NO_ERR_TRACE(WC_PENDING_E) &&
+                                ssl->options.processReply == doProcessInit) {
+                            if ((ssl->buffers.inputBuffer.idx -
+                                    ssl->curStartIdx) < ssl->curSize) {
+                                ssl->options.processReply =
+                                    runProcessingOneMessage;
+                            }
+                            else if (IsEncryptionOn(ssl, 0)) {
```
Frauschi, line 25506: False positive: the block this comment anchors to sits in the final `else` of the `if (ssl->options.dtls)` / `else if (!IsAtLeastTLSv1_3)` / `else` chain that starts at line 25399, so it only runs when `ssl->options.dtls == 0`. With DTLS off, the bypass at line 25543 short-circuits on `!ssl->options.dtls` and `processReply` returns `WC_PENDING_E` immediately — the end-of-record block at 25600 is not reached, so `padSz` is added exactly once. For DTLS, the new block never executes at all. The DTLS 1.3 async path is covered by the asynccrypt-sw-all-dtls13 CI entry added in this PR.

dgarske:
Jenkins retest this please "PRB-fips-ready-config"

src/internal.c
```diff
@@ -25480,6 +25486,36 @@ static int DoProcessReplyEx(WOLFSSL* ssl, int allowSocketErr)
                                             ssl->buffers.inputBuffer.buffer,
                                             &ssl->buffers.inputBuffer.idx,
                                             ssl->curStartIdx + ssl->curSize);
+    #if defined(WOLFSSL_ASYNC_CRYPT) && defined(WOLFSSL_POST_HANDSHAKE_AUTH)
+                        /* Post-handshake authentication runs the connect state
```
dgarske: Consider condensing the inline comments

dgarske (changes requested):

src/internal.c
Frauschi: Reduced the long ones.

dgarske (approved):

PR #11061, opened 2026-08-05
Fix buffer overruns in certificate signing and the EVP_PKEY populate paths

Frauschi:
# Fix buffer overruns in certificate signing and the EVP_PKEY populate paths

Three heap buffer overruns plus a set of related `EVP_PKEY` state defects, all reproducible on master and none introduced by recent work.

**1. `SignCert()` under-counts the signature wrapper.** The check `requestSz + MAX_SEQ_SZ * 2 + sigSz > buffSz` covers the outer SEQUENCE but not the `signatureAlgorithm` AlgorithmIdentifier or `signatureValue` BIT STRING header that `AddSignature()` also writes. The real wrapper is 24 bytes for RSA and 19 for ECDSA against a 12 byte estimate, so certificates landing in that band were accepted and written past the end. Reachable from `wolfSSL_X509_sign()` and `wolfSSL_X509_REQ_sign()`. Reproduced with plain RSA: an RSA-2048 leaf, RSA-2048 CA and 64 SANs writes 4099 bytes into the 4096 byte `WC_MAX_X509_GEN` buffer. Both call sites now ask `AddSignature()` for the exact size with a NULL buffer first, the idiom already used in `SignCrl()`.

**2. `EVP_PKEY_keygen()` on a populated `EVP_PKEY`.** The RSA branch passed `&pkey->pkey.ptr` to `wolfSSL_i2d_RSAPrivateKey()`, which treats a non-NULL `*outBuf` as a caller supplied buffer and encodes into it with no size check. Keygen on a pkey decoded from a 2048-bit public key writes about 1190 bytes into a 294 byte buffer and leaves `pkey.ptr` interior. The branch now goes through `PopulateRSAEvpPkeyDer()`, as `wolfSSL_EVP_PKEY_set1_RSA()` and the sibling EC case already do.

**3. `WOLFSSL_NO_REALLOC` populate paths.** `PopulateRSAEvpPkeyDer()` and `ECC_populate_EVP_PKEY()` allocated for the new encoding then copied `pkey_sz` bytes of the old one into it, 1192 into 294 when replacing a private key with a public one. The copy is redundant and is removed. Covered by the `opensslextra-norealloc-asan` job.

**4. `EVP_PKEY_keygen()` drops a DH key without releasing it.** The DH case assigned straight over `pkey->dh`. `EVP_PKEY_set1_DH()` takes a reference and `EVP_PKEY_assign_DH()` installs one outright, so keygen on such a pkey dropped the EVP_PKEY's only pointer to that object without releasing its reference. It now generates into a temporary and frees the previous key when the pkey owned it, matching the RSA case in the same switch.

**5. `d2i_make_pkey()` keeps the previous key's metadata, data and key object when reusing an `EVP_PKEY`.** The reuse branch replaces the key data, size and type of a caller supplied `WOLFSSL_EVP_PKEY` but leaves `pkcs8HeaderSz`, `mldsaOID`, `pkey_curve` and `save_type` describing the key the object held before. The stale `pkcs8HeaderSz` is the damaging one: `pkcs8_encode()` and `wolfssl_i_evp_pkey_get_der()` both encode from `pkey.ptr + pkcs8HeaderSz`, so decoding a PKCS#8 RSA key and then reusing that object for a traditional ECC key makes `PEM_write_bio_PKCS8PrivateKey()` report success while wrapping the ECC key with its first 26 bytes cut off, and the PEM cannot be read back. Only the `d2i_PUBKEY` and `d2i_PrivateKey_EVP` routes are affected; `d2i_PrivateKey` and `d2i_AutoPrivateKey` go through `d2i_evp_pkey()`, which allocates a fresh object and recomputes the header size. The same branch also drops the data and the key object of the previous key without releasing either: `pkey.ptr` is overwritten with a fresh allocation, and `wolfSSL_EVP_PKEY_free()` disposes only of the object matching the type currently set, so the object of a key whose type has since changed is never freed. The sequence above leaks the 1219 byte RSA encoding together with the `WOLFSSL_RSA` and its bignums, 13 allocations in all. The reuse branch now clears the metadata and disposes of the previous key, so a reused object decodes to the same state as a new one. The data is released after the new encoding has been copied in, since the caller may be decoding out of it, and the call clearing the key object is under `OPENSSL_EXTRA` because `clearEVPPkeyKeys()` is defined there while `d2i_make_pkey()` also compiles under `WOLFSSL_WPAS_SMALL`.

Also fixed: `pkey_sz` and `pkcs8HeaderSz` left describing a buffer that no longer holds an encoding on the failure returns; a stale `pkcs8HeaderSz` in the traditional private branch of `ECC_populate_EVP_PKEY()`, making export paths return a short encoding under a success return; unsigned buffer comparison; a `requestSz` bound so the signature hash cannot over-read; and `ForceZero()` on the private key DERs before they are reallocated or freed.

One commit is hardening rather than a fix. `EVP_PKEY_get1_EC_KEY()` has a branch that builds an EC_KEY when the pkey carries none, caches it and returns it without taking a second reference or setting `ownEcc`, and on load failure frees it while leaving `pkey->ecc` pointing at it; `EVP_PKEY_keygen()` claimed `ownEcc` whether or not it created the key. Both are wrong as written, but no entry point checked reaches them, since `d2i_PrivateKey()` and `d2i_PUBKEY()` both leave `pkey->ecc` populated. The commit message says so.

## Tests

`test_wc_SignCert_buffer_bounds`, `test_wc_SignCert_cb`, `test_wolfSSL_EVP_PKEY_keygen_reuse`, `test_wolfSSL_EVP_PKEY_set1_shrinking_der`, `test_wolfSSL_EVP_PKEY_set1_EC_KEY_no_pkcs8` and `test_wolfSSL_d2i_PrivateKey_reuse_resets_state`. Each was confirmed to fail against the unpatched library: restoring master's `asn.c` fails both sweeps, master's `evp.c` aborts under ASan in `sp_to_unsigned_bin_len`, dropping the `pkcs8HeaderSz` reset fails the shrinking-DER assertion, reverting the `ECC_populate_EVP_PKEY()` assignment fails on a 92 byte encoding starting mid scalar, and reverting the `d2i_make_pkey()` reset fails the `pkcs8HeaderSz == 0` assertion. The last one compares the PKCS#8 output of a reused key against a freshly decoded one, which are byte identical with the fix in place.

The reuse leak is asserted by the sanitizer jobs rather than by the test itself, since it needs an allocation tracker: any job building with AddressSanitizer and no `ASAN_OPTIONS` runs LeakSanitizer, and the same test drives the path. It was measured directly by toggling only the `d2i_make_pkey()` hunk in a static build and running the `evp_pkey` group under a leak checker: 13 leaks and 22240 bytes without it, 0 leaks with it. The 13 allocations match what LeakSanitizer reports, and the direct leak is the size of `certs/server-keyPkcs8.der`.

Two further tests do not assert their defect and are labelled as such in the source. `test_wolfSSL_EVP_PKEY_keygen_dh_reuse` drives the DH path but cannot assert the leak, which needs an allocation tracker; the smoke-test `sanitize-asan` job supplies one, since it builds with AddressSanitizer and sets no `ASAN_OPTIONS`, so LeakSanitizer runs there by default. `test_wolfSSL_EVP_PKEY_get1_EC_KEY_reuse` pins the get1 reference contract on the path a decoded pkey takes, and passes with and without the EC change.

Verified with `--enable-all --enable-certsigncb --enable-debug` under ASan (`make check` 17 passed, 6 skipped, 0 failed), an ECC-only `--disable-rsa` build, and `--enable-opensslextra -DWOLFSSL_NO_REALLOC` under ASan. All build without warnings. The last commit was additionally built under `--enable-opensslextra -DWOLFSSL_NO_REALLOC`, where its test compiles out, and under `--enable-wpas=small`, which is the build without `OPENSSL_EXTRA` that reaches `d2i_make_pkey()`.

Two guard notes, since both cost coverage when they were wrong. The bounds test gates each algorithm on its own prerequisites, because `WOLFSSL_KEY_TO_DER` and `USE_CERT_BUFFERS_2048` imply RSA and a shared guard silently removes the ECDSA and ECC halves from RSA-less builds. The ECDSA sweep asserts `BUFFER_E` or a size within the capacity plus an untouched guard region, rather than a fixed return, because the signature length varies per signing; a margin wide enough to absorb that jitter would sit outside the band entirely.

## CI

`WOLFSSL_CERT_SIGN_CB` is set only by `--enable-certsigncb`, which appears nowhere under `.github/`, so `wc_SignCert_cb()` and its test are built by no job today.


wolfcrypt/src/evp.c
```diff
@@ -9841,9 +9899,11 @@ static int ECC_populate_EVP_PKEY(WOLFSSL_EVP_PKEY* pkey, WOLFSSL_EC_KEY *key)
     else if (ecc->type == ECC_PUBLICKEY) {
```
Frauschi, line 9931: fixed

tests/api/test_asn.c
```diff
@@ -2518,6 +2518,276 @@ int test_ToTraditional_ex_mldsa_bad_params(void)
...
+     * conservative instead of correct. */
+    ExpectIntGT(bodySz = test_wc_SignCert_makeBody(&cert, &key, &rng, scratch,
+        SIGN_CERT_SCRATCH_SZ), 0);
+    if (EXPECT_SUCCESS() && (buf != NULL)) {
+        XMEMCPY(buf, scratch, (size_t)bodySz);
+        XMEMSET(buf + exactSz, SIGN_CERT_GUARD_BYTE, SIGN_CERT_GUARD_SZ);
+    }
+    ExpectIntEQ(wc_SignCert(bodySz, cert.sigType, buf, (word32)exactSz, &key,
+        NULL, &rng), exactSz);
+    for (i = 0; i < SIGN_CERT_GUARD_SZ; i++) {
```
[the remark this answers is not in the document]
Frauschi: Not reachable as written - the guard is in the macro.

`ExpectIntEQ(x, y)` expands to `ExpectInt`, which evaluates `(int)(x)` only inside `if (EXPECT_SUCCESS())` (`tests/unit.h:207-214`). If `exactSz = wc_SignCert(...)` returns <= 0, the preceding `ExpectIntGT(..., 0)` sets `_ret = TEST_FAIL`, `EXPECT_SUCCESS()` is false from that point on, and `buf[exactSz + i]` is never evaluated - the loop spins without touching the buffer. The same holds for `buf[cap + i]` in the capacity sweep, and for `buf` itself being NULL, since `ExpectNotNull(buf = XMALLOC(...))` fails first.

The explicit `if (EXPECT_SUCCESS() && (buf != NULL))` on the `XMEMSET` above is there because `XMEMSET` is a plain call with no such guard, not because the macro lacks one. Same reason the capacity sweep opens with `if (!EXPECT_SUCCESS()) break;` - it does an `XMEMCPY` and an `XMEMSET`.

Checked under ASan rather than by reading: a probe built against the real `tests/unit.h` that fails the sizing expectation and then runs both flagged loop shapes over a 16 byte heap buffer produces no report. The control - the same `buf[exactSz - 1]` read written outside the macro, on the same buffer - reports `heap-buffer-overflow`, so the clean run is not vacuous.

Closing as a false positive; no change made.

tests/api.c
```diff
@@ -25755,6 +25762,45 @@ static int test_wc_SignCert_cb(void)
...
+        /* The check must not be over-conservative either, so require the exact
+         * size to be accepted and the bytes above it left alone. */
+        ExpectIntGT(bodySz = wc_MakeCert(&cert, der, FOURK_BUF, &key, NULL,
+            &rng), 0);
+        if (EXPECT_SUCCESS()) {
+            XMEMSET(der + exactSz, 0xA5, SIGN_CERT_CB_GUARD_SZ);
+        }
+        ExpectIntEQ(wc_SignCert_cb(bodySz, cert.sigType, der,
+            (word32)exactSz, RSA_TYPE, mockSignCb, &signCtx, &rng), exactSz);
+        for (i = 0; i < SIGN_CERT_CB_GUARD_SZ; i++) {
```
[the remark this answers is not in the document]
Frauschi: Same as the `tests/api/test_asn.c` finding - not reachable, because the guard is in the macro.

`ExpectIntEQ(x, y)` expands to `ExpectInt`, which evaluates `(int)(x)` only inside `if (EXPECT_SUCCESS())` (`tests/unit.h:207-214`). If `exactSz = wc_SignCert_cb(...)` returns <= 0, the preceding `ExpectIntGT(..., 0)` sets `_ret = TEST_FAIL`, so `der[exactSz - 1 + i]` and `der[exactSz + i]` are never evaluated and the loops spin without touching `der`.

The `if (EXPECT_SUCCESS())` wrappers on the preceding `XMEMSET` calls are needed because `XMEMSET` is a plain call that carries no guard of its own; the `Expect*` macros do not need one added.

Checked under ASan: a probe built against the real `tests/unit.h`, failing the sizing expectation first, runs both flagged loop shapes clean, while the same read written outside the macro on the same buffer reports `heap-buffer-overflow`.

Closing as a false positive; no change made.

Frauschi:
Jenkins retest this please

wolfcrypt/src/evp.c
dgarske, line 9921: Does `pkey->pkcs8HeaderSz = 0;` need set/cleared here?

dgarske (changes requested):

wolfcrypt/src/evp.c
Frauschi: good catch, fixed!

wolfcrypt/src/evp_pk.c
```diff
@@ -71,13 +96,22 @@ static int d2i_make_pkey(WOLFSSL_EVP_PKEY** out, const unsigned char* mem,
...
             /* Copy in key data. */
             XMEMCPY(pkey->pkey.ptr, mem, memSz);
         }
     }
+    /* The data of the key held before is no longer referenced. */
+    if (prevData != NULL) {
+        if (prevSz > 0) {
+            ForceZero(prevData, (word32)prevSz);
+        }
+        XFREE(prevData, pkey->heap, DYNAMIC_TYPE_PUBLIC_KEY);
```
dgarske: Change XMALLOC to include the pkey->heap and fix the dynamic type on free to match the XMALLOC case `priv ? DYNAMIC_TYPE_PRIVATE_KEY : DYNAMIC_TYPE_PUBLIC_KEY`

wolfcrypt/src/evp.c
```diff
@@ -9279,10 +9303,20 @@ static int PopulateRSAEvpPkeyDer(WOLFSSL_EVP_PKEY *pkey)
         return WOLFSSL_FAILURE;
     }
 
+    /* On a private key the outgoing buffer holds a full RSA DER, so wipe it
+     * before it is reallocated or returned to the allocator. Nothing is
+     * carried over: the encoding below fills the new buffer completely. The
+     * size is dropped with the contents so a failure below cannot leave
+     * pkey_sz describing a buffer that no longer holds an encoding. */
+    if (pkey->pkey.ptr != NULL && pkey->pkey_sz > 0) {
+        ForceZero(pkey->pkey.ptr, (word32)pkey->pkey_sz);
```
dgarske: The new block `ForceZero(pkey->pkey.ptr, pkey->pkey_sz); pkey->pkey_sz = 0;` runs before the buffer is (re)allocated. In the `XREALLOC` path that ordering is forced, since realloc consumes the old pointer - the comment says as much. But in the `WOLFSSL_NO_REALLOC` path the allocation is a plain `XMALLOC` that could be done first, and on `XMALLOC` failure the function now returns `WOLFSSL_FAILURE` having already destroyed a previously valid encoding (the buffer is retained but zeroed, with `pkey_sz`/`pkcs8HeaderSz` reset). The private-key branches of `ECC_populate_EVP_PKEY()` use the opposite, allocate-first ordering and only wipe the old buffer once the new encoding has succeeded. Worth noting that a failed `wolfSSL_EVP_PKEY_set1_RSA()` on a memory-constrained target now leaves the pkey without its previous DER; the pkey is already inconsistent in that case (set1 replaces `pkey->rsa` before calling this), so this is an observation about ordering consistency rather than a defect.


Recommendation: If you want the OOM path to preserve the previous encoding under `WOLFSSL_NO_REALLOC`, move the wipe inside the `derBuf != NULL` arm; otherwise no change is needed, since the caller already treats a failed populate as leaving the pkey undefined.

dgarske (changes requested):

dgarske (approved):

PR #11065, opened 2026-08-05
Reject zero modulus in mulmod/sqrmod

### LINKED ISSUE #11063: sp_sqrmod: zero modulus rejected or accepted depending on output aliasing
### Version

master (a103f5a, 2026-08)

### Description

`mp_sqrmod()` disagrees with itself about a zero modulus. Whether it returns an error depends on where the caller points the result:

```
mp_sqrmod(a, m, r)  a=0, m=0, r distinct  ->  MP_VAL
mp_sqrmod(a, m, m)  a=0, m=0              ->  MP_OKAY, result 0
```

In `wolfcrypt/src/sp_int.c`, `sp_sqrmod()` sends the `r != m` case through `sp_sqr()` then `sp_mod()`, and `sp_mod()` rejects a zero modulus. The `r == m` case calls `_sp_sqrmod()`, which short-circuits on `sp_iszero(a)` and returns `MP_OKAY` without ever looking at `m`.

A zero modulus is undefined either way, but the two paths should agree.

Found via OSS-Fuzz 513887571 (cryptofuzz `BignumCalc SqrMod`).

### Reproduction steps

```c
#include <wolfssl/options.h>
#include <wolfssl/wolfcrypt/settings.h>
#include <wolfssl/wolfcrypt/integer.h>
#include <stdio.h>

int main(void) {
    mp_int a, m, r;
    mp_init_multi(&a, &m, &r, NULL, NULL, NULL);
    mp_set_int(&a, 0);
    mp_set_int(&m, 0);
    printf("r != m: %d\n", mp_sqrmod(&a, &m, &r));
    mp_set_int(&m, 0);
    printf("r == m: %d\n", mp_sqrmod(&a, &m, &m));
    return 0;
}
```

Built against `./configure --enable-static --disable-shared`.

### Relevant log output

r != m: -98
r == m: 0

MarkAtwood:
Fixes #11063.

`_sp_sqrmod()` and `_sp_mulmod_tmp()` short-circuit on a zero operand and store zero without checking the modulus, while the non-aliased paths reject a zero modulus through `sp_mod()`. So the return value depended on whether the caller aliased the result onto the modulus:

```
mp_sqrmod(a, m, r)  a=0, m=0, r distinct  ->  MP_VAL
mp_sqrmod(a, m, m)  a=0, m=0              ->  MP_OKAY, result 0
```

`mp_mulmod` behaves the same way through `_sp_mulmod_tmp()`. Both public wrappers already document `MP_VAL` for a zero modulus (`sp_sqrmod` and `sp_mulmod`), so the aliased path did not match its documented contract.

The check goes in the zero-operand branch of each helper rather than in the public wrappers, so the non-zero path keeps getting it from `sp_mod()` as before.

### Other backends

- `tfm.c` (`fp_mulmod`, `fp_sqrmod`) and `integer.c` (`mp_mulmod`, `mp_sqrmod`) have no zero-operand short-circuit and always reduce through `fp_div`/`mp_div`, which reject a zero divisor. Already correct in both aliasings, unchanged here.
- `hw_mulmod()` and `hw_sqrmod()` in `port/maxim/max3266x.c` carry the identical pattern, and `wolfmath.h` maps `mp_mulmod`/`mp_sqrmod` onto them when `WOLFSSL_USE_HW_MP` is defined. They get the same check. **That hunk is by inspection only** — building the port needs the Maxim SDK, so it has not been compiled or run. Happy to drop it into a separate PR if you would rather someone with the hardware take it.

### Behavior change

The only production caller reaching the changed branch is `wc_CheckRsaKey()`, at `rsa.c:927` and `rsa.c:950`, which call `mp_mulmod(&key->dP, &key->e, tmp, tmp)` with the result aliased onto the modulus. `tmp` is `p-1`/`q-1`, non-zero for any real key. For a malformed key with `p == 1` the call now returns `MP_VAL` and the key is rejected as `MP_EXPTMOD_E`, where previously it returned `MP_OKAY` with `tmp = 0` and was rejected by the following `!mp_isone(tmp)` check. Same outcome, explicit rather than incidental.

### Testing

Extends the existing aliasing test in `mp_test_mulmod_sqrmod()` to cover a zero modulus in both arrangements, including both halves of the `_sp_mulmod_tmp()` zero-operand disjunction. The test fails before the `sp_int.c` change and passes after.

Note that `mp_test()` is gated behind `WOLFSSL_PUBLIC_MP`, so a default `./configure && make check` skips the whole mp suite. Verified with `CFLAGS=-DWOLFSSL_PUBLIC_MP`. CI covers it via `cmake.yml` and `os-check-linux.json`.

Found via OSS-Fuzz 513887571 (cryptofuzz `BignumCalc SqrMod`).

### DIFF
--- a/wolfcrypt/src/port/maxim/max3266x.c
+++ b/wolfcrypt/src/port/maxim/max3266x.c
@@ -2495,6 +2495,10 @@ int hw_mulmod(mp_int* multiplier, mp_int* multiplicand, mp_int* mod,
         return MP_VAL;
     }
     if ((multiplier->used == 0) || (multiplicand->used == 0)) {
+        /* A zero modulus is invalid whatever the operands are. */
+        if (mod->used == 0) {
+            return MP_VAL;
+        }
         mp_zero(result);
         return 0;
     }
@@ -2591,6 +2595,10 @@ int hw_sqrmod(mp_int* base, mp_int* mod, mp_int* result)
         return MP_VAL;
     }
     if (base->used == 0) {
+        /* A zero modulus is invalid whatever the base is. */
+        if (mod->used == 0) {
+            return MP_VAL;
+        }
         mp_zero(result);
         return 0;
     }
--- a/wolfcrypt/src/sp_int.c
+++ b/wolfcrypt/src/sp_int.c
@@ -12292,6 +12292,7 @@ int sp_mul(const sp_int* a, const sp_int* b, sp_int* r)
  * @param [out] r  SP integer result.
  *
  * @return  MP_OKAY on success.
+ * @return  MP_VAL when m is 0.
  * @return  MP_MEM when dynamic memory allocation fails.
  */
 static int _sp_mulmod_tmp(const sp_int* a, const sp_int* b, const sp_int* m,
@@ -12300,7 +12301,15 @@ static int _sp_mulmod_tmp(const sp_int* a, const sp_int* b, const sp_int* m,
     int err = MP_OKAY;
 
     if (sp_iszero(a) || sp_iszero(b)) {
-        _sp_zero(r);
+        /* Only reached from sp_mulmod() when the result aliases the modulus.
+         * The zero-operand short-circuit would otherwise bypass the sp_mod()
+         * validation that the non-zero operand path relies on. */
+        if (sp_iszero(m)) {
+            err = MP_VAL;
+        }
+        else {
+            _sp_zero(r);
+        }
     }
     else {
         /* Create temporary for multiplication result. */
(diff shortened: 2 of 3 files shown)


SparkiDev (approved):

JacobBarthelmeh:
Retest this please Jenkins

JacobBarthelmeh (approved):

philljj:
Retest this please.

(no logs)

philljj:
Retest this please.

`TLSAbruptCloseError`

philljj (approved):

PR #11068, opened 2026-08-06
AES-GCM ASM crypto: restore tmp pointer

SparkiDev:
# Description

Assembly code for AES-GCM was not restoring tmp pointer and writing over fields of the AES structure.

# Testing

./configure --disable-shared LDFLAGS=--static --host=armv8 CC=arm-linux-gnueabihf-gcc --enable-all --disable-crl-monitor


SparkiDev:
Code generated with PR:
https://github.com/wolfSSL/scripts/pull/648

dgarske (approved):

PR #11089, opened 2026-08-06
Add native API for TLS 1.3 certificate_authorities extension

julek-wolfssl:
RFC 8446 4.2.4 `certificate_authorities` (extension type 47) was previously only accessible behind `OPENSSL_EXTRA` via the `client_ca_names` / `ca_names` / `peer_ca_names` `WOLF_STACK_OF(WOLFSSL_X509_NAME)` stacks, leaving users without the OpenSSL compat layer no way to send or inspect the extension. This adds a native API gated on `WOLFSSL_TLS13`.

- New `CertificateAuthority` node type: a singly-linked list with a flexible-array DN buffer holding the inner DER-encoded Name content (no SEQUENCE header). `TLSX_CertificateAuthorities_Add`/`FreeAll` manage the list, gated on `WOLFSSL_TLS13 && !NO_CERTS && !WOLFSSL_NO_CA_NAMES`.
- Storage added on `WOLFSSL_CTX` (`ws_ca_names`) and `WOLFSSL` (`ws_ca_names` + `ws_peer_ca_names`); SSL shadows CTX via the `WS_CA_NAMES` helper.
- New public API mirroring the UseSNI style:
  - `wolfSSL_UseCertificateAuthority` / CTX variant
  - `wolfSSL_ClearCertificateAuthorities` / CTX variant
  - `wolfSSL_GetPeerCertificateAuthorityCount`
  - `wolfSSL_GetPeerCertificateAuthority` (index-based, copy-out)

  The library prepends the DER SEQUENCE header on the wire and strips it on parse, so callers pass the raw subject content straight from `wc_GetDecodedCertSubjectRaw` (new accessor on `DecodedCert`).
- `TLSX_CA_Names_Write` is a single emitter matching the `PHA_GET_SIZE` / `PHA_WRITE` style (int return, `word16* pSz` accumulator); macros pass NULL when sizing and the output buffer when serializing. It walks the compat stack first (when `OPENSSL_EXTRA` is on) then the native list, enforcing the RFC 8446 per-DN cap and capping the whole extension payload (outer length included) at `WOLFSSL_MAX_16BIT` so the `word16` size accumulator stays exact, returning `BUFFER_ERROR` on overflow.
- `TLSX_CA_Names_Parse` always populates `ws_peer_ca_names` and, when `OPENSSL_EXTRA` is compiled in, additionally populates `peer_ca_names` through the existing `InitDecodedCert`/`GetName`/`CopyDecodedName` path. Length validation enforces the RFC 8446 `DistinguishedName<1..2^16-1>` and `authorities<3..2^16-1>` bounds.
- `TLSX_GetSize` now propagates a non-zero `ret` by breaking out of the walk, matching `TLSX_Write`.
- Teardown paths in `SSL_CtxResourceFree` and `wolfSSL_ResourceFree` free the native lists on the owning heap.

Tests added in `tests/api/test_tls_ext.c` cover argument validation, size limits, send/receive counts, handshake round-trips on SSL and CTX, a `cert_cb` scenario feeding `DecodedCert` subjects to the API (with a params loop for TLS 1.3 and DTLS 1.3), the existing `OPENSSL_EXTRA` cases, and a bad-extension regression. Documentation for all new API functions is added under `doc/dox_comments/header_files/`.


julek-wolfssl:
retest this please

philljj (changes requested):
The PRB hung because of the draft conversion. Please rebase + force-push so CI will unblock.

src/tls.c
```diff
@@ -7694,74 +7694,176 @@ int TLSX_Cookie_Use(const WOLFSSL* ssl, const byte* data, word16 len, byte* mac,
...
 /******************************************************************************/
-/* Certificate Authorities                                                       */
+/* Certificate Authorities                                                    */
 /******************************************************************************/
 
-static word16 TLSX_CA_Names_GetSize(void* data)
+/* Append a copy of dn/dnSz to the list so wire order matches call order.
+ * tail, when not NULL, caches the append point across calls: parsing a
+ * maximum sized extension adds thousands of entries and walking from the
+ * head each time would be quadratic. */
```
philljj, line 7705: This quadratic comment reads scary at first, but after looking at code paths it seems like it would rarely be realized:
- most use cases will be quick tail appends.
- when the tail is null it will be early in building the entry list, when it's small and quick to walk?

Is this correct?

Is there a better way to phrase this that the worst case scenario (n^2, with big n) won't happen unless you deliberately contrive it?

src/internal.c
philljj, line 9982: `FreeSSL_StaticMemory(ssl);` just happened here, but

```
#if !defined(NO_CERTS) && !defined(WOLFSSL_NO_CA_NAMES) && \
    defined(WOLFSSL_TLS13)
    TLSX_CertificateAuthorities_FreeAll(ssl->ws_ca_names, ssl->heap);
    ssl->ws_ca_names = NULL;
    TLSX_CertificateAuthorities_FreeAll(ssl->ws_peer_ca_names, ssl->heap);
    ssl->ws_peer_ca_names = NULL;
#endif
```

is happening down below.

Shouldn't the heap-hint frees happen first?

wolfssl/ssl.h
```diff
@@ -4756,6 +4756,45 @@ WOLFSSL_API unsigned short wolfSSL_SNI_GetRequest(WOLFSSL *ssl,
...
+/* Certificate Authorities - RFC 8446 (TLS 1.3 extension type 47).
+ *
+ * Native API, independent of the OpenSSL compatibility layer. CA DNs passed
+ * to the Use functions are the inner content of a DER-encoded Name, i.e. the
+ * bytes after the SEQUENCE tag and length; the library prepends the SEQUENCE
+ * header on the wire and strips it on parse. The bytes are copied; the caller
+ * retains ownership of the input buffer. When both the native API and the
+ * compat layer (set0_CA_list et al.) provide CAs, the two lists are
+ * concatenated on the wire. */
+#if !defined(NO_CERTS) && !defined(WOLFSSL_NO_CA_NAMES) && \
```
philljj, line 4768: My understanding is this is pulling a feature from behind OPENSSL_EXTRA and giving it broader exposure. This is cool, but embedded users might bristle at the extra 1kB - 3kB footprint.

Suggestion: wrap this in its own feature define, since you are in effect making it a cool separate feature. It could be negative define gated so users can turn it off if they want.

Feel free to disregard if not worth the define complexity for the footprint.

philljj (changes requested):
Looks cool, just a few suggestions.

src/tls.c
julek-wolfssl: The parsing side uses `TLSX_CertificateAuthorities_Add_ex` which keeps track of the tail so quadratic won't happen there. `TLSX_CertificateAuthorities_Add` is used only in `wolfSSL_UseCertificateAuthority` and `wolfSSL_CTX_UseCertificateAuthority` but that is user controlled setup. The ca_names field has a maximum length of `2^16-1` which makes the comment true. How would you prefer to change the naming?

wolfssl/ssl.h
julek-wolfssl: `WOLFSSL_NO_CA_NAMES` is already there for the users that don't want it.

julek-wolfssl:
retest this please

src/tls.c
philljj: It's probably fine

philljj (changes requested):
PRB generic config is failing:
```
FAILURES:
   1673: test_tls13_early_data_0rtt_replay
 End API Tests
 Failed/Skipped/Passed/All: 1/392/1940/2333
FAIL scripts/unit.test (exit status: 1)
```

julek-wolfssl:
retest this please

julek-wolfssl:
retest this please

julek-wolfssl:
retest this please

julek-wolfssl:
retest this please

philljj:
Retest this please.

(no logs)

julek-wolfssl:
retest this please

julek-wolfssl:
retest this please

PR #11093, opened 2026-08-06
wolfCrypt Fenrir fixes

Frauschi:
## Description

Six independent wolfCrypt fixes, one commit each, no dependency between them.

**F-7306 - PKCS#12 PBKDF mp variant, undersized buffer** (`pwdbased.c`). The `WC_PKCS12_PBKDF_USING_MP_API` variant sized the `D || S || P` working buffer as `dLen + sLen`, omitting the password block. The fill loop wrote `pLen` bytes past the end, and the hash was fed the short length, so every password over a given salt derived the same key. Sum `dLen + iLen` instead, matching the non-mp variant, which fixes the allocation, hash input length, `I` update bound and `ForceZero` coverage together.

**F-7445 - `EVP_DecodeUpdate()` heap overflow on a NUL byte** (`evp.c`). The decode loop broke out on NUL without clearing the remaining length, so the leftover-buffering block copied the full remaining count into the 48 byte `ctx->data` with an unbounded index, writing attacker-controlled data past the end of the context. Clear the length before breaking and bound the copy at one decode block, which also keeps `ctx->remaining` small enough that a reused context cannot underflow or over-read later.

**F-7438 - Renesas cryptocb slot table, unbounded index** (`renesas_common.c`). The slot index came from a free-running devId counter with no bound check, so the sixth registration in a process wrote past `gCbCtx[]`. Now searches for a free slot under the hardware lock and returns `INVALID_DEVID` when full; `Renesas_cmn_Cleanup()` releases the slot, dropping a stale pointer to a context the app may free. Two off-by-one bounds checks corrected as well.

**F-7411 - LMS key not invalidated on private key write failure** (`wc_lms.c`). The signature is computed at the current leaf before the advanced index is stored. A failed write erased the signature but left the state `OK`, so the key could sign again while storage still pointed at the consumed leaf, and a reload after restart would reuse the same LM-OTS key. Mark the key bad before signing and restore `OK` only after the write succeeds, as XMSS already does.

**F-7413 - NETOS thread stack freed while running** (`wc_port.c`, `types.h`). `wolfSSL_JoinThread()` freed the thread's stack without waiting and never unregistered from ThreadX. Waiting was impossible because `THREAD_TYPE` embedded the `TX_THREAD` by value and is passed by value, so the join saw a stale copy; it now holds a pointer to a heap-allocated control block, like the Zephyr port. The join polls `tx_thread_info_get()` for `TX_COMPLETED`/`TX_TERMINATED`, calls `tx_thread_delete()`, and frees only then. On `tx_thread_delete()` failure the block and stack are leaked deliberately, since freeing would hand ThreadX freed memory.

**F-7412 - `RSA_W_ENC` verify guard** (`signature.c`). The DigestInfo length extraction in `wc_SignatureVerifyHash()` was guarded on `WOLFSSL_RSA_PUBLIC_ONLY`, but it depends on ASN.1, not on private key support. Public-only builds still DER-encode the digest, so every valid signature was rejected with `BAD_LENGTH_E`; the same guard also broke compilation under `NO_ASN` with RSA. Guard on `NO_ASN` instead, matching the encode call site.

## Testing

- `test_wolfSSL_EVP_DecodeUpdate()`: new case feeding a NUL followed by more data than the context buffer holds, asserting `ctx->remaining == 0` (was 64 bytes into a 48 byte buffer).
- `test_wc_LmsKey_write_fail()` (new): fails the write-key callback, checks the signature is erased and that the key then refuses to sign with `BAD_STATE_E`.
- Two new `os-check-linux.json` configs: `cryptonly-no-asn-rsa` (build-only; did not compile before F-7412) and `pkcs12-pbkdf-mp-api` (the mp variant no configure option reaches).

The Renesas and NETOS changes are review-verified only - no vendor SDK, no ThreadX, and no CI job for either port.

### DIFF
--- a/wolfcrypt/src/evp.c
+++ b/wolfcrypt/src/evp.c
@@ -13925,6 +13925,8 @@ int  wolfSSL_EVP_DecodeUpdate(WOLFSSL_EVP_ENCODE_CTX* ctx,
         }
         e[0] = in[j++];
         if (e[0] == '\0') {
+            /* a NUL byte ends the input, nothing is left to buffer */
+            inLen = 0;
             break;
         }
         inLen--;
@@ -14004,6 +14006,15 @@ int  wolfSSL_EVP_DecodeUpdate(WOLFSSL_EVP_ENCODE_CTX* ctx,
             }
             if (c == '=') {
                 pad = 1;
+            }
+            if (i >= BASE64_DECODE_BLOCK_SIZE) {
+                /* More leftover data than one decode block can hold. The loops
+                 * above leave at most a partial block behind, so this is a
+                 * bound on the buffer and not a reachable input error. */
+                XMEMSET(ctx->data, 0, sizeof(ctx->data));
+                ctx->remaining = 0;
+                *outl = 0;
+                return -1;
             }
             ctx->data[i++] = c;
             ctx->remaining++;
--- a/wolfcrypt/src/port/Renesas/renesas_common.c
+++ b/wolfcrypt/src/port/Renesas/renesas_common.c
@@ -67,7 +67,6 @@
 #define INITIAL_DEVID 7890
 
 uint32_t   g_CAscm_Idx = (uint32_t)-1; /* index of CM table    */
-static int gdevId = INITIAL_DEVID;     /* initial dev Id for Crypt Callback */
 
 #ifdef WOLF_CRYPTO_CB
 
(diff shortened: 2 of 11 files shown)


wolfcrypt/src/port/Renesas/renesas_common.c
```diff
@@ -76,19 +75,61 @@ FSPSM_ST    *gCbCtx[MAX_FSPSM_CBINDEX];
...
+ * well. A slot that cannot be locked is still released, since losing it would
+ * make the devId unusable for the rest of the run.
+ *
+ * idx : index into gCbCtx of the slot to release
+ */
+static void cmn_release_cb_slot(int idx)
+{
+    int locked = (cmn_hw_lock() == 0);
+
+    gCbCtx[idx] = NULL;
```
Frauschi, line 90: fixed

wolfcrypt/src/port/Renesas/renesas_common.c
```diff
@@ -76,19 +75,62 @@ FSPSM_ST    *gCbCtx[MAX_FSPSM_CBINDEX];
...
+ *
+ * The hardware lock guards the search for a free slot, so it is taken here as
+ * well. Without the lock the slot is left in use rather than written
+ * unsynchronized, which would race with a claim in
+ * wc_CryptoCb_CryptInitRenesasCmn. A claim needs the same lock, so a slot kept
+ * this way cannot be handed out again either way.
+ *
+ * idx : index into gCbCtx of the slot to release
+ */
+static void cmn_release_cb_slot(int idx)
```
Frauschi, line 86: fixed

wolfcrypt/src/wc_lms.c
```diff
@@ -1475,6 +1480,16 @@ int wc_LmsKey_Sign(LmsKey* key, byte* sig, word32* sigSz, const byte* msg,
         ret = wc_CryptoCb_PqcStatefulSigSign(msg, (word32)msgSz, sig, sigSz,
             WC_PQC_STATEFUL_SIG_TYPE_LMS, key);
         if (ret != WC_NO_ERR_TRACE(CRYPTOCB_UNAVAILABLE)) {
+            if (ret == WC_NO_ERR_TRACE(KEY_EXHAUSTED_E)) {
+                /* Signature space exhausted. */
+                WOLFSSL_MSG("error: no LMS signatures remaining");
+                key->state = WC_LMS_STATE_NOSIGS;
+            }
+            else if (ret != 0) {
```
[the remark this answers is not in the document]
Frauschi: fixed

wolfcrypt/src/port/Renesas/renesas_common.c
```diff
@@ -507,7 +582,16 @@ void wc_CryptoCb_CleanupRenesasCmn(int* id)
...
+
+    /* Unregister before the slot is released. A new registration that claims
+     * the slot in between would otherwise be torn down here. */
     wc_CryptoCb_UnRegisterDevice(*id);
+
+    if (cbInfo != NULL)
+        cbInfo->devId = INVALID_DEVID;
+
+    /* release the slot so that it can be used by a new registration */
+    cmn_release_cb_slot(idx);
```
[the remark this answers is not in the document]
Frauschi: fixed

wolfcrypt/src/evp.c
```diff
@@ -13881,6 +13883,15 @@ int  wolfSSL_EVP_DecodeUpdate(WOLFSSL_EVP_ENCODE_CTX* ctx,
             if (c == '=') {
                 pad = 1;
             }
+            if (i >= BASE64_DECODE_BLOCK_SIZE) {
+                /* More leftover data than one decode block can hold. The loops
+                 * above leave at most a partial block behind, so this is a
+                 * bound on the buffer and not a reachable input error. */
+                XMEMSET(ctx->data, 0, sizeof(ctx->data));
```
dgarske: Should this be ForceZero since it might hold key material? Actually was this already memset above, so it this needed?

wolfcrypt/src/port/Renesas/renesas_common.c
```diff
@@ -415,6 +460,9 @@ int wc_CryptoCb_CryptInitRenesasCmn(struct WOLFSSL* ssl, void* ctx)
 
     FSPSM_ST* cbInfo = (FSPSM_ST*)ctx;
     size_t internal_sz = sizeof(FSPSM_ST_Internal);
+    int idx = 0;
```
dgarske, line 460: While you are here... can you move the (void)ssl; and (void)ctx; below the variables?

dgarske (changes requested):

wolfcrypt/src/evp.c
Frauschi: Switched to `ForceZero` — the input can be base64 of a PEM private key, so it may hold key material.

Not redundant with the `XMEMSET` above: that one runs before the loop, which then writes up to 3 bytes into `ctx->data` before this bound can trip.

Worth noting the bound itself is unreachable by construction — `Base64_SkipNewline()` only returns `BUFFER_E` when `inLen == 0`, so every `break` out of the quad loop leaves `inLen == 0`, the loop's own exit leaves `inLen <= 3`, and the leftover loop decrements `inLen` each iteration, so `i` tops out at 3. The actual overflow is fixed by the `inLen = 0` line on the NUL break; this is just asserting the invariant.

wolfcrypt/src/port/Renesas/renesas_common.c
Frauschi: fixed

dgarske (approved):

dgarske (changes requested):
Please resolve merge conflicts. Sorry

dgarske (approved):

PR #11094, opened 2026-08-06
TLS Fenrir fixes

Frauschi:
Three independent Fenrir fixes, each with a regression test that fails without its source change.

### 1. Honor the per-context X509_STORE_CTX verify callback (F-7341)

`wolfSSL_X509_STORE_CTX_set_verify_cb()` stored the callback in `ctx->verify_cb`, but every verification site read `ctx->store->verify_cb`, so the field was never consulted - a restrictive callback installed on the store context was silently ignored, and `wolfSSL_X509_verify_cert()` could succeed on a chain the callback would have rejected.

New `X509StoreGetVerifyCb()` prefers the context callback and falls back to the store's, matching OpenSSL precedence; it is used at all four call sites. `wolfSSL_X509_STORE_CTX_init()` now clears `ctx->verify_cb` so a reused context carries no stale callback.

Two notes: the pathLen and `INVALID_CA` overrides become reachable in `OPENSSL_EXTRA` builds, not just `OPENSSL_ALL`/`WOLFSSL_QT`, since the setter is guarded that way - this is the documented OpenSSL semantic. Rejection (callback returns 0) is honored only for the per-context callback; widening the store callback would be a separate behavior change.

### 2. Fix double free in `wolfSSL_X509_EXTENSION_set_data()` (F-7340)

The function freed the extension's dynamic ASN.1 string buffer but left `value.data` and `value.isDynamic` pointing at it. The following `wolfSSL_ASN1_STRING_copy()` snapshots those fields and frees the old buffer itself once the copy completes, so the stale pointer was freed twice. Any second call on a value of at least `CTC_NAME_SIZE` bytes hit this; passing the extension its own value also read freed memory. `wolfSSL_ASN1_STRING_copy()` is alias-safe, so the manual free is dropped.

### 3. Send a protocol_version alert when no version can be negotiated (F-7568)

`DoClientHello()` has four exits that fail with `VERSION_ERROR` when runtime version restrictions leave nothing acceptable at or below the offered version. Three sent no alert, the fourth only under `WOLFSSL_EXTRA_ALERTS`, so a default build just dropped the connection - the client could not tell a version mismatch from a network failure. RFC 5246 7.2.2 and RFC 8446 4.1.3 require a fatal `protocol_version` alert.

The generic fallback did not help: `SendFatalAlertOnly()` is a no-op without `WOLFSSL_EXTRA_ALERTS`, and where defined it grouped `VERSION_ERROR` with `MATCH_SUITE_ERROR` and sent `handshake_failure`, disagreeing with the TLS 1.3 mapping. All four branches now alert regardless of `WOLFSSL_EXTRA_ALERTS` (propagating `SOCKET_ERROR_E`), and `VERSION_ERROR` gets its own case in `SendFatalAlertOnly()`. Only observable on the TLS 1.2 path; TLS 1.3 already translates the error correctly.

### Tests

| Test | Group | Covers |
|---|---|---|
| `test_wolfSSL_X509_STORE_CTX_verify_cb` | `ossl_x509_store` | Context callback is consulted, wins over the store's, and is cleared by `X509_STORE_CTX_init()` |
| `test_wolfSSL_X509_EXTENSION_set_data` | `ossl_x509_ext` | Replacing a dynamic value twice, and setting a value from itself |
| `test_tls_no_acceptable_version_alert` | `tls` | Fatal `protocol_version` alert on the wire (no `WOLFSSL_EXTRA_ALERTS`) |
| `test_tls_version_error_alert_mapping` | `tls` | `SendFatalAlertOnly()` maps `VERSION_ERROR` to `protocol_version` (with `WOLFSSL_EXTRA_ALERTS`) |

The two TLS tests are complementary: with `WOLFSSL_EXTRA_ALERTS` the fallback emits a byte-identical alert, so the first would pass either way.

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -29386,6 +29386,9 @@ static const char* wolfSSL_ERR_reason_error_string_OpenSSL(unsigned long e)
     /* TODO: -WOLFSSL_X509_V_ERR_CERT_SIGNATURE_FAILURE. Conflicts with
      *       -WOLFSSL_ERROR_WANT_CONNECT.
      */
+    case WOLFSSL_X509_V_ERR_UNSPECIFIED:
+        return "unspecified certificate verification error";
+
     case WOLFSSL_X509_V_ERR_CRL_HAS_EXPIRED:
         return "CRL has expired";
 
--- a/src/internal.c
+++ b/src/internal.c
@@ -20523,8 +20523,11 @@ int SendFatalAlertOnly(WOLFSSL *ssl, int error)
     case WC_NO_ERR_TRACE(ECC_OUT_OF_RANGE_E):
         why = bad_record_mac;
         break;
-    case WC_NO_ERR_TRACE(MATCH_SUITE_ERROR):
     case WC_NO_ERR_TRACE(VERSION_ERROR):
+        why = wolfssl_alert_protocol_version;
+        break;
+    /* listed for symmetry with TranslateErrorToAlert(); default covers it */
+    case WC_NO_ERR_TRACE(MATCH_SUITE_ERROR):
     default:
         why = handshake_failure;
         break;
@@ -40681,6 +40684,7 @@ static int AddPSKtoPreMasterSecret(WOLFSSL* ssl)
         int             ret = 0;
         byte            lesserVersion;
         byte            maxMinor;
+        byte            preMaskMinor;
 
         WOLFSSL_START(WC_FUNC_CLIENT_HELLO_DO);
         WOLFSSL_ENTER("DoClientHello");
@@ -40762,9 +40766,19 @@ static int AddPSKtoPreMasterSecret(WOLFSSL* ssl)
             word16 havePSK = 0;
             int    keySz   = 0;
 
+            /* RFC 5246 7.2.2 and RFC 8446 6.2: a version that cannot be
+             * negotiated must be refused with a fatal protocol_version alert.
+             * SendFatalAlertOnly() in ProcessReply is compiled out unless
+             * WOLFSSL_EXTRA_ALERTS is defined, so alert here. */
             if (!ssl->options.downgrade) {
                 WOLFSSL_MSG("Client trying to connect with lesser version");
                 ret = VERSION_ERROR;
+                /* propagate socket errors to avoid re-calling send alert */
+                if (SendAlert(ssl, alert_fatal,
+                        wolfssl_alert_protocol_version)
+                        == WC_NO_ERR_TRACE(SOCKET_ERROR_E)) {
+                    ret = SOCKET_ERROR_E;
+                }
                 goto out;
             }
 
@@ -40778,6 +40792,12 @@ static int AddPSKtoPreMasterSecret(WOLFSSL* ssl)
             if (belowMinDowngrade) {
                 WOLFSSL_MSG("\tversion below minimum allowed, fatal error");
                 ret = VERSION_ERROR;
+                /* propagate socket errors to avoid re-calling send alert */
+                if (SendAlert(ssl, alert_fatal,
+                        wolfssl_alert_protocol_version)
+                        == WC_NO_ERR_TRACE(SOCKET_ERROR_E)) {
+                    ret = SOCKET_ERROR_E;
+                }
                 goto out;
             }
 
@@ -40835,6 +40855,13 @@ static int AddPSKtoPreMasterSecret(WOLFSSL* ssl)
                        TRUE, TRUE, TRUE, TRUE, ssl->options.side);
         }
 
+        /* Version the record layer is on before the mask walk below steps it
+         * down. It is one the client offered (or the server's own maximum
+         * when the client offered more), so the two alert exits inside that
+         * block restore it rather than alerting with the masked-off version
+         * the walk stopped at, which the peer may reject outright. */
+        preMaskMinor = ssl->version.minor;
+
         /* check if option is set to not allow the current version
          * set from either wolfSSL_set_options or wolfSSL_CTX_set_options */
         if (!ssl->options.dtls && ssl->options.downgrade &&
@@ -40874,15 +40901,28 @@ static int AddPSKtoPreMasterSecret(WOLFSSL* ssl)
                 WOLFSSL_OP_NO_SSLv3) {
                 WOLFSSL_MSG("\tError, option set to not allow SSLv3");
                 ret = VERSION_ERROR;
-#ifdef WOLFSSL_EXTRA_ALERTS
-                SendAlert(ssl, alert_fatal, wolfssl_alert_protocol_version);
-#endif
+                /* alert with a version the client will accept */
+                ssl->version.minor = preMaskMinor;
+                /* propagate socket errors to avoid re-calling send alert */
+                if (SendAlert(ssl, alert_fatal,
+                        wolfssl_alert_protocol_version)
+                        == WC_NO_ERR_TRACE(SOCKET_ERROR_E)) {
+                    ret = SOCKET_ERROR_E;
+                }
                 goto out;
             }
 
             if (ssl->version.minor < ssl->options.minDowngrade) {
                 WOLFSSL_MSG("\tversion below minimum allowed, fatal error");
                 ret = VERSION_ERROR;
+                /* alert with a version the client will accept */
+                ssl->version.minor = preMaskMinor;
+                /* propagate socket errors to avoid re-calling send alert */
+                if (SendAlert(ssl, alert_fatal,
+                        wolfssl_alert_protocol_version)
+                        == WC_NO_ERR_TRACE(SOCKET_ERROR_E)) {
+                    ret = SOCKET_ERROR_E;
+                }
                 goto out;
             }
 
--- a/src/x509.c
+++ b/src/x509.c
@@ -3776,16 +3776,14 @@ WOLFSSL_ASN1_STRING* wolfSSL_X509_EXTENSION_get_data(
 int wolfSSL_X509_EXTENSION_set_data(WOLFSSL_X509_EXTENSION* ext,
         WOLFSSL_ASN1_STRING* data)
 {
-    WOLFSSL_ASN1_STRING* current;
-
     if (ext == NULL || data == NULL)
         return WOLFSSL_FAILURE;
 
-    current = wolfSSL_X509_EXTENSION_get_data_internal(ext);
-    if (current->length > 0 && current->data != NULL && current->isDynamic) {
-        XFREE(current->data, NULL, DYNAMIC_TYPE_OPENSSL);
-    }
-
+    /* wolfSSL_ASN1_STRING_copy() defers to wolfSSL_ASN1_STRING_set(), which
+     * owns the free of any existing dynamic buffer and only releases it once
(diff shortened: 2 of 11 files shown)


src/x509_str.c
```diff
@@ -540,11 +568,18 @@ static int X509StoreVerifyCert(WOLFSSL_X509_STORE_CTX* ctx)
...
+        if (verifyCb != NULL) {
+            if (verifyCb(ret >= 0 ? 1 : 0, ctx) == 1) {
+                ret = WOLFSSL_SUCCESS;
+            }
+            else if (ret >= 0 && ctx->verify_cb != NULL) {
+                /* Returning 0 must reject a chain the cert manager accepted.
+                 * Only for the per-context callback - a store callback has
+                 * never been able to reject here, and widening it is a
+                 * separate behavior change. */
+                ret = WOLFSSL_FATAL_ERROR;
```
Frauschi: fixed

tests/api/test_ossl_x509_str.c
```diff
@@ -1140,6 +1140,130 @@ int test_wolfSSL_X509_verify_cert_pathlen_override(void)
...
+}
+#endif
+#endif
+
+/* A callback installed with X509_STORE_CTX_set_verify_cb must be honored by
+ * X509_verify_cert().  The chain verifies cleanly on its own, so a rejecting
+ * callback is the only thing that can fail it.  Also pins that the
+ * per-context callback wins over the store's, and that
+ * X509_STORE_CTX_init() clears it. */
+int test_wolfSSL_X509_STORE_CTX_verify_cb(void)
```
Frauschi, line 1289: fixed

src/x509_str.c
```diff
@@ -540,14 +571,35 @@ static int X509StoreVerifyCert(WOLFSSL_X509_STORE_CTX* ctx)
...
+                 * Only for the per-context callback - a store callback has
+                 * never been able to reject here, and widening it is a
+                 * separate behavior change. */
+                if (ctx->error == WOLFSSL_X509_V_OK) {
+                    /* Keep an error the callback recorded itself; otherwise
+                     * the rejection has no error to report. */
+                    wolfSSL_X509_STORE_CTX_set_error(ctx,
+                        WOLFSSL_X509_V_ERR_UNSPECIFIED);
+                }
+                ret = WOLFSSL_FATAL_ERROR;
```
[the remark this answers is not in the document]
Frauschi: fixed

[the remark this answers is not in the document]
Frauschi: fixed

wolfssl/ssl.h
```diff
@@ -2797,6 +2797,7 @@ WOLFSSL_API long wolfSSL_get_default_read_buffer_len(const WOLFSSL* ssl);
      */
 enum {
     WOLFSSL_X509_V_OK                                    = 0,
+    WOLFSSL_X509_V_ERR_UNSPECIFIED                       = 1,
```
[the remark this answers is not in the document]
Frauschi: fixed

src/internal.c
```diff
@@ -40751,9 +40754,19 @@ static int AddPSKtoPreMasterSecret(WOLFSSL* ssl)
...
 
+            /* RFC 5246 7.2.2 and RFC 8446 6.2: a version that cannot be
+             * negotiated must be refused with a fatal protocol_version alert.
+             * SendFatalAlertOnly() in ProcessReply is compiled out unless
+             * WOLFSSL_EXTRA_ALERTS is defined, so alert here. */
             if (!ssl->options.downgrade) {
                 WOLFSSL_MSG("Client trying to connect with lesser version");
                 ret = VERSION_ERROR;
+                /* propagate socket errors to avoid re-calling send alert */
+                if (SendAlert(ssl, alert_fatal,
```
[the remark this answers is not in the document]
Frauschi: fixed

src/x509_str.c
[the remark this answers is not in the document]
Frauschi: fixed

src/internal.c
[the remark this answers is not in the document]
Frauschi: fixed

src/internal.c
```diff
@@ -40751,9 +40757,19 @@ static int AddPSKtoPreMasterSecret(WOLFSSL* ssl)
...
```
dgarske, line 40780: This can produce a duplicate fatal alert. 
Recommendation: Add the `ssl->alert_history.last_tx.level != alert_fatal` guard — preferably at `src/tls13.c:14896`, since that fixes the whole class of problem for every error DoClientHello alerts on directly (it already double-sends for `decode_error`, `illegal_parameter` and `inappropriate_fallback` via the same route); alternatively guard the four new `SendAlert()` calls in `DoClientHello`.

wolfssl/ssl.h
dgarske, line 2801: Consider fixing `GetX509Error()` -> 'case 1: return 0' success mapping

src/x509_str.c
```diff
@@ -1195,6 +1267,15 @@ int wolfSSL_X509_verify_cert(WOLFSSL_X509_STORE_CTX* ctx)
         }
     }
 
+    /* A failure has to be reportable through X509_STORE_CTX_get_error(). A
+     * verify callback that rejects a chain the verification accepted leaves
+     * ctx->error at X509_V_OK unless it records one itself, which would tell
+     * the application the chain was fine. Fill in the generic error OpenSSL
+     * uses for this, without touching an error the callback did set. */
+    if (ret != WOLFSSL_SUCCESS && ctx->error == WOLFSSL_X509_V_OK) {
+        ctx->error = WOLFSSL_X509_V_ERR_UNSPECIFIED;
```
dgarske, line 1298: Blanket X509_V_ERR_UNSPECIFIED fallback applies to every failure path, including allocation failures. 
Recommendation: Decide which behavior is intended. If the blanket fallback is deliberate, widen the comment so it does not read as callback-specific. If not, record a specific error (OpenSSL's `X509_V_ERR_OUT_OF_MEM`) at the allocation-failure `goto exit` sites and keep the fallback scoped to the veto case.

src/x509_str.c
```diff
@@ -540,14 +576,32 @@ static int X509StoreVerifyCert(WOLFSSL_X509_STORE_CTX* ctx)
         }
 #endif
         SetupStoreCtxError(ctx, ret);
-    #if defined(OPENSSL_ALL) || defined(WOLFSSL_QT)
-        if (ctx->store->verify_cb)
-            ret = ctx->store->verify_cb(ret >= 0 ? 1 : 0, ctx) == 1 ?
-                                                        WOLFSSL_SUCCESS : ret;
-    #endif
+        if (verifyCb != NULL) {
+            if (verifyCb(ret >= 0 ? 1 : 0, ctx) == 1) {
```
dgarske, line 597: Recommendation: Snapshot `ctx->error` before calling the callback and compare afterwards, so only an error the callback genuinely recorded is preserved.

src/x509_str.c
```diff
@@ -508,11 +509,46 @@ static int X509StoreCheckCtxCrls(WOLFSSL_X509_STORE_CTX* ctx)
...
+    return NULL;
+}
+
+/* X509StoreVerifyCert() result telling the caller that the application's
+ * per-context verify callback explicitly rejected a certificate the
+ * verification itself accepted. Kept distinct from an ordinary failure
+ * (WOLFSSL_FAILURE / WOLFSSL_FATAL_ERROR) because chain building must stop on
+ * it: a veto must not be turned into a retry with another issuer, and must not
+ * be cleared by the partial-chain fallback in wolfSSL_X509_verify_cert(). */
+#define X509_STORE_CB_REJECTED (-2)
```
dgarske: Recommendation: Pick a value that cannot collide with `wolfSSL_ErrorCodes`/`WOLFSSL_CBIO_ERR_*`, or return the veto through an out-parameter instead of in band. Either way, move the `#define` up with the file's other constants.

src/x509.c
```diff
@@ -3776,16 +3776,11 @@ WOLFSSL_ASN1_STRING* wolfSSL_X509_EXTENSION_get_data(
...
-
     if (ext == NULL || data == NULL)
         return WOLFSSL_FAILURE;
 
-    current = wolfSSL_X509_EXTENSION_get_data_internal(ext);
-    if (current->length > 0 && current->data != NULL && current->isDynamic) {
-        XFREE(current->data, NULL, DYNAMIC_TYPE_OPENSSL);
-    }
-
+    /* wolfSSL_ASN1_STRING_copy() frees any existing dynamic buffer itself,
```
dgarske: Recommendation: Reword to name `wolfSSL_ASN1_STRING_set()` as the owner of the free.

src/internal.c
```diff
@@ -40863,15 +40885,24 @@ static int AddPSKtoPreMasterSecret(WOLFSSL* ssl)
...
+                    ret = SOCKET_ERROR_E;
+                }
                 goto out;
             }
 
             if (ssl->version.minor < ssl->options.minDowngrade) {
                 WOLFSSL_MSG("\tversion below minimum allowed, fatal error");
                 ret = VERSION_ERROR;
+                /* propagate socket errors to avoid re-calling send alert */
+                if (SendAlert(ssl, alert_fatal,
```
dgarske, line 40924: Recommendation: Consider restoring `ssl->version` to the client's offered version (or the server's own maximum) around the `SendAlert()` call at exits 3 and 4, so the alert record carries a version the peer will accept. At minimum, note the intent in a comment at the source rather than only in the test.

dgarske (changes requested):

src/internal.c
Frauschi: Done, in tls13.c as you suggested — `DoTls13HandShakeMsgType` skips the generic alert once a fatal one went out. New test asserts 7 bytes on the wire; without the guard it's 14.

wolfssl/ssl.h
Frauschi: Left as-is with a comment: it takes internal codes where `1` is `WOLFSSL_SUCCESS`, never an `X509_V_*` value, so the two can't meet.

src/x509_str.c
Frauschi: Deliberate — widened the comment. It's a fail-closed backstop for veto, OOM and chain-building errors alike.

src/x509_str.c
Frauschi: Fixed. Good catch — `SetupStoreCtxError()` preserves the worst error across the walk, so the old check kept stale ones. Now snapshots and compares.

src/x509_str.c
Frauschi: Went with the out-parameter. `#define` is gone, nothing left to collide.

src/x509.c
Frauschi: Reworded to name `wolfSSL_ASN1_STRING_set()`.

src/internal.c
Frauschi: Done via `preMaskMinor`. The existing test can't see it (`set_options` already lowers the version), so `test_tls_version_mask_alert_record` covers it — reverting turns `0303` into `0302`.

dgarske (dismissed):

dgarske (approved):

PR #11095, opened 2026-08-06
(d)tls13: implement session export feature

rizlik:
# DTLS 1.3: session export and import

 wolfSSL_dtls_export()/wolfSSL_dtls_import() now work for DTLS 1.3.
 Export version 7 appends the DTLS 1.3 record layer state and the
 Connection ID to the blob; version 6 blobs still import.

 - Export is refused while the connection can't be fully described:
   KeyUpdate in progress, half sent fragmented message, or work parked in
   a write dup. State only export/import is refused for DTLS 1.3

 CI/tests: session export job gains --enable-dtlscid; new API tests
 cover DTLS 1.3 export/import, v6 compatibility and the refusal paths.


doc/dox_comments/header_files/ssl.h
```diff
@@ -826,6 +826,10 @@ int wolfSSL_dtls_set_export(WOLFSSL* ssl, wc_dtls_export func);
     choice over when the session is serialized. If buffer is NULL when
     passed to function then sz will be set to the size of buffer needed
     for serializing the WOLFSSL session.
+    WARNING: buf holds the key material the connection continues with. For
+             DTLS 1.3 that includes the traffic secrets every later epoch key
+             is derived from, so it stays sensitive for the rest of the
+             connection: encrypt it if stored and wipe the buffer when done.
```
julek-wolfssl: Isn't this true for all DTLS versions?

src/dtls.c
```diff
@@ -1545,6 +1545,109 @@ const unsigned char* wolfSSL_dtls_cid_parse(const unsigned char* msg,
...
+        idx += rxSz;
+    }
+
+    exp[idx++] = txSz;
+    if (txSz > 0) {
+        XMEMCPY(exp + idx, info->tx->id, txSz);
+        idx += txSz;
+    }
+
+    return (int)idx;
```
julek-wolfssl, line 1630: Should we assert `info->negotiated` when `rxSz || txSz`?

src/dtls.c
```diff
@@ -1545,6 +1545,109 @@ const unsigned char* wolfSSL_dtls_cid_parse(const unsigned char* msg,
...
+    negotiated = exp[idx++];
+    rxSz = exp[idx++];
+    if (rxSz > DTLS_CID_MAX_SIZE || idx + rxSz + OPAQUE8_LEN > len)
+        return BUFFER_E;
+    idx += rxSz;
+
+    /* the tx id is the peer's choice, bounded only by its length field */
+    txSz = exp[idx++];
+    if (idx + txSz > len)
+        return BUFFER_E;
```
julek-wolfssl: Seperate out the `OPAQUE8_LEN` for the `txSz`. Makes it less confusing.

src/dtls.c
```diff
@@ -1545,6 +1545,109 @@ const unsigned char* wolfSSL_dtls_cid_parse(const unsigned char* msg,
...
+    }
+    if (txSz > 0) {
+        info->tx = DtlsCidNew(exp + idx, txSz, ssl->heap);
+        if (info->tx == NULL)
+            return MEMORY_ERROR;
+    }
+    idx += txSz;
+
+    info->negotiated = negotiated ? 1 : 0;
+    ssl->options.useDtlsCID = 1;
```
julek-wolfssl: Why unconditional `ssl->options.useDtlsCID = 1` when CID could not be negotiated?

src/dtls.c
```diff
@@ -1545,6 +1545,109 @@ const unsigned char* wolfSSL_dtls_cid_parse(const unsigned char* msg,
...
+            return MEMORY_ERROR;
+    }
+    if (txSz > 0) {
+        info->tx = DtlsCidNew(exp + idx, txSz, ssl->heap);
+        if (info->tx == NULL)
+            return MEMORY_ERROR;
+    }
+    idx += txSz;
+
+    info->negotiated = negotiated ? 1 : 0;
```
julek-wolfssl: Should `negotiated` be asserted to match rx/tx cid presence?

src/dtls13.c
```diff
@@ -2691,6 +2691,522 @@ int Dtls13SetRecordNumberKeys(WOLFSSL* ssl, enum encrypt_side side)
...
+
+    return (int)(WOLFSSL_EXPORT_LEN + fieldSz);
+}
+
+/* Serialize the DTLS 1.3 record layer state: the current, peer and invalidate
+ * before epoch numbers, the traffic secrets and three length prefixed epoch
+ * fields - the sending epoch, the peer epoch and the previous peer epoch - a
+ * field being empty when the connection does not have that epoch.
+ * Returns the number of bytes written to 'exp' or a negative value. */
+int ExportDtls13State(WOLFSSL* ssl, byte* exp, word32 len)
```
julek-wolfssl, line 2969: This and `ImportDtls13State` are awfully complicated. Can't we just dump the full `Dtls13Epoch dtls13Epochs[DTLS13_EPOCH_SIZE];` and call it a day? `struct Dtls13Epoch` is static (no pointers) so this whole section would boil down to a size check and a single `XMEMCPY`. I prefer the simple solution.

julek-wolfssl (changes requested):

src/dtls.c
rizlik: `rxSz` can be `> 0` if `wolfSSL_dtls_cid_set` is invoked and later the extension is not negotiated.
So I don't think we should assert, but I agree to not serialize in that case the rx CID, so I'll change this so that we gate the serialization on `info->negotiated`.

julek-wolfssl:
# Review — wolfSSL/wolfssl PR #11095 "(d)tls13: implement session export feature"

**HEAD reviewed:** `a5e11d023` (commits `a09ec82b9` tls13 refactor, `a5e11d023` dtls13 export/import). Four independent review passes: correctness/regressions, security of the deserialization path, tests/CI validation, and maintainer-comment reconciliation/simplicity. The PR is currently `CHANGES_REQUESTED`; the six inline comments dated 2026-09-01 postdate the last commit and are all unaddressed in HEAD.

**Overall verdict: OK with notes** — no P0, but six P1 items below should be fixed before release, and the maintainer review needs responses regardless.

## What checks out (evidence)

- **Refactor commit `a09ec82b9` is behavior-preserving.** `Tls13DeriveRecordKeys()` (`src/tls13.c:1522-1605`) is a verbatim move of the deleted block: same labels/secret arguments, same `i` packing incl. the deliberate "don't advance i after Server IV", `Dtls13DeriveSnKeys()` still only on `ssl->options.dtls` after a successful `StoreKeys`, `ForceZero`/`wc_MemZero_Check` pairing intact. Only observable delta: a `MEMORY_E` that was never reachable on the no-store path disappears.
- **Round-trip semantics.** Every consumed `Dtls13Epoch` field (`epochNumber`, `nextSeqNumber`, `nextPeerSeqNumber`, `dropCount`, `window`) maps to its runtime consumer (`src/dtls13.c:2377-2400, 2815`); re-deriving epoch keys from the traffic secrets is claimed only for epochs whose secret is current, and the one non-re-derivable set (previous peer epoch) is the only key material serialized (`src/dtls13.c:2697-2731`). Provision/side XOR mapping verified against `Dtls13EpochCopyKeys`.
- **Format constants match what the code writes** (`wolfssl/internal.h:3029-3105` vs `src/internal.c:913-945`); new `DTLS_EXPORT_KEY_SZ` is strictly larger than the old value, so v6 blobs still pass the length gate. No off-by-one found in the length prefixes or window `wordAdj` handling.
- **Bounds arithmetic in the new parsers is sound.** Every blob length is checked against remaining size before use; key material writes are double-bounded (`keySz == specs.key_size && <= MAX_SYM_KEY_SIZE`, `src/dtls13.c:2779-2804`), which is what stops the unvalidated word16 `specs.key_size`/`iv_size` from `ImportCipherSpecState` (`src/internal.c:1082-1083`) from overwriting `Keys::*_write_key`/`*_write_IV`. Section length prefixes cannot desync the reader (`src/internal.c:2172-2247`).
- **v6 / existing users preserved.** v6 DTLS 1.2 and TLS blobs import unchanged; the new sections are gated on DTLS version so a v6 blob never reaches them; state-only export stays stamped v6 (`src/internal.c:1836-1838`). Replay-window size mismatch fails safe: absent words marked received (`src/dtls13.c:2889-2892`), never widening the accept set.
- **Tests assert real semantics, not just `import() > 0`.** Round-trip (`tests/api/test_dtls.c:7418-7540`) checks epoch equality, non-reset sequence numbers, bidirectional app data, a post-import KeyUpdate, and second-generation export; replay (`:8023-8118`) requires a captured datagram to be discarded after import; negative tests assert specific codes (`BUFFER_E`/`VERSION_ERROR`/`BAD_STATE_E`) across 9 mutation classes (`:8464-8570`). No leak paths found (no early returns; unconditional cleanup); header/body registration is consistent (no guard mismatches in `test_dtls.h`).

## P1 — fix before release

1. **Imported epoch values are unvalidated.** `src/dtls13.c:3101-3110` stores `dtls13Epoch`, `dtls13PeerEpoch`, `dtls13InvalidateBefore` with only a `>= DTLS13_EPOCH_TRAFFIC0` floor. No `DTLS13_EPOCH_MAX` check, though the same file enforces it at `:3315` and `src/tls13.c:13578`; no relation between `dtls13InvalidateBefore` and the peer epoch; `dropCount` imported unbounded (`src/dtls13.c:2873-2876`) feeds the AEAD-failure backoff at `src/internal.c:3864-3871`. A well-formed blob can produce a session that can never rotate keys, is permanently deaf, or has its decryption-failure limit pre-loaded. Fix: reject unless both epochs ≤ `DTLS13_EPOCH_MAX` and `invalidateBefore ≤ peerEpoch`; reset/clamp `dropCount` instead of trusting it.
2. **Export during post-handshake authentication is permitted but not representable.** The guards in `src/dtls13.c:2979-3010` cover KeyUpdate, half-sent fragments, and parked dup work, but not PHA: a server that sent a CertificateRequest (`src/tls13.c:8841-8843` supports it on DTLS) or a client owing a response holds `ssl->postHandshakeAuthPending`/`certReqCtx`/`options.sendVerify` — none serialized, nothing refused. The imported side silently drops the auth round. Fix: extend the refusal to those flags, mirroring the parked-work check (incl. `dupWrite->postHandshakeAuthPending`).
3. **CID `negotiated` invariant (maintainer comments `#2/#4/#5`, VALID).** `DtlsCidImport` (`src/dtls.c:1586-1638`) accepts `negotiated=0` with non-empty rx/tx and installs it; since `wolfSSL_dtls_cid_is_enabled()` keys off extension presence (`src/dtls.c:1423-1426`), the reconstituted connection then rejects every CID-less peer record (`src/dtls13.c:1226-1236`) — one blob constructs a dead connection. And `useDtlsCID = 1` at `src/dtls.c:1636` is unconditional. The sound invariant is one-way, `(rxSz || txSz) ⇒ negotiated`: zero-length CIDs are legal per RFC 9146 and give `negotiated=1` with NULL ids (`src/dtls.c:1352-1356, 1446-1448`), so full equivalence is wrong. Fix: import rejects `!negotiated && (rxSz||txSz)`; export asserts the same.
4. **Two refusal paths ship untested in CI.** `test_dtls13_export_half_sent_fragment` and `test_dtls13_export_write_dup_pending` are gated on `WOLFSSL_DTLS_MTU`/`HAVE_WRITE_DUP` (`tests/api/test_dtls.c:7696, 7742`); the only session-export job (`sessionexport-dtls13`, `.github/configs/os-check-linux.json:103-106`) enables neither, so the gated-out bodies still return `EXPECT_RESULT()` and CI is green while the subtlest guards in the PR are never exercised. Fix: add `--enable-dtls-mtu` and `--enable-writedup` to that job.
5. **Import accepts a tx CID the send path refuses.** `txSz` is bounded only by 255 (`DTLS_CID_MAX_TX_SIZE`), while both in-tree writers of the same field cap at `DTLS_CID_MAX_SIZE` (`src/dtls.c:1451, 1471`) and the DTLS 1.2 record layer errors `DTLS_CID_ERROR` above it (`src/tls.c:1349-1354`, `src/internal.c:22192-22198`). A v7 DTLS 1.2 blob with an 11–255-byte tx CID imports "successfully" into a session that cannot send. Fix: cap `txSz` at `DTLS_CID_MAX_SIZE` on import while v7 is unreleased.
6. **Write-dup parked-work check is incomplete.** `src/dtls13.c:3002` tests only `keyUpdateRespond || sendAcks`; `WriteDup` also parks `keyUpdateWaiting` (set at `src/dtls13.c:2157`) and `keyUpdateAcked` (`:3642`). Exporting the read side with either set passes the guard, and the imported connection never completes the announced KeyUpdate — contradicting the guard's own rationale. Fix: include both bits.

## P2 — report-only / hygiene

- **Partial-size secret copy leaves stale bytes**: `src/dtls13.c:3118-3119` `XMEMCPY`s `secretSz` (< `SECRET_LEN` when hashes differ) into `clientSecret`/`serverSecret` without zeroizing the fields first. One-line `ForceZero` before the copy.
- **Failed CID import leaves the extension attached with NULL ids** — `TLSX_ConnectionID_Use()` runs before allocation can fail; call `DtlsCidClear()` on the MEMORY_ERROR exits (`src/dtls.c:1611-1636`).
- **Mid-handshake export succeeds but always produces an unimportable blob** (import rejects at `src/dtls13.c:3115-3118`), and the buffer then holds handshake traffic secrets nobody can resume into. Fail fast or document.
- **Non-dup deferred work is neither exported nor refused**: `dtls13Rtx.sendAcks`/`retransmit`/`rtxRecords`, `options.dtls13SendMoreAcks` fall outside the guards even though the dup case treats equivalent parked work as export-invalidating. ACK loss self-heals; `rtxRecords` loss is silent — the PR's own `test_dtls13_export_pending_ticket_lost` demonstrates a NewSessionTicket dropped on the floor. Document the limitation next to the key-material warning, or refuse.
- **v7 bump changes the stamp of all exports** (TLS, DTLS 1.2 included), and old importers reject v7 — upgrade the import side first in mixed deployments. Release-notes item.
- **CID wire-layout formula `(3*OPAQUE8_LEN)+rxSz+txSz` maintained in three places** (`src/dtls.c:1566`, `src/internal.c:2338-2340`, `wolfssl/internal.h:3066-3068`). One `DtlsCidExportSize(ssl)` helper.
- **`DtlsCidImport` doesn't invalidate `ssl->recordSzOverhead`** unlike every other tx-CID mutation (`src/dtls.c:1371-1376, 1479`). Harmless on a fresh object, a trap otherwise.
- **Bare `42` in `DTLS_EXPORT_KEY_SZ`** (`wolfssl/internal.h:3036`) where every other term is spelled out.
- **v6 compat tested with a same-build fixture**: `tests/api/test_dtls.c:7350-7385` rewrites a live v7 blob's version nibble, which cannot catch layout changes to the shared v6 body (precisely the region commit 1 touches). Check in a canned v6 blob captured from `origin/master` under this config.

## Maintainer-comment dispositions (`#2`-`#5` above; plus)

- **`#1` (dox "Isn't this true for all DTLS versions?")** — INVALID as a defect. The doc's second sentence is explicitly scoped "For DTLS 1.3" and states a strictly-1.3 property that is accurate (the v7 blob carries the traffic secrets precisely because future epoch keys re-derive from them, `src/dtls13.c:3040-3042, 2700-2726`). Optional reword only.
- **`#3` (separate the folded `OPAQUE8_LEN` in `DtlsCidImport`)** — VALID, style. `src/dtls.c:1601` folds the txSz length octet into the rx-body bound and then uses two offset conventions in one record; correct today, exactly the confusion cited.
- **`#6` ("dump the full `Dtls13Epoch` array, single XMEMCPY")** — complaint VALID (the three length-prefixed fields and the `recvEpoch==sendEpoch` alias special-cases at `src/dtls13.c:2912-2960, 3055-3067, 3121-3176` are the real complexity drivers), but the proposed mechanism is unsound: `Dtls13Epoch` has feature-conditional layout (`#ifdef`'d `dropCount`, `wolfssl/internal.h:6388-6390`) and field sizes that vary across conforming builds (`WOLFSSL_DTLS_WINDOW_WORDS` user-tunable, `MAX_SYM_KEY_SIZE`∈{16,24,32}, `MAX_WRITE_IV_SZ`∈{16,32,48}) — `sizeof` differs between two conforming builds, and raw bytes bake host order + padding into a persistent cross-build format whose contract is "v3-v6 blobs still import." A sound version of the simplification: fixed-layout field-wise encode of **all** `DTLS13_EPOCH_SIZE` slots including keys, dropping the derive-on-import path — no security regression (the blob already carries the secrets), cost a few hundred bytes. Worth agreeing in the review thread before implementing.

## Explicitly checked and found *not* problematic

- `keys.keyUpdateRespond` outside write-dup is not independently exploitable: `WOLFSSL_RW_THREADED` routes it through the guarded `options.sendKeyUpdate` (`src/tls13.c:13781`); non-threaded builds consume it synchronously in `DoTls13KeyUpdate`. The residual merge window is covered by P1-6.
- Early data / HRR / 0-RTT epochs: blobs with epoch < `TRAFFIC0` are refused on import; server/client asymmetry has dedicated tests.
- `ForceZero(buf, bufSz)` additions in `wolfSSL_send_session` (`src/ssl_sess.c:3618, 3623`) use the allocation-size `word32` correctly; the `MAX_EXPORT_IP + 3*WOLFSSL_EXPORT_LEN` reservation fix removes a latent short-reservation in the old code.
- Copilot's single comment (`test_dtls13.h:103`, missing `\`) is **STALE** — that file is no longer in the diff after the 08-26 restructure.

src/dtls13.c
julek-wolfssl: memcpy is not the correct approach

rizlik:
retest this please

julek-wolfssl (dismissed):

rizlik:
wait on merge, I need to double check something

rizlik:
retest this please (no logs)

rizlik:
retest this

julek-wolfssl (dismissed):

rizlik:
retest this please

PR #11096, opened 2026-08-06
Support post-quantum-only (ML-KEM + ML-DSA/SLH-DSA) TLS 1.3 builds

Frauschi:
## Description

This PR adds support for a quantum-safe-only TLS 1.3 stack: ML-KEM key exchange with ML-DSA or SLH-DSA authentication, with RSA, ECC and DH all disabled. That is the CNSA 2.0 / EO 14412 posture, and until now such a configure line did not build: the "no cipher suites" `#error` fired before anything else, and several files failed `-Werror` once every classic algorithm was compiled out. This is based on [initial patches published in our blog](https://www.wolfssl.com/white-house-eo-14412/).

It also fixes a latent TLS 1.3 bug that only shows up once certificates get large enough to span multiple TLS records, which is exactly what post-quantum certificates do.

### Library changes

- **Cipher suite gates (`src/ssl.c`, `wolfssl/internal.h`).** Treat a usable standalone ML-KEM group as a valid key exchange, and Falcon/ML-DSA/SLH-DSA as valid authentication, so the two "no cipher suites" `#error`s no longer fire in a PQC-only build.
- **`src/tls.c`.** Advertise the `psk_dhe_ke` exchange mode when a usable standalone ML-KEM group is available. Without it an ML-KEM-only build fell back to `psk_ke`, the resumption key schedules diverged, and the first AEAD record failed to authenticate (`AES_GCM_AUTH_E`), which broke session-ticket resumption.
- **`src/internal.c`.** Extend the `CopyDecodedToX509` `pkCurveOID` guard so it matches the struct member guards.
- **`wolfcrypt/src/asn.c`.** Guard the RSA/ECC-only certificate signing helpers so certificate generation still compiles with RSA and ECC disabled, while keeping `MakeSignatureCb` available for the `WOLFSSL_CERT_SIGN_CB` (TPM/HSM) offload path.
- **`src/tls13.c`.** Hoist the `hashAlgo` void cast in `EncodeSigAlg` out of the Ed25519 and Ed448 cases so a build with no classic signature algorithm still marks the parameter used.
- **`src/internal.c`, `src/tls13.c`, `src/ssl_load.c`.** Mark `ssl`, `sigOut`, `keyType` and `keySz` used. Every reader of these sits inside a per-algorithm block, so a build with no signing algorithm at all left them unused and `-Werror` rejected the library.

### Certificate chain send fix (`src/tls13.c`)

`SendTls13Certificate` keeps its chain walk cursor in the function locals `len`, `idx`, `offset` and `p`, but the only state that survives the `WANT_WRITE` return of a non-blocking send is `ssl->fragOffset`, and that is consulted for the leaf certificate alone. A send that blocked part way through the chain therefore re-primed the walk on the next call and copied the chain from its first byte again. The byte count still matched the announced payload size, so the message stayed well formed on the wire while the tail of the chain was replaced by a repeat of its head, and the peer rejected it with `BUFFER_ERROR`.

The fix rebuilds the cursor from `ssl->fragOffset` when a resume lands inside the chain. `NextCert` reads each entry's three byte length prefix and skips it, so passing over the entries already sent costs one hop per certificate and only happens on a resume. No new per-connection state is added, which keeps the embedded footprint unchanged.

This is not a regression from this branch. `git log -S "offset = OPAQUE16_LEN" -- src/tls13.c` dates the code to `2b1e9973e` (2016). Reaching it needs a Certificate message larger than one record on a non-blocking socket, and classic chains stay under one record, which is why it lay dormant. CI never caught it either: the SLH-DSA test configurations carry no `-6` cases, and every other SLH-DSA configuration keeps RSA and ECC on, so the `-6` cases in `tests/test-tls13.conf` run with small RSA certificates. The new `slhdsa-pqc-only` configuration is the first to combine large certificates with a non-blocking send.

It is kept as a separate commit so it can be reviewed apart from the post-quantum feature work.

### Example and test credentials (`wolfssl/test.h`, `examples/`, `tests/`)

Whether a build can authenticate with a post-quantum certificate is three separate questions: is the algorithm compiled in, are test credentials wired up for the enabled parameter set, and can this build sign. `test.h` now answers them with three tiers of macro rather than testing `WOLFSSL_HAVE_MLDSA` and `WOLFSSL_HAVE_SLHDSA` directly.

| Macro | Means |
|---|---|
| `TEST_HAVE_MLDSA_CERT_FILES` / `TEST_HAVE_SLHDSA_CERT_FILES` | The credential paths name real files. ML-DSA needs an enabled parameter level. SLH-DSA needs a 128s parameter set, since both families share a 128s root and only those certificates are wired up, and PEM decoding, since the entity certificates ship as PEM only. Loading a certificate needs verification support alone, so `test_tls13.c` uses this tier. |
| `TEST_HAVE_MLDSA_CERTS` / `TEST_HAVE_SLHDSA_CERTS` | The above, plus the ability to sign. A verify-only build derives `WOLFSSL_MLDSA_NO_SIGN` and drops private key support entirely, so it can neither load the key nor produce a CertificateVerify. The example credential ladders use this tier. |
| `TEST_HAVE_PQC_CERT_AUTH` | Either of those, plus `WOLFSSL_TLS13`. Both algorithms are TLS 1.3 only, so folding the version into the macro keeps the version defaults and the PSK fallbacks in the client, server and echo examples from disagreeing. Falcon is deliberately absent, having no credentials in the ladders, so a Falcon-only build keeps the historic TLS 1.2 PSK behavior. |

A build that fails a gate now falls back to PSK instead of naming credentials it cannot use. That removes two hard `#error`s that were reachable straight from configure: `--enable-slhdsa` with only non-128s parameter sets broke every build that includes `test.h`, even ones with RSA and ECC enabled that would never touch an SLH-DSA credential.

Two credential bugs fixed along the way:

- **`test.h`.** Pair the DER ML-DSA certificate with `mldsa<N>-key.der`. The `mldsa<N>_priv-only.der` files are standalone PKCS#8 encoding vectors from an unrelated key pair, so a build without PEM decoding presented a certificate and signed with a key that did not match it, and the peer rejected the CertificateVerify.
- **`examples/echoserver/echoserver.c`.** Load the SLH-DSA key with `CERT_FILETYPE` like every neighboring branch, rather than hardcoding `WOLFSSL_FILETYPE_PEM` against a certificate loaded in the build-dependent format.

`tests/api.c` skips `test_wolfSSL_PKCS8` unless RSA, ECC or PEM decoding is available. With none of them every key-loading body compiles out, leaving the file locals unused and nothing for the test to do.

## Testing

New CI entries in `.github/configs/pq-all.json`, one per signature algorithm:

```
--enable-mlkem --enable-mldsa  --enable-tls-mlkem-standalone \
  --enable-session-ticket --disable-rsa --disable-ecc --disable-dh
--enable-mlkem --enable-slhdsa --enable-tls-mlkem-standalone \
  --enable-session-ticket --disable-rsa --disable-ecc --disable-dh
```

Each exercises the build, a full TLS 1.3 handshake, and ML-KEM-only session-ticket resumption.

`tests/test-tls13-slhdsa-entity-128s.conf` gains two scenarios that drive the certificate chain send under simulated `WANT_WRITE` (`-6`), one server-auth and one mutual-auth, so the resume path is covered in both directions. Without the `src/tls13.c` hunk `./tests/unit.test` exits 1 at the new case; with it, exit 0.

### DIFF
--- a/scripts/ocsp-stapling_tls13multi.test
+++ b/scripts/ocsp-stapling_tls13multi.test
@@ -60,6 +60,18 @@ else
     dtls13=yes
 fi
 
+# The fragmented-send case needs a client that can request a small maximum
+# fragment length and a server that can simulate WANT_WRITE. Both options are
+# silently ignored otherwise, which would turn the case into a no-op that still
+# reports PASSED.
+if ! ./examples/client/client -? 2>&1 | grep -q 'Maximum Fragment Length'; then
+    fragmented_send=no
+elif ./examples/server/server -6 -? 2>&1 | grep -q 'Ignoring -6'; then
+    fragmented_send=no
+else
+    fragmented_send=yes
+fi
+
 if [[ ("$tls13" == "no") && ("$dtls13" == "no") ]]; then
     echo 'skipping ocsp-stapling_tls13multi.test because TLS1.3 is not available.' 1>&2
     exit 77
@@ -492,6 +504,27 @@ if [ "$tls13" == "yes" ]; then
         exit 1
     fi
     printf '%s\n\n' "Test successfully REVOKED!"
+
+    printf '%s\n\n' "------------- TEST CASE 8 FRAGMENTED SEND --------------------"
+    if [ "$fragmented_send" == "no" ]; then
+        printf '%s\n\n' "Test SKIPPED: needs HAVE_MAX_FRAGMENT and async I/O."
+    else
+        # A small maximum fragment length (-F 1) splits the stapled Certificate
+        # message over many records, and the server (-6) blocks on every one of
+        # them, so the send resumes from a WANT_WRITE inside the chain.
+        remove_single_rF "$ready_file5"
+        ./examples/server/server -c certs/ocsp/server3-cert.pem \
+                                -k certs/ocsp/server3-key.pem -R "$ready_file5" \
+                                -p "$port5" -v 4 -6 &
+        server_pid5=$!
+        wait_for_readyFile "$ready_file5" "$server_pid5" "$port5"
+        ./examples/client/client -C -A certs/ocsp/root-ca-cert.pem -W 1 -v 4 \
+                                -F 1 -p "$port5"
+        RESULT=$?
+        [ "$RESULT" -ne 0 ] && printf '\n\n%s\n' "Client connection 8 failed" \
+                            && exit 1
+        printf '%s\n\n' "Test PASSED!"
+    fi
 fi
 
 if [ "$dtls13" == "yes" ]; then
--- a/src/ssl.c
+++ b/src/ssl.c
@@ -49,7 +49,10 @@
 #if !defined(WOLFSSL_ALLOW_NO_SUITES) && !defined(WOLFCRYPT_ONLY)
     #if defined(NO_DH) && !defined(HAVE_ECC) && !defined(WOLFSSL_STATIC_RSA) \
                 && !defined(WOLFSSL_STATIC_DH) && !defined(WOLFSSL_STATIC_PSK) \
-                && !defined(HAVE_CURVE25519) && !defined(HAVE_CURVE448)
+                && !defined(HAVE_CURVE25519) && !defined(HAVE_CURVE448) \
+                && (!defined(WOLFSSL_TLS13) \
+                    || !defined(WOLFSSL_HAVE_MLKEM_CLIENT_SUPPORT) \
+                    || defined(WOLFSSL_TLS_NO_MLKEM_STANDALONE))
         #error "No cipher suites defined because DH disabled, ECC disabled, " \
                "and no static suites defined. Please see top of README"
     #endif
--- a/src/tls.c
+++ b/src/tls.c
@@ -4672,17 +4672,6 @@ int TLSX_UseCertificateStatusRequestV2(TLSX** extensions, byte status_type,
 
 #endif /* HAVE_CERTIFICATE_STATUS_REQUEST_V2 */
 
-/* ML-KEM client support requires generating a key pair (encapsulation key) and
- * decapsulating the server's ciphertext. */
-#if defined(WOLFSSL_HAVE_MLKEM) && !defined(WOLFSSL_MLKEM_NO_MAKE_KEY) && \
-     !defined(WOLFSSL_MLKEM_NO_DECAPSULATE)
-    #define WOLFSSL_HAVE_MLKEM_CLIENT_SUPPORT
-#endif
-/* ML-KEM server support requires encapsulating to the client's key. */
-#if defined(WOLFSSL_HAVE_MLKEM) && !defined(WOLFSSL_MLKEM_NO_ENCAPSULATE)
-    #define WOLFSSL_HAVE_MLKEM_SERVER_SUPPORT
-#endif
-
 #if defined(HAVE_SUPPORTED_CURVES) || \
     (defined(WOLFSSL_TLS13) && defined(HAVE_SUPPORTED_CURVES))
 
@@ -16752,7 +16741,9 @@ int TLSX_PopulateExtensions(WOLFSSL* ssl, byte isServer)
                     modes = 1 << PSK_KE;
                 }
             #if !defined(NO_DH) || defined(HAVE_ECC) || \
-                          defined(HAVE_CURVE25519) || defined(HAVE_CURVE448)
+                          defined(HAVE_CURVE25519) || defined(HAVE_CURVE448) || \
+                          (defined(WOLFSSL_HAVE_MLKEM_CLIENT_SUPPORT) && \
+                           !defined(WOLFSSL_TLS_NO_MLKEM_STANDALONE))
                 if (!ssl->options.noPskDheKe) {
                     modes |= 1 << PSK_DHE_KE;
                 }
--- a/src/tls13.c
+++ b/src/tls13.c
@@ -9664,33 +9664,46 @@ static int WriteCSRToBuffer(WOLFSSL* ssl, DerBuffer** certExts,
         for (extIdx = 0; extIdx < (word16)(extSz_num); extIdx++) {
             tmpSz = TLSX_CSR_GetSize_ex(csr, 0, (int)extIdx);
 
-            if (tmpSz > (OPAQUE8_LEN + OPAQUE24_LEN) &&
-                certExts[extIdx] == NULL) {
-                /* csr extension is not zero */
-                if (tmpSz > WOLFSSL_MAX_16BIT)
-                    return BUFFER_E;
-                extSz[extIdx] = (word16)tmpSz;
+            if (ssl->fragOffset != 0 && certExts[extIdx] != NULL) {
+                /* A fragmented send is being resumed and this buffer was
+                 * written by the earlier call. extSz starts over on every
+                 * call, so recover this entry's size from the length written
+                 * into the buffer. */
+                ato16(certExts[extIdx]->buffer, &extSz[extIdx]);
+                extSz[extIdx] += OPAQUE16_LEN;
+            }
+            else {
+                /* Not a resume, so anything still allocated here is left over
+                 * from a completed message and must not be reused. */
+                FreeDer(&certExts[extIdx]);
 
-                ret = AllocDer(&certExts[extIdx], extSz[extIdx] + ex_offset,
-                                                    CERT_TYPE, ssl->heap);
-                if (ret < 0)
-                    return ret;
-                der = certExts[extIdx];
+                if (tmpSz > (OPAQUE8_LEN + OPAQUE24_LEN)) {
+                    /* csr extension is not zero */
+                    if (tmpSz > WOLFSSL_MAX_16BIT)
+                        return BUFFER_E;
+                    extSz[extIdx] = (word16)tmpSz;
 
-                /* write extension type */
-                c16toa(ext->type, der->buffer
-                                + OPAQUE16_LEN);
-                /* writes extension data length. */
-                c16toa(extSz[extIdx], der->buffer
-                            + HELLO_EXT_TYPE_SZ + OPAQUE16_LEN);
-                /* write extension data */
-                extSz[extIdx] = (word16)TLSX_CSR_Write_ex(csr,
-                        der->buffer + ex_offset, 0, extIdx);
-                /* add extension offset */
-                extSz[extIdx] += (word16)ex_offset;
-                /* extension length */
-                c16toa(extSz[extIdx] - OPAQUE16_LEN,
-                            der->buffer);
+                    ret = AllocDer(&certExts[extIdx], extSz[extIdx] + ex_offset,
+                                                        CERT_TYPE, ssl->heap);
+                    if (ret < 0)
+                        return ret;
+                    der = certExts[extIdx];
+
+                    /* write extension type */
+                    c16toa(ext->type, der->buffer
+                                    + OPAQUE16_LEN);
+                    /* writes extension data length. */
+                    c16toa(extSz[extIdx], der->buffer
+                                + HELLO_EXT_TYPE_SZ + OPAQUE16_LEN);
+                    /* write extension data */
+                    extSz[extIdx] = (word16)TLSX_CSR_Write_ex(csr,
+                            der->buffer + ex_offset, 0, extIdx);
+                    /* add extension offset */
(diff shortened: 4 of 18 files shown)


src/tls13.c
```diff
@@ -10007,6 +10006,41 @@ static int SendTls13Certificate(WOLFSSL* ssl)
...
+            len = NextCert(ssl->buffers.certChain->buffer,
+                           ssl->buffers.certChain->length, &idx);
+            if (len == 0)
+                break;
+        #if defined(HAVE_CERTIFICATE_STATUS_REQUEST) && \
+                !defined(NO_WOLFSSL_SERVER)
+            if (extIdx + 1 < MAX_CERT_EXTENSIONS)
+                extIdx++;
+        #endif
+            entrySz = len + extSz[extIdx];
```
Frauschi, line 10050: fixed

src/tls13.c
```diff
@@ -10007,6 +10007,40 @@ static int SendTls13Certificate(WOLFSSL* ssl)
...
```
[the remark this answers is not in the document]
Frauschi: fixed

tests/api/test_tls13.c
```diff
@@ -84,6 +84,19 @@ int test_tls13_apis(void)
 #elif defined(HAVE_ED448)
     const char*  ourCert = ed448CertFile;
     const char*  ourKey  = ed448KeyFile;
+/* The CERT_FILES gates, not the signing ones: only a certificate on the context
+ * is needed here, so the key load below is allowed to fail. */
+#elif defined(TEST_HAVE_MLDSA_CERT_FILES)
```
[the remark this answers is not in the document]
Frauschi: fixed

src/tls13.c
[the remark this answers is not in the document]
Frauschi: fixed

src/tls13.c
```diff
@@ -10007,6 +10007,40 @@ static int SendTls13Certificate(WOLFSSL* ssl)
...
+
+            if (chainPos < entrySz) {
+                /* Resume part way through this entry. */
+                p = ssl->buffers.certChain->buffer + prevIdx;
+                offset = chainPos;
+                chainPos = 0;
+            }
+            else {
+                /* Entry already sent in full; stay primed for the next one. */
+                chainPos -= entrySz;
```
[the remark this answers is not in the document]
Frauschi: fixed

src/tls.c
```diff
@@ -16752,7 +16752,9 @@ int TLSX_PopulateExtensions(WOLFSSL* ssl, byte isServer)
                     modes = 1 << PSK_KE;
                 }
             #if !defined(NO_DH) || defined(HAVE_ECC) || \
-                          defined(HAVE_CURVE25519) || defined(HAVE_CURVE448)
+                          defined(HAVE_CURVE25519) || defined(HAVE_CURVE448) || \
+                          (defined(WOLFSSL_HAVE_MLKEM) && \
```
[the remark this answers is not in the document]
Frauschi: fixed

src/tls13.c
```diff
@@ -9692,6 +9691,13 @@ static int WriteCSRToBuffer(WOLFSSL* ssl, DerBuffer** certExts,
                 c16toa(extSz[extIdx] - OPAQUE16_LEN,
                             der->buffer);
             }
+            else if (certExts[extIdx] != NULL) {
```
dgarske: WriteCSRToBuffer infers "resumed send" from a non-NULL certExts[] pointer, which is also true for a buffer left over from a completed message

Consider using `ssl->fragOffset != 0` instead.

Recommendation: Key the branch off `ssl->fragOffset != 0`, which is the actual resume signal and is already what gates `SetupOcspResp` at src/tls13.c:9970. Separately, relax the resume block's condition at src/tls13.c:10024 from `>` to `>=` so `certExts[0]` is always released once the cursor is past the leaf; I traced that change and it is safe (with `chainPos == 0` the while loop does not execute, and `idx`/`offset`/`entrySz`/`extIdx` all stay at the values the send loop needs).

src/ssl.c
```diff
@@ -49,7 +49,9 @@
 #if !defined(WOLFSSL_ALLOW_NO_SUITES) && !defined(WOLFCRYPT_ONLY)
     #if defined(NO_DH) && !defined(HAVE_ECC) && !defined(WOLFSSL_STATIC_RSA) \
                 && !defined(WOLFSSL_STATIC_DH) && !defined(WOLFSSL_STATIC_PSK) \
-                && !defined(HAVE_CURVE25519) && !defined(HAVE_CURVE448)
+                && !defined(HAVE_CURVE25519) && !defined(HAVE_CURVE448) \
+                && (!defined(WOLFSSL_HAVE_MLKEM) || \
+                    defined(WOLFSSL_TLS_NO_MLKEM_STANDALONE))
```
dgarske: "No cipher suites defined" #error escape for ML-KEM does not require TLS 1.3

Recommendation: Add a `!defined(WOLFSSL_TLS13) ||` term to the ML-KEM escape so the diagnostic still fires for a TLS 1.2-only build. Consider also gating configure.ac's standalone auto-enable on ENABLED_TLS13 — it currently prints "auto-enabling standalone ML-KEM for TLS 1.3" even when --disable-tls13 was given.

wolfssl/internal.h
```diff
@@ -854,7 +854,9 @@
 #if !defined(WOLFCRYPT_ONLY) && defined(NO_PSK) && \
     (defined(NO_DH) || !defined(HAVE_ANON)) && \
     defined(NO_RSA) && !defined(HAVE_ECC) && \
-    !defined(HAVE_ED25519) && !defined(HAVE_ED448)
+    !defined(HAVE_ED25519) && !defined(HAVE_ED448) && \
+    !defined(HAVE_FALCON) && !defined(WOLFSSL_HAVE_MLDSA) && \
+    !defined(WOLFSSL_HAVE_SLHDSA)
```
dgarske: "No cipher suites available" #error accepts Falcon/ML-DSA/SLH-DSA auth without requiring TLS 1.3

Recommendation: Gate the three post-quantum terms on WOLFSSL_TLS13 so a TLS 1.2-only build with only PQC signatures still gets the compile-time diagnostic.

wolfssl/test.h
```diff
@@ -461,18 +468,64 @@ err_sys_with_errno_func(const char* msg, const char *file, int line)
...
+ * up, and the entity certificates ship as PEM only. */
+#if defined(WOLFSSL_HAVE_SLHDSA) && defined(WOLFSSL_PEM_TO_DER) && \
+    (defined(WOLFSSL_SLHDSA_PARAM_128S) || \
+     defined(WOLFSSL_SLHDSA_PARAM_SHA2_128S))
+    #define TEST_HAVE_SLHDSA_CERT_FILES
+#endif
+
+/* The same credentials, plus the ability to sign: a verify-only build can
+ * neither load the key nor produce a CertificateVerify. */
+#if defined(TEST_HAVE_MLDSA_CERT_FILES) && \
```
dgarske, line 497: Add `&& !defined(WOLFSSL_MLDSA_NO_VERIFY)` — the examples need both halves, since they sign their own CertificateVerify and verify the peer's chain. Reproduce with `./configure --disable-rsa --disable-ecc --disable-dh --enable-mlkem --enable-tls-mlkem-standalone --enable-psk --enable-mldsa=44,make,sign`.

scripts/ocsp-stapling_tls13multi.test
```diff
@@ -492,6 +492,23 @@ if [ "$tls13" == "yes" ]; then
...
+    # message over many records, and the server (-6) blocks on every one of
+    # them, so the send resumes from a WANT_WRITE inside the chain. Both
+    # options are ignored by builds that do not support them.
+    remove_single_rF "$ready_file5"
+    ./examples/server/server -c certs/ocsp/server3-cert.pem \
+                            -k certs/ocsp/server3-key.pem -R "$ready_file5" \
+                            -p "$port5" -v 4 -6 &
+    server_pid5=$!
+    wait_for_readyFile "$ready_file5" "$server_pid5" "$port5"
+    ./examples/client/client -C -A certs/ocsp/root-ca-cert.pem -W 1 -v 4 \
```
dgarske: New OCSP fragmented-send test silently no-ops in builds without HAVE_MAX_FRAGMENT

Recommendation: Detect HAVE_MAX_FRAGMENT with the script's existing `client -?` grep idiom and skip the case loudly rather than running a no-op that reports PASSED. The comment "Both options are ignored by builds that do not support them" documents the hole but does not make the coverage loss visible in CI output.

src/ssl.c
dgarske: Consider moving tls.c:4677 into internal.h or settings.h and using that gate.

```
/* src/tls.c:4677 */
    #if defined(WOLFSSL_HAVE_MLKEM) && !defined(WOLFSSL_MLKEM_NO_MAKE_KEY) && \
         !defined(WOLFSSL_MLKEM_NO_DECAPSULATE)
        #define WOLFSSL_HAVE_MLKEM_CLIENT_SUPPORT
    #endif
```

dgarske (changes requested):

src/tls13.c
Frauschi: Both done. `WriteCSRToBuffer` now keys the resume branch on `ssl->fragOffset != 0`, and the non-resume path `FreeDer`s any leftover buffer before rebuilding it. Relaxed the resume block to `>=` as well.

Verified the exact-boundary case actually fires rather than passing vacuously: clamping `maxFragment` to 3170 (and 1585) puts a record boundary exactly at `certSz + extSz[0]` = 3159, `chainPos == 0` is hit, and the handshake completes.

src/ssl.c
Frauschi: Added `!defined(WOLFSSL_TLS13) ||`, and gated the configure.ac auto-enable on `ENABLED_TLS13` too, so `--disable-tls13` no longer prints the "auto-enabling standalone ML-KEM for TLS 1.3" notice.

Confirmed the `#error` now fires for a TLS 1.2-only ML-KEM build and stays silent with TLS 1.3 on.

wolfssl/internal.h
Frauschi: Done, the three terms are now wrapped in `(!defined(WOLFSSL_TLS13) || (...))`.

Checked both directions: a TLS 1.2-only ML-DSA build gets the diagnostic, the TLS 1.3 build still compiles.

wolfssl/test.h
Frauschi: Added.

Your repro line also turned up three latent breaks, all fixed in the same commit: two pre-existing `-Werror` failures (unused `ssl` in `ProcessPeerCertDecodeKey`, and `ReuseKey` going unused because its guard lists `WOLFSSL_HAVE_MLDSA` while the call site also needs `!WOLFSSL_MLDSA_NO_VERIFY`), plus test.h defaulting to TLS 1.2 in an ML-KEM-only build where TLS 1.2 has no cipher suite at all. That last one is a new `TEST_NO_CLASSIC_KEX`, and both version defaults now come from one macro so they cannot drift.

Two things still fail in that exact config and I left them alone: `resume.test` needs `--enable-session-ticket` (both PQC CI entries have it), and `unit.test`'s `-j` PSK-plus-certs case wants a certificate the build has none of. Failing loudly there seemed better than silently degrading `-j` to plain PSK.

scripts/ocsp-stapling_tls13multi.test
Frauschi: Now detected and skipped loudly. `HAVE_MAX_FRAGMENT` via a `client -?` grep for the `-F` usage line; async I/O via `server -6 -?`, since getopt handles `-6` before `-?` and emits "Ignoring -6 since async I/O support not compiled in." when it is missing.

Verified both paths: the case runs and passes under `--enable-all`, and prints `Test SKIPPED` under `--disable-maxfragment`.

src/ssl.c
Frauschi: Moved both `WOLFSSL_HAVE_MLKEM_CLIENT_SUPPORT` and `WOLFSSL_HAVE_MLKEM_SERVER_SUPPORT` from tls.c into internal.h, just after the `extern "C"` block, and ssl.c's escape now uses the client one instead of bare `WOLFSSL_HAVE_MLKEM`.

dgarske (dismissed):

dgarske (approved):

PR #11097, opened 2026-08-06
TLS 1.3: fix early data decryption handling and middlebox compat build option

### LINKED ISSUE #11011: Corrupted accepted 0-RTT data does not produce the required bad_record_mac alert
# Corrupted accepted 0-RTT data does not produce the required bad_record_mac alert

## Problem Description

After wolfSSL accepts TLS 1.3 early data, a corrupted 0-RTT application_data record is treated as ignorable early data instead of causing the required fatal bad_record_mac alert. The focused memio test first minted a session ticket that permits 0-RTT, then resumed with early data, flipped one byte in the first encrypted client application_data record before the server decrypted it, and checked the server's transmitted alert history. wolfSSL returned a fatal error to the API but did not transmit or record bad_record_mac.

## Standard Requirement

- Section: 4.2.10. Early Data Indication
- Evidence: `D:\paper\idea\document\rfc8446.txt:3034-3039`

> If the server chooses to accept the "early_data" extension, then it
>    MUST comply with the same error-handling requirements specified for
>    all records when processing early data records.  Specifically, if the
>    server fails to decrypt a 0-RTT record following an accepted
>    "early_data" extension, it MUST terminate the connection with a
>    "bad_record_mac" alert as per Section 5.2.

Interpretation:

Once the server accepts the TLS 1.3 `early_data` extension, a decryption failure on a 0-RTT record is not optional recovery behavior. The server must terminate the connection with a `bad_record_mac` alert.

## Relevant Source Code

wolfSSL accepts TLS 1.3 early data when the ClientHello contains `early_data`, the server has a nonzero early-data limit, the selected ticket is usable, and the server derives early-data decrypt keys. The record processing path then has a special TLS 1.3 early-data branch on decrypt failure that returns before the generic `bad_record_mac` alert is sent.

### `src/tls13.c:6906-6935`

```c
        extEarlyData = TLSX_Find(ssl->extensions, TLSX_EARLY_DATA);
        if (extEarlyData != NULL) {
            /* Check if accepting early data and first PSK.
             * RFC 9973: early_data is not compatible with
             * cert_with_extern_psk, so skip key derivation in that case. */
            if (ssl->earlyData != no_early_data && first
                && ssl->options.maxEarlyDataSz > 0
    #ifdef WOLFSSL_CERT_WITH_EXTERN_PSK
                && !hasCertWithExternPsk
    #endif
    #if defined(HAVE_SESSION_TICKET) && !defined(NO_SESSION_CACHE)
                /* RFC 8446 section 8: evict the session from the cache.
                 * Accept 0-RTT only when the eviction found the entry
                 * (single-use). */
                && wolfSSL_SSL_CTX_remove_session(ssl->ctx, ssl->session)
                    == 1
    #endif
            ) {
                extEarlyData->resp = 1;

                /* Derive early data decryption key. */
                ret = DeriveTls13Keys(ssl, early_data_key, DECRYPT_SIDE_ONLY,
                                                                             1);
                if (ret != 0)
                    return ret;
                if ((ret = SetKeysSide(ssl, DECRYPT_SIDE_ONLY)) != 0)
                    return ret;

                ssl->keys.encryptionOn = 1;
                ssl->earlyData = process_early_data;
```

This is the server acceptance path: `extEarlyData->resp = 1` means the server will signal accepted early data, and `ssl->earlyData = process_early_data` puts the server into accepted 0-RTT processing with early-data decrypt keys installed.

### `src/internal.c:24583-24625`

```c
                else {
                    WOLFSSL_MSG("Decrypt failed");
                #ifdef WOLFSSL_DTLS
                    /* If in DTLS mode, if the decrypt fails for any
                     * reason, pretend the datagram never happened. */
                    if (ssl->options.dtls) {
                        DropAndRestartProcessReply(ssl);
                        return HandleDTLSDecryptFailed(ssl);
                    }
                #endif /* WOLFSSL_DTLS */
                #ifdef WOLFSSL_EARLY_DATA
                    if (ssl->options.tls1_3) {
                         if (ssl->options.side == WOLFSSL_SERVER_END &&
                                  ssl->earlyData != no_early_data &&
                                  ssl->options.clientState <
                                                      CLIENT_FINISHED_COMPLETE) {
                            ssl->earlyDataSz += ssl->curSize;
                            if (ssl->earlyDataSz <=
                                                  ssl->options.maxEarlyDataSz) {
                                WOLFSSL_MSG("Ignoring EarlyData!");
                                if (ssl->keys.peer_sequence_number_lo-- == 0)
                                    ssl->keys.peer_sequence_number_hi--;
                                ssl->options.processReply = doProcessInit;
                                ssl->buffers.inputBuffer.idx += ssl->curSize;
                                if (ssl->buffers.inputBuffer.idx >
                                    ssl->buffers.inputBuffer.length) {
                                    WOLFSSL_ERROR(BUFFER_E);
                                    return BUFFER_E;
                                }

                                return 0;
                            }
                            WOLFSSL_MSG("Too much EarlyData!");
                            SendAlert(ssl, alert_fatal, unexpected_message);
                            WOLFSSL_ERROR(TOO_MUCH_EARLY_DATA);
                            return TOO_MUCH_EARLY_DATA;
                        }
                    }
                #endif
                    SendAlert(ssl, alert_fatal, bad_record_mac);
                    /* Push error once we know that we will error out here */
                    WOLFSSL_ERROR(ret);
                    return ret;
```

For a TLS 1.3 server still before `CLIENT_FINISHED_COMPLETE`, any decrypt failure on an early-data record within `maxEarlyDataSz` takes the `Ignoring EarlyData!` branch and returns `0` before reaching the unconditional `SendAlert(ssl, alert_fatal, bad_record_mac)` path. This contradicts RFC 8446 once the server has accepted the `early_data` extension.

### `tests/api/test_tls13.c:5247-5295`

```c
    /* Mint a ticket that permits 0-RTT. */
    ExpectIntEQ(test_memio_setup(&test_ctx, &ctx_c, &ctx_s, &ssl_c, &ssl_s,
                    wolfTLSv1_3_client_method, wolfTLSv1_3_server_method),
                0);
    ExpectIntGE(wolfSSL_CTX_set_max_early_data(ctx_s, MAX_EARLY_DATA_SZ), 0);
    ExpectIntGE(wolfSSL_set_max_early_data(ssl_s, MAX_EARLY_DATA_SZ), 0);
    ExpectIntEQ(test_memio_do_handshake(ssl_c, ssl_s, 10, NULL), 0);
    ExpectIntEQ(wolfSSL_read(ssl_c, msgBuf, (int)sizeof(msgBuf)), -1);
    ExpectIntEQ(wolfSSL_get_error(ssl_c, -1), WOLFSSL_ERROR_WANT_READ);
    ExpectNotNull(sess = wolfSSL_get1_session(ssl_c));
...
    ExpectIntEQ(test_tls13_early_data_write_until_write_ok(ssl_c, earlyMsg,
                    (int)sizeof(earlyMsg), &written), (int)sizeof(earlyMsg));
    ExpectIntEQ(written, (int)sizeof(earlyMsg));

    /* Corrupt the first encrypted application_data record in the client's
     * outbound buffer before the server attempts to decrypt it. */
...
    if (EXPECT_SUCCESS())
        test_ctx.s_buff[fuzzOff] ^= 0x01;

    ExpectIntEQ(test_tls13_early_data_read_until_write_ok(ssl_s, msgBuf,
                    (int)sizeof(msgBuf), &read), WOLFSSL_FATAL_ERROR);
    ExpectIntEQ(wolfSSL_get_alert_history(ssl_s, &h), WOLFSSL_SUCCESS);
    ExpectIntEQ(h.last_tx.code, bad_record_mac);
    ExpectIntEQ(h.last_tx.level, alert_fatal);
```

The focused test uses wolfSSL's in-memory I/O harness, not an abstract adapter: it creates a real TLS 1.3 ticket allowing 0-RTT, resumes with early data, mutates the encrypted record bytes in the client-to-server buffer, and then asserts that the server sent fatal `bad_record_mac`.

### Round 1

- Earlier native handshake probe: positive control passed, but the abstract mutation probe could not inject raw TLS 1.3 bytes and was not decisive for this issue.
- Follow-up focused test: added and ran `test_tls13_early_data_bad_record_mac` in wolfSSL's TLS 1.3 API test group.
- Test binary: `harness-build/cmake-mingw-examples-psk-compat/tests/unit.test.exe`.
- Relevant build features: `WOLFSSL_TLS13`, `WOLFSSL_EARLY_DATA`, and `HAVE_SESSION_TICKET` are enabled in `harness-build/cmake-mingw-examples-psk-compat/wolfssl/options.h`.
- Positive control inside test: the test first completes a TLS 1.3 handshake and obtains a session ticket with nonzero early-data allowance, then successfully calls `wolfSSL_write_early_data()` on the resumed client path. This verifies that the reproducer reached the real 0-RTT path rather than only a synthetic record parser.
- Mutation: the test locates the first encrypted client `application_data` record in memio and flips one byte at the start of that record body before the server reads it.
- Expected result: server returns fatal error and records/sends fatal `bad_record_mac` alert code `20`.
- Observed result from the saved run: `tests/api/test_tls13.c line 4891 failed`, `expected: h.last_tx.code == bad_record_mac`, `result: -1 != 20`. Exit code was `1`.
- Artifact: `tasks/cand-1a405ce4b2d3-replay/runtime-artifacts/unit-test.stdout.log`.

### Round 2

- Recheck command: from `D:\paper\idea\implementions\wolfssl-master`, with `D:\paper\idea\opt\runs\rfc8446-wolfssl\harness-build\cmake-mingw-examples-psk-compat` prepended to `PATH`, ran `unit.test.exe --api -test_tls13_early_data_bad_record_mac`.
- Recheck result: exit code `1`; the same assertion failed at the current source line `tests/api/test_tls13.c:5294`.
- Recheck stdout excerpt: `expected: h.last_tx.code == bad_record_mac`; `result: -1 != 20`.
- Interpretation: wolfSSL did return `WOLFSSL_FATAL_ERROR` for the corrupted accepted 0-RTT record, but the server's transmitted alert history remained unset instead of containing fatal `bad_record_mac`.
- Artifacts: `tasks/cand-1a405ce4b2d3-replay-recheck/runtime-artifacts/unit-test-rerun-current.stdout.log` and `tasks/cand-1a405ce4b2d3-replay-recheck/runtime-artifacts/summary.json`.

## Inconsistency Reason

RFC 8446 requires fatal `bad_record_mac` when a server fails to decrypt a 0-RTT record after accepting early data. wolfSSL's accepted early-data decrypt-failure path instead counts the record size, branches as ignored early data, advances the input buffer, and returns before sending `bad_record_mac`. The focused runtime test reproduced this behavior on the current build: the server reported a fatal error to the API but did not record a transmitted `bad_record_mac` alert.

## Decision Reason

Verified standard evidence, code evidence, and current runtime evidence establish a concrete divergence. The server acceptance path sets `extEarlyData->resp = 1` and `ssl->earlyData = process_early_data`; the record decrypt-failure path then returns through the early-data ignore branch before `SendAlert(ssl, alert_fatal, bad_record_mac)`. The focused unit test exercises that path with a real ticket-based 0-RTT resumption and observes `h.last_tx.code == -1` instead of `bad_record_mac` (`20`). Final verdict remains `issue_found`.

### LINKED ISSUE #11012: Supposed middlebox-compat client build still sends an empty legacy_session_id
# Supposed middlebox-compat client build still sends an empty legacy_session_id

## Problem Description

The `cmake-mingw-examples-psk-mbox` client artifact still emits a TLS 1.3 ClientHello whose `legacy_session_id` length is `0`. If this artifact is meant to implement RFC 8446 compatibility mode, that wire behavior is inconsistent with the RFC requirement to send a non-empty 32-byte value for a non-resumption client.

## Standard Requirement

- Section: `4.1.2 ClientHello`
- Evidence: [rfc8446.txt](/D:/paper/idea/document/rfc8446.txt:1586)

> In compatibility mode (see Appendix D.4),
> this field MUST be non-empty, so a client not offering a
> pre-TLS 1.3 session MUST generate a new 32-byte value.

Interpretation:

If a client build is operating in TLS 1.3 compatibility mode, its initial ClientHello must not carry a zero-length `legacy_session_id`.

## Relevant Source Code

The concrete serialization logic lives in
[src/tls13.c](/D:/paper/idea/implementions/wolfssl-master/src/tls13.c:4544).

### `src/tls13.c:4544-4580`

```c
static void GetTls13SessionId(WOLFSSL* ssl, byte* output, word32* idx)
{
    if (ssl->session->sessionIDSz > 0) {
        /* Session resumption for old versions of protocol. */
        ...
    }
    else {
    #ifdef WOLFSSL_TLS13_MIDDLEBOX_COMPAT
        if (ssl->options.tls13MiddleBoxCompat) {
            if (output != NULL)
                output[*idx] = ID_LEN;
            (*idx)++;
            if (output != NULL)
                XMEMCPY(output + *idx, ssl->arrays->clientRandom, ID_LEN);
            *idx += ID_LEN;
        }
        else
    #endif
        {
            /* TLS v1.3 does not use session id - 0 length. */
            if (output != NULL)
                output[*idx] = 0;
            (*idx)++;
        }
    }
}
```

This path is the direct decision point for whether the ClientHello carries a 32-byte `legacy_session_id` or a zero-length one.

### Capture 1: produced `mbox` example client

- Target: [client.exe](/D:/paper/idea/opt/runs/rfc8446-wolfssl/harness-build/cmake-mingw-examples-psk-mbox/examples/client/client.exe)
- Command: `client.exe -h 127.0.0.1 -p 12345 -v 4 -d`
- Harness: local TCP listener that accepted one connection and recorded the first TLS record from the client.

Observed result:

- Captured bytes: `513`
- `legacy_session_id` length byte at ClientHello offset `43`: `0`
- First bytes:

```text
16 03 03 01 fc 01 00 01 f8 03 03 6b c8 4b 19 f7 6c e4 c8 90 e8 b4 b9 55
a3 37 d4 ed 0f 14 c2 d0 54 d1 aa 2d 4c 23 f6 26 7a 8f 9a 00 00 36 ...
```

The byte after the 32-byte ClientHello random is `00`, so the emitted `legacy_session_id` vector is empty.

### Capture 2: minimal library-linked probe

- Probe source: [mbox_probe.c](/D:/paper/idea/opt/runs/rfc8446-wolfssl/tasks/report-0009-middlebox-probe/mbox_probe.c:1)
- Linked against: [libwolfssl.a](/D:/paper/idea/opt/runs/rfc8446-wolfssl/harness-build/cmake-mingw-tests-mbox-audit/libwolfssl.a)

Observed result:

```text
initial_middlebox=0
initial_dtls=0
initial_session_id_sz=0
connect_ret=-1
connect_err=2
post_middlebox=0
captured_sends=1
captured_len=1426
captured_sid_len=0
```

This independent probe reproduces the same observable behavior: the produced client-side library emits an empty `legacy_session_id`.

## Inconsistency Reason

The purported middlebox-compatible client artifact does not place a 32-byte `legacy_session_id` on the wire. Its first ClientHello uses a zero-length vector instead.

## Decision Reason

This is a concrete wire-level divergence in the produced `mbox` artifact. The artifact that is supposed to represent compatibility-mode behavior still sends an empty `legacy_session_id`.

### LINKED ISSUE #11013: Fresh-start deployments still accept 0-RTT while the replay window overlaps startup time
# Fresh-start deployments still accept 0-RTT while the replay window overlaps startup time

## Problem Description

wolfSSL's reachable TLS 1.3 0-RTT acceptance path does not implement the RFC 8446 Section 8.2 fresh-start overlap safeguard, and the behavior is observable at runtime. In two fresh-start scenarios that preserve the resumption material a deployment would realistically keep across restart, the restarted server still resumes the session and accepts early data immediately after startup instead of rejecting 0-RTT for the overlap period.

## Standard Requirement

- Section: `8.2 Client Hello Recording`
- Evidence: `RFC 8446, lines 5585-5587`

> When implementations are freshly started, they SHOULD reject 0-RTT as
>    long as any portion of their recording window overlaps the startup
>    time.

Interpretation:

When a server instance has just started, it should reject 0-RTT during any period where its anti-replay recording window still overlaps the startup time.

## Relevant Source Code

wolfSSL's 0-RTT admission path checks whether early data is possible, verifies ticket freshness, and enforces single-use eviction, but it does not consult any startup-time or replay-window-overlap state before accepting early data.

### `src/tls13.c:4615-4627`

```c
static int EarlyDataPossible(WOLFSSL* ssl)
{
    /* Need session resumption OR PSK callback configured */
    if (ssl->options.resuming) {
        return 1;
    }
#ifndef NO_PSK
    if (ssl->options.client_psk_tls13_cb != NULL ||
        ssl->options.client_psk_cb != NULL) {
        return 1;
    }
#endif
    return 0;
}
```

This entry point gates only on resumption state or PSK callback presence.

### `src/internal.c:41604-41617`

```c
        word32 ticketSeen;        /* Time ticket seen (ms) */

        ato32(psk->it->timestamp, &ticketSeen);

        now = TimeNowInMilliseconds();
        if (now == 0)
            return GETTIME_ERROR;
        /* Difference between now and time ticket constructed
         * (from decrypted ticket). */
        diff = now;
        diff -= ticketSeen;
        if (diff > timeout * 1000 ||
            diff > (sword64)TLS13_MAX_TICKET_AGE * 1000)
            return WOLFSSL_FATAL_ERROR;
```

The ticket freshness check compares current time to ticket construction time, but it does not compare startup time to the anti-replay recording window.

### `src/tls13.c:6908-6923`

```c
            /* Check if accepting early data and first PSK.
             * RFC 9973: early_data is not compatible with
             * cert_with_extern_psk, so skip key derivation in that case. */
            if (ssl->earlyData != no_early_data && first
                && ssl->options.maxEarlyDataSz > 0
    #ifdef WOLFSSL_CERT_WITH_EXTERN_PSK
                && !hasCertWithExternPsk
    #endif
    #if defined(HAVE_SESSION_TICKET) && !defined(NO_SESSION_CACHE)
                /* RFC 8446 section 8: evict the session from the cache.
                 * Accept 0-RTT only when the eviction found the entry
                 * (single-use). */
                && wolfSSL_SSL_CTX_remove_session(ssl->ctx, ssl->session)
                    == 1
    #endif
            ) {
```

Actual 0-RTT acceptance depends on early-data enablement and single-use session eviction, with no startup-overlap guard in the acceptance condition.

## Runtime Evidence

A focused memio-based reproducer was built on top of wolfSSL's test harness to:

1. complete an initial TLS 1.3 handshake and obtain a 0-RTT-capable session;
2. destroy the original server object and construct a fresh server context;
3. preserve the resumption material that a real deployment could keep across restart;
4. immediately attempt resumption with early data on the restarted server.

`wolfSSL_get_early_data_status()` returns `2` for `WOLFSSL_EARLY_DATA_ACCEPTED`.

### Scenario 1: stateful ticket plus external session cache preserved across restart

- Fresh-start action: destroy the original server `WOLFSSL_CTX` and `WOLFSSL`, then create a new server context
- Persisted state: external session cache callbacks still retain the resumable session
- Observed output:

```text
stateful_restart: resumed=1 earlyWritten=26 earlyRead=0 earlyRet1=26 earlyRet2=0 clientStatus=2 serverStatus=2 cache(new=1 get=1 rem=1 has=0)
```

Interpretation:

- `resumed=1`: the restarted server resumed the session
- `earlyRet1=26`: the first server-side `wolfSSL_read_early_data()` consumed all 26 bytes of early data
- `clientStatus=2` and `serverStatus=2`: both sides report early data accepted

### Scenario 2: stateless ticket keys restored into a fresh server context

- Fresh-start action: destroy the original server context and create a new one
- Persisted state: export ticket keys with `wolfSSL_CTX_get_tlsext_ticket_keys()` and import them into the fresh context with `wolfSSL_CTX_set_tlsext_ticket_keys()`
- Observed output:

```text
stateless_restart_keys: resumed=1 earlyWritten=27 earlyRead=0 earlyRet1=27 earlyRet2=0 clientStatus=2 serverStatus=2
```

Interpretation:

- `resumed=1`: the restarted server resumed the session
- `earlyRet1=27`: the first server-side `wolfSSL_read_early_data()` consumed all 27 bytes of early data
- `clientStatus=2` and `serverStatus=2`: both sides report early data accepted

The final `earlyRead=0` in both lines is not a rejection signal. It reflects the second `wolfSSL_read_early_data()` call after the first one had already consumed the available early data.

## Inconsistency Reason

RFC 8446 says a freshly started implementation should reject 0-RTT while its anti-replay recording window overlaps startup time. wolfSSL instead resumes the session and accepts early data immediately after startup when the deployment preserves either the external session cache state or the session-ticket encryption keys across restart.

## Decision Reason

The source path lacks any startup-overlap check, and the focused runtime reproducer confirms that a freshly restarted server can still accept 0-RTT in realistic restart configurations. That is sufficient to upgrade this record from `suspected_issue` to `issue_found`.

### LINKED ISSUE #11014: Generated options.h advertises WOLFSSL_TLS13_MIDDLEBOX_COMPAT, but libwolfssl is built without it
# Generated options.h advertises WOLFSSL_TLS13_MIDDLEBOX_COMPAT, but libwolfssl is built without it

## Problem Description

The CMake build can generate a public `wolfssl/options.h` that defines `WOLFSSL_TLS13_MIDDLEBOX_COMPAT`, while the actual `libwolfssl` source files are compiled without that macro in their compile definitions. As a result, the publicly advertised configuration and the library's real compiled behavior diverge.

### The cache accepts the variable

In the produced `mbox` build cache (`CMakeCache.txt`, line 1):

```text
WOLFSSL_JNI:BOOL=no
WOLFSSL_TLS13_MIDDLEBOX_COMPAT:UNINITIALIZED=yes
```

So the cache contains a value for `WOLFSSL_TLS13_MIDDLEBOX_COMPAT`, but it is not a normal declared option.

### CMakeLists does not declare an independent option

The only place where `WOLFSSL_TLS13_MIDDLEBOX_COMPAT` is appended to `WOLFSSL_DEFINITIONS` is inside the JNI bundle branch in `CMakeLists.txt:390`:

```cmake
if(WOLFSSL_JNI)
    ...
    list(APPEND WOLFSSL_DEFINITIONS
        ...
        "-DWOLFSSL_TLS13_MIDDLEBOX_COMPAT")
endif()
```

There is no standalone `add_option("WOLFSSL_TLS13_MIDDLEBOX_COMPAT" ...)` declaration.

### Generated public header says the macro is enabled

The generated public header `wolfssl/options.h:782` contains:

```c
#undef WOLFSSL_TLS13_MIDDLEBOX_COMPAT
#define WOLFSSL_TLS13_MIDDLEBOX_COMPAT
```

That means downstream code including `wolfssl/options.h` sees the feature as enabled.

### The actual library compile flags do not contain the macro

The real compile definitions for `libwolfssl` in `CMakeFiles/wolfssl.dir/flags.make:5` do not include `-DWOLFSSL_TLS13_MIDDLEBOX_COMPAT`.

The same omission is visible in the separate audit build's `CMakeFiles/wolfssl.dir/flags.make:5`.

## Source-Inclusion Evidence

wolfSSL library sources enter through `libwolfssl_sources.h:36`, which pulls in `config.h` when `HAVE_CONFIG_H` is set:

```c
#if defined(HAVE_CONFIG_H) && !defined(WC_CONFIG_H_INCLUDED)
    #include <config.h>
    #define WC_CONFIG_H_INCLUDED
#endif
```

The library sources therefore follow `config.h` plus compile definitions during their build, not the generated public `wolfssl/options.h`.

In the audit build's `config.h:1`, `WOLFSSL_TLS13_MIDDLEBOX_COMPAT` is absent.

## Preprocessor Cross-Check

I verified the mismatch directly:

- Preprocessing with `#include <wolfssl/options.h>` shows
  `#define WOLFSSL_TLS13_MIDDLEBOX_COMPAT`.
- Preprocessing with `#include <wolfssl/wolfcrypt/libwolfssl_sources.h>` in the same build environment does not show that macro.

So the generated public header and the library-source build view disagree.

## Inconsistency Reason

The build system propagates `WOLFSSL_TLS13_MIDDLEBOX_COMPAT` into the generated public header, but not into the compile definitions that build `libwolfssl`.

## Decision Reason

This is a real build/configuration bug. It creates a configuration mismatch between what the build artifacts advertise and what the compiled library actually implements. That mismatch directly explains why the produced `mbox` artifact does not exhibit compatibility-mode wire behavior.

julek-wolfssl:
- Terminate the connection with a fatal `bad_record_mac` alert when a 0-RTT
  record fails to decrypt after the server accepted early data, per RFC 8446
  Section 4.2.10. Undecryptable records are only skipped when early data was
  rejected. Fixes #11011.
- Declare `WOLFSSL_TLS13_MIDDLEBOX_COMPAT` as a real CMake option and add
  `--enable-tls13-middlebox-compat` to configure. Previously CMake accepted
  `-DWOLFSSL_TLS13_MIDDLEBOX_COMPAT=yes` and advertised the macro in the
  generated `options.h` but never passed it to the library compile, so such
  a build sent an empty `legacy_session_id` instead of the non-empty 32-byte
  value RFC 8446 Section 4.1.2 requires in compatibility mode. Fixes #11012,
  #11014.
- Reject 0-RTT (but not resumption) for session tickets minted before the
  current server ctx was created, per the RFC 8446 Section 8.2 recommendation
  that freshly started servers reject 0-RTT while their anti-replay state
  cannot cover the ticket. This protects restarted servers that keep their
  session cache or ticket encryption keys from accepting early data they
  cannot check for replays. Deployments whose anti-replay state reliably
  survives restarts can opt out with
  `wolfSSL_CTX_no_early_data_fresh_start_check()`. Fixes #11013.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -725,6 +725,21 @@ if("${WOLFSSL_HRR_COOKIE}" STREQUAL "yes")
     endif()
 endif()
 
+# TLS v1.3 middlebox compatibility (RFC 8446 Appendix D.4)
+add_option("WOLFSSL_TLS13_MIDDLEBOX_COMPAT"
+    "Enable TLS v1.3 middlebox compatibility mode (default: disabled)"
+    "no" "yes;no")
+
+if(WOLFSSL_TLS13_MIDDLEBOX_COMPAT)
+    if(NOT WOLFSSL_TLS13)
+        message(WARNING "TLS 1.3 is disabled - disabling middlebox compatibility")
+        override_cache(WOLFSSL_TLS13_MIDDLEBOX_COMPAT "no")
+    else()
+        list(APPEND WOLFSSL_DEFINITIONS
+            "-DWOLFSSL_TLS13_MIDDLEBOX_COMPAT")
+    endif()
+endif()
+
 # DTLS v1.3
 add_option("WOLFSSL_DTLS13"
     "Enable wolfSSL DTLS v1.3 (default: disabled)"
@@ -3063,6 +3078,13 @@ if(WOLFSSL_ECH)
         message(FATAL_ERROR "ECH supported only with SNI (WOLFSSL_SNI)")
     endif()
     list(APPEND WOLFSSL_DEFINITIONS "-DHAVE_ECH")
+    if(WOLFSSL_TLS13_MIDDLEBOX_COMPAT)
+        message(WARNING
+            "ECH is incompatible with middlebox compatibility - disabling middlebox compatibility")
+        override_cache(WOLFSSL_TLS13_MIDDLEBOX_COMPAT "no")
+        list(REMOVE_ITEM WOLFSSL_DEFINITIONS
+            "-DWOLFSSL_TLS13_MIDDLEBOX_COMPAT")
+    endif()
 endif()
 
 if(WOLFSSL_KEYING_MATERIAL)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -706,6 +706,21 @@ if("${WOLFSSL_HRR_COOKIE}" STREQUAL "yes")
     endif()
 endif()
 
+# TLS v1.3 middlebox compatibility (RFC 8446 Appendix D.4)
+add_option("WOLFSSL_TLS13_MIDDLEBOX_COMPAT"
+    "Enable TLS v1.3 middlebox compatibility mode (default: disabled)"
+    "no" "yes;no")
+
+if(WOLFSSL_TLS13_MIDDLEBOX_COMPAT)
+    if(NOT WOLFSSL_TLS13)
+        message(WARNING "TLS 1.3 is disabled - disabling middlebox compatibility")
+        override_cache(WOLFSSL_TLS13_MIDDLEBOX_COMPAT "no")
+    else()
+        list(APPEND WOLFSSL_DEFINITIONS
+            "-DWOLFSSL_TLS13_MIDDLEBOX_COMPAT")
+    endif()
+endif()
+
 # DTLS v1.3
 add_option("WOLFSSL_DTLS13"
     "Enable wolfSSL DTLS v1.3 (default: disabled)"
@@ -2982,6 +2997,13 @@ if(WOLFSSL_ECH)
         message(FATAL_ERROR "ECH supported only with SNI (WOLFSSL_SNI)")
     endif()
     list(APPEND WOLFSSL_DEFINITIONS "-DHAVE_ECH")
+    if(WOLFSSL_TLS13_MIDDLEBOX_COMPAT)
+        message(WARNING
+            "ECH is incompatible with middlebox compatibility - disabling middlebox compatibility")
+        override_cache(WOLFSSL_TLS13_MIDDLEBOX_COMPAT "no")
+        list(REMOVE_ITEM WOLFSSL_DEFINITIONS
+            "-DWOLFSSL_TLS13_MIDDLEBOX_COMPAT")
+    endif()
 endif()
 
 if(WOLFSSL_KEYING_MATERIAL)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -738,6 +738,17 @@ if(WOLFSSL_TLS13_MIDDLEBOX_COMPAT)
         list(APPEND WOLFSSL_DEFINITIONS
             "-DWOLFSSL_TLS13_MIDDLEBOX_COMPAT")
     endif()
+elseif("-DWOLFSSL_TLS13_MIDDLEBOX_COMPAT" IN_LIST WOLFSSL_DEFINITIONS)
+    # A bundle (WOLFSSL_JNI) appends the define straight to the list. Sync the
+    # option with it, otherwise the conflict checks below - ECH - test a
+    # variable that is still "no" and leave both features enabled.
+    if(WOLFSSL_TLS13)
+        override_cache(WOLFSSL_TLS13_MIDDLEBOX_COMPAT "yes")
+    else()
+        message(WARNING "TLS 1.3 is disabled - disabling middlebox compatibility")
+        list(REMOVE_ITEM WOLFSSL_DEFINITIONS
+            "-DWOLFSSL_TLS13_MIDDLEBOX_COMPAT")
+    endif()
 endif()
 
 # DTLS v1.3
--- a/configure.ac
+++ b/configure.ac
@@ -9645,7 +9645,12 @@ then
     AM_CFLAGS="$AM_CFLAGS -DKEEP_PEER_CERT"
     AM_CFLAGS="$AM_CFLAGS -DWOLFSSL_ALWAYS_VERIFY_CB"
     AM_CFLAGS="$AM_CFLAGS -DWOLFSSL_ALWAYS_KEEP_SNI"
-    AM_CFLAGS="$AM_CFLAGS -DWOLFSSL_TLS13_MIDDLEBOX_COMPAT"
+    if test "x$ENABLED_ECH" = "xyes"
+    then
+        AC_MSG_NOTICE([ECH is incompatible with middlebox compatibility - not enabling middlebox compatibility for JNI])
+    else
+        AM_CFLAGS="$AM_CFLAGS -DWOLFSSL_TLS13_MIDDLEBOX_COMPAT"
+    fi
     AM_CFLAGS="$AM_CFLAGS -DWOLFSSL_PUBLIC_MP"
 
     # Enable openssl compat layer AES-CTS to maintain FIPS compatibility
--- a/configure.ac
+++ b/configure.ac
@@ -2558,6 +2558,28 @@ then
 fi
 
 
+# TLS v1.3 middlebox compatibility (RFC 8446 Appendix D.4)
+AC_ARG_ENABLE([tls13-middlebox-compat],
+    [AS_HELP_STRING([--enable-tls13-middlebox-compat],[Enable TLS v1.3 middlebox compatibility mode (default: disabled)])],
+    [ ENABLED_TLS13_MIDDLEBOX_COMPAT=$enableval ],
+    [ ENABLED_TLS13_MIDDLEBOX_COMPAT=no ]
+    )
+if test "$ENABLED_TLS13_MIDDLEBOX_COMPAT" = "yes"
+then
+    if test "x$ENABLED_TLS13" = "xno"
+    then
+        AC_MSG_NOTICE([TLS 1.3 is disabled - disabling middlebox compatibility])
+        ENABLED_TLS13_MIDDLEBOX_COMPAT="no"
+    elif test "x$ENABLED_ECH" = "xyes"
+    then
+        AC_MSG_NOTICE([ECH is incompatible with middlebox compatibility - disabling middlebox compatibility])
+        ENABLED_TLS13_MIDDLEBOX_COMPAT="no"
+    else
+        AM_CFLAGS="$AM_CFLAGS -DWOLFSSL_TLS13_MIDDLEBOX_COMPAT"
+    fi
+fi
+
+
 # RNG
 AC_ARG_ENABLE([rng],
     [AS_HELP_STRING([--enable-rng],[Enable compiling and using RNG (default: enabled)])],
--- a/configure.ac
+++ b/configure.ac
@@ -2696,6 +2696,28 @@ then
 fi
 
 
+# TLS v1.3 middlebox compatibility (RFC 8446 Appendix D.4)
+AC_ARG_ENABLE([tls13-middlebox-compat],
+    [AS_HELP_STRING([--enable-tls13-middlebox-compat],[Enable TLS v1.3 middlebox compatibility mode (default: disabled)])],
+    [ ENABLED_TLS13_MIDDLEBOX_COMPAT=$enableval ],
+    [ ENABLED_TLS13_MIDDLEBOX_COMPAT=no ]
+    )
+if test "$ENABLED_TLS13_MIDDLEBOX_COMPAT" = "yes"
+then
+    if test "x$ENABLED_TLS13" = "xno"
+    then
+        AC_MSG_NOTICE([TLS 1.3 is disabled - disabling middlebox compatibility])
+        ENABLED_TLS13_MIDDLEBOX_COMPAT="no"
+    elif test "x$ENABLED_ECH" = "xyes"
+    then
+        AC_MSG_NOTICE([ECH is incompatible with middlebox compatibility - disabling middlebox compatibility])
+        ENABLED_TLS13_MIDDLEBOX_COMPAT="no"
+    else
+        AM_CFLAGS="$AM_CFLAGS -DWOLFSSL_TLS13_MIDDLEBOX_COMPAT"
+    fi
+fi
+
+
 # RNG
 AC_ARG_ENABLE([rng],
     [AS_HELP_STRING([--enable-rng],[Enable compiling and using RNG (default: enabled)])],
--- a/doc/dox_comments/header_files/ssl.h
+++ b/doc/dox_comments/header_files/ssl.h
@@ -15720,6 +15720,14 @@ int  wolfSSL_read_early_data(WOLFSSL* ssl, void* data, int sz,
     call this function when the anti-replay state (session cache or
     external cache) reliably survives server restarts.
 
+    The check compares the ticket timestamp against the context creation
+    time, both taken from TimeNowInMilliseconds(). That clock is only
+    required to be millisecond accurate, not correlated to the epoch, so on
+    ports where it counts from boot (Windows QPC, Zephyr, FreeRTOS, Micrium,
+    Microchip) it restarts near zero and the check does not fire for tickets
+    minted before a reboot. Such deployments must rely on ticket key
+    rotation instead.
+
     \param [in,out] ctx a pointer to a WOLFSSL_CTX structure, created
     with wolfSSL_CTX_new().
 
--- a/doc/dox_comments/header_files/ssl.h
+++ b/doc/dox_comments/header_files/ssl.h
@@ -15319,6 +15319,40 @@ int  wolfSSL_read_early_data(WOLFSSL* ssl, void* data, int sz,
     int* outSz);
 
 /*!
+    \ingroup Setup
+
+    \brief This function is called on the server to disable the
+    RFC 8446 Section 8.2 fresh start protection. By default a freshly
+    created context rejects early data, but not resumption, for session
+    tickets minted before the context was created, since the anti-replay
+    state for those tickets may not have survived a server restart. Only
+    call this function when the anti-replay state (session cache or
+    external cache) reliably survives server restarts.
+
+    \param [in,out] ctx a pointer to a WOLFSSL_CTX structure, created
+    with wolfSSL_CTX_new().
+
+    \return BAD_FUNC_ARG if ctx is NULL or not using TLS v1.3.
+    \return SIDE_ERROR if called with a client.
+    \return 0 if successful.
+
+    _Example_
+    \code
+    int ret;
+    WOLFSSL_CTX* ctx;
+    ...
+    ret = wolfSSL_CTX_no_early_data_fresh_start_check(ctx);
+    if (ret != 0) {
+        // failed to disable the fresh start check
+    }
+    \endcode
+
+    \sa wolfSSL_CTX_set_max_early_data
+    \sa wolfSSL_read_early_data
+*/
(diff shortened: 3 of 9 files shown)


julek-wolfssl:
retest this please

philljj (changes requested):
This needs a rebase + force-push to unblock the PRB.

Also, the Fenrir mediums look valid.

philljj (changes requested):
There are a few more WOLFSSL_32BIT_MILLI_TIME corner cases to fix (see fenrir above and test below).

This test fails

```
./tests/unit.test -test_tls13_0rtt_fresh_start
starting unit tests...
 Begin API Tests
  Begin Group: tls13
   1636: test_tls13_0rtt_fresh_start                        :
ERROR - tests/api/test_tls13.c line 5585 failed with:
    expected: earlyRead == 0
    result:   17 != 0

 failed (  0.02133)
```

found with:

```
./configure CFLAGS="-DWOLFSSL_32BIT_MILLI_TIME" \
  --enable-earlydata --enable-session-ticket  && make && make check
```

philljj (approved):

PR #11100, opened 2026-08-06
Enforce RFC 8446 SHA-1 certificate rule on TLS 1.3 chains

embhorn:
# Description

`signature_algorithms_cert` was parsed into `ssl->certHashSigAlgo` but never read again, so a TLS 1.3 peer sent whatever chain was configured regardless of what the other side advertised.

RFC 8446 Section 4.4.2.2 allows a fallback chain the peer did not advertise support for, but the chain MUST NOT use SHA-1 unless the peer's advertisement permits it. Section 4.4.2.3 applies the same rule to the client. Only that requirement is enforced; a chain that merely fails to match is still sent, as the RFC intends.

Fixes zd22296

# Testing

1. SHA-1 leaf, client offers no SHA-1        fails, MATCH_SUITE_ERROR
2. same leaf, client offers rsa_pkcs1_sha1   succeeds
3. SHA-256 chain, default client             succeeds
4. SHA-256 leaf under self signed SHA-1 root succeeds
5. SHA-1 client chain, request omits SHA-1   succeeds, SEND_BLANK_CERT
6. same chain, request offers SHA-1          succeeds, SEND_CERT

# Checklist

 - [x] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/certs/check_cert_chains.sh
+++ b/certs/check_cert_chains.sh
@@ -5,24 +5,30 @@
 # rather than as a runtime TLS handshake failure. Two checks per pair:
 #   1. identity: leaf issuer DN == CA subject DN and leaf AKID == CA SKID; runs
 #      on any OpenSSL and catches the DN drift that has actually broken CI.
-#   2. crypto: openssl verify, only when OpenSSL supports the CA's algorithm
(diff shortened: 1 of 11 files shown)


embhorn:
retest this please

embhorn:
retest this please

PR #11101, opened 2026-08-06
Add separate client cookie enable config

### LINKED ISSUE #11074: TLS 1.3 cookie extension is absent in the audited default build and only available behind build/runtime gates
# TLS 1.3 cookie extension is absent in the audited default build and only available behind build/runtime gates

## Executive Summary

The earlier report stopped at a source-backed suspicion. A focused runtime recheck on August 3, 2026 resolves that uncertainty.

For the audited default wolfSSL TLS 1.3 build, the `cookie` extension is not actually implemented on the HelloRetryRequest path: a forced-HRR probe reached a real `HelloRetryRequest`, but the HRR carried no `cookie` extension. The source tree does contain working HRR-cookie support, but that support is guarded by `WOLFSSL_SEND_HRR_COOKIE`, requires a build configured with `WOLFSSL_HRR_COOKIE=yes`, and still requires the server to opt in at runtime with `wolfSSL_send_hrr_cookie()`.

So this is a real issue for the audited default build, not a proof gap. The implementation exists in the source tree, but only behind build/runtime gates rather than as an unconditional TLS 1.3 capability in the checked target.

## Standard Requirement

- RFC: [RFC 8446](https://www.rfc-editor.org/rfc/rfc8446.html)
- Section: `9.2 Mandatory-to-Implement Extensions`
- Relevant lines: `5715-5721`

Standard basis:

> In the absence of an application profile standard specifying
>    otherwise, a TLS-compliant application MUST implement the following
>    TLS extensions:
>
>    -  Supported Versions ("supported_versions"; Section 4.2.1)
>
>    -  Cookie ("cookie"; Section 4.2.2)

See [RFC 8446 Section 9.2](https://www.rfc-editor.org/rfc/rfc8446.html#section-9.2).

Interpretation:

- In the audited target, absent an application-profile exception, TLS 1.3 support must include the `cookie` extension capability.
- Because RFC 8446 Section 4.2.2 defines `cookie` for the HelloRetryRequest / second-ClientHello flow, a valid way to test implementation is to force a real HRR and observe whether the stack can actually produce and process that extension.

## Relevant Source Code

The source tree itself already shows that HRR-cookie support is not unconditional.

### 1. wolfSSL documents HRR cookie support as default-off

`src/tls13.c:38` explicitly describes the feature gate:

```c
 * WOLFSSL_SEND_HRR_COOKIE:  Send cookie in HelloRetryRequest     default: off
```

### 2. TLS 1.3 ClientHello extension construction only enables `cookie` under `WOLFSSL_SEND_HRR_COOKIE`

`src/tls.c:17485` and `src/tls.c:17720` only turn on `TLSX_COOKIE` when the macro is compiled in:

```c
#ifdef WOLFSSL_SEND_HRR_COOKIE
    TURN_ON(semaphore, TLSX_ToSemaphore(TLSX_COOKIE));
#endif
```

That means the TLS 1.3 request-size and request-write paths do not treat `cookie` as an unconditional extension in the audited build.

### 3. The server-side runtime opt-in API itself only exists inside the same macro gate

`src/tls13.c:14865` defines `wolfSSL_send_hrr_cookie()` only under `WOLFSSL_SEND_HRR_COOKIE`, and the function merely enables the behavior for a particular `WOLFSSL*`:

```c
int wolfSSL_send_hrr_cookie(WOLFSSL* ssl, const unsigned char* secret,
                            unsigned int secretSz)
{
    ...
    ssl->options.sendCookie = 1;
    ...
}
```

So even a build that compiles the code still needs explicit runtime opt-in before a server will emit an HRR `cookie`.

### Recheck goal

Resolve the old uncertainty by separating three questions:

1. Does the audited default build emit a real HelloRetryRequest without `cookie`?
2. Does a build compiled with `WOLFSSL_HRR_COOKIE=yes` still omit `cookie` unless explicitly enabled at runtime?
3. If explicitly enabled, does the rebuilt path actually emit `cookie` and pass the existing bad-cookie regression test?

### Recheck method

The test used a native in-memory TLS 1.3 client/server probe in three configurations: the audited default build, a build configured with `WOLFSSL_HRR_COOKIE=yes` without runtime opt-in, and the same enabled build after calling `wolfSSL_send_hrr_cookie()`. It forced HelloRetryRequest by sending the initial ClientHello without a key share, parsed the first server handshake message, and checked for extension type `0x002c` (`cookie`). The enabled configuration was also exercised with the existing bad-cookie regression case.

The probe forces HRR by making the client send a TLS 1.3 ClientHello with no key shares. It then inspects the first server flight and reports whether that flight is really a HelloRetryRequest and whether the HRR carries extension type `0x002c` (`cookie`).

### 1. Audited default build: real HRR, no `cookie`

Build-side macro evidence for the audited default build:

- `wolfssl/options.h:345` leaves `WOLFSSL_SEND_HRR_COOKIE` undefined.
- `CMakeCache.txt:740` records `WOLFSSL_HRR_COOKIE:STRING=undefined`.

Observed runtime output:

```text
compile_macro=off
server_first_message_is_hrr=1
cookie_present=0
```

The full log also shows the handshake really did reach the HRR point:

```text
no_key_shares_ret=1
client_ch1 ret=-1 err=2 s_msg_count=1 c_msg_count=0
server_hrr ret=-1 err=2 s_msg_count=0 c_msg_count=1
```

This is the decisive runtime confirmation that the audited default build can produce a real HelloRetryRequest but does not implement the `cookie` extension there.

### 2. Rebuilt `WOLFSSL_HRR_COOKIE=yes` build: still no `cookie` without runtime opt-in

The separate recheck build was configured with `-DWOLFSSL_TLS13=yes -DWOLFSSL_HRR_COOKIE=yes`. Configuration completed successfully, and a preprocessor check reported `WOLFSSL_SEND_HRR_COOKIE` as defined before the probe was run.

When the same probe runs against that rebuilt tree without calling `wolfSSL_send_hrr_cookie()`, the output is:

```text
compile_macro=on
enable_cookie_request=0
server_first_message_is_hrr=1
cookie_present=0
```

So compiling the feature in is still not enough by itself. The server must also enable it per-connection.

### 3. Rebuilt `WOLFSSL_HRR_COOKIE=yes` build with runtime opt-in: `cookie` appears

When the probe calls `wolfSSL_send_hrr_cookie()` before the handshake, the same rebuilt tree produces:

```text
send_hrr_cookie_ret=1
compile_macro=on
enable_cookie_request=1
server_first_message_is_hrr=1
cookie_present=1
cookie_ext_len=87
cookie_vector_len=85
```

This shows that the source tree does have a working HRR-cookie implementation, but only when both of these conditions hold:

- the build is compiled with `WOLFSSL_HRR_COOKIE=yes`, and
- the server explicitly calls `wolfSSL_send_hrr_cookie()`.

### 4. Existing bad-cookie regression test passes in the enabled build

The rebuilt enabled tree also passes the existing unit test `test_tls13_hrr_bad_cookie`:

```text
1528: test_tls13_hrr_bad_cookie : passed
Failed/Skipped/Passed/All: 0/0/1/1
```

That matters because it rules out a weaker alternative explanation such as "the feature exists only as dead code". Once compiled in and enabled, the path is alive and the implementation rejects corrupted HRR cookies.

## Why This Is a Real Issue

This is not just a documentation mismatch or a source-only caution.

The runtime recheck establishes all of the following:

- The audited default build really reaches a TLS 1.3 HelloRetryRequest path.
- In that audited build, the HRR does not contain `cookie`.
- The absence is explained by a real build/runtime gate, not by the absence of HRR traffic.
- A separate build can emit `cookie`, but only when explicitly rebuilt and opt-in enabled.

That means the checked default target does not satisfy the mandatory-to-implement `cookie` extension requirement as audited here. The problem is build/configuration-scoped, but it is still real for the audited default build.

## Decision Reason

- Source inspection showed the `cookie` extension was controlled by `WOLFSSL_SEND_HRR_COOKIE`.
- The previous report could not tell whether that was only a source-level appearance or an actual capability gap in the audited target.
- The new forced-HRR runtime probe resolves that uncertainty: the audited default build produces a real HRR with no `cookie`, while a separate `WOLFSSL_HRR_COOKIE=yes` build can produce `cookie` only after explicit runtime opt-in.

Therefore this root cause should no longer remain `suspected_issue`. It is a confirmed `issue_found` for the audited default wolfSSL TLS 1.3 build.

## Scope and Remaining Uncertainty

Confirmed scope:

- This report is runtime-confirmed for the audited default wolfSSL TLS 1.3 example build.
- The confirmed mismatch is that this audited build does not implement the TLS 1.3 `cookie` extension capability on the HRR path.

Not claimed here:

- This report does not claim that every possible wolfSSL build is missing the feature.
- The recheck directly shows the opposite: a separately rebuilt tree with `WOLFSSL_HRR_COOKIE=yes` and explicit `wolfSSL_send_hrr_cookie()` opt-in does implement and exercise the feature.

Remaining uncertainty:

- No material uncertainty remains for the audited default build.

padelsbach:
# Description

Split the existing flag which enables TLS cookie support `WOLFSSL_SEND_HRR_COOKIE` into a smaller portion `WOLFSSL_TLS13_COOKIE` which enables *replying* with a cookie echo'ed from the HelloClientResponse (HRR). This is enabled by default for clients.

Fixes #11074 

# Testing

New unit test

# Checklist

 - [x] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/src/tls.c
+++ b/src/tls.c
@@ -7546,7 +7546,7 @@ static int TLSX_SetSupportedVersions(TLSX** extensions, const void* data,
 
 #endif /* WOLFSSL_TLS13 */
 
-#if defined(WOLFSSL_TLS13) && defined(WOLFSSL_SEND_HRR_COOKIE)
+#ifdef WOLFSSL_TLS13_COOKIE
 
 /******************************************************************************/
 /* Cookie                                                                     */
@@ -7622,8 +7622,6 @@ static int TLSX_Cookie_Parse(WOLFSSL* ssl, const byte* input, word16 length,
 {
     word16  len;
     word16  idx = 0;
-    TLSX*   extension;
-    Cookie* cookie;
 
     if (msgType != client_hello && msgType != hello_retry_request) {
         WOLFSSL_ERROR_VERBOSE(SANITY_MSG_E);
@@ -7646,32 +7644,40 @@ static int TLSX_Cookie_Parse(WOLFSSL* ssl, const byte* input, word16 length,
                                &ssl->extensions);
     }
 
-    /* client_hello */
-    extension = TLSX_Find(ssl->extensions, TLSX_COOKIE);
-    if (extension == NULL) {
-#ifdef WOLFSSL_DTLS13
-        if (ssl->options.dtls && IsAtLeastTLSv1_3(ssl->version))
-            /* Allow a cookie extension with DTLS 1.3 because it is possible
-             * that a different SSL instance sent the cookie but we are now
-             * receiving it. */
-            return TLSX_Cookie_Use(ssl, input + idx, len, NULL, 0, 0,
-                                   &ssl->extensions);
-        else
-#endif
-        {
+#ifdef WOLFSSL_SEND_HRR_COOKIE
+    /* client_hello - only a server that sends cookies has one to check the
+     * echoed cookie against. Otherwise ignore it. */
+    {
+        TLSX*   extension = TLSX_Find(ssl->extensions, TLSX_COOKIE);
+        Cookie* cookie;
+
+        if (extension == NULL) {
+    #ifdef WOLFSSL_DTLS13
+            if (ssl->options.dtls && IsAtLeastTLSv1_3(ssl->version))
+                /* Allow a cookie extension with DTLS 1.3 because it is
+                 * possible that a different SSL instance sent the cookie but
+                 * we are now receiving it. */
+                return TLSX_Cookie_Use(ssl, input + idx, len, NULL, 0, 0,
+                                       &ssl->extensions);
+            else
+    #endif
+            {
+                WOLFSSL_ERROR_VERBOSE(HRR_COOKIE_ERROR);
+                return HRR_COOKIE_ERROR;
+            }
+        }
+
+        cookie = (Cookie*)extension->data;
+        if (cookie->len != len ||
+                XMEMCMP(cookie->data, input + idx, len) != 0) {
             WOLFSSL_ERROR_VERBOSE(HRR_COOKIE_ERROR);
             return HRR_COOKIE_ERROR;
         }
-    }
 
-    cookie = (Cookie*)extension->data;
-    if (cookie->len != len || XMEMCMP(cookie->data, input + idx, len) != 0) {
-        WOLFSSL_ERROR_VERBOSE(HRR_COOKIE_ERROR);
-        return HRR_COOKIE_ERROR;
+        /* Request seen. */
+        extension->resp = 0;
     }
-
-    /* Request seen. */
-    extension->resp = 0;
+#endif
 
     return 0;
 }
@@ -17603,7 +17609,7 @@ int TLSX_GetRequestSize(WOLFSSL* ssl, byte msgType, word32* pLength)
         #ifdef WOLFSSL_EARLY_DATA
             TURN_ON(semaphore, TLSX_ToSemaphore(TLSX_EARLY_DATA));
         #endif
-        #ifdef WOLFSSL_SEND_HRR_COOKIE
+        #ifdef WOLFSSL_TLS13_COOKIE
             TURN_ON(semaphore, TLSX_ToSemaphore(TLSX_COOKIE));
         #endif
         #ifdef WOLFSSL_POST_HANDSHAKE_AUTH
@@ -17839,7 +17845,7 @@ int TLSX_WriteRequest(WOLFSSL* ssl, byte* output, byte msgType, word32* pOffset)
         #ifdef WOLFSSL_EARLY_DATA
             TURN_ON(semaphore, TLSX_ToSemaphore(TLSX_EARLY_DATA));
         #endif
-        #ifdef WOLFSSL_SEND_HRR_COOKIE
+        #ifdef WOLFSSL_TLS13_COOKIE
             TURN_ON(semaphore, TLSX_ToSemaphore(TLSX_COOKIE));
         #endif
         #ifdef WOLFSSL_POST_HANDSHAKE_AUTH
--- a/src/tls.c
+++ b/src/tls.c
@@ -7622,6 +7622,10 @@ static int TLSX_Cookie_Parse(WOLFSSL* ssl, const byte* input, word16 length,
 {
     word16  len;
     word16  idx = 0;
+#ifdef WOLFSSL_SEND_HRR_COOKIE
+    TLSX*   extension;
+    Cookie* cookie;
+#endif
 
     if (msgType != client_hello && msgType != hello_retry_request) {
         WOLFSSL_ERROR_VERBOSE(SANITY_MSG_E);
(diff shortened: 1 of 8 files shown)


src/tls.c
```diff
@@ -7507,7 +7507,7 @@ static int TLSX_SetSupportedVersions(TLSX** extensions, const void* data,
 
 #endif /* WOLFSSL_TLS13 */
 
-#if defined(WOLFSSL_TLS13) && defined(WOLFSSL_SEND_HRR_COOKIE)
+#ifdef WOLFSSL_TLS13_COOKIE
```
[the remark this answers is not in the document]
padelsbach: fixed

wolfssl/internal.h
```diff
@@ -3882,6 +3882,12 @@ int TLSX_EncryptThenMac_Respond(WOLFSSL* ssl);
 #endif
 
 #ifdef WOLFSSL_TLS13
+
+/* Cookie support is mandatory per RFC 8446 9.2 */
+#if !defined(NO_WOLFSSL_CLIENT) || defined(WOLFSSL_SEND_HRR_COOKIE)
```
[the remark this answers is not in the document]
padelsbach: fixed

tests/api/test_tls13.c
```diff
@@ -7646,6 +7646,107 @@ int test_tls13_hrr_bad_cookie(void)
...
+        wolfTLSv1_3_client_method, wolfTLSv1_3_server_method), 0);
+
+    ExpectIntEQ(wolfSSL_send_hrr_cookie(ssl_s, NULL, 0), WOLFSSL_SUCCESS);
+    /* No key shares in the first ClientHello forces a HelloRetryRequest. */
+    ExpectIntEQ(wolfSSL_NoKeyShares(ssl_c), WOLFSSL_SUCCESS);
+
+    ExpectIntEQ(test_memio_do_handshake(ssl_c, ssl_s, 10, NULL), 0);
+
+    /* Client saw a cookie and the server accepted the one echoed back. */
+    ExpectIntEQ(ssl_c->options.hrrSentCookie, 1);
```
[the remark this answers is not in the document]
padelsbach: fixed

dgarske:
Jenkins retest this please

src/tls.c
```diff
@@ -7607,32 +7605,40 @@ static int TLSX_Cookie_Parse(WOLFSSL* ssl, const byte* input, word16 length,
...
-             * receiving it. */
-            return TLSX_Cookie_Use(ssl, input + idx, len, NULL, 0, 0,
-                                   &ssl->extensions);
-        else
-#endif
-        {
+#ifdef WOLFSSL_SEND_HRR_COOKIE
+    /* client_hello - only a server that sends cookies has one to check the
+     * echoed cookie against. Otherwise ignore it. */
+    {
```
dgarske: Please avoid empty brace.

src/tls.c
dgarske, line 7654: Should we add a len sanity check here? 

Enabling the client path by default means every wolfSSL TLS 1.3 client now accepts a cookie of any length permitted by the extension encoding (up to 65535 bytes, `opaque cookie<1..2^16-1>`) and heap-allocates it in `TLSX_Cookie_Use` (`src/tls.c:7676`: `XMALLOC(sizeof(Cookie) + len + macSz, ...)`), then copies it verbatim into the second ClientHello. HelloRetryRequest is unauthenticated, so any on-path attacker can inject an HRR with a maximal cookie; previously (default build) the extension was compiled out and the blob was discarded. The resulting failure is graceful — `TLSX_GetRequestSize` rejects totals above `WOLFSSL_MAX_16BIT - OPAQUE16_LEN` with `BUFFER_E` (`src/tls.c:7614-7617` region), and `CheckAvailableSize` fails if the output buffer cannot grow — so this is new attack surface and transient memory pressure rather than a memory-safety defect. It is still a change worth bounding: wolfSSL sanity-bounds most other peer-supplied lengths, and the server side already has `MAX_COOKIE_LEN`.

Recommendation: Consider a sanity bound on the accepted HRR cookie length now that this path is on by default, or state explicitly in a comment that unbounded cookies are accepted deliberately per RFC 8446 4.2.2.

dgarske, line 7635: Because `WOLFSSL_TLS13_COOKIE` is derived from `!NO_WOLFSSL_CLIENT`, a combined client+server library built WITHOUT `--enable-hrrcookie` (the default) now compiles `TLSX_Cookie_Parse`, so a SERVER runs the length validation for a cookie extension in an incoming ClientHello even though the body of the checking logic is `#ifdef WOLFSSL_SEND_HRR_COOKIE`-ed out. Previously `CKE_PARSE(a,b,c,d)` expanded to `0` and the extension was ignored wholesale. Now a ClientHello carrying an empty cookie (`00 00`, i.e. `length == 2 < OPAQUE16_LEN + 1`) or an inconsistent inner length aborts the handshake with `BUFFER_E` on a server that has no interest in cookies at all. This is arguably more correct (an empty cookie violates `opaque cookie<1..2^16-1>`), and it is probably intentional, but it is an unannounced behavior change for the default server configuration and there is no test for it. The added comment ("only a server that sends cookies has one to check the echoed cookie against. Otherwise ignore it.") says "ignore it", which does not match the fact that length errors are still fatal.

Recommendation: Confirm this is the intended default-server behavior, fix the comment to say the encoding is still validated, and add a server-side test (no `WOLFSSL_SEND_HRR_COOKIE`) that a ClientHello with a stray cookie extension is ignored while a malformed one is rejected.

dgarske (changes requested):

src/tls.c
padelsbach: removed

src/tls.c
padelsbach: Done, defaults to 4096

padelsbach: intended behavior, updated comment

dgarske (dismissed):
Testing configuration:
`--enable-tls13 --disable-tlsv12 --enable-secure-renegotiation --enable-nullcipher CFLAGS="-DNO_WOLFSSL_CLIENT"`

```
   1579: test_tls13_server_cookie_parse                     :
ERROR - tests/api/test_tls13.c line 7835 failed with:
    expected: TLSX_Parse(ssl, badExt, (word16)sizeof(badExt), client_hello, (Suites*)((const Suites*) ((ssl)->suites != ((void *)0) ? (ssl)->suites : (ssl)->ctx->suites))) == (BUFFER_E)
    result:   0 != -132
 failed (  0.00005)
ERROR - tests/api.c line 39831 failed with:
    expected: Test failed
    result:   ret 0
```

padelsbach:
jenkins retest this please

tests/api/test_tls13.c
```diff
@@ -8063,6 +7864,249 @@ int test_tls13_hrr_bad_cookie(void)
...
+    }
+
+    wolfSSL_free(ssl);
+    wolfSSL_CTX_free(ctx);
+#endif
+    return EXPECT_RESULT();
+}
+
+/* Test that a client rejects a HelloRetryRequest cookie larger than it is
+ * willing to store and echo back. RFC 8446 4.2.2 sets no upper bound.
```
philljj: nit: this is a little confusing.

RFC 8446 4.2.2 says upper bound is  `2 ** 16 - 1 = 65535`.

This test requires `WOLFSSL_MAX_TLS13_COOKIE_SZ <= 65528` because `2 ** 16 - 1 - 1 - 6 = 65528`.

philljj (changes requested):

src/tls.c
```diff
@@ -7591,12 +7593,22 @@ static int TLSX_Cookie_Parse(WOLFSSL* ssl, const byte* input, word16 length,
         return BUFFER_E;
 
     if (msgType == hello_retry_request) {
+        /* RFC 8446 4.2.2 puts no upper bound on the cookie. Limit how much a
+         * server can make us hold and echo back. */
```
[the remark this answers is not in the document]
padelsbach: comment updated

tests/api/test_tls13.c
```diff
@@ -7864,6 +7864,249 @@ int test_tls13_hrr_bad_cookie(void)
...
+ * One byte over the limit plus the six bytes of extension headers has to
+ * still fit the length TLSX_Parse() takes. */
```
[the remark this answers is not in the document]
padelsbach: comment updated

tests/api/test_tls13.c
padelsbach: comment updated

philljj (approved):

Issue #11077, opened 2026-08-06
Policy Rejection Of A Valid PSK Reports `handshake_failure` Instead Of `access_denied`


LiD0209:
# Policy Rejection Of A Valid PSK Reports `handshake_failure` Instead Of `access_denied`

## Problem Description

Focused runtime verification confirmed a real alert-mapping issue in wolfSSL's TLS 1.3 external-PSK path: when the server recognizes a valid offered PSK identity and then rejects it by local policy, the connection fails with `handshake_failure(40)` rather than `access_denied(49)`.

This revised report narrows the original claim. RFC 8446 Section 6.2 defines `access_denied` as the alert meaning for a valid certificate or PSK rejected after access control, but the surrounding text frames alert selection here as sending an appropriate fatal alert rather than an explicit per-site "`MUST send access_denied`" rule. The confirmed issue is therefore a real semantic mismatch in alert selection, not the strongest form of textual MUST-violation.

## Standard Requirement

- Official standard: [RFC 8446](https://www.rfc-editor.org/rfc/rfc8446.html)
- Section: RFC 8446 Section 6.2 Error Alerts

Relevant RFC 8446 context:

```text
Whenever an implementation encounters a fatal error condition, it
SHOULD send an appropriate fatal alert and MUST close the connection
without sending or receiving any additional data.  In the rest of
this specification, when the phrases "terminate the connection" and
"abort the handshake" are used without a specific alert it means that
the implementation SHOULD send the alert indicated by the
descriptions below.  The phrases "terminate the connection with an X
alert" and "abort the handshake with an X alert" mean that the
implementation MUST send alert X if it sends any alert.
```

```text
access_denied:  A valid certificate or PSK was received, but when
   access control was applied, the sender decided not to proceed with
   negotiation.
```

Interpretation:

- Section 6.2 defines `access_denied` as the alert description that matches the "valid certificate or PSK, then local access-control rejection" condition.
- The same section also says that, absent a more specific "abort with X alert" instruction elsewhere, the implementation `SHOULD` send the alert indicated by the alert descriptions.
- Therefore the standards basis here is best read as: `access_denied` is the semantically appropriate alert for this condition, but this specific passage is not itself phrased as an explicit "`MUST send access_denied`" command.

## Relevant Source Code

The implementation defines the `access_denied` codepoint, but the TLS 1.3 external-PSK policy-rejection path does not use it.

### `src/tls13.c:6385-6390`

```c
[...]

PR #11105, opened 2026-08-07
Send decrypt_error alert for TLS 1.3 PSK binder failures

Frauschi:
## Description

`TranslateErrorToAlert()` mapped `BAD_BINDER` to `illegal_parameter(47)`, which tells the peer it sent an invalid ClientHello. RFC 8446 Section 6.2 lists PSK binder validation failure under `decrypt_error(51)`, and that is what the condition actually is: the message was well formed and only the binder MAC failed. This PR changes the mapping.

`CheckPreSharedKeys()` also returns `BAD_BINDER` when no offered PSK matched and there is no certificate to fall back to. That reuse is deliberate (commit `089f1f7c9`): it keeps an unknown PSK identity indistinguishable from a failed binder. Both conditions still share one alert, so the change is indistinguishability preserving, and a comment at the return site now records the invariant at both ends.

Finally, the later certificate check in `CheckPreSharedKeys()` has been unreachable since `089f1f7c9` added the earlier one, so it is removed. The remaining branch falls through to the shared exit, which also restores a missing `WOLFSSL_LEAVE()` on the certificate fallback path.

## Behaviour change

Applications inspecting `wolfSSL_get_alert_history()` will see alert **51 instead of 47** when a binder fails to verify, and when no PSK matched on a server with no certificate. The returned error code (`BAD_BINDER`) and the fatal level are unchanged.

## Identity protection

Splitting `BAD_BINDER` off `illegal_parameter` exposed a second path that had to move with it. `FindPsk()` raised `PSK_KEY_ERROR`, which stays mapped to `illegal_parameter`, when the callback recognised an identity carrying a non-zero `obfuscated_ticket_age`. That ran before any binder was derived, so on a certificate-less server the two alerts would have let an unauthenticated peer enumerate valid PSK identities without holding a key.

The check is removed rather than made to fail differently. RFC 8446 Section 4.2.11: "For identities established externally, an obfuscated_ticket_age of 0 SHOULD be used, and servers MUST ignore the value." Ignoring it satisfies that requirement, closes the oracle, and lets a conformant client that sends a non-zero age complete a handshake that wolfSSL previously rejected.

`test_tls13_psk_age_no_identity_oracle()` guards this. It rewrites the age on the wire, since wolfSSL's own client always sends zero, and asserts a known and an unknown identity produce the same alert. A control run with an untouched ClientHello completes the handshake first, and each run asserts `ssl_s->options.isPSK`, which is set only after the ciphersuite match, so neither rejecting run can pass by way of a fixture that never negotiated a PSK. The accept side for a non-zero age is not covered: any on-wire rewrite invalidates the binder the client already computed, so driving it would need a hand-built ClientHello with an offline-precomputed binder.

## New CI coverage

`.github/workflows/psk.yml` already covers `NO_CERTS` with TLS 1.3 PSK via `static-psk-lowresource-tls13`, but that config disables DH, so it only reaches the `psk_ke` branch. The new `tls13-psk-no-certs` entry in `.github/configs/os-check-linux.json` leaves DH enabled and covers `psk_dhe_ke` under `NO_CERTS`. There is no `--disable-certs`; `ENABLED_CERTS` is derived, so both reach it via `--disable-asn`.

## Testing

- Full `./tests/unit.test` on `--enable-psk --enable-tls13`: `0/1664/483/2147`.
- The two updated assertions in the existing binder tests both fail if the old `illegal_parameter` expectation is restored, so they are not vacuous.
- A memio probe confirms the wire alert changes from 47 to 51, while `wolfSSL_require_psk()` still sends `handshake_failure(40)` and accepting controls still complete.
- The oracle is reproduced and closed: with a client patched to send a non-zero age, a known and an unknown identity gave 47 and 51 before the `FindPsk()` change and both give 51 after.
- Restoring only the age check, with the new test kept, fails it on two assertions: the server-side error (`-333 != -423`) and the alert comparison. The control run still passes there, which is what localises the failure to the age handling.
- A client patched to send a non-zero age with a valid binder is rejected with `illegal_parameter(47)` before the change and completes the handshake after it, confirming the RFC 8446 Section 4.2.11 behaviour.
- The removed branch was instrumented and never entered across the full suite.
- `WOLFSSL_LEAVE()` on the certificate fallback path: 0 before, 1 after.
- `tls13-psk-no-certs` passes through `parallel-make-check.py` with the workflow's own flags. Its `make check` is a genuine `NO_CERTS` run (`0/1876/235/2111`), and both `test_tls13_psk_no_cert_bad_binder` and the new test execute rather than skipping.

## Related

Found while analysing #11077, which asked for `access_denied(49)` when a server PSK callback declines a valid identity. That is not actionable and #11077 should be closed as working as intended: the callback runs before the binder is verified, so nothing is authenticated when the application decides, and `wc_psk_server_callback` cannot distinguish an unknown identity from a denied one. A distinct alert would also reopen the enumeration oracle `089f1f7c9` closed. The mapping fixed here is the real defect that analysis surfaced.


src/internal.c
```diff
@@ -38108,9 +38108,12 @@ static int DoSessionTicket(WOLFSSL* ssl, const byte* input, word32* inOutIdx,
             case WC_NO_ERR_TRACE(PSK_KEY_ERROR):
             case WC_NO_ERR_TRACE(INVALID_PARAMETER):
             case WC_NO_ERR_TRACE(HRR_COOKIE_ERROR):
-            case WC_NO_ERR_TRACE(BAD_BINDER):
             case WC_NO_ERR_TRACE(DUPLICATE_TLS_EXT_E):
                 return illegal_parameter;
+            /* RFC 8446 Section 6.2. The no-PSK-match path reuses BAD_BINDER
+             * for identity protection, so both share this alert. */
+            case WC_NO_ERR_TRACE(BAD_BINDER):
+                return decrypt_error;
```
Frauschi, line 38116: fixed

tests/api/test_tls13.c
```diff
@@ -2691,6 +2691,172 @@ int test_tls13_psk_no_cert_bad_binder(void)
...
+    wolfSSL_CTX_free(ctx_c);
+    wolfSSL_CTX_free(ctx_s);
+
+    return EXPECT_RESULT();
+}
+#endif
+
+/* A ClientHello that offers a known PSK identity with a non-zero
+ * obfuscated_ticket_age must fail exactly like one offering an unknown
+ * identity. Both abort before any binder is derived, so a differing alert
```
Frauschi: fixed.

tests/api/test_tls13.c
```diff
@@ -2691,6 +2691,172 @@ int test_tls13_psk_no_cert_bad_binder(void)
...
+    !defined(NO_PSK)
+    int knownAlert = -1;
+    int unknownAlert = -1;
+
+    ExpectIntEQ(test_tls13_psk_age_alert(TEST_TLS13_PSK_AGE_KNOWN_ID,
+        &knownAlert), TEST_SUCCESS);
+    /* Positive control: without this the test would still pass if the two
+     * identity strings ever drifted apart and both runs took the unknown
+     * path, asserting nothing. */
+    ExpectIntEQ(test_tls13_psk_age_matched, 1);
```
Frauschi: fixed.

tests/api/test_tls13.c
```diff
@@ -2691,6 +2691,172 @@ int test_tls13_psk_no_cert_bad_binder(void)
...
+    wolfSSL_CTX_free(ctx_s);
+
+    return EXPECT_RESULT();
+}
+#endif
+
+/* A ClientHello that offers a known PSK identity with a non-zero
+ * obfuscated_ticket_age must fail exactly like one offering an unknown
+ * identity. Both abort before any binder is derived, so a differing alert
+ * would let an unauthenticated peer enumerate the server's PSK identities. */
```
Frauschi: fixed.

tests/api/test_tls13.c
Frauschi: fixed.

tests/api/test_tls13.c
```diff
@@ -2691,6 +2691,211 @@ int test_tls13_psk_no_cert_bad_binder(void)
...
+int test_tls13_psk_age_no_identity_oracle(void)
+{
+    EXPECT_DECLS;
+#if defined(WOLFSSL_TLS13) && \
+    defined(HAVE_MANUAL_MEMIO_TESTS_DEPENDENCIES_BUILD) && \
+    !defined(NO_PSK)
+    int knownAlert = -1;
+    int unknownAlert = -1;
+
+    ExpectIntEQ(test_tls13_psk_age_alert(TEST_TLS13_PSK_AGE_KNOWN_ID,
```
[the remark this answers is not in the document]
Frauschi: Declining this one.

The suggested approach is not reachable from a test. `TLSX_PreSharedKey_Use()` assigns `psk->ticketAge = age` unconditionally, including when it finds an existing entry, so an age set before `wolfSSL_connect()` is overwritten by the client's own call passing 0. There is no hook between that call and binder computation. Precomputing the binder instead needs a byte-exact static ClientHello that any change to the default extensions or suite order would break.

The branch is also not unguarded. The known-identity run asserts the server error is `BAD_BINDER`, so reintroducing an age check fails it with `-333 != -423`. Only the wire-level completion is uncovered, and that was verified by hand: rejected with `illegal_parameter(47)` before the change, completing after it.

dgarske (approved):

PR #11108, opened 2026-08-07
Fix several OpenSSL compatibility API gaps

julek-wolfssl:
- Define `WOLFSSL_SIGNER_DER_CERT` for `OPENSSL_ALL` builds so the CA table keeps DER data, fixing `X509_STORE_get0_objects()` returning an empty stack after `X509_STORE_add_cert()`. `WOLFSSL_NO_SIGNER_DER_CERT` can opt back out.
- Accept FFDHE group names in `SSL_set1_groups_list()` by adding FFDHE entries to `kNistCurves`, so OpenSSL-style lists mixing curves and FFDHE groups no longer fail as a whole. `SSL_set1_curves_list()` remains EC-only.
- Add `SSL_get_negotiated_group()` and `SSL_group_to_name()`, reporting TLS group registry names (OpenSSL spelling), distinct from both `kNistCurves` names and `wolfSSL_get_curve_name()` output.
- Move the leaf domain/IP checks driven by `X509_VERIFY_PARAM` into `ProcessPeerCerts()`, alongside the existing `wolfSSL_check_domain_name()` check, instead of running them in `DoVerifyCallback()` after the app callback already ran. This lets the application see and react to name-mismatch errors via the verify callback instead of the handshake just failing afterward. As a side effect, `SSL_VERIFY_NONE` now suppresses these checks, matching the non-OpenSSL path.
- Fix `SSL_get_peer_tmp_key()` for TLS 1.3: the key share code freed the peer's ephemeral key right after deriving the shared secret regardless of `wolfSSL_KeepHandshakeResources()`, breaking the call for every TLS 1.3 group. Now the peer key is kept when requested, X25519/X448 keys are also returned, and the kept keys are freed with the other peer keys on teardown.


julek-wolfssl:
retest this please flaky

src/tls.c
philljj, line 9774: Will this stomp on a previously retained `ssl->peerX25519Key`?

TLSX_KeyShare_ProcessX448_ex is already doing the same with that key, so maybe is ok.

src/tls.c
```diff
@@ -9767,8 +9767,9 @@ static int TLSX_KeyShare_ProcessX25519_ex(WOLFSSL* ssl,
          * falls through to the cleanup code below. */
     }
 
-    /* done with key share, release resources */
-    if (ssl->peerX25519Key != NULL) {
+    /* done with key share, release resources unless the peer key was asked
+     * for - wolfSSL_get_peer_tmp_key() needs it after the handshake */
+    if ((ssl->peerX25519Key != NULL) && !ssl->options.keepResources) {
```
philljj: This appears valid.

philljj: This looks valid.

We handle `ssl->peerX448Key` with different checks than `ssl->peerX25519Key` for the same wolfSSL_get_peer_tmp_key mechanism.

philljj (changes requested):
Please review Fenrir. Two of them look real.

wolfssl/wolfcrypt/settings.h
```diff
@@ -4999,6 +4999,14 @@ blinding by defining WC_BLINDING_NO_RNG_ACKNOWLEDGE_WEAKNESS."
     #define WOLFSSL_COPY_KEY
 #endif
 
+/* X509_STORE_get0_objects() can only enumerate a store's certificates when the
+ * CA table keeps their DER. Costs one DER copy per loaded CA, so it stays
+ * optional; define WOLFSSL_NO_SIGNER_DER_CERT to opt back out. */
+#if defined(OPENSSL_ALL) && !defined(WOLFSSL_NO_SIGNER_DER_CERT)
+    #undef  WOLFSSL_SIGNER_DER_CERT
+    #define WOLFSSL_SIGNER_DER_CERT
```
philljj, line 5007: Please check this null deref one too.

julek-wolfssl:
Pushed. Shutdown threads are a decline — explained last.

- **`X509_VERIFY_PARAM` host/IP** (`1c2c2fe4c`) — `hostName`/`ipasc` checked in their own pass, not as a fallback for the WOLFSSL buffers; shared matching via `CheckPeerHostName()`. Ungated by `verifyNone`, as `DoVerifyCallback()` had them.
- **Unauthenticated TLS 1.3 alerts** (`1c2c2fe4c`) — `WOLFSSL_CB_READ_ALERT` moved below the plaintext-alert rejection.
- **X25519 peer key** (`7446e5ae8`) — @philljj: yes, it would have stomped a retained key. Now mirrors `ProcessX448_ex()`: `FreeKey()` before allocating, retain only on `(ret == 0) && keepResources`.
- **FFDHE NIDs** (`1f0a74b03`) — `WC_NID_ffdhe2048/3072/4096` already existed with OpenSSL's values; added 6144/8192 and used them in `kNistCurves`. **Behaviour change:** `wolfSSL_get_negotiated_group()` now returns the NID, not the code point, as OpenSSL does; `group_to_name()` accepts both. EC-only lookups skip these rows by RFC 7919 range, not by the NID column.
- **`ca->derCert` deref** (`3da9e1765`, `3486df080`) — reachable: `OPENSSL_ALL` defaults `WOLFSSL_SIGNER_DER_CERT` on and `cm_restore_cert_row()` inserts signers with `derCert == NULL`. Guarded; single success return.
- **Tests** (`9257beb43`, `1f0a74b03`) — helper guard matches its caller (`!NO_SHA256`); peer tmp key asserts type via `EVP_PKEY_id()` (`base_id()` returns `NID_undef` for X25519/X448 — pre-existing); regression asserts `EC_curve_nid2nist(WC_NID_md4)` is NULL.

**Shutdown-in-init `ssl->error` — not applying, it breaks shutdown.** Setting it (even only when already 0) makes a second `SSL_shutdown()` return 0 and send a close_notify on a connection that never handshook:

```
after 1st shutdown: error=-324 s_len=0
after 2nd shutdown: error=0    s_len=7
```

`wolfssl_handshake_failed()` is "`ssl->error` nonzero and not WANT_READ/WANT_WRITE", so any non-WANT_* value makes the next call take the send path — the guard poisons its own input. `wolfSSL_want()` compares `ssl->error` the same way, so `SSL_want_read()` breaks too. The info isn't missing anyway: `WOLFSSL_ERROR()` queues it via `wc_AddErrorNode()`, so `ERR_get_error()` returns it, same as OpenSSL. The real gap — `SSL_get_error()` never consulting the queue — is library-wide, not this PR.

Verified: `-Werror`; full suite green under `--enable-all`, `--enable-opensslextra`, `-DWOLFSSL_NO_SIGNER_DER_CERT`; `check-source-text` OK.

julek-wolfssl:
1. peerVerifyRet — yes, the new verify-param checks set ret but leave
   peerVerifyRet at V_OK, so a callback override hides a wrong-name cert.
   Will record HOSTNAME_MISMATCH / IP_ADDRESS_MISMATCH at both sites. The
   old check_domain_name path has the same gap, but that predates this PR
   so I'll leave it for a separate one.

2. get_negotiated_group — confirmed. InitSSL() copies ctx->ecdhCurveOID
   unconditionally, so after SSL_CTX_set_tmp_ecdh() the getter returns a
   group with no handshake at all. Will gate the fallback on an actual
   EC(DH) exchange and add a test.

3. shutdown-in-init — I tried this and it breaks two things, so I've left
   the behaviour and documented it instead. ssl->error also gates handshake
   progress (internal.c:25067) and drives wolfssl_handshake_failed()
   (ssl_api_rw.c:1041), so writing it there aborts the in-flight handshake
   (test_memio_do_handshake then fails with -324) and makes the guard skip
   itself on the next call, sending the very alert it exists to suppress.

   The caller isn't left without a signal: the existing
   WOLFSSL_ERROR(NOT_READY_ERROR) queues it via wc_AddErrorNode, so
   ERR_get_error() reports it — same mechanism OpenSSL uses for
   SSL_R_SHUTDOWN_WHILE_IN_INIT, which is on the error queue rather than in
   the object's retry state. Added tests for both properties.

philljj:
Retest this please.

(fips ready passed, but success didn't percolate up)

src/ssl_api_rw.c
```diff
@@ -1030,6 +1063,25 @@ int wolfSSL_shutdown(WOLFSSL* ssl)
...
+             (!wolfssl_handshake_failed(ssl))) {
+        WOLFSSL_MSG("Shutdown called before the handshake completed");
+        /* Queue NOT_READY_ERROR but leave ssl->error alone. ssl->error is a
+         * single slot that also gates handshake progress and drives
+         * wolfssl_handshake_failed() above, so writing it here would abort
+         * the in-flight handshake and make this guard skip itself on the
+         * next call. The queued error is what OpenSSL reports for
+         * SSL_R_SHUTDOWN_WHILE_IN_INIT, and it is separate from the retry
+         * state either way. */
+        WOLFSSL_ERROR(NOT_READY_ERROR);
```
philljj, line 1097: code comment explains why this is intended

src/ssl_api_rw.c
```diff
@@ -1021,6 +1025,35 @@ static int wolfssl_shutdown_recv_close_notify(WOLFSSL* ssl)
...
+}
+
+/* Whether the handshake failed outright rather than being still in flight.
+ * WANT_READ/WANT_WRITE (and a pending async operation) mean it can still
+ * continue, anything else recorded in ssl->error means it cannot.
+ *
+ * @param [in] ssl  SSL/TLS object.
+ * @return  1 when the handshake has failed, 0 when it can still progress.
+ */
+static int wolfssl_handshake_failed(const WOLFSSL* ssl)
```
philljj, line 1044: skoll:

**Description:** The new predicate recognizes only `WANT_READ`, `WANT_WRITE`, and optionally `WC_PENDING_E` as an in-progress handshake. Other retry states, notably `OCSP_WANT_READ` from asynchronous OCSP processing and `WOLFSSL_ERROR_WANT_X509_LOOKUP` from certificate callbacks, also leave the handshake resumable. Calling `wolfSSL_shutdown()` in either state now misclassifies the handshake as failed, bypasses the new initialization guard, and can send `close_notify` while the initial handshake is incomplete. The added shutdown test covers only `WANT_READ`, so these paths remain untested.

**Code:**
```
return (ssl->error != 0) &&
    (ssl->error != WC_NO_ERR_TRACE(WANT_READ)) &&
    (ssl->error != WC_NO_ERR_TRACE(WANT_WRITE))
#ifdef WOLFSSL_ASYNC_CRYPT
    && (ssl->error != WC_NO_ERR_TRACE(WC_PENDING_E))
#endif
    ;
```

**Recommendation:** Add shutdown tests that pause a handshake in nonblocking OCSP and certificate-lookup callbacks, then verify that shutdown fails without appending an alert.

src/tls.c
```diff
@@ -9925,8 +9936,18 @@ static int TLSX_KeyShare_ProcessX448_ex(WOLFSSL* ssl,
                     ssOutput, ssOutSz, EC448_LITTLE_ENDIAN);
     }
 
-    wc_curve448_free(peerX448Key);
-    XFREE(peerX448Key, ssl->heap, DYNAMIC_TYPE_TLSX);
+    /* Keep the peer key when it was asked for - wolfSSL_get_peer_tmp_key()
```
philljj, line 9966: skoll:

**Description:** The PR begins retaining X25519 and X448 peer keys when handshake resources are kept, but the X448 path does not clear a retained `peerX25519Key` from a previous connection. After reusing a `WOLFSSL` for an X25519 handshake followed by an X448 handshake, both keys remain populated. The newly extended `wolfSSL_get_peer_tmp_key()` checks X25519 before X448 and therefore returns the stale X25519 key instead of the current connection's X448 key. The tests create a fresh `WOLFSSL` for every group and do not cover connection reuse.

**Code:**
```
if ((ret == 0) && ssl->options.keepResources) {
    FreeKey(ssl, DYNAMIC_TYPE_CURVE448, (void**)&ssl->peerX448Key);
    ssl->peerX448Key = peerX448Key;
    ssl->peerX448KeyPresent = 1;
}
```

**Recommendation:** Add a reuse test that performs X25519 and then X448 handshakes on the same `WOLFSSL` and verifies that the returned EVP key type is X448.

src/ssl.c
```diff
@@ -9754,6 +9808,161 @@ int set_curves_list(WOLFSSL* ssl, WOLFSSL_CTX *ctx, const char* names,
...
+ * Mirrors OpenSSL's SSL_get_negotiated_group(): the NID is returned for groups
+ * wolfSSL gives one, the IANA code point otherwise. NID values are wolfSSL's
+ * WC_NID_* (not every one matches OpenSSL's), so compare against those rather
+ * than against literals. The result can be passed to wolfSSL_group_to_name().
+ *
+ * @param [in] ssl  SSL/TLS object.
+ * @return  Group identifier on success.
+ * @return  0 when ssl is NULL or no group has been negotiated.
+ */
+int wolfSSL_get_negotiated_group(const WOLFSSL* ssl)
```
philljj, line 9860: skoll:

**Description:** The new API reads `ssl->namedGroup` without establishing that it belongs to the current connection. `wolfSSL_clear()` resets the handshake state but does not reset `namedGroup`, so a cleared, unconnected object still reports the previous connection's group rather than zero. A subsequent TLS 1.3 PSK-only `psk_ke` handshake can also complete without processing a new key share, leaving the old group visible even though the current handshake negotiated no group. The test only checks a newly allocated object before its first handshake.

**Code:**
```
#if defined(WOLFSSL_TLS13) || defined(HAVE_FFDHE)
    group = ssl->namedGroup;
#endif

if (group == 0)
    return 0;
```

**Recommendation:** Test the API immediately after `wolfSSL_clear()` and after reusing the object for a PSK-only TLS 1.3 handshake.

src/ssl.c
```diff
@@ -9754,6 +9808,161 @@ int set_curves_list(WOLFSSL* ssl, WOLFSSL_CTX *ctx, const char* names,
...
+    /* Not a group OpenSSL names - fall back to the wolfSSL name. */
+    for (nist_name = kNistCurves; nist_name->name != NULL; nist_name++) {
+        if ((int)nist_name->curve == group)
+            return nist_name->name;
+    }
+
+    return NULL;
+}
+
 #endif /* (HAVE_ECC || HAVE_CURVE25519 || HAVE_CURVE448) */
```
philljj: skoll:

**Description:** The PR adds FFDHE rows and public group APIs inside a compilation block that requires `HAVE_ECC`, `HAVE_CURVE25519`, or `HAVE_CURVE448`. In an `OPENSSL_EXTRA` build supporting only FFDHE, neither function is declared or defined, while `wolfssl/openssl/ssl.h` still aliases `SSL_get_negotiated_group` and `SSL_group_to_name` unconditionally outside nginx. Applications using either advertised compatibility API then encounter an implicit declaration and ultimately a missing symbol. This also prevents the newly added FFDHE naming behavior from existing in the configuration where it is most directly needed.

**Code:**
```
#if defined(OPENSSL_EXTRA) || defined(HAVE_CURL)
#if (defined(HAVE_ECC) || defined(HAVE_CURVE25519) || defined(HAVE_CURVE448))
/* kNistCurves and the new group APIs */
...
#endif
#endif
```

**Recommendation:** Add an `OPENSSL_EXTRA` FFDHE-only compile test that calls both compatibility APIs, plus a handshake test verifying the negotiated FFDHE group name.

philljj (changes requested):

julek-wolfssl:
retest this please

julek-wolfssl:
retest this please

philljj:
Retest this please.

(no logs)

julek-wolfssl:
retest this please

julek-wolfssl:
retest this please

philljj (changes requested):
Please review fenrir item, seems straightforward

julek-wolfssl:
retest this please

julek-wolfssl:
retest this please

philljj (dismissed):
merge conflict in ssl_api_rw.c

julek-wolfssl:
retest this please

src/ssl.c
```diff
@@ -5742,6 +5785,28 @@ size_t wolfSSL_get_client_random(const WOLFSSL* ssl, unsigned char* out,
...
+         * leave it readable as this connection's. */
+        ssl->namedGroup = 0;
+    #endif
+    #if defined(HAVE_ECC) || defined(HAVE_ED25519) || \
+        defined(HAVE_CURVE25519) || defined(HAVE_ED448) || \
+        defined(HAVE_CURVE448)
+        /* The peer's ephemeral key is now kept past the handshake for
+         * wolfSSL_get_peer_tmp_key(). A next handshake that negotiates no
+         * (EC)DH would otherwise report the previous connection's key. */
+        ssl->ecdhCurveOID = 0;
```
JacobBarthelmeh: I think this should be set to `ssl->ctx->>ecdhCurveOID` instead of 0 so that the value set with `wolfSSL_CTX_set_tmp_ecdh()` is retained and that the next connection attempt when re-using a WOLFSSL structure does not drop the curve preference.

JacobBarthelmeh (commented):
One comment about the ECDH curve preference reset with wolfSSL_clear(). Otherwise it all looks good.

src/ssl.c
julek-wolfssl: Good find. Fixing.

JacobBarthelmeh (approved):

PR #11115, opened 2026-08-08
Fix UAF in DH buffers

padelsbach:
# Description

Fixes the case where `SetSSL_CTX` aliases the `serverDH_P` and `serverDH_G` buffers with `weOwnDH` still at 0. A subsequent call to `wolfSSL_CTX_SetTmpDH*` would result in a dangling pointer.

Related to #11109, but with DH buffers. Note that refcounting is not performed here since the API uses plain byte buffers instead of structs/objects.

`weOwnDH == 0` currently means either "CTX heap buffer" or "static named-group params" today. After this change, it only means the static case which removes ambiguity in the free paths.

# Testing

Added unit tests.

# Checklist

 - [x] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -7624,6 +7624,58 @@ static int SetSSL_CTX_CertsAndKeys(WOLFSSL* ssl, WOLFSSL_CTX* ctx)
 }
 #endif /* NO_CERTS */
 
+#ifndef NO_DH
+/* Give the SSL object its own copy of the context's DH parameters.
+ *
+ * The parameters are plain buffers with no reference count on them, so a
+ * session must not point at the context's: the context is free to replace
+ * them at any time. They are small and only present when the application
+ * asked for them, so a copy per session is the cheap way to keep them safe.
+ *
+ * @param [in, out] ssl  SSL object. Any parameters it owns are let go of.
+ * @param [in]      ctx  SSL context object.
+ * @return  0 on success.
+ * @return  MEMORY_E when dynamic memory allocation fails.
+ */
+int CopySSL_CTX_DhParams(WOLFSSL* ssl, WOLFSSL_CTX* ctx)
+{
+    byte* p;
+    byte* g;
+
+    if ((ctx->serverDH_P.buffer == NULL) || (ctx->serverDH_G.buffer == NULL)) {
+        return 0;
+    }
+
+    p = (byte*)XMALLOC(ctx->serverDH_P.length, ssl->heap,
+        DYNAMIC_TYPE_PUBLIC_KEY);
+    g = (byte*)XMALLOC(ctx->serverDH_G.length, ssl->heap,
+        DYNAMIC_TYPE_PUBLIC_KEY);
+    if ((p == NULL) || (g == NULL)) {
+        XFREE(p, ssl->heap, DYNAMIC_TYPE_PUBLIC_KEY);
+        XFREE(g, ssl->heap, DYNAMIC_TYPE_PUBLIC_KEY);
+        return MEMORY_E;
+    }
+
+    /* Let go of any parameters this object already had. */
+    if (ssl->buffers.weOwnDH) {
+        XFREE(ssl->buffers.serverDH_P.buffer, ssl->heap,
+            DYNAMIC_TYPE_PUBLIC_KEY);
+        XFREE(ssl->buffers.serverDH_G.buffer, ssl->heap,
+            DYNAMIC_TYPE_PUBLIC_KEY);
+    }
+
+    XMEMCPY(p, ctx->serverDH_P.buffer, ctx->serverDH_P.length);
+    XMEMCPY(g, ctx->serverDH_G.buffer, ctx->serverDH_G.length);
+    ssl->buffers.serverDH_P.buffer = p;
+    ssl->buffers.serverDH_P.length = ctx->serverDH_P.length;
+    ssl->buffers.serverDH_G.buffer = g;
+    ssl->buffers.serverDH_G.length = ctx->serverDH_G.length;
+    ssl->buffers.weOwnDH = 1;
+
+    return 0;
+}
+#endif /* !NO_DH */
+
 int SetSSL_CTX(WOLFSSL* ssl, WOLFSSL_CTX* ctx, int writeDup)
 {
     int ret = WOLFSSL_SUCCESS; /* set default ret */
@@ -7812,8 +7864,9 @@ int SetSSL_CTX(WOLFSSL* ssl, WOLFSSL_CTX* ctx, int writeDup)
         !defined(HAVE_SELFTEST)
         ssl->options.dhKeyTested = ctx->dhKeyTested;
     #endif
-    ssl->buffers.serverDH_P = ctx->serverDH_P;
-    ssl->buffers.serverDH_G = ctx->serverDH_G;
+    if (CopySSL_CTX_DhParams(ssl, ctx) != 0) {
+        return MEMORY_E;
+    }
 #endif
 
 #if defined(HAVE_RPK)
@@ -9688,7 +9741,6 @@ void wolfSSL_ResourceFree(WOLFSSL* ssl)
     }
     XFREE(ssl->buffers.serverDH_Priv.buffer, ssl->heap, DYNAMIC_TYPE_PRIVATE_KEY);
     XFREE(ssl->buffers.serverDH_Pub.buffer, ssl->heap, DYNAMIC_TYPE_PUBLIC_KEY);
-    /* parameters (p,g) may be owned by ctx */
     if (ssl->buffers.weOwnDH) {
         XFREE(ssl->buffers.serverDH_G.buffer, ssl->heap, DYNAMIC_TYPE_PUBLIC_KEY);
         XFREE(ssl->buffers.serverDH_P.buffer, ssl->heap, DYNAMIC_TYPE_PUBLIC_KEY);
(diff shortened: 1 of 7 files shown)


tests/api/test_ssl_hs.c
```diff
@@ -1055,6 +1062,115 @@ int test_wolfSSL_set_connect_state_dh(void)
...
+        ExpectIntEQ(ssl_s->buffers.serverDH_P.length, ctx_s->serverDH_P.length);
+        ExpectIntEQ(ssl_s->buffers.serverDH_G.length, ctx_s->serverDH_G.length);
+        ExpectIntEQ(ssl_s->buffers.weOwnDH, 1);
+        startP = ssl_s->buffers.serverDH_P.buffer;
+        startG = ssl_s->buffers.serverDH_G.buffer;
+    }
+
+    /* Offering no FFDHE group keeps the server on the parameters it was
+     * given, rather than switching to a named group and dropping them. */
+    ExpectIntEQ(wolfSSL_UseSupportedCurve(ssl_c, WOLFSSL_ECC_SECP256R1),
```
padelsbach, line 1151: adding missing `HAVE_SUPPORTED_CURVES`

src/internal.c
```diff
@@ -10077,13 +10129,8 @@ void FreeHandshakeResources(WOLFSSL* ssl)
...
     XFREE(ssl->buffers.serverDH_Pub.buffer, ssl->heap, DYNAMIC_TYPE_PUBLIC_KEY);
     ssl->buffers.serverDH_Pub.buffer = NULL;
-    /* parameters (p,g) may be owned by ctx */
-    if (ssl->buffers.weOwnDH) {
-        XFREE(ssl->buffers.serverDH_G.buffer, ssl->heap, DYNAMIC_TYPE_PUBLIC_KEY);
-        ssl->buffers.serverDH_G.buffer = NULL;
-        XFREE(ssl->buffers.serverDH_P.buffer, ssl->heap, DYNAMIC_TYPE_PUBLIC_KEY);
-        ssl->buffers.serverDH_P.buffer = NULL;
-    }
+    /* The parameters (p,g) are kept: a renegotiation or a reused object needs
```
padelsbach: This is the point of the PR, to keep the buffers

tests/api/test_ssl_hs.c
padelsbach, line 1151: adding missing `HAVE_SUPPORTED_CURVES`

tests/api/test_ssl_hs.c
```diff
@@ -1055,6 +1062,115 @@ int test_wolfSSL_set_connect_state_dh(void)
...
+    ExpectNotNull(ctx_s = wolfSSL_CTX_new(wolfTLSv1_2_server_method()));
+    ExpectIntEQ(wolfSSL_CTX_use_PrivateKey_file(ctx_s, svrKeyFile,
+        WOLFSSL_FILETYPE_PEM), WOLFSSL_SUCCESS);
+    ExpectIntEQ(wolfSSL_CTX_use_certificate_file(ctx_s, svrCertFile,
+        WOLFSSL_FILETYPE_PEM), WOLFSSL_SUCCESS);
+    ExpectIntEQ(wolfSSL_CTX_set_cipher_list(ctx_s, test_ctx.s_ciphers),
+        WOLFSSL_SUCCESS);
+    wolfSSL_SetIORecv(ctx_s, test_memio_read_cb);
+    wolfSSL_SetIOSend(ctx_s, test_memio_write_cb);
+    ExpectIntEQ(wolfSSL_CTX_SetTmpDH_file(ctx_s, dhParamFile,
```
padelsbach: changed to `CERT_FILETYPE`

tests/api/test_ssl_hs.c
```diff
@@ -1055,6 +1062,115 @@ int test_wolfSSL_set_connect_state_dh(void)
...
+    /* So the object can be handed a second connection. */
+    ExpectIntEQ(wolfSSL_clear(ssl_s), WOLFSSL_SUCCESS);
+    wolfSSL_free(ssl_c);
+    ssl_c = NULL;
+    ExpectNotNull(ssl_c = wolfSSL_new(ctx_c));
+    wolfSSL_SetIOWriteCtx(ssl_c, &test_ctx);
+    wolfSSL_SetIOReadCtx(ssl_c, &test_ctx);
+    ExpectIntEQ(wolfSSL_UseSupportedCurve(ssl_c, WOLFSSL_ECC_SECP256R1),
+        WOLFSSL_SUCCESS);
+    test_ctx.c_len = test_ctx.s_len = 0;
```
padelsbach: done

tests/api/test_ssl_hs.c
```diff
@@ -1055,6 +1062,115 @@ int test_wolfSSL_set_connect_state_dh(void)
...
+
+    ExpectIntEQ(test_memio_do_handshake(ssl_c, ssl_s, 10, NULL), 0);
+    ExpectIntNE(wolfSSL_get_error(ssl_s, 0), WC_NO_ERR_TRACE(NO_DH_PARAMS));
```
padelsbach: dropped and asserts added

tests/api/test_tls13.c
```diff
@@ -3909,6 +3909,110 @@ int test_tls13_pha(void)
     return EXPECT_RESULT();
 }
 
+
+/* DH parameters are not reference counted, so each session takes its own copy
+ * of them and the context is free to replace its own at any time. */
+int test_tls13_ctx_dh_rotation(void)
+{
+    EXPECT_DECLS;
+#if defined(WOLFSSL_TLS13) && !defined(NO_DH) && !defined(NO_RSA) && \
```
padelsbach, line 3918: fixed

tests/api/test_ssl_hs.c
padelsbach, line 1151: fixed

tests/api/test_tls13.c
padelsbach, line 3918: fixed

src/internal.c
padelsbach: this is the point of the PR

padelsbach: fixed

src/ssl_api_hs.c
```diff
@@ -1592,9 +1592,14 @@ void wolfSSL_set_accept_state(WOLFSSL* ssl)
...
-            ssl->buffers.serverDH_P = ssl->ctx->serverDH_P;
-            ssl->buffers.serverDH_G = ssl->ctx->serverDH_G;
-            ssl->options.haveDH = 1;
+            /* Nothing to return the failure through, so leave the object
+             * without DH rather than claim parameters it does not have. */
+            if (CopySSL_CTX_DhParams(ssl, ssl->ctx) == 0) {
+                ssl->options.haveDH = 1;
+            }
+            else {
+                WOLFSSL_MSG("Unable to copy DH parameters from context");
```
padelsbach, line 1601: We don't have this capability today

src/internal.c
padelsbach: pre-existing issue, out of scope

src/ssl_api_hs.c
```diff
@@ -1592,9 +1592,14 @@ void wolfSSL_set_accept_state(WOLFSSL* ssl)
...
         if ((!ssl->options.haveDH) && (ssl->ctx->haveDH)) {
-            ssl->buffers.serverDH_P = ssl->ctx->serverDH_P;
-            ssl->buffers.serverDH_G = ssl->ctx->serverDH_G;
-            ssl->options.haveDH = 1;
+            /* Nothing to return the failure through, so leave the object
+             * without DH rather than claim parameters it does not have. */
+            if (CopySSL_CTX_DhParams(ssl, ssl->ctx) == 0) {
+                ssl->options.haveDH = 1;
+            }
+            else {
```
padelsbach, line 1600: We don't have this capability today

src/internal.c
```diff
@@ -7624,6 +7624,58 @@ static int SetSSL_CTX_CertsAndKeys(WOLFSSL* ssl, WOLFSSL_CTX* ctx)
...
+
+    /* Let go of any parameters this object already had. */
+    if (ssl->buffers.weOwnDH) {
+        XFREE(ssl->buffers.serverDH_P.buffer, ssl->heap,
+            DYNAMIC_TYPE_PUBLIC_KEY);
+        XFREE(ssl->buffers.serverDH_G.buffer, ssl->heap,
+            DYNAMIC_TYPE_PUBLIC_KEY);
+    }
+
+    XMEMCPY(p, ctx->serverDH_P.buffer, ctx->serverDH_P.length);
```
padelsbach, line 7668: pre-existing issue, out of scope

tests/api/test_ssl_hs.c
```diff
@@ -810,10 +810,15 @@ int test_wolfSSL_set_accept_state_reinit(void)
...
+    /* Picked up as the object's own copy, not the context's buffers. */
     if ((ssl != NULL) && (ctx != NULL)) {
         ExpectIntEQ(ssl->options.haveDH, 1);
-        ExpectPtrEq(ssl->buffers.serverDH_P.buffer, ctx->serverDH_P.buffer);
-        ExpectPtrEq(ssl->buffers.serverDH_G.buffer, ctx->serverDH_G.buffer);
+        ExpectNotNull(ssl->buffers.serverDH_P.buffer);
+        ExpectNotNull(ssl->buffers.serverDH_G.buffer);
+        ExpectPtrNE(ssl->buffers.serverDH_P.buffer, ctx->serverDH_P.buffer);
+        ExpectPtrNE(ssl->buffers.serverDH_G.buffer, ctx->serverDH_G.buffer);
+        ExpectIntEQ(ssl->buffers.serverDH_P.length, ctx->serverDH_P.length);
```
padelsbach, line 820: fixed

src/ssl_api_hs.c
padelsbach, line 1600: fixed

padelsbach:
jenkins retest this please

src/ssl_api_hs.c
philljj, line 1601: At first I didn't like this, but it's actually no different than the ecc errors that can happen above in the same function.

Looks reasonable to me, we're constrained by backwards compat.

(this is a comment only, hit resolve at leisure)

src/internal.c
```diff
@@ -7624,6 +7624,58 @@ static int SetSSL_CTX_CertsAndKeys(WOLFSSL* ssl, WOLFSSL_CTX* ctx)
 }
 #endif /* NO_CERTS */
 
+#ifndef NO_DH
+/* Give the SSL object its own copy of the context's DH parameters.
```
philljj, line 7628: I think this is better / safer from a correctness point of view.

But I think embedded users might not like the extra allocs and copies.

Is there a way this can be handled like `WOLFSSL_COPY_CERT`?

E.g. `WOLFSSL_COPY_DH`:

```
int CopySSL_CTX_DhParams(WOLFSSL* ssl, WOLFSSL_CTX* ctx)
{

    #if defined WOLFSSL_COPY_DH
      alloc and deep copy
    #else
      shallow copy
    #endif /* WOLFSSL_COPY_DH */

}
```

Or, make the deep copy the default behavior, but it can be switched to the shallow copy with `WOLFSSL_NO_COPY_DH`, like `WOLFSSL_NO_COPY_CERT`.


```
wolfssl/wolfcrypt/settings.h:

/* Always copy certificate(s) from SSL CTX to each SSL object on creation,      
 * if this is not defined then each SSL object shares a pointer to the          
 * original certificate buffer owned by the SSL CTX. */                               
#if defined(OPENSSL_ALL) && !defined(WOLFSSL_NO_COPY_CERT)                        
    #undef WOLFSSL_COPY_CERT                                                        
    #define WOLFSSL_COPY_CERT                                                      
#endif
```

philljj (changes requested):

src/internal.c
padelsbach: I also considered the footprint impact, but disabling this change just re-enables the UAF. I'd say the improvement is worth the ~300 bytes.

I did find a small improvement for servers to free the buffer after the negotiation is complete. Will push shortly

philljj (approved):

PR #11120, opened 2026-08-10
SBOM dep cpe purl

sameehj:
# Description

Improve SBOM generation so dependency and subcomponent identities are usable for automated vulnerability scanning (CRA / IEC 62443 style monitoring).

- Emit registered **CPE 2.3** on dependency components (CycloneDX `cpe` / SPDX `cpe23Type`), not PURL alone.
- Emit **canonical, resolvable** `pkg:github` PURLs (lowercase namespace/name; real release tags such as `v<version>-stable` for wolfSSL).
- Model **wolfCrypt as contained in wolfSSL** (CycloneDX nested component / SPDX `CONTAINS`), not as an external `DEPENDS_ON`.
- Align `scripts/gen-sbom` with the wolfGlass generator: do not publish a `cpe` for products that are not in the NVD Official CPE Dictionary; support crypto-only builds via `--crypto-only`.

This supports corrected release SBOMs for integrators (e.g. wolfSSL 5.9.1) and matches the identity model used by downstream products such as wolfBoot.

Fixes zd#22297

# Testing

- Unit tests: `scripts/test_gen_sbom.py` (updated/extended for CPE, PURL forms, wolfCrypt containment, omit CPE when version unknown / product pending).
- Regenerated sample CycloneDX/SPDX from the updated generator and checked:
  - non-empty component inventory where expected
  - `cpe:2.3:a:wolfssl:wolfssl:…` and `cpe:2.3:a:wolfssl:wolfcrypt:…` present
  - PURLs use lowercase `pkg:github/wolfssl/…` and `@v…-stable` where applicable
  - wolfCrypt nested under wolfSSL / SPDX `CONTAINS`

# Checklist

 - [x] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/.github/workflows/sbom.yml
+++ b/.github/workflows/sbom.yml
@@ -516,15 +516,20 @@ jobs:
         # GitHub Security Advisories index is also pinned so an auditor
         # reading the SBOM has a single in-document link to the project's
         # disclosures.
+        # purl-spec lowercases the github namespace/name; wolfSSL tags are
+        # `vX.Y.Z-stable`, so the version is not digits-and-dots alone.
         run: |
           python3 - <<'PY'
           import glob, json, re
           with open(glob.glob('wolfssl-*.spdx.json')[0]) as f:
               d = json.load(f)
+          pkg = next(p for p in d['packages'] if p['name'] == 'wolfssl')
           refs = {r['referenceType']: r['referenceLocator']
-                  for r in d['packages'][0]['externalRefs']}
+                  for r in pkg['externalRefs']}
           assert re.match(r'cpe:2\.3:a:wolfssl:wolfssl:[\d.]+:', refs['cpe23Type']), refs
-          assert re.match(r'pkg:github/wolfSSL/wolfssl@v[\d.]+', refs['purl']), refs
+          assert re.match(
+              r'pkg:github/wolfssl/wolfssl@v[\d.]+(-stable)?$',
+              refs['purl']), refs
           assert refs['advisory'] == \
               'https://github.com/wolfSSL/wolfssl/security/advisories', refs
           print('identifiers ok:', refs)
--- a/.github/workflows/sbom.yml
+++ b/.github/workflows/sbom.yml
@@ -5,6 +5,7 @@ on:
   push:
     branches: [ 'master', 'main', 'release/**' ]
   pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
     branches: [ '**' ]
 
 # Defence-in-depth: the workflow does no API writes (no `gh pr`, no
@@ -28,7 +29,7 @@ jobs:
   # cheapest gate for licence/UUID/timestamp logic regressions.
   unit:
     name: gen-sbom unit tests
-    if: github.repository_owner == 'wolfssl'
+    if: ${{ (github.repository_owner == 'wolfssl') && (github.event_name != 'pull_request' || github.event.pull_request.draft == false) }}
     runs-on: ubuntu-24.04
     timeout-minutes: 5
     steps:
@@ -64,7 +65,7 @@ jobs:
   # --srcs, or --dep-version handling would silently land.
   standalone:
     name: SBOM standalone (no autotools)
-    if: github.repository_owner == 'wolfssl'
+    if: ${{ (github.repository_owner == 'wolfssl') && (github.event_name != 'pull_request' || github.event.pull_request.draft == false) }}
     runs-on: ubuntu-24.04
     needs: unit
     timeout-minutes: 10
@@ -423,7 +424,7 @@ jobs:
   # everything an external auditor or vulnerability scanner relies on.
   integration:
     name: SBOM integration (linux)
-    if: github.repository_owner == 'wolfssl'
+    if: ${{ (github.repository_owner == 'wolfssl') && (github.event_name != 'pull_request' || github.event.pull_request.draft == false) }}
     runs-on: ubuntu-24.04
     needs: unit
     timeout-minutes: 20
@@ -758,7 +759,7 @@ jobs:
   # macOS runner minutes go to portability coverage, not duplicated checks.
   integration-macos:
     name: SBOM integration (macos)
-    if: github.repository_owner == 'wolfssl'
+    if: ${{ (github.repository_owner == 'wolfssl') && (github.event_name != 'pull_request' || github.event.pull_request.draft == false) }}
     runs-on: macos-latest
     needs: unit
     timeout-minutes: 20
@@ -818,7 +819,7 @@ jobs:
   # exercised by hand; a regression in either would silently land.
   bomsh:
     name: bomsh integration (linux)
-    if: github.repository_owner == 'wolfssl'
+    if: ${{ (github.repository_owner == 'wolfssl') && (github.event_name != 'pull_request' || github.event.pull_request.draft == false) }}
     runs-on: ubuntu-24.04
     needs: unit
     timeout-minutes: 30
--- a/Makefile.am
+++ b/Makefile.am
@@ -583,6 +583,8 @@ sbom:
 	    --options-h $(abs_builddir)/wolfssl/options.h \
 	    --lib "$$sbom_lib" \
 	    --dep-libz "$(ENABLED_LIBZ)" \
+	    --dep-wolfcrypt yes \
+	    --dep-version wolfcrypt=$(PACKAGE_VERSION) \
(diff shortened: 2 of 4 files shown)


JacobBarthelmeh:
@sameehj, un-assigning wolfSSL-bot until the CI failures are resolved.

MarkAtwood (dismissed):
Reviewed against zd#22297. The design is right: the CPE/PURL identity model,
the wolfCrypt containment, and the decision to keep the wolfssl component on a
crypto-only build are all correct, and the reasoning in the comments is good.
All 334 checks pass.

One thing blocks merge. Everything else below is a follow-up, not a gate.

## Blocking: wolfCrypt loses its CPE and PURL on the wolfBoot path

`main()` inherits wolfCrypt's version before `_resolve_dep_versions()` runs:

```python
dep_version_overrides = _parse_dep_version_overrides(args.dep_version)
if 'wolfcrypt' in enabled_deps and 'wolfcrypt' not in dep_version_overrides:
    if dep_version_overrides.get('wolfssl'):        # still empty here
        ...
_resolve_dep_versions(enabled_deps, dep_version_overrides)   # runs after
```

The wolfssl version normally arrives from pkg-config rather than from
`--dep-version`, so at the inherit step there is nothing to inherit from. An
embedder that does not pass `--dep-version wolfssl=` gets a wolfcrypt component
with no version, no purl and no cpe. It warns on stderr and exits 0, so nothing
catches it.

Repro, with a `wolfssl.pc` on `PKG_CONFIG_PATH` reporting 5.9.1:

```
python3 scripts/gen-sbom --name wolfboot --version 2.9.0 \
    --component-type firmware --options-h o.h --lib lib.a \
    --license-file LICENSING --dep-wolfssl yes --dep-wolfcrypt yes \
    --cdx-out out.cdx.json --spdx-out out.spdx.json
```

This branch:

```
wolfssl    version=5.9.1  cpe=cpe:2.3:a:wolfssl:wolfssl:5.9.1:*:*:*:*:*:*:*
wolfcrypt  version=None   cpe=None                                purl=None
```

wolfGlass `share/gen-sbom` on the same inputs:

```
wolfssl    version=5.9.1  cpe=cpe:2.3:a:wolfssl:wolfssl:5.9.1:*:*:*:*:*:*:*
wolfcrypt  version=5.9.1  cpe=cpe:2.3:a:wolfssl:wolfcrypt:5.9.1:*:*:*:*:*:*:*
                          purl=pkg:github/wolfssl/wolfssl@v5.9.1-stable#wolfcrypt
```

This matters because it is half of the customer's original report. Forough's
words were "no CPE entries are present for wolfSSL/wolfCrypt". This PR fixes the
wolfSSL half and leaves the wolfCrypt half broken on the path wolfBoot uses.

wolfGlass already hit this and fixed it in `_inherit_wolfcrypt_version()`. The
fix has two parts, and the second is easy to miss:

1. Call it after `_resolve_dep_versions`.
2. Test the value, not the key. `_resolve_dep_versions` caches a `None` for
   every dep pkg-config cannot resolve, so by that point `'wolfcrypt'` is
   always present in the overrides dict and the current `not in` guard still
   skips.

Taking that function verbatim from wolfGlass is the cheapest fix.

## Follow-ups, not gating

**`--dep-*` accepts anything and silently means "no".** No `choices=`, and the
filter is `if flag.lower() == 'yes'`. A fragment wiring
`--dep-wolfssl=$(HAVE_WOLFSSL)` where the variable expands to `1`, `true` or `Y`
drops a CVE-bearing component with no diagnostic. Pre-existing for
`--dep-wolfssl` so not a regression here, but this PR extends the pattern to
`--dep-wolfcrypt`. wolfGlass's `_enabled_deps()` exits with the flag name.

**The `pending` CPE path is unreachable.** All five `PRODUCT_CPE` entries are
`registered`, so `product_cpe_requested()` and `cpe-status=pending` can never
fire, and the only test asserts it returns `None`. Fine as scaffolding, worth
either exercising or noting as deliberate.

**wolfCrypt's CPE resolves to nothing in NVD.** Querying the dictionary API,
`cpe:2.3:a:wolfssl:wolfcrypt` returns exactly one row, the versionless `-`.
No version rows, against 132 for wolfssl and 33 for wolfboot. Emitting
`:5.9.1:` is not wrong, the vendor:product pair is registered and
`resolve_crypto_only` documents why wolfssl stays top-level. But a scanner
reporting "wolfCrypt: 0 findings" reads as coverage when it is really no data.
Worth a deliberate call rather than a side effect.

**Unescaped version in the CPE string.** `_product_cpe_string` interpolates
`version` straight into a CPE 2.3 formatted string. A pkg-config version
containing `:` yields a malformed CPE. Low likelihood, cheap guard.

**`GEN_SBOM_VERSION` jumps 1.2 to 1.7** with changelog entries for 1.3 through
1.6, versions that never produced output in this tree. Harmless, but the
constant's own comment says it must change in lockstep with the output it
produces.

## Verified

The wolfBoot CPE claim in the 1.7 changelog entry checks out. The NVD dictionary
API returns 33 rows for `cpe:2.3:a:wolfssl:wolfboot`, including
`cpe:2.3:a:wolfssl:wolfboot:2.9.0`, which is the version in the ticket. Flipping
that entry to `registered` is correct.

The CycloneDX nesting and `dependencies[]` edges are right, and the SPDX
`CONTAINS` mirror handles the no-container fallback correctly.

MarkAtwood:
Correcting my own review above, on the blocking item.

I already told you on 08-12 that the wolfCrypt version-inheritance ordering is
my bug, not yours, that it comes from wolfGlass `gen-sbom` which you ported
faithfully, and that you should NOT patch it locally because I am fixing it
upstream and you re-port. That still stands. My review above asks you to take
`_inherit_wolfcrypt_version` from wolfGlass into this PR, which contradicts it.
Ignore that instruction.

What the review does add is the reproduction, which I had not done before. With
a `wolfssl.pc` on `PKG_CONFIG_PATH` reporting 5.9.1, in wolfBoot's argument
shape, this branch emits a wolfcrypt component with `version=None`, `cpe=None`,
`purl=None` and exits 0, where wolfGlass emits the full identity. So it is real
and it is half of what the customer reported in zd#22297. It just is not yours
to fix here.

Leaving this as changes-requested so the PR does not merge in the meantime, but
the reason is the sequence, not anything you need to do: wolfGlass lands the
fixes and cuts a resolvable rev, you re-vendor wolfBoot#830, then this re-ports
and gets the fix for free. Same sequence as 08-12.

The non-gating items in the review stand on their own and are worth carrying
into wolfGlass rather than here, for the same reason: the `--dep-*` flags
silently swallowing non-yes values, the unreachable `pending` CPE path, the
unescaped version in the CPE string, and the 1.2 to 1.7 version jump.

The verified parts are unchanged and good: the wolfBoot CPE really is
registered now, including a 2.9.0 row, and the CycloneDX nesting, the
`dependencies[]` edges and the SPDX `CONTAINS` mirror are all correct.

MarkAtwood:
Superseded, closing. Not rejected, and nothing here is lost.

Every headline feature in this PR already exists in wolfGlass master. I checked
symbol by symbol: `PRODUCT_CPE`, `GITHUB_TAG_FORMS`, `github_purl`,
`product_cpe_requested`, `resolve_crypto_only`, `CRYPTO_ONLY_MACRO`,
`WOLFSSL_SUBSET_CRYPTO_ONLY`, all present upstream, none of them on wolfssl
master. Measured against wolfGlass, this branch closes the engine gap from 590
differing lines to 125, so it is a 465-line catch-up rather than new work. That
catch-up is exactly what the sync does mechanically.

The sequence is the one from 08-12, unchanged:

1. wolfGlass lands its open PRs and cuts a resolvable rev
2. wolfBoot#830 re-vendors against that rev
3. wolfssl/scripts/gen-sbom syncs from wolfGlass rather than being hand-ported

The wolfCrypt version-inheritance bug gets fixed for free in step 3, because
`_inherit_wolfcrypt_version` is already upstream. So does the containment
asymmetry once that decision lands. The two `--dep-version wolfcrypt=`
workarounds in `sbom.mk` and `sbom.cmake` can go at the same time, since they
only exist to work around that bug.

The design calls in this PR were right and they carry forward as-is: registered
CPEs on dependency components, canonical resolvable pkg:github PURLs with real
release tags, wolfCrypt as contained rather than a sibling, and keeping the
wolfssl component present on crypto-only builds because it is the only
identifier NVD maps advisories to.

Reopen if the sync stalls and zd#22297 needs an artifact sooner than it lands.

PR #11125, opened 2026-08-11
SP ECC P-256 x64 asm: fix carry in triple and rsb_sub_dbl

SparkiDev:
# Description

Subtracted 0 assuming register was 0.
Changed to subtract register from register to ensure 0.

Fixes zd#22300

# Testing

./configure --disable-shared --enable-intelasm --enable-sp --enable-sp-asm
Used customers reproducer which is very rare case.


SparkiDev:
Code generated with PR:
https://github.com/wolfSSL/scripts/pull/651

JacobBarthelmeh (approved):

Issue #11087, opened 2026-08-06
Mismatch mismatch in sending tickets incompatible with advertised PSK modes


LiD0209:
#Mismatch in sending tickets incompatible with advertised PSK modes

## Problem Description

A focused runtime recheck confirmed a real RFC 8446 Section 4.2.9 mismatch in wolfSSL.

When the client advertises only `psk_ke` and the server is configured to only allow `psk_dhe_ke` for PSK-based handshakes, wolfSSL still completes the initial full TLS 1.3 certificate handshake and sends a `NewSessionTicket`. That ticket is not compatible with the only advertised mode: a follow-up resumption attempt with the saved ticket fails on the server PSK path with `PSK_KEY_ERROR`.

This report deduplicates multiple candidate-level records that resolved to the same root cause.

## Standard Requirement

- Official standard: [RFC 8446](https://www.rfc-editor.org/rfc/rfc8446.html)
- Section: Section 4.2.9 Pre-Shared Key Exchange Modes (lines 2862-2864)

> Servers SHOULD NOT
>    send NewSessionTicket with tickets that are not compatible with the
>    advertised modes;

Interpretation:

If the client advertises a PSK mode set and the server later sends `NewSessionTicket`, the ticket should be usable with at least one advertised mode. A ticket that is incompatible with every advertised mode should not be sent.

The key context from RFC 8446 Section 4.2.9 is that `psk_key_exchange_modes`
does not only constrain PSKs already offered in the current `ClientHello`; it
also constrains PSKs that the server may later create with `NewSessionTicket`.

## Relevant Source Code

The runtime reproducer matches a concrete source path:

### `src/tls.c:16724-16735`

ClientHello mode advertisement is built directly from the `noPskDheKe` / `onlyPskDheKe` options. With `wolfSSL_CTX_no_dhe_psk()` on the client, the extension advertises only `PSK_KE`.

```c
/* Some servers do not generate session tickets unless
 * the extension is seen in a non-resume client hello.
 * We used to send it only if we were otherwise using PSK.
 * Now always send it.
 */
/* Pre-shared key modes: mandatory extension for resumption. */
if (!ssl->options.onlyPskDheKe) {
    modes = 1 << PSK_KE;
}
if (!ssl->options.noPskDheKe) {
    modes |= 1 << PSK_DHE_KE;
}
```

### `src/tls13.c:13097-13217`

[...]

PR #11137, opened 2026-08-11
tls 1.3: respect advertised psk_key_exchange modes for tickets and PSKs

gasbytes:
# Description

A wolfSSL server ignored the client's advertised psk_key_exchange_modes when issuing a NewSessionTicket, handing out  tickets it would never accept , and then aborted the resuption attempt with PSK_KEY_ERROR instead of falling back to a full handshake.
This skips tickets no advertised mode can use (as mentioned in RFC 8446, around section 4.2.9) and declines unusuable PSKs rather than failing the connection (section 4.2.11).

Fixes https://github.com/wolfSSL/wolfssl/issues/11087.

# Testing

Added two regression tests: test_tls13_psk_mode_no_incompatible_ticket test_tls13_psk_mode_incompatible_falls_back. The first one asserts no ticket is sent to a psk_ke-only client when the server requires psk_dhe_ke, with a compatible-client control that still gets one.
The second offers a valid ticket under that same mode mismatch and asserts the handshake completes unresumed instead of aborting.

# Checklist

 - [x] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -8308,6 +8308,7 @@ static void InitSSL_Tls13Options(WOLFSSL* ssl, WOLFSSL_CTX* ctx)
     #endif
     #if defined(HAVE_SESSION_TICKET) || !defined(NO_PSK)
         ssl->options.noPskDheKe = ctx->noPskDheKe;
+        ssl->options.noPskDheKeCfg = ctx->noPskDheKe;
     #ifdef HAVE_SUPPORTED_CURVES
         ssl->options.onlyPskDheKe = ctx->onlyPskDheKe;
     #endif /* HAVE_SUPPORTED_CURVES */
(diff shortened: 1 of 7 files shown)


julek-wolfssl (changes requested):
This overlaps with https://github.com/wolfSSL/wolfssl/pull/11178. I don't like that this is declining to send tickets by default. The RFC says
```
If the client's hello contained a suitable "psk_key_exchange_modes" extension, the server MAY send a NewSessionTicket message at any time after it has received the client Finished message.
```
so sending the ticket when the modes are missing is not against the RFC. Not sending as a default may break use cases for users.

julek-wolfssl:
Closing in favor of https://github.com/wolfSSL/wolfssl/pull/11178

gasbytes:
> Closing in favor of #11178

Commenting for history, the fallback logic to full handshake included at this commit:
https://github.com/wolfSSL/wolfssl/pull/11178/commits/77bff353dac924ccac6fd0f2677431a27683ec09

PR #11140, opened 2026-08-11
Fix directoryName NameConstraints bypass

holtrop-wolfssl:
# Description

Excluded and permitted directoryName subtrees were compared with a byte prefix memcmp, so a leaf DN escaped an exclusion by varying case, string type or spacing.

Compare the RDN sequence per RFC 5280 Sec. 7.1 instead. UTF-8 decoding provided by wc_Utf8_DecodeChar() in coding.c.

Fixes ZD#22282

# Testing

Added CI tests.

# Checklist

 - [X] added tests
 - [ ] updated/added doxygen


holtrop-wolfssl:
The wolfBoot Integration / keytools CI failure needs this fix on the wolfBoot side: https://github.com/wolfSSL/wolfBoot/pull/857

wolfcrypt/src/asn.c
```diff
@@ -18588,16 +18588,392 @@ int ConfirmSignature(SignatureCtx* sigCtx,
...
+    switch (tag) {
+        case ASN_UTF8STRING:
+        case ASN_NUMERICSTRING:
+        case ASN_PRINTABLE_STRING:
+        case ASN_T61STRING:
+        case ASN_IA5_STRING:
+        case ASN_ISO646STRING:
+        case ASN_GENERALSTRING:
+        case ASN_UNIVERSALSTRING:
+        case ASN_BMPSTRING:
```
JacobBarthelmeh, line 18713: Should this account for the tags ASN_GRAPHICSTRING (0x19), VideotexString (0x15) and constructed forms (0x2C)?

wolfcrypt/src/asn.c
```diff
@@ -18588,16 +18588,392 @@ int ConfirmSignature(SignatureCtx* sigCtx,
...
+        return 1;
+    }
+
+    if (DirStringIsTag(aTag) && DirStringIsTag(bTag)) {
+        return DirStringEqual(aTag, a + ai, (word32)aLen,
+                              bTag, b + bi, (word32)bLen);
+    }
+
+    /* Not a character string: only an identical encoding is equal. */
+    if ((aTag != bTag) || (aLen != bLen)) {
```
JacobBarthelmeh: When exactly one of aTag/bTag is a recognised character-string tag, when either tag has ASN_CONSTRUCTED set on a UNIVERSAL string tag, or when a tag is not one RFC 5280 permits for an AttributeValue (not a DirectoryString type and not BIT STRING for x500UniqueIdentifier), return ASN_PARSE_E so the caller rejects the certificate instead of concluding "not in the excluded subtree".

wolfcrypt/src/asn.c
```diff
@@ -18588,16 +18588,392 @@ int ConfirmSignature(SignatureCtx* sigCtx,
...
+        int ret = DirStringRaw(s, &c);
+
+        if (ret != 1) {
+            /* End of value, where a buffered space run was trailing and is
+             * dropped, or a malformed encoding. */
+            return ret;
+        }
+        if (c == (word32)' ') {
+            /* Leading spaces are dropped, inner ones buffered until it is
+             * known that a further code point follows. */
```
JacobBarthelmeh, line 18915: Trimming of spaces and out of order comparison needs close review. I think we should either close that gap and add the complexity to handle those cases, or fail closed and report ASN_PARSE_E rather than continue on to report a match/missmatch.

JacobBarthelmeh, line 18915: Value consisting only of spaces folds to the empty string, diverging from RFC 4518 insignificant-space handling.

JacobBarthelmeh (changes requested):
`./configure --disable-coding && make check` fails to build now with:

```
  CCLD     src/libwolfssl.la
Undefined symbols for architecture x86_64:
  "_wc_Utf8_DecodeChar", referenced from:
      _DirStringNext in src_libwolfssl_la-asn.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

holtrop-wolfssl:
> `./configure --disable-coding && make check` fails to build now with:
> 
> ```
>   CCLD     src/libwolfssl.la
> Undefined symbols for architecture x86_64:
>   "_wc_Utf8_DecodeChar", referenced from:
>       _DirStringNext in src_libwolfssl_la-asn.o
> ld: symbol(s) not found for architecture x86_64
> clang: error: linker command failed with exit code 1 (use -v to see invocation)
> ```

Did you `./autogen.sh` first?

JacobBarthelmeh:
You are right, I did not run ./autogen.sh first, and it builds okay after. It looks like there is a couple unguarded calls to Base64_Decode() though, and can be seen with `./autogen.sh && ./configure --disable-coding --enable-opensslextra && make`. That configure fails on the master branch too though, but in a different way than after this PR. So maybe it is separate from this PR and the fix could be it's own PR if one is made.

wolfcrypt/src/asn.c
```diff
@@ -18588,16 +18588,392 @@ int ConfirmSignature(SignatureCtx* sigCtx,
...
+    if (DirStringIsTag(aTag) && DirStringIsTag(bTag)) {
+        return DirStringEqual(aTag, a + ai, (word32)aLen,
+                              bTag, b + bi, (word32)bLen);
+    }
+
+    /* Not a character string: only an identical encoding is equal. */
+    if ((aTag != bTag) || (aLen != bLen)) {
+        return 0;
+    }
+    return XMEMCMP(a + ai, b + bi, (size_t)aLen) == 0;
```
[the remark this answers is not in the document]
holtrop-wolfssl: Fixed

wolfcrypt/src/asn.c
```diff
@@ -18588,16 +18588,392 @@ int ConfirmSignature(SignatureCtx* sigCtx,
...
+        int aLen = 0;
+        int bLen = 0;
+        int ret;
+
+        if ((GetSequence(a, &ai, &aLen, aSz) < 0) ||
+                (GetSequence(b, &bi, &bLen, bSz) < 0)) {
+            return ASN_PARSE_E;
+        }
+
+        ret = MatchDirAttr(a + ai, (word32)aLen, b + bi, (word32)bLen);
```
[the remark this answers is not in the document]
holtrop-wolfssl: Fixed

wolfcrypt/src/asn.c
```diff
@@ -19211,11 +19585,18 @@ static int PermittedListOk(DNS_entry* name, Base_entry* dnsList, byte nameType)
...
-                                            current->name, current->nameSz)) {
-                match = 1; /* found the current name in the permitted list*/
-                break;
+            else {
+                int mRet = wolfssl_local_MatchBaseName(nameType, name->name,
+                        name->len, current->name, current->nameSz);
+                if (mRet < 0) {
+                    /* Neither answer is safe when the name cannot be
+                     * compared: let the caller reject the certificate. */
+                    return mRet;
```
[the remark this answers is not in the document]
holtrop-wolfssl: Fixed

wolfcrypt/src/asn.c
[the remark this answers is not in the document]
holtrop-wolfssl: Fixed

wolfcrypt/src/asn.c
holtrop-wolfssl: Those are not options for a DirectoryString object (https://datatracker.ietf.org/doc/html/rfc5280#section-4.1.2.4).

Also relevant text from section 7.1:

Conforming implementations MUST
   support UTF8String and PrintableString.  [RFC 3280](https://datatracker.ietf.org/doc/html/rfc3280) required only
   binary comparison of attribute values encoded in UTF8String, however,
   this specification requires a more comprehensive handling of
   comparison.  Implementations may encounter certificates and CRLs with
   names encoded using TeletexString, BMPString, or UniversalString, but
   support for these is OPTIONAL.

wolfcrypt/src/asn.c
holtrop-wolfssl: Can you clarify what you mean? RFC 4518 2.6.1 says "If the input string contains no non-space character ...", so I read that as "" being treated the same as " ", "  ", and "   ", which this implements.

holtrop-wolfssl: Can you clarify what you mean?

wolfcrypt/src/asn.c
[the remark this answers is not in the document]
holtrop-wolfssl: Fixed

wolfcrypt/src/asn.c
JacobBarthelmeh: Did some more testing and my concern of order after spaces are removed is handled already in this PR.

Was thinking of cases like :

```
       base RDN:  SET { CN="Forbidden Corp",    O="Forbidden Corp" }                                          
                  both AVAs encode to 0x15 octets, tie broken on the OID  -> [CN, O]                          
                                                                                                              
       leaf RDN:  SET { CN="Forbidden Corp   ", O="Forbidden Corp" }                                          
                  CN is now 0x18 octets, so DER *requires*                -> [O, CN] 
```

JacobBarthelmeh: Your understanding of this is correct. One additional thing I'd bring up is if there is other characters besides ' ' that should be stripped out. Like \<TAB\>.

wolfcrypt/src/coding.c
```diff
@@ -749,3 +750,111 @@ int Base16_Encode(const byte* in, word32 inLen, byte* out, word32* outLen)
...
+ *          continuation octets than the buffer holds.
+ * @return  ASN_INPUT_E when the octets are not a valid encoding of a code
+ *          point.
+ */
+int wc_Utf8_DecodeChar(const byte* in, word32 inLen, word32* inOutIdx,
+                       word32* cp)
+{
+    word32 idx;
+    word32 c;
+    word32 min;
```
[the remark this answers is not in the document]
holtrop-wolfssl: Fixed

wolfcrypt/src/asn.c
holtrop-wolfssl: The RFC explicitly lists only the space character: "For the purposes of this section, a space is defined to be the SPACE (U+0020) code point"

holtrop-wolfssl:
retest this please (ERROR: Build 'wolfSSL/PRB-fips-repo-and-harness-test-v3-part1' failed with result: FAILURE)

wolfcrypt/src/asn.c
JacobBarthelmeh: Sending on Slack some of test certificates that caused me to bring up \<TAB\> cases.

holtrop-wolfssl: Fixed - implemented code point mapping to SPACE and nothing.

JacobBarthelmeh (approved):

PR #11143, opened 2026-08-11
wolfCrypt SRAM PUF: health test the raw readout before accepting it

dgarske:
Fixes Fenrir 8140

## What this adds

`wc_PufReadSram()` health tests the raw SRAM readout before the context accepts it. A readout that cannot be SRAM power-on noise - a 128-bit block that is all zero or all ones, a block repeating the one before it, or a total Hamming weight outside `WC_PUF_HW_MIN_PCT`..`WC_PUF_HW_MAX_PCT` (35% to 65% by default) - is rejected with `PUF_READ_E` and never reaches the fuzzy extractor. Without the test, an all-zero region (the ordinary result of sampling after C runtime startup) passes cleanly through encoding, masking, BCH decoding and HKDF: every call returns 0, while the derived key is identical on every device and computable offline.

The test runs on the copy in `ctx->rawSram` rather than the caller's buffer, so the bytes that were tested are the bytes that get used. `WC_PUF_FLAG_SRAM_SET` goes down at the top of the call and back up only on success, so any failed read - a short buffer or a rejected readout, on a fresh context or one that had already read successfully - leaves `wc_PufEnroll()` and `wc_PufReconstruct()` refusing to run, with the rejected bytes scrubbed by `ForceZero()`. Output already derived from a readout that did pass, such as the identity or helper data, stays available.

New public accessor `wc_PufCheckSram(sramAddr, sramSz, onesCount)` runs the same test standalone, for qualifying a candidate SRAM region during board bring-up. `onesCount` is optional and is written whenever the size check passes - including when the readout is then rejected - so the measured bias of a rejected region is still available; it is left untouched when `sramAddr` is NULL or `sramSz` is short.

There is deliberately no macro that compiles the check out; only the band is tunable, through `WC_PUF_HW_MIN_PCT` / `WC_PUF_HW_MAX_PCT` (CMake: `-DWOLFSSL_PUF_HW_MIN_PCT` / `-DWOLFSSL_PUF_HW_MAX_PCT`, which reach the generated `options.h` so an application measuring a region sees the same band as the library). `wc_PufSetTestData()` still bypasses it, since that `WOLFSSL_PUF_TEST` injection path must carry adversarial vectors.

## What it does not catch

Degenerate shapes only. Ordinary firmware content left in the region - `.data` copied from flash, a string table, a previous boot stage - is device-identical yet neither constant nor strongly biased, so it can pass. Sampling from reset, before `.bss`/`.data` init, remains a requirement of correct NOLOAD placement rather than something this test enforces.

## Compatibility

No change to the helper-data format, the derived key, or any existing signature, so helper data enrolled by wolfSSL 5.9.2 stays valid; the band is deliberately not part of `WC_PUF_PROFILE_ID`. One behavior change worth calling out: an integration that was sampling its region too late now gets `PUF_READ_E` where it previously got a silent, device-independent key. Documented in `ChangeLog.md`.

## Testing

- `puf_test()` covers every rejection path (all zero, all ones, a repeated block and a constant block away from index 0, biased high and low), both edges of the band with exact-weight fixtures, the accessor's bad-args and success paths, a failed read invalidating a context that had already read successfully, and the `wc_PufSetTestData()` bypass.
- Host matrix, 14 configs green: four BCH profiles, codeword counts 1/3/32, compact helper layout, SHA3, `--enable-smallstack`, `--enable-c89`, and non-default bands at 20/80 and 55/85. `make check` clean.
- The CI matrix builds the default band and a widened one, and `puf_test()` takes its expectations from the band the build configured, so each job exercises its own band rather than re-testing the default.
- Hardware, 11/11 PASS on the STM32 bench: NUCLEO-H563ZI at t=7/10/13/15 and cw=8/16/32, plus g071 (M0+), f439 (M4F), f767 (M7) and u585 (M33).
- The default band is validated against real silicon: genuine cold-boot SRAM on an H563ZI measures 51.90% and 51.61% ones over two power cycles, while the same board after a *software* reset measures 12.79% - the previous run's stack and identity rather than power-on entropy - and is now rejected instead of silently deriving a device-independent key.

### DIFF
--- a/doc/dox_comments/header_files/puf.h
+++ b/doc/dox_comments/header_files/puf.h
@@ -17,6 +17,21 @@
     the helper data at enrollment and pass it to wc_PufReconstructEx, or
     compare it against wc_PufGetProfileId(), to catch a build mismatch.
 
+    Every readout handed to wc_PufReadSram() is health tested first (see
+    wc_PufCheckSram). A degenerate readout - all zero, all ones, or a repeating
+    block - would otherwise pass cleanly through encoding, masking, decoding
+    and HKDF and derive a key that is the same on every device.
+    WC_PUF_HW_MIN_PCT and WC_PUF_HW_MAX_PCT (default 35 and 65) set the
+    Hamming-weight band the readout must fall inside.
+
+    The test rejects degenerate readouts, not every already-written region:
+    ordinary firmware content - .data copied from flash, a string table, a
+    previous boot stage - is identical on every device yet neither constant nor
+    strongly biased, so it can pass. Sampling the region from reset, before
+    .bss/.data initialization and before it is used as stack or heap, remains a
+    requirement of correct NOLOAD placement rather than something this test can
+    enforce.
+
     For a complete bare-metal example (tested on NUCLEO-H563ZI), see
     https://github.com/wolfSSL/wolfssl-examples/tree/master/puf
 */
@@ -52,9 +67,18 @@ int wc_PufInit(wc_PufCtx* ctx);
     required size, WC_PUF_RAW_BYTES, scales with WC_PUF_NUM_CODEWORDS
     (256 bytes at the default 16 codewords).
 
+    The readout is health tested with wc_PufCheckSram() before it is accepted.
+    A degenerate readout - typically a region already cleared by .bss init, the
+    common bring-up mistake of sampling after C runtime startup - is rejected
+    with PUF_READ_E, and the context is left unusable by wc_PufEnroll() and
+    wc_PufReconstruct(). The test catches degenerate shapes, not every
+    already-written region, so it does not remove the requirement to sample
+    from reset.
+
     \return 0 on success
     \return BAD_FUNC_ARG if ctx or sramAddr is NULL
-    \return PUF_READ_E if sramSz < WC_PUF_RAW_BYTES
+    \return PUF_READ_E if sramSz < WC_PUF_RAW_BYTES, or if the readout fails
+    the health test
 
     \param ctx pointer to wc_PufCtx structure
     \param sramAddr pointer to raw SRAM memory region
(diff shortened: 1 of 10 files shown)


doc/dox_comments/header_files/puf.h
```diff
@@ -68,11 +83,49 @@ int wc_PufInit(wc_PufCtx* ctx);
...
+
+    \brief Health test a candidate raw SRAM readout without loading it into a
+    context. Rejects a readout that cannot be SRAM power-on noise: any 128-bit
+    block that is all zero or all ones, any block that repeats the block before
+    it, or a total Hamming weight outside the WC_PUF_HW_MIN_PCT to
+    WC_PUF_HW_MAX_PCT band (default 35% to 65% of WC_PUF_RAW_BITS).
+
+    wc_PufReadSram() applies this test to every readout, so calling it directly
+    is only needed to qualify a candidate SRAM region during board bring-up, or
+    to report why a read was refused. onesCount is optional and is written even
```
danielinux: it should say "written whenever the size check passes" -- two out of three failures in this functions kip the *onesCount=one assignment

danielinux (changes requested):

danielinux:
The new health-test band entry in `.github/workflows/puf.yml` doesn't have any discriminating power as written. With `-DWC_PUF_HW_MIN_PCT=20 -DWC_PUF_HW_MAX_PCT=80`, every fixture in `puf_test()` lands on the same side of the band as it does with the default 35/65:

| fixture | ones | 35/65 | 20/80 |
|---|---|---|---|
| `testSram` | 50.0% | pass | pass |
| biased high (`0xFE`) | 83.6% | reject | reject |
| biased low (`0x01`) | 13.3% | reject | reject |

Same outcome at `WC_PUF_NUM_CODEWORDS` 1 and 3, so no profile in the matrix separates the two bands.

The job would therefore still be green if `WC_PUF_HW_MIN_PCT` were misspelled, or never reached the compiler at all: it builds the tunable path but never exercises it, which is not quite what the comment above the entry claims.

Two ways to give it teeth: pick fixture biases that fall *between* the bands (~25% and ~75% ones are rejected by the default band and accepted at 20/80), or assert in `puf_test()` against the configured `WC_PUF_HW_MIN_PCT`/`WC_PUF_HW_MAX_PCT` rather than against a fixed expectation.

danielinux (approved):

PR #11144, opened 2026-08-11
Configure.ac: Limit wolfCLU features on FIPS version

lealem47:
# Description

Fix for --enable-wolfclu enabling non-fips features. Users should opt-in to MD5, Ed25519 and the ASN format relaxation as they require it.

# Testing

Tested with ./configure --enable-wolfclu --enable-fips=v5

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -12241,10 +12241,13 @@ then
         ENABLED_CERTEXT="yes"
     fi
 
-    # Requires md5
-    if test "$ENABLED_MD5" = "no"
+    # Has option for md5 hashing, enabled by default for non-FIPS 140-3
+    if test "$ENABLED_FIPS" = "no" || test $HAVE_FIPS_VERSION -lt 5
     then
         ENABLED_MD5="yes"
+    elif test "$enable_md5" != "yes"
+    then
+        ENABLED_MD5="no"
     fi
 
     # Requires aesctr
@@ -12266,7 +12269,7 @@ then
     fi
 
     # Has option for signing with ED25519
-    if test "$ENABLED_ED25519" = "no"
+    if test "$ENABLED_ED25519" = "no" && (test "$ENABLED_FIPS" = "no" || test $HAVE_FIPS_VERSION -ge 6)
     then
         ENABLED_ED25519=yes
         ENABLED_FEMATH=yes
@@ -12295,7 +12298,12 @@ then
     # Uses alt name
     ENABLED_ALTNAMES="yes"
 
-    AM_CFLAGS="$AM_CFLAGS -DHAVE_OID_ENCODING -DWOLFSSL_NO_ASN_STRICT"
+    AM_CFLAGS="$AM_CFLAGS -DHAVE_OID_ENCODING"
+
+    if test "$ENABLED_FIPS" = "no"
+    then
+        AM_CFLAGS="$AM_CFLAGS -DWOLFSSL_NO_ASN_STRICT"
+    fi
 
     # wolfCLU allows MD5 with wc_Signature*; lower the hash strength floor
     # when MD5 is compiled in.


configure.ac
sebastian-carpenter, line 12266: MEDIUM-3: New MD5 gate is defeated by --enable-opensslextra/all and yields an OPENSSL_ALL + NO_MD5 combination configure.ac elsewhere prevents [SUGGEST] (bug)
  File:       configure.ac:12244-12247
  Function:   ENABLED_WOLFCLU block
  Confidence: Medium

  The wolfCLU block unconditionally turns on OpenSSL-All a few lines below the changed MD5 gate:

```sh
# Uses functions guarded by opensslall
if test "$ENABLED_OPENSSLALL" = "no"
then
    ENABLED_OPENSSLALL="yes"
fi
```

and `ENABLED_OPENSSLALL=yes` propagates to `ENABLED_OPENSSLEXTRA="yes"` at `configure.ac:12338`, which produces `-DOPENSSL_ALL -DOPENSSL_EXTRA` at `configure.ac:12643` and `configure.ac:12664`.

But configure.ac has an explicit "Options that require MD5" rule at lines 6752-6763 which force-enables MD5 whenever `ENABLED_OPENSSLEXTRA`/`ENABLED_OPENSSLALL` is yes — with **no FIPS guard at all**. Two consequences of the new gate:

1. **The gate does not achieve its goal.** `--enable-wolfclu --enable-opensslall --enable-fips=v5` still ends with `ENABLED_MD5=yes` (set at line 6762 before the wolfCLU block runs), so MD5 is still compiled into the FIPS build. Only the bare `--enable-wolfclu --enable-fips=v5` invocation gets MD5 off — the outcome now depends on whether the user spelled out an option wolfCLU turns on anyway.
2. **It creates a combination nothing else produces.** For `--enable-wolfclu --enable-fips=v5`, the build gets `-DOPENSSL_ALL -DOPENSSL_EXTRA` together with `-DNO_MD5 -DNO_OLD_TLS` (`configure.ac:12651-12654`). Every other meta-option that turns on OPENSSL_ALL (asio at line 10015, stunnel at line 10223, and the generic rule at 6753) explicitly forces MD5 on, so `OPENSSL_ALL && NO_MD5` appears to be an untested configuration and may not compile.

The same reasoning applies to the pre-existing DES3 gate at line 12284, but that one at least does not conflict with a force-enable rule that runs earlier.

  Code:
        # Requires md5
        if test "$ENABLED_MD5" = "no" && (test "$ENABLED_FIPS" = "no" || test $HAVE_FIPS_VERSION -lt 5)
        then
            ENABLED_MD5="yes"
        fi
        ...
        # Uses functions guarded by opensslall
        if test "$ENABLED_OPENSSLALL" = "no"
        then
            ENABLED_OPENSSLALL="yes"
        fi

  Recommendation: Verify that `--enable-wolfclu --enable-fips=v5` (OPENSSL_ALL with NO_MD5) actually compiles, and decide whether the MD5-under-FIPS policy belongs on the generic "Options that require MD5" rule at line 6753 rather than only on the wolfCLU branch — otherwise the gate is trivially bypassed.

configure.ac
```diff
@@ -12242,7 +12242,7 @@ then
     fi
 
     # Requires md5
```
sebastian-carpenter: LOW-6: Comments no longer describe the conditional behavior, and the -lt 5 / -ge 6 threshold asymmetry is unexplained [NIT] (style)
  File:       configure.ac:12244-12272
  Function:   ENABLED_WOLFCLU block
  Confidence: High

  `# Requires md5` and `# Has option for signing with ED25519` were accurate when the enables were unconditional; now they read as contradictions of the code (if wolfCLU truly *requires* MD5, why is it conditional?). The two thresholds also differ — `-lt 5` for MD5 versus `-ge 6` for ED25519 — which encodes real but undocumented knowledge (MD5 is out of the FIPS 140-3 boundary from v5 on; ED25519 only entered the module at v6, per the force-enables at `configure.ac:7087` and `configure.ac:7328`). A reader hitting the odd case of FIPS v2 (MD5 on, ED25519 off) has no way to tell that is intentional.

  Code:
        # Requires md5
        if test "$ENABLED_MD5" = "no" && (test "$ENABLED_FIPS" = "no" || test $HAVE_FIPS_VERSION -lt 5)
        ...
        # Has option for signing with ED25519
        if test "$ENABLED_ED25519" = "no" && (test "$ENABLED_FIPS" = "no" || test $HAVE_FIPS_VERSION -ge 6)

  Recommendation: Extend each comment with the FIPS rationale so the differing version thresholds are self-explanatory.

configure.ac
```diff
@@ -12266,7 +12266,7 @@ then
     fi
 
     # Has option for signing with ED25519
-    if test "$ENABLED_ED25519" = "no"
+    if test "$ENABLED_ED25519" = "no" && (test "$ENABLED_FIPS" = "no" || test $HAVE_FIPS_VERSION -ge 6)
     then
```
sebastian-carpenter, line 12273: LOW-7: ED25519 clause is redundant for FIPS v6/v7 and overrides an explicit --disable-ed25519 in FIPS dev/ready builds [NIT] (question)
  File:       configure.ac:12268-12275
  Function:   ENABLED_WOLFCLU block
  Confidence: Medium

  For every FIPS flavor where `HAVE_FIPS_VERSION -ge 6` holds, the FIPS case blocks already force ED25519 on before this code runs — `configure.ac:7087` (v7/ready/dev/dev-no-post) and `configure.ac:7328` (v6):

```sh
AS_IF([test "$ENABLED_ED25519" = "no" &&
       (test "$ENABLED_FIPS_DEV" != "yes" || test "$enable_ed25519" != "no")],
    [ENABLED_ED25519="yes"; AM_CFLAGS="$AM_CFLAGS -DHAVE_ED25519 -DHAVE_ED25519_KEY_IMPORT"])
```

So `ENABLED_ED25519` is already `yes` by the time the wolfCLU block is reached, and the new `-ge 6` clause is a no-op — *except* in exactly one case: a FIPS dev/ready build where the user passed `--disable-ed25519`. There the FIPS block deliberately honors the user's opt-out, and the wolfCLU block then silently flips it back to `yes`, overriding an explicit user request. (The old unconditional code had the same wart, so this is not a new bug — but the line is being touched here and the clause could be made to respect it.)

Also note `ENABLED_FEMATH=yes; ENABLED_GEMATH=yes; ENABLED_CERTS=yes` inside this branch are already handled globally at `configure.ac:12391-12405` for any `ENABLED_ED25519 != "no"`, so they are dead weight here.

  Code:
        # Has option for signing with ED25519
        if test "$ENABLED_ED25519" = "no" && (test "$ENABLED_FIPS" = "no" || test $HAVE_FIPS_VERSION -ge 6)
        then
            ENABLED_ED25519=yes
            ENABLED_FEMATH=yes
            ENABLED_GEMATH=yes
            ENABLED_CERTS=yes
        fi

  Recommendation: Consider also checking `test "$enable_ed25519" != "no"` so an explicit `--disable-ed25519` is not silently reversed, and confirm whether the `-ge 6` clause is intended to be reachable at all given the v6/v7 FIPS blocks already enable ED25519.

sebastian-carpenter (commented):
MEDIUM-2: CMake wolfCLU option not updated — autotools and CMake --enable-wolfclu/WOLFSSL_WOLFCLU now produce different builds [SUGGEST] (convention)
  File:       configure.ac:12244-12298
  Function:   ENABLED_WOLFCLU block
  Confidence: High

  `CMakeLists.txt:2950-2959` mirrors this same wolfCLU feature block and was not touched by the diff:

```cmake
override_cache(WOLFSSL_MD5      "yes")
...
override_cache(WOLFSSL_ED25519  "yes")
override_cache(WOLFSSL_DES3     "yes")
list(APPEND WOLFSSL_DEFINITIONS "-DHAVE_OID_ENCODING" "-DWOLFSSL_NO_ASN_STRICT" "-DWOLFSSL_ALT_NAMES")
```

After this diff the two build systems disagree in two ways for a wolfCLU build: (1) CMake still defines `WOLFSSL_NO_ASN_STRICT` while autotools no longer does, so the same option yields different ASN.1 parsing strictness depending on build system; (2) CMake still force-enables MD5/ED25519/DES3 with no FIPS-version gating, so the "limit wolfCLU features on FIPS version" intent is autotools-only. `AGENTS.md` states the option rule is `--enable-foo` == `-DWOLFSSL_FOO=yes`, i.e. the two paths are expected to stay in lockstep.

  Code:
    # CMakeLists.txt:2959 (unchanged by this diff)
    list(APPEND WOLFSSL_DEFINITIONS "-DHAVE_OID_ENCODING" "-DWOLFSSL_NO_ASN_STRICT" "-DWOLFSSL_ALT_NAMES")

  Recommendation: Update `CMakeLists.txt` in the same PR so the autotools and CMake wolfCLU builds stay equivalent, or state explicitly in the commit message that CMake FIPS variants are out of scope.

configure.ac
lealem47: fixed

configure.ac
lealem47: fixed

configure.ac
lealem47: Not possible to disable that alg with fips >=v6 anyway

lealem47:
> MEDIUM-2: CMake wolfCLU option not updated — autotools and CMake --enable-wolfclu/WOLFSSL_WOLFCLU now produce different builds [SUGGEST] (convention) File: configure.ac:12244-12298 Function: ENABLED_WOLFCLU block Confidence: High
> 
> `CMakeLists.txt:2950-2959` mirrors this same wolfCLU feature block and was not touched by the diff:
> 
> ```cmake
> override_cache(WOLFSSL_MD5      "yes")
> ...
> override_cache(WOLFSSL_ED25519  "yes")
> override_cache(WOLFSSL_DES3     "yes")
> list(APPEND WOLFSSL_DEFINITIONS "-DHAVE_OID_ENCODING" "-DWOLFSSL_NO_ASN_STRICT" "-DWOLFSSL_ALT_NAMES")
> ```
> 
> After this diff the two build systems disagree in two ways for a wolfCLU build: (1) CMake still defines `WOLFSSL_NO_ASN_STRICT` while autotools no longer does, so the same option yields different ASN.1 parsing strictness depending on build system; (2) CMake still force-enables MD5/ED25519/DES3 with no FIPS-version gating, so the "limit wolfCLU features on FIPS version" intent is autotools-only. `AGENTS.md` states the option rule is `--enable-foo` == `-DWOLFSSL_FOO=yes`, i.e. the two paths are expected to stay in lockstep.
> 
> Code: # CMakeLists.txt:2959 (unchanged by this diff) list(APPEND WOLFSSL_DEFINITIONS "-DHAVE_OID_ENCODING" "-DWOLFSSL_NO_ASN_STRICT" "-DWOLFSSL_ALT_NAMES")
> 
> Recommendation: Update `CMakeLists.txt` in the same PR so the autotools and CMake wolfCLU builds stay equivalent, or state explicitly in the commit message that CMake FIPS variants are out of scope.

CMake doesn't support fips so is out of scope for the changes targetted here

sebastian-carpenter (approved):

JacobBarthelmeh (approved):

PR #11145, opened 2026-08-11
Enforce Extended Key Usage on chain-supplied intermediate CAs

embhorn:
# Description

The peer certificate is checked for the serverAuth/clientAuth EKU, but the intermediate CAs sent with it were not. A CA constrained to another purpose by a critical EKU, a code signing sub-CA for example, could issue a serverAuth leaf for any hostname and wolfSSL would complete the handshake.

ProcessPeerCerts() now applies the same purpose rule to every chain CA it validates and fails with EXTKEYUSE_AUTH_E. An absent extension and anyExtendedKeyUsage leave all purposes valid per RFC 5280 4.2.1.12, and a self-signed certificate is exempt since it can only serve as an operator-loaded trust anchor. IGNORE_KEY_EXTENSIONS opts out, as it already did for the peer certificate.

Fixes zd22309

# Testing

Added `test_wolfSSL_chain_ca_ext_key_usage()`

# Checklist

 - [x] added tests
 - [ ] updated/added doxygen


Frauschi:
The chain CA check here is a real gap worth closing, and the enforcement itself looks right. One follow-up on the self-signed exemption in `CheckChainCAExtKeyUsage()`:

```c
if (!cert->extExtKeyUsageSet || cert->selfSigned)
    return 0;
```

`DecodedCert.selfSigned` is set purely by comparing the issuer and subject name hashes in `DecodeCertInternal()`, so it means *self-issued* (RFC 5280 3.2), not *self-signed*. asn.c already notes this in the serial-0 comment: "cert->selfSigned is a subject/issuer name-hash compare ... not a validated self-signature".

That matters for a CA key-rollover link certificate (RFC 4210 OldWithNew/NewWithOld), which is self-issued but signed by the CA's previous key:

```
Subject:    CN=Acme Root      <- same name
Issuer:     CN=Acme Root      <- same name  => selfSigned == 1
Public key: K2                <- the new key
Signature:  made with K1      <- NOT self-signed
```

Such a certificate is a genuine, fully verified, non-anchor link in the path: `ProcessPeerCertParse()` finds the old root as its issuer and confirms the signature. But because the names match, the new EKU check returns 0 before ever looking at the extension.

## Repro

Built this branch with `--enable-all --enable-debug` and ran a memio handshake. Root is a self-signed CA loaded as the only trust anchor; the CA under test carries `codeSigning` only and issues a `serverAuth` leaf:

| CA under test | `selfSigned` | Handshake | `wolfSSL_get_verify_result()` |
|---|---|---|---|
| Subject name equal to the root's | 1 | **accepted (0)** | 0 (`X509_V_OK`) |
| Byte-identical cert, different subject name | 0 | rejected (`EXTKEYUSE_AUTH_E`) | 26 (`INVALID_PURPOSE`) |

Only the subject name differs between those two rows.

To be clear about scope: this is not a regression from this PR. Before it there was no chain CA EKU check at all, so every chain CA was exempt. This PR strictly narrows the hole; the self-issued case just remains open. Exploitability is also limited, since the issuing parent chooses the subject DN, so an attacker cannot generally obtain a purpose-constrained CA whose DN equals its issuer's. The realistic impact is an operator whose own EKU-constrained CA silently stops being constrained after a key rollover.

## Suggested fix

The exemption is load-bearing and should not simply be dropped: the self-signed-anchor case in the new test depends on it. But a true anchor also verified against *itself*, so matching the public key separates it from a rollover cert. This is the same trust-anchor test asn.c already uses for the RFC 5280 6.1(l) pathlen exclusion, and `Signer.publicKey` / `Signer.pubKeySize` are unconditional members, so no new build gating is needed:

```diff
-    if (!cert->extExtKeyUsageSet || cert->selfSigned)
+    if (!cert->extExtKeyUsageSet)
         return 0;
 
+    /* A trust anchor the operator loaded is exempt: RFC 5280 6.1 keeps the
+     * anchor outside the prospective certification path. cert->selfSigned is
+     * only an issuer/subject name compare, so the anchor's public key has to
+     * match too - a self-issued key-rollover CA carries the same names but a
+     * different key and stays subject to the check. */
+    if (cert->selfSigned && cert->ca != NULL && cert->publicKey != NULL &&
+            cert->ca->publicKey != NULL && cert->pubKeySize > 0 &&
+            cert->pubKeySize == cert->ca->pubKeySize &&
+            XMEMCMP(cert->publicKey, cert->ca->publicKey,
+                    cert->pubKeySize) == 0) {
+        return 0;
+    }
+
     if (ssl->options.side == WOLFSSL_CLIENT_END)
         purpose = EXTKEYUSE_SERVER_AUTH;
     else
```

On `cert->ca` being available: I instrumented the function and traced every call. `ret == 0` from `ProcessPeerCertParse()` on a chain cert implies `cert->ca != NULL`, because the `verifyNone` branch above returns before reaching this check, which leaves `verify` as `VERIFY` (or `VERIFY_NAME` under `WOLFSSL_SMALL_CERT_VERIFY`) with `type == CHAIN_CERT_TYPE`, so the lookup block always runs; and a failed lookup returns `ASN_SELF_SIGNED_E` / `ASN_NO_SIGNER_E` because the `else { no signer }` hangs off `if (cert->ca)` rather than the inner verify-mode test. The NULL guard above is defensive only.

Verified: every trace line showed the CA set, with a key match for the real anchor and no key match for the rollover CA.

## Two smaller points

1. The function's doc comment still says "A self-signed certificate is exempt: it can only take part in a path as a trust anchor the operator chose to load." That premise is what the above disproves, so it would be worth rewording to reference the anchor key match.

2. The new test cannot catch this shape. The `selfSignedCa` case sets `rootDer = interDer`, making the CA the actual loaded anchor. A regression case needs identical issuer/subject DNs with a *distinct* key. I have one written against your `test_eku_gen_ca()` helper and can post it if useful.

## Testing

- `--enable-all --enable-debug`: full `./tests/unit.test --api` is 0 failed / 1915 passed with the patch, and `test_wolfSSL_chain_ca_ext_key_usage` still passes all 16 cases.
- `--enable-altcertchains CFLAGS=-DWOLFSSL_SMALL_CERT_VERIFY`: 3 failures, all in the TLS 1.3 RPK tests. Those reproduce identically on master at 63b02a158 and with the patch reverted, so they are pre-existing and unrelated. They isolate to `WOLFSSL_SMALL_CERT_VERIFY` alone; I will report that separately.

philljj (changes requested):
See @Frauschi 's analysis.

embhorn:
retest this please

embhorn:
retest this please

Frauschi (dismissed):
LGTM

philljj (changes requested):
merge conflict in changelog

src/internal.c
```diff
@@ -18287,14 +18343,39 @@ int ProcessPeerCerts(WOLFSSL* ssl, byte* input, word32* inOutIdx,
...
                     else if (ret == 0) {
-                    #ifdef OPENSSL_EXTRA
-                        if (args->certIdx > args->untrustedDepth) {
-                            args->untrustedDepth = (char)args->certIdx + 1;
+                    #ifndef IGNORE_KEY_EXTENSIONS
+                        /* A CA restricted to some other purpose by its
+                         * Extended Key Usage must not authenticate this peer,
+                         * whether or not the certificate manager already
+                         * holds it. */
+                        ret = CheckChainCAExtKeyUsage(ssl, args->dCert);
```
Frauschi: This runs ahead of the OCSP, CRL and chain-depth checks further down, and all three are gated on `ret == 0`. A CA rejected here skips them, and `DoVerifyCallback()` at the bottom of the block can then clear the EKU error - so a callback that waives only `INVALID_PURPOSE` silently waives whatever those checks would have found.

I reproduced it with a memio handshake at this head under `--enable-all --enable-debug`. Chain is `leaf <- midCA(serverAuth,clientAuth) <- interCA`, verify depth pinned to 1, and a verify callback that returns 1 only for `WOLFSSL_X509_V_ERR_INVALID_PURPOSE` and lets everything else stand:

| interCA EKU | this branch | master |
|---|---|---|
| serverAuth | rejected, `MAX_CHAIN_ERROR` | rejected, `MAX_CHAIN_ERROR` |
| codeSigning | handshake accepted | rejected, `MAX_CHAIN_ERROR` |

Same chain, same depth limit, same callback - only the EKU on the top CA differs. Both depth paths behave this way: `ssl->verifyDepth` at line 18507, and `options.verifyDepth` via `untrustedDepth` after the loop.

What worked here was moving the check to just after the depth check and before the `WOLFSSL_ALT_CERT_CHAINS` block, guarded `if (ret == 0 && args->dCert->isCA && !ssl->options.verifyNone)` - those two guards replace the gating this `else if` chain is giving you at the current site. It also lets the original `untrustedDepth` update stand unchanged, which I think is more correct anyway: a CA the callback waives is part of the chain and ought to count toward the depth.

With that plus a case that pins the verify depth and uses the selective callback, I get 22 EKU cases passing, `unit.test --api` 0 failed, and `make check` clean; reverting just the internal.c hunk fails the new cases. Glad to send the patch if it helps.

tests/api/test_ssl_cert.c
```diff
@@ -1498,3 +1499,427 @@ int test_wolfSSL_cert_unload(void)
...
+    int midSz = 0;
+    int caSz = 0;
+    int leafSz = 0;
+    int chainSz = 0;
+    int hsRet = -1;
+    long verifyRet = -1;
+
+    /* An extra level puts the CA under test at depth 2, below a CA that does
+     * carry the TLS purposes. */
+    if (tc->extraLevel) {
```
Frauschi, line 1784: The `extraLevel` cases don't actually reach past depth 1. `midDer` is signed by the root, the CA under test is signed by `midDer`, and the leaf is then signed by the CA under test - so the transmitted chain is leaf, CA-under-test, midDer, and the constrained CA sits at depth 1 exactly like the ordinary cases. The comment here and the "not limited to the CA directly above the leaf" rows below describe something the test isn't doing.

Turning it around works: sign the CA under test with the root, sign `midDer` with the CA under test, sign the leaf with `midDer`, and transmit leaf, midDer, CA-under-test. I checked the result by issuer name - the leaf's issuer becomes "EKU Middle CA" and midDer's becomes "EKU Intermediate".

Worth saying that both depth-2 rows pass unchanged once the chain is built that way, so this is only missing coverage, not missing enforcement.

Frauschi (changes requested):
Second pass over this branch. The public-key anchor test looks right to me now, thanks for picking that up.

One of the two below is the reason I'm requesting changes: putting the purpose check ahead of the revocation and depth checks means a verify callback that waives only `INVALID_PURPOSE` also waives those, and I have a handshake where this branch accepts a chain master rejects. The other is a test-shape point with no defect behind it.

philljj (dismissed):
merge conflict in tests

JacobBarthelmeh (approved):

Frauschi (approved):

PR #4257, opened 2021-07-28
Add support for EVP_sha512_224/256

TakayukiMatsuo:
In this PR, following APIs are added:

- wolfSSL_EVP_sha512_224
- wolfSSL_EVP_sha512_256
- wc_Sha512_224/2256Hash
- wc_Sha512_224/256Hash
- wc_InitSha512_224/256
- wc_InitSha512_224/256_ex
- wc_Sha512_224/256Update
- wc_Sha512_224/256FinalRaw
- wc_Sha512_224/256Final
- wc_Sha512_224/256Free
- wc_Sha512_224/256GetHash
- wc_Sha512_224/256Copy
- wc_Sha512_224/256SetFlags
- wc_Sha512_224/256GetFlags
- wc_Sha512_224/256Transform

PR #11153, opened 2026-08-12
Fix TSP message imprint algorithm/digest length consistency

yosuke-wolfssl:
### Problem
`f-7075`: the `TspRequest` message imprint was configured by two setters that both wrote `imprint.hashSz` with no cross-validation, so the field did not distinguish "a digest was supplied" from "the algorithm implies this length".

- **Forgotten digest.** `wc_TspRequest_Init()` + `wc_TspRequest_SetHashType(WC_HASH_TYPE_SHA256)` left `hashSz = 32` with `imprint.hash` still all zero. `wc_TspRequest_Encode()` only checked `hashSz == 0`, so it returned 0 and encoded a TimeStampReq whose messageImprint was 32 zero bytes — the caller obtains a token over a fixed, attacker-known digest instead of over their data.
- **Either call order.** `SetHash(sha1Digest, 20)` then `SetHashType(SHA-256)` raised `hashSz` to 32, encoding 20 real bytes plus 12 zero bytes tagged SHA-256; the reverse order encoded a 20-byte imprint tagged SHA-256. Neither was rejected.

`wc_TspTstInfo_SetMsgImprint()` and `wc_TspTstInfo_Encode()` had the same `hashSz != 0`-only check. Medium severity, unsafe default configuration. TSP is unreleased (added after v5.9.2), so the setter contract is tightened rather than kept compatible.

### Fix (`wolfcrypt/src/tsp.c`, `wolfcrypt/src/asn_tsp.c`, `wolfcrypt/src/hash.c`)
`hashSz` now means only "a digest was supplied"; the expected length is derived from `imprint.hashAlgOID` via `wc_HashGetDigestSize(wc_OidGetHash(...))`. No new struct member, no ABI change.

| Function | Change |
|---|---|
| `wc_TspRequest_SetHashType()` | Sets the OID and **invalidates** any digest (`imprint.hash` zeroed, `hashSz = 0`) instead of resizing it; rejects a hash type whose OID does not map back to it |
| `wc_TspRequest_SetHash()` | `HASH_TYPE_E` when the algorithm is unset/unavailable, `BUFFER_E` when `hashSz` is not its digest size |
| `wc_TspTstInfo_SetMsgImprint()` | `HASH_TYPE_E` when the OID has no determinable digest size, `BUFFER_E` on a length mismatch |
| `wc_TspRequest_Encode()`, `wc_TspTstInfo_Encode()` | Reject `hashSz == 0` (never-supplied digest) and a determinable length mismatch |
| `wc_OidGetHash()` | Added the missing `SHA512_224h`/`SHA512_256h` cases so those digest lengths are determinable |

Both setters now require a determinable digest size. The encode-time check stays conditional on the algorithm being known, so a `TspTstInfo` decoded off the wire can still be re-encoded in a build lacking that algorithm. Decoders are unchanged: a verifier still sees what arrived, and a malformed imprint fails closed at `wc_TspTstInfo_Encode()` / `wc_TspTstInfo_CheckRequest()`. The OpenSSL compat layer writes the imprint fields directly, so order-independent `TS_MSG_IMPRINT_set_algo`/`_set_msg` semantics are preserved and it gains only the fail-closed encode check.

`WC_HASH_TYPE_MD5_SHA` is the one hash type whose OID does not identify it — `wc_HashGetOID()` maps the TLS 1.0/1.1 concatenation to `MD5h`. Without the round-trip check `SetHashType()` accepted it and the request encoded with a plain MD5 imprint, silently substituting the algorithm. Every other hash type round-trips, so the check costs only that one. Plain MD5 is still accepted; `WC_TSP_MIN_HASH_STRENGTH_BITS` covers strength policy.

The `wc_OidGetHash()` gap came out of review: `wc_HashGetOID()` has mapped those hash types to those OIDs since #4257 (2021), but the reverse function, written in 2018, was never updated. Deriving the expected length through it therefore made SHA-512/224 and SHA-512/256 imprints unbuildable (`SetHash()` always returned `HASH_TYPE_E`) and silently skipped the encode-time length check for them. The two cases carry the same guards `wc_HashGetDigestSize()` uses for these types, including the `HAVE_FIPS`/`HAVE_SELFTEST` terms, so the mapping never reports a type the build cannot hash.

Being the shared reverse lookup, this also reaches other callers: PKCS#7/CMS SignedData **verification** now accepts these digest algorithms (confirmed against OpenSSL-generated CMS), as do OCSP CertID hashes and the PKCS#12 MAC, all of which validate the length against the digest size before use. PKCS#7 **signing** still rejects them — `wc_PKCS7_SignedDataGetEncAlgoId()` has no signature-algorithm mapping — and the OpenSSL-compat RSA-PSS path is unaffected, since `nid2oid()` has no `oidHashType` entry for these NIDs.

Closes `f-7075`.

### Tests
`tests/api/test_tsp.c`: forgotten digest fails to encode; `SetHash()` then `SetHashType()` discards the digest and fails to encode; `SetHash()` with no algorithm → `HASH_TYPE_E`; wrong length → `BUFFER_E` for both setters; encode-time length mismatch for request and TSTInfo; an unknown algorithm → `HASH_TYPE_E` from `SetMsgImprint()`; `WC_HASH_TYPE_MD5_SHA` → `HASH_TYPE_E` while plain MD5 is accepted; SHA-512/224 imprint set and encoded end to end. `tests/api/test_hash.c`: both OIDs added to the supported and not-supported lists of `test_wc_OidGetHash()`, and the `wc_HashGetOID()`/`wc_OidGetHash()` round trip in `test_wc_HashFeatureCoverage()` no longer excludes them.

Config-sensitive assertions are guarded to match the code: the SHA-512/224 case and the `test_wc_OidGetHash()` lists use the same `(!HAVE_FIPS || FIPS_VERSION3_GE(7,0,0)) && !HAVE_SELFTEST` condition as `wc_HashGetDigestSize()`, and the SHA-256 assertions in `test_wc_TspTstInfo_Setters()` are guarded with `NO_SHA256` since `--enable-tsp` does not pull in SHA-256.

### Verification
- Build clean, no warnings; all GitHub CI checks green.
- `make check`: 6 passed, 4 skipped, 0 failed (includes `scripts/tsp.test` and the full `unit.test`); API suite 0 failed / 1071 passed of 2247.
- Builds and tests clean with `-DWOLFSSL_NOSHA512_224 -DWOLFSSL_NOSHA512_256` (0 failed / 1049 passed).
- Negative controls: restoring `hashSz = digestSz` fails the forgotten-digest test; dropping the encode length term fails both mismatch tests; removing the `SHA512_224h` case fails the round-trip and `wc_OidGetHash()` tests; swapping the `SHA512_224h`/`SHA512_256h` mappings fails `test_wc_HashFeatureCoverage()`.

### DIFF
--- a/tests/api/test_hash.c
+++ b/tests/api/test_hash.c
@@ -780,6 +780,15 @@ int test_wc_OidGetHash(void)
     #endif
     #ifdef WOLFSSL_SHA512
         SHA512h,
+        #if (!defined(HAVE_FIPS) || FIPS_VERSION3_GE(7,0,0)) && \
+            !defined(HAVE_SELFTEST)
+        #ifndef WOLFSSL_NOSHA512_224
+        SHA512_224h,
+        #endif
+        #ifndef WOLFSSL_NOSHA512_256
+        SHA512_256h,
+        #endif
+        #endif
     #endif
     #ifdef WOLFSSL_SHA3
         SHA3_224h,
@@ -813,6 +822,19 @@ int test_wc_OidGetHash(void)
     #endif
     #ifndef WOLFSSL_SHA512
         SHA512h,
+        SHA512_224h,
+        SHA512_256h,
+    #elif defined(HAVE_SELFTEST) || \
+          (defined(HAVE_FIPS) && !FIPS_VERSION3_GE(7,0,0))
+        SHA512_224h,
+        SHA512_256h,
+    #else
+        #ifdef WOLFSSL_NOSHA512_224
+        SHA512_224h,
+        #endif
+        #ifdef WOLFSSL_NOSHA512_256
+        SHA512_256h,
+        #endif
     #endif
     #ifndef WOLFSSL_SHA3
         SHA3_224h,
@@ -1088,11 +1110,7 @@ int test_wc_HashFeatureCoverage(void)
         {
             int oid = wc_HashGetOID(supportedHash[i]);
             ExpectIntGT(oid, 0);
-            /* wc_OidGetHash() has no case for SHA512_224h/SHA512_256h (no
-             * OID assigned upstream for those two truncated variants), so
-             * the round trip only holds for the other digests. */
-            if (oid > 0 && supportedHash[i] != WC_HASH_TYPE_SHA512_224 &&
-                supportedHash[i] != WC_HASH_TYPE_SHA512_256) {
+            if (oid > 0) {
                 enum wc_HashType rtType = wc_OidGetHash(oid);
                 ExpectIntEQ((int)rtType, (int)supportedHash[i]);
             }
--- a/wolfcrypt/src/asn_tsp.c
+++ b/wolfcrypt/src/asn_tsp.c
@@ -112,9 +112,8 @@ enum {
  * @param [in, out] outSz  On in, length of buffer in bytes.
  *                         On out, length of encoding in bytes.
  * @return  0 on success.
- * @return  BAD_FUNC_ARG when req or outSz is NULL, the message imprint hash
- *          is not set, a field is too long for its array or the nonce has a
- *          leading zero byte.
+ * @return  BAD_FUNC_ARG when req or outSz is NULL, or a field is unset, the
+ *          wrong length or not encodable as given.
  * @return  BUFFER_E when out is not NULL and encoding is longer than outSz.
  * @return  ASN_UNKNOWN_OID_E when the hash algorithm is not recognized.
  * @return  MEMORY_E on dynamic memory allocation failure.
@@ -124,6 +123,7 @@ int wc_TspRequest_Encode(const TspRequest* req, byte* out, word32* outSz)
     DECL_ASNSETDATA(dataASN, tspReqASN_Length);
     int ret = 0;
     word32 sz = 0;
+    int digestSz = 0;
 
     WOLFSSL_ENTER("wc_TspRequest_Encode");
 
@@ -131,10 +131,17 @@ int wc_TspRequest_Encode(const TspRequest* req, byte* out, word32* outSz)
     if ((req == NULL) || (outSz == NULL)) {
         ret = BAD_FUNC_ARG;
     }
-    /* The message imprint is the only required field. */
-    if ((ret == 0) && ((req->imprint.hashSz == 0) ||
-            (req->imprint.hashSz > sizeof(req->imprint.hash)))) {
-        ret = BAD_FUNC_ARG;
+    /* The message imprint is the only required field. Its length is checked
+     * against the hash algorithm when the algorithm is known and available. */
+    if (ret == 0) {
+        digestSz = wc_HashGetDigestSize(
+            wc_OidGetHash((int)req->imprint.hashAlgOID));
+        if ((req->imprint.hashSz == 0) ||
+                (req->imprint.hashSz > sizeof(req->imprint.hash)) ||
+                ((digestSz > 0) &&
+                    (req->imprint.hashSz != (word32)digestSz))) {
+            ret = BAD_FUNC_ARG;
+        }
     }
     /* Policy, when set, must fit. */
     if ((ret == 0) && (req->policySz > sizeof(req->policy))) {
@@ -447,11 +454,8 @@ enum {
  * @param [in, out] outSz    On in, length of buffer in bytes.
  *                           On out, length of encoding in bytes.
  * @return  0 on success.
- * @return  BAD_FUNC_ARG when tstInfo or outSz is NULL, a required field of
- *          tstInfo is not set or empty, the hash is too long, the genTime
- *          is not a valid GeneralizedTime, the tsa is empty, the serial
- *          number or nonce is empty or has a leading zero byte or accuracy
- *          millis or micros is out of range.
+ * @return  BAD_FUNC_ARG when tstInfo or outSz is NULL, or a field is unset,
+ *          the wrong length or not encodable as given.
  * @return  BUFFER_E when out is not NULL and encoding is longer than outSz.
  * @return  ASN_UNKNOWN_OID_E when the hash algorithm is not recognized.
  * @return  ASN_TIME_E when getting the current time failed.
@@ -462,6 +466,7 @@ int wc_TspTstInfo_Encode(const TspTstInfo* tstInfo, byte* out, word32* outSz)
     DECL_ASNSETDATA(dataASN, tspTstInfoASN_Length);
     int ret = 0;
     word32 sz = 0;
+    int digestSz = 0;
 #if !defined(NO_ASN_TIME) && !defined(USER_TIME) && !defined(TIME_OVERRIDES)
     byte timeBuf[ASN_GENERALIZED_TIME_SIZE];
 #endif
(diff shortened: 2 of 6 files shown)


wolfcrypt/src/tsp.c
```diff
@@ -209,6 +214,17 @@ int wc_TspRequest_SetHash(TspRequest* req, const byte* hash, word32 hashSz)
         ret = BUFFER_E;
     }
 
+    if (ret == 0) {
+        /* The digest must be the length of the algorithm already set. */
+        digestSz = wc_HashGetDigestSize(
+            wc_OidGetHash((int)req->imprint.hashAlgOID));
```
yosuke-wolfssl, line 226: Fixed

wolfcrypt/src/tsp.c
```diff
@@ -209,6 +214,17 @@ int wc_TspRequest_SetHash(TspRequest* req, const byte* hash, word32 hashSz)
...
+        if (digestSz <= 0) {
```
yosuke-wolfssl, line 227: Fixed

wolfcrypt/src/asn_tsp.c
```diff
@@ -124,17 +123,25 @@ int wc_TspRequest_Encode(const TspRequest* req, byte* out, word32* outSz)
...
-            (req->imprint.hashSz > sizeof(req->imprint.hash)))) {
-        ret = BAD_FUNC_ARG;
+    /* The message imprint is the only required field. Its length is checked
+     * against the hash algorithm when the algorithm is known and available. */
+    if (ret == 0) {
+        digestSz = wc_HashGetDigestSize(
+            wc_OidGetHash((int)req->imprint.hashAlgOID));
+        if ((req->imprint.hashSz == 0) ||
+                (req->imprint.hashSz > sizeof(req->imprint.hash)) ||
+                ((digestSz > 0) &&
```
yosuke-wolfssl, line 141: Fixed

wolfcrypt/src/hash.c
```diff
@@ -325,6 +325,20 @@ enum wc_HashType wc_OidGetHash(int oid)
             hash_type = WC_ERR_TRACE(WC_HASH_TYPE_NONE);
         #endif
             break;
+        case SHA512_224h:
+        #if defined(WOLFSSL_SHA512) && !defined(WOLFSSL_NOSHA512_224)
```
yosuke-wolfssl: Won't do

tests/api/test_hash.c
```diff
@@ -780,6 +780,12 @@ int test_wc_OidGetHash(void)
     #endif
     #ifdef WOLFSSL_SHA512
         SHA512h,
+        #ifndef WOLFSSL_NOSHA512_224
+        SHA512_224h,
```
yosuke-wolfssl, line 786: Fixed

yosuke-wolfssl:
Jenkins retest please

yosuke-wolfssl:
retest this please

wolfcrypt/src/hash.c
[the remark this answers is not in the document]
philljj: These seem valid. Most occurrences of `WC_HASH_TYPE_SHA512_224` and `WC_HASH_TYPE_SHA512_256` have FIPS define guards.

wolfcrypt/src/tsp.c
```diff
@@ -209,6 +214,17 @@ int wc_TspRequest_SetHash(TspRequest* req, const byte* hash, word32 hashSz)
         ret = BUFFER_E;
     }
 
+    if (ret == 0) {
+        /* The digest must be the length of the algorithm already set. */
+        digestSz = wc_HashGetDigestSize(
```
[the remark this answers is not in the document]
philljj: This looks valid. It doesn't look like we restrict MD5 for TSP, but I can't imagine anyone using it.

The smallest hashes we test with in wolfcrypt test.c are:

```
#ifndef NO_SHA256
    #define TSP_TEST_HASH_OID   SHA256h
    #define TSP_TEST_HASH_SZ    32                 
#elif !defined(NO_SHA)
    #define TSP_TEST_HASH_OID   SHAh 
    #define TSP_TEST_HASH_SZ    20
#endif 
```

If it could lead to an OOB read / write then let's fix it it. Maybe just reject MD5 completely, or whatever makes sense.

If it only leads to a returned error, then maybe document we don't support MD5 (which shouldn't be a big deal).

wolfcrypt/src/tsp.c
```diff
@@ -594,18 +610,27 @@ int wc_TspTstInfo_GetMsgImprint(const TspTstInfo* tstInfo, word32* hashOID,
...
     /* Validate parameters. */
     if ((tstInfo == NULL) || (hash == NULL) || (hashSz == 0)) {
         return BAD_FUNC_ARG;
     }
     if (hashSz > sizeof(tstInfo->imprint.hash)) {
         return BUFFER_E;
     }
+    /* The digest size of an unknown or unavailable algorithm is not known -
+     * the length is only checked against a known one. */
+    digestSz = wc_HashGetDigestSize(wc_OidGetHash((int)hashOID));
```
philljj, line 637: Shouldn't we return error `HASH_TYPE_E` here if digestSz <= 0?

Because that would mean something was requested that wasn't compiled in to `wc_HashGetDigestSize()`.

philljj (changes requested):
Looks good, just a few questions / suggestions.

philljj:
Retest this please.

(PRB passed, but hung fwding results).

wolfcrypt/src/hash.c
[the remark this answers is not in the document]
yosuke-wolfssl: Applied, and retracting my earlier "Won't do" - you were right. Both cases now
use the same condition as `wc_HashGetDigestSize()`, and the `test_wc_OidGetHash()`
OID lists are mirrored to match.

Confirmed against the v5.2.1 and v5.2.4 FIPS bundles: there `wc_HashGetDigestSize()`
returns `HASH_TYPE_E` for these two types, so without the guard the new mapping
would return a hash type the same file refuses to size. (Only matters for FIPS 3-6;
`types.h` already force-defines `WOLFSSL_NOSHA512_224/256` under selftest and FIPS <= 2.)

Left the inverse `wc_HashGetOID()` alone - it is unguarded in those certified
modules too and references nothing FIPS-conditional, so guarding it would diverge
from the shipped module for no gain.

wolfcrypt/src/tsp.c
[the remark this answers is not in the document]
yosuke-wolfssl: No OOB - 36 bytes fits `WC_TSP_MAX_HASH_SZ` (64) and nothing is copied on the
rejected path. But it is not only a returned error, which changes the answer:

```
SetHashType(MD5_SHA) = 0     hashAlgOID = MD5h
SetHash(36) = -132 (BUFFER_E), SetHash(16) = 0
Encode = 0, 39-byte request with a plain MD5 imprint
```

So a caller asking for MD5_SHA silently gets MD5. It is the TLS 1.0/1.1
concatenation with no OID of its own, so it can never be named in a
messageImprint. `wc_TspRequest_SetHashType()` now rejects any type whose OID does
not map back to itself. I checked every hash type through
`wc_HashGetOID()` -> `wc_OidGetHash()` - MD5_SHA is the only one that fails, so
this costs exactly that type.

Did not reject MD5 outright: it has its own OID, is self-consistent, and
`WC_TSP_MIN_HASH_STRENGTH_BITS` already covers that policy.

wolfcrypt/src/tsp.c
yosuke-wolfssl: Agreed, changed - it now returns `HASH_TYPE_E` when `digestSz <= 0`, matching
`wc_TspRequest_SetHash()` on the requester side. The test that pinned the old
behaviour now expects `HASH_TYPE_E`, and the SHA-256 assertions in that test
needed an `#ifndef NO_SHA256` guard since `--enable-tsp` does not pull in SHA-256.

Left the encode-time check in `asn_tsp.c` lenient - it also runs on a `TspTstInfo`
decoded off the wire, so failing closed there would break re-encoding a token whose
algorithm this build lacks.

yosuke-wolfssl:
Hello @philljj ,
Thank you for reviewing. I reworked and replied to your comments inline.
Could you please review this again ?

philljj (approved):

PR #11158, opened 2026-08-12
PSA AES: abort the cipher operation on wc_AesSetIV() so a new IV takes effect [F-8254]

danielinux:
# Description

0281978c44 PSA AES: abort the cipher operation on wc_AesSetIV() so a new IV takes effect

+ test adaptations

Fixes F-8254


JacobBarthelmeh (approved):
Tested using some VM's, fix looks good.

PR #11160, opened 2026-08-12
Add wc_ecc_key_new_ex

holtrop-wolfssl:
# Description

Add wc_ecc_key_new_ex

Rust wrapper: use wc_ecc_key_new_ex for ECC when available.

Fixes F-8300.

# Testing

How did you test?

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/doc/dox_comments/header_files/ecc.h
+++ b/doc/dox_comments/header_files/ecc.h
@@ -682,7 +682,8 @@ int wc_ecc_init(ecc_key* key);
 
     \param key pointer to the ecc_key object to initialize
     \param heap pointer to a heap identifier
-    \param devId ID to use with crypto callbacks or async hardware. Set to INVALID_DEVID (-2) if not used
+    \param devId ID to use with crypto callbacks or async hardware. Set to
+    INVALID_DEVID if not used
 
     _Example_
     \code
@@ -703,8 +704,11 @@ int wc_ecc_init_ex(ecc_key* key, void* heap, int devId);
     \brief This function uses a user defined heap and allocates space for the
     key structure.
 
-    \return 0 Returned upon successfully initializing the ecc_key object
-    \return MEMORY_E Returned if there is an error allocating memory
+    \param heap pointer to a heap identifier
+
+    \return Non-null returned upon successfully allocating and initializing the
+    ecc_key object
+    \return NULL returned if there is an error allocating memory
 
 
     _Example_
@@ -712,12 +716,41 @@ int wc_ecc_init_ex(ecc_key* key, void* heap, int devId);
     wc_ecc_key_new(&heap);
     \endcode
 
+    \sa wc_ecc_key_new_ex
     \sa wc_ecc_make_key
     \sa wc_ecc_key_free
     \sa wc_ecc_init
 */
 
 ecc_key* wc_ecc_key_new(void* heap);
+
+/*!
+    \ingroup ECC
+
+    \brief This function uses a user defined heap and allocates space for the
+    key structure.
+
+    \param heap pointer to a heap identifier
+    \param devId ID to use with crypto callbacks or async hardware. Set to
+    INVALID_DEVID if not used
+
+    \return Non-null returned upon successfully allocating and initializing the
+    ecc_key object
+    \return NULL returned if there is an error allocating memory
+
+
+    _Example_
+    \code
+    wc_ecc_key_new_ex(&heap, MY_DEVID);
+    \endcode
+
+    \sa wc_ecc_key_new
+    \sa wc_ecc_make_key
+    \sa wc_ecc_key_free
+    \sa wc_ecc_init
+*/
+
+ecc_key* wc_ecc_key_new_ex(void* heap, int devId);
 
 /*!
     \ingroup ECC
@@ -2297,28 +2330,6 @@ int wc_ecc_set_curve(ecc_key *key, int keysize, int curve_id);
     \sa wc_ecc_init
 */
 mp_int* wc_ecc_key_get_priv(ecc_key* key);
-
-/*!
-    \ingroup ECC
-    \brief Allocates and initializes new ECC key.
-
-    \return ecc_key pointer on success
-    \return NULL on failure
-
-    \param heap Heap hint for memory allocation
-
-    _Example_
-    \code
-    ecc_key* key = wc_ecc_key_new(NULL);
-    if (key != NULL) {
-        // use key
-        wc_ecc_key_free(key);
-    }
-    \endcode
-
-    \sa wc_ecc_key_free
-*/
-ecc_key* wc_ecc_key_new(void* heap);
 
 /*!
     \ingroup ECC
--- a/wolfcrypt/src/hpke.c
+++ b/wolfcrypt/src/hpke.c
@@ -1116,18 +1116,10 @@ static int wc_HpkeCopyPrivateKey(Hpke* hpke, void* key, void** copy)
             ret = wc_ecc_export_private_only((ecc_key*)key, eccPriv,
                 &eccPrivSz);
             if (ret == 0) {
-                *copy = wc_ecc_key_new(hpke->heap);
+                *copy = wc_ecc_key_new_ex(hpke->heap, INVALID_DEVID);
                 if (*copy == NULL)
                     ret = MEMORY_E;
             }
-        #if defined(PLUTON_CRYPTO_ECC) || defined(WOLF_CRYPTO_CB)
-            /* wc_ecc_key_new() adopts a platform default device id on some
-             * ports. The key being copied has none, so the copy must not gain
-             * one, or the shared secret would move onto a device the caller
-             * never asked for. */
-            if (ret == 0)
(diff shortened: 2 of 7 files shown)


holtrop-wolfssl:
retest this please (Calling Pipeline was cancelled, build removed)

holtrop-wolfssl:
retest this please (git ssh permission problem)

wolfcrypt/src/hpke.c
```diff
@@ -1116,18 +1116,10 @@ static int wc_HpkeCopyPrivateKey(Hpke* hpke, void* key, void** copy)
             ret = wc_ecc_export_private_only((ecc_key*)key, eccPriv,
                 &eccPrivSz);
             if (ret == 0) {
-                *copy = wc_ecc_key_new(hpke->heap);
+                *copy = wc_ecc_key_new_ex(hpke->heap, INVALID_DEVID);
```
philljj, line 1119: I'm confused by this and the copilot finding.

Previously ` wc_ecc_key_new(hpke->heap);` would force a devId with:

```
#if defined(WOLFSSL_QNX_CAAM) || defined(WOLFSSL_IMXRT1170_CAAM)
    /* assume all keys are using CAAM for ECC unless explicitly set otherwise */
    devId = WOLFSSL_CAAM_DEVID;
#endif
```

and then we would set devId back to invalid with a different define gate:
```
        #if defined(PLUTON_CRYPTO_ECC) || defined(WOLF_CRYPTO_CB)
            /* wc_ecc_key_new() adopts a platform default device id on some
             * ports. The key being copied has none, so the copy must not gain
             * one, or the shared secret would move onto a device the caller
             * never asked for. */
            if (ret == 0)
                ((ecc_key*)*copy)->devId = INVALID_DEVID;
        #endif
```

Now we are forcing invalid with `wc_ecc_key_new_ex()`, but elsewhere we are still calling `wc_exx_key_new()`

```
$ grep -r "wc_ecc_key_new" wolfcrypt/src/hpke.c 
            *keypair = wc_ecc_key_new(hpke->heap);
            *keypair = wc_ecc_key_new(hpke->heap);
            *keypair = wc_ecc_key_new(hpke->heap);
            *key = wc_ecc_key_new(hpke->heap);
                *copy = wc_ecc_key_new_ex(hpke->heap, INVALID_DEVID);
```

Will this break expected behavior for CAAM builds?

wolfssl/wolfcrypt/ecc.h
```diff
@@ -675,6 +675,7 @@ WOLFSSL_API mp_int* wc_ecc_key_get_priv(ecc_key* key);
 
 
 WOLFSSL_ABI WOLFSSL_API ecc_key* wc_ecc_key_new(void* heap);
+WOLFSSL_ABI WOLFSSL_API ecc_key* wc_ecc_key_new_ex(void* heap, int devId);
```
philljj, line 678: I know it's a trivial wrapper, but from a code coverage point of view we probably should add a `wc_ecc_key_new_ex()` test to `static wc_test_ret_t ecc_test_allocator(WC_RNG* rng)` in wolfcrypt test.c

philljj (dismissed):

wolfcrypt/src/hpke.c
holtrop-wolfssl: When I implemented this issue I did not originally have any changes for this file. skoll review pointed out that my new _ex version could be used with INVALID_DEVID here to avoid the partial workaround below (that I now removed) that "unset" the devId that wc_ecc_key_new was automatically setting for those build options, since there should not be a devId present for HPKE. It seemed like skoll's finding was correct given the comments below that were trying to force back the devId to INVALID_DEVID after initializing the ecc_key instance. Since I added a path to allocate and initialize the ecc_key without devId being automatically set to an undesired value, it made sense to use that path here rather than unsetting devId afterwards.

wolfssl/wolfcrypt/ecc.h
holtrop-wolfssl: Added `ecc_test_allocator_ex()`

wolfcrypt/test/test.c
```diff
@@ -44727,6 +44478,43 @@ static wc_test_ret_t ecc_test_allocator(WC_RNG* rng)
 }
 #endif
 
+#if !defined(HAVE_FIPS) && !defined(HAVE_SELFTEST) && \
+    !defined(WOLFSSL_NO_MALLOC) && (!defined(WOLF_CRYPTO_CB_ONLY_ECC) || defined(WOLFSSL_SWDEV)) && \
+    (!defined(NO_ECC_SECP) || defined(WOLFSSL_CUSTOM_CURVES))
+/* Test for the wc_ecc_key_new_ex() and wc_ecc_key_free() functions. */
+static wc_test_ret_t ecc_test_allocator_ex(WC_RNG* rng)
```
padelsbach, line 44481: Can we merge this new function with the existing `ecc_test_allocator()`?

doc/dox_comments/header_files/ecc.h
```diff
@@ -703,22 +703,51 @@ int wc_ecc_init_ex(ecc_key* key, void* heap, int devId);
...
 ecc_key* wc_ecc_key_new(void* heap);
 
+/*!
+    \ingroup ECC
+
+    \brief This function uses a user defined heap and allocates space for the
+    key structure.
+
+    \param heap pointer to a heap identifier
+    \param devId ID to use with crypto callbacks or async hardware. Set to INVALID_DEVID (-2) if not used
```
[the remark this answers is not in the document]
padelsbach: I agree with Copilot here. This and the existing function seem to be the only ones that specify the number.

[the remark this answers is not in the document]
padelsbach: Another nit: can you wrap the lines to >=80

doc/dox_comments/header_files/ecc.h
```diff
@@ -703,22 +703,51 @@ int wc_ecc_init_ex(ecc_key* key, void* heap, int devId);
...
     wc_ecc_key_new(&heap);
     \endcode
 
+    \sa wc_ecc_key_new_ex
     \sa wc_ecc_make_key
     \sa wc_ecc_key_free
     \sa wc_ecc_init
 */
 
 ecc_key* wc_ecc_key_new(void* heap);
```
padelsbach, line 725: Since we are in there: can you delete the duplicate doc for `wc_ecc_key_new` around line 2350?

wolfcrypt/test/test.c
```diff
@@ -44727,6 +44478,43 @@ static wc_test_ret_t ecc_test_allocator(WC_RNG* rng)
...
+{
+    wc_test_ret_t ret = 0;
+    ecc_key* key;
+#ifdef WC_NO_RNG
+    word32 idx = 0;
+#endif
+
+    key = wc_ecc_key_new_ex(HEAP_HINT, INVALID_DEVID);
```
padelsbach: Using `INVALID_DEVID` here just exercises the old path. Can we use `devId` here (already defined above)? And then, we'd want to confirm it matches `key->devId` after the call.

padelsbach (changes requested):

wolfcrypt/test/test.c
holtrop-wolfssl: @philljj specifically requested testing both APIs here.

padelsbach: Sure, `ecc_test_allocator_ex` is almost identical to `ecc_test_allocator`. Should be possible to test both with a single test function and without the copypasta

doc/dox_comments/header_files/ecc.h
holtrop-wolfssl: Removed

wolfcrypt/test/test.c
holtrop-wolfssl: Used devId defined above as suggested.

wolfcrypt/test/test.c
holtrop-wolfssl: I started to merge these but it got uglier because of the way the exit label and ERROR_OUT was used. So I kept them separate but did remove the identical `#if` / `#endif` between the functions to reduce a little redundancy.

holtrop-wolfssl:
I also removed the (-2) from the doxygen header, wrapped lines to <= 80 characters, and removed the second copy of `wc_ecc_key_new` documentation and prototype.

holtrop-wolfssl:
retest this please (Failed in branch fips_ready_part3only)

wolfcrypt/test/test.c
padelsbach: Can you add the check on `devID` after the call? Something like:

```
#if defined(WOLF_CRYPTO_CB) || defined(PLUTON_CRYPTO_ECC)
    if (key->devId != devId) {
        ERROR_OUT(WC_TEST_RET_ENC_NC, exit);
    }
#endif
```

holtrop-wolfssl: Oh yes, I forgot to add that check last round. Added!

padelsbach (approved):
LGTM, thanks

PR #11172, opened 2026-08-13
Thread safety and private key handling

dgarske:
# Thread safety and private key handling

Part 3 of 4 from an internal automated source review. Independent of parts 1 and 2. **Part 4 builds on this branch** and should merge after it.

| Report | Summary |
|---|---|
| F-1514 | ECC fixed point cache mutex initialization |
| F-1515 | ECC OID cache mutex initialization |
| F-1946 | netRandom mutex initialization |
| F-1380 | Crypto callback device table entry publication order |
| F-1930 | ECC private key checks on import |
| F-1931 | ECC raw private key checks on import |
| F-2636 | SE050 port private key buffer release |
| F-2972 | RSA blinding factor inversion |

## Notes for reviewers

**One-time mutex initialization.** Three lazy check-then-set sites could double-initialize a mutex. They now share `wc_local_InitMutexOnce()` in `wc_port.c`, using the three-state compare-exchange already used by the SP ECC cache locks (0 uninitialized, 1 in progress, 2 ready). Losers of the race wait for the winner rather than returning a retryable error, so no caller needs a retry path. Only compiled when `WOLFSSL_MUTEX_INITIALIZER` is undefined, i.e. non-pthread targets.

**ECC private key import.** The private scalar is now range checked against `[1, n-1]` on every import rather than only under `WOLFSSL_VALIDATE_ECC_IMPORT`. The check is two comparisons - no scalar multiply, no curve specification load - so the default-build cost is negligible. The expensive public key consistency check stays behind the existing macro. It reads the scalar through `ecc_get_k()` so it is correct on either side of key blinding, and is skipped for NXP CAAM black keys, whose `k` holds an encrypted blob rather than a scalar.

**RSA blinding.** The blinding factor was inverted with `mp_invmod()`, a binary extended Euclidean variant whose iteration count tracks its input. It is now inverted under a fresh multiplicative mask which is divided back out. Measured cost: no measurable change on the default and fastmath backends; -11% RSA-2048 private on the deprecated heapmath backend, which makes no timing-resistance claim.

**Crypto callback registry.** Entries are filled before `devId` is stored and cleared after it is retired. Store ordering only, no barrier - serializing register/unregister against in-flight operations remains the caller's responsibility, and the comment now says so.

Also fixed: the `HAVE_WNR` build did not compile (`wc_GenerateSeed()` referenced an identifier that exists nowhere in the tree), and `wc_InitNetRandom()` returned uninitialized stack data on success.

## Testing

`make check` passes on `--enable-all`, `--enable-all --enable-lms`, `--enable-all -DWOLFSSL_TEST_NO_MUTEX_INITIALIZER` (the configuration that actually compiles the one-time mutex code), `--enable-all -DWOLFSSL_CHECK_MEM_ZERO`, `--enable-cryptocb --enable-all`, `--enable-fastmath`, `--enable-heapmath` and `--enable-smallstack`.

New test: ECC private key range rejection without `WOLFSSL_VALIDATE_ECC_IMPORT` (`tests/api/test_ecc.c`).

The NXP SE050 port has no host build and was reviewed by hand.


wolfcrypt/src/cryptocb.c
```diff
@@ -515,9 +521,14 @@ int wc_CryptoCb_RegisterDevice(int devId, CryptoDevCallbackFunc cb, void* ctx)
         else {
             /* Error in callback register cmd. Don't register */
             wc_CryptoCb_ClearDev(dev);
+            return rc;
         }
     }
 #endif
+
+    /* Publish the entry last. */
```
douzzer: For ordered reads and writes, we want a `WC_BARRIER` to keep the optimizer from reordering things.

douzzer (changes requested):

dgarske:
Jenkins retest this please

douzzer (dismissed):
tested with
```
LIBWOLFSSL_CONFIGURE_ARGS_OVERRIDE=CFLAGS=-DWOLFSSL_TEST_NO_MUTEX_INITIALIZER BASELINE_OID=upstream/master ../testing/git-hooks/wolfssl-multi-test.sh ... --oid dgarske/fenrir_threadsafety_keys --rebase 20260807-linuxkm-ed-pqc-glue
check-source-text
clang-tidy-all-sp-all
quantum-safe-wolfssl-all-intelasm-sp-asm-helgrind
quantum-safe-wolfssl-all-intelasm-sp-asm-sanitizer
quantum-safe-wolfssl-all-noasm-clang-msan
```

(Note that `helgrind` with `WOLFSSL_TEST_NO_MUTEX_INITIALIZER` found numerous races around `test_wolfSSL_client_server_nofail_ex()`, identically found in `master`, so there's definitely more work to be done to fully race-harden !`WOLFSSL_MUTEX_INITIALIZER` targets.)

philljj:
Retest this please.

(no logs)

philljj:
Tests good with multi-test.

philljj (approved):

PR #11173, opened 2026-08-13
Constant-time handling of the ECC point at infinity

dgarske:
# Constant-time handling of the ECC point at infinity

Part 4 of 4 from an internal automated source review, and the last of the batch. Parts 1-3 (#11170, #11171, #11172) have merged; this branch is now rebased onto master and is a single commit touching one file. It was kept separate because it is the only change in the batch with a measurable performance cost.

| Report | Summary |
|---|---|
| F-2618 | ECC point add and double exceptional case handling |

`ecc_projective_add_point_safe()` and `ecc_projective_dbl_point_safe()` branched on whether an operand was the point at infinity, collapsing to a point copy instead of doing the arithmetic. Both now run the ordinary operation regardless and select the result with `mp_cond_copy()`, so an exceptional operand costs the same as an ordinary one.

The addition computes into a temporary rather than into `R`: every call site passes the first operand as the destination, so writing `R` early would destroy `A` before the selection. The temporary is on the stack unless the build asked otherwise, and inherits the key heap and the small-stack cache back-pointer from `R` so it behaves like the caller's own points.

The `A == B` equality branch is deliberately left in place. The Montgomery ladder cannot reach it, and covering it would mean computing a doubling as well as an addition on every call.

## Performance

Measured interleaved A/B with standalone binaries, four runs each, on a default `./configure` build:

| Operation | master | branch | Delta |
|---|---|---|---|
| ECDH agree | 5747.6 | 5228.0 | **-9.0%** |
| ECDSA sign | 5484.8 | 5043.8 | **-8.0%** |
| ECDSA verify | 8181.6 | 7746.1 | **-5.3%** |

This is the legacy point arithmetic only. Builds using `--enable-sp` - including `--enable-all` - route P-256/384/521 through the SP implementations and are unaffected.

Worth noting for the record: this code is **not** compiled out under SP math as the report claims. The guard is `!WOLFSSL_SP_MATH`, not `!WOLFSSL_SP_MATH_ALL`, and instrumentation confirms P-256 ECDH and ECDSA reach both functions in a default build.

A further caveat: `mp_cmp_ct()` and `mp_montgomery_reduce_ct()` are `#define`d to the non-constant-time versions on `integer.c` and `tfm.c`, so this change cannot deliver constant time on those backends regardless.

## Testing

`make check` passes on `--enable-all`, `--enable-all --enable-lms`, `--enable-fastmath`, `--enable-heapmath`, `--enable-smallstack --enable-smallstackcache` and `--enable-fpecc`. Re-verified after the rebase onto master: build, `testwolfcrypt` and the full unit suite all clean.

The source-text lint (`wolfssl-multi-test.sh check-source-text`) passes on the commit.

### DIFF
--- a/wolfcrypt/src/ecc.c
+++ b/wolfcrypt/src/ecc.c
@@ -8700,6 +8700,10 @@ static int ecc_cond_copy_point(ecc_point* S, int copy, ecc_point* D)
  *   A = B             -> Doubled by the add itself when the z values match;
  *                        all-zero result when they differ - doubled here.
  *   A + B = infinity  -> Add result has z == 0, x/y not 0; set to infinity.
+ *
+ * The infinity out-flag is set only for the last two cases - a finite pair
+ * whose sum is infinity - matching the original branching version, which
+ * never touched the flag when copying an operand out.
  */
 int ecc_projective_add_point_safe(ecc_point* A, ecc_point* B, ecc_point* R,
     mp_int* a, mp_int* modulus, mp_digit mp, int* infinity)
@@ -8708,6 +8712,7 @@ int ecc_projective_add_point_safe(ecc_point* A, ecc_point* B, ecc_point* R,
     int aInf;
     int bInf;
     int nInf;
+    int rInf = 0;
 #ifdef WOLFSSL_SMALL_STACK
     ecc_point* T = NULL;
 #else
@@ -8770,16 +8775,14 @@ int ecc_projective_add_point_safe(ecc_point* A, ecc_point* B, ecc_point* R,
         /* When only Z zero then result is infinity */
         else {
             err = ecc_set_point_infinity(T);
-            if ((err == MP_OKAY) && (infinity != NULL))
-                *infinity = 1;
+            rInf = 1;
         }
     }
 
     /* A = -B with matching z: replace the double the add produced. */
     if ((err == MP_OKAY) && nInf) {
         err = ecc_set_point_infinity(T);
-        if ((err == MP_OKAY) && (infinity != NULL))
-            *infinity = 1;
+        rInf = 1;
     }
 
     /* Select without branching on which operand was infinity: A infinity
@@ -8791,6 +8794,16 @@ int ecc_projective_add_point_safe(ecc_point* A, ecc_point* B, ecc_point* R,
         err = ecc_cond_copy_point(A, bInf, T);
     if (err == MP_OKAY)
         err = wc_ecc_copy_point(T, R);
+
+    /* An infinity operand feeds the raw formula a shape it cannot represent,
+     * so the add's infinity note may be spurious. The selection above then
+     * replaces the result with the other operand, and that selection
+     * overrides the note as well: the flag reports only a result the add
+     * itself produced. */
+    rInf &= (aInf == 0);
+    rInf &= (bInf == 0);
+    if ((err == MP_OKAY) && (infinity != NULL) && (rInf != 0))
+        *infinity = 1;
 
 #if defined(WOLFSSL_SMALL_STACK_CACHE) && !defined(WOLFSSL_ECC_NO_SMALL_STACK)
     T->key = NULL;
--- a/wolfcrypt/src/ecc.c
+++ b/wolfcrypt/src/ecc.c
@@ -8667,88 +8667,137 @@ int wc_ecc_free(ecc_key* key)
     !defined(WOLFSSL_CRYPTOCELL) && !defined(WOLFSSL_SP_MATH) && \
     (!defined(WOLF_CRYPTO_CB_ONLY_ECC) || defined(WOLFSSL_QNX_CAAM) || \
       defined(WOLFSSL_IMXRT1170_CAAM))
-/* Handles add failure cases:
+/* Set a point to the representation of infinity, (0, 0, 1). */
+static int ecc_set_point_infinity(ecc_point* P)
+{
+    int err = mp_set(P->x, 0);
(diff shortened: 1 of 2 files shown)


dgarske:
Jenkins retest this please

dgarske:
Jenkins retest this please

wolfcrypt/src/ecc.c
```diff
@@ -8662,63 +8684,101 @@ int ecc_projective_add_point_safe(ecc_point* A, ecc_point* B, ecc_point* R,
...
+     * infinity costs the same as the ordinary case. The result goes to a
+     * temporary: callers pass the first operand as the destination, so
+     * writing R early would destroy A before the selection below. */
+    aInf = (mp_iszero(A->x) == MP_YES) & (mp_iszero(A->y) == MP_YES);
+    bInf = (mp_iszero(B->x) == MP_YES) & (mp_iszero(B->y) == MP_YES);
+
+    /* Equal ordinates still branch to the doubling path - unreachable from
+     * the Montgomery ladder, and covering it would cost an extra double on
+     * every call. */
+    eq = (!aInf) && (!bInf) && (mp_cmp(A->x, B->x) == MP_EQ) &&
```
SparkiDev: use '&' instead of '&&'?

wolfcrypt/src/ecc.c
```diff
@@ -8662,63 +8684,101 @@ int ecc_projective_add_point_safe(ecc_point* A, ecc_point* B, ecc_point* R,
...
             err = _ecc_projective_dbl_point(B, R, a, modulus, mp);
         }
         else {
             /* A = -B */
-            err = mp_set(R->x, 0);
-            if (err == MP_OKAY)
-                err = mp_set(R->y, 0);
-            if (err == MP_OKAY)
-                err = mp_set(R->z, 1);
+            err = ecc_set_point_infinity(R);
```
SparkiDev: Are we trying to make this constant time?

wolfcrypt/src/ecc.c
```diff
@@ -8662,63 +8684,101 @@ int ecc_projective_add_point_safe(ecc_point* A, ecc_point* B, ecc_point* R,
...
             if ((err == MP_OKAY) && (infinity != NULL))
                 *infinity = 1;
         }
+        return err;
```
SparkiDev: Are we making this whole function constant time?
If so, we should be doing the add.

SparkiDev (changes requested):

philljj:
Retest this please.

no logs

philljj:
Retest this please.

ConnectionResetError: [Errno 104] Connection reset by peer

dgarske:
Jenkins retest this please. m1 FIPS failure

philljj:
Retest this please.

(fips tests hung and aborted)

wolfcrypt/src/ecc.c
```diff
@@ -8667,89 +8667,138 @@ int wc_ecc_free(ecc_key* key)
...
-                if ((err == MP_OKAY) && (infinity != NULL))
-                    *infinity = 1;
+                err = _ecc_projective_dbl_point(B, T, a, modulus, mp);
             }
         }
+        /* When only Z zero then result is infinity */
+        else {
+            err = ecc_set_point_infinity(T);
+            if ((err == MP_OKAY) && (infinity != NULL))
+                *infinity = 1;
```
philljj: my skoll run is angry about this. Argues it could copy finite result A over T while leaving infinity set.


### HIGH-1: Infinity flag can remain set after selecting a finite operand

**Description:** The unconditional add can set `*infinity = 1` before the subsequent conditional copies replace its result. For example, with finite `A = (0, y, 1)` where `y != 0` and infinity `B = (0, 0, 1)`, the raw formula produces `T->z == 0` with a nonzero coordinate, so lines 8771-8774 set the flag. The `bInf` selection then copies finite `A` into `T`, but the flag remains set. Before this PR, the `B`-infinity branch copied `A` without modifying the flag. Thus the function can return a finite point while reporting infinity, potentially corrupting callers that use this output as state, such as the fixed-point paths that pass `&first`.

**Code:**
```
if ((err == MP_OKAY) && mp_iszero(T->z)) {
    ...
    else {
        err = ecc_set_point_infinity(T);
        if ((err == MP_OKAY) && (infinity != NULL))
            *infinity = 1;
    }
}
...
err = ecc_cond_copy_point(A, bInf, T);
```

**Recommendation:** Fix the flag-selection ordering and add regression cases for finite `x == 0` points plus infinity in both operand orders.

philljj (dismissed):
Skoll flagged a High that looks valid, and a medium testing suggestion.

Looks good otherwise so far.

### MEDIUM-2: Rewritten exceptional-case selection lacks behavioral assertions

**Description:** Existing white-box cases exercise these functions but generally check only the return code. In particular, the existing finite `(x=0, y!=0)` plus infinity vector initializes `infinity` to zero but never verifies that it remains zero or that `R` equals the finite operand, allowing the blocking regression above to pass unnoticed.

**Code:**
```
err = ecc_cond_copy_point(B, aInf, T);
if (err == MP_OKAY)
    err = ecc_cond_copy_point(A, bInf, T);
if (err == MP_OKAY)
    err = wc_ecc_copy_point(T, R);
```

**Recommendation:** Extend the existing unit white-box tests with result and flag assertions rather than return-code-only coverage.

philljj:
Retest this please.

(PRB hung then aborted)

SparkiDev (approved):

PR #11175, opened 2026-08-14
Fix AES-GCM nonce reuse in the async encrypt path

embhorn:
# Description

`wc_AesGcmEncrypt_ex()` reports the nonce it consumed via `ivOut` and advances an internal counter so no two records share a nonce under one key. The counter only advanced when `wc_AesGcmEncrypt()` returned 0; but an async backend reports a successful submission with `WC_PENDING_E`.

Fixes zd22291

# Testing

`test_wc_AesGcmEncrypt_ex_NonceUnique`
`test_wc_AesCcmEncrypt_ex_NonceUnique`
`test_tls12_aesgcm_record_nonce_unique`

# Checklist

 - [x] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/wolfcrypt/src/aes.c
+++ b/wolfcrypt/src/aes.c
@@ -14917,12 +14917,23 @@ int wc_AesGcmEncrypt_ex(Aes* aes, byte* out, const byte* in, word32 sz,
     }
 
     if (ret == 0) {
+        /* Hand the encrypt the nonce out of ivOut rather than aes->reg. Some
+         * backends use aes->reg as scratch space, and an asynchronous backend
+         * keeps the pointer until the operation completes, so aes->reg is not
+         * a stable place to hold the nonce being consumed. */
         XMEMCPY(ivOut, aes->reg, ivOutSz);
         ret = wc_AesGcmEncrypt(aes, out, in, sz,
-                               (byte*)aes->reg, ivOutSz,
+                               ivOut, ivOutSz,
                                authTag, authTagSz,
                                authIn, authInSz);
-        if (ret == 0)
+        /* Put the nonce back unconditionally - a backend may have left scratch
+         * data in aes->reg - so a failed encrypt leaves the counter on the
+         * nonce it did not consume rather than on backend leftovers. */
+        XMEMCPY(aes->reg, ivOut, ivOutSz);
+        /* A nonce handed to an asynchronous backend has been consumed even
+         * though the operation has not finished yet - the counter must still
+         * advance or the next record would reuse this nonce. */
+        if (ret == 0 || ret == WC_NO_ERR_TRACE(WC_PENDING_E))
             IncCtr((byte*)aes->reg, ivOutSz);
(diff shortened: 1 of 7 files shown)


wolfcrypt/src/aes.c
```diff
@@ -14956,12 +14917,23 @@ int wc_AesGcmEncrypt_ex(Aes* aes, byte* out, const byte* in, word32 sz,
...
         ret = wc_AesGcmEncrypt(aes, out, in, sz,
-                               (byte*)aes->reg, ivOutSz,
+                               ivOut, ivOutSz,
                                authTag, authTagSz,
                                authIn, authInSz);
-        if (ret == 0)
+        /* Put the nonce back unconditionally - a backend may have left scratch
+         * data in aes->reg - so a failed encrypt leaves the counter on the
+         * nonce it did not consume rather than on backend leftovers. */
+        XMEMCPY(aes->reg, ivOut, ivOutSz);
```
philljj, line 14932: I'm uncomfortable with this double copy, and potential side effects with code that access aes->reg (like `linuxkm/lkcapi_aes_glue.c`).

Would like @douzzer input here.

Also see copilot comment below.

philljj (changes requested):

wolfcrypt/src/aes.c
embhorn: @douzzer the first draft for this fix modifies code in the FIPS boundary. 

```
diff --git a/wolfssl/wolfcrypt/aes.h b/wolfssl/wolfcrypt/aes.h
index 1eccc39ca..58aa5f0bf 100644
--- a/wolfssl/wolfcrypt/aes.h
+++ b/wolfssl/wolfcrypt/aes.h
@@ -482,6 +482,12 @@ struct Aes {
     cy_stc_crypto_aes_gcm_state_t aes_gcm_state;
 #endif
 #endif /* WOLFSSL_PSOC6_CRYPTO */
+#if defined(HAVE_AESGCM) || defined(HAVE_AESCCM)
+    /* Nonce counter behind wc_AesGcmSetIV(), wc_AesGcmSetExtIV() and
+     * wc_AesCcmSetNonce(). Kept apart from reg because backends use reg as
+     * scratch, and an asynchronous backend writes it after the call returns. */
+    byte nonceCtr[WC_AES_BLOCK_SIZE];
+#endif
 };
 
 #ifndef WC_AES_TYPE_DEFINED
```

douzzer (approved):
LGTM

tested with
```
wolfssl-multi-test.sh ...
quantum-safe-wolfssl-all-crypto-only-intelasm-sp-asm-sp-linuxkm-mainline-clang-tidy
quantum-safe-wolfssl-all-crypto-only-intelasm-sp-asm-sp-fips-dev-linuxkm-mainline-insmod
```

philljj (approved):

JacobBarthelmeh (approved):

PR #11178, opened 2026-08-14
Harden TLS 1.3 NewSessionTicket handling per RFC 9846

### LINKED ISSUE #11126: wolfSSL sends TLS 1.3 NewSessionTicket without client PSK mode advertisement
# wolfSSL sends TLS 1.3 NewSessionTicket without client PSK mode advertisement

## Summary

This is a real but configuration-dependent protocol compliance and interoperability issue. RFC 9846 does not state an unconditional "server MUST NOT send NewSessionTicket" rule. The permission to send a TLS 1.3 `NewSessionTicket` is tied to the client having advertised suitable `psk_key_exchange_modes`, and servers are also advised not to send tickets that are incompatible with the advertised modes.

wolfSSL parses and stores `psk_key_exchange_modes`, but the automatic TLS 1.3 session-ticket send path and the explicit `wolfSSL_send_SessionTicket()` API do not check whether the client advertised suitable PSK key exchange modes before sending a ticket. A fresh runtime rerun confirmed that a non-PSK TLS 1.3 client can legally omit extension type `45`, while an unmodified ticket-enabled wolfSSL server still completes the handshake and sends `NewSessionTicket`.

## Standard Requirement

Official standard links:

- [RFC 9846 Section 4.3.9, Pre-Shared Key Exchange Modes](https://datatracker.ietf.org/doc/html/rfc9846#section-4.3.9)
- [RFC 9846 Section 4.7.1, New Session Ticket Message](https://datatracker.ietf.org/doc/html/rfc9846#section-4.7.1)
- [RFC 8446 Section 4.6.1, New Session Ticket Message](https://datatracker.ietf.org/doc/html/rfc8446#section-4.6.1)

RFC 9846 Section 4.3.9 makes `psk_key_exchange_modes` the client's statement of which PSK modes it supports. That statement applies both to PSKs offered in the current `ClientHello` and to PSKs that could be provided later by `NewSessionTicket`.

The same section says servers should not send tickets that are incompatible with the advertised modes. RFC 9846 Section 4.7.1 also makes the server's permission to send `NewSessionTicket` conditional on the client hello containing a suitable `psk_key_exchange_modes` extension.

The precise requirement is therefore:

1. A client that does not offer `pre_shared_key` may omit `psk_key_exchange_modes`; that omission is not itself a handshake error.
2. A `NewSessionTicket` creates a future resumption PSK, so the server should respect the client's advertised PSK mode support before sending one.
3. If the `ClientHello` contains no `psk_key_exchange_modes` extension, the server has no advertised mode set that makes the ticket suitable.

Compared with RFC 8446 Section 4.6.1, RFC 9846 adds PSK-mode conditions around ticket sending. This issue should be treated as a low-severity compliance and interoperability problem rather than a direct confidentiality or integrity break.

### `src/tls.c:12900-12944`

```c
static int TLSX_PskKeModes_Parse(WOLFSSL* ssl, const byte* input, word16 length,
                                 byte msgType)
{
    int    ret;
    byte modes;

    ret = TLSX_PskKeyModes_Parse_Modes(input, length, msgType, &modes);
    if (ret == 0)
        ret = TLSX_PskKeyModes_Use(ssl, modes);
    /* ... */
}

int TLSX_PskKeyModes_Use(WOLFSSL* ssl, byte modes)
{
    int ret = 0;
    TLSX* extension = TLSX_Find(ssl->extensions,
                                TLSX_PSK_KEY_EXCHANGE_MODES);
    if (extension == NULL) {
        ret = TLSX_Push(&ssl->extensions, TLSX_PSK_KEY_EXCHANGE_MODES,
                        NULL, ssl->heap);
        /* ... */
    }

    extension->val = modes;
    return 0;
}
```

When the client sends `psk_key_exchange_modes`, wolfSSL stores an extension object and the parsed mode bits in `ssl->extensions`. If the client does not send the extension, that receive-state object is not created, so the server has enough state to distinguish "extension present" from "extension absent".

### `src/tls13.c:6942-6948`

```c
/* Get the PSK key exchange modes the client wants to negotiate. */
ext = TLSX_Find(ssl->extensions, TLSX_PSK_KEY_EXCHANGE_MODES);
if (ext == NULL) {
    WOLFSSL_ERROR_VERBOSE(MISSING_HANDSHAKE_DATA);
    return MISSING_HANDSHAKE_DATA;
}
modes = ext->val;
```

The handshake branch that actually uses PSK requires this extension. That protects the current PSK handshake path, but it does not protect the later certificate-handshake path that sends `NewSessionTicket`.

### `src/internal.c:2883-2894`

```c
#if defined(HAVE_SESSION_TICKET) && !defined(NO_WOLFSSL_SERVER)
    ctx->ticketEncCb = DefTicketEncCb;
    ctx->ticketEncCtx = (void*)&ctx->ticketKeyCtx;
    ctx->ticketHint = SESSION_TICKET_HINT_DEFAULT;
#if defined(WOLFSSL_TLS13)
    ctx->maxTicketTls13 = 1; /* default to sending a session ticket if compiled
                                in */
#endif
```

In ticket-enabled TLS 1.3 server builds, the default context installs a ticket encryption callback and defaults to sending one TLS 1.3 session ticket.

### `src/tls13.c:13037-13217`

```c
static int SendTls13NewSessionTicket(WOLFSSL* ssl)
{
    if (DefTicketHintTooLarge(ssl)) {
        return 0;
    }
    /* ... */
    if (!ssl->options.noTicketTls13) {
        if ((ret = SetupTicket(ssl)) != 0)
            return ret;
        if ((ssl->options.mask & WOLFSSL_OP_NO_TICKET) == 0) {
            if ((ret = CreateTicket(ssl)) != 0)
                return ret;
        }
    }
    /* ... build and send NewSessionTicket ... */
    ret = SendBuffered(ssl);
}
```

The shared ticket construction and send function does not centrally check the client's PSK key exchange modes. All callers inherit that missing precondition.

### `src/tls13.c:16085-16100`

```c
#ifdef HAVE_SESSION_TICKET
while (ssl->options.ticketsSent < ssl->options.maxTicketTls13) {
    if (!ssl->options.noTicketTls13 && ssl->ctx->ticketEncCb != NULL) {
        if ((ssl->error = SendTls13NewSessionTicket(ssl)) != 0) {
            WOLFSSL_ERROR(ssl->error);
            return WOLFSSL_FATAL_ERROR;
        }
    }
    ssl->options.ticketsSent++;
    /* ... */
}
#endif
```

The automatic send loop checks only the ticket switch, encryption callback, and count. It does not check whether `TLSX_PSK_KEY_EXCHANGE_MODES` exists or whether it contains a compatible mode.

### `src/tls13.c:16151-16166`

```c
int wolfSSL_send_SessionTicket(WOLFSSL* ssl)
{
    if (ssl == NULL || !IsAtLeastTLSv1_3(ssl->version))
        return BAD_FUNC_ARG;
    if (ssl->options.side == WOLFSSL_CLIENT_END)
        return SIDE_ERROR;
    if (ssl->options.handShakeState != HANDSHAKE_DONE)
        return NOT_READY_ERROR;

    if ((ssl->error = SendTls13NewSessionTicket(ssl)) != 0) {
        WOLFSSL_ERROR(ssl->error);
        return WOLFSSL_FATAL_ERROR;
    }
    ssl->options.ticketsSent++;
    return WOLFSSL_SUCCESS;
}
```

The explicit API checks version, endpoint role, and handshake state, but it also lacks the `psk_key_exchange_modes` precondition.

## Implementation Behavior

In a ticket-enabled TLS 1.3 server build:

1. A non-PSK certificate handshake can legally omit both `pre_shared_key` and `psk_key_exchange_modes`.
2. wolfSSL accepts that `ClientHello`; only the branch that negotiates PSK requires `TLSX_PSK_KEY_EXCHANGE_MODES`.
3. After the handshake completes and the server receives client Finished, the automatic ticket loop calls `SendTls13NewSessionTicket()` based on `maxTicketTls13` and `ticketEncCb`.
4. `SendTls13NewSessionTicket()` creates and sends a `NewSessionTicket` without checking whether the client advertised a suitable PSK mode.
5. `wolfSSL_send_SessionTicket()` reaches the same internal send function and has the same behavior.

`NO_PSK` does not remove this path. The rerun build defined `NO_PSK`, `HAVE_SESSION_TICKET`, and `WOLFSSL_TLS13`, and the server still sent a TLS 1.3 session ticket that creates a future resumption PSK.

## Inconsistency Reason

The standard model is: the client first advertises acceptable PSK exchange modes, and the server then uses those modes to decide whether a future resumption PSK is suitable. wolfSSL implements: if tickets are enabled, the handshake state is suitable, and the ticket callback is available, send `NewSessionTicket`.

Successful completion of a non-PSK certificate handshake does not prove that the client advertised acceptable future PSK modes. wolfSSL's default client usually sends extension `45`, which hides this missing server-side pre-send check in common interop runs. The targeted runtime rerun shows that when a client legally omits the extension, the unmodified server still sends `NewSessionTicket`.

## Runtime Evidence

Fresh rerun date: 2026-08-10.

Action: a ticket-enabled unmodified wolfSSL TLS 1.3 server was run with a transparent proxy that parsed the cleartext `ClientHello` extension list. The reproducer client performed a non-PSK initial handshake and was built so it did not advertise extension type `45` (`psk_key_exchange_modes`).

Observed result for the reproducer:

```json
{
  "client_exit_code": 0,
  "server_exit_code": 0,
  "client_hello_extension_ids": [43, 13, 10, 22, 51, 35],
  "client_hello_contains_psk_key_exchange_modes": false,
  "new_session_ticket_callback_observed": true,
  "issue_observed": true,
  "probe_passed": true,
  "ticket_marker": "Session Ticket CB: ticketSz = 210, ctx = initial session"
}
```

Interpretation: the `ClientHello` did not contain extension type `45`; the TLS 1.3 handshake still succeeded; and the unmodified ticket-enabled wolfSSL server sent `NewSessionTicket`.

Action: the same proxy was run as a positive control with the default wolfSSL client, which should advertise extension type `45`.

Observed result for the control:

```json
{
  "client_exit_code": 0,
  "server_exit_code": 0,
  "client_hello_extension_ids": [45, 43, 13, 10, 22, 51, 35],
  "client_hello_contains_psk_key_exchange_modes": true,
  "new_session_ticket_callback_observed": true,
  "issue_observed": false,
  "probe_passed": true,
  "ticket_marker": "Session Ticket CB: ticketSz = 210, ctx = initial session"
}
```

Interpretation: when extension `45` was present, the proxy recognized it. This rules out the explanation that the reproducer merely failed to detect the extension. Both client and server exited with code `0`, so the observed ticket came from a successful TLS 1.3 handshake, not from a failed path or an anomalous log.

## Impact

- wolfSSL deviates from RFC 9846's ticket-sending permission and ticket-mode compatibility guidance.
- A server can provide a resumption PSK to a client that did not advertise any acceptable PSK exchange mode, causing useless ticket state, extra network and storage cost, and possible later resumption failures or inconsistent behavior.
- The rerun did not show a direct confidentiality or integrity break; the main risk is protocol compliance and interoperability.
- Builds without `HAVE_SESSION_TICKET` have no reachable send path and are out of scope.

## Fix Direction

Put the check in shared pre-send logic used by all `NewSessionTicket` callers:

1. After `ClientHello` parsing, retain whether `psk_key_exchange_modes` was received and which mode bits were recognized.
2. Send automatic `NewSessionTicket` only when at least one advertised mode is compatible with the server's ticket and resumption policy.
3. Make `wolfSSL_send_SessionTicket()` use the same check; when the extension is absent or incompatible, return a clear documented error.
4. Add regression tests for absent extension, only unknown modes, only `psk_ke`, only `psk_dhe_ke`, both modes, and the explicit API path.

### LINKED ISSUE #11127: RFC9846 wolfSSL issue: NewSessionTicket extension TLV framing is skipped without early-data support
# RFC9846 wolfSSL issue: NewSessionTicket extension TLV framing is skipped without early-data support

## Summary
- Not applicable: builds without session-ticket/resumption support, because RFC9846 permits clients that do not support resumption to silently ignore the whole `NewSessionTicket` message.

wolfSSL parses and stores TLS 1.3 `NewSessionTicket` messages when session tickets are compiled in. In the configuration where session tickets are enabled but `WOLFSSL_EARLY_DATA` is disabled, `DoTls13NewSessionTicket` validates only the outer `extensions` vector length and skips the vector contents. Therefore malformed `Extension` TLV framing inside `NewSessionTicket.extensions` is accepted instead of causing a `decode_error`.

## Standard Requirement

Official standard links:

- [RFC 9846 Section 4.3, Extensions](https://datatracker.ietf.org/doc/html/rfc9846#section-4.3)
- [RFC 9846 Section 4.7.1, New Session Ticket Message](https://datatracker.ietf.org/doc/html/rfc9846#section-4.7.1)
- [RFC 9846 Section 6, Alert Protocol](https://datatracker.ietf.org/doc/html/rfc9846#section-6)

```text
A number of TLS messages contain tag-length-value encoded extensions
structures.

struct {
    ExtensionType extension_type;
    opaque extension_data<0..2^16-1>;
} Extension;
```

```text
There MUST NOT be more than one extension of the same type in a given
extension block.
```

```text
Clients which receive a NewSessionTicket message but do not support
resumption MUST silently ignore this message.
```

```text
struct {
    uint32 ticket_lifetime;
    uint32 ticket_age_add;
    opaque ticket_nonce<0..255>;
    opaque ticket<1..2^16-1>;
    Extension extensions<0..2^16-1>;
} NewSessionTicket;
```

```text
extensions: A list of extension values for the ticket. The "Extension"
format is defined in Section 4.3. Clients MUST ignore unrecognized
extensions.
```

```text
Peers which receive a message which cannot be parsed according to the
syntax (e.g., have a length extending beyond the message boundary or
contain an out-of-range length) MUST terminate the connection with a
"decode_error" alert.
```

Interpretation: a client that does not support resumption may ignore the entire `NewSessionTicket`. A client that does support resumption and processes the ticket must still parse `extensions` as a list of Section 4.3 `Extension` TLVs. The instruction to ignore unrecognized extensions applies to well-formed unknown extension types; it does not permit accepting bytes that cannot be parsed as `Extension` structures at all.

### `src/tls13.c:12880-12892`

```c
    if ((*inOutIdx - begin) + EXTS_SZ > size)
        return BUFFER_ERROR;
    ato16(input + *inOutIdx, &length);
    *inOutIdx += EXTS_SZ;
    if ((*inOutIdx - begin) + length != size)
        return BUFFER_ERROR;
    #ifdef WOLFSSL_EARLY_DATA
    ret = TLSX_Parse(ssl, (byte *)input + (*inOutIdx), length, session_ticket,
                     NULL);
    if (ret != 0)
        return ret;
    #endif
    *inOutIdx += length;
```

This path checks that the two-byte `extensions` vector length reaches the end of the `NewSessionTicket` body. It calls the generic extension parser only when `WOLFSSL_EARLY_DATA` is compiled in.

### `src/tls13.c:12900-12907`

```c
#else
    (void)ssl;
    (void)input;

    WOLFSSL_ENTER("DoTls13NewSessionTicket");

    *inOutIdx += size;
#endif /* HAVE_SESSION_TICKET */
```

When `HAVE_SESSION_TICKET` is not compiled in, wolfSSL ignores the whole message. That path matches the RFC9846 exception for clients that do not support resumption.

### `src/tls.c:18421-18476`

```c
        if (length - offset < HELLO_EXT_TYPE_SZ + OPAQUE16_LEN)
            return BUFFER_ERROR;

        ato16(input + offset, &type);
        offset += HELLO_EXT_TYPE_SZ;

        ato16(input + offset, &size);
        offset += OPAQUE16_LEN;

        if (length - offset < size)
            return BUFFER_ERROR;
```

`TLSX_Parse` performs the generic `Extension` TLV framing checks. It rejects a short extension header and rejects an extension whose declared `extension_data` length runs past the extension block.

### `src/tls.c:18444-18450` and `src/internal.c:37153-37167`

```c
            if (IS_OFF(seenType, TLSX_ToSemaphore(type))) {
                TURN_ON(seenType, TLSX_ToSemaphore(type));
            }
            else {
                return DUPLICATE_TLS_EXT_E;
            }
```

```c
    int TranslateErrorToAlert(int err)
    {
        switch (err) {
            case WC_NO_ERR_TRACE(BUFFER_ERROR):
                return decode_error;
            ...
            case WC_NO_ERR_TRACE(DUPLICATE_TLS_EXT_E):
                return illegal_parameter;
```

The same generic parser also catches duplicate recognized extensions. Syntax framing failures map to `decode_error`; duplicate recognized extensions map to `illegal_parameter`.

## Implementation Behavior

In `HAVE_SESSION_TICKET && !WOLFSSL_EARLY_DATA` builds, `DoTls13NewSessionTicket` parses the fixed ticket fields, stores ticket state, validates only the outer `extensions` vector length, skips all extension bytes, calls `SetupSession`, and returns success. No internal `ExtensionType` or `extension_data` length scan is performed.

In `HAVE_SESSION_TICKET && WOLFSSL_EARLY_DATA` builds, the same function calls `TLSX_Parse` for `session_ticket`. That parser rejects malformed TLV framing with `BUFFER_ERROR` and duplicate recognized extensions with `DUPLICATE_TLS_EXT_E`.

In builds without `HAVE_SESSION_TICKET`, the whole `NewSessionTicket` message is ignored. That is not an issue under the standard's no-resumption exception.

## Inconsistency Reason

RFC9846 separates three cases:

- A client that does not support resumption may silently ignore the entire `NewSessionTicket`.
- A client that supports resumption may ignore a well-formed but unrecognized extension.
- A peer that receives a message whose fields cannot be parsed according to the TLS syntax must terminate with `decode_error`.

wolfSSL's non-early-data ticket build falls into the second implementation category but behaves as if malformed extension bytes were merely unknown extensions. Because it stores and processes the ticket while skipping the `Extension` TLV framing, malformed `NewSessionTicket.extensions` payloads with a matching outer vector length are accepted.

## Runtime Evidence

Runtime was rerun on 2026-08-10 from the workspace root.

Action: the non-early-data and early-data loopback targets were rebuilt. Both rebuilds completed successfully. The loopback test established a real TLS 1.3 client/server connection, had the server send an encrypted post-handshake `NewSessionTicket` built through wolfSSL's own record builder, and then sent one application-data marker byte. Reading the marker means the client accepted or skipped the crafted ticket; rejection before the marker means the ticket parser failed the crafted input.

Observed result for a valid unknown extension in the ticket-enabled non-early-data build:

```text
mode=valid_unknown_extension
compile_have_session_ticket=defined
compile_wolfssl_early_data=not_defined
client_note=client read application marker after crafted NST
observed=accepted_post_handshake_message
issue_observed=not_applicable_control
```

Observed result for malformed ticket extensions in the ticket-enabled non-early-data build:

```text
mode=malformed_short_vector
compile_have_session_ticket=defined
compile_wolfssl_early_data=not_defined
client_note=client read application marker after crafted NST
client_alert_tx_code=-1
observed=accepted_post_handshake_message
issue_observed=yes

mode=malformed_inner_length_truncation
compile_have_session_ticket=defined
compile_wolfssl_early_data=not_defined
client_note=client read application marker after crafted NST
client_alert_tx_code=-1
observed=accepted_post_handshake_message
issue_observed=yes
```

Observed result for the same malformed ticket extensions with early-data enabled:

```text
mode=malformed_short_vector
compile_have_session_ticket=defined
compile_wolfssl_early_data=defined
client_note=client rejected before marker, read_ret=-1 ssl_error=-328
client_alert_tx_code=50
client_alert_tx_level=2
observed=rejected_post_handshake_message
issue_observed=no

mode=malformed_inner_length_truncation
compile_have_session_ticket=defined
compile_wolfssl_early_data=defined
client_note=client rejected before marker, read_ret=-1 ssl_error=-328
client_alert_tx_code=50
client_alert_tx_level=2
observed=rejected_post_handshake_message
issue_observed=no
```

Interpretation: the loopback rerun confirms that the non-early-data build accepts malformed `NewSessionTicket.extensions` and continues to read application data, while the early-data build rejects the same malformed records with alert code `50` (`decode_error`).

Action: a focused dispatcher probe directly called wolfSSL's TLS 1.3 `DoTls13HandShakeMsgType(..., session_ticket, ...)` dispatcher after placing the client object in a legitimate post-handshake receive state. This narrower parser-path check verifies the exact `DoTls13NewSessionTicket` branch.

Observed result in the ticket-enabled non-early-data build:

```text
build_WOLFSSL_EARLY_DATA=0
case=empty_vector accepted=1 ret=0 consumed=14 alert_level=-1 alert_code=-1
case=well_formed_unknown accepted=1 ret=0 consumed=18 alert_level=-1 alert_code=-1
case=well_formed_early_data accepted=1 ret=0 consumed=22 alert_level=-1 alert_code=-1
case=short_extension_header accepted=1 ret=0 consumed=15 alert_level=-1 alert_code=-1
case=truncated_extension_value accepted=1 ret=0 consumed=18 alert_level=-1 alert_code=-1
case=duplicate_early_data accepted=1 ret=0 consumed=30 alert_level=-1 alert_code=-1
probe_result=passed
```

Observed result in the ticket-enabled early-data build:

```text
build_WOLFSSL_EARLY_DATA=1
case=empty_vector accepted=1 ret=0 consumed=14 alert_level=-1 alert_code=-1
case=well_formed_unknown accepted=1 ret=0 consumed=18 alert_level=-1 alert_code=-1
case=well_formed_early_data accepted=1 ret=0 consumed=22 alert_level=-1 alert_code=-1
case=short_extension_header accepted=0 ret=-328 consumed=14 alert_level=2 alert_code=50
case=truncated_extension_value accepted=0 ret=-328 consumed=14 alert_level=2 alert_code=50
case=duplicate_early_data accepted=0 ret=-457 consumed=14 alert_level=2 alert_code=47
probe_result=passed
```

Interpretation: the dispatcher probe confirms the same configuration split. The non-early-data build accepts malformed and duplicate `NewSessionTicket` extensions. The early-data build rejects malformed TLV framing with `decode_error` and duplicate `early_data` with `illegal_parameter`.

## Impact

A nonconformant or malicious authenticated TLS 1.3 server can send a malformed encrypted post-handshake `NewSessionTicket.extensions` vector to a wolfSSL client built with session tickets but without early data. The client accepts the message and continues the connection instead of terminating with the required `decode_error`. The main impact is standards compliance and interoperability; the sender must already be the TLS peer capable of producing valid encrypted post-handshake traffic.

## Fix Direction

Validate generic `Extension` TLV framing for `NewSessionTicket.extensions` whenever `HAVE_SESSION_TICKET` is enabled, regardless of `WOLFSSL_EARLY_DATA`. A direct fix is to call `TLSX_Parse` for `session_ticket` unconditionally and keep extension-specific handling conditional where necessary. An alternative is a small structural scanner that enforces `ExtensionType` plus `extension_data<0..2^16-1>` boundaries and duplicate checks while still ignoring well-formed unrecognized extension types.

### LINKED ISSUE #11132: TLS session ticket partition key prefix aliasing in wolfSSL client cache
# TLS session ticket partition key prefix aliasing in wolfSSL client cache

## Summary

This is a real issue. RFC 9846 Appendix C.4 says client applications should not offer tickets across connections that are meant to be uncorrelated. wolfSSL provides `wolfSSL_SetServerID()` so applications can separate the client session and ticket cache with a custom key. However, the client-cache lookup compares only the requested key length and does not require the cached session's original key length to be equal.

As a result, a shorter partition key can match the prefix of a longer key in the same cache row and reuse a TLS 1.3 ticket saved under the longer key. A fresh TLS 1.3 reproducer run confirmed this: after the first connection used long key `origin|Ak`, the second connection using shorter key `origin|A` resumed, and both client and server reported `reused=1`.

## Standard Requirement

Official standard links:

- [RFC 9846 Appendix C.4, Client and Server Tracking Prevention](https://datatracker.ietf.org/doc/html/rfc9846#appendix-C.4)
- [RFC 9846 Section 1.1, Conventions and Terminology](https://datatracker.ietf.org/doc/html/rfc9846#section-1.1)

Relevant meaning:

- RFC 9846 Section 1.1 binds uppercase `SHOULD NOT` to BCP 14 interpretation.
- Appendix C.4 explains that offering a ticket lets a server correlate different connections, independent of ticket reuse count.
- Therefore, client applications should not offer tickets across connections that are meant to be uncorrelated. The RFC gives web browser network partition keys as an example.

Key excerpt:

```text
Client applications SHOULD NOT offer tickets across
connections that are meant to be uncorrelated.
```

This issue is not about the SNI or certificate resumption conditions in RFC 9846 Section 4.7.1. Section 4.7.1 controls when resumption is acceptable for server identity and certificate constraints. This report concerns the privacy partitioning rule in Appendix C.4: the application supplied different partition keys, but the library mixed them and caused a ticket to be offered and accepted across partitions.

## Relevant Source Code

Target source tree: `implementions/wolfssl-master`

`src/ssl_sess.c:331-374` shows `wolfSSL_SetServerID()` receiving the application-provided `serverID` key. When `newSession == 0`, it first calls `wolfSSL_GetSessionClient()` to look for an existing client session. If none is found, it saves the current key and length in the session:

```c
if (newSession == 0) {
    session = wolfSSL_GetSessionClient(ssl, id, len);
    if (session) {
        if (wolfSSL_SetSession(ssl, session) != WOLFSSL_SUCCESS) {
            WOLFSSL_MSG("wolfSSL_SetSession failed");
            session = NULL;
        }
    }
}

if (session == NULL) {
    ssl->session->idLen = (word16)len;
    XMEMCPY(ssl->session->serverID, id, (size_t)len);
}
```

`src/ssl_sess.c:1653-1727` places client-cache entries into a row computed from `serverID` and `idLen`:

```c
if (idLen) {
    clientRow = HashObject(serverID,
            idLen, &error) % CLIENT_SESSION_ROWS;
}
```

The defect is in `src/ssl_sess.c:998-1062`. The lookup computes the row using the requested key length, but the match predicate compares only `len` bytes and does not check whether `current->idLen == len`:

```c
if (current && XMEMCMP(current->serverID, id,
                                             (unsigned long)len) == 0) {
    WOLFSSL_MSG("Found a serverid match for client");
    if (LowResTimer() < (current->bornOn + current->timeout)) {
        WOLFSSL_MSG("Session valid");
        ret = current;
```

`wolfssl/internal.h:4905-4907` shows the session already stores enough information for strict matching:

```c
word16             idLen;             /* serverID length          */
byte               serverID[SERVER_ID_LEN]; /* for easier client lookup */
```

The missing check is therefore a length equality condition in the lookup predicate, not a missing field.

## Trigger Conditions

The tested build enables the affected path:

- `HAVE_SESSION_TICKET` is defined.
- `NO_MD5` is defined.
- `NO_SHA` is not defined.
- `WOLFSSL_TLS13` is defined.

With the default cache configuration, `CLIENT_SESSION_ROWS` is `88`. The reproducer uses two different application partition keys:

| Application partition key | Length | SHA1 first four bytes | Row modulo 88 | Behavior |
|---|---:|---|---:|---|
| `origin|Ak` | 9 | `aa099dcc` | 44 | first connection, full handshake |
| `origin|A` | 8 | `243b6a94` | 44 | second connection, incorrect resumption |

This is not a SHA1 collision. The two keys hash differently, but modulo 88 places both in the same client-cache row. At the same time, `origin|A` is a prefix of `origin|Ak`. Because lookup compares only the requested length, the short key accepts the cached session created for the long key.

## Runtime Evidence

Fresh rerun date: 2026-08-10.

Action: a TLS 1.3 reproducer was run from the workspace root using the ticket-enabled audit build. The reproducer first connected with partition key `origin|Ak`, then connected again with the distinct shorter prefix key `origin|A`.

Observed output:

```text
client partition=origin|Ak new_session=0 reused=0 marker=0
client partition=origin|A new_session=0 reused=1 marker=1
server connection=1 reused=0
server connection=2 reused=1
RESULT=ISSUE distinct_prefix_partition_key_reused=1
process_exit=0
```

Interpretation:

- The first connection used `origin|Ak` and did not resume, as expected for a first connection.
- The second connection used the different key `origin|A`, but resumed.
- Both client and server reported `reused=1` on the second connection, proving the old ticket was offered and accepted.
- Process exit code `0` means the reproducer observed the target defect condition.

Action: the control mode was run on the same build. It tested same-key resumption, different same-length partition keys, and an explicit fresh-session request.

Observed control output:

```text
client partition=origin.example|partition-A new_session=0 reused=0 marker=0
client partition=origin.example|partition-A new_session=0 reused=1 marker=1
client partition=origin.example|partition-B new_session=0 reused=0 marker=2
client partition=origin.example|partition-A new_session=1 reused=0 marker=3
server connection=1 reused=0
server connection=2 reused=1
server connection=3 reused=0
server connection=4 reused=0
RESULT=PASS same_partition_resumed=1 cross_partition_resumed=0 forced_fresh_resumed=0
process_exit=0
```

Interpretation:

- The same partition key can resume, so session tickets work normally.
- A different same-length partition key does not resume, so the test is not "all different keys resume".
- `newSession=1` forces a fresh session, so the application's explicit fresh-session path works.

## Impact

This is a conditional privacy partitioning and cache correctness issue. Applications that use `wolfSSL_SetServerID()` with variable-length partition keys can reuse tickets across partitions when all of these conditions hold:

- one partition key is a prefix of another partition key;
- both keys land in the same client-cache row after `HashObject()`;
- the build enables session tickets and client cache;
- the caller allows lookup of an existing session, meaning `newSession == 0`.

Fixed-length keys, explicit length-prefix encoding, or `wolfSSL_SetServerID(..., newSession=1)` avoid this specific trigger, but those are application-side workarounds rather than a library fix.

## Fix Direction

Add a length equality check to `wolfSSL_GetSessionClient()`:

```c
current->idLen == len &&
XMEMCMP(current->serverID, id, (unsigned long)len) == 0
```

Recommended regression coverage:

- long key cached first, short prefix key looked up second, both in the same client-cache row;
- short key cached first, long key looked up second;
- same row but non-prefix different keys;
- equal-length but different-content keys;
- `newSession=1` fresh-session path.

julek-wolfssl:
- Only send a NewSessionTicket when the client advertised `psk_key_exchange_modes`, per RFC 9846 §4.3.9/§4.7.1. The advertised modes are now recorded in `Options` since the extension object is freed with the rest of the handshake state; the automatic path skips the ticket, and the explicit API returns `MISSING_HANDSHAKE_DATA` or `PSK_KEY_ERROR`. Old behaviour is available via `WOLFSSL_TLS13_TICKET_NO_PSK_MODES`. Fixes #11126.
- Made this PSK-mode check opt-in behind `WOLFSSL_TLS13_TICKET_CHECK_PSK_MODES` (including the `Options` fields that record the advertised modes), since withholding a ticket from clients that omit `psk_key_exchange_modes` but still expect one is a behaviour change.
- Validate `NewSessionTicket.extensions` framing in `DoTls13NewSessionTicket()` even when `WOLFSSL_EARLY_DATA` is not defined, so malformed framing is rejected with a decode_error alert regardless of build config. Fixes #11127.
- Fixed `wolfSSL_GetSessionClient()` to compare the full cached server ID length instead of only the requested number of bytes, preventing a short server ID from matching an unrelated cached entry in the same cache row and crossing `wolfSSL_SetServerID()` partitions, per RFC 9846 Appendix C.4. Fixes #11132.
- Added tests asserting that unrecognized `NewSessionTicket` extensions (empty vector, one unknown type, two unknown types) are ignored per RFC 9846 §4.7.1, with the connection staying up and application data flowing both ways with no alert sent, both with and without the framing check.


julek-wolfssl:
retest this please

philljj (dismissed):
Needs a rebase + force-push to clear merge-conflicts and unblock PRB.

Also review Fenrir item please.

julek-wolfssl:
rebased + addressed

julek-wolfssl:
retest this please

julek-wolfssl:
retest this please

julek-wolfssl:
retest this please

julek-wolfssl:
retest this please

julek-wolfssl:
retest this please

julek-wolfssl:
retest this please

julek-wolfssl:
retest this please

JacobBarthelmeh (approved):

PR #11186, opened 2026-08-17
 Fix heap overflow in TLS compression receive path

embhorn:
# Description

`DoApplicationData()` decompressed a record and copied the result back over the record it arrived in with `XMEMMOVE`. The input buffer is only grown to hold the compressed wire record, so a ~40 byte record expanding to 16384 bytes wrote ~16KB past the allocation, and would also have clobbered any records queued behind it. Requires `--with-libz` plus `wolfSSL_set_compression()` on both peers; TLS 1.2 only.

The plaintext now goes into a connection owned buffer that is never the caller's, which also covers the sniffer path (it passes its own packet buffer, so a bounds check against `inputBuffer` would have been wrong there). The buffer carries one byte of headroom so a full output buffer means the record ran past the RFC 5246 6.2.2 limit; leftover input cannot be used to detect this, as `inflate()` pulls trailing bytes into its bit accumulator with output still pending. Over-limit records get a `decompression_failure` alert.

Fixes found while making the feature reachable enough to test:

- ClientHello offered zlib without the null method RFC 5246 7.4.1.2  requires, so a compressed handshake could not complete at all.
- `SendData()` advanced `sent` by the compressed length, so a write spanning more than one record resent most of the payload.
- Compression stayed on when TLS 1.3 was negotiated, which removed it.
- The receive path capped `curSize` at 2^14, rejecting a legally expanded compressed record; `GetRecordHeader()` already allows `MAX_COMP_EXTRA`.

Fixes zd22288

# Testing

Adds a `compress` test group (round trip, multi-record, incompressible, ClientHello wire bytes, TLS 1.3 off, no-writeback, over-limit) and an `all-libz` CI config, since no job enabled zlib and none of these would otherwise compile.

# Checklist

 - [x] added tests
 - [ ] updated/added doxygen


src/internal.c
```diff
@@ -809,9 +809,57 @@ int IsDtlsNotSrtpMode(WOLFSSL* ssl)
...
         return (int)ssl->d_stream.total_out - currTotal;
     }
 
+
+    /* Size the decompression buffer one byte over the largest plaintext the
+     * peer may send, so myDeCompress() can tell a record at the limit apart
+     * from one past it.  Grows if a renegotiation raised the fragment size.
+     * Not DYNAMIC_TYPE_IN_BUFFER: a static memory build with a fixed IO pool
+     * serves that type from the same buffer as ssl->buffers.inputBuffer. */
+    static int EnsureDecompBuffer(WOLFSSL* ssl)
```
philljj: My skoll flagged this, but it seems like a false positive.

### HIGH-1: Grow-only buffer bypasses a reduced max-fragment limit

**Description:** The new decompression limit is derived from `decompBuffer.length`, but that field is an allocation capacity which only grows. If the maximum fragment size is later reduced—through renegotiation, object reuse, or `WOLFSSL_ALLOW_MAX_FRAGMENT_ADJUST`—`EnsureDecompBuffer()` retains the larger allocation and `DoApplicationData()` passes that stale capacity to `myDeCompress()`. A highly compressible fragment can therefore expand past the current negotiated/configured maximum without filling the buffer and is delivered to the application. This is introduced by conflating the new buffer's capacity with the current logical decompression limit.

**Code:**
```
if (ssl->buffers.decompBuffer.length >= needed)
    return 0;
...
dataSz = myDeCompress(ssl, rawData, rawSz,
    ssl->buffers.decompBuffer.buffer,
    (int)ssl->buffers.decompBuffer.length);
```

**Recommendation:** Separate logical output size from allocation capacity and add a regression test that allocates at a large fragment size, lowers the limit, and verifies that a compressed record expanding beyond the new limit is rejected.

src/internal.c
philljj, line 29048: My skoll review also flagged this:

### HIGH-2: Compression expansion is not reserved in DTLS records
- **File:** `src/internal.c:29035-29060`

**Description:** The PR reserves `MAX_COMP_EXTRA` only in the non-DTLS branch. The DTLS branch still sizes the output and checks the MTU using the uncompressed plaintext length. For an incompressible payload whose ordinary encrypted record fits the MTU, `myCompress()` can increase `buffSz`, after which `BuildMessage()` requires more space than `outputSz` and returns an error. The compliant ClientHello added by this PR also applies to DTLS 1.2, making compression negotiable between wolfSSL peers while this newly reachable send path cannot handle legal compression expansion.

**Code:**
```
if (ssl->options.dtls) {
    outputSz = wolfssl_local_GetRecordSize(ssl, (word32)buffSz, 1);
    ...
}
else {
    outputSz = wolfssl_local_GetRecordSize(ssl, buffSz + compExtra, 1);
}
```

**Recommendation:** Correct DTLS sizing and add a DTLS 1.2 test using an incompressible payload near the MTU, including the `WOLFSSL_NO_DTLS_SIZE_CHECK` splitting configuration.

src/internal.c
```diff
@@ -28988,7 +29052,12 @@ int SendData(WOLFSSL* ssl, const void* data, size_t sz)
             int maxFrag = wolfSSL_GetMaxFragSize(ssl);
             if (maxFrag > 0)
                 buffSz = min((word32)buffSz, (word32)maxFrag);
-            outputSz = wolfssl_local_GetRecordSize(ssl, (word32)buffSz, 1);
+            /* No MTU to respect here, so the record is simply allocated big
```
philljj, line 29068: ### HIGH-3: wolfSSL_GetOutputSize under-reports records now emitted by SendData

**Description:** This change allows `SendData()` to emit a compressed record larger than the plaintext by sizing for `buffSz + MAX_COMP_EXTRA`, but the public `wolfSSL_GetOutputSize(ssl, inSz)` path still calculates a record from `inSz` alone. Consequently, an application asking how many transport bytes a write requires can receive a value smaller than the record that is subsequently emitted. Previously such an expanded compressed write failed the output bound; the PR makes it succeed without updating the associated size-query contract.

**Code:**
```
outputSz = wolfssl_local_GetRecordSize(ssl, buffSz + compExtra, 1);
```

**Recommendation:** Update the size-query API alongside `SendData()` and extend the incompressible test to assert that the actual emitted record never exceeds the reported output size.

philljj (dismissed):
Review skoll and fenrir items.

If the DTLS change is out of scope then that could be done in another PR.

embhorn:
retest this please

embhorn:
retest this please

JacobBarthelmeh (approved):

src/internal.c
```diff
@@ -34147,12 +34255,18 @@ static void MakePSKPreMasterSecret(Arrays* arrays, byte use_psk_key)
...
 
-        /* last, compression */
-        output[idx++] = COMP_LEN;
-        if (ssl->options.usingCompression)
+        /* last, compression.  RFC 5246 7.4.1.2 requires the list to always
+         * include CompressionMethod.null, so zlib is offered alongside it
+         * rather than on its own - a list of just zlib is rejected. */
+        if (ssl->options.usingCompression) {
+            output[idx++] = COMP_LEN + ENUM_LEN;
             output[idx++] = ZLIB_COMPRESSION;
```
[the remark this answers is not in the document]
JacobBarthelmeh: out of scope of this PR

PR #11195, opened 2026-08-18
asn.c: check the OCSP responder certificate's signature

holtrop-wolfssl:
# Description

asn.c: check the OCSP responder certificate's signature

The VERIFY_OCSP_CERT -> VERIFY/VERIFY_OCSP mapping lived only in the non-template ASN.1 path, so in the default WOLFSSL_ASN_TEMPLATE build the mode reached the ConfirmSignature() gate unmatched and the responder certificate embedded in an OCSP response was accepted without its signature ever being checked against the issuing CA.

Hoist the mapping to where both implementations converge, after decoding and before the signer lookup, so one copy serves both. This also covers responder certificates with no extensions, which skipped the mapping in the non-template path as well. Add VERIFY_OCSP_CERT to the gate as a safety net for re-entry, where the parse block is skipped.

Adds test_ocsp_forged_responder_cert_rejected, which fails without this change.

Fixes ZD#22329

# Testing

How did you test?

# Checklist

 - [X] added tests
 - [ ] updated/added doxygen


holtrop-wolfssl:
retest this please (Timeout has been exceeded)

philljj (approved):

PR #11197, opened 2026-08-18
Add a Linux user space backend for the CAAM port

dgarske:
## Summary

Adds a Linux user-space backend to the wolfSSL CAAM driver, enabling NXP SEC hardware acceleration from standard user processes. Developed and tested on a QorIQ T1040 (PowerPC).

## Key Features & Components

* **New Drivers:** Added `caam_linux.[ch]` (11-function port seam) and `wolfcaam_linux.[ch]` (in-process dispatch replacing QNX IPC).
* **Build System:** Integrated `--enable-caam=linux` into `configure.ac` and `include.am` files.
* **Documentation:** Added setup guidelines, platform constraints, and benchmark notes in `wolfcrypt/src/port/caam/README.md`.
* **Offload Scope:** Accelerates AES-CBC, AES-CTR, AES-ECB, and TRNG. All other algorithms default to software.

## Technical & Architecture Notes

* **Unbinding Native Driver:** Requires unbinding the kernel `caam` driver first. The port masks ring interrupts and polls for completion to bypass the kernel handler.
* **Memory Management:** Operates outside standard heap via reserved physical memory (`mem=` boot parameter) to bypass 36-bit/4GB physical address limits and `CONFIG_STRICT_DEVMEM` restrictions.
* **Staging & Job Splitting:** Oversized requests are chunked into contiguous jobs. Intermediate IV state is preserved to ensure proper CBC/CTR chaining without returning `CRYPTOCB_UNAVAILABLE`. Mutual exclusion protects shared staging buffers across multi-job requests.

## Fixes to Shared CAAM Core

* **Interrupt Masking:** Fixed `caamGetJobStatus()` unintentionally clearing `JRCFGR_JR` interrupt masks.
* **Output Ring Alignment:** Rounded output ring offsets up to 8-byte boundaries to prevent SEC timeout errors.
* **AES-CTR Error Handling:** Prevented `wc_CAAM_AesCtrEncrypt()` from processing leftover bytes on failed calls.
* **Guards & Target Flags:** Introduced `WOLFSSL_CAAM_NO_SM`, `WOLFSSL_LP_ONLY_CAAM_AES`, and feature exclusion guards to prevent destructive fallback behavior (e.g., in-place AES-CCM truncation). Fixes target compilation warnings across standard `./configure` options.

## Testing & Verification

* **Target Hardware:** Validated on NXP T1040D4RDB (e5500, 32-bit Big Endian, Linux 3.12).
* **Validation:** Verified via hardware offload failure modes and large buffer chunking comparisons against software implementations. Passed `make check` on x86_64 (`--enable-all` and `--enable-caam=linux`).
* **Performance Note:** Hardware AES offload is slower than e5500 software execution due to per-job overhead; primary benefits are TRNG entropy and offloading CPU cycles.

## Scope Exclusions

* No descriptor support for hashing, i.MX secure memory (blobs/black keys), or public key operations in this release. 
* Separate from the `sec_qoriq` branch PR.


wolfcrypt/src/port/caam/wolfcaam_linux.c
```diff
@@ -0,0 +1,374 @@
...
+
+    return ret;
+}
+
+int SynchronousSendRequest(int type, unsigned int args[4], CAAM_BUFFER* buf,
+    int sz)
+{
+    int ret;
+
+    if (wc_CAAMInitInterface() != 0) {
```
aidangarske: **CAAM can be torn down between initialization and request locking**

`wc_CAAMInitInterface()` acquires and releases `caamLock` before this function later acquires it for staging and dispatch. Another thread can call `wc_CAAMFreeInterface()` in that gap, run `CleanupCAAM()`, and unmap the SEC register window. The first thread then dispatches using the driver's stale ring base address, causing an access to unmapped memory. This contradicts the new teardown synchronization guarantee and affects normal concurrent use with `wolfCrypt_Cleanup()`.

**Fix:** Hold one lock continuously across initialization, request staging, dispatch, and copy-back. Refactor initialization into a lock-held helper or add lifecycle reference counting so cleanup cannot unmap the backend after initialization succeeds but before the request starts.

wolfcrypt/src/port/caam/caam_linux.c
```diff
@@ -0,0 +1,328 @@
...
+
+    *base = (CAAM_ADDRESS)caamSecMap + CAAM_LINUX_JR_OFFSET;
+
+    /* Mask this ring's interrupt: the kernel driver leaves its handler
+     * installed after an unbind, and a completion interrupt would then be
+     * serviced against memory it has already freed. */
+    CAAM_WRITE(*base + JRCFGR_JR,
+        CAAM_READ(*base + JRCFGR_JR) | JRCFGR_IMSK);
+
+    if (caamRingVirt == NULL) {
```
aidangarske, line 234: **Job-ring allocation failure is ignored before ring memory is dereferenced**

The new port correctly returns `-1` when the reserved DMA pool cannot provide the ring. However, `SetupJobRing()` does not check the return value of `CAAM_SET_JOBRING_ADDR()` before doing pointer arithmetic and `memset()` on `r->VirtualIn`. A denied `/dev/mem` mapping, unavailable reserved pool, or exhausted pool therefore becomes a null/invalid-pointer dereference during initialization rather than a clean hardware-initialization failure.

**Fix:** Make `SetupJobRing()` check and propagate the result of `CAAM_SET_JOBRING_ADDR()` before accessing any ring pointers, and have `InitCAAM()` unwind any earlier setup on that error path.

wolfcrypt/src/port/caam/wolfcaam_linux.c
```diff
@@ -0,0 +1,374 @@
...
+            return WC_HW_E;
+        }
+        caamLockInit = 1;
+    }
+#endif
+
+    if (wc_LockMutex(&caamLock) != 0) {
+        return WC_HW_E;
+    }
+    if (!caamInitDone) {
```
aidangarske, line 122: **Failed initialization leaves mappings and mutex state behind**

When `InitCAAM()` fails after mapping the SEC window or initializing the driver's job-ring mutex, this function leaves `caamInitDone` clear and returns. `wc_CAAMFreeInterface()` only calls `CleanupCAAM()` when `caamInitDone` is set, so the mapping and `/dev/mem` descriptor are never released. A later retry remaps over the saved mapping and attempts to initialize the already initialized driver mutex again, which is undefined for `pthread_mutex_init()`.

**Fix:** Unwind partial CAAM initialization before returning an error, including the register mapping, file descriptor, and initialized driver mutex. Ensure a subsequent initialization attempt starts from a fully clean state.

configure.ac
```diff
@@ -4785,6 +4785,14 @@ then
             AM_CFLAGS="$AM_CFLAGS -DWOLFSSL_IMX6UL_CAAM"
             ;;
 
+        linux)
+            # Drive a job ring from Linux user space. Needs a reserved
+            # physical DMA pool (boot with mem=) and the in-tree caam driver
+            # unbound; see wolfcrypt/src/port/caam/README.md.
+            AM_CFLAGS="$AM_CFLAGS -DWOLFSSL_CAAM_LINUX -DWOLFSSL_CAAM_NO_SM"
```
aidangarske, line 4807: **Autotools CAAM defines are not propagated to installed consumers**

The new backend defines `WOLFSSL_CAAM`, `WOLFSSL_CAAM_LINUX`, and `WOLFSSL_CAAM_NO_SM` only in library `AM_CFLAGS`. They are absent from the generated Autotools options header and from `wolfssl.pc` consumer Cflags. Consequently, an application built with the normal installed headers and `pkg-config --cflags wolfssl` sees a different `Aes` layout than the library: the library includes the CAAM-only `blackKey` member while the application does not. Calls such as `wc_AesInit()`, which clears `sizeof(Aes)` in the library, can then write past an application-allocated `Aes`.

**Fix:** Expose all CAAM Linux feature defines through the installed consumer configuration, preferably the generated options header, and add an installed-library consumer test that compiles using only the advertised pkg-config or CMake flags.

wolfcrypt/src/port/caam/wolfcaam_linux.c
```diff
@@ -0,0 +1,374 @@
...
+    dma[outIdx].BufferType = DataBuffer | LastBuffer;
+    dma[outIdx].TheAddress = (CAAM_ADDRESS)caamOutBuf;
+
+    /* Anything past the staging buffer goes as successive jobs. The engine
+     * stores the updated IV back to caamIvBuf on each one, so CBC and CTR
+     * chain across the split on their own, and ECB has no state to carry.
+     * Each job stays on a block boundary so the chaining lines up. */
+    total = buf[inIdx].Length;
+    done  = 0;
+    while (done < total) {
```
aidangarske, line 318: **New chunking and failure paths have no automated coverage**

The new Linux-specific AES dispatcher adds multi-job CBC/CTR chaining, shared scratch-buffer handling, and per-job failure behavior, but no test source exercises `WOLFSSL_CAAM_LINUX` or `CAAM_LINUX_*`. Hardware validation is useful, but this logic can be tested with a mocked CAAM submission layer.

**Fix:** Add mocked tests for CBC and CTR requests immediately below, at, and above `CAAM_LINUX_AES_MAX`; verify IV/counter continuity across chunks and correct error propagation when a later submitted job fails.

aidangarske (changes requested):
## Skoll Code Review

**Scan type:** `review`

### Posted findings
- [High] **CAAM can be torn down between initialization and request locking** — `wolfcrypt/src/port/caam/wolfcaam_linux.c:340-360`
- [High] **Job-ring allocation failure is ignored before ring memory is dereferenced** — `wolfcrypt/src/port/caam/caam_linux.c:234-238`
- [High] **Failed initialization leaves mappings and mutex state behind** — `wolfcrypt/src/port/caam/wolfcaam_linux.c:113-120`
- [High] **Autotools CAAM defines are not propagated to installed consumers** — `configure.ac:4792-4793`
- [Medium] **New chunking and failure paths have no automated coverage** — `wolfcrypt/src/port/caam/wolfcaam_linux.c:290-321`

wolfcrypt/src/port/caam/caam_linux.c
```diff
@@ -0,0 +1,328 @@
...
+    off_t off = (off_t)phys;
+
+    /* belt and braces for a platform where the request above did not take */
+    if ((unsigned long long)off != phys) {
+        WOLFSSL_MSG("caam: physical address does not fit an mmap offset");
+        return NULL;
+    }
+
+    if (caamMemFd < 0) {
+        caamMemFd = open(CAAM_LINUX_MEM_DEV, O_RDWR | O_SYNC);
```
aidangarske: **The /dev/mem capability remains open across exec** · _Security_

BEFORE this PR, the CAAM port did not open `/dev/mem` from Linux user space. AFTER, `caamMapPhys` retains a process-global `/dev/mem` descriptor without `O_CLOEXEC`. If a privileged service initializes CAAM and subsequently executes a less-privileged or attacker-controlled program, that program inherits the already-authorized descriptor and can use it to map arbitrary physical memory even after credentials are dropped. Other wolfSSL device-opening paths use close-on-exec handling to prevent this capability leak.

**Fix:** Open the descriptor with `O_CLOEXEC`, or use wolfSSL's close-on-exec open helper, and retain that protection for every fallback implementation.

**Suggestion:**
```suggestion
caamMemFd = open(CAAM_LINUX_MEM_DEV, O_RDWR | O_SYNC | O_CLOEXEC);
```

wolfcrypt/src/port/caam/caam_linux.c
```diff
@@ -0,0 +1,328 @@
...
+static unsigned char* caamPoolAlloc(int sz, CAAM_ADDRESS* physOut)
+{
+    unsigned int need;
+    unsigned char* out;
+
+    if (sz <= 0) {
+        return NULL;
+    }
+
+    if (caamPoolVirt == NULL) {
```
aidangarske, line 139: **Initialization can overwrite live physical memory** · _Security_

BEFORE this PR, wolfSSL did not map or modify a fixed Linux physical-memory range. AFTER, the first allocation maps `CAAM_LINUX_POOL_PHYS` and immediately clears all `CAAM_LINUX_POOL_SZ` bytes without verifying that the range is reserved and outside Linux's managed RAM. On systems where `/dev/mem` permits RAM access, a missing or mismatched boot reservation causes initialization to erase live kernel or process pages and then expose them to CAAM DMA. Documentation requiring `mem=` is not an executable validation and does not make this failure safe.

**Fix:** Obtain the DMA region through a kernel-owned reserved-memory, UIO, or VFIO interface that can prove ownership. At minimum, validate the complete physical range against an authoritative reserved-memory resource before mapping or writing it, and fail closed when ownership cannot be established.

wolfcrypt/src/port/caam/caam_linux.c
```diff
@@ -0,0 +1,328 @@
...
+/* The job ring memory, taken from the pool on the first init and then kept.
+ * The pool has no free, so allocating a fresh ring on every init would let
+ * repeated init/teardown cycles walk through the reserved range. */
+#define CAAM_LINUX_RING_SZ 1024
+static unsigned char* caamRingVirt = NULL;
+static CAAM_ADDRESS   caamRingPhys = 0;
+
+/* The job ring registers are inside the block CAAM_SET_BASEADDR already
+ * mapped, so this only offsets into it. The ring itself comes from the pool. */
+int CAAM_SET_JOBRING_ADDR(CAAM_ADDRESS* base, CAAM_ADDRESS* ringInPhy,
```
aidangarske, line 218: **The job ring and DMA pool have no system-wide owner** · _Security_

BEFORE this PR, there was no Linux in-process backend competing for the hardware. AFTER, each process independently maps the same CAAM job-ring registers and fixed DMA pool. The mutex added by the PR is process-local, and the code neither verifies that the kernel driver is detached nor acquires an interprocess ownership lease. Two CAAM-enabled processes, or one process alongside the kernel driver, can therefore overwrite one another's descriptors, keys, plaintext, output, and ring state while DMA is active. This can produce incorrect cryptographic results, cross-process data exposure, or arbitrary memory corruption by the engine.

**Fix:** Move ownership and mapping behind a kernel, UIO, or VFIO driver that grants one exclusive owner and verifies the native CAAM driver is detached. A fallback design must hold a system-wide exclusive lock for the entire mapping lifetime and validate ownership before touching MMIO.

wolfcrypt/src/port/caam/wolfcaam_linux.c
```diff
@@ -0,0 +1,422 @@
...
+            buf[outIdx + 1].TheAddress != 0) {
+        XMEMCPY((void*)buf[outIdx + 1].TheAddress, caamIvBuf, ivSz);
+    }
+
+    /* Clear the key material on every request: it is the secret, and it is
+     * only 48 bytes. The data buffers are cleared at teardown instead -
+     * wiping up to 2 x CAAM_LINUX_AES_MAX per call costs about a fifth of
+     * the throughput, and the next request overwrites them before the engine
+     * reads them anyway. */
+    ForceZero(caamKeyBuf, CAAM_LINUX_KEY_MAX);
```
aidangarske, line 357: **AES plaintext remains in persistent DMA scratch buffers** · _Zeroization_

BEFORE this PR, these persistent Linux bounce buffers did not exist. AFTER, every AES request copies its input and output through `caamInBuf` and `caamOutBuf`, but completion clears only the key and IV. Encryption inputs and decryption outputs can contain plaintext, and unused tails from shorter subsequent requests remain in the reserved physical mapping until global CAAM cleanup. This unnecessarily extends the lifetime of sensitive application data and makes it recoverable through another mapping or process that gains access to the shared pool.

**Fix:** Zero both AES data staging buffers on every success and failure path after the final copy. Preserve the existing key and IV clearing.

**Suggestion:**
```suggestion
ForceZero(caamKeyBuf, CAAM_LINUX_KEY_MAX);
ForceZero(caamIvBuf, CAAM_LINUX_IV_MAX);
ForceZero(caamInBuf, CAAM_LINUX_AES_MAX);
ForceZero(caamOutBuf, CAAM_LINUX_AES_MAX);
```

wolfssl/wolfcrypt/port/caam/wolfcaam_linux.h
```diff
@@ -0,0 +1,92 @@
...
+#define NoActivityReady -1
+#define MemoryOperationNotPerformed -1
+#define CAAM_ARGS_E -3
+
+/* unique devId for CAAM use on crypto callbacks */
+#ifndef WOLFSSL_CAAM_DEVID
+    #define WOLFSSL_CAAM_DEVID 7
+#endif
+
+#include <wolfssl/wolfcrypt/port/caam/wolfcaam_ecdsa.h>
```
aidangarske: **The Linux header breaks valid no-ECC configurations** · _Regression_

BEFORE this PR, the Linux-specific header did not exist. AFTER, it includes ECC, CMAC, and hash CAAM headers even though `settings.h` explicitly disables those offloads for this backend. In a valid `--enable-caam=linux --disable-ecc` configuration, `wolfcaam_hash.h` declares `wc_CAAM_ShaHash(wc_Sha*, ...)` before any header defines `wc_Sha`, causing compilation to fail. The default ECC include chain merely hides this dependency accidentally.

**Fix:** Include only the AES and crypto-callback interfaces supported by the Linux backend, or make every disabled subsystem header independently type-complete.

**Suggestion:**
```suggestion
#include <wolfssl/wolfcrypt/port/caam/wolfcaam_aes.h>
#include <wolfssl/wolfcrypt/cryptocb.h>
```

wolfcrypt/src/port/caam/wolfcaam_linux.c
```diff
@@ -0,0 +1,422 @@
...
+        return Failure;
+    }
+    if (caamLinuxScratch() != 0) {
+        return Failure;
+    }
+
+    /* Pass the driver's status straight back: wc_caamAddAndWait() turns
+     * CAAM_WAITING into RAN_BLOCK_E and wc_GenerateSeed() retries with a
+     * delay, so do not spin here. */
+    ret = caamEntropy(caamEntBuf, buf[0].Length);
```
aidangarske, line 232: **Partial-word entropy requests always fail on the T1040 path** · _Logic_

BEFORE this PR, Linux callers could not reach the shared `caamTRNG` implementation. AFTER, the new entropy dispatcher calls it on pre-era-9 hardware such as the stated T1040 target. For a request whose length is not divisible by four, `caamTRNG` compares the absolute mapped register address against the relative `CAAM_RTENT_MAX` offset (`reg > CAAM_RTENT_MAX`), so the remaining one to three bytes always produce an error. The Linux backend consequently violates the entropy callback's arbitrary-length contract and reports hardware failure for otherwise valid requests.

**Fix:** Change the shared boundary comparison to `reg > caam.baseAddr + CAAM_RTENT_MAX`, audit the following equality check for the same absolute-versus-relative mistake, and test entropy lengths such as 1, 3, 5, and 43 bytes on the legacy TRNG path.

aidangarske (changes requested):
## Skoll Code Review

**Findings:** 6 total — 6 posted, 0 skipped

### Posted findings
- [High] **The /dev/mem capability remains open across exec** — `wolfcrypt/src/port/caam/caam_linux.c:94`
- [High] **Initialization can overwrite live physical memory** — `wolfcrypt/src/port/caam/caam_linux.c:139-147`
- [High] **The job ring and DMA pool have no system-wide owner** — `wolfcrypt/src/port/caam/caam_linux.c:218-246`
- [Medium] **AES plaintext remains in persistent DMA scratch buffers** — `wolfcrypt/src/port/caam/wolfcaam_linux.c:357-358`
- [Medium] **The Linux header breaks valid no-ECC configurations** — `wolfssl/wolfcrypt/port/caam/wolfcaam_linux.h:72-76`
- [Medium] **Partial-word entropy requests always fail on the T1040 path** — `wolfcrypt/src/port/caam/wolfcaam_linux.c:232`

wolfcrypt/src/port/caam/wolfcaam_linux.c
```diff
@@ -0,0 +1,422 @@
...
+ * wc_GenerateSeed() asks for at a time. */
+#define CAAM_LINUX_ENT_MAX 64
+
+static int caamInitDone = 0;
+
+/* Statically initialized where the platform supports it, so there is no window
+ * in which two threads racing into wc_CAAMInitInterface() both try to create
+ * it. The mutex is never destroyed: teardown can run while another thread is
+ * blocked on it, and re-init has to find it usable. */
+static wolfSSL_Mutex caamLock WOLFSSL_MUTEX_INITIALIZER_CLAUSE(caamLock);
```
aidangarske, line 83: Since caamLock is process local but the CAAM engine and DMA pool are system-global, what prevents two processes from initializing and using the same CAAM resources at the same time? Do we need a system-wide ownership lock here?

wolfcrypt/src/port/caam/caam_linux.c
```diff
@@ -0,0 +1,328 @@
...
+
+/* Offset of an address within the pool, or -1 if it is not a pool address.
+ * Linux must have been told not to manage the range (mem= on the kernel
+ * command line), which also makes CONFIG_STRICT_DEVMEM allow the mapping:
+ * the range is no longer reported as System RAM. */
+static long caamPoolOffset(const void* v, int sz)
+{
+    const unsigned char* p = (const unsigned char*)v;
+
+    if (caamPoolVirt == NULL || p < caamPoolVirt || sz < 0 ||
```
aidangarske, line 118: Looks like these are only well defined when in the same C object. Could we convert both addresses to uintptr_t and validate using integer offset/subtraction checks instead? That would also let us avoid overflow in the size bound.

wolfssl/wolfcrypt/port/caam/caam_linux.h
```diff
@@ -0,0 +1,113 @@
...
+#ifndef CAAM_LINUX_POOL_SZ
+    #define CAAM_LINUX_POOL_SZ (256 * 1024)
+#endif
+
+#ifndef CAAM_LINUX_MEM_DEV
+    #define CAAM_LINUX_MEM_DEV "/dev/mem"
+#endif
+
+/* Yield to other threads while polling for job completion. */
+#define CAAM_CPU_CHILL() sched_yield()
```
aidangarske, line 111: should the polling path also have a bounded completion deadline?

wolfcrypt/src/port/caam/wolfcaam_linux.c
```diff
@@ -0,0 +1,422 @@
...
+            break;
+        }
+
+        XMEMCPY((unsigned char*)buf[outIdx].TheAddress + done, caamOutBuf,
+            chunk);
+        done += chunk;
+    }
+
+    /* hand back the updated IV so chaining continues correctly */
+    if (ret == Success && ivSz > 0 && sz > (outIdx + 1) &&
```
aidangarske, line 347: Is partial output/context state undefined after an AES error in the wolfCrypt API? I think output/state need to be committed consistently.

wolfcrypt/src/port/caam/caam_driver.c
```diff
@@ -1844,17 +1880,35 @@ void caamDescInit(DESCSTRUCT* desc, int type, unsigned int args[4],
...
+     * every job times out. Round up instead of assuming. */
+    outOfst = (outOfst + 7U) & ~7U;
 
     /* register the in/out and sizes of job ring */
-    r->Desc     = r->JobIn + (CAAM_JOBRING_SIZE * sizeof(unsigned int));
-    r->JobOut   = r->Desc + (CAAM_DESC_MAX * CAAM_JOBRING_SIZE);
+    r->Desc     = r->JobIn + descOfst;
+    r->JobOut   = r->JobIn + outOfst;
 
     CAAM_INIT_MUTEX(&caam.ring.jr_lock);
```
aidangarske, line 1908: Seems like this assumes a physical range was successfully exuded from the linux ram using the boot config. Could we make the pool configuration explicitly required/validated before clearing it or fail safely when reservation cannot be established?

aidangarske (commented):
Looks good just some questions

aidangarske (approved):

PR #11198, opened 2026-08-18
Rework post-handshake checks for resumed connections.

kareem-wolfssl:
# Description

Fixes zd#22349

# Testing

Built in tests

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/src/tls13.c
+++ b/src/tls13.c
@@ -16686,6 +16686,9 @@ int wolfSSL_request_certificate(WOLFSSL* ssl)
     ssl->msgsReceived.got_certificate = 0;
     ssl->msgsReceived.got_certificate_verify = 0;
     ssl->msgsReceived.got_finished = 0;
+    /* Each round must prove possession again; these are only ever set to 1. */
+    ssl->options.havePeerCert = 0;
+    ssl->options.havePeerVerify = 0;
 
     ret = SendTls13CertificateRequest(ssl, &certReqCtx->ctx, certReqCtx->len);
     if (ret == WC_NO_ERR_TRACE(WANT_WRITE))
--- a/src/tls13.c
+++ b/src/tls13.c
@@ -13064,13 +13064,12 @@ exit_dcv:
 #endif /* !NO_CERTS */
 
 #ifdef WOLFSSL_POST_HANDSHAKE_AUTH
-/* The message being processed belongs to a post-handshake exchange rather than
- * to the enclosing handshake. Whatever resumption, PSK or deferred
- * (post-handshake) verification excused during that handshake, a later
- * post-handshake exchange has to stand on its own. */
-#define TLS13_POST_HANDSHAKE(ssl)   ((ssl)->options.handShakeDone)
+/* Message is being processed after the enclosing handshake completed. Whatever
+ * resumption, PSK or deferred (post-handshake) verification excused during that
+ * handshake, a later post-handshake exchange has to stand on its own. */
+#define TLS13_AFTER_HANDSHAKE(ssl)  ((ssl)->options.handShakeDone)
 #else
-#define TLS13_POST_HANDSHAKE(ssl)   0
+#define TLS13_AFTER_HANDSHAKE(ssl)  0
 #endif
 
 /* Parse and handle a TLS v1.3 Finished message.
@@ -13098,10 +13097,10 @@ int DoTls13Finished(WOLFSSL* ssl, const byte* input, word32* inOutIdx,
 #if !defined(NO_CERTS) && !defined(WOLFSSL_NO_CLIENT_AUTH)
     /* verify the client sent certificate if required */
     if (ssl->options.side == WOLFSSL_SERVER_END &&
-            (!ssl->options.resuming || TLS13_POST_HANDSHAKE(ssl)) &&
+            (!ssl->options.resuming || TLS13_AFTER_HANDSHAKE(ssl)) &&
             (ssl->options.mutualAuth || ssl->options.failNoCert)) {
 #ifdef OPENSSL_COMPATIBLE_DEFAULTS
-        if (ssl->options.isPSK && !TLS13_POST_HANDSHAKE(ssl)) {
+        if (ssl->options.isPSK && !TLS13_AFTER_HANDSHAKE(ssl)) {
             WOLFSSL_MSG("TLS v1.3 client used PSK but cert required. Allowing "
                         "for OpenSSL compatibility");
         }
@@ -13112,7 +13111,7 @@ int DoTls13Finished(WOLFSSL* ssl, const byte* input, word32* inOutIdx,
             /* Exempt only the enclosing handshake; a post-handshake exchange
              * still requires a peer certificate and a valid
              * CertificateVerify. */
-            (!ssl->options.verifyPostHandshake || TLS13_POST_HANDSHAKE(ssl)) &&
+            (!ssl->options.verifyPostHandshake || TLS13_AFTER_HANDSHAKE(ssl)) &&
         #endif
             (!ssl->options.havePeerCert || !ssl->options.havePeerVerify)) {
             ret = NO_PEER_CERT; /* NO_PEER_VERIFY */
@@ -14681,7 +14680,7 @@ static int SanityCheckTls13MsgReceived(WOLFSSL* ssl, byte type)
         #if defined(HAVE_SESSION_TICKET) || !defined(NO_PSK)
             /* RFC 8446 4.3.2: a server authenticating with a PSK must not send
              * this in the main handshake, but may send it post-handshake. */
-            if (ssl->options.pskNegotiated && !TLS13_POST_HANDSHAKE(ssl)
+            if (ssl->options.pskNegotiated && !TLS13_AFTER_HANDSHAKE(ssl)
 #ifdef WOLFSSL_CERT_WITH_EXTERN_PSK
                 && !ssl->options.certWithExternPsk
 #endif
@@ -14843,7 +14842,7 @@ static int SanityCheckTls13MsgReceived(WOLFSSL* ssl, byte type)
         #if defined(HAVE_SESSION_TICKET) || !defined(NO_PSK)
             if (!ssl->options.pskNegotiated ||
                 (ssl->options.side == WOLFSSL_SERVER_END &&
-                 TLS13_POST_HANDSHAKE(ssl))
+                 TLS13_AFTER_HANDSHAKE(ssl))
 #ifdef WOLFSSL_CERT_WITH_EXTERN_PSK
                 || ssl->options.certWithExternPsk
 #endif
@@ -14863,7 +14862,7 @@ static int SanityCheckTls13MsgReceived(WOLFSSL* ssl, byte type)
                      * verify mode (FAIL_IF_NO_PEER_CERT), exactly as for
                      * first-handshake client authentication. */
                     (!ssl->options.verifyPostHandshake ||
-                     TLS13_POST_HANDSHAKE(ssl)) &&
+                     TLS13_AFTER_HANDSHAKE(ssl)) &&
                 #endif
                                            !ssl->msgsReceived.got_certificate) {
                     WOLFSSL_MSG("Finished received out of order - "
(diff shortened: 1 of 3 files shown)


tests/api/test_tls13.c
```diff
@@ -3986,6 +3986,195 @@ int test_tls13_pha(void)
...
+            ExpectIntEQ(ssl_s->options.havePeerVerify, 0);
+        }
+
+        wolfSSL_free(ssl_c);
+        ssl_c = NULL;
+        wolfSSL_free(ssl_s);
+        ssl_s = NULL;
+    }
+
+    wolfSSL_SESSION_free(sess);
```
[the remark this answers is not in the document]
kareem-wolfssl: wolfSSL_SESSION_free has a NULL check and other wolfSSL_SESSION_free calls in the same file are not guarding their calls to wolfSSL_SESSION_free.

GSoJC234:
Confirmed—this fixes the issue. Thanks!

tests/api/test_tls13.c
```diff
@@ -3986,6 +3986,199 @@ int test_tls13_pha(void)
...
+    wolfSSL_free(ssl_c);
+    wolfSSL_free(ssl_s);
+    wolfSSL_CTX_free(ctx_c);
+    wolfSSL_CTX_free(ctx_s);
+#endif
+    return EXPECT_RESULT();
+}
+
+/* Server must reject a post-handshake Finished that omits the Certificate
```
Frauschi, line 4280: This `failNoCert` branch is the only thing that stops a client answering the post-handshake request with an *empty* Certificate on a resumed connection, and neither new test reaches it: `test_tls13_pha_resumption_bare_finished` omits the Certificate flight entirely, so `SanityCheckTls13MsgReceived()` already rejects it with `OUT_OF_ORDER_E` before `DoTls13Finished()` runs.

Could you add a variant of `test_tls13_pha_resumption` that loads no client cert/key on `ctx_c`, so the client replies with `SEND_BLANK_CERT`, and assert `NO_PEER_CERT` on the server read? I checked against the base commit and the server does accept the blank certificate there, so this is a real hole the PR closes with nothing guarding it.

src/tls13.c
```diff
@@ -12749,21 +12758,22 @@ int DoTls13Finished(WOLFSSL* ssl, const byte* input, word32* inOutIdx,
...
         if (
         #ifdef WOLFSSL_POST_HANDSHAKE_AUTH
-            /* Exempt only the initial handshake; a pending post-handshake
-             * CertificateRequest (certReqCtx != NULL) still requires a peer
-             * certificate and a valid CertificateVerify. */
-            (!ssl->options.verifyPostHandshake || ssl->certReqCtx != NULL) &&
+            /* Exempt only the enclosing handshake; a post-handshake exchange
+             * still requires a peer certificate and a valid
+             * CertificateVerify. */
+            (!ssl->options.verifyPostHandshake || TLS13_AFTER_HANDSHAKE(ssl)) &&
```
Frauschi, line 13112: These two reworked exemption terms are the security-relevant half of the change, but nothing evaluates them. Both new tests configure the server with only `WOLFSSL_VERIFY_PEER | WOLFSSL_VERIFY_FAIL_IF_NO_PEER_CERT`, so `verifyPostHandshake` is 0 and `!ssl->options.verifyPostHandshake` short-circuits the whole term regardless of what `TLS13_AFTER_HANDSHAKE()` returns.

I confirmed it by mutating both terms back to the permanently-exempting form, i.e. reintroducing the bypass this PR closes: the full API suite still passes. Setting `WOLFSSL_VERIFY_POST_HANDSHAKE` on the server in `test_tls13_pha_resumption_bare_finished` would cover it.

tests/api/test_tls13.c
```diff
@@ -3986,6 +3986,199 @@ int test_tls13_pha(void)
     return EXPECT_RESULT();
```
Frauschi, line 4076: The comments in `test_tls13_pha` still describe the exemption in terms of `certReqCtx == NULL`, which this PR replaces with `handShakeDone`. You already rewrote the equivalent comments in `tls13.c`; worth updating these two sites as well so the test file does not point at a variable these paths no longer consult.

Frauschi (commented):
Went through the post-handshake rework. The fix itself reads correctly to me: the outer `pskNegotiated` guard is what was letting a bare post-handshake Finished through on a resumed connection, and that is now closed.

Three comments inline, all about test coverage rather than the change itself. Nothing blocking.

kareem-wolfssl:
Retest this please

kareem-wolfssl:
Retest this please

kareem-wolfssl:
Retest this please

kareem-wolfssl:
Retest this please

kareem-wolfssl:
Retest this please

tests/api/test_tls13.c
```diff
@@ -4142,6 +4126,93 @@ int test_tls13_ctx_dh_rotation(void)
...
+    return EXPECT_RESULT();
+}
+
+/* Post-handshake auth over a resumed (ticket-PSK) connection. Client
+ * credentials go on the CTX: FreeHandshakeResources() unloads SSL-owned
+ * certificates, leaving nothing to answer a later request with. */
+int test_tls13_pha_resumption(void)
+{
+    EXPECT_DECLS;
+#if defined(HAVE_MANUAL_MEMIO_TESTS_DEPENDENCIES) && defined(WOLFSSL_TLS13) && \
```
Frauschi, line 4139: All three new tests hard-assert that the second connection resumes, so they need the server to have issued a NewSessionTicket. With `WOLFSSL_NO_DEF_TICKET_ENC_CB` there is no default ticket encryption callback and `CreateTicket()` returns `BAD_TICKET_ENCRYPT`, so no ticket is sent and the tests fail rather than skip. I built with the define and all three die at `wolfSSL_set_session()`. `settings.h` implies that macro from `NO_ASN_TIME`, and the other ticket-resumption tests in this file already carry the guard.

Same one-line addition is needed on the guards of the other two new tests. Fair warning that four unrelated tests in this group already fail in that configuration on master, so this is not the only thing broken there - but the guard matches the convention and costs nothing.

```suggestion
#if defined(HAVE_MANUAL_MEMIO_TESTS_DEPENDENCIES) && defined(WOLFSSL_TLS13) && \
    !defined(WOLFSSL_NO_DEF_TICKET_ENC_CB) && \
```

src/tls13.c
```diff
@@ -13085,21 +13094,22 @@ int DoTls13Finished(WOLFSSL* ssl, const byte* input, word32* inOutIdx,
...
```
Frauschi, line 13114: `havePeerCert` and `havePeerVerify` are never cleared between post-handshake rounds. `wolfSSL_request_certificate()` resets the `msgsReceived` flags but not these two, and outside `wolfSSL_clear()` they are only ever assigned 1. So on a second post-handshake round a client can send a Certificate and skip the CertificateVerify, and the stale `havePeerVerify == 1` carries this check and the equivalent one in `SanityCheckTls13MsgReceived()`.

To be clear, this is pre-existing on the non-resumed path and not something you introduced - but this PR is what makes the block reachable for resumed connections, so it looks like the natural place to clear those two alongside the `msgsReceived` flags.

tests/api/test_tls13.c
```diff
@@ -4142,6 +4126,93 @@ int test_tls13_ctx_dh_rotation(void)
...
+        WOLFSSL_FILETYPE_PEM), WOLFSSL_SUCCESS);
+    /* PHA has to be offered in the ClientHello. */
+    ExpectIntEQ(wolfSSL_CTX_allow_post_handshake_auth(ctx_c), 0);
+    ExpectIntEQ(wolfSSL_CTX_load_verify_locations(ctx_s, cliCertFile, NULL),
+        WOLFSSL_SUCCESS);
+    /* No client certificate wanted for the handshake itself. */
+    wolfSSL_CTX_set_verify(ctx_s, WOLFSSL_VERIFY_NONE, NULL);
+
+    for (i = 0; i < 2 && EXPECT_SUCCESS(); i++) {
+        test_ctx.c_len = test_ctx.s_len = 0;
```
Frauschi: `test_ctx.c_len = test_ctx.s_len = 0;` resets two of the fields the memio harness tracks. `msg_count`, `msg_pos` and `msg_sizes[]` describe the record boundaries, and `test_memio_read_cb()` takes its read length from `msg_sizes[*msg_pos]` rather than from `*len`. `test_memio_clear_buffer()` resets all of them together and `test_dtls.c` uses it consistently.

This is latent today - both buffers happen to be fully drained at the end of the first iteration, so the partial reset is currently a no-op. Worth using the helper anyway so it stays that way, and the same applies to the other two new tests.

```c
        test_memio_clear_buffer(&test_ctx, 0);
        test_memio_clear_buffer(&test_ctx, 1);
```

src/tls13.c
```diff
@@ -14666,8 +14676,9 @@ static int SanityCheckTls13MsgReceived(WOLFSSL* ssl, byte type)
             }
         #endif
         #if defined(HAVE_SESSION_TICKET) || !defined(NO_PSK)
-            /* Server's authenticating with PSK must not send this. */
-            if (ssl->options.pskNegotiated
+            /* RFC 8446 4.3.2: a server authenticating with a PSK must not send
```
Frauschi, line 14681: This relaxes the RFC 8446 4.3.2 check so a PSK-authenticating server may send CertificateRequest post-handshake. The new tests cover the positive direction, but nothing pins the other half: that a CertificateRequest arriving during a resumed handshake still returns `SANITY_MSG_E`.

The discriminator looks sound to me - `handShakeDone` is only set on the client in `SendTls13Finished()`, after which `serverState` cannot return to `SERVER_ENCRYPTED_EXTENSIONS_COMPLETE` - but that is exactly the kind of invariant a later change to the post-handshake state rewind could break silently. A test that injects the CertificateRequest before the client's Finished and asserts `SANITY_MSG_E` would bound the relaxation on both sides.

tests/api/test_tls13.c
```diff
@@ -4142,6 +4126,93 @@ int test_tls13_ctx_dh_rotation(void)
 
     wolfSSL_free(ssl);
     wolfSSL_CTX_free(ctx);
+    #endif
```
Frauschi: This `#endif` and the one closing `test_tls13_accept_state_dh_copy` picked up four spaces of indentation with the insertion; both sit at column 0 on master.

```suggestion
#endif
```

Frauschi (changes requested):
Second pass over the post-handshake rework. The three points from my last review are all addressed, thanks. I re-ran the mutation check on a `--enable-all --enable-debug --enable-postauth` build: reverting the outer `pskNegotiated` guard, or either `verifyPostHandshake` term, is now caught by `test_tls13_pha_resumption_bare_finished`.

One thing worth knowing about the new blank-certificate test. The rejection it observes comes from `ProcessPeerCerts()`, which errors on an empty Certificate regardless of resumption state, so the test does not actually exercise the `(!ssl->options.resuming || TLS13_AFTER_HANDSHAKE(ssl))` change in `DoTls13Finished()` - I mutated that guard back to `(!ssl->options.resuming)` and nothing fails. Its only reachable trigger is a client that sends a real Certificate and then omits the CertificateVerify. I do not think that needs a fourth test; just flagging that the coverage is not quite where the comment suggests.

Comments inline.

**MC/DC whitebox vectors still drive the removed certReqCtx operand; handShakeDone is uncovered** (`tests/unit-mcdc/test_tls13_whitebox.c:591-615`)
`tests/unit-mcdc/test_tls13_whitebox.c` includes `src/tls13.c` and exists to complete MC/DC pairs on this exact `finished` guard. Its comment still documents the decision as `(side == SERVER && certReqCtx != NULL)`, and the vector that sets `ssl->certReqCtx = &reqCtx` no longer touches anything the decision reads: `handShakeDone` is 0 on the fixture and does not appear anywhere in that file, so the vector now takes the same branch as its partner and the independence pair is lost. `wb_sc_reset()` does not reset `handShakeDone` either, so the fixture base is incomplete.

Could you drive `options.handShakeDone` instead of `certReqCtx` there, fix the guard comment, and add `handShakeDone = 0` to `wb_sc_reset()`? The file still compiles and passes because the calls are `(void)`-cast, which is what makes this easy to miss.

Frauschi (approved):

JacobBarthelmeh (approved):

PR #11199, opened 2026-08-18
Add NXP QorIQ SEC (CAAM) hardware crypto port for PowerPC T-series

dgarske:
wolfSSL has had no PowerPC QorIQ SEC support. The existing `wolfcrypt/src/port/caam/` code builds only for i.MX under QNX, INTEGRITY or the i.MX8 SECO firmware, so on a T-series part the only acceleration available was the scalar PowerPC assembly. This adds a self-contained port under `wolfcrypt/src/port/nxp/` that drives the SEC through wolfCrypt's crypto callback layer, selected at runtime by `devId`.

Three commits, kept separate because the first two touch shared API rather than the port:

1. **A crypto callback hook for finite field Diffie-Hellman.** `WC_PK_TYPE_DH` was in the enum but had no entry in `wc_CryptoInfo`, no `wc_CryptoCb_Dh()` and no call site in `dh.c`, and `DhKey` had no `devId`, so no port could offload DH at all. Only key agreement is routed; the SP 800-56A group and peer-key validation is hoisted above the dispatch so a device cannot skip it.
2. **`WOLFSSL_NO_DH_GEN_PARAMS`**, with `--enable-keygen=no-dh-params` and a matching CMake option, so a constrained target can drop safe-prime search while keeping DH key generation.
3. **The QorIQ SEC port itself.** AESA (AES-128/192/256 CBC/CTR/ECB, and AES-GCM with AAD where the engine checks the tag on decrypt), PKHA (ECDSA sign/verify and ECDH on any prime curve wolfCrypt carries, RSA public and private to 4096 bit, FFDHE agreement), MDHA (driver API only; hashing is not routed through the callback yet) and RNG4 for DRBG seeding. Everything else returns `CRYPTOCB_UNAVAILABLE` and runs in software, and that fallback covers run-time failures too, so an unsupported case is a performance question rather than a correctness one.

**Measured on real silicon.** T2080E at 1.2 GHz against plain C: AES-128-GCM 8.07 to 107.42 MiB/s (13.3x), ECDSA P-256 verify 49.4 to 592.0 ops/s (12.0x), RSA-2048 private 1.88 to 58.29 ops/s (31x), DH-2048 agree 16.5 to 477.1 ops/s (28.9x). T1040E at 1.4 GHz against PowerPC assembly, the more conservative baseline: AES-128-GCM 35.80 to 91.92 MiB/s (2.6x), RSA-2048 private 1.71 to 17.19 ops/s (10.1x), ECDSA verify 43.5 to 350.2 ops/s (8.1x).

**Verified** on a Curtiss-Wright VPX3-152 (T2080E) and an NXP T1040D4RDB: `wolfcrypt_test()` passes in full on both, including AES, AES-GCM, RSA, DH and ECC, pushing about 3.4 million descriptors per run. The T2080 has PKHA v2 and the T1040 v1; the same driver covers both unchanged. Also builds under `--enable-all` on x86-64 and compiles clean under `NO_RSA`, `WOLFSSL_SEC_QORIQ_NO_PKHA`, `WOLFSSL_SEC_QORIQ_NO_RSA`, `WOLFSSL_RSA_PUBLIC_ONLY` and `WOLFSSL_SMALL_STACK`.

**The PKHA built-in curve table does not work on this silicon.** A descriptor is supposed to be able to name a preloaded curve by index, which the i.MX port relies on. On the QorIQ parts every flag position in bits 25:17 crossed with every curve index 0..31 (288 combinations swept on hardware) is refused with DECO error `0x82`, and the "already hashed" PROTINFO flag with `0x81`. The port supplies the domain parameters explicitly instead, four extra descriptor words for any prime curve rather than the five built in.

**One fix outside the port.** `DecodeECC_DSA_Sig_Ex()` ends with `mp_clear(r); mp_clear(s);` on its error path, but with `init` set the halves are only initialized as items parse, so an early decode failure reaches those calls with uninitialized values. Under SP math `sp_clear()` then zeroes a garbage-sized run of the caller's stack, and the crash lands far from the cause. Found on a T2080 with a hardware write watchpoint, traced back from a branch to address zero. It is latent for any caller passing `init` non-zero, the async ECC verify path included. Happy to split it out if you would rather review it separately.

**Known limitations**, all in `wolfcrypt/src/port/nxp/README.md`: hashing is not routed through the callback; buffers must be physically contiguous with no scatter-gather; GCM takes a 12-byte IV and a full 16-byte tag only; 32-bit descriptor pointers only; one job ring, polled, serialised on the hardware mutex. An ECDSA digest wider than the curve order uses the leftmost-order-bits reduction, except on a curve whose order does not fill the fixed-width block, which is declined because wolfCrypt finishes that truncation with a sub-byte shift the layout cannot express (no curve in common use is affected).

**This port is bare metal.** Its `/dev/mem` backend initialises on a stock Linux but offloads nothing: 32-bit descriptor pointer mode cannot reach user pages, which sit above 4 GB on a 36-bit part. Rather than push that further, the Linux path is handled by a separate change that gives the existing `wolfcrypt/src/port/caam/` driver a Linux user space backend (`--enable-caam=linux`), validated on the same T1040D4RDB. That one accelerates AES-CBC/CTR/ECB and the TRNG at roughly 0.6x software throughput, because operands have to be staged into a reserved physical pool; the reason to enable it is the TRNG and moving work off the core, not speed. Bare metal remains the fast path, and the READMEs say so.


dgarske:
Jenkins retest this please

dgarske:
Jenkins retest this please

wolfcrypt/src/dh.c
```diff
@@ -2137,6 +2145,47 @@ static int wc_DhAgree_Sync(DhKey* key, byte* agree, word32* agreeSz,
...
+     * Never on the ct path: wc_DhAgree_ct() exists to give the caller a
+     * constant-time agreement, and a callback makes no such promise. Offload
+     * there would silently drop the guarantee the caller asked for, so that
+     * entry point stays in software. */
+    if (ct == 0)
+    #ifndef WOLF_CRYPTO_CB_FIND
+    if (key->devId != INVALID_DEVID)
+    #endif
+    {
+        ret = wc_CryptoCb_Dh(key, priv, privSz, otherPub, pubSz, agree,
```
aidangarske, line 2164: **Clear cached DH validation after callback completion**

The PR sets `key->nb->pubKeyValidated` before dispatching the new callback, but every callback result other than `CRYPTOCB_UNAVAILABLE` returns without clearing it. The existing nonblocking software paths clear this flag after terminal success or failure because it is only valid for one agreement. After one successful callback-backed agreement, a second call using the same `DhKey` skips `wc_DhCheckPubKey_ex()` and sends a different peer value directly to the device. The QorIQ `secCbDh()` path relies on this outer validation and does not repeat it.

**Fix:** Clear `pubKeyValidated` on every terminal callback result, including the shared-secret-is-one path. Preserve it only for a callback result that genuinely represents an in-progress operation. Add a regression that completes one callback agreement and then attempts another with an invalid peer value.

wolfcrypt/src/port/nxp/sec_qoriq_cb.c
```diff
@@ -0,0 +1,887 @@
...
+
+    #ifdef NO_INLINE
+        #include <wolfssl/wolfcrypt/misc.h>
+    #else
+        #define WOLFSSL_MISC_INCLUDED
+        #include <wolfcrypt/src/misc.c>
+    #endif
+#endif
+
+/* The fallback promise has to cover run time failures too, not just what
```
aidangarske: **Do not fall back after hardware may have mutated caller state**

The new fallback policy converts every `WC_HW_E` into `CRYPTOCB_UNAVAILABLE`, including errors returned after `IRJA` submitted a descriptor. At that point AES may already have overwritten `out` and the CBC/CTR IV, while RSA private operations explicitly run in place. The generic wolfCrypt entry points immediately retry in software when they receive `CRYPTOCB_UNAVAILABLE`; an in-place operation can therefore consume hardware-modified output instead of the original input, and CBC/CTR can retry with an advanced IV. Timeouts and late job-status failures cannot safely be treated like pre-submission capability failures.

**Fix:** Distinguish failures proven to occur before submission from failures after submission. Propagate post-submission failures as hard errors, or execute hardware operations into private bounce buffers and commit output and chaining state only after clean completion.

wolfcrypt/src/port/nxp/sec_qoriq.c
```diff
@@ -0,0 +1,577 @@
...
+        return ret;
+    }
+
+    /* Publish the entry, then tell the engine one job was added. */
+    secWrite(dev->jr, SEC_QORIQ_IRJA, 1);
+    dev->inIdx = (dev->inIdx + 1) % SEC_QORIQ_RING_SIZE;
+
+    waited = 0;
+    while (secRead(dev->jr, SEC_QORIQ_ORSF) == 0) {
+        if (++waited > SEC_QORIQ_POLL_MAX) {
```
aidangarske, line 404: **Do not return while a failed reset can still DMA through stack descriptors**

On timeout, the code correctly notes that the engine may still own the stack-resident descriptor. If `secJrReset()` also times out, however, it only sets `initialized` to zero and returns. The descriptor and caller buffers can then leave scope and be reused while the engine still has their physical addresses, allowing later DMA into unrelated stack or application memory. Setting the flag only prevents new submissions; it does not quiesce the current one. It also makes `wc_SecQoriqFree()` return immediately, leaving the callback, mappings, and ring allocations installed.

**Fix:** Guarantee controller-level DMA quiescence before returning from this path. If ring reset cannot do so, use a confirmed controller/DMA reset or retain all referenced storage with persistent lifetime. Unregister the callback and preserve enough state for cleanup rather than making `wc_SecQoriqFree()` a no-op.

wolfcrypt/src/port/nxp/sec_qoriq_linux.c
```diff
@@ -0,0 +1,325 @@
...
+         * the process lacks CAP_SYS_ADMIN and the kernel zeroed the PFN.
+         * Aiming engine DMA at physical 0 is not an acceptable guess. */
+        WOLFSSL_MSG("sec_qoriq: pagemap PFN is zero, need CAP_SYS_ADMIN");
+        return 0;
+    }
+
+    return (pfn * SEC_QORIQ_PAGE_SZ) + (vaddr % SEC_QORIQ_PAGE_SZ);
+}
+
+/* A buffer is only usable by the engine if it is physically contiguous.
```
aidangarske: **Do not DMA through unpinned pagemap translations**

The Linux backend snapshots PFNs from `/proc/self/pagemap` and hands them directly to the SEC, but it never pins or DMA-maps ordinary caller pages. Only the separately allocated ring pages use `MAP_LOCKED`; stack descriptors and caller key, IV, input, output, AAD, and tag buffers do not. Such pages can be migrated or undergo copy-on-write after translation, and an untouched output can still refer to a shared zero page. The SEC may consequently DMA through a stale PFN or into memory not exclusively owned by the caller.

**Fix:** Use a kernel-mediated DMA mapping or a dedicated pinned, physically contiguous bounce pool exposed by a driver. Copy caller data into that pool and copy results back after completion; pagemap observation alone is not a valid DMA ownership mechanism.

wolfcrypt/src/port/nxp/sec_qoriq_cb.c
```diff
@@ -0,0 +1,887 @@
...
+    mp_int mpr;
+    mp_int mps;
+    SEC_CB_DECL(buf, MAX_ECC_BYTES * SEC_QORIQ_PK_SCRATCH_UNITS);
+
+    if (info->pk.eccsign.in == NULL || info->pk.eccsign.out == NULL ||
+            info->pk.eccsign.outlen == NULL) {
+        return CRYPTOCB_UNAVAILABLE;
+    }
+
+#if defined(WOLFSSL_ECDSA_DETERMINISTIC_K) || \
```
aidangarske, line 348: **Honor explicitly configured ECDSA nonces**

The callback declines deterministic ECDSA but does not detect a nonce installed through the public `wc_ecc_sign_set_k()` API under `WOLFSSL_ECDSA_SET_K` or `WOLFSSL_ECDSA_SET_K_ONE_LOOP`. `wc_ecc_sign_hash()` invokes the callback before its software path consumes `key->sign_k`, so the QorIQ path signs with an unrelated RNG4-generated nonce and reports success. This breaks fixed-vector and caller-controlled signing semantics and leaves the configured nonce pending for an unexpected later software signature.

**Fix:** Also return `CRYPTOCB_UNAVAILABLE` whenever `sign_k` or `sign_k_set` indicates an explicitly supplied nonce. Add a callback regression using `wc_ecc_sign_set_k()` and verify that the expected deterministic signature is produced by the software fallback.

wolfcrypt/src/port/nxp/sec_qoriq_rng.c
```diff
@@ -0,0 +1,211 @@
...
+    SecQoriqDev* dev = wc_SecQoriqGetDev();
+    word32 entDelay = SEC_QORIQ_RTSDCTL_ENT_DLY_MIN;
+    word32 scfgr;
+    int ret = WC_HW_E;
+
+    if (dev == NULL) {
+        return WC_HW_E;
+    }
+
+    /* Do not touch the TRNG parameters if a handle is already live. */
```
aidangarske, line 126: **Serialize RNG4 initialization**

The lazy initialization sequence is not atomic. Two independent signing or seeding threads can both observe `RDSTA[IF0]` clear, interleave `secKickTrng()` register writes, and submit competing instantiation jobs. After one thread establishes the handle, the other can continue changing TRNG parameters while that handle is live, which the implementation itself says is invalid. The job-ring mutex only covers each `secInstantiateRng()` submission, not the surrounding status check and register programming.

**Fix:** Protect the complete initialization state transition with a separate initialization mutex/state machine, or perform it once during `wc_SecQoriqInit()` before callback registration and permanently record whether hardware seeding is available.

wolfcrypt/src/port/nxp/sec_qoriq_aes.c
```diff
@@ -0,0 +1,440 @@
...
+    if (out == NULL && inSz > 0) {
+        return BAD_FUNC_ARG;
+    }
+    if (aad == NULL && aadSz > 0) {
+        return BAD_FUNC_ARG;
+    }
+    if (keySz != 16 && keySz != 24 && keySz != 32) {
+        return BAD_FUNC_ARG;
+    }
+    if (tagSz == 0 || tagSz > SEC_QORIQ_AES_BLOCK) {
```
aidangarske: **Reject truncated GCM tags in the direct driver API**

The router restricts GCM to `SEC_QORIQ_GCM_TAG_SZ`, and the header and README state that the engine supports only a full 16-byte tag. The underlying driver entry point instead accepts every size from 1 through 16. Direct-driver builds using `WOLFSSL_SEC_QORIQ_NO_CRYPTOCB` therefore reach a descriptor shape that the documented hardware contract says is unsupported.

**Fix:** Enforce the same exact 16-byte requirement at the driver boundary and add direct encrypt/decrypt tests for rejected truncated tags.

**Suggestion:**
```suggestion
if (tagSz != SEC_QORIQ_GCM_TAG_SZ) {
```

wolfcrypt/src/port/nxp/sec_qoriq_rng.c
```diff
@@ -0,0 +1,211 @@
...
+    if (wc_SecQoriqRead(dev->regs, SEC_QORIQ_RDSTA) & SEC_QORIQ_RDSTA_IF0) {
+        dev->rngReady = 1;
+        return 0;
+    }
+
+    /* A failure usually means the statistical checks rejected the entropy
+     * at this sample length, so widen it and retry. */
+    while (entDelay < SEC_QORIQ_RTSDCTL_ENT_DLY_MAX) {
```
aidangarske: **Attempt the configured maximum entropy delay**

The retry loop stops at 12400 and never tries `SEC_QORIQ_RTSDCTL_ENT_DLY_MAX` itself because it uses `<`. The constant is documented as the maximum usable delay, and the existing CAAM RNG initialization loop includes its maximum. Hardware that rejects the 12400 sample but accepts 12800 will unnecessarily lose RNG offload.

**Fix:** Make the upper bound inclusive and cover the final retry in a mocked failure-sequence test.

**Suggestion:**
```suggestion
while (entDelay <= SEC_QORIQ_RTSDCTL_ENT_DLY_MAX) {
```

wolfcrypt/src/port/nxp/sec_qoriq.c
```diff
@@ -0,0 +1,577 @@
...
+    }
+    if (dev->initialized == 0) {
+        return WC_HW_E;
+    }
+
+    /* The failure paths below return before a new status is read, so a
+     * caller classifying the outcome must not see a stale one. */
+    dev->lastStatus = 0;
+
+    dev->jobCount++;
```
aidangarske: **Count jobs only after submission**

`jobCount` is documented as the number of descriptors actually submitted and is recommended as evidence that offload occurred, but it is incremented before descriptor translation, cache maintenance, ring-space validation, and the `IRJA` write. In the Linux fallback case, for example, an unreachable stack descriptor increments this counter even though the engine never saw a job.

**Fix:** Move the increment immediately after the successful `IRJA` publication, or rename and document it as an attempted-job counter.

wolfcrypt/src/port/nxp/README.md
```diff
@@ -48,7 +49,436 @@ NOTE: Both can be defined with no problem.
...
+- AES-CBC/CTR/ECB require whole 16-byte blocks. Partial-block CTR, where
+  wolfCrypt keeps leftover keystream, falls back to software.
+- One job ring, polled, serialised on the wolfSSL hardware mutex. No
+  interrupt support and no concurrent submission.
+- Output buffers should not share a cache line with data another thread
+  writes while a job is in flight. The driver flushes before submitting and
+  invalidates afterwards, which is safe for its own accesses, but a
+  concurrent writer to an adjacent object in the same line can still lose an
+  update on a non-coherent path.
+- There is no in-tree test for the port yet. It was validated with an
```
aidangarske: **Add reproducible in-tree tests for the QorIQ state machine**

The PR adds more than 3,400 lines of descriptor construction, DMA/cache handling, fallback behavior, status classification, reset logic, and mutable singleton state, but explicitly provides no in-tree test. The external hardware harness cannot protect normal CI from regressions in the most failure-prone logic. The new generic DH callback test does not execute this port.

**Fix:** Add a fake environment backend with controlled MMIO, DMA translations, completion statuses, and timeouts. Exercise pre- versus post-submission failures, reset failure, in-place AES behavior, GCM authentication, descriptor limits, RNG retry/initialization, and PKHA status decoding; retain a separate real-hardware smoke gate.

wolfcrypt/src/port/nxp/sec_qoriq_hash.c
```diff
@@ -0,0 +1,188 @@
...
+    if (inSz > 0) {
+        ret = wc_SecQoriqCacheFlush((void*)in, inSz);
+        if (ret != 0) {
+            return ret;
+        }
+    }
+
+    blockSz = secHashBlockSz(algo);
+
+    /* Feed the message in chunks a single command can describe. Only the
```
aidangarske, line 118: **Bound or stream large single-shot hashes**

Each hash chunk consumes two descriptor words, while the operation and final digest store also need descriptor space. The 64-word limit therefore allows only 30 data loads, placing an undocumented aggregate input limit near 1.96 MB. Larger valid `word32` inputs fill the descriptor and return `BUFFER_E`, despite the API and README describing chunking as the solution for messages above the per-command 64 KB limit.

**Fix:** Implement extended-length or multi-descriptor streaming, or add an explicit checked maximum with matching header/README documentation and boundary tests.

wolfcrypt/src/include.am
```diff
@@ -255,6 +263,17 @@ endif
 EXTRA_DIST += wolfcrypt/src/port/sealsq/README.md
 
 
+if BUILD_SEC_QORIQ
```
aidangarske, line 266: **Wire the QorIQ backend into CMake**

The PR adds an Autotools option, conditional, installed header, and source list for QorIQ, but adds no corresponding CMake option or entries in `generate_lib_src_list()`. Consequently the new port cannot be enabled through the project's CMake interface. Manually defining `WOLFSSL_SEC_QORIQ` is also insufficient because the implementation files are absent from the target, leaving references such as `wc_SecQoriqInit()` unresolved.

**Fix:** Add a CMake backend selector, the required definitions and crypto-callback enablement, all QorIQ source files, and the generated options-header macro, matching the Autotools behavior.

aidangarske (changes requested):
## Skoll Code Review

**Scan type:** `review`

**Findings:** 12 total — 12 posted, 0 skipped
**12 finding(s) posted as inline comments** (see file-level comments below)

### Posted findings
- [High] **Clear cached DH validation after callback completion** — `wolfcrypt/src/dh.c:2164-2183`
- [High] **Do not fall back after hardware may have mutated caller state** — `wolfcrypt/src/port/nxp/sec_qoriq_cb.c:65-80`
- [High] **Do not return while a failed reset can still DMA through stack descriptors** — `wolfcrypt/src/port/nxp/sec_qoriq.c:321-334`
- [High] **Do not DMA through unpinned pagemap translations** — `wolfcrypt/src/port/nxp/sec_qoriq_linux.c:216-238`
- [High] **Honor explicitly configured ECDSA nonces** — `wolfcrypt/src/port/nxp/sec_qoriq_cb.c:337-344`
- [High] **Serialize RNG4 initialization** — `wolfcrypt/src/port/nxp/sec_qoriq_rng.c:121-138`
- [Medium] **Reject truncated GCM tags in the direct driver API** — `wolfcrypt/src/port/nxp/sec_qoriq_aes.c:270-271`
- [Medium] **Attempt the configured maximum entropy delay** — `wolfcrypt/src/port/nxp/sec_qoriq_rng.c:129`
- [Medium] **Count jobs only after submission** — `wolfcrypt/src/port/nxp/sec_qoriq.c:288-316`
- [Medium] **Add reproducible in-tree tests for the QorIQ state machine** — `wolfcrypt/src/port/nxp/README.md:477-480`
- [Medium] **Bound or stream large single-shot hashes** — `wolfcrypt/src/port/nxp/sec_qoriq_hash.c:102-145`
- [Medium] **Wire the QorIQ backend into CMake** — `wolfcrypt/src/include.am:266-275`

aidangarske:
Jenkins retest this please 
(logs lost)

wolfcrypt/src/port/nxp/sec_qoriq.c
```diff
@@ -0,0 +1,714 @@
...
+    }
+
+    dev->outRing = (word32*)wc_SecQoriqDmaAlloc(outSz, &dev->outRingPhys);
+    if (dev->outRing == NULL) {
+        wc_SecQoriqDmaFree(dev->inRing, dev->inRingPhys, inSz);
+        dev->inRing = NULL;
+        return MEMORY_E;
+    }
+
+    if ((dev->inRingPhys >> 32) != 0 || (dev->outRingPhys >> 32) != 0) {
```
aidangarske, line 521: **Above-4GB ring allocation makes the Linux port fatally fail global initialization** · _Logic_

The new Linux allocator obtains ordinary locked mappings and returns their physical addresses without ensuring they fit the engine's 32-bit descriptor-pointer mode. `secRingsAlloc()` detects an above-4GB ring address but returns `MEMORY_E`. The new `wolfCrypt_Init()` hook only converts `NOT_COMPILED_IN` into software fallback; any other result marks the global wolfCrypt state bad and returns an error. Thus a valid but unreachable ring allocation prevents all wolfCrypt initialization instead of declining QorIQ offload.

**Fix:** Classify addresses unreachable in 32-bit pointer mode as an unavailable backend so `wolfCrypt_Init()` retains software operation. Add a Linux-backend test that supplies an above-4GB ring physical address and verifies successful software-only initialization.

**Suggestion:**
```suggestion
return NOT_COMPILED_IN;
```

wolfcrypt/src/port/nxp/sec_qoriq.c
```diff
@@ -0,0 +1,714 @@
...
+    if (((chanumMs >> SEC_QORIQ_CHANUM_MS_JRNUM_SHIFT) & 0xF) <=
+            dev->jrIndex) {
+        WOLFSSL_MSG("sec_qoriq: requested job ring does not exist");
+        wc_SecQoriqUnmapRegs(dev->regs);
+        return BAD_FUNC_ARG;
+    }
+
+    /* Refuse rather than hand the engine addresses it reads as 64-bit. */
+    mcfgr = secRead(dev->regs, SEC_QORIQ_MCFGR);
+    if (mcfgr & SEC_QORIQ_MCFGR_LONG_PTR) {
```
aidangarske, line 634: **Nonfatal unsupported-pointer-mode initialization leaves an unmapped register pointer for cleanup** · _Resource Leak_

When the engine is already in unsupported long-pointer mode, the new initializer unmaps `dev->regs` and returns `NOT_COMPILED_IN`, but leaves both `dev->regs` and `dev->jr` populated with dangling pointers. The new `wolfCrypt_Init()` path treats that result as success, so normal `wolfCrypt_Cleanup()` subsequently calls `wc_SecQoriqFree()`. That function tests only `dev->regs != NULL` and invokes `secJrReset()` through the unmapped register address. On the Linux backend this dereferences unmapped memory during cleanup.

**Fix:** Clear the device structure after every post-map initialization failure, preferably through one shared cleanup path. Add a simulated long-pointer-mode initialization test followed by `wolfCrypt_Cleanup()`.

**Suggestion:**
```suggestion
wc_SecQoriqUnmapRegs(dev->regs);
        XMEMSET(dev, 0, sizeof(*dev));
        return NOT_COMPILED_IN;
```

aidangarske (commented):
## Skoll Code Review

### Posted findings
- [Medium] **Above-4GB ring allocation makes the Linux port fatally fail global initialization** — `wolfcrypt/src/port/nxp/sec_qoriq.c:516-518`
- [Medium] **Nonfatal unsupported-pointer-mode initialization leaves an unmapped register pointer for cleanup** — `wolfcrypt/src/port/nxp/sec_qoriq.c:613-616`

aidangarske (dismissed):

aidangarske (dismissed):

dgarske:
Jenkins retest this please. FIPS

aidangarske (approved):

aidangarske (approved):

PR #11202, opened 2026-08-19
TI C2000: hardware AES accelerator and oscillator-jitter entropy source

dgarske:
`Add TI C2000 AESA hardware AES port and fix CHAR_BIT!=8 AES defects` brings up the on-chip AESA accelerator (a TI EIP-120t) for ECB/CBC/CTR at 128/192/256 bits as a crypto-callback device gated on `WOLFSSL_C2000_AES`, so software AES stays compiled in, one image can run identical NIST vectors through both paths and compare, and anything unsupported returns `CRYPTOCB_UNAVAILABLE`. Measured 5.2x-7.3x on ECB/CBC and 3.6x-4.4x on CTR. Two device behaviours are worth a look, both documented inline: octet packing is done explicitly (driverlib takes `uint32_t*`, and staging deliberately uses `uint32_t` rather than `word32`, which is only 32-bit under `WC_16BIT_CPU`), and CTR is driven through hardware ECB with a software counter because the block's own counter does not carry across octet boundaries the way `IncrementAesCounter()` does. The commit also fixes four `CHAR_BIT != 8` defects that all reproduce in a pure software build with no device registered and are no-ops at `CHAR_BIT == 8`: counter carry never propagating (`if (++inOutCtr[i])` sees `0xFF + 1` as truthy `0x100`) across five increment helpers plus `DecrementKeyWrapCounter`, `shiftLeftArray()` letting CFB1 feedback cells exceed `0xFF`, `FlattenSzInBits()` building the GCM length block at 16 bits because `sizeof` counts cells, and `roll_auth()` putting oversized cells into the CCM CBC-MAC input.

`Add TI C2000 oscillator-jitter entropy source and harden wc_RNG_TestSeed for wide bytes` replaces the dev-only seed. The part has no TRNG, so the noise bit is the LSB of a Dual-Clock Comparator measurement -- PLL edges counted inside a window of INTOSC cycles, i.e. the relative phase drift of two independent oscillators -- measured on hardware at 0.924 bits of min-entropy per bit (SP800-90B 6.3.1 MCV with bias and correlation screening, not a full non-IID assessment), oversampled about 4x and fed to the existing Hash-DRBG. Almost none of that is C2000-specific, so the reusable half is a new platform-agnostic module in `random.c` behind `WOLFSSL_NOISE_SRC`: a port supplies one `wc_NoiseSampleCb` returning a raw octet plus a config struct, and `wc_NoiseSrc_*` owns the SP800-90B 4.3 startup test, the 4.4.1/4.4.2 continuous tests, the entropy budget, the SHA-256 conditioner and the latched fail-closed state, with no mallocs and no locks. Only test verdicts latch, not sampler errors. The payoff is that those failure paths, which previously only fired on broken silicon, now run on every host build via `noisesrc_test()` with synthetic stuck, biased, sampler-error and periodic sources. Also fixes `wc_RNG_TestSeed()`, which indexes a 256-entry array with a byte cell that can exceed 255 where `CHAR_BIT != 8` -- unreachable until a real entropy source returns unmasked values, which is exactly what this commit adds. Verified with `make check` at `--enable-all`, again with `-DWOLFSSL_NOISE_SRC`, and again with `--enable-smallstack` (the C2000 config); `IDE/C2000/compile.sh` compiles both port files and `random.c` with the entropy gate under cl2000 at `CHAR_BIT == 16`; on-hardware runs cover 14 HW-vs-SW AES cross-KATs and the entropy startup, liveness and DRBG checks, though they predate the `wc_NoiseSrc_*` refactor and a re-run is queued.


wolfcrypt/src/port/ti/ti-c2000-entropy.c
```diff
@@ -0,0 +1,293 @@
...
+    DCC_enableSingleShotMode(base, DCC_MODE_COUNTER_ZERO);
+
+    /* DONE only latches with the done/error signals enabled. */
+    DCC_enableErrorSignal(base);
+    DCC_enableDoneSignal(base);
+    DCC_enableModule(base);
+
+    limit = ((uint32_t)WOLFSSL_C2000_ENTROPY_WINDOW * 256UL) + 100000UL;
+    for (guard = 0; guard < limit; guard++) {
+        if (DCC_getSingleShotStatus(base) || DCC_getErrorStatus(base)) {
```
SparkiDev: Why is it when there is an error it is a success?

wolfcrypt/src/port/ti/ti-c2000-entropy.c
```diff
@@ -0,0 +1,293 @@
...
+            *outCnt = (C2000_CNT1_SEED -
```
SparkiDev: Why have a large range when you only take the bottom bit?

wolfcrypt/src/port/ti/ti-c2000-entropy.c
```diff
@@ -0,0 +1,293 @@
...
+        s0   = WOLFSSL_C2000_ENTROPY_SRC1_CLK;
+    }
+
+    acc = 0;
+    for (b = 0; b < 8; b++) {
+        ret = c2000_dccSample(base, s0, &cnt);
+        if (ret != 0) {
+            return ret;
+        }
+        acc = (word16)(acc | (word16)((cnt & 1U) << b));
```
SparkiDev: You can shift acc and or in the bottom bit to be more efficient.

wolfcrypt/src/port/ti/ti-c2000-entropy.c
```diff
@@ -0,0 +1,293 @@
...
+ * still valid; driverlib itself waits on (DONE || ERR) and reads it either
+ * way.  Only a real timeout returns WC_HW_E - folding one in as a zero would
+ * silently inject a deterministic bit. */
+static int c2000_dccSample(uint32_t base, DCC_Count0ClockSource src0,
+                           uint32_t* outCnt)
+{
+    uint32_t guard;
+    uint32_t limit;
+
+    DCC_clearErrorFlag(base);
```
SparkiDev, line 177: A lot going on here.
I don't know anything about the platform but want you to be sure all of it is necessary.

dgarske:
Jenkins retest this please

SparkiDev (dismissed):

dgarske:
Jenkins retest this please

dgarske:
Jenkins retest this please. History lost (so lame)

philljj:
Retest this please.

(no logs)

philljj:
multi-test failed:
```
failed: check-source-text
```

I think it's these:
```
[check-source-text] [1 of 3] [58e516cbe3]
    autogen.sh 58e516cbe3...   real 0m8.404s  user 0m0.003s  sys 0m0.005s
    configure...   real 0m14.694s  user 0m0.002s  sys 0m0.006s
wolfcrypt/test/test.c:27222:    if (wc_NoiseSrc_Init(&src) != ENTROPY_RT_E)
wolfcrypt/test/test.c:27231:    if (wc_NoiseSrc_Init(&src) != WC_HW_E)
...
```

philljj (dismissed):
PRB-multi-test-script is failing. I think it's just from wolfcrypt/test/test.c. See note above

SparkiDev (dismissed):

dgarske:
Jenkins retest this please

SparkiDev (approved):

PR #11205, opened 2026-08-19
tests: cover extended key usage OID sum collisions in non-wolfSSH builds

gasbytes:
# Description

The KeyPurposeId OID byte check landed upstream in https://github.com/wolfSSL/wolfssl/pull/11361/. Its regression test is gated on WOLFSSL_WOLFSSH, so a default build has no coverage for the check.

Fixes zd#22337

# Testing

Added a regression test that tests the behaviour with an actual id-kp-serverAuth as oid and one that must collide and check that it still doesn't allow server authentication. Also updated the MC/DC white-box coverage.

# Checklist

 - [x] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/tests/api/test_asn.c
+++ b/tests/api/test_asn.c
@@ -3685,6 +3685,105 @@ int test_wc_DecodeExtKeyUsage_ssh_oid_collision(void)
     return EXPECT_RESULT();
 }
 
+#if !defined(NO_ASN) && !defined(NO_RSA) && !defined(NO_CERTS) && \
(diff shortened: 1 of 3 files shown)


gasbytes:
Fixed some conflicts against master.

The fix was superseded by:
https://github.com/wolfSSL/wolfssl/pull/11361/

Rebased to use https://github.com/wolfSSL/wolfssl/pull/11361/changes/11a6e284e11dc8bc7d744ac9eb9837141ca9f354 instead as fix, and left the coverage for non-wolfSSH builds.

PR #11206, opened 2026-08-19
SE050 CI: update pinned simulator for PKCS7 coverage

LinuxJedi:
Pin bump to the simulators commit that adds --enable-pkcs7 to the wolfcrypt CI build, so the suite's pkcs7signed_test and pkcs7enveloped_test regression-guard the SE050 RSA verify buffer fix and the ECDH software-key fallback. The pinned commit is fetchable by CI once the corresponding simulators change merges.


philljj (approved):

philljj:
note: the SE050 simulator tests all passed.

The ThreadXBuild Test is hung, and unrelated to this PR.

PR #11210, opened 2026-08-19
drbg: add thread-safe option, --disable-rng-autolock and --disable-rng-autofork to opt out

kaleb-himes:
# Description

Feature addition

# Testing

Test case added in wolfcrypt/test/test.c

### Lock cost: master c74d6c669 vs this PR (eddacc65a)

Same default `./configure` for both. macOS (M1 Max, native) is the inline pthread mutex variant; Linux (arm64 container on the same box) is the semaphore variant with the fork handlers on. Medians of 3 to 7 runs, master and PR interleaved.

### NOTE: Numbers re-generated after merge conflict resolution on 18 September 2026

**1. TLS: no measurable change.** In a threaded build every `WOLFSSL` owns its `WC_RNG`, so TLS never contends the lock; it pays one uncontended enter/exit per RNG call (a few per handshake, one per CBC record). Measured with `examples/server -B` / `client -B` and `-b`, and with a thread-per-connection echo tool at 1, 4 and 8 concurrent sessions, TLS 1.3 AES-GCM and TLS 1.2 AES-CBC, 16 KB and 1 KB records.

| Config | Metric | master | PR | Delta |
|---|---|---|---|---|
| TLS 1.3 AES128-GCM-SHA256 | TX MBps | 164.4 | 164.5 | +0.1% |
| TLS 1.3 AES128-GCM-SHA256 | RX MBps | 81.2 | 82.4 | +1.5% |
| TLS 1.2 ECDHE-RSA-AES128-SHA256 | TX MBps | 186.9 | 186.8 | -0.1% |
| TLS 1.2 ECDHE-RSA-AES128-SHA256 | RX MBps | 223.4 | 223.9 | +0.2% |
| TLS 1.3 handshake | connect ms | 5.59 | 5.33 | within noise |

Repeats of the same tree scatter more than this (351 to 401 MB/s, 4.2 to 7.4 ms in one config), so these are noise.

**2. What one uncontended lock costs per `wc_RNG_GenerateBlock` call.**

| Primitive | Linux | macOS |
|---|---|---|
| mutex lock + unlock | 2.4 ns | 9.7 ns |
| `pthread_setcancelstate` pair | 13.8 ns | 249.1 ns |
| `sem_wait` + `sem_post` | 12.7 ns | unavailable, `sem_init` is ENOSYS |
| **lock path total** | **26.5 ns** (semaphore + cancel state) | **258.8 ns** (mutex + cancel state) |

### End to End:
| Measurement | Linux | macOS |
|---|---|---|
| 32-byte generate, master to PR | 0.702 to 0.726 us (+3.4%, **fixed 24 ns**) | 1.06 to 1.31 us (+23.8%) |
| 1 MiB, chunked | 2778 to 2793 us/MiB (+0.5%) | 4509 to 4620 us/MiB (+2.5%, inside a 4.3% spread) |

Fixed cost per call, so it only shows on tiny generates. On macOS it is Apple's `pthread_setcancelstate`, not the mutex.

**3. Scaling.** With an instance per thread (what TLS does) the RNG scales with cores: aggregate throughput rises 6.4x on 8 threads, and each thread keeps ~80% of its solo rate, which is ordinary core sharing. A single instance shared by N threads serializes on its lock, as any shared locked resource does. On master that sharing is an unlocked data race, so there is no master column.

| PR, calls/s | 1 thread | 2 | 4 | 8 |
|---|---|---|---|---|
| own instance per thread, aggregate | 1.45M | 2.90M | 5.50M | 11.0M |
| own instance per thread, per thread | 1.45M | 1.45M | 1.37M | 1.37M |
| one shared instance, aggregate | 1.45M | 863k | 855k | 773k |
| one shared instance, per thread | 1.45M | 387k | 190k | 65k |

### NOTE: This increased from 80% of solo to 95% of solo now!

(Linux; macOS is within a few percent of these.) A shared instance suits occasional use across threads, not a hot loop.

Other tests performed:

Platforms: Linux (Ubuntu, kernel 6.17.0-1032-oem, Intel Core Ultra 9 285K with
RDRAND/RDSEED) and macOS 26.4.1 (Apple clang 21). Kernel modules built against
the running kernel's headers on the Linux box.

## Kernel space versus user space

* Kernel and user space were built and compared as separate targets, not assumed equivalent
  * kernel builds define neither `WC_RNG_HAVE_AUTO_LOCK` nor `WC_RNG_LOCK_ATFORK`; verified.
  * `nm` on the built `libwolfssl.ko`: zero occurrences of `RngLockEnter`, `wc_ForkLock_Enter`, `wc_ForkLock_New`, `ForkPrepare`, `forkMine`. No lock code reaches the kernel module
  * the module was diffed against one built from a fresh `upstream/master` (`git archive`, not a stale copy): `.text` differs by 256 bytes and 3 symbols, which are the three `_local` helper functions plus their CFI prologues, and contain no locking
  * before the DRBG-mutator split, that same comparison was byte-identical `.text` and identical symbol tables
* Defaults differ by design and both were exercised
  * user space: the lock is on by default, the fork handlers are on where the platform supports them
  * kernel space: off, unconditionally, by the gate
  * `--enable-rng-extras` (upstream's own lock) is off by default in user space and on in the kernel build; both states tested

## Configuration matrix

* 14 user-space configurations, each configured, built and run: default, extras, nolock, nolock_extras, noatfork, noatfork_extras, nolock_noatfork, nolock_noatfork_extras, bank, bank_extras, singlethreaded, cryptonly_extras, all, all_extras
  * every one records which gate macros it actually produced, so a silently-disabled feature cannot pass as tested
* CMake build path built and `ctest` run separately from autotools
* `--enable-intelrand` built and run on real Intel hardware with RDRAND
* FIPS v7 (ready) built and run three ways: this PR, this PR plus extras, and an upstream control; `wc_RunAllCast_fips passed` in each
* `make check` on both Linux and macOS

## Coexistence and isolation

* 8 configurations check that upstream's `wc_RNG_lock_get` still links whenever `--enable-rng-extras` is on, including with our lock disabled
  * confirms disabling our feature never removes upstream's API
* `bank` and `bank_extras`: our lock, upstream's lock, next-seed and RBGC machinery all live in one binary, with the bank's own test suite passing
* `fullmutex_shared`: where an instance already carries upstream's full mutex, our lock stands down and reuses it rather than adding a second; run under TSan as well

## Deadlock testing

* Reseed, Linux
  * 50 sequential reseed plus generate cycles on one instance
  * 1.4 to 1.6 million concurrent reseed plus generate cycles across 4 threads on one shared instance, watchdogged so a stall is reported rather than hanging
  * the suite itself runs 24 concurrent reseeds per execution (4 threads, 96 draws, 2 reseeders reseeding every 8 draws)
* Fork, Linux
  * 200 forks while 4 threads generate: 200/200, zero hangs
  * fork while the forking thread holds the lock (what a seed callback doing `fork()` looks like): returns cleanly
  * fork from a signal handler: measured, documented as the one uncovered case
  * fork whose `fork()` itself fails (forced with `RLIMIT_NPROC`): parent handlers still run, lock still usable
* No Recursive locking possible
  * all 7 functions that acquire the lock walked programmatically; none calls another
  * every call site of a locking entry in `random.c` enumerated and individually cleared
  * a deliberate recursive acquisition is included as a control and does deadlock, so the detector test is known to work, PR is green, no recursion.

## Race conditions

* ThreadSanitizer on Linux across 5 configurations including the full-mutex coexistence build: 0 warnings
* AddressSanitizer plus LeakSanitizer: 0 errors
* Concurrent `wc_RNG_DRBG_Stir()` against concurrent generates: 48 TSan data races before the fix, 0 after
* 400,000 generated blocks collected under 4.5 million concurrent stirs and compared pairwise for duplicates
* fork divergence: a forked child never reproduces the parent's bytes, with and without the fork handlers

## Performance

* TLS, Linux, fresh `upstream/master` versus this branch with identical flags, 3 repeats
  * TLS 1.3 AES-GCM: TX +0.1%, RX +1.5%; TLS 1.2 AES-CBC: TX -0.1%, RX +0.2%; handshake within noise
  * repeats of the same tree scatter more than the gap between trees
* Lock cost on the bare primitives: Linux 26.5 ns, macOS 258.8 ns of which 249 ns is Apple's `pthread_setcancelstate`
* 32-byte generate: Linux 0.702 to 0.726 us (fixed 24 ns), macOS 1.06 to 1.31 us
* 1 MiB in legal chunks: Linux +0.5%, macOS inside run-to-run spread
* Scaling, 1/2/4/8 threads: instance per thread scales 7.6x on 8 threads keeping 95% of solo rate; one shared instance serializes, as any shared locked resource does
* The DRBG-mutator locks added no cost to the generate path (729 to 722 ns, noise); `wc_RNG_DRBG_Stir()` itself pays 41 ns for its new lock

## Method

* Every fix has a negative control: the test was run against the unfixed tree first and had to fail, then against the fixed tree and had to pass
  * examples: the fork wedge wedged `fork()` by fork 3 and is now clean over 12,000 forks; the reseed race returned success on a condemned instance 3/3 and now refuses 3/3; the ABI pin fails the build when a flag bit is renumbered
* Controls were also used to prove a detector works before trusting a negative result, and one of them caught a flaw in the test rather than the code
* Static analysis: skoll `multi review,bugs,concurrency` over the branch: 26 findings, 0 blockers, 0 Critical, 0 High, all addressed or adjudicated against POSIX, Windows, macOS, Android, and other OS's threading documentation.

### Reseed on a failing instance (the status read that used to sit outside the lock)

- Both entry points covered, wc_RNG_DRBG_Reseed() and wc_RNG_DRBG_Reseed_Now(), run as two passes of one test
- The window is forced rather than waited for: a holder thread takes the instance lock, signals over a pipe that it holds it, then condemns the instance one fifth of the way into the hold, so the reseed always arrives while the instance is healthy and completes against a condemned one
- The reseed must refuse with RNG_FAILURE_E; returning 0 fails the test
- Negative control: against the tree before the fix it returned success on a condemned instance 3/3, and refuses 3/3 after
- Test lives in the suite as rng_reseed_status_test, so it runs on every execution rather than on demand

### Cygwin 3.5.3, gcc 12.4.0, Windows 11 (third platform, and the first outside glibc to run the fork-handler path)

- ./configure --disable-examples, build, and testwolfcrypt each returned 0, with RANDOM, RNGFLAG, RNGTHRD and RNGSVC all passing
- Not a token build: that configure produces WC_RNG_ATFORK, WC_RNG_HAVE_AUTO_LOCK and WC_RNG_LOCK_ATFORK, so the lock and the fork handlers are both compiled in and exercised
- The build was checked for real work rather than a no-op: 74 objects compiled, cygwolfssl-45.dll and testwolfcrypt.exe produced
- The _GNU_SOURCE define this platform needs was verified directly, negative control first
without it, dladdr and Dl_info are undeclared, a hard error
- dlfcn.h:49 gates both behind #if __GNU_VISIBLE, newlib's feature-test macro
RTLD_NOLOAD and RTLD_NODELETE are both present
- At runtime dladdr resolves and the RTLD_NOLOAD | RTLD_NODELETE reopen returns a handle

# Checklist

 - [X] added tests
 - [X] updated/added doxygen


wolfcrypt/src/random.c
```diff
@@ -686,9 +686,44 @@ static int Hash_DRBG_Reseed(DRBG_internal* drbg, const byte* seed, word32 seedSz
         *              and array_add_one (shared utility) which both must
         *              remain available to SHA-512-only builds */
 
+#ifndef WC_NO_DRBG_THREAD_SAFE
+/* Thread-safe DRBG support.  A compare-exchange flag
+ * rather than a lock, so it is legal in every context the DRBG runs in and
+ * contention waits with WC_RELAX_LONG_LOOP().  Returns 1 when this call took
```
douzzer: This is precisely what is impermissible in the kernel.

Spin-locking a DRBG in a kernel atomic context is inherently destabilizing.  The current native kernel RNG implementation allocates one generator per CPU to systematically avoid it.

With `LINUXKM_DRBG_GET_RANDOM_BYTES`, currently in use by multiple customers, our DRBG replaces that native kernel RNG, using callbacks.  Our DRBG then inherits the strictest caller population in the kernel: under `CONFIG_SLAB_FREELIST_RANDOM` every fresh slab draws from it — including `GFP_ATOMIC` allocations in hardirq and softirq — along with page-allocator shuffling under `zone->lock`, neighbour-timer softirq work, and per-packet IPv4 ID selection.

As for the lock mechanism implemented here, it's a hand-rolled spinlock missing both properties that make spinlocks legal. The holder takes no `preempt_disable()` and no bh/irq exclusion, so a softirq/hardirq contender on the same CPU spins unboundedly against an interrupted or preempted holder that can never run — a hard lockup, the exact hazard `spin_lock_bh()`/`spin_lock_irqsave()` exist to preclude.  With `LINUXKM_DRBG_GET_RANDOM_BYTES`, that is a single-CPU self-deadlock requiring no concurrency at all: an interrupt arriving mid-generate need only make a `GFP_ATOMIC` allocation that brings in a fresh slab, and freelist randomization re-enters the very DRBG instance whose holder it interrupted.  Cross-CPU, a preempted holder parks atomic-context spinners for scheduling-quantum timescales — softlockup/RCU-stall watchdog territory even when it eventually completes.

douzzer (dismissed):

douzzer:
The changes in this PR are viable in non-kernel builds, but need to be gated out in kernel module build.  See attached report.
[pr11210-technical-review.pdf](https://github.com/user-attachments/files/31272699/pr11210-technical-review.pdf)

kaleb-himes:
retest this please.

wolfcrypt/src/random.c
kaleb-himes: Changes were made to make this off by default with linuxkm. Resolved.

douzzer (dismissed):
@kaleb-himes and I will work on refactoring this to coexist/leverage conflicting in-flight  changes to the DRBG.  "stand by"

kaleb-himes:
retest this please...

wolfcrypt/src/random.c
```diff
@@ -695,9 +695,86 @@ static int Hash_DRBG_Reseed(DRBG_internal* drbg, const byte* seed, word32 seedSz
...
+/* Returns 0 once the caller may use the instance, or a negative error. */
+static int RngLockEnter(WC_RNG* rng)
+{
+#if !defined(WC_PLATFORM_NO_FORK) && \
+    defined(HAVE_GETPID) && !defined(WOLFSSL_NO_GETPID)
+    WC_RNG_lock_arg_t self = (WC_RNG_lock_arg_t)getpid();
+#else
+    /* No getpid(), so every process stamps the same value and a lock left by
+     * a fork cannot be told apart. */
+    WC_RNG_lock_arg_t self = 1;
```
kaleb-himes: Suggestion not taken, some threading implies fork is possible even if HAVE_GETPID not explicitly set. Fix implemented was an opt out of fork with `WC_PLATFORM_NO_FORK`

kaleb-himes:
retest this please (no logs)

wolfcrypt/src/random.c
```diff
@@ -485,6 +497,226 @@ static int UnlockDrbgState(void)
...
+ * the flag is decided under drbgStateMutex, the pin taken outside it. */
+int wc_RngAtForkInit(void)
+{
+    int ret = LockDrbgState();
+    if (ret != 0)
+        return ret;
+    if (!rngAtForkSet) {
+        /* pin outside the lock: dlopen() takes the loader lock */
+        (void)UnlockDrbgState();
+        wc_RngPinImage((void*)(wc_ptr_t)RngAtForkPrepare);
```
kaleb-himes, line 585: I would argue with Fenrir's finding here.

A NULL from `dlopen(name, RTLD_NOLOAD | RTLD_NODELETE)` means the loader has no handle for this image, not that the pin failed ([dlopen(3), RTLD_NOLOAD](https://man7.org/linux/man-pages/man3/dlopen.3.html)).

- On glibc that is what you get when wolfSSL is linked into the executable: glibc refuses to dlopen an executable ([dl-load.c:1260,1302](https://sourceware.org/git/?p=glibc.git;a=blob;f=elf/dl-load.c)), and dlclose only unmaps objects that came in through dlopen ([dl-close.c:124](https://sourceware.org/git/?p=glibc.git;a=blob;f=elf/dl-close.c)). Nothing can unmap it.

- Tested with glibc 2.35: shared object = handle (pinned), static libwolfssl.a in the executable = NULL, fully static binary = dladdr fails. macOS returns a handle for both.
- The proposed fix (skip pthread_atfork on NULL) would therefore silently disable the fork handlers for every static link on Linux.
- glibc removes a shared object's fork handlers itself when it is unloaded ([cxa_finalize.c:117-120](https://sourceware.org/git/?p=glibc.git;a=blob;f=stdlib/cxa_finalize.c), [pthread_atfork.c:51](https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/pthread/pthread_atfork.c));
 
FreeBSD does the same ([thr_fork.c:109-123](https://github.com/freebsd/freebsd-src/blob/main/lib/libthr/thread/thr_fork.c)); musl never unloads ([dlclose is a no-op](https://wiki.musl-libc.org/functional-differences-from-glibc.html)); macOS never unloads a library the executable links against (`man 3 dlclose`), and a dlopen'ed one gets a handle, so NODELETE pins it.

- Builds without RTLD_NOLOAD/RTLD_NODELETE fail at compile time in wc_port.c; configure and CMake enable the handlers only after linking this exact pin.

Line numbers as of glibc and FreeBSD git HEAD, 2026-09-07.

In my humble opinion Fenrir is wrong on this one.

kaleb-himes: Additional: I checked back on older versions as well, glibc 2.3.6, 2.12, 2.17, 2.28 in addition to HEAD; only the PIE refusal is newer (2.30), and it does not affect this.

kaleb-himes: Final Notes: This threading feature addition has been a real twist with LLM's doing the checking for us.

The earlier suggestion, from Fenrir that I had to revert entirely after doing more reading on the subject was not just bad, it was undefined behavior. It suggested to reclaim a lock held by a thread that no longer exists in the child, reading the documentation was a wakeup call to just how wrong Fenrir had been on that one (sources linked below).

This latest suggestion from Fenrir is a silent security regression masquerading as a fail-close. Both are the same pattern: a fix shaped by the finding rather than by the platform.

https://pubs.opengroup.org/onlinepubs/9699919799/functions/fork.html
https://pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_mutex_lock.html
https://pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_mutex_destroy.html
https://pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_atfork.html

Note the nuance of the child handler: glibc resets its malloc arena locks in the child with a fresh lock init (malloc/arena.c, __malloc_fork_unlock_child, https://sourceware.org/git/?p=glibc.git;a=blob;f=malloc/arena.c). That is the same thing I did here. The difference from "reclaim" is that under prepare-acquire the forking thread is the owner of every lock at fork time. This means nothing in the child is touching a lock owned by a vanished thread, and the libc that defines the behavior uses the same approach.

wolfcrypt/src/wc_port.c
```diff
@@ -415,6 +422,25 @@ static WC_DECLARE_INIT_STATE(wolfcrypt_init_state);
...
+#endif
+/* The fork handlers can never be unregistered, so the image that holds them
+ * is pinned against dlclose() before they are registered. */
+WOLFSSL_LOCAL void wc_RngPinImage(void* fn)
+{
+    Dl_info info;
+    const char* name;   /* a pointer on most libcs, an array on Cygwin */
+    if (dladdr(fn, &info) == 0 || (name = info.dli_fname) == NULL ||
+        name[0] == '\0' ||
+        dlopen(name, RTLD_NOLOAD | RTLD_NODELETE | RTLD_LAZY) == NULL) {
```
kaleb-himes, line 437: Adjudicated: "RTLD_NOLOAD returns NULL when the object is not resident", is exactly the executable case; glibc refuses to dlopen an executable at all, so no handle can ever exist for a dlclose to use. The suggested fix disables the handlers for every static Linux link. Citations already on the PR, I'm surprised you got this wrong again @wolfSSL-Fenrir-bot

wolfcrypt/src/random.c
```diff
@@ -485,6 +492,257 @@ static int UnlockDrbgState(void)
...
+{
+    WC_RNG_LOCK* n;
+    if (rngListDead)
+        return;
+    if (RngSemWait(&rngListSem) != 0) {
+        rngListDead = 1;   /* nothing held, and never again */
+        return;
+    }
+    for (n = rngList; n != NULL; n = n->next) {
+        if (!n->broken && RngSemWait(&n->sem) != 0)
```
kaleb-himes, line 531: Adjudicated: "prepare runs in the forking thread", is why this cannot be fixed without owner tracking, which is the design POSIX made undefined and this PR has since removed after following advice on this topic before from @wolfSSL-Fenrir-bot. The suggested undefined behavior had to be removed after reading the documentation more closely. How are you still getting this wrong?

wolfcrypt/src/random.c
```diff
@@ -485,6 +492,257 @@ static int UnlockDrbgState(void)
...
+    int ret = 0;
+    int old = PTHREAD_CANCEL_ENABLE;
+    (void)pthread_setcancelstate(PTHREAD_CANCEL_DISABLE, &old);
+    while (sem_wait(s) != 0) {
+        if (errno != EINTR) {
+            ret = BAD_MUTEX_E;
+            break;
+        }
+    }
+    (void)pthread_setcancelstate(old, NULL);
```
kaleb-himes, line 516: Adjudicated: XSH 2.9.5: cancelling a thread with asynchronous cancellation enabled inside any function that is not async-cancel-safe is undefined, and only three pthread functions are required to be. Under deferred cancellation the registry sections contain no cancellation point: the sem_wait runs with cancellation off, and the rest is sem_init, stores, and sem_post.

wolfcrypt/src/random.c
```diff
@@ -485,6 +492,284 @@ static int UnlockDrbgState(void)
...
```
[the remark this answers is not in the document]
kaleb-himes: Adjudicated: Repeat with a new basis, "disabled cancellation remains pending." True, but XSH 2.9.5 says a pending request under the default deferred type acts only "until a cancellation point is reached," and nothing executed while a registry semaphore is held is one. Neither 2.9.5.2 list contains sem_init, sem_post, pthread_atfork or fork, and implementations "shall not introduce cancellation points into any other functions.

wolfcrypt/src/random.c
```diff
@@ -485,6 +492,284 @@ static int UnlockDrbgState(void)
...
+{
+    WC_RNG_LOCK* n;
+    for (n = rngList; n != NULL; n = n->next) {   /* forward links stay whole */
+        if (rngListDead) {
+            n->broken = 1;
+            continue;
+        }
+    #ifndef NO_SHA256
+        if (n->drbg != NULL)
+            ((DRBG_internal*)n->drbg)->reseedCtr = WC_RESEED_INTERVAL;
```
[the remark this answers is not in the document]
kaleb-himes: Adjudicated: The handler itself does only stores and sem_post. The allocation is in the application's later generate call in the child, and under small stack every draw allocates anyway, so excluding those builds would remove no allocation, only the lock release. Master already forces the same first-draw reseed in a child via getpid on every default build. The docs already state the allocator must work after fork. And on every platform where the handlers are enabled, fork itself takes and releases the malloc locks in both processes: glibc posix/fork.c, musl fork.c, FreeBSD thr_fork.c, all cited by line.

wolfcrypt/src/wc_port.c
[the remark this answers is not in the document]
kaleb-himes: Adjudicated: Nothing has changed since last time. "RTLD_NOLOAD returns NULL when the object is not resident", is exactly the executable case; glibc refuses to dlopen an executable at all, so no handle can ever exist for a dlclose to use. The suggested fix disables the handlers for every static Linux link. Citations already on the PR, I'm surprised you keep getting this wrong @wolfSSL-Fenrir-bot

wolfcrypt/test/test.c
```diff
@@ -27298,12 +27335,457 @@ WOLFSSL_TEST_SUBROUTINE wc_test_ret_t random_test(void)
...
+        args[i].ret = 0;
+        if (wolfSSL_NewThread(&threads[i], &rng_thread_test_worker,
+                              &args[i]) != 0) {
+            break;
+        }
+        started++;
+    }
+
+    for (i = 0; i < started; i++) {
+        if (wolfSSL_JoinThread(threads[i]) != 0)
```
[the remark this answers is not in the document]
kaleb-himes: Bikeshedding: This is the mirror image of the Jenkins clang-tidy finding that removed the deliberate leak in cf4c38f73. POSIX pthread_join has a single error, EDEADLK, and only under "may fail"; joining a non-joinable or already-joined thread is undefined behavior rather than an error. The workers are created joinable, joined once, and join nothing, so a failure return has no defined meaning.

JacobBarthelmeh (dismissed):

wolfcrypt/src/wc_port.c
```diff
@@ -420,6 +430,218 @@ static WC_DECLARE_INIT_STATE(wolfcrypt_init_state);
...
+    int ret = 0;
+    int old = PTHREAD_CANCEL_ENABLE;
+    (void)pthread_setcancelstate(PTHREAD_CANCEL_DISABLE, &old);
+    while (sem_wait(s) != 0) {
+        if (errno != EINTR) {
+            ret = BAD_MUTEX_E;
+            break;
+        }
+    }
+    (void)pthread_setcancelstate(old, NULL);
```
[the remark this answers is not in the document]
kaleb-himes: Fenrir cited POSIX accurately: the rationale does say re-enabling with a pending request cancels the thread. But I measured it to be sure, and two independent things kill the finding:

It does not happen on glibc. A thread with a pending cancel survived the restore, ran the next line, joined normally, and its cleanup handler never ran. pthread_setcancelstate is not a cancellation point.
Even if an implementation behaved as the rationale warns, the thread cannot die holding a semaphore: between acquiring one and posting it back there is no cancellation point anywhere (sem_init, pointer writes, sem_post, atomic loads are none of them).

And wc_ForkLock_Enter already keeps cancellation disabled for the entire hold, which is exactly the discipline POSIX's rationale recommends.

wolfcrypt/test/test.c
```diff
@@ -27988,6 +28030,446 @@ WOLFSSL_TEST_SUBROUTINE wc_test_ret_t noisesrc_test(void)
...
+            close(fd[1]);
+    }
+    if (piped >= 2) {
+        close(hfd[0]);
+        if (!started)
+            close(hfd[1]);
+    }
+    if (piped >= 3 && !started)
+        close(gfd[0]);
+    XFREE(h, HEAP_HINT, DYNAMIC_TYPE_TMP_BUFFER);
```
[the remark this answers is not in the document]
kaleb-himes: POSIX specifies exactly one error for pthread_join: [EDEADLK], "A deadlock was detected." Non-joinable, self-join and double-join are undefined behavior, not error returns. Our threads are default-joinable, joined exactly once, from a different thread, acyclic. The branch cannot be taken.

The suggested fix guards an unreachable path and introduces a real leak path that LeakSanitizer would flag if it ever executed. Declining this suggestion.

wolfcrypt/test/test.c
```diff
@@ -27988,6 +28030,446 @@ WOLFSSL_TEST_SUBROUTINE wc_test_ret_t noisesrc_test(void)
...
+
+    /* Fewer than two threads tested nothing. */
+    if (started < 2)
+        ERROR_OUT(WC_TEST_RET_ENC_NC, out_free);
+
+    /* All pairs, as a smoke test. */
+    nblocks = started * WC_RNG_THREAD_TEST_DRAWS;
+    for (i = 1; i < nblocks; i++) {
+        for (j = 0; j < i; j++) {
+            if (XMEMCMP(out + ((size_t)i * WC_RNG_THREAD_TEST_BLKSZ),
```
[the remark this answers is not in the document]
kaleb-himes: Fenrir's claim: "Removing RngLockEnter() can still produce unique blocks and WC_DRBG_OK."

I ran the mutation, which is the only way to settle it: stubbed out RngLockEnter/RngLockExit in a --disable-rng-atfork build and ran the suite 20 times. Caught 20 out of 20, every single time at the duplicate-block check, test.c:28725. The mutex path is covered, and the claim is false.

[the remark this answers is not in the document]
kaleb-himes: Actually on second thought on this one... The narrow version of the point is accurate: that assertion is a race detector, empirically 20/20 but not deterministic by construction, and the deterministic hold-and-time test is gated to the atfork build. That is cheap to close, and with public API. I'll do it. The holder can take wc_LockMutex(&rng->mutex) when rng->autoLockInited, since the field is visible, about 12 lines in test.c.

Worth doing as coverage, not as a defect fix.

wolfssl/wolfcrypt/random.h
```diff
@@ -34,6 +34,33 @@
...
+ * what makes the child throw away the pooled and banked bytes it inherited,
+ * which the handlers themselves do not touch. */
+#if defined(WC_RNG_HAVE_AUTO_LOCK) && defined(WOLFSSL_PTHREADS) && \
+    defined(WC_RNG_ATFORK) && !defined(__APPLE__) && \
+    defined(HAVE_GETPID) && !defined(WOLFSSL_NO_GETPID) && \
+    !defined(WOLFSSL_NO_MALLOC) && !defined(HAVE_ENTROPY_MEMUSE) && \
+    !defined(WC_RNG_BANK_SUPPORT) && !defined(WOLFSSL_STATIC_MEMORY) && \
+    !defined(HAVE_WNR) && !defined(WOLFSSL_CHECK_MEM_ZERO) && \
+    !defined(WOLFSSL_TRACK_MEMORY) && !defined(WOLFSSL_MEM_FAIL_COUNT)
+    #define WC_RNG_LOCK_ATFORK
```
douzzer: These gates are really fidgety and specific -- the RNG-specific clauses (e.g. `defined(WC_NO_HASHDRBG)`) seem fine, but the references to non-RNG stuff are not ideal for an inside-the-boundary file and should be more generic.  E.g. the tests for kernel module builds should be `defined(WOLFSSL_KERNEL_MODE_DEFAULTS)`, not specific kernel flavors.

But better is to move all the fidgety moving-target non-RNG-specific tests to settings.h and wc_port.h, where arcane dependencies are the norm, and are represented out-of-boundary where we have freedom of maneuver.  The random.h setup clauses can then default the features off, with default-on provided by the easy-to-tweak settings.h/wc_port.h clauses.

The in-boundary setup clauses should also have positive overrides to allow forcing on the features, e.g. explicit `WC_RNG_WANT_AUTO_LOCK`, and when the feature isn't activated there should be an explicit undef, e.g. `#undef WC_RNG_HAVE_AUTO_LOCK`.

wolfssl/wolfcrypt/random.h
```diff
@@ -609,6 +637,20 @@ struct WC_RNG {
...
+#ifdef WC_RNG_LOCK_ATFORK
+    /* NULL until wc_InitRng succeeds.  Initialize only a new or freed WC_RNG:
+     * wc_InitRng over a live one leaks this and grows the fork registry that
+     * every fork() walks. */
+    struct wc_ForkLock* autoLock;   /* defined in wc_port.c */
+#elif defined(WC_RNG_HAVE_AUTO_LOCK)
+    #ifndef WC_RNG_HAVE_LOCK_FULL_MUTEX
+    /* One RNG mutex: the full mutex build declares it above, so it is
+     * declared here only when that build did not. */
+    wolfSSL_Mutex mutex;
```
douzzer: There's no need to have `.mutex` appear twice in the `struct` def.  We can just "hoist" the member like this:
```
#if defined(WC_RNG_HAVE_LOCK_FULL_MUTEX) || defined(WC_RNG_HAVE_AUTO_LOCK)
    wolfSSL_Mutex mutex;
#endif
#ifdef WC_RNG_HAVE_LOCK
[...]
```

wolfcrypt/src/random.c
```diff
@@ -558,6 +565,107 @@ static int UnlockDrbgState(void)
...
+    if (!rng->autoLockInited)
+        return;
+#ifdef PTHREAD_CANCEL_DISABLE
+    old = rng->autoLockCancel;   /* read before the unlock hands the slot on */
+#endif
+    (void)wc_UnLockMutex(&rng->mutex);
+#ifdef PTHREAD_CANCEL_DISABLE
+    (void)pthread_setcancelstate(old, NULL);
+#endif
+}
```
douzzer: This is port-specific logic that belongs somewhere like wc_port.c, as usual.

wolfcrypt/src/random.c
```diff
@@ -5399,22 +5533,35 @@ int wc_RNG_GenerateBlock(WC_RNG* rng, byte* output, word32 sz)
...
 
+    ret = RngLockEnter(rng);   /* held across every other backend */
+    if (ret != 0)
+        return ret;
+
 #ifdef HAVE_INTEL_RDRAND
-    if (IS_INTEL_RDRAND(intel_flags))
-        return wc_GenerateRand_IntelRD(NULL, output, sz);
+    if (IS_INTEL_RDRAND(intel_flags)) {
+        ret = wc_GenerateRand_IntelRD(NULL, output, sz);
```
douzzer: `wc_GenerateRand_IntelRD()` doesn't want a lock around it -- it's a pure native intrinsic machine instruction with no exposed state -- and if it did want a lock, it couldn't be locked from here anyway.  The `WC_RNG` is one of potentially many instances each with its own `.mutex`; the RDRAND generator is a common resource shared by all `WC_RNG`s.

wolfcrypt/src/random.c
```diff
@@ -5399,22 +5533,35 @@ int wc_RNG_GenerateBlock(WC_RNG* rng, byte* output, word32 sz)
...
+        RngLockExit(rng);
+        return ret;
+    }
 #endif
 
 #if defined(WOLFSSL_SILABS_SE_ACCEL) && defined(WOLFSSL_SILABS_TRNG)
-    return silabs_GenerateRand(output, sz);
+    ret = silabs_GenerateRand(output, sz);
```
douzzer: `silabs_GenerateRand()` doesn't want a lock around it -- it's an opaque generator with no exposed state -- and if it did want a lock, it couldn't be locked from here anyway.  The `WC_RNG` is one of potentially many instances each with its own `.mutex`; the "silabs" generator is a common resource shared by all `WC_RNG`s.

wolfcrypt/src/random.c
```diff
@@ -558,6 +565,107 @@ static int UnlockDrbgState(void)
 
 #endif /* !HAVE_SELFTEST && (!HAVE_FIPS || FIPS v7+) */
 
+#ifdef WC_RNG_LOCK_ATFORK
+/* The lock and its fork handlers live in wc_port.c, outside the FIPS module
+ * boundary.  Only the DRBG's reaction to a fork belongs in here. */
+static int RngLockInit(WC_RNG* rng)
```
douzzer: The `RngLock*` functions should probably be named `RngOpLock*` to avoid confusion with the extensive existing infrastructure around `WC_RNG.lock`, which is an orthogonal mechanism distinguished be being _not_ per-op / automatic.

wolfcrypt/src/wc_port.c
```diff
@@ -26,6 +26,9 @@
 #elif defined(__FreeBSD__)
     /* for __FreeBSD_version */
     #include <sys/param.h>
+#elif (defined(__CYGWIN__) || defined(__MSYS__)) && !defined(_GNU_SOURCE)
+    /* dladdr and Dl_info, for the RNG fork handler pin, hide behind it */
+    #define _GNU_SOURCE 1
```
douzzer, line 31: I think this is problematic to globally assert `_GNU_SOURCE` on these targets -- has portability implications that need extensive testing, given a long history of volatility and version-specific fidgetiness around language/runtime features on Windows.

wolfcrypt/src/wc_port.c
```diff
@@ -420,6 +430,234 @@ static WC_DECLARE_INIT_STATE(wolfcrypt_init_state);
 int aarch64_use_sb = 0;
 #endif
 
+#ifdef WC_RNG_LOCK_ATFORK
+#if !defined(RTLD_NOLOAD) || !defined(RTLD_NODELETE)
+    #error "WC_RNG_ATFORK needs RTLD_NOLOAD and RTLD_NODELETE"
+#endif
+/* The fork handlers can never be unregistered, so the image that holds them
+ * is pinned against dlclose() before they are registered. */
+WOLFSSL_LOCAL void wc_RngPinImage(void* fn)
```
douzzer: Is there something RNG-specific about these?  If not, they shouldn't have "Rng" in their names.

douzzer (changes requested):
Seems on the right track, but needs some tweaks, and the implementation is incomplete.

See `random.c`  in http://github.com/douzzer/wolfssl/tree/20260917-WC_RNG_FLAG_OP_MUTEX for a tentatively complete implementation of the op lock get/release cycles (follows this PR's code pattern for convenient adoption, each critical patch wrapped in `RngOpLockEnter()`/`RngOpLockExit()`).

wolfssl/wolfcrypt/random.h
kaleb-himes: random.h:61 (Gates): Adopted using WOLFSSL_KERNEL_MODE instead of _DEFAULTS to ensure all kernel builds are covered cleanly.

wolfssl/wolfcrypt/random.h
kaleb-himes: random.h:649 (Struct Clean-up): Consolidated the .mutex declaration while excluding atfork builds to avoid extra memory overhead.

wolfcrypt/src/random.c
kaleb-himes: Fantastic suggestion and I agree whole heartedly!

random.c:663 (Port Logic): Moved platform-specific thread-cancellation logic out of random.c into wc_port.c.

wolfcrypt/src/random.c
kaleb-himes: random.c:5542/5549 (Hardware RNG): Agreed; hardware devices are shared, not per-instance. Fixed.

wolfcrypt/src/wc_port.c
kaleb-himes: I tested on Cygwin 3.5.3, gcc 12.4.0, and agree with the concern, wanted to be sure:

Without _GNU_SOURCE: dladdr and Dl_info undeclared, hard errors
dlfcn.h:49 gates both behind #if __GNU_VISIBLE, newlib's own feature test
The pin is genuinely compiled there, and works at runtime
Deleting the block breaks the Cygwin build

| Test | Result |
|---|---|
| Negative control (undeclared function) | fails as required, so the probe discriminates |
| dladdr + Dl_info, no _GNU_SOURCE | hard error: unknown type name 'Dl_info', implicit declaration of 'dladdr' |
| Same with _GNU_SOURCE | compiles |
| RTLD_NOLOAD / RTLD_NODELETE | both present |
| Runtime dladdr=1, and the RTLD_NOLOAD/RTLD_NODELETE | pin returns a handle |

And the header itself is the documentation:

/usr/include/dlfcn.h:49:  #if __GNU_VISIBLE
                    :50:  typedef struct Dl_info Dl_info;
                    :60:  extern int dladdr (const void *addr, Dl_info *info);
                    
Good to be sure, thank you for thinking about this, I hadn't considered it.

wolfcrypt/src/wc_port.c
kaleb-himes: Great thought! We may want to use these elsewhere and they are absolutely not RNG specific.

wc_port.c:439 (Function Name): Renamed to wc_PinImage() since the logic isn't RNG-specific.

wolfcrypt/src/random.c
kaleb-himes: Good catch, RngLock* sitting next to WC_RNG.lock was ambiguous. Renamed to RngAutoLockEnter/Exit/Init/Free.

I landed on Auto rather than Op, and your PR's framing is what pointed me there. Your PR described WC_RNG.lock as "not per-op / automatic", and of those two, automatic seemed like the cleaner separator, since a caller bracketing a single generate with wc_RNG_lock_get()/put() is arguably per-op as well. The part that really distinguishes this one is that the caller never requests it.

It also lines up with the rest of the feature, which already uses that vocabulary in 6 names across 45 places (WC_RNG_HAVE_AUTO_LOCK, rng->autoLock, and --enable-rng-lock resolving to -DWC_RNG_NO_AUTO_LOCK), so git grep autoLock turns up the implementation alongside the gate and the struct member.

A side benefit is that it leaves RngOpLock* free for the op mutex on your branch, which is opt-in rather than automatic.

kaleb-himes:
Thanks @douzzer for the review, and for putting up [20260917-WC_RNG_FLAG_OP_MUTEX](https://github.com/douzzer/wolfssl/tree/20260917-WC_RNG_FLAG_OP_MUTEX). Having a concrete implementation to build and run made it easy to evaluate.

I built it and worked through it. Three things I would want to resolve (this PR already does):

1. A lock release that can be skipped

wc_RNG_DRBG_NextSeedGenerate_local() takes the root lock at random.c:4875. On the non-error path the only release is at random.c:4981, inside if (n > 0). The complete-but-unpublished branch at random.c:4955 sets n = 0, so that block is skipped and the function returns with the lock still held; the next operation on that instance then waits on it.

Your suite passes, so it takes a specific sequence to reach: bank once with no root so the health test runs and its allocation fails (returns NOT_READY_E), then bank once with a root. Same binary both runs, WC_RNG_INIT_FLAG_USE_OP_MUTEX the only variable:

op mutex off: completes
op mutex on : hangs in wc_LockMutex

Related, wc_FreeRng() at random.c:5954 calls (void)wc_FreeMutex(&rng->mutex). pthread_mutex_destroy returns EBUSY on a held mutex, so the cast discards it and the caller still sees 0. The release just above it is guarded on WC_RNG_FLAG_FULL_MUTEX, which an op-mutex instance never matches.

2. Public flag bit values

The change renumbers five values across both flag spaces, so an application compiled against master passing WC_RNG_INIT_FLAG_RECOVER_AND_PROMOTE_FROM_NEXT_SEED would receive WC_RNG_INIT_FLAG_USE_OP_MUTEX instead, with a 0 return. 1U<<4 looks free in both spaces.

Separately, under WC_RNG_USE_OP_MUTEX_BY_DEFAULT the flag is OR'ed in at random.c:2698, above the exclusion check at 2702, so callers passing WC_RNG_INIT_FLAG_USE_FULL_MUTEX get BAD_FUNC_ARG in that build.

3. The rng_bank.c guards

Your guard fix looks right to me; those #ifdefs on the flag macros are always true as written. It is unrelated to RNG locking though, and this PR does not otherwise touch that file. Would you be open to landing it as its own PR? It would go in faster without waiting on this one.

wolfcrypt/src/random.c
kaleb-himes: Agreed and fixed!

wolfcrypt/src/wc_port.c
```diff
@@ -420,6 +430,218 @@ static WC_DECLARE_INIT_STATE(wolfcrypt_init_state);
...
+static pthread_once_t forkOnce = PTHREAD_ONCE_INIT;
+static int forkOnceRet = 0;
+
+/* sem_wait() is a cancellation point; a cancel here would strand the lock. */
+static int ForkSemWait(sem_t* s)
+{
+    int ret = 0;
+    int old = PTHREAD_CANCEL_ENABLE;
+    (void)pthread_setcancelstate(PTHREAD_CANCEL_DISABLE, &old);
+    while (sem_wait(s) != 0) {
```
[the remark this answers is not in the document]
kaleb-himes: Adjudicated: A fork() called from a signal handler can hang the caller, if the signal caught a thread part-way through taking or releasing a lock; POSIX does not allow fork() from a signal handler with handlers like these anyway. Values stay safe either way, with or without the handlers: the child reseeds on its pid change.

wolfcrypt/src/random.c
```diff
@@ -1150,6 +1258,12 @@ int wc_RNG_DRBG_Reseed_Nonce(WC_RNG* rng, const byte* seed, word32 seedSz,
     }
 #endif /* WC_RNG_HAVE_LOCK */
 
+    /* The internal reseed paths run with the lock already held, so it is
+     * taken here, at the public entry, and not in Hash_DRBG_Reseed(). */
+    ret = RngLockEnter(rng);
```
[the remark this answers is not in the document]
kaleb-himes: Fixed:
- Both entry points covered, wc_RNG_DRBG_Reseed() and wc_RNG_DRBG_Reseed_Now(), run as two passes of one test
- The window is forced rather than waited for: a holder thread takes the instance lock, signals over a pipe that it holds it, then condemns the instance one fifth of the way into the hold, so the reseed always arrives while the instance is healthy and completes against a condemned one
- The reseed must refuse with RNG_FAILURE_E; returning 0 fails the test
- negative control: against the tree before the fix it returned success on a condemned instance 3/3, and refuses 3/3 after
- lives in the suite as rng_reseed_status_test, so it runs on every execution rather than on demand

wolfcrypt/src/wc_port.c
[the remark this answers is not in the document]
kaleb-himes: Just a little more info on how I tested. Thank you Fenrir. Only three things precede the lock, and none of them is the racy read described: two NULL parameter checks, and rng_lock_required_check(), which reads only rng->lock through WOLFSSL_ATOMIC_LOAD. That is an atomic access, so it does not meet C17 §5.1.2.4's "conflicting non-atomic accesses" condition, and we separately proved with a negative control that it takes no lock. The comment in the code states the reasoning: "These checks read state a generate writes under the lock, so the lock comes first."

Commit b6d9c61d6, with a regression test (rng_reseed_status_test) that returned success on a condemned instance 3/3 before the fix and refuses 3/3 after.

wolfcrypt/src/wc_port.c
```diff
@@ -420,6 +430,321 @@ static WC_DECLARE_INIT_STATE(wolfcrypt_init_state);
...
+        if (forkMine[i] == lock) {
+            return 1;
+        }
+    }
+    return 0;
+}
+#else
+#define ForkMineAdd(lock)  WC_DO_NOTHING
+#define ForkMineDrop(lock) WC_DO_NOTHING
+#define ForkMineHeld(lock) 0
```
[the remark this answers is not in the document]
kaleb-himes: Fixed, thanks Fenrir.

tests/api/test_ossl_rand.c
```diff
@@ -213,35 +216,46 @@ int test_wolfSSL_RAND_bytes(void)
...
         byte childrand[8] = {0};
         int waitstatus = 0;
+        int reaped;
 
         close(pipefds[1]);
         ExpectIntEQ(RAND_bytes(randbuf, sizeof(randbuf)), 1);
         ExpectIntEQ(read(pipefds[0], childrand, sizeof(childrand)),
             sizeof(childrand));
-    #ifdef WOLFSSL_NO_GETPID
+    #if defined(WOLFSSL_NO_GETPID) && !defined(WC_RNG_LOCK_ATFORK)
```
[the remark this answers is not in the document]
kaleb-himes: Fixed.

wolfcrypt/src/random.c
```diff
@@ -5266,6 +5452,26 @@ int wc_RNG_DRBG_NextSeedNow_Nonce(WC_RNG* rng, const byte* nonce,
...
+    ret = RngAutoLockEnter(rng);
+    if (ret != 0)
+        return ret;
+    ret = wc_RNG_DRBG_NextSeedNow_Nonce_local(rng, nonce, nonceSz);
+    RngAutoLockExit(rng);
+    return ret;
+}
+
+/* For the generate path, which holds the lock already. */
+static WARN_UNUSED_RESULT int wc_RNG_DRBG_NextSeedNow_local(WC_RNG* rng) {
```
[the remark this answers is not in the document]
kaleb-himes: Fixed.

douzzer (approved):
tested locally with
```
wolfssl-multi-test.sh ...
'all-rng-.*'
```
with new targeted tests
```
all-rng-op-mutex-default-on-cryptonly
all-rng-op-mutex-default-off-cryptonly
all-rng-no-op-mutex-cryptonly
all-rng-only-op-mutex-default-on-cryptonly
all-rng-only-op-mutex-default-off-cryptonly
```
alongside existing disable-one and enable-one scenarios for all existing RNG-extras features.

JacobBarthelmeh (approved):

PR #11216, opened 2026-08-20
random: add WC_RNG_SEED_DEVICE to seed from a nominated device

dgarske:
Adds `WC_RNG_SEED_DEVICE`, an optional macro naming a device to seed the DRBG from before the default sources, for platforms that expose a hardware RNG such as `/dev/hwrng` (the NXP i.MX95 EdgeLock Secure Enclave, among others). Its value is a quoted string, set either by autotools `--with-rng-seed-device=PATH`, by CMake `-DWOLFSSL_RNG_SEED_DEVICE=PATH`, or directly as `#define WC_RNG_SEED_DEVICE "/dev/hwrng"` in `user_settings.h`; the two build systems take an unquoted path and add the quoting for you. The device attempt is self-contained and strictly additive: any failure to open or fill it falls through to the untouched `getrandom()` -> `/dev/urandom` -> `/dev/random` chain, so a default build is unchanged. Also fixes two pre-existing bugs in `wc_GenerateSeed()` (happy to split these out): `WOLFSSL_KEEP_RNG_SEED_FD_OPEN` with `NO_DEV_URANDOM` does not compile on master due to a dangling `else`, and the seed read loop spins forever if `read()` returns 0. Two new `os-check-linux` CI configs cover the device and fallback paths.


dgarske:
Jenkins retest this please

wolfcrypt/src/random.c
```diff
@@ -5933,6 +5941,12 @@ int wc_GenerateSeed(OS_Seed* os, byte* output, word32 sz)
     int wc_GenerateSeed(OS_Seed* os, byte* output, word32 sz)
     {
         int ret = 0;
+    #ifdef WC_RNG_SEED_DEVICE
```
philljj: should these 4 vars be define guarded the same as the block below? or else you can get unused var errors:
```
./configure --with-rng-seed-device=/dev/urandom  \
  CFLAGS="-DHAVE_ENTROPY_MEMUSE -DENTROPY_MEMUSE_FORCE_FAILURE" \
  && make
...
wolfcrypt/src/random.c: In function 'wc_GenerateSeed':
wolfcrypt/src/random.c:5948:16: error: unused variable 'devLen' [-Werror=unused-variable]
 5948 |         int    devLen;
      |                ^~~~~~
wolfcrypt/src/random.c:5947:16: error: unused variable 'devFd' [-Werror=unused-variable]
 5947 |         int    devFd;
      |                ^~~~~
wolfcrypt/src/random.c:5946:16: error: unused variable 'devSz' [-Werror=unused-variable]
 5946 |         word32 devSz;
      |                ^~~~~
wolfcrypt/src/random.c:5945:16: error: unused variable 'devOut' [-Werror=unused-variable]
 5945 |         byte*  devOut;
```

Feel free to disregard if this is too contrived of a configure scenario.

philljj (changes requested):

philljj:
Retest this please.

(PRB passed, but hung & aborted when reporting success)

philljj (approved):

PR #11219, opened 2026-08-20
CI: bound apt calls with timeouts and install ccache from staged bundle

julek-wolfssl:
- `ccache-setup` ran a raw `apt-get update` even when the ghcr `.deb` bundle
  had already installed cleanly, so jobs still hit the mirror. On 2026-08-19
  this stalled on archive.ubuntu.com for 9-43 min and took out 11 jobs.
- Install ccache with `--no-download` from the staged bundle in
  `/var/cache/apt/archives` first (already staged by `install-apt-deps`), and
  only fall back to the mirror if that fails.
- Add ccache to the 22.04-minimal and linuxkm bundles, where it was missing.
- Add Acquire timeouts plus `timeout` to every raw apt call in
  `install-apt-deps`'s fallback and the linuxkm producer job, matching what
  `ci-deps-image` already does, so a wedged mirror fails fast instead of
  hanging with no retry.
- Recover from a dpkg interrupted by a kill, and bound the bundle pull.

### DIFF
--- a/.github/actions/ccache-setup/action.yml
+++ b/.github/actions/ccache-setup/action.yml
@@ -42,9 +42,37 @@ runs:
         if command -v ccache >/dev/null 2>&1; then
           echo "ccache already installed: $(ccache --version | head -1)"
         elif [ "${{ runner.os }}" = "Linux" ]; then
-          sudo apt-get update -q
-          sudo DEBIAN_FRONTEND=noninteractive apt-get install -y \
-            --no-install-recommends ccache
+          export DEBIAN_FRONTEND=noninteractive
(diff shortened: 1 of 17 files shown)


julek-wolfssl:
retest this please

philljj (changes requested):
skoll feedback shared offline

julek-wolfssl:
feedback addressed

julek-wolfssl:
retest this please

julek-wolfssl:
retest this please

julek-wolfssl:
retest this please

padelsbach (approved):

philljj (approved):

PR #11220, opened 2026-08-20
Fix ASN guards for RSA key DER export

LinuxJedi:
WOLFSSL_SE050 enables WOLFSSL_KEY_TO_DER, but the original ASN implementation only compiled SetASNIntMP and SetBitString when unrelated features such as ECC or key generation were enabled. An ECC-free SE050 build therefore compiled the RSA DER callers without these helpers and failed at link time.

Include WOLFSSL_KEY_TO_DER in both helper guards, while retaining the NO_CERTS constraint for the public-key BIT STRING encoder. This allows SE050 RSA builds to disable HAVE_ECC without unresolved ASN symbols.

ZD 22321

### DIFF
--- a/.github/workflows/disable-pk-algs.yml
+++ b/.github/workflows/disable-pk-algs.yml
@@ -76,7 +76,8 @@ jobs:
            "configs": [
           {"name": "rsa-dh", "minutes": 1.2,
            "configure": ["--enable-rsa", "--enable-dh"]},
-          {"name": "rsa-key-to-der-original-asn", "minutes": 1.2,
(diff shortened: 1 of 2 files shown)


philljj (changes requested):
Should there be a workflow added to test this build scenario? Or is that covered in the simulators repo?

Otherwise this could break again.

LinuxJedi:
> Should there be a workflow added to test this build scenario? Or is that covered in the simulators repo?
> 
> Otherwise this could break again.

Good point, I'll add a build scenario for this in a bit.

philljj (approved):

PR #11226, opened 2026-08-20
Intel x64 ASM: intrinsic output

SparkiDev:
# Description

Add another implementation of the assembly code for Intel x64 that uses intrinsics and C code.

# Testing

Regression tested with intrinsic implementation.

Code generated with PR:
https://github.com/wolfSSL/scripts/pull/661


SparkiDev:
Jenkins: retest this please

PR #11227, opened 2026-08-20
CI: Address occasional ccache install hangs

padelsbach:
# Description

# Description

CI can hang until the job timeout when the apt mirror stalls.

- add ccache to ubuntu-22.04-minimal (matches 24.04) and install it from the
  bundle rather than apt
- bound the apt fallback in `install-apt-deps` with per-command timeouts and a
  retry count -- affects all 51 workflows using it
- fips-dev-no-post opts out (`ghcr-debs-tag: none`): it pulls the linuxkm
  bundle, so -minimal would be a second, larger download just for ccache

After merge, run `ci-deps-image` via workflow_dispatch to republish the 22.04
bundle; until then aesgcm-siv and multi-arch fall back to apt (no breakage).

# Testing

CI run

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/.github/actions/ccache-setup/action.yml
+++ b/.github/actions/ccache-setup/action.yml
@@ -1,12 +1,15 @@
 name: 'Set up ccache'
 description: >
-  Install ccache (Ubuntu via apt, macOS via brew), restore the ccache
-  directory from a previous run, and prepend the ccache compiler-symlink
-  dir to PATH. Subsequent gcc/cc/g++/c++/clang invocations are
-  transparently intercepted by ccache, so no other workflow step needs to
-  change. On scheduled (cron) runs the cache is reseeded from clean
-  compiles (CCACHE_RECACHE) instead of only being updated incrementally,
-  so it can't drift indefinitely.
+  Install ccache (Ubuntu from the ghcr bundle or apt, macOS via brew),
+  restore the ccache directory from a previous run, and prepend the
+  ccache compiler-symlink dir to PATH. Subsequent gcc/cc/g++/c++/clang
+  invocations are transparently intercepted by ccache, so no other
+  workflow step needs to change. The install is best-effort: if it fails
+  the job builds without ccache, so callers must tolerate a missing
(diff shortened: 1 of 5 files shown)


padelsbach:
jenkins retest this please

.github/actions/ccache-setup/action.yml
```diff
@@ -36,26 +46,73 @@ inputs:
...
+      if: steps.check.outputs.present != 'true' && runner.os == 'Linux'
+      uses: ./.github/actions/install-apt-deps
+      with:
+        packages: ccache
+        no-install-recommends: 'true'
+        ghcr-debs-tag: ${{ steps.check.outputs.tag }}
+        apt-budget: '120'
+        apt-update-timeout: '120'
+        apt-install-timeout: '120'
+        optional: 'true'
```
philljj, line 95: some workflows have hardcoded ccache compile commands and will fail without it.

e.g. .github/workflows/multi-compiler.yml

philljj (changes requested):

.github/actions/ccache-setup/action.yml
```diff
@@ -36,26 +46,73 @@ inputs:
...
+      id: have
+      shell: bash
+      run: |
+        if ccache --version >/dev/null 2>&1; then
+          echo "ccache=true" >> "$GITHUB_OUTPUT"
         else
-          echo "::error::ccache install not supported on ${{ runner.os }}"
-          exit 1
+          echo "ccache=false" >> "$GITHUB_OUTPUT"
+          echo "::warning::ccache unavailable; building without it"
```
philljj, line 112: I guess this could aggressively speed up workflows where it was optional but blocked on installing ccache

and the workflows where it was required would just error out later (but instantly) when ccache isn't there?

Maybe that would be better than hanging a lot...

padelsbach:
Superseded by https://github.com/wolfSSL/wolfssl/pull/11219. closing

PR #11232, opened 2026-08-21
Guard wolfSSL_CTX_set_alpn_protos with HAVE_ALPN

holtrop-wolfssl:
# Description

This was found while investigating F-10717.

This does not change the default build options which is the main thing that F-10717 is identifying. Rather, a secondary finding related was that the `wolfSSL_CTX_set_alpn_protos` function was not guarded by `HAVE_ALPN`. I moved the guard in `ssl_api_ext.c` up so that `wolfSSL_CTX_set_alpn_protos` is guarded the same as `wolfSSL_set_alpn_protos` below it.

# Testing

How did you test?

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/src/ssl_api_ext.c
+++ b/src/ssl_api_ext.c
@@ -2659,6 +2659,9 @@ int wolfSSL_set1_curves_list(WOLFSSL* ssl, const char* names)
 
 #ifdef OPENSSL_EXTRA
 
+#ifdef HAVE_ALPN
+#ifndef NO_BIO
+
 /* Set the ALPN protocol list, in wire format, on the context.
  *
  * @param [in] ctx    SSL/TLS context object.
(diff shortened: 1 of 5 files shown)


holtrop-wolfssl:
retest this please (ERROR: Build 'PRB-generic-config-parser' failed with result: FAILURE)

src/ssl_api_ext.c
```diff
@@ -2659,6 +2659,9 @@ int wolfSSL_set1_curves_list(WOLFSSL* ssl, const char* names)
 
 #ifdef OPENSSL_EXTRA
 
+#ifdef HAVE_ALPN
+#ifndef NO_BIO
```
SparkiDev, line 2663: wolfSSL_CTX_set_alpn_protos() doesn't need !NO_BIO.
Is NO_BIO required in order to use ALPN?

SparkiDev (changes requested):

src/ssl_api_ext.c
holtrop-wolfssl: I moved these two guards up from below to match the preexisting build options used to guard the call to `wolfSSL_CTX_set_alpn_protos` from src/internal.c. But, I think you're right, that NO_BIO should not be required to use ALPN, so perhaps it makes sense to just remove the NO_BIO guard there and here.

PR #11246, opened 2026-08-24
IoT-Safe: serialize APDU transactions with a port mutex

danielinux:
# Description

All IoT-Safe operations share the file-scope static command/response buffers on a single CSIM channel. Concurrent operations (e.g. a random generation and an ECDSA sign callback from different TLS sessions) could overwrite each other's buffers and consume the wrong modem reply, corrupting cryptographic output or leaving the applet in a persistent error state.

Add a port-level mutex (active in multi-threaded builds, no-op when SINGLE_THREADED) held across each APDU transaction. The seven single-transaction operations get thin serialized wrappers. GetRandom locks after its lazy-init check, and the ECDH callback locks its APDU branch, calling the locked impls directly since it already holds the mutex (the software fallback never takes the lock). The mutex is created in iotsafe_init(); document the single-threaded init contract in the doxygen of wolfSSL_CTX_iotsafe_enable() and the CSIM callback setters.

Fixes F-10045.

# Testing

Tested via existing iotsafe demos build

# Checklist

 - [x] updated/added doxygen
 - [x] Updated manual and documentation

### DIFF
--- a/doc/dox_comments/header_files/iotsafe.h
+++ b/doc/dox_comments/header_files/iotsafe.h
@@ -6,11 +6,14 @@
     \return 0 on success
     \return WC_HW_E on hardware error
 
-    This function loads the IoT-Safe applet and must be called from a single
-    thread, before the context is used concurrently from multiple threads.
-    Once initialized, individual APDU transactions are serialized internally
-    in multi-threaded builds, so concurrent TLS sessions can safely share the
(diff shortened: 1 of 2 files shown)


danielinux:
jenkins retest this please

philljj (approved):

PR #11267, opened 2026-08-25
Add Silicon Labs EFR32xG25 Secure Element crypto callback port

dgarske:
## Summary

Adds **`WOLFSSL_SILABS_CRYPTOCB`**, a **crypto callback** port for the EFR32 Series 2 **Secure Element**. Unlike the existing direct-hook `WOLFSSL_SILABS_SE_ACCEL` port, operations route by **`devId`** and anything the SE cannot do **falls back to software** instead of failing. Both ports share one set of SE Manager helpers.

## New crypto callback support

Two algorithms gain crypto callback dispatch in the **core framework**. Neither had it before, and both are available to any callback device, not just this one.

- **ChaCha20-Poly1305.** New `wc_Chacha_SetKey_ex()` and `wc_ChaCha20Poly1305_Init_ex()` bind a **`devId`** to the cipher object, and `wc_ChaCha20Poly1305_Encrypt_ex()` / `_Decrypt_ex()` dispatch on it. That is the API the **TLS record layer** uses, so a connection can offload its ChaCha records. The legacy one-shot entry points carry no key object and stay in software.
- **PBKDF2.** New `WC_KDF_TYPE_PBKDF2`, dispatched from `wc_PBKDF2_ex()` on the caller's **`devId`**.

Both follow the existing `_ex(..., void* heap, int devId)` idiom. No public signature changes.

## Features

- **AES** ECB/CBC/CTR/GCM/CCM, **AES-CMAC**, **ChaCha20-Poly1305**
- **SHA-1/224/256**, **TRNG**, **HKDF**, **PBKDF2**
- **ECDSA** sign/verify, **ECDH**, **key generation**
- **Secure Vault High keys**: **wrapped keys** and **built-in slots** - the SE uses the key, the application never sees the material
- **Per-engine selection**: offload a subset, the rest stays in software
- **Host compile test** (`--enable-silabs-cryptocb`): builds and passes `make check` with **no Simplicity SDK installed**

## Testing

**EFR32FG25B222F1920IM56**, Secure Vault High, SE firmware 2.2.0: `wolfcrypt_test` **full PASS**. Because `WOLFSSL_SILABS_CRYPTOCB` sets `WC_USE_DEVID`, every algorithm the port claims runs on the SE, so the pass is direct evidence against the known-answer vectors.

| | HW vs SW |
|---|---|
| AES-256-GCM | **9.4x** |
| AES-CCM | **5.3x** |
| AES-256-ECB / CTR | **4.3x** |
| AES-256-CBC | **4.2x** |
| AES-256-CMAC | **4.1x** |
| ECDSA verify | **3.9x** |
| SHA-256 | **3.0x** |
| ECDH agree | **2.7x** |
| ECDSA sign | **2.6x** |

Full per-algorithm table in `wolfcrypt/src/port/silabs/README.md`. Both columns come from a single benchmark run on one part: the stock benchmark measures each algorithm twice and labels the rows **HW** and **SW**.

**TLS 1.3** handshake and record layer, both peers on the device, pass for **`TLS13-AES128-GCM-SHA256`**, **`TLS13-AES256-GCM-SHA384`** and **`TLS13-CHACHA20-POLY1305-SHA256`**. That last one is what exercises the new ChaCha dispatch through the record layer.

Host: `--enable-all --enable-cryptocb` and `--enable-silabs-cryptocb` both `make check` clean, plus Secure Vault Mid, single-engine, async, and a staged `make install` consumer compile. `check-source-text` clean.

## Notes

- No **RSA** engine - Series 2 has no RSA hardware. No **HMAC** engine - wolfCrypt HMAC gives its inner and outer hash contexts the `Hmac`'s own `devId`, so every block already runs on the SE through the hash engine.
- **SHA-384/512** are always left to software. `wc_ShaXXXGetHash()` copies the context and finalizes the copy, which the SE context does not survive.
- **AES-CTR** offloads only whole blocks on a block boundary; the SE and wolfCrypt track the keystream remainder differently. Under **`WOLFSSL_ARMASM`** CTR is left entirely to software, since the Thumb2 remainder and the SE's do not line up.
- Binding an SE-resident key onto an `ecc_key` **scrubs any software scalar** it held. The object stays typed `ECC_PRIVATEKEY`, so without the scrub a caller could export the secret the object held before the bind.
- **Session tickets are not offloaded.** The AES-GCM ticket variants deliberately use `wc_AesInit(aes, NULL, INVALID_DEVID)`, so threading a `devId` into the ChaCha variant alone would be inconsistent.
- Fixes latent bugs in the shared SiLabs code, including **`pubkey.z` never set after SE key generation** and CCM reporting `AES_GCM_AUTH_E`.
- The worked example lives in **wolfssl-examples**: https://github.com/wolfSSL/wolfssl-examples/pull/627


dgarske:

dgarske:
retest this please

wolfcrypt/src/port/silabs/silabs_cb_pk.c
```diff
@@ -0,0 +1,299 @@
...
+    if (key->silabsKeySet) {
+        return 0;
+    }
+
+    /* silabs_ecc_import() rewrites key->type from the pub/priv flags. That is
+     * right when wolfCrypt is genuinely importing a key, but here it only
+     * refreshes the SE buffer before an operation: a verify needs the public
+     * part alone and must not downgrade a key that holds a private scalar,
+     * which would break a later private-key export. */
+    savedType = key->type;
```
aidangarske, line 139: **SiLabs ECC buffers inherit another backend's smaller maximum**

The new callback passes every supported curve's `dp->size` to `silabs_ecc_import()`, but `ecc_key.key_raw` and the callback's temporary signature buffers use the single `ECC_MAX_CRYPTO_HW_SIZE` constant. That constant is selected by an `#elif` chain, so a simultaneously enabled backend such as ATECC can select 32 bytes while the SiLabs Vault-High callback still accepts P-384 and P-521. `silabs_ecc_import()` then writes X||Y||D using 48- or 66-byte strides into `3 * 32` bytes. The PR exposes this memory corruption by making SiLabs a devId-routed callback that can coexist with other registered hardware while retaining the old one-backend sizing scheme.

**Fix:** Define a SiLabs-specific maximum derived from the SiLabs device capabilities and use it for `key_raw` and every SiLabs temporary buffer, or calculate the maximum across all enabled backends. Add a combined SiLabs-plus-ATECC configuration test.

wolfcrypt/src/port/silabs/silabs_cb_pk.c
```diff
@@ -0,0 +1,299 @@
...
+                info->pk.eccsign.inlen)) {
+            return WC_NO_ERR_TRACE(CRYPTOCB_UNAVAILABLE);
+        }
+    #ifndef WC_ALLOW_ECC_ZERO_HASH
+        if (silabs_pk_hash_is_zero(info->pk.eccsign.in,
+                info->pk.eccsign.inlen)) {
+            return ECC_BAD_ARG_E;
+        }
+    #endif
+        /* silabs_ecc_sign_hash() substitutes the SE vault attestation key when
```
aidangarske: **Resident ECC keys cannot be used for signing**

The new wrapped- and built-in-key binders deliberately leave the software private scalar empty, but the ECDSA callback rejects every key with an empty scalar before considering `silabsKeySet`. It therefore returns `CRYPTOCB_UNAVAILABLE`, after which software signing also fails because no private scalar exists. Simply removing this check is unsafe: the shared `silabs_ecc_sign_hash()` helper redirects any empty-scalar key to `private_device_key`, which would sign with the attestation key instead of the descriptor selected by `wc_SilabsSe_EccUseWrappedKey()` or `wc_SilabsSe_EccUseBuiltInKey()`.

**Fix:** Treat `silabsKeySet` as a valid private-key source and update `silabs_ecc_sign_hash()` to use the bound descriptor in that case. Retain the attestation-key substitution only for the legacy direct-port behavior that explicitly requires it.

wolfcrypt/src/port/silabs/silabs_key.c
```diff
@@ -0,0 +1,469 @@
...
+    if (wrappedSz != need) {
+        return BAD_LENGTH_E;
+    }
+
+    ret = silabs_key_aes_type(keyBits, &type);
+    if (ret != 0) {
+        return ret;
+    }
+
+    aes->ctx.cmd_ctx = cc;
```
aidangarske, line 233: **AES key rebinding can silently select stale key material**

The new binders set `ctx.keySet` but leave any existing software schedule, `devKey`, and `keyInstalled` state intact. Conversely, a later `wc_AesSetKey()` installs a new plaintext key without clearing `ctx.keySet`. Binding resident key B after plaintext key A therefore makes supported modes use B while a deliberately declined mode falls back with stale key A. Setting plaintext key A after binding B has the opposite result: `silabs_cipher_setkey()` continues trusting B because `ctx.keySet` remains set. Both sequences can encrypt with a different key than the caller most recently selected.

**Fix:** Make plaintext and resident key state mutually exclusive. The binders should invalidate and scrub software key state, while every normal AES set-key path should clear the resident descriptor and `ctx.keySet`. Resident-key operations that cannot be offloaded must fail rather than use stale software state.

wolfcrypt/src/port/silabs/silabs_key.c
```diff
@@ -0,0 +1,469 @@
...
+    }
+    ret = wc_SilabsSe_EccGetWrappedKeySize(curveId, &need);
+    if (ret != 0) {
+        return ret;
+    }
+    if (wrappedSz != need) {
+        return BAD_LENGTH_E;
+    }
+
+    /* Give wolfCrypt the curve so sizes and signature encoding are right,
```
aidangarske, line 426: **ECC key rebinding can use a different key than requested**

The resident-key binders install a descriptor and set `silabsKeySet`, but do not clear an existing software private scalar or public point. Normal ECC import paths likewise do not clear the new flag. Binding key B over software key A can therefore make an operation declined by the callback fall back and use stale A. Importing A over an object previously bound to B leaves `silabsKeySet` set, so `silabs_pk_load()` continues using descriptor B even though the public `ecc_key` now represents A.

**Fix:** Clear and zero incompatible software key components when binding a resident key, and clear the resident descriptor and `silabsKeySet` from all normal import and key-generation paths. Add same-object rebinding tests in both directions.

wolfcrypt/src/port/silabs/silabs_cb_cmac.c
```diff
@@ -0,0 +1,194 @@
...
+    cmac = info->cmac.cmac;
+    if (cmac == NULL) {
+        return WC_NO_ERR_TRACE(CRYPTOCB_UNAVAILABLE);
+    }
+    /* The SE does AES-CMAC only. */
+    if (info->cmac.type != WC_CMAC_AES) {
+        return WC_NO_ERR_TRACE(CRYPTOCB_UNAVAILABLE);
+    }
+
+    if (info->cmac.key != NULL) {
```
aidangarske, line 232: **CMAC callback violates output-length and pointer contracts**

`wc_CmacFinalNoFree()` accepts truncated tags from 4 through 16 bytes, but the new callback rejects every output smaller than 16 and always writes back a size of 16. Since earlier updates were handled only by the SE, finalization cannot safely fall back to software. The one-shot callback also runs before `wc_AesCmacGenerate_ex()` performs its pointer/length validation: `in == NULL && inSz > 0` is silently treated as an empty message, while `out == NULL` with a positive requested size can return success without a tag and leave the allocated device context live.

**Fix:** Validate pointer/length pairs before starting or updating the SE operation. Finalize into a local 16-byte buffer and copy the caller-requested 4-16 byte prefix, preserving `*outSz`. Ensure terminal argument errors release newly allocated one-shot state.

wolfcrypt/src/port/silabs/silabs_cb_kdf.c
```diff
@@ -0,0 +1,163 @@
...
+#if defined(HAVE_HKDF) && !defined(NO_HMAC)
+static int silabs_kdf_hkdf(wc_CryptoInfo* info)
+{
+    sl_se_command_context_t cmd = SL_SE_COMMAND_CONTEXT_INIT;
+    sl_se_key_descriptor_t inKey;
+    sl_se_key_descriptor_t outKey;
+    sl_status_t status;
+    int seHash;
+
+    if (info->kdf.hkdf.inKey == NULL || info->kdf.hkdf.out == NULL) {
```
aidangarske: **KDF callbacks reject valid zero-length inputs**

The HKDF callback rejects `inKey == NULL` even when `inKeySz == 0`, although the public HKDF implementation and its API tests explicitly accept that as zero-length input keying material. PBKDF2 similarly rejects `passwd == NULL, pLen == 0` and changes the core `kLen < 0` check to `kLen <= 0`; empty passwords and a zero-length result are accepted by the non-FIPS software API. Because callback dispatch precedes software derivation, these valid calls now return `BAD_FUNC_ARG` instead of falling back.

**Fix:** Match the public API's pointer/length validation exactly. If the SE cannot represent an otherwise valid empty input, return `CRYPTOCB_UNAVAILABLE` so software preserves the existing contract.

wolfcrypt/src/port/silabs/silabs_cb_hash.c
```diff
@@ -0,0 +1,202 @@
...
+#include <wolfssl/wolfcrypt/error-crypt.h>
+#include <wolfssl/wolfcrypt/hash.h>
+#include <wolfssl/wolfcrypt/port/silabs/silabs_cryptocb.h>
+#include <wolfssl/wolfcrypt/port/silabs/silabs_hash.h>
+
+/* Resolve a crypto callback hash request to the SE context embedded in the
+ * caller's wolfCrypt hash object, and to the digest length. Returns
+ * CRYPTOCB_UNAVAILABLE for anything the SE cannot do, so software runs.
+ *
+ * SHA-384 and SHA-512 are left to software unless
```
aidangarske: **Optional SHA-384/512 offload knowingly returns incorrect digests**

The PR exposes `WOLFSSL_SILABS_CRYPTOCB_SHA512_HW` and conditionally enables SHA-384/SHA-512 handling while the adjacent comment states that the public `wc_ShaXXXGetHash()` path returns a wrong digest. The SDK-v3 copy fixups in `wc_Sha512Copy()` and `wc_Sha384Copy()` also remain gated by the direct-port-only `WOLFSSL_SILABS_SHA512` and `WOLFSSL_SILABS_SHA384` macros, which the callback port does not define. An optional configuration that produces incorrect cryptographic output is not safe to ship.

**Fix:** Remove or reject the opt-in macro until context copying and `GetHash` are correct, or implement complete copy/finalization support for every supported SDK interface. Test `GetHash`, continued use of the original context, and SHA-384/SHA-512 variants.

wolfcrypt/src/random.c
```diff
@@ -4468,7 +4468,13 @@ int wc_GenerateSeed(OS_Seed* os, byte* output, word32 sz)
      */
     #define USE_TEST_GENSEED
 
-#elif defined(WOLFSSL_SILABS_SE_ACCEL)
+#elif defined(WOLFSSL_SILABS_SE_TYPES)
```
aidangarske: **Host compile-test configuration disables usable entropy**

`--enable-silabs-cryptocb` defines both `WOLFSSL_SILABS_HOST_TEST` and `WOLFSSL_SILABS_SE_TYPES`. The broadened seed branch therefore replaces the normal host entropy implementation with `silabs_GenerateRand()`. The host shim's `sl_se_get_random()` always returns `SL_STATUS_NOT_SUPPORTED`, and `silabs_GenerateRand()` converts that to `WC_HW_E`, so software RNG initialization fails instead of using the host OS source. This contradicts the new option's stated `make check` fallback behavior.

**Fix:** Exclude the host shim from the hardware seed branch so host builds continue to use their normal OS entropy implementation.

**Suggestion:**
```suggestion
#elif defined(WOLFSSL_SILABS_SE_TYPES) && \
    !defined(WOLFSSL_SILABS_HOST_TEST)
```

wolfcrypt/src/port/silabs/silabs_cb_cipher.c
```diff
@@ -0,0 +1,371 @@
...
+#ifndef NO_AES
+#if defined(HAVE_AES_ECB) || defined(WOLFSSL_AES_DIRECT)
+    case WC_CIPHER_AES_ECB:
+        if (info->cipher.aesecb.aes == NULL) {
+            return BAD_FUNC_ARG;
+        }
+        if ((info->cipher.aesecb.sz % WC_AES_BLOCK_SIZE) != 0) {
+            return BAD_LENGTH_E;
+        }
+        ret = silabs_cipher_setkey(info->cipher.aesecb.aes);
```
aidangarske, line 301: **AES helper error translation prevents software fallback**

The new callback routes ECB and CBC through shared direct-port helpers that collapse every non-success SE status to `WC_HW_E`; the GCM and CCM helpers similarly collapse statuses to authentication errors. Consequently `SL_STATUS_NOT_SUPPORTED` and `SL_STATUS_INVALID_PARAMETER` never reach `silabs_cb_status()` and can never become `CRYPTOCB_UNAVAILABLE`. A missing firmware command or unsupported request therefore hard-fails instead of taking the software fallback that distinguishes this new port from the direct port.

**Fix:** Add raw-status variants, as was done for hashing, or perform callback-specific SE calls and translate unsupported statuses with `silabs_cb_status()`. Preserve authentication-failure mappings separately from unsupported-command mappings.

wolfcrypt/src/port/silabs/silabs_cb_pk.c
```diff
@@ -0,0 +1,299 @@
...
+         * would silently discard that binding. Use
+         * wc_SilabsSe_EccGenerateWrappedKey() for that instead. */
+        if (info->pk.eckg.key->silabsKeySet) {
+            return BAD_FUNC_ARG;
+        }
+        /* Use the curve's own field size, not the size the caller asked for.
+         * wc_ecc_make_key_ex() rounds a requested size up to the next curve,
+         * so eckg.size can be smaller than dp->size (40 selects P-384); using
+         * it would read the SE's key back at the wrong stride. */
+        ret = silabs_ecc_make_key(info->pk.eckg.key,
```
aidangarske: **ECC helper error translation prevents software fallback**

The new ECC callback reuses direct-port key-generation, signing, verification, import, and ECDH helpers. Those helpers collapse unsupported SE statuses to `WC_HW_E` rather than returning the raw status or `CRYPTOCB_UNAVAILABLE`. Plaintext-key operations that could run in software therefore fail when an SE command is unavailable, defeating the callback port's advertised fallback behavior.

**Fix:** Provide callback-aware/raw-status ECC helpers and map unsupported commands through `silabs_cb_status()`, while retaining the existing direct-port wrapper contract.

wolfcrypt/src/port/silabs/silabs_cb_pk.c
```diff
@@ -0,0 +1,299 @@
...
+            info->pk.eckg.key->dp->size);
+        break;
+
+    case WC_PK_TYPE_ECDH:
```
aidangarske, line 202: **ECDH callback bypasses the public API's key validation**

`wc_ecc_shared_secret()` dispatches to the callback before checking that the private key has a private-key type, both domain parameters are valid, and the two curve IDs match. The new handler checks only the private curve before importing both objects and submitting them to the SE. Invalid or mismatched keys therefore reach hardware and return backend-dependent errors instead of the established `ECC_BAD_ARG_E` result.

**Fix:** Mirror the type, domain-parameter, and matching-curve checks performed by `wc_ecc_shared_secret()` before invoking the SE, or decline early so the core path performs them.

tests/api/test_async.c
```diff
@@ -698,10 +697,10 @@ int test_wc_CryptoCb_AsyncPollChachaUnimpl(void)
...
         registered = 1;
 
-    /* no crypto callback dispatch for ChaCha: runs in software, cb untouched */
+    /* offered to the device, declined, then completed in software */
     ExpectIntEQ(wc_ChaCha20Poly1305_Encrypt(key, iv,
         test_async_aad12(), 12, test_async_plain32(), 32, out, tag), 0);
-    ExpectIntEQ(hsm.invocations, 0);
+    ExpectIntEQ(hsm.invocations, 1);
     ExpectIntEQ(hsm.submits, 0);
     ExpectIntEQ(hsm.polls, 0);
```
aidangarske, line 716: **ChaCha async test assumes its callback owns slot zero**

The new one-shot dispatch always calls callback slot zero, but this test registers its `TestHsm` without checking whether an automatically registered device already occupies that slot. In configurations combining async callbacks with SiLabs, Versal ASU, or another auto-registered backend, ChaCha reaches that earlier device and `hsm.invocations` remains zero, making the new assertion fail. The modified `cryptocb_test()` already handles this exact situation, but this test does not.

**Fix:** Isolate the callback registry for this test or condition the invocation assertion on `wc_CryptoCb_GetDevIdAtIndex(0) == TEST_ASYNC_HSM_DEVID`, as the other updated test does.

wolfcrypt/src/cryptocb.c
```diff
@@ -2101,8 +2103,75 @@ int wc_CryptoCb_PqcSignatureCheckPrivKey(void* key, int type,
...
 
+#if defined(HAVE_CHACHA) && defined(HAVE_POLY1305)
+int wc_CryptoCb_Chacha20Poly1305Encrypt(const byte* inKey, const byte* inIV,
+    const byte* inAAD, word32 inAADSz, const byte* in, word32 inSz,
+    byte* out, byte* outAuthTag)
+{
+    int ret = WC_NO_ERR_TRACE(CRYPTOCB_UNAVAILABLE);
+    CryptoCb* dev;
+
+    /* No key object and so no devId on the public API: use the first
```
aidangarske, line 2115: **Legacy ChaCha API dispatches through an unrelated global device**

The existing public one-shot API has no device selection, yet the PR changes every `WOLF_CRYPTO_CB` build to send the operation and raw key to whichever callback happened to register first. Registration order now determines behavior, a capable device in a later slot is never considered, and an unrelated existing callback's error can replace the previously deterministic software result. Existing applications also have no per-call way to opt out.

**Fix:** Keep the legacy API on its existing software path and add an explicit `_ex` form carrying a `devId`, or require an explicit capability/opt-in macro before enabling slot-zero dispatch.

wolfssl/wolfcrypt/port/silabs/silabs_cryptocb.h
```diff
@@ -0,0 +1,137 @@
...
+#include <wolfssl/wolfcrypt/types.h>
+#include <wolfssl/wolfcrypt/cryptocb.h>
+#ifndef NO_AES
+    #include <wolfssl/wolfcrypt/aes.h>
+#endif
+#ifdef HAVE_ECC
+    #include <wolfssl/wolfcrypt/ecc.h>
+#endif
+
+/* Declare the wrapped-key API unless the application says the part has no
```
aidangarske: **Vault Mid headers advertise functions that are not linked**

The implementation compiles wrapped-key functions only when `_SILICON_LABS_SECURITY_FEATURE` identifies Vault High, but the public header declares them by default unless the application manually defines `WOLFSSL_SILABS_NO_VAULT_KEYS`. A normal Vault Mid build therefore compiles calls against public prototypes and then fails at link time because no definitions exist. Requiring users to duplicate a hardware capability already visible to the port makes the API configuration fragile.

**Fix:** Derive declarations from the same hardware-capability condition as the implementation, or always provide definitions that return `NOT_COMPILED_IN` on unsupported parts.

wolfssl/wolfcrypt/port/silabs/silabs_settings.h
```diff
@@ -0,0 +1,128 @@
...
+/* Device id for the SE crypto callback; set WOLFSSL_SILABS_DEVID (or
+ * WC_USE_DEVID) to any int but INVALID_DEVID (-2), an id not an address. */
+#ifndef WOLFSSL_SILABS_DEVID
+    #define WOLFSSL_SILABS_DEVID 0x5345 /* 'SE' */
+#endif
+
+/* Let the unmodified wolfcrypt test and benchmark route every operation
+ * through this device by giving their devId the SE value.
+ *
+ * Not under the host compile-test: there the SE Manager is a stub that does no
```
aidangarske, line 107: **Host gate does not execute the new SiLabs handlers**

The host configuration explicitly suppresses `WC_USE_DEVID`, and every shim crypto operation immediately returns unavailable. No added test directly invokes the SiLabs AES, hash, CMAC, ECC, KDF, or resident-key APIs. Generic callback tests exercise their own callback at another device ID, so the port's state transitions, argument contracts, and fallback translations remain compile-only. That leaves the resident-key, CMAC, and KDF regressions above undetected.

**Fix:** Make the shim behavior injectable and add direct handler tests for unsupported-status translation, truncated CMAC, empty KDF inputs, SHA context copying, and plaintext/resident key rebinding.

aidangarske (changes requested):
## Skoll Code Review

**Scan type:** `review`

**Findings:** 15 total — 15 posted, 0 skipped
**15 finding(s) posted as inline comments** (see file-level comments below)

### Posted findings
- [Critical] **SiLabs ECC buffers inherit another backend's smaller maximum** — `wolfcrypt/src/port/silabs/silabs_cb_pk.c:128-130`
- [High] **Resident ECC keys cannot be used for signing** — `wolfcrypt/src/port/silabs/silabs_cb_pk.c:209-216`
- [High] **AES key rebinding can silently select stale key material** — `wolfcrypt/src/port/silabs/silabs_key.c:226-241`
- [High] **ECC key rebinding can use a different key than requested** — `wolfcrypt/src/port/silabs/silabs_key.c:409-426`
- [High] **CMAC callback violates output-length and pointer contracts** — `wolfcrypt/src/port/silabs/silabs_cb_cmac.c:153-187`
- [High] **KDF callbacks reject valid zero-length inputs** — `wolfcrypt/src/port/silabs/silabs_cb_kdf.c:70-106`
- [High] **Optional SHA-384/512 offload knowingly returns incorrect digests** — `wolfcrypt/src/port/silabs/silabs_cb_hash.c:39-45`
- [High] **Host compile-test configuration disables usable entropy** — `wolfcrypt/src/random.c:4471-4482`
- [High] **AES helper error translation prevents software fallback** — `wolfcrypt/src/port/silabs/silabs_cb_cipher.c:218-238`
- [High] **ECC helper error translation prevents software fallback** — `wolfcrypt/src/port/silabs/silabs_cb_pk.c:167-168`
- [Medium] **ECDH callback bypasses the public API's key validation** — `wolfcrypt/src/port/silabs/silabs_cb_pk.c:171-187`
- [Medium] **ChaCha async test assumes its callback owns slot zero** — `tests/api/test_async.c:695-705`
- [Medium] **Legacy ChaCha API dispatches through an unrelated global device** — `wolfcrypt/src/cryptocb.c:2114-2116`
- [Medium] **Vault Mid headers advertise functions that are not linked** — `wolfssl/wolfcrypt/port/silabs/silabs_cryptocb.h:38-43`
- [Medium] **Host gate does not execute the new SiLabs handlers** — `wolfssl/wolfcrypt/port/silabs/silabs_settings.h:115-123`

wolfcrypt/src/port/silabs/silabs_cb_hash.c
```diff
@@ -0,0 +1,168 @@
...
+        return WC_NO_ERR_TRACE(CRYPTOCB_UNAVAILABLE);
+    }
+
+    return 0;
+}
+
+/* Start the SE context on first use. wolfCrypt has no crypto callback hook on
+ * wc_InitShaXXX, and that call zeroes the object, so "started" is clear on a
+ * freshly initialized context. */
+static int silabs_hash_start(wc_silabs_sha_t* ctx, int hashType)
```
aidangarske, line 95: **Fresh hash contexts can skip Secure Element initialization**

The PR adds `silabsCtx.started` to the SHA objects and uses it as the lazy-initialization sentinel, but the normal `wc_InitSha*_ex()` paths initialize individual software fields rather than clearing the new `silabsCtx`. A fresh stack or heap object can therefore contain a nonzero `started` byte, causing the callback to skip `wc_silabs_se_hash_init_status()` and pass an uninitialized SE context to update or final.

**Fix:** Initialize `silabsCtx` in every applicable SHA-1, SHA-224, SHA-256, SHA-384, and SHA-512 initialization path, or introduce a reliably initialized sentinel. Add tests that initialize hash objects whose storage was deliberately filled with nonzero bytes.

wolfcrypt/src/port/silabs/silabs_cb_cmac.c
```diff
@@ -0,0 +1,219 @@
...
+
+    if (cmac == NULL || key == NULL) {
+        return BAD_FUNC_ARG;
+    }
+    if (keySz != AES_128_KEY_SIZE && keySz != AES_192_KEY_SIZE &&
+        keySz != AES_256_KEY_SIZE) {
+        return WC_NO_ERR_TRACE(CRYPTOCB_UNAVAILABLE);
+    }
+
+    silabs_cmac_free(cmac);
```
aidangarske, line 88: **One-shot CMAC can free an uninitialized device context**

One-shot CMAC dispatch occurs before `wc_InitCmac_ex()` initializes the supplied `Cmac`. Existing callers such as the two-step CMAC KDF pass fresh, uninitialized stack or heap storage to `wc_AesCmacGenerate_ex()`. The new callback immediately reads `cmac->devCtx`, may pass its garbage value to `XFREE`, and then uses the likewise uninitialized `cmac->aes.heap` for allocation.

**Fix:** Do not inspect object-owned callback state until the `Cmac` has been initialized. Initialize the object before the first callback or use temporary one-shot state with an explicitly supplied heap. Add a regression test using nonzero fresh `Cmac` storage and the two-step CMAC KDF.

wolfcrypt/src/port/silabs/silabs_cb_hash.c
```diff
@@ -0,0 +1,168 @@
...
+    if (ret != 0) {
+        return ret;
+    }
+
+    ret = wolfSSL_CryptHwMutexLock();
+    if (ret != 0) {
+        return ret;
+    }
+
+    ret = silabs_hash_start(ctx, info->hash.type);
```
aidangarske, line 149: **Hash fallback loses data after the SE acquires the context**

The callback maps `SL_STATUS_NOT_SUPPORTED` and `SL_STATUS_INVALID_PARAMETER` from every update and final to `CRYPTOCB_UNAVAILABLE`. The core then falls back to software for that individual call. Once an earlier start or update succeeded in hardware, however, the software hash state has not consumed the preceding input, so resuming there produces an incorrect digest. The inverse transition is also possible: a declined update can advance software while `started` remains set, allowing a later update to resume the divergent SE state.

**Fix:** Only return `CRYPTOCB_UNAVAILABLE` before any SE state or input has been accepted. After hardware processing begins, return a hard error unless the implementation maintains a mirrored software context or can replay the complete message. Add fault-injection tests for failures after start and after one successful update.

wolfcrypt/src/port/silabs/silabs_cb_cmac.c
```diff
@@ -0,0 +1,219 @@
...
+    }
+
+    c = silabs_cmac_ctx(cmac);
+    if (c == NULL) {
+        /* No SE state: the key was set before this device took over, so let
+         * software finish the operation. */
+        return WC_NO_ERR_TRACE(CRYPTOCB_UNAVAILABLE);
+    }
+
+    if (info->cmac.in != NULL) {
```
aidangarske, line 250: **CMAC fallback is invalid after hardware initialization succeeds**

After the SE successfully handles CMAC start, `wc_InitCmac_ex()` returns without initializing the software AES/CMAC state. The new update and final paths can nevertheless translate a later SE `NOT_SUPPORTED` or `INVALID_PARAMETER` result into `CRYPTOCB_UNAVAILABLE`, causing the core to resume an uninitialized software operation. This can return misleading errors or compute a MAC without the prior key and message state.

**Fix:** Once CMAC start succeeds in hardware, do not return `CRYPTOCB_UNAVAILABLE` from subsequent update or final calls. Treat later failures as hard errors or keep an equivalent software state from the beginning. Test failures at each multipart transition.

wolfcrypt/src/port/silabs/silabs_key.c
```diff
@@ -0,0 +1,491 @@
...
+    aes->ctx.key.size = (uint32_t)(keyBits / 8);
+    aes->ctx.key.storage.method = SL_SE_KEY_STORAGE_EXTERNAL_WRAPPED;
+    /* The descriptor references the caller's blob, which must outlive the
+     * Aes. Nothing is copied: that is the point of a wrapped key. */
+    aes->ctx.key.storage.location.buffer.pointer = (uint8_t*)wrapped;
+    aes->ctx.key.storage.location.buffer.size = wrappedSz;
+    /* Scrub any plaintext key this object was carrying. The Secure Element
+     * holds the key from here on, and leaving software material behind would
+     * let a path that never consults ctx.keySet encrypt with the old key. */
+    ForceZero(aes->devKey, sizeof(aes->devKey));
```
aidangarske, line 246: **Resident-key APIs succeed when their callback engine is disabled**

The configuration explicitly supports selecting only a subset of engines, but the AES resident-key APIs are compiled whenever the overall port is enabled rather than requiring `WOLFSSL_SILABS_CRYPTOCB_CIPHER`; the ECC equivalents likewise do not require the ECC engine. These APIs return success, erase the software key material, and mark the resident binding active. If the corresponding engine is disabled, dispatch returns unavailable and software runs with the erased AES schedule or absent ECC scalar, yielding failure or incorrect output after a successful bind call.

**Fix:** Gate AES resident-key declarations and definitions on the cipher engine and ECC resident-key declarations and definitions on the ECC engine, or reject the bind before modifying the object. Add tests for each documented partial-engine configuration.

wolfssl/wolfcrypt/include.am
```diff
@@ -125,6 +125,8 @@ noinst_HEADERS+= \
                          wolfssl/wolfcrypt/port/silabs/silabs_ecc.h \
                          wolfssl/wolfcrypt/port/silabs/silabs_hash.h \
                          wolfssl/wolfcrypt/port/silabs/silabs_random.h \
+                         wolfssl/wolfcrypt/port/silabs/silabs_settings.h \
+                         wolfssl/wolfcrypt/port/silabs/silabs_cryptocb.h \
```
aidangarske: **The new configure profile installs unusable public headers**

`--enable-silabs-cryptocb` records `WOLFSSL_SILABS_CRYPTOCB` and `WOLFSSL_SILABS_HOST_TEST` in the generated, installed `wolfssl/options.h`. Installed `settings.h` consequently includes `wolfssl/wolfcrypt/port/silabs/silabs_settings.h`, but this PR adds that header and `silabs_cryptocb.h` only to `noinst_HEADERS`. The host shim is also distributed without being installed. A downstream consumer of this configure-and-install profile therefore fails while preprocessing the public wolfSSL headers.

**Fix:** Install every header required by the generated profile and place the shim on an installed include path, or prevent the host-test defines from entering installed `options.h`. Add a staged `make install` consumer-compilation test for this configure option.

IDE/SimplicityStudio/xg25/build.sh
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env bash
+# Generate and build the EFR32xG25 wolfCrypt test/benchmark project headlessly.
+# Requires the Silicon Labs toolchain installed via SLT (see README.md).
+set -euo pipefail
+
+HERE="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+BOARD="${BOARD:-brd4270b}"
+BUILD_DIR="${BUILD_DIR:-$HERE/build}"
```
aidangarske, line 8: **Unvalidated BUILD_DIR can erase arbitrary directories**

The new script accepts an unrestricted `BUILD_DIR` environment value and recursively deletes it before generation. Values such as `/`, the user's home directory, the repository root, or an unrelated existing project directory are accepted without validation and can cause irreversible data loss.

**Fix:** Canonicalize and validate the destination before deletion. Reject root, home, the repository root, empty or unresolved paths, and preferably require the directory to be beneath the example directory or contain a project-specific ownership marker.

wolfcrypt/src/port/silabs/silabs_key.c
```diff
@@ -0,0 +1,491 @@
...
+    }
+    *keySz = (word32)sz;
+
+    return 0;
+}
+#endif /* HAVE_ECC */
+
+#if !defined(NO_AES) && defined(WOLFSSL_SILABS_WRAPPED_KEYS)
+
+int wc_SilabsSe_AesGetWrappedKeySize(int keyBits, word32* outSz)
```
aidangarske, line 137: **Secure Vault public APIs have no behavioral coverage**

The PR adds seven public wrapped-key and built-in-key APIs with substantial sizing, generation, export, binding, and state-replacement logic, but no test calls any `wc_SilabsSe_*` function. The existing wolfCrypt run exercises plaintext keys and therefore cannot detect failures specific to these new public interfaces.

**Fix:** Add tests for short-buffer size reporting, invalid key sizes and curves, generate-bind-use flows, public-key export, rebind behavior, disabled engines, incorrect device routing, and cleanup. Use a controllable fake backend for host coverage and retain on-device tests for actual wrapped-key behavior.

IDE/SimplicityStudio/xg25/app.c
```diff
@@ -0,0 +1,198 @@
...
+#else
+    printf("Software path: C\n");
+#endif
+    print_se_info();
+
+#ifdef DEBUG_WOLFSSL
+    wolfSSL_Debugging_ON();
+#endif
+
+    ret = wolfCrypt_Init();
```
aidangarske, line 185: **TLS mode leaves a wolfSSL initialization reference outstanding**

The example calls `wolfCrypt_Init()` directly. When TLS mode creates its first context, wolfSSL performs its own automatic `wolfSSL_Init()`, which takes another wolfCrypt initialization reference. The example only calls `wolfCrypt_Cleanup()` once and never calls `wolfSSL_Cleanup()`, leaving the TLS globals and one wolfCrypt/SE callback reference initialized after the application reports completion.

**Fix:** In TLS mode, pair `wolfSSL_Init()` with `wolfSSL_Cleanup()` and avoid an unbalanced independent wolfCrypt reference, or explicitly pair both initialization layers.

wolfcrypt/test/test.c
```diff
@@ -83920,6 +84001,113 @@ WOLFSSL_TEST_SUBROUTINE wc_test_ret_t cryptocb_test(void)
...
+            XMEMCPY(badCipher, cpCipher, sizeof(badCipher));
+            XMEMCPY(badTag, cpTag, sizeof(badTag));
+            if (i == 0)
+                badTag[0] ^= 0x01;
+            else if (i == 1)
+                badCipher[0] ^= 0x01;
+            else
+                badAAD[0] ^= 0x01;
+
+            XMEMSET(cpOut, 0x5A, sizeof(cpOut));
```
aidangarske, line 84185: **ChaCha callback test does not verify output zeroization**

The new negative-tag test initializes `cpOut` to `0x5A` but only rejects output equal to the complete original plaintext. A callback that leaves the sentinel unchanged, exposes a partial plaintext, or clears only one byte still passes, even though the implementation and test comment require failed decryption to clear the output.

**Fix:** Compare the entire output against an all-zero buffer after every authentication failure. This directly verifies the fail-closed behavior implemented by both the core and the new callback.

aidangarske (changes requested):
## Skoll Code Review

**Scan type:** `review`

**10 finding(s) posted as inline comments** (see file-level comments below)

### Posted findings
- [High] **Fresh hash contexts can skip Secure Element initialization** — `wolfcrypt/src/port/silabs/silabs_cb_hash.c:95-108`
- [High] **One-shot CMAC can free an uninitialized device context** — `wolfcrypt/src/port/silabs/silabs_cb_cmac.c:88-91`
- [High] **Hash fallback loses data after the SE acquires the context** — `wolfcrypt/src/port/silabs/silabs_cb_hash.c:145-160`
- [High] **CMAC fallback is invalid after hardware initialization succeeds** — `wolfcrypt/src/port/silabs/silabs_cb_cmac.c:184-212`
- [High] **Resident-key APIs succeed when their callback engine is disabled** — `wolfcrypt/src/port/silabs/silabs_key.c:245-247`
- [High] **The new configure profile installs unusable public headers** — `wolfssl/wolfcrypt/include.am:124-129`
- [High] **Unvalidated BUILD_DIR can erase arbitrary directories** — `IDE/SimplicityStudio/xg25/build.sh:8-35`
- [Medium] **Secure Vault public APIs have no behavioral coverage** — `wolfcrypt/src/port/silabs/silabs_key.c:136-487`
- [Medium] **TLS mode leaves a wolfSSL initialization reference outstanding** — `IDE/SimplicityStudio/xg25/app.c:170-192`
- [Medium] **ChaCha callback test does not verify output zeroization** — `wolfcrypt/test/test.c:84097-84106`

wolfcrypt/src/port/silabs/silabs_key.c
```diff
@@ -0,0 +1,492 @@
...
+    if (ret != 0) {
+        return ret;
+    }
+    if (wrappedSz != need) {
+        return BAD_LENGTH_E;
+    }
+
+    /* Give wolfCrypt the curve so sizes and signature encoding are right,
+     * even though it never sees the private scalar. */
+    ret = wc_ecc_set_curve(key, (int)keySz, curveId);
```
aidangarske, line 444: **Resident ECC binding leaves the prior software private key exportable** · _Zeroization_

The new resident-key binding APIs reuse an existing `ecc_key`, change its type to `ECC_PRIVATEKEY`, and mark the SE descriptor non-exportable, but only wipe the new SiLabs `key_raw` buffer. They do not clear the existing software private scalar in `key->k` (or its blinding state). Before this PR, these SE-resident binding APIs did not exist. After binding a wrapped or built-in SE key onto an `ecc_key` that previously held a software private key, `wc_ecc_export_private_only()` still accepts the new `ECC_PRIVATEKEY` type and exports `ecc_get_k(key)`, recovering the old private scalar. This retains secret material across the key replacement and exposes it through a public export API despite the object now representing a non-exportable key.

**Fix:** Apply this replacement in both `wc_SilabsSe_EccUseWrappedKey()` and `wc_SilabsSe_EccUseBuiltInKey()`. Add a regression test that creates a software ECC key, binds a resident key to the same object, and verifies the former scalar cannot be exported.

**Suggestion:**
```suggestion
ForceZero(key->key_raw, sizeof(key->key_raw));
ecc_forcezero_k(key);
key->silabsKeySet = 1;
```

aidangarske (commented):
## Skoll Code Review

### Posted findings
- [Medium] **Resident ECC binding leaves the prior software private key exportable** — `wolfcrypt/src/port/silabs/silabs_key.c:428-446`

aidangarske (dismissed):

wolfcrypt/src/chacha20_poly1305.c
```diff
@@ -442,6 +446,17 @@ WOLFSSL_API int wc_ChaCha20Poly1305_Encrypt_ex(ChaCha* chacha, Poly1305* poly,
         return BAD_FUNC_ARG;
     }
 
+#ifdef WOLF_CRYPTO_CB
+    /* devId comes from wc_Chacha_SetKey_ex(). This is the path the TLS
+     * record layer uses. Software runs if the device declines. */
+    if (chacha->devId != INVALID_DEVID) {
```
aidangarske: **Callback path mishandles 128-bit ChaCha keys**

`wc_Chacha_SetKey_ex()` still accepts both 16- and 32-byte ChaCha keys, but this new dispatch sends only `devKey`, not `devKeySz`, to the callback. For a 16-byte key, only the first half of `devKey` is initialized. The SiLabs handler then always describes it as a 32-byte `SL_SE_KEY_TYPE_CHACHA20` key. Consequently, a callback device can encrypt or decrypt with stale/uninitialized trailing bytes rather than the 128-bit ChaCha key that the existing software path uses. Gate the AEAD callback on the required 32-byte key length so 16-byte contexts retain their prior software behavior.

**Fix:** Apply the same guard to decrypt and add a callback-device test that pre-keys a ChaCha context with a 16-byte key, verifies no hardware result is accepted, and compares software encryption/decryption with the prior behavior.

**Suggestion:**
```suggestion
if (chacha->devId != INVALID_DEVID &&
        chacha->devKeySz == CHACHA20_POLY1305_AEAD_KEYSIZE) {
```

configure.ac
```diff
@@ -3830,6 +3830,31 @@ then
...
 
 
+# Silicon Labs EFR32 Secure Element crypto-callback port.
+# On-target the application supplies the Simplicity SDK include path. This
+# option is a host compile-test of the port: it swaps the SE Manager headers
+# for a shim (WOLFSSL_SILABS_HOST_TEST) so the cryptocb dispatch and wiring
+# build without the vendor SDK. It forces crypto callbacks on (see the
+# cryptocb block).
+# Example: "./configure --enable-silabs-cryptocb"
+ENABLED_SILABS_CRYPTOCB="no"
```
aidangarske, line 3868: **No supported production build includes the callback port**

The sole Automake source path for the new port is `BUILD_SILABS_CRYPTOCB`, but its only enable option unconditionally defines `WOLFSSL_SILABS_HOST_TEST`. That profile therefore always compiles the shim, whose operations all return `SL_STATUS_NOT_SUPPORTED`, rather than a target port. Conversely, defining `WOLFSSL_SILABS_CRYPTOCB` through a target `user_settings.h` compiles new calls such as `wc_SilabsCryptoCb_RegisterDevice()` in `wc_port.c` without including the new SiLabs source files. CMake has neither a SiLabs callback option nor source-list wiring, so its user-settings integration has the same unresolved-symbol failure. There is currently no canonical production build path for the advertised port.

**Fix:** Keep the shim-only option as a host gate, but add a separate production target configuration that includes the SiLabs sources without `WOLFSSL_SILABS_HOST_TEST`. Wire the same source set and definitions into CMake, then add a non-shim target/consumer build test.

aidangarske (changes requested):
## Skoll Code Review

**Scan type:** `review`

### Posted findings
- [High] **Callback path mishandles 128-bit ChaCha keys** — `wolfcrypt/src/chacha20_poly1305.c:452-456`
- [High] **No supported production build includes the callback port** — `configure.ac:3840-3855`

wolfcrypt/src/port/silabs/silabs_key.c
```diff
@@ -0,0 +1,518 @@
...
+            (int)sl_se_export_public_key(&cmd, &desc, &pubDesc));
+        if (ret == 0) {
+            *pubOutSz = keySz * 2;
+        }
+    }
+
+    return ret;
+}
+
+int wc_SilabsSe_EccUseWrappedKey(ecc_key* key, const byte* wrapped,
```
aidangarske, line 447: When binding this wrapped ECC private key, where does `key->pubkey` get updated to the corresponding public point? Since `wc_SilabsSe_EccGenerateWrappedKey()` already gets the public key from the SE should this bind path also populate key->pubkey from the wrapped key's corresponding SE public key?

wolfcrypt/src/include.am
```diff
@@ -288,6 +296,21 @@ if BUILD_RTL8735B
 src_libwolfssl@LIBSUFFIX@_la_SOURCES += wolfcrypt/src/port/realtek/rtl8735b.c
 endif
 
+if BUILD_SILABS_CRYPTOCB
```
aidangarske, line 299: do we need to wire this cmake source as well? I dont see a `BUILD_SILABS_CRYPTOCB` in the cmake. do we expect this port to use cmake and user_settings.h?

wolfcrypt/src/port/silabs/silabs_cb_kdf.c
```diff
@@ -0,0 +1,182 @@
...
+    if (seHash == SL_SE_HASH_NONE) {
+        return WC_NO_ERR_TRACE(CRYPTOCB_UNAVAILABLE);
+    }
+
+    silabs_kdf_plain_key(&inKey, info->kdf.pbkdf2.passwd,
+        (word32)info->kdf.pbkdf2.pLen);
+    silabs_kdf_plain_key(&outKey, info->kdf.pbkdf2.output,
+        (word32)info->kdf.pbkdf2.kLen);
+
+    status = sl_se_derive_key_pbkdf2(&cmd, &inKey,
```
aidangarske, line 154: Should we return a `BAD_FUNC_ARG` or something here if the salt == NULL or if sLen is != 0 I dont know if this is checked in the dispatcher

wolfcrypt/src/port/silabs/silabs_cb_kdf.c
```diff
@@ -0,0 +1,182 @@
...
+
+    seHash = silabs_cb_hash_type(info->kdf.hkdf.hashType, NULL);
+    if (seHash == SL_SE_HASH_NONE) {
+        return WC_NO_ERR_TRACE(CRYPTOCB_UNAVAILABLE);
+    }
+
+    silabs_kdf_plain_key(&inKey, info->kdf.hkdf.inKey, info->kdf.hkdf.inKeySz);
+    silabs_kdf_plain_key(&outKey, info->kdf.hkdf.out, info->kdf.hkdf.outSz);
+
+    status = sl_se_derive_key_hkdf(&cmd, &inKey,
```
aidangarske, line 104: should we mirror the checks here that are in wolfcrypt `wc_HKDF_ex()` we do info == NULL && infoSz > 0, and Expand enforces RFC 5869 L <= 255 * HashLen perhaps we should check that here?

wolfcrypt/src/port/silabs/silabs_cb_cipher.c
```diff
@@ -0,0 +1,469 @@
...
+                              SL_SE_AES_BLOCK_SIZE];
+    uint32_t ncOff = 0;
+    sl_status_t status;
+    int ret;
+
+    if (aes == NULL || info->cipher.aesctr.out == NULL ||
+        info->cipher.aesctr.in == NULL) {
+        return BAD_FUNC_ARG;
+    }
+    if (aes->left != 0 ||
```
aidangarske, line 109: does this work for wrapped / built in keys? I see that those bindings intentionally remove software key schedule so a valid partial CTR request or continuation cant actually fallback to software here 

this could make valid ctr usage fail for resident keys maybe

wolfcrypt/src/port/silabs/silabs_cryptocb.c
```diff
@@ -0,0 +1,199 @@
...
+     * are not flattened into WC_HW_E. */
+    if (slStatus < 0) {
+        return slStatus;
+    }
+
+    switch ((sl_status_t)slStatus) {
+    case SL_STATUS_OK:
+        return 0;
+    case SL_STATUS_NOT_SUPPORTED:
+    case SL_STATUS_INVALID_PARAMETER:
```
aidangarske, line 62: Is SL_STATUS_INVALID_PARAMETER guaranteed by SE Manager to be returned before any output/context/IV mutation? We globally convert it to CRYPTOCB_UNAVAILABLE, which tells wolfCrypt to retry in software. I think that  `SL_STATUS_INVALID_PARAMETER` should mean safe to fall back to unless we can document `INVALID_PARAMETER` is side effect free

wolfssl/wolfcrypt/port/silabs/silabs_settings.h
```diff
@@ -0,0 +1,110 @@
...
+
+/* Pulls the SE context members into Aes, ecc_key and wc_Sha*, shared with the
+ * direct port. Defined in settings.h for WOLFSSL_SILABS_SE_ACCEL as well. */
+#ifndef WOLFSSL_SILABS_SE_TYPES
+    #define WOLFSSL_SILABS_SE_TYPES
+#endif
+
+/* If the port is on but no specific engine was requested, enable the full
+ * supported set. */
+#if !defined(WOLFSSL_SILABS_CRYPTOCB_TRNG)   && \
```
aidangarske, line 74: **Default engine selection breaks reduced configurations** · _Regression_

Before this PR, reduced configurations such as `user_settings_min_ecc.h` could omit AES, DES3, ChaCha, RNG, hashes, and KDFs independently. The new default block enables the TRNG, hash, cipher, and KDF callback handlers without checking whether their corresponding `wc_CryptoInfo` members exist. Those members are conditionally compiled in `cryptocb.h`, while the new handlers unconditionally reference `info->rng`, `info->hash`, `info->cipher`, or `info->kdf`. For example, a minimum-ECC configuration defines `NO_AES`, `NO_DES3`, `NO_SHA`, and `NO_PWDBASED` without enabling ChaCha or HKDF; adding the advertised `WOLFSSL_SILABS_CRYPTOCB` macro selects cipher and KDF handlers whose required union members are absent, causing compilation to fail.

**Fix:** Enable each default engine only when its algorithms and corresponding `wc_CryptoInfo` member are available. Add reduced builds covering minimum ECC, `WC_NO_RNG`, no-hash, and no-KDF configurations.

wolfcrypt/src/pwdbased.c
```diff
@@ -255,6 +258,16 @@ int wc_PBKDF2_ex(byte* output, const byte* passwd, int pLen, const byte* salt,
         return BAD_FUNC_ARG;
     }
 
+#ifdef WOLF_CRYPTO_CB
+    if (devId != INVALID_DEVID) {
+        ret = wc_CryptoCb_Pbkdf2(output, passwd, pLen, salt, sLen, iterations,
+            kLen, hashType, devId);
```
aidangarske, line 272: **PBKDF2 callback receives invalid pointer-length pairs before validation** · _Security_

Before this PR, `passwd == NULL` with a positive password length was rejected by `wc_HmacSetKey_ex`, and `salt == NULL` with a positive salt length was rejected by `wc_HmacUpdate`. The new callback dispatch occurs before either validation point. Consequently, a registered callback receives invalid pointer-length pairs. The new SiLabs callback checks the password pair but passes `salt` directly to `sl_se_derive_key_pbkdf2` without validating `salt == NULL && sLen > 0`, exposing the invalid pointer to the external SDK call instead of returning `BAD_FUNC_ARG` consistently.

**Fix:** Validate both pointer-length pairs before callback dispatch while continuing to permit NULL pointers for zero-length inputs. Add callback-enabled tests for NULL pointers paired with nonzero lengths.

**Suggestion:**
```suggestion
if (output == NULL || (passwd == NULL && pLen > 0) ||
        (salt == NULL && sLen > 0) || pLen < 0 || sLen < 0 || kLen < 0) {
```

aidangarske (commented):
## Skoll Code Review

### Posted findings
- [Medium] **Default engine selection breaks reduced configurations** — `wolfssl/wolfcrypt/port/silabs/silabs_settings.h:74-87`
- [Medium] **PBKDF2 callback receives invalid pointer-length pairs before validation** — `wolfcrypt/src/pwdbased.c:231-264`
- [Medium] **Resident AES keys cannot perform zero-length AEAD operations** — `wolfcrypt/src/port/silabs/silabs_cb_cipher.c:324-327`

wolfcrypt/src/port/silabs/silabs_cb_cipher.c
```diff
@@ -0,0 +1,469 @@
...
+        if (info->cipher.enc) {
+            /* The SE computes only full-length GCM tags. */
+            if (info->cipher.aesgcm_enc.authTagSz <
+                    WC_AES_BLOCK_SIZE) {
+                return WC_NO_ERR_TRACE(CRYPTOCB_UNAVAILABLE);
+            }
+            /* GMAC authenticates AAD with no plaintext, so wolfCrypt passes a
+             * zero length and NULL in/out. The SE helper requires both, so
+             * leave that case to software. */
+            if (info->cipher.aesgcm_enc.sz == 0 ||
```
aidangarske: **Resident AES keys cannot perform zero-length AEAD operations** · _Logic_

Before this PR, wolfCrypt's GCM and CCM APIs accepted zero-length messages and required input/output pointers only when `sz != 0`. The shared SiLabs helpers also support this case by substituting a dummy buffer. The new callback instead returns `CRYPTOCB_UNAVAILABLE` whenever `sz == 0` for GCM and CCM encryption and decryption. Plaintext keys can fall back to software, but wrapped and built-in keys set `aes->ctx.keySet`; `wc_SilabsCipher` then converts the unavailable result to `WC_HW_E` because no software key exists. Authentication-only AEAD operations therefore fail specifically for the new resident-key APIs, despite being valid public API operations and supported by the underlying helper.

**Fix:** Apply the same conditional-pointer check to GCM decryption and CCM encryption/decryption, allow `sz == 0` to reach the existing helpers, and add zero-length AEAD tests using wrapped or built-in keys.

**Suggestion:**
```suggestion
if (info->cipher.aesgcm_enc.sz != 0 &&
        (info->cipher.aesgcm_enc.in == NULL ||
         info->cipher.aesgcm_enc.out == NULL)) {
    return WC_NO_ERR_TRACE(CRYPTOCB_UNAVAILABLE);
}
```

aidangarske (approved):

PR #11281, opened 2026-08-26
[STSAFE] Normalize the digest to field size in ECDSA verify

danielinux:
# Description
F-11238: normalize the digest to field size in STSAFE ECDSA verify
    
    stsafe_verify copied key_sz bytes from the caller's digest buffer and
    told the SE the digest was key_sz bytes long. For a digest shorter than
    the field size (P-256 with a 28-byte SHA-224 hash, P-384 with SHA-256)
    that reads past the end of the buffer and submits a wrong digest, so a
    valid signature fails to verify.
    
    The A120 expects a field-size prehash on the wire, so the digest is
    normalized host-side before submission: truncate if longer, left-pad if
    shorter - the same convention the sign path already applies. The SE
    still receives key_sz bytes, so the wire contract is unchanged; only
    the buffer contents for short digests are corrected.
    
    Both variants (A120 STSELib, A100/A110 legacy SDK) and both call sites
    (crypto-cb, TLS verify-peer callback) now use the real digest length.
    

# Testing

    - Built: host x86_64, --enable-stsafe=a120, STSELib v1.1.7 + Rust
      A120 simulator.
    - Tested: wolfCrypt STSAFE-A120 sim test (wolfSSL/simulators
      wolfcrypt-test): new P-256 + 28-byte-digest verify case fails on the
      pre-fix build (valid signature rejected) and passes post-fix; rng,
      on-device keygen, sign and 32-byte verify round-trip unchanged.
    - A100/A110 variant not runnable here (closed SDK, no silicon); change
      mirrors the A120 logic on the existing buffer and keeps Length=key_sz.

### DIFF
--- a/.github/workflows/stsafe-a120-sim.yml
+++ b/.github/workflows/stsafe-a120-sim.yml
@@ -37,7 +37,7 @@ permissions:
 #      CFLAG to the new --enable-stsafe=a120 option.
 
 env:
-  SIMULATORS_REF: 7e4a50ac3316680c523ec2f66711ea8403c94468
+  SIMULATORS_REF: 9f9900295b34116e9f9cba978c9143c608d36f7d
 
 jobs:
   stsafe_a120_sim:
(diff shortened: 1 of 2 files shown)


danielinux:
jenkins retest this please

philljj (approved):

PR #11282, opened 2026-08-26
[NXP DCP port] Fix four fenrir claims

danielinux:
f52198314 F-11239: give copied DCP hash contexts their own channel
d67eb0d5f F-11244: release the reserved DCP channel when hash init fails
5a8127d2a F-8261: check DCP lock results before touching shared state
f4ad17ce8 F-7127: restructure dcp_key_slot as mutually exclusive branches

### DIFF
--- a/wolfcrypt/src/port/nxp/dcp_port.c
+++ b/wolfcrypt/src/port/nxp/dcp_port.c
@@ -198,10 +198,12 @@ int wc_dcp_init(void)
     return 0;
 }
 
-/* Release a channel without taking the DCP lock. For the dcp_lock()
- * failure branches only: a contended lock blocks rather than fails, so
- * a failure there means no other thread can be inside the critical
- * section and the clear cannot race a live access. */
+/* Release a channel without taking the DCP lock. Safe whenever
+ * dcp_lock() is known to be unavailable: a contended lock blocks
+ * rather than fails, so a failure means no other thread can be
+ * inside the critical section (or the caller already holds the
+ * mutex), and the owner check prevents releasing a channel held by
+ * another context. */
 static void dcp_free_unlocked(void* owner, int ch)
 {
 #ifndef SINGLE_THREADED
@@ -220,12 +222,17 @@ static void dcp_free_unlocked(void* owner, int ch)
 #endif
 }
 
+/* Total: always releases, so callers may drop their handle record
+ * unconditionally. A lock failure falls back to the unlocked release
+ * (safe per dcp_free_unlocked) instead of orphaning the reservation. */
 static void dcp_free(void* owner, int ch)
 {
-    if (dcp_lock() != 0)
-        return;
+    int locked;
+
+    locked = (dcp_lock() == 0);
     dcp_free_unlocked(owner, ch);
-    dcp_unlock();
+    if (locked)
+        dcp_unlock();
 }
 
 /*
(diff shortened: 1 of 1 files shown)


wolfcrypt/src/port/nxp/dcp_port.c
```diff
@@ -419,10 +513,41 @@ int wc_Sha256GetFlags(wc_Sha256* sha256, word32* flags)
...
         return BAD_FUNC_ARG;
-    dcp_lock();
+    if (src == dst)
+        return 0;
+    handleWord = dcp_hash_handle_word();
+    if (handleWord >= DCP_HASH_CTX_SIZE)
+        return WC_HW_E;
+    /* The copy replaces dst (software contract: free dst resources,
+     * then clone); drop any channel dst owns. */
+    dcp_free(dst, dst->handle.channel);
```
danielinux: Rejected - no code change. Three reasons:

1. **Not a leak.** For the copy to lose its channel *and* still fail, the pool must be exactly full **and** a concurrent thread must reserve the channel dst just released in the free-to-reserve window. In that case the channel is legitimately held by the other thread (first-come), and `WC_HW_WAIT_E` is the correct resource-wait error for the resource genuinely being unavailable at reservation time - the clear-error semantics this port now uses for exhaustion (CAAM returns it for `ResourceNotAvailable`, Atmel for a failed slot allocation).
2. **The stale `handle.channel` is inert.** The owner-validated release in this PR makes every later free/init/copy of that struct a no-op for a value it does not own, so the stale value cannot double-free or free another context's channel.
3. **The proposed ordering is strictly worse.** Reserve-first cannot see dst's own channel as free while dst still holds it: in the exactly-full case, a copy that succeeds today by recycling dst's channel would fail instead, and the ordering holds two of the four channels transiently.

The scenario also requires multithreaded operation (SINGLE_THREADED builds have no pool at all); its worst outcome is a retryable resource-wait error, not corruption or loss.

danielinux:
retest this please
(unrelated FIPS failure)

danielinux:
retest this please
(different timeouts)

wolfcrypt/src/port/nxp/dcp_port.c
```diff
@@ -419,10 +534,47 @@ int wc_Sha256GetFlags(wc_Sha256* sha256, word32* flags)
 
 int wc_Sha256Copy(wc_Sha256* src, wc_Sha256* dst)
```
philljj, line 585: would this ever be used with Hmac and `WOLFSSL_HMAC_COPY_HASH`? That would result in `struct Hmac` having two extra `wc_HmacHash` fields to `wc_Sha256Copy()`, which might exhaust this.

philljj (changes requested):

wolfcrypt/src/port/nxp/dcp_port.c
danielinux: This is now partially addressed in 2abecd6. HmacUpdate/Final is fixed now, dst keeps its channel. Unfortunately wolfSSL_HmacCopy cannot work because it needs 3+3 channels (6) and HW has 4.

philljj:
Retest this please.

  [0] Incorrect return -308

danielinux:
retest this please

philljj:
Retest this please.

(no logs)

philljj (approved):

PR #11289, opened 2026-08-26
Send unexpected message alert for UNKNOWN_RECORD_TYPE

rlm2002:
# Description

`DoProcessReplyEx()` rejected records with an undefined `ContentType` but returned `UNKNOWN_RECORD_TYPE` without putting an alert on the wire. RFC 8446 section 5 and RFC 9846 section 5 both require that a TLS implementation receiving an unexpected record type MUST terminate the connection with an unexpected_message alert. RFC 5246 section 6 states the same for TLS 1.2.

Two rejection sites were affected:

1. The error mapping switch in case `getRecordLayerHeader`, which handled `VERSION_ERROR` and `LENGTH_ERROR`, but let `UNKNOWN_RECORD_TYPE` fall through to `default: break`.
2. The record dispatch `default: arm`, reached by types the whitelist admits but nothing on a TLS connection handles. `dtls12_cid` (25) and `ack` (26) pass the header check whenever `WOLFSSL_DTLS_CID` or `WOLFSSL_DTLS13` are built, with no `ssl->options.dtls` guard. It is also reached with the inner `ContentType` recovered by `removeMsgInnerPadding()`.

Both now send a fatal `unexpected_message` before returning.

Fixes F-10722

# Testing

Added new unit test: `test_tls_unknown_record_type_alert`

# Checklist

 - [x] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -25404,6 +25404,10 @@ static int DoProcessReplyEx(WOLFSSL* ssl, int allowSocketErr)
                     /* invalid record length, RFC 8446 section 5.1 */
                     SendAlert(ssl, alert_fatal, record_overflow);
                     break;
+                case WC_NO_ERR_TRACE(UNKNOWN_RECORD_TYPE):
+                    /* undefined record type, RFC 8446/9846 section 5 */
+                    SendAlert(ssl, alert_fatal, unexpected_message);
+                    break;
                 default:
                     break;
                 }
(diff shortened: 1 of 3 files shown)


rlm2002:
Retest this please Jenkins: history lost

night1rider:
Jenkins retest this please

rlm2002:
Retest this please Jenkins

PR #11290, opened 2026-08-26
MASM: non-AVX VEX fix, x64 home space, vzeroupper on VAES exits

kaleb-himes:
# Description @SparkiDev 

1. vzeroupper: After using the wide 256-bit registers, this instruction wipes their upper halves before returning. Without it, the next ordinary SSE code in the program runs slow. Speed only, nothing computes wrong. (Intel Optimization Reference Manual, "Mixing AVX and SSE code".)

2. vmovd → movd: Three ECC routines are the versions used on old CPUs that lack AVX, but each had an AVX-only instruction in it. Those CPUs crash. Proved it: under an emulated Nehalem the old library dies with SIGILL, the fixed one passes. (Intel SDM Vol. 2A §2.3.)

3. sub rsp,32 before calls: Windows requires the caller to set aside 32 bytes of stack for the function it calls. We weren't, so the callee could overwrite our data. Windows only. (Microsoft, "x64 calling convention".)

4. The big block-copy change: Copies now use plain integer registers instead of vector ones. We didn't need this fix; it came along because you require the committed asm to match the generator exactly, and that's what the generator emits. I can remove item 4 if there is an outstanding PR on the scripts/ repo that fixes the generators and simply hasn't merged yet.

Pairs with: https://github.com/wolfSSL/scripts/pull/671

# Testing

Emulation and virtualized.

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen

### DIFF
--- a/wolfcrypt/src/aes_x86_64_asm.S
+++ b/wolfcrypt/src/aes_x86_64_asm.S
@@ -2463,6 +2463,7 @@ L_AES_ECB_encrypt_vaes_16_aes_enc_block_last:
         cmpl	%r9d, %eax
         jb	L_AES_ECB_encrypt_vaes_enc_16
 L_AES_ECB_encrypt_vaes_done_enc:
+        vzeroupper
         repz retq
 #ifndef __APPLE__
 .size	AES_ECB_encrypt_vaes,.-AES_ECB_encrypt_vaes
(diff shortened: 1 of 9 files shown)


kaleb-himes:
retest this please. (no logs).

kaleb-himes:
retest this please (aborted, no test failure)...

kaleb-himes:
Pairs with: https://github.com/wolfSSL/scripts/pull/671

kaleb-himes:
retest this please.

SparkiDev (approved):

PR #11291, opened 2026-08-26
Create the persisted session cache file with owner-only permissions

gasbytes:
# Description

On POSIX builds wolfssl_save_session_cache now opens the file with open in 0600 mode + fdopen.

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


gasbytes:
Jenkins retest this please (lost logs)

gasbytes:
Jenkins retest this please

gasbytes:
Jenkins retest this please

src/ssl_sess.c
```diff
@@ -595,6 +595,45 @@ int wolfSSL_memrestore_session_cache(const void* mem, int sz)
...
+{
+    XFILE file;
+    int fd;
+
+    fd = open(fname, O_RDWR | O_CREAT | O_TRUNC | O_CLOEXEC,
+              S_IRUSR | S_IWUSR);
+    if (fd < 0)
+        return XBADFILE;
+
+    if (fchmod(fd, S_IRUSR | S_IWUSR) != 0) {
```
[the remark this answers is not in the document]
philljj: I think this is fine. If we're saving the cache then the old cache was probably stale (or had already been restored from).

src/ssl_sess.c
```diff
@@ -595,6 +595,45 @@ int wolfSSL_memrestore_session_cache(const void* mem, int sz)
...
+    #define O_CLOEXEC 0
+#endif
+
+/* Open the save file for writing, owner-only whether created or overwritten. */
+static XFILE SaveSessionCacheOpen(const char* fname)
+{
+    XFILE file;
+    int fd;
+
+    fd = open(fname, O_RDWR | O_CREAT | O_TRUNC | O_CLOEXEC,
```
philljj: We have a wc_port wrapper function that does this with fallback in case O_CLOEXEC is not supported on the system:

```
int wc_open_cloexec(const char* path, int flags)                                
{                                                                               
    int fd = open(path, flags | O_CLOEXEC);                                     
#ifdef FD_CLOEXEC                                                               
    if (fd < 0 && errno == EINVAL) {                                            
        fd = open(path, flags);                                                 
        wc_set_cloexec(fd);                                                     
    }                                                                           
#endif                                                                          
    return fd;                                                                  
}
```

src/ssl_sess.c
philljj: amazingly, this seems to be the first appearance of fchmod / chmod in our library source:

```
$grep -ir "chmod(" * --include=*.[c,h]
src/ssl_sess.c:    if (fchmod(fd, S_IRUSR | S_IWUSR) != 0) {
```

This `fchmod()` and `open()` might belong in `wc_port.c` in a wrapper. I doubt this will be the last time we need to do this.

We usually don't do straight `open()` calls in normal src code, instead using `wc_open*()` wrappers.

In wc_port we would know it's safely guarded with the right defines:
```
#if (defined(__unix__) || defined(__APPLE__)) && \                              
    !defined(WOLFSSL_KERNEL_MODE) && !defined(WOLFSSL_ZEPHYR) && \              
    !defined(WOLFSSL_SGX)                                                       
                                                                                
#include <fcntl.h>                                                              
#include <errno.h>                                                              
#include <sys/socket.h>
...
```

philljj (changes requested):
I think this looks good.

I would just suggest to use the wc_open_cloexec wrapper, and see if the fchmod can be refactored to a new or existing wrapper in wc_port.c

src/ssl_sess.c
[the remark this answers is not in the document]
gasbytes: addressed the second part regarding the fhmocd, added a flag WOLFSSL_NO_FCHMOD for that in case fchmod is not available etc...

src/ssl_sess.c
gasbytes: as decided, created a second version that takes the mode, since O_CREAT without a mode specified is UB (https://linux.die.net/man/2/open)

src/ssl_sess.c
gasbytes: thanks, moved both into wc_fopen_owner_only() in wc_port.c.

gasbytes:

gasbytes:
Jenkins retest this please (lost logs)

gasbytes:

gasbytes:

gasbytes:

philljj (approved):

PR #11294, opened 2026-08-26
Require a key for 3DES ECB operations

dgarske:
Fixes F-11142

## Problem

The 3DES ECB entry points ran without checking whether a key had ever been configured. A `Des3` that had been through `wc_Des3Init()` (or `wc_Des3Free()`) but never through `wc_Des3_SetKey()` was encrypted with the zeroed key schedule left behind, and the call returned 0. The caller got back something that looks like ciphertext, with no indication anything was wrong.

This is the same fail-open class that PR #10632 (F-5379) closed for `wc_Des3_CbcEncrypt()` and `wc_Des3_CbcDecrypt()` in June, which added the `keySet` marker to `struct Des3`. That change guarded the two software CBC entry points and missed ECB entirely -- in the software path and in both hardware backends that implement it.

The gap is reachable from the OpenSSL compatibility layer. `wolfSSL_EVP_CipherInit()` only calls `wc_Des3_SetKey()` when a key is supplied (`evp.c:8595`), so `EVP_CipherInit(ctx, EVP_des_ede3_ecb(), NULL, NULL, enc)` followed by `EVP_CipherUpdate()` previously ran against an unkeyed context and reported success.

## Testing

Built with `--enable-des3 --enable-opensslall --enable-pkcs7 CFLAGS="-DWOLFSSL_DES_ECB"`.

### DIFF
--- a/doc/dox_comments/header_files/des3.h
+++ b/doc/dox_comments/header_files/des3.h
@@ -171,8 +171,11 @@ int  wc_Des_EcbEncrypt(Des* des, byte* out,
     directly whenever possible.
 
     \return 0 Returned upon successfully encrypting the given plaintext
+    \return BAD_FUNC_ARG If des, out, or in is NULL.
+    \return MISSING_KEY If no key has been set on the Des3 structure with
+    wc_Des3_SetKey.
 
-    \param des3 pointer to the Des3 structure to use for encryption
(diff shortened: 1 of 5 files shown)


dgarske:
Jenkins retest this please. History lost.

philljj:
Retest this please.

fips

philljj (approved):

PR #11303, opened 2026-08-27
PQ algorithm PCT's and RSA compliance fixes

kaleb-himes:
# Description

Branch D: FIPS compliance fixes inside the crypto boundary for v7

RSA

- Salt for PSS signatures can't be longer than the hash. FIPS 186-5 §5.4(g) says so with no exception, so it now holds even in builds that allow long salts. Checked on signing and on verifying.
- Key validation now runs the full set of checks NIST specifies (SP 800-56B §6.4.1.4.3): primes the right size and far enough apart, actually prime, exponent in range, private exponent big enough, CRT parts sane.
- Messages outside the legal range are rejected before the math runs.
- Only 2048/3072/4096-bit keys accepted, the sizes we hold certificates for.
- Exponent must be above 65537 when generating a key.
- A random-generator failure now reports itself instead of being reported as "couldn't find a prime."
- Recovered plaintext gets wiped on the way out, including on errors.

Post-quantum key generation

- ML-KEM, ML-DSA and SLH-DSA now test every key pair they generate before handing it back, which FIPS requires. A key that fails the test is destroyed so a caller who ignores the error can't use it by accident.
- For SLH-DSA we picked the cheap test: recompute the public half from the private half and compare. Signing and verifying would also work but costs about 10x more in wall clock time per key. We measured it.

Supporting

- New fips_cast_bench tool that times the module's self-tests, so a customer on slow hardware can budget start-up time.
- One test expectation updated: it asked for the longest possible salt, which v7 correctly refuses now and limits to hash length.
- Windows build placement fixes so ML-KEM and SLH-DSA land inside the integrity-checked region.
- Comments cleaned up and two wrong standards references corrected.

# Testing

Testing against the FIPS standards with linux user and kernel space solutions

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


kaleb-himes:
retest this please (no logs)..

kaleb-himes:
retest this please.......

wolfcrypt/src/wc_mlkem.c
```diff
@@ -1042,8 +1002,88 @@ int wc_MlKemKey_MakeKeyWithRandom(MlKemKey* key, const unsigned char* rand,
 #endif
 #endif
 
-    /* Note: PCT is performed in wc_MlKemKey_MakeKey() which calls this
-     * function and has the RNG parameter needed for encapsulation. */
+#if FIPS_VERSION3_GE(7,0,0)
```
kaleb-himes: Will add an opt in for non-FIPS builds but these must be gated on v7 only since that is the first module claiming FIPS compliance for these algorithms. Widening to HAVE_FIPS would result in invalid claims if compiled with older modules where these algorithms were outside the module boundary and no FIPS claim can be made for them until the v7 module.

wolfcrypt/src/wc_mldsa.c
```diff
@@ -11228,71 +11297,60 @@ int wc_MlDsaKey_MakeKey(wc_MlDsaKey* key, WC_RNG* rng)
...
+    else if (key->params == NULL) {
+        ret = BAD_STATE_E;
+    }
+    else {
+        ret = mldsa_make_key_from_seed(key, seed);
     }
 
-    /* Note: PCT is performed in wc_MlDsaKey_MakeKey() which calls this
-     * function and has the RNG parameter needed for signing. */
+#if FIPS_VERSION3_GE(7,0,0)
```
kaleb-himes: Will add an opt in for non-FIPS builds but these must be gated on v7 only since that is the first module claiming FIPS compliance for these algorithms. Widening to HAVE_FIPS would result in invalid claims if compiled with older modules where these algorithms were outside the module boundary and no FIPS claim can be made for them until the v7 module.

kaleb-himes:
wolfSSL Inc. would like to extend our gratitude to @Arpan0995 for a detailed and well-written report identifying a FIPS 204 conformance edge case in ML-DSA hint validation (mldsa_check_hint). Their contribution included an accurate root-cause analysis, a standalone reproducer, a proposed fix, and verification of the final PR patch.

SparkiDev (approved):

Issue #9156, opened 2025-09-02
[Bug] TLS 1.3 issue with HAProxy


feld:
### Version

5.8.0

### Description

I am not sure if this is related to #8793 at all, but with a WolfSSL-flavored HAProxy I get this TLS error when sending requests from Erlang/Elixir:

```
[warning] Description: ~c"Failed to assert middlebox server message"
     Reason: [missing: {:change_cipher_spec, 1}]

[notice] TLS :client: In state :hello_middlebox_assert at ssl_gen_statem.erl:803 generated CLIENT ALERT: Fatal - Unexpected Message
 - {:unexpected_msg,
 {:internal,
  {:encrypted_extensions,
   %{
     elliptic_curves: {:supported_groups,
      [:secp521r1, :secp384r1, :secp256r1, :x25519, :ffdhe2048]}
   }}}}
{:error,
 {:tls_alert,
  {:unexpected_message,
   ~c"TLS client: In state hello_middlebox_assert at ssl_gen_statem.erl:803 generated CLIENT ALERT: Fatal - Unexpected Message\n {unexpected_msg,\n     {internal,\n         {encrypted_extensions,\n             \#{elliptic_curves =>\n                   {supported_groups,\n                       [secp521r1,secp384r1,secp256r1,x25519,ffdhe2048]}}}}}"}}}
```

This issue is documented upstream in Erlang as a TLS implementation bug on the server side: https://github.com/erlang/otp/issues/8470

I have not tried building against WolfSSL 5.8.2 yet, but if it is believed the issue may be fixed I can do a custom build to validate as well as ensure those changes get pushed upstream in FreeBSD which is has an HAProxy variant built against WolfSSL

### Relevant log output

```shell

```

anhu:
Hi @feld , I can't really tell if this is related to the ticket that you pointed out.  I'll be helping you with this issue as well as looking into the other ticket. Can you give me step by step instructions on how to reproduce what you are seeing?  Please note that I have limitted experience with HAProxy and Erlang. 

Warm regards, Anthony

anhu:
Hi, 

I've successfully connected to the server with the `examples/client/client` and   I cannot reproduce what you are seeing.  I've built with both `WOLFSSL_TLS13_MIDDLEBOX_COMPAT` defined and undefined. 

```
./configure 
```

```
./configure CFLAGS=-DWOLFSSL_TLS13_MIDDLEBOX_COMPAT
```

Both yield the same results. 

```
[...]

PR #11306, opened 2026-08-28
Send server ChangeCipherSpec when client offers a session id

Frauschi:
## Description

RFC 8446 Appendix D.4 says a TLS 1.3 server answers a ClientHello carrying a non-empty `legacy_session_id` with a ChangeCipherSpec. Both server send paths sat behind `WOLFSSL_TLS13_MIDDLEBOX_COMPAT`, which only `--enable-tls13-middlebox-compat` and `--enable-jni` define and neither enables by default, so a default-built server never sent one. Peers that require it, such as Erlang's `ssl`, abort with an `unexpected_message` alert.

Only wolfSSL-as-server was affected: the wolfSSL client sends an empty session id by default, so no peer is ever obliged to answer back. Reported as an SLH-DSA interop bug, but nothing in this path touches the certificate or sigalg and the regression test reproduces it with the stock test certs.

## Changes

Both server sends now compile unconditionally, driven by the runtime `tls13MiddleBoxCompat` flag the server already clears for an empty session id, and are factored into one static `SendTls13ServerChangeCipher()`.

- `sentChangeCipher` is compiled in for every TLS 1.3 build. `SendChangeCipher()` is now compiled for a TLS 1.3 server (or a middlebox-compat client), where before it went missing from a TLS 1.3-only build such as `examples/configs/user_settings_tls13.h`.
- Under TLS 1.3 the record no longer moves `serverState` in the `OPENSSL_EXTRA` block. That ran before any send could fail, leaving the server not waiting for the second ClientHello.
- A HelloRetryRequest is flushed as it is built, so the record behind it needs its own flush; without one a server grouping its messages sent a bare HelloRetryRequest.
- `wolfSSL_clear()` resets the flag, or a reused server answers only its first peer.
- The client sites take the same ordering, and the one answering a HelloRetryRequest keeps `buildingMsg` set when the send stops short; without that a refused write stranded the client handshake.

DTLS stays excluded, QUIC clears the flag per RFC 9001 Section 8.4, and with ECH the outer ClientHello decides since the inner hello is rebuilt carrying its session id. The configure and CMake option text now says client-side, since only that half stays behind the macro.

## Behavioral change

A TLS 1.3 server now sends a ChangeCipherSpec it did not before, whenever the client offered a non-empty `legacy_session_id`; an empty one still gets none. One API difference: `wolfSSL_get_state()` no longer passes through `WOLFSSL_SS_SERVER_CHANGECIPHERSPEC` or `WOLFSSL_SS_CLIENT_CHANGECIPHERSPEC` on TLS 1.3, though the info callback still fires. What a client puts on the wire is unchanged. Written up in `ChangeLog.md`.

## Testing

Eight tests in `tests/api/test_tls13.c`, seven of them ungated since default-build behaviour is the point: a non-empty session id gets exactly one record, an empty one gets none (#9156), a cleared server still answers its second peer, the HelloRetryRequest flight carries the record grouped and ungrouped while the flight after it carries none, and a refused-then-resumed write leaves exactly one on the wire on the straight-through, retry and client paths.

Each was verified to fail without its fix: reverting the source fails the two server tests, dropping the grouped-messages flush fails the HelloRetryRequest test, marking `sentChangeCipher` after a successful send fails the server retry test, and reverting the client HelloRetryRequest site fails the client test with the handshake never completing.

Clean in `--enable-all`, a default build, `--enable-jni`, `--enable-tls13-middlebox-compat --enable-opensslextra --enable-hrrcookie`, `--enable-ech --enable-quic --enable-hrrcookie`, a client-only TLS 1.3 build, `user_settings_tls13.h`, and CMake. `make check` passes on `--enable-all`.

## Notes for reviewers

Two pre-existing things this surfaces but does not fix:

- **ECH**: `ssl->session->sessionIDSz` is reassigned on every ClientHello parse including the inner one, so with ECH accepted and a non-empty outer session id the ServerHello echoes an empty `legacy_session_id_echo` alongside the record. Predates this change, a conformant client rejects on that mismatch alone, and it is unreachable wolfSSL-to-wolfSSL.
- **Sniffer**: `sniffer.c`'s `change_cipher_spec` case sets `handShakeDone` with no TLS 1.3 exclusion, but the branch above it sets `serverCipherOn = 0` when `srvKs.key_len == 0`, so the recovery path it looks like it would trip never fires. Checked against a real capture carrying both a server and a client ChangeCipherSpec.


Frauschi:
Jenkins retest this please.

philljj (changes requested):
Skoll had a testing coverage nit (see below), otherwise ran clean.

Would like @julek-wolfssl to review this because he also just worked on middlebox-compat.

### MEDIUM-1: Exercise the changed early-data CCS retry path
- **File:** `src/tls13.c:15778-15784`
- **Function:** `wolfSSL_connect_TLSv13`

**Description:** The PR changes the early-data path to guard and set `sentChangeCipher` before calling `SendChangeCipher()`, but the new client retry test covers only ordinary and HelloRetryRequest handshakes. It never enables early data or enters this branch. Moreover, the existing early-data WANT_WRITE callback in `tests/api/test_tls13.c` deliberately lets a standalone CCS write pass, so the default non-grouped configuration cannot detect a duplicate or stranded CCS here.

**Code:**
```
if (!ssl->options.dtls && !ssl->options.sentChangeCipher
        && ssl->options.tls13MiddleBoxCompat) {
    ssl->options.sentChangeCipher = 1;
    if ((ssl->error = SendChangeCipher(ssl)) != 0) {
        return WOLFSSL_FATAL_ERROR;
    }
}
```

**Recommendation:** Cover this third modified client send site with the same retry assertions already applied to the normal and HelloRetryRequest paths.

Frauschi:
@philljj fixed your finding.

Frauschi:
Jenkins retest this please

philljj:
Might be a real unit test failure:

```
FAILURES:
   1686: test_tls13_early_data_0rtt_replay
 End API Tests
 Failed/Skipped/Passed/All: 1/396/1942/2339
FAIL scripts/unit.test (exit status: 1)
```

Frauschi:
I cannot reproduce it. Jenkins retest this please.

src/internal.c
```diff
@@ -26028,13 +26023,17 @@ int SendChangeCipher(WOLFSSL* ssl)
...
         if (ssl->CBIS != NULL)
             ssl->CBIS(ssl, WOLFSSL_CB_ACCEPT_LOOP, WOLFSSL_SUCCESS);
     }
     else {
-        ssl->options.clientState =
-            CLIENT_CHANGECIPHERSPEC_COMPLETE;
+        /* As above; nothing on the TLS 1.3 connect path reads clientState. */
+        if (!IsAtLeastTLSv1_3(ssl->version))
+            ssl->options.clientState =
+                CLIENT_CHANGECIPHERSPEC_COMPLETE;
```
julek-wolfssl: multi-line `if` requires `{}`.

src/tls13.c
```diff
@@ -15765,13 +15775,13 @@ int wolfSSL_connect_TLSv13(WOLFSSL* ssl)
...
-                    if (!ssl->options.dtls &&
-                           ssl->options.tls13MiddleBoxCompat) {
+                    if (!ssl->options.dtls && !ssl->options.sentChangeCipher
+                            && ssl->options.tls13MiddleBoxCompat) {
+                        ssl->options.sentChangeCipher = 1;
                         if ((ssl->error = SendChangeCipher(ssl)) != 0) {
                             WOLFSSL_ERROR(ssl->error);
                             return WOLFSSL_FATAL_ERROR;
                         }
-                        ssl->options.sentChangeCipher = 1;
```
julek-wolfssl, line 15807: Why is `ssl->options.sentChangeCipher = 1` set before `SendChangeCipher` now?

src/tls13.c
```diff
@@ -15817,11 +15827,15 @@ int wolfSSL_connect_TLSv13(WOLFSSL* ssl)
...
                     if ((ssl->error = SendChangeCipher(ssl)) != 0) {
+                        /* The second ClientHello still has to follow, so hold
+                         * the state machine on this case while the record
+                         * drains. */
+                        ssl->options.buildingMsg = 1;
                         WOLFSSL_ERROR(ssl->error);
                         return WOLFSSL_FATAL_ERROR;
                     }
-                    ssl->options.sentChangeCipher = 1;
                 }
```
julek-wolfssl, line 15858: again

src/tls13.c
```diff
@@ -15887,11 +15901,11 @@ int wolfSSL_connect_TLSv13(WOLFSSL* ssl)
         #if defined(WOLFSSL_TLS13_MIDDLEBOX_COMPAT)
             if (!ssl->options.sentChangeCipher && !ssl->options.dtls
                 && ssl->options.tls13MiddleBoxCompat) {
+                ssl->options.sentChangeCipher = 1;
                 if ((ssl->error = SendChangeCipher(ssl)) != 0) {
                     WOLFSSL_ERROR(ssl->error);
                     return WOLFSSL_FATAL_ERROR;
                 }
-                ssl->options.sentChangeCipher = 1;
```
julek-wolfssl, line 15927: again

julek-wolfssl (changes requested):

src/internal.c
Frauschi: Fixed

src/tls13.c
Frauschi: Fixed all occurrences by moving it after the call, setting it only on success and on WANT_WRITE as it is already queued in this case.

src/tls13.c
Frauschi: See above

src/tls13.c
Frauschi: See above

julek-wolfssl (dismissed):
@Frauschi for the future push review fixes in new commits instead of amending.

julek-wolfssl:
@Frauschi merge conflict + jenkins

Frauschi:
Jenkins retest this please

julek-wolfssl (dismissed):

philljj (dismissed):
merge conflict in changelod md

Frauschi:
Jenkins retest this please.

Frauschi:
Jenkins retest this please

JacobBarthelmeh (approved):

PR #11310, opened 2026-08-28
ECH: Take the AAD from the parsed ClientHello offset

gasbytes:
# Description

Sets the ECH AAD base to input + args->begin and the inner ClientHello re-entry index to that hello's reserved handshake-header size (in the case of DTLS, we set DTLS13_HANDSHAKE_HEADER_SZ), and adds an ECH handshake test that runs with read-ahead enabled on the server.

# Testing

Added associated regression test `test_wolfSSL_Tls13_ECH_read_ahead`.

# Checklist

 - [x] added tests
 - [ ] updated/added doxygen


gasbytes:
Jenkins retest this please

SparkiDev (approved):

PR #11318, opened 2026-08-31
Fix the libssh2 test retry

Frauschi:
The libssh2 PRB fails intermittently, always with a single random test out of 46 reporting `libssh2_session_handshake failed (-43): Failed getting banner`. Recent examples: runs 33378491609, 33222310715, 33215954763 and 33215953834, which failed on four different tests.

The retry meant to absorb this never ran. `make recheck` was invoked from the libssh2 repo root, but automake generates `recheck` only in the directory that defines `TESTS`, which for libssh2 is `tests/` alone. Every flaky run ended with `make: *** No rule to make target 'recheck'. Stop.` printed twice and failed the job on the first transient error. This runs it as `make -C tests recheck`.

The step comment also described the race incorrectly. `-43` is `LIBSSH2_ERROR_SOCKET_RECV`, not `LIBSSH2_ERROR_BANNER_RECV` (-2); "Failed getting banner" is just the wrapper message around any `banner_receive()` error. The real failure is an ECONNRESET: docker-proxy binds the published port as soon as the container starts, so `connect()` succeeds and sshd resets the connection because it is not listening inside the container yet. libssh2 cannot recover on its own here, since its fixture retries only `connect()`, which never fails in this situation.

Finally, `make -j` was unbounded. The failing runs had roughly 20 containers finishing inside a 1.3 second window on a four-CPU runner, which is what starves sshd of the time it needs to come up, so the parallelism is now capped at `nproc`.

This is a mitigation rather than a cure, and deliberately so. Both root-cause fixes were investigated and dropped, because each needs changes inside the libssh2 checkout rather than in our own workflow: adding a handshake retry to the test fixture (`tests/session_fixture.c`), or pointing every test at a single pre-warmed sshd. The second is also not the drop-in it looks like, since `OPENSSH_NO_DOCKER` does not merely select an external server - it additionally switches the fixture to authenticate as `$USER` instead of the container's hardcoded `libssh2` account (`tests/test_read.c:39`, `tests/session_fixture.c:475`), so every pubkey test would fail against a container.


Frauschi:
Jenkins retest this please

philljj (approved):

PR #11319, opened 2026-08-31
Fix sniffer decryption, threaded snifftest, and the shared-state teardown

Frauschi:
## Description

Nine sniffer fixes. Three of them mean whole classes of capture could not be decrypted at all, and one means a sniffer build could not complete a DTLS 1.3 handshake at all.

**Encrypt-Then-MAC was never implemented.** `EXT_ENCRYPT_THEN_MAC` was defined in `src/sniffer.c` but never handled in the `ProcessServerHello` extension switch, so `startedETMRead` was never set and the trailing MAC was passed to `wc_AesCbcDecrypt()` along with the ciphertext. The length is then not a block multiple and every record fails with `BAD_FUNC_ARG`. wolfSSL peers negotiate RFC 7366 by default, so no TLS 1.2 CBC capture between wolfSSL peers could be decrypted. A build without Encrypt-Then-MAC support now says so through a new error instead of failing every record generically; the complaint is deferred until the suite is known, since RFC 7366 covers only block ciphers.

**Reported plaintext lengths included the MAC or AEAD tag.** The length came from the record size without removing `ssl->keys.padSz`, so 14 bytes were reported as 30 under TLS 1.2 AES-GCM. The plaintext was correct, which is why this survived. **For `WOLFSSL_SNIFFER_STORE_DATA_CB` this is a heap buffer overread** - the callback pointer already sits at the plaintext start, so the extra bytes run off the decrypt output buffer. ASan reports a `READ of size 68` past the allocation before the fix, clean after.

**The X25519 static ephemeral key had no RNG.** curve25519 blinding is on by default for the C implementation and draws from the key's own RNG, but the sniffer built its key with `wc_curve25519_init_ex()` and `wc_Curve25519PrivateKeyDecode()`, neither of which sets one. `wc_RNG_GenerateBlock()` then returned `BAD_FUNC_ARG` with every argument valid, and no X25519 traffic could be read. The library's own static ephemeral path and the sniffer's ECC path already called the matching setter; only X25519 was missed.

The rest:

- **A specific error was reported as a key mismatch.** The `client_key_exchange` handler and the TLS 1.3 ServerHello path overwrote whatever `SetupKeys()` had reported with the generic "Server Client Key Mismatch", discarding the real reason every time. This is what made the X25519 failure look like a key problem.
- **`snifftest` could not report a failed capture.** The read loop assigned `hadBadPacket` instead of accumulating it, so an early error was erased by any later clean packet. A new opt-in `-expectdata` additionally fails a run that decrypted nothing at all.
- **`THREADED_SNIFFTEST` decoded nothing.** The worker checked its shutdown flag before draining its queue, and reading a capture file sets that flag before the worker is first scheduled. Separately, the server and secret tables are thread-local, so workers had to repeat `main()`'s keylog setup and never did.
- **`ssl_FreeSniffer()` tore down process-wide state from every thread.** The trace file, mutexes and crypto device are shared, but every call released them, so a worker finishing early closed the trace file under threads still running. The init entry points now count their callers and only the last free releases the shared state.
- **A sniffer build could not complete a DTLS 1.3 handshake.** With `WOLFSSL_SNIFFER` defined, the example client and server pin themselves to a static RSA and static ECC cipher list so a capture can be decrypted. The guard for that read `version < 4`, meant to leave TLS 1.3 alone, but the examples encode a DTLS version as a negative number and DTLS 1.3 is `-4`, so a `-u -v 4` run was handed a TLS 1.2 only cipher list. The server then found no common TLS 1.3 suite in `MatchSuite_ex()` while processing the ClientHello and failed with `MATCH_SUITE_ERROR`, reaching the client as a `missing_extension` alert. This also took out the DTLS cases of `scripts/ocsp-stapling_tls13multi.test`. Only the examples were affected; an application that does not set that cipher list was always able to handshake.
- **The example store-data callback overflowed on a multi-record packet.** It sized its buffer from the first record while the callback offset restarts at zero for each one.

## Testing

Nothing failed before because three things swallowed it: `hadBadPacket` was assigned rather than accumulated, the keylog leg read `tee`'s exit status instead of snifftest's, and every other leg checked only the exit code, never the lengths. On top of that both `sniffer-tls13-x25519*.pcap` were checked in at 24 bytes - a bare pcap header with no packets - so those legs had always tested nothing.

- New `sniffer-tls12-etm.pcap` and `-etm-keylog.pcap`, with a `tls12-etm` branch in `sniffer-gen.sh`. Two suites deliberately: HMAC-SHA1 makes the pre-fix decrypt fail outright, HMAC-SHA256 silently decrypts the wrong byte range.
- Both X25519 captures regenerated and non-empty; `sniffer-static-rsa.pcap` and `sniffer-ipv6.pcap` regenerated Encrypt-Then-MAC-free, so IPv6 is no longer gated on an unrelated feature.
- The Encrypt-Then-MAC and keylog legs now diff reported lengths against a stored `.out` oracle. Reverting the `padSz` fix makes the EtM leg report 36/52/48/64 where 14/22/14/22 are correct.
- `sniffer-gen.sh` regenerates the oracles itself, takes capture names, and fails on a name that matched nothing. Six keylog assets had never been in any `.am`, so `make dist` shipped the test but not its captures; fixed.

Three CI entries added to `.github/configs/os-check-linux.json`, each covering a path no other entry compiles: `sniffer-keylog-storedata` (`WOLFSSL_SNIFFER_KEYLOGFILE` and `WOLFSSL_SNIFFER_STORE_DATA_CB`), `sniffer-no-enc-then-mac` (the new diagnostic and its assertion), and `sniffer-threaded` (`THREADED_SNIFFTEST`, session tickets off so the concurrent path is actually exercised).

The DTLS 1.3 fix is verified separately, since no CI entry reaches it - the only `--enable-sniffer` leg has no `--enable-dtls13`, so `ocsp-stapling_tls13multi.test` skips its DTLS cases there. Built `--enable-all --enable-debug --enable-sniffer`: before the fix a plain `server -u -v 4` / `client -u -v 4` pair fails deterministically and `ocsp-stapling_tls13multi.test` fails at `TEST CASE DTLS-1`; after it, both pass, TLS 1.3 is unchanged, and `sniffer-testsuite.test` still passes.


src/sniffer.c
```diff
@@ -232,6 +232,71 @@ BOOL APIENTRY DllMain( HMODULE hModule,
...
+/* Take a reference to the shared state.
+   returns 1 when this caller is the one that has to set it up */
+static int SnifferInitAcquire(void)
+{
+    if (SNIFFER_INIT_REF_INC() == 1)
+        return 1;
+
+#ifndef WOLFSSL_MUTEX_INITIALIZER
+    /* Another caller got there first; wait for it to finish. */
+    while (!SharedInitDone) {
```
Frauschi: fixed

Frauschi: fixed

src/sniffer.c
```diff
@@ -3823,6 +3969,15 @@ static int ProcessServerHello(int msgSz, const byte* input, int* sslBytes,
         return WOLFSSL_FATAL_ERROR;
     }
 
+    /* Encrypt-Then-MAC has to be re-negotiated in every ServerHello, so drop
+     * any earlier decision before parsing this one. */
+#if defined(HAVE_ENCRYPT_THEN_MAC) && !defined(WOLFSSL_AEAD_ONLY)
+    session->sslServer->options.startedETMRead = 0;
```
Frauschi: fixed

dgarske (approved):

PR #11323, opened 2026-08-31
Fil-C improvements

Frauschi:
### Summary

Three commits. The first is an independent correctness fix; the other two are the Fil-C CI rework and what it exposed.

**1. `sp_div_word` rdx clobber.** `divq` and `divl` write the remainder to `dx`, but `dx` was listed only as an input and was neither an output nor a clobber, so the compiler may assume it still holds `hi` after the asm block. This is a latent miscompile in every x86_64 and x86 build using SP math, which is the default. Fil-C's inline assembly validation refuses to run it, which is how it surfaced.

**2. Fixed size SP code on capability targets.** Regenerated `sp_c32.c` and `sp_c64.c`. Comes from https://github.com/wolfSSL/scripts/pull/667 and must merge with it, or the next regeneration reverts these files silently.

**3. Fil-C job.** The job pinned v0.674 and ran two configurations, and its release pattern matches both the x86_64 and the aarch64 tarball, so it would break on any release from v0.682 on. Moves to v0.684, picks the tarball by `uname -m`, and grows the matrix to eighteen legs.

It also passed `-DWC_NO_CACHE_RESISTANT`, which compiled out the cache resistant paths, so CI was not testing the code that ships. Those paths build a pointer by masking and adding two addresses, which a capability target cannot do. `WC_NO_PTR_INT_CAST` already exists for this and is defined for CHERI purecap, so it is now defined for Fil-C too and the flag is dropped.

That exposed three places the macro did not cover:

* `tfm.c` never honored it. New `fp_cond_copy` replaces the masked address selections in the timing resistant exptmod ladders.
* `RsaDec` and `DoClientKeyExchange` selected a pointer with `ctMaskCopy`, which copies byte by byte and so drops the capability. This fires on the success path, not only on error. New `ctMaskSelPtr` in `misc.c` indexes a two entry table aligned so both candidates share a cache line, and still does not branch on the decryption result.
* `sp_int.c`'s x86_64 assembly uses `"rm"` and `"m"` operands, which Fil-C does not allow. These now route through `SP_ASM_RM` and `SP_ASM_M`, which differ only under `__FILC__`. Verified byte identical object code on other targets.

### Testing

`make check` passes for `--enable-all`, and for `--enable-sp`, `--enable-sniffer` and `--enable-fastmath` built with `-DWC_NO_PTR_INT_CAST`. The sniffer configuration is the one that reaches both `ctMaskSelPtr` call sites, through static RSA.

The rdx fix was exercised on an x86_64 build with assembly enabled, confirming `divq` is still emitted and `testwolfcrypt` passes.

### Notes for review

`--disable-asm` stays on most legs because Fil-C cannot link a `.S` file at all: it gives every function a capability carrying entry point that a raw asm symbol lacks. `--enable-intelasm` fails on 227 symbols, most of them ML-KEM and ML-DSA. The `default, asm` and `default, asm, arm64` legs do build the inline assembly in `sp_int.c`, which is what most users compile.

The `fp_exptmod_nb` ladder is still uncovered. It needs `--enable-rsa=nonblock`, which fails together with `--enable-fastmath` on master with plain gcc as well.

### DIFF
--- a/.github/workflows/fil-c.yml
+++ b/.github/workflows/fil-c.yml
@@ -1,7 +1,7 @@
 name: Fil-C Tests
 
 env:
-  FIL_C_VERSION: v0.674
+  FIL_C_VERSION: v0.684
 
 # START OF COMMON SECTION
 on:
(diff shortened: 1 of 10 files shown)


Frauschi:
Jenkins retest this please

Frauschi:
Jenkins retest this please - flaky test.

SparkiDev (approved):

PR #11324, opened 2026-08-31
Fully disabled TLS 1.2

Frauschi:
`--disable-tlsv12` never compiled TLS 1.2 out. The option set `ENABLED_TLSV12=no`, which the configure summary printed and a couple of derived settings read, but nothing defined `WOLFSSL_NO_TLS12`, so the whole TLS 1.2 implementation was still built and a peer could still negotiate it. The single place `configure.ac` adds that define sits in the FIPS lean-aesgcm bundle, behind a condition that is false exactly when the user asked for the option, so no autotools configuration ever reached it. Only the CMake `WOLFSSL_TLSV12=no` path honoured the switch.

This makes the option do what it says, and fixes what that exposed.

### Behavioral change

`--disable-tlsv12` now defines `WOLFSSL_NO_TLS12`. A build that passes it and still expects to negotiate TLS 1.2 has to stop passing it. The define is added after the bundles have had their say, so `--enable-tinytls13` and the other bundles that turn the version off pick it up too.

Because the pre-TLS-1.3 handshake is now actually compiled out, configure rejects the four combinations that depend on it rather than producing a build that cannot work:

* `--enable-oldtls`, since TLS 1.0 and 1.1 use the TLS 1.2 handshake
* TLS 1.3 disabled, which would leave no version to negotiate
* `--enable-mcast`, which rides on DTLS 1.2 and its NULL cipher suite
* DTLS without `--enable-dtls13`, for the same reason on the datagram side

CMake rejects the same four for `-DWOLFSSL_TLSV12=no`. It previously accepted `-DWOLFSSL_OLD_TLS=yes` alongside it and defined `NO_OLD_TLS` anyway, so the reported option and the build disagreed.

The sniffer, the examples and the test suite now pick their code paths by version where they used to assume TLS 1.2 was present. Four `no-tlsv12-*` entries were added to `.github/configs/os-check-linux.json` covering the option on its own axes: OCSP stapling plus the OpenSSL compatibility layer, status_request_v2 without status_request, DTLS 1.3 with connection ID and session tickets, and the sniffer.

### Three bugs this uncovered

These are independent of the TLS 1.2 work and are kept as their own commits, so they can be split out if preferred. All three reproduce on master.

**PK callbacks could not complete a TLS 1.3 handshake.** `myEccKeyGen()` generated the key share straight into the library's `ecc_key`, while `myEccSharedSecret()` is handed only the peer's key on the client side, so the client had no private key left to reach: `wolfSSL_connect error -170, ECC input argument wrong type`. The example callbacks now keep the private half application-side for TLS and DTLS 1.3, which is what `TEST_PK_PRIVKEY` already modelled and what a real PK-callback application has to do anyway. `scripts/pkcallbacks.test` passed no `-v`, so it always negotiated TLS 1.2 and never covered this; it now runs the build's default plus every version the build reports.

**Chain OCSP stapling failed the handshake when a responder was unreachable.** The chain loops suppress the responder errors that mean "no status to staple", but `OCSP_INVALID_STATUS` was missing from that set, and it is what `CheckOcspRequest()` returns when it cannot reach the responder at all. A TLS 1.3 server built with `WOLFSSL_TLS_OCSP_MULTI` therefore dropped every handshake whenever the responder for any intermediate was down, having already obtained the leaf status the peer asked about. The leaf path in `CreateOcspResponse()` deliberately still fails hard, and `OCSP_CERT_REVOKED` stays fatal everywhere.

**`dtls_bench` did not build without one side.** `examples/benchmark/dtls_bench.c` called `wolfSSL_accept()` and `wolfSSL_connect()` unconditionally, so a `NO_WOLFSSL_SERVER` build failed to link. That is the existing `dtls13-client-minimal` os-check entry, which passes `CPPFLAGS=-DNO_WOLFSSL_SERVER` and gets the benchmark built because `os-check.yml` does not pass `--disable-examples`.

### Testing

Built and `make check`ed clean (0 failures) on macOS/arm64 with clang, each with
`-pedantic -Wdeclaration-after-statement -Wnull-dereference -Wno-overlength-strings -Werror`:

| Configuration | Covers |
|---|---|
| `--disable-tlsv12 --enable-ocspstapling --enable-opensslextra` | the option plus the two axes that reach TLS 1.2-only code from a TLS 1.3 build |
| `--enable-tls13 --disable-tlsv12 --enable-ocspstapling2` (`-Wall -Wextra -Werror`) | status_request_v2 without v1, the one stapling shape with no TLS 1.3 chain-status path |
| `--disable-tlsv12 --enable-dtls --enable-dtls13 --enable-dtlscid --enable-session-ticket` | DTLS 1.3 both sides with the TLS 1.2 handshake gone |
| `--enable-sniffer --disable-tlsv12` | the sniffer's TLS 1.3-only paths; `snifftest` reports `tls_v13` without `tls_v12` and decodes the ecc, dh and hrr pcaps |
| `--enable-tinytls13` | picks up the define through the bundle |
| `--enable-all --disable-mcast --disable-oldtls --disable-tlsv12` | the widest no-TLS-1.2 surface |
| `--enable-all --disable-mcast --disable-oldtls` | TLS 1.2 still on, no regression |
| `--enable-all --enable-pkcallbacks`, with and without `-DTEST_PK_PRIVKEY` | both PK-callback key-ownership models |

Rejection paths checked by hand: configure errors on `--disable-tlsv12` with each of `--enable-oldtls`, `--disable-tls13`, `--enable-mcast` and `--enable-dtls` without dtls13; CMake `FATAL_ERROR`s on `WOLFSSL_OLD_TLS=yes`, `WOLFSSL_TLS13=no`, `WOLFSSL_MCAST=yes` and `WOLFSSL_DTLS=yes` without `WOLFSSL_DTLS13`, and accepts the valid combinations.

For the PK-callback fix specifically: the new `-v 4` leg of `scripts/pkcallbacks.test` fails against pre-fix binaries and passes after, and the HelloRetryRequest re-keygen path (`server -Y`) and a two-curve client both complete. For the OCSP fix, a revoked server certificate is still rejected with `-360 OCSP Cert revoked`, so the suppression does not mask revocation.


Frauschi:
Jenkins retest this please

SparkiDev (dismissed):

wolfssl/test.h
```diff
@@ -3420,47 +3420,123 @@ static WC_INLINE int wolfSSL_PrintStatsConn(WOLFSSL_MEM_CONN_STATS* stats)
...
+    if (cbInfo != NULL && myEccKeepPrivKey(ssl)) {
+        /* A key we already kept for this curve is stale - a TLS v1.3 client
+         * regenerates its key share when the server names a group in a
+         * HelloRetryRequest - so release it and take its slot back. */
+        new_key = myEccKeptKey(cbInfo, ecc_curve);
+        if (new_key != NULL) {
+            wc_ecc_free(new_key);
+        }
+        else if (cbInfo->keyGenCnt < PKCB_MAX_ECC_KEYGEN) {
+            new_key = (ecc_key*)XMALLOC(sizeof(ecc_key), NULL,
```
Frauschi, line 3527: fixed.

wolfssl/test.h
```diff
@@ -3666,12 +3745,7 @@ static WC_INLINE int myEccSharedSecret(WOLFSSL* ssl, ecc_key* otherKey,
     #endif
     }
 
-#ifdef TEST_PK_PRIVKEY
-    if (cbInfo && cbInfo->hasKeyGen) {
-        wc_ecc_free(&cbInfo->keyGen.ecc);
-        cbInfo->hasKeyGen = 0;
-    }
-#endif
+    myEccFreeKeptKeys(cbInfo);
```
Frauschi: fixed

philljj (changes requested):
merge conflict in src/sniffer.c

CMakeLists.txt
```diff
@@ -2571,6 +2565,34 @@ if (WOLFSSL_TLS13)
         )
 endif()
 
+if(NOT WOLFSSL_TLSV12)
```
philljj: I think this needs a `NOT WOLFSSL_CRYPT_ONLY` guard (or similar) to prevent NO_TLS / wolfcrypt only builds from failing

philljj (dismissed):
Looks good, just one nit so far.

Would also like @julek-wolfssl review

CMakeLists.txt
Frauschi: Fixed. This also uncovered another issue with the CMake build system that is now fixed in the latest commit.

julek-wolfssl (approved):

JacobBarthelmeh (approved):

PR #11327, opened 2026-08-31
Harden wc_EncryptPKCS8Key_ex to check return val

embhorn:
# Description

Fixes the ignored SetShortInt() error; the local NULL check makes the invariant explicit and unblocks -Werror=nonnull. Return values are unchanged on every path, so this isn't a behavior change.

Fixes zd22390

# Testing

Customer confirmed

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


embhorn:
retest this please

SparkiDev (approved):

PR #11330, opened 2026-09-01
ssl.c: clear SSL-owned credentials when switching CTX

yosuke-wolfssl:
### Problem
`wolfSSL_set_SSL_CTX()` only wrote an `ssl->buffers` slot when the replacement CTX supplied a value, so switching to a CTX without credentials left the previous ones in place. This is the API an SNI/ClientHello callback uses to pick a per-host identity, so what stays behind is what goes on the wire.

- **With `WOLFSSL_COPY_CERT`** (implied by `OPENSSL_ALL`): a CTX with a leaf but no chain kept the previous CTX's chain, and TLS 1.2 sent it alongside the new leaf. Under `WOLFSSL_BLIND_PRIVATE_KEY` the old private key survived too, so the handshake completed under the previous identity. (f-12567)
- **Without it** (`--enable-opensslextra`): `ssl->buffers.certificate = ctx->certificate` ran without clearing `weOwnCert`, so an SSL holding its own certificate freed the CTX's `DerBuffer` at `wolfSSL_free()` — double free at `wolfSSL_CTX_free()`.

### Fix (`src/ssl.c`)
- **`wolfSSL_certs_clear()`** releases `ssl->ourCert` and `ssl->ourCertChain` with the buffers they mirror, and zeroes `certChainCnt` in every build. This also fixes a leak through the public `SSL_certs_clear()`, which cleared `weOwnCert` — the flag teardown uses to decide whether to free `ourCert`.
- **`wolfSSL_set_SSL_CTX()`** calls it before copying from the new CTX, so each buffer is released exactly once and a credential-less CTX leaves nothing behind.
- **Alt-key ownership**: `weOwnAltKey` is set after the copy in `wolfSSL_set_SSL_CTX()` and `SetSSL_CTX_CertsAndKeys()` (`src/internal.c`), and `SendTls13CertificateVerify()` (`src/tls13.c`) clears it when it moves the alt key into `buffers.key`. Without that transfer both fields own one buffer and teardown frees it twice.

Gives the same semantics as OpenSSL, which replaces the whole `CERT` with a dup of the new context's. Closes f-12567.

### Tests (`tests/api.c`)

| Test | Covers |
|---|---|
| `test_wolfSSL_set_SSL_CTX` | Switches to a CTX with no chain and to one with no credentials; buffers, `weOwn*` flags, `certChainCnt`, `wolfSSL_get_certificate()`; an SSL owning its certificate freed before the CTX |
| `test_wolfSSL_certs_clear` | Same state after `SSL_certs_clear()`, including `ourCert` and `ourCertChain` |
| `test_dual_alg_support`, `test_dual_alg_ctx_alt_key` | ALTERNATIVE-signature connection; alt key copied and owned via both `wolfSSL_new()` and a CTX switch |

### Verification
- `--enable-opensslall`: full `./tests/unit.test` passes. `--enable-opensslextra` and `--disable-tls13` builds pass.
- Dual-alg + `WOLFSSL_BLIND_PRIVATE_KEY` under ASan + UBSan: clean.
- Negative controls per fix: without the clear, the stale-chain assertions fail and ASan reports heap-use-after-free in `FreeDer()`; without the `tls13.c` line, ASan reports use-after-free in `wolfSSL_UnloadCertsKeys()`.
- Note: no CI job enables `--enable-dual-alg-certs` with the unit tests, so the alt-key coverage above was verified locally.

### DIFF
--- a/src/ssl.c
+++ b/src/ssl.c
@@ -8409,6 +8409,13 @@ void wolfSSL_certs_clear(WOLFSSL* ssl)
     /* ctx still owns certificate, certChain, key, dh, and cm */
     if (ssl->buffers.weOwnCert) {
         FreeDer(&ssl->buffers.certificate);
+    #ifdef KEEP_OUR_CERT
+        /* ourCert is only made when this SSL owns the certificate. */
+        if (ssl->ourCert != NULL) {
+            wolfSSL_X509_free(ssl->ourCert);
+            ssl->ourCert = NULL;
+        }
+    #endif
         ssl->buffers.weOwnCert = 0;
     }
     ssl->buffers.certificate = NULL;
@@ -8417,8 +8424,12 @@ void wolfSSL_certs_clear(WOLFSSL* ssl)
         ssl->buffers.weOwnCertChain = 0;
     }
     ssl->buffers.certChain = NULL;
-#ifdef WOLFSSL_TLS13
     ssl->buffers.certChainCnt = 0;
+#ifdef KEEP_OUR_CERT
+    if (ssl->ourCertChain != NULL) {
+        wolfSSL_sk_X509_pop_free(ssl->ourCertChain, NULL);
+        ssl->ourCertChain = NULL;
+    }
 #endif
     if (ssl->buffers.weOwnKey) {
         FreeDer(&ssl->buffers.key);
@@ -8828,33 +8839,26 @@ WOLFSSL_CTX* wolfSSL_set_SSL_CTX(WOLFSSL* ssl, WOLFSSL_CTX* ctx)
     ssl->ctx = ctx;
 
 #ifndef NO_CERTS
+    /* Release the certificate, chain and keys this SSL owns so that a ctx
+     * without them leaves none of the old ones behind. */
+    wolfSSL_certs_clear(ssl);
 #ifdef WOLFSSL_COPY_CERT
     /* If WOLFSSL_COPY_CERT defined, always make new copy of cert from ctx */
     if (ctx->certificate != NULL) {
-        if (ssl->buffers.certificate != NULL) {
-            FreeDer(&ssl->buffers.certificate);
-            ssl->buffers.certificate = NULL;
-        }
         ret = AllocCopyDer(&ssl->buffers.certificate, ctx->certificate->buffer,
             ctx->certificate->length, ctx->certificate->type,
             ctx->certificate->heap);
         if (ret != 0) {
-            ssl->buffers.weOwnCert = 0;
             return NULL;
         }
 
         ssl->buffers.weOwnCert = 1;
     }
     if (ctx->certChain != NULL) {
-        if (ssl->buffers.certChain != NULL) {
-            FreeDer(&ssl->buffers.certChain);
-            ssl->buffers.certChain = NULL;
-        }
         ret = AllocCopyDer(&ssl->buffers.certChain, ctx->certChain->buffer,
             ctx->certChain->length, ctx->certChain->type,
             ctx->certChain->heap);
         if (ret != 0) {
-            ssl->buffers.weOwnCertChain = 0;
             return NULL;
         }
 
@@ -8865,20 +8869,14 @@ WOLFSSL_CTX* wolfSSL_set_SSL_CTX(WOLFSSL* ssl, WOLFSSL_CTX* ctx)
     ssl->buffers.certificate = ctx->certificate;
     ssl->buffers.certChain = ctx->certChain;
 #endif
-#ifdef WOLFSSL_TLS13
     ssl->buffers.certChainCnt = ctx->certChainCnt;
-#endif
 #ifndef WOLFSSL_BLIND_PRIVATE_KEY
 #ifdef WOLFSSL_COPY_KEY
-    if (ssl->buffers.key != NULL && ssl->buffers.weOwnKey) {
-        FreeDer(&ssl->buffers.key);
-    }
     if (ctx->privateKey != NULL) {
         ret = AllocCopyDer(&ssl->buffers.key, ctx->privateKey->buffer,
             ctx->privateKey->length, ctx->privateKey->type,
             ctx->privateKey->heap);
         if (ret != 0) {
-            ssl->buffers.weOwnKey = 0;
             return NULL;
         }
         ssl->buffers.weOwnKey = 1;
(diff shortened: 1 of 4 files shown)


src/ssl.c
```diff
@@ -8828,33 +8828,37 @@ WOLFSSL_CTX* wolfSSL_set_SSL_CTX(WOLFSSL* ssl, WOLFSSL_CTX* ctx)
...
+        ssl->ourCert = NULL;
+    }
+    if (ssl->ourCertChain != NULL) {
+        wolfSSL_sk_X509_pop_free(ssl->ourCertChain, NULL);
+        ssl->ourCertChain = NULL;
+    }
+#endif
+    /* Release the certificate, chain and keys this SSL owns so that a ctx
+     * without them leaves none of the old ones behind. */
+    wolfSSL_certs_clear(ssl);
```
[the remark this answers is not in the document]
yosuke-wolfssl: Fixed. The root cause was `wolfSSL_certs_clear()` zeroing `certChainCnt` only under `WOLFSSL_TLS13`, while `wolfSSL_clear_chain_certs()` zeroes it unconditionally and the field itself isn't TLS 1.3 gated. Dropped that guard, so the count is cleared together with the chain on every early return, and `wolfSSL_set_SSL_CTX()` now assigns it in all builds as well.

tests/api.c
```diff
@@ -30867,14 +30868,65 @@ static int test_wolfSSL_set_SSL_CTX(void)
...
+#endif
+#ifdef KEEP_OUR_CERT
+    ExpectNull(wolfSSL_get_certificate(ssl));
+#endif
+
+    wolfSSL_free(ssl);
+    ssl = NULL;
+
+    /* An ssl holding a certificate of its own must not end up owning the one
+     * of ctx1: ctx1 is used again after the ssl is free'd. */
```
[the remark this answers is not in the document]
yosuke-wolfssl: Fixed.

src/ssl.c
yosuke-wolfssl, line 8844: Confirmed — this one was a regression from the new `wolfSSL_certs_clear()` call, since `wolfssl_priv_der_blind()` used to free the old mask. Both `wolfSSL_set_SSL_CTX()` and `SetSSL_CTX_CertsAndKeys()` now set `weOwnAltKey` after the copy, so `altKey` and `altKeyMask` are both freed.

Setting the flag alone was not safe, though. `SendTls13CertificateVerify()` moves `altKey` into `buffers.key` and copies `weOwnAltKey` into `weOwnKey` without clearing the source, so with the flag set both fields own one buffer and teardown runs `ForceZero()` and `FreeDer()` on it twice — ASan reports heap-use-after-free in `wolfSSL_UnloadCertsKeys()`. The swap now clears `weOwnAltKey`, leaving a single owner. Note that path is reachable today, before this PR, for any application that loads an SSL-level alt key and negotiates the ALTERNATIVE signature.

yosuke-wolfssl, line 8844: Same fix as the sibling comment: `weOwnAltKey` is now set after the `AllocCopyDer()` in both `wolfSSL_set_SSL_CTX()` and `SetSSL_CTX_CertsAndKeys()`, plus an ownership transfer in the `SendTls13CertificateVerify()` swap so the flag can be set without causing a double free. Covered by a new ALTERNATIVE-signature connection in `test_dual_alg_support()` and by `test_dual_alg_ctx_alt_key()` for the `wolfSSL_new()` path; both verified under ASan with `WOLFSSL_DUAL_ALG_CERTS` + `WOLFSSL_BLIND_PRIVATE_KEY`, including negative controls.

src/ssl.c
```diff
@@ -8828,33 +8826,37 @@ WOLFSSL_CTX* wolfSSL_set_SSL_CTX(WOLFSSL* ssl, WOLFSSL_CTX* ctx)
     ssl->ctx = ctx;
 
 #ifndef NO_CERTS
+#ifdef KEEP_OUR_CERT
+    /* buffers.weOwnCert tells whether ourCert is this SSL's or the CTX's. */
+    if (ssl->buffers.weOwnCert && ssl->ourCert != NULL) {
```
SparkiDev: Should this go into wolfSSL_certs_clear()?

SparkiDev (changes requested):

src/ssl.c
yosuke-wolfssl: You are correct.
I moved it. wolfSSL_certs_clear() now frees ssl->ourCert inside the existing if (weOwnCert) block, the same shape as ProcessBuffer() in src/ssl_load.c, and ssl->ourCertChain next to the chain buffer, mirroring wolfSSL_clear_chain_certs(). wolfSSL_set_SSL_CTX() is now just the one wolfSSL_certs_clear() call.

It turned out to fix a leak on its own, too: wolfSSL_certs_clear() clears weOwnCert, which is the flag wolfSSL_UnloadCertsKeys() tests before freeing ssl->ourCert — so after a public SSL_certs_clear() nothing could ever free it. (ourCertChain was fine; wolfSSL_ResourceFree() frees it unconditionally.) A new test_wolfSSL_certs_clear() covers this.

tests/api.c
```diff
@@ -1797,6 +1817,61 @@ static int test_dual_alg_crit_ext_support(void)
...
+
+    /* Switching the ctx makes the ssl copy the alternative key again. */
+    ExpectNotNull(ctx2 = wolfSSL_CTX_new(wolfTLSv1_3_server_method()));
+    ExpectTrue(wolfSSL_CTX_use_certificate_file(ctx2, svrCertFile,
+        WOLFSSL_FILETYPE_PEM));
+    ExpectTrue(wolfSSL_CTX_use_PrivateKey_file(ctx2, svrKeyFile,
+        WOLFSSL_FILETYPE_PEM));
+    ExpectIntEQ(wolfSSL_CTX_use_AltPrivateKey_buffer(ctx2, altKey,
+        (long)altKeySz, WOLFSSL_FILETYPE_ASN1), WOLFSSL_SUCCESS);
+    ExpectNotNull(wolfSSL_set_SSL_CTX(ssl, ctx2));
```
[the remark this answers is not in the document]
yosuke-wolfssl: Checked this against the exact configuration named. ./configure --enable-experimental --enable-dual-alg-certs CFLAGS="-Werror" builds clean with no errors, and test_dual_alg_ctx_alt_key passes rather than skips, so the wolfSSL_set_SSL_CTX() call does execute — no implicit declaration or undefined reference.

WOLFSSL_DUAL_ALG_CERTS force-defines OPENSSL_EXTRA in wolfssl/wolfcrypt/settings.h, so the guard term would always be true. In that build wolfssl/options.h contains no OPENSSL_EXTRA at all — it comes only from settings.h — which is why the config this finding predicts would fail is the one that demonstrates it compiles.

!defined(NO_RSA) is likewise unreachable: settings.h #errors when NO_RSA and WOLFSSL_DUAL_ALG_CERTS are both defined. Both terms were removed in an earlier round for exactly this reason.

SparkiDev (approved):

PR #11336, opened 2026-09-01
cryptocb: add WC_ALGO_TYPE_KEYSTORE for hardware key store operations

Frauschi:
## Why

wolfCrypt can already *use* a key that lives in a hardware key store: attach a `WOLF_PRIVATE_KEY_ID` identifier to a key object and the crypto callback does the rest. It cannot manage that key's lifetime. There is no way to put a key into the store, take one out, derive one, destroy one, or ask what a slot holds.

The existing callbacks cannot be stretched to cover it: `WC_ALGO_TYPE_SETKEY` and `WC_ALGO_TYPE_EXPORT_KEY` are bound to a wolfCrypt key object and carry material for that object's own use. Neither can name a stored key, ask for one to be created exportable, or make one outlive the object. So today every secure element port invents its own vendor API for this, and applications using one are welded to that vendor.

**This is a prerequisite for the NXP RW612 EdgeLock (ELS + PKC) crypto callback port**, which follows in a separate PR and cannot express its key store at all without it. That port is the first consumer and is the reason this lands now, but nothing here is NXP-specific.

## What this adds

`WC_ALGO_TYPE_KEYSTORE`, with seven operations:

| Operation | |
| --- | --- |
| `IMPORT_PLAIN` / `EXPORT_PLAIN` | plaintext key material in and out |
| `IMPORT_WRAPPED` / `EXPORT_WRAPPED` | wrapped blob in and out, never plaintext on this side |
| `DERIVE` | slot to slot, without touching RAM |
| `DELETE` | destroy a stored key |
| `GET_INFO` | what does this slot hold? |

Public entry points in `wolfssl/wolfcrypt/wc_keystore.h`, dispatch in `cryptocb.c`, all behind `WOLF_CRYPTO_CB_KEYSTORE` (`--enable-cryptocbutils=keystore`). Off by default; no existing code path changes.

## Backends it was validated against

Three: the EdgeLock port above, `wc_Pkcs11StoreKey()`, and wolfHSM's key cache. The type vocabulary and the plaintext encoding rule come from the latter two.

`wc_Pkcs11StoreKey()` is the same operation as `IMPORT_PLAIN`, reachable only through a `Pkcs11Token*`, so an application using it is bound to PKCS#11 even though the operation is generic. Its AES and HMAC cases map directly onto this API.

Neither backend is fully covered yet. PKCS#11's asymmetric cases and several of wolfHSM's key cache features have no expression here, and each of those is purely additive: appending an operation or an enum value breaks no existing device, so none has to be settled now.

## Design points

**Delete is separate from `WC_ALGO_TYPE_FREE`.** Folding it in would make `wc_ecc_free()` silently destructive for any hardware-backed key.

**A key reference is not a new naming scheme.** It is the identifier `WOLF_PRIVATE_KEY_ID` already uses, so the bytes naming a key here are the bytes `wc_ecc_init_id()` takes and the bytes read back from `key->id`.

**`keyType` and `attrs` are on every operation that creates a key**, and neither export. `keyType` also fixes the plaintext encoding: raw bytes for a symmetric type, DER for an asymmetric one. Where a format carries no metadata (a bare RFC 3394 wrap, or raw bytes) both are the device's only source. Where a container carries its own, the container wins: `keyType` becomes a cross-check a device must refuse a mismatch on, and `attrs` is ignored, since a container's attributes are inside the wrap and unknowable until the key exists.

**RSA and ECC are split by purpose**, because a key store grants their two purposes separately (`CKA_SIGN` against `CKA_DECRYPT`, and `CKA_SIGN` against `CKA_DERIVE`) and one key doing both is the key-separation problem. Everything else is single-purpose. Note that wolfCrypt's own PKCS#11 layer honours this for ECC but sets both `CKA_DECRYPT` and `CKA_SIGN` unconditionally for RSA, for want of an `RsaKey` flag; having the distinction here is what would let that be tightened.

**Argument order is a safety property.** Two interchangeable control words are never left adjacent, because a transposition between them compiles cleanly and surfaces much later as an unrelated-looking failure. `attrs` therefore sits at the tail of every creating operation rather than beside `keyType`: `WC_KEYSTORE_ATTR_EXPORTABLE` and `WC_KEYSTORE_KEY_WRAP` are both 1, so a swap would be silent.

## Testing

Tests in `wolfcrypt/test` and `tests/api` assert that each operation reaches the device **carrying the arguments the caller passed**, not merely that it returned zero, plus argument validation before any device is consulted, the size-query path, the cleared-output contract, and `CRYPTOCB_UNAVAILABLE` from every entry point when nothing is registered. Verified to fail as well as pass: transposing `keyType` and `attrs` at a call site makes `unit.test` exit non-zero.

### DIFF
--- a/doc/dox_comments/header_files/wc_keystore.h
+++ b/doc/dox_comments/header_files/wc_keystore.h
@@ -1,0 +1,346 @@
+/*!
+    \ingroup CryptoCb
+    \brief Place plaintext key material into a device's key store. This is the
+    counterpart to wc_KeyStore_ImportWrapped for devices that accept a key in
+    the clear; a device holding keys that may never appear in plaintext
+    declines it and offers only the wrapped form.
+
+    keyType says what the material is for, which is what lets the device set
+    the stored key's properties: a raw byte string cannot tell an AES key from
+    an HMAC key of the same length. It also fixes the encoding, so no further
+    argument is needed to say: a symmetric type takes the raw key bytes, and an
+    asymmetric type takes DER, a private key as PKCS#8 PrivateKeyInfo and a
+    public key as SubjectPublicKeyInfo. keySz is the length of that encoding
+    in bytes either way.
+
+    The caller owns the plaintext copy and should zeroize it once the import
+    succeeds, for example with ForceZero().
+
+    Key references are opaque byte strings, interpreted only by the device.
+    They are the same identifier WOLF_PRIVATE_KEY_ID uses, so the bytes that
+    name a key here are the bytes wc_ecc_init_id() or wc_AesInit_Id() take to
+    bind a wolfCrypt object to that key.
+
+    attrs sits at the end of the argument list rather than beside keyRef,
+    because WC_KEYSTORE_ATTR_EXPORTABLE and WC_KEYSTORE_KEY_WRAP are both 1 and
+    a swap between two adjacent control words would compile cleanly.
+
+    \return 0 on success
+    \return CRYPTOCB_UNAVAILABLE if no registered device handles the operation
+    \return BAD_FUNC_ARG if keyRef or key is NULL, or either size is zero
+
+    \param devId crypto callback device ID
+    \param keyRef opaque reference naming where the key should land
+    \param keyRefSz length of keyRef in bytes
+    \param keyType what the key is for, from enum wc_KeyStoreKeyType
+    \param key plaintext key material
+    \param keySz length of key in bytes
+    \param attrs attributes for the key being created, WC_KEYSTORE_ATTR_*
+    \param ctx read-only caller context passed through to the device
+
+    _Example_
+    \code
+    byte keyRef[8];
+    byte aesKey[32];
+
+    ret = wc_KeyStore_ImportPlain(devId, keyRef, sizeof(keyRef),
+                                  WC_KEYSTORE_KEY_AES, aesKey, sizeof(aesKey),
+                                  WC_KEYSTORE_ATTR_PERSISTENT, NULL);
+    ForceZero(aesKey, sizeof(aesKey));
+    if (ret != 0) {
+        // handle error
+    }
+    \endcode
+
+    \sa wc_KeyStore_ExportPlain
+    \sa wc_KeyStore_ImportWrapped
+    \sa wc_KeyStore_Delete
+*/
+int wc_KeyStore_ImportPlain(int devId,
+    const byte* keyRef, word32 keyRefSz,
+    word32 keyType, const byte* key, word32 keySz,
+    word32 attrs, const void* ctx);
+
+/*!
+    \ingroup CryptoCb
+    \brief Read a stored key back as plaintext. Typically requires the key to
+    have been created with WC_KEYSTORE_ATTR_EXPORTABLE, and many devices refuse
+    this operation entirely; wc_KeyStore_ExportWrapped is the form that keeps
+    the material inside the boundary.
+
+    The encoding matches wc_KeyStore_ImportPlain(): raw bytes for a symmetric
+    key, DER for an asymmetric one. Call wc_KeyStore_GetInfo() to learn which
+    the stored key is.
+
+    keySz is in/out: the capacity of key on entry, the number of bytes written
+    on return. Passing key as NULL is a size query, which returns the size the
+    device would produce through keySz. keySz is cleared first in that form,
+    so any value passed in is discarded rather than preserved.
+
+    Call wc_KeyStore_GetInfo() first to learn whether the key is exportable,
+    rather than attempting the export and interpreting the failure.
+
+    \return 0 on success
+    \return CRYPTOCB_UNAVAILABLE if no registered device handles the operation
+    \return BAD_FUNC_ARG if keyRef or keySz is NULL, keyRefSz is zero, or key
+    is non-NULL with a zero capacity
+
+    \param devId crypto callback device ID
+    \param keyRef opaque reference naming the key to export
+    \param keyRefSz length of keyRef in bytes
+    \param key buffer receiving the plaintext key, or NULL to query the size
+    \param keySz in: capacity of key in bytes, out: bytes written
+    \param ctx read-only caller context passed through to the device
+
+    _Example_
+    \code
+    byte   out[32];
+    word32 outSz = sizeof(out);
+
+    ret = wc_KeyStore_ExportPlain(devId, keyRef, sizeof(keyRef),
+                                  out, &outSz, NULL);
+    if (ret != 0) {
+        // handle error
+    }
+    \endcode
+
+    \sa wc_KeyStore_ImportPlain
+    \sa wc_KeyStore_ExportWrapped
+    \sa wc_KeyStore_GetInfo
+*/
+int wc_KeyStore_ExportPlain(int devId,
+    const byte* keyRef, word32 keyRefSz,
+    byte* key, word32* keySz, const void* ctx);
+
+/*!
+    \ingroup CryptoCb
+    \brief Unwrap a key blob directly into a device's key store. The key
+    material is never plaintext on this side of the boundary: the device
+    unwraps it internally and the result exists only at keyRef.
+
+    keyType says what the wrapped key is for. A bare RFC 3394 wrap
+    (WC_KEYWRAP_FORMAT_AESKW) is a pure data transformation carrying no
+    metadata, so for that format keyType is the device's only source. A vendor
+    container that carries its own property word stays authoritative, and
+    keyType is then a cross-check the device must refuse a mismatch on, since
+    a mismatch means the right blob is going into the wrong kind of slot.
+
+    Key references are opaque byte strings, interpreted only by the device.
+    They are the same identifier WOLF_PRIVATE_KEY_ID uses, so the bytes that
+    name a key here are the bytes wc_ecc_init_id() or wc_AesInit_Id() take to
+    bind a wolfCrypt object to that key.
+
+    \return 0 on success
+    \return CRYPTOCB_UNAVAILABLE if no registered device handles the operation
+    \return BAD_FUNC_ARG if keyRef or blob is NULL, if either size is zero, or
+    if wrapKeyRef is NULL with a non-zero wrapKeyRefSz
+
+    \param devId crypto callback device ID
+    \param keyRef opaque reference naming where the key should land
+    \param keyRefSz length of keyRef in bytes
+    \param keyType what the wrapped key is for, from enum wc_KeyStoreKeyType,
+    or WC_KEYSTORE_KEY_NONE to leave it to the container
+    \param wrapKeyRef opaque reference naming the wrapping key, or NULL when
+    the device uses an implicit one
+    \param wrapKeyRefSz length of wrapKeyRef in bytes, zero when wrapKeyRef
+    is NULL
+    \param format container format, from enum wc_KeyWrapFormat
+    \param blob the wrapped key blob
+    \param blobSz length of blob in bytes
+    \param attrs WC_KEYSTORE_ATTR_* for the key being created. Used only when
+    the format carries no attributes of its own; a container that carries them
+    wins and this is ignored. Call wc_KeyStore_GetInfo() afterwards to confirm
+    what the import actually produced
+    \param ctx read-only caller context passed through to the device
+
+    _Example_
+    \code
+    byte keyRef[8];
+    byte kekRef[8];
+
+    ret = wc_KeyStore_ImportWrapped(devId, keyRef, sizeof(keyRef),
+                                    WC_KEYSTORE_KEY_AES,
+                                    kekRef, sizeof(kekRef),
+                                    WC_KEYWRAP_FORMAT_AESKW,
+                                    blob, blobSz,
+                                    WC_KEYSTORE_ATTR_EXPORTABLE, NULL);
+    if (ret != 0) {
+        // handle error
+    }
+    \endcode
+
+    \sa wc_KeyStore_ExportWrapped
+    \sa wc_KeyStore_Delete
+*/
+int wc_KeyStore_ImportWrapped(int devId,
+    const byte* keyRef, word32 keyRefSz, word32 keyType,
+    const byte* wrapKeyRef, word32 wrapKeyRefSz,
+    word32 format, const byte* blob, word32 blobSz,
+    word32 attrs, const void* ctx);
+
+/*!
+    \ingroup CryptoCb
+    \brief Wrap a stored key under another stored key and emit the blob.
+    Typically requires the key to have been created with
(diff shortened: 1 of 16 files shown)


Frauschi:

tests/api/test_keystore.c
```diff
@@ -0,0 +1,413 @@
...
+    KsSeen* seen = (KsSeen*)ctx;
+
+    (void)devIdArg;
+
+    if (info == NULL || info->algo_type != WC_ALGO_TYPE_KEYSTORE) {
+        return WC_NO_ERR_TRACE(CRYPTOCB_UNAVAILABLE);
+    }
+
+    seen->calls++;
+    seen->op = info->keystore.type;
```
Frauschi: fixed

wolfcrypt/test/test.c
```diff
@@ -68426,6 +68439,230 @@ WOLFSSL_TEST_SUBROUTINE wc_test_ret_t she_test(void)
...
+    if (ret != 0 || ctx.lastOp != WC_KEYSTORE_DELETE) {
+        ret = WC_TEST_RET_ENC_NC; goto out;
+    }
+
+    /* the device answers keySz only; keyType and attrs must come back cleared,
+     * never left holding whatever was on the caller's stack */
+    keyType = 0xDEADBEEF;
+    keySz   = 0xDEADBEEF;
+    attrs   = 0xDEADBEEF;
+    ret = wc_KeyStore_GetInfo(ksDevId, kref, (word32)sizeof(kref),
```
Frauschi, line 69417: fixed

tests/api/test_keystore.c
```diff
@@ -0,0 +1,413 @@
...
+    /* blob == NULL is the size query and must still reach the device */
+    outSz = 0;
+    ExpectIntEQ(wc_KeyStore_ExportWrapped(KS_TEST_DEVID,
+        ksKeyRef, (word32)sizeof(ksKeyRef),
+        ksOtherRef, (word32)sizeof(ksOtherRef),
+        WC_KEYWRAP_FORMAT_AESKW, NULL, &outSz, NULL), 0);
+    ExpectIntEQ(outSz, 40);
+
+    /* a buffer with no size is not a size query, it is a mistake */
+    ExpectIntEQ(wc_KeyStore_ExportWrapped(KS_TEST_DEVID,
```
Frauschi, line 349: fixed

tests/api/test_keystore.c
```diff
@@ -0,0 +1,413 @@
...
+    return EXPECT_RESULT();
+}
+
+int test_wc_KeyStore_NoDevice(void)
+{
+    EXPECT_DECLS;
+#if defined(WOLF_CRYPTO_CB) && defined(WOLF_CRYPTO_CB_KEYSTORE)
+    word32 keyType = 0, keySz = 0, attrs = 0;
+
+    /* Nothing registered: every entry point must decline. */
```
Frauschi, line 522: fixed

tests/api/test_keystore.c
```diff
@@ -0,0 +1,413 @@
...
+    KsSeen seen;
+    byte   out[64];
+    word32 outSz = (word32)sizeof(out);
+
+    XMEMSET(&seen, 0, sizeof(seen));
+    ExpectIntEQ(wc_CryptoCb_RegisterDevice(KS_TEST_DEVID, KsCb, &seen), 0);
+
+    ExpectIntEQ(wc_KeyStore_ExportPlain(KS_TEST_DEVID,
+        ksKeyRef, (word32)sizeof(ksKeyRef), out, &outSz, NULL), 0);
+    ExpectIntEQ(seen.op, WC_KEYSTORE_EXPORT_PLAIN);
```
Frauschi, line 216: fixed

tests/api/test_keystore.c
```diff
@@ -0,0 +1,413 @@
...
+            seen->deriv    = info->keystore.op.derive.deriv;
+            seen->derivSz  = info->keystore.op.derive.derivSz;
+            break;
+        case WC_KEYSTORE_DELETE:
+            seen->keyRef   = info->keystore.op.deleteKey.keyRef;
+            seen->keyRefSz = info->keystore.op.deleteKey.keyRefSz;
+            break;
+        case WC_KEYSTORE_GET_INFO:
+            seen->keyRef = info->keystore.op.getInfo.keyRef;
+            *info->keystore.op.getInfo.keyType = WC_KEYSTORE_KEY_WRAP;
```
Frauschi: fixed

wolfcrypt/test/test.c
```diff
@@ -68426,6 +68439,275 @@ WOLFSSL_TEST_SUBROUTINE wc_test_ret_t she_test(void)
...
+            c->lastBlobSz  = info->keystore.op.importWrapped.blobSz;
+            c->lastFormat  = info->keystore.op.importWrapped.format;
+            c->lastAttrs   = info->keystore.op.importWrapped.attrs;
+            break;
+        case WC_KEYSTORE_EXPORT_WRAPPED:
+            c->lastKeyRef  = info->keystore.op.exportWrapped.keyRef;
+            break;
+        case WC_KEYSTORE_DERIVE:
+            c->lastKeyRef    = info->keystore.op.derive.keyRef;
+            c->lastSrcKeyRef = info->keystore.op.derive.srcKeyRef;
```
Frauschi, line 69282: fixed

tests/api/test_keystore.c
```diff
@@ -0,0 +1,486 @@
...
+            /* answer the size query so the caller can check it propagates */
+            if (info->keystore.op.exportPlain.key == NULL) {
+                *info->keystore.op.exportPlain.keySz = 32;
+            }
+            break;
+        case WC_KEYSTORE_IMPORT_WRAPPED:
+            seen->keyRef   = info->keystore.op.importWrapped.keyRef;
+            seen->keyRefSz = info->keystore.op.importWrapped.keyRefSz;
+            seen->keyType  = info->keystore.op.importWrapped.keyType;
+            seen->otherRef = info->keystore.op.importWrapped.wrapKeyRef;
```
Frauschi: fixed

tests/api/test_keystore.c
```diff
@@ -0,0 +1,486 @@
...
+    ExpectIntEQ(wc_KeyStore_ImportWrapped(KS_TEST_DEVID, NULL, 0,
+        WC_KEYSTORE_KEY_AES, ksOtherRef, (word32)sizeof(ksOtherRef),
+        WC_KEYWRAP_FORMAT_VENDOR, ksBlob, (word32)sizeof(ksBlob), 0, NULL),
+        WC_NO_ERR_TRACE(BAD_FUNC_ARG));
+    ExpectIntEQ(wc_KeyStore_ImportWrapped(KS_TEST_DEVID,
+        ksKeyRef, (word32)sizeof(ksKeyRef), WC_KEYSTORE_KEY_AES,
+        ksOtherRef, (word32)sizeof(ksOtherRef),
+        WC_KEYWRAP_FORMAT_VENDOR, NULL, 0, 0, NULL),
+        WC_NO_ERR_TRACE(BAD_FUNC_ARG));
+    ExpectIntEQ(seen.calls, 0);
```
Frauschi, line 289: fixed

tests/api/test_keystore.c
```diff
@@ -0,0 +1,486 @@
...
+    word32 outSz = (word32)sizeof(out);
+
+    XMEMSET(&seen, 0, sizeof(seen));
+    ExpectIntEQ(wc_CryptoCb_RegisterDevice(KS_TEST_DEVID, KsCb, &seen), 0);
+
+    ExpectIntEQ(wc_KeyStore_ExportWrapped(KS_TEST_DEVID,
+        ksKeyRef, (word32)sizeof(ksKeyRef),
+        ksOtherRef, (word32)sizeof(ksOtherRef),
+        WC_KEYWRAP_FORMAT_AESKW, out, &outSz, ksCallerCtx), 0);
+    ExpectIntEQ(seen.op, WC_KEYSTORE_EXPORT_WRAPPED);
```
Frauschi, line 320: fixed

tests/api/test_keystore.c
```diff
@@ -0,0 +1,486 @@
...
+    ExpectIntEQ(seen.attrs, WC_KEYSTORE_ATTR_PERSISTENT);
+    ExpectPtrEq(seen.keyRef, ksKeyRef);
+    ExpectPtrEq(seen.otherRef, ksOtherRef);
+    ExpectIntEQ(seen.format, WC_KEYWRAP_FORMAT_VENDOR);
+    ExpectIntEQ(seen.blobSz, (word32)sizeof(ksBlob));
+    ExpectPtrEq(seen.ctx, ksCallerCtx);
+
+    /* arguments are validated before the device is consulted */
+    seen.calls = 0;
+    ExpectIntEQ(wc_KeyStore_ImportWrapped(KS_TEST_DEVID, NULL, 0,
```
Frauschi, line 273: fixed

tests/api/test_keystore.c
```diff
@@ -0,0 +1,539 @@
...
+    XMEMSET(&seen, 0, sizeof(seen));
+    ExpectIntEQ(wc_CryptoCb_RegisterDevice(KS_TEST_DEVID, KsCb, &seen), 0);
+
+    /* Asserting each control word separately is what catches a
+     * transposition between them. */
+    ExpectIntEQ(wc_KeyStore_Derive(KS_TEST_DEVID,
+        ksKeyRef, (word32)sizeof(ksKeyRef), WC_KEYSTORE_KEY_AES,
+        ksOtherRef, (word32)sizeof(ksOtherRef),
+        WC_KDF_TYPE_HKDF, ksDeriv, (word32)sizeof(ksDeriv),
+        WC_KEYSTORE_ATTR_EXPORTABLE, ksCallerCtx), 0);
```
Frauschi: fixed

tests/api/test_keystore.c
```diff
@@ -0,0 +1,539 @@
...
+            seen->kdfType  = info->keystore.op.derive.kdfType;
+            seen->deriv    = info->keystore.op.derive.deriv;
+            seen->derivSz  = info->keystore.op.derive.derivSz;
+            break;
+        case WC_KEYSTORE_DELETE:
+            seen->keyRef   = info->keystore.op.deleteKey.keyRef;
+            seen->keyRefSz = info->keystore.op.deleteKey.keyRefSz;
+            break;
+        case WC_KEYSTORE_GET_INFO:
+            seen->keyRef = info->keystore.op.getInfo.keyRef;
```
Frauschi: fixed

Frauschi:
Jenkins retest this please

wolfssl/wolfcrypt/cryptocb.h
```diff
@@ -803,6 +856,70 @@ typedef struct wc_CryptoInfo {
...
+            } derive;
+            struct {                       /* WC_KEYSTORE_DELETE */
+                const byte* keyRef;
+                word32      keyRefSz;
+            } deleteKey;
+            struct {                       /* WC_KEYSTORE_GET_INFO */
+                const byte* keyRef;
+                word32      keyRefSz;
+                word32*     keyType;       /* out: enum wc_KeyStoreKeyType */
+                word32*     keySz;         /* out: key size in BITS */
```
padelsbach: Nit: can we use `keyBits`, `keySzBits` or similar to avoid conflating with other uses of `keySz` in this union?

doc/dox_comments/header_files/wc_keystore.h
```diff
@@ -0,0 +1,341 @@
...
+    if (ret == 0 && (attrs & WC_KEYSTORE_ATTR_EXPORTABLE)) {
+        // the key may be wrapped out
+    }
+    \endcode
+
+    \sa wc_KeyStore_ExportWrapped
+    \sa wc_KeyStore_Derive
+*/
+int wc_KeyStore_GetInfo(int devId, const byte* keyRef, word32 keyRefSz,
+    word32* keyType, word32* keySz, word32* attrs, const void* ctx);
```
padelsbach: Nit: should probably have a blank line at the end of this file

wolfcrypt/src/cryptocb.c
```diff
@@ -3694,6 +3704,264 @@ int wc_CryptoCb_ExportKey(int devId, int type, const void* obj, void* out)
...
+    CryptoCb* dev;
+
+    if (keyRef == NULL || keyRefSz == 0 || blob == NULL || blobSz == 0) {
+        return BAD_FUNC_ARG;
+    }
+    if (wrapKeyRef == NULL && wrapKeyRefSz != 0) {
+        return BAD_FUNC_ARG;
+    }
+
+    dev = wc_CryptoCb_FindDevice(devId, WC_ALGO_TYPE_KEYSTORE);
```
padelsbach, line 3802: the `GetInfo()` call below clears out the params it is about to modify:

```
    if (keyType != NULL) {
        *keyType = 0;
    }
    if (keySz != NULL) {
        *keySz = 0;
    }
    if (attrs != NULL) {
        *attrs = 0;
    }
```

Should we do that here too? And any vars in the other new functions?

padelsbach (dismissed):
Nice addition of this helpful feature. A few minor nits from me. Over to @bigbrett

wolfssl/wolfcrypt/cryptocb.h
Frauschi: Fixed. Moved to keyBits

doc/dox_comments/header_files/wc_keystore.h
Frauschi: Fixed

wolfcrypt/src/cryptocb.c
Frauschi: Fixed where appropriate

.skoll-known.md
```diff
@@ -0,0 +1,14 @@
+# Known issues
```
padelsbach: Is this file relevant to the PR?

zephyr/samples/wolfssl_tls_sock/prj-no-malloc.conf
```diff
@@ -36,12 +36,9 @@ CONFIG_LOG=y
 CONFIG_LOG_MODE_IMMEDIATE=y
 
 # TLS configuration
+# The settings file below is authoritative, so the feature Kconfigs
```
padelsbach: Is this change relevant?

.skoll-known.md
Frauschi: Sorry, I apparently messed up with the fixup commits. Fully removed.

zephyr/samples/wolfssl_tls_sock/prj-no-malloc.conf
Frauschi: Also a fixup error, thanks for catching! Removed completely.

padelsbach (approved):

PR #11342, opened 2026-09-01
Store an alt name with its entry in one allocation

Frauschi:
One of six independent branches that cut allocations in the TLS and certificate layers. Self-contained and touches no TLS code.

### What changes

Parsing the subject alternative names of a certificate allocated a `DNS_entry` and then a second buffer for the name it points at - a shape the code itself flagged with a "consider one malloc" note. A chain carrying six names cost twelve allocations on every handshake that verifies it.

`AltNameNewEx()` allocates the entry with room for the name behind it and copies the name in, so a six name chain costs six allocations. `wolfSSL_X509_add_altname_ex()` uses the same helper.

The failure path in `SetDNSEntry()` now hands the entry to `FreeAltNames()` instead of freeing two buffers by hand, so any `ipString` or `ridString` a later step allocates is released with it. Nothing leaked before this - no in-tree path can fail after those buffers are built - it is hardening against a future failure path.

### API note

An entry built this way reports `nameStored == 0`, which parsing already used for a name borrowed from the input DER. Either way the name is not a separate allocation and must not be freed on its own. Application code that walks `WOLFSSL_X509->altNames` or `DecodedCert->altNames` and frees `name` only when `nameStored` is set is unaffected; code that frees it unconditionally was already wrong.

`AltNameNewEx()` rejects a negative length, and a NULL name with a positive length, rather than leaving `len` covering bytes that were never written - which the old inline code did. `SetDNSEntry()` screens those arguments before allocating so a NULL return really does mean out of memory, rather than reporting a bad argument as `MEMORY_E` and sending a caller that retries on `MEMORY_E` round forever. `wolfSSL_X509_add_altname_ex()` bounds its `word32` length against `INT_MAX` explicitly rather than relying on what an out-of-range value narrows to.

### DIFF
--- a/src/x509.c
+++ b/src/x509.c
@@ -3119,7 +3119,6 @@ int wolfSSL_X509_add_altname_ex(WOLFSSL_X509* x509, const char* name,
         word32 nameSz, int type)
 {
     DNS_entry* newAltName = NULL;
-    char* nameCopy = NULL;
 
     if (x509 == NULL)
         return WOLFSSL_FAILURE;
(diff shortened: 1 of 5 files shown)


Frauschi:

PR #11343, opened 2026-09-01
Leave TLS 1.2 transcript state out of TLS 1.3 only builds

Frauschi:
One of six independent branches that cut per-object memory in the TLS layer. Self-contained; the only one that adds build options.

### What changes

Every build hashed each handshake message with SHA-512 and reserved two `Hashes` structures for the TLS 1.2 Finished and CertificateVerify digests, whether or not TLS 1.2 was compiled in. TLS 1.3 derives both differently and never reads them.

`verifyHashes` and `certHashes` are now kept only when TLS 1.2 is enabled, and the SHA-512 transcript only when TLS 1.2 or `WOLFSSL_TLS13_SHA512` is. A TLS 1.3 only build drops the handshake hashes from 864 to 352 bytes per handshake and stops running SHA-512 over every handshake message.

### New options

`--enable-tls13-sha512` and the matching CMake `WOLFSSL_TLS13_SHA512` are new. The macro already existed in `src/tls13.c`; giving it a configure and CMake surface means a macro that now decides the layout of the handshake hashes is recorded in the generated `wolfssl/options.h` like every other one.

It keeps the SHA-512 transcript available to TLS 1.3. No cipher suite in the table sets `mac_algorithm` to `sha512_mac`, so nothing negotiates it today; the one path that reaches the hash is the HelloRetryRequest cookie HMAC, and only in a build with neither SHA-256 nor SHA-384.

Both prerequisite checks are placed after the values they read have settled, which matters here: `ENABLED_TLS13` can still be forced off by the key exchange prerequisite check much further down `configure.ac`, and `WOLFSSL_SHA512` is not written until its own `add_option()` and can be raised again by the `WOLFSSL_CLU` block. A `user_settings.h` build reaches neither check, so `wolfssl/internal.h` carries an `#error` for the same combination - without it the failure surfaces as five "no member named hashSha512" errors from deep inside `src/tls13.c`.

### CI

Three entries in `.github/configs/os-check-linux.json`:

- `tls13-only-sha512-transcript` - `WOLFSSL_NO_TLS12` with SHA-512 still enabled, the only shape where `WOLFSSL_HS_HASH_SHA512` is false while `WOLFSSL_SHA512` is true.
- `tls13-sha512-transcript-only` - the other side of that macro, `--enable-tls13-sha512` without TLS 1.2.
- `tls13-sha512-runtime` - the option with TLS 1.2 left in, so examples and unit tests build and run with the macro set. It proves the option does not break an otherwise ordinary build; it cannot exercise the `sha512_mac` arms, since no suite selects them.

### Memory impact

Measured on a 64-bit TLS 1.3 only build (`--enable-tls13 --disable-tlsv12 --disable-oldtls CPPFLAGS=-DWOLFSSL_NO_TLS12`):

| | master | this branch |
|---|---|---|
| `sizeof(HS_Hashes)` | 864 | **352** |

**512 bytes saved per handshake**, allocated for the life of the handshake on every connection. A build that keeps TLS 1.2 is unchanged, as it needs the state.

There is a CPU saving alongside it: such a build no longer runs SHA-512 over every handshake message. That hash was computed on every connection and never read.


Frauschi:

JacobBarthelmeh (approved):

PR #11344, opened 2026-09-01
TI C2000 C28x: 16-bit-byte fixes, octet helpers, and an optional precomputed ML-DSA matrix A

dgarske:
Support work for 16-bit-byte targets (TI C2000 C28x, `CHAR_BIT == 16`). All of it is a no-op where `CHAR_BIT == 8`, and the ML-DSA addition is off unless explicitly enabled.

**SP regression.** `sp_c32.c` is generated, and a regeneration reverted two `CHAR_BIT != 8` fixes it carried: the `& 0xFF` octet masks in `sp_*_to_bin_*`, and `sizeof(mp_digit) * CHAR_BIT` (not `* 8`) in the `sp_*_from_mp` constant-time mask. Without them ECDSA sign/verify, ECDH, RSA-2048 verify and DH FFDHE-2048 all fail on the C28x. Replaced with the output of the corrected generator, so the next regeneration is a no-op; the template fix is a companion PR in wolfSSL/scripts.

**Octet boundary helpers.** A wolfCrypt `byte*` buffer holds one octet per `byte` cell. Data arriving from flash or a byte-oriented link is packed two octets per cell instead, and callers had no supported way to convert. Adds `wc_UnpackOctets()`/`wc_PackOctets()` -- both bounds-check source and destination -- plus `WC_OCTETS_PER_BYTE`/`WC_PACKED_CELLS()`. The functions are declared only under `WOLFSSL_WIDE_BYTE`: on an 8-bit-byte target the two layouts are the same buffer, so there is nothing to convert and no new API surface. Also masks `mldsa_encode_w1_88_c()`, whose 6-bit inputs reach 2795.

**SHA-3 absorb.** `misc.c`'s `readUnalignedWord32/64()` mask every cell under `WOLFSSL_WIDE_BYTE`; `sha3.c`'s `Load64Unaligned()`/`Load64BitLittleEndian()` did not, so SHA-3/SHAKE silently trusted callers to hand it clean octets. Now masked to match.

**Optional precomputed ML-DSA matrix A.** ML-DSA verify expands the public matrix A from `rho` by SHAKE128 rejection sampling on every call, which dominates verify on a part with no SHA-3 acceleration. A depends only on the public key, so where the verification key is fixed at build time it can be expanded once off target and stored in flash. Adds `WOLFSSL_MLDSA_VERIFY_PRECOMP_A` and `wc_MlDsaKey_SetPrecompA()`, which attaches a caller-owned, read-only matrix that is read in place and costs no RAM. It is consulted in the shared verify core, so it applies to every verify mode rather than one entry point. The setter binds the matrix to the key by comparing `rho` and requires the public key to have been imported first, so a stale or mismatched matrix is rejected rather than silently verified against. With nothing attached the behavior is bit-identical to before, including when the gate is enabled.

Security note: A is derived from the public key and must be integrity-protected exactly as the public key is, since substituting it can influence verification. This is stated at the API and in the header.

Also adds a CI job that fails if `sp_c32.c` loses its octet masks or reverts to `sizeof(mp_digit) * 8`, since that regression passes every 8-bit-byte build and all hardware-free CI. And sets `WC_16BIT_CPU` in `IDE/C2000/user_settings.h` -- the C28x has a 16-bit `int` as well as a 16-bit `char`, and without it the 32-bit OID sums in `wc_OidGetHash()` collide once truncated. That lets `hash.c`, which HashML-DSA depends on, join the compile guard.

## Test status

Hardware-validated on a TI LAUNCHXL-F28P55X (cl2000): the example's full 15-config matrix passes, including ML-DSA-44/65/87 verify, `wc_MlDsaKey_VerifyCtxHash()` over SHA-256/SHA-512, a verify from a packed key and signature, and a pure-mode verify of a packed firmware image with a corrupted-image rejection. The five SP KATs fail before this change and pass after. The precomputed matrix was regenerated independently from `rho` and matched the stored one exactly. Host `make check` and `testwolfcrypt` pass with the ML-DSA gate both on and off, and `IDE/C2000/compile.sh` is clean.

## Notes

The `mldsa_encode_w1_88_c()` mask is defensive rather than a live fix. Its overflow bits are, by construction, the same bits the next cell contributes at the same position in the 64-bit Keccak lane, and the absorb ORs octets into that lane -- so the stray bits are either idempotent or shifted out of the word entirely. Verified on hardware (the level-44 commitment matches with the mask reverted) and exhaustively in simulation (192,000 lanes, zero differences). The invariance depends on the absorb staying OR-based and on nothing else consuming the buffer, so the mask removes the dependency rather than fixing a defect.

Generating the matrix is left to the application. A build-time generator and a complete worked example are in the companion wolfssl-examples PR https://github.com/wolfSSL/wolfssl-examples/pull/617


wolfcrypt/src/wc_mldsa.c
```diff
@@ -11809,6 +11830,39 @@ int wc_MlDsaKey_VerifyCtxHash(wc_MlDsaKey* key, const byte* sig, word32 sigLen,
...
+        return BAD_FUNC_ARG;
+    }
+    if (rhoLen != MLDSA_PUB_SEED_SZ) {
+        return BAD_FUNC_ARG;
+    }
+    /* Row-major k x l polynomials of MLDSA_N coefficients. */
+    if (aLen != (word32)key->params->k * (word32)key->params->l * MLDSA_N) {
+        return BAD_FUNC_ARG;
+    }
+    /* A is a function of rho alone.  Reject a matrix built for a different
```
aidangarske, line 11908: **Precomputed matrix survives key replacement**

The PR stores `aPre` after a one-time check against the current `rho`, but does not invalidate it when `wc_MlDsaKey_SetParams()`, `wc_MlDsaKey_ImportPubRaw()`, or key generation replaces the key. Verification subsequently indexes the pointer using the current `k` and `l`. For example, attaching a level-44 matrix, changing to level 87, and importing a level-87 key makes verification read 14,336 coefficients from a 4,096-coefficient allocation. Even same-level key replacement uses the stale matrix and produces incorrect verification results. Existing internal matrix caches are explicitly invalidated on these transitions; the new borrowed pointer must receive equivalent lifecycle handling.

**Fix:** Clear `aPre` whenever parameters or public-key material change, including successful key generation and import paths. Add same-level re-import and cross-level reuse tests; alternatively retain the bound length, parameters, and `rho` and revalidate them before every use.

wolfcrypt/src/wc_mldsa.c
```diff
@@ -10942,11 +10949,25 @@ static int mldsa_verify_with_mu(wc_MlDsaKey* key, const byte* mu,
                 zt = z;
             #endif
                 /* Step 3: Create polynomial from hashing seed. */
+            #ifdef WOLFSSL_MLDSA_VERIFY_PRECOMP_A
```
aidangarske, line 11010: **Default verifier ignores the attached matrix**

The new `aPre` selection is compiled only in the `WOLFSSL_MLDSA_VERIFY_SMALL_MEM` branch. The default verifier branch at lines 10597-10753 still allocates a matrix and calls `mldsa_expand_a()` without consulting `key->aPre`. Because `WOLFSSL_MLDSA_VERIFY_PRECOMP_A` neither implies nor requires `WOLFSSL_MLDSA_VERIFY_SMALL_MEM`, a normal build can enable the advertised feature, receive success from `wc_MlDsaKey_SetPrecompA()`, and then silently regenerate and use a different buffer. The new public feature is therefore a no-op in the default supported configuration.

**Fix:** Integrate `aPre` into the non-small-memory verifier and skip its matrix allocation/expansion when attached. If the feature is intentionally small-memory-only, reject incompatible configurations at compile time and document that restriction instead of allowing a successful no-op setter.

wolfcrypt/src/wc_mldsa.c
```diff
@@ -11809,6 +11830,39 @@ int wc_MlDsaKey_VerifyCtxHash(wc_MlDsaKey* key, const byte* sig, word32 sigLen,
     return ret;
 }
 
+#ifdef WOLFSSL_MLDSA_VERIFY_PRECOMP_A
```
aidangarske, line 11891: **Precomputed-A API has no regression tests**

No in-tree test invokes `wc_MlDsaKey_SetPrecompA()`, and no committed test configuration enables `WOLFSSL_MLDSA_VERIFY_PRECOMP_A`. Running existing tests with the gate defined exercises only the `aPre == NULL` fallback, leaving the attached path, validation branches, build variants, and key lifecycle uncovered. This omission allowed both blocking defects above to pass unnoticed.

**Fix:** Add an API test using a known expanded matrix and run it in both default and small-memory verifier configurations. Cover successful equivalence with ordinary expansion, NULL and length errors, mismatched `rho`, fallback without attachment, public-key re-import, parameter changes, and key generation.

.github/workflows/ti-c2000-compile.yml
```diff
@@ -157,6 +157,33 @@ jobs:
...
+    # sp_c32.c is generated.  Two CHAR_BIT != 8 fixes live in it and have been
+    # dropped by a regeneration before, which breaks ECDSA/ECDH/RSA/DH on
+    # 16-bit-byte targets while every 8-bit build stays green.  Fail loudly.
+    name: sp_c32.c CHAR_BIT != 8 masks
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Octet masks present in sp_*_to_bin_*
+        run: |
+          n=$(grep -c 'a\[j--\] |= (byte)(((sp_uint32)r\[i\] << s) & 0xFF)' \
```
aidangarske, line 176: **SP mask guard passes partial regressions**

The new guard checks only that the two mask-pattern counts are nonzero and equal. If one serializer loses both masks, the counts fall from 8/8 to 7/7 and the job passes; adding another unmasked serializer also leaves the counts unchanged. The `CHAR_BIT` check similarly rejects only one exact spelling of the old expression rather than asserting that every `sp_*_from_mp` implementation contains the corrected expression. Consequently, the job does not enforce the invariant it claims to protect.

**Fix:** Enumerate every generated `sp_*_to_bin_*` and `sp_*_from_mp` function and verify the required expression within each function, or compare each mask count against the independently counted number of serializer/converter definitions.

wolfssl/wolfcrypt/wc_mldsa.h
```diff
@@ -759,6 +765,24 @@ WOLFSSL_API
 int wc_MlDsaKey_VerifyMu(wc_MlDsaKey* key, const byte* sig, word32 sigLen,
     const byte* mu, word32 muLen, int* res);
 
+#ifdef WOLFSSL_MLDSA_VERIFY_PRECOMP_A
```
aidangarske, line 772: **New ML-DSA API is absent from canonical documentation**

The PR introduces a public, feature-gated API with important ownership and integrity requirements, but adds it only as a normal header comment. The canonical `doc/dox_comments/header_files/wc_mldsa.h` documents all neighboring verification APIs and has no entry for this function, while `doc/ALGORITHM_DEFINES.md` does not describe `WOLFSSL_MLDSA_VERIFY_PRECOMP_A`. Users of the generated API documentation therefore cannot discover the function or its caller-owned lifetime and integrity contract.

**Fix:** Add a full Doxygen entry to `doc/dox_comments/header_files/wc_mldsa.h`, including element-count units, ownership, lifetime, key-replacement behavior, integrity requirements, errors, and an example. Document the enabling define and supported verifier configurations in `doc/ALGORITHM_DEFINES.md`.

aidangarske (dismissed):
## Skoll Code Review

**Scan type:** `review`

### Posted findings
- [High] **Precomputed matrix survives key replacement** — `wolfcrypt/src/wc_mldsa.c:11850-11862`
- [High] **Default verifier ignores the attached matrix** — `wolfcrypt/src/wc_mldsa.c:10952-10958`
- [Medium] **Precomputed-A API has no regression tests** — `wolfcrypt/src/wc_mldsa.c:11833-11864`
- [Medium] **SP mask guard passes partial regressions** — `.github/workflows/ti-c2000-compile.yml:170-182`
- [Medium] **New ML-DSA API is absent from canonical documentation** — `wolfssl/wolfcrypt/wc_mldsa.h:768-783`

dgarske:
Jenkins retest this please

dgarske:
Jenkins retest this please

SparkiDev:
Jenkins: retest this please

SparkiDev:
unrecognized macros used:
WOLFSSL_MLDSA_VERIFY_PRECOMP_A

dgarske:
Jenkins retest this please. FIPS failure

SparkiDev (approved):

PR #11345, opened 2026-09-01
Size the pre-master secret buffer by the key exchanges, and fix a DHE-PSK overflow

Frauschi:
One of six independent branches that cut allocations and per-object memory in the TLS layer. This one also carries a memory-safety fix.

### The fix

A DHE-PSK exchange stores a two byte length, the DH shared secret, another two byte length and the PSK in the pre-master secret buffer. Sized with `ENCRYPT_LEN`, that is **two bytes short** for the largest DH parameters the build accepts together with a maximum length PSK, and the last two bytes of the key were written past the allocation on both peers. It is reached before authentication completes, from parameters the peer supplies.

`test_dhe_psk_max_premaster()` drives the handshake that produces the largest secret: 4096 bit parameters on the server and a `MAX_PSK_KEY_LEN` PSK. The client asks for an ECC named group so FFDHE negotiation does not substitute a smaller group, and the negotiated key size is checked to confirm it did not. Against the old sizing the assertion fails and a sanitizer build reports a heap buffer overflow in `MakePSKPreMasterSecret()`.

### What else changes

**The buffer is part of the `Arrays` struct.** It used to be its own allocation - made eagerly for TLS 1.3 and the sniffer, lazily elsewhere, and freed by `CleanPreMaster()` or `FreeArrays()` - for a buffer whose lifetime is exactly that of the `Arrays` it belongs to. It is now a member, so the two are created, wiped and released together, and nothing tracks or frees it separately.

**It is sized by what the build can actually produce.** `ENCRYPT_LEN` also has to cover an RSA signature, so it follows the largest RSA key the build supports: a build with RSA 4096 certificates reserved 512 bytes for a secret ECDHE fills 32 to 98 bytes of. `MAX_PREMASTER_SZ` takes the largest secret any key exchange in the build can produce - the RSA pre-master, the DH parameters a peer may send, an ECDHE secret plus the 32 byte ML-KEM half of a hybrid key share, the length prefixes and key the PSK suites append, the TLS 1.3 handshake secret that reuses the buffer - with the master secret length as a floor.

A build with DH is unchanged, since the peer chooses the parameter size. A build without it, the usual TLS 1.3 configuration, holds 98 rather than 512 bytes per handshake. A build with both DH and PSK gains two bytes, which is the overflow above.

`WOLFSSL_MAX_GEN_PREMASTER_SZ` lets an integration that knows what its `GenPreMasterCb` writes opt down; it defaults to `ENCRYPT_LEN` so no existing callback can be overrun, and the capacity handed to the callback is unchanged.

### Behaviour changes worth flagging

`wolfSSL_[CTX_]SetMaxDhKey_Sz()` now returns `BAD_FUNC_ARG` for a ceiling above what the build's math supports, rather than accepting it and silently reducing it. This is load-bearing: without it an application could raise `maxDhKeySz` above the buffer size and reopen the overflow. An application that defensively calls `wolfSSL_CTX_SetMaxDhKey_Sz(ctx, 4096)` on a build whose math tops out lower now gets an error where it previously got success. The check sits after the crypto-policy block so a request that violates both is still reported as the policy violation.

The minimum setters are deliberately unchanged: a floor above what the build supports is a tightening the caller asked for, it simply refuses every key a peer can offer, and it cannot overrun anything.

`wolfSSL_dtls_import()` treats the serialized DH sizes the same way: the exported ceiling is clamped, and an exported floor or negotiated size above what this build supports fails the import rather than silently applying a policy neither side asked for.

### Memory impact

Measured on 64-bit builds. `Arrays` is one allocation per handshake; on master the pre-master secret was a second one beside it.

**TLS 1.3 without DH or PSK** (`--enable-tls13 --disable-dh --disable-psk`), the usual TLS 1.3 configuration:

| | master | this branch |
|---|---|---|
| `sizeof(Arrays)` | 208 | 300 |
| separate pre-master buffer | 512 | none |
| **total per handshake** | **720 bytes in 2 allocations** | **300 bytes in 1** |

That is **420 bytes and one allocation/free pair saved per handshake**, because `MAX_PREMASTER_SZ` comes out at 98 where `ENCRYPT_LEN` was 512.

**Default build** (DH enabled, so the peer chooses the parameter size and the buffer cannot shrink):

| | master | this branch |
|---|---|---|
| **total per handshake** | 720 bytes in 2 allocations | **712 bytes in 1** |

No meaningful size change, but still **one fewer allocation and free per handshake**, and the buffer is now wiped together with the arrays rather than needing its own `CleanPreMaster()` path.

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -1513,6 +1513,15 @@ static int ImportOptions(WOLFSSL* ssl, const byte* exp, word32 len, byte ver,
 {
     int idx = 0;
     Options* options = &ssl->options;
+    byte sendVerify;
+    byte verifyPeer;
+    byte verifyNone;
+    byte downgrade;
+#ifndef NO_DH
+    word16 minDhKeySz;
+    word16 maxDhKeySz;
+    word16 dhKeySz;
+#endif
 
     switch (ver) {
         case WOLFSSL_EXPORT_VERSION:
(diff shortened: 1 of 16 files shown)


Frauschi:

src/keys.c
```diff
@@ -4077,8 +4079,6 @@ static void CleanPreMaster(WOLFSSL* ssl)
     wc_MemZero_Check(ssl->arrays->preMasterSecret, sz);
```
Frauschi: fixed

tests/api/test_dtls.c
```diff
@@ -2998,6 +2998,96 @@ static unsigned char version_3[] = {
...
+        at = (int)i;
+        hits++;
+    }
+    ExpectIntEQ(hits, 1);
+    if (at < 0)
+        goto done;
+
+    /* A ceiling this build cannot honour is clamped, and the import stands. */
+    XMEMCPY(blob, version_3, sizeof(blob));
+    blob[at + 2] = (byte)((16000 / 8) >> 8);
```
Frauschi: fixed

tests/api.c
```diff
@@ -10797,10 +10447,9 @@ static int test_wolfSSL_clear_zeroizes_secrets(void)
         /* The pre-master secret is not part of that contract. */
         ExpectNotNull(ssl_s->arrays->preMasterSecret);
```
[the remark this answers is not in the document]
Frauschi: Fixed

Frauschi:
Jenkins retest this please

src/internal.c
```diff
@@ -1554,19 +1563,46 @@ static int ImportOptions(WOLFSSL* ssl, const byte* exp, word32 len, byte ver,
...
+    ato16(exp + idx, &minDhKeySz); idx += OPAQUE16_LEN;
+    ato16(exp + idx, &maxDhKeySz); idx += OPAQUE16_LEN;
+    ato16(exp + idx, &dhKeySz);    idx += OPAQUE16_LEN;
+    /* The blob is not trusted to respect what the pre-master secret buffer
+     * was sized for. */
+    if (maxDhKeySz > MAX_DHKEY_SZ)
+        maxDhKeySz = MAX_DHKEY_SZ;
+    /* Clamping the ceiling must not drag the floor down with it, which would
+     * accept keys the exporting peer's policy refused. */
+    if (minDhKeySz > maxDhKeySz) {
```
Frauschi: Correct, removed. `minDhKeySz` is a policy floor only - every use is a rejection test, it never sizes a buffer - so a floor above the ceiling just refuses more keys, which is the policy the peer exported. Carrying it over verbatim now; the clamp on `maxDhKeySz` and the rejection of `dhKeySz` stay, since those are what bound the pre-master write. Didn't compare the floor against `MAX_DHKEY_SZ` either: `wolfSSL_SetMinDhKey_Sz()` deliberately accepts a floor above the build's math.

Frauschi:
Jenkins retest this please - flaky test

PR #11350, opened 2026-09-02
PKCS#7: decode a message addressed to more than one recipient

Frauschi:
## Description

Four related defects in the PKCS#7 RecipientInfo walk. None involve post-quantum algorithms - the reproduction throughout is two plain RSA recipients. The commit messages carry the detailed rationale.

**AuthEnvelopedData with more than one recipient has never worked.** The decoder stops at the RecipientInfo that matches and, unlike the EnvelopedData decoder, never steps over the ones after it, so the `EncryptedContentInfo` that follows the set is parsed from the wrong offset. `wc_PKCS7_DecodeAuthEnvelopedData()` has behaved this way since `bc94cdc11`, first released in v3.15.5-stable. The equivalent fix landed for EnvelopedData alone in `6987304f4` (v5.8.4-stable), which introduced `PKCS7State.recipientSz` and the step-over; AuthEnvelopedData was not covered by it.

**The recipient search was not bounded by the RecipientInfo set.** After exhausting the recipients the walk read the `EncryptedContentInfo` that follows as though it were another RecipientInfo, so a message carrying no recipient for the reader came back as a parse error rather than as "no recipient found". `wc_PKCS7_ParseToRecipientInfoSet()` already returns the size of the set, so it is passed down as a new `setEnd` argument and the walk stops at the end of the set.

**An indefinite-length RecipientInfo set rewound the index.** `wc_PKCS7_ParseToRecipientInfoSet()` reports a BER set carrying an indefinite length as a size of zero, so stepping the index over the set by that size moves it back to the start of the set instead of past the matched recipient, and the `EncryptedContentInfo` that follows is parsed from the wrong offset. The assignment in the EnvelopedData decoder has been unguarded since `6987304f4`, so that one predates this change and is fixed in its own commit; the AuthEnvelopedData step-over added here carries the same guard from the start.

**The stream accounting was left inconsistent after a rejected RecipientInfo.** A rejected handler stops counting the moment it decides the recipient is not the reader's, but the index has already stepped over the whole structure, so the next handler's bookkeeping compounded the difference and the decode asked for input it had already consumed. The two stream modes need opposite handling, hence the new `wc_PKCS7_StreamResyncRecipient()` helper.

## Behaviour change

`wc_PKCS7_DecodeEnvelopedData()` returns **`PKCS7_RECIP_E` instead of `ASN_PARSE_E`** when no recipient in the set is the reader's. The old value was a symptom of the unbounded search rather than an answer; the one expectation in `tests/api/test_pkcs7.c` that asserted it is updated in the commit that makes it wrong.

`wc_PKCS7_DecryptRecipientInfos()` no longer discards the error a rejected RecipientInfo handler returned. Walking on to the next recipient overwrote it, so a resumed streaming decode - where the caller's `recipFound` starts at zero again while `WC_PKCS7_DECRYPT_KTRI_3` resumes past the issuer compare - could report `PKCS7_RECIP_E` in place of the real failure. The last error is kept and returned when the walk ends with no recipient found. This does not undo the value above: `wc_PKCS7_DecryptKtri()` returns `PKCS7_RECIP_E` for a recipient that is not the reader's, so a set in which none match still answers with it.

`wc_PKCS7_DecryptRecipientInfos()` takes an eighth argument. It is `static` to `pkcs7.c`, so no public or internal API changes; the only other callers are the MC/DC white-box programs under `tests/unit-mcdc/`, which `#include` the source directly and are updated here.

## Testing

New `test_wc_PKCS7_MultipleRecipients` encodes one message to two `KeyTransRecipientInfo` recipients and opens it as each of them in turn, for EnvelopedData and AuthEnvelopedData, whole and fed in 128-byte chunks. The second recipient is the interesting one - decoding as the first passes even with the search unbounded.

`test_wc_PKCS7_IndefiniteRecipientSet` rewrites the RecipientInfo set header of the existing `berContent` vector to an indefinite length and checks that the walk still reaches the recipient. Dropping the guard on the search bound makes it fail with `PKCS7_RECIP_E`.

## Out of scope

The chunked leg is scoped to EnvelopedData and the second recipient. Two gaps in the partial-input path predate this change and are untouched: matching on the *first* recipient in chunks fails with `ASN_PARSE_E` on the merge-base and still does, and AuthEnvelopedData fed in chunks has never opened for a recipient other than the first. A `NOTE` on the EnvelopedData step-over marks the arithmetic that is only right while the whole set is read in one piece; the AuthEnvelopedData step-over refers back to it.

An indefinite-length RecipientInfo set still cannot be decoded through to the plaintext. Nothing consumes the end-of-contents pair that closes it, so `EncryptedContentInfo` is parsed from those two zero bytes and the decode ends at `ASN_PARSE_E` - the behaviour on the merge-base as well. Doing better means walking the remaining RecipientInfos to find the end of the set, which is a change of its own; `test_wc_PKCS7_IndefiniteRecipientSet` asserts the current value exactly, so it fails loudly if that is ever added.

### DIFF
--- a/tests/api/test_pkcs7.c
+++ b/tests/api/test_pkcs7.c
@@ -4443,11 +4443,12 @@ int test_wc_PKCS7_EncodeDecodeEnvelopedData(void)
         pkcs7->singleCert = NULL;
     }
   #ifndef NO_RSA
-    /* With corrupted singleCert, decode should fail with a parse error.
-     * State is properly reset on error so re-decode starts from scratch. */
+    /* With singleCert cleared no KeyTransRecipientInfo can match, so the
+     * decode says so. Was ASN_PARSE_E before the search was bounded by the
+     * RecipientInfo set. State resets, so re-decode starts from scratch. */
     ExpectIntEQ(wc_PKCS7_DecodeEnvelopedData(pkcs7, output,
         (word32)sizeof(output), decoded, (word32)sizeof(decoded)),
-        WC_NO_ERR_TRACE(ASN_PARSE_E));
+        WC_NO_ERR_TRACE(PKCS7_RECIP_E));
   #endif /* !NO_RSA */
     if (pkcs7 != NULL) {
         pkcs7->singleCert = tmpBytePtr;
@@ -6011,6 +6012,93 @@ int test_wc_PKCS7_BER(void)
 #endif
     return EXPECT_RESULT();
 } /* END test_wc_PKCS7_BER() */
+
+/* A BER RecipientInfo SET may carry an indefinite length, which
+ * wc_PKCS7_ParseToRecipientInfoSet reports as a set size of 0. A search bound
+ * computed from that lands on the start of the set and stops the walk before
+ * the first recipient, so such a message decodes to PKCS7_RECIP_E.
+ *
+ * berContent has indefinite outer structures but a definite RecipientInfo SET,
+ * so the vector is rewritten here: 31 82 01 54 becomes 31 80, the content
+ * shifts down two bytes and an end-of-contents pair closes the set. Every
+ * enclosing structure is already indefinite, so no other length moves and the
+ * message stays the same size.
+ *
+ * The decode then unwraps the recipient key and stops at the end-of-contents
+ * pair, which nothing steps over, so EncryptedContentInfo is read from the two
+ * zero bytes: ASN_PARSE_E, as on the merge base. Decoding an indefinite set
+ * through to the plaintext needs that step-over and is not part of this change.
+ * The assertion is exact so that both a walk that stops early (PKCS7_RECIP_E)
+ * and a later step-over fail here rather than pass quietly.
+ *
+ * SP math is excluded: it cannot do the vector's 1024-bit RSA key, and the fake
+ * CEK substituted for that failure makes the outcome non-deterministic. */
+int test_wc_PKCS7_IndefiniteRecipientSet(void)
+{
+    EXPECT_DECLS;
+#if defined(HAVE_PKCS7) && !defined(NO_RSA) && !defined(NO_FILESYSTEM) && \
+    defined(ASN_BER_TO_DER) && !defined(NO_DES3) && !defined(NO_SHA) && \
+    !defined(NO_PKCS7_STREAM) && !defined(WOLFSSL_SP_MATH)
+    wc_PKCS7* pkcs7 = NULL;
+    byte* ber = NULL;
+    byte  decoded[2048];
+    byte  cert[2048];
+    byte  key[2048];
+    word32 certSz = 0;
+    word32 keySz = 0;
+    XFILE f = XBADFILE;
+    int ret = 0;
+    word32 setOff = 20;   /* 31 82 01 54, the RecipientInfo SET header */
+    word32 setLen = 340;  /* 0x0154 */
+
+    ExpectTrue((f = XFOPEN("./certs/1024/client-cert.der", "rb")) != XBADFILE);
+    ExpectTrue((certSz = (word32)XFREAD(cert, 1, sizeof(cert), f)) > 0);
+    if (f != XBADFILE) {
+        XFCLOSE(f);
+        f = XBADFILE;
+    }
+    ExpectTrue((f = XFOPEN("./certs/1024/client-key.der", "rb")) != XBADFILE);
+    ExpectTrue((keySz = (word32)XFREAD(key, 1, sizeof(key), f)) > 0);
+    if (f != XBADFILE) {
+        XFCLOSE(f);
+        f = XBADFILE;
+    }
+
+    /* Confirm the vector still has the header this rewrite expects, so the
+     * test fails loudly rather than silently checking nothing. */
+    ExpectIntEQ(berContent[setOff], 0x31);
+    ExpectIntEQ(berContent[setOff + 1], 0x82);
+    ExpectIntEQ(berContent[setOff + 2], 0x01);
+    ExpectIntEQ(berContent[setOff + 3], 0x54);
+    ExpectIntGE((int)sizeof(berContent), (int)(setOff + setLen + 4));
+
+    ExpectNotNull(ber = (byte*)XMALLOC(sizeof(berContent), NULL,
+        DYNAMIC_TYPE_TMP_BUFFER));
+    if (ber != NULL) {
+        XMEMCPY(ber, berContent, sizeof(berContent));
+        ber[setOff + 1] = 0x80;
+        XMEMMOVE(ber + setOff + 2, ber + setOff + 4, setLen);
+        ber[setOff + 2 + setLen]     = 0x00;
+        ber[setOff + 2 + setLen + 1] = 0x00;
+    }
+
+    ExpectNotNull(pkcs7 = wc_PKCS7_New(HEAP_HINT, testDevId));
+    ExpectIntEQ(wc_PKCS7_InitWithCert(pkcs7, cert, certSz), 0);
+    if (pkcs7 != NULL) {
+        pkcs7->privateKey   = key;
+        pkcs7->privateKeySz = keySz;
+    }
+    if (EXPECT_SUCCESS()) {
+        ret = wc_PKCS7_DecodeEnvelopedData(pkcs7, ber, sizeof(berContent),
+            decoded, sizeof(decoded));
+        ExpectIntEQ(ret, WC_NO_ERR_TRACE(ASN_PARSE_E));
+    }
+
+    wc_PKCS7_Free(pkcs7);
+    XFREE(ber, NULL, DYNAMIC_TYPE_TMP_BUFFER);
+#endif
+    return EXPECT_RESULT();
+}
 
 int test_wc_PKCS7_signed_enveloped(void)
 {
--- a/tests/api/test_pkcs7.c
+++ b/tests/api/test_pkcs7.c
@@ -4699,6 +4699,281 @@ int test_wc_PKCS7_SetAESKeyWrapUnwrapCb(void)
 /*
  * Testing wc_PKCS7_GetEnvelopedDataKariRid().
  */
+/* Open one message as each of its two recipients. Decoding as the second is
+ * what exercises the walk past a RecipientInfo that is not the reader's;
+ * decoding as the first passes even with the search unbounded. */
+int test_wc_PKCS7_MultipleRecipients(void)
+{
+    EXPECT_DECLS;
+#if defined(HAVE_PKCS7) && !defined(NO_RSA) && !defined(NO_AES) && \
+    defined(HAVE_AES_CBC) && defined(WOLFSSL_AES_256) && \
+    !defined(NO_FILESYSTEM) && !defined(NO_SHA256)
+    byte* cert1 = NULL;
+    byte* key1  = NULL;
+    byte* cert2 = NULL;
+    byte* key2  = NULL;
+    byte* cert3 = NULL;
+    byte* key3  = NULL;
+    byte* out   = NULL;
+    byte decoded[128];
+    word32 cert1Sz = 0, key1Sz = 0, cert2Sz = 0, key2Sz = 0;
+    word32 cert3Sz = 0, key3Sz = 0;
+    XFILE f = XBADFILE;
+    int outSz = 4096;
+    int encodedSz = 0;
+    int i, j;
+    WOLFSSL_SMALL_STACK_STATIC const byte content[] = {
+        0x74,0x77,0x6F,0x20,0x6F,0x66,0x20,0x75,0x73   /* "two of us" */
+    };
+
+    cert1Sz = key1Sz = cert2Sz = key2Sz = FOURK_BUF;
+    cert3Sz = key3Sz = FOURK_BUF;
+    ExpectNotNull(cert1 = (byte*)XMALLOC(FOURK_BUF, HEAP_HINT,
+        DYNAMIC_TYPE_TMP_BUFFER));
+    ExpectNotNull(key1 = (byte*)XMALLOC(FOURK_BUF, HEAP_HINT,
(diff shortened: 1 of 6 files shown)


wolfcrypt/src/pkcs7.c
```diff
@@ -14447,10 +14575,16 @@ int wc_PKCS7_DecodeEnvelopedData(wc_PKCS7* pkcs7, byte* in,
             decryptedKey   = pkcs7->stream->aad;
             decryptedKeySz = MAX_ENCRYPTED_KEY_SZ;
             tmpIdx = idx;
+        #endif
+        #ifndef NO_PKCS7_STREAM
+            setEnd = pkcs7->stream->recipientStart +
```
Frauschi: fixed

wolfcrypt/src/pkcs7.c
```diff
@@ -519,8 +520,71 @@ static int wc_PKCS7_StreamEndCase(wc_PKCS7* pkcs7, word32* tmpIdx, word32* idx)
...
+        /* Buffered: the consumed bytes are shifted out rather than counted,
+         * so the next RecipientInfo starts at 0 and every index is rebased. */
+        ret = wc_PKCS7_StreamEndCase(pkcs7, tmpIdx, idx);
+        if (ret == 0) {
+            *idx = 0;
+            *savedIdx = 0;
+            *tmpIdx = 0;
+            /* setEnd is an offset into the same buffer; left un-rebased the
+             * search window grows on every rejected recipient. */
+            if (*setEnd > shifted) {
```
Frauschi, line 555: fixed

tests/unit-mcdc/test_pkcs7_fault_whitebox.c
```diff
@@ -595,41 +595,43 @@ static void wb_ktri_recipinfos(void)
...
         ret = wc_PKCS7_DecryptRecipientInfos(&pkcs7, in, sizeof(in), &idx,
-                decKey, NULL, &recipFound);
+                decKey, NULL, &recipFound, &setEndWb);
         WB_CHECK(ret == WC_NO_ERR_TRACE(BAD_FUNC_ARG),
                 "decryptedKeySz==NULL true");
         idx = 0;
         ret = wc_PKCS7_DecryptRecipientInfos(&pkcs7, in, sizeof(in), &idx,
-                decKey, &decKeySz, NULL);
+                decKey, &decKeySz, NULL, &setEndWb);
         WB_CHECK(ret == WC_NO_ERR_TRACE(BAD_FUNC_ARG), "recipFound==NULL true");
```
Frauschi, line 635: fixed

wolfcrypt/src/pkcs7.c
```diff
@@ -13863,20 +13929,73 @@ static int wc_PKCS7_DecryptRecipientInfos(wc_PKCS7* pkcs7, byte* in,
     }
 
     if (ret < 0) {
-        return ret;
+        /* A resumed decode re-enters here rather than through the walk, so
+         * each type must keep the same policy its arm in the walk uses. */
+        int retry = 0;
+
+        if ((*recipFound == 0) &&
```
Frauschi, line 13942: fixed

wolfcrypt/src/pkcs7.c
```diff
@@ -519,8 +520,71 @@ static int wc_PKCS7_StreamEndCase(wc_PKCS7* pkcs7, word32* tmpIdx, word32* idx)
...
+         * so the next RecipientInfo starts at 0 and every index is rebased. */
+        ret = wc_PKCS7_StreamEndCase(pkcs7, tmpIdx, idx);
+        if (ret == 0) {
+            *idx = 0;
+            *savedIdx = 0;
+            *tmpIdx = 0;
+            /* setEnd is an offset into the same buffer; left un-rebased the
+             * search window grows on every rejected recipient. */
+            if (*setEnd > shifted) {
+                *setEnd -= shifted;
```
Frauschi, line 556: Same finding twice, and not a regression: before this PR the walk had no set bound at all, it ran until `GetSequence_ex()` failed, which is the behavior described here. A partially-rebased bound is strictly better than none.

The gap is real and already flagged in the code: the NOTE on the step-over says the arithmetic "is only right while the whole set is read in one piece". Rebasing `setEnd` across every `wc_PKCS7_StreamEndCase()` means threading it through `wc_PKCS7_DecryptKtri()`, a larger change to shared streaming code, better as its own PR. No memory-safety consequence: when the buffer drains, `pkiMsg` and `pkiMsgSz` are refreshed together, so the parse stays in bounds and fails cleanly.

wolfcrypt/src/pkcs7.c
```diff
@@ -13863,20 +13929,73 @@ static int wc_PKCS7_DecryptRecipientInfos(wc_PKCS7* pkcs7, byte* in,
...
+        pkiMsgSz = inSz;
+    }
 #endif
 
     /* when looking for next recipient, use first sequence and version to
      * indicate there is another, if not, move on */
     while (*recipFound == 0) {
 
+        /* stop at the end of the set, not at the EncryptedContentInfo */
+        if ((*setEnd != 0) && (*idx >= *setEnd)) {
```
Frauschi, line 13976: See reply on the first comment

wolfcrypt/src/pkcs7.c
```diff
@@ -519,8 +520,77 @@ static int wc_PKCS7_StreamEndCase(wc_PKCS7* pkcs7, word32* tmpIdx, word32* idx)
...
+static int wc_PKCS7_StreamResyncRecipient(wc_PKCS7* pkcs7, word32* idx,
+    word32* savedIdx, word32* tmpIdx, word32* setEnd)
+{
+    int ret;
+    word32 shifted = *idx;
+
+    if (*setEnd == 0) {
+        /* No bound to rebase, and 0 must stay 0: the walk reads any other
+         * value as a real end of set. */
+        return wc_PKCS7_StreamEndCase(pkcs7, tmpIdx, idx);
```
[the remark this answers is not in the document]
Frauschi: Fixed

tests/api/test_pkcs7.c
```diff
@@ -5998,6 +6180,85 @@ int test_wc_PKCS7_BER(void)
...
+ * enclosing structure is already indefinite, so no other length moves and the
+ * message stays the same size.
+ *
+ * The assertion is only that the walk reached the recipient. What the DES3 and
+ * 1024-bit RSA decrypt then does is what the berContent test above already
+ * covers, with all its SP-math caveats. */
+int test_wc_PKCS7_IndefiniteRecipientSet(void)
+{
+    EXPECT_DECLS;
+#if defined(HAVE_PKCS7) && !defined(NO_RSA) && !defined(NO_DES3) && \
```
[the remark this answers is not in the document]
Frauschi: Fixed

tests/api/test_pkcs7.c
```diff
@@ -5998,6 +6180,85 @@ int test_wc_PKCS7_BER(void)
...
+    ExpectNotNull(pkcs7 = wc_PKCS7_New(HEAP_HINT, testDevId));
+    ExpectIntEQ(wc_PKCS7_InitWithCert(pkcs7, cert, certSz), 0);
+    if (pkcs7 != NULL) {
+        pkcs7->privateKey   = key;
+        pkcs7->privateKeySz = keySz;
+    }
+    if (EXPECT_SUCCESS()) {
+        ret = wc_PKCS7_DecodeEnvelopedData(pkcs7, ber, sizeof(berContent),
+            decoded, sizeof(decoded));
+        ExpectIntNE(ret, WC_NO_ERR_TRACE(PKCS7_RECIP_E));
```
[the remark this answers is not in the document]
Frauschi: Fixed

Frauschi:
Jenkins retest this please

wolfcrypt/src/pkcs7.c
```diff
@@ -15852,9 +16019,26 @@ int wc_PKCS7_DecodeAuthEnvelopedData(wc_PKCS7* pkcs7, byte* in,
...
+
+                if ((ret = wc_PKCS7_StreamEndCase(pkcs7, &tmpIdx, &idx)) != 0) {
+                    break;
+                }
+            }
+
             tmpIdx = idx;
             pkcs7->stream->expected = MAX_SEQ_SZ;
+        #else
+            idx = recipientSetStart + recipientSetSz;
```
Frauschi: Fixed. The `recipientSetSz > 0` guard is now on the step-over as well.

The identical assignment in the EnvelopedData decoder (`WC_PKCS7_ENV_2`) had the same rewind and predates this PR; it is fixed here too, in its own commit.

wolfcrypt/src/pkcs7.c
```diff
@@ -519,8 +520,77 @@ static int wc_PKCS7_StreamEndCase(wc_PKCS7* pkcs7, word32* tmpIdx, word32* idx)
...
+         * so the next RecipientInfo starts at 0 and every index is rebased. */
+        ret = wc_PKCS7_StreamEndCase(pkcs7, tmpIdx, idx);
+        if (ret == 0) {
+            *idx = 0;
+            *savedIdx = 0;
+            *tmpIdx = 0;
+            /* setEnd is an offset into the same buffer; left un-rebased the
+             * search window grows on every rejected recipient. */
+            if (*setEnd > shifted) {
+                *setEnd -= shifted;
```
Frauschi: Same finding as the round-two comment on this function, and the answer has not changed: before this PR the walk had no set bound at all, it ran until `GetSequence_ex()` failed, so a partially rebased bound is strictly better than none. Rebasing across every `wc_PKCS7_StreamEndCase()` means threading `setEnd` through `wc_PKCS7_DecryptKtri()`, a larger change to shared streaming code that belongs in its own PR. The gap is stated in the NOTE on the step-over.

No memory-safety consequence: when the buffer drains, `pkiMsg` and `pkiMsgSz` are refreshed together, so the parse stays in bounds and fails cleanly.

wolfcrypt/src/pkcs7.c
Frauschi: Fixed. The early return is gone: both mode fixups now always run and only the `setEnd` rebase is conditional on there being a bound to rebase.

Frauschi: Same as the other comment on this line; fixed there.

wolfcrypt/src/pkcs7.c
```diff
@@ -519,8 +520,77 @@ static int wc_PKCS7_StreamEndCase(wc_PKCS7* pkcs7, word32* tmpIdx, word32* idx)
...
+         * is the caller's because the next recipient measures from it. */
+        *tmpIdx = pkcs7->stream->idx;
+        ret = wc_PKCS7_StreamEndCase(pkcs7, tmpIdx, idx);
+    }
+    else {
+        /* Buffered: the consumed bytes are shifted out rather than counted,
+         * so the next RecipientInfo starts at 0 and every index is rebased. */
+        ret = wc_PKCS7_StreamEndCase(pkcs7, tmpIdx, idx);
+        if (ret == 0) {
+            *idx = 0;
```
Frauschi: Fixed. When the shift empties the buffer the walk reads from `in`/`inSz` again, so the indices now rebase onto `pkcs7->stream->idx` rather than 0, and `setEnd` is carried into that space with them.

tests/api/test_pkcs7.c
```diff
@@ -6012,6 +6194,86 @@ int test_wc_PKCS7_BER(void)
...
```
Frauschi: The reasoning is wrong, but the point stands and the test is stronger now.

That `BUFFER_E` is at pkcs7.c:15119, after the recipient walk *and* the content decrypt, so the assertion did discriminate the case it was written for: dropping the `recipientSz > 0` guard makes the test fail with `PKCS7_RECIP_E` (-196), verified.

Asserting the plaintext is not reachable, though. Instrumenting `WC_PKCS7_ENV_3` shows the index sitting on `00 00 30 80` - the end-of-contents pair that closes the indefinite set, which nothing steps over, so `EncryptedContentInfo` is parsed from those two zero bytes and the decode ends at `ASN_PARSE_E`. That is the behaviour on the merge base as well; decoding an indefinite set through to the plaintext needs the step-over and is a separate change.

The test now asserts `ASN_PARSE_E` exactly, so both a walk that stops early and a later step-over fail here rather than pass quietly, with `decoded` at 2048 so the output size is not what stops it, and SP math excluded - it cannot do the vector's 1024-bit RSA key, and the fake CEK substituted for that failure makes the outcome non-deterministic.

wolfcrypt/src/pkcs7.c
```diff
@@ -519,8 +520,80 @@ static int wc_PKCS7_StreamEndCase(wc_PKCS7* pkcs7, word32* tmpIdx, word32* idx)
...
+        ret = wc_PKCS7_StreamEndCase(pkcs7, tmpIdx, idx);
+        if (ret == 0) {
+            *idx      = (pkcs7->stream->length == 0)? pkcs7->stream->idx : 0;
+            *savedIdx = *idx;
+            *tmpIdx   = *idx;
+
+            if (*setEnd != 0) {
+                /* setEnd is an offset in the same space; left un-rebased the
+                 * search window grows on every rejected recipient. */
+                *setEnd = *idx + remain;
```
Frauschi: fixed

wolfcrypt/src/pkcs7.c
```diff
@@ -15793,18 +15955,28 @@ int wc_PKCS7_DecodeAuthEnvelopedData(wc_PKCS7* pkcs7, byte* in,
             if (ret < 0)
                 break;
 
+            /* Decryption stops at the matching RecipientInfo, so remember
+             * the set's start and length to bound the search and step over
+             * the rest. Buffer all of it, as WC_PKCS7_ENV_2 does. */
         #ifndef NO_PKCS7_STREAM
             tmpIdx = idx;
+            pkcs7->stream->recipientSz    = ret;
+            pkcs7->stream->expected       = (word32)ret;
```
Frauschi: fixed

PR #11352, opened 2026-09-02
ML-KEM: add ASN.1 key encoding and X.509 certificate support

Frauschi:
## Description

ML-KEM (FIPS 203) key encoding and X.509 certificate support in wolfCrypt: key OIDs (`2.16.840.1.101.3.4.4.{1,2,3}`), SubjectPublicKeyInfo and PKCS#8 encoders and decoders, `MLKEM_TYPE` public keys in certificates, and issuing one with `wc_MakeCert_ex()`.

This is the key-material half of CNSA 2.0 S/MIME; the CMS side that consumes it (RFC 9629 KEMRecipientInfo) is a separate PR and is not needed to review this one. Nothing here touches TLS or PKCS#7.

Two things beyond the plumbing:

- **All three RFC 9935 Section 6 private key CHOICE forms**, where only the expanded shape was accepted before. For the "both" shape the halves are compared and a mismatch rejected per Section 8. A `WOLFSSL_MLKEM_NO_MAKE_KEY` build cannot run that comparison, so it rejects every key carrying a seed rather than trusting the expanded half.
- **A DER decode can name the parameter set.** `wc_MlKemKey_Init()` accepts the new `WC_ML_KEM_TYPE_UNSET`, so a caller holding a certificate or key file no longer has to read the algorithm OID out first.

Test material is in `certs/mlkem/`, signed by the ML-DSA-87 test CA with `keyUsage` critical asserting `keyEncipherment` alone per RFC 9935 Section 5, regenerated by `certs/renewcerts.sh`.

## Why a flag rather than renumbering

`WC_ML_KEM_512` is `0`, so a zeroed `MlKemKey` is indistinguishable from one holding a valid ML-KEM-512 parameter set. Renumbering is the wrong trade: `WC_ML_KEM_512` and `KYBER512` are public, the enum is structured (`MLKEM_KYBER` = `0x10` is an OR-able variant flag), and the constants are referenced by wolfHSM, wolfProvider and wolfPKCS11 - so moving them silently changes which algorithm an already-compiled application asks for.

`MLKEM_FLAG_TYPE_SET = 0x0010` records the distinction instead. Additive: no public constant moves, no ABI break, no cross-repo coordination.

Worth a look during review: of the four `key->flags = 0` sites, `wc_MlKemKey_Free()` is right to clear everything, but `wc_MlKemKey_MakeKeyWithRandom()` and the `PrivateKeyDecode` scrub path discard key *material*, not the parameter set, so both now use `key->flags &= MLKEM_FLAG_TYPE_SET`.

## Fixes carried along

- **`DecodedCert.pkCurveOID`'s build guard listed only signature algorithms** (`asn.h:1987`) while `GetCertKey()` writes it for FrodoKEM (`asn.c:14133`), so FrodoKEM certificates without ECC/Ed25519/Ed448/ML-DSA/Falcon/SLH-DSA/LMS/XMSS fail to compile on master today. The guard now names both KEMs.
- **`wc_CheckPrivateKey()` freed an ML-KEM key `wc_MlKemKey_Init()` had rejected**, handing a zeroed key to a crypto callback at device id 0. Now gated on a successful init.
- **The crypto callback test faked entropy with a plain counter.** `_InitRng()` health-tests the seed it gets back with the SP 800-90B repetition count and adaptive proportion tests, and once the counter reaches `0x010101xx` three of every four bytes are identical - 384 in a 512-byte window against a cutoff of 325 - so the seed is rejected with `ENTROPY_APT_E`. Whether a build trips it depends only on how many seeds `cryptocb_test` has already drawn: master stops at 11, one short of the edge, and the RNG init in the new `mlkem_asn1_test` makes it 12. That is the `all-amdrand` CI failure, where the entropy scale factor is 512 and each seed is 24592 bytes. The stub now steps an xorshift.
- **The OID sum generator handles guarded entries**, so `oid_sum.h` round-trips instead of losing its hand-added `ACME_IDENTIFIER_OID` block on regeneration.

`WOLFSSL_MLKEM_NO_ASN1` is derived in `settings.h` from `WOLFSSL_ASN_ORIGINAL` and `NO_ASN`, **not** `WOLFSSL_ASN_TEMPLATE` - autotools puts that on the command line but CMake never defines it, so testing it there would switch the feature off in every CMake build.

## Testing

New: `mlkem_asn1_test` (round trips plus parameter-set adoption, negative-controlled), `mlkem_cert_test`, `mlkem_seed_consistency_test` (all three CHOICE forms, wrong-length seed, disagreeing halves), `mlkem_certgen_test`. Seed vectors from RFC 9935 Appendix C/C.4.1; the expansion was confirmed against the public key RFC 9936 Appendix C derives from the same seed.

## Out of scope

`--enable-mlkem=original` (Kyber-only) fails to build, identically to master.

The `NOT_COMPILED_IN` path for adopting a parameter set the build lacks has no test; it needs an embedded DER fixture rather than a generated one.


Frauschi:
Jenkins retest this please

philljj:
Retest this please.

(no logs. known issue, resolved by: https://github.com/wolfSSL/testing/pull/1028)

philljj:
Retest this please.

wolfcrypt/test/test.c
```diff
@@ -57213,6 +57213,837 @@ static wc_test_ret_t mlkem1024_kat(void)
...
+    n->commonNameEnc = CTC_UTF8;
+}
+
+/* Issue one certificate for the given ML-KEM parameter set, then read it back
+ * and confirm what was written. */
+static wc_test_ret_t mlkem_certgen_one(WC_RNG* rng, ecc_key* ca, int level,
+    int expKeyOID, const char* cn)
+{
+    wc_test_ret_t ret = 0;
+    MlKemKey kem[1];
```
Frauschi: fixed

wolfcrypt/test/test.c
```diff
@@ -57244,6 +57244,847 @@ static wc_test_ret_t mlkem1024_kat(void)
...
+        WC_ML_KEM_768,
+#endif
+#if defined(WOLFSSL_WC_ML_KEM_1024) && !defined(WOLFSSL_NO_ML_KEM)
+        WC_ML_KEM_1024,
+#endif
+        -1      /* WC_ML_KEM_512 is 0, so 0 cannot terminate this list */
+    };
+
+    if (levels[0] == -1)
+        return 0;
```
philljj: multi-test clang-tidy yelled at this:

```sh
[clang-tidy-asn-template-sp-all-small-stack] [2 of 4] [f13889c7b8]
    configure with clang-tidy...   real 0m19.069s  user 0m0.003s  sys 0m0.005s
    build...Error while processing /tmp/wolfssl_test_workdir.3364664/wolfssl/wolfcrypt/test/test.c.
/tmp/wolfssl_test_workdir.3364664/wolfssl/wolfcrypt/test/test.c:57305:16: error: 'return' will never be executed [clang-diagnostic-unreachable-code-return]
57305 |         return 0;
|                ^
wolfcrypt/test/test.c:57305:16: error: 'return' will never be executed [-Werror,-Wunreachable-code-return]
 57305 |         return 0;
       |                ^
1 error generated.
```

i guess the define gates are turning this into a dead check

philljj (changes requested):

wolfcrypt/test/test.c
Frauschi: Good catch, fixed. `levels[]` is `static const`, so with any parameter set compiled in clang folds `levels[0] == -1` to `false` and the `return 0` really is dead. Dropped the check instead of working around it: the loop below already terminates on the `-1` sentinel and runs zero times on an empty list, and `out_seed:` frees unconditionally, so a no-parameter-set build behaves the same.

PR #11360, opened 2026-09-02
Fenrir Fix: Added cheap bounds checking to FREESCALE des

aidankeefe2022:
Check that the bottom 3 bits of any input size are 0.

# Description

Bounds are already checked at internal call sights but this is a public api and an easy OOB error to hit. 

Fenrir Fix: https://fenrir.wolfssl.com/finding/12171

IMO test not needed due to simplicity of change.


aidankeefe2022:
retest this please

SparkiDev (approved):

PR #11367, opened 2026-09-03
Add WOLFSSL_CHAIN_VERIFY_CB to replace peer chain verification

julek-wolfssl:
Introduces an opt-in callback, set via `wolfSSL_CTX_SetChainVerifyCb()` or `wolfSSL_SetChainVerifyCb()`, that takes over the peer certificate trust decision entirely. When enabled, wolfSSL builds no chain and performs no signature, date, revocation, key-usage or hostname checks, and the regular verify callback is not invoked. The callback receives only the DER certificates from the Certificate message (as `WOLFSSL_BUFFER_INFO` entries, certs[0] first), and is consulted even under `WOLFSSL_VERIFY_NONE`; rejection always fails the handshake.

Key details:

- Every certificate is still parsed with `ParseCert(NO_VERIFY)` before the callback runs, so malformed DER still fails the handshake without reaching the callback. Content wolfSSL doesn't understand (unknown critical extensions, unsupported key/signature algorithms, algorithm mismatches) is left for the callback to judge, and the check runs only once per certificate.
- The callback may return `CHAIN_VERIFY_WANT_E` to suspend the handshake; it is re-invoked with the same certificates on re-entry into `wolfSSL_connect()`, `wolfSSL_accept()`, `wolfSSL_read()` or `wolfSSL_write()`, including for post-handshake authentication in TLS 1.3.
- Rejection fails with `CHAIN_VERIFY_CB_E`, sends one bad_certificate alert, and sets `peerVerifyRet` so `wolfSSL_get_verify_result()` no longer reports `X509_V_OK`.
- Existing mutual-auth policy for an empty Certificate message is preserved (callback not called), and minimum peer key size checks still apply to the peer's own key.
- Not supported together with DTLS, raw public keys, or OCSP stapling; the setters return `CHAIN_VERIFY_UNSUPPORTED_E` if those are already configured, and a connection attempting to combine them fails with the same error plus an internal_error alert before the callback runs. DTLS code paths are otherwise unchanged.
- Callback and user context can be configured on the context or the SSL object, with the object's setting taking precedence.
- Included in `--enable-all`, with a CMake option and Doxygen documentation.

Also refactors the handshake-suspend machinery shared by `WOLFSSL_NONBLOCK_OCSP` and `WOLFSSL_ASYNC_CRYPT`: replaces open-coded checks against `WC_PENDING_E`/`OCSP_WANT_READ` with a new `IsHsSuspendErr()` helper guarded by `WOLFSSL_HAVE_HS_SUSPEND`. This fixes three related bugs affecting `WOLFSSL_NONBLOCK_OCSP` builds:

- `DoHandShakeMsg()` freeing `ssl->pendingMsg` on a non-`WC_PENDING_E` suspend while `ProcPeerCertArgs` still referenced it.
- `wolfSSL_connect()`/`wolfSSL_accept()` (and their TLS 1.3 counterparts) discarding saved certificate state via `FreeAsyncCtx()` when buffered output flushed.
- `ProcessPeerCerts()` freeing `args` in `WOLFSSL_SMALL_STACK` builds without `WOLFSSL_ASYNC_CRYPT`, where `args` points inside `ssl->async`.


julek-wolfssl:
retest this please

julek-wolfssl:
retest this please

PR #11383, opened 2026-09-04
20260904-configure-all-nonfips-crypto-etc

douzzer:
`configure.ac`: add `--enable-all-nonfips-crypto` and `--enable-all-legacy-crypto`, gathering the non-FIPS and legacy algs formerly enabled piecemeal by `--enable-all-crypto`:

* `all-crypto` now implies both bundles iff !FIPS, so FIPS `all-crypto` builds get no non-FIPS or legacy algs from the bundle (FIPS <v5 builds still get DES3, from the flavor setup rather than all-crypto).

* either bundle can be explicitly enabled in FIPS builds, and explicitly disabling one alongside `all-crypto` allows subtractive selection, e.g. `--enable-all-crypto` `--disable-all-legacy-crypto`.

* `all-nonfips-crypto` initially comprises siphash, blake2b, blake2s, poly1305, chacha, xchacha, curve25519, curve448, pkcs7, nullcipher, aessiv, aeseax, sakke, atomicuser, cryptocb, pkcallbacks, scrypt and argon2 (excluded from kernel-mode builds as before), eccsi, and, gated on !sp-math, ecccustcurves, brainpool, and srp.

* `all-legacy-crypto` initially comprises camellia, ripemd, arc4, md2, md4, md5, dsa (gated on !sp-math), and des3.

* also move EdDSA out from the !`ENABLED_FIPS` gate to the FIPS>=v6 gate (Ed25519/Ed448 are FIPS 186-5 algs, in the v6+ boundary),

* recast the rsapss gate from `FIPS_VERSION`!=v1 to the equivalent !FIPS||FIPS>=v2,

* keep SHA-1 in `all-crypto` with a note anticipating its move to legacy, and

* annotate why anon remains enabled even in FIPS builds (OSP dependents: WPAS, NGINX, HAPROXY, RSYSLOG).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1659,7 +1659,6 @@ AC_ARG_ENABLE([all-crypto],
     )
 if test "$ENABLED_ALL_CRYPT" = "yes"
 then
-    test "$enable_atomicuser" = "" && enable_atomicuser=yes
     test "$enable_aesecb" = "" && enable_aesecb=yes
     test "$enable_aesgcm" = "" && enable_aesgcm=yes
     test "$enable_aesccm" = "" && enable_aesccm=yes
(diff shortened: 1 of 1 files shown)


kaleb-himes (approved):
Good changes, catches most of the problematic algorithms and we can plug the final gaps in a subsequent PR!

(retest this please)..

kaleb-himes:
retest this please

JacobBarthelmeh (approved):

PR #11387, opened 2026-09-06
Add a verbose logging tier and move the highest-volume traces onto it

dgarske:
# Description

tests/unit.test emits 722 MB across 29.0 million lines under DEBUG_WOLFSSL, which makes the output impractical to read and truncates CI job logs before anything useful can be recovered.

WOLFSSL_DEBUG_OPENSSL was already this idea, but only as 23 hand-rolled #ifdef/ENTER/#endif blocks. This promotes it into a real tier: WOLFSSL_MSG_VERBOSE() / WOLFSSL_ENTER_VERBOSE() / WOLFSSL_LEAVE_VERBOSE(), gated on WOLFSSL_VERBOSE_LOGGING — the sibling of the existing WOLFSSL_VERBOSE_ERRORS / WOLFSSL_ERROR_VERBOSE pair (that controls the error queue, this controls the log). WOLFSSL_DEBUG_OPENSSL is honored as an alias, so existing users see no change.

# Testing

Result: 722 MB → 271 MB (62% smaller), in a diff that is net −10 lines, because promoting the idiom removes the boilerplate.

# Checklist

 - [ ] added tests
 - [ ] updated/added doxygen


JacobBarthelmeh (approved):

JacobBarthelmeh:
Retest this please Jenkins

JacobBarthelmeh (approved):

PR #11388, opened 2026-09-06
tests: fix test_client_nofail() ignoring callback_functions CA/cert/key

thiagodoregosousa:
Client half of test_wolfSSL_client_server_nofail()'s driver pair silently ignores callback_functions.caPemFile/certPemFile/keyPemFile and always loads the hardcoded classical defaults, even though its server-side counterpart already honors the equivalent fields for certPemFile/keyPemFile.

First commit adds a regression test (fails on the parent commit, using an ECC server cert the default classical CA can't verify). Second commit adds the fix: caFile/certFile/keyFile locals, overridden when callback_functions supplies them.

Tested locally: built and ran the full tests/unit.test --api suite before and after (0 failures either way), and confirmed the new test fails on the first commit and passes on the second.

### DIFF
--- a/tests/api.c
+++ b/tests/api.c
@@ -7410,6 +7410,9 @@ int test_client_nofail(void* args, cbType cb)
     int  sharedCtx = 0;
     int  doUdp = 0;
     const char* cipherName1, *cipherName2;
+    const char* caFile;
+    const char* certFile;
+    const char* keyFile;
 
     wolfSSL_SetLoggingPrefix("client");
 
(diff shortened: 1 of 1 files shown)


JacobBarthelmeh:
@thiagodoregosousa thanks for opening a pull request! Can you tell us a little more about the use case with wolfSSL? I'm not seeing a contributor agreement on file, please reach out to us through support@wolfssl.com to get that process started.

thiagodoregosousa:
Hi Jacob, thanks for the quick reply.

Use case: I'm a researcher at CESAR working on integrating post-quantum algorithms (ML-KEM and ML-DSA) into TLS/DTLS. Rather than wolfSSL's built-in PQC, I expose a custom PQC library through the crypto callback interface (wc_CryptoCb, via SetDevId), so it runs inside a real handshake. wolfSSL was a natural fit because of its native DTLS 1.3 support, which is the target transport for the project. This fix came up while writing DTLS 1.3 tests that drive the handshake through the callback path, mirroring the existing TLS 1.3 tests.

On the contributor agreement: I'll check with CESAR on how to proceed and get back to you by email, it may take a little time on my side.

kareem-wolfssl:
OK to test - contributor agreement approved

PR #11390, opened 2026-09-07
sniffer: use the wire session id length in both hello parsers

yosuke-wolfssl:
### Problem
`src/sniffer.c` was the only TLS parser in the tree that ignored the wire session-ID length: both hello parsers validated, copied and recorded a hardcoded `ID_LEN` (32). RFC 5246 §7.4.1.2 permits 0–32.

- **Legal short IDs are dropped.** The guard asked whether 32 bytes fit (`ID_LEN > *sslBytes`), so a compact hello — 16-byte ID, one cipher suite, one compression method, 22 bytes remaining — is torn down with a fatal error.
- **Non-session-ID bytes reach the cache key.** For a short ID that is accepted, the 32-byte copy pulls in the cipher suite, compression and extension bytes that follow, and `ProcessFinished()` feeds that value to `AddSession()` as a session cache key.
- **Over-long lengths are silently truncated** to 32 instead of rejected.
- The `sessionIDSz == ID_LEN` tests on the resumption comparison are tautologies, since both sides are forced to `ID_LEN`.

Closes f-13334.

### Fix (`src/sniffer.c`)
`ProcessServerHello()` and `ProcessClientHello()` reject a declared length above `ID_LEN`, copy exactly that many bytes, and record it as `sessionIDSz`. `ProcessClientHello()` gains a read-through check against the declared length, which it never had. This matches `DoServerHello()` and `DoClientHello()` in `src/internal.c` and both TLS 1.3 parsers.

The resumption comparison is **unchanged from master**; its two `== ID_LEN` tests only become meaningful now that `sessionIDSz` holds the wire length. A comment records why the full length is required: the session cache matches only full-length IDs (`src/ssl_sess.c:1173`), as do the client and server resumption decisions (`src/internal.c:34543`, `src/internal.c:41480`).

**Not in scope:** short-ID resumption. wolfSSL treats resumption as 32-only throughout, so supporting shorter IDs means changing the cache and both peers' resumption decisions — a separate change. Short IDs are now parsed and recorded correctly; they simply do not resume.

### Tests (`tests/api.c`)
`test_sniffer_hello_session_id_len` drives `ssl_DecodePacket()` with hand-built IPv4/TCP packets — no pcap needed, as the sniffer verifies no checksums.

| Scenario | Expected |
|---|---|
| ClientHello, length `ID_LEN + 1` | reject |
| ClientHello, length runs past the record | reject |
| ServerHello, length `ID_LEN + 1`, after a zero-ID ClientHello | reject |
| ClientHello + ServerHello, differing 16-byte IDs | accept |
| ClientHello + ServerHello, matching 16-byte IDs | accept, no resumption |
| ClientHello + ServerHello, matching 32-byte IDs | resumption attempted |

### Verification
Every guard was confirmed load-bearing by reverting it individually and checking the test fails — including a control that restores the length-based comparison, pinning the `ID_LEN` requirement. The `bLen > *sslBytes` bound is ASan-only: whenever it fires, the next check rejects the same packet, so the over-read is the sole observable, and the hello packet is allocated at its exact size to expose it.

Clean under `-Werror` with and without TLS 1.3. `make check` 6 passed, 0 failed, 5 skipped. `scripts/sniffer-testsuite.test` passes all 10 pcaps with all 9 resume decrypts intact. ASan + UBSan clean on both the unit test and the pcap suite.

### DIFF
--- a/src/sniffer.c
+++ b/src/sniffer.c
@@ -3947,17 +3947,17 @@ static int ProcessServerHello(int msgSz, const byte* input, int* sslBytes,
         SetError(SERVER_HELLO_INPUT_STR, error, session, FATAL_ERROR_STATE);
         return WOLFSSL_FATAL_ERROR;
     }
+    if (b > ID_LEN) {
+        SetError(SERVER_HELLO_INPUT_STR, error, session, FATAL_ERROR_STATE);
+        return WOLFSSL_FATAL_ERROR;
+    }
     if (b) {
-        if (ID_LEN > *sslBytes) {
-            SetError(SERVER_HELLO_INPUT_STR, error, session, FATAL_ERROR_STATE);
-            return WOLFSSL_FATAL_ERROR;
-        }
     #ifdef WOLFSSL_TLS13
-        XMEMCPY(session->sslServer->session->sessionID, input, ID_LEN);
-        session->sslServer->session->sessionIDSz = ID_LEN;
+        XMEMCPY(session->sslServer->session->sessionID, input, b);
+        session->sslServer->session->sessionIDSz = b;
     #endif
-        XMEMCPY(session->sslServer->arrays->sessionID, input, ID_LEN);
-        session->sslServer->arrays->sessionIDSz = ID_LEN;
+        XMEMCPY(session->sslServer->arrays->sessionID, input, b);
+        session->sslServer->arrays->sessionIDSz = b;
         session->sslServer->options.haveSessionId = 1;
     }
     input     += b;
@@ -4194,6 +4194,8 @@ static int ProcessServerHello(int msgSz, const byte* input, int* sslBytes,
 #endif
 
     if (session->sslServer->options.haveSessionId) {
+        /* The session cache only matches a full length id, so a shorter
+         * one is not treated as a resumption. */
         if (session->sslServer->arrays->sessionIDSz == ID_LEN &&
                 session->sslClient->arrays->sessionIDSz == ID_LEN &&
                 XMEMCMP(session->sslServer->arrays->sessionID,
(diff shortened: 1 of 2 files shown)


src/sniffer.c
```diff
@@ -3916,17 +3916,17 @@ static int ProcessServerHello(int msgSz, const byte* input, int* sslBytes,
         SetError(SERVER_HELLO_INPUT_STR, error, session, FATAL_ERROR_STATE);
         return WOLFSSL_FATAL_ERROR;
     }
+    if (b > ID_LEN) {
+        SetError(SERVER_HELLO_INPUT_STR, error, session, FATAL_ERROR_STATE);
+        return WOLFSSL_FATAL_ERROR;
+    }
     if (b) {
```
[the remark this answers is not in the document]
yosuke-wolfssl: This is pre-existing behavior, not introduced here: the previous code
had the same `if (b) { ... sessionIDSz = ID_LEN; haveSessionId = 1; }`
shape, and this change alters what is copied, not when.

It is also only reachable with a non-conformant peer -- TLS 1.3
requires legacy_session_id_echo to mirror the client's value in both
the HelloRetryRequest and the real ServerHello.

Clearing sessionIDSz alone would leave haveSessionId stale, which is
what drives the ticket-based resumption branch, so a complete fix is a
separate change with its own coverage of the second-hello path.
Tracking it separately rather than widening this one.

tests/api.c
```diff
@@ -39857,6 +39859,215 @@ static int test_sniffer_chain_input_overflow(void)
 }
 #endif /* WOLFSSL_SNIFFER && WOLFSSL_SNIFFER_CHAIN_INPUT */
 
+#if defined(WOLFSSL_SNIFFER) && !defined(WOLFSSL_SNIFFER_WATCH) && \
```
yosuke-wolfssl, line 40125: Applied: !defined(WOLFSSL_NO_TLS12) added to both the test definition
guard and the TEST_DECL guard.

For the record, the impact is not reachable today. Autotools
--disable-tlsv12 does not define WOLFSSL_NO_TLS12 -- the AM_CFLAGS
line for it sits inside the FIPS lean-aesgcm case -- and the test
passes there. Under CMake with -DWOLFSSL_NO_TLS12=yes the library
builds, but unit.test does not compile at all: wolfTLSv1_2_server_method
is undeclared in tests/api/test_tls13.c, plus errors in
tests/api/test_tls13_bounds.c. Neither file is touched by this PR.

The guard is still correct -- SnifferTestHello() builds TLS 1.2
records -- so it is applied as cheap insurance for when those
pre-existing build failures are fixed.

src/sniffer.c
```diff
@@ -3916,17 +3916,17 @@ static int ProcessServerHello(int msgSz, const byte* input, int* sslBytes,
...
     #ifdef WOLFSSL_TLS13
-        XMEMCPY(session->sslServer->session->sessionID, input, ID_LEN);
-        session->sslServer->session->sessionIDSz = ID_LEN;
+        XMEMCPY(session->sslServer->session->sessionID, input, b);
+        session->sslServer->session->sessionIDSz = b;
     #endif
-        XMEMCPY(session->sslServer->arrays->sessionID, input, ID_LEN);
-        session->sslServer->arrays->sessionIDSz = ID_LEN;
+        XMEMCPY(session->sslServer->arrays->sessionID, input, b);
+        session->sslServer->arrays->sessionIDSz = b;
```
yosuke-wolfssl, line 3960: Confirmed. Fixed by gating doResume on a full length id on both sides,
rather than padding the cached id or relaxing the cache.

Resumption in wolfSSL is 32-only throughout, not just in the cache
(client src/internal.c:34543, server src/internal.c:41480, cache
src/ssl_sess.c:1173), so the sniffer now applies the same rule as the
client it models and no longer attempts a lookup that cannot succeed.

The comparison is back to master's text; what changed is its operands,
which now hold the wire length instead of a forced ID_LEN.

Short session id resumption is deliberately not added: it would need
the cache and both peers' decisions changed, and it cannot be verified
here, since neither the wolfSSL client nor server emits a short id.

PR #11395, opened 2026-09-07
Fix X509_NAME_print_ex flag handling and unsafe value copying

julek-wolfssl:
- `XN_FLAG_RFC2253` shared bit 1 with `ASN1_STRFLGS_ESC_2253`, so requesting only RFC 2253 escaping also reversed the DN order. `XN_FLAG_RFC2253` and `XN_FLAG_MULTILINE` are now defined as OpenSSL does, and DN order is only reversed when `XN_FLAG_DN_REV` is set.
- Implement `ASN1_STRFLGS_ESC_2253`, `ASN1_STRFLGS_ESC_CTRL` and `ASN1_STRFLGS_ESC_MSB` escaping to match OpenSSL behavior.
- Write values byte for byte instead of relying on string length, since an embedded NUL previously truncated the value and leaked uninitialized heap bytes into the BIO.
- Reject value lengths where escaping (which can triple the length) would overflow the output buffer.
- Entries are still separated by ", " and printed with their short name; `XN_FLAG_MULTILINE` (previously `0xFFFF`, behaving like `XN_FLAG_RFC2253`) now escapes control and non-ASCII bytes without reversing order.

This fixes OpenVPN master's `ssl_testdriver`, which calls `X509_NAME_print_ex()` with `XN_FLAG_SEP_CPLUS_SPC | XN_FLAG_FN_SN | ASN1_STRFLGS_ESC_2253 | ASN1_STRFLGS_UTF8_CONVERT` (fix for CVE-2026-84790) and previously failed against wolfSSL.

### DIFF
--- a/src/x509.c
+++ b/src/x509.c
@@ -15748,8 +15748,8 @@ int wolfSSL_X509_NAME_print_ex(WOLFSSL_BIO* bio, WOLFSSL_X509_NAME* name,
         valSz = wolfssl_x509_name_esc_value(str->data, str->length, flags,
                                             NULL);
 
-        /* attribute, '=', value, ", " and null terminator */
-        tmpSz = attrSz + eqSz + valSz + 3;
+        /* attribute, '=', value and ", " */
+        tmpSz = attrSz + eqSz + valSz + 2;
         tmp = (char*)XMALLOC((size_t)tmpSz, NULL, DYNAMIC_TYPE_TMP_BUFFER);
         if (tmp == NULL)
             return WOLFSSL_FAILURE;
@@ -15764,13 +15764,6 @@ int wolfSSL_X509_NAME_print_ex(WOLFSSL_BIO* bio, WOLFSSL_X509_NAME* name,
         if (i < count - 1) {
             XMEMCPY(tmp + idx, ", ", 2);
             idx += 2;
-        }
-        else if (bio->type != WOLFSSL_BIO_FILE &&
-                 bio->type != WOLFSSL_BIO_MEMORY) {
-            /* include the terminating null when not writing to a file or
-             * memory BIO */
-            tmp[idx] = '\0';
-            idx++;
         }
 
         if (wolfSSL_BIO_write(bio, tmp, idx) != idx) {
--- a/src/x509.c
+++ b/src/x509.c
@@ -15675,17 +15675,24 @@ static int wolfssl_x509_name_esc_value(const char* in, int inSz,
  *
  * bio    - output BIO to place name string. Does not include null terminator.
  * name   - input name to convert to string
- * indent - number of indent spaces to prepend to name string
- * flags  - flags to control function behavior. Not all flags are currently
- *          supported/implemented. Currently supported are:
- *              XN_FLAG_DN_REV  - print name reversed.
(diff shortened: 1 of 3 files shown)


wolfssl/openssl/x509.h
```diff
@@ -65,7 +64,17 @@
 #define WOLFSSL_XN_FLAG_DUMP_UNKNOWN_FIELDS (1 << 24)
 #define WOLFSSL_XN_FLAG_FN_ALIGN        (1 << 25)
 
-#define WOLFSSL_XN_FLAG_MULTILINE       0xFFFF
+#define WOLFSSL_XN_FLAG_RFC2253 (WOLFSSL_ASN1_STRFLGS_RFC2253 | \
```
Frauschi, line 69: XN_FLAG_RFC2253 now expands to OpenSSL's exact bit pattern, including XN_FLAG_SEP_COMMA_PLUS and XN_FLAG_FN_SN, but wolfSSL_X509_NAME_print_ex never reads the SEP_*/FN_* selectors - it always emits ", " between entries. OpenSSL and RFC 2253 2.2 use ",", so X509_NAME_print_ex(bio, name, 0, XN_FLAG_RFC2253) gives `CN=wolfssl.com, C=US` where OpenSSL gives `CN=wolfssl.com,C=US`. Same gap for SEP_MULTILINE and FN_LN in the new XN_FLAG_MULTILINE: output stays on one line with short names.

That matters for anyone porting from OpenSSL who renders a DN and compares it against a string generated there. Could you either honour XN_FLAG_SEP_MASK, or note the limitation here in x509.h? The caveat is currently only in the doxygen block on wolfSSL_X509_NAME_print_ex, and the header is what a caller reads when picking the flag.

Frauschi (commented):
Looks good overall - the byte-for-byte value copy fixes a real uninitialized-heap disclosure for DN values with an embedded NUL, and the flag bits no longer collide. One comment on the header side of XN_FLAG_RFC2253.

wolfssl/openssl/x509.h
julek-wolfssl: Addressed

julek-wolfssl:
https://github.com/wolfSSL/osp/pull/360 is now necessary to get the grpc tests passing.

src/x509.c
```diff
@@ -15727,67 +15780,69 @@ int wolfSSL_X509_NAME_print_ex(WOLFSSL_BIO* bio, WOLFSSL_X509_NAME* name,
...
-                            str->length, escaped, sizeof(escaped));
-            if (escapeSz < 0)
+        if (fnOpt == WOLFSSL_XN_FLAG_FN_NONE) {
+            /* no attribute name */
+        }
+#ifdef OPENSSL_EXTRA
+        else if (fnOpt == WOLFSSL_XN_FLAG_FN_LN) {
+            attr = wolfSSL_OBJ_nid2ln(ne->nid);
+            if (attr == NULL)
                 return WOLFSSL_FAILURE;
```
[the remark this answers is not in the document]
julek-wolfssl: We are moving towards having all OSP ports using `OPENSSL_EXTRA` at least. This define is enough.

Frauschi (approved):

